PR #335, opened 2020-07-09
Add seasonal variants of strategies to NaiveForecaster

Piyush1729:
Fixes #227 #253 #288 
Added seasonal mean to NaiveForecaster

PR #337, opened 2020-07-12
Add drift strategy to NaiveForecaster

### LINKED ISSUE #227: Extend NaiveForecaster
Extend `NaiveForecaster` to include all common naive forecasting strategies. For an overview, see this [chapter](https://otexts.com/fpp2/simple-methods.html#simple-methods).

- [x] introduce `seasonal` as boolean kwarg, refactor "seasonal_last" and implement "seasonal_mean", so that we can set `seasonal=True` and `strategy="mean"` for example
- [x] add "drift" strategy, the forecasts should be similar to the forecasts from `ReducedRegressionForecaster(regressor=LinearRegression(), ...)`
- [ ] implement more efficient `update_predict` routine

sri1419:
Fixes #227  
Also see #335


sktime/forecasting/tests/test_naive.py
```diff
@@ -101,9 +101,11 @@ def test_strategy_seasonal_mean(fh, sp, window_length):
         np.testing.assert_array_equal(y_pred, expected)
```
Piyush1729, line 102: add another blank line

sktime/forecasting/tests/test_naive.py
```diff
@@ -101,9 +101,11 @@ def test_strategy_seasonal_mean(fh, sp, window_length):
...
 
 @pytest.mark.parametrize("fh", TEST_OOS_FHS)
-@pytest.mark.parametrize("window_length", TEST_WINDOW_LENGTHS)
+@pytest.mark.parametrize("window_length",
+                         [*TEST_WINDOW_LENGTHS,None])
```
Piyush1729, line 105: add space after comma

sktime/forecasting/tests/test_naive.py
```diff
@@ -101,9 +101,11 @@ def test_strategy_seasonal_mean(fh, sp, window_length):
...
 def test_strategy_drift(fh, window_length):
-    f = NaiveForecaster(strategy="drift", window_length=window_length)
+    f = NaiveForecaster(strategy="drift",
+                        window_length=window_length)
     f.fit(y_train)
     y_pred = f.predict(fh)
```
Piyush1729, line 111: add blank line at the end of file,
remove redundant backslash at line 121

sktime/forecasting/naive.py
```diff
@@ -183,3 +191,16 @@ def _predict_last_window(self, fh, X=None, return_pred_int=False,
...
             return last_window[fh_idx]
+
+        elif self.strategy == "drift":
+            if any(last_window) is not np.nan:
+                drift = np.mean(np.diff(last_window))
+
+                # get zero-based index by subtracting the minimum
+                fh_idx = fh.index_like(self.cutoff)
+
+                last_window =np.arange(last_window[-1],
```
Piyush1729, line 202: add space after operator

sktime/forecasting/naive.py
```diff
@@ -183,3 +191,16 @@ def _predict_last_window(self, fh, X=None, return_pred_int=False,
...
+                                       last_window[-1]+\
```
Piyush1729, line 203: add space after operator and remove redundant backslash

sktime/forecasting/naive.py
```diff
@@ -139,5 +154,53 @@ def _predict_last_window(self, fh, X=None, return_pred_int=False,
...
+                last_window =\
+                    np.tile(last_window,
+                            reps=reps)
+
+            # get zero-based index by subtracting the minimum
+            fh_idx = fh.index_like(self.cutoff)
+            return last_window[fh_idx]
+
+        elif self.strategy == "drift":
+            if any(last_window) is not np.nan:
```
mloning, line 196: what happens if this condition fails? perhaps we could just use `np.nanmean` to handle missing values? Or raise an error?

mloning:
Dear @sri1419, we've now merged @Piyush1729's PR, so you should be able to implement the drift strategy now, but you have to sync your branch and resolve the merge conflicts.

sri1419:
Hi @mloning I had implemented drift strategy

sktime/forecasting/naive.py
```diff
@@ -176,3 +193,20 @@ def _predict_last_window(self, fh, X=None, return_pred_int=False,
                 # get zero-based index by subtracting the minimum
                 fh_idx = fh.index_like(self.cutoff)
                 return y_pred[fh_idx]
+
+        elif self.strategy == "drift":
+            if self.window_length_ != 1:
+                # if NaNs exists they are counted as zero and jump
+                # in linear difference is splitted among
+                # the missing values
+                drift = np.mean(np.diff(np.nan_to_num(last_window)))
```
mloning, line 202: I don't understand why we can just replace missing values with zeros? We need to compute the difference between the last value and the first value, no? Check out the formula here: https://otexts.com/fpp3/simple-methods.html

I think we want something like this: 
```python
slope = (last_window[last] - last_window[first]) / (last - first)  # formula for slope
fh_idx = fh.index_like(self.cutoff)  # forecasting horizon
y_pred = last_window[last] + fh_idx * slope  # linear extrapolation
return y_pred 
```
This seems to be all we need for the case with no missing values. If we have missing values, and if the first or last value is a missing value, we need to take the first or last non-missing value. So we could do something like this:
```python
idx = np.argwhere(~np.isnan(last_window))  # get the index of non-missing values
first = idx[0]
last = idx[-1]
```
In the case where the last value is missing we still need to do one more thing: we need to estimate the missing value(s) given the last non-missing value and slope: 
```
# last value is non-missing
if last == (len(last_window) - 1):
    last_value = last_window[last]
# last value is missing
else:
    last_value = last_window[last] + slope * len(last_window) - last - 1

y_pred = last_value + fh_idx * slope
```

mloning: This does seem to be a lot of code for a simple method, there may be a better way to implement it. We could also just raise an Error if there are missing values.

sktime/forecasting/tests/test_naive.py
```diff
@@ -109,3 +109,29 @@ def test_strategy_mean_seasonal_simple(n_seasons, sp):
...
+    if window_length != 1:
+        f = NaiveForecaster(strategy="drift",
+                            window_length=window_length)
+        f.fit(y_train)
+        y_pred = f.predict(fh)
+
+        if window_length is None:
+            window_length = len(y_train)
+
+        window_length = y_train.iloc[-window_length:]
```
mloning, line 127: It's easy to write unit tests for the drift method without using the same code that we use inside the drift method. For example, we know the solution for a number of easy cases and can check if the forecaster returns the expected solution: 

1) a flat time series with zero slope, in this case strategy="drift" and "last" should be the same
2) a time series with constant slope 1, e.g. `np.arange(n_timepoints)` or `-np.arange(n_timepoints)`

If you want to support missing values too, we should add some tests for that as well.

sktime/forecasting/naive.py
```diff
@@ -176,3 +193,22 @@ def _predict_last_window(self, fh, X=None, return_pred_int=False,
...
+        else:
+            if self.window_length_ != 1:
+                if np.any(np.isnan(last_window[0], last_window[-1])):
+                    raise ValueError(f"For {self.strategy},"
+                                     f"first and last elements in the last "
+                                     f"window must not be missing a value.")
+                else:
+                    # formula for slope
+                    slope = (last_window[-1] -
+                             last_window[0]) / (len(last_window) - 1)
```
mloning, line 207:
```suggestion
                             last_window[0]) / (self.window_length_ - 1)
```

sktime/forecasting/naive.py
```diff
@@ -176,3 +193,22 @@ def _predict_last_window(self, fh, X=None, return_pred_int=False,
                 # get zero-based index by subtracting the minimum
                 fh_idx = fh.index_like(self.cutoff)
                 return y_pred[fh_idx]
+
+        # if self.strategy == "drift":
+        else:
+            if self.window_length_ != 1:
+                if np.any(np.isnan(last_window[0], last_window[-1])):
```
mloning, line 200: this may work
```suggestion
                if np.any(np.isnan(last_window[[0, -1]])):
```

sktime/forecasting/tests/test_naive.py
```diff
@@ -109,3 +109,48 @@ def test_strategy_mean_seasonal_simple(n_seasons, sp):
...
+                            window_length=window_length)
+        f.fit(y_train)
+        y_pred = f.predict(fh)
+
+        if window_length is None:
+            window_length = len(y_train)
+
+        g = NaiveForecaster(strategy="last")
+        g.fit(y_train)
+        expected = g.predict(fh)
```
mloning, line 154: we don't have to use `NaiveForecaster(strategy="last") ` to find out what's expected, no? We can just use `np.ones(len(fh))`, no?

mloning: adding the other simple case of a line will also be useful: `y_train = pd.Series(np.arange(20))`

sktime/forecasting/tests/test_naive.py
```diff
@@ -109,3 +109,48 @@ def test_strategy_mean_seasonal_simple(n_seasons, sp):
...
+        np.testing.assert_array_equal(y_pred, expected)
+
+
+@pytest.mark.parametrize("fh", TEST_OOS_FHS)
+@pytest.mark.parametrize("window_length",
+                         [*TEST_WINDOW_LENGTHS, None])
+def test_strategy_drift_simple(fh, window_length):
+    # test for flat time series data
+    if window_length != 1:
+        y_train = np.ones((1, 20))
```
mloning, line 143:
```suggestion
        y_train = pd.Series(np.ones(20))
```

sktime/forecasting/tests/test_naive.py
sri1419: > adding the other simple case of a line will also be useful: `y_train = pd.Series(np.arange(20))`

It is already implemented in the test_strategy_drift()

mloning: I see, could you still make it more explicit, in the function name, also we may change the data at the beginning of the script but this function should still work when we change the data, we also don't have to compute the slope, right? We can just use `np.arange(start, end)` where start is the last point in the training series, no?

sri1419: I think start is the successor to last point when we use check_fh

sktime/forecasting/naive.py
```diff
@@ -37,6 +37,9 @@ class NaiveForecaster(OptionalForecastingHorizonMixin,
                      When sp is not 1, mean of all values
                      in a season from last window will be
                      forecasted for each season.
+        * "drift": forecast the values increasing or
```
mloning, line 40: I'd say: "forecast by fitting a line between the first and last point of the window and extrapolating it into the future"

sktime/forecasting/naive.py
```diff
@@ -176,3 +193,22 @@ def _predict_last_window(self, fh, X=None, return_pred_int=False,
...
                 fh_idx = fh.index_like(self.cutoff)
                 return y_pred[fh_idx]
+
+        # if self.strategy == "drift":
+        else:
+            if self.window_length_ != 1:
+                if np.any(np.isnan(last_window[[0, -1]])):
+                    raise ValueError(f"For {self.strategy},"
+                                     f"first and last elements in the last "
+                                     f"window must not be missing a value.")
```
mloning, line 203: "must not be a missing value"

sktime/forecasting/tests/test_naive.py
```diff
@@ -109,3 +109,46 @@ def test_strategy_mean_seasonal_simple(n_seasons, sp):
...
+        fh = check_fh(fh)
+
+        expected = y_train.iloc[-1] + np.arange(0, max(fh) + 1)[fh]
+        np.testing.assert_array_equal(y_pred, expected)
+
+
+@pytest.mark.parametrize("fh", TEST_OOS_FHS)
+@pytest.mark.parametrize("window_length",
+                         [*TEST_WINDOW_LENGTHS, None])
+def test_strategy_drift_simple(fh, window_length):
```
mloning, line 138:
```suggestion
def test_strategy_drift_flat_line(fh, window_length):
```

sri1419:
For the unit test, using random numbers for window is ok?

sri1419:
@mloning that's all for now

mloning:
@allcontributors please add @sri1419 for code

fkiraly:
... this bot doesn't seem to be very good at understanding English, @mloning ?

mloning:
Yes, has been reported already here: https://github.com/all-contributors/all-contributors/issues/362

PR #378, opened 2020-08-18
Fix ci

mloning:
Most recent Numba version fails on manylinux 64x. No wheel seems to be available. Fix Numba version to `0.50.*` for now. 

For easier debugging in the future, we could use `pip install --platform $PLATFORM` to raise error if no wheel for platform is found.


mloning:
Closed in favour of #379

Issue #353, opened 2020-07-20
Data set generators 


mloning:
As discussed in #320, it would be great to have functions that generate/simulate data sets, they are useful for testing and explaining the difference between estimators. 

* We currently have a very simple data set generator here: [`make_classification_problem`](https://github.com/alan-turing-institute/sktime/blob/67c56be8b1e838f2628df829946f795b7dba9aed/sktime/utils/_testing/series_as_features.py#L33)
* TSML has some here in Java: https://github.com/uea-machine-learning/tsml/tree/master/src/main/java/statistics/simulators

Kerybas:
I am interested in this issue.  
I would like to propose the following approach:  
1. Generate a bunch of randomised sine waves (10 here):
![image](https://user-images.githubusercontent.com/42703200/89106120-92cddf00-d427-11ea-8c07-e54d6df290c4.png)
2. Pass them through a sklearn MLPRegressor randomly fitted, to obtain a number of "observations":
![image](https://user-images.githubusercontent.com/42703200/89106156-d1fc3000-d427-11ea-9a19-760ce8e7413f.png)
3. Optionally, discretize some of the signals in 2. to obtains classes (here, the red signal is discretized in 3 bins via KBinsDiscretizer):
![image](https://user-images.githubusercontent.com/42703200/89106248-8d24c900-d428-11ea-99d5-5a442ed4cce0.png)  

Step 2. should ensure that the data set is complex while keeping a realistic relationship between the signals.  

We could also add some noise, under-sampling, etc. to step 2 to make it even more realistic.

What do you think?  
Is it worth investigating that approach?

mloning:
@ABostrom @TonyBagnall any thoughts on this?

ABostrom:
There's an arxiv paper did ages ago here: https://arxiv.org/abs/1703.09480
 
An old SimulationExperiment class from tsml: https://github.com/uea-machine-learning/tsml/blob/master/src/main/java/experiments/SimulationExperiments.java

I think our angle for simulation was, what type of problem are you looking to simulate within TSC.

Can you design a simulated problem that an interval classifier should be best on, same for DTW, dictionary, shapelet etc.

fkiraly:
Hm, @TonyBagnall, @Kerybas, I thought it would be useful to get a quick way to generate synthetic test datasets for the different *tasks*, e.g., time series classification, regression, forecasting, that sktime supports?

mloning:
[...]

Issue #361, opened 2020-08-03
[DOC] User guide


mloning:
#### Purpose of a user guide
In addition to our API reference, we want to create a user guide. The API reference serves as a technical software reference. The user guide should serve as a scientific reference. In the user guide, we want to give an overview of sktime, including the 
* taxonomy of learning tasks, 
* how they are related and the concept of reduction, 
* sktime's package structure, 
* an overview of task-specific submodules (classification, forecasting, etc),
* guidance/explanations of algorithms 

#### Table of contents
1. Introduction
2. Taxonomy of learning tasks & reduction 
3. Classification 
5. Forecasting

#### Implementation
* the user guide should live in the documentation on our [website](https://sktime.org) on a separate page
* we need to add a new `user_guide.rst` file to [/docs/source](https://github.com/alan-turing-institute/sktime/tree/master/docs/source), and link and collect chapters in `/docs/source/user_guide/`.

mloning:
Hi @fkiraly and @lynnssi, I've create a PR with a framework for the user guide #377 and added our preliminary table of contents to the wiki (see link above).

mloning:
Another idea: We could also invite people to upload their blogposts in form of Jupyter notebooks to a blog post repo, something along the lines of https://github.com/fastai/fastpages for sktime.

evanmiller29:
I'm happy to contribute a notebook or two if that would help?

I'm only familiar with forecasting but I'd be happy to do one for TS Classification too if anyone can point me in a good real world example.

mloning:
Hi @evanmiller29, if you want to contribute to the docs directly, best to work off and merge into #377. The preliminary table of contents above should give a basic scaffolding for the sections. For the notebooks, best to look at the examples and improve them.

mloning:
Hi @lynnssi and @fkiraly, I've now merged the basic structure of the user guide (#377) with a note that it's under development and that we're looking for contributions. 

@lynnssi I've added a page for extra resources - perhaps it's best to start the curated list of blog posts there? Feel free to edit/move around pages as you see fit.

Lovkush-A:
@mloning. Hello Marcus and Merry Christmas! As a christmas present to you and the team, I would like to contribute significantly to this package, and in particular, to the user guide. Here are some questions I have and other relevant points:

[...]

PR #384, opened 2020-08-25
Initial work on generators for #353 Data set simulators/generators

sjmiller8182:
#### Reference Issues/PRs

Initial work on generators based on mention in #353 with @mloning.

This introduces univariate generators for random normal noise, ARMA process, and linear process with noise.

No new dependencies.

#### PR checklist for new estimators 

Not a new estimator.


mloning:
Thanks for the PR! 

Quick comment: Linting failed, you need to include `__all__ = [...]` in the the `__init__.py` file, otherwise flake8 will complain about unused imports.

mloning:
Thanks again @sjmiller8182 for the PR! Looks all good to me. 

@fkiraly any comments?

sjmiller8182:
@mloning Thanks for taking a look!

I noticed a major copy and paste error in the docs - I'll fix that this weekend. 

Also, I want to push a few more items. I've come up with a way to change the RandomState sampling method and pass the kwargs to the sampling method. I'd also like to implement something for seasonal data generation. I'm hoping to have something to add this weekend on these two items.

I am not sure if the current code needs some kind of unit tests, since its rather simple. But, I can put something in if that is recommended.

mloning:
Unit tests would be nice to have, but since we're using statsmodels under the hood not we don't have to test the sampling itself I think. It would be nice to integrate the generators in some tutorial or user guide (see #361)

fkiraly:
>@fkiraly any comments?

* in terms of module structure, I would separate data sets from data generators - these are different things in my opinion, the former are fixed data sets, the latter are stochastic samplers of data sets. `sktime.datagenerators`?
* this introduces a dependency on `typing` and `abc` which we've tried to avoid so far. May be worth a discussion going forward, i.e., whether to integrate these in the general package design, but if not we may want to remove these as deps.
* I don't think 100 is a sensible default for `n_sample`, since arbitrary, should we rather not have a default?
* default for `burnin` in ARMA is 0, which is a "bad" default - typically, you would have a longer burn-in to ensure you are sampling from something closer to the stationary process
* the `scale` parameter should be supported, since this is another typical parameter for the mentioned processes
* `statsmodels` also allows generation of multiple series, through the `nsample` param which can be tuple valued. This might be useful for the time series classification context, so I wonder whether we also want to support this.

mloning:
Thanks @fkiraly!

* `typing` and `abc` come with the standard Python distribution and don't require us to install other packages, I also tend to view this as maintainer preference now and wouldn't mind having different conventions across modules, but happy to discuss this more generally
* the generators are implemented in their own file, we simply expose them to user through `sktime.datasets`, I think that's fine, the function/class names distinguish clearly whether it's *loading* or *generating* data, that way we'd follow sklearn too.

sjmiller8182:
* Seems fine to me to drop `typing` since it was added in python 3.5. Is there an official list of supported python versions? `abc` is supported by python 3.
* I will add support for the kwargs of `np.random.normal` (explicitly mentioned method) into `NoiseGenerator`. However, would it make sense to create a more general version of the noise generator that can generate from any of the methods of `np.RandomState` with a `**kwargs`-like interface for setting the arguments?
* For `burnin`, should a user input be required or is there a recommendation for a default?
* I will extend the generators to accept tuples for `n_sample` and fine by me to remove the default value.

mloning:
Great! FYI We support Python 3.6-3.8, so typing and abc are fine.

fkiraly:
>Great! FYI We support Python 3.6-3.8, so typing and abc are fine.

Fine with me to use `typing`, given @mloning´s comment.
I´d like to have consistent use of `typing` and/or types at some point, but there´s no need to be that strict right now.

>However, would it make sense to create a more general version of the noise generator that can generate from any of the methods of np.RandomState with a **kwargs-like interface for setting the arguments?

Hm, that actually makes sense - as long as it does not lead to interface overkill (or unclear defaults).

>For burnin, should a user input be required or is there a recommendation for a default?

hm, `statsmodels` also has 0 as default for `burnin`. 
We could leave it at that - any other choice might require justification anad not generalize well to all use cases.

fkiraly (changes requested):
Given discussion above, I'll wait for the one change that is outstanding (`n_sample` support), please ask for a review then.
The other changes (removal/handling of `typing`, `burnin` default change) are not necessary for my approval since I'm happy with current state after discussion.

fkiraly:
@sjmiller8182, we are discussing general design of data set generators in #759 - are you still working on this?

There was just a minor change outstanding here.

sjmiller8182:
@fkiraly, Sorry for the delay on this. If this still fits in the general scope of data generators, I will work on getting it finished.

fkiraly:
no worries, of course!

You may also want to chip in to the design discussion and/or ensure API consistency on your PR.

sjmiller8182:
@fkiraly, @mloning,

I've implemented allowing a tuple input to `n_sample` for `NoiseGenerator` and `ArmaGenerator`, but I am not sure of a good way to implement the change for `LinearGenerator`. I may just not be looking at it correctly. I also removed the default value for `n_sample` since we had discussed previously that it doesn't make sense. 

Right now I am seeing a failure on the macos-tests, but it doesn't seem to be related to my additions so I am not sure what is causing the problem on those builds at the moment. I would appreciate it if someone could help me understand what is failing on those builds.

I also still need to look at how the unit tests are implemented here. I am not familiar with the testing structure in this project so it may take me some time to implement the tests. However, the doc tests are passing.

fkiraly:
Can you explain what your speicifc question about `LinearGenerator` is?

Questions from my side (not change requests):
* should the `sample` method not have two arguments, `n_sample` and `n_instance`? One for the length of the time series, one for how many there are.
* should the base class inherit from `sklearn's` `BaseEstimator`? That way, you get parameter handling for free.

Also, I strongly suggest adhering to a joint interface specificationm i.e., type and arguments of `sample` should be the same across the generators as long as it generates the same data object(s). For example, `burnin` in `sample` of `ARMAGenerator` should move to `__init__`; the types of all `sample` methods should be the same, why is this different between `ARMAGenereator` and `LinearGenerator`?

sjmiller8182:
Ah yes, agreed the `sample` method should be the same between the classes. 

Actually, if we change `sample` to take two arguments, `n_sample` and `n_instance` that eliminates my issue with the `LinearGenerator`. I'll take a in cooperating `BaseEstimator` from `sklearn`.

sktime/datasets/generators.py
```diff
@@ -12,21 +12,37 @@
...
+    -----
+    All generator parameters should be specified as keyword arguments in
+    their `__init__`.
+    All generators should implement the unified interface given below.
+
     """
 
     @abstractmethod
-    def sample(self, n_sample):
+    def sample(self, n_sample, n_instance):
```
mloning:
```suggestion
    def sample(self, n_instances=None, n_variables=None, n_timepoints=None):
```
with some useful default values

Alternatively we could distinguish between generators for different kinds of data (time series and panel data)?

mloning (commented):
Thanks @sjmiller8182 - we also need to write some unit tests to make sure they work as expected.

sjmiller8182:
@mloning, 

I've added tests. 

For my understanding, by time series you mean a single instance of a univariate or multivariate time series and by panel data you mean multiple instances of a time series? If that is the case, I think we could add a method `sample_panel` to generate the panels.

TonyBagnall:
hi, @sjmiller8182 we are having a tidy up of pull requests. Are you still working on this? If not, is it ok to close it? We would welcome a new pull request if you update it later

sjmiller8182:
@TonyBagnall, I have implemented the requests mentioned above. I believe it is ready to be integrated.

TonyBagnall:
hi @sjmiller8182 sorry for the delay, if you want it to go in could you change the status so it is no longer a draft, we will then review it

PR #376, opened 2020-08-13
Multivariate regression

SebasKoel:
Draft PR

Started working on adding multivariate regression support
Not finished yet

PR #385, opened 2020-08-25
added multivariate dataset  with us consumption over time

SebasKoel:
Reference Issues/PRs
#376, #380, #382 

added multivariate dataset so we can add support for multivariate/exogenous data of #382


sktime/datasets/base.py
```diff
@@ -614,3 +615,56 @@ def load_airline():
...
+
+    # Sort by Quarter then set simple numeric index
+    # TODO add support for period/datetime indexing
+    # data.index = pd.PeriodIndex(data.index, freq='Y')
+    data = data.sort_values('Quarter')
+    data = data.reset_index(drop=True)
+    data.index = pd.Int64Index(data.index)
+    data.name = name
+    y = data[y_col]
+    if y_col != 'Quarter':
```
mloning, line 667: Quarter is the time index, no? But predicting the time index doesn't make much sense. So it'd set Quarter as the index of the data, but given that we don't support period/datetime indices at the moment, I'd just drop it.

sktime/datasets/base.py
```diff
@@ -614,3 +615,56 @@ def load_airline():
...
+        data = data.drop('Quarter', axis=1)        
+    X = data.drop(y_col, axis=1)
+    return X, y
```
mloning, line 670: linting (see flake8) probably failed because there's an no empty line at the end

sktime/datasets/base.py
```diff
@@ -614,3 +615,56 @@ def load_airline():
     data.index = pd.Int64Index(data.index)
     data.name = name
     return data
+
+def load_uschange(y_col='Consumption'):
```
mloning, line 619: I'd call it `y_name`, please add it to the doc string 
```suggestion
def load_uschange(y_name="Consumption"):
```

mloning (commented):
Thanks for the PR! Looks good in general, left three minor comments. Happy to merge afterwards!

sktime/datasets/base.py
SebasKoel: Yes Quarter is just a string. It's just being sorted on it as a double check (the csv is already sorted). It's being dropped. But it's also possible to get it back as y column, indeed it doesn't really make sense but else it should return an error when 'Quarter' is chosen which I don't see as better then having the option of having it returned as y

mloning (commented):
There are a few pkl files on the PR now which we probably don't need, right?

sktime/datasets/tests/test_uschange.py
```diff
@@ -0,0 +1,19 @@
...
+__all__ = []
+
+import pandas as pd
+from sktime.datasets import load_uschange
+
+
+def test_uschange():
+    X, y = load_uschange()
+
+    X_test = pd.read_pickle('uschange_X.pkl')
```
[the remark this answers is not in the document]
SebasKoel: The pickle files are the X and y in pickle files and then it just compares these with the dataframe/serie returned by the function. Now the entire dataset is in it but for larger datasets I would've done i.e. the first 20 rows. I discussed this with Omar and he preferred not to have binary files, but just assert the column names, shapes and dtypes.  I think it's better this way because it checks the actual data as well, if someone accidentally changed the csv. But I don't mind changing it to what Omar suggested. Do you want me to change it?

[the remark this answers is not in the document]
mloning: Ah okay, I think it's enough to have checks for what @big-o suggested, I don't want to have two copies of the data, the data itself is not that important I think, it's just toy data. Also someone may accidentally change the copy of the data, right? So a better test would be to download and compare the dataset from a reference source but I think this is not necessary. If you want to write unit tests for the loading function, it would be very useful to generalise them to test the other loading functions too (only the output object types, data types).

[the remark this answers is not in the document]
SebasKoel: Okay just removed the pickle files and only use assertions. Also tried to make a general dataset testing function which can be used for other datasets as well.

sktime/datasets/tests/test_uschange.py
```diff
@@ -0,0 +1,55 @@
...
+# copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+
+__author__ = ["Sebastiaan Koel"]
+__all__ = []
+
+import pandas as pd
+from sktime.datasets import load_uschange
+
+
+def check_dataset(checks, y=None, X=None):
```
mloning, line 12: This look good! The only thing I'd change is to replace `len` with `shape`.

sktime/datasets/tests/test_uschange.py
```diff
@@ -0,0 +1,55 @@
...
+        'columns' : list with column names in correct order,
+        'len_y'   : lenght of the y series (int),
+        'len_X'   : lenght of the X series/dataframe (int),
+        'data_types_X' : dictionary with column name keys and dtype as value,
+        'data_type_y'  : dtype if y column (string)
+    y:  y series of the dataset
+    X:  X dataframe or series from dataset
+    """
+    if y is not None:
+        assert(isinstance(y, pd.Series))
```
mloning, line 26: I've only seen `assert` being used as a statement rather than function, please replace here and throughout, or is this a new thing? 
```suggestion
        assert isinstance(y, pd.Series)
```

sktime/datasets/tests/test_uschange.py
```diff
@@ -0,0 +1,55 @@
...
+
+        assert(X.columns.values.tolist() == checks['columns'])
+
+        for col, dt in checks['data_types_X'].items():
+            assert(X[col].dtype == dt)
+
+        assert(len(X) == checks['len_X'])
+
+
+def test_uschange():
```
mloning, line 44: Could you run this on `load_airline` too? You could use 
```python
@pytest.mark.parametrize("data_loader", [load_airline, load_uschange, ...])
def test_data_loaders(data_loader):
    ...
```
This will iterate over `data_loader` and run a separate test for each one.

mloning: Of course we would have to iterate over the checks too ... perhaps we could add the `checks` dictionary to the docstring and retrieve it via `data_loader.__doc__`, just a thought

mloning (commented):
Looks great! Left some minor comments and there's still this file: sktime/datasets/data/Uschange/Uschange.pkl

sktime/datasets/tests/test_uschange.py
big-o: Maybe it would be cleaner to just declare the checks dictionary at the top of the module and then you would just be parametrising over the dictionary keys? e.g.:

```python
_CHECKS = {
    "uschange": {
        'loader': load_uschange,
        'len_y': 187,
        ...
    }
}

@pytest.mark.parametrize("dataset", [load_airline, load_uschange, ...])
def test_data_loaders(dataset):
    data_loader = _CHECKS[dataset]["loader"]
    ...
```

mloning (approved):
Looks good to me! @big-o I'll wait for your approval before we merge.

big-o:
As mentioned in the above comment is be happy to approve now; the above change is very minor and could either be done in this PR or at a later date, if you think it's needed.

mloning:
Hi @SebasKoel, could you confirm if you want to implement @big-o's suggestion? I wouldn't mind merging it now.

mloning (changes requested):
Hi @SebasKoel, 

There's still the `Uschange.pkl` file in the PR and the changes to the `sktime/forecasting/compose/_reduce.py` which we wanted to have in a separate PR. 

I believe you should be able to remove them as follows:
* `git rm sktime/datasets/data/Uschange/Uschange.pkl`
* `git checkout master -- sktime/forecasting/compose/_reduce.py`
See if that works!

mloning (approved):

mloning:
Thanks @SebasKoel! 🎉

big-o (commented):
All looks good to me - as mentioned before I don't think the comment about the unit test really matters here, but maybe it should be considered when more unit tests are added in the future.

PR #387, opened 2020-08-27
Move prediction interval API to base classes

big-o:

Most of the code in `ThetaForecaster` for computing prediction intervals is applicable to all forecasters. I've moved that code out into the base classes (`_base.py` and `_sktime.py`), leaving only the parts that are specific to the Theta algorithm in `ThetaForecaster`.

An added benefit is that all forecasters will now have a `compute_pred_int` method, allowing for a common API. If the method hasn't been implemented for a forecaster, a `NotImplementedError` will be raised.

I wasn't sure whether my changes belonged in `_base.py` or `_sktime.py`. Let me know if I've got it wrong and I can change it.

No

#### Any other comments?

The old `compute_pred_int` code had a minor bug in it. It's supposed to return a list of intervals if you provide an iterable of alphas. However if you provide an iterable of length 1, `compute_pred_int` just gives you a single set of intervals, not in a list. This has been fixed now, so the output type is based on the type of input provided for `alpha`, rather than the number of alphas provided.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,5 +1,0 @@
-# Even if empty this file is useful so that when running from the root folder
-# the local codebase is added to sys.path by pytest. See
-# https://docs.pytest.org/en/latest/pythonpath.html for more details.  For
-# example, this allows us to build extensions in place and run pytest using the package
-# from the local folder rather than the one from site-packages.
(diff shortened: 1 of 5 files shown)


mloning (approved):
Thanks @big-o, looks all good to me, will merge those changes now!

PR #389, opened 2020-09-04
Refactored SFA implementation (additional features and speed improvements)

patrickzib:
I am the author of the SFA paper.

I have refactored the SFA implementation:
- Add anova f-test, 
- equi-width-binning, 
- simplified information-gain binning, 
- use NUMBA for faster MFT, 
- added a test case for SFA, 
- speed optimizations
- replaced Fourier Transform O(n^2) with fast fourier transform (n log n)

This improves the speed of the implementation and adds additional features as well as a test case

To Numba (but Rocket has the same dependency)

#### Any other comments?

I will add WEASEL based on this in the future.


MatthewMiddlehurst (approved):
No issues with any of the changes here. Thanks for contributing Patrick.

Some changes in TDE output (I assume from the IGB change), but I doubt it would be significant and there aren’t any baseline results from the previous version to compare it against anyway.

Will be interesting to see how much of an impact the speed ups have.

sktime/transformers/series_as_features/tests/test_sfa.py
```diff
@@ -0,0 +1,132 @@
...
+                                alphabet_size=alphabet_size,
+                                window_size=window_size, norm=norm,
+                                binning_method=binning).fit(X, Y)
+
+                        print("Norm", norm, "Anova", anova)
+                        print(np.shape(p.breakpoints), word_length,
+                              window_size)
+                        print("dft_length", p.dft_length, "word_length",
+                              p.word_length)
+                        # assert(np.shape(p.breakpoints)==
```
mloning, line 131: Is this meant to be commented?

mloning (approved):
Looks good to me, thanks for the update! Left one comment in the code below, but I'm happy to merge it.

sktime/transformers/series_as_features/tests/test_sfa.py
patrickzib: Thanks for pointing it out. Yes, the assert was kind of hard to formulate thus I uncommented it. My purpose of the test was to check that there is no exception thrown. I am not sure how to formulate that using pytest (other than just running the test and waiting for one to be thrown as I do now ;)

TonyBagnall:
thanks Patrick

PR #390, opened 2020-09-04
Add support for exogenous variables to forecasting reduction

SebasKoel:
Reference Issues/PRs
#376, #380, #382

Updated ReducedRegressionForecaster so for strategies recursive and direct we can now use multivariate/exogenous variables. 

PR #385 needs to be merged in order to create an Unit test for this since now it doesn't have a dataset available for this.


SebasKoel:
I'm not sure if the assertion of the coefficients in the test with actual values is the right way to do it. To prevent rounding errors for different versions of python or used packages I test a small range.

mloning:
Check out the `np.testing` module, they have something like `np.testing.assert_almost_equal`

sktime/datasets/base.py
```diff
@@ -639,4 +639,4 @@ def load_uschange(y_name="Consumption"):
     if y_name != "Quarter":
         data = data.drop("Quarter", axis=1)
     X = data.drop(y_name, axis=1)
-    return y, X
+    return X, y
```
mloning, line 642: Check out the `load_longley` function above in the same file, we need to make sure both are consistent in terms of input arguments and output.

mloning: We now also support date/time indexes so we can replace the `pd.Int64Index` above with a `pd.PeriodIndex`

sktime/forecasting/compose/tests/test_multivariate_regression.py
```diff
@@ -0,0 +1,62 @@
...
+    y_pred = forecaster.predict(fh=fh)
+
+    assert len(y_pred) == len(y_test)
+
+    coefs = forecaster.regressor_.coef_
+    intercept = forecaster.regressor_.intercept_
+
+    assert len(coefs) == (len(X.columns) + 1) * window_length
+    assert isinstance(intercept, float)
+    assert 0.33 < coefs[0] < 0.34
```
mloning, line 37: Where do these values come from?

sktime/forecasting/compose/tests/test_multivariate_regression.py
```diff
@@ -0,0 +1,62 @@
...
+    assert -0.09 > coefs[-1] > -0.10
+    assert 0.39 < intercept < 0.40
+
+    forecaster = ReducedRegressionForecaster(
```
mloning, line 41: This should probably be a separate test for the direct strategy

mloning (commented):
Sorry @SebasKoel for the delay, wasn't aware that this PR was ready for review! 

Left a few minor comments.

sktime/datasets/base.py
SebasKoel: I didn't see this, will change load_longley to be consistent with the rest. _load_dataset in base.py uses the X, y order as well.

mloning: But that's classification, I think it makes more sense to have y, X no, because you always have the y in forecasting but not always X?

SebasKoel: Maybe. I don't think it's that big of an issue, but y, X is fine by me as well. However I think the important part is to choose one order and apply this through the entire package. We shouldn't have classification, regression and forecasting use different orders. Also internal class methods should hold this order.

mloning: I don't agree, we follow sklearn for classification/regression and use X, y. For forecasting, because there's isn't always an X, it doesn't make sense to have X, y because that would change the first output if X isn't there, so I suggest to have y, X for forecasting.

mloning:
Hi @SebasKoel, are you still thinking about completing this PR? Let me know if you need help with anything here!

SebasKoel:
> Hi @SebasKoel, are you still thinking about completing this PR? Let me know if you need help with anything here!

PR #391, opened 2020-09-08
WEASEL implementation

patrickzib:
This is a WEASEL classifier implementation. 

The pull-request contains:
- The WEASEL classifier using the supervised SFA implementation
- Some performance fixed for SFA: One limitation was the use of Pandas.Series as a return type, which I removed for speed improvements
- A basic WEASEL test case
- A Bugfix for BOSS
- Updated the jupyter-notebook for dictionary classifiers


docs/source/api_reference.rst
```diff
@@ -46,6 +46,7 @@ Dictionary-based
 
     BOSSIndividual
     BOSSEnsemble
+    WEASEL
```
mloning, line 49: Seems like we forgot to add TDE to the docs
```suggestion
    WEASEL
    TemporalDictionaryEnsemble
    IndividualTDE
```

sktime/classification/dictionary_based/tests/test_boss.py
```diff
@@ -35,6 +35,19 @@ def test_individual_boss_on_gunpoint():
     testing.assert_array_equal(probas, individual_boss_gunpoint_probas)
 
 
+def test_boss_on_power_demand():
```
mloning, line 38: I appreciate this new test and the one below for WEASEL too, I'm only a bit worried about limited run times on free CI services, if there's anything you can do to make them quicker that would be great e.g. using a subset of the data, fixing certain parameters, but haven't checked yet how much time they actually need

sktime/transformers/series_as_features/dictionary_based/_sax.py
```diff
@@ -185,10 +190,19 @@ def push(self, letter):
 
     def shorten(self, amount):
         # shorten a word by set amount of letters
-        self.word = self.right_shift(self.word, amount * 2)
+        _BitWord.shorten_word(self.word, amount)
```
mloning, line 193: why not use self?
```suggestion
        self.shorten_word(self.word, amount)
```
here and below in `shorten_word()` too, also not sure why making it a static method but then passing in `self.word` but I may not understand the full context here, perhaps leave a short comment to clear this up :)

mloning (approved):
Thanks @patrickzib, code looks very clean in general, left a few minor comments, but happy to merge from my side.

sktime/transformers/series_as_features/dictionary_based/_sax.py
patrickzib: Very good spot on :) 

My main reason was to get rid of the _BitWord representation within BOSS and WEASEL. Thus these calls to _BitWord are static from outside the class now and I do only store lists of "int"-representation of the words. I though this safes some memory compared to storing lists of BitWords, which was the old representation.

To maintain backward-compatibility, I didnt want to remove the old methods though - maybe make them deprecated? What do you think?

sktime/classification/dictionary_based/tests/test_boss.py
patrickzib: Oh, sure. I was not aware of this issue. I noticed a problem while editing the jupyter-notebook, which was not covered by tests. Thus, I added a test. 

The additional BOSS test takes 9s. I am fine to uncomment the test again - what do you think?

The two tests in WEASEL take only 4.5s, so they are quite fast. Is that ok?

docs/source/api_reference.rst
patrickzib: Just added :)

sktime/classification/dictionary_based/tests/test_boss.py
mloning: Yeah that seems fine!

sktime/transformers/series_as_features/dictionary_based/_sax.py
mloning: I'm not too worried about backward compatibility at this point, a lot of our API is still somewhat experimental :)

patrickzib: I see :)

the BitWord instances were never really used other than as an intermediate step. I have removed that intermediate step. Therefor, I have removed the member variables of the BitWord and made it a static library + refactored SAX and SFA.

sktime/classification/dictionary_based/tests/test_boss.py
patrickzib: I have reduced the runtime by using a subset of the test data, too.

sktime/transformers/series_as_features/dictionary_based/_sax.py
```diff
@@ -166,29 +173,24 @@ def _generate_breakpoints(self):
...
 
-    def shorten(self, amount):
+    @staticmethod
+    def shorten_word(word, amount):
         # shorten a word by set amount of letters
-        self.word = self.right_shift(self.word, amount * 2)
-
-    def create_bigram(self, other_word, length):
-        return (self.word << length) | other_word.word
+        return _BitWord.right_shift(word, amount * 2)
```
mloning, line 193:
```suggestion
        return self.right_shift(word, amount * 2)
```

mloning (dismissed):
I suggest one minor change, otherwise looks good to go. Happy to merge once @TonyBagnall approves it :)

sktime/transformers/series_as_features/dictionary_based/_sax.py
patrickzib: Thank you for the suggestion. 

There was no self in the static methods to call self.right_shift. Thus, I have changed the methods to @classmethod such that I can call cls.right_shift(...)

```python    
   @classmethod
    def shorten_word(cls, word, amount):
        # shorten a word by set amount of letters
        return cls.right_shift(word, amount * 2)
```

TonyBagnall (dismissed):
all looks good to me Patrick, thanks for the contribution

patrickzib:
The failed test on alan-turing-institute.sktime is due to a problem with _rise.RandomIntervalSpectralForest ... I did not touch this. Is it possible to rerun the test without be pushing any new commits?

sktime/transformers/series_as_features/dictionary_based/_sax.py
patrickzib: is that ok with you?

mloning: Still seems complicated to me, I'd either just have standalone functions or a simple class, but maybe I'm missing something, happy to leave it as it is.  :)

patrickzib:
I am sorry, I had to make a new commit to make the tests pass ...

TonyBagnall (approved):

sktime/transformers/series_as_features/dictionary_based/_sax.py
patrickzib: thank you :)

mloning (approved):

mloning:
Thanks @patrickzib for the great contribution! 🎉

Issue #347, opened 2020-07-19
Supporting datetime indices


shruthibalaji2307:
**Is your feature request related to a problem? Please describe.**
When we load a pandas series as input data, Python datetime indices are not supported by sktime, only simple numerical indices are supported.  

**Describe the solution you'd like**
The datetime index provides much more clarity while plotting and analyzing the data. Sktime must support series that have a datetime index and add appropriate computations in the datetime context.

**Describe alternatives you've considered**
I have tried using the numerical index, but it does not clearly indicate the unit of the time, whether it represents seconds, minutes, days, weeks, or months.

**Additional context**
More details about this are posted in this Bug : https://github.com/alan-turing-institute/sktime/issues/346

mloning:
Thanks for creating the issue and volunteering to work on this! 

I think you already identified the main parts of the code where changes would have to be implemented: the handling of observation (oh) and forecasting horizons (fh). 

Adding support for date time indices will also require a lot more extensive unit testing. Some package that we interface may not support date time indexing (e.g. [pmdarima](https://github.com/alkaline-ml/pmdarima) as far as I remember).

shruthibalaji2307:
As far as I've seen, the arima class of the pmdarima package wraps the statsmodel's ARIMA. And it appears that statsmodels supports date time indexing. Source : https://alkaline-ml.com/pmdarima/1.0.0/_modules/pmdarima/arima/arima.html

I will do elaborate Unit testing anyway after making appropriate code changes. Can you please assign this to me?

fkiraly:
@mloning, would it perhaps also be a good time to think about the API and use of indices? Currently, the API is integer index (`iloc`) based, and time indices are ignored both in application of algorithms, and in the horizon interface, if I'm not mistaken. Shall we think about specs?

fkiraly:
Alright @mloning, @shruthibalaji2307, I made a proposal on how support for datetime indices could look here, please have a look at [this wiki page](https://github.com/alan-turing-institute/sktime/wiki/Design-proposal:-forecasting-API-re-work-including-support-for-datetime-indices).

There are two versions, and the "luxurious" version relies on a bit o a re-work of input/output checks that allow for multiple input/output types, I've also added a [design for that](https://github.com/alan-turing-institute/sktime/wiki/Design-proposal%3A-multi-type-input-output) to the wiki.

Comments and criticism very much appreciated. 
I'd prefer the luxurious version, but it might be a bit of work to implement.

[...]

Issue #220, opened 2020-01-28
Forecasting wish list


mloning:
# Forecasting wish list

### Atomic 
- [ ] DummyForecaster (strategies={"last", "seasonal-last", "mean"})
- [ ] TrendForecaster (linear, polynomial, etc)
- [ ] ThetaForecaster
- [ ] ExponentialSmoothingForecaster
- [ ] AutoARIMAForecaster

### Reduction
- [ ] ReducedRegressionForecaster
- [ ] ReducedTimeSeriesRegressionForecaster

## Transformers
- [ ] Detrender (series to series, meta-estimator that works with any forecaster)
- [ ] Imputer

## Composition
- [ ] TransformedTargetForecaster (pipeline with target transformations)

### Temporal CV
- [x] RollingWindowSplit
- [ ] ExpandingWindowSplit
- [ ] ManualWindowSplit

### Tuning
- [ ] ForecastingGridSearchCV

big-o:
Due to the ThetaForecaster's dependency on SES, #198 also (re-)implements the ExpSmoothingForecaster in the new framework.

hweiTPR:
Is there a way to tune both sklearn & sktime hyperparameters at the same time, when using regressors from sklearn? If not. It could really useful to have it.

mloning:
Hi @hweiTPR, yes you can do something like this: 
```python

regressor_param_grid = {"n_estimators": [100, 200, 300]}
forecaster_param_grid = {"window_length": [10, 12, 14]}
regressor = GridSearchCV(RandomForestRegressor(), param_grid=regressor_param_grid)
forecaster = RecursiveRegressionForecaster(regressor)
cv = SlidingWindowSplitter(initial_window=100)
gscv = ForecastingGridSearchCV(forecaster, cv=cv, param_grid=forecaster_param_grid)
gscv.fit(y_train)

print(gscv.best_params_, gscv.best_forecaster_.regressor_.best_params_)
>>> {'window_length': 12} {'n_estimators': 100}
```

hweiTPR:
That looks great! Would be helpful to have this in the documentation!

mloning:
@hweiTPR I mention it already in the forecasting tutorial below cell 27: https://github.com/alan-turing-institute/sktime/blob/master/examples/01_forecasting.ipynb but without showing the code! Would appreciate a PR if you'd like to add it! 🙂

huayicodes:
[...]

PR #392, opened 2020-09-09
Add date-time support for forecasting framework

mloning:
#### Reference Issues/PRs
#347, #220

* Support for `pd.DatetimeIndex`, `pd.PeriodIndex` in data and forecasting horizon, including helper functions to patch mathematical operations on pandas date/time objects for our purposes (mostly returning integers for known time unit, rather than time delta objects)
* Included forecasting data sets are now returned with `pd.PeriodIndex`
* Support for absolute forecasting horizon specification via refactored `ForecastingHorizon` class
* Easy importing via `sktime.forecasting.all` module enabling `from sktime.forecasting.all import *` following [fastai](https://www.fast.ai/2020/02/13/fastai-A-Layered-API-for-Deep-Learning/)
* A few minor changes (typos, formatting, etc)
* Corresponding updates to forecasting notebook
* New unit tests for supported combinations of data index types and forecasting types

* seaborn (soft dependency, for plotting) 

#### Key files to review 
Since the forecasting horizon was used in basically every forecasting algorithm, this PR makes changes to many files. Key changes include:

* `sktime/forecasting/base/_fh.py` (implementation of forecasting horizon class)
* `sktime/forecasting/base/tests/test_fh.py` (unit tests for forecasting horizon class)
* `sktime/forecasting/tests/_config.py` (check SUPPORTED_INDEX_FH_COMBINATIONS)
* `sktime/forecasting/tests/test_all_forecasters.py` (unit testing framework for all forecasters)
* `sktime/utils/datetime.py` (date/time helper functions)

#### Comments
* Supporting only `pd.PeriodIndex` and not `pd.DatetimeIndex` would make the code a bit simpler as `pd.DatetimeIndex` requires additional patches to return integers for durations/time deltas and additional type distinctions (dispatch), however it seems like `pd.DatetimeIndex` supports some functionality which is not supported by `pd.PeriodIndex` (e.g. `is_month_end`)

#### Future work
* refactor `BaseLastWindowForecaster` and reduce inheritance as much as possible


mloning:
@fkiraly I've made some changes after our conversation, mostly refactored the helpers functions, and added a list of key files to review.

fkiraly (commented):
All in all, this seems to be working.
Some comments below, as discussion/questions.
One comment for other reviewers: I think the changes in `_sktime` are also important to look at.

On the design level, I am a bit worried about:
(i) unclear scityping
(ii) no support for time like relative indices

I'd suggest to have a uniform external interface, which has only the following distinctions in the type:
* relative/absolute
* time like vs index like

and think about machine type defaults. Subclassing, maybe?
Currently, there is a lot of overhead due to what I think is not a very clear distinction between index like and time like.

More detailed comments:

`_fh.py`
* there are a lot of type based case distinctions in the code. Would it be cleaner, and more extensible, to use `functools.singledispatch`?
* I'm disturbed by the fact that I cannot come up with a combination of decorators or inheritance that would solve the problem with delegating methods and wrapping of the pandas objects - maybe @big-o has an idea?
It seems to work though.
* my gut feeling is also a bit mixed about the decision to have multiple internal representations (the `SUPPORTED_INDEX_TYPES`, instead a "standardized internal state". Though again, it is hard to come up with cases or types that do not somehow duplicate the separation in different types of indices in `pandas`.
* do we not want to support time or time difference like indices for relative indices? Any reason why we don't want `timedelta` to be an option in relative?
* as a side effect to disallowing relative time like horizons, we get `_coerce_duration_to_int`, is that correct?
* instead of manually looping through `__len__` and `__getitem__`, should these not be added to `DELEGATED_METHODS`?
* for partial downwards compatibility, should `to_out_of_sample` remain `out_of_sample`?
* instead of `len(fh_oos) == len(fh):` check etc, perhaps introduce a `is_past` or so check?

`utils.datetime`
* this is very hard to read, since there are no docstring, and the devil is in the details: which argument types (integer, datetime) is expected, which are supported? It's easy to introduce subtle bugs, and lack of specs makes maintenance by others very difficult.
* again, due to numerous case distinctions by type, seems a candidate for `functools.singledispatch`?

mloning:
Thanks @fkiraly for the review. 

* Re adding support for more types like `pd.TimeDelta`: (1) it would add more case distinctions (e.g. because pandas does not consistently handle collections of duration types (see below)), (2) specifying a forecasting horizon with durations seems redundant as the time unit will always be specified in the data, (3) I don't think using duration types is very intuitive from a user's perspective and can't think of a situation where I would prefer passing duration types rather than integers
* We need `_coerce_duration_to_int()` because some of our algorithms internally and other package we interface require integer forecasting horizons, not duration types.  
* One reason why we need so many case distinctions is that pandas doesn't seem to be very consistent with the typing. For example, multiple pd.TimeDelta (duration for timestamp type) are collected in a pd.TimeDeltaIndex, but multiple pd.DateOffsets (from duration for pd.Period type) are collected in a simple pd.Index with individual pd.DateOffsets inside. 
* I'm going to move `__getitem__` and `__len__` to `DELEGATED_METHODS`
* Re downward compatibility, at this stage I prefer consistency of method names over compatibility, adding deprecation cycles will complicate and slow dev work considerably I think, also the ForecastingHorizon was mainly used internally, it was included in the notebooks for example
*  I'm going to add some unit tests for the date-time utility functions and better docstrings
* I'm going to add something like `has_out_of_sample_values` and `has_in_sample_values` and caching to avoid computing the same things twice

mloning:
I've implemented a few suggested changes. Let me know if there's anything else you would definitively change.

fkiraly (approved):
Tentatively happy with the state and recent changes.

Even if issues remain (unspotted etc), it´s downwards compatible in signature, and this will allow us to get feedback from users.

PR #350, opened 2020-07-19
Draft PR for ets

HYang1996:
#### Reference Issues/PRs

Fixes #327 

#### What does this implement/fix? Explain your changes.

Implements automatic ETS based on R version.

#### PR checklist for new estimators 

- [ ] I've added unit tests and made sure they pass locally. 
- [ ] I've updated the existing example notebooks or provided a new one to showcase how my estimator works.
- [ ] I've updated sktime's [estimator overview](https://github.com/alan-turing-institute/sktime/blob/master/ESTIMATOR_OVERVIEW.md) and I'm committed to maintain my contribution and provide bug fixes if necessary.
- [ ] I've added myself to the list of contributors

#### Any other comments?

This is a draft PR, the class is not completely finished.

PR #393, opened 2020-09-10
ETSModel with auto-fitting capability similar to R implementation

### LINKED ISSUE #327: Implement Automatic Exponential Smoothing

Add the Automatic Exponential Smoothing feature to BaseSktimeForecaster.

Translate [the fable version on R](https://github.com/tidyverts/fable/blob/master/R/ets.R) to Python and add to sktime.

There might be existing Python packages that implement this feature.

**Additional context**

R documentation of the [ETS function](https://www.rdocumentation.org/packages/fable/versions/0.0.0.9100/topics/ETS).

HYang1996:
#### Reference Issues/PRs

Fixes #327, replacing #350 .

Created a wrapper for statsmodels' latest implementation of ETSModel, and added an auto-fitting option which is similar to the implementation in R.

#### PR checklist for new estimators 

- [x] I've updated sktime's [CODEOWNERS](https://github.com/alan-turing-institute/sktime/blob/master/CODEOWNERS) and I'm committed to maintain my algorithm.
- [x] I've added myself to the list of contributors

#### Any other comments?

Simple tests have been performed to compare results from this forecaster against those from the statsmodels' version, and auto-fitting capability is tested against R implementation on the airline dataset.


mloning (commented):
Hi @HYang1996, thanks for the great work! Code is very clean and readable! 

* the continuous integration tests fail at the moment, not entirely sure what the problem is
* I think we need to update the statsmodels version to `0.12.x` here: [binder](https://github.com/alan-turing-institute/sktime/blob/master/.binder/requirements.txt), [build](https://github.com/alan-turing-institute/sktime/blob/master/build_tools/requirements.txt) and [setup.py](https://github.com/alan-turing-institute/sktime/blob/6557fd8aee0b0fbdb6667413254a820e37050598/setup.py#L22)
* could you add some unit tests where we check results against R's values, I'm happy if you just copy and paste results from R and check if our version returns more or less the same values, the `np.testing` module has some good tools do to that)
* Just another idea: I recently came across this function for storing attributes: https://fastcore.fast.ai/utils.html#store_attr perhaps we could use something similar, that would make it easier to maintain the init functions of all our estimators

sktime/forecasting/ets.py
```diff
@@ -0,0 +1,314 @@
...
+            if self.allow_multiplicative_trend:
+                trend_range = ['add', 'mul', None]
+            else:
+                trend_range = ['add', None]
+            seasonal_range = ['add', 'mul', None]
+            damped_range = [True, False]
+
+            # Fit model, adapted from:
+            # https://github.com/robjhyndman/forecast/blob/master/R/ets.R
+            for error in error_range:
```
mloning, line 220: This is probably something we want to parallelise. You could try to make this into a single for loop using [`itertools.product()`](https://docs.python.org/2/library/itertools.html#itertools.product) and joblib.

mloning:
Seems like the CI servers time out because we hit the one hour mark. I may have to speed them up a bit.

mloning:
Okay, now they seem to work, just restarted them. Not sure, I believe Azure had some maintenance work going on over the last few days, perhaps that caused the slow downs.

sktime/forecasting/ets.py
mloning:
```python
from itertools import product
from joblib import delayed, Parallel

def _iter(x, y):
    for a, b in product(x, y):
        if a == 1:
            continue
        yield a, b

def _fit(a, b):
    print(a, b)
    return a + b

x = [1, 2, 3]
y = [4, 5, 6]

results = Parallel(n_jobs=1)(delayed(_fit)(a, b) for a, b in _iter(x, y))
```

mloning:
The failure now seems to be related to the changed module structure in statsmodels.

mloning:
Failing docs due to pmdarima and statsmodels )0.12. incompatibility due to this bug: https://github.com/statsmodels/statsmodels/issues/7019

HYang1996:
Unit tests passed locally but failed in checks since rpy2 is not included. The test importing dataset from R is therefore removed for now.

HYang1996:
I'm not exactly sure why the pre-commit check is failing.

docs/source/api_reference.rst
```diff
@@ -200,6 +200,17 @@ Exponential Smoothing
 
     ExponentialSmoothing
 
+ETS Models
```
mloning, line 203: Shall we not add it to the family of exponential smoothing models?

sktime/forecasting/tests/test_ets.py
```diff
@@ -0,0 +1,42 @@
...
+
+    forecaster = AutoETS(autofitting=True, sp=12, n_jobs=-1,
+                         allow_multiplicative_trend=True)
+    forecaster.fit(y)
+    fitted_forecaster = forecaster._fitted_forecaster
+    fit_result = [fitted_forecaster.error,
+                  fitted_forecaster.trend,
+                  fitted_forecaster.seasonal]
+
+    assert_array_equal(fit_result_R, fit_result)
```
mloning, line 42: I'd make this a separate function/test, makes it easier to debug if one of them fails

sktime/forecasting/ets.py
```diff
@@ -0,0 +1,338 @@
...
+        dates=None,
+        freq=None,
+        missing="none",
+        start_params=None,
+        maxiter=1000,
+        full_output=True,
+        disp=True,
+        callback=None,
+        return_params=False,
+        autofitting=False,
```
mloning, line 171: I'd just call it "auto" rather than "auto fitting"

sktime/forecasting/ets.py
```diff
@@ -0,0 +1,338 @@
...
+    References
+    ----------
+    [1] Hyndman, R.J., & Athanasopoulos, G. (2019) *Forecasting:
+        principles and practice*, 3rd edition, OTexts: Melbourne,
+        Australia. OTexts.com/fpp3. Accessed on April 19th 2020.
+
+    The following parameters are adapted from the ones on R
+    (https://www.rdocumentation.org/packages/forecast/versions/8.12/topics/ets),
+    and are used for automatic model selection:
+
```
mloning, line 120: I would move the section `References` to the end of the docstring, I see that you want to reference the different parameters to the different implementations, but I think it's okay to just mention in the references that they come from both model implementations

sktime/forecasting/ets.py
```diff
@@ -0,0 +1,338 @@
...
+                    self.information_criterion != 'bic' and \
+                    self.information_criterion != 'aicc':
+                raise ValueError('information criterion must '
+                                 'either be aic, bic or aicc')
+
+            # Fit model, adapted from:
+            # https://github.com/robjhyndman/forecast/blob/master/R/ets.R
+
+            # Initialise iterator
+            def _iter(error_range, trend_range, seasonal_range, damped_range):
```
mloning, line 239: nice, looks good!

sktime/forecasting/ets.py
```diff
@@ -0,0 +1,338 @@
...
+
+            # Fit models
+            _fitted_results = Parallel(n_jobs=self.n_jobs)(
+                delayed(_fit)(error, trend, seasonal, damped)
+                for error, trend, seasonal, damped in _iter(
+                        error_range, trend_range, seasonal_range, damped_range
+                    ))
+
+            # Select best model based on information criterion
+            for result in _fitted_results:
```
mloning, line 298: We can do this in a loop, but I wonder if there's a faster way using `np.arrray` where we can simply call something like `np.argmax(_ics)` to find the index of the model with the highest IC, but it's always a bit tricky to get the returned list (`_fitted_results`) into an array ...

sktime/forecasting/ets.py
```diff
@@ -0,0 +1,338 @@
...
+                )
+                _fitted_forecaster = _forecaster.fit(
+                    start_params=self.start_params,
+                    maxiter=self.maxiter,
+                    full_output=self.full_output,
+                    disp=self.disp,
+                    callback=self.callback,
+                    return_params=self.return_params
+                )
+                return _forecaster, _fitted_forecaster
```
mloning, line 288: Why do we return the unfitted `_forecaster` here?

sktime/forecasting/ets.py
```diff
@@ -0,0 +1,338 @@
...
+            self._fitted_forecaster = self._forecaster.fit(
+                start_params=self.start_params,
+                maxiter=self.maxiter,
+                full_output=self.full_output,
+                disp=self.disp,
+                callback=self.callback,
+                return_params=self.return_params
+            )
+
+    def summary(self):
```
mloning, line 337: Since this is a method exposed to users, could you add a docstring?

mloning (commented):
@HYang1996 looks great! I left a few minor comments, so it's almost good to go now! 

Don't worry about the pre-commit CI checks, I changed the other day and we can ignore them on this PR. If you want to find out more on how they work, check out the [CONTRIBUTING.md](https://github.com/alan-turing-institute/sktime/blob/master/CONTRIBUTING.md#Continuous-integration)

HYang1996:
AutoETS doesn't seem to be appearing in the online documentation.

HYang1996:
> AutoETS doesn't seem to be appearing in the online documentation.

Fixed

mloning (commented):
Hi @HYang1996 

Looks great! I'm very happy with it, excellent work! 

One last question: Currently, the unit tests only check if we select the right kind of parameters but not exact numerical values. Is this something we could do using the results from R even if they are just approximately equal?

mloning (approved):
Excellent work, thanks a lot @HYang1996! 

I'll leave the PR open for a bit longer before I merge in case others want to comment.

mloning (approved):

mloning:
Thanks again @HYang1996, now merged! 🎉

PR #399, opened 2020-09-21
Update 01_forecasting.ipynb

akanz1:
While this does not change the plot, it improves clarity since it explicitly uses the output of forecaster.predict(fh), i.e., "y_last" instead of the numpy interpolation from the previous cell (y_pred).


mloning:
Alright, make sense, thanks!

Issue #325, opened 2020-06-30
Raise appropriate error when soft dependencies are missing


mloning:
We have a number of soft dependencies that we don't require to install sktime or to run most of our functionality. But, they are require to run some modules (e.g. tsfresh, numba or pmdarima). Currently, when users don't have them installed, they run into a module not found error which isn't really informative. For example, #322 

We could add checks and more appropriate error messages at the beginning of the files suggesting to run `pip install <soft_dependency>`. 

We could add the following function:
```python
from importlib import import_module

def _check_imports(package):
    try:
        import_module(package)
    except ModuleNotFoundError:
        msg = f"No module named {package}. {package} is a soft dependency in sktime and is not included in the sktime installation. Please run `pip install {package}`."
        raise ModuleNotFoundError(msg)
```

We could then call this function at the top of modules with soft dependencies:
```python
soft_dependency = "tsfresh"
_check_imports(soft_dependency)
```

PR #407, opened 2020-09-30
Resolved warnings in ipynb examples

alwinw:

When running the example ipynb in a clean environment, `ModuleNotFoundError` are raised for pmdarima (sktime/forecasting/arima.py#269) and tsfresh (sktime/transformers/series_as_features/summarize/_tsfresh.py).

Reran and cleaned up .ipynb outputs (example of previous warnings below):

image

image

#### Any other comments?

Really liking this package! Hoping to contribute more than fixing typos and updating docs in the future.


mloning:
Hi @alwinw, thanks for your work on this. We treat tsfresh and pmdarima as a soft dependency to keep our core dependencies to a minimum. That's why we haven't included it in `setup.py`. But perhaps we should add a `!pip install tsfresh` and ``!pip install pmdarima` line to the notebook?

And thanks for cleaning the notebook output!

alwinw:
Hi @mloning, thanks for explaining that to me. Interesting approach since a user can `pip install sktime` and hit a `ModuleNotFoundError` when trying to run selected functions which could become annoying. I'll remove it as a core dependency and add it into the notebook. 

Just changed my PR into a WIP one until I've fixed that up and have it passing the CI pipeline (I'll install precommit locally as well so my commit history isn't full of flake8 or black fixes!)

mloning:
Since there are a number of other packages we'd like to support without blowing up our list of hard dependencies, we opted for that middle way. But happy to hear about alternatives! 

One suggestion we've received in the past is to add import checks and raise a more helpful error message when soft dependencies are missing (see #325).

mloning:
Closed in favour of #410

PR #408, opened 2020-10-01
Parallelized TimeSeriesForest using joblib.

### LINKED ISSUE #381: Parallelize algorithms
- [ ] [Time series forest](https://github.com/alan-turing-institute/sktime/blob/master/sktime/classification/interval_based/_tsf.py)

* Replace loops with joblib parallelized computations or numba.

kkoziara:
#### Reference Issues/PRs

Resolves #381.

Implementation parallelizes `sktime.classification.interval_based.TimeSeriesForest` training and prediction using joblib library.


sktime/classification/interval_based/_tsf.py
```diff
@@ -97,6 +109,66 @@ def __init__(self,
         # We need to add is-fitted state when inheriting from scikit-learn
         self._is_fitted = False
 
+    @classmethod
```
mloning, line 112: What's the reason for using @classmethod here? To me, it seems like a @staticmethod, since `cls` isn't used inside the function. But perhaps a stand-alone function would be easiest.

sktime/classification/interval_based/_tsf.py
```diff
@@ -97,6 +109,66 @@ def __init__(self,
...
+        intervals = np.zeros((n_intervals, 2), dtype=int)
+        for j in range(n_intervals):
+            intervals[j][0] = rng.randint(series_length - min_interval)
+            length = rng.randint(series_length - intervals[j][0] - 1)
+            if length < min_interval:
+                length = min_interval
+            intervals[j][1] = intervals[j][0] + length
+        return intervals
+
+    @classmethod
```
mloning, line 129: Same here, why do you use @classmethod here? :)

sktime/classification/interval_based/_tsf.py
```diff
@@ -97,6 +109,66 @@ def __init__(self,
...
+        of input data X.
+        """
+        n_instances, _ = X.shape
+        n_intervals, _ = intervals.shape
+        transformed_x = np.empty(shape=(3 * n_intervals, n_instances), dtype=np.float32)
+        for j in range(n_intervals):
+            X_slice = X[:, intervals[j][0]: intervals[j][1]]
+            means = np.mean(X_slice, axis=1)
+            std_dev = np.std(X_slice, axis=1)
+            slope = cls._lsq_fit(X_slice)
```
mloning, line 142: I think `_lsq_fit` can probably also be a stand-alone helper function which lives in this file but is not attached to the class. The same for `_get_transformed_data` which I would rename to simply `_transform`. 

Unless we want to expose the `_transform` as a public `transform` method to users , but that may introduce some issues with estimator types ...

sktime/classification/interval_based/_tsf.py
```diff
@@ -97,6 +109,66 @@ def __init__(self,
...
+            transformed_x[3 * j] = means
+            transformed_x[3 * j + 1] = std_dev
+            transformed_x[3 * j + 2] = slope
+
+        return transformed_x.T
+
+    @classmethod
```
mloning, line 149: See above, once we detach the class methods we can make this to a stand-alone function too.

sktime/classification/interval_based/_tsf.py
```diff
@@ -128,37 +200,24 @@ def fit(self, X, y):
...
-            tree.set_params(**{"random_state": self.random_state})
-            transformed_x = transformed_x.T
-            tree.fit(transformed_x, y)
-            self.classifiers.append(tree)
+
+        self.classifiers = []
+        self.intervals = []
+
+        for estimator, intervals in Parallel(n_jobs=self.n_jobs)(
+            delayed(self.__class__._fit_estimator)(
```
mloning, line 208: With the suggested changes above, we can then simply call `delayed(_fit_estimator)` here

sktime/classification/interval_based/_tsf.py
```diff
@@ -180,6 +239,15 @@ def predict(self, X):
         proba = self.predict_proba(X)
         return np.asarray([self.classes_[np.argmax(prob)] for prob in proba])
 
+    @classmethod
```
mloning, line 242: see above

sktime/classification/interval_based/tests/test_tsf.py
```diff
@@ -0,0 +1,60 @@
...
+    tsf.fit(X_train, y_train)
+
+    n_classes = len(np.unique(y))
+    proba = tsf.predict_proba(X_test)
+
+    assert proba.shape == (
+        X_test.shape[0],
+        n_classes,
+    ), "Incorrect shape of classes probabilities"
+    np.testing.assert_array_equal(
```
mloning, line 26: This is a general test which should pass for all probabilistic classifiers, perhaps we could add this here: https://github.com/alan-turing-institute/sktime/blob/master/sktime/series_as_features/tests/test_all_series_as_features_estimators.py

sktime/classification/interval_based/tests/test_tsf.py
```diff
@@ -0,0 +1,60 @@
...
+        np.ones(X_test.shape[0]),
+        np.sum(proba, axis=1),
+        "Incorrect classes probabilities",
+    )
+
+    tsf_score = tsf.score(X_test, y_test)
+
+    dummy = DummyClassifier(strategy="most_frequent")
```
mloning, line 34: Not sure if we need the DummyClassifier comparison here, but I appreciate the added test for the overall score

sktime/classification/interval_based/tests/test_tsf.py
```diff
@@ -0,0 +1,60 @@
...
+    dummy.fit(X_train, y_train)
+
+    dummy_score = dummy.score(X_test, y_test)
+
+    assert tsf_score > dummy_score, "Classifier performs worse than dummy classifier"
+
+    assert tsf_score >= 0.95, "Classifier performs worse than expected"
+
+
+def test_multiproc_determinism():
```
mloning, line 44: Again this seems like a general tests which all our estimators with the `n_jobs` parameter should pass ... perhaps add it here: https://github.com/alan-turing-institute/sktime/blob/master/sktime/tests/test_all_estimators.py

mloning (commented):
Great work @kkoziara, I really appreciate these changes! It's now much more readable and modular. And thanks for adding the unit tests!

I've added some in-line comments, mostly on the use of @classmethod 

If you're looking for extra work: You could also take a look at the [`RandomIntervalSpectralEnsemble`](https://github.com/alan-turing-institute/sktime/blob/master/sktime/classification/frequency_based/_rise.py) which has a very similar structure. But perhaps best to focus on this one first🙂

sktime/classification/interval_based/_tsf.py
kkoziara: Hey, I used **@classmethod** here and in a bunch of other places to provide a way to modify the algorithm behavior through subclassing. If you think that won't be necessary or should not happen I can refactor most of this to be standalone methods?

sktime/classification/interval_based/_tsf.py
kkoziara: I agree with respect to `_lsq_fit` and the naming for `_get_transformed_data`. For **@classmethod** see my other comment.

sktime/classification/interval_based/tests/test_tsf.py
kkoziara: My thinking was as this is a randomized classifier, some internal changes can make it perform a bit worse for the given `random_state` and break the latter test but this part should always pass, but maybe it's a bit unnecessary :) I can drop this part.

sktime/classification/interval_based/tests/test_tsf.py
kkoziara: OK

sktime/classification/interval_based/tests/test_tsf.py
kkoziara: OK

sktime/classification/interval_based/_tsf.py
mloning: Why not use staticmethod for subclassing? Subclassing may actually be relevant if you look at the RandomIntervalSpectralEnsemble, which essentially only applies different transformations. You could also easily convert these algorithms into regressors by simply changing the final decision tree to a regressor.

sktime/classification/interval_based/tests/test_tsf.py
mloning: Just to be sure, these suggestions are optional, feel free to ignore them, happy to focus on TSF only!

sktime/classification/interval_based/_tsf.py
kkoziara: I agree I could use a mix of @classmethod, @staticmethod and plain methods. Maybe something like this:
`_get_intervals` - static of plain method, depending on if someone might want to select intervals some other way (probably not),
`_get_transformed_data` - static,
`_fit_estimator` - class, as it needs to call proper `_get_transformed_data` when inheriting,
`_predict_proba_for_estimator` - same as above,
`_lsq_fit` - plain

mloning: I still don't understand why we need a class method for inheritance to work, even if it's a plain method it can call `self._get_tranformed_data` no? Perhaps I'm missing something!

kkoziara: It could be either a classmethod or a plain method, but not static or stand-alone method, and I thought it's better to use classmethod as plain methods are serialized with the whole object (if I am not mistaken) when passed to other process, whereas I would expect classmethods to only go with the class object (not so sure about it though). Also I mixed up some names in my previous comment (sorry!) What I meant was:

`_get_intervals` - static method or ~~plain~~ stand-alone function, depending on if someone might want to select intervals some other way (probably not),
`_get_transformed_data` - static,
`_fit_estimator` - class, as it needs to call proper _get_transformed_data when inheriting,
`_predict_proba_for_estimator` - same as above,
`_lsq_fit` - ~~plain~~ stand-alone function

mloning: Not sure about serialisation, just always find it strange to see `cls.method` instead of `self.method`. I'd try to make it as simple as possible and stand-alone functions are probably the simplest construct, class method the most complex and the others in between, no? So I'd suggest to make them all stand-alone functions unless you want to also refactor the `RandomIntervalSpectralForest` or a `TimeSeriesForestRegressor` :)

kkoziara: OK, then. I'll refactor them to be stand-alone functions for now. Let's not introduce unnecessary complexity. If the need for subclassing etc arises, it could be reverted. Not sure if I have the bandwidth to refactor the other classes but who knows :)

sktime/classification/interval_based/tests/test_tsf.py
kkoziara: I think it's a good suggestion. I'll check first that the other classes actually pass this, as making them to do so would be out of scope here.

kkoziara: I moved this test to check all of the estimators and it seems that for `StackingForecaster` and `EnsembleForecaster` the `n_jobs` parameter is overwritten in subclass initialization, so is always set to `None`. I have code with the test which fixes this. Should I make another PR for that?

mloning: Yes, please! Good catch!

sktime/classification/interval_based/_tsf.py
mloning: Sure, thanks! Ping me when I should review again!

kkoziara: Hey, I think it's ready.

mloning (commented):
Hi @kkoziara, could we add a tests to check if results are the same as before? The problem is that this algorithm didn't have a test before even though we did some substantial testing of it against a Java implementation. So I suggest we run it on one of our datasets and compare the `predict_proba` results and add that as a test. Let me know if you can do that, otherwise I'm happy to add this before we merge!

kkoziara:
@mloning, they unfortunately won't be the same. I moved the initialization of the intervals inside the parallelized function whereas before intervals were initialized sequentially from the shared `RandomState`.
In order to have the same results as before the interval initialization would have to be moved out of the parallel calls.
Interval initialization is not very costly and I pondered keeping it not parallel exactly for this reason. I can easily rewrite this. What do you think?

mloning:
@kkoziara Yes moving the interval generation out of the parallel loop is a good idea! It'll also simplify the code a bit because the parallel call only returns a simple list of estimators! Thanks!

sktime/classification/interval_based/tests/test_tsf.py
```diff
@@ -0,0 +1,50 @@
...
+        [1.0, 0.0],
+        [0.0, 1.0],
+        [0.0, 1.0],
+    ]
+)
+
+
+@pytest.mark.parametrize("n_jobs", [1, 4])
+def test_on_gunpoint(n_jobs):
+    np.random.seed(42)
```
mloning, line 37: We don't need to call `np.random.seed` here, but need to pass it to `train_test_split`

sktime/classification/interval_based/tests/test_tsf.py
```diff
@@ -0,0 +1,50 @@
...
+        [0.95, 0.05],
+        [1.0, 0.0],
+        [1.0, 0.0],
+        [0.0, 1.0],
+        [0.0, 1.0],
+    ]
+)
+
+
+@pytest.mark.parametrize("n_jobs", [1, 4])
```
mloning, line 35: Don't think we need to iterate over n_jobs here given the other test added in #414

sktime/classification/interval_based/_tsf.py
```diff
@@ -128,37 +214,23 @@ def fit(self, X, y):
...
-                transformed_x[3 * j] = means
-                transformed_x[3 * j + 1] = std_dev
-                transformed_x[3 * j + 2] = slope
-            tree = clone(self.base_estimator)
-            tree.set_params(**{"random_state": self.random_state})
-            transformed_x = transformed_x.T
-            tree.fit(transformed_x, y)
-            self.classifiers.append(tree)
+
+        self.intervals = [
```
mloning, line 218: Could we add trailing underscores to variables generated in fit following sklearn's convention, so 

```suggestion
        self.intervals_ = [
```

Same for classifiers below

mloning (commented):
Alright, thanks for making these changes! Three last comments, will merge it once they are fixed!

mloning:
I quickly made the changes myself, will merge once all tests pass! Thanks @kkoziara for the good work, especially the refactoring, now looks a lot better and much more readable!

mloning:
Thanks again @kkoziara great stuff! 🥳

kkoziara:
@mloning, thanks for all the feedback!

PR #410, opened 2020-10-01
Resolves #325 ModuleNotFoundError for soft dependencies

### LINKED ISSUE #325: Raise appropriate error when soft dependencies are missing

alwinw:
#### Reference Issues/PRs

Resolves #325 for `tsfresh`, `pmdarima`, `scikit_posthocs`. If there are more, please let me know as I'm only just starting on `sktime`. Also references #407

* Adds try/except wrappers around soft dependencies. Example:  
```py
# import inside method to avoid hard dependency
try:
    from pmdarima.arima import AutoARIMA as _AutoARIMA
except ModuleNotFoundError as e:
    raise Exception(
        f"{e}, please run `pip install pmdarima` to install pmdarima"
    ) from e
```

* Adds extra_requirements to setup_package()
  * Note again it was only for ones I have been exposed to so far. If there are more, please let me know
  * I took a best guess at what to name the optional features. Please review and let me know
  * To install a soft dependency, run `pip install .[<soft dep, e.g. auto_ARIMA>]`

* General house cleaning with pre-commit on changed files
  * I would suggest running pre-commit over the entire repo, however it may break many open PRs

N/A

##### For all contributions

#### Any other comments?

If this is an accepted solution to #325, I'm happy to update the docs with using `extras_require` in `setup.py`


setup.py
```diff
@@ -23,7 +24,12 @@
     "numpy": "1.18.0",
     "pandas": "1.0.0",
     "scikit-learn": "0.23.0",
-    "statsmodels": "0.12.0"
+    "statsmodels": "0.12.0",
+}
+EXTRAS_REQUIRE = {
```
mloning, line 29: What's the benefit of adding them here as EXTRAS_REQUIRE? I know that you could now run `pip install sktime[series_as_features]`. 

I'm a little wary of specifying the extra requirements in another place, we already need to keep .binder/requirements.txt and build_tools/requirements.txt in sync.

sktime/forecasting/arima.py
```diff
@@ -266,26 +294,52 @@ def __init__(self, start_p=2, d=None, start_q=2, max_p=5,
         super(AutoARIMA, self).__init__()
 
         # import inside method to avoid hard dependency
-        from pmdarima.arima import AutoARIMA as _AutoARIMA
+        try:
```
mloning, line 297: I'd move this to the top of the file, not inside the class, to make it easier to see with the other imports. I'd also write a helper function to do this which we can use in multiple places. I've updated #325 to give an example! 

Checking for importing the main package should also be enough instead of trying all submodules, no? Making sure people install the correct version may be another issue. That's where EXTRA_REQUIRES may come in handy ...

mloning (commented):
Looks good, I've added a few comments! Thanks for working on this! 👍

mloning:
We also use matplotlib and seaborn in `sktime.utils.plotting`, we actually use lazy imports there, but best to put them through the import check function.

setup.py
alwinw: Initially I was going to do `pip install sktime[all_extras]` instead of specifying each individual soft dependency before I realised there were so many. I know it's another place to keep track and sync dependencies, but personally, I think it would make it easier for new users to `sktime`. Rather than having to hunt through the `sktime` module itself to find additional requirements.txt, or run various modules and consistently get `ModuleNotFoundError`s, they can simply run a standard `pip install sktime[all_extras]` command. We could even add `pip install sktime[all_extras]` to the `ModuleNotFoundError` when soft dependencies aren't installed

sktime/forecasting/arima.py
alwinw: Yes, I agree with this. I did see your suggestion of performing a check at the top of files, but didn't realise there'd be so many dependencies scattered throughout sorry. I'll change them to function calls at the top of files.

alwinw:
I've added a `_check_imports` function now and used it in the top of files. Let me know your thoughts

setup.py
mloning: Okay, let's do that! We could go through the CI scripts to use `pip install -e .[all_extras]` there too, I'm happy to look into this if you don't have the time, but would wait for you to finish your work.

setup.py
```diff
@@ -32,69 +42,79 @@
...
     else:
         raise RuntimeError("Unable to find version string.")
 
 
-WEBSITE = 'https://alan-turing-institute.github.io/sktime/'
-DISTNAME = 'sktime'
-DESCRIPTION = 'scikit-learn compatible Python toolbox for machine learning ' \
-              'with time series'
-with codecs.open('README.rst', encoding='utf-8-sig') as f:
+WEBSITE = "https://alan-turing-institute.github.io/sktime/"
```
mloning, line 58: Oh we should have changed that a while ago!

```suggestion
WEBSITE = "https://www.sktime.org"
```

setup.py
```diff
@@ -32,69 +42,79 @@
...
+DISTNAME = "sktime"
+DESCRIPTION = (
+    "scikit-learn compatible Python toolbox for machine learning " "with time series"
```
mloning, line 61: Another quick update here since we're on it:
```suggestion
    "A unified Python toolbox for machine learning with time series"
```

setup.py
mloning: We need to add: numba (sorry, forgot about it, see build_tools/requirements.txt)

sktime/utils/check_imports.py
```diff
@@ -0,0 +1,13 @@
+# -*- coding: utf-8 -*-
+from importlib import import_module
+
+
+def _check_imports(package):
+    try:
+        import_module(package)
+    except ModuleNotFoundError as e:
+        raise Exception(
```
mloning, line 9:
```suggestion
        raise ModuleNotFoundError(
```
I'd prefer to raise a more specific exception and adjust the unit tests correspondingly.

sktime/utils/check_imports.py
```diff
@@ -0,0 +1,13 @@
+# -*- coding: utf-8 -*-
+from importlib import import_module
+
+
+def _check_imports(package):
```
mloning, line 5: Since we have to check multiple package in some modules, how about we accept multiple package names and loop over them? 

```python
def _check_soft_dependencies(*packages):
    for package in packages:
        ...
```

mloning: `check_soft_dependencies` may also be a more meaningful name, would also appreciate a short/informal docstring!

sktime/benchmarking/evaluation.py
```diff
@@ -7,14 +8,18 @@
...
+from scipy.stats import ranksums, ttest_ind
+
 from sktime.benchmarking.base import BaseResults
 from sktime.exceptions import NotEvaluatedError
+from sktime.utils.check_imports import _check_imports
 
 plt.style.use("seaborn-ticks")
 
 
+_check_imports("scikit_posthocs")
```
mloning, line 20: Looks like we're also importing matplolib in line 7 above and use it in line 17, so we need to move the import check a bit further up

mloning (commented):
Thanks @alwinw, excellent work, it's great that you're sorting this out! I left a few minor comments!

alwinw:
Thanks for the feedback @mloning, appreciate it! Happy to help out where I can.  I think I'm touching quite a few files that weren't run through pre-commit or other linters so the changelist is getting quite long.

I'd be up to update the build to `pip install -e .[all_extras]`, but maybe split it out from this PR to make it more manageable. Your thoughts?

sktime/transformers/series_as_features/rocket.py
```diff
@@ -1,16 +1,20 @@
...
+from sktime.transformers.series_as_features.base import BaseSeriesAsFeaturesTransformer
+from sktime.utils.check_imports import _check_soft_deps
 from sktime.utils.data_container import nested_to_3d_numpy
 from sktime.utils.validation.series_as_features import check_X
 
 __author__ = "Angus Dempster"
 __all__ = ["Rocket"]
 
 
+_check_soft_deps("numba")
```
mloning, line 15: We need to make sure to run the checks before importing numba

mloning (commented):
@alwinw yes, let's update the build tools in a separate PR! 

One minor change, everything else looks good to go!

alwinw:
Whoops my mistake! I had `isort` setup in vscode from another project and it automatically moved `from numba import njit, prange` above `from numba import njit, prange`.

It should be good to go now hopefully! Thanks again for your feedback @mloning

mloning (approved):
Yes, looks good, will merge once the status checks pass!

mloning:
Thanks @alwinw great work, really appreciate your contribution!

mloning:
Have you added yourself to our list of contributors?

alwinw:
> Have you added yourself to our list of contributors?

I haven't added myself directly, but I think you added me as a docs contributor when I was fixing some typos in the example notebooks

mloning:
Alright, just updated it to properly reflect the work of this PR! 👍

PR #414, opened 2020-10-03
Global test checking if estimators behave properly for n_jobs > 1. Fixed the n_jobs param handling.

kkoziara:
#### Reference Issues/PRs

References discussion in #408  PR

* Added global test which checks if all eligible estimators behave in a proper way for many parallel jobs.
* Fixed n_jobs param setting for `EnsembleForecaster` and `StackingForecaster` estimators.


sktime/utils/_testing/estimator_checks.py
```diff
@@ -426,3 +429,31 @@ def check_persistence_via_pickle(Estimator):
...
+    if 'n_jobs' in signature(Estimator.__init__).parameters:
+        estimator = _construct_instance(Estimator)
+        fit_args = _make_args(estimator, "fit")
+        ops = ["predict", "predict_proba", "transform"]
+        for op in ops:
+            if hasattr(estimator, op):
+                X_test = _make_args(estimator, op)[0]
+                result_set = []
+                for n_jobs in [1, 4]:
+                    estimator = _construct_instance(Estimator, n_jobs=n_jobs)
```
mloning, line 444: I think I'd prefer to call `estimator.set_params(*{"n_jobs": n_jobs})` here rather than changing the `_construct_instance` method if that works? This way we wouldn't have to reconstruct the instance but could simply set the new params and would have to do all of this only if the estimator has the n_jobs attribute. What do you think?

sktime/utils/_testing/estimator_checks.py
```diff
@@ -426,3 +429,31 @@ def check_persistence_via_pickle(Estimator):
...
+
+
+def check_multiprocessing_determinism(Estimator):
+    if 'n_jobs' in signature(Estimator.__init__).parameters:
+        estimator = _construct_instance(Estimator)
+        fit_args = _make_args(estimator, "fit")
+        ops = ["predict", "predict_proba", "transform"]
+        for op in ops:
+            if hasattr(estimator, op):
+                X_test = _make_args(estimator, op)[0]
```
mloning, line 441: It's not always `X_test`, I'd just call it args (e.g. may be the forecasting horizon for forecasters), just to avoid confusion.

sktime/utils/_testing/estimator_checks.py
```diff
@@ -426,3 +429,31 @@ def check_persistence_via_pickle(Estimator):
...
+                    if hasattr(estimator, 'n_jobs'):
+                        assert estimator.n_jobs == n_jobs
+                    set_random_state(estimator)
+                    estimator.fit(*fit_args)
+                    result_set.append(getattr(estimator, op)(X_test))
+
+                if isinstance(result_set[0], pd.DataFrame):
+                    assert_frame_equal(result_set[0], result_set[1])
+                else:
+                    np.testing.assert_array_almost_equal(
```
mloning, line 454: should that be `assert_array_equal` in line with `assert_frame_equal`?

mloning (commented):
Looks good! Thanks for working on this! Left a few minor comments in the code!

sktime/utils/_testing/estimator_checks.py
kkoziara: Yes, that makes sense. Switched to using `set_params` and removed changes in `_construct_instance`.

sktime/utils/_testing/estimator_checks.py
kkoziara: Renamed this.

sktime/utils/_testing/estimator_checks.py
kkoziara: Yes, you are right. The arrays should be just equal.

sktime/utils/_testing/estimator_checks.py
```diff
@@ -426,3 +429,31 @@ def check_persistence_via_pickle(Estimator):
     for method in results:
         unpickled_result = getattr(unpickled_estimator, method)(*args[method])
         _assert_almost_equal(results[method], unpickled_result)
+
+
+def check_multiprocessing_determinism(Estimator):
+    if 'n_jobs' in signature(Estimator.__init__).parameters:
+        estimator = _construct_instance(Estimator)
+        fit_args = _make_args(estimator, "fit")
+        ops = ["predict", "predict_proba", "transform"]
```
mloning, line 438: One last comment: We have a NON_STATE_CHANGING_METHODS list at the top of the file, so we use that here:

```python
for method in NON_STATE_CHANGING_METHOD:
    ...
```

Sorry, I thought I had posted this comment before but can't find it anymore. Everything else looks good to go!

mloning (approved):

mloning:
This one looks good to go, will merge later once all tests pass!

mloning:
Thanks @kkoziara, great work!

Issue #356, opened 2020-07-24
Interface to DICOM data


TonyBagnall:
**Is your feature request related to a problem? Please describe.**

DICOM for data is standard for storing images in hospitals. However, it can also be used to store "waveform" data such as ECG. Analysing this data with sktime requires expertise in converting and formatting. It would be great to embed tools to transform DICOM time series into ts format.

**Describe the solution you'd like**

there are existing DICOM interfaces in python. These should be explored and hopefully facilitate the translation. 

1) This should involve code to form a single file including time series for a single patient record, where meta information is stored in the .ts file header.  
2) post processing code to form different problem types such as clustering, classification and anomaly detection. 

there is a lot of leg work to make this happen and we will need to define some more meta info in the .ts format. 

The whole issue of data extraction/formatting might be better in a sister package at sktime.org?

zbenmo:
I'm new here, yet happy to jump in. This seems like a useful feature(s) / tool(s) / tutorial. Happy for additional advice from sktime maintainers and for where to obtain DICOM that relate to ECG for example. Believe I can find something in Kaggle  or elsewhere.

PR #415, opened 2020-10-05
example for reading ECG from DICOM

zbenmo:
#### Reference Issues/PRs

Issue #356 

An example is given how to read DICOM waveform and bringing it quickly to a format where one can use sktime.

In order to to run the example (notebook) 'pydicom' is needed.
Also note that a .dcm file was introduced in the example directory (the file was taken from https://github.com/marcodebe/dicom-ecg-plot)

##### For all contributions

##### For new estimators

#### Any other comments?

The original issue suggests to create a .ts file. This was not done. Only the minimum demonstration is included; a DICOM file is loaded and is used in memory quickly with sktime.
The issue also mentions a desire for classification / clustering demonstration, yet my understanding is that this would be unnecessary repeat of other sktime tutorials.

I'm happy to create more relevant and complete code / utilities / examples, yet maybe above is just what needed, as the DICOM format is very flexible and one may be introduced with different settings.

This PR can be either merged as is, or can be considered a WIP. Happy for any guidance.
Maybe makes sense to merge this PR and to open new relevant issue(s) to be implemented (by me) in additional increments.

WKR, Oren.


mloning:
Hi @zbenmo, thanks for the PR! Just some quick initial comments:

I have no experience of working with DICOM data, @TonyBagnall may want to comment here too. 

Ideally, I think we'd like some re-usable `from_dicom_to_3d_numpy` function that allow users to convert DICOM data into 3d numpy arrays with shape = (n_instances, n_variables, n_timepoints) (or alternatively our nested pd.DataFrame structure). 

But I'm not sure how many variations there are of DICOM file formats, I'm open to suggestions. What do you think would make sense here?

zbenmo:
I will definitely look at https://dicom-numpy.readthedocs.io/en/latest/, thank you @fkiraly .

zbenmo:
Thank you @mloning. I'll wait for some advice from @TonyBagnall . I'm new to DICOM. From what I saw you can put there whatever you want. There is some guidance, and common practices, but still a lot of freedom. But I may be wrong. Maybe I do need to add at least one level of sophistication, by indeed showing how to put into 3d numpy arrays with shape = (n_instances, n_variables, n_timepoints). I'll work on that while hoping that Tony will have a look and share his thoughts.

zbenmo:
Have improved some the demo. Still all in a single notebook. Need some advice. Please note that for example,  https://dicom-numpy.readthedocs.io/en/latest/, is about images. here we're trying to load ECG (wave) readings. Also note that there is some device specific logic.

mloning (commented):
Hi @zbenmo, thanks again for working on this. Here are some suggested changes: 

* Add more text/descriptions to the notebook so that users can quickly understand what the notebook is about and what each steps does 
* Is it possible to download the data in the notebook rather than including it in the repository? 
* The plots look strange and are overcrowded because of `plot_series` plotting style, I suggest to either change that, so that `plot_series` doesn't plot the individual data points or simply use matplotlib or seaborn. 

I think it's worth exploring a bit more how DICOM files differ and how we can generalise the data loading functionality. In the end, we'd need some functionality that reads in DICOM files from multiple patients. Hope this helps!

zbenmo:
Believe I've addressed the issues raised above.
Open for any additional tips. 
A bit missing in my mind is to actually use sktime in the notebook, at least in a glance, yet not sure what would be straight-forward there.

mloning:
Hi @zbenmo, yes I agree, at the moment this seems like a nice blog post explaining a DICOM dataset and how to preprocess it to use it with sktime, but not tightly integrated with sktime. I think we need to better understand the different formats DICOM data can take and provide a function that can read most formats and transform them into the required sktime input format (3d numpy or nested data frame). The function could live within `sktime.utils` and then we could have a notebook that shows the complete workflow of time series classification with some DICOM dataset. What do you think?

zbenmo:
My current thoughts are that instead of trying to tame DICOM, we need a generic way to create (3d numpy or nested data frame) in sktime. Generic in the sense, that if you create your data structure using those methods, you're guaranteed that other functionality in sktime will work with it. Then we can show for example, how to take a DICOM file from a specific manufacture, in a specific setting, and push into the expected structure by sktime. This same utilities, can then be used for other source formats, etc. The right place for such functionality seems to be be indeed sktime.util, and may be as trivial as a simple wrapper around numpy/pandas.  Will be happy to join in such an effort. If makes sense.

My current understanding re: DICOM, at least from the single example I was working on, is that there is no "common" DICOM setting. It is very easy to work with DICOM with the 'pydicom' package, numpy, and pandas, but you do need to understand the application / manufacture / device, specifics.

mloning:
Okay, I think the 3d numpy array with shape = (n_instances, n_variables, n_timepoints) is relatively easy for people to understand and use with their own data, no? Much easier than the nested data frame anyway. We have a number of utility functions in `sktime.utils.data_container`. Please take a look and let us know if there's anything important that's missing (I'm sure there is)!

zbenmo:
Have made use of 'sktime.utils.data_container.from_3d_numpy_to_nested'.

PR #422, opened 2020-10-10
Fixed small bug in performance_metrics/forecasting/_functions.py

krumeto:
#### Reference Issues/PRs
There is no currently open issue or PR on this bug
In the mase_loss function, there is a check if the training set is prior to test set. In the latest sktime version, the check was comparing y_train.index.max() >= y_test.min() , rather than y_train.index.max() >= y_test.index.min(). In addition, the ValueError raised a comment which was referring to y_pred, while the check performed was against y_test. Both were straightforward fixes.

No new dependency.

##### For all contributions

##### For new estimators


riteshahuja13 (dismissed):
This is causing the function to break in my use case as well. #426 
With these changes it works well.

mloning:
Thanks for the fix @krumeto and the feedback @riteshahuja13! Synced with master and will merge once all tests pass!

Issue #434, opened 2020-10-18
Restructuring of the utils directory


TonyBagnall:
this directory contains a lot of legacy code that could usefully be tidied up and restructured. This is ongoing work with UEA project students, and we will list changes as we do them, this is an issue simply to help manage it. Suggestions welcome

this is the starting structure
![image](https://user-images.githubusercontent.com/9594042/96365996-a0e4ef00-113c-11eb-8b96-24c11777f279.png)

1. All these files need better commenting, including authorship 
2. files boxcox, seasonality and time_series are all related to forecasting and can be removed or refactored. 
3. check_imports, comparison, confidence are small utility functions, which means they should be here, but possibly better grouped/different file structure
4. load_data and results_writing can be here, but need tidying and refactoring
5. Directory plotting is empty
6. Directory _testing, tests and _testing/tests needs tidying up 
7. Directory validation: thje role of this directory needs clarifying
8. Directory maintenance contains show_versions, which is used in sktime/__init__.py, although the init accesses a protected file

whackteachers:
Hi I am Jason. Got some questions on this.
For 2, I see that there are still usages on them, how do I handle the dependencies on them?
similarly when restructuring the directory, I assume I have to refactor the usages as well?

TonyBagnall:
Hi jason, can we discuss this on teams? I am not expecting you to do this without guidance, the above is for discussion

whackteachers:
sure

mloning:
Thanks @TonyBagnall! 

I agree with most of this! I'm planning to make a new minor release, and then work towards a new major release by adding transformer types (#420) and restructuring this. 

5. Plotting isn't empty but can be made a file rather than a module (there's a function in the __init__)
6. Files/module that start with "test" are recognised by pytest as unit tests. The module "_testing" contains utils for running unit tests, some of which are themselves tested. I had a quick look at the module itself which seems fairly organised I think.
7. contains input checks and is also fairly organised I think

Adding docstrings to the top of the files in addition to authorship would help clarify the purpose. For authorship please also use CODEOWNERS in the root dir.

mloning:
This should now be fixed in the new release https://github.com/alan-turing-institute/sktime/releases/tag/v0.5.0

PR #438, opened 2020-10-19
Added write dataset to ts file functionality

whackteachers:
#### Reference Issues/PRs

This is partly addressed to #434 which is a functionality requested by @TonyBagnall 

I have implmented a new function that can writes a dataframe to a .ts file which is very useful for saving transformed dataset,
along with some unit testing.
N/A

##### For all contributions

##### For new estimators

#### Any other comments?

Just a side note, the function right now does not support timestamp data, as I have not seen an example of it in the dataset nor in the docs. I have also mention this in the doc string.


sktime/utils/validation/tests/test_write_dataframe_to_ts.py
```diff
@@ -0,0 +1,41 @@
+__author__ = ["Jason Pong"]
+__all__ = ["test_write_dataframe_to_ts_success", "test_write_dataframe_to_ts_fail"]
+
+import os
+
+import numpy as np
+import pandas as pd
+import pytest
+from _pytest.tmpdir import tmp_path
```
mloning, line 9: I don't think we need this import, pytest automatically recognises `tmp_path` as a test fixture see https://docs.pytest.org/en/stable/tmpdir.html

sktime/utils/results_writing.py
```diff
@@ -80,6 +85,126 @@ def write_results_to_uea_format(path, strategy_name, dataset_name, y_true,
     file.close()
 
 
+def write_dataframe_to_tsfile(data, path, problem_name,
```
mloning, line 88: Regarding the restructuring of utils, this function should probably go into `data_container` and not `results_writing`, no?

mloning: I'd also set a default argument for `problem_name` so that users don't have to specify it whenever they just want to write out some data

sktime/utils/results_writing.py
```diff
@@ -80,6 +85,126 @@ def write_results_to_uea_format(path, strategy_name, dataset_name, y_true,
...
+        Representation for missing value, default is NaN.
+    comment: {None, str}, optional
+        Comment text to be inserted before the header in a block.
+
+    Returns
+    -------
+    None
+    -------
+    Notes
+    -----
```
mloning, line 133: some superfluous horizontal lines here

mloning (commented):
Looks generally good to me, very clean and readable code, thanks @whackteachers!

sktime/utils/results_writing.py
TonyBagnall: hi markus, I asked him to put it here for now, we are delaying all refactoring until you give us the green light 👍

TonyBagnall:
I think the code quality issue is just some of my code in experiments.py and not relevant to this PR. The travis issue may be caused by the fact that @whackteachers does not have travis installed on his fork? We encountered something similar with tsml.

mloning:
The travis build didn't error, just stalled, I re-started and should pass now. 

To fix the code-quality check, see this post: https://gist.github.com/mloning/6f4b8d412743968af6b2da8b4eaaf2bf

whackteachers:
> The travis build didn't error, just stalled, I re-started and should pass now.
> 
> To fix the code-quality check, see this post: https://gist.github.com/mloning/6f4b8d412743968af6b2da8b4eaaf2bf

I have ran pre-commit before I push and it was fine, also the command given is not working, give smth like --name-only master is not a valid parameter error

mloning:
Okay happy to ignore this, since it's only on the contrib/experiments.py file

TonyBagnall (dismissed):
looks good to me

mloning:
Hi @whackteachers, I just merged another big PR which created a merge conflict here, once you resolve it, we can merge this PR too. Let me know if you need help with that!

sktime/utils/results_writing.py
```diff
@@ -80,6 +109,144 @@ def write_results_to_uea_format(path, strategy_name, dataset_name, y_true,
     file.close()
 
 
+def write_dataframe_to_tsfile(
+    data,
+    path,
```
mloning, line 114: Does path not include the file name? Below it says that `problem_name` gives the file name. So does this give the directory?

sktime/utils/results_writing.py
```diff
@@ -80,6 +109,144 @@ def write_results_to_uea_format(path, strategy_name, dataset_name, y_true,
...
+    References
+    ----------
+    The code for writing series data into file is adopted from
+    https://stackoverflow.com/questions/37877708/
+    how-to-turn-a-pandas-dataframe-row-into-a-comma-separated-string
+    """
+    if class_value_list is None:
+        class_value_list = []
+    # ensure data provided is a dataframe
+    if not isinstance(data, pd.DataFrame):
```
mloning, line 180: Could we use the input validation functions from `sktime.utils.validation.panel`, i.e. `check_X` with `coerce_to_pandas=True` so that this also works when passing a 3d numpy array?

sktime/utils/validation/tests/test_write_dataframe_to_ts.py
```diff
@@ -0,0 +1,56 @@
...
+        comment="""
+          The data was derived from twelve monthly electrical power demand
+          time series from Italy and first used in the paper "Intelligent
+          Icons: Integrating Lite-Weight Data Mining and Visualization into
+          GUI Operating Systems". The classification task is to distinguish
+          days from Oct to March (inclusive) from April to September.
+        """,
+    )
+    # load data back from the ts file
+    result = f"{tmp_path}/ItalyPowerDemand/ItalyPowerDemand_transform.ts"
```
mloning, line 42: Why is this called `ItalyPowerDemand_transform.ts`, where does the `transform` come from?

mloning (commented):
Three minor comments, everything looks good!

sktime/utils/results_writing.py
whackteachers: no, path just refers to the directory to store the file, while problem_name is used to name the ts file at the same time.

sktime/utils/results_writing.py
whackteachers: yeah sure, I wasn't aware there is already such module exist

sktime/utils/validation/tests/test_write_dataframe_to_ts.py
whackteachers: the write_dataframe_to_tsfile appends _transformat the end of the problem name as the file name, see
https://github.com/whackteachers/sktime/blob/3591a48577a7cda6fb2b5fffa8cdf1f6597107c2/sktime/utils/results_writing.py#L194-L201

TonyBagnall: hi, this is fine, but fyi general format for storing results we use is 
<problem name>/<problem_name>.ts
If I were doing a transform on multiple data I would either have a whole directory and keep the original name
<transform name>/<problem name>/<problem_name>.ts
or append the transform name to all
<problem name_transform name>/<problem_name_transform name>.ts

TonyBagnall (approved):

sktime/utils/validation/tests/test_write_dataframe_to_ts.py
whackteachers: ok thanks for the reminder, I will keep this in mind next time

sktime/utils/results_writing.py
mloning: Okay I think this is worth clarifying in the doc strings, usually input args are data and path including file name (e.g. `np.savetxt` or `pd.to_csv`)

sktime/utils/validation/tests/test_write_dataframe_to_ts.py
mloning: from a user perspective, I find it strange if I give a path (directory) and problem name but then can't find the generated file, no? would probably prefer to have a user friendly function which takes data and file (path with file name, e.g. "my_data.ts") and a wrapper for your benchmarking purposes

TonyBagnall: one reason for this is that we need to organise multiple problems, each of which has 6 or more files associated with it. These may be predefined splits into train and test or resamples. This structure allows us to automate much of the internal data complexity without bothering the user. For example, if you load a problem it looks to detect if there is already an predefined train test split, and it also looks to see if the resamples are already generated.

mloning: yes I think that makes sense, but can that not be handled by a wrapper which defines additional behaviour for your purposes, so that we can still have a more generic function for writing a data frame to a ts file for users, i.e. the counterpart to `load_from_tsfile_to_dataframe`

TonyBagnall: Ultimately, sure, but that's beyond the scope of this PR

mloning:
Alright, agree, let's merge this PR first and work on the more generic function in a separate PR. 

One of the travis build stalled again, just restarted it now and will merge later today once it passes. 

Thanks again @whackteachers - great work!

PR #446, opened 2020-10-24
Fix soft dependency imports

mloning:
- [x] added a test to check if soft dependencies are properly isolated 
- [x] made `numba` a hard dependency
- [x] fix soft dependency issue where soft dependencies were not properly isolated and instead require to load unrelated modules
- [x] minor fixes

We need to organise our dependencies a bit better, currently they are specified in multiple locations (.binder/requirements.txt, setup.py, build_tools/requirements.txt, etc).


TonyBagnall (approved):
looks good to me

PR #317, opened 2020-06-23
[ON HOLD] Implementation of the catch22 transformer

MatthewMiddlehurst:
ON HOLD DUE TO VARYING RESULTS ACROSS OPERATING SYSTEMS (see discussion below for details)

#### Reference Issues/PRs

Resolves #129
Work towards #294

#### What does this implement/fix? Explain your changes.

Implements the catch22 set of features as a transformer. Additionally, ports over the catch22 forest classifier put together for sktime (https://github.com/chlubba/sktime-catch22).

Replaces the previous dictionary based example notebook dataset with ItalyPowerDemand (a better dataset for the problem anyway).

#### Does your contribution introduce a new dependency? If yes, which one? 

Requires the catch22 project by the original authors containing C code wrapped for Python3 (https://github.com/chlubba/catch22).
Thanks to Carl Lubba for his work making the package Windows compatible so it could be included in sktime.

#### PR checklist for new estimators 

- [x] I have added unit tests and made sure they pass locally. 
- [x] I have updated the existing example notebooks or provided a new one to showcase how my algorithm works.
- [x] I have updated sktime's [estimator overview](https://github.com/alan-turing-institute/sktime/blob/master/ESTIMATOR_OVERVIEW.md) and I'm committed to maintain my contribution and provide bug fixes if necessary.

PR #453, opened 2020-10-28
Implementation of catch22 transformer, CIF classifier and dictionary based clean-up

MatthewMiddlehurst:
(See PR #317 for previous catch22 discussion)

Separate cBOSS from BOSS. Small clean-up of BOSS, TDE and related documentation/examples/tests.

Implementation of the catch22 transformer and Canonical Interval Forest classifier.  ~~The Windows OS issue from PR #317 has still not been resolved, as such both will raise an OSError if called in Windows. Tests for both will be skipped on Windows also, though im unsure if more will be needed for the general estimator tests.~~ 
Edit: I think Linux may be the broken one with Windows being more faithful actually. The original was tested on MacOS, with neither of the other OS fully replicating it. Putting all this stuff in contrib for now with disclaimers, very messy.

Documentation for the above and interval based classifiers.

##### For all contributions

##### For new estimators


mloning:
Hi @MatthewMiddlehurst, re linux see (#456), I'm working on it! 

How did you end up skipping the common estimator checks on Windows for catch22?

MatthewMiddlehurst:
Moved it to contrib, where I assume it doesn’t even get checked. 

I have updated the original PR text, but with differences from multiple OS I think it is a bit too unstable for anywhere but contrib.

mloning (commented):
It looks like you forgot about an empty folder for `classification/hybrid` when you moved everything to `contrib` - everything else looks good!

mloning (approved):

TonyBagnall (approved):
looks good

PR #469, opened 2020-11-05
Bump pre-commit versions, run again on notebooks

MarcoGorelli:
Hey :wave:

couple of things:
1. there are new versions available for a couple of hooks, so I've bumped their version numbers
2. it looks like a couple of notebooks got left out from the previous round of formatting, so I've included them here


mloning (approved):
Alright, great, looks good to me

MarcoGorelli:
Also, there's a number of files which don't get changed when running `pre-commit run --all-files`, I think they're being missed by only checking `pre-commit` on changed files (just flagging this up in case you weren't aware, but if that's expected then sorry for the noise :smile: )

mloning:
Yes, I should have just run pre-commit on all files when I implemented it instead of running it only on the changed files in each PR. But by now most files should have been through pre-commit I hope.

mloning (approved):
Looks good to me, thanks for keeping this up-to-date @MarcoGorelli!

PR #475, opened 2020-11-09
update basic motion description

vollmersj:
fix in doc string - before was a copy paste error


mloning (approved):
Thanks @vollmersj, also noticed the datasets weren't in the online docs, so added them here too! Will merge once all tests pass.

PR #479, opened 2020-11-12
Add parallelization for training the `RandomIntervalSpectralForest`

AaronX121:

The training stage of `RandomIntervalSpectralForest` is quite slow because decision trees are fitted sequentially within a loop.

This PR uses `joblib.Parallel` and `joblib.delayed` to parallelize the training stage of `RandomIntervalSpectralForest`. Following the routine in `_forest.py` from `scikit-learn`, a private function named `_parallel_build_trees` is added to support parallelization. [Reference](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/ensemble/_forest.py#L140).

This PR also includes improvements on the docstrings for better readability and style consistency.

Changes:
In my personal experiment, by setting `n_jobs` to `-1` (i.e., using all processors), fitting a `RandomIntervalSpectralForest` with `100` decision trees on `Adiac` dataset is able to speed up the training time by over 6 times on a CPU with 10 physical cores, and the results with the random state fixed are exactly the same.

Since `joblib` is required by `scikit-learn`, and `scikit-learn` is required by `sktime`, the answer is No ;-)

##### For all contributions

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -606,6 +606,7 @@
       "avatar_url": "https://avatars2.githubusercontent.com/u/22359569?v=4",
       "profile": "https://github.com/AaronX121",
       "contributions": [
+        "code",
         "doc"
       ]
     },
--- a/sktime/classification/frequency_based/_rise.py
+++ b/sktime/classification/frequency_based/_rise.py
@@ -2,12 +2,13 @@
 """ Random Interval Spectral Forest (RISE).
 Implementation of Deng's Time Series Forest, with minor changes
 """
-__author__ = ["Tony Bagnall"]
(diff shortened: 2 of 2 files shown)


AaronX121:
oops, sorry for some coding style problems : (

mloning (commented):
Hi @AaronX121 - this is great work! 

I like how you factored out the parallisable code in `fit`, could you do the same in `predict`?  We have done this already for `TimeSeriesForest` which has a very similar algorithm structure, take a look [here](https://github.com/alan-turing-institute/sktime/blob/master/sktime/classification/interval_based/_tsf.py).

AaronX121:
That would be very nice. I will close this PR, and open another one with parallelization on `predict_proba` latter.

PR #482, opened 2020-11-13
Add parallelization for `RandomIntervalSpectralForest`

AaronX121:
#### Reference Issues/PRs

This PR is an extended version of #479.

In addition to the parallelization for the training stage implemented in #479, this PR further adds the parallelization for the evaluating stage in `predict_proba`.

Below are some results on the time efficiency improvements brought by parallelization, with each cell corresponding to results before/after parallelization.
| Dataset Name | Test Acc (%) |  Training (s) |   Testing (s)   |
|:------------:|:------------:|:-------------:|:---------------:|
|     Adiac    |  75.19/75.19 | 66.278/10.595 |   63.555/8.768  |
|     Beef     |  76.67/76.67 |  7.374/1.041  |   7.315/1.038   |
|      Car     |  80.00/80.00 |  15.305/2.122 |   15.167/2.106  |
|      CBF     |  93.89/93.89 |  3.698/0.537  |  106.791/15.221 |
|   ChlorineConcentration|  76.93/76.93 |  71.050/9.793 | 559.179/209.677 |

No.

##### For all contributions


mloning (commented):
Hi @AaronX121 - great work, thanks! 

There's still some duplicate code for the transformation procedure.
```python
    acf_x = np.empty(shape=(n_instances, lag))
     ps_len = (interval[1] - interval[0]) / 2
     ps_x = np.empty(shape=(n_instances, int(ps_len)))
     for j in range(0, n_instances):
         acf_x[j] = acf(X[j, interval[0] : interval[1]], lag)
         ps_x[j] = ps(X[j, interval[0] : interval[1]])
     transformed_x = np.concatenate((acf_x, ps_x), axis=1)
```

Could this into a function? Take a look at how we've done it for time series forest [here](https://github.com/alan-turing-institute/sktime/blob/master/sktime/classification/interval_based/_tsf.py). 

Apart from that, it's very well written, thanks again!

AaronX121:
Got it, I will fix this : )

AaronX121:
Hi @mloning, I have updated the code as you suggested. There are also some modifications on the import statements and docstrings for the better consistency of coding style.

sktime/classification/frequency_based/_rise.py
```diff
@@ -202,13 +204,13 @@ def fit(self, X, y):
             delayed(_parallel_build_trees)(
                 X,
                 y,
-                self.acf_lag_,
+                self.base_estimator,
```
mloning, line 207: Can we pass a cloned the estimator here? I mean `clone(self.base_estimator)` or we clone all of them beforehand `self.estimators_ = [clone(self.base_estimator) for _ in range(self.n_estimators)]` and then index into them `self.estimators_[i]` in the loop

mloning (commented):
Excellent! 

One last thing if you have time: it would be great to add a unit test similar to the one [here](https://github.com/alan-turing-institute/sktime/blob/master/sktime/classification/interval_based/tests/test_tsf.py) where the expected result comes from the "unchanged" estimator code (on master). This way we can test that we haven't broken anything and add a unit tests in case we want to make further changes in the future. 

Thanks again for the good work!

FYI: there will be a dev sprint tonight as part of the PyData Global: https://github.com/alan-turing-institute/sktime/wiki/PyData-Global-2020-sprint in case you're interested

sktime/classification/frequency_based/_rise.py
AaronX121: Let's follow your suggestion👍 , it looks like cloning all trees before the parallelization is also the convention in scikit-learn [Here](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/ensemble/_forest.py#L377).

AaronX121:
It looks like the test module for `RandomIntervalSpectralForest` is missing on the master branch. I will take time to add a unit test to make it better.

AaronX121:
Hi @mloning, I have followed your suggestion on cloning all base estimators before parallelization. As to the unit test, the expected result in `test_rise.py` is computed using sktime with stable version `0.4.3`.

No idea on why building the docs fails, maybe there are too many on-going tasks : (

sktime/classification/frequency_based/_rise.py
```diff
@@ -80,36 +130,56 @@ class RandomIntervalSpectralForest(ForestClassifier, BaseClassifier):
...
         self.min_interval = min_interval
         self.acf_lag = acf_lag
         self.acf_min_values = acf_min_values
+        self.n_jobs = n_jobs
+        self.random_state = random_state
 
         # We need to add is-fitted state when inheriting from scikit-learn
         self._is_fitted = False
 
+    def _make_estimator(self, append=True, random_state=None):
```
mloning, line 153: I'm not sure if we need to add a method for this, does this one come from scikit-learn? They may need more complex functionality in their case because they support warm-starts. 

But in our case a simple function should be enough, as far as I can see, no? 

```python
def _make_estimator(base_estimator, random_state=None):
    estimator = clone(estimator)
    estimator.set_params(**{"random_state": random_state})
    return estimator 
```

And then in `fit`, we can call:
```python
self.estimators_ = [_make_estimator(self.base_estimators, rng.randint(np.iinfo(np.int32).max) for _ in range(self.n_estimators)]
```

I think functions are a lot easier to understand than methods. We may also want to re-use that function in some other classes, I had a related discussion with someone who's working on other algorithms, having it as a stand-alone function makes it easy to re-use it!

sktime/classification/frequency_based/_rise.py
```diff
@@ -133,118 +203,128 @@ def fit(self, X, y):
...
-                )
-                ps_x[j] = ps(X[j, self.intervals[i][0] : self.intervals[i][1]])
-            transformed_x = np.concatenate((acf_x, ps_x), axis=1)
-            #            transformed_x=acf_x
-            tree = clone(self.base_estimator)
-            # set random state, but not the same, so that estimators vary
-            tree.set_params(**{"random_state": rng.randint(np.iinfo(np.int32).max)})
-            tree.fit(transformed_x, y)
+
+        trees = [
```
mloning, line 215: see comments above

sktime/classification/frequency_based/_rise.py
```diff
@@ -133,118 +203,128 @@ def fit(self, X, y):
...
+            )
+            for i in range(self.n_estimators)
+        ]
+
+        # Parallel loop
+        worker_rets = Parallel(n_jobs=self.n_jobs)(
+            delayed(_parallel_build_trees)(
+                X,
+                y,
+                t,
```
mloning, line 227: Please replace `t` with `tree` - much easier to understand with more meaningful variable names :)

mloning (commented):
Looks good, I think everything we need is there now! 

I've left a few minor comments in the code, let me know what you think about them! If you agree, we can merge once you implement the suggested changes, let me know if you need help with anything!

sktime/classification/frequency_based/_rise.py
AaronX121: Thanks! You are right, the parameter `append` is not necessary for the current implementation of `RandomIntervalSpectralForest`. Let's remove it, and keep the code as succinct as possible.

AaronX121:
Hi @mloning, thanks for your suggestions! I have isolated `_make_estimator` from the body of class.

What I am a little confused about is your code snippet: `self.estimators_ = [_make_estimator(self.base_estimators, rng.randint(np.iinfo(np.int32).max) for _ in range(self.n_estimators)]`:

If we store a total number of `self.n_estimators` **unfitted** trees in `self.estimators_` before the parallelization, we will have to remove these unfitted trees after the parallelization, because the **fitted** tree returned by each worker in joblib will not be stored in its original location (i.e., `self.estimators_`), but in a standalone memory space. For a fask check, if we do nothing after the parallelization, the program raises `NotFittedError` for decision tree, indicating that trees returned by joblib are not those in `self.estimators_`.

 Therefore, an auxiliary list `trees` is used before the parallelization, and we manually append each **fitted** tree onto `self.estimators_` after the parallelization.

mloning (approved):
@AaronX121 you're right about that, I didn't think of that! 

Thanks for simplifying the `_make_estimators` function, everything looks good now! 

I'll restart the Travis CI job (errored for some unknown reason) and will merge once it passes!

AaronX121:
Thanks @mloning, All checks have passed now ; )

AaronX121:
By the way, how about improving the time efficiency of other algorithms implemented in `sktime` @mloning ?

This great lib has helped me a lot during my own research, and I am very willing to help to improve it :)

mloning:
@AaronX121 I'm happy to hear that you want to contribute more! Feel free to take a look at some of the other algorithms to see if you can add improvements, I'd also be more than happy to discuss other areas to work on, ping me on Gitter and we can discuss what you'd be interested in 👍

PR #485, opened 2020-11-13
Add HCrystalBall wrapper

MichalChromcak:

Brings the possibility to use [HCrystalBall wrappers](https://github.com/heidelbergcement/hcrystalball) as time-series forecasters (fbprophet, theta, ensembles, pmdarima, sklearn, statsmodels, tbats, ...)
hcrystalball>=0.1.8

##### For all contributions

##### For new estimators


MichalChromcak:
_Note:_ An idea of interfacing a HolidayTransformer or other ones (Seasonality, ...) came up in the middle of discussions - selected to be postponed now, but might have a later look.
[HolidayTransformer](https://github.com/heidelbergcement/hcrystalball/blob/master/src/hcrystalball/feature_extraction/_holiday_transformer.py) (uses https://pypi.org/project/workalendar/ inside)

MichalChromcak:
@mloning Can you check the factored out part now?
Still more tests to be written and make a nicer showcase notebook (incorporate to 01_forecasting or create a separate one?)

mloning:
I think it looks much better now, much more readable! Regarding the notebooks, I'd prefer a separate one, something like `forecasting_with_hcrystalball.ipynb`. 

I'll review more thoroughly over the next few days!

mloning:
CI tests are failing because you need to add the `hcrystalball` dependency to the `build_tools/requirements.txt` (check the dependency section in the CONTRIBUTING.md)

MichalChromcak:
@mloning any idea how to solve a missing version of HCrystalBall in CI pipelines? 
[Here](https://dev.azure.com/mloning/sktime/_build/results?buildId=1468&view=logs&j=26d7f9dc-5e93-57c4-b0e5-b9e4f7c61a28&t=bb047f14-6077-515c-c7fb-f6394be0c469&l=59) I see only old versions of hcrystalball (yet the current release is 0.1.8) ([pypi](https://pypi.org/project/hcrystalball/#history)). Side note we support python>=3.7.

MichalChromcak:
@mloning Where would you see the best fit for HCrystalballForecaster in the [online documentation of estimators](https://github.com/alan-turing-institute/sktime/blob/master/ESTIMATOR_OVERVIEW.md)?

MichalChromcak:
@mloning There is one fix I need to do in HCrystalBall, but otherwise test, that still keep failing, are caused by not supported Int based index, which is something we do not plan to support. What would be the best way of dealing with these failures?

mloning:
Hi @MichalChromcak, I had a quick look, everything looks quite nice now I think. I need to look into the int index issue a bit more in detail to see how we can fix that. Thanks for working on this! 👍

mloning:
@MichalChromcak not really sure what's happening with the CI now: 

> ERROR: Could not find a version that satisfies the requirement hcrystalball>=0.1.8 (from -r build_tools/requirements.txt (line 3)) (from versions: 0.1a0, 0.1a12, 0.1a13, 0.1.0, 0.1.1, 0.1.2)

It does seem to work locally. Have you seen this before? Otherwise I'll play around to debug.

MichalChromcak:
@mloning I had similar problem right after the release was created even locally, but thought it was just solved by time. Also is there any chance it refers to python3.6 environment? Wouldn't have a specific recommendation on how to solve it.

mloning:
Ah yes - forgot about the Python 3.6 issue - I think that's what's causing it. What was the reason again for not supporting version 3.6? 

So we need to figure out how to

- [ ] skip `hcrystalball` for Python 3.6 (that one may be tricky without changing the CI scripts) 
- [ ] skip the unit tests that use int indices 

I'll try to look into it.

MichalChromcak:
@mloning I added support for python 3.6 and created v0.1.9 release (might take a while until it's out there). You might try against it maybe....?

mloning:
Thanks @MichalChromcak will try again later today!

mloning:
Okay, tests now run and we hit the `NotImplementedError` for int indices as expected, will try to fix this over the next couple of days.

mloning:
Hi @MichalChromcak - I've changed the default time index type we use for unit testing to `pd.Datetime` but annoyingly we now run into a few other errors (some are related to in-sample predictions), I had a look but wasn't really sure how to best fix this. To make things a bit simpler perhaps, I think it's fine if we only support date time indices and leave it to the user to get it into the required format.

MichalChromcak:
@mloning I agree. I might try to have a look next week into the issues if you won't find the time/solution - what is your plan?

mloning:
@MichalChromcak okay sounds good, I'll take another look over the next few days. How do you handle in-sample predictions in hcrystalball?

mloning:
Hi @MichalChromcak - so this is a minimal working version (no support for in-sample predictions, prediction intervals, or index types other than date-time). The code is however a lot simpler. 

I'd suggest we merge this for now and then add more support when needed. What do you think?

docs/requirements.txt
```diff
@@ -1,3 +1,4 @@
+hcrystalball>=0.1.8
```
MichalChromcak, line 1: @mloning should we have also here 0.1.9?

MichalChromcak:
Hi @mloning, yes happy to merge it

mloning (approved):

mloning:
thanks again @MichalChromcak!

MichalChromcak:
> thanks again @MichalChromcak!

Big thanks to you @mloning, especially for taking over and finishing it up.

Issue #141, opened 2019-09-04
Benchmarking design & implementation 


mloning:
## Evaluation 
- [ ]  Extend evaluation API and functionality to single dataset case, currently only multiple dataset case is supported

## Performance metrics 
A few implementation notes:
   - Vectorised vs iterative computations
   - Callable classes vs classes with methods for computation
   - Use of jackknife by default for non point-wise metrics
   - Computation of standard error as decorator/mix-in
   - Have separate classes for point-wise metrics which can be wrapped by aggregation functions (e.g. mean)

Also see https://github.com/JuliaML/LossFunctions.jl.

### Should have
- [ ] Allow orchestrator to be persisted to replicate benchmarking studies
- [ ] add unit tests for `evaluator` methods
- [ ] update all methods on evaluator to work on new internal data representation, also see  https://www.statsmodels.org/stable/stats.html for some additional test implementations, e.g. the sign test, to improve readability, so that we can deprecate `_get_metrics_per_estimator_dataset` and `_get_metrics_per_estimator` methods
- [ ] for saving results inside the `orchestrator` and for loading results in results classes use `_ResultsWrapper` to simply/unify interface, `_ResultsWrapper` needs to have slots for at least: y_true, y_pred, y_proba, index, fit_time, predict_time, strategy_name, dataset_name, cv_fold, train_or_test
- [ ]  No timing of fit and predict available, see https://docs.python.org/3/library/time.html#time.perf_counter, potentially have new `save_timings` and `load_timings` method
- [ ] `orchestrator` cannot make probabilistic predictions, orchestrator tries to make probabilistic predictions using `predict_proba`, but (i) this will only works for some but not all classifiers and it won't work in regression, (ii) strategies currently don't even have a `predict_proba` (not even `TSCStrategy`), and (iii) current computation of `y_proba` fails if `y_pred` contains strings instead of integers which however is an accepted output format for classification I believe, add `predict_proba` to `TSCStrategy`
- [ ] handling of probabilistic metrics in `evaluator`
- [ ] no longer sure that saving results object as a master file is a good idea, as it may cause problems when multiple processes try to update it and because it needs to reflect the state of the directory somehow, maybe better to have a method on results object that allow to infer datasets, strategies and so on, something like a `register_results` method, instead of loading a fully specified dumped result object 
- [ ] separate `predict` method on `orchestrator` which loads and uses already fitted strategies
- [ ] fix UEA results class  

### Could have
[...]

PR #492, opened 2020-11-16
Benchmarking issue141

ViktorKaz:
#### Reference Issues/PRs
Reference Issues/PRs
Initial PR dealing with issue #141.

* added additional unit tests for evaluator
* introduced functionality for calculating the runtime for fit and predict methods
* added some docstrings

No

##### For all contributions


sktime/benchmarking/base.py
```diff
@@ -182,7 +183,10 @@ def __init__(self, strategy_name, dataset_name, index, y_true, y_pred,
         self.y_true = y_true
         self.y_pred = y_pred
         self.y_proba = y_proba
-
+        self.fit_estimator_start_time = fit_estimator_start_time
```
mloning, line 186: saving the fit and predict time is probably enough, or are the absolute start and end time points interesting in some cases?

ViktorKaz: For my specific use case I don't currently need the absolute start and end time points. However, my thinking was that:

1. We have the data so why not save it in case it might be useful in the future. It is not a big drag on resources, especially if results are saved in RAM. If results are saved on HDD, my current implementation is not ideal as it saves an entire column of the same copy of the values which is wasteful. However, I have not found a better way to persist the data but will investigate further.
2. The design is such that the orchestrator saves the data and the data is subsequently processed by the evaluator. It is true that taking the difference of two numbers is not a very complicated calculation so it is totally possible to do this in the orchestrator but this still deviates a little bit from the overall design logic.

mloning (approved):
Thanks @ViktorKaz for cleaning this up! Looks good to me!

PR #500, opened 2020-11-20
added mape_loss and cosmetic fixes to notebooks (removed kernel)

### LINKED ISSUE #499: MAPE loss function implementation
A need for MAPE loss function to be able to compare forecasting results.

MAPE loss function compatible with sktime

None

**Additional context**
NA

tch:
#### Reference Issues/PRs
Fixes #499.

Implements MAPE loss function.

No new dependencies introduced

##### For all contributions

#### Any other comments?
Another option of implementing this solution would be to re-use sk-learn implementation via simple wrapper. But its small enough that perhaps its better to keep it as a separate implementation.


sktime/performance_metrics/tests/test_performance_metrics_forecasting.py
```diff
@@ -0,0 +1,65 @@
...
+import pytest
+
+from sktime.performance_metrics.forecasting import mape_loss
+from sktime.performance_metrics.forecasting import smape_loss
+
+from sktime.performance_metrics.tests._config import TEST_YS, TEST_YS_ZERO
+
+
+@pytest.mark.parametrize("test_y", TEST_YS_ZERO)
+def test_all_losses_near_zero(test_y):
```
mloning, line 16: I'd either rename the function to `test_mape_near_zero` or add the other losses to the function.

mloning (commented):
Hi @tch - thanks for the PR, looks all good to me! Left a few minor inline comments.

Will leave it open for a bit longer to give others a chance to take a look!

sktime/performance_metrics/tests/test_performance_metrics_forecasting.py
tch: Thanks Markus, I have changed the name to test_mape_loss_near_zero but will revisit it to introduce near zero checks for all losses as this would be helpful. However, there is small discrepancy between the way sMAPE/MASE and MAPE handle such situation. The former return nan when encountering 0 denominator while MAPE returns large value (that comes from division by a small value - machine epsilon). The second approach seem pragmatic and taken by scikit learn as well but happy to change that behaviour, either way, to be consistent. What the view here?

mloning: I haven't thought much about this, but following scikit-learn seems sensible.

tch: OK. Lets get this PR in and I will do next one with proposed changes to sMAPE (handle special case where y_test and y_pred are both 0 -> return 0 instead of nan) and MASE (special case where reference, in-sample mae is 0 -> return relatively large error instead of nan) this should make it consistent and follow the approach in scikit learn

mloning (approved):
I'm happy with the changes, we're currently having some issue with macOS builds because travis changed their CI plans, I'll try to fix this as soon as possible and leave this PR open for a bit longer in case others want to comment - but looks good to go to me.

mloning:
FYI I fixed the CI issues, should be good to go once all tests pass!

mloning:
Thanks again @tch - great stuff!

Issue #503, opened 2020-11-21
Add FbProphet wrapper


robomotic:
Hi there,
it would be nice to add a wrapper for FbProphet (maybe as an optional dependency during pip install) that satisfies the SK interface in this way I will be able to compare it with other models in a uniform manner.

Cheers.

mloning:
Hi @robomotic, yes, that'd be great! It's already on our forecasting Wishlist #220. 

Any help is much appreciated. 

There is also an interesting technical solution to handle these wrapper which one could try out, more details [here](https://gist.github.com/mloning/34aa0eb84cb14bedb9f878d446d51e5a).

aiwalter:
I am working on this and will open a PR next days. The problem will be the dependency of `fbprophet` to `pystan` as the latter requires a C++ compiler. I am using Windows and already faced problems when installing `pystan`, so lets see how this will work :)

robomotic:
That is great news, thanks for efforts and yes I know pystan is a challenge in terms of multiplatform deployments.

mloning:
We usually handles this via soft dependencies, see for example the [`tsfresh`](https://github.com/alan-turing-institute/sktime/blob/master/sktime/transformers/panel/tsfresh.py) file. That way, sktime as a whole does not depend on it and only user who want to work with fbprophet have to install it.

aiwalter:
@mloning great, thx for the hint. I will do it like that. 
Another question: `fbprophet` can also work with exogenous data (multivariate), should this also be considered? Afaik `sktime` provides currently only univariate time series regression algorithms?

mloning:
Exogenous variables should work, some algorithms support it already e.g. `AutoARIMA`, the interface is also defined in the BaseForecaster

aiwalter:
@mloning I have created a class `Prophet` with a `fit()` function. I am a bit stuck in writing the `predict()` function as I see you are using a couple of functions e.g. in `arima.py`. In this file, there is no docstring for some functions, so I was not 100% sure if I got everything.
Here my questions:

- what is `_predict_fixed_cutoff()` for?
- Is it mandatory to do a recursion in `_predict_in_sample()`, in `fbprophet` I can do it without. Is this ok?
- which functions are mandatory for the class `Prophet` to be either provided or inherited?

Thx, Martin

mloning:
[...]

PR #515, opened 2020-12-01
Add fbprophet wrapper

aiwalter:
#### Reference Issues/PRs

Issue: Enhancement as in #503 

Added wrapper for fbprophet. Works for univariate as well as multivariate (exog)

Soft dependency to `fbprophet`

##### For all contributions

##### For new estimators


mloning:
Thanks for the PR! 

To fix the code-style errors, check out the pre-commit workflow described in CONTRIBUTING.md. To apply it retrospectively, read this post: https://gist.github.com/mloning/6f4b8d412743968af6b2da8b4eaaf2bf

mloning:
Thanks again, there's a lot of good work in this PR! - I'll add a more thorough review over the next few days.

aiwalter:
@mloning great, I would appreciate detailed feedback and I am also open to improve/change things if required. I was doing `pre-commit` but I was facing an error when commiting as it was not finding my Python. I had to use Conda for this PR in order to get `fbprophet` running with C++ on Windows. It seems there are maybe issues with `pre-commit`  and Conda.

mloning:
Hi, I'm running pre-commit with conda, all I have to do is make sure I activate my environment first before committing. We can jump into a voice channel on discord if you're around to discuss any questions.

aiwalter:
@mloning ok how can I join?

mloning:
Invite link: https://discord.com/invite/gqSab2K

sktime/forecasting/base/_base.py
```diff
@@ -184,3 +186,18 @@ def get_fitted_params(self):
         fitted_params : dict
         """
         raise NotImplementedError("abstract method")
+
+    def _get_model_params(self):
```
mloning, line 190: Why do we need this? Isn't this already handled through sklearn's hyper-parameter interface via `get_params`?

sktime/forecasting/prophet.py
```diff
@@ -0,0 +1,341 @@
...
+    ----------
+    https://facebook.github.io/prophet
+    https://github.com/facebook/prophet
+
+    """
+
+    def __init__(
+        self,
+        # Args due to wrapping
+        freq,
```
mloning, line 95: `freq` should have a sensible default argument. If `freq` refers to the seasonal periodicity, it should be renamed to `sp` with a default value of `1` (no seasonality) - at least that's what we use elsewhere.

sktime/forecasting/prophet.py
```diff
@@ -0,0 +1,341 @@
...
+        holidays=None,
+        seasonality_mode='additive',
+        seasonality_prior_scale=10.0,
+        holidays_prior_scale=10.0,
+        changepoint_prior_scale=0.05,
+        mcmc_samples=0,
+        interval_width=1 - DEFAULT_ALPHA,
+        uncertainty_samples=1000,
+        stan_backend=None,
+        **kwargs
```
mloning, line 115: What are those additional kwargs?

sktime/forecasting/prophet.py
```diff
@@ -0,0 +1,341 @@
...
+            seasonality_mode=seasonality_mode,
+            seasonality_prior_scale=seasonality_prior_scale,
+            holidays_prior_scale=holidays_prior_scale,
+            changepoint_prior_scale=changepoint_prior_scale,
+            mcmc_samples=mcmc_samples,
+            interval_width=interval_width,
+            uncertainty_samples=uncertainty_samples,
+            stan_backend=stan_backend,
+            **kwargs
+        )
```
mloning, line 171: I think I'd move this to the beginning of `fit` and pass on the values from `self`, otherwise values may not be passed on correctly if I use `set_params()` instead of re-instantiating the object.

sktime/forecasting/prophet.py
```diff
@@ -0,0 +1,341 @@
...
+        stan_backend=None,
+        **kwargs
+    ):
+        self.freq = freq
+        self.add_seasonality = add_seasonality
+        self.add_country_holidays = add_country_holidays
+
+        self.growth = growth
+        self.changepoints = changepoints
+        if self.changepoints is not None:
```
mloning, line 123: I'd wrap these checks into a function - something like `_check_changepoints` - call it in fit, and assign the return value to self.

sktime/forecasting/prophet.py
```diff
@@ -0,0 +1,341 @@
...
+            Returns series of predicted values.
+
+        Raises
+        ------
+        Exception
+            Error when merging data
+        """
+        if type(fh._values) == pd.DatetimeIndex:
+            df = pd.DataFrame()
+            df["ds"] = fh._values
```
mloning, line 241:
```suggestion
            df["ds"] = fh.to_pandas()
```

sktime/forecasting/prophet.py
```diff
@@ -0,0 +1,341 @@
...
+        -------
+        y_pred : pandas.Series
+            Returns series of predicted values.
+
+        Raises
+        ------
+        Exception
+            Error when merging data
+        """
+        if type(fh._values) == pd.DatetimeIndex:
```
mloning, line 239:
```suggestion
        if isinstance(fh.to_pandas(), pd.DatetimeIndex):
```

sktime/forecasting/prophet.py
```diff
@@ -0,0 +1,341 @@
...
+        else:
+            # Try to create pd.DatetimeIndex
+            try:
+                periods = fh.to_pandas().max()
+                periods = fh.to_pandas().max()
+                df = self._forecaster.make_future_dataframe(
+                    periods=periods + 1, freq=self.freq, include_history=False)
+                df = df.iloc[fh.to_pandas()]
+            except Exception:
+                raise TypeError("Type of fh values must be int, np.array, list or pd.DatetimeIndex")
```
mloning, line 251: I'd wrap this into a helper function - makes it more readable - `_coerce_to_datetime_index()` perhaps?

sktime/utils/validation/series.py
```diff
@@ -61,11 +66,11 @@ def check_series(Z, enforce_univariate=False, allow_empty=False, allow_numpy=Tru
         _check_is_univariate(Z)
 
     # check time index
-    check_time_index(Z.index, allow_empty=allow_empty)
+    check_time_index(Z.index, allow_empty=allow_empty, index_type=index_type)
     return Z
 
 
-def check_time_index(index, allow_empty=False):
+def check_time_index(index, allow_empty=False, index_type=None):
```
mloning, line 73: I like the idea of adding this here, I'd rename `index_type` to `enforce_index_type`

sktime/utils/validation/series.py
```diff
@@ -83,13 +90,20 @@ def check_time_index(index, allow_empty=False):
...
+    if index_type:
+        if type(index) is not index_type:
+            raise ValueError(
+                f"{type(index)} is not supported, use "
+                f"type {index_type} instead."
+            )
+    else:
+        # We here check for type equality because isinstance does not
+        # work reliably because index types inherit from each other.
+        if not type(index) in VALID_INDEX_TYPES:
```
mloning, line 102: I'd check for `VALID_INDEX_TYPES` first and regardless of `index_type` is given, we wouldn't want to support other index types not in VALID_INDEX_TYPES, but be more restrictive in some cases.

sktime/forecasting/prophet.py
```diff
@@ -0,0 +1,341 @@
...
+                if len(X) == len(fh.to_pandas()):
+                    X = X.set_index(df.index)
+                    df = pd.concat([df, X], axis=1)
+                else:
+                    df.index = df["ds"]
+                    df = df.merge(X, left_index=True, right_on=X.index)
+            except Exception:
+                raise TypeError(merge_error)
+        if df.empty:
+            raise TypeError(merge_error)
```
mloning, line 268: again I'd wrap this into a separate helper function

sktime/forecasting/prophet.py
mloning: Since we're enforcing a `pd.DateTimeIndex` in `fit` for the training data, I'd also enforce it here for the absolute representation of the `fh`, but the `fh` may also be relative, so we need to check `fh.to_absolute()`

mloning (commented):
Hi @aiwalter - thanks again, this is great work, really appreciate this! 

I've left a number of minor comments throughout the code and still need to check the notebook. Let me know what you think!

sktime/forecasting/base/_base.py
aiwalter: I wrote it to get the params that have been passed when initializing an `_SktimeForecaster` child. This params are used by `Prophet` in the `update()` function to initialize the updated (actually a new) model. You are right, I can just use `get_params()`, so I will fix this.

sktime/forecasting/prophet.py
aiwalter: I am removing `**kwargs`

sktime/forecasting/prophet.py
aiwalter: I actually dont understand this point, in https://github.com/alan-turing-institute/sktime/blob/4422a66a2749eaea41d3fbda723bc8c64f334787/sktime/forecasting/arima.py#L306-L310 it is also done like I did it?

sktime/forecasting/prophet.py
aiwalter: done

sktime/forecasting/prophet.py
aiwalter: I actually did not 100% get what `fh.to_absolute()` is doing and why it is relevant to check it here. Would be great if you can explain?

sktime/forecasting/prophet.py
aiwalter: done

sktime/utils/validation/series.py
aiwalter: done

sktime/utils/validation/series.py
aiwalter: done

sktime/forecasting/prophet.py
aiwalter: done

sktime/forecasting/prophet.py
aiwalter: freq is related to the DatetimeIndex frequency and has to be a string, see here:
        https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#timeseries-offset-aliases
This link is also there in the Docstring. I was also thinking about setting it to "D" (daily) as default, however I think it is better not to set a default value because the user will not keep in mind to set it. `freq` is needed to create the future DatetimeIndex in case no future index was given. `fbprophet` can not make predictions without a DatetimeIndex. If we set it to "D" and the user actually has a DatetimeIndex with hourly data, the predictions will be on daily basis if he forgets to set `freq`...

aiwalter:
Btw I have found a workaraound to fix the `pre-commit` error, just writing it here in case this topic will pop up again. Error is:
```
(sktime) C:\Users\Martin\Desktop\sktime\sktime>git commit -m "implemented change requests"
[INFO] Installing environment for https://github.com/psf/black.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
An unexpected error has occurred: CalledProcessError: command: ('c:\\users\\martin\\miniconda3\\envs\\sktime\\python.exe', '-mvirtualenv', 'C:\\Users\\Martin\\.cache\\pre-commit\\repo7lrolriz\\py_env-python3')
return code: 1
expected return code: 0
stdout:
    FileNotFoundError: [Errno 2] No such file or directory: 'c:\\users\\martin\\miniconda3\\envs\\sktime\\Lib\\venv\\scripts\\nt\\python.exe'
```
The solution is described here: https://github.com/pre-commit/pre-commit/issues/1645
Basically just doing the following to fix it:
```
pip uninstall -y virtualenv
pip install virtualenv==20.0.33
```

aiwalter:
@mloning in `continuous-integration/appveyor/pr` https://ci.appveyor.com/project/mloning/sktime/builds/36678854/job/v1cbcke7w3x2uvto I got the same error as I was facing locally with `pystan`. Seems there is missing a C++ compiler, we can solve this by doing `conda install libpython m2w64-toolchain -c msys2
` before installing `pystan`, any idea how to include this? This would at least solve problems for Windows..
See here: https://pystan.readthedocs.io/en/latest/windows.html#

sktime/forecasting/prophet.py
```diff
@@ -0,0 +1,394 @@
...
+        if X is not None:
+            df = df.merge(X, left_index=True, right_on=X.index)
+            for col in X.columns:
+                self._forecaster.add_regressor(col)
+
+        self._forecaster.fit(df=df, **fit_params)
+        self._is_fitted = True
+        return self
+
+    def _predict(self, fh, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
```
aiwalter, line 202: todo: @aiwalter 
if alpha != None, raise exception. because alpha can only be provided when fitting.

sktime/forecasting/prophet.py
mloning: I see, in that case I'd set it to None and infer it from the given data and if it cannot be inferred, raise an error

sktime/forecasting/prophet.py
mloning: It should be in fit there too I believe. I worry about the following situation:
```python

f = AutoARIMA()
f._forecaster.get_params()["max_d"]
>>> 2  # default value
f.set_params(**{"max_d": 1})
f._forecaster.get_params()["max_d"]
>>> 2  # internal value not updated 

y = load_airline()
f.fit(y)
f._forecaster.get_params()["max_d"]
>>> 2  # the wrong value is used in fit 
```

sktime/forecasting/prophet.py
mloning: The forecasting horizon can be represented in two ways: relative or absolute, that is, (integer-valued) steps ahead *relative* to some reference point (e.g. the end of the training set) or the exact, *absolute* time points to predict (e.g. time stamps). Different libraries expect different specifications. The idea of our `ForecastingHorizon` class is to make it easy to switch between these representations. `fbprophet` expect the absolute one as far as I understand, as they encode it in the (time) index of the future dataframe, and the index is always absolute. To get the absolute representation, you can call inside the forecaster: `fh.to_absolute(self.cutoff)` where the `cutoff` is the reference point that allows you to translate one representation to the other.

.appveyor.yml
```diff
@@ -65,6 +65,7 @@ install:
 
   # Install requirements from inside conda environment
   - cmd: activate testenv
+  - cmd: conda install libpython m2w64-toolchain -c msys2
```
mloning, line 68: Could we add some comment here and below in for azure-pipelines why we need this for others to understand?

sktime/forecasting/prophet.py
mloning: I believe statsmodels have a function to go between "sp" values and pandas "freq" values, here in case that's helpful: https://github.com/jseabold/statsmodels/blob/master/statsmodels/tsa/tsatools.py#L780

So how is seasonal periodicity set for prophet? Would be nice to have a unified way of doing that for all forecasters (so far we've used the sp keyword) - but probably best to separate that into its own issue/PR.

aiwalter:
@aiwalter  
**todo**: `update()` has to return `self` instead of a separate model, so that its possible to do `model.update()`

sktime/forecasting/prophet.py
aiwalter: yes thats actually a nice idea. As we enforce DatetimeIndex, the will also be a index.freq value.

mloning:
@aiwalter I can take a look at this one in the next couple of days, want to include it in the next release.

aiwalter:
@mloning that sounds good and would be nice to have it in the next release. **Let me know when you start with working on this branch**, I am trying meanwhile more things to get it running on all environments. The code should be fine now, I didnt run tests locally due to an issue of my setup.

The `neuralpeophet` (#529)  package looks also very promising to me, with a minimal config I got even better results in `smape_loss` compared to the `AutoARIMA` example in the `01_forecasting.ipynb` notebook. Should we try to include this into the next release already? So I might then work on this and you take the `fbprophet` part? Btw I was thinking about reaming the `prophet.py` file to `fbprophet.py` as this is the Python package name, its up to you.

aiwalter:
@mloning you might have a look [here ](https://pystan.readthedocs.io/en/latest/getting_started.html) or [here ](https://pystan.readthedocs.io/en/latest/windows.html#windows) in how to install `pystan` as this is where the propblem come from :/ . Is this AppVeyor actually a Windows VM, right? I have fixed now the merge conflicts and other minor things. The yaml files are now as in master as this didnt work out what tried. I will handover this branch to you now, as it is probably more easy for you to just login to the platforms and check how to install `fbprophet` with `pystan` there sucessfully... Puzzling this out with the yaml files is not the best idea.. When this installation issues are fixed, I might fix issues from code side if needed, just let me know..

mloning:
Hey, I've managed to fix the compilation errors on Windows.

Note it's still not ideal because we need to have a special line just for fbprophet in `.appveyor.yml`, but it works for now and it can later be fixed by using conda and an `environment.yml` file.

Some tests are still failing, will look into that in the next few days.

aiwalter:
Good news, then let me fix those issues

aiwalter:
The test issues seem to come from two main issues:

1. There is an issue with doing `self.update_predict_single(y_test)`. There is no refit happening and that is why the returned forecast is not correct. The returned forecast is based on the old train data as this is stored in `self._forecaster.history_dates.max()`, which is basically the way how `fbpropohet` is saving what we call `self.cutoff`. I think this `update_predict_single()` function must call the `self.fit()` function first? Or I need to implement a different `update()` function? Issue comes from test `forecasting/tests/test_all_forecasters.py::test_predict_time_index[1-int-int-True-Prophet]`

2. The enforcement of index type to `pd.DatetimeIndex` in `y, X = check_y_X(y, X, enforce_index_type=pd.DatetimeIndex)` in fitting causes problems in the tests, as the tests dont expect exceptions here when another index type is used. Any idea? Exemplary test here is `sktime/forecasting/tests/test_all_forecasters.py::test_predict_time_index[1-int-int-True-Prophet]`

mloning:
@aiwalter okay raising `NotImplementedErrors` for those cases should get you through the tests, let me know if that works, otherwise I can take a look again if everything else seems to work. We could also allow for except `TypeError` with an appropriate error message if that would easier.

aiwalter:
seems it doesnt like the `update` function with `NotImplementedError`

mloning:
@aiwalter alright, I can take a look tomorrow

mloning (approved):
Hopefully all tests pass now. Here's a short summary of the changes: 

* added `Prophet` forecaster and `_ProphetAdapter` class
* added `verbose` flag to `Prophet` to suppress pystan output
* minor changes to unit testing framework (less strict check on prediction intervals, updated `_construct_instance`)
* updated requirements (increasing some minimum required versions, clean up after #538)
* restructured adapters file into module to keep related classes and functions together

mloning:
Alright @aiwalter - looks good to go now in my opinion! Anything you want to review or add before we merge?

aiwalter:
@mloning thx for doing the finish! Looks good now, also with the separation of the adapters 👍🏻 . I have just added an example in the notebook, so I think we can merge it.

ngupta23:
I think Prophet is not installed using this command

```
pip install sktime[all_extras]
```

...although the error states that it will be.

```
ModuleNotFoundError: No module named 'fbprophet'. 'fbprophet' is a soft dependency and not included in the sktime installation. Please run: `pip install fbprophet`. To install all soft dependencies, run: `pip install sktime[all_extras]`
```

aiwalter:
@ngupta23 this is not yet in PyPI. In master it is correctly done here:
 https://github.com/alan-turing-institute/sktime/blob/58abf4de44715a6b57fda8effb5aad20cf9e5acb/setup.py#L42

So pls try to install `sktime`from Github if you already want to use `fbprophet`.

Issue #342, opened 2020-07-15
Integrate tbats package


mloning:
**Is your feature request related to a problem? Please describe.**
Integrate tbats package: https://pypi.org/project/tbats/

**Describe the solution you'd like**
Adding tbats as a soft dependency and wrapping it or copying code.

mloning:
@aiwalter in case you're interested!

aiwalter:
@mloning  I am working on this. Can this be a regular dependency or should it also be a soft dependency?

mloning:
I generally prefer soft dependencies, unless it's used in several places in the code base.

mloning:
Closed by #527

PR #527, opened 2020-12-06
Tbats

aiwalter:
#### Reference Issues/PRs

Enhancement as described in #342

Added BATS, TBATS is going to follow in this PR after resolving question ( see below)

soft dependecy: tbats

##### For all contributions

##### For new estimators

#### Any other comments?

@mloning I have 3 questions:
___

**Question 1:**
regarding your understanding of the future in relation to `ForecastingHorizon`. 
```
fh = ForecastingHorizon([1,3,5])
pred = model.predict(fh=fh)
pred

0    178.673409
1    178.673409
2    178.673409
3    178.673409
4    178.673409
dtype: float64
```
Now I have to do `pred[fh.to_pandas()]` in order to get only those three time points. The question that comes here is which index has the first prediction e.g. when doing `model.predict(1)`. From my understanding, this has index 1 and not 0? So just did `pred.index = pd.RangeIndex(start=pred.index.start + 1, stop = pred.index.stop + 1)
`. Is this ok for you?

___
**Question 2:**
I haave seen that when I do e.g. `model.predict(5)`, the fh is set to `ForecastingHorizon([5], dtype='int64', is_relative=True)` automatically. Is this a design pupose? As a user I would expect to get all 5 steps predicted, so `fh` could be set to `ForecastingHorizon([1,2,3,4,5], dtype='int64', is_relative=True)` instead. What do you think?

___
**Question 3:**
This PR yet contains only BATS. To add TBATS there are two options. Both, BATS and TBATS have the same args and fit/predict function. Options:
1. Adding an arg like `type=BATS` 
2. Creating a new file `tbats.py` and inherit from `bats.py`. This is my preffered option :) Because then the user will not have to deal with a new arg that is unknown from original `tbats` packe interface.


mloning:
Hi @aiwalter! 

* Q1: I'd return this: `pd.Series(y_pred[fh.to_indexer(self.cutoff)], index=fh.to_absolute(self.cutoff))` assuming y_pred is a np.array or using `y_pred.iloc[...]` otherwise - this selects the correct values and sets the correct index of the data given the forecasting horizon.
* Q2: Yes, that's just a convention on our side to make it a bit more convenient for users, I'm happy to re-consider how we handle this, but the potential problem with your solution is that `ForecastingHorizon([5])` and `ForecastingHorizon(5)`
would no longer mean the same which may be even more confusing. 
* Q3: Yes, I'd prefer having to classes with a shared parent class - feel free to create a `sktime/forecasting/bats` with multiple private files in there and an init that exposes all public functionality to users on the module.

aiwalter:
@mloning Could you help me in understanding the failure of _appveyor_?

mloning:
Hi @aiwalter, I synced with master and restarted the tests. The problem doesn't seem to be on our side, it fails during the installation of conda. Let's ignore it for now, it usually sorts itself out within a few days.

sktime/forecasting/base/_adapters.py
```diff
@@ -12,6 +12,72 @@
...
+            The forecasters horizon with the steps ahead to to predict.
+        X : pd.DataFrame, optional (default=None)
+            Exogenous variables (ignored)
+
+        Returns
+        -------
+        self : returns an instance of self.
+        """
+        if X is not None:
+            raise NotImplementedError("BATS/TBATS don't support exog or endog data.")
```
mloning, line 38:
```suggestion
            raise NotImplementedError("BATS/TBATS don't support exogenous data.")
```

mloning: Thinking about it, we probably shouldn't raise an error in this case. I see it this way: all forecasters accept X but some choose to ignore it. The problem is that this breaks composition workflows like ensembles where some forecasters may work with exogenous data but others don't. I believe that for the same reason, the `DummyClassifier` in scikit-learn also doesn't raise an error when passing X even though X is ignored. 

We could think about raising a warning. In that case, we should implement that in `check_y_X` with an additional keyword (something like `warn_if_exogenous`) and a corresponding unit tests based on estimator tags for `endogenous_only`. But not sure if that's necessary and probably best done in a separate PR.

sktime/forecasting/base/_adapters.py
```diff
@@ -12,6 +12,72 @@
...
+        self._set_y_X(y, X)
+        self._set_fh(fh)
+
+        self._forecaster = self._forecaster.fit(y)
+        self._is_fitted = True
+        return self
+
+    def _predict(self, fh, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
+        if not fh.is_relative:
+            fh = fh.to_relative(cutoff=self.cutoff)
```
mloning, line 50: No need to check for if it is relative, that's handled internally. So you can simply call `fh = fh.to_relative(cutoff=self.cutoff)` and if it is relative already it ignores the cutoff and just returns `self`.

sktime/forecasting/base/_adapters.py
```diff
@@ -12,6 +12,72 @@
...
+
+        if not fh.is_all_in_sample(cutoff=self.cutoff):
```
mloning, line 52: I feel like we can handle the in-sample vs out-of-sample predictions better on a higher-level in the class hierarchy but happy to leave it as it is for now.

sktime/forecasting/bats.py
```diff
@@ -0,0 +1,96 @@
...
+        self.n_jobs = n_jobs
+        self.multiprocessing_start_method = multiprocessing_start_method
+        self.context = context
+
+        super(BATS, self).__init__()
+
+        # import inside method to avoid hard dependency
+        from tbats import BATS as _BATS
+
+        self._forecaster = _BATS(
```
mloning, line 85: Given the discussion in the fbprophet PR #515, I prefer doing that in `fit` and not `__init__`, what do you think?

sktime/forecasting/tbats.py
```diff
@@ -0,0 +1,97 @@
...
+        self.n_jobs = n_jobs
+        self.multiprocessing_start_method = multiprocessing_start_method
+        self.context = context
+
+        super(TBATS, self).__init__()
+
+        # import inside method to avoid hard dependency
+        from tbats import TBATS as _TBATS
+
+        self._forecaster = _TBATS(
```
mloning, line 86: same as above - regarding design: the `_TbatsAdapter` could have an abstract `_instantiate_model` method which is called in fit and implemented by the concrete `TBATS` and `BATS` classes.

mloning (commented):
Thanks @aiwalter - again all of this looks very good to me, happy to merge this once the tests pass (there seem to some issues on the side of the CI platforms but hopefully they sort themselves out soon).

sktime/forecasting/base/_adapters.py
aiwalter: I have added a warning as you described in the code, but not yet in the tests.

sktime/forecasting/base/_adapters.py
aiwalter: yes I agree, this higher-level function could just receive the in-sample and out-sample predictions and return a combined prediction.

sktime/forecasting/bats.py
aiwalter: but then I cant share the fit function of `_TbatsAdapter` anymore but have to overwrite it?

sktime/forecasting/tbats.py
aiwalter: I did that, however pls recheck.

sktime/forecasting/bats.py
aiwalter: Ok I have solved this by passing the import `from tbats import TBATS as _TBATS` also as `self.ModelClass = _TBATS`. Looks maybe a bit strange but works :)

mloning:
This may be related to the appveyor issue: https://stackoverflow.com/questions/65221052/my-fixed-pyopenssl-version-19-0-0-suddenly-broke (let's wait a bit longer before we adjust our requirements file)

aiwalter:
@mloning seems Azure pipeline is stuck? I increased the timeout to 120 mins as it has had a timeout issue before. Now I dont know whats the problem..

mloning:
Ah but appveyor is back on! 😄 

Azure doesn't seem to like the timeout value: https://dev.azure.com/mloning/sktime/_build/results?buildId=1638&view=results

aiwalter:
@mloning the default timeout is 60 mins. The 120 mins was even not enough, so I increase it again. I would appreciate if there is any way to speed it a bit up :)

sktime/forecasting/tbats.py
```diff
@@ -0,0 +1,62 @@
...
+        How many jobs to run in parallel when fitting BATS model.
+        When not provided BATS shall try to utilize all available cpu cores.
+    multiprocessing_start_method: str, optional (default='spawn')
+        How threads should be started. See also:
+        https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods
+    context: abstract.ContextInterface, optional (default=None)
+        For advanced users only. Provide this to override default behaviors
+    """
+
+    def __init__(self, **kwargs):
```
mloning, line 56: Having `**kwargs` in the constructor is usually not a good idea - makes it hard for users to understand.

mloning:
We could split it up by creating separate builds for all Python versions but setting up the pipeline for manylinux wheels wasn't the easiest and it has worked before on master below 50 mins (so I'm assuming it's again an issue on the azure side rather than our side), but speeding it up would be nice.

sktime/forecasting/tbats.py
aiwalter: This `**kwargs` are required here because some parts of the `__init__` are defined in the parent `_TbatsAdapter` and one of the `self` arguments is defined in the children. I dont know any other way to do that and this is the most minimal way. Otherwise I would have to keep the whole `__init__` in the children and do copy paste.

aiwalter:
@mloning so should I rename `seasonal_periods` to `sp`? I think you mentioned that somewhere but I didnt find it again? There are several options we could do in renaming or accpeting args that `sktime` would like to unify with a same name, here with the example of `seasonal_periods`:

1. Just rename to `sp`
2. Rename to `sp` but accept `**kwargs` to that users could still use `seasonal_periods` as this is `tbats` original naming. If both raise exception.
3. Using `seasonal_periods` butt accept `**kwargs` so that user can give also `sp`. If both raise exception.
4. Having both, `seasonal_periods` and `sp` as args and chosing the one that is given. If both raise exception.

I would prefer option 2 or 3, what do you think? Is there any other arg inspite `seasonal_periods` that should be renamed here?

sktime/forecasting/tbats.py
mloning: I'd say that's preferable to having `*kwargs`. 

Could the `_ModelClass` not be a class variable so that we don't have to overwrite the `init` at all if it has the same signature for both methods? Something like this:
```python
class TBATS(_TbatsAdapter):
    _ModelClass = _TBATS
```

aiwalter: it has to be `self._ModelClass` because in `fit()` I am doing then `self._instantiate_model()` which requires to read then `self._ModelClass`

mloning:
@aiwalter Okay this is looking really good now, all tests pass! 

Regarding `sp` I prefer option 1 and avoid `*kwargs` if at all possible and multiple options for doing the same thing following the Zen of Python: "There should be one-- and preferably only one --obvious way to do it." 

`*kwargs` for a user is just not helpful at all, but see my comment above. I'm sure there's a way to handle this without much duplication.

sktime/forecasting/tbats.py
mloning: Yes but class variables are accessible to objects via `self`, no? This for example works:

```python
class A:
    _x = 1

    def get_x(self):
        return self._x

a = A()
a.get_x()
>>> 1
```

mloning:
All other arg names seem fine.

sktime/forecasting/base/_adapters.py
```diff
@@ -63,7 +63,7 @@ def _instantiate_model(self):
         )
         return self
 
-    def _get_sp(sp):
+    def _get_sp(self, sp):
```
mloning, line 66: Wouldn't make this a method but a simple standalone function something like `_check_sp`. We also have a `check_sp` function [here](https://github.com/alan-turing-institute/sktime/blob/28646ab32d083f799d7723efcf514ee558c43694/sktime/utils/validation/forecasting.py#L151) which you could use in `fit`

sktime/forecasting/tbats.py
aiwalter: I just tried this, but doing to I would have to do `_ModelClass = TBATS` in `tbats.py` but this raises an issue `local variable '_ModelClass ' is assigned to but never used flake8(F841)`

aiwalter: Ok I got it resolved, now its even more minimalistic, probably the most minimal wrapper out there in the ML world :D So lets got for it!

aiwalter:
seem the tests dont like such minimal classes. What should I do @mloning ? Going back to the version with `**kwargs`?

```python
C:\Users\Martin\Desktop\sktime\sktime\sktime>pytest tests/test_all_estimators.py::test_estimator[BATS]
=============================================================== test session starts ================================================================
platform win32 -- Python 3.8.5, pytest-6.1.2, py-1.9.0, pluggy-0.13.1
rootdir: C:\Users\Martin\Desktop\sktime\sktime, configfile: setup.cfg
collected 1 item

tests\test_all_estimators.py F                                                                                                                [100%]

===================================================================== FAILURES =====================================================================
_______________________________________________________________ test_estimator[BATS] _______________________________________________________________

Estimator = <class 'sktime.forecasting.bats.BATS'>

    @pytest.mark.parametrize("Estimator", ALL_ESTIMATORS)
    def test_estimator(Estimator):
        # We run a number of basic checks on all estimators to ensure correct
        # implementation of our framework and compatibility with scikit-learn.
>       check_estimator(Estimator, EXCLUDED_TESTS.get(Estimator.__name__, []))

tests\test_all_estimators.py:24:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
utils\_testing\estimator_checks.py:53: in check_estimator
    check(Estimator)
utils\_testing\estimator_checks.py:480: in check_multiprocessing_idempotent
    estimator.fit(*args["fit"])
forecasting\base\_adapters.py:87: in fit
    self._forecaster = self._forecaster.fit(y)
..\..\..\..\miniconda3\lib\site-packages\tbats\abstract\Estimator.py:98: in fit
    best_model = self._do_fit(y)
..\..\..\..\miniconda3\lib\site-packages\tbats\bats\BATS.py:75: in _do_fit
    return self._choose_model_from_possible_component_settings(y, components_grid)
..\..\..\..\miniconda3\lib\site-packages\tbats\abstract\Estimator.py:143: in _choose_model_from_possible_component_settings
    with self.context.multiprocessing().Pool(processes=self.n_jobs) as pool:
..\..\..\..\miniconda3\lib\multiprocessing\context.py:119: in Pool
    return Pool(processes, initializer, initargs, maxtasksperchild,
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <multiprocessing.pool.Pool state=INIT pool_size=0>, processes = -1, initializer = None, initargs = (), maxtasksperchild = None
context = <multiprocessing.context.SpawnContext object at 0x0000010E24D58FA0>

    def __init__(self, processes=None, initializer=None, initargs=(),
                 maxtasksperchild=None, context=None):
        # Attributes initialized early to make sure that they exist in
        # __del__() if __init__() raises an exception
        self._pool = []
        self._state = INIT

        self._ctx = context or get_context()
        self._setup_queues()
        self._taskqueue = queue.SimpleQueue()
        # The _change_notifier queue exist to wake up self._handle_workers()
        # when the cache (self._cache) is empty or when there is a change in
        # the _state variable of the thread that runs _handle_workers.
        self._change_notifier = self._ctx.SimpleQueue()
        self._cache = _PoolCache(notifier=self._change_notifier)
        self._maxtasksperchild = maxtasksperchild
        self._initializer = initializer
        self._initargs = initargs

        if processes is None:
            processes = os.cpu_count() or 1
        if processes < 1:
>           raise ValueError("Number of processes must be at least 1")
E           ValueError: Number of processes must be at least 1

..\..\..\..\miniconda3\lib\multiprocessing\pool.py:205: ValueError
================================================================= warnings summary =================================================================
..\..\..\..\miniconda3\lib\site-packages\win32\lib\pywintypes.py:2
  c:\users\martin\miniconda3\lib\site-packages\win32\lib\pywintypes.py:2: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
    import imp, sys, os

..\..\..\..\miniconda3\lib\site-packages\patsy\constraint.py:13
  c:\users\martin\miniconda3\lib\site-packages\patsy\constraint.py:13: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
    from collections import Mapping

-- Docs: https://docs.pytest.org/en/stable/warnings.html
============================================================= short test summary info ==============================================================
FAILED tests\test_all_estimators.py::test_estimator[BATS] - ValueError: Number of processes must be at least 1
==================================================== 1 failed, 2 warnings in 157.19s (0:02:37) =====================================================
```

mloning:
Okay bats and tbats is just really, really slow, even with the most restricted model and tiny amount of toy data. That explains why you had to increase the time out. Not sure what to do about that. It does slow down everything. Perhaps moving it into a separate companion repo may be a better idea. Let's see what they say https://github.com/intive-DataScience/tbats/issues/23

The error you posted is because it doesn't like `-1` for `n_jobs` but we have an `check_n_jobs` function to take of that. 

I checked out the code locally and made some changes to get everything to work. All tests passed locally, let's see what happens on CI.

aiwalter:
thanks for having a look into it :)

aiwalter:
Did you see [this?]( https://github.com/intive-DataScience/tbats/issues/17). We could try with `n_jobs=1`?

And [here ](https://github.com/intive-DataScience/tbats/issues/1) they recommend `use_box_cox=False` and `use_arma_errors=False`

mloning:
Yes I've set these parameters already in `sktime/tests/_config.py` - still pretty slow compared to the other algorithms.

aiwalter:
This was already much faster now, seems it was running about 90 minutes only..

mloning:
Yes but it usually runs in less then 50 mins (see e.g. https://github.com/alan-turing-institute/sktime/pull/485) - doubling the overall run time isn't great.

aiwalter:
Then I would propose to only run tests on either BATS or TBATS as they share the same code, only different imports. Is this ok for you? How could we exclude one of them from the tests? Would otherwise be a bit unfortunate to have coded this...

mloning:
@aiwalter yes that sounds like a good idea. I don't want to waste the code either! Just thinking that perhaps it's better to set up a separate repo since the sktime's forecasting API is relatively stable now. But let's wait first to hear back from them. 

Have you ever tried the R version, is that also so slow?

aiwalter:
I did not try the R version yet. But I just tested this relative comparison on my (slow) local machine:

Setting `n_jobs=1` did a MAJOR speed up:

- 0.36 seconds for the following:
`model = TBATS(sp=[], use_arma_errors=False, use_box_cox=False, use_trend=False, use_damped_trend=False, n_jobs=1)
`

- 19.03 seconds with default `n_jobs=None`:
`model = TBATS(sp=[], use_arma_errors=False, use_box_cox=False, use_trend=False, use_damped_trend=False)
`

mloning:
@aiwalter yes just saw that too, will push the `n_jobs=1` changes - should be the default value for `n_jobs=None` anyway following scikit-learn's convention.

aiwalter:
great, yes lets set `n_jobs=1` as default in our code, not only in tests :)

mloning (approved):

PR #542, opened 2020-12-16
Add MiniRocket and MiniRocketMultivariate

angus924:

Implements MiniRocket (https://github.com/angus924/minirocket) and MiniRocketMultivariate (a simple/naive extension to multivariate data), a new method which is much faster (up to 75x) than Rocket, and largely deterministic.  The paper should be online imminently.

No.  (Same as for Rocket.)

##### For all contributions

##### For new estimators

#### Any other comments?


mloning (commented):
Hi @angus924 - looks great already! 

We made `numba` a core dependency, so you no longer have to take care of it with the `_check_soft_dependencies` function. 

Make sure to take a look at the code style section in CONTRIBUTING.md - I'm happy to fix it if you get stuck.

mloning:
@angus924 could you also please add you to the code owners file to make sure that you get a notification in case someone touches the algorithms you wrote?

angus924:
Thanks very much @mloning.

> We made `numba` a core dependency, so you no longer have to take care of it with the `_check_soft_dependencies` function.

I'll fix this.

> Make sure to take a look at the code style section in CONTRIBUTING.md - I'm happy to fix it if you get stuck.

I'll fix this (or try to...).  It looks like this is where the build is tripping up?  I'll let you know if I get stuck.  Sorry, I just haven't had the chance to work through the code yet to address this.

> @angus924 could you also please add you to the code owners file to make sure that you get a notification in case someone touches the algorithms you wrote?

Will do.

angus924:
Hi @mloning, it looks like it's building now.

What else do you need at this stage?

sktime/transformers/panel/minirocket.py
```diff
@@ -0,0 +1,842 @@
...
+    """
+
+    _tags = {"univariate-only": True}
+
+    def __init__(
+        self, num_features=10_000, max_dilations_per_kernel=32, random_state=None
+    ):
+        self.num_features = num_features
+        self.max_dilations_per_kernel = max_dilations_per_kernel
+        self.random_state = (
```
mloning, line 47: this should happen in `fit` rather than `init` in case it's changed later via `set_params()`

sktime/transformers/panel/minirocket.py
```diff
@@ -0,0 +1,842 @@
...
+
+@njit(
+    "float32[:](float32[:,:],int32[:],int32[:],float32[:],optional(int32))",
+    fastmath=True,
+    parallel=False,
+)
+def _fit_biases(X, dilations, num_features_per_dilation, quantiles, seed):
+
+    if seed is not None:
+        np.random.seed(seed)
```
mloning, line 106: there's the `check_random_state` function from sklearn which returns an rng object which you can then use via `rng.randn` for example

sktime/transformers/panel/minirocket_multivariate.py
```diff
@@ -0,0 +1,931 @@
...
+    max_dilations_per_kernel : int, maximum number of dilations per kernel (default 32)
+    random_state             : int, random seed (optional, default None)
+    """
+
+    def __init__(
+        self, num_features=10_000, max_dilations_per_kernel=32, random_state=None
+    ):
+        self.num_features = num_features
+        self.max_dilations_per_kernel = max_dilations_per_kernel
+        self.random_state = (
```
mloning, line 45: same here as above

mloning (commented):
Thanks @angus924 - I suggest to create a submodule `transformers/panel/rocket/` which contains the various rocket related files and exposes all public classes via an `transformers/panel/rocket/__init__.py`.

mloning:
@angus924 I introduced a quite a lot of changes on `master` yesterday for the `v0.5.0` release, let me know if you need help resolving conflicts.

sktime/transformers/panel/minirocket.py
angus924: > this should happen in `fit` rather than `init` in case it's changed later via `set_params()`

Sorry, I don't quite understand what you are asking here...  Which part(s) exactly should move from `init` to `fit`?

sktime/transformers/panel/minirocket.py
angus924: > there's the `check_random_state` function from sklearn which returns an rng object which you can then use via `rng.randn` for example

This is the same problem we have with rocket...  You can't pass a `RandomState` into jit code.  However, unlike with rocket, I think there is a potential workaround here.  We could pass the `RandomState` to `_fit(...)` (regular python), generate a big list of indices using the `RandomState`, and pass the indices into `_fit_biases(...)`.

angus924:
> @angus924 I introduced a quite a lot of changes on `master` yesterday for the `v0.5.0` release, let me know if you need help resolving conflicts or have any questions.

No problem, I'll have a go at it.

sktime/transformers/panel/minirocket.py
mloning: Sorry for not being clearer, the processing of the random state before you assign it to `self.random_state`, I believe that's best done in `fit` via some `check_random_state` function, I now remember the problem you describe with scikit-learn (forgot about that) but we could have our own ROCKET-specific `_check_random_state` function here

sktime/transformers/panel/minirocket.py
mloning: sounds good, also see my comment above on `random_state` in `__init__`

angus924:
> Thanks @angus924 - I suggest to create a submodule `transformers/panel/rocket/` which contains the various rocket related files and exposes all public classes via an `transformers/panel/rocket/__init__.py`.

This should be fine, a possible complication is if numba decides to compile everything imported into `__init__.py`, which would be highly undesirable...  I'll try it out and see what happens.

sktime/transformers/panel/minirocket.py
angus924: All good, I'll have a think about a solution here.

angus924:
> > Thanks @angus924 - I suggest to create a submodule `transformers/panel/rocket/` which contains the various rocket related files and exposes all public classes via an `transformers/panel/rocket/__init__.py`.
> 
> This should be fine, a possible complication is if numba decides to compile everything imported into `__init__.py`, which would be highly undesirable... I'll try it out and see what happens.

Ok, so this does indeed happen...  Incuding the imports in `__init__.py` induces Numba to compile everything (Rocket, MiniRocket, and MiniRocketMultivariate).

Accordingly, I have moved Rocket, MiniRocket, and MiniRocketMultivariate to a new `transformations/panel/rocket` submodule, and set `cache = True` on all the jit functions.  The first import will be quite slow (as Rocket, MiniRocket, and MiniRocketMultivariate all get compiled, even if you are just importing, e.g., Rocket), but any subsequent imports should be more or less instantaneous.

mloning (approved):
Thanks @angus924 - everything looks good to me now!

Issue #541, opened 2020-12-15
[ENH] Plotting confidence intervals


aiwalter:
**Is your feature request related to a problem? Please describe.**
Plotting confidence intervalls with the current plot function `from sktime.utils.plotting import plot_series` is not well suitable in case all lines are very close, see picture.

![image](https://user-images.githubusercontent.com/29627036/102230635-6d7cd180-3eed-11eb-82b1-54bbc9f73bac.png)

**Describe the solution you'd like**
Plot like this would be great, with the points as an option ideally. The function could just receive the DataFrame `pred_int` from `y_pred, pred_int = model.predict(fh=fh, return_pred_int=True)
` as an input so this is already well structured and named.

![image](https://user-images.githubusercontent.com/29627036/102231098-e419cf00-3eed-11eb-97ab-9101f6eaa195.png)

mloning:
Yes we could add an `pred_int=None` input argument to `plot_series` with the same time index as one of the other series. As you say, it'd optionally accept a `pd.DataFrame` and plot the intervals.

timcashion:
I'd love to take this on!

aiwalter:
@timcashion cool, I have just seen that in examples `01_forecasting.ipynb` at the end there  is a plot like I proposed. So basically you could just copy this into the function :)

aiexplorations:
1. Would it be a good idea to have CIs at different confidence levels (say, 90%, 95%, 99%) also graphically represented in these plots? 
2. Would it also be good to have a related feature to store the CIs in the forecast as series (series of tuples, or as individual columns)?

fkiraly:
@aiwalter, @timcashion, one of the ideas in the sktime ecosystem was to use interval and distribution returns compatible with the skpro (probabilistic prediction) toolbox https://github.com/alan-turing-institute/skpro 
Unfortunately skpro has gone on hiatus after the disappearance of its maintainer half a year ago - any appetite for taking parts of the development/prototyping on with the ultimate aim of getting nice probabilistic returns for sktime?

aiwalter:
@aiexplorations 
1. We would have to fit multiple models  because most models can only return CIs for a single alpha value. But actually it could work to accept also a list like `alpha = [0.95, 0.9, 0.5]` and loop 3 times over `_fit()` and `predict()`.
[...]

PR #545, opened 2020-12-17
Update plotting to handle prediction intervals

timcashion:
#### Reference Issues/PRs
Fixes Issue [#541](https://github.com/alan-turing-institute/sktime/issues/541)

Allows plot_series to take pred_int (2nd output of model.predict()) as an option variable. If present, plots lower and upper bound confidence intervals for prediction using plt.fill_between. The index of pred_int is also checked to make sure it fits within same axis as existing series. 

No

##### For all contributions


sktime/utils/plotting/__init__.py
```diff
@@ -75,6 +77,25 @@ def plot_series(*series, labels=None):
 
         plot_func(x=x, y=y, ax=ax, marker="o", label=label, color=color)
 
+    # plot prediction intervals if present
+    if pred_int is not None:
+        # check same conditions as for earlier indices
+        if not type(index) is type(pred_int.index):  # noqa
+            raise TypeError("Found series with different index types.")
+        elif all([x in index for x in pred_int.index]):
+            plt.fill_between(
```
mloning, line 86:
```suggestion
            ax.fill_between(
```

sktime/utils/plotting/__init__.py
```diff
@@ -75,6 +77,25 @@ def plot_series(*series, labels=None):
...
+                pred_int.index,
+                pred_int.lower,
+                pred_int.upper,
+                alpha=0.3,
+                color=ax.get_lines()[-1].get_c(),
+                label="95% prediction intervals",
```
mloning, line 92: Alpha may be different so we should assign a more general label here:
```suggestion
                label="prediction intervals"
```

mloning (commented):
Hi @timcashion - thanks for contributing! I had a quick look and something still seems to not work as expected. 

Adding a simple unit test that calls the function would be nice too. 

When I run this, it throws an error: 
```python
y = load_airline()
y_train, y_test = temporal_train_test_split(y, test_size=36)
forecaster = ThetaForecaster(sp=12)
forecaster.fit(y_train)
y_pred, pred_int = forecaster.predict(fh, return_pred_int=True, alpha=0.05)
fig, ax = plot_series(y_train, y_test, y_pred, 
                      labels=["y_train", "y_test", "y_pred"], pred_int=pred_int)
```
Seems like it doesn't like the `pred_int.index` for the x-values in `fill_between`.

sktime/utils/plotting/__init__.py
timcashion: Makes sense!

mloning:
I think we need to use this line something similar to what we do in this line: `xs = [np.argwhere(index.isin(y.index)).ravel() for y in series]` to get the x values for the prediction intervals.

timcashion:
Added two unit tests for plot_series, and this now works with integer and datetime indices.

mloning:
@timcashion have you forgotten to push the latest changes, can't see them?

mloning:
Closed in favour of #802

PR #559, opened 2020-12-23
Add ARIMA

### LINKED ISSUE #412: Interface ARIMA from pmdarima/statsmodels
We already have AutoARIMA, but it would be nice to also interface ARIMA, either from statsmodels or pmdarima.

HYang1996:
#### Reference Issues/PRs

Resolves #412.

Interfaces [`ARIMA`](https://alkaline-ml.com/pmdarima/modules/generated/pmdarima.arima.ARIMA.html) from pmdarima, which internally wraps [`SARIMAX`](https://www.statsmodels.org/stable/generated/statsmodels.tsa.statespace.sarimax.SARIMAX.html) from statsmodels.

Soft dependency on `pmdarima` similar to that in `AutoARIMA`.

##### For all contributions

##### For new estimators


mloning (commented):
Hi @HYang1996 - looks great! 

* One suggestion: add the new class to the `sktime/forecasting/arima.py` or create a new `arima` module with different hidden files inside and expose all methods in the `sktime/forecasting/arima/__init__.py`
* To fix the soft dependency error, adding it to `arima.py` should solve it already, otherwise you have to add it [here](https://github.com/alan-turing-institute/sktime/blob/08c06b0afb4e6f92bd0a012ca84d10da88fe2c66/build_tools/azure/check_soft_dependencies.py#L13)
* How similar is the code to the AutoARIMA wrapper? We could think about creating a common parent class but happy to merge this PR first

HYang1996:
> * How similar is the code to the AutoARIMA wrapper? We could think about creating a common parent class but happy to merge this PR first

- `fit` is essentially the same but involves different instantiation of the estimator objects.
- I copied over  `_predict`, `_predict_in_sample` and `_predict_fixed_cutoff`. Is there a way to verify this was done correctly?
- `get_fitted_params` is slightly different since `arima_res_` attribute is under `self._forecaster.model_` for `AutoARIMA` but `self._forecaster` for `ARIMA`

I think a common parent class could be implemented, but since AutoARIMA and ARIMA seem to be the only estimators offered in pmdarima, the parent class may not have benefits in the future.

mloning:
@HYang1996 alright, I think a common (abstract) parent class still makes sense to avoid duplications, you could add it in `sktime/forecasting/base/adapters/_pmdarima.py`. Model instantiation should take place in fit but you could have a `_instantiate_model` method which is implemented in by the concrete classes (see e.g. how we did it [here](https://github.com/alan-turing-institute/sktime/blob/08c06b0afb4e6f92bd0a012ca84d10da88fe2c66/sktime/forecasting/base/adapters/_tbats.py#L18)), what do you think?

HYang1996:
> @HYang1996 alright, I think a common (abstract) parent class still makes sense to avoid duplications, you could add it in `sktime/forecasting/base/adapters/_pmdarima.py`. Model instantiation should take place in fit but you could have a `_instantiate_model` method which is implemented in by the concrete classes (see e.g. how we did it [here](https://github.com/alan-turing-institute/sktime/blob/08c06b0afb4e6f92bd0a012ca84d10da88fe2c66/sktime/forecasting/base/adapters/_tbats.py#L18)), what do you think?

@mloning Sounds good!

mloning:
Thanks @HYang1996 - I had a quick look and it looks greats! I'll give it a more thorough review over the next couple of days!

mloning (approved):
Hi @HYang1996, I hope you're well and safe and are able to enjoy the holidays. I had a deeper look into the PR now and it all looks very good. I've made some very minor changes (see the latest commit) and will merge once all tests pass again! 

Thanks again for the excellent work!

HYang1996:
Hi @mloning , thank you for the review and it's always a pleasure to contribute!

mloning:
Hi @HYang1996 - looks like there an error unrelated to the changes in this PR, but you may still be able to help since it involves AutoETS. Seems like it's no longer in line with the R results, possibly due to a change in statsmodels? Any ideas?

HYang1996:
> Hi @HYang1996 - looks like there an error unrelated to the changes in this PR, but you may still be able to help since it involves AutoETS. Seems like it's no longer in line with the R results, possibly due to a change in statsmodels? Any ideas? Strangely it's only on Linux.

hmm, I ran the test code manually and it was producing the correct results, and the tests passed locally for me too. Let's see if it fails again.

mloning:
@HYang1996 yes strangely it's only on Linux ...

HYang1996:
> @HYang1996 yes strangely it's only on Linux ...

@mloning Do you think it might be related to [this](https://github.com/statsmodels/statsmodels/commit/877b5b50c21850147574770f96c1a670a7b122d0)?

mloning:
Possibly, only Python 3.6 is affected though. We could just skip these tests on Linux and Python 3.6 ...

mloning:
Seems like pandas==1.2.0 is causing the issue, seems to work with pandas==1.1.5

mloning (approved):

mloning:
Thanks again @HYang1996 - now merged! 🎉

PR #587, opened 2020-12-31
Tidy code in _distance_to_exemplars_inst in stump class

Lovkush-A:
- Remove unnnecessary calculation of minimum distances
- Increase readability by re-writing for loop as `for exemplar in exemplars`, as opposed to introducing variable for n and looping through indices.
- I tried removing the if loop checking if `exemplar.name == instance.name`, but it caused the unittests to fail. Seems unusual to me that a distance function can't deal with equal inputs. Might be something worth changing in the future.

- [ ] Was I correct in thinking minimum distances were unnecessary?
- [ ] Is `for exemplar in exemplars` better than what was there before, or was the previous version better?
- [ ] Is it worth creating an issue to change distance functions to handle equal inputs? 

#### Any other comments?
I realise this is not what I said I would work on, but I while I was skimming code, I just noticed these potential improvements, so thought I might as well do this as my first pull request.

##### For all contributions


TonyBagnall (commented):
Hi thanks for this, I'll ask George to look next week. I suspect the min distance is used for early abandon

mloning:
Any updates on this? 

One problem is that `ProximityForest` is currently not included in the automatic unit tests because it doesn't fully comply with our API spec. Just need to make sure we don't break anything.

TonyBagnall:
no word from george I'm afraid, I dont want to approve until he has looked at it, I will chase him

goastler:
Hi, sorry for the delay, I'm slowing working through my backlog haha!

Anyway, back to sktime. I have a couple of points:

1. The main difference is you're using native python lists in your code, whereas the original uses a numpy array. The numpy array is preallocated and therefore quicker, but I doubt there's much in it so doesn't *particularly* warrant changing but whatever works :)
2. You've removed the min_distance mechanism. This isn't currently in use, but will be. The min_distance is passed to distance measures as a lower bound to early abandon distance calculation, which is the most costly part of elastic based classifiers. Currently, the distance measure implementations in sktime don't use a limit (as far as I know) but they will do once I've updated them (also on my todo list) so we need to leave this in if possible
3. It's concerning that the unit tests are no longer passed. I can't see how this code would change the way PF runs, but if it's not passing the tests then it's not got the same output as before. @mloning mentioned the tests for PF have been turned off so the failure of the tests may have been caused by something else

Lovkush-A:
@goastler

1. OK. So it's a choice between slightly better speed vs. slightly better readability.
2. OK, I will add the calculation of min_distances back in.
3. The unittests only failed when I removed the if loop `if exemplar == instance`. In its current form, the unittests all passed. Do you know why the distance function would break if exemplar == instance?

mloning:
@TonyBagnall @goastler can you please review again and merge if you're happy with the changes?

mloning:
@TonyBagnall @goastler can we merge this?

TonyBagnall:
he, I think I'd like an experimental assessment of the efficiency loss of using native python in this. It can sometimes cause horrendous time overhead

Lovkush-A:
Hello all,
Probably best just to ditch the PR. It does not achieve anything - most of the changes I made originally have been undone, and the one remaining change is very minor and unclear if it is net positive.

One thing I think is worth noting somewhere is there is a need for an if loop checking if `exemplar.name == instance.name`. I think this should be within distance functions, rather than outside of them. Should I create a new issue for that?

PR #593, opened 2021-01-05
Relax `numba` pinning

dhirschfeld:

I'm trying to package `sktime` for `conda` but it turns out that the version of `llvmlite` required by the `numba` pinning wasn't able to be built for py38/win64.

If `sktime` doesn't strictly *require* `numba =0.50.*` then the easiest solution is to simply relax the pinning.


mloning:
Hi @dhirschfeld thanks for the PR, happy to make adjustments.

Are you saying that you'll create the recipe for distributing sktime via conda-forge? That would awesome, it's been on our todo list for some time now!

dhirschfeld:
Hi @mloning! 👋 

I've already opened a `staged-recipes` PR over at:
https://github.com/conda-forge/staged-recipes/pull/13568

I've currently skipped Windows due to the `numba` dependency issue but can easily fix that later and bump the build number... so long as `sktime` works fine with a newer `numba`

If you (or any other `sktime` devs) would like to be added as maintainers of the recipe just make a comment to that effect on the PR and I can add you. I can also do that later so no big rush to do it right now if you're not sure.

mloning:
Great work! Thanks!

dhirschfeld:
Looks like `appveyor` isn't too happy with `pystan` though :/
```
forecasting\tests\test_all_forecasters.py .............Windows fatal exception: access violation
Current thread 0x00000ad4 (most recent call first):
  File "c:\miniconda36-x64\envs\testenv\lib\site-packages\pystan\model.py", line 550 in optimizing
  File "c:\miniconda36-x64\envs\testenv\lib\site-packages\fbprophet\models.py", line 245 in fit
  File "c:\miniconda36-x64\envs\testenv\lib\site-packages\fbprophet\forecaster.py", line 1166 in fit
  File "c:\miniconda36-x64\envs\testenv\lib\site-packages\sktime\forecasting\base\adapters\_fbprophet.py", line 61 in fit
```

mloning:
Looks like there's some issue with the Windows builds ...

mloning:
Yes not sure what changed compared to the previous builds.

dhirschfeld:
An access violation is fairly nasty - it means there's a problem in compiled code, or that different compiled dlls are incompatible.

That's (mostly) a solved problem with `conda`, but if `wheels` are correctly built using the manylinux image then IIUC they should also be compatible. I still think that `wheels` are much more likely to have problems on Windows though.

dhirschfeld:
I'll try tightening the pinning slightly... 🤞

dhirschfeld:
Looks like master failed with the same error:
https://ci.appveyor.com/project/mloning/sktime/builds/37116378/job/isx6vivkryko2xqk#L2742

dhirschfeld:
No joy with 0.51.2:
image

```
forecasting\tests\test_all_forecasters.py .............Windows fatal exception: access violation
Current thread 0x00000ad4 (most recent call first):
```

mloning:
Yes @dhirschfeld not sure where this comes from, will have to look into it and compare it with the last successful build ...

dhirschfeld:
Well, that seems to have "fixed" the CI.

In my experience it's ~always a *terrible* idea to mix `pip` and `conda` installed, compiled packages. Unfortunately, not all deps are on `conda-forge` yet so there's no way around it, but to minimise the chance of exactly these sort of errors I'd strongly recommend installing everything possible with `conda` (as is done here) and only installing what you absolutely have to with `pip`

----

mloning:
Okay great - let me know if you can fix it by using conda, if not, I'll take a closer look!

.appveyor.yml
```diff
@@ -65,7 +65,7 @@ install:
 
   # Install requirements from inside conda environment
   - cmd: activate testenv
-  - cmd: conda install -c conda-forge fbprophet  # prior fbprophet installation
+  - cmd: conda install -c conda-forge "matplotlib>=3.3.2" "seaborn>=0.11.0" "tsfresh>=0.17.0" "hcrystalball>=0.1.9" "stumpy>=1.5.1" "fbprophet>=0.7.1" "numba>=0.50,<0.53" "numpy>=1.19" "pandas>=1.1.0,<1.2" "statsmodels>=0.12.1" "scikit-learn>=0.23.0" "scikit-posthocs>=0.6.5" pystan
```
mloning, line 68: Can we not install from `%REQUIREMENTS`, i.e. `build_tools/requirements.txt` here? We're already specifying our dependencies in way to many places - don't want to add another one here! 🤔

mloning (commented):
Hi @dhirschfeld - thanks again for working on this! Looks good to go, just left one comment.

.appveyor.yml
dhirschfeld: The problem is that not all requirements are available from `conda-forge` (yet). So you'd have to also maintain a `conda` `environment.yml` or inline it as above.

*Aside:* In general `requirements.txt` won't always work as `conda` packages can have different names to their `pip`  counterparts.

dhirschfeld: ☝️ I'm not too sure what can be done about that - I can't see any *good* options. Probably a separate `environment.yml` is easier to maintain (more visible) than inlining the deps and constraints here.

mloning: Yes, okay. Should we try to move away from `requirements.txt` and to `environment.yml` where you can separately specify pip installables?

mloning: Probably best to do that in a separate PR though to merge this one first

dhirschfeld: That wouldn't work for `pip` installing though. Eventually when all deps are on CF you wouldn't need to maintain two files. I'm happy to create the `environment.yml` but that will have to wait until tomorrow night. That can easily be done in a separate PR though, if you want to merge this to fix the Access Violation on master.

mloning: Yeah let's do that! Re the env file, I have something like [this](https://github.com/heidelbergcement/hcrystalball/blob/25f186dc72d4e273c6696a5c822f601d54bab734/environment.yml#L28) in mind where you specify pip installs in the conda env file.

mloning (approved):

mloning:
Thanks again for fixing this! Much appreciated!

.appveyor.yml
dhirschfeld: That would work for *`conda`* installing the deps but you wouldn't be able to *`pip`* install them from that file. If you don't care about `pip` installing the deps then that would be the way to go.

PR #599, opened 2021-01-09
Uea dev

TonyBagnall:
small refinement for dictionary based classifiers. Previously, if series length was less than min_length (default 10), no classifiers were built and they failed on predict(). Now a ValueError is raised in fit()


TonyBagnall:
adds a check to dictionary classifiers to make sure min_window is admissible for a given problem

PR #635, opened 2021-01-28
Create new factory method for forecasting via reduction

Lovkush-A:
#### Reference Issues/PRs
See #220.

Created new factory method that combines the two current factory methods into one.

Is this new factory method tidier, for both users and contributors?

##### For all contributions


mloning (commented):
Sorry for the delay @Lovkush-A - looks all good to me. I left one minor comment. 

I'm happy to include this!

Lovkush-A:
Should I add tests for this factory method?

Lovkush-A:
Also, technical question about github, could I accept your suggested change and have auto-formatting within github?

mloning:
Hi @Lovkush-A - yes, if you want to add a test that would be much appreciated!

Lovkush-A:
@mloning
Testing. OK, I will add some sort of tests. I just looked and there is currently only one test (of pipelines) for the compose subfolder in forecasting. Does that mean we don't know whether all the other stuff in compose actually works?  Also, the test for the pipeline does not use pytest, unlike the other tests I have seen for forecasting. Should I create an issue for lack of testing in compose?

Formatting. Alright, good to know.

Next step. I still need to do multioutput! Multioutput should be straightforward though, I think I can copy and paste the direct reducer and change a few lines.  I am happy to work on exogenous variables after that. For exogenous variables, am I correct in thinking that we just need to add extra features to the dataframe - ensuring we line up the rows appropriately?

mloning:
@Lovkush-A oh yes forgot about multi-output! 

Regarding exogenous features yes, that's right, I had an idea how to restructure the code (see #390).

Regarding testing, a simple test that calls the function with different input arguments would be sufficient for now.

mloning (approved):

mloning:
@Lovkush-A let me know if you're happy with this to be merged, will leave it open a bit longer in case others want to comment.

Lovkush-A:
@mloning Not ready for a PR yet. I have written some tests but am struggling to get VSCode's python testing to work. I will push another commit later today with the tests later today.

Lovkush-A:
Are there any time series regression algorithms? I ask because I would like to test the ts-regressor scitype of the factory method.

mloning:
@Lovkush-A there is a `TimeSeriesForestRegressor` in the `regression` module I believe, otherwise using a pipeline with a `Tabularizer` and any regressor from scikit-learn should also work.

Lovkush-A:
@mloning I used pipeline with tabularizer. Satisfying when you finally get all the tests to come up green! As far as I am concerned, this is ready to be pulled in.

mloning (approved):
Thanks @Lovkush-A - will merge now!

PR #636, opened 2021-01-29
Euclidean distance for KNNs

goastler:

This implements ED to be used with the KNNs. Sk-learn's current ED doesn't work with our KNN.

#### Any other comments?

This is not implemented in cython, unlike most of the other distance functions. ED can be performed entirely using numpy operations, therefore cython becomes redundant. I've ran benchmarks and confirmed this is the case.

##### For all contributions

##### For new estimators


TonyBagnall (approved):

mloning (approved):
Thanks @goastler, happy to merge this! 

Just as a note, none of this code is tested as far as I know, so hopefully we don't break more things while we're trying to fix it.

Issue #603, opened 2021-01-10
Is there any imputation method in sktime?


zhenlingcn:
Currently, I'm encountering a prediction problem with some missing values. In the current implementation of sktime, when there is a missing value in the timing window, sktime will output nan. However, rather than tuning the size of the timing window, I prefer to fill these missing values. Thus, is there any method to do so?

TonyBagnall:
hi, we do not currently have imputation methods in sktime that I am aware of, sorry. We would very much like to have some though!

fkiraly:
You can get imputation through composition for any reduction method to a `sklearn` method, such as `DirectRegressionForecaster` - simply wrap a pipeline whose first element is a `sklearn` imputer, e.g., `SimpleImputer`.

Depending on your data format, you may also consider using `transformations.panel.interpolate` which can be used as imputer depending on the setting.

As @TonyBagnall, contributions are welcome - e.g., an imputation transformer based on the interpolator.

Let's maybe briefly also check with @mloning whether we didn't forget another potential solution to your problem.

TonyBagnall:
thanks Franz, my knowledge of sktime is localised to the classification side, I am not really best place to comment .... so probably shouldnt :)

mloning:
@zhenlingcn is this in the context of forecasting?

zhenlingcn:
Yes, the context is forecasting. I can use pandas to simply fill in the missing values, such as filling in the missing values by mean value. However, the filling function provided by pandas is rather simple. In fact, I hope I can use more advanced filling methods, such as linear or non-linear interpolation.

fkiraly:
Hm, I thought interpolators (padding, last value carry forward, linear/smoothing) were on the forecasting wishlist #220 or some roadmap, but I can't find them. Should definitely be there, since there are time series specific missing value handling strategies that you won't get from scikit-learn. @mloning, @TonyBagnall, do we have this recorded in some roadmap issue?

mloning:
There is a related issue on handling missing values #69

mloning:
Linear and non-linear interpolation are also provided by numpy, SciPy or pandas. See for example: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.interpolate.html

[...]

Issue #69, opened 2019-05-21
How to handle missing values


mloning:
* transformers for distribution of missing values (see #6)
* adapting existing algorithms to internally handle missing values if possible
* add details about missing values in meta-data

fkiraly:
just to add from discussion:

there could be two chains of transformers to do this, depending on whether missingness is 
(a) explicit (at time t, measured value NA) or (b) not (simply not measured, irregularly sampled)

(a) [by series] 1. apply binning (missing Y/N, "is na" logical), series-to-primitive 2. table/histogram, series-to-primitive (or series-to-bag)
(b) [by series] 1. get time points (series-to-sequence, "forget values), 2. first differences, 3. histogram/binning (or 5-num summary), series-to-primitive

TonyBagnall:
I'm implementing a transformer to simply interpolate missing in the simplest case

fkiraly:
I noticed this may be dangerous: while ok in the TSC/TSR case, in forecasters it may leak information "from the future". Perhaps let's add a boolean trait "temporal": if it's true, it doesn't leak information from the future.

aiwalter:
solved by #637 . I think it can be closed then.

mloning:
Thanks @aiwalter

PR #637, opened 2021-01-29
Added Imputer for missing values

aiwalter:
#### Reference Issues/PRs

#69 #603 

Transformation of `pd.Series` to impute missing values (`np.nan`) with different imputation strategies.

No

The imputation with `method="drift"` might be renamed to `"trend"`, however I adopted this naming from `sktime.NaiveForecaster`, so maybe we can rename it there also?

#### Any other comments?

1. I might implement further imputation  strategies, if s.o. has more ideas of other methods. 
2. I might also add an example to the notebook. 
3. What about imputation of `pd.DataFrame`, should this also be considered?

Example:
```python

y = load_airline()

y[0] = np.nan
y[4] = np.nan
y[5] = np.nan
y[8] = np.nan

# with trend
imputer = Imputer(method="drift")
imputer.fit_transform(y)

# with any forecaster
imputer = Imputer(method="forecaster", forecaster=TBATS(sp=12, use_trend=True, use_box_cox=False))
imputer.fit_transform(y)
```

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/transformations/series/impute.py
+++ b/sktime/transformations/series/impute.py
@@ -7,10 +7,10 @@ __all__ = ["Imputer"]
 from sktime.transformations.base import _SeriesToSeriesTransformer
 from sktime.utils.validation.series import check_series
 from sktime.forecasting.trend import PolynomialTrendForecaster
+from sklearn.utils import check_random_state
+
 
 import numpy as np
-import pandas as pd
-import random
 
 
 class Imputer(_SeriesToSeriesTransformer):
@@ -35,8 +35,8 @@ class Imputer(_SeriesToSeriesTransformer):
     value : int/float, optional
         Value to fill NaN, by default None
     forecaster : Any Forecaster based on sktime.BaseForecaster, optinal
-        Use given Forecaster to impute by insample predictions. Before
-        fitting, data missing data is imputed with method="ffill"/"bfill"
+        Use a given Forecaster to impute by insample predictions. Before
+        fitting, missing data is imputed with method="ffill"/"bfill"
         as heuristic.
     random_state : int/float/str, optional
         Value to set random.seed() if method="random", default None
@@ -108,7 +108,7 @@ class Imputer(_SeriesToSeriesTransformer):
             z = z.interpolate(method=self.method)
         else:
             raise ValueError(f"method {self.method} not available")
-        return pd.Series(z)
+        return z
 
     def _check_method(self):
         if (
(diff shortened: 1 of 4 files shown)


mloning:
A fix for the failing master is on its way: #643

aiwalter:
@mloning could you have a look, if you find time? :) Also regarding above comments

sktime/transformations/series/impute.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def _get_random(self, z):
+        """Create a random int or float value
+
+        :param z: Series
+        :type z: pd.Series
+        :return: Random int or float between min and max of z
+        :rtype: int/float
+        """
+        random.seed(self.random_state)
```
mloning, line 145: Can we use `check_random_state` from scikit-learn and numpy's random state object here? So something like this works: 
```python
rng = check_random_state(self.random_state)
rng.randint(z.min(), z.max())
```

mloning (commented):
Hi @aiwalter - looks all good except for one minor comment regarding the random state. 

All tests pass again on master, so should we able to merge this. 

Please add this to the docs too, in `docs/source/api_reference/transformations.rst`.

sktime/transformations/series/impute.py
aiwalter: @mloning thx, this is much better  👍🏻 ! Because otherwise doing `method="random"` would impute all `np.nan` with the same value. So now, if user sets seed, there are different values imputed

aiwalter:
I will add docs too, sry for re-requesting review, I forgot about it 😞

mloning:
@aiwalter no worries, ping me when you're done with the docs!

aiwalter:
@mloning from my side we can merge now

sktime/transformations/series/impute.py
```diff
@@ -0,0 +1,150 @@
...
+        """Create a random int or float value
+
+        :param z: Series
+        :type z: pd.Series
+        :return: Random int or float between min and max of z
+        :rtype: int/float
+        """
+        rng = check_random_state(self.random_state)
+        # check if series contains only int or int-like values (e.g. 3.0)
+        if (z.dropna() % 1 == 0).all():
```
mloning, line 147: Perhaps we should use data type checks here rather than inspecting the actual values.

mloning (approved):
Great - looks all good from my side. I left one minor comment, but happy to merge as it is. We can address this later if necessary. 

Will leave it open for a bit longer and then merge later today.

mloning:
Now merged - thanks @aiwalter! 🎉

PR #643, opened 2021-02-01
Pin NumPy 1.19

mloning:

* Fixes NumPy type-errors for v1.20
* Windows compilation still fails, so pinning NumPy to 1.19 for now


mloning:
@TonyBagnall this should fix the failure on master

Issue #403, opened 2020-09-28
Implement multi-output reduction strategy for ReducedRegressionForecaster


mloning:
We currently support the following strategies for reduction from forecasting to regression (see this [file](https://github.com/alan-turing-institute/sktime/blob/master/sktime/forecasting/compose/_reduce.py)): 

- [x] direct
- [x] recursive
- [ ] dirrec (see #226)
- [ ] multi-output (see #440)

For an overview of forecasting strategies for multi-step forecasting horizons see this [paper](https://d1wqtxts1xzle7.cloudfront.net/41231210/Machine_Learning_Strategies_for_Time_Ser20160114-28513-1pqgoe9.pdf20160115-19908-phg1xc.pdf?1452888125=&response-content-disposition=inline%3B+filename%3DMachine_Learning_Strategies_for_Time_Ser.pdf&Expires=1602491710&Signature=V6cHqybz3g9idBMTkbC4kwDMppSA4FnO9cd~R0VY8JVmVUKrK28MGRJlKhMfODdsSxcWsT5wvu2y3elqoES3ILHn1vlka20fO9z~dYmKB4oPsV04fG3lRhDe1Sx5QKgSmU4PMpWxa3bAwE8DKC4dlDcuLY7wequBDkTlItBFeYqOrJUhc91tMJxMpJ7gVtDzmAFyP58Fw5iXQN19fhCe2RoMtQ2uLRJrWvmlEdRvOhrffgnOR1g9N71Nt01I3X5~DON8gmKEmaGKYWv1wYg7JrSmLXnyZ2gVBUgdhERUxZ7VKSOMp-z0KFsuPV1bL7MbIZrHs7eHMnXK9v4vqT47ig__&Key-Pair-Id=APKAJLOHF5GGSLRBV4ZA).  This issue is about the multi-output strategy. Some models can directly predict multiple outputs, see e.g. LinearRegression:
```python
import numpy as np
from sklearn.linear_model import LinearRegression

y = np.random.normal(size=(10, 3))  # target has multiple columns
X = np.random.normal(size=(10, 5))

estimator = LinearRegression()
estimator.fit(X, y)
y_pred = estimator.predict(X)  # we predict multiple columns
print(y_pred.shape)
>>> (10, 3)
```
 
[...]

PR #659, opened 2021-02-05
Multioutput reduction strategy for forecasting

Lovkush-A:
#### Reference Issues/PRs

See #403 

Created MultioutputRegressionForecaster class via a _MultioutputReducer base class. Almost identical to Direct strategy.

- [ ] Have I understood Multitoutput reduction?
- [ ] Is the test sufficient?
- [ ] Do I need to modify any other files?
- [ ] Should I also have Multioutput for Time series?
- [ ]  Does this count as a new estimator?

##### For all contributions

##### For new estimators


mloning:
Thanks @Lovkush-A - looks very good, I'll take a closer look in the next few days!

sktime/forecasting/compose/_reduce.py
```diff
@@ -264,6 +265,76 @@ def _predict_in_sample(self, fh, X=None, return_pred_int=False, alpha=None):
...
+            The forecasters horizon with the steps ahead to to predict.
+        X : pd.DataFrame, optional (default=None)
+            Exogenous variables are ignored
+        Returns
+        -------
+        self : returns an instance of self.
+        """
+        self._set_y_X(y, X)
+        if X is not None:
+            raise NotImplementedError()
```
mloning, line 288:
```suggestion
            raise NotImplementedError("Exogenous variables `X` are not yet supported.")
```

sktime/forecasting/compose/_reduce.py
```diff
@@ -264,6 +265,76 @@ def _predict_in_sample(self, fh, X=None, return_pred_int=False, alpha=None):
...
+            window_length=self.window_length_,
+            step_length=self.step_length_,
+            start_with_window=True,
+        )
+
+        # transform data using rolling window split
+        X, Y_train = self._transform(y, X)
+
+        # fit regressor to training data
+        self.regressor.fit(X, Y_train)
```
mloning, line 310: What's the error here if the the regressor does not support multi-output? Is it user friendly or should we include some kind of check on the regressor if it is compatible with the multi output strategy?

mloning (approved):
Thanks @Lovkush-A looks all good to me! 

Only thing that we still have to do is to add it to the docs at `docs/source/api_reference/forecasting.rst`. 

Will merge it once that's done.

sktime/forecasting/compose/_reduce.py
Lovkush-A: I do not know, nor do I know how I would find out. I am trusting the error is user-friendly...

1. How do I find a regressor that does not support multi-output? I searched on google and somebody asked and got no reponse. https://datascience.stackexchange.com/questions/25829/scikit-learn-which-regressors-natively-support-multi-target-regression
2. How would we check if a regressor is compatible?

sktime/forecasting/compose/_reduce.py
Lovkush-A: Done (also in other parts of the file)

Lovkush-A:
Should I add multitoutput for time series too?

Lovkush-A:
Should I add multitoutput for time series too?

sktime/forecasting/compose/_reduce.py
mloning: I think ultimately we can only have heuristics for finding out whether an estimator supports multiouput or not. One thing to look for is the multi output mixin class or multiouput tag from scikit-learn: https://github.com/scikit-learn/scikit-learn/blob/95119c13af77c76e150b753485c662b7c52a41a2/sklearn/base.py#L758

mloning:
Thanks again @Lovkush-A, really appreciate your work on this! I think that adding support for exogenous variables is more important at the moment, leaving multioutput time series regression as a good first issue for other perhaps

sktime/forecasting/compose/_reduce.py
fkiraly: to add a detail: if a `sklearn` estimator does not support multioutput, you can always make it support it by wrapping it in `MultiOutputRegressor` or `MultiOutputClassifier` (a kind of reduction/composition).

Lovkush-A: Responding to the detail, if they estimator does not support multioutput, they should just use direct!

Lovkush-A:
Alright, works for me. I will go onto exogenous next.

sktime/forecasting/compose/_reduce.py
mloning: @Lovkush-A yes, wrapping in `MultiOutputRegressor` is the same as using `direct` - perhaps something we should add to the docstring. Similarly, there is [`RegressorChain`](https://scikit-learn.org/stable/modules/generated/sklearn.multioutput.RegressorChain.html) for recursive (or direct?) strategy.

Lovkush-A: I am not sure if docstring is correct place for that. Feels like the kind of thing that should go in the user-guide.

fkiraly: hm, this starts souding like an inconsistency!

I.e., if the wrapped estimator supports multi-output that is not identical with looping over output variables, what should the `direct` strategy do?

I think there are actually two direct strategies!
* `direct` - loops over the forecasting horizon, using single-output versions of the wrapped estimator; fits multiple single-output versions of the estimator internally
* `direct-multioutput` - uses the wrapped estimator's multi-output mode; fits a single multi-output estimator internally

While `direct-multioutput` after applying `MultiOutputRegressor` is always the same as `direct`, it is not true that `direct-multioutput` is always the same as `direct`, hence these are two different strategies for reduction!

fkiraly: @mloning implementing it this way, i.e., properly splitting up the two reduction strategies here, would also remove the need to check whether the estimator supports multi-output. The docstring can say that multi-output is assumed to be supported if you use `direct-multioutput` as your reduction strategy.

Lovkush-A: I do not follow. Your `direct-multioutput` sounds like `multioutput`.

fkiraly: ok, I'm confused now, probably entirely my fault ... what are the current options for strategies?
And which option did your original comment refer to ("if not support multioutput")?

Lovkush-A: We have three strategies. `direct`, `recursive` and (in this PR) `multioutput`. 

Markus was interested in knowing what error message arises if person X tries to use the multioutput strategy with a regressor R that does not support multioutput. You then suggested that X could wrap the regressor R using sklearn's `MultiOutputRegressor`, so you'd get something like: `MultioutputReductionForecaster(MultiOutputRegressor(R),...)`

I then said that you do not need this multitoutput as we can just use direct:
`DirectionReductionForecaster(R,...)`

fkiraly: Get it now.
So, I'm saying: in that case (regressor does not have multi-output support), I'd *still* wrap it in `MultiOutputRegressor`.
Why: because then overall behaviour is consistent and does not raise an error when using the `multioutput` strategy

Even if in that special case `multioutput` would give you the same behaviour as `direct`, it's fine - since it's not the case in general (just in that case). Both ways get you the expected result rather than an error message which wouldn't be clear why it is raised, if the user just starts with specifying either `multioutput` or `direct`.

mloning: Okay, I like that solution too. But the problem still remains that we need to check if the passed regressor support multioutput natively or not. And I would add a warning if it doesn't, since we change the input and wrap it automatically in the `MultioutputRegressor` we should let the user know about it.

mloning:
@Lovkush-A can you ping me again whenever you think this is good to go?

sktime/forecasting/compose/_reduce.py
Lovkush-A: My opinion is that using a non-multitoutput regressor in the multioutput regression should raise an error (as it currently will). Maybe there is a principle of software design I do not know about, but I think a user should be incentivised to use the tools properly. If they want direct strategy, they should choose it, if they want multioutput, they should choose an appropriate regressor.

Lovkush-A:
> @Lovkush-A can you ping me again whenever you think this is good to go?

@mloning I am fine with it.

sktime/forecasting/compose/_reduce.py
mloning: I agree @Lovkush-A, that's why I think raising the warning may be a compromise. I suggest to leave it as it is for now, merge the PR and wait for some user feedback, that also saves us from having to check the estimator type.

sktime/forecasting/compose/_reduce.py
```diff
@@ -57,15 +58,17 @@ def update(self, y, X=None, update_params=False):
         self : an instance of self
         """
         if X is not None or update_params:
-            raise NotImplementedError()
+            raise NotImplementedError(
+                "Exogenous variables `X` and `update_params` are not yet supported."
+            )
```
mloning, line 63: Once we have merged #660, we need to remove the `or update_params` here and update the error message.

mloning (commented):
Okay thanks @Lovkush-A - this is good to go with one caveat below depending another PR which I like to merge first.

sktime/forecasting/compose/_reduce.py
mloning: Now merged

mloning (approved):

mloning:
Thanks @Lovkush-A - now merged! 🎉

PR #660, opened 2021-02-06
Make update_params=true by default

### LINKED ISSUE #647: Change update method
https://github.com/sktime/enhancement-proposals/pull/8

The `update` method should 
* change to `update_params=True` as default, 
* implement a complete re-fit in the base forecaster as a fall-back option if not overridden in concrete implementations, but raise a warning if fall-back option is used

pabworks:
#### Reference Issues/PRs
Fixes #647. See also sktime/enhancement-proposals#8

- Makes `update_params=True` by default for all `update`-related forecasting methods.
- Implements a refit in the base Forecaster class when `update_params=True`

No

- [x] I have set the Theta forecaster `update_params=False` for the `super().update` call since the function then does a custom  routine if `update_params=True`, so, presumably, we don't want to refit. I have also set the `Deseasonalizer` transform to use the updated `y` as opposed to just the passed `y`
- [x] Assumption has been made that `update_predict` on a CV with `step_length>1` when `update_params=True` is not a desired use case for testing since the time-series is not continuous

#### Any other comments
The behaviour of the Forecaster w.r.t. update is as follows:
First, the Forecaster calls `fit` on a time period `y_a=y1,y2,...,yN`
Then, when we call `update`, the Forecaster uses the new information, e.g. `y_b=yN+1,...,yN+k` along with the old information to update the model. If a custom `update` routine is not implemented, then we simply fit a new model on `y=concat(y_a,y_b)`.

Original test failures have been isolated to when `step_length > 1`. In this case, when `update_params=True`, the model attempts to refit on a concatenation of the previous data (e.g. daily from "2000-01-01" to "2000-02-01") with the new data point as selected by the CV splitter. If the CV splitter has a `step_length > 1` then the frequency of the time series is no longer valid (e.g. daily up to "2000-02-01" then stepping multiple days into the future, e.g. "2000-02-07", before continuing at the same frequency)

The testing of `update_predict` where `step_length>1` and `update_params=True` has been left out, as a result.

#### PR checklist

- [x] set `update_params=True` for all methods
- [x] ensure all tests still pass when `update_params=False` (maintain original behaviour)
- [x] implement refit method in base Forecaster using original and update data
- [x] ensure all tests pass when `update_params=True` and `step_length=1`
- [ ] check desired behaviour for `update_params=True` and `step_length>1`

##### For all contributions

### DIFF
--- a/sktime/forecasting/compose/_reduce.py
+++ b/sktime/forecasting/compose/_reduce.py
@@ -13,6 +13,8 @@ __all__ = [
     "ReducedForecaster",
 ]
 
+from warnings import warn
+
 import numpy as np
 import pandas as pd
 from sklearn.base import clone
@@ -43,23 +45,32 @@ class BaseReducer(_BaseWindowForecaster):
         self.step_length_ = None
         self._cv = None
 
-    def update(self, y, X=None, update_params=False):
+    def update(self, y, X=None, update_params=True):
         """Update fitted parameters
 
         Parameters
         ----------
         y : pd.Series
         X : pd.DataFrame
-        update_params : bool, optional (default=False)
+        update_params : bool, optional (default=True)
 
         Returns
         -------
         self : an instance of self
         """
-        if X is not None or update_params:
+        if X is not None:
             raise NotImplementedError()
         self.check_is_fitted()
         self._update_y_X(y, X)
+        if update_params:
+            # default to re-fitting if update is not implemented
+            warn(
+                f"NotImplementedWarning: {self.__class__.__name__} "
+                f"does not have a custom `update` method implemented. "
+                f"{self.__class__.__name__} will be refit each time "
+                f"`update` is called."
+            )
+            self.fit(self._y, self._X, self.fh)
         return self
 
     def _transform(self, y, X=None):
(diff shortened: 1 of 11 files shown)


sktime/forecasting/compose/_reduce.py
```diff
@@ -43,23 +45,32 @@ def __init__(self, regressor, window_length=10, step_length=1):
         self.step_length_ = None
         self._cv = None
 
-    def update(self, y, X=None, update_params=False):
+    def update(self, y, X=None, update_params=True):
```
mloning, line 48: Thinking about this, can we not simply remove this here as we inherit `update` from the base forecaster? I probably overrode it to handle the error when X is passed, but this should already be handled in `fit` and `update` only works on fitted forecasters, so we should be able to simply remove it here, no?

mloning (commented):
Thanks @pabworks - looks all good! I left one comment which we can perhaps resolve in this PR too if it works. Otherwise happy to merge it!

mloning:
@pabworks I'll work on this today to make a few more minor changes.

mloning (approved):

pabworks:
Thanks @mloning! (I was off ill last week so didn't see your comments until this morning, sorry!)

mloning:
@pabworks no worries - thanks for doing most of the work on this one!

Issue #694, opened 2021-02-19
[BUG] ElasticEnsemble with revised TS-KNN


TonyBagnall:
the transpose change in KNN has effected the ElasticEnsemble. We are now effectively transposing twice. The quickest solution is for EE to transform the pandas into [n][d][m] numpy instead of [n][m][d] numpy, and to remove the coercion to pandas. Assigned to Aiden

TonyBagnall:
fixed by #697

PR #697, opened 2021-02-20
Fixed bugs mentioned in #694 

AidenRushbrooke:
#### Reference Issues/PRs

Fixed bug mentioned in #694 for Elastic Ensemble. Also fixed other bugs caused by recent changes to KNN classifiers.

Removed the double transposing issue, and removed coercion to pandas for ElasticEnsemble. Also removed algorithm parameter from classifiers used in cross validation parameter searches, and fixed some naming issues regarding usage of the term distance rather than metric.

No

Checking if I accidentally missed anything
#### Any other comments?

Ive compared the current elastic ensemble implementation to that from tsml for a few datasets and found they matched, but there may still be some issues regarding distance parameters being set through cross validation.
##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/classification/distance_based/_elastic_ensemble.py
+++ b/sktime/classification/distance_based/_elastic_ensemble.py
@@ -141,7 +141,7 @@ class ElasticEnsemble(BaseClassifier):
         self : object
         """
 
-        X, y = check_X_y(X, y, enforce_univariate=True, coerce_to_pandas=True)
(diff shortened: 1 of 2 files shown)


TonyBagnall (commented):
I think the reshape should only happen if EE is passed pandas. If it is given a numpy array it should just leave it alone as the format is assumed correct after check_X_y

AidenRushbrooke:
It should now only reshape if X is a pandas dataframe

TonyBagnall (approved):
looks good

PR #713, opened 2021-02-28
Fix cross validation for KNN classifiers

AidenRushbrooke:
#### Reference Issues/PRs

Related to issue #596.

Rolls back changes relating to the terms "distance_params" and "metric_params" to fix cross validation for KNN classifers.
Also includes setting the defaults for each parameter manually, if none have been provided.

No

Are there better ways to fix this issue while keeping the same naming conventions
#### Any other comments?

Due to the way the classifier is cloned during cross validation, the distance metric was not being set correctly. This is because the distance functions still look for the parameters in metric_params, following the scikit-learn implementation. There may be better ways of resolving this issue whilst keeping the same naming convention, but I found this method the cleanest of options.

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/classification/distance_based/_elastic_ensemble.py
+++ b/sktime/classification/distance_based/_elastic_ensemble.py
@@ -318,7 +318,7 @@ class ElasticEnsemble(BaseClassifier):
             best_model = KNeighborsTimeSeriesClassifier(
                 n_neighbors=1,
                 distance=this_measure,
-                distance_params=grid.best_params_["distance_params"],
+                metric_params=grid.best_params_["metric_params"],
             )
             preds = cross_val_predict(
                 best_model, full_train_to_use, y, cv=LeaveOneOut()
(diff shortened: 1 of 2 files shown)


TonyBagnall:
hi aiden, it fails one of my tests here, set "g":0.05 

 elif distance == wdtw_distance or distance == wddtw_distance:
            return {"g": 0}

AidenRushbrooke:
Hi, just made that change,
Not sure about the other issue, I think its to do with metric_params being set when the test expects it to be empty when using default values.

TonyBagnall:
> Hi, just made that change,
> Not sure about the other issue, I think its to do with metric_params being set when the test expects it to be empty when using default values.

could you send me the code you are using to cross validate on teams? ta

mloning:
@AidenRushbrooke @TonyBagnall The tests fail because the hyper-parameter specified in the constructor is not stored in the object.

PR #715, opened 2021-03-01
Multiplexer forecaster

### LINKED ISSUE #651: Implement Multiplexer
Arising from the call on Feb 2 on forecasting model selection & evaluation.

The multiplexer is a component that allows easy tuning for selection over a list of estimators with potentially distinct parameters

The high-behaviour is as follows.

The multiplexer takes a named list (dict) of components at initialization, `components`.
It also has a single `select` parameter that should be identical to one of the components´ names.

The behaviour of the estimator is identical to the component `components[select]`, the other components are ignored.

Why is this useful:
this is arising from a suggestion by @aiwalter that model selection should be possible over inhomogenous (not the same) estimators.

Better than the "naive" solution to write a custom grid search and random search is the solution to wrap a multiplexer in a tuner - this way, model selection can be done using any evaluation strategy, e.g., grid search, random search, Bayesian optimization, etc.

**Usage and implementation sketch**

```python
class Multiplexer
    def __init__(self, components, select=None):
        self.components = components
        self.select = select
        
    def fit(self, ...):
        self.components[select].fit(...)

    def predict(self, ...):
        self.components[select].predict(...)
        
    def update(self, ...):
        self.components[select].update(...)
```

To get model selection, one would then write the following:

```python
mpx = Multiplexer({"ARIMA": ARIMA(sth), "ExpS": ExpSmoothing(sth)])
grid = {"select": ["ARIMA", "ExpSmoothing"]}
gscv = ForecastingGridSearchCV(mpx, grid, ...)
```
A technical challenge might be to check whether this works well with nested `get_params`/`set_params`.
Easy solution might just be to "unpack" the list in the constructor.

koralturkk:
#### Reference Issues/PRs

Fixes #651

A Multiplexer class to provide a frame to ForecastingGridSearchCV for performing model selection over different sktime forecastors

No

##### For all contributions

##### For new estimators

#### Author and Legal Imprint:
Kutay Koralturk kutay.koralturk@daimler.com, Mercedes-Benz AG on behalf of Daimler TSS GmbH.


koralturkk:
@aiwalter

fkiraly:
@koralturkk, nice!

I had a quick look and noticed one thing:
you have `component_fit_params` to set the parameters of the wrapped estimators, but that's "wrong" in the sense of not interface compliant. Parameters of nested estimators should be initialized through *their respective constructors* (in the passed argument `components`), and support access via `get_params` and `set_params`.

Since `Multiplexer` inherits from `sklearn`'s `BaseEstimator` - via `_SktimeForecaster` - it may already automatically support the nested indexing via `get_params` and `set_params`, though I'm not sure whether it looks in dictionaries. So you may have to implement `get_params` and `set_params` yourself (you may want to check first whether it works out-of-the-box). If it's needed, you could do this directly, or with a mixin that extends `BaseEstimator` to look in `dict`s.

@mloning, any opinions/comments on what I just said?

koralturkk:
@fkiraly the reason for `component_fit_params` is that we are providing fit_params of each forecaster at fit function not in the estimator constructor. So far, it seems like a rare occasion where fit_params is given. I wanted to isolate it for the ease of use. Otherwise, if we want to provide that in components each component key will map to another dict which is consisted two keys, constructor and a fit params. Did I understand you correctly and do you have any comments on it? 

---
Kutay Koralturk kutay.koralturk@daimler.com, Mercedes-Benz AG on behalf of Daimler TSS GmbH.

fkiraly:
@koralturkk, ah, I see - it is not the construction params, but parameters passed in fitting.

Please ignore what I said, you are right.

Though, is there a plausible use case where not all `fit_params` are identical?

koralturkk:
is there a plausible use case where not all component_fit_params are identical?

@fkiraly Although, I have never used before, the fit functions of estimators which we wrap are different among each other. I would assume that fit_params should be different as well. 

could you explain your reason why you think it's better to pass it in the constructor than in fit? 

I dont have a strong argument against this. I will change it so that it gets aligned with the overall design

---
Kutay Koralturk kutay.koralturk@daimler.com, Mercedes-Benz AG on behalf of Daimler TSS GmbH.

fkiraly:
@koralturkk, I'm not sure whether this is the best solution - it seems to go against the "unified interface" pattern.
Why: typically, you would pass the parameters in `fit`. But for the composite (using `Multiplexer`), you have to pass them in the constructor. I.e., it is inconsistent between different forecasters where these would be passed.

Any thoughts, @mloning?

koralturkk:
@fkiraly As I have shared previously, I don't have a strong argument against it. I have made the changes to fit function to include fit_params.

fkiraly:
@koralturkk, sorry if I caused a misunderstanding - this was not a change request or criticism, but inviting you to a design discussion. I wanted to understand why you thought it would be better to put the argument in the constructor, rather than asking you to move it.

This is since I also see a potential issue with a nested structure passed to `fit`, because otherwise there is no nesting in the default interface.

koralturkk:
@fkiraly No problem at all, we had the same discussion with @aiwalter as well. I was proposing to have component_fit_params in constructor because of the extra hurdle in it. But both of you have suggested the same approach (fit_params in fit function). I agree that, this is more aligned with the overall design. 

The reason why I thought it would be a better idea is, when we provide fit_params in fit function, user has a higher chance to misinterpret it with other fit functions of forecasters. This may cause a confusion. To implicitly attract his/her extra attention, I wanted to move it to the constructor. Neither of the options are fully satisfactory to me. I would also wonder what @mloning and @aiwalter think about this as well.

aiwalter:
I would say `fit_params` should be given in `fit`, thats why they are called like this 🙂
___

docs/source/api_reference/forecasting.rst
```diff
@@ -104,6 +104,17 @@ BATS/TBATS
 
     TBATS
 
+Multiplexer
+----------
+
+.. currentmodule:: sktime.forecasting.bats
```
mloning, line 110: I'd add this to `sktime/forecasting/compose`

sktime/forecasting/base/_sktime.py
```diff
@@ -495,6 +495,64 @@ def _predict(self, fh, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
         raise NotImplementedError("abstract method")
 
 
+class _OptionalDictParamsMixin:
```
mloning, line 498: Do we still need this? If we want to keep it, this seems to be a major API change ...

sktime/forecasting/multiplexer.py
```diff
@@ -0,0 +1,158 @@
...
+from sktime.forecasting.base._base import DEFAULT_ALPHA
+from sktime.forecasting.base._sktime import _OptionalForecastingHorizonMixin
+from sktime.forecasting.base._sktime import _SktimeForecaster
+import copy
+
+__author__ = ["Kutay Koralturk"]
+__all__ = ["Multiplexer"]
+
+
+class Multiplexer(_OptionalForecastingHorizonMixin, _SktimeForecaster):
```
mloning, line 14:
```suggestion
class MultiplexerForecaster(_OptionalForecastingHorizonMixin, _SktimeForecaster):
```

sktime/forecasting/multiplexer.py
```diff
@@ -0,0 +1,158 @@
...
+        _forecaster_fit_params: float
+            Fit params for the forecaster that is
+            going to be used by the forecaster in the
+            _forecaster
+        """
+
+        self.components = components
+        self.select = select
+        self._check_components()
+        self._forecaster_fit_params = None
```
mloning, line 64: Yes, I agree with @fkiraly and @aiwalter, best to keep `fit_params` in `fit`

sktime/forecasting/multiplexer.py
```diff
@@ -0,0 +1,158 @@
...
+import copy
+
+__author__ = ["Kutay Koralturk"]
+__all__ = ["Multiplexer"]
+
+
+class Multiplexer(_OptionalForecastingHorizonMixin, _SktimeForecaster):
+    def __init__(
+        self,
+        components: dict,
```
mloning, line 17: I'd follow scikit-learn's composition API here and make it a list of tuples in which each tuple is of the form (name, estimator), see e.g. the Pipeline in scikit-learn. 

Instead of "components", I'd call it "estimators" for the same reason as above.

sktime/forecasting/multiplexer.py
```diff
@@ -0,0 +1,158 @@
...
+            forecaster that multiplexer will currently
+            forecast with.
+        _forecaster_fit_params: float
+            Fit params for the forecaster that is
+            going to be used by the forecaster in the
+            _forecaster
+        """
+
+        self.components = components
+        self.select = select
```
mloning, line 62: Can we make "select" more explicit, for example "selected_estimator"?

mloning (commented):
Hi @koralturkk - looks quite nice already, I left a few comments in the code.

koralturkk:
@mloning I believe its completed now, ready to be merged

fkiraly:
(you need to re-request a review for this to appear on Markus' review list - click the 
image top right)

sktime/tests/_config.py
```diff
@@ -181,6 +182,16 @@
         "max_q": 2,
         "seasonal": False,
     },
+    MultiplexerForecaster: {
+        "components": [
+            ("Naive", NaiveForecaster(strategy="mean")),
```
fkiraly, line 187: maybe also add other naive forecasters here?

sktime/forecasting/multiplexer.py
```diff
@@ -0,0 +1,171 @@
...
+import copy
+
+__author__ = ["Kutay Koralturk"]
+__all__ = ["MultiplexerForecaster"]
+
+
+class MultiplexerForecaster(_OptionalForecastingHorizonMixin, _SktimeForecaster):
+    def __init__(
+        self,
+        components: list,
```
fkiraly, line 17: should this perhaps be called `component_estimators`?

sktime/forecasting/multiplexer.py
fkiraly: are we still passing this in the init?

fkiraly (changes requested):
Nice work, great!

Some comments:
* it does not seem that you moved the `fit_params` from `__init__` to `fit` - have you pushed the newest commit?
* do we want to add the Multiplexer foreceaster to the 1st sktime tutorial? Might be a good idea, since AutoML is a highly sought after features
* But if we do this, I think the description of what it does needs to be much clearer! In terms of terminology, the section header should perhaps be "AutoML using tuning plus multiplexer"; then explain, what the code does - it does not do evaluation, but in fact model selection! Also explain tha the multiplexer can be combined with any tuner, e.g., the online hedge tuners.

koralturkk:
@fkiraly fit_params is passed in fit function. The _forecaster_fit_params is just an attribute of the object that the user can access. The value is either none or the fit_params being passed in the fit function. That was my idea. 

Regarding 1st sktime tutorial, can you please elaborate further. I am not sure if I understand fully what you mean by 1st sktime tutorial, I might be missing sth.

fkiraly:
>Regarding 1st sktime tutorial, I might be missing sth. Can you please elaborate further or share a link of the tutorial?

your changes to `01_forecasting.ipynb` in this PR

>The _forecaster_fit_params is just an attribute of the object that the user can access. The value is either none or the value associated with selected_estimator key of fit_params dict which is used during fitting the selected estimator. That was the idea.

But why is that necessary, and why do we want this to be an attribute of the estimator?

In my mind, the `fit_params` of `fit` and the `update_params` of `update` should just be the same (by convention/signature) as the same parameters for the selected estimator. Why would you pass parameters for estimators that are not selected, given that after construction they are never read?

koralturkk:
Because every estimator contained in components_estimator has a different fit_params dict. When we are performing a grid search, I thought it is the only way to search over different estimators with different fit_params to find the best performing one. As we are switching between different estimators we have to switch between different fit params that are associated with selected estimator. @fkiraly

fkiraly:
>Because every estimator contained in components_estimator has a different fit_params dict. When we are performing a grid search, I thought it is the only way to search over different estimators with different fit_params to find the best performing one. As we are switching between different estimators we have to switch between different fit params that are associated with selected estimator. 

Yes, I think you are right and this makes sense.
It should behave like `sklearn`'s `Pipeline`.

But, given this: 
* why are we then not simply unpacking the `fit_params` in `fit`?
I.e., `fit` internal hands `fit_params[estimator_key]` to the wrapped `fit`. If the key does not exist, the wrapped `fit_params` default to `None`.
* why is this not happening in `update` with `update_params`? I.e., why is `update_params` not a keyed `dict` but passed directly to the wrapped `update`?

fkiraly:
>I believe thats the way it is already. We unpack the fit_params with _update_forecaster_fit_params and fit_params[estimator_key] as _forecaster_fit_params

Yes, I see - that's a lot of edge case checks, but I guess "morally" it is fine, and probably safer to delegate to a designated private method as you do.

>update_params is a boolean, so basically it is a paramater to allow changes to parameters not to hyperparameters as far as I understand. @fkiraly

Yes, but you could still have this differ for the individual estimators.
Though, I'm also fine with it being equal for all individual estiamtors.

fkiraly:
This discussion resolves my questions around the `fit_params` leaves us with one major comment form my side:
better framing and presentation of autoML (via multiplexer) in the tutorial notebook

sktime/forecasting/multiplexer.py
mloning: I wouldn't store this in the object

sktime/forecasting/multiplexer.py
```diff
@@ -0,0 +1,173 @@
...
+        for component in self.component_estimators:
+            name, estimator = component
+            if not isinstance(estimator, _SktimeForecaster):
+                raise Exception(
+                    "Each component has to be an "
+                    "sktime forecaster object. "
+                    "Please check {} with name:{}".format(estimator, name)
+                )
+
+    def _check_fit_params(self, fit_params):
```
mloning, line 87: is this from scikit-learn? how do they do this?

sktime/forecasting/multiplexer.py
```diff
@@ -0,0 +1,173 @@
...
+            name, _ = component
+
+            if name not in fit_params.keys():
+                raise KeyError(
+                    "If you provide fit_params for models "
+                    "dictionary key of fit params need to "
+                    " match the associated component key."
+                )
+
+    def _check_selected_estimator(self):
```
mloning, line 101: please check out the meta-estimator [here](https://github.com/alan-turing-institute/sktime/blob/d4f3292e310670018658f56b239aac08ce240ab1/sktime/base/_meta.py#L13), we need to make sure that parameters are handled correctly too

sktime/forecasting/multiplexer.py
```diff
@@ -0,0 +1,173 @@
...
+            ]
+        else:
+            self._forecaster_fit_params = None
+
+    def _update_forecaster(self):
+        self._check_selected_estimator()
+        if self.selected_estimator is not None:
+            for name, estimator in self.component_estimators:
+                if self.selected_estimator == name:
+                    self._forecaster = copy.deepcopy(estimator)
```
mloning, line 128:
```suggestion
                    self._forecaster = clone(estimator)
```
where `clone` comes from `sklearn.base`

mloning (commented):
Hi @koralturkk - please find a few comments below

sktime/forecasting/multiplexer.py
koralturkk: I removed the _update_fit_params and moved all the content in to _check_fit_params following the sklearn's practice
given here: 

https://github.com/scikit-learn/scikit-learn/blob/95119c13a/sklearn/pipeline.py#L37

---

Kutay Koralturk kutay.koralturk@daimler.com, Mercedes-Benz AG on behalf of Daimler TSS GmbH.

sktime/forecasting/multiplexer.py
koralturkk: I believe you meant handling of get_params and set_params. I have extended the class with  `_HeterogenousMetaEstimator  `so that both methods work.

aiwalter (approved):
From my side all to do's are done and the example notebook looks also fine. Well done @koralturkk 
___

sktime/forecasting/compose/_multiplexer.py
```diff
@@ -0,0 +1,188 @@
...
+            _forecaster
+        """
+
+        self.forecasters = forecasters
+        self.selected_forecaster = selected_forecaster
+        self._check_components()
+
+        super(MultiplexForecaster, self).__init__()
+
+    def _check_components(self):
```
mloning, line 74: Could we make use of the existing checks from [`_check_forecasters`](https://github.com/alan-turing-institute/sktime/blob/77f34fc09d1bfa6abac613048fe5407c1be0c387/sktime/forecasting/base/_meta.py#L29) and [`_check_names`](https://github.com/alan-turing-institute/sktime/blob/77f34fc09d1bfa6abac613048fe5407c1be0c387/sktime/base/_meta.py#L64) if possible?

sktime/forecasting/compose/_multiplexer.py
```diff
@@ -0,0 +1,188 @@
...
+        for component in self.forecasters:
+            name, forecaster = component
+            if not isinstance(forecaster, _SktimeForecaster):
+                raise Exception(
+                    "Each component has to be an "
+                    "sktime forecaster object. "
+                    "Please check {} with name:{}".format(forecaster, name)
+                )
+
+    def _check_fit_params(self, fit_params):
```
mloning, line 89: Does this establish a new convention or is this something from scikit-learn? What's the expected format of the dictionary?

sktime/forecasting/compose/_multiplexer.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def _check_selected_forecaster(self):
+        component_names = [name for name, _ in self.forecasters]
+        if self.selected_forecaster not in component_names:
+            raise Exception(
+                "Please check the selected_forecaster argument provided "
+                " Valid selected_forecaster parameters: {}".format(component_names)
+            )
+
+    def _update_forecaster(self):
```
mloning, line 118:
```suggestion
    def _set_forecaster(self):
```
to avoid ambiguity with `update`

sktime/forecasting/compose/_multiplexer.py
```diff
@@ -0,0 +1,188 @@
...
+        """
+        self._set_params("forecasters", **kwargs)
+        return self
+
+    def _predict(self, fh, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
+        return self._forecaster.predict(
+            fh, X, return_pred_int=return_pred_int, alpha=alpha
+        )
+
+    def update(self, y, X=None, update_params=False):
```
mloning, line 183:
```suggestion
    def update(self, y, X=None, update_params=True):
```
we need a unit test for this

sktime/forecasting/compose/_multiplexer.py
mloning: Just read the docstring below, can we not just pass the dictionary with param, value pairs without the estimator name assuming the estimator is the "selected_estimator"?

mloning (commented):
Hi @koralturkk, looks much better now! A few more comments below.

sktime/forecasting/compose/_multiplexer.py
mloning: Now also read your comment above on sklearn's pipeline: https://github.com/scikit-learn/scikit-learn/blob/95119c13af77c76e150b753485c662b7c52a41a2/sklearn/pipeline.py#L249 - but still, can we not just pass the ones for the selected estimator?

mloning: Could we just follow scikit-learn's convention here instead of creating our own?

koralturkk: I think we need to create our own. Multiplexer switches btw diferent forecasters to find the best performing one. It may be that in this model selection process, user wants to run the process while taking into account different fit_params of different estimators. In scikit_learn, the fit_params argument changes only a param of a single estimator

sktime/forecasting/compose/_multiplexer.py
koralturkk: I decided to use _check_forecasters of `_HeterogenousEnsembleForecaster`

sktime/forecasting/compose/_multiplexer.py
koralturkk: Why? It is simply calling the update functions of other forecasters, as they already have a unit test for it, I believe there is no need for an additional one.

sktime/forecasting/compose/_multiplexer.py
mloning: @koralturkk yes you're right!

sktime/forecasting/compose/_multiplexer.py
```diff
@@ -0,0 +1,143 @@
...
+        ----------
+        _forecaster : Sktime forecaster
+            forecaster that MultiplexForecaster will currently
+            forecast with.
+        """
+
+        super(MultiplexForecaster, self).__init__(forecasters=forecasters, n_jobs=None)
+        self.selected_forecaster = selected_forecaster
+        self._forecaster = None
+        self._check_forecasters()
```
mloning, line 66: That should be called at the beginning of `fit` rather than `__init__`, the reason is that one may change the forecasters later through `set_params` without calling `__init__` again.

mloning (commented):
One last change and then we are ready merge this!

sktime/forecasting/compose/_multiplexer.py
koralturkk: I agree

mloning (approved):
Great work @koralturkk! Will leave this open a bit longer and merge in the next few days!

mloning:
@fkiraly since you requested changes, do you want to take another look before we merge?

fkiraly (approved):
Looks good to me now!

Minor comment: in the `01_forecasting.ipynb`, make sure you use a grave to surround names of code objects, i.e., write `fit` instead of fit. If you still have time to add this, would be appreciated, but it's so minor I'd also accept like this.

koralturkk:
@mloning I don't understand why the tests are failing right now. I didn't make any changes related to pystan

mloning:
@koralturkk yes failures seem unrelated to this PR, will have to look into this

mloning (approved):

mloning:
@koralturkk now merged! 🎉

PR #720, opened 2021-03-12
update baked in data to match archive

TonyBagnall:
some of these problems were rounded to four dps, whereas the archive data is not rounded. This PR resolves that for GunPoint, ArrowHead and ACSF1

<!--


TonyBagnall:
ah I see that this breaks the data loading tests, so maybe I will defer doing this and just note the difference

PR #725, opened 2021-03-15
Binder build fix

### LINKED ISSUE #662: Binder build fails
Binder seems to fail at the moment (thanks to Ammo for raising this issue on Discord), seems like one of the dependencies installations goes wrong. 

Perhaps we should add Binder to our CI to check if it builds (see #196).

ddorf:
Fixes #662 

Updated `.binder/requirements.txt` so that binder-build finishes successfully. Have tested this by using `https://mybinder.org/` to build from my fork.

The changed requirement for `tsfresh` avoids an attempt to install `numpy-1.20.1` which is incompatible with `python 3.6` installed on binder and causes failed build. Extra line to address separate issue around `llvm` (see below).

`llvm-lite-0.16.0` is the default version installed on Binder. Including the extra line avoids a doomed re-installation attempt out of the `/.binder` subfolder, where a certain `llvm-config` can't be found. Later in the build process, `llvm-lite` is successfully updated to `0.35.0`, presumably because the path to `llvm-config` works from the main directory.

This is my first pull request, so any feedback is very welcome!

##### For all contributions

##### For new estimators

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -239,7 +239,7 @@
         "maintenance",
         "doc"
       ]
-    },    
+    },
     {
       "login": "angus924",
       "name": "angus924",
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -231,6 +231,16 @@
       ]
     },
     {
+      "login": "ddorf",
+      "name": "Dan Mittendorf",
+      "avatar_url": "https://avatars.githubusercontent.com/u/28767641?v=4",
+      "profile": "https://github.com/ddorf",
+      "contributions": [
+        "maintenance",
+        "doc"
+      ]
+    },    
+    {
       "login": "angus924",
       "name": "angus924",
       "avatar_url": "https://avatars0.githubusercontent.com/u/55837131?v=4",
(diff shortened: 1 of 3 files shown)


fkiraly:
Nice! I guess we should upgrade the python version sometime.

Regarding this PR: given your explanation, this seems almost ready to go, except that the linting check does not run through (see above). Could you remove your trailing whitespaces?

mloning (dismissed):
Thanks @ddorf, much appreciated! 

Just to verify, I also checked the [binder build based on your branch](https://mybinder.org/v2/gh/ddorf/sktime/binder-build-fix?filepath=examples) - it worked as described. 

Happy to merge this once the code style changes are fixed. 

A related issue is to make the builds faster by prebuilding the Docker image as described here in case you're interested: https://github.com/jupyterhub/repo2docker-action#use-github-actions-to-cache-the-build-for-binderhub

ddorf:
Based on test 2 it looks like my suggested fix unfortunately causes a wider dependency issue. I've tried other ways to get the update of `llvm` working, but unless we can locate the `llvm-config` on the Binder instance it may be hard.

Can look into this some more, but it may not be as simple as first thought.

fkiraly:
Oh no - can you explain though what precisely the secondary issue here is? Not sure whether I understand what's going on with `llvm`.

mloning:
Closed in favour of #1266

Issue #464, opened 2020-11-02
[DOC] Developer guide (extension guidelines and scitype templates)


fkiraly:
From my comments in #420:

I think we very much need to write up template designs, type conventions, extension conventions. This information is typically not suitably covered by an auto-generated API reference, as it is distributed across classes - the generic template, and the conventions implemented in the concrete desendants. Sth like in sklearn dev guidelines could work, but I'd actually prefer to also have "formal" template definitions collected in one place. While the typing makes things clearer, it also makes the necessity to document the interface cleanly greater.

Opinions, @mloning, @TonyBagnall?

I'm not sure what the "final form" should be here, but I feel this needs to be manually written and curated, with any modification to template patterns and conventions.
Could live on the wiki, linked from user docs, or directly in the official docs.

Happy to contribute to this, but let's discuss first where it should live.

mloning:
Hi @utsavcoding, this may be interesting to you too

mloning:
Hi @fkiraly, I agree, we need better extension guidelines. I think we should write a "developer guide" in addition to the planned "user guide" and the scitype template should be part of that.

It should live in the docs. The wiki isn't very good for suggesting and reviewing changes collaboratively. I'm trying to slowly move away from the wiki for most things. 

We should merge in the user guide (#377) first though, even with only some basic information and place holders, that way other people can see it and may be interested in contributing. Having a work-in-progress warning at the top of the page should be fine. Similarly, I'm happy to merge in a basic structure of the developer guide. I believe some sections from CONTRIBUTING.md should eventually also be moved into the developer guide.

fkiraly:
@mloning, writing on the extension guidelines.

Quick question on forecasting extension pattern.
Can you confirm the following:

To implement a new forecaster, one should inherit from `_SktimeForecaster` and precisely one of the two `[...]ForecastingHorizonMixin`-s.

The extender needs to implement:
* `fit`, writing to `self`
* `_predict` (not `predict`!), reading from `self`
* `update` (if the model can be run in update mode)
* `update_predict` (if the model can be run in update mode)

Optional implements:
* `compute_pred_int` for prediction intervals
* `_update_predict_single` for single-step update/predict
* `get_fitted_params` for fitted parameter inspection

noting that some of these methods are implemented in `_SktimeForecaster` and will hence not raise `NotImplementedError`?

fkiraly:
@mloning, @big-o, can you please confirm?

[...]

PR #730, opened 2021-03-19
Add Croston's method

### LINKED ISSUE #729: Implement Croston's method for intermittent demand forecasting
# **Functions for  forecasting intermittent demand time series.**
Conventional methods for modeling(like ARIMA) are not suited for time series which is Intermittent in nature.
We need specialized methods to model such Demands/Time Series

**Methods like [Croston's](https://otexts.com/fpp2/counts.html) can be used to model such series**
A rough sketch of the API
```
croston=Croston(...)
croston.fit(y)
```
[Package in R](https://cran.r-project.org/web/packages/tsintermittent/index.html)

Riyabelle25:
Fixes #729 by adding requested functionality.

This is a draft pull request. 
I have implemented Croston's Method, by defining a function Croston() at sktime/forecasting/croston.py, which takes historical data as input and returns forecasted data in an array. 

Nope.

This is my time contributing here, so the following would be helpful:
- [ ] code readability
- [ ] modularity

##### For all contributions:

##### For new estimators


fkiraly (changes requested):
Nice work, @riyabelle25!

But please adhere to the `sktime` template interface - each `sktime` estimator should adhere to the `BaseForecaster` template.

We're currently working on extension guidelines, please see a draft here:
https://github.com/alan-turing-institute/sktime/wiki/Scitype-interface-specification:-forecaster
Most important but is to implement `fit`, `predict`, and `update` methods.

Let us know if you have any questions

Riyabelle25:
Cool, thanks a lot- have started working on it!

mloning:
Great start @Riyabelle25!

Riyabelle25:
Guys, I created a class Croston, inheriting from `BaseForecaster`, and implemented `Fit`, `Predict and` `Update` methods.
This work is still in progress, I've got to put the input checks in for each method, but thought I should give my updates and get some review.

mloning:
We renamed master to main today. You have to rename the branches locally:
```
git branch -m master main
git fetch origin
git branch -u origin/main main
```
More info here https://github.com/alan-turing-institute/sktime/issues/644

mloning (commented):
Hi @Riyabelle25, this is going in the right direction! 

A few comments:
* Check out the [NaiveForecaster](https://github.com/alan-turing-institute/sktime/blob/d4f3292e310670018658f56b239aac08ce240ab1/sktime/forecasting/naive.py#L19), especially how we use `self._set_y_X(...)` and `self._set_fh(...)` in `fit`, this takes care of setting the data, fh and the cutoff
* You don't need to inherit from sklearn's `BaseEstimator`, this is taken care of in `BaseForecaster`
* You do need to also inherit from `_OptionalForecastingHorizonMixin`, assuming that Croston does not need the forecasting horizon during fitting
* We need to separate the estimation of parameters (`fit`) and prediction based on the estimated parameters (`predict`)
* We need to add unit tests, ideally comparing results against the implementation in R
* I think we can ignore `update` for, that's taken care of in the BaseForecaster class
* the docstring should be placed just below the class definition rather than at the top of the file
* I'd suggest to rename alpha because it is already used for specifying prediction intervals

Let us know if you need help with any of this

fkiraly:
@mloning, @riyabelle25, the issues with inheritance templates may be caused by my specs - I should perhaps be clearer in "inheritance conventions" what concretely people should inherit from.

In any case, a "fill-in-template" file as suggested in #464 may be useful for the future.
Should I open an issue about that, @mloning?

Riyabelle25:
Yup, I've been debugging errors thrown by the unit tests since then. Should I try inheriting from ` _SktimeForecaster` 
rather than `BaseForecaster`? These errors are associated with faulty checks for invalid input.

mloning:
@Riyabelle25 oh yes that's right, please inherit from `_SktimeForecaster` and then implement abstract methods

Riyabelle25:
Hi guys, I've inherited from `_SktimeForecaster` and `_OptionalForecastingHorizonMixin`, took some time solving the thrown errors.
Though tedious, debugging helped me get familiar with the codebase, which was nice :))
I did all this keeping in mind the pointers @mloning gave earlier. Could someone take a look and give suggestions?

sktime/forecasting/croston.py
```diff
@@ -0,0 +1,107 @@
...
+            Smoothing parameter
+
+    Returns:
+    --------
+        forecast: array-like
+            Forecasted demand (on average per period) diff."""
+
+    def __init__(self, alpha=DEFAULT_ALPHA):
+        # hyperparameter
+        self.alpha = alpha
```
mloning, line 37: Could we rename `alpha` to avoid ambiguity with the alpha parameter for confidence intervals? `smoothing` should work.

sktime/forecasting/croston.py
```diff
@@ -0,0 +1,107 @@
...
+        # training data
+        self._y = None
+        self._X = None
+
+        # forecasting horizon
+        self._fh = None
+        # self.cutoff = None  # reference point for relative fh
+
+        # set _is_fitted to False
+        self._is_fitted = False
```
mloning, line 48: We should be able to remove these, as they are set by the base class.

sktime/forecasting/croston.py
```diff
@@ -0,0 +1,107 @@
...
+
+        return self
+
+    def predict(self, fh=None, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
+        if return_pred_int or X is not None:
+            raise NotImplementedError()
+        self.check_is_fitted()
+        self._set_fh(fh)
+
+        future_periods = len(self._fh.to_numpy())
```
mloning, line 68: My impression is that some of the variable naming makes it hard to read. For example:
* `cols` is what we usually call `n_timepoints`, 
* `future_periods` is `len_fh`,
* `d` is renamed, first it's just `y`, then it's the entire series with preallocated space for the predictions

sktime/forecasting/croston.py
```diff
@@ -0,0 +1,107 @@
...
+        p = 1  # periods since last demand observation
+
+        # Initialization:
+        first_occurrence = np.argmax(d[:cols] > 0)
+        q[0] = d[first_occurrence]
+        a[0] = 1 + first_occurrence
+        f[0] = q[0] / a[0]
+
+        # Create t+1 forecasts:
+        for t in range(0, cols):
```
mloning, line 88: Is this the training loop? If so, this should go into `fit`

mloning (commented):
Thanks @Riyabelle25 - great work! 

A few general comments:
* we need to add docstrings, please stick to the numpy docstring format, a reference to the literature would also be nice
* we need to add a unit test, it's probably a good idea to add that before we make further changes, I suggest to compare our results against results from R
* we need to separate fitting from prediction, currently `fit` is empty, which parameters are fitted on the training data? I'm not very familiar with this method, but it seems like the loop from line 88-98 is the training loop.

Let me know if you need help with any of this! 

I've left a few more detailed comments in the code below.

Riyabelle25:
Thanks, @mloning! I'll first resolve your comments on my code and update the PR, then work on adding the unit test

sktime/forecasting/croston.py
Riyabelle25: Did that, but got `TypeError: predict() got an unexpected keyword argument 'alpha'` from 20 failed unit tests

sktime/forecasting/croston.py
Riyabelle25: Yes, the block of code from 69-98 is handling the training of parameters! However (if I got this correctly); for shifting that block of code to `fit()`, we will have to make `fh` a required argument for `fit()` instead of optional.
As the training of parameters needs `len_fh` for the predictions

mloning: Or we here only loop until the end of the training set?

sktime/forecasting/croston.py
mloning: Not sure where the error comes from. Perhaps push the latest changes to see what's going on?

sktime/forecasting/croston.py
Riyabelle25: Oh thanks, figured this one out!

sktime/forecasting/croston.py
Riyabelle25: Yep, I just pushed the changes. They'll throw these tests `TypeError: predict() got an unexpected keyword argument 'alpha'`

Riyabelle25:
Can someone take a look at the failed tests to see why they're failing on renaming parameter `alpha` to `smoothing` please?
The code is functioning well otherwise (when named `alpha`) and clearing all tests

sktime/forecasting/croston.py
```diff
@@ -0,0 +1,95 @@
...
+                q[t + 1] = q[t]
+                a[t + 1] = a[t]
+                f[t + 1] = f[t]
+                p += 1
+        self.f = f
+
+        return self
+
+    def predict(
+        self, fh=None, X=None, return_pred_int=False, smoothing=DEFAULT_SMOOTHING
```
mloning, line 80:
```suggestion
self, fh=None, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA
```
This is a different alpha! This is why we needed to rename the alpha in the constructor (`__init__`) to avoid the confusion!

mloning:
You can see the failing tests here: https://github.com/alan-turing-institute/sktime/pull/730/checks?check_run_id=2222563283 - some of them return (somewhat) helpful error messages and hints about what needs changing. Does that help?

Riyabelle25:
Yeah, I've been piping the output of `pytest sktime/sktime/forecasting` to an output.log file, to debug better :smile:

mloning (commented):
Great work @Riyabelle25! 

At this point, I'd look for a simple data set, compute some results in R and write a unit test to check the output from the Python implementation against the results from R for a number of different train/test splits. I'd look here: https://otexts.com/fpp3/counts.html. You can either add a data set to sktime in `sktime/datasets/` or generate some toy data.

Riyabelle25:
Sounds good. I'll resume work next week, my mid-semester exams are right around the corner now:"(

Riyabelle25:
That turned out to be one loong week, sorry :disappointed_relieved:! Was unwell for a while there.

I have computed results for both the R package and my Python method implementations; using the **Australian Pharmaceutical Benefits Scheme (`PBS`)** [dataset](https://otexts.com/fpp3/counts.html) (which is used in the fpp3 book example itself).

For starting off, I worked in a [Google Colab notebook](https://colab.research.google.com/drive/1bqRMzqNH5FTZ8U-kSCsPkaXGzyt5Wj33?usp=sharing) as it's easier to work in both R and Python to compare results- do check them out!

My aim now is to write a unit test for sktime as @mloning suggested. 
However, how do I write a python unit test that compares the results between the R package (since it's written in R) and sktime's Croston forecaster class? 
Do we have similar pre-existing unit tests from other forecasters which I could maybe study?

mloning:
@Riyabelle25 would love to get this merged. For the tests against the R implementation, take a look at how @GuzalBulatova implemented the test here: https://github.com/alan-turing-institute/sktime/blob/8f15cc3b400ec5d557ddd6de95ae9aba89770168/sktime/transformations/series/tests/test_boxcox.py#L38

Riyabelle25:
Yay, thanks! I'll work on this now

RNKuhns:
@Riyabelle25 I've got a couple minor comments on this one. 

See the individual comments, I've put in the code. But these generally fall into 2 categories:
1. Updating the docstring
2. Whether the `predict()` method should accept the smoothing parameter and also whether some of the functionality there should actually be in `_predict()` method called by `predict()` in `_SktimeForecaster`

@mloning and @fkiraly can you confirm about the `predict()` and `_predict()` part? I believe @fkiraly might be updating the base forecasting object(s) currently.

sktime/forecasting/croston.py
```diff
@@ -0,0 +1,101 @@
...
+from sktime.forecasting.base._sktime import _OptionalForecastingHorizonMixin
+
+
+DEFAULT_SMOOTHING = 0.05
+DEFAULT_ALPHA = 0.05
+
+
+class Croston(_OptionalForecastingHorizonMixin, _SktimeForecaster):
+    """
+    Implementation of Croston's Method
```
RNKuhns, line 14: I think the docstring needs to be updated. See info on [NumPy docstrings](https://numpydoc.readthedocs.io/en/latest/format.html). 

Instead of "Implementation of Croston's Method" followed by underscrores, that probably needs to be a short one line summary.

sktime/forecasting/croston.py
```diff
@@ -0,0 +1,101 @@
...
+        self.f = f
+
+        return self
+
+    def predict(
+        self,
+        fh=None,
+        X=None,
+        return_pred_int=False,
+        smoothing=DEFAULT_SMOOTHING,
```
RNKuhns, line 86: Predict should probably not accept `DEFAULT_SMOOTHING`. Shouldn't that get referenced from `self.smoothing` which is already set in the class constructor as opposed to a parameter being passed to `predict()`? 

Also, the default `predict()` method in `_SktimeForecaster`  includes

```python
        self.check_is_fitted()
        self._set_fh(fh)
        return self._predict(self.fh, X, return_pred_int=return_pred_int, alpha=alpha)
``` 

To follow the `_SktimeForecaster`  convention, shouldn't the rest of the functionality be included in `_predict()`

```python 
len_fh = len(self._fh.to_numpy())
f = self.f        
# Predicting future forecasts:        
y_pred = np.full(len_fh, f[-1])        
index = self.fh.to_absolute(self._cutoff)        
return pd.Series(y_pred, index=index)
```

sktime/datasets/data/PBS_dataset/PBS_dataset
```diff
@@ -0,0 +1,205 @@
+Month,Scripts
```
mloning, line 1: Please remove this file (it's the same as the CSV file)

sktime/forecasting/tests/test_croston.py
```diff
@@ -0,0 +1,35 @@
...
+# test the Croston's Method against the R package
+
+
+@pytest.mark.parametrize(
+    "fh, r_forecast",
+    [
+        (10, 0.8688921),
+    ],
+)
+def test_Croston_against_r_implementation(fh, r_forecast):
```
mloning, line 16: This one looks good, can we also iterate over one or two additional smoothing parameters?

sktime/forecasting/tests/test_croston.py
```diff
@@ -0,0 +1,35 @@
...
+        demand=ts(y)
+        forecast <- croston(y,h = 10)
+    Output:
+        0.8688921
+    """
+    y = load_PBS_dataset()
+    forecaster = Croston(smoothing=0.1)
+    forecaster.fit(y)
+    y_pred = forecaster.predict(fh=fh)
+    np.testing.assert_almost_equal(y_pred.item(), r_forecast, decimal=4)
```
mloning, line 35: Why only `decimal=4`? Could we see where the deviations come from? Perhaps they are already resolved when fixing the comments pointed out above by @RNKuhns

mloning (commented):
I agree with the review by @RNKuhns, added a few additional comments below. After that it should be good to be merged!

sktime/forecasting/tests/test_croston.py
Riyabelle25: Sure!
Shall I type in another method in this test file to test against different smoothing parameters?

sktime/forecasting/tests/test_croston.py
Riyabelle25: Oh alright. Which higher value of decimal should I consider?

Riyabelle25:
Thanks @RNKuhns for your detailed review!
I shall update the docstrings, and resolve the comments as it goes

sktime/datasets/data/PBS_dataset/PBS_dataset
Riyabelle25: Done!

Riyabelle25:
Locally, the tests are passing just fine, but for some reason the build is failing on pushing with this error:
 `FileNotFoundError: [Errno 2] No such file or directory: '/opt/_internal/cpython-3.6.13/lib/python3.6/site-packages/sktime/datasets/data/PBS_dataset/PBS_dataset.csv'`
I don't understand why though- the commits pushed do have the PBS_dataset.csv in the desired location, and I didn't get this error while running pytest locally..

sktime/forecasting/tests/test_croston.py
mloning: I'd add it to the pytest parametrisation, perhaps also for different fhs:
```python
@pytest.mark.parametrize("smoothing, fh, r_forecast",  [(0.1, 10, 0.8688921), (0.5, 5, ...), (...), ...])
```
Three different cases in total should be enough.

sktime/forecasting/tests/test_croston.py
mloning: I'd try fixing the suggestion by @RNKuhns above first in case smoothing isn't properly used inside predict. Highest number of decimals that pass the test, 4 should be fine in the end.

sktime/forecasting/tests/test_croston.py
```diff
@@ -0,0 +1,36 @@
...
+import pytest
+import numpy as np
+
+# test the Croston's Method against the R package
+
+
+@pytest.mark.parametrize(
+    "smoothing, fh, r_forecast",
+    [
+        (0.1, 10, 0.8688921),
```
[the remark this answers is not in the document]
Riyabelle25: Ah, varying fh would still give the same forecast point value; 
Croston's method does not compute Forecast distributions (This model does not have an underlying stochastic model)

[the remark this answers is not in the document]
mloning: Yes but worth testing that our implementation does that too

sktime/datasets/base.py
```diff
@@ -755,3 +756,49 @@ def load_uschange(y_name="Consumption"):
...
+    References
+    ----------
+    ..fpp3: Data for "Forecasting: Principles and Practice" (3rd Edition)
+    """
+
+    name = "PBS_dataset"
+    fname = name + ".csv"
+    MODULE, filename = os.path.split(os.path.realpath(__file__))
+    # MODULE variable to be removed once dataset merged with sktime package
+    path = os.path.join(MODULE, DIRNAME, name, fname)
```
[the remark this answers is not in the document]
Riyabelle25: Thanks @mloning. Was worrying my head over this silly 404 error :relieved:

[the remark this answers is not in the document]
mloning: Yes perhaps we should add a short paragraph to the developer's guide: 
* website: https://www.sktime.org/en/latest/developer_guide.html
* source files: https://github.com/alan-turing-institute/sktime/tree/main/docs/source/developer_guide

A need section on "Adding data sets":
* include CSV file or supported other format,
* add `load_<dataset-name>(...)` function to `sktime/datasets/base.py`
* add to `sktime/datasets/__init__.py`
* add to `sktime/sktime/datasets/setup.py`

sktime/forecasting/croston.py
```diff
@@ -0,0 +1,131 @@
+# -*- coding: utf-8 -*-
+import numpy as np
+import pandas as pd
+from sktime.forecasting.base._sktime import _SktimeForecaster
+from sktime.forecasting.base._sktime import _OptionalForecastingHorizonMixin
+
+
+DEFAULT_SMOOTHING = 0.1
```
mloning, line 8: I suggest to make this private `_DEFAULT_SMOOTHING` or if it's only used here just use the value directly without defining a variable for it

sktime/forecasting/croston.py
```diff
@@ -0,0 +1,131 @@
...
+DEFAULT_ALPHA = 0.05
```
mloning, line 9: isn't that defined already somewhere else, in `forecasting/base` perhaps?

sktime/forecasting/croston.py
```diff
@@ -0,0 +1,131 @@
...
+    [1]  J. D. Croston. Forecasting and stock control for intermittent demands.
+        Operational Research Quarterly (1970-1977), 23(3):pp. 289–303, 1972.
+    [2]  Forecasting: Principles and Practice,
+        Otext book by Rob J Hyndman and George Athanasopoulos
+    """
+
+    def __init__(self, smoothing=DEFAULT_SMOOTHING):
+        # hyperparameter
+        self.smoothing = smoothing
+        self.f = None
```
mloning, line 43: I suggest to make this private: `self._f`

sktime/forecasting/croston.py
```diff
@@ -0,0 +1,131 @@
...
+        super(Croston, self).__init__()
+
+    def fit(self, y, X=None, fh=None):
```
mloning, line 46: Can I ask you to refactor this one more time? We did some refactoring in the mean time, please see #912 and especially the extension template in `sktime/extension_templates/forecasting.py`.

sktime/forecasting/tests/test_croston.py
```diff
@@ -0,0 +1,36 @@
...
+        demand=ts(y)
+        forecast <- croston(y,h = 10)
+    Output:
+        0.8688921
+    """
+    y = load_PBS_dataset()
+    forecaster = Croston(smoothing)
+    forecaster.fit(y)
+    y_pred = forecaster.predict(fh=fh)
+    np.testing.assert_almost_equal(y_pred.item(), r_forecast, decimal=5)
```
mloning, line 36: why do we use `y_pred.item()`? instead I suggest to create a full array of expected values for comparison: `np.full(len(fh), r_forecast)`

mloning (commented):
@Riyabelle25 thank you for the latest changes! Here's another round of comments, let me know if you have any questions or need help!

sktime/forecasting/croston.py
Riyabelle25: yes found it in forecasting.base._base

sktime/forecasting/croston.py
```diff
@@ -0,0 +1,123 @@
...
+            Exogenous variables are ignored
+        Returns
+        -------
+        forecast : pd.series
+                   predicted forecasts
+        """
+        if return_pred_int or X is not None:
+            raise NotImplementedError()
+
+        len_fh = len(self._fh.to_numpy())
```
mloning, line 116:
```suggestion
        len_fh = len(self.fh)
```

mloning: That should work - the `fh` is exposed as a public attribute

sktime/forecasting/croston.py
```diff
@@ -0,0 +1,123 @@
...
+        f = self._f
+
+        # Predicting future forecasts:
+        y_pred = np.full(len_fh, f[-1])
+
+        index = self._fh.to_absolute(self._cutoff)
```
mloning, line 122: same for the `cutoff`
```suggestion
        index = self.fh.to_absolute(self.cutoff)
```

mloning (commented):
Two quick comments

mloning:
Hi @fkiraly, you initially suggested changes, I think with my two quick comments above I'm happy with this PR to be merged. Would you like to review again?

sktime/forecasting/croston.py
Riyabelle25: Oh, nice, the tests pass on removing .`to_numpy()` now
 When I had coded this back in April, it would fail for the case when `fh` was int

mloning (approved):

mloning:
Now merged @Riyabelle25 🎉

Issue #388, opened 2020-09-04
Proximity Forest


moradisten:
I've been developing a Proximity Forest Technique in Python based on the authors' project for my degree thesis when I found your toolkit by chance. The thing is, I was testing your Proximity Forest benchmark and I found it's not scalable. It takes too much time for simple datasets. I made a comparison between your Sktime PForest and mine, and I've found mine is much more efficient. 

You can check my repository if you want to have a look and don't hesitate if you want me to contribute or give you a hand: https://github.com/moradisten/PForests.git

[accuracies.pdf](https://github.com/alan-turing-institute/sktime/files/5173784/accuracies.pdf)
[time_exec.pdf](https://github.com/alan-turing-institute/sktime/files/5173786/time_exec.pdf)
<img width="1196" alt="comparisontable" src="https://user-images.githubusercontent.com/29915156/92222289-13588300-ee9f-11ea-88bd-b4187e3a59d5.png">

I'm also developing the LbEnhanced distance measure in Python, based on the project implemented by the original authors in Java: https://github.com/moradisten/KNN-LB.git

Best Regards :)

TonyBagnall:
hi, thanks for that, I'm tagging George in this who implements PF @goastler , he will take a look. Are you at Monash?

moradisten:
No :P. I'm currently finishing my degree thesis based on time series
classifiers comparison. I'm a computer science student in the University of
Murcia, Spain

Whatever you need, just say it :)

TonyBagnall:
ah great, I'll let Chang Wei know, he will be pleased :) We would welcome your contributing to sktime, I'll talk to George and Chang Wei and get back to you. We have a wrapped version of his original code and a ported version in tsml, the java toolkit, so we can do a good comparison.

moradisten:
Alright :)

moradisten:
By the way, pleased to meet you. I've read a lot of your articles, which
are pretty interesting and I've referenced some in my report

fkiraly:
@moradisten, nice to meet you - while you're here and given that you're already working on it for your thesis - would you be interested to implement an overarching distance/kernel interface for `sktime`, and composition patterns that involve distances/kernels? 
I imagine an interface similar to the GP/kernels interface in `scikit-learn`, and an updated version of #52 (which is a "dormant" piece of work currently).

moradisten:
Interesting. Tell me more about it

mloning:
[...]

PR #733, opened 2021-03-21
Uploading new Proximity forest version

moradabaz:
Uploading new version of Proximity Forest (Issue #388 )
As commented in the issue, the current implementation of Proximity Forest is too slow, even for small datasets, so I modified some code and added new functionality in order to make it faster.

Proximity Forests still contains same classes, but functionality is different and some methods are added, as well as the way to calculate the probabilities. 

Improve time of performance of a Proximity Forest classifier. So far, so good, I manage to reduce performance time, even though accuracy/score might have decreased a little bit, it's worth it. 

* The methods `ProximityForest.fit(X,y)` and `ProximityForest.predict_proba(X,y)` transform the Pandas Dataframe X into 3D numpy arrays in order to improve the management and time performance of data management.
* Some attributes of the classes have been removed because of their deprecation

My first pull request, please any feedback to make me improve my contribution

##### For all contributions

##### For new estimators

Kind Regards to the team


mloning:
Hi @moradisten thanks for the PR, there seem to be some merge conflicts, could you try to resolve them? Let me know if you need help. We also need to make sure that Proximity Forest passes the standard unit tests. We currently exclude it [here](https://github.com/alan-turing-institute/sktime/blob/77f34fc09d1bfa6abac613048fe5407c1be0c387/sktime/tests/_config.py#L86).

moradisten:
Alright, I'll take a look at it

TonyBagnall:
hi, thanks for this, could you PR this into contrib/distance_based in the first instance? I can then run it on all the archive data to make sure that there is no significant difference in accuracy and quantify any train and test time differences before it supersedes the old version. If it is significantly less accurate than the old version we could include it as an option/extra classifier, but I dont want to remove the old one if it is on average more accurate. thanks again. Tony

moradisten:
Alrigh, do you want me to PR right now or after I solve the conflicts?

Kind Regards

MatthewMiddlehurst:
Hi @moradisten, I think there may have been a bit of confusion about Tonys comment. It would be best if you PR it into the main branch as you originally intended, but as a separate file under sktime/contrib/distance_based/ so we still have both versions for now. 

This will make it easier for us to compare both versions results, and make a decision on how to integrate it further (i.e. replacing the current version or including as an alternate). This normally isn’t the case, but as you have shown both accuracy and time have quite large differences from the original. I am personally quite interested to see what has changed to cause this!

moradisten:
OK @MatthewMiddlehurst , I'm sorry about the misunderstanding, I've just changed the target branch

MatthewMiddlehurst:
No need to apologise. FYI the contrib directory is exempt from the unit testing as far as I am aware, so you may want to make sure that is all ok before making the move.

moradisten:
Oh yes, I forgot PF was excluded from the is excluded from unit tests.

moradisten:
Hi everyone,

After spending some time adapting my implementation to the sktime interface and researching why it's apparently better than the original one, I got to the conclusion that the main reason mine looks better it's because I used the DTW distance measure from the `dtaidistance` package. My implementation resulted to be as slow as the original one of the sktime when using another distances functions implemented by the contributors. But the DTW distance contained in the package I mentioned definitely improves my performances, at least in terms of time. 

However, the original Proximity Forest implementation doesn't improve its performance when the `dtaidistance` distance measure, in other words and making it simpler:

* My implementation and the original implementation are too slow when using the contributors distance measures they have implemented
* My implementation gets better time results when using the distance measures provided by `dtadistance` package, especially DTW.
* However, this package doesn't have any impact in the original implementation of Proximity Forest.

So, that's the conclusion of my research. I suggest considering this package I mentioned because its performance is really good. Here is the link: [https://pypi.org/project/dtaidistance/](url) 

```{python}
import dtaidistance.dtw as dtw
...
distance dtw.distance_fast(first, second)
```

Moving on, I uploaded my latest version so you can check it yourselves (Some feedback would be helpful) and I think my contribution in this issue is nearly done :P

TonyBagnall:
hi morad, thanks for looking into this. Currently we use cython based distance measures, but are thinking of moving over to numba, as it avoids the need for a c++ compiler. We are investigating how to improve performance. The dtai project is really good, there are a fair few dtw alternatives in python, but they are all really slow compared to java or c++ implementations. 

In terms of performance, our approach is to contract the classifiers. i.e. you give a maximum amount of time, build the best model you can. This can include limiting the number of trees and/or cases. We hope a stable PF version will be available over the summer.

moradisten:
Exactly @TonyBagnall , but as far as I've researched in this case, the dtai project is excelent and show great results in terms of time. Here are some plots of some datasets showing the accuracy and execution time using DTW distance measure provided by *dtai project*. Even with a slightly lower accuracy values, the time it takes to compute the distances is way lower than the original one. 
Just an example: Let's pick the `ItalyPowerDemand` dataset, and use a model of 100 trees and 5 candidates:
* Original Proximity Forest offers a **score value of 0.96** and an **execution time of 1682.03 seconds**
* My implementation using dtaidistance  offers a **score value of 0.873** and an **execution time of 13.29 seconds**
As you see, the original one offers an accuracy which is 17% better than my implementation, but the time it takes is more than 100 times higher than mine. So it's all up to you to decide whether it's better to chose a classifier that offers the best score value no matter the time it takes, or if it's better to have a classifier that runs faster

Another thing I've noticed in my implementation using dtai DTW is that the increase of trees and candidates doesn't necessarily improve the score even though it takes much time. 

I'll upload my changes so that you can be able to test it yourselves and here you have some pics of the scores and execution times of the datasets I run. 

I'll continue my research to see if I could improve the accuracy value

[Chinatown_score_results.pdf](https://github.com/alan-turing-institute/sktime/files/6216545/Chinatown_score_results.pdf)
[Chinatown_time_results.pdf](https://github.com/alan-turing-institute/sktime/files/6216547/Chinatown_time_results.pdf)
[ItalyPowerDemand_score_results.pdf](https://github.com/alan-turing-institute/sktime/files/6216548/ItalyPowerDemand_score_results.pdf)
[ItalyPowerDemand_time_results.pdf](https://github.com/alan-turing-institute/sktime/files/6216549/ItalyPowerDemand_time_results.pdf)
[Plane_score_results.pdf](https://github.com/alan-turing-institute/sktime/files/6216550/Plane_score_results.pdf)
[Plane_time_results.pdf](https://github.com/alan-turing-institute/sktime/files/6216551/Plane_time_results.pdf)
[SmoothSubspace_score_results.pdf](https://github.com/alan-turing-institute/sktime/files/6216552/SmoothSubspace_score_results.pdf)
[SmoothSubspace_time_results.pdf](https://github.com/alan-turing-institute/sktime/files/6216553/SmoothSubspace_time_results.pdf)

TonyBagnall:
hi @moradisten, we are trying to tidy up the PRs to stem the tide. Are you still working on this version of PF? If we are going to have it in I would like to see some equivalence experiments against the current version. We are currently reworking our distance measures to use numba instead of cython, so this may impact your work? It may be best to close this PR then open a new one once its reworked?

TonyBagnall:
ah sorry, Matthew just pointed out that this goes into contrib so its not overwriting anything. I'm happy to bring it in to there if all the tests pass, we can try it out then.

MatthewMiddlehurst (approved):
In contrib for testing so no issue assuming it passes the tests.

moradisten:
Hi everyone. Long Time no see :D.

Yes, those were my last changes, so feel free to run all the tests you want, detect anomalies and decide wether my version could be better or not. For anything else, just let me know. 

Kind Regards  :)

TonyBagnall (approved):

Issue #724, opened 2021-03-14
provide pypi wheels for MacOS < 10.15


fkiraly:
**Is your feature request related to a problem? Please describe.**
MacOS wheels should be provided for actively maintained versiosn of MacOS <10.15, see
https://github.com/alan-turing-institute/sktime/discussions/723

**Describe the solution you'd like**
According to @hoesler who raised this:
"Well, I don't think that I am really able to contribute to your build and publishing process without access to the CI/CD environment. I might be wrong, but I think all you need to do is to set the mac os x runner to an earlier OS version."

fkiraly:
@mloning FYI

TonyBagnall:
hey @fkiraly and @mloning can we close this?

fkiraly:
Not done, #748 is not merged - @hoseler4dh is either still working on this or has disengaged.

mloning:
Closed for now due to inactivity. I'm happy to revisit the idea if this is still relevant.

PR #748, opened 2021-03-24
Supporting macosx older than 10.15

freddyaboulton:
#### Reference Issues/PRs

Fixes #723 

Builds wheels for mac osx 10.14 and 10.15 in the deploy-to-pypi github action.

##### For all contributions

##### For new estimators


mloning (commented):
Thanks @freddyaboulton for the PR! Seems to work nicely! 

From what I understand, we need to change both the [build-and-test workflow](https://github.com/alan-turing-institute/sktime/blob/fd927f3c8109e2dd033f6747fc6a9c229f36f42e/.github/workflows/build-and-test.yml) and the upload-to-pypi workflow.

freddyaboulton:
@mloning I updated the build-and-test workflow! Thanks for the tip!

mloning (dismissed):
Thanks @freddyaboulton - looks all good to me now. 

Could you share a link where you found out about using `MACOS_DEPLOYMENT_TARGET`? Quickly searched for it but couldn't find any official documentation on it.

mloning:
Not sure about the build failures now, seem unrelated to this PR. Will look into this.

freddyaboulton:
@mloning Thanks for the approval and hunting down the pystan issue! 

I learned about the `MACOSX_DEPLOYMENT_TARGET` env variable from this [thread](https://github.com/actions/virtual-environments/issues/1256#issuecomment-770270252)!

fkiraly:
Fascinating ... https://en.wikipedia.org/wiki/Linus%27s_law

riddell-stan:
PyStan 3.0 just appeared which makes backwards incompatible changes, https://discourse.mc-stan.org/t/pystan-3-0-is-released/21500

You likely want to pin your version of PyStan to a version less than 3.

mloning:
Thanks @riddell-stan!

Pinned the version here: #751

freddyaboulton:
Thank you @mloning !

hoesler4dh:
Hey @freddyaboulton,
thanks for helping with this issue. I did a litte bit of experiments on my own and it seems like simply setting `MACOSX_DEPLOYMENT_TARGET` is not enough.
You actually have to compile python with this env var set. Here is an old blog post on this: https://lepture.com/en/2014/python-on-a-hard-wheel

I can confirm this works.
After installing python using pyenv on my mac 10.14 with `MACOSX_DEPLOYMENT_TARGET=10.13`, bdist_wheel was producing a 10.13 wheel.

For the workflow it might be sufficient to just install an official mac pkg, which seems to be compiled against 10.9:
https://www.python.org/ftp/python/3.8.8/python-3.8.8-macosx10.9.pkg

So I guess something like this could work:

```
    strategy:
      matrix:
        os: [ macos-latest ]
        python-version: [ 3.6.9, 3.7.9, 3.8.8 ]
    steps:
      - uses: actions/checkout@v2
      - name: Download Python
        run: |
          curl https://www.python.org/ftp/python/${{ matrix.python-version }}/python-${{ matrix.python-version }}-macosx10.9.pkg -o "python.pkg"
      - name: Install Python
        run: |
          sudo installer -pkg python.pkg -target /
```

hoesler4dh:
fixes #724

.github/workflows/build-and-test.yml
```diff
@@ -53,22 +53,22 @@ jobs:
     strategy:
       matrix:
         os: [macos-latest]
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [ 3.6.8, 3.7.9, 3.8.8 ]
```
mloning, line 56: Do we have to pin the exact Python version here? I'm assuming it's necessary for the URL in the curl command below?

mloning (commented):
Thanks for figuring this out!

Why is macOS 10.9 sufficient here if we want 10.14?

I'll try to get this in before the next release.

.github/workflows/build-and-test.yml
hoesler4dh: If I’m not mistaken, the 10.9 python distribution is the official release for all macos versions >= 10.9. So I was thinking, that we just could use that. If you want to limit to 10.14, we need to compile Python for the target.

Regarding the python version for the curl command, I don’t think omitting the patch version would work.

fkiraly:
@freddyaboulton , @hoesler4dh, this PR is still open - is this superseded, or is this still work in progress?
Or, would it be ready for review/merge?

mloning:
Closed for now due to inactivity. I'm happy to revisit the idea if this is still relevant.

PR #749, opened 2021-03-24
Upload source tarball to PyPI during releases

### LINKED ISSUE #607: Upload source code (.tar.gz) to PyPI
See [SO question](https://stackoverflow.com/questions/65648658/why-is-an-older-version-of-sktime-getting-installed-using-pip); .tar.gz files are only available for some earlier releases, but not the latest ones: https://pypi.org/simple/sktime/

Make sure we upload wheels but also .tar.gz files.

dsherry:
#### Reference Issues/PRs
Fixes #607 

Relates to #723 #748 (@freddyaboulton FYI)

I see sktime [was uploading a source tarball in v0.4.3](https://pypi.org/project/sktime/0.4.3/#files) and prior, but [has not done so since that version](https://pypi.org/project/sktime/0.5.3/#files).

As a result, running `pip install sktime` on platforms for which sktime has not built a wheel (like mac OS 10.14) will install that older version of sktime, v0.4.3.

Uploading a source tarball to PyPI will enable `pip`-installation of future sktime releases on platforms for which sktime didn't upload a wheel.

Nope!

Check out [the console output from the last run of the PyPI upload action](https://github.com/alan-turing-institute/sktime/runs/1844683304?check_suite_focus=true) (for the v0.5.3 release). You can see it ran `python setup.py sdist` meaning it created a source tarball, but that tarball was simply never uploaded to PyPI.

#### Any other comments?
I also suggest the project owners (@mloning ) manually upload a source tarball for the last version (v0.5.3), to allow users of more platforms to install the latest package version without having to wait for the next release. I believe this would do it, from looking at [the twine docs](https://twine.readthedocs.io/en/latest/):
```bash
git checkout v0.5.3
python3 setup.py sdist
twine upload dist/sktime-0.5.3.tar.gz
```

##### For all contributions

##### For new estimators

### DIFF
--- a/.github/workflows/upload-to-pypi.yml
+++ b/.github/workflows/upload-to-pypi.yml
@@ -54,6 +54,7 @@ jobs:
           python -m pip install --upgrade twine
           python -m twine upload -r pypi --non-interactive --skip-existing --verbose dist/*.whl
       - name: Upload source tarball to PyPI
+        env:
           TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
           TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
         # for simplicity only do the upload once
--- a/.github/workflows/upload-to-pypi.yml
+++ b/.github/workflows/upload-to-pypi.yml
@@ -52,7 +52,7 @@ jobs:
         run: |
           ls -l dist/*
           python -m pip install --upgrade twine
-          python -m twine upload -r pypi --non-interactive --skip-existing --verbose dist/*.whl
+          python -m twine upload -r pypi --non-interactive --skip-existing --verbose dist/*
       - name: Upload source tarball to PyPI
         env:
           TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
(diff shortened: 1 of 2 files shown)


.github/workflows/upload-to-pypi.yml
```diff
@@ -45,11 +45,19 @@ jobs:
...
           pytest --showlocals --durations=10 --pyargs --cov-report=xml --cov=sktime sktime
-      - name: Upload to PyPI
+      - name: Upload wheel to PyPI
         env:
           TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
           TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
         run: |
           ls -l dist/*
           python -m pip install --upgrade twine
           python -m twine upload -r pypi --non-interactive --skip-existing --verbose dist/*.whl
```
mloning, line 55: Would this work too? 
```suggestion
          python -m twine upload -r pypi --non-interactive --skip-existing --verbose dist/*
```

mloning (commented):
Thanks @dsherry - looks great! Quick question below:

.github/workflows/upload-to-pypi.yml
dsherry: @mloning yes this is a fine simplification! I figured it would be nice to speed up the job by avoiding the duplicate uploads, but simpler code is also great :) I'll add this suggestion now.

mloning (approved):

dsherry:
@mloning thanks for the approval! What should I do to get a second approval?

mloning:
@dsherry now merged - thanks for the PR! :tada:

Issue #196, opened 2019-11-12
Update CI / distribution


mloning:
A few useful change to improve our release/distribution cycle:

- [ ] include Changelog.md
- [ ] compile manylinux wheels (see #188)
- [ ] include Makefile to automise distribution workflow
- [ ] set up Travis and Appveyor to directly upload wheels to Github releases and PyPI

See this [issue](https://github.com/alan-turing-institute/the-turing-way/issues/746) for useful links.

xuyxu:
Hi @mloning, I would like to work on using `cibuildwheel` to simplify the workflow on package building. You can check our [package](https://github.com/LAMDA-NJU/Deep-Forest/blob/master/.github/workflows/build-wheels.yml) that uses cibuildwheel, it is quite convenient, and we can even build wheels for architectures like `aarch64`.

I have checked all supported python versions and platforms of sktime on [PyPI](https://pypi.org/project/sktime/#files), all of them can be built with cibuildwheel.

mloning:
@xuyxu sounds good

TonyBagnall:
hi @mloning Im doing a tidy up, 200+ issues triggers me. Can this be closed?

PR #752, opened 2021-03-26
Add doc-quality tests on changed files

mloning:
#### Reference Issues/PRs
#196 

* Adds pydocstyle checks on changed files to enforce numpy docstring format

* pydocstyle for docstring quality checks as a dependency in CI, not for installation/usage


mloning (commented):
Use available precommit hook instead of new CI script

PR #756, opened 2021-03-27
Modify tests for forecasting reductions

Lovkush-A:
#### Reference Issues/PRs

NA

Currently, tests check for relative consistency - do two different methods give the same answer.
Changed to give absolute consistency - do individual tests product the expected outputs. This was done in preparation for refactoring of reductions.

[ ] - Are the new tests good? Are they comprehensive enough?

##### For all contributions

##### For new estimators


mloning (commented):
Hi @Lovkush-A, thanks for the PR! 

* Could we separate the test fixture from the test logic? The logic seems to be always the same, we only change the estimators and expected data.   
* How did you generate these expected values? 
* I'd be useful to think about properties that these estimators should meet, simple cases where we know the expected solution (e.g. linear extrapolation as I added here: https://github.com/alan-turing-institute/sktime/pull/757)

Lovkush-A:
@mloning

1. Sure, I will try to combine them
2. The expected values were created by running the code locally and printing the outputs.
3. I will have a think. The tests in this PR do implicitly check for some internal consistency because many of them should produce the same output.

Lovkush-A:
@mloning The error in the tests is incomprehensible to me.
`ERROR: Failed building wheel for fbprophet`
`ModuleNotFoundError: No module named 'numpy'`

I googled and found these discussions which seem relevant, but 
the technicalities are beyond my knowledge.

https://github.com/facebook/prophet/issues/1551
https://github.com/facebook/prophet/issues/401

Any tips on what I can do sort out this issue?
Thanks!

mloning:
@Lovkush-A it should have been resolved on `main` - I synced your branch now, had to do with a new pystan release, one of the dependencies of fbprophet.

Lovkush-A:
Great, thanks for sorting that out so quickly.

mloning:
@Lovkush-A do you want to still integrate these tests? It would be worth checking if they still pass with the new changes.

Lovkush-A:
@mloning Yes, I was planning to update it to the newest api.

sktime/forecasting/compose/tests/test_reduce.py
```diff
@@ -381,3 +381,110 @@ def test_multioutput_direct_equivalence_tabular_linear_regression(fh):
...
+            DirectTimeSeriesRegressionForecaster(
+                make_pipeline(Tabularizer(), LinearRegression())
+            ),
+            EXPECTED_AIRLINE_LINEAR_DIRECT,
+        ),
+        (
+            MultioutputTimeSeriesRegressionForecaster(
+                make_pipeline(Tabularizer(), LinearRegression())
+            ),
+            EXPECTED_AIRLINE_LINEAR_DIRECT,
```
mloning, line 468: Since multioutput and direct are equivalent for linear regression, we could remove them. There is already a test that checks their equivalence. If you want to leave them in, I would add a comment stating that they are expected to be equivalent. Otherwise it may be confusing to see that we check multioutput against results for direct.

Lovkush-A: I have no strong thought either way; I added some comments.

mloning: I suspect these tests are reasonably fast, so let's leave them in.

mloning (approved):

mloning:
Thanks again @Lovkush-A - now merged! 🎉

PR #769, opened 2021-04-01
ThetaForecaster now works with initial_level

### LINKED ISSUE #692: ThetaForecaster does not work with "initial_level" different from None
**Describe the bug**
When calling
`forecaster = ThetaForecaster(initial_level=0.5, deseasonalize=True, sp=1)` then fitting via `forecaster.fit(y_train)` leads to the following error:

```python
`Traceback (most recent call last):

  File "<ipython-input-362-6ef348c52370>", line 1, in <module>
    forecaster.fit(y_train)

  File "/Users/muhlbach/opt/anaconda3/envs/experimental/lib/python3.8/site-packages/sktime/forecasting/theta.py", line 131, in fit
    super(ThetaForecaster, self).fit(y, fh=fh)

  File "/Users/muhlbach/opt/anaconda3/envs/experimental/lib/python3.8/site-packages/sktime/forecasting/base/adapters/_statsmodels.py", line 47, in fit
    self._fit_forecaster(y, X)

  File "/Users/muhlbach/opt/anaconda3/envs/experimental/lib/python3.8/site-packages/sktime/forecasting/exp_smoothing.py", line 79, in _fit_forecaster
    self._forecaster = _ExponentialSmoothing(

  File "/Users/muhlbach/opt/anaconda3/envs/experimental/lib/python3.8/site-packages/pandas/util/_decorators.py", line 199, in wrapper
    return func(*args, **kwargs)

  File "/Users/muhlbach/opt/anaconda3/envs/experimental/lib/python3.8/site-packages/statsmodels/tsa/holtwinters/model.py", line 291, in __init__
    self._initialize()

  File "/Users/muhlbach/opt/anaconda3/envs/experimental/lib/python3.8/site-packages/statsmodels/tsa/holtwinters/model.py", line 439, in _initialize
    raise ValueError(msg.format("level"))

ValueError: initialization method is estimated but initial_level has been set.`
```

**To Reproduce**

```python
y = load_airline()
forecaster = ThetaForecaster(initial_level=0.5).fit(y)
```

**Expected behavior**
It should estimate Theta with initial smoothing level.

**Additional context**
I think the problem relates to the class ExponentialSmoothing and the fact that users cannot provide the `initialization_method` argument. See https://github.com/alan-turing-institute/sktime/blob/060af89907d135e41491d61f0966e285289a805b/sktime/forecasting/exp_smoothing.py#L10

yashlamba:
#### Reference Issues/PRs
Fixes #692

ThetaForecaster works when passed an initial_level, passed initialization_method parameter to ExponentialSmoothing

NA

Is this a better fix or should I receive it from the user? 
Refer: https://github.com/alan-turing-institute/sktime/issues/692#issuecomment-811197573

##### For all contributions

### DIFF
--- a/sktime/forecasting/theta.py
+++ b/sktime/forecasting/theta.py
@@ -94,13 +94,17 @@ class ThetaForecaster(ExponentialSmoothing):
 
         self.sp = sp
         self.deseasonalize = deseasonalize
-
+        self.initialization_method_ = "known" if initial_level else "estimated"
         self.deseasonalizer_ = None
         self.trend_ = None
         self.initial_level_ = None
         self.drift_ = None
         self.se_ = None
-        super(ThetaForecaster, self).__init__(initial_level=initial_level, sp=sp)
+        super(ThetaForecaster, self).__init__(
+            initial_level=initial_level,
+            initialization_method=self.initialization_method_,
+            sp=sp,
+        )
 
     def fit(self, y, X=None, fh=None):
         """Fit to training data.
--- a/sktime/forecasting/theta.py
+++ b/sktime/forecasting/theta.py
@@ -94,17 +94,12 @@ class ThetaForecaster(ExponentialSmoothing):
 
         self.sp = sp
         self.deseasonalize = deseasonalize
-        self.initialization_method_ = "known" if initial_level else "estimated"
         self.deseasonalizer_ = None
         self.trend_ = None
         self.initial_level_ = None
         self.drift_ = None
         self.se_ = None
-        super(ThetaForecaster, self).__init__(
-            initial_level=initial_level,
-            initialization_method=self.initialization_method_,
-            sp=sp,
-        )
+        super(ThetaForecaster, self).__init__(initial_level=initial_level, sp=sp)
 
     def fit(self, y, X=None, fh=None):
         """Fit to training data.
@@ -130,6 +125,7 @@ class ThetaForecaster(ExponentialSmoothing):
             self.deseasonalizer_ = Deseasonalizer(sp=self.sp, model="multiplicative")
             y = self.deseasonalizer_.fit_transform(y)
 
+        self.initialization_method = "known" if self.initial_level else "estimated"
         # fit exponential smoothing forecaster
         # find theta lines: Theta lines are just SES + drift
         super(ThetaForecaster, self).fit(y, fh=fh)
--- a/sktime/forecasting/tests/test_theta.py
+++ b/sktime/forecasting/tests/test_theta.py
@@ -43,3 +43,15 @@ def test_pred_errors_against_y_test(fh):
     for ints in intervals:
         assert np.all(y_test > ints["lower"])
         assert np.all(y_test < ints["upper"])
+
+
+def test_forecaster_with_initial_level():
+    y = np.log1p(load_airline())
+    y_train, y_test = temporal_train_test_split(y)
+    fh = np.arange(len(y_test)) + 1
+
+    f = ThetaForecaster(initial_level=0.1, sp=12)
+    f.fit(y_train)
+    y_pred = f.predict(fh=fh)
+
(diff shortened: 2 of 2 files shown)


sktime/forecasting/theta.py
```diff
@@ -94,13 +94,17 @@ def __init__(self, initial_level=None, deseasonalize=True, sp=1):
 
         self.sp = sp
         self.deseasonalize = deseasonalize
-
+        self.initialization_method_ = "known" if initial_level else "estimated"
```
mloning, line 97: This should happen in `fit` rather than in `__init__`.

mloning (commented):
Hi @yashlamba - thanks for the PR! Please find one small comment below.

sktime/forecasting/theta.py
yashlamba: I am not sure, why? The super class requires this a parameter when initializing right. Should I override the intialization_method variable instead of passing it?

Sorry, I am just confused and want to make sure.

mloning: Yes, we should probably change it in the super class then. The problem is that if someone uses `set_params(...)` to change hyper-parameters after construction, `__init__` isn't called again and hence the `initialization_method` wouldn't be updated correctly.

yashlamba: Wouldn't that create the same problem for ExponentialSmoothing? Should I add it to the fit method of the super class (that might make the most sense)?

mloning: Yes sounds good!

yashlamba: @mloning So I was changing this and realized what will happen if someone explicitly passes initialization_method to ExponentialSmoothing (since it gives an option to that while ThetaModel doesn't). It should use what the user passed instead of what we need for ThetaModel. If I add it to the fit method of super class, that would always change the initialization_method, which isn't something we want.

In statsmodel, they do this in their ThetaModel and call ExponentialSmoothing with the correct parameter, which is what we are doing when we change it in __init__.

Now the most viable option would be to update the super class variable in fit method of ThetaModel. Let me know if I missed something here.

mloning: Hi @yashlamba yes, makes sense, that should work if we set the variable before the call to the fit of the super class I think

mloning (commented):
Hi @yashlamba - I played around with this, but changing the `initial_level` doesn't seem to have any noticeable effect? I'm not too familiar with the model itself, could you find out if that's expected?

yashlamba:
@mloning, yeah, I noticed that too. I'll look around and get back to you.

yashlamba:
Update: I was busy with some college work + assignments for the past few days. Sorry for the delay, this is on my radar this week.

mloning:
Thanks for the update @yashlamba and no worries

yashlamba:
Hi @mloning, I can confirm that the model is passing the correct parameters to ExponentialSmoothing and it does make some impact on the predictions (not very significant to fail our test). I debugged and checked this within the statsmodel ExponentialSmoothing call and the initial_level is correctly set and I am assuming that nothing will go wrong after that.

One thing I would like to point out is, this isn't something other implementations pass as a parameter, for instance, statsmodel implicitly sets data[0] as the initial level.

https://github.com/statsmodels/statsmodels/blob/557ee695c09542103e0da37dc9b6ed23b766c9be/statsmodels/tsa/forecasting/theta.py#L259-L261

and darts doesn't use this as a parameter too, https://unit8co.github.io/darts/generated_api/darts.models.theta.html.

My suggestions: 
1. We can change our implementation to inherit ThetaModel from statsmodel.
2. We can merge this PR to fix the current API and open an issue for this.

mloning:
Thanks @yashlamba - I suggest to merge this now. There is some ongoing discussion on refactoring the Theta model in #854, perhaps we can take that into account there.

mloning (approved):

mloning:
Ok thanks @yashlamba - now merged!

PR #757, opened 2021-03-27
Add support for exogenous variables to forecasting reduction

mloning:
#### Reference Issues/PRs
Closes #345 
Related PR: #699, #792 
Contributors: @mloning, @Lovkush-A

#### What does this implement/fix? Explain your changes.
* adds support for exogenous variables to forecasting reduction (lagged values only, no contemporaneous values yet)
* adds more efficient sliding window transforms for tabular/panel output
* implements `MultioutputTimeSeriesRegressionForecaster`
* works towards unifying tabular and time-series reduction 
* adds more extensive unit tests for reduction approaches

#### What's still missing
For discussion, see #792 
* for exogenous variables, only considers lagged values for same time points as target variable, not contemporaneous values
* more consistent handling of X (#787)
* simplification of checks on forecasting horizon 
* factor out transformation and formatting, so that same transform/format function can be used in fit and predict 
* raising an error if last window contains insufficient data, rather than returning nan values

#### Does your contribution introduce a new dependency? If yes, which one?
No.

#### What should a reviewer concentrate their feedback on?
Main changes are in
* `sktime/forecasting/compose/_reduce.py`
* `sktime/forecasting/compose/tests/test_reduce.py`

Other files are mostly formatting/minor changes. 

##### For all contributions
- [x] I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/main/.all-contributorsrc).
- [x] Optionally, I've updated sktime's [CODEOWNERS](https://github.com/alan-turing-institute/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
- [x] I've added unit tests and made sure they pass locally.

##### For new estimators
- [x] I've added the estimator to the online documentation.
- [x] I've updated the existing example notebooks or provided a new one to showcase how my estimator works.

PR #784, opened 2021-04-07
Refactor the factory method for reduction forecasters

Lovkush-A:
#### Reference Issues/PRs
Refactors #757 

Refactoring the factory method in reduction forecasters.
* creating new method which includes helper functions to check user inputs
* keeping the old method (and reintroducing an even older method) but adding deprecation warning
* updating various files and test with new methods name
* updating forecasting tutorial notebook

### DIFF
--- a/docs/source/api_reference/forecasting.rst
+++ b/docs/source/api_reference/forecasting.rst
@@ -126,12 +126,11 @@ Composition
 
     EnsembleForecaster
     TransformedTargetForecaster
-    DirectRegressionForecaster
+    DirectTabularRegressionForecaster
     DirectTimeSeriesRegressionForecaster
-    MultioutputRegressionForecaster
-    RecursiveRegressionForecaster
+    MultioutputTabularRegressionForecaster
+    RecursiveTabularRegressionForecaster
     RecursiveTimeSeriesRegressionForecaster
-    ReducedForecaster
     StackingForecaster
     MultiplexForecaster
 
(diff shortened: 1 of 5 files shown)


Lovkush-A:
@mloning What was the reason for adding ReducedRegressionForecaster in your PR? It looks identical to ReducedForecaster, and seems redundant.

mloning:
I included it again because we didn't deprecate it properly when we removed it and a lot of tutorials are based on it.

Lovkush-A:
Alright. Can you point me to an example tutorial, to check the latest deprecation works as expected?

mloning:
Can't think of a tutorial on deprecation. We could add additional warning if the signature changed. But this way at least user don't get a `ModuleNotFound` error but a somewhat more helpful invalid input argument error and can then look up the new signature.

Lovkush-A:
@mloning Sorry, I meant an example sktime tutorial in which ReducedRegressionForecaster is used. I will then locally run the sktime tutorials and see if the error message/deprecation message makes sense.

mloning:
Ah - the forecasting tutorial in the examples folder uses it.

Lovkush-A:
@mloning I looked at the forecasting notebook in examples. `ReducedRegressionForecaster` had already been updated to `ReducedForecaster`. 

More importantly, the deprecation warning did come up for `ReducedForecaster`, so looks like that works fine, and it should work similarly for `ReducedRegressionForecaster`.  However, the notebook will need to be updated as the scitype names have changed. I will update it now (and update it to `make_reduction`)

fkiraly:
a common motif is to print a warning message "is deprecated in v.123 and will be removed in future major release v.234", see also
https://docs.python.org/3/library/exceptions.html#DeprecationWarning

mloning:
Closed in favour of #757

Issue #744, opened 2021-03-24
inverse_transform function for Imputer and HampelFilter


aiwalter:
**Is your feature request related to a problem? Please describe.**
I want to be able to use `Imputer` and `HampelFilter` inside `TransformedTargetForecaster` to be able to tune them together with a forecaster. However, this is currently not possible because there is not `inverse_transform()` function implemented. 

**Describe the solution you'd like**
I have checked the `sklearn.Imputer` and they also dont have an `inverse_tranform`. My idea however is to just implement a dummy `inverse_tranform` function that does nothing, so just returning input as output.

My thoughts on implementing this:
PRO
- enable tuning

CONTRA:
- inverting would not lead to the same original data (but we could store the original data inside the transformer to be able to do this at least for the insample transformations). Otherwise we could just raise a warning that inverting `Imputer` or `HampelFilter` is not unambiguously.

Would be interested in your opinion @mloning @fkiraly
___
Martin Walter <martin_friedrich.walter@daimler.com>, Mercedes-Benz AG on behalf of Daimler TSS GmbH.
https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md

fkiraly:
Hm, yes - or we just want the effect of not having `inverse_transform` implemented to default to the identity transform?
We could do this inside `TransformedTargetForecaster`, or perhaps - better - in the abstract template class (whichever that is)?

mloning:
Instead of doing the identity transform we could just skip them in the pipeline here:
https://github.com/alan-turing-institute/sktime/blob/bcce82cbabff00878d5347225d5679c8d30cf239/sktime/forecasting/compose/_pipeline.py#L130-L131
by adding something like this:
```python
 for _, _, transformer in self._iter_transformers(reverse=True): 
     if hasattr(transformer, "inverse_transform"):
         y_pred = transformer.inverse_transform(y_pred) 
```

Using the identity transform has the danger that for some transformers the inverse transform may simply not be implemented yet, even though it exists. 

I'm also not sure if we're missing something here. There seem to be some kinds of transforms where we do want to take the inverse, and others where we don't, e.g. for box-cox we do, for imputation we don't. Not sure if there's a better way of handling this.

fkiraly:
I feel it's important here to try spell out a few example cases explicitly before we make a design decision - shall we do that? @aiwalter, would you mind explaining how you would like `Imputer` and `HampelFilter` to behave, specifically?

[...]

PR #788, opened 2021-04-08
Added skip-inverse-transform tag for Imputer and HampelFilter

aiwalter:
#### Reference Issues/PRs
#744 

Skip `inverse_transform` for `HampelFilter` and `Imputer` in `TransformedTargetForecaster`
___


sktime/forecasting/compose/_pipeline.py
```diff
@@ -122,6 +121,9 @@ def fit(self, y, X=None, fh=None):
         return self
 
     def _predict(self, fh=None, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
+        # import here to avoid circular dependency
+        from sktime.utils._testing.estimator_checks import _has_tag
```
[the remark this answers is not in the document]
aiwalter: you mean the function `_has_tag`?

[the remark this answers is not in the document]
mloning: Yes!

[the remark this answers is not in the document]
aiwalter: ok now I have another issue with `VALID_ESTIMATOR_TAGS`. I think the general issue is the `_conf.py` file because it contains all those imports but also the configurations. So I cant import the configurations without having a circular import. Can I put  `VALID_ESTIMATOR_TAGS` and `NON_STATE_CHANGING_METHODS` to another file?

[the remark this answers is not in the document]
aiwalter: or I create a separate function somewhere to avoid calling `_has_tag()` in `_pipeline.py`?

sktime/utils/__init__.py
```diff
@@ -176,3 +178,21 @@ def _get_err_msg(estimator_type):
...
+    ----------
+    Estimator : Estimator class
+    tag : str
+        An Estimator tag like "skip-inverse-transform"
+
+    Returns
+    -------
+    bool
+    """
+    assert tag in VALID_ESTIMATOR_TAGS
```
mloning, line 196: I think we should remove the assertion here and write a unit test for it instead. So a unit test that checks if for all estimators, all tags are valid tags. Would that work?

mloning: I'm also not sure why I used an Estimator class here rather than an instance - that should cause any issue though (I hope).

mloning: Perhaps even better if we move `VALID_ESTIMATOR_TAGS` somewhere else, into `sktime/base/__init__.py`? That way, we could still have a check on `tag` here.

sktime/forecasting/compose/_pipeline.py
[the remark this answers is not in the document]
mloning: Please see my comments below - started a new thread :D

aiwalter:
@mloning seems there is an issue with `brew install libomp`, unrelated to my PR.
___

aiwalter:
> @mloning seems there is an issue with `brew install libomp`, unrelated to my PR.
> 
> Martin Walter [martin_friedrich.walter@daimler.com](mailto:martin_friedrich.walter@daimler.com), Mercedes-Benz AG on behalf of Daimler TSS GmbH.
> https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md

seems `libomp` is back working. Will fix the other issues then.

sktime/transformations/series/compose.py
```diff
@@ -70,7 +68,7 @@ class OptionalPassthrough(_SeriesToSeriesTransformer):
         "fit-in-transform": True,
     }
 
-    def __init__(self, transformer, passthrough=True):
+    def __init__(self, transformer, passthrough=False):
```
mloning, line 71: Shouldn't this be True by default?

sktime/utils/__init__.py
```diff
@@ -9,6 +9,18 @@
 from pathlib import Path
 
 
+# We use estimator tags in addition to class hierarchies to further distinguish
+# estimators into different categories. This is useful for defining and running
+# common tests for estimators with the same tags.
+VALID_ESTIMATOR_TAGS = (
```
mloning, line 15: Since we're not checking VALID_ESTIMATOR_TAGS in `_has_tag(...)`, could we move them back to the `test/_config.py` file or are there still circular imports?

mloning (commented):
Hi @aiwalter - I had a quick look. 

We also need to add a test to the pipeline class to make sure if works correctly.

sktime/transformations/series/compose.py
aiwalter: so this `OptionalPassthrough` will most likely only be used in combination with a pipeline like `TransformedTargetForecaster` and then a grid search. I actually wanted not set a default value for `passthrough` because user might forget to add this to the grid as `passthrough: [True, False]` which is the only way it makes sense to use this class. However, not giving a default doesnt work either. So I thought to make `passthrough=False` is better than `True` in case user forgets to add it to grid

sktime/forecasting/compose/tests/test_pipeline.py
```diff
@@ -49,3 +52,24 @@ def compute_expected_y_pred(y_train, fh):
 
     expected = compute_expected_y_pred(y_train, fh)
     np.testing.assert_array_equal(actual, expected)
+
+    def test_skip_inverse_transform():
```
mloning, line 56: The indent seems to be wrong here, assuming this is meant to be as a separate test

aiwalter:
ready to merge :)
___

mloning (approved):
Thanks @aiwalter - will merge soon

PR #791, opened 2021-04-08
Added Guerrero method for lambda estimation to BoxCoxTransformer (#778)

### LINKED ISSUE #778: Implement Guerrero method for Box-Cox lambda estimation
Add Guerrero method for lambda estimation to [BoxCoxTransformer](https://github.com/alan-turing-institute/sktime/blob/main/sktime/transformations/series/boxcox.py).

You can find an example here: https://otexts.com/fpp3/transformations.html
R code from forecast is available here: https://github.com/robjhyndman/forecast/blob/master/R/guerrero.R

GuzalBulatova:
#### Reference Issues/PRs

Fixes #778

Adds `_guerrero` function for lambda estimation to `BoxCoxTransformer`.
This method is tested in `test_boxcox.py` in `test_lambda_bounds` and in `test_guerrero_against_r_implementation`.

No.
##### For all contributions

##### For new estimators


GuzalBulatova:
Hi @mloning, could you please take a look at this draft PR?
The Guerrero method needs seasonal periodicity. I've seen it passed as a parameter in other functions like `autocorrelation_seasonality_test` in `sktime.utils.seasonality`.
If we allow people to specify it here too, how should we pass it through `BoxCoxTransformer`? All other methods like` _pearsonr` don't have any additional parameters.

If people do not provide it, should we estimate it? Is there a method in sktime to do it?
[guerrero.R](https://github.com/robjhyndman/forecast/blob/master/R/guerrero.R) uses `frequency` that seems to come from `findfrequency`  in [forecast.R](https://github.com/robjhyndman/forecast/blob/master/R/forecast.R). It appears to rely on an arbitrary threshold as they specify in line 150. Do we need to implement that too?

mloning (commented):
Hi @GuzalBulatova, we usually pass the seasonal periodicity `sp` to the constructor, so we could have `BoxCoxTransformer.__init__(..., sp=None)` and check if it's set whenever `method=guerrero`. We would then have to move the function definition of `_guerrero` inside the `fit` method and use `self.sp` inside that function. 

Having a function that can infer the `sp` from the data would be nice too, but seems like a separate, more general issue to work on. If I remember correctly, something similar is already implemented in statsmodels or pandas.

GuzalBulatova:
Thank you for the feedback! 
OK, then we'll require` sp `if `_guerrero` is used.
Then providing methods to estimate `sp` should be in a different issue.

sktime/transformations/series/tests/test_boxcox.py
```diff
@@ -25,9 +25,11 @@ def test_boxcox_against_scipy():
 
 
 @pytest.mark.parametrize("bounds", [(0, 1), (-1, 0), (-1, 2)])
-@pytest.mark.parametrize("method", ["mle", "pearsonr"])
-def test_lambda_bounds(bounds, method):
+@pytest.mark.parametrize(
+    "method_sp", [("mle", None), ("pearsonr", None), ("guerrero", 5)]
```
mloning, line 29:
```suggestion
    "method, sp", [("mle", None), ("pearsonr", None), ("guerrero", 5)]
```

sktime/transformations/series/tests/test_boxcox.py
```diff
@@ -25,9 +25,11 @@ def test_boxcox_against_scipy():
...
+)
+def test_lambda_bounds(bounds, method_sp):
```
mloning, line 31:
```suggestion
def test_lambda_bounds(bounds, method, sp):
```

sktime/transformations/series/tests/test_boxcox.py
mloning: pytest allows to separate these arguments

sktime/transformations/series/boxcox.py
```diff
@@ -158,18 +165,63 @@ def _boxcox(x, lmbda=None, bounds=None, alpha=None):
...
 
     def fit(self, Z, X=None):
+        self._check_method()
         z = check_series(Z, enforce_univariate=True)
-        self.lambda_ = _boxcox_normmax(z, bounds=self.bounds, method=self.method)
+        if self.method != "guerrero":
+            self.lambda_ = _boxcox_normmax(z, bounds=self.bounds, method=self.method)
+        else:
+
+            def _guerrero(x, sp):
```
mloning, line 182: I'd prefer adding this to the above `_boxcox_normmax` function as a possible input for the `method` argument. We could even ask SciPy if they'd be interested in including it.

mloning: That way, we can avoid the if-else clause and also get rid of the `self._check_method` helper function, as this would move inside the `_boxcox_normmax` function

mloning: We also have a helper function `check_sp` in `sktime/utils/validation/forecasting` I believe

mloning (commented):
Thanks for the updates @GuzalBulatova!

Please find below a few more comments. I would also like to see a test against results from R's implementation.

fkiraly:
quick question/remark, @guzalbutova and @mloning:

since the Guerrero method is basically a re-sampling based parameter estimator, should this be a separate estimator, in line with `sklearn`'s convention of separating these?
Compare for example `LogisticRegression` and `LogisticRegressionCV`.

mloning:
We don't really have parameter estimators as separate classes anywhere else, do we? We could think about specifying an API for that, but I suggest to finish this PR first using the keyword argument and then think about refactoring later if more modularity or consistency with related methods is required.

sktime/transformations/series/boxcox.py
GuzalBulatova: Thanks for the review, @mloning!
I moved `_guerrero` definition from `_boxcox_normmax` to the `fit` method of `BoxCoxTransformer` because of how I understood [this comment](https://github.com/alan-turing-institute/sktime/pull/791#pullrequestreview-632255503), did I get it wrong?
So it's better if I move the method back inside  `_boxcox_normmax` and pass `sp` there as a parameter? I personally agree that `_guerrero` fits better in there.

Another question is about a test against results from R's implementation: I'm not quite sure how to get these results. Is there a way to run R code here in sktime or sklearn? Or do you mean hard-coded values taken from running R code locally and putting them into the test?

mloning: Oh yes, I forgot about that comment, probably makes sense to keep it separate, but I would still define the `_guerrero(...)` function outside of the `fit` function as a standalone function and pass arguments as necessary inside fit. That would make the if-else clause more readable.

mloning: Yes hard-coded values would be fine I think, perhaps pasting the R code to generate the values as a comment.

GuzalBulatova: Hi @mloning! 
I've just moved it outside the `fit` as a standalone function and added testing against R implementation with hardcoded values.

mloning (commented):
Thanks @GuzalBulatova for the updates! - I had a quick look, need to take a closer look later.

sktime/transformations/series/tests/test_boxcox.py
```diff
@@ -25,9 +26,38 @@ def test_boxcox_against_scipy():
...
+        airline_file <- read.csv(file = 'Airline.csv')[,c('Passengers')]
+        airline.ts <- ts(airline_file)
+        guerrero(airline.ts, lower=-1, upper=2, nonseasonal.length = 20)
+    Output:
+        -0.156981228426408
+    """
+    y = load_airline()
+    t = BoxCoxTransformer(bounds=bounds, method="guerrero", sp=20)
+    t.fit(y)
+    assert t.lambda_ == approx(r_lambda, abs=1e-4)
```
mloning, line 63: Oh nice didn't know that pytest provides `approx` but `np.testing` seems to be preferred in the data science ecosystem, e.g. `np.testing.assert_almost_equal`

GuzalBulatova: Ok thanks, I'll replace `approx` with `np.testing` :)

GuzalBulatova: Hi @mloning!
I've replaced `approx` with `np.testing.assert_almost_equal`.

mloning (approved):
Thanks again @GuzalBulatova - looks all good to me! 

I've updated your branch with the latest changes on `main` and will merge once all tests have passed again.

mloning:
Now merged! 🎉

GuzalBulatova:
Thank you @mloning  for updating the branch and merging it! I'm very glad it's approved! 🚀

PR #699, opened 2021-02-20
Add handling of exogeneous data to forecasting reductions

Lovkush-A:
#### Reference Issues/PRs

see #345 

#### What does this implement/fix? Explain your changes.

Allow reduction of forecasting to regression to handle exogenous data

##### For all contributions
- [ ] I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/master/.all-contributorsrc).
- [ ] Optionally, I've updated sktime's [CODEOWNERS](https://github.com/alan-turing-institute/sktime/blob/master/CODEOWNERS) to receive notifications about future changes to these files.
- [ ] I've added unit tests and made sure they pass locally.

##### For new estimators
- [ ] I've added the estimator to the online documentation.
- [ ] I've updated the existing example notebooks or provided a new one to showcase how my estimator works.

PR #792, opened 2021-04-08
Adjustments to PR757

Lovkush-A:
#### Reference Issues/PRs
See #757 

* Various minor changes (e.g. moving an import statement)
* Introducing a new test with explicit example to aid new contributors

Note this PR is not complete.

##### For all contributions

##### For new estimators


sktime/forecasting/compose/tests/test_reduce.py
```diff
@@ -130,6 +130,48 @@ def test_sliding_window_tranform_panel(n_timepoints, window_length, n_variables,
...
+        [
+            [[0.0, 1.0], [100.0, 101.0], [200.0, 201.0]],
+            [[1.0, 2.0], [101.0, 102.0], [201.0, 202.0]],
+            [[2.0, 3.0], [102.0, 103.0], [202.0, 203.0]],
+            [[3.0, 4.0], [103.0, 104.0], [203.0, 204.0]],
+            [[4.0, 5.0], [104.0, 105.0], [204.0, 205.0]],
+        ]
+    )
+
+    assert (yt_actual == yt_expected).all()
```
mloning, line 170: Please use the assert function from `np.testing`, e.g. `np.testing.assert_array_equal(...)`

sktime/forecasting/compose/tests/test_reduce.py
```diff
@@ -130,6 +130,48 @@ def test_sliding_window_tranform_panel(n_timepoints, window_length, n_variables,
     assert np.all(Xt < yt[:, np.newaxis, [0]])
 
 
+def test_sliding_window_transform_explicit():
```
mloning, line 133: I wanted to write something similar, this is great!

sktime/forecasting/compose/__init__.py
```diff
@@ -13,7 +13,6 @@
     "MultioutputTimeSeriesRegressionForecaster",
     "RecursiveTabularRegressionForecaster",
     "RecursiveTimeSeriesRegressionForecaster",
-    "ReducedForecaster",
```
mloning, line 16: Why do you want to remove this here? 👀

mloning (commented):
I had a quick look already - thanks for the adjustments!

sktime/forecasting/compose/__init__.py
Lovkush-A: I assumed that deprecated stuff don't need to go in.  If this should be kept, then we should also add in ReducedReductionForecaster

mloning: I want to leave it in for the next release and then remove it afterwards.

Lovkush-A:
@mloning. Here is first batch of feedback.  There is quite a lot here. Let me know if something does not make sense

* Variables names.
    * Inconsistent with our markdown file - using z to represent y concatenated with X. I would prefer to call that yX. 
    * Given that y and X have conventions in sktime already, probably better to change the markdown file to match up with sktime conventions, than to change _reduce.py to match markdown
    * I am not fan of yt and Xt, because 'yt' suggest 'transformed version of y' and Xt suggests 'transformed version of X', which is not what is going on.  The suggestion I had in 699 (which I do not 100% like either) was reduction_X and reduction_y. Regardless, either option is better than only having y and X which is what we had previously. Whatever you decide, please update the markdown to match it.

* Docstrings. Bunch of functions do not have docstrings, so those need to be added. Some docstrings are incomplete (e.g. sliding_window_transform does not explain what the parameters/inputs should be)

* checking/setting/preparing fh.
  * This happens (at least) three times:
     * `fh = self.fh.to_relative(self.cutoff)` inside _transform
     * `fh = _prepare_fh(fh)` inside _sliding_window_transform
     * `self._set_fh(fh)` inside fit
  * This seems odd to me. Surely there is a way to do this where it is only checked/set/prepared once.
  * This is one example of the kind of thing I wanted to disentangle in #699 by rewriting all the classes without helper functions.

* checking of scitype seems redundant inside sliding_window_transformer.

* the transformations of data in predict_last_window should be factored out, and ought to be dealt with using _transform.  Whatever transformations we did during fit, we should be going in predict. The difference is that in predict we only output Xt whereas in fit we output both yt and Xt.

* I think, but i am unsure, that _is_predictable is redundant. Why would last_window's length not be equal to window_length? Isn't _get_last_window defined to to produce a window of the appropriate length? And surely we ought to check for nans and infs earlier on (wouldn't the fit stage break if there were nans or infs in the data?)

* The logic of fitting and predicting in the three different strategies looks correct on first glance, but I would have to play around in detail to make sure.

* The removal of the mixin classes for the scitype is nice. This version is cleaner as a result.

* _infer_scitype is a nice touch. My own preference is that user should know what scitype they want, but that is a very minor preference. I like that it is possible to infer and I like that I can see how the inference might be done.

* testing consistency with cv is good way to automatically create many tests and test various settings.

* test_sliding_window_transform tabular and panel does not look like a convincing test to me; it tests some necessary conditions but no sufficient conditions. This test is probably redundant given comparison to cv and given new explicit test.

* `test_make_reduction_construct_instance` is nice; much cleaner than what I did.

* testing infer.
  * why is TimeSeriesForestRegressor() a  tabular-regressor? The name sounds like it ought to be a time series regressor.

mloning:
> @mloning. Here is first batch of feedback. There is quite a lot here. Let me know if something does not make sense

Thanks @Lovkush-A for the detailed feedback! 

> * Variables names.
>   
>   * Inconsistent with our markdown file - using z to represent y concatenated with X. I would prefer to call that yX.
>   * Given that y and X have conventions in sktime already, probably better to change the markdown file to match up with sktime conventions, than to change _reduce.py to match markdown
>   * I am not fan of yt and Xt, because 'yt' suggest 'transformed version of y' and Xt suggests 'transformed version of X', which is not what is going on.  The suggestion I had in 699 (which I do not 100% like either) was reduction_X and reduction_y. Regardless, either option is better than only having y and X which is what we had previously. Whatever you decide, please update the markdown to match it.


> * Docstrings. Bunch of functions do not have docstrings, so those need to be added. Some docstrings are incomplete (e.g. sliding_window_transform does not explain what the parameters/inputs should be)

Agreed, will have to be added (note that `_sliding_window_transform` is a private function and perhaps doesn't need proper docstrings) - I'll try to add more docstrings/notes before merging.

> * checking/setting/preparing fh.
>   
>   * This happens (at least) three times:
>     
>     * `fh = self.fh.to_relative(self.cutoff)` inside _transform
>     * `fh = _prepare_fh(fh)` inside _sliding_window_transform
>     * `self._set_fh(fh)` inside fit
>   * This seems odd to me. Surely there is a way to do this where it is only checked/set/prepared once.
>   * This is one example of the kind of thing I wanted to disentangle in #699 by rewriting all the classes without helper functions.

They all do different things. One issue is that cutoff is only available in the forecaster, not the forecasting horizon itself. The sliding window transform needs the relative representation of the horizon. `_set_fh()` is the common interface for forecasters to check/set the horizon. I agree that there these aspects could be simplified further, but probably not without making changes to the `ForecastingHorizon` class. 

> * checking of scitype seems redundant inside sliding_window_transformer.

Yes, that's probably true, helped me catch errors but can be removed now I think. 

> * the transformations of data in predict_last_window should be factored out, and ought to be dealt with using _transform.  Whatever transformations we did during fit, we should be going in predict. The difference is that in predict we only output Xt whereas in fit we output both yt and Xt.

I'm not sure if that's true. We only look at the last window in predict and do not need to do the sliding window transformation again, as we're dealing with only a single instance. There is the formatting aspect though, converting between tabular and panel data, which I thought about factoring out into mixin classes, but preferred the conditional statements in predict. That seemed simpler than having to deal with mixin classes. 

> * I think, but i am unsure, that _is_predictable is redundant. Why would last_window's length not be equal to window_length? Isn't _get_last_window defined to to produce a window of the appropriate length? And surely we ought to check for nans and infs earlier on (wouldn't the fit stage break if there were nans or infs in the data?)

Yes, that's legacy code to handle iterative update-predict procedures, will have to be cleaned up in a separate PR. I think that if there isn't enough data in the last window, we should simply raise an error instead of returning nan.

> * The logic of fitting and predicting in the three different strategies looks correct on first glance, but I would have to play around in detail to make sure.

Sure, let me know if you find any inconsistencies. 

> * The removal of the mixin classes for the scitype is nice. This version is cleaner as a result.

I agree, but comes at the cost of conditional statements in predict (see comment above).

> * _infer_scitype is a nice touch. My own preference is that user should know what scitype they want, but that is a very minor preference. I like that it is possible to infer and I like that I can see how the inference might be done.

It's just convenience. Technically, users provide the scitype already with the choice of algorithm, so asking them again for a string seems redundant, but scityping isn't perfect in scikit-learn or sktime, so if inference fails, users can specify it. 

> * test_sliding_window_transform tabular and panel does not look like a convincing test to me; it tests some necessary conditions but no sufficient conditions. This test is probably redundant given comparison to cv and given new explicit test.

Yes, they are bit hard to follow, but cover many configurations, they're also fast, so I'd leave them in. 

> * testing infer.
>   * why is TimeSeriesForestRegressor() a  tabular-regressor? The name sounds like it ought to be a time series regressor.

Seems like a bug. Not even sure why the test passes, perhaps inference doesn't work after all. Will have to look into this.

Lovkush-A:
@mloning. Thanks for detailed response!

> note that _sliding_window_transform is a private function and perhaps doesn't need proper docstrings

Sounds like you're treading dangerous territory...  Docstrings are helpful for developers, just as much as end-users!  In fact, during this review, a docstring would have helped me know what types were accepted for y and X. I used numpy arrays in my first attempts rather than Series/DataFrames.

> I agree that there these aspects could be simplified further, but probably not without making changes to the ForecastingHorizon class.

Three suggestions:
* whatever you do, can you document somewhere in the file what the differences are between the three? It is not obvious from the names; they all sound like they do the same thing.
* can you have all the checks happen at the same place, rather then spreading them out?
* leave it for now but create a new issue to improve it later.

>I'm not sure if that's true. We only look at the last window in predict and do not need to do the sliding window transformation again, as we're dealing with only a single instance. There is the formatting aspect though, converting between tabular and panel data, which I thought about factoring out into mixin classes, but preferred the conditional statements in predict.

* Sliding window transformation applied to the final bunch of rows in y and X should do the necessary formatting, no?
* I don't think you need mixin classes. Presumably we could create a helper function to factor out relevant bits.
* For me personally, conditionals are not an issue, but maybe there is a downside about them I am unaware of.

mloning:
Thanks @Lovkush-A - I've added some notes on #757 and will try to make issues out of them later. 

Could you try to fix the failing tests on this PR so that we can merge it? Let me know if you need help.

mloning (approved):
Thanks @Lovkush-A - let me know if you're happy with these changes as they are and I'll merge them.

Lovkush-A:
@mloning Yes I'm happy with it.

mloning:
Thanks again for the review and improvements!

PR #796, opened 2021-04-11
Use joblib to parallelize ensemble fitting for Rocket classifier

kachayev:

`ROCKETClassifier` now excepts `n_jobs` parameter to parallelize ensemble fitting using `joblib`.

Similar to implementation for `TimeSeriesForest` introduced in #408. 

##### For all contributions


sktime/classification/shapelet_based/tests/test_rocket_classifier.py
```diff
@@ -41,6 +41,27 @@ def test_rocket_ensemble_on_gunpoint():
     testing.assert_array_equal(probas, rocket_e_gunpoint_probas)
 
 
+def test_rocket_parallel_ensemble_on_gunpoint():
```
mloning, line 44: This seems to be the same test logic as the above test, could we use `pytest.mark.parametrize("n_jobs", [1, 8])` to iterate over it rather than duplicating the code?

sktime/classification/shapelet_based/_rocket_classifier.py
```diff
@@ -17,6 +18,14 @@
 from sktime.utils.validation.panel import check_X_y
 
 
+def _fit_pipeline(X, y, num_kernels, random_state):
```
mloning, line 21: I suggest to construct the instance once outside the function in the main `fit` method, and then have this function accept a cloned instance and X and y, rather than passing hyper-parameters and constructing it inside the function (for cloning you can use `from sklearn.base import clone`)

```python
def _fit_estimator(estimator, X, y):
    return estimator.fit(X, y)
```

sktime/classification/shapelet_based/_rocket_classifier.py
```diff
@@ -109,23 +122,16 @@ def fit(self, X, y):
...
-                rocket_pipeline.fit(X, y)
-                self.classifiers.append(rocket_pipeline)
-                self.weights.append(rocket_pipeline.steps[1][1].best_score_)
-                self.weight_sum = self.weight_sum + self.weights[i]
-        else:
-            rocket_pipeline = make_pipeline(
-                Rocket(num_kernels=self.num_kernels, random_state=self.random_state),
-                RidgeClassifierCV(alphas=np.logspace(-3, 3, 10), normalize=True),
+            self.classifiers = Parallel(n_jobs=self.n_jobs)(
+                delayed(_fit_pipeline)(X, y, self.num_kernels, self.random_state)
```
mloning, line 126: Here we would pass the cloned instance
```suggestion
                delayed(_fit_estimator)(clone(estimator), X, y)
```

mloning (commented):
Thanks @kachayev for the PR! Please find a few minor comments below.

sktime/classification/shapelet_based/_rocket_classifier.py
mloning: If I understand this correctly, we also want to make sure that ensemble members use different random states (though fixed if passed), to avoid having an ensemble in which all members are the same, I remember that scikit-learn has some helper function for this but don't remember where exactly

kachayev: Regarding `random_state`, absolutely. I've been thinking about it but didn't find how exactly this is handled for other ensembles throughout the project. E.g. here https://github.com/alan-turing-institute/sktime/blob/30cb0fc3ae1001ed7936b82331627c734ea1698d/sktime/series_as_features/base/estimators/interval_based/_tsf.py#L162-L172 -- the same `random_state` is passed a parameter into all estimators. Let me check what does `sklearn` provide as an option

kachayev: Okay, I found how `sklearn` sets random state internally for ensembles. Working on update to the PR. I'm going to make `n_jobs=None` by default and have another branch where the processing happens exactly as it was done before (without any manipulations with `random_state`). Otherwise this update will break reproducibility of ensembles trained previously.

kachayev:
@mloning Thanks a lot for the review! I pushed a change with `random_state` being managed the same way `sklearn` does it internally. Though I have a couple of questions/concenrs:

* is it okay to rely on "internal" `_set_random_state` function from `sklearn`? I can simply copy-paste the code if we don't want to introduce this subtle dependency
* do we need to re-work sktime/sktime/series_as_features/base/estimators/interval_based/_tsf.py using similar approach?

mloning:
Hi @kachayev 
* Setting `n_jobs=None` makes sense in line with scikit-learn's default (we do have a `check_n_jobs` helper function in `sktime/utils/validation` to enforce scikit-learn's convention if needed
* Yes, using `_set_random_state(...)` seems fine
* Yes, updating TSF accordingly would be nice in a separate PR

kachayev:
Hi @mloning! I've added validator for `n_jobs` arguments. I will work on TSF separately, and this PR is good to go 😀

sktime/classification/shapelet_based/_rocket_classifier.py
```diff
@@ -70,11 +76,13 @@ def __init__(
         ensemble=False,
         ensemble_size=25,
         random_state=None,
+        n_jobs=None,
     ):
         self.num_kernels = num_kernels
         self.ensemble = ensemble
```
mloning, line 82: Convention is to use `n_estimators=None` rather than `ensemble_size`, which could default to 1 if not given, making the boolean ensemble option redundant

sktime/classification/shapelet_based/_rocket_classifier.py
```diff
@@ -102,31 +110,35 @@ def fit(self, X, y):
...
-                    RidgeClassifierCV(alphas=np.logspace(-3, 3, 10), normalize=True),
+            if n_jobs is None:
+                for _ in range(self.ensemble_size):
+                    base_estimator = _make_estimator(
+                        self.num_kernels, self.random_state
+                    )
+                    self.classifiers.append(_fit_estimator(base_estimator, X, y))
+            else:
+                base_estimator = _make_estimator(self.num_kernels, self.random_state)
+                self.classifiers = Parallel(n_jobs=n_jobs)(
```
mloning, line 129: `self.classifiers` should be `self.estimators_` with the trailing underscore indicating fitted parameters

mloning (commented):
Thanks @kachayev - re-reading the code I noticed a few other things which would help make the classifier more consistent with other code in sktime/sklearn. Not sure if we want to change them as part of this PR. Let me know what you think.

sktime/classification/shapelet_based/_rocket_classifier.py
kachayev: Renaming of the parameter will introduce potential incompatibility with the existing code. Should we have both parameters as of now and a `DeprecationWarning` that the old parameter will be removed in future releases?

sktime/classification/shapelet_based/_rocket_classifier.py
kachayev: Same comment about compatibility after renaming. Should be less of an issue in this case though `classifiers` is mentioned in docs as a public attribute of the `ROCKETClassifier` class. Which means it could be used outside of the scope of the library.

sktime/classification/shapelet_based/_rocket_classifier.py
mloning: Yes deprecating it would be nice (we don't have a policy on deprecation cycles at the moment, too much is still in flux at this stage)

sktime/classification/shapelet_based/_rocket_classifier.py
mloning: Yes, see above comment

kachayev:
@mloning Updated the code with new parameter names and deprecation warnings when using old convention

sktime/classification/shapelet_based/_rocket_classifier.py
```diff
@@ -102,37 +123,41 @@ def fit(self, X, y):
...
-                rocket_pipeline = make_pipeline(
-                    Rocket(
-                        num_kernels=self.num_kernels, random_state=self.random_state
-                    ),
-                    RidgeClassifierCV(alphas=np.logspace(-3, 3, 10), normalize=True),
+        for index, class_val in enumerate(self.classes_):
+            self.class_dictionary[class_val] = index
+
+        if self.n_estimators is not None:
+            if n_jobs is None:
```
mloning, line 134: I don't think we need the if-else case distinction here, `n_jobs=None` is interpreted as `n_jobs=1` (unless inside a context manager)

sktime/classification/shapelet_based/_rocket_classifier.py
```diff
@@ -28,14 +34,15 @@ class ROCKETClassifier(BaseClassifier):
     ----------
     num_kernels             : int, number of kernels for ROCKET transform
     (default=10,000)
-    ensemble                : boolean, create ensemble of ROCKET's (default=False)
-    ensemble_size           : int, size of the ensemble (default=25)
+    n_estimators            : int, ensemble size, optional (default=None)
```
mloning, line 37: we should say that None means 1

sktime/classification/shapelet_based/_rocket_classifier.py
```diff
@@ -28,14 +34,15 @@ class ROCKETClassifier(BaseClassifier):
...
     random_state            : int or None, seed for random, integer,
     optional (default to no seed)
+    n_jobs                  : int, the number of jobs to run in parallel for `fit`,
```
mloning, line 40: Do you also want to add parallelisation to `predict`?

mloning (commented):
Thanks for the changes - a few more comments

sktime/classification/shapelet_based/_rocket_classifier.py
kachayev: Yeah, that was previously the case (described here https://github.com/alan-turing-institute/sktime/pull/796#discussion_r612109920). Right now if `n_jobs` is not `None`, we fit a clone of the estimator rather than estimator itself. And random state is re-defined for all clones. Which  means that setting `n_jobs=1` be default would break reproducibility with previously trained classifiers. If it's okay to ignore this fact, I can re-work the code to remove one layer of if-else branches.

sktime/classification/shapelet_based/_rocket_classifier.py
kachayev: I'm not sure about that, predict is straightforward and reasonably fast (at least when compared to fitting). I don't know if the overhead of `Parallel` is even worth it. If we do want to parallelize predict as well, it might be good to have different `n_jobs` for this, e.g. I have `n_jobs=25` for fitting and `n_predict_jobs=2` for predict. Or... maybe an additional parameter for `predict` call itself? What do you think?

sktime/classification/shapelet_based/_rocket_classifier.py
kachayev: Updated documentation. Plus, I changed the execution to treat `n_estimators=1` exactly as `n_estimators=None`

sktime/classification/shapelet_based/_rocket_classifier.py
mloning: I prefer fixing the bug here and have a single training loop with properly set random states (breaking reproducibility). Reproducibility can still be ensure by reference to the older sktime release. @TonyBagnall @ABostrom @jasonlines, any thoughts on this as code owners?

TonyBagnall:
hi, I've not followed this, sorry, but I thought the ROCKET transform was already threaded? I am also not a fan of randomised seeding, it is deceptive to the user to say you can set seed but get different results. If you need repeated RNG from a single seed, just multiply the seed by a different constant for each RNG.

sktime/classification/shapelet_based/_rocket_classifier.py
kachayev: Okay, I will revert to initializing `n_jobs` to 1

kachayev: Done!

sktime/classification/shapelet_based/_rocket_classifier.py
mloning: Okay let's leave it at fit for now

mloning:
@TonyBagnall based on your comment, would you prefer rewriting `_set_random_state` rather than importing it from scikit-learn. It played around with `_set_random_state`, it seems like you do get always the same random state for ensemble members. 

When I rerun the code below, I always get the same random states for the `DecisionTreeRegressor`s.
```python
import numpy as np

estimator = RandomForestRegressor(n_estimators=3, random_state=123)
estimator.fit(np.array([[1, 2]]), np.array([1]))
estimator.estimators_
>>> [DecisionTreeRegressor(max_features='auto', random_state=1872583848),
 DecisionTreeRegressor(max_features='auto', random_state=794921487),
 DecisionTreeRegressor(max_features='auto', random_state=111352301)]
```

Is that ok then?

kachayev:
Yeah, removing all the logic for traversing params for a given estimators, what the code is doing is basically the following:

```python
            to_set[key] = random_state.randint(np.iinfo(np.int32).max)
```

(source code [here](https://github.com/scikit-learn/scikit-learn/blob/e4bb9fa86b0df873ad750b6d59090843d9d23d50/sklearn/ensemble/_base.py#L76)).

So, it derives random states for all estimators in the ensemble from `random_state` given. Thus remains deterministic with response to initial `random_state` + number of estimators in the ensemble.

mloning:
@TonyBagnall I'll merge this in the next few days unless you want any other changes

TonyBagnall:
can I just ask @MatthewMiddlehurst to have a look?

mloning:
@MatthewMiddlehurst @TonyBagnall could you suggest changes or approve?

MatthewMiddlehurst:
No issues if it passes all the tests. Though at some point I may separate the ensemble into a different classifier and have the original ROCKETClassifier with the new ROCKET addons and Arsenal (the ensemble version used for HC2). May be easier just to do that now. 

Worried it may be a be misleading currently to users, its not really a parameter you want to use unless you need to produce probability estimates as we do with HC2. Regular ROCKET will be faster and just as accurate if you don't need them.

mloning:
@MatthewMiddlehurst agreed - the ensembling doesn't really add much at this level. I suggest to merge this now and then think about changing/deprecating it later. So I'll merge in the next few days unless there are more comments.

mloning:
Thanks again @kachayev - now merged!

Issue #712, opened 2021-02-26
Uniform handling of y_train in forecasting performance metrics


mloning:
**Is your feature request related to a problem? Please describe.**
Some forecasting performance metrics require `y_train`, e.g. [MASE](https://robjhyndman.com/papers/mase.pdf). This adds some complication to higher-level functionality that expects a common interface for metrics, like the `evaluate` function or `ForecastingGridSearchCV`, and in unit testing (see #672).

**Current problem**
This currently fails because `y_train` is not passed internally when calling `scoring`. 
```python
from sktime.forecasting.all import *
from sktime.forecasting.model_evaluation import evaluate

y = load_airline()
f = NaiveForecaster()
cv = SlidingWindowSplitter()
scoring = MASE()
out = evaluate(f, cv, y, scoring=scoring)
```

**Possible solutions**
1. Change interface for all performance metrics to optionally accept `y_train`, but only those that require it use it. This requires wrapping metrics from scikit-learn.  
2. Add case distinctions in higher-level functionality to separately handle those metrics that require `y_train` and those that do not. This requires adding a `requires_y_train` attribute to metric classes.
3. Adapt metrics interface at run time to, making case distinctions inside adapter, exposing uniform interface to higher-order functionality (suggested by @fkiraly). This also requires adding a `requires_y_train` attribute to metric classes.

**Describe the solution you'd like**
```python
from sktime.forecasting.all import *
from sktime.forecasting.model_evaluation import evaluate

y = load_airline()
fh = np.arange(1, 10)
y_train, y_test = temporal_train_test_split(y, fh=fh)
f = NaiveForecaster()
f.fit(y_train)
y_pred = f.predict(fh)

# uniform interface
scoring = MASE()
scoring.requires_y_train = True
scoring = check_scoring(scoring)
scoring(y_test, y_pred, y_train)
>>> 3.577770878609128

scoring = sMAPE()
scoring.requires_y_train = False
scoring = check_scoring(scoring)
scoring(y_test, y_pred, y_train)
>>> 0.1780237534499896
```

Here's a rough implementation of the adapter-based solution:
```python
class _MetricAdapter:
    """
    Adapter for performance metrics to uniformly handle 
    y_train requirement of some metrics.
    """

    def __init__(self, metric):
        # wrap metric object
        self.metric = metric
        
    def __call__(self, y_true, y_pred, y_train, *args, **kwargs):
        """Compute metric, uniformly handling those metrics that 
[...]

PR #801, opened 2021-04-12
New forecasting metrics

### LINKED ISSUE #671: Add Additional Forecasting Evaluation Functions/Classes

RNKuhns:
#### Reference Issues/PRs

Fixes #671. See discussion on closed PR #672 that this pull request will complete. 

Adds additional forecasting metrics as discussed in #671. 

This is a draft pull request. I need to finish the docstrings (I'm adding examples to docstrings) and also write unit tests but wanted to give you a heads up of the functionality.

The pull request adds new forecasting performance metrics (see list below) that can each work with multivariate series (but default to work with univariate forecasts). Also changes naming conventions to align with scikit-learn (e.g. mean_absoluate_error rather than mae or mae_loss).

The end result would be the following performance metric functions (each with a corresponding class version for scoring):

relative_error,
mean_asymmetric_error,
mean_absolute_scaled_error,
median_absolute_scaled_error,
mean_squared_scaled_error,
median_squared_scaled_error,
mean_absolute_error,
mean_squared_error,
median_absolute_error,
median_squared_error,
median_absolute_percentage_error,
mean_squared_percentage_error,
median_squared_percentage_error,
mean_relative_absolute_error,
median_relative_absolute_error,
geometric_mean_relative_absolute_error,
geometric_mean_relative_squared_error

None. Pinning Scikit-learn dependency to >= 0.24.* was done separately.

Focus on changes in ./performance_metrics/forecasting/_functions.py and ./performance_metrics/forecasting/_classes.py. 

Finishing fixes elsewhere in module where old performance metric names (e.g. sMAPE) are imported. 

##### For all contributions

##### For new estimators


docs/source/user_guide/introduction.rst
```diff
@@ -41,8 +41,8 @@ can simply write:
     forecaster = ReducedRegressionForecaster(regressor, window_length=12)
     forecaster.fit(y_train)
     y_pred = forecaster.predict(fh)
-    smape_loss(y_test, y_pred)
-    >>> 0.12726230426056875
+    SymmetricMeanAbsolutePercentageError(y_test, y_pred)
+    >>> 12.574429324402637
```
mloning, line 45: Could we scale them so that they are between 0 and 1? I think this is how MAPE is defined in scikit-learn too: https://scikit-learn.org/stable/modules/generated/sklearn.metrics.mean_absolute_percentage_error.html

sktime/utils/validation/forecasting.py
```diff
@@ -101,22 +100,32 @@ def check_X(
     )
 
 
-def check_y(y, allow_empty=False, allow_constant=True, enforce_index_type=None):
+def check_y(
```
mloning, line 103: Could we still revert the changes here (removing the option to enforce_univariate) as discussed in the previous PR?

mloning (commented):
I had a quick look - let me know if you need any help with debugging. 

I'll take a closer look once all tests pass and you're happy with the PR.

docs/source/user_guide/introduction.rst
fkiraly: MAPE in general won't be between 0 and 1 (it can be greater than 1), but what you mean is that 1 should correspond to 100% percentage error?

sktime/utils/validation/forecasting.py
RNKuhns: @mloning and @fkiraly I'm using the scikit-learn checks for the metric input validations, which allow multivariate inputs. However, the scikit-learn checks won't work for validating the input `y_train` info for those metrics that need it (MASE and its related metrics). 

One option would be to keep the enforce_univariate option (which is also useful down the line for multivariate forecasting problems). 

Other option is that I create some separate validation that just applies to `y_train` metric inputs.

sktime/utils/validation/series.py
```diff
@@ -34,18 +34,19 @@ def check_series(
...
+    Z : pd.Series, pd.DataFrame, np.ndarray
+        Univariate or multivariate time series. NumPy ndarray only allowed
+        if `allow_numpy` = True.
+    enforce_univariate : bool, default = False
         If True, multivariate Z will raise an error.
-    allow_empty : bool
-        If False, empty Z will raise an error
-    enforce_index_type : type, optional (default=None)
+    allow_empty : bool, default = False
+    allow_numpy : bool, default = True
```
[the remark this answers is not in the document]
RNKuhns: See my comment above about the `y_train` validation in metrics that need it (e.g. MASE).

sktime/utils/validation/series.py
```diff
@@ -69,10 +70,12 @@ def check_series(
     if enforce_univariate:
         _check_is_univariate(Z)
 
-    # check time index
-    check_time_index(
-        Z.index, allow_empty=allow_empty, enforce_index_type=enforce_index_type
-    )
+    # check time index if input data is not an NumPy ndarray
+    if not isinstance(Z, np.ndarray):
```
RNKuhns, line 74: @mloning  and @fkiraly note that this function already had option to accept a ndarray. However, the call to check_time_index was using the `.index` attribute that a ndarray does not have. Just added guardian code to avoid this if a ndarray is being checked. 

Also added extra doc to make clearer that ndarrays were already in the allowed input type.

docs/source/user_guide/introduction.rst
RNKuhns: I saw that in the scikit docs. I'm actually not sure why they mention the 0 to 100 range in their documentation. Bad predictions can definitely have arbitrarily large MAPE values (e.g. prediction is always 4 and true value is 1 would yield a MAPE of 300 or 3, depending on how you scale). Right now I'm showing 100% error as 100 (Hyndman paper referenced in docstring multiplies by 100), but we could just as easily not scale by 100.

sktime/contrib/distance_based/elastic_ensemble_from_file.py
```diff
@@ -1,13 +1,21 @@
+# -*- coding: utf-8 -*-
 import numpy as np
 import os
-from sklearn.metrics.scorer import accuracy_score
```
RNKuhns, line 3: This wasn't working with update to scikit-learn being pinned to 0.24.*. Rest of changes to this file are from the CI auto-reformatting.

sktime/utils/validation/series.py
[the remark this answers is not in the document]
RNKuhns: Not that the function already had arguments to accept multivariate data and/or ndarrays.

sktime/forecasting/base/_base.py
```diff
@@ -169,13 +169,15 @@ def score(self, y, X=None, fh=None):
...
         """
         # no input checks needed here, they will be performed
         # in predict and loss function
-        from sktime.performance_metrics.forecasting import smape_loss
+        from sktime.performance_metrics.forecasting import (
+            symmetric_mean_absolute_percentage_error,
+        )
 
-        return smape_loss(y, self.predict(fh, X))
+        return symmetric_mean_absolute_percentage_error(y, self.predict(fh, X))
```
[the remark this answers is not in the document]
RNKuhns: We are passing it as a keyword. The last step was to fix all the references to the functions/classes and update them. This should be corrected now.

sktime/performance_metrics/forecasting/_classes.py
```diff
@@ -1,54 +1,381 @@
...
-def make_forecasting_scorer(fn, name=None, greater_is_better=False):
-    """Factory method for creating metric classes from metric functions"""
-    return MetricFunctionWrapper(fn, name=name, greater_is_better=greater_is_better)
+class PercentageErrorMixIn:
+    def __call__(self, y_true, y_pred):
+        return self.fn(y_true, y_pred, symmetric=self.symmetric)
 
 
-class MASE(MetricFunctionWrapper):
+class SquaredErrorMixIn:
```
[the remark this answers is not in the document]
RNKuhns: I wasn't sure if we should treat these as private or not. I've added the underscore.

sktime/utils/validation/forecasting.py
RNKuhns: @mloning should be backed out now

sktime/performance_metrics/forecasting/_classes.py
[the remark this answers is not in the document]
mloning: Yes anything that we don't want users to use directly should be private

RNKuhns:
@mloning I just pushed an update I am hoping will fix the kernel issue I created in the example notebook. I'm running into an issue with the tests passing locally but failing here. The error appears related to `median_absolute_errror` as this is used in `median_absolute_scaled_error` that is also having an issue. 

Errors indicate undefined `sample_weight` parameter being used. The function `median_absolute_error` is a wrapper around scikit-learn version that has that parameter (we've defined a `horizon_weight` parameter I pass to that one in the underlying scikit-learn function). I checked scikit-learn docs and everything seems to match up with the usage ([Scikit-learn's median_absolute_error docs](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.median_absolute_error.html#sklearn.metrics.median_absolute_error))

I thought I set scikit-learn to use vesrion 0.24 onward. How can I see what version is being used in the build? Any other thought besides that on what could be happening there?

mloning:
Hi @RNKuhns, yes you're right, it's still using scikit-learn version 0.23 - I suggest to create a separate PR to bump up the scikit-learn version as we may have to make a few additional adjustments. We describe some of the things that need to change here: https://github.com/alan-turing-institute/sktime/blob/c2d754f38228c9d426983f8adb26cd76900b6976/CONTRIBUTING.md#dependencies

mloning:
I've created one here: https://github.com/alan-turing-institute/sktime/pull/810

sktime/contrib/distance_based/elastic_ensemble_from_file.py
mloning: see https://github.com/alan-turing-institute/sktime/pull/810

docs/source/user_guide/introduction.rst
mloning: Yes 1 meaning 100% is what I meant

mloning: Let's follow scikit-learn rather than the Hyndman paper, I remember there were some inconsistencies in the M4 study too which always required to go back and forth between dividing/multiplying by 100

sktime/utils/validation/forecasting.py
```diff
@@ -332,17 +331,32 @@ def check_scoring(scoring):
...
-        return sMAPE()
+        return MeanAbsolutePercentageError(symmetric=True)
 
     if not callable(scoring):
         raise TypeError("`scoring` must be a callable object")
 
-    allowed_base_class = MetricFunctionWrapper
-    if not isinstance(scoring, allowed_base_class):
-        raise TypeError(f"`scoring` must inherit from `{allowed_base_class.__name__}`")
+    allowed_base_classes = (
```
mloning, line 349: Thinking about this, I'm not sure if we need the inheritance check here, in principle users could pass their custom error functions/objects as long as they adhere to the expected API

RNKuhns: @mloning I was thinking about that when writing the update. Benefit I could see is that it would ensure the user's custom error function adheres to the same `__call__` signature, but I agree it seems like a restrictive way to do that. 

If we wanted, I'm guessing we could check the calIable classes args more directly if we prefer to flag custom classes that don't adhere to the API expectations here rather than when they get called. 

I'll take this check out though when I push my last commit.

RNKuhns:
@mloning thanks for opening up the PR for the update of the scikit-learn. I think I've taken care of the things we've talked about and the build appears to be working. I'll switch this over to ready for review.

docs/source/user_guide/introduction.rst
RNKuhns: Switched this to make 100 percent error = 1.

sktime/performance_metrics/tests/test_performance_metrics_forecasting.py
```diff
@@ -2,64 +2,524 @@
...
+    loss_func = LOSS_RESULTS[loss_func_name]["func"]
+    true_loss = 0
+    y_true = Y_TEST_CASES[f"test_case_{n_test_case}"]["test"]
+    y_train = Y_TEST_CASES[f"test_case_{n_test_case}"]["train"]
+    # Testing case of perfect forecaster
+    y_pred = y_true
+    # Also perfect benchmark
+    y_pred_benchmark = y_true
 
+    loss_func_args = inspect.getfullargspec(loss_func).args
```
mloning, line 228: perhaps wrap this into a helper function `_requires_y_train(func)` which returns a boolean, that way we can directly write `if _requires_y_train(loss_func):`

mloning: or even `_requires_arg(func, arg)` to also handle `y_pred_benchmark`

sktime/performance_metrics/tests/test_performance_metrics_forecasting.py
```diff
@@ -2,64 +2,524 @@
...
+    if "y_train" in loss_func_args:
+        function_loss = loss_func(y_true, y_pred, y_train)
+        class_loss = loss_class.fn(y_true, y_pred, y_train)
```
mloning, line 231: shouldn't we here test whether `__call__` works as expected rather than `fn` attribute?

mloning: I was also thinking whether we should make `fn` private and rename it to `_func`, more readable in my opinion, and avoids users using it rather than going through `__call__`

sktime/performance_metrics/tests/test_performance_metrics_forecasting.py
```diff
@@ -2,64 +2,524 @@
...
+    loss_func_args = inspect.getfullargspec(loss_func).args
+    if "y_train" in loss_func_args:
+        if len(class_attrs) > 0:
+            function_loss = loss_func(y_true, y_pred, y_train, **class_attrs)
+            class_loss = loss_class.fn(y_true, y_pred, y_train, **class_attrs)
+        else:
+            function_loss = loss_func(y_true, y_pred, y_train)
+            class_loss = loss_class.fn(y_true, y_pred, y_train)
+    elif "y_pred_benchmark" in loss_func_args:
+        if len(class_attrs) > 0:
```
mloning, line 278: this also seems to be something we can factor out into helper function, e.g. `_generate_loss(loss, y_true, y_pred, y_pred_benchmark, class_attrs)` to avoid repeating the if-clause

sktime/performance_metrics/tests/test_performance_metrics_forecasting.py
```diff
@@ -2,64 +2,524 @@
...
+            function_loss = loss_func(y_true, y_pred)
+
+        is_num = is_numeric_dtype(function_loss)
+        is_scalar = np.isscalar(function_loss)
+        assert is_num and is_scalar, " ".join(
+            ["Loss function with univariate input should return scalar number"]
+        )
+
+
+# @pytest.mark.parametrize("loss_func", LOSS_RESULTS.keys())
```
mloning, line 382: why are they commented out? shall we remove them or uncomment them?

mloning (commented):
Hi @RNKuhns - this is a huge contribution - thank you! Generally very well written and documented, I've left a number of minor comments below, haven't had a chance to look at all of it yet, but will resume the review later! 

We also need to add the metrics to the docs in `docs/source/api_reference` - currently metrics are not included at all as it seems.

sktime/performance_metrics/tests/test_performance_metrics_forecasting.py
RNKuhns: Yea, we need to remove them. The use of the scikit-learn input validations rendered those moot since their functionality converts array-like things to arrays. I'll delete these in the last push I make.

sktime/performance_metrics/tests/test_performance_metrics_forecasting.py
RNKuhns: That does seem better. I'll update that. Right now we can't test the call because the call doesn't accept y_train (I think our plan was to complete this pull request based on #712, and follow-up right away with a second on that to update the classes with that). 

Note that the second pull request will also make the test function part much easier (no need to conditionally calculate loss, etc).

sktime/performance_metrics/tests/test_performance_metrics_forecasting.py
RNKuhns: Same as above. Will be fixed in the PR related to #712.

sktime/performance_metrics/forecasting/_functions.py
```diff
@@ -1,139 +1,2181 @@
...
+    "mean_relative_absolute_error",
+    "median_relative_absolute_error",
+    "geometric_mean_relative_absolute_error",
+    "geometric_mean_relative_squared_error",
+]
+
+EPS = np.finfo(np.float64).eps
+
+
+def weighted_geometric_mean(x, sample_weight=None, axis=None):
```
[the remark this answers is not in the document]
RNKuhns: Yep, I"ll update that in final push.

sktime/performance_metrics/tests/test_performance_metrics_forecasting.py
```diff
@@ -2,64 +2,524 @@
...
+    MedianSquaredError,
+    MeanAbsolutePercentageError,
+    MedianAbsolutePercentageError,
+    MeanSquaredPercentageError,
+    MedianSquaredPercentageError,
+    MeanRelativeAbsoluteError,
+    MedianRelativeAbsoluteError,
+    GeometricMeanRelativeAbsoluteError,
+    GeometricMeanRelativeSquaredError,
+    # relative_loss,
```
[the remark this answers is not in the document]
RNKuhns: I'll update in final push.

sktime/performance_metrics/tests/test_performance_metrics_forecasting.py
```diff
@@ -2,64 +2,524 @@
...
+    # TEST_YS,
+    # TEST_YS_ZERO,
+    Y_TEST_CASES,
+)
 
-from sktime.performance_metrics.forecasting import mape_loss
-from sktime.performance_metrics.forecasting import smape_loss
 
-from sktime.performance_metrics.tests._config import TEST_YS, TEST_YS_ZERO
+# Dictionary mapping functions to the true loss values
```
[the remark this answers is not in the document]
RNKuhns: Sure, I'll add that in final push from my local version. The "official" ones came from me calculating the same functions in excel.

sktime/performance_metrics/tests/_config.py
```diff
@@ -2,21 +2,155 @@
...
+    _make_series(n_timepoints=50, random_state=RANDOM_SEEDS[0]),
+    _make_series(n_timepoints=1, random_state=RANDOM_SEEDS[0]),
     pd.Series([0.0000001, 0.0000002, 0.0000003]),
 ]
 
-
 TEST_YS_ZERO = [pd.Series([0.0, 0.0, 0.0])]
+# Create specific test series to verify calculated performance metrics match
+# those calculated externally
+Y1 = np.array(
```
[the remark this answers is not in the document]
RNKuhns: I've got this done locally. Will include in final push.

sktime/performance_metrics/tests/test_performance_metrics_forecasting.py
RNKuhns: This will be handled by functionality in follow-on PR related to #712.

examples/01_forecasting.ipynb
```diff
@@ -360,7 +376,7 @@
     "forecaster.fit(y_train)\n",
     "y_pred = forecaster.predict(fh)\n",
     "plot_series(y_train, y_test, y_pred, labels=[\"y_train\", \"y_test\", \"y_pred\"])\n",
-    "smape_loss(y_pred, y_test)"
+    "mean_absolute_percentage_error(y_pred, y_test, symmetric=True)"
```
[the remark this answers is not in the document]
RNKuhns: Updated default to symmetric=True

README.rst
```diff
@@ -114,16 +114,16 @@ Forecasting
...
 
     y = load_airline()
     y_train, y_test = temporal_train_test_split(y)
     fh = ForecastingHorizon(y_test.index, is_relative=False)
     forecaster = ThetaForecaster(sp=12)  # monthly seasonal periodicity
     forecaster.fit(y_train)
     y_pred = forecaster.predict(fh)
-    smape_loss(y_test, y_pred)
-    >>> 0.08661468139978168
+    mean_absolute_percentage_error(y_test, y_pred, symmetric=True)
```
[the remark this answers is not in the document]
RNKuhns: I had symmetric=False as the default. I've switched it though so now symmetric=True is the default. I'll edit this file and other examples/docs.

RNKuhns:
@mloning I updated the PR to fix the docs stuff we talked about but now it isn't building successful. Examples error seems to be related to NumPy array error related to AutoArima and Pyramid' arima soft dependency. Seeing similar error messages in a forecaster's in the regular build. 

Didn't change anything that would affect that on my end.

mloning:
@RNKuhns main should now be fixed after merging #828

mloning (dismissed):

mloning:
LGTM - @fkiraly @aiwalter let us know if you still want to review

fkiraly (commented):
@RNKuhns, really nice work! A lot of metrics supported here, impressive!

I have one question though - I'm slightly worried about interface inconsistency. There are classes in `_classes` and functions in `_functions`, with a separate interface. Why that?

My other comment on the class design is that the constructor in the classes does not seem to adhere with the sklearn convention that:
* named arguments are hyper-parameters
* tags are not constructor arguments but class attributes

For example `greater_is_better` should be a property that isn't set by the constructor, because it's implied by which loss/score it is and which parameters you choose (possibly). Similarly, shouldn't `name` be also completely determined and not constructor settable?

RNKuhns:
@fkiraly want to make sure I'm following on the class design part, so thanks in advance for putting up with all my questions! 

For the _classes.py versus _functions.py I maintained the separation there so I'll defer to @mloning or you about what led to that setup to start; likewise, I'll defer to @mloning on the name part being completely determined (we could probably do that pretty easily, but I don't know the backstory of why the current setup was decided on). 

You mention that that per sklearn convention that named arguments are hyper-parameters and tags are not constructor arguments but class attributes. I want to clarify your meaning there.

* In terms of tags being class attributes and not constructor arguments, I've got a couple follow-up questions:
    * Which arguments are you referring to beyond `greater_is_better` and `name`?
    * Since the classes themselves  redefine `__init__` and don't include `greater_is_better` and `name` aren't we effectively excluding these from the user-facing constructor arguments?
        * The way its set up a metric like `MeanSquaredScaledError` (see below) only has `square_root` as user facing constructor arg, but has `greater_is_better` and `name` as attributes since they aren't exposed to end-user but passed to constructor of the underlying `ScaledSquaredMetricFunctionWrapper`.
  
Note that Scikit-learn looks like it is passing `sign` variable as argument to the underlying scorer constructors in `make_scorer` and it is derived from `greater_is_better` parameter in `make_scorer`.
    
```python
class MeanSquaredScaledError(ScaledSquaredMetricFunctionWrapper):
    """Mean squared scaled error (MSSE) `square_root` is False or
    root mean squared scaled error (RMSSE) if `square_root` is True.
    MSSE and RMSSE output is non-negative floating point. The best value is 0.0.
    """

    def __init__(self, square_root=False):
        name = "MeanSquaredScaledError"
        func = mean_squared_scaled_error
        greater_is_better = False
        super().__init__(
            func=func,
            name=name,
            greater_is_better=greater_is_better,
            square_root=square_root,
        )
```
In terms of the hyper-parameter part of your question, what are we considering a hyper-parameter? Things like `square_root` control the returned metric in so far as the parameter affects the returned value. Does that make it a "hyper-parameter" even though it isn't something that one would reasonably tune?

On a seperate but related note, do we want to make the underlying different MetricFunctionWrapper classes private? For instance `_ScaledSquaredMetricFunctionWrapper`  instead of since `ScaledSquaredMetricFunctionWrapper` we aren't really expecting these to be user facing?

mloning:
Thanks for the review @fkiraly!

Having classes and functions comes from scikit-learn I think, where all performance metrics are principally functions, but you wrap them in a scorer when passing them to higher-level functionality like model evaluation or tuning. 

In terms of software design, even if we only expose classes, it still makes sense to encapsulate performance metrics in functions I think, and then wrap them in classes to attach additional information like `greater_is_better`. This seems to be the common pattern in other libraries like `pytorch` too.

Having said that, `MetricFunctionWrapper` should be private and only used internally for constructing classes. We should provide a public `make_scorer` (or some other name) function to allow users to construct classes out of functions.

fkiraly:
>Having classes and functions comes from scikit-learn I think, where all performance metrics are principally functions, but you wrap them in a scorer when passing them to higher-level functionality like model evaluation or tuning.

I do agree with that, but I feel they are not homogenously exposed - which contributes to interface inhomogenity.

>Having said that, `MetricFunctionWrapper` should be private and only used internally for constructing classes.

That's indeed an instance of what I mean above.
Template metrics can be public, but factories should probably be private. There should also be public classes corresponding to concrete metrics - for every function as well, no?

>In terms of software design, even if we only expose classes, it still makes sense to encapsulate performance metrics in functions I think, and then wrap them in classes to attach additional information like `greater_is_better`. This seems to be the common pattern in other libraries like `pytorch` too.

Also agreed, but properties shouldn't be constructor parameters (it's not something you can choose/set, but an intrinsic property - a static class variable).

sktime/performance_metrics/forecasting/_classes.py
```diff
@@ -69,16 +68,23 @@ def __call__(self, y_true, y_pred):
...
         -------
         loss : float
-            Calculated loss metric
+            Calculated loss metric. If `greater_is_better` attribute is True
+            returns the negative of the metric. If `greater_is_better` attribute
+            is False the metric is returned.
         """
-        return self._func(y_true, y_pred)
+        sign = 1 if self.greater_is_better else -1
+        return sign * self._func(y_true, y_pred)
```
mloning, line 76: Can we undo this change? 

We currently use `greater_is_better` directly when selecting results in tuning. 

https://github.com/alan-turing-institute/sktime/blob/7343dc36566935b42167878c45d570099af96dab/sktime/forecasting/model_selection/_tune.py#L287-L288

That seems to be the only place where we currently use `greater_is_better`.

mloning (commented):
One last change from my side!

mloning (approved):
Looks good to be merged to me! 

@fkiraly let me know if you want to take another look before we merge.

sktime/performance_metrics/forecasting/_classes.py
```diff
@@ -1,54 +1,1304 @@
...
+        -------
+        loss : float
+            Calculated loss metric. If `greater_is_better` attribute is True
+            returns the negative of the metric. If `greater_is_better` attribute
+            is False the metric is returned.
+        """
+        return self._func(y_true, y_pred)
+
+
+class _PercentageErrorMixIn:
```
fkiraly, line 78: minor point: "I" should not be capitalized in "mixin"

sktime/performance_metrics/forecasting/_classes.py
```diff
@@ -1,54 +1,1304 @@
...
+    will not give infinite or undefined values unless the training data
+    is a flat timeseries. In this case the function returns a large value
+    instead of inf.
+
+    Works with multioutput (multivariate) timeseries data
+    with homogeneous seasonal periodicity.
+
+    Parameters
+    ----------
+    sp : int
```
fkiraly, line 333: why does this `sp` parameter appear in docstrings, like here and elsewhere? It does not seem to be a hyper-parameter (settable in `__init__`)

fkiraly: (understanding that it's an attribute set by some mixins)

fkiraly:
Nice!

I have a few comments:
* the non-optional call signature of the functions is inconsistent and violates the "uniformity" pattern - some functions require `y_pred_benchmark`. I would recommend to let all functions to accept this as an argument (e.g., by using a kwargs catch argument), and simply ignoring it.
* I would strongly suggest to loop `multioutput` and `sp` parameters through to the class constructors - otherwise you can't use classes with multi-output metrics; perhaps even `horizon_weight` (though I'm not sure about that one)
* I would suggest that classes inherit from `sklearn` `BaseEstimator`, otherwise you can't tune parameters in metrics, e.g., when using regularization or penalty parameters in the metric to tune for the same, unregularized/unpenalized metric as a generalization criterion

Minor issues:
* strange capitalization of mixin (should be `Mixin` not `MixIn`)
* `sp` appears in docstrings but not in constructor

RNKuhns:
@fkiraly thanks for the feedback. I've made most of the changes you've discussed.

- Good catch on `sp` in docstrings and not constructor. I've got that updated in the 4 scaled metrics that use it.
- I'm updating the mixin capitalization now (I was thinking 2 words like mix in are MixIn per CapWords Python class naming conventions, but I see that Scikit-learn uses Mixin).
- _MetricFunctionWrapper is already setup to inherit from BaseEstimator. The other metrics inherit from _MetricFunctionWrapper, so they should all have its attributes/methods built-in.

I haven't made the update with `y_pred_benchmark` because Markus and I are planning to discuss how to deal with `y_train` and y_pred_benchmark` as a follow-on PR (see #712, I believe we noted `y_pred_benchmark` in the discussion even though the title just mentions `y_train`).  

We could also include the potential  inclusion of `multioutput` and `horizon_weight` in this PR; however, I'm open to adding them to each functions constructor since they are parameters of all the underlying functions. It would be really quick to do that (but for the time being we aren't actually using that functionality so that could also be a future PR to add it).

mloning (approved):
I'd prefer to merge this now and add additional functionality in separate PRs.

I'm not sure if all functions should accept `y_train` and `y_pred_benchmark`, but let's discuss this in a separate issue/PR.

mloning:
@fkiraly any more comments? Otherwise I'll merge this tomorrow :)

mloning:
Now merged @RNKuhns - thanks again 🎉 👍

mloning:
@jerronl `smape_loss` was renamed to `mean_absolute_percentage_error`

mloning:
Perhaps we should add deprecation warnings for the old loss functions?

PR #811, opened 2021-04-14
Fix counting for Github's automatic language discovery

### LINKED ISSUE #809: [BUG] problematic language statistics shown in the homepage
Hi, I have noticed that `.ts` files in the folder `datasets` are tracked by Github Linguist, which is not the case in the past.

As a result, the language statistics shown in the homepage are somehow mis-leading. Maybe we should add `jquery.js -linguist-vendored` as a new line in `.gitattributes`, according to this [markdown file](https://github.com/github/linguist/blob/master/docs/overrides.md#vendored-code).

xuyxu:
#### Reference Issues/PRs

resolves #809

* Update config in `.gitattributes` for Github Linguist to conduct language discovery more appropriately

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,3 @@
+docs/* linguist-documentation
 sktime/datasets/data/* linguist-vendored
+jquery.js -linguist-vendored
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,2 @@
-docs/* linguist-documentation
 sktime/datasets/data/* linguist-vendored
 jquery.js -linguist-vendored
--- a/sktime/forecasting/compose/tests/test_reduce.py
+++ b/sktime/forecasting/compose/tests/test_reduce.py
@@ -12,6 +12,7 @@ from sklearn.base import RegressorMixin
 from sklearn.dummy import DummyRegressor
 from sklearn.linear_model import LinearRegression
 from sklearn.pipeline import make_pipeline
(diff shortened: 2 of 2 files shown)


xuyxu:
cc @mloning

mloning:
Thanks @xuyxu

mloning:
Looking again at the documentation: do we need to add this last line too?

```
docs/* linguist-documentation
docs/formatter.rb -linguist-documentation
```

mloning:
Actually seems like docs is excluded automatically - so perhaps just adding your originally suggested fix should do the work.

xuyxu:
> Actually seems like docs is excluded automatically - so perhaps just adding your originally suggested fix should do the work.

You are right, `.rst` files are not tracked now. So shall we remove the first line?

mloning:
Yes let's just add the `jquery.js -linguist-vendored` - the entire `docs` folder is excluded actually.

xuyxu:
Done. I am not sure if this problem arises from the missing line on `jquery.js`. Lingust has released the version 7.14 two days ago, and it looks like many repos also suffer from this problem after that release.

mloning (approved):

mloning:
@xuyxu I merged it now, let's see if this fixes the issue

xuyxu:
Well, it looks like the problem still exists :-(

I will open an issue in Github linguist to see if they could help us ;-)

PR #812, opened 2021-04-14
Fix counting for Github's automatic language discovery

### LINKED ISSUE #809: [BUG] problematic language statistics shown in the homepage

xuyxu:
#### Reference Issues/PRs

resolves #809

Take suggestions from this [discussion](https://github.com/github/linguist/discussions/5323#discussioncomment-610376) to recursively mark all downstream directories as vendored.

As mentioned in this discussion, in order to trigger the language discovery appropriately, we have to push other changes along with this PR. Therefore, I have modified the config for all-contributors, removing my previous account that causes the duplicates in [CONTRIBUTORS.md](https://github.com/alan-turing-institute/sktime/blob/main/CONTRIBUTORS.md).

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -606,17 +606,6 @@
       ]
     },
     {
-      "login": "AaronX121",
-      "name": "Yi-Xuan Xu",
-      "avatar_url": "https://avatars2.githubusercontent.com/u/22359569?v=4",
-      "profile": "https://github.com/AaronX121",
-      "contributions": [
-        "code",
-        "test",
-        "doc"
-      ]
-    },
-    {
       "login": "huayicodes",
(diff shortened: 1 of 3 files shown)


mloning (approved):
Thanks @xuyxu for following-up on this!

xuyxu:
Great, it looks like the problem solved.

PR #814, opened 2021-04-15
Accommodating Windows Length as a float

thayeylolu:
https://github.com/alan-turing-institute/sktime/issues/795

**Task** 

- [x] Understand the use of windows_length

- [x] Edited the windows_length in the codebase to include float
- [x] included condition for windows length as a float data type in the check_window_length function.

### DIFF
--- a/sktime/forecasting/compose/_reduce.py
+++ b/sktime/forecasting/compose/_reduce.py
@@ -83,7 +83,7 @@ def _sliding_window_transform(
     # There are different ways to implement this transform. Pre-allocating an
     # array and filling it by iterating over the window length seems to be the most
     # efficient one.
-    window_length = check_window_length(window_length)
+    window_length = check_window_length(y, window_length)
 
     z = _concat_y_X(y, X)
     n_timepoints, n_variables = z.shape
@@ -163,7 +163,7 @@ class _Reducer(_BaseWindowForecaster):
         self._set_fh(fh)
 
         self.step_length_ = check_step_length(self.step_length)
-        self.window_length_ = check_window_length(self.window_length)
+        self.window_length_ = check_window_length(y, self.window_length)
 
         self._fit(y, X)
         self._is_fitted = True
--- a/sktime/forecasting/compose/_reduce.py
+++ b/sktime/forecasting/compose/_reduce.py
@@ -83,7 +83,7 @@ def _sliding_window_transform(
     # There are different ways to implement this transform. Pre-allocating an
     # array and filling it by iterating over the window length seems to be the most
     # efficient one.
-    window_length = check_window_length(window_length, y)
+    window_length = check_window_length(window_length, y=y)
 
     z = _concat_y_X(y, X)
     n_timepoints, n_variables = z.shape
@@ -163,7 +163,7 @@ class _Reducer(_BaseWindowForecaster):
         self._set_fh(fh)
 
         self.step_length_ = check_step_length(self.step_length)
-        self.window_length_ = check_window_length(self.window_length, y)
+        self.window_length_ = check_window_length(self.window_length, y=y)
 
         self._fit(y, X)
         self._is_fitted = True
--- a/sktime/forecasting/compose/_reduce.py
+++ b/sktime/forecasting/compose/_reduce.py
@@ -83,7 +83,7 @@ def _sliding_window_transform(
     # There are different ways to implement this transform. Pre-allocating an
     # array and filling it by iterating over the window length seems to be the most
     # efficient one.
-    window_length = check_window_length(window_length)
+    window_length = check_window_length(window_length, y)
 
     z = _concat_y_X(y, X)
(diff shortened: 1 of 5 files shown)


sktime/forecasting/base/_sktime.py
```diff
@@ -589,9 +589,10 @@ def _set_fh(self, fh):
 class _BaseWindowForecaster(_SktimeForecaster):
     """Base class for forecasters that use """
 
-    def __init__(self, window_length=None):
+    def __init__(self, length_y, window_length=None):
         super(_BaseWindowForecaster, self).__init__()
-        self.window_length = window_length
+        self.length_y = len(length_y)
+        self.window_length = window_length * length_y
```
mloning, line 595: Rather than passing y to the constructor (`__init__`), we would want to compute the window length when we call `split`

Ifeanyi30 (commented):
The `.__init__()` method in the `super()` should also have the `length_y` as an argument.

sktime/forecasting/base/_sktime.py
thayeylolu: Thank you @mloning.

thayeylolu: I am thinking of adding the n_timepoints as an argument in check_windows_length. This way, I would not have to edit all instances where `windows_length = check_windows_length()`

Ifeanyi30 (commented):
looks great, you can just remove the `0 <` in `if not valid and (isinstance(window_length, float) and 0 < window_length < 1):` It does not really do much in the code.

And the string for the `raise ValueError()` should be a single statement like :
`f"`{name}` must be a positive integer >= 1,\
   float between 0 and 1, or None but found: {window_length}"`

Ifeanyi30 (commented):
Please add documentation for the function arguments. Thanks

thayeylolu:
@Ifeanyi30  Thank you for reviewing the changes I included. I shall make necessary corrections and inclusions.

thayeylolu:
> looks great, you can just remove the `0 <` in `if not valid and (isinstance(window_length, float) and 0 < window_length < 1):` It does not really do much in the code.
> 
> And the string for the `raise ValueError()` should be a single statement like :
> `f"`{name}` must be a positive integer >= 1,\ float between 0 and 1, or None but found: {window_length}"`

The condition `0 < window_length` checks if window_length is a negative. And we do not want negative floats.
pre-commit flake8 hook checks if the length of each line is more than 88 characters. Which is why I had to write the formatted strings on different lines. Although `\` is a typo.

Ifeanyi30:
Well that's true for the window_length range but it would be better the
error message is enclosed in just one pair of quotation marks.

thayeylolu:
image
Initially, I enclosed the error message in one pair of quotation marks. It failed the check and this is the error message I got

Ifeanyi30:
You could use a multiple line comment quote to enclose the string (""" """)

Ifeanyi30 (dismissed):

Ifeanyi30:
In the file sktime/transformations/panel/segment.py there should be an if/else statement for the line 179 since the `y` argument could be a None type

thayeylolu:
Thank you @Ifeanyi30 . I thought that could be cause of the failed test as well.

Ifeanyi30:
<pre>
If y is not None:
    Self.min_window_length = check_window_length(y,
self.min_window_length...
    Self.max_window_length = check_window_length(y,
self.max_window_length...
else:
    Self.min_window_length = check_window_length(self.min_window_length...
    Self.max_window_length = check_window_length(y,
self.max_window_length...</pre>

thayeylolu:
@mloning @Ifeanyi30 I added `y` as an optional parameter and set it's default as `None` in the `check_windows_length` function. Prior to this recent commit, I thought refactoring `fit()` method in `RandomIntervalSegmenter` class would be a good decision but it broke the code. So, I ended up editing `check_windows_length`.

Ifeanyi30:
I think you should revert back to the first commit with no errors, then you
can first again from there.

thayeylolu:
All checks have passed. I don't think I need to revert.

Ifeanyi30:
` self.min_length = check_window_length(self.min_length, y, "min_length")
    self.min_length = check_window_length(self.max_length, y, "max_length")`

you passed both instances to the self.min_length, the max_length should be passed to self.max_length

thayeylolu:
thank you for notifying me. I have fixed it

thayeylolu:
@mloning  @Ifeanyi30  I cannot seem to point out why the checks- test example was not successful. The changes I made seem fine to me. I would appreciate your feedbacks. Thank you

thayeylolu:
image
It is failing the test because of this dependency.

aiwalter:
@thayeylolu see #825 . I have a hot fix there fot the `pmdarima` issue... Should be approved soon I hope :)
___

fkiraly:
now worked on in #827

PR #825, opened 2021-04-18
Added cutoff to BaseGridSearch to use any grid search inside evaluate…

aiwalter:
`BaseGridSearch` did not have an attribute `cutoff` and could therefore not be used as a forecaster in `evaluate` funciton.


aiwalter:
opened a bug issue at `pmdarima`:
https://github.com/alkaline-ml/pmdarima/issues/423
___

aiwalter:
@mloning this contains a hotfix for the `pmdarima` issue.
___

sktime/forecasting/model_selection/_tune.py
```diff
@@ -302,6 +302,16 @@ def evaluate_candidates(candidate_params):
...
 
+    @property
+    def cutoff(self):
+        """The time point at which to make forecasts
+
+        Returns
+        -------
+        cutoff : pd.Period, pd.Timestamp, int
+        """
+        return self.best_forecaster_.cutoff
```
mloning, line 313: what happens if the algorithm hasn't been fitted yet?

aiwalter: it would throw an error that there is no `best_forecaster_` i think, is this ok like that?

aiwalter: probably best to check `self.check_is_fitted()`

mloning (approved):
I'll merge this now, but think that we still need to factor out meta algorithms better, rather than implement this for each algorithm.

PR #827, opened 2021-04-19
Including floating numbers as a window length

thayeylolu:
https://github.com/alan-turing-institute/sktime/issues/795

**Task** 

- [x] Understand the use of windows_length

- [x] Edited the windows_length in the codebase to include float
- [x] included condition for windows length as a float data type in the check_window_length function.
- [x] write a test case to validate check_window_length function


sktime/utils/validation/__init__.py
```diff
@@ -38,12 +38,45 @@ def check_n_jobs(n_jobs):
         return n_jobs
 
 
-def check_window_length(window_length, name="window_length"):
+def check_window_length(window_length, y=None, name="window_length"):
```
mloning, line 41: rather than passing `y`, I would pass `n_timepoints` and compute `n_timepoints = y.shape[0]` outside of the function.

mloning (commented):
Thanks @thayeylolu - I will take a closer look later. 

We need some tests for this too, to make sure it works as expected. We use pytest for testing. I would start by writing a test for the `check_window_length` function.

sktime/utils/validation/__init__.py
thayeylolu: I have edited the function and computed `n_timepoints` outside the function. Should I go ahead with the test?

sktime/utils/validation/__init__.py
```diff
@@ -13,6 +13,12 @@ def is_int(x):
     return isinstance(x, (int, np.integer)) and not isinstance(x, bool)
 
 
+def is_float(x):
+    """Check if x is of integer type, but not boolean"""
+    # boolean are subclasses of integers in Python, so explicitly exclude them
+    return isinstance(x, (float, np.floating)) and not isinstance(x, bool)
```
mloning, line 19: booleans are not floats 
```
    return isinstance(x, (float, np.floating))
```

sktime/utils/validation/__init__.py
```diff
@@ -38,12 +44,44 @@ def check_n_jobs(n_jobs):
...
+        The number of training set used for splitting
+        Window length for transformed feature variables
+
+    Returns
+    -------
+    window_length: int
+    """
     if window_length is not None:
-        if not is_int(window_length) or window_length < 1:
+
```
mloning, line 61: Looking at this function, it still seems a bit complicated to me. I would try to simplify it further. One thing I would try to get rid of is the `valid` flag.

mloning (commented):
Thanks again @thayeylolu - I left a few comments below. 

We also need to write some unit tests to make sure it works as expected - let me know if you need help with that.

sktime/utils/validation/__init__.py
mloning: Yes please write some unit testing for this! 👍

thayeylolu: Hello Markus. I have started writing the test
but I am unsure if I am also going to include asserts when the window_length is not a valid input.

sktime/utils/validation/__init__.py
thayeylolu: Thank you @mloning . I have made changes to it

sktime/utils/validation/__init__.py
thayeylolu: I have refactored the function and removed `valid` from it

sktime/utils/validation/__init__.py
thayeylolu: image

mloning: For invalid inputs you can use something like this: 
```python
with pytest.raises(ValueError): 
    check_window_length("bad_window_length")
```

mloning: You can also use `pytest.mark.parametrize()` to iterate over a test function

mloning: `from sktime.utils.validation.__init__` is the same as `from sktime.utils.validation`

thayeylolu:
commit `24ff0ee` failed because `AttributeError: 'list' object has no attribute 'shape'`. I decided to assign `Len(y)` to `n_timepoints` instead of `y.shape[0]`

thayeylolu:
Hello @mloning . I have made adjustments and I am ready for the review. :)

thayeylolu:
@Ifeanyi30 Apologies for not notifying you earlier. I created another branch because the previous one was rough. Kindly review. Thank you

sktime/utils/validation/tests/test_check_window_length.py
```diff
@@ -1,10 +1,31 @@
...
-def test_check_window_length(window_length):
-    assert check_window_length(0.2) == 1  # true
-    assert check_window_length(43) == 43  # True
+def test_check_window_length():
+    # assert check_window_length(0.2) == 1  # true
+    # assert check_window_length(43) == 43  # True
     assert check_window_length(None) is None  # True
+
+
+def test_raises_exception_on_string_arg():
```
mloning, line 20: Take a look at `pytest.mark.parametrize()` to see how you can further simplify these tests

Something like this should work:
```python
@pytest.mark.parametrize("arg", ["str", -1, -0.1])"
def test_window_length_bad_arg(arg):
    ...
```

thayeylolu: Thats interesting. I would take a look at it

sktime/utils/validation/__init__.py
```diff
@@ -38,12 +43,33 @@ def check_n_jobs(n_jobs):
...
+        if n_timepoints is None:
```
mloning, line 61: Why do we set n_timepoints to 1 here? I think perhaps we should:
* ignore n_timepoints if window_length is a positive integer
* raise an error if n_timepoints is None if window_length is a positive float (instead of setting it to 1)

mloning (commented):
Thanks for the latest changes @thayeylolu - I left a few more comments

sktime/utils/validation/tests/test_check_window_length.py
```diff
@@ -0,0 +1,23 @@
...
+# -*- coding: utf-8 -*-
+import pytest
+
+from sktime.utils.validation import check_window_length
+
+
+@pytest.mark.parametrize(
+    "arg,expected",
+    [
+        (0.2, 1),
```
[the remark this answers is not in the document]
thayeylolu: Noted

sktime/utils/validation/__init__.py
thayeylolu: Noted

thayeylolu:
No errors :]

Ifeanyi30 (approved):

mloning (approved):

mloning:
@thayeylolu the failures seem unrelated to the changes in this PR, I'll check back later.

mloning:
Now merged @thayeylolu 🎉

Issue #798, opened 2021-04-11
Implement PyOD transformer adapter


satya-pattnaik:
**Is your feature request related to a problem? Please describe.**
A sktime Series Transformer which wraps a ADTK Pipeline for outlier detection. 

**Describe the solution you'd like**
This [notebook](https://github.com/satya-pattnaik/sktime-mentorship-program/blob/main/anomaly_detection/sktime_adtk.ipynb) has an initial approach

**Additional context**
https://arundo-adtk.readthedocs-hosted.com/en/stable/

fkiraly:
Unfortunately, ADTK is not permissive open source, so we can't make it a dependency.
I'm also not sure how well it is supported still.

Though, if you want outlier detection, I suggest `pyod`:
https://github.com/yzhao062/pyod

So, we could wrap the `pyod` interface (for i.i.d./tabular data) for use with temporal scenarios via reduction?

satya-pattnaik:
This looks to be a good one, it has got some techniques specific for time series - https://github.com/SeldonIO/alibi-detect

fkiraly:
yes, indeed! Thanks for making us aware of this. @mloning FYI, also good to have on the related work page?

They do not seem to have a pipeline structure though - so it might be a good idea to construct:
* a seldon interface, with a cleaner API design for scitype handling (e.g., time series input or tabular input? annotators and detectors)
* some functionality similar to sklearn `Pipeline` or ADTK `PipeNet`; on that note, I believe @viktorkaz wanted to build something similar to that

mloning:
@satya-pattnaik looks interesting, please add to `docs/source/related_software.rst`

ViktorKaz:
For the last several weeks I've been working on a design of a pipeline which can also be used for outlier detection. You can find the preliminary code [here](https://github.com/ViktorKaz/sktime/tree/pipeline/sktime/pipeline). This is related to Issue https://github.com/alan-turing-institute/sktime/issues/763.

fkiraly:
so... how do we proceed with this?
There are multiple non-exclusive options I can see:
* interface to `pyod`
* interface to `alibi-detect`
* extend @ViktorKaz's pipeline design to cover generic temporal on-line anomaly detection

satya-pattnaik:
I am already on it. I am developing the interface to PyOD, I had a call with @mloning  to get the design sketch(which I will update as an enhancement proposal ).

RNKuhns:
[...]

PR #838, opened 2021-04-25
Added Annotation Base Class

satya-pattnaik:
#### Reference Issues/PRs

With regards to #798 

Creates a Base Class for Annotations.

If the Base Estimator contains the proper code and is in the proper directory.
##### For all contributions

##### For new estimators


sktime/annotation/tests/test_all_annotators.py
```diff
@@ -0,0 +1,41 @@
...
+        Zt.iloc[1] = True
+        return Zt
+
+
+test_annotator = TestAnnotator()
+
+
+def test_output_is_series():
+
+    data = pd.Series(range(5))
```
mloning, line 26:
```python
ANNOTATORS = all_estimators("annotator", return_names=False)

@pytest.mark.parametrize("Estimator", ANNOTATORS):
def my_test_function(Estimator):
  
  args = _make_args(estimator, "fit")
  estimator.fit(*args)
  # assert somthing

  args = _make_args(estimator, "transform")
  estimator.transform(*args)
  # assert something
```
* need to update `all_estimators(...)`
* need to update `_make_args(...)` to handle annotators as estimators

sktime/annotation/tests/test_all_annotators.py
```diff
@@ -0,0 +1,41 @@
+# -*- coding: utf-8 -*-
+import pandas as pd
+import numpy as np
+from sktime.annotation.base._base import BaseAnnotator
+
+
+class TestAnnotator(BaseAnnotator):
```
mloning, line 7: create mock annotator class in `sktime/annotation/base/_mock.py` to make sure all the unit testing functionality works
```python
class MockAnnotator(BaseAnnotator):
    # mock class, implement basic interface fit/predict
```

mloning (commented):
Hi @satya-pattnaik as discussed in our meeting:

sktime/annotation/base/_mock.py
```diff
@@ -0,0 +1,18 @@
...
+
+class MockAnnotator(BaseAnnotator):
+    def __init__(self):
+        pass
+
+    def fit(self, Z, X=None):
+        self._is_fitted = True
+        return self
+
+    def transform(self, Z, X=None):
```
mloning, line 13: Should be `predict` as in BaseAnnotator above

sktime/utils/_testing/estimator_checks.py
```diff
@@ -556,6 +559,8 @@ def _make_fit_args(estimator, **kwargs):
         fh = 1
         X = None
         return y, X, fh
+    elif isinstance(estimator, BaseAnnotator):
```
mloning, line 562: Yes, this is good, we also need to modify `_make_predict_args` accordingly

```python
def _make_predict_args(estimator, **kwargs):    
    ...
    elif isinstance(estimator, BaseAnnotator):
        return make_annotation_problem(n_timepoints=10, **kwargs)
```

mloning (commented):
Comments as discussed in our call. 

For the test on bad input arguments: 

```python
@pytest.mark.parametrize("arg", BAD_INPUT_ARGS)
def test_bad_input_raises_error(arg):
    msg = "expected error msg"
    with pytest.raises(ValueError, match=msg):
        check_X(arg)

```

fkiraly:
@satya-pattnaik, @mloning, great work!

I would, however, suggest to think about the following:

* I would strongly suggest that the annotated time series `z` does *not* need to have the same indices!
* I don't think we need exogeneous data - can we not just assume that the series to segment is multivariate to start with? If the annotator insists to focus on one variable, then this can be passed as info to init.
* I would suggest to cover both supervised and unsupervised annotators, that should be easy with a "meta-base-class". Supervised ones accept a `Y` which are training annotations.
* I would suggest to call the object being annotated `X`, not `Z`.
* I would suggest to adopt the `fit/_fit` design from forecasters - this will make input checks etc easier. We'll probably move there too with TSC, so let's make this uniform!
* for segmentation, I would suggest the convention returns a series with an `pandas.IntervalIndex`; more generally, the docstrings should describe the convention of the input or return objects, if any - implement tests accordingly.
* I strongly suggest that we also think about the panel case - this is important, and covered ,methodologically by key annotation packages in the segmentation domain, like `hmmlearn`. The format I'd suggest is "list-of-X", and for supervised learners "list-of-Y" as a label. Such annotators should also accept the "single-sample" case.

mloning:
Closed in favour of #1021

PR #848, opened 2021-04-28
Fix AutoETS handling of infinite information criteria

### LINKED ISSUE #847: -inf AIC/BIC in AutoETS
Some ETS models from the [statsmodels.tsa.exponential_smoothing.ets.ETSModel](https://www.statsmodels.org/stable/generated/statsmodels.tsa.exponential_smoothing.ets.ETSModel.html) can yield an infinite likelihood. As a result, information criteria like AIC, AICc and BIC will be negative infinity. The AutoETS implementation is picking the model that minimises the chosen information criterion, therefore, in the above case it would choose the infinite likelihood model. For the time series I was playing with this results in many obscure forecasts with a mean MAPE (across splits) of 168%.

Simply ignoring models with -inf Information Criteria is an effective solution to the problem.
For my time-series, applying this fix causes a reduction in mean MAPE (across splits) from 168% to 13%

**pre vs post -inf IC fix**
image  image

ltsaprounis:
#### Reference Issues/PRs
Fixes #847 

AutoETS now ignores models with -inf AIC/BIC/AICc and selects models with finite values in the chosen information criterion.

No new dependencies

##### For all contributions


fkiraly (changes requested):
Thanks, interesting observation on the infinite IC.

My main comment comes from the fact that this is a change in algorithmic behaviour from the default coming out of the model being interfaced.

So instead of mutating functionality in the interface, I would strongly suggest to do one of the following:
* add or change the functionality "upstream", i.e., directly in `statsmodels`
* instead of *replacing* default functionality, make it optional functionality; that means, adding a hyper-parameter `ignore_inf_IC` or similar which if `True` runs your new behaviour. Given the nature of the issue, I might be easily convinced to make this the default even.

@HYang1996, as the original author, what do you think?

HYang1996:
@ltsaprounis Thank you for raising the issue, and @fkiraly thank you for the comment.

@fkiraly When I implemented the best model selection method I was referring to the [R forecast version](https://github.com/robjhyndman/forecast/blob/f0965d594a23fa90b3b56152f52f3a5fbc5240e9/R/ets.R#L424-L433) which does not check for -inf cases. I just checked the implementation in the [fable version](https://github.com/tidyverts/fable/blob/5b953984ff084d2acc314e586355f9b28c849e56/R/ets.R#L70-L72) and it does seem to be excluding the model if the IC is infinite. I think it's reasonable to exclude the -inf cases in the algorithm and set it as the default case of a hyper-parameter as you suggested.

ltsaprounis:
Thanks @fkiraly and @HYang1996  for taking a close look at this PR and for the great feedback.

As described in [[Hyndman, Rob & Khandakar, Yeasmin. (2008)] ](https://www.researchgate.net/publication/222105759_Automatic_Time_Series_Forecasting_The_forecast_Package_for_R/citation/download) some ETS models can be unstable and should be handled with caution, I think that's the source of the problem!

I agree 100%, it's better to add it as a hyperparameter and not "hardcode" it. I've implemented the change and added the default parameter ignore_inf_ic in the model.

sktime/forecasting/ets.py
```diff
@@ -299,14 +305,20 @@ def _fit(error, trend, seasonal, damped):
                 )
             )
 
+            # Store IC values for each model in a list
+            # Ignore infinite likelihood models if ignore_inf_ic is True
+            _ic_list = []
+            for result in _fitted_results:
+                ic = getattr(result[1], self.information_criterion)
+                if ic == -np.inf and self.ignore_inf_ic:
```
mloning, line 313:
```suggestion
                if self.ignore_inf_ic and np.isinf(np.inf):
```

mloning (dismissed):
Looks good to me, one minor comment. Thanks for the quick fix @ltsaprounis!

sktime/forecasting/ets.py
ltsaprounis: Thanks @mloning! np.isinf is more elegant indeed.

mloning (commented):
Thanks @ltsaprounis - before we merge, do you have a small example that recreates the problem which we could use as a unit test?

ltsaprounis:
> Thanks @ltsaprounis - before we merge, do you have a small example that recreates the problem which we could use as a unit test?

Thanks @mloning I've made up a dummy time series that results in infinite ICs and added 2 unit tests to make sure we get the expected behaviour when toggling ignore_inf_ic from True to False. Hope the I've added the tests correctly and in the right place!

sktime/forecasting/tests/test_ets.py
```diff
@@ -52,3 +60,29 @@ def test_airline_allow_multiplicative_trend():
...
+    fitted_forecaster = forecaster._fitted_forecaster
+    # check that none of the information criteria are finite
+    assert (
+        np.isfinite(fitted_forecaster.aic)
+        and np.isfinite(fitted_forecaster.bic)
+        and np.isfinite(fitted_forecaster.aicc)
+    )
+
+
+# Don't ignore infinite IC models when ignore_inf_ic is True
```
mloning:
```suggestion
# Don't ignore infinite IC models when ignore_inf_ic is False
```

sktime/forecasting/tests/test_ets.py
```diff
@@ -52,3 +60,29 @@ def test_airline_allow_multiplicative_trend():
...
 
     assert_array_equal(fit_result_R, fit_result)
+
+
+# Ignore infinite IC models when ignore_inf_ic is True
+def test_inf_ic_true():
+    forecaster = AutoETS(auto=True, sp=52, n_jobs=-1, ignore_inf_ic=True)
+    forecaster.fit(inf_ic_ts)
+    fitted_forecaster = forecaster._fitted_forecaster
+    # check that none of the information criteria are finite
```
mloning:
```suggestion
    # check that all of the information criteria are finite
```

mloning (commented):
Great @ltsaprounis - thanks, two minor changes and then we can merge from my side

sktime/forecasting/tests/test_ets.py
ltsaprounis: I think those and the above comments are fixed in the latest commit (called "fix comments"), unless I'm too sleepy now to tell the difference.

mloning (dismissed):
Yes thanks again @ltsaprounis - LGTM

mloning:
@fkiraly you requested changes, could you please re-review?

ltsaprounis:
Sorry @mloning made a new commit to fix a typo in the doc string. I think you'll need to re-review along with @fkiraly.

fkiraly (approved):
Fine from my side, nice work!

Though, before you commit, you may like to ensure that the rare edge case is captured where all elements of `_ic_list` end up as `nan`, which happens if all IC values are positive or negative infinity.

ltsaprounis:
Thanks @fkiraly!
I think this is a very unlikely case but, better safe than sorry. 

I'm not sure what the right behaviour is in this case but in my opinion we shouldn't return a model and should raise an error.
With the code "as is" the error would be the one from np.nanargmin which is: 
```python
np.nanargmin([np.nan, np.nan])

# ValueError: All-NaN slice encountered
```
If we think that's not informative enough, I would write something like this: 
_(more informative but also more operations)_
```python
information_criterion = "bic"
_ic_list = [np.nan, np.nan]
if all(np.isnan(ic) for ic in _ic_list):
    # if all models have infinite IC raise an error
    raise ValueError("None of the fitted models have finite %s"
                     %information_criterion
                     )
else:
    # Get index of best model
    _index = np.nanargmin(_ic_list)

# ValueError: None of the fitted models have finite bic
``` 

What do you think? Keep it as is or add the additional check?

fkiraly:
I guess that would work? The alternative would be picking one of the parameters. Anyway, seems like a very rare edge case.

ltsaprounis:
> I guess that would work? The alternative would be picking one of the parameters. Anyway, seems like a very rare edge case.

Yes the issue is with a subset of models with a combination of multiplicative and additive parameters. I don't think something like SES will have issues with a non-defined likelihood. 

So @fkiraly / @mloning  what's the verdict add the additional check or leave it as is with the error from np.nanargmin?

mloning:
@ltsaprounis let's add the more informative error message. 

You can you use: `np.all(np.isnan(ic))`

mloning (approved):

ltsaprounis:
Sorry guys - I need to change the author name for the commits to this github profile. Please don't merge yet, I'll make a new commit.

ltsaprounis:
@mloning / @fkiraly / @aiwalter - Sorry for the super mess everyone, I had to use interactive rebase to change my details in the commits. Hope this is alright now. Let me know if I need to do anything else.
Thanks!

sktime/forecasting/ets.py
```diff
@@ -299,14 +305,27 @@ def _fit(error, trend, seasonal, damped):
...
+
             # Select best model based on information criterion
-            # Get index of best model
-            _index = np.argmin(
-                [
-                    getattr(result[1], self.information_criterion)
-                    for result in _fitted_results
-                ]
-            )
+            if np.all(np.isnan(_ic_list)):
```
mloning, line 319: What happens if `self.ignore_inf_ic` is True, then `_ic_list` would be empty right? So should we write:
```suggestion
            if np.all(np.isnan(_ic_list)) or len(_ic_list) == 0:
```

mloning (commented):
Hi @ltsaprounis, one last comment from my side. 

Ignore the failing manylinux builds for now, they will be fixed in #870.

sktime/forecasting/ets.py
ltsaprounis: Yes, that's right. I missed this completely.

mloning (dismissed):
Looks all good now - thanks @ltsaprounis

mloning (approved):
Will merge once all tests have passed again!

ltsaprounis:
Thanks everyone! This was my first pull-request for an open source package and the sktime community and documentation made up for a very smooth and positive experience!

mloning:
@ltsaprounis if you want to stay involved, we will run an onboarding workshop starting on Wed, more details here: https://github.com/sktime/community-council/blob/main/events/sktime-onboarding-days.md

Issue #817, opened 2021-04-16
Add convenience function to plot series, ACF and PACF


RNKuhns:
**Is your feature request related to a problem? Please describe.**
Reviewing a plot of a series, as well as its ACF and PACF functions is a common exploratory data analysis step in time series analysis. 

Currently users would have to individually plot the series using sktime's `plot_series` function and then either import and use statsmodels' `plot_acf `and `plot_pacf` functions. Providing an off-the-shelf method to easily generate a nice looking figure containing all three plots would make basic time series exploratory analysis easier for users.

**Describe the solution you'd like**
A clear and concise description of what you want to happen, ideally taking into consideration the existing toolbox design, classes and methods. 

Creation of a convenience function `plot_correlations` to generate a plot of a time series and its ACF and PACF. 

**Describe alternatives you've considered**

Alternative is to have users continue to use separate functionality from sktime's `plot_series` and statsmodels' `plot_acf `and `plot_pacf`  functions. But this is more cumbersome for users performing a routine exploratory data analysis task. 

**Additional context**

RNKuhns:
I'm planning to submit a Pull Request that contributes this functionality shortly.

RNKuhns:
This was closed by #850.

PR #850, opened 2021-04-30
Add plot_correlations() to plot series and acf/pacf

RNKuhns:
#### Reference Issues/PRs

Adds functionality requested in #817. 

This adds a function to plot with subplots that show a series (re-uses plot_series(), and its ACF/PACF plots. 

Adds dependency on [pytest-mpl](https://github.com/matplotlib/pytest-mpl) for testing plotting functionality against baseline plot output. 

#### Any other comments?

In addition to plot_correlations() the PR makes a tweaks to plot_series() so that if a user supplies an Axe object the plot_series() plot will be created there instead of in a new figure (makes it easy to use plot_series() in other figures with additional subplots). 

##### For all contributions

##### For new estimators


RNKuhns:
@mloning and @aiwalter let me know if the update to `plot_series()` seems reasonable.  The new plotting functionality is pretty straight-forward. 

It just takes a series and:
1. Returns its plot using `plot_series()`
2. Returns plots of ACF and PACF with confidence intervals using statsmodels functionality

The args/kwargs beyond the series to be plotted are just to allow users to manipulate titles of subplots or change underlying ACF/PACF plot options. 

I'm looking into including unit testing for plots and will update this once I've got that figure out.

mloning (approved):
@RNKuhns looks good to me

RNKuhns:
@mloning I believe I've got some useful working tests for the plotting functionality. These include input/output tests as well as tests to compare the generated plots against baseline plots. To do the baseline plot comparison, I'm using [pytest-mpl](https://github.com/matplotlib/pytest-mpl). 

Everything is passing locally on my machine, but I've installed pytest-mpl and we'd need this as part of the build-test process.

Since this would introduce a new build dependency, I've got a few questions on how best to do that: 
1. Should I make a seperate PR or include that in this PR?  
2. To add the testing dependency, would that just be adding `pytest-mpl` to `sktime/build_tools/requirements.txt`?

mloning:
Hi @RNKuhns, I had a look at `pytest-mpl` and a few other discussion threads on the topic. I'm not sure if we want to actually compare images (see for example the answer here https://stackoverflow.com/questions/27948126/how-can-i-write-unit-tests-against-code-that-uses-matplotlib). Running numerical tests on the returned axis object may be a better idea. What do you think? 

I'm happy to try it out on this PR, but I'm always hesitant to add dependencies and potential complications of the CI routines. I think `build_tools/requirements.txt` would be the only place where you would have to add `pytest-mpl`

RNKuhns:
@mloning I've pushed the code with the pytest-mpl dependency and unit test checks for you to take a look at. Note that I also added an option to let users customize the x-axis and y-axis labels in the call to `plot_series`

I was reading more and it seems like the biggest issue with the image comparison tests is the text varying with different versions of FreeType. I've used the option to remove text in pytest-mpl. I also opened up a Issue with them because that option wasn't removing the axis label text (I have a workaround in test_plotting.py that does it for now). 

After you take a look, we can decide whether we want to actually do the image comparison checks. If not, I can remove that dependency and the image comparison unit tests. 

I think the other unit tests give us reasonable coverage if we decide not to use `pytest-mpl`, since they are checking correct errors are raised and that the plotting functions execute without error.

RNKuhns:
@mloning I forgot to mention that I was thinking about the numerical tests on the axis. I'm actually not sure we need to go down that road. I'm also open to not adding the pytest-mpl dependency based on your thoughts after you take a look. 

The approach I took, for the tests outside of the `pytest-mpl` tests is to assume `matplotlib` and `seaborn` will draw the plots correctly based on what is passed as `x` and `y`. So I focused my tests on making sure that the `plot_series` and `plot_correlations` run without error and can be drawn, as well as some checks to verify expected function input/output and correct error raising. Given that those are both pretty heavily used and tested, I don't see that as a bad strategy. 

The only other thing I didn't really check in `plot_series` that I could see being a useful check to include is verifying that the creation of the combined `x` values from the series matches expectation. Let me know if you think it would be useful to add that test too.

RNKuhns:
@mloning I'm fine with removing dependency on `pytest-mpl`. I think we are fine assuming that checks that the plots execute without error (and raise errors as appropriate) should be fine. 

I'll push an update without that later today.

fkiraly:
FYI, @mloning, you can compare to hashes in `pytest-mpl`, that would solve the issue with the images carried around.

But generally, I don't lean either way on this issue.

RNKuhns:
I wanted to explore whether it made sense to include unit tests of the plot themselves. But I'm leaning towards excluding them for the time being. Whether the tests are on the images themselves or hashes, we are really testing the functionality of the underlying plotting library (e.g. matplotlib).

For our purposes, I think that tests that the plotting functions run without error (and the plots can be drawn) and any of our non-plotting logic works as expected should suffice. 

So I'd vote that we exclude the pytest-mpl dependency for the time being. Caveat being that we always have the ability to add it later on if we find we are running into any issues. 

@mloning and @fkiraly are you fine if I go that route and remove the pytest-mpl part of the request? I think other than that, this is probably largely ready to go.

mloning:
@RNKuhns yes I agree with your comments!

RNKuhns:
@mloning I've made the changes I mentioned. Let me know if you see anything you want me to change after you take a look.

sktime/utils/tests/test_plotting.py
```diff
@@ -0,0 +1,164 @@
...
+from sktime.utils.validation.series import VALID_DATA_TYPES
+
+ALLOW_NUMPY = False
+y_airline = load_airline()
+y_airline_true = y_airline.iloc[y_airline.index < "1960-01"]
+y_airline_test = y_airline.iloc[y_airline.index >= "1960-01"]
+series_to_test = [y_airline, (y_airline_true, y_airline_test)]
+
+
+def _plot_series(series, ax=None, **kwargs):
```
mloning, line 18: Why do we need to redefine it here?

sktime/utils/plotting.py
```diff
@@ -105,7 +128,98 @@ def format_fn(tick_val, tick_pos):
...
+        Used to set title of PACF plot.
+
+    Returns
+    -------
+    fig : matplotlib.figure.Figure
+
+    axes : np.ndarray
+        Array of the figure's Axe objects
+    """
+    _check_soft_dependencies("matplotlib", "seaborn")
```
mloning, line 204: `seaborn` does seem to be used here

sktime/utils/tests/test_plotting.py
```diff
@@ -0,0 +1,164 @@
...
+
+@pytest.fixture
+def valid_data_types():
+    valid_data_types = tuple(filter(lambda x: x is not np.ndarray, VALID_DATA_TYPES))
+    return valid_data_types
+
+
+@pytest.mark.parametrize("series_to_plot", series_to_test)
+def test_plot_series_runs_without_error(series_to_plot):
+    _check_soft_dependencies("matplotlib", "seaborn")
```
mloning, line 33: why do we need to import this here, isn't it imported in the function?

mloning: `seaborn` doesn't seem to be needed

sktime/utils/plotting.py
```diff
@@ -105,7 +128,98 @@ def format_fn(tick_val, tick_pos):
...
+    suptitle=None,
+    series_title=None,
+    acf_title="Autocorrelation",
+    pacf_title="Partial Autocorrelation",
+):
+    """Plot series and its ACF and PACF values.
+
+    Parameters
+    ----------
+    lags : int, default = 24
```
mloning, line 163: `series` is missing from docstring

sktime/utils/tests/test_plotting.py
```diff
@@ -0,0 +1,164 @@
...
+def test_plot_series_invalid_marker_kwarg_len_raises_error(series_to_plot):
+    match = """There must be one marker for each time series,
+                but found inconsistent numbers of series and
+                markers."""
+    with pytest.raises(ValueError, match=match):
+        # Generate error by creating list of markers with length that does
+        # not match input number of input series
+        if isinstance(series_to_plot, pd.Series):
+            markers = ["o", "o"]
+        elif isinstance(series_to_plot, pd.DataFrame):
```
mloning, line 70: does it work with data frames? if so, we need to add it to the docstring for `plot_series`, but `series_to_test` doesn't seem to include any dataframe either

sktime/utils/tests/test_plotting.py
```diff
@@ -0,0 +1,164 @@
...
+def test_plot_series_invalid_label_kwarg_len_raises_error(series_to_plot):
+    match = """There must be one label for each time series,
+                but found inconsistent numbers of series and
+                labels."""
+    with pytest.raises(ValueError, match=match):
+        # Generate error by creating list of labels with length that does
+        # not match input number of input series
+        if isinstance(series_to_plot, pd.Series):
+            labels = ["Series 1", "Series 2"]
+        elif isinstance(series_to_plot, pd.DataFrame):
```
mloning, line 88: same as above

sktime/utils/tests/test_plotting.py
```diff
@@ -0,0 +1,164 @@
...
+            labels = [f"Series {i}" for i in range(series_to_plot.shape[1] - 1)]
+        elif isinstance(series_to_plot, tuple):
+            labels = [f"Series {i}" for i in range(len(series_to_plot) - 1)]
+
+        _plot_series(series_to_plot, labels=labels)
+
+
+@pytest.mark.parametrize("series_to_plot", series_to_test)
+def test_plot_series_output_type(series_to_plot):
+    _check_soft_dependencies("matplotlib", "seaborn")
```
mloning, line 98: same as above

sktime/utils/tests/test_plotting.py
```diff
@@ -0,0 +1,164 @@
...
+        [
+            "plot_series with plt.Axes object passed to kwarg ax",
+            f"should return plt.Axes, but returned: {type(ax)}",
+        ]
+    )
+
+
+@pytest.mark.parametrize("series_to_plot", [y_airline])
+def test_plot_correlations_runs_without_error(series_to_plot):
+    _check_soft_dependencies("matplotlib")
```
mloning, line 130: same as above

sktime/utils/tests/test_plotting.py
```diff
@@ -0,0 +1,164 @@
...
+    # TODO: Is it possible to dynamically create the matching str if it includes
+    #       characters that need to be escaped (like .)
+    # match = f"Data must be a one of {valid_data_types}, but found type: {type(Z)}"
+    with pytest.raises(TypeError):
+        plot_correlations(series_to_plot)
+
+
+@pytest.mark.parametrize("series_to_plot", [y_airline])
+def test_plot_correlations_output_type(series_to_plot):
+    _check_soft_dependencies("matplotlib")
```
mloning, line 150: same as above

mloning (commented):
Great @RNKuhns - looks very good already, a few minor comments. Happy to merge afterwards.

RNKuhns:
@mloning, I pushed commits with the changes you mentioned in your comments.

- In plotting.py
    - Added `series` to `plot_correlations()` docstring
    - Removed unnecessary import of `seaborn` in `plot_correlations()`
- test_plotting.py
    - Added note to clarify why I defined `_plot_series()`
    - Removed unnecessary import `seaborn` in some tests
    - Got rid of the `elif` statement for `pd.DataFrame` case in label and marker tests (pd.DataFrames are not valid input to functions)
    - Updated invalid input tests to also check pd.DataFrame case

Note that I left the import of `matplotlib` in the unit tests it was included in. The import was needed in those tests because the tests needed to reference the library outside of the function being tested (for example, to check the output of the function is the correct matplotlib type). Without importing the `matplotlib` library in the test this errored out. 

Also note, when adding the pd.DataFrame invalid input case, I discovered that `check_y()` is raising a `ValueError` for `pd.DataFrame` input (b/c that is what `_check_univariate()` is raising, but `TypeError` for other types of invalid input. I've handled that in the plotting tests. This may be something we want to unifying (pd.DataFrame should raise `TypeError` in `check_y()`), but that should be another PR since it could affect other tests.

mloning (approved):
Great - looks all good to me. Will merge in a few days in case anyone else wants to comment.

mloning:
Now merged!

PR #844, opened 2021-04-27
Added ForecastingPipeline

aiwalter:
#### Reference Issues/PRs
https://github.com/sktime/enhancement-proposals/pull/10

#### What does this implement/fix? Explain your changes.
Draft for multivariate pipelining.
Example:
```python
pipe = ForecastingPipeline(steps=[
    ("imputer", Imputer(), ["y"]),
    ("scaler", TabularToSeriesAdaptor(MinMaxScaler()), ["y", "X"]),
    ("forecaster", NaiveForecaster())
    ]
)
pipe.fit(y_train, X_train)
```
To do:
- [ ] Enable grid search: solve issues with `get_params` and `set_params`
- [ ] Add validation for the `identifier` like `["y", "X"]` has to be checked if `X` is given at all and if `identifier` contains valid strings.

##### For all contributions
- [ ] I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/main/.all-contributorsrc).
- [ ] Optionally, I've updated sktime's [CODEOWNERS](https://github.com/alan-turing-institute/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
- [ ] I've added unit tests and made sure they pass locally.

##### For new estimators
- [ ] I've added the estimator to the online documentation.
- [ ] I've updated the existing example notebooks or provided a new one to showcase how my estimator works.

___
Martin Walter <martin_friedrich.walter@daimler.com>, Mercedes-Benz AG on behalf of Daimler TSS GmbH.
https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md

Issue #763, opened 2021-03-30
Pipeline for Online Directional Forecasts 


ViktorKaz:
# Problem

In financial applications it is common to be interested in predicting at time $t$ the direction of change for an asset at time $t+m$, where $m \geq 1$ is a future time period. Generally speaking, there are three possibilities. In the time period $[t,t+m]$ either the price can go up, down or there will be no change. 

This problem can be decomposed in several smaller, more manageable ones, namely:

1. Dimensionality Reduction and/or Feature Extraction
1. Feature Transformation
1. Changepoint Detection
1. Annotation (Labelling)
1. Temporal cross validation and estimator tuning

 
# Solution
Ideally, we would have a pipeline which chains together the different elements of the algorithm. In contrast to `sklearn` pipelines we will need the capability to chain the different steps in the pipeline not only sequentially but in any giver order. Furthermore, we will need a functionality for transforming and processing both the features and the labels in our dataset.

# Alternatives
`mlfinlab` has functionality with which directional forecasts can be made. However, it is not implemented as a pipeline but rather as a collection of loose functions organised in modules. Furthermore, `mlfinlab` is not open source. 

Other open source toolboxes that tackle at least part of the problem include:
* `ruptures` deals with change point detection, 
* `ADTK` with anomaly detection (specific case of change point detection), 
* `PyOD` was covers outlier detection (specific case of change point detection)

# Additional context

## Background: `mlfinab` example

```Python
#feature reduction and annotation
daily_vol = ml.util.volatility.get_daily_vol(close=dollar_bars['close'], lookback=5)
cusum_events = ml.filters.cusum_filter(dollar_bars['close'], threshold=-0.1)
vertical_barriers = ml.labeling.add_vertical_barrier(t_events=cusum_events, close=dollar_bars['close'], num_days=5)
triple_barrier_events = ml.labeling.get_events(close=dollar_bars['close'],
                                                t_events=cusum_events,
                                                pt_sl=[1, 1], # sets the width of the barriers
                                                target=daily_vol * 2, # twice the daily vol
                                                min_ret=0.005,
                                                num_threads=4,
                                                vertical_barrier_times=vertical_barriers,
                                                side_prediction=None,
                                                verbose=False)
triple_labels = ml.labeling.get_bins(triple_barrier_events, dollar_bars['close'])

#tune and fit an estimator
train_idx, test_idx = train_test_split(np.arange(dataset.shape[0]), shuffle=False, test_size=0.25)
[...]

PR #856, opened 2021-05-02
Network Pipeline

ViktorKaz:
### Reference Issues/PRs
Aimed at addressing [Forecasting Pipelining enhancement proposal.](https://github.com/sktime/enhancement-proposals/blob/main/steps/01_forecasting_api/forecasting-with-exogenous-variables.md#forecasting-pipelining)

This is also with reference to Issue #763 and Issue #844.

### What does this implement/fix? Explain your changes.

This is a draft PR proposing a design for a more flexible "network" pipeline. The general idea of this network pipeline is that the steps comprising the pipeline are not chained, i.e. the output of a step is not necessarily the input for the next step. Instead, the output of any step can serve as input to any other step in the pipeline. 

The steps of the pipeline are specified as tuple following the convention below:
```python
('step_name', Algorithm, arguments)
```
`step_name` is a string with the name of the step
`Algorithm` can be `sktime.transformations.base.BaseTransformer` or `sktime.base.BaseEstimator`.
`arguments` is a dictionary that is substituted in the `fit`, `fit_transform` or `predict` method of the step.

The `arguments` dictionary can use the following key words: `original_X` and `original_y` which refer to the `X` and `y` arguments of  `NetworkPipeline.fit` and `NetworkPipeline.predcit` methods.

The `arguments` dictionary can refer to the output of any step of the pipeline by the name of the step.

The `arguments` dictionary can specify different arguments for `NetworkPipeline.fit` or `NetworkPipeline.predict`

```python
# Forecasting example. Note convention for specifying different arguments 
# depending on whether NetworkPipeline.fit or  NetworkPipeline.predict is called.

pipe = NetworkPipeline(
    steps=[
        ("imputer", Imputer(method="drift"), {"fit": {"Z": "original_y"}, "predict": None}),
        ("forecaster", NaiveForecaster(strategy="last", sp=12), {"fit": {"y": "imputer"}, "predict": {"fh": "original_y"}})
    ]
)

# Classification example
pipe = NetworkPipeline(
        steps=[
            ("tabularizer", Tabularizer(), {"X": "original_X"}),
            ( "classifier", DummyClassifier(strategy="prior"), {
                         fit": {"X": "tabularizer", "y": "original_y"},
                         "predict": {"X": "tabularizer"} })
        ]
    )
```

### Questions

- Is proposed interface intuitive?
- Should the steps be executed sequentially or in any order defined by the user? This would be achieved if the `arguments` dictionary can refer to future steps. This feature is not currently supported in my design.

### TODO
- [ ] Improve examples by adding more than two steps.
- [ ] Write checks and tests.


mloning:
I suggest to implement this for a single task (i.e. forecasting or classification) first, since neither API design nor implementation is obvious here.

ViktorKaz:
@mloning what are your main concerns about the API?

My thinking is that the main feature of the pipeline is that the output of a step is not necessarily the input of the next step. This I believe is relevant for all tasks. If on top of this you add the ability to:

1. Have different behaviour for the `fit` and `predict` methods of the pipeline (have different arguments at `fit` and `predict`, be able to skip `predict`).
1. Make  `X` and `y` (but not both) optional.

We would cover the tasks of:
1. classification
2. forecasting
3. online directional forecasts

I can focus on one of the tasks above in the first place. However, if we want to avoid numerous revisions, we should agree on what the main requirements for the API.

mloning:
@ViktorKaz I worry that if we start off too general it will be hard to integrate, review and maintain. I actually prefer numerous smaller changes. For example, if we make it a forecaster, there are a number of tests that are run automatically on the class making sure that it complies with the basic forecasting API. 

Also note that there is no API for directional forecasting yet. So we would have to develop that in a separate PR.

ViktorKaz:
@mloning ok, I will start with the forecasting case to test the concept. I will have the pipeline inherit from `sktime.forecasting.base. BaseForecaster`, this is what you mean, right?

mloning:
@ViktorKaz yes that's right

sktime/forecasting/compose/_networkpipeline.py
```diff
@@ -0,0 +1,267 @@
...
+                continue
+            # go through all steps and look for returned values
+            returned_step_value = self._check_steps_for_values(argument_value)
+            if returned_step_value is not None:
+                returned_arguments_kwarg[argument_name] = returned_step_value
+
+        return returned_arguments_kwarg
+
+    def fit(self, y, X=None, fh=None):
+        self._X = X
```
mloning, line 145: use `self._set_y_X()` for this

sktime/forecasting/compose/_networkpipeline.py
```diff
@@ -0,0 +1,267 @@
...
+            # estimators have fit and predict methods
+            if hasattr(alg, "fit") and hasattr(alg, "predict"):
+                alg.fit(**processed_arguments)
+                self._is_fitted = True
+
+        return self
+
+    def predict(self, fh=None, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
+        if fh is not None:
+            self._set_fh(fh)
```
mloning, line 182: just use `self._set_fh(fh)` without the if-statement, None values are handled internally

sktime/forecasting/compose/_networkpipeline.py
```diff
@@ -0,0 +1,267 @@
...
+
+__author__ = ["Viktor Kazakov"]
+__all__ = ["NetworkPipelineForecaster"]
+
+
+class NetworkPipelineForecaster(
+    _OptionalForecastingHorizonMixin,
+    _SktimeForecaster,
+    _HeterogenousMetaEstimator,
+    _SeriesToSeriesTransformer,
```
mloning, line 18: if you inherit from `_SeriesToSeriesTransformer` you have to implement `transform`,  
take a look at `TransformedTargetForecaster` for an example, but this may cause some issues still, so perhaps easiest to start without it being a transformer

mloning (commented):
Hi @ViktorKaz - just a few quick comments to make some progress on the failing tests.

ViktorKaz:
@mloning thanks for the comments. I still get the error that the cutoff has not `.freq` property at which I have been stuck for the last couple of days. Will investigate further.

sktime/forecasting/compose/_networkpipeline.py
```diff
@@ -0,0 +1,265 @@
...
+                del processed_arguments["X"]
+            # Transformers are instances of BaseTransformer and BaseEstimator
+            # Estimators are only instances of BaseEstimator
+            if hasattr(alg, "fit_transform"):
+                out = alg.fit_transform(**processed_arguments)
+                self._step_results[name] = out
+                self._is_fitted = True
+            # estimators have fit and predict methods
+            if hasattr(alg, "fit") and hasattr(alg, "predict"):
+                alg.fit(**processed_arguments)
```
mloning, line 171: * You should pass the forecasting horizon as a kwarg to `alg.fit` here
* For variable naming, instead of `alg` we usually use `estimator`

mloning (commented):
A quick comments, perhaps it helps!

sktime/forecasting/compose/_networkpipeline.py
```diff
@@ -0,0 +1,269 @@
...
+            # Transformers are instances of BaseTransformer and BaseEstimator
+            # Estimators are only instances of BaseEstimator
+            if hasattr(alg, "fit_transform"):
+                out = alg.fit_transform(**processed_arguments)
+                self._step_results[name] = out
+                self._is_fitted = True
+            processed_arguments["fh"] = self._fh
+            # estimators have fit and predict methods
+            if hasattr(alg, "fit") and hasattr(alg, "predict"):
+                alg.fit(**processed_arguments)
```
mloning, line 171: Are you cloning the algorithm somewhere? Otherwise you will update the object in the list which may explain the `check_fit_does_not_overwrite_hyper_params` error.

sktime/forecasting/compose/_networkpipeline.py
```diff
@@ -0,0 +1,269 @@
...
+                self._is_fitted = True
+
+        return self
+
+    def _predict(self, fh=None, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
+        self._set_fh(fh)
```
mloning, line 177: Having this line here isn't necessary because it's already being taken care of in the `predict` method which then calls the internal `_predict` method: https://github.com/alan-turing-institute/sktime/blob/f71f373aa8f89497d654661976d745c747c1cb0e/sktime/forecasting/base/_sktime.py#L206

Perhaps that fixes the other error?

sktime/forecasting/compose/_networkpipeline.py
ViktorKaz: thanks, this fixed the error.

sktime/forecasting/compose/_networkpipeline.py
ViktorKaz: I removed the line `self._set_fh(fh)` from the `_predict` method but the test about the cutoff not having .freq attribute still fails.

mloning: I'm not sure why this information is lost? I suggest to step through it with the debugger. I know that the functionality around the forecasting horizon is a bit brittle given the required date/time conversions but haven't seen this error before.

mloning: It also only happens in specific cases in `update_predict`. The test is defined here: https://github.com/alan-turing-institute/sktime/blob/8d0b94b4992083f3c1e388afaeaf9f644759e68f/sktime/forecasting/tests/test_all_forecasters.py#L289

mloning: Please also update with latest changes on `main`.

ViktorKaz: @mloning I think I found the problem but I am not sure that I have the right solution.

The dataset that is being used for the test has daily observations starting on 2000-01-01 and finishing on 2000-02-06. When the second prediction is made, a prediction for 2000-02-11 is produced, i.e. skipping the dates 2000-02-07 to 2000-02-10.

When calling the line `self._update_y_X(y, X)` at the next `update()`, `self._y` is set with an `index.freq=None` I presume because of the gap in the index.

My temporary fix was to do:

```Python
y_index_frequency = self._y.index.freq
self._update_y_X(y, X)
self._y = self._y.asfreq(y_index_frequency)
```

This sets the index frequency to the correct one and the test passes but `self._y` has some missing values.

Do you think this behaviour is correct, i.e. `predict()` skipping certain days? For the purposes of this test I am just using a NaiveForecaster.

mloning: I see, what I don't understand yet is, why this fails in your implementation but not the other algorithms?

ViktorKaz: @mloning to be honest I am not 100% sure either, my implementation is fairly simple at the moment, I think.

Maybe this is linked to the chain of inheritance and method calls which I could not fully follow. While I was debugging the code I remember seeing that `predict()` was called more than once for the same date, this might be somehow linked to the issue but this would need to be investigated.

From what I see the code now passes the tests and can be merged. I will experiment with the pipeline some more with some real datasets to see whether it produces meaningful forecasts. Feel free to review and let me know if you have comments at this stage.

sktime/forecasting/compose/_networkpipeline.py
```diff
@@ -0,0 +1,278 @@
...
+                self._fitted_estimators[name] = f
+
+        self._is_fitted = True
+
+        return self
+
+    def _predict(self, fh=None, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
+        if X is not None:
+            self._X = X
+        for name, est, arguments in self._iter():
```
mloning, line 178: Do we iterate in reverse order here?

sktime/forecasting/compose/_networkpipeline.py
```diff
@@ -0,0 +1,278 @@
...
+    ...    ])
+    """
+
+    _required_parameters = ["steps"]
+    _tags = {"univariate-only": True}
+
+    def __init__(self, steps):
+        self.steps = steps
+        self._step_results = {}
+        self._fitted_estimators = {}
```
mloning, line 76:
```suggestion
        self.steps_
```
(in line with scikit-learn's trailing underscore convention for fitted variables)

sktime/forecasting/compose/_networkpipeline.py
```diff
@@ -0,0 +1,278 @@
...
+            if returned_step_value is not None:
+                returned_arguments_kwarg[argument_name] = returned_step_value
+
+        return returned_arguments_kwarg
+
+    def fit(self, y, X=None, fh=None):
+        self._set_y_X(y, X)
+        self._set_fh(fh)
+
+        for name, est, arguments in self._iter():
```
mloning, line 136: I find it hard to follow what's happening in this loop. Could we add some comments?

sktime/forecasting/compose/_networkpipeline.py
```diff
@@ -0,0 +1,278 @@
...
+            processed_arguments = {}
+
+            if "fit" in arguments and arguments["fit"] is None:
```
mloning, line 139: Does that mean we may not fit some estimators?

sktime/forecasting/compose/_networkpipeline.py
```diff
@@ -0,0 +1,278 @@
...
+
+            else:
+                processed_arguments = self._process_arguments(arguments)
+            # Transformers are instances of BaseTransformer and BaseEstimator
+            # Estimators are only instances of BaseEstimator
+
+            if hasattr(est, "update"):
+                out = est.update(**processed_arguments)
+                self._step_results[name] = out
+                self._is_fitted = True
```
mloning, line 240: We shouldn't update the `_is_fitted` state anywhere but in `fit(...)`.

sktime/tests/_config.py
```diff
@@ -275,8 +276,25 @@
...
+                Imputer(method="drift"),
+                {"fit": {"Z": "original_y"}, "predict": None, "update": None},
+            ),
+            (
+                "forecaster",
+                NaiveForecaster(strategy="mean"),
+                {
+                    "fit": {"y": "imputer", "X": "original_X", "fh": "original_fh"},
+                    "predict": {"fh": "original_fh"},
+                    "update": {"y": "original_y"},
```
mloning, line 292: Why would we want to update the estimator without prior transformation when we used transformations in fitting?

mloning (commented):
I had a first look at the code. What's the simplest use case that is enabled by this PR which is not possible with the `TransformedTargetForecaster` or a simple `PipelineForecaster`? 

I left a few questions and suggestions, mostly to aid my understanding at this stage.

sktime/forecasting/compose/_networkpipeline.py
ViktorKaz: ok, will change it. I saw the `self.steps` convention being used in `TransformedTargetForecaster`.

sktime/forecasting/compose/_networkpipeline.py
ViktorKaz: No, `self._iter()` goes through the steps in the order in which they were created.

sktime/forecasting/compose/_networkpipeline.py
ViktorKaz: ok

sktime/forecasting/compose/_networkpipeline.py
ViktorKaz: The design of the pipeline allows to skip `fit`, `predict` and `update` for each step. I had in mind the use case where you might not want to update the transformers but only the forecaster (last step).

In my prototype for this pipeline I had a step which only had `transform()` without a `fit()`. It was a custom class for compiling a new dataset from extracted features and processed labels in previous steps in the pipeline. The design has changed a little bit since then so we should probably discuss whether it makes sense to have steps that do not implement `fit()`. The line that you highlighted is a legacy of this first prototype.

sktime/tests/_config.py
ViktorKaz: this is an oversight, I'll fix it.

sktime/forecasting/compose/_networkpipeline.py
ViktorKaz: ok

sktime/forecasting/compose/_networkpipeline.py
ViktorKaz: Yes, you are right, `predict()` will need to be in reverse order. I missed this.

ViktorKaz: Actually, thinking a little bit more about this, does predict need to always go through the steps in reverse order?

I am planning to ultimately use this pipeline for online one step ahead supervised classification. For this use case at the model fitting stage `y` is transformed, features are extracted, etc. The same transformations and feature extractions need to happen at the prediction stage as well in the same order as for `fit`.

For forecasting, let's assume that you fit your estimator after you transformed `y`. In this case, at the prediction stage would you not need to:

1. transform `y`, 
2. predict, 
3. inverse transform y.

I checked the code of  `TransformedTargetForecaster`, indeed the steps are executed in reverse order there. It will be quite easy to implement the same logic for the Network Pipeline but I just want to understand what we are trying to achieve. Is the pipeline supposed to be executing `predict()` if the estimator is a forecaster and `inverse_transform()` if it is a transformer? How do we handle the case if the forecaster was fitted on transformed `y`?

mloning: I'm not sure about online one step ahead supervised classification (is that directional forecasting?). For forecasting, if the forecaster is fitted on transformed data, we first generated the predictions and then inverse transform the predictions. 

I think it's really important to write down what the simplest use case is that is enabled by this PR which is not possible with the `TransformedTargetForecaster` or a simple `PipelineForecaster`.

ViktorKaz: yes, this is directional forecasting. 

The pipeline allows to execute any number of steps in almost any given order, as opposed to sequentially. This design was initially inspired by the directional forecasting use case in which you might need to extract different features from the original raw time series and labelling the observations based on some rule. For directional forecasts, I don't think this can easily be achieved with a simple pipeline. 

Regarding forecasting, maybe a simple use case of this pipeline would be to build some sort of an ensemble forecasting algorithm. For example, fitting two forecasters on the original `y` and having some sort of weighting or voting rule to produce an ensemble forecast. I can think a little bit more about it and propose some classification and forecasting examples for the documentation.

ViktorKaz:
@mloning I made some tweaks as we discussed, let me know what you think about it. 

One thing to note is that there is still an issue with the cutoffs at `update` that I could not follow track fully. I am setting the cutoff for estimators manually:

```Python
 out._cutoff = self.cutoff
```

I am also setting the index frequency manually otherwise `self.cutoff` has no `.freq` attribute
```Python
y_index_frequency = self._y.index.freq
self._update_y_X(y, X)
self._y = self._y.asfreq(y_index_frequency)
```

This solution works, at least for passing the tests. 

@fkiraly @mloning I wanted to update the forecasting notebook with an example of how this pipeline can be used for ensemble forecasts. As discussed in #948 this requires the option to not invert the steps at `predict`. 

I was thinking of changing the signature of `_predict` to add one additional parameter, in this example `inverse_steps`:

```Python
    def _predict(self, fh=None, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA, inverse_steps=False):
        if X is not None:
            self._X = X
        for name, est, arguments in self._iter(reverse=inverse_steps): #this is the suggested change
           ....
```

Do you agree with this?

fkiraly:
>I was thinking of changing the signature of _predict to add one additional parameter, in this example inverse_steps

I don't think that belongs in `_predict` - the signature of `predict`/`_predict` should be universal.
What are you planning to solve by this?

mloning:
@ViktorKaz could we try to first discuss the scope before going into more details of the code (see https://github.com/alan-turing-institute/sktime/discussions/948)? To me this isn't very clear, but perhaps I have missed something.

ViktorKaz:
@mloning @fkiraly ok, do you want to have a call or do you want to discuss this in another format?

fkiraly:
let´s have a call.

mloning:
Hi @ViktorKaz, not sure if you saw the discussion already, but I think we identified a use case which cannot be easily covered with the existing functionality in https://github.com/alan-turing-institute/sktime/pull/967#pullrequestreview-687842603, basically to specify a forecasting pipeline that does transformations on y and X, I think that would be very useful and the next logical step in developing a more expressive pipeline syntax.

ViktorKaz:
@mloning I am on holiday at the moment and have not been following the discussion closely. However, the network pipeline can definitely handle transformations on both `X` and `y`. By the way, the network pipeline code is ready for review, I just need to update the docstring with more relevant examples. If you or @fkiraly want to review please feel free to do so.

mloning:
@fkiraly could you take a look? I'm still not really sure about the scope and use case, or did you describe them somewhere @ViktorKaz and I missed it?

ViktorKaz:
@mloning I started writing an enhancement proposal. You can find it here: https://github.com/ViktorKaz/enhancement-proposals/blob/network_pipeline/steps/10_network_pipeline/step.md

The current version of the pipeline also supports specifying the steps without explicitly specifying the `fit`, `predict` and `update` arguments. I don't have an example of this in the enhancement proposal but will update it.

sktime/forecasting/compose/_networkpipeline.py
```diff
@@ -0,0 +1,398 @@
...
+        deep : boolean, optional
+            If True, will return the parameters for this estimator and
+            contained subobjects that are estimators.
+        Returns
+        -------
+        params : mapping of string to any
+            Parameter names mapped to their values.
+        """
+
+        return {"steps": self.steps}
```
fkiraly, line 382: I think this should go through the steps and return their parameters, similar to how a `sklearn` pipeline does it - the steps are not parameters.

fkiraly:
@ViktorKaz, I have looked at this and I would say this is almost good to go - I would suggest, let's include this as an "experimental" pipeline and see what users say. We can later add functionality like select-by-variable-name or feature union type operations.

However, I'd like to ask you to do two things first:

* during the summer dev days, we changed the forecaster interface - all base classes for forecasters have been merged in `BaseForecaster`, and the base class should also take care of all input checks etc. See #955 for a recipe to how to change from "old" to "new" interface.
* I think `get_params` is not quite right. This should never return components, only a dict of hyper-parameters. for example, if your steps are called `("a" : myest1, "b" : myest2}`, then the expected dict to return would have keys `"a__param1"`, `"a__param2"`, `"b__param2"` and so on, with double underscores. See how `sklearn` pipeline does it, including how it does the deep/recursive parameter access, you should do exactly the same.

fkiraly (changes requested):
see above

ViktorKaz:
@fkiraly thanks for the comments, I'll implement the changes you request.

ViktorKaz:
I fixed the doc quality checks.

fkiraly (commented):
Great work, @ViktorKaz, really nice!

Some comments from our Friday meeting:
* I think we don't need the `Converter` in the example, if the `Concatenator` also takes series
* I would rename `Concatenator` to `SeriesUnion` or `SeriesFeatureUnion`
* the example worksheet needs more explanation, e.g., a plain English description of the pipeline we are implementing, perhaps even a small picture

fkiraly (changes requested):
see above.

aiwalter:
@ViktorKaz 
didnt have time to look much into this PR yet, but short side note: maybe it makes sense to rename `NetworkPipelineForecaster` to `NetworkPipeline`? As we already have a `ForecastingPipeline`, so would fit better together with the name then

sktime/forecasting/compose/_networkpipeline.py
```diff
@@ -0,0 +1,423 @@
...
+    Forecasting example. Note convention for specifying
+    different arguments depending on whether
+    NetworkPipelineForecaster.fit or
+    NetworkPipelineForecaster.predict is called.
+    >>> from sktime.transformations.panel.dataset_manipulation import Selector
+    >>> from sktime.transformations.panel.dataset_manipulation import Converter
+    >>> from sktime.transformations.panel.dataset_manipulation import Concatenator
+    >>> from sktime.transformations.series.boxcox import BoxCoxTransformer
+    >>> from sktime.transformations.series.adapt import TabularToSeriesAdaptor
+    >>> from sklearn.preprocessing import MinMaxScaler
```
aiwalter, line 51:
```suggestion
```

sktime/forecasting/compose/_networkpipeline.py
```diff
@@ -0,0 +1,423 @@
...
+    >>> from sktime.forecasting.arima import AutoARIMA
+    >>> from sktime.forecasting.model_selection import temporal_train_test_split
+    >>> from sktime.datasets import load_longley
```
aiwalter, line 54:
```suggestion
    >>> from sktime.datasets import load_longley
    >>> from sklearn.preprocessing import MinMaxScaler
```

sktime/forecasting/compose/_networkpipeline.py
```diff
@@ -0,0 +1,423 @@
...
+    Parameters
+    ----------
+    steps : array of lists
+        list comprised of three elements:
+            1. name of step (string),
+            2. estimator (object),
+            3. arguments (dictionary) key value paris for
+            fit_transform or predict method of estimator
+
+    Examples
```
aiwalter, line 33:
```suggestion
            fit_transform or predict method of estimator

    See Also
    --------
    ForecastingPipeline, TransformedTargetForecaster

    Examples
```

sktime/forecasting/compose/_networkpipeline.py
ViktorKaz: I am using MinMaxScaler in the example below:

```
    ... ("ft2", TabularToSeriesAdaptor(MinMaxScaler()), { "Z": "feature_X2"}),
```
and also

```
    ... ("new_y", TabularToSeriesAdaptor(MinMaxScaler()), {"Z":"original_y"}),
```

ViktorKaz:
> @ViktorKaz
> didnt have time to look much into this PR yet, but short side note: maybe it makes sense to rename `NetworkPipelineForecaster` to `NetworkPipeline`? As we already have a `ForecastingPipeline`, so would fit better together with the name then. Seems you even used this name previously? :)

@aiwalter I don't have strong opinion about the name of the pipeline but I will start working on a network pipeline classifier next. The intention was to refactor most of this code in a NetworkPipeline base class and have two child classes: NetworkPipelineForecaster and NetworkPipelneClassifier

ViktorKaz:
> * I think we don't need the `Converter` in the example, if the `Concatenator` also takes series
> * I would rename `Concatenator` to `SeriesUnion` or `SeriesFeatureUnion`
> * the example worksheet needs more explanation, e.g., a plain English description of the pipeline we are implementing, perhaps even a small picture

@fkiraly I introduced these changes.

sktime/forecasting/compose/_networkpipeline.py
```diff
@@ -0,0 +1,425 @@
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
ViktorKaz: ok

sktime/forecasting/compose/_networkpipeline.py
```diff
@@ -0,0 +1,425 @@
...
+# from sktime.forecasting.base._sktime import _SktimeForecaster
+# from sktime.forecasting.base._sktime import _OptionalForecastingHorizonMixin
+# from sktime.base import _HeterogenousMetaEstimator
+from sktime.forecasting.base._base import DEFAULT_ALPHA
+from sklearn.base import clone
+from sktime.forecasting.base import BaseForecaster
+
+__author__ = ["Viktor Kazakov"]
```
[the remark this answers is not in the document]
ViktorKaz: ok

sktime/forecasting/compose/_networkpipeline.py
```diff
@@ -0,0 +1,425 @@
...
+    Each step in the pipeline can use the output of any of the previous steps or
+    the original input values to the fit and predict methods of the pipeline.
+
+    The arguments of the steps in the pipeline must be provided as a dictionary.
+    The dictionary is substituted in the fit_transform, fit or predict function calls
+    of the step transformer or estimator.
+
+    Parameters
+    ----------
+    steps : array of lists
```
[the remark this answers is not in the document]
ViktorKaz: yes, will change it

sktime/forecasting/compose/_networkpipeline.py
```diff
@@ -0,0 +1,425 @@
...
+        list comprised of three elements:
+            1. name of step (string),
+            2. estimator (object),
+            3. arguments (dictionary) key value paris for
```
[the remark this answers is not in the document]
ViktorKaz: typo

sktime/forecasting/compose/_networkpipeline.py
```diff
@@ -0,0 +1,425 @@
...
+            fit_transform or predict method of estimator
```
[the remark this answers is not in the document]
ViktorKaz: will expand it

sktime/forecasting/compose/_networkpipeline.py
```diff
@@ -0,0 +1,425 @@
...
+    The arguments of the steps in the pipeline must be provided as a dictionary.
+    The dictionary is substituted in the fit_transform, fit or predict function calls
+    of the step transformer or estimator.
+
+    Parameters
+    ----------
+    steps : array of lists
+        list comprised of three elements:
+            1. name of step (string),
+            2. estimator (object),
```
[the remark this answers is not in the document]
ViktorKaz: ok, will specify this

sktime/forecasting/compose/_networkpipeline.py
```diff
@@ -0,0 +1,425 @@
...
+            3. arguments (dictionary) key value paris for
+            fit_transform or predict method of estimator
+
```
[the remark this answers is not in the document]
ViktorKaz: ok, will write more details

sktime/forecasting/compose/_networkpipeline.py
```diff
@@ -0,0 +1,425 @@
...
+    ... ("new_y", TabularToSeriesAdaptor(MinMaxScaler()), {"Z":"original_y"}),
+    ... ("y_out", AutoARIMA(suppress_warnings=True), {
+    ...    "fh":"original_fh", "y": "new_y", "X": "concat"})
+    ... ])
+    >>> pipeline = pipe.fit(y_train,X_train)
+    >>> predictions = pipe.predict(fh=[1,2,3,4], X=X_test)
+    """
+
+    _required_parameters = ["steps"]
+    _tags = {"univariate-only": False, "requires-fh-in-fit": False}
```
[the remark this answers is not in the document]
ViktorKaz: ok, will think about it. Maybe we can discuss the purpose of the tags

sktime/forecasting/compose/_networkpipeline.py
```diff
@@ -0,0 +1,425 @@
...
+            returned_step_value = self._check_steps_for_values(argument_value, method)
+            if returned_step_value is not None:
+                returned_arguments_kwarg[argument_name] = returned_step_value
+
+        return returned_arguments_kwarg
+
+    def _fit(self, y, X=None, fh=None):
+        self.steps_ = self._check_steps_for_consistency()
+        self._set_y_X(y, X)
+        self._set_fh(fh)
```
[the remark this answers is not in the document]
ViktorKaz: will investigate, if nothing breaks will remove it.

sktime/forecasting/compose/_networkpipeline.py
```diff
@@ -0,0 +1,425 @@
...
+                f.fit(**processed_arguments)
+                self._fitted_estimators[name] = f
+
+        self._is_fitted = True
+
+        return self
+
+    def _predict(self, fh=None, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
+        if X is not None:
+            self._X = X
```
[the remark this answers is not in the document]
ViktorKaz: the idea is that the "X seen so far" is overwritten as it is used by the pipeline depending on the context. We can have `self._X_fit`, `self._X_update` and `self._X_predict` if this is required. Let's discuss.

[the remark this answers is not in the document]
fkiraly: sure - I think this needs a verbal discussion. I don't think `predict` should do update `X`, only `fit` and `update` should.

sktime/forecasting/compose/_networkpipeline.py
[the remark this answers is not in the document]
fkiraly: key purpose is steering the behaviour of input/output conversions and checks, testing (e.g., a univariate model won't raise an error if it crashes with multivariate data), and findability/search for users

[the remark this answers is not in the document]
ViktorKaz: Ok, I found a solution using static and dynamic tags. The dynamic tags are set by looping through all steps, inspecting the tags of the estimators and updating the dictionary.

sktime/forecasting/compose/_networkpipeline.py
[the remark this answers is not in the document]
ViktorKaz: I removed it

sktime/forecasting/compose/_networkpipeline.py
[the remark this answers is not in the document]
ViktorKaz: changed code as per discussion. `self._X` and `self._y` are not used any longer.

ViktorKaz:
@fkiraly can you please have another look and merge the PR unless you have more comments.
Thanks

fkiraly (changes requested):
Almost ready to go in, but I think there are two problems that need to be resolved first:

* in the `__init__`, you update the `_tags` class variable - this changes the `_tags` for all objects of that class! Class variables should not be written to. Further, using the latest tags is not a good idea. This needs to be removed.
* I recommend, just remove any manipulation of `_tags` and instead give it the `"requires-fh-in-fit": True` tag, and the `scitype:y = pd.DataFrame` tag, and assume that `y` internally is always a `pd.DataFrame`.
* The notebook has much improved, it's really nice. Though, I'd suggest: please add some explanations to the simple example, too - you can do this along the lines of the advanced one.
* Please remove outputs when you commit the notebook.

ViktorKaz:
@fkiraly if the tag`"scitype:y": "pd.DataFrame"` returns the following error for all tests : 
```
E KeyError: 'pd.DataFrame'
```

can you please send me the link to the list of all supported tags  as I could not find them.

fkiraly:
>can you please send me the link to the list of all supported tags as I could not find it.

it's in `registry/_tags`, and I think it's a supported tah value.

Also, I can't see any tests failing? Everything seems to pass?

ViktorKaz:
@fkiraly The tests pass because I did not include the `"scitype:y":"pd.DataFrame"` tag. This is what I see under `registry/_tags` for the key `scitype:y`

```
(
        "scitype:y",
        "forecaster",
        ("str", ["univariate", "multivariate", "both"]),
        "which series type does the forecaster support? multivariate means >1 vars",
    ),
```

fkiraly:
oh, sorry - I meant `y_inner_type`. `scitype:y` should be `both`.

ViktorKaz:
@fkiraly the tag `"y_inner_mtype":"pd.DataFrame"` results in several failed unit tests and I am not sure how to solve this.

As far as I can see, the problem originates in the fact that at `update`, `self._update_y_X(y, X)` tries to concatenate the new `X` and `y` with `self._X` and `self._y`. However, this fails with the error:

```
ValueError: Must specify axis=0 or 1
```

However, `_update_y_X ` does not have axis parameter. Therefore, the method will probably need to be changed.

I can't find any other estimators in the code base that use the tag ``"y_inner_mtype":"pd.DataFrame"`` to see how this is solved.

We can discuss this at the meeting tonight.

fkiraly:
>However, this fails with the error:

Can you show the entire error? Not sure why it would fail, since `_update_y_X` just calls `combine_first` which does not need `axis`.

ViktorKaz:
@fkiraly please see below
```
============================= test session starts ==============================
platform linux -- Python 3.9.5, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
rootdir: /home/viktor/sktime, configfile: setup.cfg
plugins: anyio-3.2.1
collected 1 item

sktime/forecasting/tests/test_all_forecasters.py F                       [100%]

=================================== FAILURES ===================================
__ test_update_predict_predicted_index[False-1-1-1-NetworkPipelineForecaster] __

Forecaster = <class 'sktime.forecasting.compose._networkpipeline.NetworkPipelineForecaster'>
fh = 1, window_length = 1, step_length = 1, update_params = False

    @pytest.mark.parametrize("Forecaster", FORECASTERS)
    @pytest.mark.parametrize("fh", TEST_OOS_FHS)
    @pytest.mark.parametrize("window_length", TEST_WINDOW_LENGTHS)
    @pytest.mark.parametrize("step_length", TEST_STEP_LENGTHS)
    @pytest.mark.parametrize("update_params", [False])
    def test_update_predict_predicted_index(
        Forecaster, fh, window_length, step_length, update_params
    ):
        """Check predicted index in update_predict with update_params=False."""
>       _check_update_predict_predicted_index(
            Forecaster, fh, window_length, step_length, update_params
        )

sktime/forecasting/tests/test_all_forecasters.py:323:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
sktime/forecasting/tests/test_all_forecasters.py:306: in _check_update_predict_predicted_index
    y_pred = f.update_predict(y_test, cv=cv, update_params=update_params)
sktime/forecasting/base/_base.py:487: in update_predict
    return self._predict_moving_cutoff(
sktime/forecasting/base/_base.py:998: in _predict_moving_cutoff
    y_pred = self._update_predict_single(
sktime/forecasting/base/_base.py:926: in _update_predict_single
    self.update(y, X, update_params=update_params)
sktime/forecasting/base/_base.py:420: in update
    self._update(y=y_inner, X=X_inner, update_params=update_params)
sktime/forecasting/compose/_networkpipeline.py:356: in _update
    self._update_y_X(y, X)
sktime/forecasting/base/_base.py:614: in _update_y_X
    self._y = y.combine_first(self._y)
venv_sktime/lib/python3.9/site-packages/pandas/core/frame.py:6246: in combine_first
    return self.combine(other, combiner, overwrite=False)
venv_sktime/lib/python3.9/site-packages/pandas/core/frame.py:6111: in combine
    this, other = self.align(other, copy=False)
venv_sktime/lib/python3.9/site-packages/pandas/core/frame.py:3962: in align
    return super().align(
venv_sktime/lib/python3.9/site-packages/pandas/core/generic.py:8565: in align
    return self._align_series(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self =                    0
2000-02-07  3.735624
other = 2000-01-01    4.452024
2000-01-02    4.499818
2000-01-03    2.536776
2000-01-04    4.816629
2000-01-05    2.641968
200...765
2000-02-04    4.591978
2000-02-05    1.806085
2000-02-06    3.609645
2000-02-07    3.735624
Freq: D, dtype: float64
join = 'outer', axis = None, level = None, copy = False, fill_value = None
method = None, limit = None, fill_axis = 0

    def _align_series(
        self,
        other,
        join="outer",
        axis=None,
        level=None,
        copy: bool_t = True,
        fill_value=None,
        method=None,
        limit=None,
        fill_axis=0,
    ):
    
        is_series = isinstance(self, ABCSeries)
    
        # series/series compat, other must always be a Series
        if is_series:
            if axis:
                raise ValueError("cannot align series to a series other than axis 0")
    
            # equal
            if self.index.equals(other.index):
                join_index, lidx, ridx = None, None, None
            else:
                join_index, lidx, ridx = self.index.join(
                    other.index, how=join, level=level, return_indexers=True
                )
    
            left = self._reindex_indexer(join_index, lidx, copy)
            right = other._reindex_indexer(join_index, ridx, copy)
        else:
            # one has > 1 ndim
            fdata = self._mgr
            if axis == 0:
                join_index = self.index
                lidx, ridx = None, None
                if not self.index.equals(other.index):
                    join_index, lidx, ridx = self.index.join(
                        other.index, how=join, level=level, return_indexers=True
                    )
    
                if lidx is not None:
                    fdata = fdata.reindex_indexer(join_index, lidx, axis=1)
    
            elif axis == 1:
                join_index = self.columns
                lidx, ridx = None, None
                if not self.columns.equals(other.index):
                    join_index, lidx, ridx = self.columns.join(
                        other.index, how=join, level=level, return_indexers=True
                    )
    
                if lidx is not None:
                    fdata = fdata.reindex_indexer(join_index, lidx, axis=0)
            else:
>               raise ValueError("Must specify axis=0 or 1")
E               ValueError: Must specify axis=0 or 1

venv_sktime/lib/python3.9/site-packages/pandas/core/generic.py:8700: ValueError
----------------------------- Captured stdout call -----------------------------
Backend TkAgg is interactive backend. Turning interactive mode on.
-------------- generated xml file: /tmp/tmp-267991G3ohJiruIOV.xml --------------
=========================== short test summary info ============================
FAILED sktime/forecasting/tests/test_all_forecasters.py::test_update_predict_predicted_index[False-1-1-1-NetworkPipelineForecaster]
============================== 1 failed in 10.97s ==============================

```

fkiraly:
Hm, it looks like the problem is trying to concatenate a `pd.DataFrame` with a `pd.Series`. The reason seems to be the lack of conversion/checking functionality inside `update_predict`. Let me have a look at this.

fkiraly:
#1436 is now merged and should unblock work on this.

ViktorKaz:
@fkiraly will look into it.

fkiraly (approved):
Comments from earlier change request seems addressed.
Further work is perhaps best done in incremental PR since thie is already large/complex.

fkiraly (changes requested):
just updated from main and it is failing tests again.
Not sure what is going on, could you kindly investigate?

It's related to the time indices returned in `predict`.

ViktorKaz:
@fkiraly I'll investigate and fix the failing tests.

PR #858, opened 2021-05-04
Update metric to handle y_train

RNKuhns:
#### Reference Issues/PRs

This addresses functionality in #712.

Add functionality to accept `y_train` in `__call__` method and pass to underlying function if it requires `y_train`. 

Also updated underlying metric classes to inherit from BaseMetric. 

##### For all contributions

##### For new estimators


RNKuhns:
@mloning and @fkiraly the more I'm looking at this while looking to simplify the unit tests for these metrics, it makes more and more sense to me that the metric functions in `_functions.py` should also get updated to accept `y_train` via 1**kwargs` so we have a uniform interface. 

Difference to the user passing info to functions will be relatively minimal. 

Benefit is that `y_train` would be passed same way to function as the `__call__` of the class version of the metric: 

```python
mean_absolute_scaled_error(y_true, y_pred, y_train=y_train, ... # Any other keyword args like multioutput)
mase = MeanAbsoluteScaledError()
mase(y_true, y_pred, y_train=y_train)
```
versus the current function setup accepting y_train as arg and class version of metric accepting it as keyword:

```python
mean_absolute_scaled_error(y_true, y_pred, y_train, ... # Any other keyword args like multioutput)
mase = MeanAbsoluteScaledError()
mase(y_true, y_pred, y_train=y_train)
```

Note that I'll do whatever we decide for `y_train` for metrics that accept `y_pred_benchmark`. 

I'll start working on this, but check back here before pushing to github. Let me know if you don't think we should implement this change and I'll roll it back.

sktime/performance_metrics/forecasting/_classes.py
```diff
@@ -45,13 +45,24 @@
...
-        self._func = func
-        self.name = name if name is not None else func.__name__
-        self.greater_is_better = greater_is_better
+class _BaseForecastingErrorMetric(BaseMetric):
+    """Base class for defining forecasting error metrics in sktime. Extends
+    sktime's BaseMetric to the forecasting interface.
+    """
+
+    _tags = {
+        "requires_y_train": False,
```
mloning, line 54: Let's only use hyphens for separating words in tags for consistency

sktime/performance_metrics/forecasting/_classes.py
```diff
@@ -309,11 +397,13 @@ def make_forecasting_scorer(func, name=None, greater_is_better=False):
     scorer:
         Metric class that can be used as forecasting scorer.
     """
-
-    return _MetricFunctionWrapper(func, name=name, greater_is_better=greater_is_better)
+    if greater_is_better:
+        return _BaseForecastingErrorMetric(func, name=name, multioutput=multioutput)
```
mloning, line 401: Just read in scikit-learn that 
> All scorer objects follow the convention that higher return values are better than lower return values.

https://scikit-learn.org/stable/modules/model_evaluation.html#common-cases-predefined-values

Not sure if we also want to adopt that convention or give users the flexibility. I'd say we leave it as it is for now. Having a single convention would be simpler for higher-level functionality though.

sktime/performance_metrics/forecasting/_functions.py
```diff
@@ -250,6 +253,9 @@ def mean_absolute_scaled_error(
             "The M4 Competition: 100,000 time series and 61 forecasting methods",
             International Journal of Forecasting, Volume 3
     """
+    y_train = kwargs.pop("y_train", None)
+    if y_train is None:
+        raise ValueError("y_train must be supplied to compute this metric.")
```
mloning, line 258: Perhaps something like this?
```suggestion
        raise ValueError(f"{self.name} requires `y_train`. Please pass `y_train` as a keyword argument when calling the metric")
```

sktime/performance_metrics/forecasting/_functions.py
```diff
@@ -524,6 +535,9 @@ def mean_squared_scaled_error(
             "Another look at measures of forecast accuracy", International
             Journal of Forecasting, Volume 22, Issue 4.
     """
+    y_train = kwargs.pop("y_train", None)
+    if y_train is None:
+        raise ValueError("y_train must be supplied to compute this metric.")
```
mloning, line 540: Since we're using this in multiple places, would it be worth it to write a helper function for this, e.g. `_get_y_train(**kwargs)` or even `_get_kwarg(**kwargs, "y_train")` so that we can use it for `y_pred_benchmark` too?

sktime/performance_metrics/tests/test_performance_metrics_forecasting.py
```diff
@@ -368,6 +367,25 @@
 }
 
 
+def _call_metrics(loss_func, loss_class, y_true, y_pred, y_train, y_pred_benchmark):
```
mloning, line 370:
```suggestion
def _call_metric(metric_func, metric_class, y_true, y_pred, y_train, y_pred_benchmark):
```

sktime/utils/validation/forecasting.py
```diff
@@ -328,13 +328,24 @@ def check_scoring(scoring):
...
+    NotImplementedError
+        if metric requires y_pred_benchmark to be passed
     """
     # Note symmetric=True is default arg for MeanAbsolutePercentageError
     from sktime.performance_metrics.forecasting import MeanAbsolutePercentageError
 
     if scoring is None:
         return MeanAbsolutePercentageError()
 
+    if scoring._tags["requires_y_pred_benchmark"]:
```
mloning, line 340: I think we have a `_has_tag(...)` or `has_tag(...)` function in utils somewhere for this, makes it a bit more readable.

sktime/utils/validation/forecasting.py
```diff
@@ -328,13 +328,24 @@ def check_scoring(scoring):
...
+        msg = "".join(
+            [
+                "Metrics requiring benchmark forecasts",
+                "are not fully implemented in the sktime API yet",
```
mloning, line 344: Perhaps this error message?
"`scoring` requiring benchmark forecasts (`y_pred_benchmark`) are not fully supported yet. Please use a performance metric that does not require `y_pred_benchmark` as a keyword argument in its call signature."

sktime/utils/validation/forecasting.py
mloning, line 340: I would also add a kwarg, i.e. `check_scoring(..., allow_y_pred_benchmark=False)` and add that to the if condition.

mloning (commented):
Thanks @RNKuhns - the manylinux CI will hopefully be fixed in #870. 

I had a first look at the code and left a few minor comments below.

sktime/performance_metrics/forecasting/_classes.py
RNKuhns: Will do; switching "requires_y_train" to "requires-y-train".

sktime/performance_metrics/forecasting/_classes.py
RNKuhns: I agree. I was looking at that at the end of the original PR (that's why I briefly had the code added to return the negative of a metric if `greater_is_better=True`). I believe scikit-learn does something like that. 

I think you mentioned that all of that is currently handled when the metrics are called.  I think in the long-run it makes sense to move that into the metrics? They way I have it written just the call in `_BaseForecastingSuccessMetric()` would need the negation logic. I'd be down to hold that to follow-on PR (we don't have any metrics where `greater_is_better=True` currently).

sktime/performance_metrics/forecasting/_functions.py
RNKuhns: I like that. Updated

sktime/performance_metrics/forecasting/_functions.py
RNKuhns: Yep, I've moved all three of those line into a non-public function.

sktime/performance_metrics/tests/test_performance_metrics_forecasting.py
RNKuhns: Updated in the function and also changed "loss" to "metric" throughout `test_performance_metrics_forecasting.py`

sktime/utils/validation/forecasting.py
RNKuhns: Sounds good. Is that so we can just toggle on the ability to allow those metrics later?

sktime/utils/validation/forecasting.py
RNKuhns: I've updated the error message.

sktime/utils/validation/forecasting.py
mloning: Yes, and people may want to use `check_scoring` in other contexts I guess

sktime/performance_metrics/forecasting/_classes.py
mloning: Yes okay let's do that in a separate PR when necessary, currently it's handled when the metrics are called/compared.

sktime/performance_metrics/base/_base.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def __init__(self, func, name=None):
+        self._func = func
+        self.name = name if name is not None else func.__name__
+
+    def __call__(self, y_true, y_pred, **kwargs):
+        NotImplementedError("abstract method")
+
+    @classmethod
+    def _all_tags(cls):
```
mloning, line 25: Perhaps we should refactor our current `BaseEstimator` in `sktime.base._base` into 
* a `BaseObject` inheriting from scikit-learn's `BaseEstimator` for handling of hyper-parameters and tags, 
* a separate `BaseEstimator` inheriting from `BaseObject` for estimators (classifiers, forecasters, etc) which additionally has the `is_fitted` and `check_is_fitted` functionality.

@RNKuhns @fkiraly any thoughts on this? If we do this, this should go into a separate PR. Happy to leave it as it is in this one.

mloning: But perhaps add a comment that `_all_tags` is copied from `BaseEstimator`

sktime/performance_metrics/forecasting/_functions.py
```diff
@@ -46,6 +48,20 @@
 EPS = np.finfo(np.float64).eps
 
 
+def _get_kwarg(kwarg_, **kwargs):
+    kwarg_ = kwargs.pop(kwarg_, None)
+    if kwarg_ is None:
+        calling_function = inspect.stack()[1].function
```
mloning, line 54: I remember reading up on how to get the name of the calling function a while ago and came across a few warnings using something like this (see e.g. here https://stackoverflow.com/questions/2654113/how-to-get-the-callers-method-name-in-the-called-method). I don't fully understand all the consequences, but a simpler option may be to just pass the name (or some context object) into the function, may not be as elegant but potentially more reliable/efficient (see e.g. `check_window_length` in `utils.validation.forecasting`). 

```python
def _get_kwarg(**kwargs, kwarg, name):
    ...
    msg = f"{name} requires ..."
    ...
```

sktime/utils/validation/forecasting.py
```diff
@@ -328,13 +329,22 @@ def check_scoring(scoring):
...
+    NotImplementedError
+        if metric requires y_pred_benchmark to be passed
     """
     # Note symmetric=True is default arg for MeanAbsolutePercentageError
     from sktime.performance_metrics.forecasting import MeanAbsolutePercentageError
 
     if scoring is None:
         return MeanAbsolutePercentageError()
 
+    if _has_tag(scoring, "requires-y-pred-benchmark") and allow_y_pred_benchmark:
```
mloning, line 341: Just a comment: Python uses some shortcuts to evaluate if statements, so if the the first one evaluate to False and there is an AND operator it stops and returns False, so it may be faster to put cheaper evaluations first, but it doesn't really matter here since both conditions are cheap (https://docs.python.org/3/library/stdtypes.html#boolean-operations-and-or-not).

mloning (commented):
@RNKuhns I had look at the changes, I think we're almost there, I only have one more suggestion to change something, the other ones don't require changes in this PR.

sktime/performance_metrics/base/_base.py
fkiraly: Yes, I've always thought that the parameter/tag functionality in `BaseEstimator` should be split off into a separate package that implements "parametric classes" :-) 
So I'm onboard with the proposed factoring - *but* what does this have to do with the metrics module? This sounds more like an independent refactor effort.

sktime/performance_metrics/forecasting/_functions.py
fkiraly: ah, I also would recommend not to do this - function behaviour should be fully defined by arguments passed to it.
Lookup of calling environment etc just invites hard-to-detect errors to happen...

sktime/performance_metrics/base/_base.py
mloning: It's related to the metrics because metrics now inherit from `BaseEstimator` which equips them with hyper-parameter and tag functionality which we want but also functionality for managing states (fitted vs unfitted) which we don't want.

sktime/utils/validation/forecasting.py
RNKuhns: Good point. I'll keep that in mind.

sktime/performance_metrics/base/_base.py
RNKuhns: I'm for it. I can add an issue about this later.  As part of the PR to close that, we can pretty easily switch the `BaseMetric` to inherit from the new `BaseObject`.

sktime/performance_metrics/forecasting/_functions.py
RNKuhns: @mloning and @fkiraly, I"m with you on this. Reading more about it, it definately seems like we want to avoid it. 

However, manually passing the name seems like a bit of a pain (albeit a minor one), so I considered one of the other solutions from stackoverflow. What are your thoughts about defining a decorator? 

The following works to pass metric name to the metrics that need it. 

```python
from functools import wraps
def expose_metric_name(func):
    @wraps(func)
    def tmp(*args, metric_name=func.__name__, **kwargs):
        return func(*args, metric_name=metric_name, **kwargs)
    return tmp


def _get_kwarg(kwarg, metric_name="Metric", **kwargs):
    kwarg_ = kwargs.pop(kwarg, None)
    if kwarg_ is None:
        msg = "".join(
            [
                f"{metric_name} requires `{kwarg}`.",
                f"Pass `{kwarg}` as a keyword argument when calling the metric.",
            ]
        )
        raise ValueError(msg)
    return kwarg_


@expose_metric_name
def mean_absolute_scaled_error(
    y_true, y_pred, sp=1, horizon_weight=None, multioutput="uniform_average", **kwargs
):
    metric_name = kwargs.pop("metric_name", None)
    y_train = _get_kwarg("y_train", metric_name=metric_name, **kwargs)
    ... # rest of function
```

mloning: That seems like a nice solution and I'm also not happy about having to pass the name into the function, but I don't think the code is very readable anymore. Compare below:
```python
def mean_absolute_scaled_error(
    y_true, y_pred, sp=1, horizon_weight=None, multioutput="uniform_average", **kwargs
):
    y_train = _get_kwarg("y_train", name="mean_absolute_scaled_error", **kwargs)
    ... # rest of function
```
The name is also unlikely to change in the future I guess. Alternatively, we could simply raise an error without reference to the function name, that may be the simplest solution.

RNKuhns: I have no problem adding the names -- I've pushed an update that includes those; i'd rather the error message was more informative with the name. 

I was more interested in testing out a way to do that more automatically. Would have been satisfying to be able to have one, but I agree that using the decorator makes that a bit hard to read.

RNKuhns:
@mloning I think I've incorporated all your comments. As a bonus I've tweaked all the docstrings to better align with NumPy conventions and pass the pydocstyle checks.

sktime/performance_metrics/forecasting/_functions.py
mloning: Okay great!

sktime/utils/validation/forecasting.py
mloning, line 341: The condition needs to be fixed here, right? 
```suggestion
    if _has_tag(scoring, "requires-y-pred-benchmark") and not allow_y_pred_benchmark:
```

RNKuhns: Yep -- good catch!

mloning (approved):
Looks all good to me - I'll merge in the next few days in case anyone else wants to take a look!

mloning (commented):
Hi @RNKuhns, we should also update `evaluate` in `sktime.forecasting.model_evaluation` to pass `y_train` as a keyword argument to the scoring: https://github.com/alan-turing-institute/sktime/blob/c611e6a3587d7b3a44cb7deefd4b6baa4897fc9b/sktime/forecasting/model_evaluation/_functions.py#L106

If we do that here, we should also add a metric that requires `y_train` to the current test cases here:  https://github.com/alan-turing-institute/sktime/blob/c611e6a3587d7b3a44cb7deefd4b6baa4897fc9b/sktime/forecasting/model_evaluation/tests/test_evaluate.py#L80

RNKuhns:
@mloning I'm working on update to evaluate. 

Ran into two minor hitches. First is that in the call to scoring within `evaluate()` the `y_true` and `y_pred` arguments were flipped (doesn't really matter for most metrics, but I fixed it). 

The bigger hitch is that the test cases include in-sample predictions. In MeanAbsoluteScaledError() we have a check to make sure that y_train is before y_true. This makes sense in the context of forecast evaluation in general (and MASE's definition). But some of the test cases are looking at in-sample predictions and in those cases the check is failed in those cases; hence, the tests fail. 

I can think of two options for approaching this: 
1. We have separate test function for MASE that just tests the common configs with only the out-of-sample forecasting horizons
2. We remove the check in `mean_absolute_scaled_error` to require `y_train` prior to `y_true`

I am in the camp that MASE is designed to be applied when evaluating out-of-sample forecasts so we have the separate test function (plus in-sample evaluation is a bad estimate of model performance on new data). But what do you think?

mloning:
@RNKuhns I agree, but perhaps the check that `y_train` is prior to `y_true` is a bit over eager and we make everything a bit simpler by not enforcing that. In principle MASE still works for in-sample predictions, so perhaps we should exclude that option. This moves the responsibility to users for making sure they're evaluating models on genuine forecasts which I think is fine. But happy to follow your lead here @RNKuhns.

fkiraly:
@RNKuhns, I'm with @mloning on this one - in my opinion, the metric should just model the "bare" mathematical/scientific object, not make any checks regarding a plausible use case. This is because user expectations - they will expect the code object to behave as the scientific one, and because of the domain driven design principle to follow this mapping.

A secondary point - in my opinion not the main argument here, but it's to the same effect - is that it can make sense for a metric in-principle to have training and test time points overlapping, for example when you are trying to estimate over-optimism of in-sample estimates. We shouldn't exclude a rare sensible use case.

sktime/forecasting/model_evaluation/tests/test_evaluate.py
```diff
@@ -77,10 +78,14 @@ def _check_evaluate_output(out, cv, y, scoring):
 @pytest.mark.parametrize("step_length", TEST_STEP_LENGTHS)
 @pytest.mark.parametrize("strategy", ["refit", "update"])
 @pytest.mark.parametrize(
-    "scoring", [MeanAbsolutePercentageError(symmetric=True), MeanSquaredError()]
+    "scoring",
+    [
+        MeanAbsolutePercentageError(symmetric=True),
+        MeanSquaredError(),
```
mloning, line 84: I suggest to remove this to speed up tests, since MSE doesn't test anything new given MAPE and MASE. Perhaps add a comment why we use MAPE and MASE (i.e. one with and one without `y_train`)

mloning: Or MSE instead of MAPE as you do below

RNKuhns: @mloning I've got that done and added the comment. Forgot to remove MeanSquaredError() after changing up the testing for MASE after removing that validation. 

Hopefully this is looking good to go.

mloning (approved):
Thanks @RNKuhns - looks all good to me now! Will merge in the next few days in case anyone else wants to take a look.

mloning:
Now merged - great work @RNKuhns 🎉

Issue #129, opened 2019-08-15
Implement Catch22 transformer


TonyBagnall:
Implement Catch22, described here https://link.springer.com/article/10.1007%2Fs10618-019-00647-x, as an sktime transformer.

Hopefully Carl Lubba and Bun Fulcher will help us with this. They have C implementations, I'm not sure whether we can just port in C or whether it needs to be made into Cython. Perhaps someone who knows more can comment, with the constraint we want to minimise dependencies

mloning:
Closed by #453

MatthewMiddlehurst:
Re-opening this. Road testing a number of classifiers that use the catch22 transform, they use a lot of memory even on relatively small datasets. Just calling the catch22 method from the C wrapper we use without storing anything rapidly increases memory used. Most likely a memory leak in the wrapper itself. Best option IMO would be to reimplement the catch22 features in Python for sktime using numba or something.

I may do this myself at some point, but if anyone else would like to take it on I can lend a hand. There are implementations in C and Matlab by the author, and a Java implementation by myself to draw from if necessary.

TonyBagnall:
hi matt, I was thinking of asking an MSc student to work on this. I may also be able to fund a UEA intern, and this could be good for them too.

TonyBagnall:
although if someone wants to take it on before that its completely fine! It is a good, fairly technical, first issue

MatthewMiddlehurst:
Working on replacing the wrapper currently. Will hopefully make a PR this week.

TonyBagnall:
@MatthewMiddlehurst this is done isnt it? Can we close?

MatthewMiddlehurst:
Yeah I would consider this done. Any problems with the revamped transformer can be raised as new issues

PR #864, opened 2021-05-08
catch22 Remake

MatthewMiddlehurst:
Implementation of catch22 features using numba. Removal of the previous catch22 dependancy. See #129 

Updates and fixes for CIF and DrCIF. Implementation of the tree classifier used by the original versions.


TonyBagnall:
@mloning hi markus, this is failing because of problems running under mac os. I've asked about finding a mac to debug since none of us own one, but thought Id ask if you have any ideas?

mloning:
@MatthewMiddlehurst @TonyBagnall no idea, sorry, I'm fixing the manylinux builds at the moment (#870), so once they're up and running again we can see if it passes on Linux. If this doesn't help I can try to look into it if I find some time.

TonyBagnall:
no problem, I'm tracking down a mac so we can debug it!

MatthewMiddlehurst:
Assuming the tests pass (they did before!), this should be ready for review.

Sidenote: I updates the estimator overview for classifiers. Is this still in use? It seems a little dated.

TonyBagnall (approved):
this all looks good to me

sktime/classification/interval_based/vector/__init__.py
```diff
@@ -0,0 +1,8 @@
+# -*- coding: utf-8 -*-
```
mloning, line 1: Quick question about the module, is `sktime/classification/interval_based/vector/` meant to be a public module for users? I feel like the folder structure becomes a bit deep.

mloning (commented):
Hi @MatthewMiddlehurst, I had a very quick look at it, just a few things I noted.

sktime/classification/interval_based/_cif.py
```diff
@@ -303,52 +302,59 @@ def _fit_estimator(self, X, y, idx):
...
 
         return [tree, intervals, dims, atts]
 
-    def _predict_proba_for_estimator(
-        self, X, classifier, intervals, dims, atts, test_size
-    ):
-        c22 = Catch22()
+    def _predict_proba_for_estimator(self, X, classifier, intervals, dims, atts):
+        c22 = Catch22(outlier_norm=True)
+        if self.tree is ContinuousIntervalTree:
```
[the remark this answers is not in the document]
MatthewMiddlehurst: Yes good catch

sktime/classification/interval_based/vector/__init__.py
MatthewMiddlehurst: Not really. Its mostly just for the interval forests (TSF/DrCIF/CIF/STSF) and its not a time series classifier so it does not need to be public. I just had no idea where to put it really. It currently isent used by default, because it is quite a bit slower than the sklearn tree and i need to figure out why.

I will put it where you think is best.

mloning: Alright, in that case, I'd just make it private (by convention) using the leading underscore `_vector`. 

As far as I know, the key bits of the scikit-learn tree are written in Cython.

MatthewMiddlehurst: Would I have to change the init at all or just the package name?

Yeah i expected it to be, but the differences were quite large! I will take another look.

TonyBagnall: Not that keen on the directory name vector myself, its not informative. Perhaps the class ContinuousIntervalTree should just be within a file, _tsf would be my choice

TonyBagnall (approved):
moving to contrib seems like a good solution

sktime/transformations/panel/catch22.py
```diff
@@ -0,0 +1,1126 @@
...
+                continue
+
+            out = peaks[i]
+            break
+
+        return out
+
+
+@njit(fastmath=True, cache=True)
+def histogram_mode(X, num_bins, smin, smax):
```
mloning, line 596: for the functions below, I suggest to make them private (leading underscore)

mloning: besides that I think this is good to go

MatthewMiddlehurst: Done

mloning:
@MatthewMiddlehurst @TonyBagnall I haven't taken another look, but I think you can merge this if you're happy with it.

Issue #866, opened 2021-05-10
ExponentialSmoothing results in error for a series with w-sun freq


eyalshafran:
I'm using sktime 0.6.0

If I use the following code:

```
forecaster = ExponentialSmoothing()
date = pd.date_range(start='01/07/2018',end='05/09/2021',freq='W')
value = np.random.normal(size=175)
ts = pd.Series(value, index=date)
forecaster.fit(ts)
forecaster.predict(fh=np.arange(1,10))
```

I get the error below. If I replace the forecaster with something else (e.g. NaiveForecaster) I do not get an error.

---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
pandas/_libs/tslibs/timedeltas.pyx in pandas._libs.tslibs.timedeltas.parse_timedelta_unit()

KeyError: 'w-sun'

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
/opt/anaconda3/lib/python3.8/site-packages/sktime/utils/datetime.py in _coerce_duration_to_int(duration, freq)
     43             if isinstance(duration, pd.TimedeltaIndex):
---> 44                 return (duration / pd.Timedelta(1, freq)).astype(np.int)
     45         except ValueError:

pandas/_libs/tslibs/timedeltas.pyx in pandas._libs.tslibs.timedeltas.Timedelta.__new__()

pandas/_libs/tslibs/timedeltas.pyx in pandas._libs.tslibs.timedeltas.parse_timedelta_unit()

ValueError: invalid unit abbreviation: W-SUN

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
<ipython-input-594-c72ddc4c380b> in <module>
      4 ts = pd.Series(value,index=date)
      5 forecaster.fit(ts)
----> 6 forecaster.predict(fh=np.arange(1,10))

/opt/anaconda3/lib/python3.8/site-packages/sktime/forecasting/base/_sktime.py in predict(self, fh, X, return_pred_int, alpha)
    226         self.check_is_fitted()
    227         self._set_fh(fh)
--> 228         return self._predict(self.fh, X, return_pred_int=return_pred_int, alpha=alpha)
    229 
    230     def compute_pred_int(self, y_pred, alpha=DEFAULT_ALPHA):

/opt/anaconda3/lib/python3.8/site-packages/sktime/forecasting/base/adapters/_statsmodels.py in _predict(self, fh, X, return_pred_int, alpha)
[...]

PR #867, opened 2021-05-11
- Fix issue #866 and some other small bugs found during implementation.

xiaobenbenecho:
- Fix bug in function plot_series in utils.plotting to plot for both DataFrame and Series. 
- Fix issue  #866, in _get_fre in utils.datetime to extract variant frequency of the series, e.g. 'w-sun'. 
- Fix bug in check_series in utils.validation.series to force all y to series if univariate. 
- Fix bug in _fbprophet and _statsmodels in forecasting.base.adpters to force a univariate data frame to series.

### DIFF
--- a/sktime/utils/datetime.py
+++ b/sktime/utils/datetime.py
@@ -63,8 +63,12 @@ def _coerce_duration_to_int(duration, freq=None):
 def _get_freq(x):
     """Get unit for conversion of time deltas to integers"""
     if hasattr(x, "freqstr"):
-	#'W-SUN,W-SAT,etc'
-        return x.freqstr[0]
+        if x.freqstr == None:
+            return None
+        elif '-' in x.freqstr:
+            return x.freqstr.split('-')[0]
+        else:
+            return x.freqstr
     else:
         return None
 
--- a/sktime/utils/tests/test_datetime.py
+++ b/sktime/utils/tests/test_datetime.py
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+import pandas as pd
+import numpy as np
+import datetime
+from sktime.utils.datetime import _get_freq
+
+
+def test_get_freq():
+    x = pd.Series(index=pd.date_range(start='2017-01-01', periods=700, freq='W'), data=np.random.randn(700))
+    x1 = x.index
+    x2 = x.resample('W').sum().index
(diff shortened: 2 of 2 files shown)


sktime/forecasting/base/adapters/_fbprophet.py
```diff
@@ -35,6 +36,7 @@ def fit(self, y, X=None, fh=None, **fit_params):
         self._instantiate_model()
         self._check_changepoints()
         self._set_y_X(y, X, enforce_index_type=pd.DatetimeIndex)
+        y, X = check_y_X(y, X)
```
mloning, line 39: This happens inside `self._set_y_X(...)` so not necessary here or any of the other files where this is added in this PR.

sktime/utils/datetime.py
```diff
@@ -53,7 +53,8 @@ def _coerce_duration_to_int(duration, freq=None):
 def _get_freq(x):
     """Get unit for conversion of time deltas to integers"""
     if hasattr(x, "freqstr"):
-        return x.freqstr
+        #e.g. when freq = 'W-SUN'
+        return x.freqstr[0]
```
mloning, line 57: So this change would simply ignore the `-SUN` information, is my understanding correct? If so, is that a good idea? In either case, we need some testing to make sure this works as expected.

mloning (commented):
Thanks for the PR @xiaobenbenecho - but could we limit this to the fix of the frequency handling? Please see my comments below.

mloning:
Hi @xiaobenbenecho I'd love to get the frequency fix merged, let me know if I can help with anything.

xiaobenbenecho:
Hi thanks.
Should I just use the change I have made or you prefer more testing code
regarding the the frequency fix?
Bests

Markus Löning ***@***.***> 于2021年6月12日周六 下午5:52写道：

mloning:
Hi @xiaobenbenecho - let's make this PR a bit more focused, there are currently main files being changed. I suggest to only focus on the frequency fix. Yes, a unit test would be good too.

xiaobenbenecho:
Hi thanks.

Submitted with corresponding test file.

Best regards

xiaobenbenecho:
Hi, yes please. A new PR has been used to fix this issue. Best.

Issue #868, opened 2021-05-13
[BUG] KeyError when forecasting using non-unit frequencies


ckastner:
**Describe the bug**

I happen to have a time series with a frequency of 30 minutes. Attempting to use this during forecasting triggers a `KeyError` in pandas.

This is because the datetime module assumes that the frequency is always 1 of some unit, ie `(1, freq)`:

https://github.com/alan-turing-institute/sktime/blob/4be4866ac50f0c6ce422716d529e95bfbf25411d/sktime/utils/datetime.py#L41-L44

However, the frequency in my case in `30T`, which is called as `pd.Timedelta(1, "30T")`, which triggers the `KeyError`. Instead, The expected call is `pd.Timedelta(30, "T")`.

The `freqstr` returned by pandas seems to be parsed into a compact one-letter representation. At least, when I manually specified a frequency of `30mins`, reading the frequency returned `30T`. If this is generally true, then a `freqstr` of length > 1 implies that `freqstr[:-1]` hold the number of actual units.

However, a cleaner solution might be to fix this in `_get_freq`:
https://github.com/alan-turing-institute/sktime/blob/4be4866ac50f0c6ce422716d529e95bfbf25411d/sktime/utils/datetime.py#L53-L58

**To Reproduce**

```python
import numpy as np
import pandas as pd

from sktime.forecasting.base import ForecastingHorizon
from sktime.forecasting.ets import AutoETS

s = pd.Series(
    np.random.rand(100),
    index=pd.date_range('20130101 09:00:00', freq='30T', periods=100)
)

y_train = s[:80]
y_test = s[80:]
fh = ForecastingHorizon(y_test.index, is_relative=False)
forecaster = AutoETS(auto=True, sp=5, n_jobs=-1)
forecaster.fit(y_train)

# KeyError
y_pred = forecaster.predict(fh)
```

**Versions**

```
System:
    python: 3.7.3 (default, Jan 22 2021, 20:04:44)  [GCC 8.3.0]
executable: /home/christian/TUWien/Energy-efficient Distributed Systems/venv/bin/python3
   machine: Linux-5.10.0-0.bpo.5-amd64-x86_64-with-debian-10.9

Python dependencies:
          pip: 18.1
   setuptools: 40.8.0
      sklearn: 0.24.2
       sktime: 0.6.0
  statsmodels: 0.12.2
        numpy: 1.19.5
        scipy: 1.6.3
       Cython: 0.29.23
       pandas: 1.2.4
[...]

PR #873, opened 2021-05-15
datetime: Construct Timedelta from parsed pandas frequency

ckastner:
Add support for non-unit frequencies (eg: 30 minutes) by parsing the pandas frequency and extracting the relevant values to pass on to pd.Timedelta().

This works for frequencies like `D` and `3D`, as demonstrated by the test.

This does not work with [anchored offsets](https://pandas.pydata.org/docs/user_guide/timeseries.html#anchored-offsets), like the `W-SUN` mentioned in #866, as it's not immediately obvious to me how one would construct a correct Timedelta from this. I suspect that the anchor can simply be dropped (7 days are 7 days, no matter whether they start on Sunday or Monday), but I'm not entirely sure, and I had a hard time constructing a test example.

Fixes: #868

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1001,7 +1001,17 @@
       "contributions": [
         "code"
       ]
-    }
+    },
+    {
+      "login": "ckastner",
+      "name": "Christian Kastner",
+      "avatar_url": "https://avatars.githubusercontent.com/u/15859947?v=4",
+      "profile": "https://github.com/ckastner",
+      "contributions": [
+        "code",
+        "bug",
+      ]
+    },
   ],
   "projectName": "sktime",
   "projectOwner": "alan-turing-institute",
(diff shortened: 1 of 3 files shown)


mloning (commented):
Thanks for the PR @ckastner - much appreciated! Make sure you add yourself to the contributors (otherwise I can do that too if that's easier). 

Please also take a look at the CONTRIBUTING.md to set up precommit for code quality checks. Let me know if you need any help or have questions.

ckastner:
@mloning, I fixed the issue that pyflake8 complained about earlier. I also added myself to .all-contributorsrc.

I previously had issues running the full test suite (now, it's even segfaulting), but running the single test file wtih my local version of pytest worked fine.

mloning:
@ckastner thanks - failures seem unrelated to the changes in this PR, I'll check back later (they usually sort themselves out).

mloning (approved):
Changes look good to me. Thanks @ckastner!

mloning:
Thanks @ckastner - now merged!

Issue #367, opened 2020-08-11
Rocket Transform across multiple cores


jmrichardson:
**Is your feature request related to a problem? Please describe.**
Is there a plan to make Rocket.transform parallelized?  Hoping to distribute the load across multiple cores if possible.

**Describe the solution you'd like**
specify n_jobs when calling rocket.transform(X, n_jobs=x)

mloning:
@angus924 any thoughts on this?

angus924:
Hi @jmrichardson and @mloning, this should already be happening automatically, it's always been part of the implementation (provided it's working properly).  Is this not working for you, i.e., can you can you see that only one or some subset of cores is being used?

mloning:
Perhaps we could expose that choice to the user using the familiar `n_jobs` argument from scikit-learn and internally numba's [`set_num_threads`](https://numba.pydata.org/numba-doc/latest/user/threading-layer.html#numba.set_num_threads)?

angus924:
That's a possibility.  This is new to me, I'd have to play around with it a bit.  In the meantime, if I understand correclty, it should already be possible (as a kind of workaround) to do this by simply calling, e.g., `set_num_threads(2)` anywhere in your code before you run the transform.

PR #875, opened 2021-05-18
Kernel based refactor

MatthewMiddlehurst:
Separating the ensembling out of ROCKETClassifier into Arsenal as was mentioned at the end of #796. Included the workaround mentioned in #367 for the ROCKET transform, setting the number of numba jobs to a set n_jobs parameter during the transform then reverting it.

Both ROCKETClassifier and the new Arsenal aren't really shapelet classifiers, so I have moved them to their own kernel-based package at the suggestion of @TonyBagnall.


TonyBagnall (approved):
looks fine to me

Issue #877, opened 2021-05-19
Create BaseObject 


RNKuhns:
**Is your feature request related to a problem? Please describe.**
Sktime's `BaseEstimator` inherits from scikit-learn's `BaseEstimator` and adds functionality for getting estimator tags and checking if an estimator has been fitted. 

The functionality to check if an estimator is fitted (and the `is_fitted` property) are not needed for "non-fitable" Sktime classes (for instance performance_metrics). However, these Sktime would benefit from inheriting from a base Sktime object that includes the estimator tag functionality. Additionally, if these classes inherit from Sktime's BaseEstimator it causes problems for unit testing that is designed to test "fitable" estimators.

**Describe the solution you'd like**
Create an Sktime `BaseObject` class that inherits from scikit-learn's `BaseEstimator` and only adds the method for returning estimator tags. 

Sktime's BaseEstimator can then inherit from the `BaseObject` and add the functionality that applies to estimators (e.g. `is_fitted` and `check_is_fitted`).

Add appropriate unit tests for BaseObject.

**Describe alternatives you've considered**
Primary alternative is to copy functionality needed functionality from BaseEstimator (e.g. #858), but inheriting the functionality is much cleaner and less error prone.

mloning:
Functionality that the `BaseObject` and tests on it should cover:
* tags
* get/set params
* pretty printing (`__repr__`)
* constructor, immutable hyper-parameters

fkiraly:
`BaseObject` has now been added in #1134, though not all features are implemented above.

Perhaps it's time to close this and move additional desired interface points to a new issue and/or new discussion.

RNKuhns:
I think that sounds reasonable.

PR #891, opened 2021-05-20
Add BaseObject to sktime.base

RNKuhns:
#### Reference Issues/PRs

#877

Creates Sktime BaseObject class that inherits from scikit-learn's BaseEstimator and only adds the method for returning estimator tags.

No

Changes in _base.py. The changes in _meta.py are just to comply with pydocstyle checks (so all of base is compliant).

#### Any other comments?

Still need to set up the unit testing (want to apply applicable the unit tests that are currently applied to BaseEstimator except those dealing with fit functionality). 

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/base/_base.py
+++ b/sktime/base/_base.py
@@ -12,9 +12,40 @@ from sklearn.base import BaseEstimator as _BaseEstimator
 from sktime.exceptions import NotFittedError
 
 
-class BaseEstimator(_BaseEstimator):
-    """Base class for defining estimators in sktime. Extends scikit-learn's
-    BaseEstimator.
+class BaseObject(_BaseEstimator):
+    """Base class for defining other classes in sktime.
+
+    Extends scikit-learn's BaseEstimator to include sktime interface for tags.
+    """
+
+    @classmethod
+    def _all_tags(cls):
+        """Get tags from estimator class and all its parent classes.
+
+        Creates a separate sktime tag interface in addition to the one in
+        scikit-learn to make sure it does not interfere with scikit-learn's tag
+        interface when inheriting from scikit-learn classes. Sktime's
+        estimator tags are class rather than object attribute as in scikit-learn.
+        """
+        collected_tags = dict()
+
+        # We exclude the last two parent classes: sklearn.base.BaseEstimator and
+        # the basic Python object.
+        for parent_class in reversed(inspect.getmro(cls)[:-2]):
+            if hasattr(parent_class, "_tags"):
+                # Need the if here because mixins might not have _more_tags
+                # but might do redundant work in estimators
+                # (i.e. calling more tags on BaseEstimator multiple times)
+                more_tags = parent_class._tags
+                collected_tags.update(more_tags)
+
+        return collected_tags
+
+
+class BaseEstimator(BaseObject):
+    """Base class for defining estimators in sktime.
+
+    Extends sktime's BaseObject to include basic functionality for fitted estimators.
     """
 
     def __init__(self):
@@ -22,7 +53,7 @@ class BaseEstimator(_BaseEstimator):
 
     @property
     def is_fitted(self):
-        """Has `fit` been called?"""
+        """Wheter `fit` been called."""
         return self._is_fitted
 
     def check_is_fitted(self):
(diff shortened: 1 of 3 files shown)


sktime/base/_base.py
```diff
@@ -12,17 +12,48 @@
...
+    Extends sktime's BaseObject to include basic functionality for fitted estimators.
     """
 
     def __init__(self):
         self._is_fitted = False
 
     @property
     def is_fitted(self):
-        """Has `fit` been called?"""
+        """Wheter `fit` been called."""
```
mloning, line 56:
```suggestion
        """Whether `fit` has been called."""
```

mloning (commented):
Thanks @RNKuhns, I think it makes sense to have this conceptually, so happy to merge this. @fkiraly what's your opinion?

RNKuhns:
@mloning do you want me to add any unit tests for `BaseObject`?

I started to look through the tests in test_all_estimators and could work up tests that make sense for `BaseObject` based on those.

fkiraly:
Yes, quite happy with this - great idea, and it gives us a place to control functionality that all objects get.

This is out of date with the main branch, but in-principle I'm happy to approve.

mloning:
Hi @RNKuhns, yes adding unit tests would be a good idea. Do we want to test all public objects (not just estimators) inherit from the `BaseObject`? This would require writing a new functions to get all classes from the repo, similar to `all_estimators` but more general.

RNKuhns:
@mlonging, I had been thinking about something along those lines. Beyond that is there anything else I should look to create a unit test for?

I'll try to get this knocked out this weekend.

mloning:
In that case, I suggest to refactor `all_estimators` so that it can internally call `all_classes` and then does some extra filtering on the classes, not sure how much work this involves. Note that the automated unit testing depends on this, so we need to be extra careful! 🙂

If we enforce these tests, we also need to modify a few existing classes which currently do not inherit from the `BaseObject` such as the cross-validators. 

I think we would want to take another look at the tags again at some point in the future, but that seems to be a separate PR.

sktime/base/_base.py
```diff
@@ -12,17 +12,48 @@
...
+
+    @classmethod
+    def _all_tags(cls):
+        """Get tags from estimator class and all its parent classes.
+
+        Creates a separate sktime tag interface in addition to the one in
+        scikit-learn to make sure it does not interfere with scikit-learn's tag
+        interface when inheriting from scikit-learn classes. Sktime's
+        estimator tags are class rather than object attribute as in scikit-learn.
+        """
```
fkiraly, line 29: this should properly describe that this implements a kind of inheritance, no?
As far as I understand, if multiple estimators in the inheritance tree have the same tag, the descendant tag overrides the parent tag.
This is important to document, perhaps with an example of what tags are returned.

mloning: I'd also like to rename the `_all_tags` method to `get_tags` and make it public if that doesn't interfere with scikit-learn's tags. But perhaps changes to the tag system are best addressed in a separate PR.

fkiraly: minor comment: I was wondering, upon a vague feeling, whether this could be just solved by using inheritance smartly? Namely, if we were just looking at the dictionary, and its keys as attribute names, this is precisely inheritance of attributes.

Inside the list, the name for this pattern is inheritance of nested attributes - it doesn't seem python supports this natively. A quick google search indicates it's an occasional topic in various other programming languages as well (mostly ending in the conclusion that the language doesn't support it).

fkiraly:
FYI, @mloning, @rnkuhns, I made some suggestions in #1099 related to tag handling.

The gist of this is:
* public `get_tag` interface point
* dynamic tag setting via setters
please review.

fkiraly:
In the next step, I would like to add a `get_summary` interface point to `BaseObject`, as discussed in #971. But that should be in the next PR. FYI @ltsaprounis, for checking/commenting whether `BaseObject` would be the right place to locate `get_summary`

ltsaprounis:
Thanks for pinging me @fkiraly. 
`BaseObject` should be a good place for the `get_summary` method. 
Is the idea to define it as an abstract method of `BaseObject` and then overwrite? 

I had a look at `statsmodels` to see how this is handled and although I got a bit lost, it looks like they define it as an abstract method of their base class [Results](https://github.com/statsmodels/statsmodels/blob/081fc6e85868308aa7489ae1b23f6e72f5662799/statsmodels/base/model.py#L977) and then they implement it for each model (or family of models) e.g. [StateSpaceMLEResults](https://github.com/statsmodels/statsmodels/blob/081fc6e85868308aa7489ae1b23f6e72f5662799/statsmodels/tsa/exponential_smoothing/base.py#L834)

fkiraly:
>Is the idea to define it as an abstract method of BaseObject and then overwrite?

Well, in this case one would probably say "implement" rather than overwrite/override since it's abstract.

But yes, exactly!

>I had a look at statsmodels to see how this is handled and although I got a bit lost, it looks like they define it as an abstract method of their base class Results 

Hm, this looks really weird - the model is wrapped in the `Results`, rather than the other way round? Do you understand the rationale behind that? I'm sure there must be a good and well-thought out reason?

ltsaprounis:
> Hm, this looks really weird - the model is wrapped in the `Results`, rather than the other way round? Do you understand the rationale behind that? I'm sure there must be a good and well-thought out reason?

Nope - I share your confusion here. Further investigation is required to get to the bottom of this.

fkiraly:
superseded by the now merged #1134

Issue #863, opened 2021-05-07
The 2021 sktime summer programme


mloning:
We invite sktime community members, new joiners, and interested data scientists to participate in the 2021 summer programme, held virtually.

The summer programme consists of events and activities which can be joined separately:

| Event | Date | Short description
|---|---|---|
| [sktime onboarding days](https://github.com/sktime/community-council/blob/main/events/sktime-onboarding-days.md) | May 19-21 |  3 days of workshops, tutorials and a sprint for new joiners and people interested in contributing |
| [sktime dev days](https://github.com/sktime/community-council/blob/main/events/sktime-dev-days.md) | June 21-25 | 5 days of workshops, tutorials and a sprint | 
| [Meet-ups](https://github.com/sktime/community-council/blob/main/events/meet-ups.md) | May 20, June 22, August 20 (tbc) | Social activities, informal discussions   | 
| [Internships & mentoring](https://github.com/sktime/mentoring#readme) | May-Aug | Internship and mentoring activities within the sktime community | 
| [SciPy 2021](https://www.scipy2021.scipy.org) sprint | July 18 | sktime sprint at the SciPy 2021 conference | 
| [sktime doc sprint](https://github.com/sktime/community-council/blob/main/events/doc-sprint.md) | Aug 16-19 | 4-day documentation sprint | 
| [Summer project presentation](https://github.com/sktime/community-council/blob/main/events/project-presentation.md) | Aug 20 | Presentation and celebration of community projects and intern projects |

thePworeo:
Hello,
in context of my semester thesis i try to use sktime for time series classification. Unfortunately i have missed the onboarding days. Would it be possible to get the documents used for tutorials and similar?

fkiraly:
sure, material from past workshops is here: https://github.com/sktime/sktime-workshop

For the introduction, I´d recommend to watch our youtube video: https://www.youtube.com/watch?v=wqQKFu41FIw

thePworeo:
> sure, material from past workshops is here: https://github.com/sktime/sktime-workshop
> 
> For the introduction, I´d recommend to watch our youtube video: https://www.youtube.com/watch?v=wqQKFu41FIw

Thank you very much, I am sorry to answer that late.
Unfortunately the link to the workshop materials does not work, github raises an 404 error.
I already watched the youtube video, but I was looking for some deeper information. I am especially interested in time series classification since i am currently working on my semester thesis about condition monitoring of machine tools through analysing different kinds of data.
[...]

PR #892, opened 2021-05-20
Update plot_series to handle pd.Int64 and pd.Range index uniformly

### LINKED ISSUE #882: The function 'plot_series' should accept series with different index types
The plot_series function throws an error when multiple series with different index types are passed as parameters.

The function should include implicit type casting of the series' indexes  

Manually converting the datatype of the series' indexes before passing them to the plot_series function using pandas 

**Additional context**
Screenshot 2021-05-20 at 4 48 00 PM

Dbhasin1:
#### Reference Issues/PRs
Fixes #882 

added the provision of typecasting different types of series indexes such as Int64, Float64, UInt64 and RangeIndex into each other 

No


sktime/utils/plotting.py
```diff
@@ -64,13 +62,21 @@ def plot_series(*series, labels=None, markers=None):
...
+    def _type(id):
+        return str(type(id)).split(".")[-1].split("'")[0]
+
     # create combined index
     index = series[0].index
+    idx_type = type(index)
     for y in series[1:]:
         # check types, note that isinstance() does not work here because index
         # types inherit from each other, hence we check for type equality
+        acc_dtypes = ["RangeIndex", "Int64Index", "UInt64Index", "Float64Index"]
```
mloning, line 74: Note that we already check if the index is valid above in `check_y(y)`.

sktime/utils/plotting.py
```diff
@@ -64,13 +62,21 @@ def plot_series(*series, labels=None, markers=None):
...
         if not type(index) is type(y.index):  # noqa
```
mloning, line 75: I would try something like this: 
```python
def _is_int_index(index):
    return type(index) in (pd.Int64Index, pd.RangeIndex)


def _check_consistent_index_type(*series):
    
    index = series[0].index    
    for y in series[1:]: 
        
        msg = "Found `series` with inconsistent index types. Please make sure that all series have the same index type."
        
        if _is_int_index(y.index):
            if not _is_int_index(index):
                raise TypeError(msg)
                
        else:
            if type(index) is not type(y.index):
                raise TypeError(msg)
```

We can then use the function `_check_consistent_index_type` at the beginning of the `plot_series` function. What do you think?

mloning (commented):
Thanks @Dbhasin1 for the PR and participation today in the summer programme (#863)!

sktime/utils/plotting.py
Dbhasin1: Yep, I just went through the code for this function and it seems like the only accepted index types as of now are Int64Index, RangeIndex, PeriodIndex and pd.DatetimeIndex. Can we add Float64Index and UInt64Index in this list too?

sktime/utils/plotting.py
Dbhasin1: Yes, this makes complete sense. I realised that the 'union' method includes implicit type casting, so there should be no need for manually doing so in the code now.

sktime/utils/plotting.py
mloning: Float64Index and UInt64Index index types don't really make sense for time indices I think, at least not for time series with equidistant time points. Why would you want to add them?

Dbhasin1: I was thinking more along the lines of how the function shouldn't throw an error if the user passes a series with index values as '1.0, 2.0, 3.0, and so on.' But yeah, otherwise, including them doesn't really make sense, so we could leave them off

mloning: Ah okay, I see, but I think it's probably safer to tell people to use an integer index in that case rather than trying to convert a float index.

sktime/utils/validation/series.py
```diff
@@ -153,3 +158,45 @@ def check_equal_time_index(*ys):
 
         if not first_index.equals(y.index):
             raise ValueError("Some (time) indices are not the same.")
+
+
+def _is_int_index(index):
+    """Check if index type is one of pd.RangeIndex or pd.Int64Index"""
+    return type(index) in VALID_INDEX_TYPES[:2]
```
mloning, line 165: Let's make this more explicit here, because the order of `VALID_INDEX_TYPES` may change

sktime/utils/validation/series.py
```diff
@@ -153,3 +158,45 @@ def check_equal_time_index(*ys):
...
+
+        if _is_int_index(y.index):
+            if not _is_int_index(index):
+                raise TypeError(msg)
+
+        else:
+            if not type(index) is type(y.index):  # noqa
+                raise TypeError(msg)
+
+        index = index.union(y.index)
```
mloning, line 201: I wouldn't include this operation into this function, it's not directly related to the checking of consistent index types.

sktime/utils/plotting.py
```diff
@@ -2,17 +2,20 @@
 # -*- coding: utf-8 -*-
 # copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
 
-__all__ = ["plot_series"]
-__author__ = ["Markus Löning"]
+"""Function for plotting one or more time series."""
```
mloning, line 5:
```suggestion
"""Functionality for plotting one or more time series."""
```

sktime/utils/plotting.py
```diff
@@ -2,17 +2,20 @@
...
 
 import numpy as np
 
 from sktime.utils.validation._dependencies import _check_soft_dependencies
 from sktime.utils.validation.forecasting import check_y
+from sktime.utils.validation.series import check_consistent_index_type
+
+__all__ = ["plot_series"]
```
mloning, line 13: This should be above the imports.

sktime/utils/plotting.py
```diff
@@ -22,7 +25,7 @@ def plot_series(*series, labels=None, markers=None):
         Names of series, will be displayed in figure legend
     markers: list, optional (default=None)
         Markers of data points, if None the marker "o" is used by default.
-        Lenght of list has to match with number of series
+        lenght of list has to match with number of series
```
mloning, line 28:
```suggestion
        The length of the list has to match with the number of series. 
```

sktime/utils/validation/series.py
```diff
@@ -153,3 +158,34 @@ def check_equal_time_index(*ys):
...
+    ----------
+    idx : pd.Index
+        Indices being checked for compatibility
+
+    Raises
+    ------
+    TypeError
+        If index types are inconsistent
+    """
+    # check types, note that isinstance() does not work here because index
```
mloning, line 181: could we move this comment to line 189 below

sktime/utils/validation/series.py
```diff
@@ -153,3 +158,34 @@ def check_equal_time_index(*ys):
...
             raise ValueError("Some (time) indices are not the same.")
+
+
+def _is_int_index(index):
+    """Check if index type is one of pd.RangeIndex or pd.Int64Index."""
+    return type(index) in (pd.Int64Index, pd.RangeIndex)
+
+
+def check_consistent_index_type(*idx):
+    """Check that index types are compatible for plotting.
```
mloning, line 169:
```suggestion
    """Check that index types are consistent with each other.
```

mloning (commented):
Great @Dbhasin1, the doc tests are passing now, just a few more minor comments and we're good to go after that I think!

Dbhasin1:
> Great @Dbhasin1, the doc tests are passing now, just a few more minor comments and we're good to go after that I think!

That's awesome, I'll incorporate your suggested changes asap

Dbhasin1:
hey, @mloning I don't know why the build (macOS-latest, 3.6) check got canceled. Could you give any pointers on how I could resolve this?

mloning (commented):
Great @Dbhasin1 - looks all good now!

The last thing I would add is a unit test to make sure we address the issue in https://github.com/alan-turing-institute/sktime/issues/882. I would just use the code snippet in the issue and see if it now passes without any errors. Please also take a look at the time index (x-axis) and check that it makes sense.

mloning (dismissed):
Hi @Dbhasin1, I've added the unit test and am happy to merge this now. Let me know if you have any final comments before merging?

mloning (approved):

Dbhasin1:
> Hi @Dbhasin1, I've added the unit test and am happy to merge this now. Let me know if you have any final comments before merging?

Hi @mloning, sorry I couldn't add the unit test in time, got a little engaged elsewhere. yeah it looks good to me, may I know what you did to check whether the time index made sense?

mloning:
Hi @Dbhasin1 take a look at the file changes here: https://github.com/alan-turing-institute/sktime/pull/892/files

mloning:
Thanks again @Dbhasin1 for the PR, great work! As said, please reach out again if you want to join the mentoring scheme or have any questions.

Dbhasin1:
> Thanks again @Dbhasin1 for the PR, great work! As said, please reach out again if you want to join the mentoring scheme or have any questions.

Thank you @mloning! Yep, will definitely do so!

Issue #880, opened 2021-05-19
fit method argument does not accept numpy.ndarray


thayeylolu:
**Is your feature request related to a problem? Please describe.**
fit method argument of ARIMA throws error if argument is a numpy array.

**Describe the solution you'd like**
Allowing automatic casting of numpy.ndarray argument to pd.Series argument in fit method

**Describe alternatives you've considered**
converting the numpy array to a series `forecaster.fit(pd.Series(y))`

**Additional context**
<img width="1112" alt="image" src="https://user-images.githubusercontent.com/13348874/118844122-9d783880-b8c2-11eb-8b17-8778b3c465b6.png">
Docstring
```python
    """Validate input data.

    Parameters
    ----------
    Z : pd.Series, pd.DataFrame, np.ndarray
        Univariate or multivariate time series. NumPy ndarray only allowed
        if `allow_numpy` = True.
    enforce_univariate : bool, default = False
        If True, multivariate Z will raise an error.
    allow_empty : bool, default = False
    allow_numpy : bool, default = True
    enforce_index_type : type, default = None
        type of time index

    Returns
    -------
    Z : pd.Series when univariate
         pd.DataFrame when multivariate
        Validated time series

    Raises
    ------
    ValueError, TypeError
        If Z is an invalid input
    """
```

mloning:
This is now working.

PR #897, opened 2021-05-21
Accept numpy arrays in check_y and coerce to pd.Series 

thayeylolu:
#880 

There is a little bug in the test. I entered the argument 

```python 

good_numpy_input = (np.array([3, 4]))

@pytest.mark.parametrize("arg", good_numpy_input)
def test_check_y_numpy_input(arg):
    with raises(TypeError):
        check_series(arg)
```

to test if `good_numpy_input` is a valid input , which it is, but it the test passes. However, it should fail. I am a bit unsure of why it is an invalid input

### DIFF
--- a/sktime/utils/validation/forecasting.py
+++ b/sktime/utils/validation/forecasting.py
@@ -111,7 +111,7 @@ def check_y(y, allow_empty=False, allow_constant=True, enforce_index_type=None):
 
     Parameters
     ----------
-    y : pd.Series
+    y : pd.Series, np.ndarray(1)
     allow_empty : bool, optional (default=False)
         If False, empty `y` raises an error.
     allow_constant : bool, optional (default=True)
@@ -132,7 +132,7 @@ def check_y(y, allow_empty=False, allow_constant=True, enforce_index_type=None):
         y,
         enforce_univariate=True,
         allow_empty=allow_empty,
-        allow_numpy=False,
+        allow_numpy=True,
         enforce_index_type=enforce_index_type,
     )
 
--- a/sktime/utils/validation/forecasting.py
+++ b/sktime/utils/validation/forecasting.py
@@ -140,7 +140,7 @@ def check_y(y, allow_empty=False, allow_constant=True, enforce_index_type=None):
         if np.all(y == y.iloc[0]):
             raise ValueError("All values of `y` are the same.")
 
-    return y
+    return pd.Series(y)
 
 
 def check_cv(cv, enforce_start_with_window=False):
--- a/sktime/forecasting/base/_base.py
+++ b/sktime/forecasting/base/_base.py
@@ -363,6 +363,7 @@ class BaseForecaster(BaseEstimator):
         X : pd.DataFrame, optional (default=None)
             Exogenous time series
         """
+        y = pd.Series(y)
         # set initial training data
         self._y, self._X = check_y_X(
             y, X, allow_empty=False, enforce_index_type=enforce_index_type
--- a/sktime/utils/plotting.py
+++ b/sktime/utils/plotting.py
@@ -56,8 +56,12 @@ def plot_series(
     from matplotlib.cbook import flatten
     import seaborn as sns
 
+    # creates a new series of type pd.Series
+    new_series = []
     for y in series:
-        check_y(y)
+        y = check_y(y)
+        new_series.append(y)
+    series = new_series
 
     n_series = len(series)
     _ax_kwarg_is_none = True if ax is None else False
--- a/sktime/utils/validation/tests/test_forecasting.py
+++ b/sktime/utils/validation/tests/test_forecasting.py
@@ -24,4 +24,4 @@ def test_check_fh_empty_input(arg):
 
 # @pytest.mark.parametrize("arg", bad_input)
 def test_check_y():
-    check_y(valid_numpy_input)
+    isinstance(check_y(valid_numpy_input), pd.Series)
--- a/sktime/utils/validation/tests/test_forecasting.py
+++ b/sktime/utils/validation/tests/test_forecasting.py
@@ -10,11 +10,18 @@ import pytest
 from pytest import raises
 
 from sktime.utils.validation.forecasting import check_fh
+from sktime.utils.validation.forecasting import check_y
 
 empty_input = (np.array([]), [], pd.Int64Index([]))
+valid_numpy_input = np.array([3, 4, 5, 6])
 
 
 @pytest.mark.parametrize("arg", empty_input)
 def test_check_fh_empty_input(arg):
     with raises(ValueError):
         check_fh(arg)
+
+
+# @pytest.mark.parametrize("arg", bad_input)
+def test_check_y():
+    check_y(valid_numpy_input)


mloning:
We need to make sure that the output from `check_y` is still a `pd.Series`, even if a numpy array is passed. Not sure what the problem is with the test, will have to look into this in more detail.

thayeylolu:
I edited the output of check_y in the doctstring to be only a Series. I also made`pd.Series(y)` the returned value in the `check_y` function

thayeylolu:
> We need to make sure that the output from `check_y` is still a `pd.Series`, even if a numpy array is passed. Not sure what the problem is with the test, will have to look into this in more detail.
`check_y` also returns `np.array` 
```python

if coerce_to_numpy and isinstance(y, pd.Series):
        y = y.to_numpy()

```
should I make it strictly a Series?

thayeylolu:
> We need to make sure that the output from `check_y` is still a `pd.Series`, even if a numpy array is passed. Not sure what the problem is with the test, will have to look into this in more detail.

I would also like to suggest that we edit `coerce_to_numpy` argument from the `check_y`  
```
python

def check_y(y, enforce_min_instances=1, coerce_to_numpy=False)

```

sktime/utils/validation/panel.py
```diff
@@ -108,17 +112,18 @@ def check_X(
...
     enforce_min_instances : int, optional (default=1)
         Enforce minimum number of instances.
     coerce_to_numpy : bool, optional (default=False)
-        If True, y will be coerced to a numpy array.
+         If True, y will be coerced to a numpy array.
 
     Returns
     -------
-    y : pd.Series or np.array
+    y : pd.Series
```
fkiraly, line 126: should this not be: `pd.Series` if `coerce_to_numpy=True`, `np.array` if `coerce_to_numpy=False`

sktime/utils/validation/forecasting.py
```diff
@@ -127,7 +130,8 @@ def check_y(y, allow_empty=False, allow_constant=True, enforce_index_type=None):
         y,
         enforce_univariate=True,
         allow_empty=allow_empty,
-        allow_numpy=False,
+        # allows numpy
+        allow_numpy=True,
```
fkiraly, line 134: is this the only non-docstring related change?
In any case, the PR preamble should make reference to the docstring fixes addressing the new docstring checks.

sktime/utils/validation/panel.py
thayeylolu: Thank you @fkiraly . I will make necessary corrections.

sktime/utils/validation/tests/test_series.py
```diff
@@ -0,0 +1,31 @@
...
+@pytest.mark.parametrize("arg", good_numpy_input)
+def test_check_series_numpy_input(arg):
+    """Check numpy array is a valid series."""
+    with raises(TypeError):
+        check_series(arg)
+
+
+def test_check_series_input():
+    """Check series input."""
+    np.testing.assert_allclose(numpy_input, check_series(numpy_input))
```
mloning, line 31: By default, if we pass a numpy array, we would still expect a series as output from `check_series`, no? So this tests should be failing as it is currently.

sktime/utils/validation/tests/test_series.py
```diff
@@ -0,0 +1,31 @@
...
+good_numpy_input = np.array([3, 4])
+
+numpy_input = np.array([3, 2, 45, 66, 34, 22])
+
+
+@pytest.mark.parametrize("arg", good_numpy_input)
+def test_check_series_numpy_input(arg):
+    """Check numpy array is a valid series."""
+    with raises(TypeError):
+        check_series(arg)
```
mloning, line 26: After the changes, this won't fail any more, right?

sktime/utils/validation/tests/test_series.py
mloning: Ah, it doesn't fail because `np.testing` accesses the values inside the series. Perhaps to make this more explicit, we can write: 
```python
np.testing.assert_allclose(numpy_input, check_series(numpy_input).to_numpy())
```

thayeylolu: `check_series` can accept a numpy nd array, but must not return a numpy ndarray?

mloning: Okay, is think that's ok, in that case, I would make the test more explicit by adding `assert isinstance(output, np.ndarray)`.

thayeylolu:
changing check_series to return a `np.ndarray` if a`np.ndarray` is the datatype of `z` raises :
AttributeError: 'numpy.ndarray' object has no attribute 'index' . At this point, I think it is best to leave check_series as it originally was to avoid the codebase breaking.

sktime/utils/validation/series.py
```diff
@@ -62,6 +62,9 @@ def check_series(
     else:
         valid_data_types = VALID_DATA_TYPES
 
+    if isinstance(Z, np.ndarray):
+        Z = pd.Series(Z)
```
mloning, line 66: @thayeylolu I believe that Z can also be multivariate series, represented as a `pd.DataFrame` or 2-dimensional array, so if it is a 2-dimensional array, we need to cast it into a `pd.DataFrame` rather than a `pd.Series`

thayeylolu: oh! I would return a dataframe in that case. Thank you Markus

thayeylolu:
image

Hi @mloning. While running the test on locally, I ran into this error. could this be due to newly added changes. I updated the branch.

mloning:
Hi @thayeylolu to fix this error you have to locally install sktime, take a look [here](https://github.com/alan-turing-institute/sktime/blob/main/CONTRIBUTING.md#unit-testing) and [here](https://www.sktime.org/en/latest/installation.html#development-version).

thayeylolu:
> Hi @thayeylolu to fix this error you have to locally install sktime, take a look [here](https://github.com/alan-turing-institute/sktime/blob/main/CONTRIBUTING.md#unit-testing) and [here](https://www.sktime.org/en/latest/installation.html#development-version).

Hello @fkiraly and @mloning. I have installed sktime locally. But I still get the same error.

mloning:
@thayeylolu why did you close this PR? Are you going to open a new one?

thayeylolu:
Yes. I would open another one.

fkiraly:
@thayeylolu, I would recommend to just continue using this one.
Even if your dev set-up has changed, you can still use the same branch? If you have a fork, your work on the remote should not be lost.

thayeylolu:
The great news first. I had been making the write changes to the wrong `check_y` function. There is a `check_y` function in ` sktime/utils/validation/forecasting.py`. This is the function that actually checks if the `forecaster.fit` argument is a valid_input.

The not so great news....
But the reason why the `check_y` function set it's default as `allow_numpy = False` in `sktime/utils/validation/forecasting.py` is because .

``` python 
# sktime/sktime/forecasting/base/_sktime.py in _set_y_X(self, y, X, enforce_index_type)
# set initial cutoff to the end of the training data
    self._set_cutoff(y.index[-1])
``` 

If y is an np.ndarray, then it does not have an index attribute.

thayeylolu:
There are lot of scripts in the code base this change would affect if implemented. Hence, alot of broken codes. I would like propose that we don't go on with the modification to allow numpy as a valid input in the `check_y` function present in `sktime/utils/validation/forecasting.py`

fkiraly:
@thayeylolu, I still think this is a good idea, and suitable for you to work on.

I would suggest three things:

* merge the current updated interface first, and have a look at the new forecasting extension template (in `extension_templates` directory) - this is where we will migrate the interface to. This ensures that checks/etc happen only in one place - the `BaseForecaster`, and can be changed in one place, as opposed to "in every forecaster" as it was before.
* perhaps, in the meantime, help migrate current estimators to the new interface, to ensure we can easily make changes to checks etc.
* regarding the issue with the index: I would suggest, just coerce the `numpy` to `pd.Series` and assume that the index is identical with the (integer) array index.

thayeylolu:
Thank you @fkiraly . I have been going through the new changes. 
- I would have to conform the STLforecaster by adding the mandatory implementation of `_fit` and `_predict` methods.
- I honestly don't know how to go about with `assume that the index is identical with the (integer) array index.`

fkiraly:
superseded by #980

PR #902, opened 2021-05-22
Arsenal fix, extended capabilities and HC1 unit tests

MatthewMiddlehurst:
Fix so the Arsenal classiifer can produce probabilities correctly.

Classifier capabilities extended to include contracting and train probability estimates

Removal of HC1 from unit test exclusion. As HC1 uses STC there some are excluded due to contracting, These may be pretty expensive in terms of time spent, I reduced the amount of time STC uses in unit tests.


sktime/classification/kernel_based/_arsenal.py
```diff
@@ -22,10 +21,9 @@
 
 
 class Arsenal(BaseClassifier):
```
mloning, line 23: Can we rename that to `ArsenalClassifier` so that we can distinguish it from its regressor counterpart? I think most classifiers now have Classifier in their name, it's just clearer and what people will expect coming from scikit-learn.

mloning (commented):
Thanks @MatthewMiddlehurst - I appreciate the docstring fixes (hope it wasn't too painful)!

sktime/classification/kernel_based/_arsenal.py
TonyBagnall: No

TonyBagnall:
It's was painful, but my OCD side kept me going :)

MatthewMiddlehurst:
As the commits suggest it was all Tony, a good chunk of them out of the way not thankfully! Going to put this as ready to review since it passes all the tests.

sktime/classification/kernel_based/_arsenal.py
mloning: Ok

PR #911, opened 2021-05-29
Fix https://github.com/alan-turing-institute/sktime/issues/910

### LINKED ISSUE #910: [BUG] Prophet adapter fails to clone after setting parameters
**Describe the bug**

Using sklearn's `clone` on `Prophet` after either `seasonality_prior_scale`, `changepoint_prior_scale` or `holidays_prior_scale` were set with a non-float value through `set_params` will raise a `RuntimeError` during cloning. This is due to the `float` conversion happening in `__init__`, which is against what sklearn expects during cloning - it expects that all of the values passed as arguments to `__init__` will not be modified in any way when being set as object attributes.

**To Reproduce**

```python
model = Prophet()
model.set_params(changepoint_prior_scale=1)
clone(model)
```

This will raise a sklearn `RuntimeError`.

**Expected behavior**

No exception is raised.

**Additional context**

**Versions**
<details>

```
System:
    python: 3.8.5 (default, Sep  4 2020, 07:30:14)  [GCC 7.3.0]
executable: /home/baum/anaconda3/bin/python
   machine: Linux-5.11.0-7614-generic-x86_64-with-glibc2.10

Python dependencies:
          pip: 20.2.4
   setuptools: 50.3.1.post20201107
      sklearn: 0.23.2
       sktime: 0.5.3
  statsmodels: 0.12.2
        numpy: 1.20.1
        scipy: 1.6.1
       Cython: 0.29.17
       pandas: 1.2.3
   matplotlib: 3.4.1
       joblib: 1.0.1
        numba: 0.51.2
     pmdarima: 1.8.0
      tsfresh: None
```
</details>

Yard1:
#### Reference Issues/PRs

Fixes #910 

This PR fixes an issue where a `RuntimeError` will be raised by sklearn's `clone` after either `seasonality_prior_scale`, `changepoint_prior_scale` or `holidays_prior_scale` were set with a non-float value through `set_params` for a `Prophet` object.

No

#### Any other comments?

I'd suggest checking other estimators for similar issues. The attributes set in `__init__` must not be modified in any way from what's passed as arguments, otherwise sklearn will raise an error during cloning.

##### For all contributions

##### For new estimators


RNKuhns:
@Yard1 it looks like your getting some docstring quality check errors. Now that this has been turned on that validation in the CI routine, any changed files need to have docstrings that comply with pydocstyle's checks (and docstrings should follow NumPy docstring conventions). 

I've copied a screen shot of the details from the doc-quality CI check below. 
image

Note that you can run the tests locally by running `pydocstyle` from the command line on the files you want to check. 

In this case, I think the updates you'll have to make to pass the docstring quality test will be really simple. 

To fix the D100 error you just need to include a short docstring (potentially 1-liner) at top of the code file. 

I think the D205 and D411 are actually the same issue. 

Try adding a space between the summary docstring line and the parameters section of the Prophet class's docstring:

```python
"""Prophet forecaster by wrapping fbprophet.

  Parameters
  ----------
  freq: String of DatetimeIndex frequency. See here for possible values:
      https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html
      #timeseries-offset-aliases
...
"""
```

fkiraly:
>Note that you can run the tests locally by running pydocstyle from the command line on the files you want to check.

There's also a much more convenient way, if you are using vs code: just install `pydocstyle` in the env you are using, and activate `pydocstyle` linting in files/preferences

Yard1:
Sure, I can do that.

RNKuhns:
@Yard1 the docstring quality checks have just been added to the CI routine. It creates some overhead in the short-run, but in the long-run will help ensure good docstring practices.

mloning (approved):

mloning:
@aiwalter could you review please since you implemented the model originally?

aiwalter:
@Yard1 I currently dont remember why I added this type conversion, do you think we could remove it instead?

Yard1:
@aiwalter I don't know, can we? I'm not the best authority on Prophet, just wanted to fix an exception in another project 😅 I'd imagine Prophet does its own type conversion internally, but on the other hand, leaving it as it is in this PR won't have any downsides I can see.

aiwalter:
Ok it was just copied from here https://www.github.com/facebook/prophet/tree/master/python%2Fprophet%2Fforecaster.py

So lets go for your solution

aiwalter (approved):

mloning:
Now merged - thanks @Yard1 :tada:

Issue #510, opened 2020-11-29
Proliferation of forecasting base classes?


fkiraly:
@mloning, @big-o (creditef as contributor in the base classes):

I wanted to work on #464 and looked at the base classes in forecasting, amongst others.
I looked at the base classes and feel there is some unnecessary proliferation of classes going on.

Why: we seem to have two nested template base classes, `BaseForecaster` and `_SktimeForecaster`. 

The (undocumented) extension pattern seems unnecessarily complicated.
To give one example, implementation of prediction (can you confirm that this is the intended pattern, see new post in #464):

The `BaseForecaster` expects from extender to inherit and implement `predict`. 
However, if you implement a sktime native forecaster (which most extenders will be concerned with), you have to inherit from `_SktimeForecaster`, which inherits `BaseForecaster` and implements `predict`, but now you have to implement `_predict`.

I would strongly favour there being only one base class - with all private functions - and a clean docstring that states what to implement.
Only vendor adaptors and similar overrides should have nested template schemes.
Mixins (like for the forecasting horizon) are ok i.m.o.
That way, for example, one wouldn't have to track overrides through two parent classes...

What do you think, @mloning, @big-o? Next major version to refactor this?
Considering use of `abc` since it's getting a little complex?

tasmn:
I think my contribution to those classes was when I tidied up the prediction intervals API, so I only have one very specific perspective on this. That said, I felt very much the same at the time I worked on it - I was unsure what the justification was for two separate classes (there was probably a good reason at the time, but it seems less clear now). It was difficult to figure out where new methods should live.

Something I liked in earlier versions of sktime that perhaps we could return to in this refactor is a simplification of the docstrings. At the moment, the documentation is spread across both private and public methods. Not only does this clutter up the codebase a little (and make it less readable), but it doesn't play as nicely with documentation tools like sphinx. I think it could be better to keep all of the meaningful user-focussed documentation in the public methods, and the private methods can contain very light documentation notes for developers (if absolutely necessary), or be undocumented (comments where they are called in the public methods can explain what they are being used for if needed). For example:

```python
def predict(self, fh, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
    """Meaningful, generic, napoleon-formatted user documentation."""
    # Call class-specific implementation.
    return self._predict(fh=fh, X=X, return_pred_int=return_pred_int, alpha=alpha)

@abstractmethod
[...]

PR #912, opened 2021-05-30
Forecast base class refactor and extension template

fkiraly:
This PR is addressing the forecaster base class proliferation discussed in #510, by simplifying the base class inheritance tree and the streamlining the private method logic.

Specifically, I've moved the logic from `_SktimeForecaster` and the forecasting horizin mixins into `BaseForecaster`, and adopted uniformly the principle of having a public method (`fit`, `predict`, etc) with checks/plumbing, dispatching to a "core logic" version (`_fit`, `_predict`, etc) where validated arguments can be assumed.

The burden on extenders becomes much lighter, since it is now possible to only focus on the "core logic" implementation, instead of having to keep in mind a myriad of inconsistent and constantly shifting conventions in checks and other plumbing.

As side effects, if we get this right, this should make a few things on the roadmap easier:
* the extension guidelines #464. Right now, the aforementioned implicit conventions are too many and intricate to write useful extension guidelines, in my opinion.
* extending to the multivariate case
* input/output checks and the eternal data container discussion, this can go in the plumbing

As a proof-of-concept regarding ease of extension, this PR also contains a highly annotated extension template in the `extension_templates` folder.

In terms of review, the key file is `forecasting/base/_base`, with corresponding changes (contraction and deletion) in `base/_sktime`.

I've tried to keep the interface consistent as much as possible (only changing internal logic).

Interface contracts with all the earlier estimators are still honoured, via loopthroughs and default behaviour that ensures that everything still works if `fit`, `predict` etc are overridden by the current descendants, as opposed to `_fit` and `_predict`.
The only change I had to made to descendants is set `self._is_fitted=False` at the start of `fit`, which is a minimally invasive change that's also separately reviewable as PR #941.

There should only be a single change that breaks the tagging/inspection interface, which is replacing the choice of which forecasting horizon mixin one inherits from with a choice of the new `fh_in_fit` tag (can be `'requried'` or `'optional'`).
Temporarily, I've left both mixins and the `_SktimeForecaster` in and handled their behaviour as expected (inheriting from a mixin sets the `fh_in_fit` tag), so existing classes should still run, unless there is interface inconsistency or bugs in my refactor. Mid-term, we may want to remove the fh-mixins and `_SktimeForecaster` in favour of a single `BaseForecaster` class.


fkiraly:
note to self and @mloning: handling of "is fitted?" state checks and state indicator variables seems to be inconsistent across the various estimators. For instance, this creates an error in the first jupyter notebook in the first "tuning" cell: `_Reducer` in `compose._reduce` does not set `_is_fitted` to `False` when `fit` is called, before doing the `_set_fh` check. That way, `_set_fh` cannot distinguish whether it is in `fit` when it sees an old `fh` or in `predict`, in a case where grid search `ForecastingGridSearchCV` fits the same estimator twice.

This can be solved by adding the line `self._is_fitted = False` at the start of `_Reducer._fit` (before `_set_fh` is called), though I'm not sure whether it makes sense right now to do a bigger refactoring, before the draft has been looked at.

fkiraly:
I believe the tests are now all passing (at least locally), so this is ready for review.

I suggest looking at #941 first, merging it would reduce the changes to the template interface related files.

fkiraly:
actually, some "estimator" related tests were faililng, while "forecasting" tests all seem to be passing.

I believe the reason is that `check_estimator_tags` requires tags to be boolean.
The tag I added for the forecasting horizon requirement is a string.
Why can't tags be string? I've proposed a change to the test, but I'm happy to revert and change the tag to boolean.

sktime/forecasting/base/adapters/_fbprophet.py
```diff
@@ -32,6 +32,7 @@ def fit(self, y, X=None, fh=None, **fit_params):
         -------
         self : returns an instance of self.
         """
+        self._is_fitted = False
```
mloning, line 35: Why do we set this at the beginning of every `fit` rather than in the constructor?

extension_templates/forecasting.py
```diff
@@ -0,0 +1,244 @@
+# -*- coding: utf-8 -*-
+"""
```
mloning, line 2: This is nice, could this go into the extension guidelines?

extension_templates/forecasting.py
```diff
@@ -0,0 +1,244 @@
...
+    # todo: add any hyper-parameters and components to constructor
+    def __init__(self, parama, paramb="default", paramc=None):
+
+        # todo: write any hyper-parameters and components to self
+        self.parama = parama
+        self.paramb = paramb
+        self.paramc = paramc
+
+        # todo: initialize None parameters, where necessary
+        if paramc is None:
```
mloning, line 72: hyper-parameters should be stored as they are and interpreted in fit

sktime/forecasting/base/_base.py
```diff
@@ -23,10 +59,31 @@ class BaseForecaster(BaseEstimator):
...
+
+        self._y = None
+        self._X = None
+
+        # forecasting horizon
+        self._fh = None
+        self._cutoff = None  # reference point for relative fh
+
+        # defaults for estimator tags
+        if not hasattr(self, "_tags"):
```
mloning, line 71: Can we handle tags on the class level rather than object level? See the tag functionality on the `BaseEstimator`

sktime/forecasting/base/_base.py
```diff
@@ -23,10 +59,31 @@ class BaseForecaster(BaseEstimator):
...
+            self._tags = dict()
+
+        if hasattr(self, "fhmixinflag"):
+            self._tags["fh_in_fit"] = self.fhmixinflag
+
+        if "fh_in_fit" not in self._tags.keys():
+            self._tags["fh_in_fit"] = "required"
```
mloning, line 78: Could we not just keep the binary tags and make it `fit-requires-fh`? And could that tag not be defined by the mixin classes, so that we don't have to use the `fhmixinflag`? Or am I missing something here?

mloning (commented):
@fkiraly this looks great, I had a quick look and left a few comments.

sktime/forecasting/base/adapters/_fbprophet.py
mloning: see #941

extension_templates/forecasting.py
fkiraly: I think it should be linked at the top of - the file is meant to be completely self-explanatory, accompanied by the more mathematical conceptual model.

fkiraly: I'm not going to change the docs though in this PR, that would be dependent on this PR in order to keep feature additions well-delineated.

extension_templates/forecasting.py
fkiraly: mh, but isn't it common practice to move complex defaults into the constructor? E.g., if the default of `estimator_param` is some pipeline construct, you would set the argdefault to `None` and replace it with the pipeline construct initialized in the constructor if `estimator_param is None`?

sktime/forecasting/base/_base.py
fkiraly: this is necessary to make the mixin workaround work, to ensure compatibility with the current interface. I would argue to refactor this later and accept a temporary inconsistency in the `_tags` interface, to allow dependent code to work.

Mid-term, I agree that tags should all live in class variables.

sktime/forecasting/base/adapters/_fbprophet.py
fkiraly: in terms of interface: I'd say, let's keep this in.
Then, in next PR, move individual forecaster logic to `_fit` and so on, so the initial `self._is_fitted = False` appears only in the `BaseForecaster` and nowhere else. Then, refactor so state handling is nicer (however that will look like).

Much of what you dislike is caused by ensuring downwards compatibility of the interface refactor.

extension_templates/forecasting.py
fkiraly: is that a sklearn convention?

sktime/forecasting/base/adapters/_fbprophet.py
fkiraly: >Why do we set this at the beginning of every fit rather than in the constructor?

See answer in #941. If we set it in the constructor, it does not prevent unexpected behaviour when `fit` is called twice, e.g., in a grid search that calls `fit` twice (and/or isn't careful and makes a deepcopy first).

extension_templates/forecasting.py
mloning: The root directory doesn't seem like the right place for it. Why don't we move this to a new PR for doc changes?

extension_templates/forecasting.py
mloning: I believe so, I think otherwise settings hyper-parameters in `estimator.set_params()` will be treated differently than in the constructor. The constructor should be largely free of any logic on hyper-parameters.

sktime/forecasting/base/_base.py
mloning: But we have tags working already, so why not use that? Which part didn't work with the current system?

fkiraly: Yes, using tags for "fh needed in fit" would work, and that is my long-term goal.

But I did not want to make invasive changes to existing forecasters, so I built a workaround that the tag gets created on creation when something inherits from the mixin. Long-term, I´d want to remove the mixins, and work with tags only.

Short-term, I decided to make mixins work like that so we can first do the interface refactor and then slowly remove the "patches" that make it work with the current inheritance structure. That would see the mixins and `_SktimeForecaster` fully removed - but not in this PR to avoid the "too many files" problem.

extension_templates/forecasting.py
fkiraly: Why isn't the root directory the right place? The extender user doesn't want to look into the nested folder structure of sktime. They just want to get a template for the file they need to write, and it should work plug&play with sktime. I'd advocate to have one folder which collects all extension templates. The extension guidelines tell people how to build a local estimator for extension. Contributing back to sktime is an advanced case of that, but separate.

extension_templates/forecasting.py
fkiraly: not sure whether I see this - I know that get/set params look at the `__init__` signature, and there should be no kwargs. Otherwise, I don't see how initialization would affect the `BaseEstimator` functionality - would you mind explaining?

The principle by which I was going is that initialization (including of default parameters) should happen in ... the constructor, also named `__init__` for that reason.

extension_templates/forecasting.py
mloning: I see this as documentation and that should live in the root directory, but somewhere in the docs, no?

extension_templates/forecasting.py
mloning: Take a look at scikit-learn's dev guide where they state this: https://scikit-learn.org/stable/developers/develop.html#apis-of-scikit-learn-objects

extension_templates/forecasting.py
fkiraly: it's not documentation, it's not an md or an rst file, it's a py file that the extender would fill in.
Putting it in the documentation folder would "hide" it and make it less useful. I think it is its own "kind" of file, like the examples which also deserves its own folder.

The idea is:

* have extension guidelines in the docs
* the extension guidelines give a description and point to the template file
* the user then takes the template file, copies it in the intended location (sktime or own repo) and starts filling it out, following instructions in the file and the extension guidelines

If the file is "hidden", this workflow is unnecessarily complicated.

extension_templates/forecasting.py
fkiraly: I think the (undocumented) exception is estimators (not parameters!) passed to `__init__`, which can have validation and/or init logic? For example the `Pipeline` validates its `steps` in `__init__`, see https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/pipeline.py

For initialization of default estimators, I think there's no precedent for that in scikit-learn. But, in extension, I'd also think it's fine, and even necessary. Because `get/set_params` looks at the composite estimator, and wouldn't "see" the parameters of the default estimator unless it's already set in `__init__`.

sktime/forecasting/base/_base.py
mloning: see #950

sktime/forecasting/base/_base.py
mloning: see #950

extension_templates/forecasting.py
mloning: Yes, interesting, they do call it again though in [`_fit`](https://github.com/scikit-learn/scikit-learn/blob/9a13bdfaf1a47188d2e1262f0308f317e6662e8d/sklearn/pipeline.py#L278)

mloning: So I suggest to remove the parameter initialisation from the template.

extension_templates/forecasting.py
mloning: I like the idea, but prefer not to add the templates to the root directory, especially since there are more templates to come. Why don't we create a `templates/` folder?

fkiraly: slightly confused, it already is? It's in a folder called `extension_templates`

extension_templates/forecasting.py
fkiraly: ah, I see - yes, will do.

fkiraly: I've updated the extension template now along these lines, have a look.

extension_templates/forecasting.py
mloning: Ah yes, though the file itself was in the root directory, would prefer `templates` as it's shorter, but either way it's ok.

fkiraly: but `templates` is less descriptive :smiley: - I'm not too set on this though.

mloning: We should mention them in the README too and perhaps add a short README to the folder itself too but let's do that in a separate PR

mloning: either name works for now

mloning: ideally we also import them into the sphinx rst doc files somehow and have some basic tests on them to make sure they are correct

fkiraly: yes, agreed - my proposed steps for that future PR are:

* take the current markdown for forecaster specs and extension templates
* simplify, since the interface has simplified; re-work with reference to the py template
* move to sphinx rst, integrate to current docs

extension_templates/forecasting.py
mloning: looks good now

mloning (commented):
With the suggested changes in #950, I'm happy to merge this as an intermediate step in the refactoring.

mloning (approved):
Happy for this to be merged now.

fkiraly:
hm, the checks didn't finish after I merged your PR - I'd assume everything is still alright, @mloning?

mloning:
@fkiraly in general please wait for the CI to complete before merging so that we don't end up breaking the `main` branch

fkiraly:
well, they *did* pass on your PR which was ahead of this one, so that technically counts? Will be more careful next time.

Issue #922, opened 2021-06-03
Implement ThetaLines transformer


GuzalBulatova:
**References re-design of Theta Forecaster #854**

**Describe the solution you'd like**
Implement `ThetaLines` series to series transformer to decompose original data (univariate series) into theta lines for Theta forecaster .

**Related issues:**
Poor Theta model predictions #421
Implementation of AutoTheta Forecaster #738

**Existing implementations:**
- in sktime `sktime/forecasting/theta.py` (theta parameter is assumed)

other implementations in R:
- Theta https://rdrr.io/cran/forecast/man/thetaf.html
- AutoTheta https://github.com/vangspiliot/AutoTheta/blob/master/code
- M4 AutoTheta https://github.com/Mcompetitions/M4-methods/tree/master/260 - KaterinaKou
- 4Theta https://github.com/Mcompetitions/M4-methods/blob/master/4Theta method.R
- (boxcox + theta) https://github.com/Mcompetitions/M4-methods/blob/master/260 - KaterinaKou/final_M4.R
- Theta in statsmodels https://www.statsmodels.org/devel/examples/notebooks/generated/theta-model.html
- differencing is implemented in https://alkaline-ml.com/pmdarima/_modules/pmdarima/utils/array.html#diff

Lovkush-A:
@GuzalBulatova Very thorough description! I look forward to seeing how this progresses. Happy to help out if I can.

GuzalBulatova:
@Lovkush-A Thank you Lovkush, I appreciate it, reviews and general feedback would help a lot! 🚀

GuzalBulatova:
ThetaLines transformer is implemented: #923 (merged)

PR #923, opened 2021-06-03
Add ThetaLines transformer

GuzalBulatova:
#### Reference Issues/PRs
References #922. See also #854.

Implements ThetaLines transformer.

No.

##### For all contributions


sktime/transformations/series/theta.py
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env python3 -u
+# -*- coding: utf-8 -*-
+"""copyright: sktime developers, BSD-3-Clause License (see LICENSE file)."""
+
+__author__ = ["Markus Löning"]
```
Lovkush-A, line 5: *You* are the author of this!, not Markus.

sktime/transformations/series/theta.py
```diff
@@ -0,0 +1,59 @@
...
+import numpy as np
+import pandas as pd
+
+from sktime.transformations.base import _SeriesToSeriesTransformer
+from sktime.forecasting.base import ForecastingHorizon
+from sktime.forecasting.trend import PolynomialTrendForecaster
+
+
+class ThetaLines(_SeriesToSeriesTransformer):
+    """Decompose the original data into two or more ThetaLines."""
```
Lovkush-A, line 17: Maybe include reference to footnote `[1]` here, too?

sktime/transformations/series/theta.py
```diff
@@ -0,0 +1,59 @@
...
+        ----------
+        Z : pd.Series
+            Series to transform.
+        X : pd.DataFrame, optional (default=None)
+            Exogenous data used in transformation.
+
+        Returns
+        -------
+        pd.DataFrame
+            Theta lines[1].
```
Lovkush-A, line 36: maybe worth including more detail for returns part of docstring:
* name of variable that is returned
* how the shape of the dataframe is determined by the inputs. Something like number of rows = len(Z) and number of columns = len(theta).

Lovkush-A:
@GuzalBulatova. Just had a quick look, and it looks good! Logic makes sense and it looks like it should work. I made a few minor comments/suggestions.

GuzalBulatova:
@Lovkush-A thank you very much for your feedback! I agree with all and I'll make the changes 👍

sktime/transformations/series/theta.py
Lovkush-A: I don't know if sktime has 'rules' for how to reference.  Here is example where reference is in the docstring of the class: https://github.com/alan-turing-institute/sktime/blob/main/sktime/forecasting/theta.py

@mloning. For a references to papers, is there a norm for it to be in the docstirng of the class, or docstring of a method within the class, or ... ? Or does it basically not really matter as long as it is somewhere

sktime/transformations/series/theta.py
```diff
@@ -32,28 +42,37 @@ def transform(self, Z, X=None):
...
         References
         ----------
         [1] E.Spiliotis et al., "Generalizing the Theta method for
         automatic forecasting ", European Journal of Operational
         Research, vol. 284, pp. 550-558, 2020.
         """
+        self.check_is_fitted()
+        z = check_series(Z, enforce_univariate=True)
+
+        if len(self.theta) < 2:
```
mloning, line 59: I suggest to write a helper function `_check_theta` for this so that users can either pass a single theta or a list or tuple. If users pass a single float, I suggest to still wrap it in a list internally, so that we don't have to change the for-loop below. Should we also enforce non-negative values?

sktime/transformations/series/theta.py
```diff
@@ -32,28 +42,37 @@ def transform(self, Z, X=None):
...
+        thetas = np.zeros((z.shape[0], len(self.theta)))
         for i, theta in enumerate(self.theta):
-            thetas[:, i] = _theta_transform(Z, trend, theta)
+            thetas[:, i] = _theta_transform(z, trend, theta)
         return pd.DataFrame(thetas, columns=self.theta)
 
 
 def _theta_transform(Z, trend, theta):
+    # obtain one Theta-line
+    z = check_series(Z)
```
mloning, line 75: since this is a private helper function, we don' have to check `z` again here

mloning (commented):
Hi @GuzalBulatova, looks very good! Just two minor comments.

sktime/transformations/series/theta.py
GuzalBulatova: Thanks so much for the feedback, @mloning !
I thought that it doesn't make sense for ThetaForecaster to have less than two Theta-lines. Shall we wrap it in a list and add second element if a single theta is passed?
I don't know if anybody would like to use negative values, because they basically gives the same transformation as corresponding positive value, just reflected over linear regression. 
Untitled

Lovkush-A: * Single theta. The thetaforecaster always has at least two lines, but maybe there are other situations where a user only wants to create a single theta line?
* Negative theta. I also think it is unlikely that negative theta is useful, so worth enforcing theta>=0.

mloning (commented):
@GuzalBulatova as discussed:

We still need to do:
* add `_tags = {"transform-returns-same-time-index": True, "univariate-only": True}`
* add `_check_theta` function for input validation and conversion if necessary

For unit testing, we should cover the following cases:
* theta=0 - check values
* theta=1 - check values
* optionally, compare against 2nd equation based on differencing
* theta=(t1, t2, ..., tn) - check shape of the output = (y.shape[0], n)

@Lovkush-A @fkiraly currently all series-to-series transform return a series if they receive a series as input, the `ThetaLinesTransformer` on the other hand may return a data frame. Should we add a new tag for this or allow all series-to-series transformers to return data frames?

Lovkush-A:
> Should we add a new tag for this or allow all series-to-series transformers to return data frames?

Good question! My preference is to do what is most consistent with the name "series-to-series transformers". To me, this means changing the ThetaLines transformer to only accept a single theta, and have the creation of multiple theta lines done by end-users / done within the ThetaForecaster.

However, big unknown for me is how this interacts with various pipeline tools / composition tools in sktime.

sktime/transformations/series/theta.py
fkiraly: I wouldn't actually enforce theta >=0 (why? the line is perfectly defined), but I agree on one line

fkiraly: the "recommendations" are encoded in sensible defaults, we shouldn't restrict the potential choices purely on a vague belief that it might not be useful (as opposed to, e.g., it being mathematically nonsensical).

fkiraly:
>Should we add a new tag for this or allow all series-to-series transformers to return data frames?

My opinion is that "data frame" is just the machine type for the scitype "multivariate series", so it would still be series-to-series.

However, we want to refactor transformers similar to forecasters so that data frame type is supported.

For now, we can perhaps return a data frame in the genuinely multivariate case (i.e., multiple theta are requested, theta is a list) but make a note that that use is not interface compliant.

mloning:
Yes, I agree with @fkiraly. 

@GuzalBulatova So we can do the simplest fix where we add `pd.DataFrame` to the accepted output types for series-to-series transformers.

sktime/transformations/series/theta.py
GuzalBulatova: @fkiraly thank you for review! The thinking I had was this: the scenario when user knowingly passes negative theta in order to get the mirrored series seemed way less probable than one when someone passes negative values by mistake and gets mirrored forecasts that don't make sense. It's non mathematically nonsensical though, so I absolutely agree that the defaults should be enough in this case.

fkiraly: well, in my opinion, it's better to let 9 users pass a negative value by mistake than nanny the 1 user who may want to do this, potentially for a very good reason (e.g., trying it out in a sensible tuning composite)

But I think this might be a matter of opinion and design principle - about how "opinionated" `sktime` should be in prescribing what it thinks are the best choices. I'm more "laissez faire" on this point...

sktime/transformations/series/theta.py
```diff
@@ -66,7 +67,10 @@ def transform(self, Z, X=None):
         theta_lines = np.zeros((z.shape[0], len(theta)))
         for i, theta in enumerate(theta):
             theta_lines[:, i] = _theta_transform(z, trend, theta)
-        return pd.DataFrame(theta_lines, columns=theta)
+        if isinstance(self.theta, (float, int)):
+            return theta_lines
```
mloning, line 71:
```suggestion
            return pd.Series(theta_lines, index=y.index)
```

sktime/transformations/series/theta.py
```diff
@@ -66,7 +67,10 @@ def transform(self, Z, X=None):
...
+        else:
+            return pd.DataFrame(theta_lines, columns=[self.theta])
```
mloning, line 73:
```suggestion
            return pd.DataFrame(theta_lines, columns=theta, index=y.index)
```
We can use the validated `theta` here.

sktime/transformations/series/tests/test_theta.py
```diff
@@ -6,28 +6,41 @@
...
 
 def test_theta_1():
+    # with theta = 1 Theta-line is equal to the original time-series
     y = load_airline()
     t = ThetaLinesTransformer(1)
     t.fit(y)
-    actual = t.transform(y)
-
+    line = t.transform(y)
+    actual = line[:, 0]
```
mloning, line 36: this shouldn't be necessary anymore now that we return pd.Series for a single theta

sktime/transformations/series/tests/test_theta.py
```diff
@@ -6,28 +6,41 @@
...
 
 
 def test_theta_0():
+    # with theta = 0
     y = load_airline()
     t = ThetaLinesTransformer(0)
     t.fit(y)
-    actual = t.transform(y)
+    line = t.transform(y)
+    actual = line[:, 0]
```
mloning, line 22: same as below

mloning (commented):
@GuzalBulatova I think these are my last comments, looks very good overall!

GuzalBulatova:
@fkiraly @mloning @Lovkush-A I think this PR is ready. Once the tests are done running, could you please take a look?

mloning (approved):
This looks very clean, I think it's ready to be merged! Great work! 

A few comments about what we could add going forward:
* let users set a forecaster for the trend estimation
* support multivariate data (but already addressed through #1044)

@Lovkush-A do you want to review before we merge?

mloning:
Now merged - great work @GuzalBulatova! 🎉

Lovkush-A:
Woohoo! Nice one, @GuzalBulatova. FYI, I did not see the comment above about whether I wanted to review, but it is unlikely I would have anything to add based on feedback already provided.

Issue #704, opened 2021-02-22
Automated learning algorithm overview


mloning:
Contributors: @fkiraly, @mloning, @afzal442, @abdulelahsm, @sparkingdark 

## idea:
web data base for learning algorithms with fields for: 
* class name, 
* scitype (e.g. forecaster, classifier, etc.), 
* original contributor,
* maintainer/current [code owner](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners),
* link to API reference,

optional additional fields: 
* links to literature references, 
* package dependency, 
* code health/status (under development, mature, etc), 
* link to example notebooks

## use cases:
* search for suitable algorithms given a problem
* search for information about a given algorithm

## reference example from mlr3 library:
* https://mlr3extralearners.mlr-org.com/articles/learners/learner_status.html
* https://mlr3extralearners.mlr-org.com/articles/learners/list_learners.html
* GitHub action to automatically generate table: https://github.com/mlr-org/mlr3extralearners/blob/main/.github/workflows/update_learner_table.yml

## related code in sktime:
```python
from sktime.utils import all_estimators
all_estimators(estimator_types="classifier")
```
https://github.com/alan-turing-institute/sktime/blob/ee7a06843a44f4aaec7582d847e36073a9ab0566/sktime/utils/__init__.py#L16
## questions:
* integrate it with API reference: https://www.sktime.org/en/latest/api_reference.html (adding columns to that table)?

## notes:
* make clear that not all algorithm are reference implementations, even though we link them with the original research paper, some may be re-implemented, adapted or interfaced from other packages
* encourage people to ping algorithm maintainers directly when they have questions about algorithm using the GitHub user name
* Adding col for maintainer using GH

mloning:
Related issue: https://github.com/alan-turing-institute/sktime/issues/354

fkiraly:
@mloning, would you mind assigning to whoever is currently looking at this?

mloning:
@fkiraly still the same people, @afzal442, Abdul and @sparkingdark

fkiraly:
I mean, assigning to the issue

mloning:
@fkiraly I think they need to comment on the issue first before they can be assigned for some reason.

afzal442:
Thanks @fkiraly.
Hi @mloning , since we have been working on this issue, you can assign it to us as @abdulelahsm, @afzal442 , @sparkingdark. Thanks.

[...]

PR #930, opened 2021-06-05
Added Automated learning algorithm for estimator overview

afzal442:
#### Reference Issues/PRs

Fixes: #704 

Implemented Automated learning algorithm with HTML code and sphinx generated docs
No

##### For all contributions

##### For new estimators


mloning:
A questions about deployment: Do we want the generated overview in the repo or generate it on the fly during the docs generation?

afzal442:
> Do we want the generated overview in the repo or generate it on the fly during the docs generation?

We want `est_overview.html `file in the repo which is required to be added/included on the .rst file. For another file `algorithm_overview.html`, it will be auto-generated by running `algo_list.py`. 

Since sphinx doesn't provide any feature to reproduce those files in the /built folder, I think, we have to keep those files also with` algo_list.py`. 
Once that .py file is run, it will update algorithm_overview.html. And using .rst file we can see the rendered .HTML file on the go.

est_overview.html
```diff
@@ -0,0 +1,127 @@
...
+</head>
+
+<body>
+  <center>
+    <input type="text" id="myInput" placeholder="Search for names..">
+  </center>
+  <br>
+  <br>
+
+  <div w3-include-html="../../algorithm_overview.html"></div>
```
afzal442, line 78: Fortunately, I made it. I just figured it out. TTG!

mloning:
As far as I understand, it is possible to run other commands/files before running sphinx. See for example: 
https://github.com/alan-turing-institute/sktime/blob/50765a6a5bf3a78096dcca10223b3320cf2aa5f0/docs/Makefile#L25-L26
But I'm not sure if this is best practice.

docs/source/estimator_overview.rst
```diff
@@ -0,0 +1,16 @@
...
+Estimator Overview
+==================
+
+Welcome to sktime's software development plan.
+
+.. note::
+
+    The estimator overview is under development. We have created a basic
+    structure and are looking for further improvements to develop the guide
+    better which can include status, references, etc.
```
mloning, line 13: We can remove these lines here

mloning (commented):
Thanks again for the call today, a few quick comments on the PR, I'll follow up on the deployment questions. 

We could also think about adding a short paragraph about the overview here: https://github.com/alan-turing-institute/sktime/blob/main/CONTRIBUTING.md#documentation

docs/source/estimator_overview.rst
afzal442: Yes fine, we can. Or we can edit it by adding something in the note section that might be good for reflecting upon what it does.

mloning:
Just a quick follow up: 

These lines are actually not executed on readthedocs, I thought they were, but turns out they're not. 
https://github.com/alan-turing-institute/sktime/blob/50765a6a5bf3a78096dcca10223b3320cf2aa5f0/docs/Makefile#L25-L26

It does seem to be possible to add things to the sphinx build through the `conf.py` file (https://github.com/readthedocs/readthedocs.org/issues/2276)

afzal442:
https://github.com/scikit-learn/scikit-learn/blob/5f32f8b846382a595aa97f4227e8673aa640840e/doc/conf.py#L467

afzal442:
@mloning let me know if we add this line `python3 source/algo_list.py` at line no. 28 to Makefile of docs folder so we can get the updated result once we run the command `make docs` or `make html`.
https://github.com/alan-turing-institute/sktime/blob/50765a6a5bf3a78096dcca10223b3320cf2aa5f0/docs/Makefile#L24-L29

mloning:
@afzal442 yes we can add it there, the problem is that this is not used when it is deployed on readthedocs. They use their own commands and do not allow to run custom scripts.

afzal442:
Thanks. Should I try out using conf.py file? 
Ref: https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx-core-events

```
def setup(app):
  app.connect('______',algo_list.py )
```

afzal442:
What I propose is I can define a method (`algo_list.py`) namely **generate_estimator_overview_table** inside the code of `conf.py`

Plz have a look here below:

```python
def setup(app):
    app.connect('builder-inited', generate_estimator_overview_table)
    def adds(pth):
        print("Adding stylesheet: %s" % pth)
        ..........................................

```
something like this. Markus plz tell me if we can do it this way.

mloning:
Hi @afzal442, I'm also new to this way of using sphinx. So I suggest you try it out and see what happens. We will have a section on docs during the dev days next week, perhaps we can fix it there!

fkiraly:
quick question, who is this an action on? We're at the dev days but this has not come up yet.

mloning:
@fkiraly this is part of the docs work stream, I wanted to bring this up with @RNKuhns. The main challenge is that it's not straightforward to automate in the current setup with Readthedocs.

afzal442:
Apologise for that. Could we merge it for the sake of simplicity for now? And make a follow up/ subsequent PR if something comes up/solves that left part later.

fkiraly:
I think we need to have a meeting - I'm a little out of my depth with the `html` stuff, but what I would need to precisely understand how the update logic works. I.e., when is what generated based on what, and how (in-code)?

Could we schedule sth in the next weeks?

afzal442:
> Could we schedule sth in the next weeks?

Thanks @fkiraly. It would be great

> Once that `algo_list.py `file is run, it will update `algorithm_overview.html` file. And using `estimator_overview.rst` file we can see the rendered `est_overview.html` file on the fly.

fkiraly:
>Thanks @fkiraly. It would be great

Let's connect on slack with @mloning.

afzal442:
As far as I understood from other sources this above works on a **live server**, so for now, I have put those required HTML files into my own repo to check if it works on my `github.io` server which is this https://afzal442.github.io/. Plz, click on that. Thanks.

Thus it seems maybe it can work on **https://www.sktime.org** `/en/latest/estimator_overview.html` server or it can be exceptional not working here `https://sktime--930.org.readthedocs.build/en/930/estimator_overview.html`. 

I don't know if it is good practice or not but I need you to give me a favor that you can merge this PR and see if it works or not. If not then you can revert it and we will be happy to review that alternative case suggested by you.

mloning:
Hi @afzal442 if it doesn't work on https://sktime--930.org.readthedocs.build/en/930/ now it won't work after we merge it as far as I understand. The build/deployment process is exactly the same as far as I'm aware, with the only difference that another branch is checked out.  

Note that the search bar still doesn't seem to work on your demo (https://afzal442.github.io). 

I'd prefer to work with the static version first as implemented in #1138 and as suggested at the very beginning. I understand that it's a shame that some of your work may not end up being merged, but it's already a huge contribution to add the static table. There are also other things to improve on the static version first (e.g. adding links to publications, linking authors with their GitHub profiles, etc). Moving away from readthedocs eventually is an option, but does involve a lot of extra work.

afzal442:
Thanks, Markus, that really sounds pretty good. Makes sense now. 
> the search bar still doesn't seem to work on your demo

Yes, I saw that after your notice. Sorry for that.

> There are also other things to improve on the static version

To handle links to publications I think we have to do it manually creating an empty extra column produced by that table likewise for others.

> Moving away from readthedocs eventually is an option, but does involve a lot of extra work.

Really weird, readthedocs does that matter on top of that.

mloning:
Closed in favour of #1138

PR #896, opened 2021-05-21
Add STL forecaster

thayeylolu:
#### What does this implement/fix? Explain your changes.
This fix implements the STL forecaster

#### Does your contribution introduce a new dependency? If yes, which one?
No.

##### For all contributions
- [ ] I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/main/.all-contributorsrc).
- [ ] Optionally, I've updated sktime's [CODEOWNERS](https://github.com/alan-turing-institute/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
- [ ] I've added unit tests and made sure they pass locally.

##### For new estimators
- [ ] I've added the estimator to the online documentation.
- [ ] I've updated the existing example notebooks or provided a new one to showcase how my estimator works.

Issue #782, opened 2021-04-07
Implement STL forecaster


mloning:
**Is your feature request related to a problem? Please describe.**
STL forecasting seems a common composite forecaster, so it will be useful to provide a class for it. 

There is a [STLForecaster](https://www.statsmodels.org/v0.12.0/generated/statsmodels.tsa.forecasting.stl.STLForecast.html#statsmodels.tsa.forecasting.stl.STLForecast) in statsmodels.

**Describe the solution you'd like**
It should be a composite forecaster consisting of a TransformedTargetPipeline which accepts any other forecaster and applies the forecaster to deseasonalized and detrended data.  

```python
forecaster = STLForecaster(MyForecaster(...), ...)
```

This should be equivalent to this pipeline:
```python
forecaster = TransformedTargetForecaster(
    [
        ("deseasonalise", Deseasonalizer(...)),
        ("detrend", Detrender(forecaster=PolynomialTrendForecaster(degree=1))),
        ("forecast", MyForecaster(...)),
    ]
)
```

fkiraly:
Interesting!

There is an obvious "easy" way to do this - but...

I was wondering, is it possible in python to create a contraction operation that is class-level, i.e., a factory operation that uses class inspection and manipulation? Sth that allows

```python
class STLForecaster = EstimatorContract(forecaster, 
        fixed_params = {'detrend__degree' : 1},
        mapped_params = {'foo' : 'bar'}
    )
```

smriti111:
Can I work on this issue, if it hasn't been taken up by anyone.

mloning:
@smriti111 yes no-one is working on this as far as I'm aware. Creating a draft PR as soon as possible is usually a good idea to let others know you've started working on this and to get early feedback.

smriti111:
Sure am working on it, will update soon

thayeylolu:
Hello @smriti111 . Have you started working on this ?

thayeylolu:
Hello @mloning . @smriti111 has not responded and I would like to go ahead with a draft

mloning:
Great @thayeylolu - feel free to go ahead!

thayeylolu:
I am on it

thayeylolu:
> **Is your feature request related to a problem? Please describe.**
> STL forecasting seems a common composite forecaster, so it will be useful to provide a class for it.
> 
[...]

PR #933, opened 2021-06-08
Implement Stlforecaster

thayeylolu:
##### For all contributions

##### For new estimators


thayeylolu:
Hello @fkiraly and @mloning . I am having this error `ValueError: `index` must contain at least some values, but found empty index: DatetimeIndex([], dtype='datetime64[ns]', freq='D')` . I don't have a way around it. kindly assist

fkiraly:
one of the tests is failing - @thayeylolu has described it in #782.
The gist is, it seems to be caused by the `TransformedTargetForecaster`?

fkiraly:
this seems like a duplication of #896.
If yes, can you please close one of the two PR?

thayeylolu:
> this seems like a duplication of #896.
> If yes, can you please close one of the two PR?
> Perhaps the older one.

Yes. I would close the older duplicate

sktime/forecasting/compose/_loess.py
```diff
@@ -0,0 +1,66 @@
...
+    >>> y = load_airline()
+    >>> y_train, y_test = temporal_train_test_split(y)
+    >>> fh = np.arange(len(y_test)) + 1
+    >>> forecaster = STLForecaster(estimator, degree, sp)
+    >>> forecaster.fit(y_train, fh)
+    >>> y_pred = forecaster.predict()
+    """
+
+    _tags = {"univariate-only": True}
+    _required_parameters = ["forecaster", "degree", "sp"]
```
mloning, line 49: Note that these parameters are not required. By "required parameters" we mean those parameters for which there are no default values in the constructor meaning that we have to set them before doing automated unit testing on them. But here we have default values for all parameters in the constructor.

sktime/forecasting/compose/_loess.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def __init__(self, forecaster=None, degree=1, sp=1):
+        if forecaster is None:
```
mloning, line 52: I'm not sure if we should do this kind of logic in the constructor, typically this happens in `fit`.

sktime/forecasting/compose/tests/test_loess.py
```diff
@@ -0,0 +1,58 @@
...
+    ("estimator", estimator),
+]
+transformed_target_forecaster_steps_naive = [
+    ("deseasonalise", Deseasonalizer()),
+    ("detrend", Detrender(forecaster=PolynomialTrendForecaster(degree=1))),
+    ("estimator", naive_forecaster),
+]
+ttf = TransformedTargetForecaster(transformed_target_forecaster_steps)
+ttf_naive = TransformedTargetForecaster(transformed_target_forecaster_steps_naive)
+ttf.fit(y_train, fh=fh)
```
mloning, line 41: I would move the fitting to the test functions itself

mloning (commented):
Hi @thayeylolu and @fkiraly - I left a few comments on the code.

I also had a look at the error but I'm not exactly sure where it comes from. It looks as if it tries to update a transformer, this may be something which hasn't been implemented yet. I'm not sure if it is a bug with `TransformedTargetForecaster` because this one passes the tests on `main`.  I suggest to step through it with the debugger to see what's going on. Note that the `update` implementation is quite complex but perhaps @fkiraly can guide you through it if you get stuck?

sktime/forecasting/compose/_loess.py
thayeylolu: Yes. because the stlforecaster is inheriting from the `transformed target forecaster` , it is also inheriting the `_required_parameter` which is currently set to `step`. Although, I removed the default parameters from the `_required_parameter`. But removing `_required_parameter` total throws an error

sktime/forecasting/compose/_loess.py
thayeylolu: How would you suggest I go about this?

sktime/forecasting/compose/tests/test_loess.py
thayeylolu: Done

fkiraly:
Personally, I would suggest to park this and do #982, together with a review of the `transformer` interface.

thayeylolu:
> Personally, I would suggest to park this and do #982, together with a review of the `transformer` interface.

Thank you @fkiraly. I would look at it

sktime/forecasting/compose/_pipeline.py
```diff
@@ -202,10 +202,10 @@ def _update(self, y, X=None, update_params=True):
         self : an instance of self
         """
 
-        for step_idx, name, transformer in self._iter_transformers():
```
thayeylolu, line 205: @mloning I commented this block of code out and the stlforecaster passes the tests. As well as the transformedTargetForecaster. 
My understanding of this block of code is : 
we want to update the transformers if they have update methods. I want to know why this is necessary. Please, let me know if I got this wrong. thank you

fkiraly:
>Thank you @fkiraly. I would look at it

Actually, it would seem that the tests now pass - did the refactoring fix the issue?
Can you check by running your tests locally and see whether the problem re-appears?

I've updated the remote branch with main so you have to pull first.

sktime/forecasting/compose/_pipeline.py
fkiraly: this might be important to keep - logically, the transformers should update.

What this indicates though is that we may want to look closely at the transformers and do a similar refactor.

thayeylolu:
> Personally, I would suggest to park this and do #982, together with a review of the `transformer` interface.

HI @fkiraly is there an open issue for `transformer`?

fkiraly:
>HI @fkiraly is there an open issue for transformer?

I think this is a major refactor - we should discuss next week.

sktime/forecasting/compose/_loess.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+
+    _tags = {"univariate-only": True}
+    _required_parameters = ["forecaster"]
+
+    def __init__(self, forecaster=None, degree=1, sp=1):
+        self.forecaster = forecaster
+        self.sp = sp
+        self.degree = degree
+        self.detrender = Detrender(
```
mloning, line 42: I think we should wrap up the settings of defaults in a function and call it at the end of init but also at the beginning of fit.

aiwalter:
@thayeylolu is this work going to be continued?

mloning:
Superseded by #1696

PR #941, opened 2021-06-10
minor bugfix - setting _is_fitted to False before input checks in forecasters

fkiraly:
This is a minuscule change which is a preemptive fix.

The change sets `self._is_fitted = False` in the first line of `fit`, of a number of forecasters, before the input checks.

This prevents some unexpected behaviour when `fit`, of the estimator, is called multiple times in sequence, e.g., in grid search (as in the tutorial notebook).

It also prevents exceptions caused by this unexpected behaviour in any future refactor (e.g., #912) where `_set_fh` uses `_is_fitted` to check whether an `fh` is expected.

This should be done in all forecasters.

### DIFF
--- a/sktime/forecasting/base/adapters/_fbprophet.py
+++ b/sktime/forecasting/base/adapters/_fbprophet.py
@@ -32,6 +32,7 @@ class _ProphetAdapter(_OptionalForecastingHorizonMixin, _SktimeForecaster):
         -------
         self : returns an instance of self.
         """
+        self._is_fitted = False
         self._instantiate_model()
         self._check_changepoints()
         self._set_y_X(y, X, enforce_index_type=pd.DatetimeIndex)
--- a/sktime/forecasting/base/adapters/_pmdarima.py
+++ b/sktime/forecasting/base/adapters/_pmdarima.py
@@ -37,6 +37,7 @@ class _PmdArimaAdapter(_OptionalForecastingHorizonMixin, _SktimeForecaster):
         -------
         self : returns an instance of self.
         """
+        self._is_fitted = False
         self._set_y_X(y, X)
         self._set_fh(fh)
         self._forecaster = self._instantiate_model()
--- a/sktime/forecasting/base/adapters/_statsmodels.py
+++ b/sktime/forecasting/base/adapters/_statsmodels.py
@@ -41,7 +41,7 @@ class _StatsModelsAdapter(_OptionalForecastingHorizonMixin, _SktimeForecaster):
         # so we coerce them here to pd.RangeIndex
         if isinstance(y, pd.Series) and type(y.index) == pd.Int64Index:
             y, X = _coerce_int_to_range_index(y, X)
-
+        self._is_fitted = False
         self._set_y_X(y, X)
         self._set_fh(fh)
         self._fit_forecaster(y, X)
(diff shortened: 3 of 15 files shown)


mloning:
Could you explain why this is needed for forecasters but not other estimator types? Is this related to the setting of the forecasting horizon?

fkiraly:
>Is this related to the setting of the forecasting horizon? 

Correct. It is to allow `_set_fh`, which gets `self` and `fh` as arguments, to distinguish a situation in which `fit` has been called twice from a situation in which `fit` and then `predict` has been called; in both situations, the method calling `_set_fh` internally.

>If that's the case, an alternative may be to pass the name of the caller method to the set_fh function, e.g. set_fh(fh, "fit").

Yes, but it feels less python zen, and less consistent with the `checkxyz` functions that also exhibit behaviour based on the object state.

fkiraly:
do you want to look at this separately and/or accept this in prep for #912, @mloning, @aiwalter?

mloning (commented):
I've thought about this a bit more and I prefer the option where we pass the name of the calling method to `_set_fh`, e.g. `_set_fh(fh, method="fit")`. For me, the main reasons are:
* it keeps the functionality for checking/setting the `fh` contained to that function and does not assume state changes elsewhere in the code, like at the beginning of `fit`,
* it keeps it closer to the implementation of other estimator types where we do not re-set the state at the beginning of `fit`,
* it avoids the redundancy of having to set `_is_fitted=False` in the constructor and at the beginning of `fit`.

What do you think, @fkiraly?

fkiraly:
>What do you think, @fkiraly?

Ok, happy to consider this as an endstate.

But, practically, goign there would be much more more invasive than the current proposed change, since it deviates from the current signature (which you introduced).
Changing it would require a lot of manipulations in descendants - in every instance where `_set_fh` is called, it would have to change. Where, under the current change, only a line has to be added, and no logic needs (additional) change.

So, I would strongly recommend *first* carrying out this minimally invasive change, then the inheritance refactor, and then any well-isolated changes to the setters and checkers.
Note that after completing the interface refactor, this will have to be done in only a single location, namely the `BaseForecaster`.

To reiterate my suggestion: let's get with minimal invasive effort to the position where we can just refactor `BaseForecaster` and don't need to change every descendant class. Then, I'm absolutely happy to work on `_set_fh` and/or with other interfaces changes.

mloning (approved):
The change in the logic of `_set_fh` seem to be separate from the interface refactor, no? This PR was still easy to review for me, so adding a logic change to `_set_fh` could be included. In terms of descendants, we would still only have to change one line.  Doing the change here would avoid having intermediate code on `main`. 

But happy to follow your sequence of changes.

fkiraly:
>This PR was still easy to review for me, so adding a logic change to _set_fh could be included. In terms of descendants, we would still only have to change one line. 

That's not entirely true sine `_set_fh` appears in multiple places in descendants. I'm only adding a line to `fit`, and not changing existing lines or existing interfaces. But agreed with the rationale.

fkiraly:
so... @mloning - this needs a second review? Or are we find to merge?

mloning:
I think it's fine to merge this as an intermediate solution.

mloning:
@fkiraly FYI I tend to use "squash and merge" for merging into `main`, that keeps `main` a bit more organised I think.

PR #942, opened 2021-06-10
Feature/information criteria get_fitted_params

### LINKED ISSUE #931: Exposing AIC, BIC and HQIC from sktime models via get_fitted_params
After fitting a time series model (i.e. ARIMA, AutoETS...etc), there is no way to retrieve individual fitting scores (i.e. AIC, BIC, HQIC..etc) for evaluation.

The following key does not exists
model.get_fitted_params()['aic']
model.get_fitted_params()['bic']
model.get_fitted_params()['hqic']

ltsaprounis:
#### Reference Issues/PRs

Fixes #931

Adding the available information criteria (AIC, AICc, BIC, HQIC) for various forecasting algorithms.
TBATS and BATS did not have a get_fitted_params method so I added it in the adapter there. Maybe we'll want to add additional parameters returned there. 

No

Is get_fitted_params the right place to return those? maybe a similar function e.g. get_fit_results is more appropriate.

##### For all contributions


fkiraly (approved):
Nice! Looks good from my side. Must have been fiddly to find where the parameters are...

fkiraly:
Regarding this question:

> Is get_fitted_params the right place to return those? maybe a similar function e.g. get_fit_results is more appropriate.

A very valid point - do we want to separate fitted parameters from things like summaries or inspection?

The MLJ package, where I've co-designed the interfaces, does separate this, and I think it's a good idea - see
https://alan-turing-institute.github.io/MLJ.jl/stable/machines/#Inspecting-machines

During the dev days later this month, we'll probably refactor the interfaces, and this may be one thing we add - feel free to join.
I've also opened an issue #944 so we don't forget.

fkiraly:
@mloning, have you had a look at this?
This affects the relatively new `get_fitted_params` interface, would like to have a 2nd opinion.

mloning:
I agree @fkiraly, fitted parameters are different from summary statistics of the model, so suggest to create a separate interface for them.

ltsaprounis:
Thanks @mloning and @fkiraly. I think this is the right call, it felt wrong to put AIC etc. in the same place as the fitted model parameters. Do you think I should keep going with this PR and implement the function to get summary statistics here? or is it more appropriate to do it in a separate PR (maybe linked to #944 ).

mloning:
I suggest to work on the interface first in a separate PR, before we start adding it to concrete implementations.

fkiraly:
hm, @mloning, I would suggest we merge this and then perhaps later move it to the other interface - or this nice contribution may be buried under refactors and commits on top?

Rationale being, moving it from `get_fitted_params` to `get_report_params` or similar is almost trivial once we have the interface point, while finding the PR again, updating it, and remembering to do this later may not be easy.

mloning (approved):

fkiraly:
2 approvals, so can be merged now.

fkiraly:
@ltsaprounis, I think @mloning and I have aligned that we'll merge this and then later re-factor along #971.

Would you like to merge?
Minor suggestion, please add the names of the classes you've changed in the squashed commit message.

fkiraly:
merged due to dependencies and 2 approvals

ltsaprounis:
Sorry @fkiraly I was on holidays for the past few days and nowhere near my laptop. Just realised Github has an "out of office" option...

Looks good to me, thanks for merging!

Issue #935, opened 2021-06-09
Implement Series-to-Series Transformer to difference a time series


RNKuhns:
**Is your feature request related to a problem? Please describe.**
Sktime currently does not include a stand alone transformer to apply a set of differences to a time series. Differencing is common in time series forecasting (to among other things ensure a time series is stationary before using a model that assumes stationarity). 

This feature would be useful in forecasting pipelines (e.g. `TransformedTargetForecaster()`) and also useful in the work on the `Theta` and/or `AutoTheta` forecasters (eg. #922).

**Describe the solution you'd like**
A `Differencer()` should be a series-to-series transformer that takes in an input timeseries and transforms it to apply a set of differences. 

An important feature is that the `Differencer()` should be able to apply multiple differences at different lags sequentially (as a list, tuple or array in the constructor). This allows users to easily use the `Differencer()` for use cases like iteratively applying a first-difference and seasonal-difference to a series. 

Some implementations such as in [pmdarima](https://alkaline-ml.com/pmdarima/_modules/pmdarima/utils/array.html#diff) only allow users to specify higher-order differences, but restrict the iterative differences to use the same lag. 

The `Differencer()` should also implement an `inverse_transform()` method to reverse the transformations. 

**Describe alternatives you've considered**
One difficulty in implementing the `Differencer()` is the slightly different approaches to reversing the transformation on "in-sample" data and predictions. 

@mloning and I have discussed two potential solutions (as listed below). Interested to hear everyone's thoughts on these options and if anyone has other ideas on the different inverse_transform application with predictions versus training data.

One option that is being considered is to have a constructor parameter that indicates how the `Differencer` will be used. 

```python
class Differencer(_SeriesToSeriesTransformer):

    def __init__(self, lags = 1, use_with_predict=False):
        self.lags = lags
        self.use_with_predict=use_with_predict
        ...
       
    def transform(...):
        # Code to do transformation and store info to do inverse transform
        ...
        
    def inverse_transform(...):
        # Uses self.use_with_predict to diffentiate inverse_transform case
        if self.use_with_predict:
            ...
        else:
            ...
```
The other option is to create a `Differencer` that applies the transformation to a forecaster.
```python
class Differencer(BaseForecaster, _SeriesToSeriesTransformer):

    def __init__(self, forecaster = None, lags = 1):
        self.forecaster = forecaster
        self.lags = lags
        ...

    def fit(...):
         y_transform = self.transform(...)
[...]

PR #945, opened 2021-06-11
Create Differencer Transformation

RNKuhns:
#### Reference Issues/PRs

This implements the functionality described in #935.

Adds a series-to-series transformer class that lets users specify one or more differences to iteratively apply to an input time series.

No.

- Did I follow the API design correctly (leave base `fit` method as is)?
- Potential solution for not using `use_with_predict` hyperparameter

#### Any other comments?

Once we have this in a good place, I can update the forecasting notebook to include at least one example where the difference transformer is used (and ideally improves forecasting accuracy). 

##### For all contributions

##### For new estimators


RNKuhns:
@aiwalter, @fkiraly and @mloning I've got my initial crack at a difference transformer completed. I've got my implementation working with some unit tests to verify that the inverse transformation reconstructs the input data correctly, and also can convert predictions into the undifferenced unit of the original series. 

That being said, I am hoping there is a better way for the `inverse_transform` method to work so that we don't need the nuisance `use_with_predict` parameter. 

In terms of the inverse, I can think of 4 different use-cases:

1. Apply inverse transformation to entire input time series
2. Apply inverse transformation to a slice of the input time series that includes the start of the series (e.g. `series[0:some_number]`)
3. Apply inverse transformation to some slice of the input series that doesn't include the start of the series
4. Apply inverse transformation to out-of-sample predictions

Right now, the `Differencer` should return correct results for cases 1, 2 and 4. Case 3 is difficult. To reverse a difference at lag `m` you need the `m` observations prior to the slice. Its easy enough to store the first `m` observations or last `m` observations to handle cases 1, 2 and 4. But I can't think of a way to handle case 3 other than storing the entire input series. 

Use-case 3 applies to in-sample predictions, so it would be good to handle it. What are the thoughts on the differencer storing the entire input series (instead of just the first `m` and last `m` observations currently)?

RNKuhns:
@fkiraly I'd like to add fractional differencing to this, but want to get the initial functionality working first.

RNKuhns:
@mloning, @fkiraly or @aiwalter any ideas on the test case I keep failing. 

The failing test looks to be related to `inverse_transform` changing the tranformers `__dict__`. 

I have tried to see if `inverse_transform` is doing this locally and haven't seen an issue. But I also might not be fully understanding what the check is doing. I'm under the assumption it is doing something like below to verify that `inverse_transform` doesn't update the `transformer.__dict__`. 

```python

y = load_airline()
transformer = Differencer(lags=np.array([1]))
y_hat = transformer.fit_transform(y)

p = transformer.__dict__.copy()

y_reconstructed = transformer.inverse_transform(y_hat)

p2 = transformer.__dict__

p == p2
```

fkiraly:
>@fkiraly I'd like to add fractional differencing to this, but want to get the initial functionality working first.

Very sensible, I agree with you. Let's not over-scope at the start.

fkiraly:
>In terms of the inverse, I can think of 4 different use-cases:

I think indeed 1,2,4 are the most common and can be handled in the way you say.
When would 3 happen? Seems a bit artificial. We could simply return a warning saying that `inverse_transform` is valid only if certain conditions are met? Or an informative error message?

RNKuhns:
@fkiraly the only way I see case 3 applying is in-sample predictions. Then you'd need to inverse transform a slice of the original series. 

To do that I can't think of a way other than storing the entire input series (instead of just the first and last observations). 

What are your thoughts on having the `Differencer` store the entire input series in a `_y` attribute?

If I do that then we need logic that is something to the effect of:
1. If the series to be inverse transform has same exact index as `_y` then return `_y` 
2. Otherwise, get the last `n` observations from `_y` whose index comes prior to  the first index of the series to be inverse transformed

That would work for all 4 cases and allow us to get rid of the `use_with_predict`.

mloning (commented):
Thanks @RNKuhns, I left a few comments on the code. 

Perhaps it's easiest if we simply store the entire series and raise an error if there is no prior data available for the data passed to `inverse_transform`. The storing would have to happen in `fit`. 

Alternatively, there is the `fit-in-transform` for handling some transformers that do some fitting during `transform` but I think it's preferable to use `fit` in this case.

sktime/transformations/series/difference.py
```diff
@@ -0,0 +1,168 @@
...
+import numpy as np
+import pandas as pd
+from sklearn.utils import check_array
+
+from sktime.transformations.base import _SeriesToSeriesTransformer
+from sktime.utils.validation import is_int
+from sktime.utils.validation.series import check_series
+
+
+def _check_differencer_lags(lags):
```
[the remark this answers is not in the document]
RNKuhns: I've made this change.

sktime/transformations/series/difference.py
```diff
@@ -0,0 +1,168 @@
...
+    """
+
+    _tags = {
+        "fit-in-transform": True,
+        "transform-returns-same-time-index": True,
+        "univariate-only": True,
+    }
+
+    def __init__(self, lags=1, use_with_predict=False):
+        self.lags = _check_differencer_lags(lags)
```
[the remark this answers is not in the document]
RNKuhns: I've moved the check to fit.

sktime/transformations/series/difference.py
```diff
@@ -0,0 +1,168 @@
...
+        self.use_with_predict = use_with_predict
+        self._cumalative_lags = self.lags.cumsum()
```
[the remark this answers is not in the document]
RNKuhns: I've initialized the attributes as None in the constructor and I'm initializing the arrays in fit now.

RNKuhns:
@mloning I've updated the Differencer code. Note that this is failing a test in `test_all_transformers` because `inverse_transform` doesn't match original series. It reverses the transformation, but by default the `transform` method removes the first `n` observations that are missing due to the differencing; therefore, `inverse_transform`  correctly reverses the transformation, but doesn't have the first `n` observations (as expected).

Note that the Differencer has the tag "transform-returns-same-time-index": False.  When that tag is false, the tests I've copied from `test_all_transformers.py` might not apply. 

```python
def check_transform_returns_same_time_index(Estimator):
    assert issubclass(Estimator, _SeriesToSeriesTransformer)
    fit_args = _make_args(estimator, "fit")
    estimator.fit(*fit_args)
    for method in ["transform", "inverse_transform"]:
        if hasattr(estimator, method):
            X = _make_args(estimator, method)[0]
            Xt = estimator.transform(X)
            np.testing.assert_array_equal(X.index, Xt.index)


def check_transform_inverse_transform_equivalent(Estimator):
    X = _make_args(estimator, "fit")[0]
    Xt = estimator.fit_transform(X)
    Xit = estimator.inverse_transform(Xt)
    _assert_array_almost_equal(X, Xit)

```

sktime/transformations/series/difference.py
```diff
@@ -0,0 +1,317 @@
...
+
+        is_contained_by_fitted_z = False
+        is_future = False
+
+        if first_idx < orig_first_idx:
+            msg = [
+                "Some indices of `Z` are prior to timeseries used in `fit`.",
+                "Reconstruction via `inverse_transform` is not possible.",
+            ]
+            raise ValueError(" ".join(msg))
```
mloning, line 118: What's the different compared to passing the `msg` directly?
```suggestion
            raise ValueError(msg)
```

sktime/transformations/series/difference.py
```diff
@@ -0,0 +1,317 @@
...
+            is_contained_by_fitted_z = True
+
+        elif first_idx > orig_last_idx:
+            is_future = True
+
+        pad_z_inv = self.remove_missing or is_future
+
+        cutoff = Z.index[0] if pad_z_inv else Z.index[self._cumulative_lags[-1]]
+        fh = ForecastingHorizon(
+            np.array([*range(-1, -(self._cumulative_lags[-1] + 1), -1)])
```
mloning, line 130: Would that work?
```suggestion
            np.arange(-1, -(self._cumulative_lags[-1] + 1), -1)
```

sktime/transformations/series/difference.py
```diff
@@ -0,0 +1,317 @@
...
+        )
+
+        index = fh.to_absolute(cutoff).to_pandas()
+        index_diff = index.difference(self._Z.index)
+
+        if index_diff.shape[0] != 0:
```
mloning, line 136: in which case would that happen?

sktime/transformations/series/difference.py
```diff
@@ -0,0 +1,317 @@
...
+
+        lags : np.ndarray
+            Lags to be used in applying differences.
+
+        Returns
+        -------
+        diff :
+            Differenced series.
+        """
+        diff = Z.copy()
```
mloning, line 185: for readability, so that we don't have `diff.diff(...)`
```suggestion
        Zt = Z.copy()
```

sktime/transformations/series/difference.py
```diff
@@ -0,0 +1,317 @@
...
+
+        if len(lags) == 0:
+            return diff
+
+        else:
+            for lag in lags:
+                diff = diff.diff(lag)
```
mloning, line 192: what's the scenario where you want to apply different lags iteratively?

mloning (commented):
Thanks @RNKuhns - do you want to skip those tests based on the tags by adding an if condition in the test function you posted above? I think that would make sense.

RNKuhns (commented):
> Thanks @RNKuhns - do you want to skip those tests based on the tags by adding an if condition in the test function you posted above? I think that would make sense.

Yep, that would be my proposed solution.

sktime/transformations/series/difference.py
```diff
@@ -0,0 +1,168 @@
...
+    >>> from sktime.transformations.series.difference import Differencer
+    >>> from sktime.datasets import load_airline
+    >>> y = load_airline()
+    >>> transformer = Differencer()
+    >>> y_hat = transformer.fit_transform(y)
+    """
+
+    _tags = {
+        "fit-in-transform": True,
+        "transform-returns-same-time-index": True,
```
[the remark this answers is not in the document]
RNKuhns: These have been updated now to resolve this.

sktime/transformations/series/difference.py
```diff
@@ -0,0 +1,168 @@
...
+
+        Returns
+        -------
+        diff :
+            Differenced series.
+        """
+        diff = series.copy()
+        for i, lag in enumerate(self.lags):
+            prior_cum_lag = self._prior_cum_lags[i]
+            self._lag_fit_info.append(_difference_info(diff, lag, prior_cum_lag))
```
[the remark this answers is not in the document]
RNKuhns: I had been using the fit-in-transform tag, but like we talked about I'm now storing info in fit.

sktime/transformations/series/difference.py
RNKuhns: That makes sense to me. I've updated.

sktime/transformations/series/difference.py
RNKuhns: Most use cases just need a single difference, but there are definitely cases where more than one difference (and at different lags) may be needed. 

The classic scenario is a series with trend and seasonality. In that case you would consider a first difference to help deal with the trend and a seasonal difference to deal with the seasonal dependence. 

The other case is if something is non-seasonal, but integrated with an order greater than one. That case you more than likely are going to apply iterative differences at the same lag (e.g. lags = [1, 1] for 2nd order). 

Some other implementations allow the flexibility to specify a difference at a single lag, but also specify an order parameter to account for the second case. But then you'd need to have 2 `Differencer's` in some pipelines to deal with the first case. 

Implementing it this way lets a user accomplish both cases with a single `Differencer` in their pipeline, which I think is nice.

I'll a note on these use cases in the docstring.

RNKuhns:
@mloning I also realized there is another test not passing because of the `"transform-returns-same-time-index"` issue. I'm going to add the logic to the transformer tests that will handle them, but I'm not sure how to proceed on this one. 

The test `check_fit_idempotent` is actually not getting to the part that checks that the fit is actually `idempotent` because when it is first looping through the `NON_STATE_CHANGING_METHODS` it is passing a timeseries to inverse transform that has the same indices as the fitted data. But `inverse_transform` is assuming that a series that has been differenced (with lags specified in constructor). To perform the `inverse_transform` on a series that has been differenced you would need stored values prior to the series prior to the indices passed in `inverse_transform`. 

I think an example makes a bit easier to follow: 
If `check_fit_idempotent` is constructing a series starting from January 1, 2000. This series is passed in `fit` and then `transform`. Assuming `lags=1` was specified in the constructor (this generally holds but a specific example makes it easier to talk about), then `transform` will return the first-differenced series with indices starting on January 2, 2000. 

`inverse_transfrom` is expecting a first-differenced series will be passed and reconstruct the underlying series. `check_fit_idempotent` is passing a series with indices starting on January 1, 2000 to `inverse_transfrom` which is assumed to be a series of first-differenced values based on the prior `fit`. To correctly reconstruct that series you'd need to know the value on December 31, 1999 (because the value on January 1, 2000 is assumed to be a first difference) -- hence, the error being raised by the `inverse_transform` logic.

sktime/transformations/series/difference.py
RNKuhns: In the step above, `ForecastingHorizon` to calculate, based on the lags used in the differencing, the index values prior to the series passed to `inverse_transform` that would be needed to reconstruct the series. For example, a first-difference would need 1 prior indices, a difference at lag 12 (seasonal difference on monthly data) would need 12 prior indices, etc. 

`index_diff` is any indices that are in the index of indices needed for the reconstruction but not not present in the fitted data (`self._Z`). If those are not there then it is not possible to do the reconstruction.

RNKuhns:
I ended up adding code related to the "transform-returns-same-time-index" tag to the following transformer tests:
check_transform_returns_same_time_index
check_transform_inverse_transform_equivalent
check_series_to_series_transform_multivariate

RNKuhns:
Merged main into the branch.

sktime/transformations/series/difference.py
```diff
@@ -0,0 +1,315 @@
...
+    >>> y_transform = transformer.fit_transform(y)
+    """
+
+    _tags = {
+        "fit-in-transform": False,
+        "transform-returns-same-time-index": False,
+        "univariate-only": False,
+    }
+
+    def __init__(self, lags=1, remove_missing=True):
```
mloning, line 114: in line with common usage in pandas
```suggestion
    def __init__(self, lags=1, drop_na=True):
```

mloning (commented):
Hi @RNKuhns, I just checked out the code locally and played around with it. 

I think there is a small bug in there somewhere. This one gives an error: 
```python

y = load_airline()
transformer = Differencer(remove_missing=True)
yt = transformer.fit_transform(y)
yit = transformer.inverse_transform(yt)
np.testing.assert_array_equal(y, yit)
```
When switching `remove_missing=False` it works. Could we add this as a test? I'm also not sure why this hasn't been picked up by the `check_transform_inverse_transform_equivalent` automated unit test.

sktime/transformations/series/difference.py
RNKuhns: I belive it does and I've switched the code to that.

sktime/transformations/series/difference.py
RNKuhns: I'm down for that. Will push updated version shortly.

RNKuhns:
@mloning on the unit testing that is actually the expected behavior. One of the tweaks in the PR was to the unit testing for transformers. The test `check_transform_inverse_transform_equivalent` needs to apply differently to transformers where the tag "transform-returns-same-time-index" is False. In that case we can only check that the indices that are in the original series and the inverse transformed series are equal.

RNKuhns:
@mloning I know you'll be busy with the docsprint, but this is now ready for a final look when you have a chance.

mloning (approved):
Thanks @RNKuhns - had another look, looks all good to me, so we can merge this now!

mloning:
Now finally merged @RNKuhns - sorry for the delay on this! 🎉

PR #950, opened 2021-06-12
Suggested changes to PR 912

mloning:
#### Reference Issues/PRs
#912

### DIFF
--- a/sktime/forecasting/base/_base.py
+++ b/sktime/forecasting/base/_base.py
@@ -35,6 +35,7 @@ from warnings import warn
 import numpy as np
 import pandas as pd
 
+from sktime.utils import _has_tag
 from sktime.utils.datetime import _shift
 from sktime.utils.validation.forecasting import check_X
 from sktime.utils.validation.forecasting import check_alpha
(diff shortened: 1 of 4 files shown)


mloning:
Hi @fkiraly, here are my suggested changes, let me know what you think!

fkiraly:
I think this will mess up the tags, since assignments of the `_tags` class variable will override each other and not give you the set union of dictionary elements.

I would really really like to go with the current, working workaround, and we make changes like this later on?

mloning:
@fkiraly I prefer this solution, this way, when we remove the mixins, we only have to copy the tags to the concrete forecasters,  plus we don't change the constructor. Note that tags are not overridden, that's handled in `_all_tags`.

mloning:
@fkiraly this is part of your current work around, but introduces less unnecessary intermediate code and instead uses the tags as we would use them later, no?

mloning:
So I suggest to merge this and then go ahead with the refactoring of the concrete forecasters.

fkiraly:
hm, but then, you would have to change the "tags" inspection in `_set_fh` to `_all_tags`, no?

mloning:
that's what's happening inside `_has_tag`

fkiraly:
not sure I understand what you mean, but I mean that current line 524, `optfh = self._tags["fh_in_fit"] == "optional"` would have to use `_all_tags` instead, no?

fkiraly:
oh, I see, that's already happening *inside* `_has_tag`, got it.

fkiraly (approved):
fine with this (see above)

mloning:
I changed that line in this PR, take a look at the suggested changes in this PR. They would be merged into your branch. Does that help?

fkiraly:
yes, I just overlooked the change in line 524, all fine

PR #960, opened 2021-06-14
Clustering

TonyBagnall:
First PR of clustering into main. 

This is the work of Chris Holder. For further info, see 
https://github.com/alan-turing-institute/sktime/issues/501

No new dependencies.


fkiraly:
@TonyBagnall, some remarks on the base class interface design:

* I think returning list-of-list-of integers is not a good idea in `predict`. This should be consistent with the `sklearn` clustering design. I would propose a `pd.DataFrame` or `np.array` with a single column, and the `i`-th entry is the cluster assugnment.
* I would suggest to adopt the nested design, with public facing `fit`/`predict` and core logic `_fit`/`_predict`. See the new forecasting design https://github.com/alan-turing-institute/sktime/blob/main/sktime/forecasting/base/_base.py
I feel that will help us avoid a lot of pain when people are extending the module.

On user friendliness:
* I strongly suggest to write an extension template for clustering. See the one for forecasting as an example: https://github.com/alan-turing-institute/sktime/blob/main/extension_templates/forecasting.py
* `base_types` .... ouch. I really don't like it, as per earlier discussion, but I won't be too religious about this, since this is only used for developer-sided type hints which aren't runtime artefacts. Will likely confuse people who try to extend this though, you're cutting your own flesh.
* examples should go in jupyter notebooks, and the jupyter notebooks should go in the examples folder. Not in `py` files with `if equals main` and that.

On other design choices:
* distance matrices and kernels I'd propose to make degree 2 transformers, like the `sklearn` GP kernels 😄

fkiraly (changes requested):
I wrote a few change requests above.

The ones I would like to insist on are:
* return type of `predict` and related methods
* `fit`/`_fit` design
* jupyter notebook

The others I wouldn't see as blockers.

sktime/clustering/__init__.py
```diff
@@ -0,0 +1,6 @@
+# -*- coding: utf-8 -*-
+__all__ = ["TimeSeriesKMeans", "TimeSeriesKMedoids", "TimeSeriesKShapes"]
+
+from sktime.clustering._time_series_k_means import TimeSeriesKMeans
```
mloning, line 4: I'd prefer shorter files names
```suggestion
from sktime.clustering._k_means import TimeSeriesKMeans
```

sktime/clustering/_time_series_k_means.py
```diff
@@ -0,0 +1,159 @@
...
+    BarycenterAveraging,
+    MeanAveraging,
+)
+from sktime.clustering.partitioning._time_series_k_partition import TimeSeriesKPartition
+
+
+class TimeSeriesKMeans(TimeSeriesKPartition, BaseCluster):
+    """Time Series K-Means Clusterer.
+
+    This is a work in progress.
```
mloning, line 23:
```suggestion

```

sktime/clustering/_time_series_k_means.py
```diff
@@ -0,0 +1,159 @@
...
+        self,
+        n_clusters: int = 8,
+        init_algorithm: Init_Algo = "random",
+        max_iter: int = 300,
+        verbose: bool = False,
+        metric: Metric_Parameter = "dtw",
+        averaging_algorithm: Averaging_Algo = "mean",
+        averaging_algorithm_iterations: int = 50,
+    ):
+        """Constructor for TimeSeiresKMeans clusterer
```
mloning, line 41: In Python, the convention is that the docstring for the constructor goes below the class definition above, there's usually no separate docstring for the constructor.

sktime/clustering/_time_series_k_means.py
```diff
@@ -0,0 +1,159 @@
...
+
+    __averaging_algorithm_dict: Averaging_Algo_Dict = {
+        "dba": BarycenterAveraging,
+        "mean": MeanAveraging,
+    }
+
+    def __init__(
+        self,
+        n_clusters: int = 8,
+        init_algorithm: Init_Algo = "random",
```
mloning: For types and classes can we just use camel case here and throughout the code? 
```suggestion
        init_algorithm: InitAlgo = "random",
```

sktime/clustering/_time_series_k_means.py
```diff
@@ -0,0 +1,159 @@
...
+from sktime.clustering.partitioning._time_series_k_partition import TimeSeriesKPartition
+
+
+class TimeSeriesKMeans(TimeSeriesKPartition, BaseCluster):
+    """Time Series K-Means Clusterer.
+
+    This is a work in progress.
+    """
+
+    __averaging_algorithm_dict: Averaging_Algo_Dict = {
```
mloning, line 26: double underscores are typically reserved in Python for special methods, I wouldn't use them here, single underscore indicates private methods (by convention)
```suggestion
    _averaging_algorithm_dict: Averaging_Algo_Dict = {
```

sktime/clustering/_time_series_k_means.py
```diff
@@ -0,0 +1,159 @@
...
+        verbose: bool = False,
+        metric: Metric_Parameter = "dtw",
+        averaging_algorithm: Averaging_Algo = "mean",
+        averaging_algorithm_iterations: int = 50,
+    ):
+        """Constructor for TimeSeiresKMeans clusterer
+
+        Parameters
+        ----------
+            n_clusters: int, default = 8
```
mloning, line 45: Additional indent here? 
```suggestion
        n_clusters: int, default = 8
```

sktime/clustering/_time_series_k_means.py
```diff
@@ -0,0 +1,159 @@
...
+            init_algorithm=init_algorithm,
+            max_iter=max_iter,
+            verbose=verbose,
+            metric=metric,
+        )
+        metric_str = None
+        if isinstance(metric, str):
+            metric_str = metric
+
+        if isinstance(averaging_algorithm, str):
```
mloning: I believe these input checks/conversion should happen in `fit` rather than `init`. I suggest to write a helper function for them (e.g. `_check_averaging_algorithm` and run it at the beginning of `fit`)

sktime/clustering/_time_series_k_means.py
```diff
@@ -0,0 +1,159 @@
+# -*- coding: utf-8 -*-
+"""Time Series K-Means Clusterer."""
+
+__author__ = ["Christopher Holder", "Tony Bagnall"]
+__all__ = ["TimeSeriesKMeans"]
+
+from typing import List
+from sktime.clustering.base.base_types import (
```
mloning: We need to make sure these are consistent with the existing types in `sktime/transformations/base.py` and the ones suggested in PR #964

sktime/clustering/_time_series_k_means.py
```diff
@@ -0,0 +1,159 @@
...
+                    averaging_algorithm = "mean"
+
+            self.averaging_algorithm = TimeSeriesKMeans.__averaging_algorithm_dict[
+                averaging_algorithm
+            ]
+        else:
+            self.averaging_algorithm = averaging_algorithm
+        self.averaging_algorithm_iterations = averaging_algorithm_iterations
+
+    def fit(self, X: Data_Frame) -> None:
```
mloning: Question: Should we follow the same interface as for classification as in scikit-learn, keeping `y` in the signature but ignoring it? Not sure if that's necessary for re-using some of the higher-level composition functionality. @fkiraly any idea?

mloning: We would want to accept 3d numpy arrays as they seem to be much easier to understand and work with than the nested data frame and we have all the functionality for input validation/conversions if required. 
```suggestion
    def fit(self, X: Panel) -> None:
```
where `Panel = Union[pd.DataFrame, np.ndarray]` (e.g. 3d or nested array), as in `sktime/transformations/base.py`

sktime/clustering/_time_series_k_means.py
```diff
@@ -0,0 +1,159 @@
...
+        means model on dataset X
+
+        Parameters
+        ----------
+        X: Data_Frame
+            sktime data_frame to train the model on
+        """
+        super(TimeSeriesKMeans, self).fit(X)
+
+    def predict(self, X: Data_Frame) -> List[List[int]]:
```
mloning: Lists are not good for handling data arrays in Python, numpy or pandas is the standard. 
```suggestion
    def predict(self, X: Data_Frame) -> Numpy_Array:
```

sktime/clustering/_time_series_k_means.py
```diff
@@ -0,0 +1,159 @@
...
+
+        Returns
+        -------
+        List[List[int]]
+            2d array, each sub list contains the indexes that
+            belong to that cluster
+        """
+        return super(TimeSeriesKMeans, self).predict(X)
+
+    def calculate_new_centers(self, cluster_values: Numpy_Array) -> Numpy_Array:
```
mloning, line 121: I suggest to make this method private, unless you think users would want to users to use this directly. 
```suggestion
    def _calculate_new_centers(self, cluster_values: Numpy_Array) -> Numpy_Array:
```

sktime/clustering/base/base.py
```diff
@@ -0,0 +1,198 @@
...
+
+        Returns
+        -------
+        int
+            Index position of the approximation in the series
+        """
+        raise NotImplementedError("abstract method")
+
+
+Init_Algo = Union[str, BaseClusterCenterInitializer]
```
mloning, line 194: I suggest to have a separate file for these (e.g. `sktime/clustering/base/_typing.py`) or move them to the top of the file

sktime/clustering/base/base_types.py
```diff
@@ -0,0 +1,35 @@
+# -*- coding: utf-8 -*-
```
mloning, line 1: Suggest to rename the file: `sktime/clustering/base/_typing.py`

sktime/clustering/evaluation/_plot_clustering.py
```diff
@@ -0,0 +1,51 @@
...
+
+    axes.plot(center, color="r")
+
+
+def plot_cluster_algorithm(
+    model: BaseCluster, predict_series: Numpy_Or_DF, k: int, plt, mpatches=None
+):
+    if isinstance(predict_series, pd.DataFrame):
+        predict_series = from_nested_to_2d_array(predict_series, return_numpy=True)
+    plt.figure(figsize=(5, 10))
```
mloning, line 27: is there a missing import for `matplotlib` or am I not seeing it? Note that matplotlib is a soft dependency, please check the contributing guide to make sure it's handled correctly. Let me know if you need any help with that!

sktime/clustering/partitioning/_averaging_metrics.py
```diff
@@ -0,0 +1,220 @@
...
+
+        medoid_ind = self.__approximate_medoid_index(cost_mat, delta_mat)
+        center = self.series[medoid_ind]
+
+        for _ in range(self.n_iterations):
+            center = self.__DBA_update(center, cost_mat, path_mat, delta_mat)
+
+        return center
+
+    def __approximate_medoid_index(self, cost_mat, delta_mat):
```
mloning, line 104: same as above, here and below
```suggestion
    def _approximate_medoid_index(self, cost_mat, delta_mat):
```

sktime/clustering/partitioning/_time_series_k_partition.py
```diff
@@ -0,0 +1,250 @@
...
+
+    Probably want to rerun the algorithm multiple times and take the
+    best result as the init can be so impactful on final result
+
+    Multivariate support. While I've coded this with the intention to
+    support it I doubt it will work with multivariate so need to code
+    and improve that
+    """
+
+    __metric_dict: Metric_Function_Dict = {
```
mloning, line 57: same as above, here and below
```suggestion
    _metric_dict: Metric_Function_Dict = {
```

sktime/clustering/partitioning/_time_series_k_partition.py
```diff
@@ -0,0 +1,250 @@
...
+
+    def __init__(
+        self,
+        n_clusters: int = 8,
+        init_algorithm: Init_Algo = "forgy",
+        max_iter: int = 300,
+        verbose: bool = False,
+        metric: Metric_Parameter = "dtw",
+    ):
+        """
```
mloning, line 82: see above, docstring should be below class definition

sktime/clustering/base/base.py
```diff
@@ -0,0 +1,198 @@
...
+        ----------
+        X: Numpy array or Dataframe
+            sktime data_frame or numpy array to train the model on
+
+        Returns
+        -------
+        self
+            Fitted estimator
+        """
+        self._is_fitted = False
```
mloning, line 44: I don't think we need this here. As far as I know, it was only necessary for forecasters as a temporary fix, it will be set to False during construction in the `BaseEstimator`.
```suggestion
```

sktime/clustering/partitioning/_time_series_k_partition.py
```diff
@@ -0,0 +1,250 @@
...
+            Numpy_Array containing the time series to
+            predict clusters for
+
+        Returns
+        -------
+        Numpy_Array
+            Index of the cluster each sample belongs to
+        """
+
+        if self.__centers is None:
```
mloning, line 161: This should be handled via `self.check_is_fitted()`. If you still want to raise a separate exception here, please make it more specific, e.g. `AttributeError`

sktime/clustering/tests/partitioning/test_center_initializers.py
```diff
@@ -0,0 +1,11 @@
+# -*- coding: utf-8 -*-
```
mloning, line 1: I suggest to change the folder structure here to: `sktime/clustering/partitioning/tests/test_center_initializers.py`, makes it easier to find the corresponding tests and is in line with the rest of the repo

sktime/clustering/tests/partitioning/test_cluster_averaging.py
```diff
@@ -0,0 +1,47 @@
...
+from sktime.clustering.partitioning._averaging_metrics import (
+    BarycenterAveraging,
+    MeanAveraging,
+)
+from sktime.utils.data_io import load_from_tsfile_to_dataframe
+from sktime.utils.data_processing import from_nested_to_2d_array
+
+# import matplotlib.pyplot as plt
+
+DATA_PATH = os.path.join(os.path.dirname(sktime.__file__), "datasets/data")
```
mloning, line 14: Please use `load_arrow_head()` instead from `sktime/datasets`

sktime/clustering/tests/test_time_series_k_means.py
```diff
@@ -0,0 +1,54 @@
...
+# X, Y = load_from_tsfile_to_dataframe(
+#     os.path.join(DATA_PATH, "ArrowHead/ArrowHead_TRAIN.ts")
+# )
+#
+# X_test, Y_test = load_from_tsfile_to_dataframe(
+#     os.path.join(DATA_PATH, "ArrowHead/ArrowHead_TEST.ts")
+# )
+
+
+def __run_test(model: TimeSeriesKMeans):
```
mloning, line 21:
```suggestion
def _run_test(model: TimeSeriesKMeans):
```

sktime/clustering/tests/test_time_series_k_means.py
```diff
@@ -0,0 +1,54 @@
...
+    n, sz = 100, 10
+    rng = np.random.RandomState(0)
+    X = rng.randn(n, sz)
+    rng = np.random.RandomState(1)
+    X_test = rng.randn(n, sz)
+
+    model.fit(X)
```
mloning, line 28: I suggest to automate these tests, using `all_estimators` and a common set of unit tests for API compliance, especially types/dimension of input and output to the main interface points (`fit`, `predict`)

sktime/clustering/utils/_utils.py
```diff
@@ -0,0 +1,67 @@
...
+__all__ = ["compute_pairwise_distances"]
+
+from typing import Callable
+import numpy as np
+
+from sktime.clustering.base.base_types import Numpy_Array, Metric_Function
+from sklearn.metrics.pairwise import pairwise_distances
+
+
+def compute_pairwise_distances(
```
mloning, line 14:
```suggestion
def _compute_pairwise_distances(
```

sktime/clustering/utils/_utils.py
```diff
@@ -0,0 +1,67 @@
+# -*- coding: utf-8 -*-
```
mloning, line 1: move to `sktime/clustering/utils.py`, no need to have a folder here

sktime/utils/_testing/estimator_checks.py
```diff
@@ -347,6 +349,12 @@ def check_raises_not_fitted_error(Estimator):
 def check_fit_idempotent(Estimator):
     # Check that calling fit twice is equivalent to calling it once
     estimator = _construct_instance(Estimator)
+
+    # Due to how clustering algorithms work, cant guarantee same
```
mloning, line 353: Could you comment on this a bit more why we cannot handle this with random states?

mloning (changes requested):
@chrisholder @TonyBagnall this is a huge amount of work! 

A few general comments: 
* please don't use double underscores, they are reserved for special methods in Python, single underscore is for private methods by convention
* think above which methods should be private and which ones public (e.g. I'd think we would want to make `calculate_new_centers` private?)
* automate tests using `all_estimators` and a common test suit for clusterers in `sktime/clustering/tests/test_all_clusterers.py`

sktime/utils/_testing/estimator_checks.py
chrisholder: So with k-means and k-medoids how the clusters will form depend massivly on where the cluster centers initialise. As all of the cluster initilisation algorithms are based on probability it means we'll never have the same clusters being formed. 

I could create a new optional parameter for random state and use it for the initialisation of the clusters so we get conistent behaviour. In addition or as a alternate (although I was thinking of doing this at a later date) we could take 'starting_centers' as a optional parameter that will be used as the starting centers.

TonyBagnall: if you could seed them that would be best, we do it with the classifiers, I think thats all thats needed at the moment

mloning: Yes, passing a random state via `check_random_state` from scikit-learn would ensure reproducibility and is inline with the other algorithm implementations.

sktime/clustering/_k_medoids.py
```diff
@@ -4,63 +4,66 @@
...
     def __init__(
         self,
         n_clusters: int = 8,
-        init_algorithm: Init_Algo = "forgy",
+        init_algorithm: InitAlgo = "forgy",
         max_iter: int = 300,
         verbose: bool = False,
-        metric: Metric_Parameter = "dtw",
+        metric: MetricParameter = "dtw",
+        random_state: NumpyRandomState = 1,
```
mloning, line 55: Usually the default in scikit-learn is to have `random_state=None` meaning it's not seeded and may produce different results every time you call it

sktime/clustering/partitioning/_k_partition.py
```diff
@@ -67,58 +83,34 @@ class TimeSeriesKPartition(BaseCluster, ClusterMixin):
...
     def __init__(
         self,
         n_clusters: int = 8,
-        init_algorithm: Init_Algo = "forgy",
+        init_algorithm: InitAlgo = "forgy",
         max_iter: int = 300,
         verbose: bool = False,
-        metric: Metric_Parameter = "dtw",
+        metric: MetricParameter = "dtw",
+        random_state: NumpyRandomState = 1,
```
mloning, line 97: same here 
```suggestion
        random_state: NumpyRandomState = None,
```

PR #965, opened 2021-06-15
re-factoring existing forecasters - Stacking,  Multiplexer, Ensemble and TransformedTarget Forecasters

thayeylolu:

- [ ] NaiveForecasters
- [x] EnsembleForecaster
- [x] MultiplexForecaster
- [x] TransformedTargetForecaster
- [ ] _Reducer and related code
- [x] StackingForecaster
- [ ] ForecastingGridSearchCV, RandomizedGridSearchCV, and BaseGridSearch
- [ ] OnlineEnsembleForecaster and descendants
- [ ] _PmdArimaAdapter and descendants ARIMA, AutoARIMA
- [ ] BATS, TBATS, and _Tbatsadapter
- [ ] _StatsModelAdapter and descendants AutoETS, ExponentialSmoothing, ThetaForecaster
- [ ] HCrystalBallForecaster
- [ ] Prophet and _ProphetAdapter
- [ ] PolynomialTrendForecaster


fkiraly-shell:
Good!

I think you still need to copy the tag from the forecasting horizon mixin to the refactored base classes.

fkiraly-shell (approved):
see above, tags from the mixin should be added as class variables, compare the naive forecaster

fkiraly-shell:
whoops, I meant "request changes"

fkiraly-shell (changes requested):
[ removed, used the wrong account ]

fkiraly (changes requested):
sorry, used the wrong account.

Looks good, but you need to add the estimator tags from the mixin to the concrete class, see the `NaiveForecaster` refactor.

thayeylolu:
Thank you @fkiraly . the refracted NaiveForecaster uses `_tags = {"requires-fh-in-fit": False}`. But the `extension_template uses ` _tags = { "fh_in_fit": True}`. Which should I go with?

fkiraly:
> Which should I go with?

That depends, namely on the mixin the forecaster had inherited from before the refactor:

* if it inherited from `_OptionalForecastingHorizonMixin`, it should get the tag `"requires-fh-in-fit": False`
* if it inherited from `_RequiredForecastingHorizonMixin`, it should get the tag `"requires-fh-in-fit": True`

thayeylolu:
Thank you @fkiraly

thayeylolu:
Hi @fkiraly , I want to include the missing_data tag and the univariate tag in the forecaster's tag , but I honestly don't know if a forecaster can handle missing data or if it is univariate, unless it is stated in the doctstring. Is there a way I can go about this?

fkiraly:
>Hi @fkiraly , I want to include the missing_data tag and the univariate tag in the forecaster's tag , but I honestly don't know if a forecaster can handle missing data or if it is univariate, unless it is stated in the doctstring. Is there a way I can go about this?

That's a very good question.

I would recommend, always go with the "negative" version of the tag, i.e., assume the forecaster can do nothing. That way it will not break when the user tries to ask it to do something.

The "negative" version of the tags should be the ones in the extension guidelines.

If, despite this, the forecasters can deal with xyz, the I'd assume one of the authors or code owners will see this and comment that the "correct" tag should be different.

Lovkush-A:
@thayeylolu @fkiraly 

What is the plan for this PR? Is the plan for this single PR to update *all* the forecasters?  If yes, I recommend changing the plan to split up the task into multiple PRs. The main reason is that it is easier to review and to debug small PRs than a single large PR.

fkiraly:
I agree with @Lovkush-A - to make review and work easy, let's separate this into PR, @thayeylolu, along the lines of packages that are interfaced. E.g., all `pmdarima` in one PR, all `tbats` in one, etc.

The current changes can remain in here for the time being, but let's open new PR for other forecasters.

fkiraly:
@mloning, this is very odd - this PR didn´t do anything to time series classifiers, but that´s where the tests are now failing. 

We initially suspected it had to do with the latest commits, so I suggested to create #977 which is without the latest changes.

Locally, according to @thayeylolu, the tests are *not* failing, which may indicate that this is perhaps due to the TSC code being affected by a package version change?

fkiraly:
this looks done, do you want to update from `main` and make this a proper PR?

Lovkush-A:
> this looks done, do you want to update from `main` and make this a proper PR?

There are two version of this PR. #977 is the other one and I think 977 is the latest version.

thayeylolu:
Yes. Thats why I closed this. I will have to write the reason for closing PRs in the future. Sorry for the misunderstanding

Issue #955, opened 2021-06-13
re-factoring existing forecasters to new interface


fkiraly:
We need to slowly re-factor all forecasters to be compliant with the new interface (see PR #912). i.e., no overrides of `fit` etc and descendant classes only implementing core logic, while the `BaseForecaster` implements check etc logic.

Currently this is running due to various downwards compatibility workarounds, which should be removed once all forecasters are ready (in "no-override" state)

The forecasters that still need a downstream refactor are:
 * [x] `NaiveForecasters` #953
 * [x]  `EnsembleForecaster` #977
 * [x]  `MultiplexForecaster` #977
 * [x]  `TransformedTargetForecaster` #977
 * [x]  `_Reducer` and related code #1031
 * [x]  `StackingForecaster` #977
 * [x]  `ForecastingGridSearchCV`, `RandomizedGridSearchCV`, and `BaseGridSearch` #1034
 * [x]  `OnlineEnsembleForecaster` and descendants #1015
 * [x]  `_PmdArimaAdapter` and descendants `ARIMA`, `AutoARIMA` #1016 - adapter only
 * [x]  `BATS`, `TBATS`, and `_Tbatsadapter` #1017 adapter only
 * [x]  `_StatsModelAdapter` and descendants `AutoETS`, `ExponentialSmoothing`, `ThetaForecaster`  #1029
 * [x]  `HCrystalBallForecaster` #1004
 * [x]  `Prophet` and `_ProphetAdapter` #1005
 * [x]  `PolynomialTrendForecaster` #1003

For a how-to, useful resources are the forecaster extension template in `extension_templates` and PR #953 as an example for `NaiveForecaster`.

Based on discussions, we need to do the following two things in a capstone refactor that removes the last traces of the workarounds and goes over the above for consistency:
- [x] remove the `_SktimeForecaster` and mixins for good
- [ ] go through all forecasters and check the tags for correctness

thayeylolu:
Thank you @fkiraly . I shall begin work on this. And make a draft shortly

fkiraly:
For convenience, I'm adding a recipe here:

* go through all the public methods, `fit`, `predict`, etc, and change them to the new private version `_fit`, `_predict`, etc
* remove input checks (e.g., `check_X`) and output operations (e.g., set cutoff) from the private version; make sure these are already all present in the `BaseForecaster`.
[...]

PR #967, opened 2021-06-16
ForecastingPipeline for pipelining with exog data

aiwalter:
#### Reference Issues/PRs
Enhancement proposal: https://github.com/sktime/enhancement-proposals/blob/main/steps/01_forecasting_api/forecasting_pipeline.md

ForecastingPipeline for pipelining with exog data

No

1. Inheritance from new class `_Pipeline`
2. There is an issue to use the `ForecastingPipeline` just for the purpose of a transformer pipeline, bcz in the transformer interface, we allow in `transform(Z, X)` that `Z` can be a `pd.DataFrame` but in the `fit(y, X)` function `y` must be a `pd.Series`. Solutions for this which I have considered:
    1)  It would only be allowed to use `ForecastingPipeline` as a transformer pipeline if `y` and `X` are given like `fit_transform(y, X)`. This is the status when I open this PR.
    2)  We could add an if-statement in `fit()` to catch cases where `y` is a `pd.DataFrame`, however this would be an impact to the interface that we allow `y` to be a `pd.DataFrame` which is never the case so far. 
    3) We dont implement `transform()` for `ForecastingPipeline`

General question of using `ForecastingPipeline` or `TranformedTargetForecaster` as a transformer pipeline: when calling `fit_transform`, the `fit` and `transform` functions are called and this means that also the given `forecaster` is fitted. So also a forecaster has to be given, but we might want to allow `forecaster=None`. This might probably best be done in a separate PR.

##### For all contributions

##### For new estimators

___


aiwalter:
this is ready for review. The test failure will be solved after above question regarding transformer pipeline is solved.
@mloning 
___

fkiraly:
@aiwalter, regarding your question:
I would like to move the forecaster interface to a stage where the `y` in `fit` can also be a `pd.DataFrame` - we need this for multivariate forecasting.

Using the new forecasting interface, this change should be easy once #955 is complete.
That would partly alleviate your problem, no?

aiwalter:
yes, but maybe then we should write `Y` everywhere? Bcz in transformer interface we also write `Z` which can be a `pd.Series` or a `pd.DataFrame`? However this would be a big interface impact.
___

fkiraly:
>However this would be a big interface impact.

Yes, that's what frightens me - we would have to replace `y` everywhere by `Y`, and we can't do a bulk replace, because who knows whether it's always the right `y`. I'd rather leave the variable name at is to avoid the risk of breaking things.

aiwalter:
hm not sure whats the best solution, but I think we could manage to replace `y` with `Y`. The question would be how to depreciate it so that for some time both would work for the interface? Not sure if this would even be possible..

fkiraly:
>Not sure if this would even be possible..

```python
def myfunction(stuff, y, Y=None):
    if Y is none:
        Y=y
otherstuff
```

aiwalter:
Yes but the issue is when only positional args are given like `fit(y, X)`. This would be in your solution understood as `fit(y=y, Y=X)`? So we could solve it maybe when we always put `Y` as the last arg like `fit(y, X, Y)`? When we remove `y` finally we would habe to put `Y` again in first position.

sktime/forecasting/compose/_pipeline.py
```diff
@@ -13,55 +13,17 @@
...
 from sktime.transformations.base import _SeriesToSeriesTransformer
-from sktime.utils.validation.forecasting import check_y
+from sktime.utils.validation.forecasting import check_y_X, check_y, check_X
 from sktime.utils.validation.series import check_series
 from sktime.utils import _has_tag
 
 
-class TransformedTargetForecaster(
+class _Pipeline(
     _OptionalForecastingHorizonMixin,
```
mloning, line 22: We need to adjust this to the new forecasting API introduced in #https://github.com/alan-turing-institute/sktime/pull/912, instead of the mixin we now use the tag `requires-fh-in-fit`

sktime/forecasting/compose/_pipeline.py
```diff
@@ -114,6 +76,196 @@ def named_steps(self):
...
+    >>> y_pred = pipe.predict(fh=fh, X=X_test)
+    """
+
+    _required_parameters = ["steps"]
+
+    def __init__(self, steps):
+        self.steps = steps
+        self.steps_ = None
+        super(ForecastingPipeline, self).__init__()
+
```
mloning, line 145: @fkiraly and I found out that we need to run this in init too so that people can set nested parameters before calling fit.  
```suggestion
        self.steps_ = self._check_steps()
        
```

sktime/forecasting/compose/_pipeline.py
```diff
@@ -114,6 +76,196 @@ def named_steps(self):
...
+    def fit(self, y, X=None, fh=None):
```
mloning, line 146: has to be adjusted to the new API

sktime/forecasting/compose/_pipeline.py
```diff
@@ -114,6 +76,196 @@ def named_steps(self):
...
+        f = clone(forecaster)
+        f.fit(y, Xt, fh)
+        self.steps_[-1] = (name, f)
+
+        self._is_fitted = True
+        return self
+
+    def _predict(self, fh=None, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
+        Xt = check_X(X)
+        forecaster = self.steps_[-1][1]
```
mloning, line 183: what happens if we only pass transformers without a final forecaster?

mloning (commented):
Hi @aiwalter, I had a quick look at it and left a few in-line comments.

Regarding your questions:
1. Looks good to me
2. Not sure if I can follow, isn't the idea that `ForecastingPipeline` only operates on the `X` given in `fit(y, X, ...)` or `predict(..., X=X)`? I see that this is what you have implemented too, as you call `transform(X)` inside the pipeline, ignoring `y`. Creating multivariate forecasters would be a separate issue, no? Perhaps I'm missing something. 
3. I agree, both pipelines should work without a final forecaster and then only have `transform` and not the `predict` method.

sktime/forecasting/compose/_pipeline.py
aiwalter: this is not yet working, I thought I might do it in a separate PR also then for `TransformedTargetForecaster`

sktime/forecasting/compose/_pipeline.py
aiwalter: is this `_tags` also inheritable so that I can write it only in `_Pipeline`? I think I can set here `_tags = {"requires-fh-in-fit": False}`  right?

sktime/forecasting/compose/_pipeline.py
aiwalter: ok, so just `def _fit(self, y, X=None, fh=None):` instead right?

aiwalter:
@mloning thanks for review. There is an issue with the tests for multivariate bcz `ForecastingPipeline` would work only if `X is not None` and wouldn't make sense either otherwise. So I added a `NotImplementedError` but seems the tests dont like it. Further this would mean to skip many of existing tests, so should I make all those tests ready for multivariate, meaning we always give `y_train, X_train` instead of `y_train` only. Same for `predict` then. Not sure if this would break existing tests for univariate stuff? Actually most univariate classes/models would just ignore the given `X`, right? What do you think? Or how could we ensure to give for some classes always also an `X` in the tests?

sktime/forecasting/compose/_pipeline.py
mloning: Yes, see `sktime/extension_templates/forecasting.py`

sktime/forecasting/compose/_pipeline.py
mloning: There is an ongoing discussion on tags in #981 but for now `{"requires-fh-in-fit": False}` seems to be correct

fkiraly: well, I'd actually think it's `{requires-fh-in-fit": True}`, because formally (scitype level) it's whatever the forecaster at the end requires, and that could be `True` - in which case the pipeline would break. The "safe" option is hence always asking for `fh` in `fit`.

fkiraly: >is this `_tags` also inheritable so that I can write it only in `_Pipeline`?

Yes, it's inheritable - use of the method `_all_tags` mimicks inheritance.

mloning: But not all pipelines require fh in fit? Let's defer this discussion when we implement object-level tags (see #981)

aiwalter:
@fkiraly @mloning I would like to merge this soon. Could s.o. pls have a look?

aiwalter:
Also @ViktorKaz

fkiraly:
@aiwalter, why does this not work if `X` is `None`? Wouldn't you just skip anything for the exogeneous part? I.e., behave as if it were a data frame with zero columns?

fkiraly:
Question: would this problem

>There is an issue to use the ForecastingPipeline just for the purpose of a transformer pipeline, bcz in the transformer interface, we allow in transform(Z, X) that Z can be a pd.DataFrame but in the fit(y, X) function y must be a pd.Series. Solutions for this which I have considered: 

Be resolved with #980? That way, `y` could be a `pd.Series` or a `pd.DataFrame`.

fkiraly:
>Yes but the issue is when only positional args are given like `fit(y, X)`. This would be in your solution understood as `fit(y=y, X=Y)`? So we could solve it maybe when we always put Y as the last arg like `fit(y, X, Y)`? When we remove y finally we would habe to put `Y` again in first position.

Yes - this could be achieved by a decorator as well. I would be for doing this, and deprecate in 1.0 or similar.

sktime/forecasting/compose/_pipeline.py
fkiraly: well, I think you need to decide now whether you want object level tags or not...
You can't be at the same time be for a tag with different value for a given concrete pipeline, and against dynamic tag override at construction. I.e., either we must say that `requires-fh-in-fit` is always true, or have a mechanism that inspects the tags of the pipeline at construction (or later) and assigns the right dynamic tag.

fkiraly:
is there manual testing code for this anywhere?

The lack of examples makes this difficult to review.

aiwalter:
> why does this not work if X is None? Wouldn't you just skip anything for the exogeneous part? I.e., behave as if it were a data frame with zero columns?
yes thats right, I implemented this.

> Be resolved with #980? That way, y could be a pd.Series or a pd.DataFrame.
yes, that would be resolved and we could implement it then so `ForecastingPipeline` would also inherit from `_SeriesToSeriesTransformer`. So we would catch the case when `isinstance(y, pd.DataFrame)` then it would have to be a transformer, but in case there is a vector forecaster like VAR, then the user will give only `y` as a `DataFrame` and not `X`, right? We would then just check if last step is a forecaster or not. So this is related to issue #1090 .  We can make it working.

> is there manual testing code for this anywhere?
Yes its there in docstring as example. Here I have one more complex one:
```python

from sktime.forecasting.compose import ForecastingPipeline
from sktime.transformations.series.adapt import TabularToSeriesAdaptor

y, X = load_longley()

y_train, _, X_train, X_test = temporal_train_test_split(y, X)
fh = ForecastingHorizon(X_test.index, is_relative=False)

pipe_y = TransformedTargetForecaster(steps=[
    ("scaler", TabularToSeriesAdaptor(MinMaxScaler())),
    ("imputer", Imputer()),
    ("forecaster", NaiveForecaster())
])

pipe_X = ForecastingPipeline(steps=[
    ("imputer", Imputer()),
    ("scaler", TabularToSeriesAdaptor(MinMaxScaler())),
    ("forecaster", pipe_y)
])
pipe_X.fit(y_train, X_train)
y_pred = pipe_X.predict(fh=fh, X=X_test)
```

sktime/forecasting/compose/_pipeline.py
aiwalter: new forecasting API was adapted by merge

sktime/forecasting/compose/_pipeline.py
aiwalter: this has to be done when we accept `isinstance(y, pd.DataFrame)`

aiwalter:
any other comments?

sktime/forecasting/compose/_pipeline.py
```diff
@@ -135,17 +162,168 @@ def _fit(self, y, X=None, fh=None):
...
+        Returns
+        -------
+        y_pred : pd.Series
+            Point predictions
+        y_pred_int : pd.DataFrame - only if return_pred_int=True
+            Prediction intervals
+        """
+        forecaster = self.steps_[-1][1]
+        # copy to avoid transformation on original variable
+        Xt = _copy(X)
```
mloning, line 214: I don't think we should copy the data here, I think the component transformers can take care of that, no? Otherwise we end up copying the data more often than we need to.

sktime/forecasting/compose/_pipeline.py
```diff
@@ -135,17 +162,168 @@ def _fit(self, y, X=None, fh=None):
...
+        X : pd.DataFrame, required
             Exogenous variables are ignored
         Returns
         -------
         self : returns an instance of self.
         """
+        # Some transformers can not deal with X=None, therefore X is mandatory
+        self._set_y_X(y, X)
+        # copy to avoid transformation on original variable
+        Xt = _copy(self._X)
```
mloning, line 174: See my comment below on copying.

sktime/forecasting/compose/_pipeline.py
```diff
@@ -135,17 +162,168 @@ def _fit(self, y, X=None, fh=None):
...
+        X : pd.DataFrame, optional (default=None)
+            Exogenous variables are ignored
+        Returns
+        -------
+        self : returns an instance of self.
+        """
+        self._set_y_X(y, X)
+
+        # copy to avoid transformation on original variable
+        yt = _copy(self._y)
```
mloning, line 325: See my comment on copying.

sktime/forecasting/compose/_pipeline.py
```diff
@@ -204,11 +382,11 @@ def _update(self, y, X=None, update_params=True):
 
         for step_idx, name, transformer in self._iter_transformers():
             if hasattr(transformer, "update"):
```
mloning, line 384: Mh, shouldn't all transformers have an `update`? By default they should fall back to re-fitting similar to what we do for forecasters.

mloning (commented):
Hi @aiwalter, a few more comments. 

I'm not entirely sure about the copying, what do you think?

sktime/forecasting/compose/_pipeline.py
aiwalter: this is not defined in the parent classes, but would be good to have it proably. I there already an issue for refactoring transformers where we can note this down? Also the `y = y.copy()` should be moved to parent.

aiwalter:
> I don't think we should copy the data here, I think the component transformers can take care of that, no? Otherwise we end up copying the data more often than we need to.

Yes, seems to be fine without copy again here. I removed the copy stuff, also in `TransformedTargetForecaster`.

I was thinking about the name `ForecastingPipeline` and that we might want to reserve this name for a different class like this solution:

```python
pipe = ForecastingPipeline(
    steps_y=[
        ("outlier", HampelFilter()),
        ("imputer", Imputer())
        ],
    steps_X=[
        ("standardize", TabularToSeriesAdaptor(StandardScaler())),
        ("imputer", Imputer()),
        ],
    forecaster=AutoARIMA()
)
```
Also, what I wrote as `ForecastingPipeline` is just the equivalent of the `TansformedTargetForecaster` but for `X`. So I came up with the idea to change the name to `TansformedExogForecaster` or `TansformedExogenousForecaster`. What do you think @mloning @fkiraly . I think it just described better what is done there and would fit well with the name `TansformedTargetForecaster`.
___

sktime/forecasting/compose/_pipeline.py
mloning: Yes let's do this in a separate issue

sktime/forecasting/compose/_pipeline.py
```diff
@@ -223,29 +393,21 @@ def inverse_transform(self, Z, X=None):
...
-    def set_params(self, **kwargs):
-        """Set the parameters of this estimator.
-        Valid parameter keys can be listed with ``get_params()``.
-        Returns
-        -------
-        self
-        """
-        self._set_params("steps", **kwargs)
-        return self
+def _copy(Z=None):
```
mloning, line 401: so can this go then?

mloning (commented):
Everything else looks good to go

sktime/forecasting/compose/_pipeline.py
aiwalter: ah yes

mloning (approved):
I've reviewed twice, I think this is good to go. 

There are still a few uncertainties around how to best handle transformers inside the pipeline, but I think that should be addressed in a separate issue. @fkiraly is there already an issue for the transformer refactoring, I couldn't find one? 

Does anyone else want to review before we merge?

aiwalter:
@fkiraly I would be interested in your opinion regarding the class name, what I wrote above. Also would be great if we can merge this then if you dont have any further comments :)

aiwalter:
can we merge this @mloning . I assume we have "lazy approval" here :)

mloning:
Not sure if @fkiraly wanted to take another look? I'll merge tomorrow if not.

mloning:
Now merged @aiwalter 🎉

PR #975, opened 2021-06-18
Enhancement on RISE

whackteachers:

added fft padding in ps function using a new rounding function
ps now has more parameters to increase flexibility
added function to select single interval and multiple intervals
decorated acf with numba jit function and simplified the math

now requires numpy version >1.20

- anything in _rise.py
#### Any other comments?

I may have included some local things in experiments.py which can be safely discarded
Remember to use the updated code when upgrading numpy to 1.20 for the fft function
##### For all contributions

##### For new estimators


sktime/classification/interval_based/_rise.py
```diff
@@ -374,34 +513,125 @@ def matrix_acf(x, num_cases, max_lag):
...
+    fft = fft[:-1] if x_is_1d else fft[:, :-1]
+    return np.abs(fft)
+
+
+@jit('int64(int64)', cache=True, nopython=True)
+def _round_to_nearest_power_of_two(n):
+    return int64(1 << round(np.log2(n)))
+
+
+if __name__ == '__main__':
```
mloning, line 622: I suggest to remove the `if __name__ == '__main__':` section

sktime/classification/interval_based/_rise.py
```diff
@@ -374,34 +513,125 @@ def matrix_acf(x, num_cases, max_lag):
...
+    #     #     y[lag - 1] = 0
     return y
 
-
-def ps(x):
-    """Power spectrum transform.
-
-    Calculated using np function. It would be worth looking at ff implementation,
-    see difference in speed to java.
+# def ps(x):
```
mloning, line 551: if we don't use the commented-out functions here and throughout the file, I suggest to remove them rather than comment them out, git allows you to always go back and look at the old code if necessary

mloning (commented):
Hi @whackteachers, thanks for the updates, just two quick comments from my side.

whackteachers:
it seems it somehow fail the tests due to import errors and not using updated numpy version

sktime/classification/interval_based/_rise.py
```diff
@@ -374,34 +504,87 @@ def matrix_acf(x, num_cases, max_lag):
...
+            x_in_power_2 = np.pad(x, pad_length, mode=pad)
+        else:
+            x_in_power_2 = x[:n] if x_is_1d else x[:, :n]
+    else:
+        x_in_power_2 = x
+    # use sign to determine inverse or normal fft
+    # using the norm in numpy fft function
+    # backward = normal fft, forward = inverse fft (divide by n after fft)
+    norm = "backward" if sign > 0 else "forward"
+    fft = np.fft.rfft(x_in_power_2, norm=norm)
```
MatthewMiddlehurst, line 583: Already commented but a bit more specific, seems to be failing tests here due to the norm options. Error for most is "ValueError: Invalid norm value backward, should be None or "ortho".

The numpy version used by sktime is 1.19 currently, though it seems those options are valid as of 1.2.

TonyBagnall: hi @whackteachers Im not sure when we will update or indeed what our schedule for doing so is, but it is probably easier to just make it compatible with 1.19. Otherwise it all looks good to me

whackteachers: I have made the change, plz check if I m doing it the right way

TonyBagnall:
hi Jason, this is failing because you have changed the algorithm (which we approved), but the test fails as the values in it are based on the old algorithm. This means the test file needs updating, since the probabilities will have changed. You can run a version of test_y_proba_on_gunpoint locally, get the expected probas, update values in test_rise, and make the PR again. 
image

sktime/classification/interval_based/_rise.py
```diff
@@ -185,20 +249,30 @@ def fit(self, X, y):
         X = X.squeeze(1)
 
         n_instances, self.series_length = X.shape
+        if self.max_interval not in range(1, self.series_length):
+            self.max_interval = self.series_length
+        if self.min_interval not in range(1, self.series_length + 1):
+            self.min_interval = self.series_length // 2
```
whackteachers, line 255: uhm... it seems it is not happy with me doing with this :confused:

TonyBagnall (approved):
I think its good to go in now once updated to main, thanks for following through with this

whackteachers:
Thank you for your assistance along too :blush:

PR #976, opened 2021-06-18
Refactoring Reducer

thayeylolu:

- [ ] _Reducer and related code


fkiraly:
this should probably be a draft until the tests run?

Lovkush-A:
@thayeylolu Taiwo, may we delete or close this PR? Two reasons are:
* It is easier for me to make changes in a new PR, than to make changes in somebody elses PR
* The discussion with tags means that many of the changes in this PR might not be valid

Let me know what you think! Thanks

thayeylolu:
Yes, please @Lovkush-A

fkiraly:
where is the new PR?

thayeylolu:
I was hoping @Lovkush-A would like to work on it

PR #972, opened 2021-06-18
Forecasting tutorial rework

fkiraly:
This pull request contains a re-work of the forecasting tutorial which should make it more structured, complete, and accessible - from what right now looks like a patchwork piece that has organically proliferated...

Main changes:

* re-structuring in chapters - basic workflows, selected estimators, composition (reduction, pipelines)
* expanded explanation of what happens in sections/chapters, added some code for object inspection
* factored out the "pitfalls" in a separate notebook, also expanded - focus is now on use of `sktime` primarily

PR #977, opened 2021-06-18
Refactoring Stacking, Multiplexer, Ensembler and TransformedTarget Forecasters

thayeylolu:

-- This is a copy of 965 without changes to reducer.py (a time when there was no bug)

- [x] EnsembleForecaster
- [x] MultiplexForecaster
- [x] TransformedTargetForecaster
- [x] StackingForecaster


fkiraly:
looks like checks are all fine, so do you want to convert to "proper" PR and request reviews?

fkiraly:
and you may want to close #955 of which this is now a duplicate

sktime/forecasting/base/_base.py
```diff
@@ -105,6 +105,7 @@ def fit(self, y, X=None, fh=None):
         self._X = X
         self._y = y
 
+        self._set_cutoff(y.index[-1])
```
fkiraly, line 108: this is now included in `_base`, make sure you update from `main`

sktime/forecasting/compose/_multiplexer.py
```diff
@@ -81,6 +78,12 @@ class MultiplexForecaster(
     'arima'
     """
 
+    _tags = {
+        "univariate-only": True,
+        "requires-fh-in-fit": False,
+        "handles-missing-data": False,
+    }
+
```
fkiraly, line 86: this is not correct - if one of the components requires the `fh` in `fit`, then so does the `MultiplexForecaster` - or, more precisely, if the selected forecaster does.

Question for you, also in general in terms of design for @mloning,@aiwalter, @koralturkk : we seem to have three options here:
* always set the tags to the "worst" - this is the "no-error" setting at class level
* set at construction, based on components but not on value of the multiplexer - this would look at the tags of the components, and set each tag to the "bad" version if it appears in at least one of the components
* set whenever parameters are set using `set_params` - then, set the tags to the tag of the component it points to

I slightly prefer option no.2 (set at construction, look only at tags of components)

fkiraly: also, @koralturkk, why did it have the `_OptionalForecastingHorizonMixin` to start with? That looked incorrect, for the same reason as above.

sktime/forecasting/compose/_pipeline.py
```diff
@@ -19,7 +18,6 @@
 
 
 class TransformedTargetForecaster(
-    _OptionalForecastingHorizonMixin,
     _SktimeForecaster,
```
fkiraly, line 21: I think you should also remove the `_SktimeForecaster`

sktime/forecasting/compose/_stack.py
```diff
@@ -12,23 +12,39 @@
...
+    n_jobs : int or None, optional (default=None)
+        The number of jobs to run in parallel for fit. None means 1 unless
+        in a joblib.parallel_backend context.
+        -1 means using all processors.
+    """
+
     _required_parameters = ["forecasters", "final_regressor"]
+    _tags = {
+        "univariate-only": True,
+        "requires-fh-in-fit": True,
```
fkiraly, line 38: hm, similar as for the `MultiplexForecaster` - does whether a `fh` is required in `fit` not depend on the components?

fkiraly (changes requested):
This is great, thanks for your work!

Although the tests pass, I have some questions about the right tags, and the potential to dynamically attach tags based on components, to @mloning, @aiwalter, @koralturkk.

Further, I think the pre-refactor mixins for `MultiplexForecaster` and `EnsembleForecaster` were incorrect, `fh` could be required in `fit` if the components do require it. Why did the tests not catch this?

sktime/forecasting/compose/_pipeline.py
fkiraly: I updated the recipe in #955, I forgot to talk about `SktimeForecaster` there

mloning:
I'm not really sure what the best solution is. A few thoughts: 
* I see tags as being defined on the class (not object) level, conceptually I think of them as describing algorithm properties, so I would prefer not to update them dynamically
* In the cases of the above mentioned (composition) algorithms, for the implementation (i.e. input validation of the forecasting horizon), the tag should be set to False, the composite doesn't require the forecasting horizon during fitting, it merely passes it on to the components, which have their own tag and raise an error accordingly if they require the forecasting horizon during fitting
*  For documentation, when users may use tags for finding relevant models, setting it to False may be misleading, but conceptually still makes sense, as it refers to the composition algorithm, rather than a particular instance of the composite

sktime/forecasting/compose/_stack.py
fkiraly: ok, tags issue is now in #981.
For now, this is fine.

sktime/forecasting/compose/_pipeline.py
fkiraly: this still should be done, it´s not a tags issue

sktime/forecasting/compose/_multiplexer.py
fkiraly: this should be changed to `"requires-fh-in-fit" : True`, that's safe until the tags are sorted out.

fkiraly:
ok, agreed - let's split off the larger tags discussion, and just do a "safe" refactoring here.
Outstanding changes:

* the `_SktimeForecaster` removal is not tags related and should be done
* multiplexer tag should be changed to requires fh in fit (see above)

sktime/forecasting/compose/_multiplexer.py
thayeylolu: Thank you @fkiraly . I would set it to true

sktime/forecasting/base/_base.py
thayeylolu: I will do that

sktime/forecasting/compose/_pipeline.py
thayeylolu: noted

thayeylolu (commented):
I have made the necessary changes and I would I to convert the draft for review

fkiraly:
the test examples are now failing - they ran before.
Can you please check what's happening?

fkiraly:
I think it's the multiplexer - the tutorial notebook passes the `fh` too late

fkiraly:
I'd suggest:

* don't change the notebook, since #972 is currently under review, and the change to notebook will just cause very fiddly merge issues
* instead, change the tag back, and open a new issue that it needs to be changed once #972 is merged, together with the notebook

What do you think, @mloning?

sktime/forecasting/compose/_multiplexer.py
thayeylolu: I have set it to True, but it fails the test in the true state

thayeylolu:
@fkiraly Setting `requires-fh-in-fit`: to `True` is the cause

fkiraly:
>@fkiraly Setting requires-fh-in-fit: to True is the cause for the failure of the multiplexer

You mean, to `False`, no? We want to set it back to `True` according to my suggestion - please confirm?

sktime/forecasting/compose/_multiplexer.py
```diff
@@ -81,6 +78,12 @@ class MultiplexForecaster(
     'arima'
     """
 
+    _tags = {
+        "univariate-only": True,
+        "requires-fh-in-fit": True,
```
thayeylolu, line 83: @fkiraly . it  `requires-fh-in-fit` is currently set to `True`. When it is `False` it did not fail. Originally, Multiplexer inherits from `_OptionalForecastingHorizonMixin`.

fkiraly: yes, let's move it back, and make note that this may need to be changed (see discussion), in another issue.

fkiraly:
pinging @ninamiolane

sktime/forecasting/compose/_multiplexer.py
thayeylolu: Done

sktime/forecasting/compose/_multiplexer.py
```diff
@@ -147,24 +150,49 @@ def fit(self, y, X=None, fh=None, **fit_params):
...
+        y : pd.Series
+        X : pd.DataFrame, optional (default=None)
+        update_params : bool, optional (default=True)
+
+        Returns
+        -------
+        self : an instance of self
+        """
         self.check_is_fitted()
         self._update_y_X(y, X)
```
mloning, line 197: This can go, right?
```suggestion
```

sktime/forecasting/compose/_stack.py
```diff
@@ -71,10 +85,10 @@ def fit(self, y, X=None, fh=None):
         # refit forecasters on entire training series
         self._fit_forecasters(forecasters, y, fh=self.fh, X=X)
 
-        self._is_fitted = True
         return self
 
-    def update(self, y, X=None, update_params=True):
+    def _update(self, y, X=None, update_params=True):
+
```
mloning, line 91: Same here, we can remove: 
```python
        self.check_is_fitted()
        self._update_y_X(y, X)
```

mloning (commented):
Hi @thayeylolu, you need to check the `update` methods again, I think some of the code can still be removed there. Otherwise looks good to me!

sktime/forecasting/compose/_stack.py
thayeylolu: Ah!. Noted. Thank you

sktime/forecasting/compose/_multiplexer.py
fkiraly: indeed, but as always, also please check in `BaseForecaster` whether there's corresponding code there.

thayeylolu: Yes it is

thayeylolu:
@mloning this I ready for review :)

fkiraly (approved):
Looks good to me - the earlier request is partly related to tags and will be addressed separately.

Updated from main and let's wait for tests to pass.

Issue #974, opened 2021-06-18
Type hints


TonyBagnall:
add standardised type hints to classifiers, clusterers and forecasters

https://github.com/alan-turing-institute/sktime/issues/928

https://github.com/alan-turing-institute/sktime/pull/964

fkiraly:
is this dependent on #973?

TonyBagnall:
I'm closing this, issues have overtaken it

Issue #973, opened 2021-06-18
Type aliasing


TonyBagnall:
include unified and standardised type aliasing for forecasting, clustering and classification, started by @ABostrom 

https://github.com/alan-turing-institute/sktime/pull/964

fkiraly:
hm, I don't think this is a good first issue! This requires intimate knowledge of the existing codebase and the scitype taxonomy...

TonyBagnall:
Feel free to change the tag, Aaron has a first version based on the forecasting one

fkiraly:
So, @ABostrom, @chrisholder, @jambo6, @TonyBagnall, @aiwalter, @mloning, I wanted to hear your opinion on a suggestion how to deal with descriptive type hints and type checks in one.

What I would propose is that our base class `fit` etc look like as follows (below without conversion etc), in the example of the forecaster base class:

```python

@scitype_decorator
def fit(X: Series, Y: Series):
    
    X = check(X)
    Y = check(Y)

    stuff(X,Y)

    self._fit(X.inner, Y.inner)
```

About the ingredients above:
* `Series` is a custom type, or perhaps better class that denotes `scitype`, not machine type
* what `scitype_decorator` does is wrap `X` and `Y` in the declarative type (ligh-weight wrapper), and does part of what `check_series` does: it annotates `X`, `Y`, with the machine type. E.g.,: `X = Series(X)`, which moves `X` to `X.inner`
* `check` inspects `X` for validity, and adds a `X.scitype = "DataFrame"` or similar annotation. So after `check`, we know that `X` is valid, and we can read the scitype and the mtype using `X.scitype` and `X.mtype`.
* the remaining `stuff` has access to these declarative type annotations. It could be compatibility checks, conversions, etc.

This proposal has some adtanvages:

* type annotation in the base class and the user docstrings will be with the "scientific type"
* the scientific type of any object is inspectable, per declarative annotation
* we have a consistent way, across modules, to track and inspect alias mtypes for the same scitype
* this could be eaily combined with a central "register" location of scitypes, mtypes, and/or a central location for validity checking, and the conversions module as sketched in #980

chrisholder:
I like the look of this a lot @fkiraly. 

My only comment is (and Im sure this isn't final anyways) is the naming of the type 'Series' to me that implies a pandas Series. I ran into this issue of not knowing what to call my type describing pandas numpy and the other potential input types for clustering and ended up landing on:

[...]

PR #980, opened 2021-06-19
Forecasting support for multivariate y and multiple input/output types - working prototype

fkiraly:
This is a merge/review ready sketch for how multivariate `y` could be supported for multivariate forecasters, alongside the possibility to pass `np.array` and `pd.Series`. This is based on the design in [STEP no.5](https://github.com/sktime/enhancement-proposals/tree/main/steps/05_scitype_based_IO_checks), adapted to the new forecasting base class.

The key ingredient are:

* converters parameterized by from, to, and as - in the `convertIO`module. Besides the obvious conversion functionality, the converters can be given access to a dictionary via reference in the `store` argument, where information for inverting lossy conversions (like from `pd.DataFrame` to `np.array`) can be stored
* a new tag `y_type` which encodes the type of `y` that the private `_fit`, `_predict`, and `_update` assume internally - for now, it's just one type and not a list of compatible types
* some logic in the public "plumbing" area of `fit`, `predict`, `update`, which converts inputs to the public layer to the desired input of the logic layer and back

This is for discussion, and I expect major change before we merge.

It might be especially interesting to discuss the following interaction with typing:
the third argument of the converter, the "as" argument, is a scitype, and designed to have the same name as the respective `typing` composite type. Also see discussion in #965, #973 and #974.

Thinking of the logical continuation of this:

* extend this to the other arguments, e.g., `X`
* the individual argument checks like `check_X` etc should be replaced by checks of the kind `check(what, as : scitype)`, e.g., `check(y, "Series")` should replace the less generic `check_y`.
* extend this to time series classification, to support 3D numpy arrays, awkward arrays, nested data frames

A "visionary" pathway may see this working with type annotations that are scitypes, not `typing` machine types, e.g.,
```python
fit(X : Series, Y : Series)
etc
_fit(X: Series, Y : UnivariateSeries)
```
and checks/conversions being done automatically.


fkiraly:
pinging @chrisholder since otherwise I can't add him reviewer

fkiraly:
note: this fails with some of the forecasters that are not yet refactored, since it can happen that `fit` is overridden but `predict` is not, in which case the variables written in `fit` are not addressed.

I see two ways to address this:

* wait until the re-factoring #955 is done
* add some safety checks that turn off the new code in `predict` if it wasn't called in `fit`

fkiraly:
I've found a better option no.3 - initialize the `fit` variables with "safe" defaults that are what they would be for all of the forecasters that haven't been refactored. We can change that later, though it might be in general a good idea to have these "safe" defaults.

fkiraly:
ready for review

mloning:
Hi @fkiraly, I had a quick look now, this seems to be a major API change, will need some more time to look into it.

fkiraly:
sure - since it's a dev days priority, shall we try to get any questions and change requests out of the way soon-ish? Meeting later today?

aiwalter:
what about renaming `y` to `Y`? Is it postponed or decided if we want to do it in general? See also discussion here #967 
___

fkiraly:
>what about renaming `y` to `Y`? Is it postponed or decided if we want to do it in general? See also discussion here #967

I'd say variable renaming is a very invasive core interface change and will affect many files.

Therefore, it's best kept separate from *this* change, which is *not* breaking core interface contracts (as it works only on the inside).

Due to the breakage of the core interface, I think it's also worth considering not changing `y` to `Y` at all, ever.

aiwalter:
I think it could work fine, if we depreciate it properly so that users are aware of it.

fkiraly:
hm, what's going wrong with manylinux?

fkiraly:
@aiwalter, re. y/Y - let's do this in a different PR.

Following the principle that a PR should either be localized in interface or localized in scope/logic. I.e., it can be broad in terms of logic while affecting interfaces minimally, or being broad in terms of the files/places it affects while having a minimal logic scope, but not both, unless it really can't be avoided.

aiwalter:
yes sure, this would be a different PR. I was just curious :)

mloning:
Is this a blocker for anyone at the moment? If I remember correctly, in our sprint planning, the goal for this was to come closer to an API design. I certainly need more time to think about this. It seems like a big change from the more familiar scikit-learn interface and I'm not sure about the consequences. Perhaps we can have a discussion session on this?

fkiraly:
>It seems like a big change from the more familiar scikit-learn interface and I'm not sure about the consequences. Perhaps we can have a discussion session on this?

How is that? Not sure whether I understand why you think this.

As you see the tests are all running, and - in addition - all outputs are the same when compared to pre-PR behaviour.
All this adds is the ability to input and receive data types that previously would have raised errors.

fkiraly:
>Is this a blocker for anyone at the moment? 

I feel it's *crucial* to get to our self-set goal of multivariate forecasting by end of week.

mloning (commented):
I think this PR introduces two separate things:
* support for multivariate forecasting
* type conversions 

I'm less sure about the type conversion. It introduces a new module for type conversion, a new non-boolean tag, there aren't any tests yet, and I'm not really sure what the consequences are. Any thoughts @aiwalter @RNKuhns? If everyone is convinced of the type conversions at this stage, I'm happy to make a few comments on the code and then get it merged. 

I think the easy way to introduce support for multivariate forecasting is to work with the existing `check_y` function and give it an argument for whether it handles univariate or multivariate series based on a boolean tag.

fkiraly:
I think the type conversion is necessary to have multivariate forecasting under a compatible interface.

Why: otherwise we have to decide per-estimator whether we want it to rely on `pd.DataFrame` or `pd.Series`. This either/or seems a bit unnecessary and contrived, since it would split the forecaster interface down the middle.

>I think the easy way to introduce support for multivariate forecasting is to work with the existing check_y function and give it an argument for whether it handles univariate or multivariate series based on a boolean tag.

See above. This would introduce a lot of boilerplate code.

Also note that we are not talking about hypotheticals here, but a working prototype in the PR, which does not change existing interfaces, adds the capability to pass `pd.DataFrame` and `np.array` to forecasters (as attempted in #897 by more manual hacking of the `check_xyz` etc), and makes any future multivariate forecaster downwards compatible.

fkiraly:
So, to summarize our discussion at the dev days:

There's no consensus on merging this - @mloning does not like this, based on arguments that are perhaps best explained by himself here. If I were to paraphrase it:

* @mloning proposes an alternative solution that, instead of converting, just throws an error if one of the formats is not supported
* this would mean we will have forecaster which support `np.array`, `pd.Series`, or `pd.DataFrame`, or combinations thereof. For example, if you pass a `pd.Series` to a forecaster that supports `pd.DataFrame`, it would not be converted, but an error would be thrown, and the user will have to know that they have to convert the `pd.Series` to a one-column-`pd.DataFrame` if their data is univariate

In addition, @mloning proposed that we work through all forecasters and make sure they support both `pd.Series` and `pd.DataFrame` both; hoping that we can use joint interface points where possible, and otherwise make any necessary conversions inside individual forecasters, and not automated in the base class.

I disagree with this way forward, since:

* working through individual forecasters has nothing to do with base class behaviour - i.e., error throw or convert
* it's better to convert to the user expected type rather than throw an error and expect the user to make the conversion easily
* complex pipeline composites are much less likely to work out of the box
* @mloning's suggestion would lead to an "interface split" between `pd.Series` and `pd.DataFrame` forecasters going forward with multivariate implementations, along lines that are not obvious to users

fkiraly:
To get this involved, I would be invoking our governance document, which we wisely wrote in anticipation of a situation like this, to professionally get to a democratic resolution 😃 
https://github.com/alan-turing-institute/sktime/blob/main/GOVERNANCE.rst

As per "decision making" section, @mloning (or another core contributor) should now either, asap:

* veto this PR, based on STEP05, to formally express disagreement
* not veto this PR, and agree that this is merged soon

If a veto happens, core devs have 5 days (excluding week-ends) to look at this with care and cast a public vote, here in the PR. Deadline would be those 5 days after the veto.

Lovkush-A:
I wrote up my view of the discussion in the hackmd file for the forecasting group of dev days: https://hackmd.io/84rpzQEiQiWQy5oQjs7FYw?both. The summary can be found at the bottom of the hack md.

I highly recommend having a look at it, because it presents a different perspective on ML's views than what FK describes above. (And of course, ML is the best person to say what his perspective is)

TonyBagnall (commented):
I'm coming to this quite cold, so does this capture the issue?

The question is how to handle different data format  inputs. The solutions are:

1. delegate this to the subclass forecasters, throwing exceptions if they do not have the ability to process the specific data type

2. make the base class convert all formats into a type determined by the objects tags via the method convert_to, and require all subclasses to work with the type specified with their particular tags 

is this (broadly) correct?

Lovkush-A:
@TonyBagnall. I think that is broadly correct!

But, from my understanding, the fundamental disagreement is more about time. Markus did not say he outright disliked option 2, but he wanted us to take more time to consider any downsides thoroughly, whereas Franz does not think there are any downsides that require significant time and is concerned that if we do not merge it quickly, it will just continuously get pushed back. Hence (again from my understanding) Franz invoked the governance document to ensure a timely decision is made.

fkiraly:
@Lovkush-A, it's not a good idea to post a hackmd link here when the doc is still editable. A random person or malicious bot from the internet can access this, and, for instance, replace links with malicious links.

Can whoever owns this make the hackmd doc read-only? I'll put the content on the wiki in the meanwhile.

fkiraly:
>is this (broadly) correct?

In my view, yes, @TonyBagnall.

However, the discussion with @mloning convinced me that this may be sub-optimal, becaues it does not allow individual instances of estimators to accept multiple types, e.g., when it can handle series and pandas by only using joint intervace points (like the imputer).

What I would hence like to change this to, ultimately, is a tag that is a "pass" list. If the input is of a type on the list, `fit` passes the input to `_fit` unchanged. If the tag is not on the list, it converts it to the first element that's on the list.

fkiraly:
@Lovkush-A, I think this of yours is a very interesting and relevant point:

>Thought by LA later, this sounds like a reduction, of multivariate forecasting to univariate forecasting. This suggests a reduction style interface, where user creates multivariate forecaster using reduction function and a univariate forecaster object. Separate to this, can have reductions from multivariate forecasting to tabular regression

I think you indeed identify univariate forecasters correctly as a different scitype than multivariate forecaster.

The subtle point is here that I think it's a sub-scitype, and the inclusion is too "weak" to justify a separate base class.

What do I mean with sub-scitype: multivariate forecasters have the same sci-interface as univariate forecasters whenever presented with a univariate series. When presented with a multivariate series, only multivariate forecasters can do this.

Other examples of this "sub-scitype" relation are:
* binary and multi-class classification
* transformers that have an inverse transform vs those that don't
* supervised regressors that can deal with multivariate target vs those that can't

Note that neither of these have distinct base classes in `scikit-learn`, since the interfaces are too close (I reckon) - a key design principle of `scikit-learn` is to avoid the anti-pattern of (base) class proliferation. That's what I mean with "weak" sub-scitype - the added complexity by covering both cases in one base class is small.

Another interesting point: even if you have the same base class, you still can do reduction, it just doesn't mutate the general interface, it only removes the "lack of capability".
Examples are:

* the ["one-vs-rest-classification" reduction strategy](https://scikit-learn.org/stable/modules/generated/sklearn.multiclass.OneVsRestClassifier.html), which creates a multi-class capable classifier out of a binary one
* the ["apply per output variable" reduction strategy](https://scikit-learn.org/stable/modules/generated/sklearn.multioutput.MultiOutputRegressor.html) which makes a multi-target regressor out of any regressor

I think univariate forecasting vs multivariate forecasting is precisely such a case, most similar to the multi-output regression example.

TonyBagnall:
> What I would hence like to change option no.2 to, ultimately, is a tag that is a "pass" list. If the input is of a type on the list, `fit` passes the input to `_fit` unchanged. If the tag is not on the list, it converts it to the first element that's on the list.

so maintain a list of allowable inputs. You will still have to prioritise conversions, so if it can take two types and pass a third, which of the two it allows would be chosen? 

have you mocked up the use cases in an example base and subclass forecasters? That may help convince others?
1. Standard use case 1: passed a type the subclass can handle unary and many to many
2. Standard use case 2: passed a type the subclass cannot handle, where subclass can handle only one type. Convert and pass
3. Edge case (?):  passed a type the subclass cannot handle, where subclass can handle more than one type. Convert and pass

Lovkush-A:
> @Lovkush-A, it's not a good idea to post a hackmd link here when the doc is still editable.

My bad... I have removed the link from my comment. The description should be enough information for people in this group to find the link. (But I suppose there is still the small risk the link has already been scraped...)

> The subtle point here is that it's a sub-scitype, and the inclusion is too "weak" to justify a separate base class.

Makes sense to me. Note we do not need any separate base classes to still have a reduction style interface. We can still have something like `MakeMultivariate(UnivariateForecaster(), reduction_strategy)`

fkiraly:
>My bad... I have removed the link from my comment.

You would't have needed to, I've changed it to a static wiki page.

fkiraly:
To answer your questions, @TonyBagnall 

>so maintain a list of allowable inputs. You will still have to prioritise conversions, so if it can take two types and pass a third, which of the two it allows would be chosen?

I'd say there is always a "preferred" type if it can take many, that's what conversion would be attempted to. Easy to implement by making the `y_type` field being a list; check containedness, if yes pass unconverted; if not, convert to first element.

>have you mocked up the use cases in an example base and subclass forecasters? That may help convince others?

This PR implements the functionality already (one internal type only).

It does not change behaviour that currently does not throw an error - all tests pass - your case no.1. It's basically all of the current tests, and the example sheets.

It adds functionality where users can pass more inputs - to *all* forecasters. - your case no.2.

regarding no.3: there currently does not exist, to my knowledge, a forecaster that can handle more than one case. To my knowledge, @mloning's discussion of this case is purely hypothetical given the current code base.

fkiraly:
>Makes sense to me. Note we do not need any separate base classes to still have a reduction style interface. We can still have something like MakeMultivariate(UnivariateForecaster(), reduction_strategy)

@Lovkush-A, absolutely agreed! That's what #1038 should do.

TonyBagnall:
so I have just discussed this with my research group, and not wanting to muddy the water, but one suggestion (from chris, supported by aaron) was it could be better to achieve this effect with decorators, along the lines of below (wont let my upload a note book). He can comment more himself if this is of any interest. 

import numpy as np
import pandas as pd


def support_df(estimator_class):
    
    estimator_fit = estimator_class.fit
    
    def fit(self, X):
            print("Converting df to numpy")
            X = X.to_numpy()
        return estimator_fit(self, X)
    
    estimator_class.fit = fit
    return estimator_class

def support_list(estimator_class):
    
    estimator_fit = estimator_class.fit
    
    def fit(self, X):
        if isinstance(X, list):
            print("Converting list to numpy")
            X = np.array(X)
        return estimator_fit(self, X)
    
    estimator_class.fit = fit
    return estimator_class

@support_list
@support_df
class Example:
    
    def __init__(self, name: str):
        self.name = name
        
    def fit(self, X):
        return isinstance(X, np.ndarray)


numpy_example = np.array([1,2,3,4,5])
df_example = pd.DataFrame({"A": [1, 2], "B": [3, 4]})
list_example = [1,2,3,4,5]

numpy_example_cls = Example("numpy")
print(numpy_example_cls.fit(numpy_example))

print(numpy_example_cls.fit(df_example))

print(numpy_example_cls.fit(df_example))

fkiraly:
>was it could be better to achieve this effect with decorators, along the lines of below (wont let my upload a note book). He can comment more himself if this is of any interest.

Yes, we've also discussed decorators extensively, see the original [STEP05](https://github.com/sktime/enhancement-proposals/tree/main/steps/05_scitype_based_IO_checks) as well.

The problem with that lies with the *extender user archetype* (not the notebook/deployment user archetype):

* the extender user would have to remember to add/modify decorators. That's similarly error prone as remembering to having to add the checks. It's better to keep "info about the method" localized to tags/capabilities, just one table to fill in, just one failure point, not multiple.
* the functionality is not out-of-the-box, i.e., you need to add the decorators
* for an extender user, decorators are much harder to debug than the nested function calls, since it's a more advanced programming concept, and obscures the traceback

This is why I believe @mloning was against decorators, and I agreed with him.

chrisholder:
I would argue (and potentailly this isn't valid as ML is very different to web dev) that Flask, which is considered one of the most user friendly web development packages in Python; part of the reason it is seen as so user friendly is because of its use of decorators. Not only is it super simple to use, but it's also self documenting. Here is an example:

```
from flask import Flask, url_for
from markupsafe import escape

app = Flask(__name__)

@app.route('/')
def index():
    return 'index'

@app.route('/login')
def login():
    return 'login'
```

Just looking at this code with no knowledge of the application you know just from the decorator there are two GET methods for the route '/' and '/login'.

I also think using a decorator approach much better and consistently allows people to support their own custome datatypes. Lets say we support numpy and dataframe like so as a base in sktime.
`
@dtype_support.dataframe
@dtype_support.numpy
class SomeClassifier
    def __init__():
    ..... etc
`

If someone then comes along and wants to use a pytorch tensor or something, if all the logic is in the base class they're going to have to add a check to the top of their fit() that accounts for it, which is then confusing because if you were just looking at the code I'd wonder where are the rest of the checks?

By using decorators it still allows users to define their method to handle custom datatypes, but allows them to use it consistent manner to the rest of sktime. So for example:

We define our own custom tensor handler data type.
`
def dtype_support_tensor(estimator_class):
    estimator_fit = estimator_class.fit
    
    def fit(self, X):
        if torch.is_tensor(X)
            X = torch_tensor.cpu().detach().numpy()
        return estimator_fit(self, X)
 
    estimator_class.fit = fit
return estimator_class
`

And then add it to their model in a consistent manner that makes sense.

`
@dtype_support.dataframe
@dtype_support.numpy
**@dtype_support_tensor**
class SomeClassifier
    def __init__():
    ..... etc
`

My biggest concern with having to many decorators as you can see above 3 is already getting a bit ugly, so you can group ones togethers. So for example you can group a bunch together say we just define 'dtype_support.univariate' and 'type_support.multivariate' and these just encompass everything that @fkiraly (dataframes, numpy etc.) wants and then when looking at our code base it's super easy to determine what the model supports.
`
@dtype_support.multivariate
class SomeClassifier
    def __init__():
    ..... etc
`

fkiraly:
>If someone then comes along and wants to use a pytorch tensor or something.
>This is now confusing because if you were just looking at the code wouldn't you wonder where are the rest of the checks are?

No, not at the top of `fit` - you wouldn't touch that.

To add a new supported data type, you would add it in the converters module.

Once it is "registered" there, you can make use of pytorch tensors, and potentially write individual methods that use them natively (without conversion); while all already implemented methods would be able to use pytorch tensors by conversion (potentially inefficient, of course, but it runs).

>This is much more consistent and just by glancing at my class I know it supports numpy, dataframe and tensor. It also doesn't clutter the body of the class.

In this design, you see whether the algorithm supports xyz by querying a tag - that's consistent with querying any other property.
In your design:

* you have at least two sources of information about the forecaster: tags, and decorators. The decorators cannot be queried by the convenient `all_estimators` mechanism, but you have to sift through the code to find all estimators that can do `numpy`, say.
* it is not clear which sequence to add the decorators. It is also not clear how the decorators would make sure that support is automatically added? The internal logic seems speculative, and potentially highly brittle.

>and then when looking at our code base it's super easy to determine what the model supports.

but that's not what the "user archetype" is willing to do! They want to call the `all-estimators` method, say, to be able to query all estimators with a property. They will for sure not go manually through all code files and look at the decorator block.

fkiraly:
>This is not a great way of doing it I admit but the point being you can do what you like in the decorator and therefore there is definitly a solution for this.

@chrisholder, I agree that decorators are an attractive way of doing this - see the original STEP05 which was written in terms of decorators.

But, as said, the main argument against (brought up by @mloning and which I now follow) is the difficulty for the extender user archetype. The extension guidelines become more difficult to follow than with this suggestion, in my opinion. The less clear interface for inspection (tags vs decorators) is also an argument for the other solution, in my opinion.

fkiraly:
So, to speed up this discussion: I think the community and core devs have had ample time to look at this - it's the longest thread in a while if you exclude code annotations; we've also had a dedicated 1+ hour meeting at the dev days. I'd explicitly brought up the opportunity for veto 2 working days ago. I'd say this plus having one more day next week satisfies the "reasonable time" requirement for the "lazy consensus".

So, if no one vetos by end of Monday, I'll merge this. FYI, @mloning - please veto if you don't like this.

fkiraly:
(updating from main to prepare for merge - checks should pass but let's wait)

mloning:
@fkiraly In the governance doc it says:
* "To accept proposed changes, we require approval by one core developer (lazy consensus) and no rejection by a core developer (veto right)." - Currently there isn't an even a single approval yet? So why do you think you can merge? What's worse is that you know that there is disagreement. I think your comments above are not helpful at all. Note that on GitHub, we have we even require two approvals (I think that needs to be updated in the governance doc).
* "Core developers are expected to give reasonable time to others to give their opinion on the pull request if they’re not confident others would agree." Given that this PR is only 7 days old, that I have asked for more time repeatedly and that the dev days were happening, I don't think that you're giving reasonable time to others.
* "When no consensus can be found, any core developer can call for a vote at any point during the discussion." So seems like you can just call a vote. The more important point is that I'd like someone else to review it too. 

I will take some more time now to look at the PR, read through the discussions above and write some comments on it.

mloning:
I've been thinking about the proposed changes in the PR over the last few days. I feel that the discussion has been a bit tense at times, so I'll try to write down my reasons here for why I'm hesitant about this PR. 

In general, I think that the proposed solution in this PR is a bit over-engineered: it solves a problem at a level of generality which isn't necessary for supporting multivariate forecasting in sktime. My preference is to work with the current (less general but simpler) input validation/conversion functions.

I think it's a bit over-complicated because we only want to support a few internal data types for the target series. That is, `pd.Series` for univariate forecasting and `pd.DataFrame` for multivariate series (supporting pandas data containers of course also gives access to the underlying NumPy arrays). I don't think we want to support other types in the univariate/multivariate forecasting module, or do we?

I believe that supporting these types internally can be solved more easily with the existing input validation/conversion functions (e.g. `check_y`) and keyword arguments which can be set automatically in the base class based on tags. For example, in `fit` in the base class, we could call (inside `check_y_X`):
```python
y = check_y(
    y, 
    enforce_univariate=has_tag(self, "univariate-only"),
    enforce_multivariate=has_tag(self, "multivariate-only"))
```
with one new boolean tag `multivariate-only` (as suggested by @aiwalter in some other PR) and a few more lines in `check_y` to handle the new keyword arguments. I could open a PR to see if that works in practice. 

The solution in this PR adds:
* non-boolean tags (not even string-valued tags, but type-valued tags, possibly with sets of types)
* a new module for conversions, with functions relying on machine types, scitypes and type-valued tags
* possibility to support multiple output types 

My main worries are the following: 
* these changes may not be obvious for developers/users familiar with the PyData ecosystem, to me it departs significantly from scikit-learn in terms of input validation/conversion,
* in order to understand how the data passes through the estimators (e.g. in inspecting, extending or debugging code), users and developers now have to have to be aware of: 
    * the internal type (which may no longer be the same for all forecasters)
    * non-boolean tags
    * scitypes 
    * the various conversion functions
* non-boolean tags and scitypes are not self-explanatory at all I think. So it may make estimators less transparent and less extendible. In the worst case, undoing much of the work of the recent refactoring.
* I also don't think that we want to support multiple output types. Wouldn't that complicate higher-level functionality such as composition (e.g. ensembling), or model evaluation (e.g. `evaluate`), as they would now have to handle different output types (we have seen the complications of that with the current `pred_int` interface)?


Do we want to support more internal types than the ones above? My preference is to stick to `pd.Series` and `pd.DataFrames`. We have seen that supporting multiple internal data types can lead to costly data copies, making it difficult to write efficient composition models, since each model may have to convert first the data it receives from the previous model (we have seen this in the classification setting). 

To me, the proposed type conversions seems more relevant for the panel data setting, especially for user convenience, but even there I wouldn't want to support more internal types than we currently support if avoidable, with the exception of awkward array perhaps. But I would be happy to revisit the idea in that context. 
Also note that panel and supervised forecasting probably requires a different API altogether (multiple stages of fit), so may not directly build on the univariate/multivariate forecasting API discussed in this PR.

If this PR finds broader approval from others, I'm happy to support the PR, review the code in more detail and suggest improvements (I agree with @lovkush-A that some minor things can still be improved).

Let me know what you think and if my reasons make sense. I'm happy to discuss them in more detail and I'm open to other arguments.

sktime/forecasting/base/convertIO/_convertIO.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+-----------------------------------------
+
+Function signature of
+convert[(from_type, to_type, as_scitype)]
+
+Parameters
+----------
+what : from_type - object to convert
```
Lovkush-A, line 24: strongly dislike name `what`.

Some alternatives: `object` (used in python docs for generic inputs, e.g. in `isinstance`), `data` (because I think we will always be using this function to convert data, as opposed to other kinds of things)

Is there a standard 'formal' name for the kind of thing that `y`, `X`, `Z` etc. are?

sktime/forecasting/base/convertIO/_convertIO.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+__author__ = ["fkiraly"]
+
+
+import numpy as np
+import pandas as pd
+
+
+# dictionary indexed by triples of types
```
Lovkush-A, line 68: these comments potentially redundant, given explanation at the top of the file?

sktime/forecasting/base/convertIO/_convertIO.py
```diff
@@ -0,0 +1,190 @@
...
+#  1st element = convert from - type
+#  2nd element = convert to - type
+#  3rd element = considered as this scitype - string
+# elements are conversion functions of machine type (1st) -> 2nd
+
+convert = dict()
+
+
+def convert_identity(what, store=None):
```
Lovkush-A, line 77: minor preference for just `identity`. `convert_identity` sounds like identity is being converted.

But I realise `identity` is not a verb... I feel like `identity` is such a standard name of a function that the verb convention can be ignored in this instance

sktime/forecasting/base/convertIO/_convertIO.py
```diff
@@ -0,0 +1,190 @@
...
+    return what
+
+
+# assign identity function to type conversion to self
+for tp in [pd.Series, pd.DataFrame, np.array]:
+    convert[(tp, tp, "Series")] = convert_identity
+
+
+def convert_UvS_to_MvS_as_Series(what: pd.Series, store=None) -> pd.DataFrame:
+
```
Lovkush-A, line 88: needs docstring?

sktime/forecasting/base/convertIO/_convertIO.py
```diff
@@ -0,0 +1,190 @@
...
+    if not isinstance(what, pd.Series):
+        raise TypeError("input must be a pd.Series")
```
Lovkush-A, line 90: I wonder why you wrote `input must be a pd.Series` instead of `what must be a pd.Series`...

😛

sktime/forecasting/base/convertIO/_convertIO.py
```diff
@@ -0,0 +1,190 @@
...
+
+    if isinstance(store, dict) and "cols" in store.keys and len(store["cols"]) == 1:
```
Lovkush-A, line 92:
```suggestion
    if isinstance(store, dict) and "cols" in store.keys() and len(store["cols"]) == 1:
```

Lovkush-A, line 92: * checking that the length of columns is 1 might be redundant? in what situation would this particular convert function be called and the checking of length be important?
* secondly, if length of stored columns is greater than 1, then it suggests something has gone wrong somewhere (I think), and maybe we should raise an error?

sktime/forecasting/base/convertIO/_convertIO.py
```diff
@@ -0,0 +1,190 @@
...
+
+    return what
+
+
+# assign identity function to type conversion to self
+for tp in [pd.Series, pd.DataFrame, np.array]:
+    convert[(tp, tp, "Series")] = convert_identity
+
+
+def convert_UvS_to_MvS_as_Series(what: pd.Series, store=None) -> pd.DataFrame:
```
Lovkush-A, line 87: maybe this should be called `convert_Series_to_Pandas_as_Series`? Do we know for certain that pandas frames will be the only machine type for MvS?

sktime/forecasting/base/convertIO/_convertIO.py
```diff
@@ -0,0 +1,190 @@
...
+convert[(pd.Series, pd.DataFrame, "Series")] = convert_UvS_to_MvS_as_Series
+
+
+def convert_MvS_to_UvS_as_Series(what: pd.DataFrame, store=None) -> pd.Series:
+
+    if not isinstance(what, pd.DataFrame):
+        raise TypeError("input is not a pd.DataFrame")
+
+    if len(what.columns) != 1:
+        raise ValueError("pd.DataFrame must be pd.DataFrame with one column")
```
Lovkush-A, line 109: Don't you mean "what must be a pd.DataFrame with one column"? 😛 This is too easy!

TonyBagnall:
All I will say at this point is that it is a complex issue in code, in engineering and in toolkit strategy. I have discussed it in my group and we were not sure. This haste, trying to force a decision on a Saturday, is I think inappropriate. There is no actual urgency, and it seems like impatience is creating the impression of Franz trying to steam roller compliance or confrontation. Consensus is better. Let's talk next week, I don't want this issue to damage team relationships.

sktime/forecasting/base/convertIO/_convertIO.py
```diff
@@ -0,0 +1,190 @@
...
+
+    if isinstance(store, dict):
+        store["cols"] = what.columns[[0]]
```
Lovkush-A, line 112: I have slight preference for using `columns` rather than `cols`. (because I have general preference to avoid abbreviations and also in this case it would be consistent with pandas)

sktime/forecasting/base/convertIO/_convertIO.py
```diff
@@ -0,0 +1,190 @@
...
+
+    return pd.DataFrame(what).to_numpy()
+
+
+convert[(pd.Series, np.array, "Series")] = convert_UvS_to_np_as_Series
+
+
+def convert_np_to_MvS_as_Series(what: np.array, store=None) -> pd.DataFrame:
+
+    if not isinstance(what, np.array) and len(what.shape) != 2:
```
Lovkush-A, line 147: just checking that you have explicitly chosen to exclude 1D numpy arrays?

It is feasible a user tries to use a multivariate forecaster to do univariate forecasting with a 1D numpy input, and the current form of code would not allow that. I have no strong preference either way, but thought it is checking if this is what you intended.

sktime/forecasting/base/convertIO/_convertIO.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+# conversion based on queriable type to specified target
+def convert_to(what, to_type: type, as_scitype: str, store=None):
+
+    from_type = type(what)
+
+    key = (from_type, to_type, as_scitype)
+
+    ckys = list(convert.keys())
```
Lovkush-A, line 183: Dislike`ckys` as variable name; abbreviation is not necessary. Could probably just skip the creation of the variable anyway and have `if key not in convert.keys()`

fkiraly:
>In general, I think that the proposed solution in this PR is a bit over-engineered: it solves a problem at a level of generality which isn't necessary for supporting multivariate forecasting. My preference is to work with the current (less general but simpler) input validation/conversion functions.

Disagreed, in the sense that supporting multivariate forecasting will be cumbersome, unintuitive, or both, without this. It will lead to writing of boilerplate code and user frustration with unclear interfaces.

There's a clear motivation to avoid this, so I disagree that this is "overengineered" - it is fit-for-purpose and elegant in the sense that it makes all existing forecasters ready for the multivariate extension, without breaking interface contracts.

>I believe that supporting these types internally can be solved more easily with the existing input validation/conversion functions (e.g. `check_y`) and keyword arguments which can be set automatically in the base class based on tags. For example, in fit in the base class, we could call (inside `check_y_X`):

As said above: this would essentially split the forecasters into "`pd.DataFrame` only`" and a "`pd.Series` only" sub-types. Or it will require extensive rework of all forecasters manually. This is not desirable, so it is not a good suggestion.

>these changes may not be obvious for developers/users familiar with the PyData ecosystem, to me it departs significantly from scikit-learn in terms of input validation/conversion,

The changes are "under the hood", and both the standard user archetype (in `fit`, `predict` etc) as well as the extender user archetype (in `_fit`, _predict`) will deal with the familiar interfaces. We also avoid user frustration through rejection of `pd.Series` or `pd.DataFrame` on unclear/uninspectable criteria. I don't see a major or "significant" divergence here.

>users and developers now have to be aware of: 

I think your list incorrectly frames `sktime` design principles as problems, or incorrectly "blames" existing problems of `sktime` on this PR. Let me go through your list.

>the internal type (which may no longer be the same for all forecasters)

It *already isn't* so this is not a change this PR introduces. Search for occurrences of the `enforce_numpy` flag, we *already* have `numpy` vs `pandas` internally, and you can't blame this on this PR.

>scitypes

this is a key design principle of `sktime`, and advanced users should already be familiar with this. See, e.g., [this paper](https://arxiv.org/abs/2101.04938) we wrote together. Or, search the code for the string "scitype".

This is nothing that this PR introduces - in fact, it follows core `sktime` design principles.

>the various conversion functions

Look at `utils/data_processing` - there are a dozen or so from/to conversion functions (for panel time series data) in there, inconsistently named and with inconsistent function signatures. Developers are expected to understand this mess. This is i.m.o. ludicrous, instead there should be `convert` function, see my new PR #1061.

Here's the list of the converters for your edification:
```python
    from_3d_numpy_to_2d_array,
    from_3d_numpy_to_nested,
    from_nested_to_2d_array,
    from_2d_array_to_nested,
    from_nested_to_3d_numpy,
    from_nested_to_long,
    from_long_to_nested,
    from_multi_index_to_3d_numpy,
    from_3d_numpy_to_multi_index,
    from_multi_index_to_nested,
    from_nested_to_multi_index,
    convert_from_dictionary,
```

>non-boolean tags

This is indeed new, but you agreed in #1013 that non-boolean tags (e.g., string tags) are fine, and they are also needed in your alternate suggestion above.

>non-boolean tags and scitypes are not self-explanatory at all I think. So it may make estimators less transparent and less extendible. In the worst case, undoing much of the work of the recent refactoring.

Are they more explanatory than the mess of converters? I think yes.
Is a single tag that lists the data types supported more explanatory than 5 boolean tags? I think yes.
Does it "undo" the recent refactoring? No, because that was done with the purpose to put all the boilerplate stuff in the base forecaster, and not have all the checks and conversions in 100 places in the concrete classes, precisely as this PR suggests.

>I also don't think that we want to support multiple output types.

Disagreed - I think the user wants a `pd.Series` forecast when they pass a `pd.Series`, and a `pd.DataFrame` forecast when they pass a `pd.DataFrame`. Saying "no" in case a forecaster supports multivariate is a source of user frustration. Same for `numpy.array`s.

>I also agree with @chrisholder that decorators would help here, especially if we want to handle output types, as it keeps the conversion functionality outside of the code of the class and things like converter_store separate from the other class attributes.

Perhaps, but that's an orthogonal point (see #973 and #974) - whether the conversions are in a decorator or explicit in the base class does not touch the point of conversions and simultaneous support of univariate and multivariate data container types.

Lovkush-A (changes requested):
* I like this design and approach and would be in favour of PR being merged (after some changes). I have not considered the downsides, but the elegance of the idea makes me think there are unlikely to be big downsides. But, I also do not want to rush other people into merging the PR, and would be more comfortable knowing that the downsides have been thought through thoroughly by others.

* Most of my inline suggestions and comments are, surprise surprise, about variable names. But I do it because I think it makes big difference to future readability and changing them in the first PR is best place to suggest the changes.

Some other things:
* there should be some basic unittests before any PR is merged.
* might need to consider the `.name` attribute of a pd.Series object as being a candidate for dataframe column name. if this is a concern, then id personally be ok with this being turned into an issue for future consideration
* in discussion, you (Franz) say that you think having lists of types for `y_type` tag might be better. I agree. Personally id be ok with this being turned into an issue for future consideration/adjustments

fkiraly:
>This haste, trying to force a decision on a Saturday, is I think inappropriate. There is no actual urgency, and it seems like impatience is creating the impression of Franz trying to steam roller compliance or confrontation. Consensus is better. Let's talk next week, I don't want this issue to damage team relationships.

@TonyBagnall, sorry if there was a misunderstanding - I'm not forcing a decision on a Saturday. I do not expect anyone to work on week-ends, and I have said that either someone should veto, or accept that this is merged on Monday (end of day).

That gives enough time to decide whether to veto this.
Once there is a veto, there are 5 more working days to think this through, I think that's ample time.

My fear is that we keep the multivariate issue indefinitely off the table, due to decision making paralysis because none of the potential solutions are perfect. We have been discussing the data container issues since the beginning of `sktime`, and there is no silver bullet. Perfect is the enemy of good, so if we want to get somewhere with multivariate, we need to stop stalling and get going.

And, of course, friendly working relations in the team should never be in question here. I do appreciate the work that @mloning and you, @TonyBagnall, are doing; I respect and value you as persons and collaborators.

sktime/forecasting/base/convertIO/_convertIO.py
fkiraly: >just checking that you have explicitly chosen to exclude 1D numpy arrays?

No, but I did not want to complicate an already difficult discussion.
With this design, they would be easy to add, but let's not mix the "should we support 1D numpy arrays" discussion into this. See also #897 for the corresponding conversion.

sktime/forecasting/base/convertIO/_convertIO.py
fkiraly: Yes, I know. Happy to change this, will do later on once the discussion is resolved

sktime/forecasting/base/convertIO/_convertIO.py
fkiraly: that sounds like a weird riddle and not like an error message

sktime/forecasting/base/convertIO/_convertIO.py
fkiraly: >Do we know for certain that pandas frames will be the only machine type for MvS?

no, already the panel data conversions module allows numpy, so this may be something to align later on. I did not want to complicate things too much.

>maybe this should be called

It's supposed to be private and not called by the user. The only public exports are `convert` and `convert_to`

sktime/forecasting/base/convertIO/_convertIO.py
```diff
@@ -0,0 +1,190 @@
...
+
+-----------------------------------------
+
+Function signature of convert_to
+
+Parameters
+----------
+what : object to convert (any type)
+to_type : type, the type to convert "what" to
+as_scitype : str - name of scitype the object "what" is considered as
```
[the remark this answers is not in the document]
fkiraly: see #1061, the third argument is `"Panel"` there.
We have `pd.DataFrame` and `np.array` appearing there, and even multiple different formats per machine type.

sktime/forecasting/base/convertIO/_convertIO.py
```diff
@@ -0,0 +1,190 @@
+# -*- coding: utf-8 -*-
+"""
+Machine type converters for scitypes
+
+Exports
+-------
+convert : dictionary of functions, with signature as below
```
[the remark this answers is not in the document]
fkiraly: how about writing a wrapper function `convert` where `convert(args)` is an alias for what is now `convert[(args)]`)?

sktime/forecasting/base/_base.py
```diff
@@ -68,6 +72,11 @@ def __init__(self):
         self._fh = None
         self._cutoff = None  # reference point for relative fh
 
+        self.converter_store = dict()  # storage dictionary for input/output conversions
+
+        # "safe" initialization in case fit is overridden
```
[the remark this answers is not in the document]
fkiraly: `fit` should not be overwritten anymore, according to the new extension guidelines.

The safe initialization is for downwards compatibility currently.
Going forward, we may want to think of a set of informative error messages if extender users overwrite `fit` accidentally.

What do you propose as a concrete change?

fkiraly:
@Lovkush-A, to reply to your high-level comments:

>there should be some basic unittests before any PR is merged.

Good point - the current unit test suite tests downwards compatibility, but not the new cases where the input/output data type is flexible, or multivariate cases.
This will be a bit of a project and may also involve some refactoring of the current test suite, so I'll wait for resolution of this discussion.

> might need to consider the .name attribute of a `pd.Series` object as being a candidate for dataframe column name. if this is a concern, then id personally be ok with this being turned into an issue for future consideration

Good idea - in the converters, I guess? Or is there another place this would be relevant?

> in discussion, you (Franz) say that you think having lists of types for `y_type` tag might be better. I agree. Personally id be ok with this being turned into an issue for future consideration/adjustments

Yes, I think I've already decided I'd add this once the discussion is resolved positively.

sktime/forecasting/base/_base.py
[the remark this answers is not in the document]
Lovkush-A: I see! Just to check, you have this safe initialisation because the refactoring of forecasters is not yet complete, and you intend to remove it once the refactoring is complete?

sktime/forecasting/base/convertIO/_convertIO.py
fkiraly: `lovkush_variable_name_suggestions.append(this)`

sktime/forecasting/base/convertIO/_convertIO.py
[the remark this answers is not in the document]
Lovkush-A: I think that suggestion is overly complex. Renaming seems simpler and clearer to me.

sktime/forecasting/base/convertIO/_convertIO.py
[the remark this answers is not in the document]
Lovkush-A: Wow - that is one thorough PR!

I'll have a closer look at that later, but do you know an example of the top of your head in which we have convert_dict(from_type, to_type, scitype1) != convert_dict(from_type, to_type, scitype2)?

sktime/forecasting/base/convertIO/_convertIO.py
Lovkush-A: `lovkush_variable_name_suggestions` implies I suggested a variable name, whereas in this instance, I suggested removing the variable altogether. I think you need a new variable name to contain all my suggestions. 😆

sktime/forecasting/base/convertIO/_convertIO.py
Lovkush-A: You either misunderstood my suggestion or are joking...and I can't tell which

Lovkush-A:
> [Unittests] will be a bit of a project and may also involve some refactoring of the current test suite, so I'll wait for resolution of this discussion.

I was only suggesting *very* basic checks for this PR - e.g. that the converter functions you created run without error on simple inputs - as opposed to testing how this interacts with everything else. 

> > might need to consider the .name attribute of a `pd.Series`
> 
> Good idea - in the converters, I guess? Or is there another place this would be relevant?

My first instinct is it only matters in converters.

fkiraly:
All, I looked at this again after having talked to @mloning and @TonyBagnall in the CC meeting and getting feedback on how this discussion felt from @mloning's perspective.

I think I did forget my good manners, regarding tone. My enthusiasm for multivariate and impatience might have gotten the better of me. Very sorry, @mloning.

I do think that disagreement, and even passionate disagreement, is a core ingredient of seeking scientific truth and engineering insight, but it needs to happen in a civil and friendly manner. I'll give it my best.

Regaring the force-merge, that would not be in line with the rules of governance, since as per reasonable reading, the PR requires at least one approval for that. I could call a vote without such approval in-principle, but I've decided not to do this right now, in the interest of giving core developers more time to review (and potentially approve or ask for amendments).

fkiraly:
FYI, for those following this discussion: I think the new PR #1074 is meant as a counterproposal to this one.

Lovkush-A:
My thoughts on Marcus' concerns above.

> it solves a problem at a level of generality which isn't necessary for supporting multivariate forecasting.

To me, this is the big attraction! It is not just about multivariate forecasting, but a general framework for all of sktime. E.g. one of somebody's 'frustrations' in the final dev days meeting (not to do with forecasting) would be straightforwardly solved by this input/output framework. I can't remember the example, but you might remember I said 'That would be solved by Franzs input/output suggestion' and Franz said 'I was just about to say the same thing'

> I don't think we want to support other types in the univariate/multivariate forecasting module, or do we?

Unlikely no, but this input/output framework makes it easy to adapt the interface if we do want more (*or less*) supported types


> I believe that supporting these types internally can be solved more easily with the existing input validation/conversion functions (e.g. `check_y`)

That might be true. On the other hand, Franz has already coded this up and it does not look too complex.
 
> * non-boolean tags (not even string-valued tags, but type-valued tags, possibly with sets of types)

If this is an issue, we could make all the tags strings, and have a function that parses the string. E.g. instead of `{y_inner_type: set([pd.Series, pd.DataFrame])}` could have `{y_inner_type: "series;dataframe"}` and have function `lambda tag: set(tag.split(";"))`

> * a new module for conversions, with functions relying on machine types, scitypes and type-valued tags

I like this. It means all conversions are in one place in the codebase, as opposed to being split up in various check functions spread out across sktime.

> * possibility to support multiple output types (not sure if I have understood this correctly?)

I think that is part of the intended design. If a user inputs a dataframe for univariate forecasting, we convert it into series, do the forecasting, then convert it back to a dataframe.

> * these changes may not be obvious for developers/users familiar with the PyData ecosystem

For me, the main difficulty for developers/users is understanding the actual mathematics and subtleties of time series, rather than the interface.

> to me it departs significantly from scikit-learn in terms of input validation/conversion

Yup, certainly a disadvantage. But I think it is a departure in a good way. E.g. I find it annoying in sklearn that I have to manually convert things back to dataframes, and think about whether index and/or column information has been lost. (Though, maybe the issue is I should be using Pipelines more...)


> * in order to understand how the data passes through the estimators (e.g. in inspecting, extending or debugging code), users and developers now have to be aware of:
>   * the internal type (which may no longer be the same for all forecasters)
>   * non-boolean tags
>   * scitypes
>   * the various conversion functions

I think most of these are fundamental issues of having a comprehensive time series package.

> * I also don't think that we want to support multiple output types. Wouldn't that complicate higher-level functionality such as composition (e.g. ensembling), or model evaluation (e.g. `evaluate`)

This is certainly worth thinking about! I don't know enough about these items to judge.

mloning:
It looks like @fkiraly and I have come to an agreement, you can find our suggestion below. Feedback is very welcome! And thanks again for all the above comments:

* which input types do we want to support for which scitype?
    * univariate/multivariate series: pd.Series, pd.DataFrame, np.ndarray?
    * FK opinion: there should just be one concept "series" - pd.Series, pd.DataFrame, 2D np.ndarray, 1D np.dnarray
        * np.ndarray are interpreted as axis 0 indexed by integers
* which internal types do we want to support?
    * FK opinion: internally (in `_fit`, `_predict` etc *without* conversion), we should have "tiers" of support
        * low tier: pd.Series or pd.DataFrame (uni/multi)
        * mid tier: support *both* pd.Series and pd.DataFrame
        * high tier: supports all of these interally - target state
* which output types do we want to support? 
    * same as internal types?
    * FK opinion: 
        * default is type in = type out
        * type out can be controlled by tag (override), this is useful when building pipelines, so conversions can happen only at start/end of long pipeline - push towards pandas internally
        * no strong opinion about type in = type out
        * ML does not like this
    * ML opinion:
        * target state: output should always be pd.DataFrame
            * FK would be happy with this
        * short/mid term: output should be pd.Series for all "univariate" tagged estimators, pd.DataFrame for all "multivariate" tagged estimators, input converted to internal format (usually pd.Series for "univariate")
            * FK is happy with this
* how do we want to indicate whether a forecaster handles univariate data, multivariate data or both? 
    * tags?
    * FK suggestions if we insist on bool:
        * `scitype:multivariate` - can handle multivariate series
        * `requires_y_multivariate` - cannot handle univariate series
    * univariate-only, multivariate-only
    * `scitype:y` {"univariate", "multivariate", "both"}
        * univariate: error if dim>1
        * multivariate: error if dim<2
        * both: no error from dim
    * `scitype:y` {`"univariate"`, `"multivariate"`, `"multi_but_not_univariate"`}
    * agreed on:`scitype:y` {"univariate", "multivariate", "both"}
* edge cases
    * fbprophet which is univariate but requires pd.DataFrame
    * combined uni/multi forecasters 
* unit testing
    * new tags: test forecasters valid string values for `y:scitype` tag {"univariate", "multivariate", "both"}
        * maybe just do this for all tags? registry could also record expected values
    * input conversion: accepts all accepted inputs: pd.Series, pd.DataFrame, 2D np.ndarray, 1D np.dnarray
    * simply add "variants" of valid inputs to input checks
    * univariate/multivariate check: error should be raised if dimension too high/low - 2d np.ndarray and pd.dataframe on one side, 1d np.ndarray on other side

fkiraly:
with minor comment, we aim for "mid tier" internal type support only, for the next half year or so.

so as steps, I would now:
* remove or turn off the output checks, since the solution *assumes* (at least for short term) that uni/multivariate forecasters return the "right thing"
* add the new tag and concomitant checks
* add or modify tests to check multi-input-type compatibility and the tag working properly

Lovkush-A:
Sounds good to me. Does that mean @markus' draft PR is being closed and work will continue on this PR?

mloning:
Yes, I closed #1074. 

@fkiraly let me know if you need help with the implementation.

fkiraly:
I'm on it, just need to deal with `BaseObject` first since this one needs tags.

sktime/forecasting/base/convertIO/_convertIO.py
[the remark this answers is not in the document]
fkiraly: `(np.ndarray, pd.DataFrame, Series)` and `(np.ndarray, pd.DataFrame, Panel)`

sktime/forecasting/base/convertIO/_convertIO.py
fkiraly: well, it's not the identity function, it has two arguments

sktime/forecasting/base/convertIO/_convertIO.py
fkiraly: I'm not sure why this would be funny

sktime/forecasting/base/convertIO/_convertIO.py
fkiraly: yes

sktime/forecasting/base/convertIO/_convertIO.py
fkiraly: what do you mean?

fkiraly:
@mloning, @aiwalter , @TonyBagnall, I think this is now ready to review.
The one failing test is due to the newly enabled doctests, which insists I fill in all the docstrings for the tests because I just changed on line.

This should implement the agreed plan, with minor departures:
* I also added conversions for `X`, while I was at it.
* I changed some of the check functions to accommodate for `np.ndarray` as discussed - e.g., `check_equal_time_index`
* I change the check functionality to allow for `None` (e.g., when `X=None`) - the general philosophy is to ignore `None` series and let them pass.
* I replaced forecaster specific checks with scitype specific checks where possible, e.g., `check_X` by `check_series`. I think this looks much cleaner and much more general now.
* I introduced an `mtype` function which returns a string representation of the machine type. This is to avoid type valued tags, and will also help later when there are two different representations of the same machine type.
* I removed the output conversions by explicitly switching them off rather than removing them entirely, using the constant setting `OUTPUT_CONVERSIONS = "off"`. I think that's a neat compromise for the other side of the "python in python out" ideological fence.

sktime/forecasting/base/convertIO/_convertIO.py
fkiraly: also, it's private, no one sees it except if you specifically look for it

sktime/forecasting/base/convertIO/_convertIO.py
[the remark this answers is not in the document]
Lovkush-A: This conversation is now resolved; original and followup questions all answered and no changes needed.

sktime/forecasting/base/convertIO/_convertIO.py
Lovkush-A: The error message you raise is `input must be a pd.Series`. I was wondering (in a tongue-in-cheek way), why you did not use the error message `what must be be a pd.Series`, given that the variable name is `what`.

Less cryptically, I am implying that `what must be a pd.Series` is a confusing error message, and thus, `what` is a confusing variable name.

sktime/forecasting/base/convertIO/_convertIO.py
Lovkush-A: > > maybe this should be called `convert_Series_to_Pandas_as_Series`
> 
> It's supposed to be private and not called by the user. The only public exports are `convert` and `convert_to`

You misinterpreted how I used the word `called` - I meant it in the sense of `named` or `labelled`, whereas you interpreted it as a call to a function/method. I couldn't tell if you genuinely misunderstood or if you were trying to make a pun.

sktime/forecasting/base/convertIO/_convertIO.py
fkiraly: oh, I see - what would be a better name than what?

Lovkush-A (commented):
Just some minor typos, I think.

sktime/forecasting/base/convertIO/_convertIO.py
Lovkush-A: I made a couple of suggestions in a comment above, https://github.com/alan-turing-institute/sktime/pull/980#discussion_r659212321. Copied here:

Some alternatives:
* `object` (used in python docs for generic inputs, e.g. in isinstance)
* `data` (because I think we will always be using this function to convert data, as opposed to other kinds of things)

To help think of ideas, is there a standard name for the kind of thing that y, X, Z etc. are?

extension_templates/forecasting.py
```diff
@@ -68,9 +68,11 @@ class MyForecaster(BaseForecaster):
     # todo: fill out estimator tags here
     #  delete the tags that you *didn't* change - these defaults are inherited
     _tags = {
-        "requires-fh-in-fit": True,  # is forecasting horizon already required in fit?
+        "scitype:y": "univariate",  # which y are fine? univariate/multivariate/both
+        "univariate-only": True,  # does estimator use the exogeneous X?
         "handles-missing-data": False,  # can estimator handle missing data?
-        "univariate-only": True,  # can estimator deal with multivariate series y?
+        "y_inner_type": "pd.Series",  # which types do _fit, _predict, assume for y?
```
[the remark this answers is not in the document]
fkiraly: `scitype:y` refers to the number of columns `y` can have, this is a scitype descriptor. `y_inner_mtype` refers to the machine types that `_fit` can cope with.

Note that a `pd.DataFrame` can have 1 column, and that would be ok for an estimator with `scitype:y == "univariate"`.

sktime/forecasting/base/convertIO/_convertIO.py
```diff
@@ -0,0 +1,335 @@
...
+""" key objects in this section:
+
+convert: dict indexed by triples of str
+  1st element = convert from - str
+  2nd element = convert to - str
+  3rd element = considered as this scitype - str
+elements are conversion functions of machine type (1st) -> 2nd
+
+Function signature of all elements
+convert[(from_type, to_type, as_scitype)]
```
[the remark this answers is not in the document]
fkiraly: yes, fixed

sktime/forecasting/base/convertIO/_convertIO.py
fkiraly: ah yes, you are right. Not sure what I was thinking there.

sktime/forecasting/base/_base.py
```diff
@@ -107,7 +116,16 @@ def fit(self, y, X=None, fh=None):
 
         self._set_cutoff(y.index[-1])
 
-        self._fit(y=y, X=X, fh=fh)
+        self.y_in_type = type(y)
+
+        y_inner = convert_to(
```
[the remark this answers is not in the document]
fkiraly: inner is shorter and easier to read, that´s why I prefer it.

fkiraly:
from verbal discussion with @mloning, @mloning reviewed this and did not like two things:

* that output conversions were still implemented, just turned off - he thought the option should be removed entirely. That I highlighted as deviation from the agreed plan. @mloning's argument here is the unnecessary code complexity when looking at `fit`/`predict`. I concurred with this and agreed to remove entirely.
* that I didn't remove the tag for the internal machine type, `y_inner_mtype` after our discussion. @mloning thought I deviated from the agreement here, though it didn't occur to me that this should have been removed (see list above). @mloning thought that the inner type should be fully determined by the `y:scitype` tag, and some translation between the `y:scitype` tag and the `convert` functionality should happen in `fit` and `update`. @mloning thought this was a strong point of contention. I disagreed that the tag should be removed, since I didn't think it was part of the agreement to change this part, and it would add the "glue code" to `fit`/`update` as unnecessary complexity in there.

Overall, we agreed that I can make the call to merge this PR - for instance, we can have detail discussions on the necessary tags later, and see it as a delta to this PR rather than a blocker for this PR.

The call I made is as follows:

* I remove the output conversions entirely, in line with the spirit of our agreement above, and the argument of code complexity.
* I'm keeping `y_inner_mtype` for the moment, to keep the conversion logic and the content of `fit`/`update` simpler; also, in consideration of the time series classification case and facebook `prophet` where the supposed 1:1 mapping of inner type and scitype is broken. Also, in consideration of potentially slowly adding `geomstats` like data container functionality to `sktime`, and development activity where it's not entirely clear yet whether this mapping would be easy to adhere to.

FYI, @aiwalter, @TonyBagnall.

sktime/forecasting/base/_base.py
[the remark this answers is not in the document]
fkiraly: superseded, `y_in_type` is now entirely removed since we removed output conversions.

sktime/forecasting/base/_base.py
```diff
@@ -107,7 +116,16 @@ def fit(self, y, X=None, fh=None):
 
         self._set_cutoff(y.index[-1])
 
-        self._fit(y=y, X=X, fh=fh)
+        self.y_in_type = type(y)
```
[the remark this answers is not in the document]
fkiraly: superseded, removed with output conversions

sktime/forecasting/base/convertIO/_convertIO.py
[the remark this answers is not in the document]
fkiraly: changed to `convert_dict`

sktime/forecasting/base/convertIO/_convertIO.py
```diff
@@ -0,0 +1,190 @@
...
+Exports
+-------
+convert : dictionary of functions, with signature as below
+        indexed by triples of types
+         1st element = convert from - type
+         2nd element = convert to - type
+         3rd element = considered as this scitype - string
+        elements are conversion functions of machine type (1st) -> 2nd
+
+convert_to : function/dispatch version of convert, with 1st elt inferred
```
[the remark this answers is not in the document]
fkiraly: couldn't think of sth better, sorry

sktime/forecasting/base/convertIO/_convertIO.py
fkiraly: changed to `obj` consistently, docstrings read much better now

sktime/forecasting/base/convertIO/_convertIO.py
fkiraly: dict moved out of module exports (no longer exported), so no longer redundant

sktime/forecasting/base/convertIO/_convertIO.py
fkiraly: it's `obj` now, that's the arg name that appears in other python type related things like `isinstance`

sktime/forecasting/base/convertIO/_convertIO.py
fkiraly: bug fixed, thanks

fkiraly: >in what situation would this particular convert function be called and the checking of length be important?

when the columns in the store, for some reason, have the wrong length, the constructor with `columns=` throws an error

sktime/forecasting/base/convertIO/_convertIO.py
fkiraly: discussion resolved with Markus, and renaming done

sktime/forecasting/base/convertIO/_convertIO.py
fkiraly: >Do we know for certain that pandas frames will be the only machine type for MvS?

I really hope so, but if we add another one, we can simply call it different from MvS.

sktime/forecasting/base/convertIO/_convertIO.py
fkiraly: nothing to resolve since clarification

sktime/forecasting/base/convertIO/_convertIO.py
Lovkush-A: You have given two opposite answers to the question, "Do we know for certain that pandas frames will be the only machine type for MvS?" But I don't think it requires any immediate changes to the code though

> no, already the panel data conversions module allows numpy, so this may be something to align later on. I did not want to complicate things too much.

> I really hope so, but if we add another one, we can simply call it different from MvS.

Lovkush-A:
@fkiraly Great effort, Franz! This is one monster of a PR.

I have seen all your comments/resolutions to my review conversations and am happy with decisions made. There was one conversation in which you gave two opposite answers to the same question (whether MvS is always pandas frames or not), but I do not think that is actionable in this PR.

fkiraly:
FYI, @mloning: I also made changes to the test which parallel yours in #1074 - the `enforce` functions were not working properly.

fkiraly:
@GuzalBulatova, @thayeylolu, this is now merged - and it should be useable in your work on multivariate forecasters.

PR #1003, opened 2021-06-21
Refactors polynomial trend forecaster

thayeylolu:

- [x] PolynomialForecaster


sktime/forecasting/trend.py
```diff
@@ -69,14 +74,11 @@ def fit(self, y, X=None, fh=None):
         -------
         self : returns an instance of self.
         """
-        self._is_fitted = False
 
         if X is not None:
```
fkiraly, line 78: while we're here, let's just remove this error raised, see #1014

fkiraly (changes requested):
Looks great!

This is a full refactor and well done, but I think we can easily also address #1013 in this PR, by removing the superfluous error right with it.

sktime/forecasting/trend.py
mloning: Agreed, probably legacy code before we had agreed on how to handle exogenous data.

thayeylolu: Hello @fkiraly , Is there something I am suppose to resolve here? I cannot seem to see it.

fkiraly: I meant, remove the `NotImplementedError` being thrown when `X` is passed. The problem is described in #1014.

This is not strictly part of the refactoring, but let's do this while we're already here.

thayeylolu: noted

fkiraly (approved):
looks ready to go, let's just wait for the checks to finish.

thayeylolu:
@mloning this is ready for review :)

PR #1005, opened 2021-06-21
Refactor Prophet

thayeylolu:

- [x] FProphet

### DIFF
--- a/sktime/forecasting/base/adapters/_fbprophet.py
+++ b/sktime/forecasting/base/adapters/_fbprophet.py
@@ -10,15 +10,20 @@ import os
 import pandas as pd
 
 from sktime.forecasting.base._base import DEFAULT_ALPHA
-from sktime.forecasting.base._sktime import _OptionalForecastingHorizonMixin
-from sktime.forecasting.base._sktime import _SktimeForecaster
+from sktime.forecasting.base import BaseForecaster
 from contextlib import contextmanager
 
 
-class _ProphetAdapter(_OptionalForecastingHorizonMixin, _SktimeForecaster):
+class _ProphetAdapter(BaseForecaster):
     """Base class for interfacing fbprophet and neuralprophet"""
 
-    def fit(self, y, X=None, fh=None, **fit_params):
+    _tags = {
+        "univariate-only": True,
+        "requires-fh-in-fit": False,
+        "handles-missing-data": False,
+    }
+
+    def _fit(self, y, X=None, fh=None, **fit_params):
         """Fit to training data.
         Parameters
         ----------
@@ -32,11 +37,9 @@ class _ProphetAdapter(_OptionalForecastingHorizonMixin, _SktimeForecaster):
         -------
         self : returns an instance of self.
         """
-        self._is_fitted = False
         self._instantiate_model()
         self._check_changepoints()
         self._set_y_X(y, X, enforce_index_type=pd.DatetimeIndex)
-        self._set_fh(fh)
 
         # We have to bring the data into the required format for fbprophet:
         df = pd.DataFrame({"y": y, "ds": y.index})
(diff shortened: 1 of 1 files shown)


thayeylolu:
@fkiraly  Removing `self._set_y_X(y, X, enforce_index_type=pd.DatetimeIndex)` fails test with time_index. It is currently not removed. What do you suggest is the way forward?

fkiraly:
hm, interesting - I would introduce a tag called `enforce_index_type`.
If this tag exists, the value is passed to `check_y_X` in the base class.

The second line is done in the base class, the first (addition of tag) in the concrete class.

fkiraly:
I've suggested addition of this as a tag in #1013.

Once that is agreed and merged, you could set the `enforce_index_type=pd.DatetimeIndex` tag to the tags, and modify the base forecaster that the tag is passed to `check_y_X`.

fkiraly:
hm, I guess a temporary solution is to leave the conversion inside, as you did.

Opinions, @mloning?

fkiraly (approved):
happy with this to be merged as is, assuming we'll deal later with the `enforece_index_type` tag

sktime/forecasting/base/adapters/_fbprophet.py
```diff
@@ -10,15 +10,20 @@
...
 from contextlib import contextmanager
 
 
-class _ProphetAdapter(_OptionalForecastingHorizonMixin, _SktimeForecaster):
+class _ProphetAdapter(BaseForecaster):
     """Base class for interfacing fbprophet and neuralprophet"""
 
-    def fit(self, y, X=None, fh=None, **fit_params):
+    _tags = {
+        "univariate-only": True,
```
aiwalter, line 21: why is this univariate only? `fbprophet` is can deal with exog data? also `neuralprophet` which might inherit from this class if s.o. writes a wrapper.

aiwalter (commented):
I left one comment @thayeylolu

sktime/forecasting/base/adapters/_fbprophet.py
thayeylolu: Thank you @aiwalter. I would edit it soon

fkiraly: yes, good point!

Can prophet also do multivariate?

If it can, I would say: let's restrict ourselves to refactoring here, and let's make another issue that addresses the multivariate case after the multivariate extension.

aiwalter: afaik it can do only use `X`, but not do vector forecast. S.o. would have to check this. If it can forecast also `Y`, this would be definitely a separate PR, but changing the tag for now to accept `X` should be possible I guess

mloning (dismissed):

sktime/forecasting/base/adapters/_fbprophet.py
fkiraly: I thought the "univariate" tag was about `y`, not `X`

thayeylolu: @aiwalter  does s.o. mean so?

aiwalter: hm, I think `univariate-only` is about if `X` is used or not. Where are those tags described?

aiwalter: s.o. = someone

thayeylolu: So, the conclusion is to leave it as `True`

aiwalter: you can make it `True` just now, as we know this is the case

thayeylolu:
Hello @aiwalter . I would like to suggest leaving the univariate only tag as True for the meantime. then, when we resolve the tag issues, we can go ahead and set it as False

aiwalter (approved):

fkiraly (approved):
happy with this - @aiwalter changed the tags by direct commit, so I think they are now good? And we agreed to change them later potentially in a new PR.

Issue #957, opened 2021-06-13
[ENH] Define forecaster tags


fkiraly:
We should introduce tags for the following properties of forecasters:

- [x] can make multivariate forecasts
- [x] can make use of exogeneous data (doesn't just ignore it)
- [x] can deal with non-equally spaced time series
- [x] assumes that time stamps of exogeneous data are a super-set of endogeneous data
- [x] can deal with variables sampled at different time stamps
- [ ] can deal with exogeneous data that are not known throughout the future

All of these properties should map on input checks which raise informative errors.

We should also think of standard adapters that turn the restrictive version of the tag in the non-restrictive one.

For example:

* a wrapper that turns algorithms into multivariate versions by iterating over variables in `y`. There are two options here: consider the other `y` variables as exogeneous or not (latter should be the default).
* a wrapper that uses an interpolator to reduce to the equally spaced case; this needs to have transform (re-sampling equally) and inverse transform (interpolating between equally spaced forecasts)
* removing the requirement to see `fh` in `fit` by moving `fit` into `predict` at the price of inefficiency

fkiraly:
done except for the last, see `registry._tags`

We may like to introduce th last one if we have examples of atomic forecasters of this kind.

PR #1013, opened 2021-06-21
default tags in BaseForecaster; added some new tags

fkiraly:
I've added some default tags in `BaseForecaster` - these are automatically inherited via `_all_tags`.

I've also added two new tags that I anticipate we will need:
* `"X-y-must-have-same-index"` - do `X` and `y` need to have the same index?
* `"enforce-index-type"` argument to `check_X`/`check_y` if needed - this has to be a tag now, see the problem in #1005 .

i've also added these new tags to the extension template.

This PR also:
* corrects a minor bug: mis-spelled tags in the extension template
* adds the new tags to the allowed tags in tests
* removes the requirement for tags to be boolean


mloning (commented):
As discussed, I think this can go in as an intermediate step if it is helpful in the overall refactoring. Afterwards, we need to revisit the tags (see #957).

fkiraly:
ok - will merge once the tests pass

PR #1015, opened 2021-06-22
Refactors ensembler, online-ensembler-forecaster and descendants

thayeylolu:

- [x] Online Ensembler Forecaster , Descendant and Ensembler

### DIFF
--- a/sktime/forecasting/online_learning/_online_ensemble.py
+++ b/sktime/forecasting/online_learning/_online_ensemble.py
@@ -23,6 +23,11 @@ class OnlineEnsembleForecaster(EnsembleForecaster):
     """
 
     _required_parameters = ["forecasters"]
+    _tags = {
+        "univariate-only": True,
+        "requires-fh-in-fit": False,
+        "handles-missing-data": False,
+    }
 
     def __init__(self, forecasters, ensemble_algorithm=None, n_jobs=None):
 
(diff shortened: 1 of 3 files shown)


sktime/forecasting/online_learning/_online_ensemble.py
```diff
@@ -47,15 +52,12 @@ def fit(self, y, X=None, fh=None):
...
-        self._set_y_X(y, X)
-        self._set_fh(fh)
         names, forecasters = self._check_forecasters()
         self.weights = np.ones(len(forecasters)) / len(forecasters)
         self._fit_forecasters(forecasters, y, X, fh)
-        self._is_fitted = True
         return self
 
-    def _fit_ensemble(self, y, X=None):
+    def fit_ensemble(self, y, X=None):
```
mloning, line 60: This should still be private, no?
```suggestion
    def _fit_ensemble(self, y, X=None):
```

thayeylolu:
Could you point me to where you are referring to @fkiraly

mloning (commented):
Hi @thayeylolu, we also need to update the `update` method by removing the following I think: 
```python
        self.check_is_fitted()
        self._update_y_X(y, X)
```

sktime/forecasting/online_learning/_online_ensemble.py
fkiraly: yes, why was this changed?

fkiraly:
>Could you point me to where you are referring to @fkiraly

the `update`/`_update` function, currently line 59.

sktime/forecasting/online_learning/_online_ensemble.py
thayeylolu: Must have been an error

fkiraly (approved):
looks ready to merge, I'll just wait for the checks to finish.

PR #1016, opened 2021-06-22
refactors PmdArimaAdapter

thayeylolu:

- [x] PmdArimaAdapter and descendants ARIMA, AutoARIMA


fkiraly (approved):
looks good to me. I guess `univariate-only` should be `False`, but we can deal with this later.

PR #1017, opened 2021-06-22
Refactors TbatAdapter

thayeylolu:

- [x] BATS, TBATS, and _Tbatsadapter


thayeylolu:
@mloning Kindly review this too

fkiraly (approved):
looks good to me.
Maybe we want to change the `univariate-only` tag (or not?), but I'd leave that either to @aiwalter or @mloning to directly commit, or later PR.

aiwalter (approved):
`tbats` package deals only with univariate, so its fine.

PR #1019, opened 2021-06-22
Added two new related software packages

aiwalter:


fkiraly (approved):
all those `sktime` clones :-)

we may like to re-order the list a little - e.g., is the package an individual algorithm (e.g., `greykite`) or a framework? Not in this PR though.

aiwalter:
Yes, some are already described as "toolkit" and some just as "library"... We need a definition then however what is what.

mloning (approved):

Issue #993, opened 2021-06-20
Refactor classifiers for fit/_fit model


TonyBagnall:
adopt the forecasting interface model for base abstract classifier so that rather than have abstract fit, have base fit call abstract _fit. This means that the validation of the input can be called in the base classifier. 

![image](https://user-images.githubusercontent.com/9594042/122681397-10d7d780-d1ec-11eb-95c3-449f865cb753.png)

However, for the classifiers this will require a reworking of capabilities, so that they are overridden and not shadowed. This is not a bad thing, but may well make it hard to roll capabilities into sklearn tags (which Im not sure I want to do in any case). 

Tony

TonyBagnall:
needs to be done in conjunction with the refactor of tags/capabilities https://github.com/alan-turing-institute/sktime/issues/981

TonyBagnall:
I'm going to close this for now until the tags issue is resolved with forecasting

PR #1021, opened 2021-06-22
Add series annotation and PyOD adapter

fkiraly:
Update: this is now split into two PR, this one for `BaseSeriesAnnotator` (renamed from `BaseStremAnnotator` and one for `BasePanelAnnotator` (PR #1036).

This also includes the `pyod` adapter from #1022, and hence introduces a `pyod>=0.8.0.` soft dependency.

---

This is an extension of @satya-pattnaik's excellent annotation base template #838 that should allow sub-classing or other treatment of non-point-annotation and multi-sample/panel annotation use cases as not uncommon in neuroscience and beyond.

I've created two base classes:

* `BaseStreamAnnotator` which follows most closely @satya-pattnaik's original design
* `BasePanelAnnotator` which extends the design to panel annotation tasks

I've thought long after the first proposal (of mine, i.e., the first commits below) and I think it's not a good idea I had after all, to "merge" the characteristics of both into one. This becomes most clear when you look at what the `update` mode should be - temporal, or adding more sequences?

Changes in both classes:
* rename of series to be annotated from `Z` to `X` - the user expectation is `predict(X)` not `predict(Z)`
* `X` is by default a `pd.DataFrame`, not `pd.Series`. Annotation is typically multivariate, and this is in compliance with the target interface for forecasting.
* adopted nested `fit/_fit` architecture from `forecasting` refactor - it's consensus we're moving towards this, see also #993
* additional, optional argument in `fit`: `Y`, ground truth annotations - for supervised annotation
* allow annotations to be of any `pandas` `dtype`, this allows complex annotations like BIDS (see sktime neuro https://github.com/sktime/sktime-neuro/issues/7)

Additions in the `BaseStreamAnnotator`:

* `update` and `update_predict` functionality, so we can use it in time series streams and directly with forecasters

Additions in `BasePanelAnnotator`:
* `X` can be a list of sequences to be annotated (all `pd.DataFrame`)
* additional, optional argument in `fit` and `predict`: `Z`, `pd.DataFrame` of same `len` as `X` - any metadata on the sequences to annotate.


fkiraly:
pinging @satya-pattnaik

fkiraly:
@mloning, @satya-pattnaik, I've made a major change since my first commit.

I've meditated about this, and I think we should have two annotator base types - one for panels, one for streams. This splits most naturally by the update mode and use of the argument `Z` (yes/no)

sktime/annotation/base/_base.py
```diff
@@ -0,0 +1,435 @@
...
+
+    The base annotator specifies the methods and method
+    signatures that all annotators have to implement.
+
+    Specific implementations of these methods is deferred to concrete
+    annotators.
+    """
+
+    # default tags
+    _tags = {
```
mloning, line 70: Let's add tags in a separate PR once we have agreed on them.

sktime/annotation/base/_base.py
```diff
@@ -0,0 +1,435 @@
...
+        "handles-panel": True,  # can handle panel annotations, i.e., list X/y?
+        "handles-missing-data": False,  # can handle missing data in X, Y, Z
+        "annotation-type": "point",  # can be point, segment or both
+        "annotation-labels": "none",  # can be one of, or list-subset of
+        #   "label", "outlier", "change"
+    }
+
+    def __init__(self):
+
+        self._is_fitted = False
```
mloning, line 80: Doesn't this happen in the `BaseEstimator`?

sktime/annotation/base/_base.py
```diff
@@ -0,0 +1,435 @@
...
+
+        super(BasePanelAnnotator, self).__init__()
+
+    def fit(self, X, Y=None, Z=None):
```
mloning, line 84: Based on @Lovkush-A comment, I suggest:
```suggestion
    def fit(self, Z, Y=None, X=None):
```

mloning (commented):
I generally agree with the interface refactoring. I also prefer the separation into different kinds of annotators based on input data (e.g. panel and streaming).  

Here are a few suggestions:
* move panel annotators to separate PR, focus this PR on the series annotators, I believe the panel annotation warrants a bit more discussion, ideally in a separate PR. 
* rename `StreamingAnnotator` to `SeriesAnnotator`, this corresponds to the input type and is contrasted with `PanelAnnotator`, also note that panel data may be streaming data. 
* defer implementation of tags until we have agreed on them and can test them. 

I'm happy to merge this quickly, even if we keep a few things open in this PR (e.g. type of annotations, sparse vs dense) and then iterate over it.

sktime/annotation/base/_base.py
fkiraly: I disagree - I think this should be close to:

* the supervised setting as in `scikit-learn` - we use `fit` and `predict`
* the endogeneous/exogeneous labelling, where information also flows `X`->`Y`

fkiraly: also, I think you annotated the wrong annotator? This is the panel annotator, which moved to #1036

sktime/annotation/base/_base.py
fkiraly: ok, I commented them out for now

sktime/annotation/base/_base.py
fkiraly: based on the suggestion in #1023, I introduced a `transform` method which asliases `predict`.

The `transform` method is compliant with the interface in our transformers, while `predict` mirrors forecasting and `scikit-learn`.

sktime/annotation/base/_base.py
fkiraly: yes, but no harm done? It's the "safe" thing to do, also looking at the impending `BaseObject` refactor here #891.

sktime/annotation/adapters/_pyod.py
```diff
@@ -0,0 +1,113 @@
+# -*- coding: utf-8 -*-
+import numpy as np
+from sktime.annotation.base._base import BaseSeriesAnnotator
+
+__author__ = ["mloning", "satya-pattnaik", "fkiraly"]
+
+import pandas as pd
+
+
+class SeriesAnnotatorPyOD(BaseSeriesAnnotator):
```
mloning, line 10: Let's follow the naming conventions that we use for transformers already (estimator scitype at the end, ignoring subtypes)
```suggestion
class PyODAnnotator(BaseSeriesAnnotator):
```

sktime/annotation/adapters/_pyod.py
```diff
@@ -0,0 +1,113 @@
...
+        "annotation-labels": "outlier",  # can be one of, or list-subset of
+        #   "label", "outlier", "change"
+    }
+
+    def __init__(
+        self, estimator, annotation_format="sparse", annotation_values="indicator"
+    ):
+
+        if annotation_format not in ["sparse", "dense"]:
+            raise ValueError('annotation_format must be "sparse" or "dense"')
```
mloning, line 39: * Perhaps also worth shorting them to `format` and `values` since the annotation context is clear. 
* Should `values` be renamed to `labels`? Values may be a bit generic, labels will be familiar from the classification setting. 
* I suggest to write helper functions for input checks, e.g. `_check_format(format)`, and add them to `sktime/utils/validation/annotation.py`

sktime/annotation/adapters/_pyod.py
```diff
@@ -0,0 +1,113 @@
...
+        ------------
+        creates fitted model (attributes ending in "_")
+        """
+
+        X_np = X.to_numpy()
+
+        if len(X_np.shape) == 1:
+            X_np = X_np.reshape(1, -1)
+
+        self.estimator.fit(X_np)
```
mloning, line 75: Since estimator is technically a hyper-parameter we don't want to overwrite it here (none of the methods should overwrite hyper-parameters except the constructor and `set_params`), so instead we usually do:
```suggestion
        self.estimator_ = clone(self.estimator)
        self.estimator_.fit(X_np)
```
with trailing underscores indicating fitted attributes.

sktime/annotation/base/_base.py
```diff
@@ -0,0 +1,256 @@
...
+        """
+
+        X = check_series(X)
+
+        self.update(X=X)
+        Y = self.predict(X)
+
+        return Y
+
+    def transform(self, Z, Y=None):
```
mloning, line 179: If we want to treat this as a transformer, should it not also inherit from one of the transformer types? It may be easier to implement this as a wrapper.

sktime/annotation/base/_mock.py
```diff
@@ -0,0 +1,37 @@
+# -*- coding: utf-8 -*-
+from sktime.annotation.base._base import BaseSeriesAnnotator
+
+
+class MockSeriesAnnotator(BaseSeriesAnnotator):
```
mloning, line 5: Since we have the `PyODAnnotator` in this PR, I think we can remove the mock class

sktime/annotation/tests/test_all_annotators.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
```
mloning, line 1: can we add more tests on the output, i.e. if format=dense, it should have the same index as X

sktime/utils/_testing/estimator_checks.py
```diff
@@ -129,12 +132,7 @@ def check_estimator_tags(Estimator):
     assert hasattr(Estimator, "_all_tags")
     all_tags = Estimator._all_tags()
     assert isinstance(all_tags, dict)
-    assert all(
-        [
-            isinstance(key, str) and isinstance(value, bool)
-            for key, value in all_tags.items()
-        ]
-    )
+    assert all([isinstance(key, str) for key, _ in all_tags.items()])
```
mloning, line 135: Are the changes to the tag system necessary for this PR already?

mloning (commented):
Thanks @fkiraly and @satya-pattnaik looks very good already, I left a few comments below, should be quick to fix, let me know if you have any questions!

sktime/annotation/base/_base.py
fkiraly: what would we get from inheriting?

sktime/annotation/adapters/_pyod.py
fkiraly: do you know whether this works with `pyod` models out-of-the-box?

sktime/annotation/adapters/_pyod.py
fkiraly: >since the annotation context is clear.

yes, but the context that this is something that applies to the output is not clear.
How about `y_format` etc?

fkiraly: >I suggest to write helper functions for input checks,

With this design, would that not be identical with `check_series`?

fkiraly: >Values may be a bit generic, labels will be familiar from the classification setting.

Yes, good idea

sktime/annotation/adapters/_pyod.py
fkiraly: sure - generally I'd prefer the scitypes at the start, since then the lexicographic ordering clusters scitypes, but obviously this is the wrong place to change that. I'll comply with current standard.

fkiraly:
Any good idea where this is coming from?
` ValueError: Invalid estimator type: <class 'sktime.annotation.base._mock.MockSeriesAnnotator'>`

I suspect it's coming from incorrect registration of the estimator in some test config, but can't track it down easily.

sktime/annotation/tests/test_all_annotators.py
fkiraly: hm, whould we move the `format` arguments into the base class and pass them via `super`? That way it would be an expected interfacve point there, and can be read in the base class for checks.

sktime/annotation/base/_base.py
fkiraly: maybe, but do we get anything from the wrapper that isn't already addressed by this interface alias?

sktime/utils/_testing/estimator_checks.py
fkiraly: no, but to revert this I would have to go through multiple files at the cost of time and the risk of breaking things - it was already a hassle to split the PR into two

sktime/annotation/adapters/_pyod.py
fkiraly: it already did `check_series` when you wrote the above, so can you please explain what you would want on top of that?

fkiraly: >should values be renamed to labels

there actually already is a variable that is called "labels", it's the one which tells you the type of labels

sktime/annotation/base/_base.py
fkiraly: this was inconsistent with the `fit`, so I added aliasing (of `X` and `Z`) to all `fit`-like methods

sktime/annotation/adapters/_pyod.py
mloning: I would think so, they're closely aligned with scikit-learn, but worth checking.

fkiraly: seems to work - good

sktime/annotation/base/_base.py
fkiraly: I actually removed that agin. I think now it was a bad idea.

We may want to have some aliasing or not, but first we should refactor the transformers modules.
I decided this is out of scope for this PR, probably you agree.

sktime/annotation/base/_mock.py
fkiraly: will do once the `PyODAnnotator` works

satya-pattnaik (commented):
Really nice work @fkiraly . Will need to add more unit tests to it. Let me know if you have some in your mind. 
Also to remove the mock annotator.

sktime/annotation/tests/test_all_annotators.py
mloning: Yes, moving them to `super` sounds good to me

sktime/annotation/adapters/_pyod.py
mloning: For readability, I'd write:
```python
def check_fmt(fmt):
    valid_fmts = ["sparse", "dense"]
    if fmt not in valid_fmts:
        raise ValueError(f"`fm` must be in: {valid_fmts}, but found: {fmt}.")
    return fmt
```
We shouldn't use `format`, that's already a Python function.

mloning (commented):
Hi @fkiraly, let me know if you need any help with this PR, I wanted to work on series annotation anyway.

fkiraly:
>let me know if you need any help with this PR, I wanted to work on series annotation anyway.

sure - I think there's not much outstanding:

* tests, potentially writing some specifically for annotators
* moving the sparse/dense setting to the base class constructor, and passing the settings from concrete classes to the base class constructor
* updating the extension template with any additions

mloning:
Okay, I'll take a look at this now

mloning:
Please see #1093 for additional changes, this should address the open conversations on this PR. 

Once that is merged, we can merge #1064.

Quick questions: You can only set "sparse" with "indicator", right, or does "sparse" also work with "score"?

mloning (approved):
I think this is good to go, but would wait after the release, any other comments?

fkiraly:
>but would wait for after the release

Why that?

Doesn't hurt to have another base class and extension templates in.
That way, the community can more easily help extending the annotation module.

mloning:
@fkiraly I just though we would want to get the basic unit testing done first before we publish it. 

Are you happy with this being merged as it is?

mloning:
Thinking about it, perhaps it would be nice to get this out with the new release? Feel free to merge this @fkiraly.

fkiraly:
>Thinking about it, perhaps it would be nice to get this out with the new release? Feel free to merge this @fkiraly.

Yes - might be a good template for people to contribute to. The individual contributions may then lead to base class refactor, which may lead to a major version up. So, agreed.

mloning:
Now merged @satya-pattnaik @fkiraly - great work, it took us some time, but we finally have some basic annotation framework set up!  🎉

PR #1029, opened 2021-06-23
refactors statsmodels and  theta forecaster

thayeylolu:

- [x] _StatsModelAdapter and descendants AutoETS, ExponentialSmoothing, ThetaForecaster


sktime/forecasting/base/adapters/_statsmodels.py
```diff
@@ -41,11 +45,7 @@ def fit(self, y, X=None, fh=None):
         # so we coerce them here to pd.RangeIndex
         if isinstance(y, pd.Series) and type(y.index) == pd.Int64Index:
```
fkiraly, line 46: hm, this should perhaps be a `BaseForecaster` functionality - let's make a note of this (but no action required)

thayeylolu:
@mloning this is ready for review :)

fkiraly (approved):
looks good to me, we need to update from main and let the tests pass

fkiraly:
tests passed overnight - tests have re-started due to a PR that only fixed a broken link, so force-merging

PR #1031, opened 2021-06-23
Refactor reducer

Lovkush-A:
#### Reference Issues/PRs
#955 

* removed forecasting mixins and replaced with tags.
* moved `fit` in `_Reducer`. there was one check in this `fit` that was moved to all the other `_fit` methods
* added `fh` as parameter to all the `_fit` methods and adjusted docstring as such

Note that locally I checked only the tests for reduce, and those all passed. But who knows if other tests will fail... :/


Lovkush-A:
@fkiraly Please review this

mloning (approved):
@Lovkush-A Looks all good to me! Happy to merge this as it is.

Just a comment: Having a `_fit` method in the `_Reducer` class which implements the `check_window_length` call and then delegates to another `_internal_fit`(or whatever you want to call it) would be an alternative but seems a bit overcomplicated.

Lovkush-A:
> Just a comment: Having a `_fit` method in the `_Reducer` class which implements the `check_window_length` call and then delegates to another `_internal_fit`(or whatever you want to call it) would be an alternative but seems a bit overcomplicated.

I had similar thoughts, but also think it would have been overcomplicated

fkiraly (approved):
happy too, let's just wait for tests to pass

fkiraly:
tests passed overnight - tests have re-started due to a PR that only fixed a broken link, so force-merging

Issue #982, opened 2021-06-19
re-factoring forecaster.update, update_predict, update_predict_single


fkiraly:
The rolling prediction interface of forecasters is currently confusing and unintuitive.

I suggest we re-factor this as follows:

* `update` remains as is currently
* `update_predict` becomes what `update_predict_single` is currently, a short-hand for "one `update` then one `predict`
* current functionality in `update_predict` gets replaced by an `update_predict_playback`, as below.

`update_predict_playback(y, cutoffs, X=None, fh=None, update_params=True) -> list`

`y` is the full series; `cutoffs` (a list of indices) are indices on the same axis as `y` (but not necessarily a sub-set of `y.index).
It is assumed (and asserted/tested) that `cutoff` is monotonous. `fh` must be relative.

The return is a list of same length as `cutoffs`; the 0-th element is the return of `predict` after `fit` to `y[:cutoffs[0]]` (temporal slice); the i-th element is the return of the i-th `update_predict`, with `y[cutoffs[i-1]:cutoffs[i]]`.
All predictions are with the `fh` passed.

fkiraly:
@mloning, @aiwalter, thoughts?

Lovkush-A:
maybe `..._multiple` instead of `..._playback`? You are making multiple updates and predicts, and it pairs well with the method name `update_predict_single`

fkiraly:
>and it pairs well with the method name `update_predict_single`

but the proposal includes the removal of `update_predict_single`.
Further, it should be clear through naming that multiple application is basically the key step in backtesting, i.e., you already need your entire data ready. "multiple" isn't so clear here.

mloning:
> * `update` remains as is currently
> * `update_predict` becomes what `update_predict_single` is currently, a short-hand for "one `update` then one `predict`

I think we have discussed this before, I fully agree. 

> * current functionality in `update_predict` gets replaced by an `update_predict_playback`, as below.

* I prefer @Lovkush-A's suggestion: `update_predict_multiple` or `update_predict_cv` would perhaps be clearer still. 
* What's the thinking behind changing the signature from `cv` to `cutoffs`? Don't CV objects encode different sets of cutoffs? Note that there is a `CutoffSplitter` which you can pass a list of cutoffs and it will do what you suggest. 
[...]

PR #1034, opened 2021-06-23
Forecasters re-factoring: BaseGridSearch, ForecastingGridSearchCV, ForecastingRandomizedSearchCV

### LINKED ISSUE #955: re-factoring existing forecasters to new interface

GuzalBulatova:
#### Reference Issues/PRs

References #955 

- ❗ **added `self.check_is_fitted()` to `BaseForecaster`'s `update_predict` fn**
- changed `fit` to private `_fit`
- made sure `\forecasting\model_selection\tests\test_tune.py` pass
- added `_tags`: used the (conservative) values from the extension template

No

`_check_is_fitted` (line 170) - is it necessary?


fkiraly:
I think the test fails because the forecaster does not raise a "not fitted" error when you call `update_predict`.
Normally, a call `self.check_is_fitted()` would take care of this.

Now, it would appear this is missing in the base forecaster already. Can you add it there? That should fix this issue.

fkiraly:
Also, please make sure to update from main.

fkiraly:
To be more precise:

I think you should try to add the line
`self.check_is_fitted()` at the start of `BaseForecaster`'s `update_predict` - that would be my best guess.

fkiraly:
this looks done - if you think it's done, can you please convert to PR from draft PR and request reviews?
Make sure to update from main first.

GuzalBulatova:
Thank you @fkiraly , I've updated and opened it for review.

fkiraly (changes requested):
@GuzalBulatova, thanks!

I just noticed a problem: please return the public methods that were not moved "one layer deep" to their original state. This would be `get_fitted_params`, `inverse_transform`, `score`, and `evaluate_candidates`.

GuzalBulatova:
@fkiraly thanks, I've updated the PR. 👍 
Should I remove `self.check_is_fitted("score")` from  `score` method? 
In `_base.py` the method `score` has commentary  "no input checks needed here, they will be performed in predict and loss function".

fkiraly:
>Should I remove self.check_is_fitted("score") from score method?

The mentioned functions should remain entirely unchanged, since we did not change the base classes there.

fkiraly:
I think there is still a problem with `transform` and `inverse_transform`? These should be as they were before.

GuzalBulatova:
@fkiraly thank you, I un-did the changes and I think it should all be ready now.

GuzalBulatova:
@fkiraly a question about `self.check_is_fitted()` in `cutoff`. I took it out, but it's not in base - should we add it to the `BaseForecaster` or should I return it in `BaseGridSearch`?

fkiraly:
>@fkiraly a question about `self.check_is_fitted()` in `cutoff`. I took it out, but it's not in base - should we add it to the `BaseForecaster` or should I return it in `BaseGridSearch`?

Hm, looking at this:

* it should remain in `BaseGridSearch`, since it overloads the cutoff with a good reason (mirror the best forecaster)
* but it should *also* be added in the `BaseForecaster`

I.e., it should be in *both* places in the end. In the description of the PR, make a note where you changed the `BaseForecaster`.

GuzalBulatova:
Hi @fkiraly! I need your advice. After adding fit check in `BaseForecaster`'s `cutoff` a lot of tests failed. (If add it only to the cutoff in `BaseGridSearch` all the tests ran fine). As I understand it, there are many estimators that don't call `fit`. Is there an elegant solution to this? I couldn't figure it out. Thank you!

fkiraly:
>Hi @fkiraly! I need your advice. After adding fit check in `BaseForecaster`'s `cutoff` a lot of tests failed. (If add it only to the cutoff in `BaseGridSearch` all the tests ran fine). As I understand it, there are many estimators that don't call `fit`. Is there an elegant solution to this? I couldn't figure it out. Thank you!

Hm, looks like a larger number of forecasters check the `cutoff` when it's not yet fitted.
So I would suggest, remove it from `BaseForecaster` again, to complete this refactor, and let's discuss separately whether it makes sense that some forecaster access the field itself in `fitting`, or whether that should be done differently, e.g., via a temporary variable.

Other opinions, @mloning, @aiwalter? I don't think we need to solve the issue in this PR, so let's finish the refactoring without further changes to `BaseForecaster` and re-visit later.

sktime/forecasting/model_selection/_tune.py
```diff
@@ -46,14 +53,13 @@ def __init__(
         super(BaseGridSearch, self).__init__()
 
     @if_delegate_has_method(delegate=("best_forecaster_", "forecaster"))
-    def update(self, y, X=None, update_params=False):
+    def _update(self, y, X=None, update_params=False):
```
mloning, line 56: Hi @fkiraly and @GuzalBulatova, quick question: How does this work with the decorator, would the object still have `update` from the base class even if the delegate doesn't have an `_update` method? Perhaps these decorators aren't even necessary anymore if all forecasters have these methods or otherwise raise a `NotImplementedError`?

mloning: I have to look into this again but I think they should still be the public methdods (without the underscore), as we delegate the call to the `best_forecaster_`.

fkiraly: hm, but both the best forecaster and the grid search would inherit the same `update`, so it wouldn't really matter?

sktime/forecasting/model_selection/_tune.py
```diff
@@ -65,8 +71,8 @@ def update_predict(
         """Call update_predict on the forecaster with the best found
         parameters.
         """
-        self.check_is_fitted("update_predict")
-        return self.best_forecaster_.update_predict(
+
+        return self.best_forecaster_._update_predict(
```
fkiraly, line 75: currently, there is no `_update_predict` in the base class, so thsi should be delegated to `update_predict`

fkiraly (changes requested):
looks good, with the exception that an `_update_predict` is called.

Currently, this method does not exist in the template, so it should be reverted to `update_predict`.
I left it that way due to the issues and discussions around #982.

sktime/forecasting/model_selection/_tune.py
mloning: Yes that's what I thought too, but I have a slight preference for delegating the public method, rather than assuming that the forecaster always inherits from `BaseForecaster`

fkiraly: yes, good point

fkiraly: but the clean way would be for `_update` to delegate to `update`, whic the delegate decorator can't do.
I'd suggest to leave it as is then?

mloning: Yes, I'd leave it as it is

fkiraly: ... but then the tests break, apparently? Why? @GuzalBulatova I don't understand this, do you have a good explanation?

mloning (approved):
I think this makes sense now as it is, let's see if the tests pass.

fkiraly:
wait, why? Why do we need to override almost everything? This doesn't look right.

fkiraly (approved):
ok, I've looked at the code.

Indeed it makes sense, since we need to check tags etc in the delegates.

fkiraly:
merged, great! Last refactor item done, thanks @GuzalBulatova!

fkiraly:
FYI, some follow-on work is happening in #1088, #1091, #1092

Lovkush-A:
I just just about to have a look at this, but couldn't find it in the list of PRs. Good stuff for getting this sorted! 😄

fkiraly:
ah, but it's not sorted yet, @Lovkush-A!

Look at #1088, #1091 and #1092.
Seems like @mloning and I disagree on how to proceed.

My fix is in #1091, and @mloning then suggested to do it differently in #1092.

PR #1037, opened 2021-06-23
Forecasting tutorial rework - with cell content

fkiraly:
This is the forecasting tutorial with all cell contents, see #972.

No other changes except cell content in the main forecasting tutorial added.


mloning (commented):
Can we remove the setup instructions at the top? As said, this should go somewhere else, either an automatically generated header in the docs or simply a link to the install instructions.

fkiraly:
>Can we remove the setup instructions at the top? As said, this should go somewhere else, either an automatically generated header in the docs or simply a link to the install instructions.

Sure - do we already have any of that? If yes, please provide link

Merging to get the "big bit" out of the way (figures), we can always work on individual changes to bits and pieces in separate PR.

aiwalter (approved):

Issue #1038, opened 2021-06-23
Make multivariate compositor/reducer, that creates multivariate method by applying transformers/forecasters per-variable


fkiraly:
We should make the following compositors:

- [ ] a transformer that applies an univariate (or any other) series-to-series transformer to each series/variable
- [ ] a forecaster that applies an univariate (or any other) series-to-series forecaster to each series/variable

The first should be possible right now, the second will require #980.

The first one should solve #1030.

SveaMeyer13:
Happy to work on this!

aiwalter:
great idea, would this be sth equivalent to `sklearn.ColumnTransformer`, right? So would it be possible to select some columns or would it apply a transformer just to all columns?
___
Martin Walter <martin_friedrich.walter@daimler.com>, Mercedes-Benz AG on behalf of Daimler TSS GmbH.
https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md

SveaMeyer13:
@fkiraly would smth like this go into the direction you have in mind: 

class MultivariateCompositor(_SeriesToSeriesTransformer):
    """
    Parameters
    ----------
    transformer : series-to-series transformer to be applied to each series

    """
    _required_parameters = ["transformer"]


    def __init__(self, transformer):
        self.transformer = transformer
        self.transformers_ = None
        super(MultivariateCompositor, self).__init__()

    def fit(self, Z, X=None):
        """
        Iterates over columns (series) and applies the fit function of the transformer
        """
        if not isinstance(Z, pd.DataFrame):
            raise ValueError("Z needs to be a multivariate series as a pd.DataFrame")
        self._is_fitted = False
        z = check_series(Z)
        self.transformers_ = []
        for col in z:
            transformer = clone(self.transformer)
            self.transformers_.append(transformer)
        self._is_fitted = True
        return self

    def transform(self, Z, X=None):
        """
        calls transform on every single transformer (one transformer per series)
        """
        self.check_is_fitted()
        if not isinstance(Z, pd.DataFrame):
            raise ValueError("Z needs to be a multivariate series as a pd.DataFrame")
        z = check_series(Z)
        for col, transformer in enumerate(self.transformers_):
            transformer.check_is_fitted()
            z[col] = transformer.transform(Z,X)
        return z

@aiwalter Happy to change it in a way that it only iterates over specified columns


[...]

PR #1042, opened 2021-06-24
Multivariate Detrending

### LINKED ISSUE #1030: Make transformer for detrending support multivariate series
For the EEG applications we need a transformer that detrends each channel. Currently the detrending transformer only supports univariate time series.

Detrend each univariate series separately

SveaMeyer13:
#### Reference Issues/PRs

Fixes #1030 . See also #1038 .

This would be a quick fix to #1030 as it adds functionality for multivariate series to the detrender. As discussed with @mloning I took inspiration from how this was done for the imputer. But I also like the idea to have compositions for multivariate series as it is described in #1038


fkiraly:
Really nice!

What I would suggest though to address: the columns of all the data frames passed need not be in the same order.
So I would suggest:

* check whether all columns appearing later than `fit` have already been seen
* index `forecaster_` copies by column name, using a `dict`, not by integer index, that makes lookup easier

fkiraly (changes requested):
see above - suggest to robustify this with regards to:

* columns in different order in different calls after `fit`
* different set of columns passed

SveaMeyer13:
Thank you for your feedback @fkiraly - it's really helpful. 

Just to make sure I understood correctly: 

a) currently, forecaster_ is a list of forecasters and I will change it to a dict that contains the column name as keys and the forecasters as values

b) I will check if a column has already been seen, by checking whether the forecaster that belongs to that column is fitted, right?

fkiraly:
>a) currently, forecaster_ is a list of forecasters and I will change it to a dict that contains the column name as keys and the forecasters as values

That's precisely what I would suggest, yes.

>b) I will check if a column has already been seen, by checking whether the forecaster that belongs to that column is fitted, right?

What I would have done is simply check whether the list of columns is a sub-set of the `dict`'s keys, which should work if those keys are generated in `fit`. Alternatively, you could store the columns you see in `fit` in an attribute and use that.

fkiraly (changes requested):
Looks good, almost ready to go.

The one thing I would recommend to change is the error message `"not all columns are known"` - this is vague and not very helpful. Not all columns *of what*, and known *in which respect*?

Suggest: `Z` contains columns that have not been seen in `fit` (maybe even print those columns by including them in the error msg), and change the error to `ValueError` (since it's about the contents/values of `Z` - it is locally about a key, but this should be in reference to the inputs, not local objects).

SveaMeyer13:
@fkiraly, thanks for the feedback! I'll take care of it

sktime/transformations/series/detrend/_detrend.py
```diff
@@ -107,10 +117,22 @@ def transform(self, Z, X=None):
...
-        z = check_series(Z, enforce_univariate=True)
+        z = check_series(Z)
         fh = ForecastingHorizon(z.index, is_relative=False)
-        z_pred = self.forecaster_.predict(fh, X)
-        return z - z_pred
+
+        # multivariate
+        if isinstance(z, pd.DataFrame):
+            # check if all columns are known
+            if not all(k in self.forecaster_ for k in z.columns):
```
fkiraly, line 126: for the check or printing missing columns, you could do sth like `set(z.columns).difference(set(self.forecaster_.keys()))`; for readability you could first define variables `Z_fit_keys = set(self.forecaster_.keys())`, `Z_new_keys = ...`

SveaMeyer13: Thanks! I implemented it as you suggested

fkiraly (approved):
looks great!

We can merge once the tests complete.
Do you want to have a look too, @mloning?

mloning:
You can merge if you think it's all good, happy to take a look if you're unsure.

aiwalter:
@SveaMeyer13 I just tested this and I got a warning:
```
C:\...\sktime\sktime\transformations\series\detrend\_detrend.py:136: SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame.
Try using .loc[row_indexer,col_indexer] = value instead
See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
  z[colname] = z[colname] - z_pred
```
Could you have a look into it?
___

SveaMeyer13:
Thanks, @aiwalter! 

For some reason I can not reproduce this warning when running the following code:

```python
y, X = base.load_longley()
transformer = Detrender(forecaster=PolynomialTrendForecaster(degree=1))
y_hat = transformer.fit_transform(X)
print(y_hat)
```

But I do see the problem and I'm looking into it. 

For now, I think I should have made a deep copy of z in the transform/inverse-transform function before applying changes to it.

aiwalter:
Will post the code tomorrow to reproduce

aiwalter:
@SveaMeyer13 seems the warning was caused by sth in my `ForecastingPipeline` which I have fixed now. So all is fine with your PR, was my mistake! I just used the new `Detrender` there for some tests...
___

PR #1044, opened 2021-06-24
Add ColumnwiseTransformer (multivariate compositor for series-to-series transformer)

SveaMeyer13:
#### Reference Issues/PRs
addresses #1038 

This is a very first draft to apply a single series-transformer to a multivariate series in a column-wise manner

No

I'm not sure if this even goes in the direction you had in mind @fkiraly?

The main point I'm not sure how to tackle, is what to do about series transformers that also have functions that are not fit, transform, or inverse transform, e.g. the "update" in the detrender. My solution would be to have an extra "apply" function - but here I am currently relying on the fact, that any additional functions would return "self"  

#### Any other comments?
This is just a draft. Tests will fail as I haven't set the _required_parameters = ["transformer"] in the test config yet 
Just wanted to get early feedback


SveaMeyer13:
@mloning or @fkiraly can I rely on the fact that every transformer has at most four functions (fit, transform, inverse-transform and update)?

fkiraly:
wait a minute, I don't think it's a good idea to introduce a `multivariate-only` tag here. My opinion is, the transformer should still take `pd.Series`, but cast it to a one-column `pd.DataFrame`. For now, we can put that logic in `fit`.

In any case, before we introduce new tags, I would say we need sth like the #1067 registry to avoid having "personal interpretations" of tags.

What's your opinion, @mloning?

SveaMeyer13:
thanks for your feedback, @fkiraly! 

I don't fully understand (yet). Maybe I misunderstood the task. 


> wait a minute, I don't think it's a good idea to introduce a `multivariate-only` tag here. My opinion is, the transformer should still take `pd.Series`, but cast it to a one-column `pd.DataFrame`. For now, we can put that logic in `fit`.

Fom my understanding current univariate series-to-series transformers work on `pd.Series`. Which is why we can not simply apply them to multivariate series - as multivariate series are `pd.DataFrames`. So the task of this transformer would be to iterate over the columns of a multivariate series (`pd.DataFrame`), thereby getting `pd.Series` and then applying the univariate transformers to each of these columns or selected columns which are `pd.Series`. 

So in this case the multivariate-compositor-transformer would take a `pd.DataFrame`

> In any case, before we introduce new tags, I would say we need sth like the #1067 registry to avoid having "personal interpretations" of tags.
> 
> What's your opinion, @mloning?

fkiraly:
Sorry for creating confusion, @SveaMeyer13.

I think what you are doing makes sense, I'd just not introduce new tags at this moment.
We'll have a technical meeting where we'll settle on a set of tags, until then let's not make more...

sktime/tests/_config.py
```diff
@@ -85,6 +85,7 @@
 from sktime.transformations.series.compose import OptionalPassthrough
 from sktime.transformations.series.outlier_detection import HampelFilter
 from sktime.transformations.series.boxcox import BoxCoxTransformer
+from sktime.transformations.series.multivariate_compositor import MultivariateCompositor
```
aiwalter, line 88: Maybe it makes sense to have this in `sktime.transformations.series.compose`? Now there is a second path that contains `compositor` which is a bit rudimentary?

sktime/transformations/series/multivariate_compositor.py
```diff
@@ -0,0 +1,118 @@
...
+from sktime.transformations.base import _SeriesToSeriesTransformer
+from sktime.utils.validation.series import check_series
+
+
+class MultivariateCompositor(_SeriesToSeriesTransformer):
+    """
+    Parameters
+    ----------
+    transformer : series-to-series transformer to be applied to each series
+    columns : names of columns that are supposed to be transformed
```
aiwalter, line 19: here we should also add the type following the `numpy` docstring format. So e.g.
```
columns : list of str
       Names of columns that are supposed to be transformed
```

sktime/transformations/series/multivariate_compositor.py
```diff
@@ -0,0 +1,118 @@
...
+        self.columns = columns
+        self.transformers_ = None
+        super(MultivariateCompositor, self).__init__()
+
+    def fit(self, Z, X=None):
+        """
+        Iterates over columns (series) and applies the fit function of the transformer
+        """
+        if not isinstance(Z, pd.DataFrame):
+            raise ValueError("Z needs to be a multivariate Pandas Series")
```
aiwalter, line 38:
```suggestion
            raise ValueError("Z needs to be a pd.DataFrame")
```

sktime/transformations/series/multivariate_compositor.py
```diff
@@ -0,0 +1,118 @@
...
+        if len(difference) != 0:
+            raise ValueError("Missing columns" + str(difference) + "in Z.")
+
+        self.transformers_ = {}
+        for colname in self.columns:
+            transformer = clone(self.transformer)
+            self.transformers_[colname] = transformer
+            self.transformers_[colname].fit(z[colname], X)
+            # self.transformers_[colname].is_fitted = True
+        self._is_fitted = True
```
aiwalter, line 57:
```suggestion
```

aiwalter: this is already contained in the parent class `BaseTransformer`

sktime/transformations/series/multivariate_compositor.py
```diff
@@ -0,0 +1,118 @@
...
+        return self
+
+    def transform(self, Z, X=None):
+        """
+        calls transform on every single transformer (one transformer per series)
+        """
+        if not isinstance(Z, pd.DataFrame):
+            raise ValueError("Z needs to be a multivariate Pandas Series")
```
aiwalter, line 65: this should be included in `check_series()` function with a new param `enforce_multivariate`. Also above you have this check which can be replaced then as well with simply z = check_series(Z, enforce_multivariate=True).
Also the error should rather be `"Z needs to be a pd.DataFrame"`

sktime/transformations/series/multivariate_compositor.py
```diff
@@ -0,0 +1,118 @@
...
+
+        self.check_is_fitted()
```
aiwalter, line 67: I think this should also be in parent class, but we can do this probably best in a separate transformer refactoring

sktime/transformations/series/multivariate_compositor.py
```diff
@@ -0,0 +1,118 @@
...
+        return z
+
+    def inverse_transform(self, Z, X=None):
+        """
+        if the base transformer has an inverse-transform this
+        inverse transform is called on every single transformer
+        (one transformer per series)
+        """
+        if not isinstance(Z, pd.DataFrame):
+            raise ValueError("Z needs to be a multivariate Pandas Series")
```
aiwalter, line 91: same as above, to be replaced with `check_series`

sktime/transformations/series/multivariate_compositor.py
```diff
@@ -0,0 +1,118 @@
...
+
+        if not hasattr(self.transformer, "inverse_transform"):
+            raise NotImplementedError(
+                "this transform does not have an inverse_transform method"
+            )
```
aiwalter, line 96:
```suggestion
```

aiwalter: I think this is not needed, as it will anyway raise an exception if the object `self.transformer` does not have `inverse_transform`. So its rudimentary to catch this exception again here.

sktime/transformations/series/multivariate_compositor.py
```diff
@@ -0,0 +1,118 @@
...
+
+            # make sure z contains all columns that the user wants to transform
+            Z_wanted_keys = set(self.columns)
+            Z_new_keys = set(z.columns)
+            difference = Z_wanted_keys.difference(Z_new_keys)
+            if len(difference) != 0:
+                raise ValueError("Missing columns" + str(difference) + "in Z.")
+            for colname in self.columns:
+                # self.columns : columns that are supposed to be transformed
+                self.transformers_[colname].check_is_fitted()
```
aiwalter, line 112:
```suggestion
```
`self.check_is_fitted()` is in each transfromer already in `inverse_transform` or we should also best have it in the parent class when we refactor transformers. But we can remove this here as we dont need to check twice

sktime/transformations/series/multivariate_compositor.py
```diff
@@ -0,0 +1,118 @@
...
+        if not hasattr(self.transformer, "inverse_transform"):
+            raise NotImplementedError(
+                "this transform does not have an inverse_transform method"
+            )
+        self.check_is_fitted()
+        z = check_series(Z)
+
+        if isinstance(Z, pd.DataFrame):
+            # make copy of z
+            z = z.copy()
```
aiwalter, line 102: this will also be later in parent ideally, but we can keep it for now

sktime/transformations/series/multivariate_compositor.py
```diff
@@ -0,0 +1,118 @@
...
+__all__ = ["MultivariateCompositor"]
+__author__ = ["Svea Meyer"]
+
+import pandas as pd
+from sklearn.base import clone
+from sktime.transformations.base import _SeriesToSeriesTransformer
+from sktime.utils.validation.series import check_series
+
+
+class MultivariateCompositor(_SeriesToSeriesTransformer):
```
aiwalter, line 14: why not to name this class `ColumnTransformer` as in `sklearn`? It would be a name conflict maybe if s.o. wants to use both together. So what do you think about `ColumnComposition`. I think the term "multivariate" is in general weakly defined and often confusing if you talk about endog or exog multivariate, thats why I would propose to remove the name "multivariate" from the class name. What do you think?

aiwalter (commented):
@SveaMeyer13 I just did a short review :)

sktime/tests/_config.py
SveaMeyer13: Thanks for your feedback, @aiwalter!
I'll move it to the compose directory.

sktime/transformations/series/multivariate_compositor.py
aiwalter: ok it seems you overwrite `fit()` here, so in this case we need `self._is_fitted = True`. So this is also a task for refactoring, but not your task now I think :)

SveaMeyer13: hmm here I might be having a misconception, but my train of thought was that I would want to be able to check whether this overall transformer is fitted. I am aware that I am already setting is_fitted to true for all the single series transformers when calling fit() on them. But this then only contains information about the single series transformers being fitted and not whether fit() has been applied to this multivariate_compositor already.

sktime/transformations/series/multivariate_compositor.py
SveaMeyer13: Agreed, this might still change depending on what we decided about the question of whether it's ok to have a multivariate-only tag after all (see discussion above)

sktime/transformations/series/multivariate_compositor.py
SveaMeyer13: agreed!

sktime/transformations/series/multivariate_compositor.py
aiwalter: after refactoring `BaseTransformer`, we would have there a `_fit()` function that will contain `self._is_fitted = True`. But you can ignore my comment above, as I have seen that there is just a `fit()` function which is overwritten by each transformer currently. Didnt want to make confusion :)

SveaMeyer13: Ah, I did not see your second comment! yes, exactly I'm overwriting fit()

sktime/transformations/series/multivariate_compositor.py
aiwalter: I think its not dependent to the tags, but rather assessing the input? Tags are mainly for tests

sktime/transformations/series/multivariate_compositor.py
SveaMeyer13: Will do, once I know whether it's ok to have to multivariate-only tag :)

sktime/transformations/series/multivariate_compositor.py
SveaMeyer13: Agreed!

sktime/transformations/series/multivariate_compositor.py
SveaMeyer13: I wasn't aware of that, thanks @aiwalter! So far `check_series` however, doesn't have an argument `enforce_multivariate` (only enforce_univariate) so I would need to introduce that. And I think that would only make sense if we end up having the tag `multivariate-only`. Otherwise, this function would also have to accept pd.Series.

sktime/transformations/series/multivariate_compositor.py
SveaMeyer13: yes, for sure - I don't mind renaming it

sktime/transformations/series/multivariate_compositor.py
aiwalter: i think to add `enforce_multivariate` to `check_series` is fine, would be great if you can do it with this PR.?

SveaMeyer13: sure, will do :)

aiwalter: its not dependent to any tag discussion, bcz having it will help to replace your code here and in other places where we want to do `check_series(Z, enforce_multivariate=True)` in the future.

sktime/transformations/series/multivariate_compositor.py
aiwalter: ok, but I am not sure which name would be best. Any ideas @mloning @fkiraly ?

SveaMeyer13:
Hi @aiwalter, I implemented your feedback. Thanks again, it was really helpful! 

I had a short discussion with @fkiraly and @mloning yesterday and we decided that it is best to also accept `pd.Series` as inputs. 

Therefore there is no longer a need to check whether a series is multivariate and I think that I don't need the `enforce_multivariate` in `check_series()` for this PR. Happy to implement it though if you see a place in this PR where it might be necessary :)

aiwalter:
okay thats fine, not sure about the use case with a `pd.Series` however. But then we indeed dont need to change `check_series`

sktime/transformations/series/compose.py
```diff
@@ -96,3 +97,143 @@ def inverse_transform(self, Z, X=None):
         if not self.passthrough:
             z = self.transformer_.inverse_transform(z, X=None)
         return z
+
+
+class ColumnComposition(_SeriesToSeriesTransformer):
+    """
+    Parameters
+    ----------
+    transformer : series-to-series transformer to be applied to each series
```
aiwalter, line 106:
```suggestion
    transformer : Estimator
        scikit-learn-like or sktime-like transformer to fit and apply to series
```

sktime/transformations/series/compose.py
```diff
@@ -96,3 +97,143 @@ def inverse_transform(self, Z, X=None):
...
+                self.transformers_[colname] = transformer
+                self.transformers_[colname].fit(z[colname], X)
+                # self.transformers_[colname].is_fitted = True
+            self._is_fitted = True
+            return self
+
+    def transform(self, Z, X=None):
+        """
+        calls transform on every single transformer (one transformer per series)
+        """
```
aiwalter, line 155: I think we need also a proper docstring here, see e.g.  `HampelFilter`:
https://github.com/alan-turing-institute/sktime/blob/859ac7206ad8aad8ce418b391c521064083e30c3/sktime/transformations/series/outlier_detection.py#L63-L74

sktime/transformations/series/compose.py
```diff
@@ -96,3 +97,143 @@ def inverse_transform(self, Z, X=None):
...
+                self.transformers_[colname].check_is_fitted()
+                z[colname] = self.transformers_[colname].transform(z[colname], X)
+            return z
+
+    def inverse_transform(self, Z, X=None):
+        """
+        If the base transformer has an inverse-transform this
+        inverse transform is called on every single transformer
+        (one transformer per series)
+        """
```
aiwalter, line 185: also here, you can just copy the docstring about the params from another transformer

sktime/transformations/series/compose.py
```diff
@@ -96,3 +97,143 @@ def inverse_transform(self, Z, X=None):
...
+                    z[colname], X
+                )
+            return z
+
+    def update(self, Z, X=None, update_params=True):
+        """
+        If the base transformer has an update function this
+        it is called on every single transformer
+        (one transformer per series)
+        """
```
aiwalter, line 217: same here

SveaMeyer13:
Thanks, @aiwalter! I updated the docstrings accordingly.

aiwalter:
Cool, just a minor thing I forgot is to have a docstring example (see other transformers). You could use the `load_longley` data which has also an `X`. Let me know if I can support here.

sktime/transformations/series/compose.py
```diff
@@ -96,3 +97,204 @@ def inverse_transform(self, Z, X=None):
         if not self.passthrough:
             z = self.transformer_.inverse_transform(z, X=None)
         return z
+
+
+class ColumnComposition(_SeriesToSeriesTransformer):
```
mloning, line 102: We need to think about the name, `ColumnComposition` seems very general. We have a `RowwiseTransformer`, so perhaps `ColumnwiseTransformer`? Or `ColumnApplier` or `ColumnApplyTransformer`?

sktime/transformations/series/compose.py
```diff
@@ -96,3 +97,204 @@ def inverse_transform(self, Z, X=None):
...
+        self._is_fitted = False
+
+        # univariate case
+        if isinstance(z, pd.Series):
+            self.transformer.fit(z, X)
+            self._is_fitted = True
+            return self
+        # multivariate case
+        else:
+            if self.columns == "all":
```
mloning, line 162: What are the other options for `columns`? For now I would always apply it to all columns, not giving users the option. For the more expressive design I would prefer something like the `ColumnTransformer` from scikit-learn where you can specify different transformers for different columns.

sktime/transformations/series/compose.py
```diff
@@ -96,3 +97,204 @@ def inverse_transform(self, Z, X=None):
...
+            z = z.copy()
+            # make sure z contains all columns that the user wants to transform
+            z_wanted_keys = set(self.columns)
+            z_new_keys = set(z.columns)
+            difference = z_wanted_keys.difference(z_new_keys)
+            if len(difference) != 0:
+                raise ValueError("Missing columns" + str(difference) + "in Z.")
+            for colname in self.columns:
+                # self.columns : columns that are supposed to be transformed
+                self.transformers_[colname].check_is_fitted()
```
mloning, line 214: This will happen inside the transform call in the next line and we have checked it at the beginning of this method
```suggestion
```

sktime/transformations/series/compose.py
```diff
@@ -96,3 +97,204 @@ def inverse_transform(self, Z, X=None):
...
+        Returns
+        -------
+        Z : pd.Series, pd.DataFrame
+            Transformed time series(es).
+        """
+        self.check_is_fitted()
+        z = check_series(Z)
+
+        # univariate case
+        if isinstance(z, pd.Series):
```
mloning, line 198: Is there a way to write this without the if-else case distinction? For example, if it's a series, we could just convert it into a data frame (e.g. via `z.to_frame()`) and back to series before returning it. That way, we could use the main loop below for data frames also for series input.  The same applies to the other methods (fit, inverse-transform and update).

sktime/transformations/series/compose.py
```diff
@@ -96,3 +97,204 @@ def inverse_transform(self, Z, X=None):
...
+        # multivariate case
+        else:
+            # make copy of z
+            z = z.copy()
+            # make sure z contains all columns that the user wants to transform
+            z_wanted_keys = set(self.columns)
+            z_new_keys = set(z.columns)
+            difference = z_wanted_keys.difference(z_new_keys)
+            if len(difference) != 0:
+                raise ValueError("Missing columns" + str(difference) + "in Z.")
```
mloning, line 211: this check appears in multiple locations in the class, could we wrap this into a helper function, e.g. `_check_columns(columns, new_columns)`

sktime/transformations/series/compose.py
```diff
@@ -96,3 +97,204 @@ def inverse_transform(self, Z, X=None):
...
+
+    Example
+    -------
+    >>> from sktime.datasets import base
+    >>> from sktime.transformations.series.detrend import Detrender
+    >>> from sktime.transformations.series.compose import ColumnComposition
+
+    >>> y, X = base.load_longley()
+    >>> transformer = ColumnComposition(Detrender())
+    >>> y_hat = transformer.fit_transform(X)
```
mloning, line 127:
```suggestion
    >>> yt = transformer.fit_transform(y, X)
```
y_hat is typically used for estimates/forecasts

mloning (commented):
Hi @SveaMeyer13, looks very good already! I've left a few comments below. 

* We should also add it to the docs in `sktime/docs/source/api_reference/transformations.rst`.
* We need to make sure `get_params` and `set_params` work properly, e.g. can I get and set the params of `self.transformer`?

SveaMeyer13:
> Hi @SveaMeyer13, looks very good already! I've left a few comments below.

Thanks, @mloning. Your comments were really helpful. I think I changed everything as discussed. One thing I was unsure about in the end: should I make ` _revert_to_series()` static?

>     * We should also add it to the docs in `sktime/docs/source/api_reference/transformations.rst`.

done :) 

>     * We need to make sure `get_params` and `set_params` work properly, e.g. can I get and set the params of `self.transformer`?

Yes, I think it should be fine: 
image

sktime/transformations/series/compose.py
```diff
@@ -96,3 +97,210 @@ def inverse_transform(self, Z, X=None):
...
+    columnwise to multivariate series.
+
+    Parameters
+    ----------
+    transformer : Estimator
+        scikit-learn-like or sktime-like transformer to fit and apply to series
+    columns : list of str
+            Names of columns that are supposed to be transformed
+
+     Attributes
```
aiwalter, line 114:
```suggestion
    Attributes
```

sktime/transformations/series/compose.py
```diff
@@ -96,3 +97,210 @@ def inverse_transform(self, Z, X=None):
...
+    ----------
+    transformers_ : dict of {str : transformer}
+        Maps columns to transformers
+
+    Example
+    -------
+    >>> from sktime.datasets import base
```
aiwalter, line 121:
```suggestion
    >>> from sktime.datasets import load_longley
```

sktime/transformations/series/compose.py
```diff
@@ -96,3 +97,210 @@ def inverse_transform(self, Z, X=None):
...
+    >>> from sktime.transformations.series.detrend import Detrender
+    >>> from sktime.transformations.series.compose import ColumnwiseTransformer
+
+    >>> y, X = base.load_longley()
```
aiwalter, line 125:
```suggestion
    >>> y, X = load_longley()
```

sktime/transformations/series/compose.py
```diff
@@ -96,3 +97,210 @@ def inverse_transform(self, Z, X=None):
...
+    >>> transformer = ColumnwiseTransformer(Detrender())
+    >>> yt = transformer.fit_transform(X)
+    """
+
+    _required_parameters = ["transformer"]
+
+    def __init__(self, transformer, columns=None):
+        self.transformer = transformer
+        self.columns = columns
```
aiwalter, line 134: why do we have this here? I could find any other place where you read this. Below you only read/use `self.columns_` always

sktime/transformations/series/compose.py
```diff
@@ -96,3 +97,210 @@ def inverse_transform(self, Z, X=None):
...
+
+        # check that columns are None or list of strings
+        if columns is not None:
+            if not isinstance(columns, list) and all(
+                isinstance(s, str) for s in columns
+            ):
+                raise ValueError("Columns need to be a list of strings or None.")
+
+        self.columns_ = columns
+        self.transformers_ = None
```
aiwalter, line 144: I am wondering if a trailing underscore is required here or if we should rather make it a private variable like `_columns` 
 and `_transformers`? I cant think of a use case where you would want to get a particular transformer after fitting. Also it seems e.g. `EnsembleForecaster` does not have sth like `self.forecaster_` where the parts of the ensemble are exposed.

sktime/transformations/series/compose.py
```diff
@@ -96,3 +97,210 @@ def inverse_transform(self, Z, X=None):
...
+        the fit function of the transformer.
+
+        Parameters
+        ----------
+        Z : pd.Series, pd.DataFrame
+
+        Returns
+        -------
+        self : an instance of self
+
```
aiwalter, line 159:
```suggestion
```

sktime/transformations/series/compose.py
```diff
@@ -96,3 +97,210 @@ def inverse_transform(self, Z, X=None):
...
+            name_of_z = z.name
+            z = z.to_frame()
+
+        # make copy of z
+        z = z.copy()
+
+        # make sure z contains all columns that the user wants to transform
+        self._check_columns(z)
+        for colname in self.columns_:
+            # self.columns_ : columns that are supposed to be transformed
```
aiwalter, line 212: I think this docstring can be put to the top where this attribute is initiated. Actually if we keep `self.columns_` we would ideally need also an `Attributes` docstring for it like
```
Attributes
----------
columns_ : something
     My desciption

```

sktime/transformations/series/compose.py
```diff
@@ -96,3 +97,210 @@ def inverse_transform(self, Z, X=None):
...
+            is_series = True
+
+        # make copy of z
+        z = z.copy()
+
+        # make sure z contains all columns that the user wants to transform
+        self._check_columns(z)
+
+        for colname in self.columns_:
+            # self.columns_ : columns that are supposed to be transformed
```
aiwalter, line 255:
```suggestion
```

sktime/transformations/series/compose.py
```diff
@@ -96,3 +97,210 @@ def inverse_transform(self, Z, X=None):
...
+        z = check_series(Z)
+
+        # univariate case
+        if isinstance(z, pd.Series):
+            z = z.to_frame()
+
+        # make sure z contains all columns that the user wants to transform
+        self._check_columns(z)
+        for colname in self.columns_:
+            # self.columns_ : columns that are supposed to be transformed
```
aiwalter, line 288:
```suggestion
```

sktime/transformations/series/compose.py
```diff
@@ -96,3 +97,210 @@ def inverse_transform(self, Z, X=None):
         if not self.passthrough:
             z = self.transformer_.inverse_transform(z, X=None)
```
aiwalter, line 98: I have just seen a silent issue in my code 😮 . Possibly, if we would have a transformer that uses `X` to transform `y`. Could you just fix it like this:
```suggestion
            z = self.transformer_.inverse_transform(z, X=X)
```
Thx! Dont want to open a PR for changing one symbol 😅

aiwalter (commented):
Hi @SveaMeyer13 , I just found some additional minor things

SveaMeyer13:
Thank you @aiwalter! I'll take care of it next Wednesday. Currently, I'm preparing for an exam.

sktime/transformations/series/compose.py
SveaMeyer13: From how I understand the scikit learn developer guide, the idea is to not change the hyperparameters in fit/in the constructor or somewhere else. This ensures that `get_params` and `set_params` don't break. We are however changing `self.columns` in case it is `None` to all the column names. This is why I think that we need `self.columns` and `self.columns_` . The developer guide also says that this should happen in fit and and not in the constructor. Therefore, I moved this logic further down to `fit` now. The guide is here: https://scikit-learn.org/stable/developers/develop.html

sktime/transformations/series/compose.py
SveaMeyer13: I'm not sure about this one. I looked at other transformers and for example in `_detrend` there is also a `self.forecaster_` and in adapt there is a `self.transformer_`. Maybe @mloning can clarify?

aiwalter: ok, then in my opinion lets keep it here also to be consistent

SveaMeyer13:
Thanks again for your feedback @aiwalter, I made changes accordingly. Also, pydocstyle was complaining about incorrect/missing docstrings. So I added/changed them in your code as well. Please double-check if they are correct.

aiwalter:
@SveaMeyer13 thx for fixing docstring also in `OptionalPassthrough`

sktime/transformations/series/compose.py
```diff
@@ -91,8 +124,239 @@ def transform(self, Z, X=None):
...
+        if isinstance(z, pd.Series):
+            z = z.to_frame()
+
+        # make sure z contains all columns that the user wants to transform
+        self._check_columns(z)
+        for colname in self.columns_:
+            self.transformers_[colname].update(z[colname], X)
+        return self
+
+    def _check_columns(self, z):
```
mloning, line 346: this could be a stand-alone function into which you pass the `self.columns_` - the idea is that stand-alone functions are usually easier to understand than methods

sktime/transformations/series/compose.py
```diff
@@ -91,8 +124,239 @@ def transform(self, Z, X=None):
...
+        # make sure z contains all columns that the user wants to transform
+        z_wanted_keys = set(self.columns_)
+        z_new_keys = set(z.columns)
+        difference = z_wanted_keys.difference(z_new_keys)
+        if len(difference) != 0:
+            raise ValueError("Missing columns" + str(difference) + "in Z.")
+
+    def _revert_to_series(self, name_of_z, z):
```
mloning, line 354: this one too could be a stand-alone function

sktime/transformations/series/compose.py
```diff
@@ -91,8 +124,239 @@ def transform(self, Z, X=None):
...
+        # create series from one dimensional pd.DataFrame
+        # make sure not to loose column name
+        if name_of_z is not None:
+            z = z[name_of_z]
+        else:
+            z = z[0]
```
mloning, line 360: Would this work? 
```suggestion
            z = z.squeeze("columns")
```
https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.squeeze.html

sktime/transformations/series/compose.py
```diff
@@ -91,8 +124,239 @@ def transform(self, Z, X=None):
...
+        Returns
+        -------
+        Z : pd.Series, pd.DataFrame
+            Transformed time series(es).
+        """
+        self.check_is_fitted()
+        z = check_series(Z)
+
+        # transform to pd.Dataframe in univariate case
+        is_series = False
```
mloning, line 253: there's some code duplication here, lines 253-257 are used below again, right?

mloning (commented):
Hi @SveaMeyer13 - just a few minor comments, I think it looks very good now!

sktime/transformations/series/compose.py
SveaMeyer13: Thanks for your feedback @mloning! 

If I wrap this into a function it would look smth like this: 
``` python
def check_is_series(z): 
    name_of_z = none
    if isinstance(z, pd.Series):
            is_series = True
            name_of_z = z.name
            z = z.to_frame() 
    return z, name_of_z
```      
and then above we would have 
                `z, name_of_z = check_is_series(z)`
even though we only need `name_of_z` if the user passes a pd.series. So I'm not sure whether wrapping this into a function makes it less readable? But I'm happy to change it.

SveaMeyer13: Thanks @mloning, I moved it into a wrapper function now :)

mloning (dismissed):
This looks great - no more comments from my side!

aiwalter (approved):

aiwalter (approved):
just fixed a merge conflict

mloning (approved):

mloning:
Now finally merged @SveaMeyer13 🎉

PR #1051, opened 2021-06-24
Revert "Forecasting tutorial rework - with cell content"

fkiraly:
Reverts alan-turing-institute/sktime#1037

seems to be too large - seems like it can't be viewed in browser.
We may want to remove some of the images.


mloning:
In general we want cell content to be shown here: https://www.sktime.org/en/latest/examples/01_forecasting.html

PR #1066, opened 2021-06-27
An sktime registry

fkiraly:
This PR proposes to move the following things to a central `sktime` registry:

* scitype strings, estimator base class references and look-up
* tag names, the scitype which the tag is for, expected tag types

To all this, the registry should also record a plain English explanation, so it's runtime inspectable without having to dive through code files and comments in the code.

This should simplify test workflows (e.g., in `_config`), and retrieval workflows such as `all_estimators` or the automated overview of learning algorithm #704, filtered, say, by scitype or tag.

This PR also moves the `all_estimators` lookup utility to the `registry` module, from the `__init__` in the `utils` module where it previously was located. This is not interface breaking since the old location points to the new location (until deprecation).

note: the notebook update is due to one reference to `all_estimators` that got updated.

A version without moving `all_estimators` is in #1067 in case that proves controversial (or test breaking).


fkiraly:
FYI @afzal442, @abdulelahsm and @tobyhodges, comments are appreciated on the interaction of this and the automated estimator overview.

fkiraly:
Huh, issue number [1066](https://en.wikipedia.org/wiki/Battle_of_Hastings) - nice!

afzal442 (commented):
FOI, **`from sktime.utils import all_estimators`** => **`from sktime.registry import all_estimators`**

fkiraly:
yes, that should fix the failing tests, but don´t have time right now.
Until then, instead have a look at #1067, version without the `all_estimators` move

mloning:
Can this PR be closed?

fkiraly:
yes, I'll revisit `all_estimators` separately.

PR #1067, opened 2021-06-27
sktime-registry - without moving all_estimators

fkiraly:
This is #1066, without moving the `all_estimators` utility.


sktime/registry/_base_classes.py
```diff
@@ -0,0 +1,108 @@
...
+    (
+        "s-to-s-trafo",
+        _SeriesToSeriesTransformer,
+        "time-series-to-time-series transformer",
+    ),
+    ("p-to-t-trafo", _PanelToTabularTransformer, "panel-to-tabular transformer"),
+    ("p-to-p-trafo", _PanelToPanelTransformer, "panel-to-panel transformer"),
+]
+
+TRANSFORMER_MIXIN_SCITYPE_LIST = pd.DataFrame(TRANSFORMER_MIXIN_REGISTER)[0].values
```
mloning, line 102: None of the "lists" are actually lists, but numpy arrays, is this intended? 
```suggestion
TRANSFORMER_MIXIN_SCITYPE_LIST = pd.DataFrame(TRANSFORMER_MIXIN_REGISTER)[0].tolist()
```

mloning: I see that in the test files we use tuples, so why not define them as tuples here: 
```suggestion
TRANSFORMER_MIXIN_SCITYPE_LIST = tuple(pd.DataFrame(BASE_CLASS_REGISTER)[0])
```

sktime/registry/_base_classes.py
```diff
@@ -0,0 +1,108 @@
...
+BASE_CLASS_SCITYPE_LIST = pd.DataFrame(BASE_CLASS_REGISTER)[0].values
+
+BASE_CLASS_LIST = pd.DataFrame(BASE_CLASS_REGISTER)[1].values
+
+BASE_CLASS_LOOKUP = dict(zip(BASE_CLASS_SCITYPE_LIST, BASE_CLASS_LIST))
+
+
+TRANSFORMER_MIXIN_REGISTER = [
+    (
+        "s-to-prim-trafo",
```
mloning, line 89: I prefer the longer names
```suggestion
        "series-to-primitive-transformer",
```

mloning (commented):
I like the idea of having a registry, makes a lot of sense. We should later add the type hints here too. 

Should we move this in before we make a new release?

sktime/registry/_base_classes.py
fkiraly: which now?

sktime/registry/_base_classes.py
fkiraly: it's supposed to be a shorthand... but I'm not too precious about this

fkiraly: made them longer

sktime/registry/_base_classes.py
fkiraly: went with first suggestion

sktime/tests/_config.py
```diff
@@ -281,16 +279,7 @@
...
-    "fit-in-transform",  # fitted in transform or non-fittable
-    "univariate-only",
-    "transform-returns-same-time-index",
-    "handles-missing-data",
-    "skip-inverse-transform",
-    "requires-fh-in-fit",
-    "X-y-must-have-same-index",
-    "enforce-index-type",
-)
+VALID_ESTIMATOR_TAGS = tuple(ESTIMATOR_TAG_LIST)
```
mloning, line 282: Since we just wrap them in tuples here now, should we perhaps rename `VALID_ESTIMATOR_TAGS` to `ESTIMATOR_TAG_LIST` everywhere (and the same for the other variables)? Could also do that in a separate PR

fkiraly: I'm using the refactor principle "avoid proliferation of scope" which means to keep changes localized behind natural interfaces or localized at some semantic principle. That is in general safer, if you don't refactor the monkey with the banana and the entire jungle, for using a `sklearn` metaphor.

I'm only changing how the variables are imported/created, not how they are used. Your suggestion would additionally require to ensure that the refactor would have to go carefully through the entire testing module to ensure all calls to `VALID_ESTIMATOR_TAGS` don't break.

Variable renaming and/or type change is a textbook candidate for a single change PR.

fkiraly: so, yes, why not, but in a separate PR?

fkiraly:
merged since it sounds like we agreed except on variable names - so I can work on #1071 now. Will open an issue for the variable renaming.

Lovkush-A:
> merged since it sounds like we agreed except on variable names - so I can work on #1071 now. Will open an issue for the variable renaming.

Wait, did you merge with this without getting a formal Approval? If so, it seems like you're setting a bad precedent and vearing away from standard github workflows/good practice.

mloning:
I was happy with the change in the end, but I agree it's usually best to wait for approval.

Issue #981, opened 2021-06-19
Refactor tags


mloning:
**Is your feature request related to a problem? Please describe.**
* use of "capabilities" rather than "tags" in classification module
* unclear naming of `_all_tags` method
* inconsistent tags for composition algorithms (see #977)
* integration in `all_estimators` function to make them usable in finding/filtering models

**Describe the solution you'd like**
* refactor capabilities into tags
* rename `_all_tags` to `get_tags` and make public
* add tag kwarg to `all_estimators`
* add tests to catch inconsistent tags

mloning:
From #977:

> A few thoughts:
> * I see tags as being defined on the class (not object) level, conceptually I think of them as describing algorithm properties, so I would prefer not to update them dynamically
> * In the cases of the above mentioned (composition) algorithms, for the implementation (i.e. input validation of the forecasting horizon), the tag should be set to False, the composite doesn't require the forecasting horizon during fitting, it merely passes it on to the components, which have their own tag and raise an error accordingly if they require the forecasting horizon during fitting
> * For documentation, when users may use tags for finding relevant models, setting it to False may be misleading, but conceptually still makes sense, as it refers to the composition algorithm, rather than a particular instance of the composite

fkiraly:
>I see tags as being defined on the class (not object) level, conceptually I think of them as describing algorithm properties, so I would prefer not to update them dynamically

Yes, but do you see the problem? The properties of composites may depend on components.
Especially in pipelines this is important, so it would be good if there is something like a "dynamic tag".
At least, it should be settable or overridable at construction.

>the tag should be set to False, the composite doesn't require the forecasting horizon during fitting, it merely passes it on to the components, which have their own tag and raise an error accordingly if they require the forecasting horizon during fitting

disagreed!

I think each forecaster should be able to tell the user in advance whether it will raise an error when no `fh` is passed to fit. What else is the purpose of these tags? That's also what you are testing for.

Now the stage at which the forecaster should be able to tell that is up for discussion though, i.m.o - my preference would be when constructed, i.e., as an object.

TonyBagnall:
[...]

PR #1068, opened 2021-06-27
Bugfix - forecasters should use internal interface _all_tags for self-inspection, not _has_tag

fkiraly:
Forecasters were, strangely, using the *external* interface point `_has_tags` to inspect their own tags, instead of the already existing internal interface point `_all_tags`.

I believe this must be a mistake? Why should any class call an external function to inspect its own attributes?

I have fixed this ins two places - the `_set_fh` in `BaseForecaster` and the `TransformedTargetForecaster`.

Furthermore, is the "key not found" default in `TransformedTargetForecaster` for `"skip-inverse-transform"` correct? Should it be `True` not `False` perhaps?

I believe this is also what must have been causing problems with #933?
@thayeylolu, is this the same block of code that you removed to make it work?


thayeylolu:
N o @fkiraly , it is the `update of transform` block I commented

mloning (commented):
I think `_has_tag` reads better but don't have a strong preference.

fkiraly:
>I think `_has_tag` reads better but don't have a strong preference. 

Just to clarify, this PR is not about the naming, and the problem is not about the naming.

It's about an external method being called from inside a class to inspect an attribute of the same class or object. That should not happen, it's bad design. If inspection of an attribute is needed, use a getter or inspect the public attribute directly. There should not be any "floating functions" whose purpose it is to inspect class/object attributes and which you need to import.

`_has_tag` is such a floating function and hence should be removed entirely, to be replaced by use of class/object method that is bundled with the class/instance - such as `_all_tags`, or `_has_tags` moved into the class if you really insist (but then the handling of tag defaults would become duplicated, once by `_all_tags` inheritance, and once by the default in `_has_tags`).

mloning:
I see your point, but these functions are quite common in Python, see e.g. `check_is_fitted` in scikit-learn or the built-in function `getattr`, I think they're good for readability and setting of defaults. 

Perhaps we could implement `has_tag` as a public method on the `BaseObject`?

fkiraly:
>I see your point, but these functions are quite common in Python, see e.g. `check_is_fitted` in scikit-learn or the built-in function `getattr`, I think they're good for readability and setting of defaults.

I don't think we need to obsessively copy all of `scikit-learn`'s mistakes...

>Perhaps we could implement `has_tag` as a public method on the `BaseObject`? 

Yes, that's a better idea, so I would suggest to accept this PR and deprecate `_has_tag`, and include a `get_tag` in #891.
Though let's consider the alternative: what's wrong with `_all_tags().get("name", default_value)`? Would it not be un-zen to have two obvious ways to do things, i.e., also `get_tag("name", default_value)`?

Btw, if you call it `has_tag`, the user expectation is that the return is `bool` and means whether the object has the tag, which is not what the function does. So it should be named `get_tag`.

fkiraly:
oops, accidentally edited your comment instead of posting a new one - trying to revert.

mloning:
Does `_has_tag` do something else? It asks if the given tag is true for the given object/class and if it doesn't have the tag it sets it to False. I think that makes sense, at least for now, where not all estimators have all tags. I think that's useful and more readable than `_all_tags().get("name", default_value)`. There is also the built-in `hasattr` function, so that naming and syntax should be familiar to many. 

I believe this is conditional on the tags refactoring in #981.

fkiraly:
> believe this is conditional on the tags refactoring in #981.

No - what I'm removing is just self-inspection, that should use no internal method.

Regarding external inspection, we can discuss, no strong opinion. But a class should not inspect its own attributes using an external method that needs importing, really.

mloning:
> Regarding external inspection, we can discuss, no strong opinion. But a class should not inspect its own attributes using an external method that needs importing, really.

As said, this is really common, `hasattr` is used everywhere, though I agree with the overall principle, especially since tags should be a public interface point. But in its current state I think the readability is worse now in this PR. It also changes the default values, so why don't we wait until we have figured out #981, so that we first agree on how to handle defaults and public interface points?

fkiraly:
Could we do *something* before we wait for the length discussion in #981?

This ties the registry and individual methods and the test suite together via this `_has_tag` method that now sits with the test suite. This looks really brittle to me, could we please at least uncouple the utility from the individual estimators?

So any later changes to the tag system would not interfere with `has_tag`? There are really just these two instances that couple the estimators with the `_has_tag`, and the rest doesn't in cases where tags are read.

We can easily change this later by searching for all the instances where `_all_tags` or `_tags` is called.

mloning (commented):
I'll leave this PR to you @fkiraly, happy if you think it's good to merge this now, otherwise I'd close it until we discuss tags more thoroughly.

fkiraly:
I'd prefer to merge this and uncouple `_has_tags` before the release. I'll do so once tests pass.

Issue #1062, opened 2021-06-26
Sktime metrics rework


chrisholder:
**Is your feature request related to a problem? Please describe.**
When compared to the vast metric library that sklearn has, sktime is missing a great deal of structure. Additionally some of the metrics in Sktime are inconsistent with sklearn. See: https://scikit-learn.org/stable/modules/classes.html#module-sklearn.metrics for details on what sklearn has to offer.

In particular as someone working on the clustering side of sktime the metrics are imperative to many of the algorithms for both funcationality and offering evaluation. 

If you consider the current directory of sktime we have distances, benchmarking, performance_metrics in the upper most level, then some plotting stuff in with the utils, some evaluation methods within a type of time series analysis (i.e. forecasting has model_evluation) and some like classifiction where it isn't obvius where the evaluation metrics are. Point being while this is working for now the inconsistency and departure from the structure of sklearn is confusing for both developers and users of the package.

**Describe the solution you'd like**
My proposal is a restructure of sktime so that model selection, classification metrics, regression metrics, forecasting metrics, clustering metrics, pairwise metrics (distances) and plotting are located under the sktime/metrics/ directory which is in line with sklearn. In addition to the restructure I would propose some parts (in particular distance metrics) are reworked and integrated with numba rather than cython for future ease of development (see http://numba.pydata.org/).

In particular I personally would like to prioritise working on the pairwise distance. Currently the distances I'm using for clustering  are located in distances/elastic.py and distances/elastic_cython.pyx. While these distance work and are very well optimised,  they don't follow the interface set out by sklearn (they do not allow for 2 1d arrays to be past). While im not propsing scrapping these overnight I believe overtime they should be redeveloped to follow sklearns interface and move to numba over Cython.

**Describe alternatives you've considered**
Alternativly without this structure I would just develop the distance and evaluation methods I need inside the sktime/clustering directory which further makes the structure of sktime confusing and inconsistent.

**Additional context**
This is a large proposal and restructure. However, I think if we do it incrementally we can get to a good point of consistency. In particular while talking with @TonyBagnall the distance metrics we agreed that accepting 1d arrays is a necessity for Sktime.

TonyBagnall:
Whilst I largely agree about performance measures being grouped, I think your are conflating performance  metrics with distances/metrics, and this is really two issues
1. Redesign distances with numba. I think this would be great if it does not degrade performance, but I think it makes sense to keep distances a separate module
[...]

Issue #52, opened 2019-05-07
Design 2nd degree transformer


mloning:
As discussed in #5

fkiraly:
Implemented here?
https://github.com/goastler/sktime-kernels/blob/master/sktime/transformers/kernels.py

Looks nice, @goastler !

fkiraly:
ok, based on discussion on slack, here's a generic design and how it should interact with pipeline composition and tuning

fkiraly:
slightly modified from #5 since decided to not use xpandas

Part 3. Pairwise transformer aka degree 2 transformer.

should be a class with parameters, attributes, methods.
Does not use abstract base class. Inherit from sklearn.BaseEstimator, or transformer

Parameters - private variables. Correspond to "hyper-parameters" to be set or tuned. Like in supervised estimator.
Attributes - private variables. Correspond to "model parameters" set by the fit method. Not to be set by the user or via an interface.

Constructor _ _ init _ _
arguments: all parameters, explicitly, with sensible default setting.
behavior: Sets self.parameters to the values provided.

public method fit
arguments:
X - a pandas data frame. no default.
optimal argument:
sample_weight - a vector of weights, equal length to X.
behavior:
fits model and stores it in attribute variables. May access but not modify parameters.

public method transform
arguments:
X - an pandas data frame. no default. Column headers and types should be the same as of fit's X argument
Xnew - a pandas data frame. no default. Column headers and types should be the same as of fit's X argument
optimal argument:
sample_weight - a vector of weights, equal length to X.
behavior:
Returns an named 3D array K. First dimension is indexed by rows of X. Second dimension is indexed by rows of Xnew. Third dimension is indexed, with transformed columns having headers defined by the transformer.
May access attributes and hyper-parameters. May not modify attributes and hyper-parameters.

public method get_params
arguments:
deep - a boolean indicating whether parameters of nested estimators should be returned
behaviour:
returns string -> value dictionary of parameters, following the sklearn naming convention (nested estimators' parameters by < estimatorname >__< parametername >, returned if and only if deep = true)

public method set_params
arguments:
string -> value dictionary of parameters, following the sklearn naming convention (nested estimators' parameters by < estimatorname >__< parametername >)
behaviour:
sets parameters, and nested parameters if provided, to the values as defined by the dictionary mapping

fkiraly:
The counterpart of this, in composition is a "distance method", or "kernel method".

Part II: kernel/distance methods

[...]

PR #1071, opened 2021-06-28
Pairwise transformers, kernels/distances on tabular data and panel data - base class, examples, extension templates

fkiraly:
Related to #52 (oldest open issue!) and #1062.

This PR contains:

* base classes for a new scitype: pairwise transformers, i.e., distance matrix makers and kernel matrix makers; there are two base classes for two different inputs, tabular (classical distances/kernels) and time series panels (e.g., time warping distances)
* an example tabular distance `ScipyDist` that interfaces the `scipy` distances
* an example panel distance `AggrDist` that uses any tabular distance to create an aggregate sample distance between time series, e.g., the mean Euclidean distance
* extension templates for both scitypes

This may be useful for future extension of the clustering module, or refactorings and extensions of the TSC module, @chrisholder, @TonyBagnall.

Also pinging @jasonlines since this is a proposal how the existing distances (due to @jasonlines) could be refactored.

Notably, currently this contains no tests. Please look at the design first.


fkiraly:
@goastler was also assigned to #52, so including FYI. Not sure whether that's still current.

fkiraly:
Tests fail because the distances are not yet in the registry (base classes etc). Will deal with that later, please review the design first.

TonyBagnall:
my first comment is I dont like the naming conventions. I have no idea what a TrafoPw is, I dont mind longer names if they can retain some meaning

fkiraly:
>  I have no idea what a TrafoPw is, I dont mind longer names if they can retain some meaning

a pairwise transformer! What would you suggest instead?

Note that this has the same structure for distances and kernels, so choosing either would be misleading.

fkiraly:
Maybe @Lovkush-A can help. He usually has strong but sensible opinions about variable names.

TonyBagnall:
> > I have no idea what a TrafoPw is, I dont mind longer names if they can retain some meaning
> 
> a pairwise transformer! What would you suggest instead?
> 
err, PairwiseTransformer?

fkiraly:
Only if we can call it `PairwiseseTransformerer`, @chrisholder.

chrisholder:
I apologise in advance if I've oversimplfied this or if I am missing an obvious use case for this which is why you've done it the way you have. From my understanding really we just want users to be writing the distance/kernal function rather than the transformer itself to get the matrix.

So my comment is about

`def _transform(self, X, X2=None):`

I think it's potentially a better idea to ask the user to write a distance/kernal function and leave the logic and optimisation of how the matrix is computed to sklearn. This isn't fully done but just quick mock up example (untested and written in github so sorry for any syntax errors)

```
from sklearn.metrics.pairwise import pairwise_distances

class BaseTrafoPw(BaseEstimator):

    def __init__(self, metric, n_jobs=none, kwds=none):
        self.metric = metric
        self.njobs = n_jobs
        self.kwds = kwds

    def transform(self, X, X2=None):
        X = check_series(X)

        if X2 is None:
            X = X2
            self.symmetric = True
        else:
            X2 = check_series(X2)
            self.symmetric = False

        return pairwise_distances(X=X, 
                                      X2=X2, 
                                      metric=self._calculate_distance, 
                                      n_jobs=self.n_jobs,
                                      kwds=self.kwds)

    def _calculate_distance(self, X, Y):
        raise NotImplementedError
```

The sklearn pairwise_distances is incredibly well optimised (compiled to low level) and allows for threading as well so could be a better solution than making the user write the logic for it. In addition the pairwise_distance function accepts a string (that are the scipy distance metrics) or a distance function of the for d(X, Y), and so it would mean we only need this one class rather than another class just to accept string input. Here is the full documentation for pairwise_distance:

https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise_distances.html#sklearn.metrics.pairwise_distances

There are also further Kernal pairwise metrics in the package and so pairwise_distances can be swapped out for something like sigmoid_kernel(X[, Y, …]) etc. very easily. Full list see here:

https://scikit-learn.org/stable/modules/classes.html#module-sklearn.metrics.pairwise

Again apologise if this is quite a narrow perspective or missing the bigger picture of how you intended this to be used.

Lovkush-A:
> I have no idea what a TrafoPw is

@fkiraly How do you come up with these variable names?!

fkiraly:
Silly question, why do you think we're *not* leaving the implementation of the matrix computation to the concrete class or the interface package? `_transform` is where you would implement or interface it, in a concrete child class.

`transform` is just for input checking logic, no computation happens there.

fkiraly:
>@fkiraly How do you come up with these variable names?!

Haruspicy on weekdays, falling asleep on my keyboard on week-ends.

chrisholder:
> Silly question, why do you think we're _not_ leaving the implementation of the matrix computation to the concrete class or the interface package? `_transform` is where you would implement or interface it, in a concrete child class.
> 
> See the `scipy.dist` interface as an example, the logic happens all in the highly optimized `scipy` functions.
> 
> `transform` is just for input checking logic, no computation happens there.

In def _transform() you have the comment  

`returns pairwise distance/kernel matrix between samples in X and X2 (equal to X if not passed) core logic`

which implies you're letting a user implement this

fkiraly:
>which implies you're letting a user implement this

or interface from a package like `scipy`, have a look at the `ScipyDist` class which is an example for precisely this.

That you're in a function doesn't mean you can't call other functions? Getting increasingly confused there, mayby I'm grossly overlooking sth obvious.

chrisholder:
O I see where we're missing each other. You're in favour of giving the user the option to do what they want. Im saying well how many people are going to need a custom solution that is not one of these prebuilt pairwises like sklearn or scipy and so just put it as default behaviour. If they want to use something custom then give them the option to override it but thats a more explicit choice.

fkiraly:
>You're in favour of giving the user the option to do what they want. Im saying well how many people are going to need a custom solution that is not one of these prebuilt pairwises like sklearn or scipy and so just put it as default behaviour. If they want to use something custom then give them the option to override it but thats a more explicit choice.

Not sure what the alternative is that you are comparing it to - just call some distance function from somewhere?

That alternative would not allow parameter inspection in composites, and tuning of parameters in distances.
Time warping distances and kernels notoriously have parameters that should be tuned, so they should be exposed.

E.g., if you write sth like `KNeighborsClust(dist=DtwDist(skip_penalty=9), k=10)`, you want to be able to access the `skip_penalty` for tuning via the `sklearn` type deep parameter access convention by reference to `dist__skip_penalty`.

Another advantage of this design is that it makes any distances and kernels universally composable with methods that take distances or kernels. This is similar to how the the Gaussian Process module in `sklearn` works.

Given the above, and that time series distances will be a recurring motif across modules rather than just localized to one, I think this makes a lot of sense...

chrisholder:
Yep actually I agree with you when you put it like that. Happy with that then looks good! How close is this to be being ready to pr just so I have an idea of timeline as I have a few distances I'd quite like to wrap in what you've done here.

fkiraly:
>How close is this to be being ready to pr just so I have an idea of timeline as I have a few distances I'd quite like to wrap in what you've done here.

I guess it's ready except if you want tests...

help is appreciated, but let's coordinate.

chrisholder:
I can hop on the branch and write some tests if that would make your life easier?

fkiraly:
yes, indeed - though I've had hoped that #1071 is merged first, that would make it easier to add the base class etc and the tests.

Lovkush-A:
@chrisholder These templates are intended for developers, not end users (I think). Sounds like you thought Franz wanted users to create their own distance-matrix functions.

@fkiraly I am not a fan of the name 'PairwiseTransformers'. Calculating the distance between X[i] and X2[j] is not the same kind of thing as doing a log-transform or re-scaling or other transforms we already have. I couldn't think of a better alternative by myself, but fortunately the naming conventions in sklearn seem decent to me: https://scikit-learn.org/stable/modules/metrics.html#

Just use the word 'Pairwise' by itself for base items, then have PairwiseMetric or PairwiseKernel once you know what kind of thing it is.

fkiraly:
>while you rename all your variables

If you keep making fun of me like this, I will rename all variables to match the regular expression `__*`

chrisholder:
Well im in no rush with it getting 100% finalised as I can still write my stuff and just merge it into contrib for now until this is finalised

fkiraly:
> am not a fan of the name 'PairwiseTransformers'

So, how about my newest suggestion, `__________________________`?

chrisholder:
Make them into faces

-________-

x_______x

fkiraly:
@chrisholder, ok, let me merge #1071 - it sounded like @mloning was all happy with it anyway, and then you can get started on tests while I work on variable names and registry, how about that?

Lovkush-A:
> So, @Lovkush-A, how about my newest suggestion, `__________________________`?

I'd prefer it if there were 42 underscores

Lovkush-A:
> @chrisholder, ok, let me merge #1071 - it sounded like @mloning was all happy with it anyway, and then you can get started on tests while I work on variable names and registry, how about that?

1071 is *this* PR...?

fkiraly:
>1071 is this PR...?

Sorry, I meant #1067 which I just merged.

fkiraly:
Alright, @chrisholder, happy for you to write some tests - I'll register the base class and will rename it to `BasePairwiseTransformer` but will call it a day then. Would you like to PR to this branch?

chrisholder:
Yea ill pr into the branch think that makes the most amount of sense

Lovkush-A:
> Alright, @chrisholder, happy for you to write some tests - I'll register the base class and will rename it to `BasePairwiseTransformer` but will call it a day then. Would you like to PR to this branch?

Does that mean you don't like `BasePairwise`?

extension_templates/dist_kern_panel.py
```diff
@@ -0,0 +1,105 @@
...
+Mandatory implements:
+    transforming    - _transform(self, X, X2=None)
+
+State:
+    none, this is a state-free scitype
+
+copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+"""
+
+from sktime.dists_kernels import BaseTrafoPw
```
Lovkush-A, line 25: You imported tabular base class, not panel.

extension_templates/dist_kern_tab.py
```diff
@@ -0,0 +1,105 @@
...
+Mandatory implements:
+    transforming    - _transform(self, X, X2=None)
+
+State:
+    none, this is a state-free scitype
+
+copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+"""
+
+from sktime.dists_kernels import BaseTrafoPwPanel
```
Lovkush-A, line 25: should be tabular import

fkiraly: yes, was wrong way round in a number of places in the extension template, fixed it

fkiraly:
>Does that mean you don't like BasePairwise?

Pairwise what? I find this not descriptive enough and slightly misleading

sktime/dists_kernels/compose_tab_to_panel.py
```diff
@@ -0,0 +1,78 @@
...
+    pairwise distance based on aggregation function applied to sample-wise distance
+        example: AggrDist(ScipyDist()) is mean Euclidean distance between series
+
+    Components
+    ----------
+    transformer: pairwise transformer of BaseTrafoPw scitype (tabular pairwise)
+
+    Hyper-parameters
+    ----------------
+    aggfun: aggregation function 2D np.array -> float
```
Lovkush-A, line 24:
```suggestion
    aggfunc: aggregation function 2D np.array -> float
```
Only because aggfunc is used elsewhere in sktime, e.g. https://github.com/alan-turing-institute/sktime/blob/main/sktime/forecasting/compose/_ensemble.py

sktime/dists_kernels/compose_tab_to_panel.py
```diff
@@ -0,0 +1,78 @@
...
+"""
+
+__author__ = ["fkiraly"]
+
+import numpy as np
+
+from sktime.dists_kernels._base import BaseTrafoPwPanel
+
+
+class AggrDist(BaseTrafoPwPanel):
```
Lovkush-A:
```suggestion
class PairwiseAggrDist(BaseTrafoPwPanel):
```
I think Pairwise should be part of the name of any of these classes.

sktime/dists_kernels/compose_tab_to_panel.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+    pairwise distance based on aggregation function applied to sample-wise distance
+        example: AggrDist(ScipyDist()) is mean Euclidean distance between series
+
+    Components
+    ----------
+    transformer: pairwise transformer of BaseTrafoPw scitype (tabular pairwise)
```
Lovkush-A, line 20: the fact you included '(tabular pairwise)' is indication that having 'tabular' as part of the baseclass name would be helpful

sktime/dists_kernels/_base.py
```diff
@@ -0,0 +1,198 @@
...
+
+__author__ = ["fkiraly"]
+
+
+from sktime.base import BaseEstimator
+
+from sktime.utils.validation.series import check_series
+
+
+class BaseTrafoPw(BaseEstimator):
```
Lovkush-A: The word `Tabular` should be added to the end of the base class name, in my view.  (So either `BasePairwiseTransformerTabular` or `BasePairwiseTabular`).  Things are starting to get long and unwieldy, but I think its just about ok...

sktime/dists_kernels/compose_tab_to_panel.py
fkiraly: I don't know, I would find it confusing - better have a base concept, and then `Panel` is appended for modifying the meaning.

sktime/dists_kernels/compose_tab_to_panel.py
fkiraly: ah, no fun(c)

sktime/dists_kernels/compose_tab_to_panel.py
```diff
@@ -0,0 +1,78 @@
...
+        symmetric = self.symmetric
+        aggfun = self.aggfun
+
+        distmat = np.zeros((n, m), dtype="float")
+
+        for i in range(n):
+            for j in range(m):
+
+                if symmetric and j < i:
+                    distmat[i, j] = distmat[j, i]
```
Lovkush-A, line 74: Neat way to enforce symmetry!

sktime/dists_kernels/scipy_dist.py
```diff
@@ -0,0 +1,80 @@
...
+__author__ = ["fkiraly"]
+
+import pandas as pd
+
+from scipy.spatial.distance import cdist
+
+from sktime.dists_kernels._base import BaseTrafoPw
+
+
+class ScipyDist(BaseTrafoPw):
```
Lovkush-A:
```suggestion
class PairwiseScipyDist(BaseTrafoPw):
```
Similar to above, I think Pairwise should be part of class names

sktime/dists_kernels/compose_tab_to_panel.py
fkiraly: note to self: there may be a todo here - not all pairwise transformers are symmetric. So there should be a tag, and this should only be done if `self.transformer` has the "symmetric" tag

sktime/dists_kernels/scipy_dist.py
fkiraly: a distance is always pairwise, I find this redundant

fkiraly: maybe the word "pairwise" is not good - what this is supposed to mean is "of pairs"

extension_templates/dist_kern_panel.py
fkiraly: yes, thanks, fixed it

sktime/dists_kernels/scipy_dist.py
```diff
@@ -0,0 +1,80 @@
...
+                'rogerstanimoto', 'russellrao', 'seuclidean', 'sokalmichener',
+                'sokalsneath', 'sqeuclidean', 'yule'
+            if function, should have signature 1D-np.array x 1D-np.array -> float
+        p: if metric='minkowski', the "p" in "p-norm", otherwise irrelevant
+        colalign: string, one of 'intersect' (default), 'force-align', 'none'
+            controls column alignment if X, X2 passed in fit are pd.DataFrame
+            columns between X and X2 are aligned via column names
+            if 'intersect', distance is computed on columns occurring both in X and X2,
+                other columns are discarded; column ordering in X2 is copied from X
+            if 'force-align', raises an error if the set of columns in X, X2 differs
```
Lovkush-A, line 36:
```suggestion
            if 'force-align', raises an error if the set of columns in X, X2 differs;
            if set of columns are the same, column order in X2 is copied from X
```

sktime/dists_kernels/scipy_dist.py
```diff
@@ -0,0 +1,80 @@
...
+
+        Args:
+            X: pd.DataFrame of length n, or 2D np.array of 'float' with n rows
+
+        Optional args:
+            X2: pd.DataFrame of length m, or 2D np.array of 'float' with m rows
+
+        Returns:
+            distmat: np.array of shape [n, m]
+                (i,j)-th entry contains distance/kernel between X.iloc[i] and X2.iloc[j]
```
Lovkush-A, line 65:
```suggestion
                (i,j)-th entry contains distance between X.iloc[i] and X2.iloc[j]
```

sktime/dists_kernels/scipy_dist.py
Lovkush-A: Pairwise is standard term for doing stuff to each pair of two sets/lists, I think.

sktime/dists_kernels/compose_tab_to_panel.py
Lovkush-A: > > Does that mean you don't like BasePairwise?
> 
> Pairwise what? I find this not descriptive enough and slightly misleading

I agree that Pairwise byitself leaves something missing, but:
* Pairwise is used in sklearn (does not mean it is a good name, but it *is* an advantage of Pairwise)
* I think PairwiseTransformer is more misleading than just Pairwise, because it makes it sound like it is a similar kind of thing as the other Transformers in sktime, which I dont think it is.

Lovkush-A:
> > Does that mean you don't like BasePairwise?
> 
> Pairwise what? I find this not descriptive enough and slightly misleading

> > Does that mean you don't like BasePairwise?
> 
> Pairwise what? I find this not descriptive enough and slightly misleading

I agree that Pairwise byitself leaves something missing, but:
* Pairwise is used in sklearn (does not mean it is a good name, but it *is* an advantage of Pairwise)
* I think PairwiseTransformer is more misleading than just Pairwise, because it makes it sound like it is a similar kind of thing as the other Transformers in sktime, which I dont think it is.

sktime/dists_kernels/compose_tab_to_panel.py
Lovkush-A: You should compete in the UK pun championships: https://www.youtube.com/watch?v=hriVI0x_OEc

sktime/dists_kernels/scipy_dist.py
Lovkush-A: E.g. see use of pairwise in these contexts:
https://mathworld.wolfram.com/Pairwise.html
https://en.wikipedia.org/wiki/Pairwise_independence
https://en.wikipedia.org/wiki/Disjoint_sets

fkiraly:
>I think `PairwiseTransformer` is more misleading than just `Pairwise`, because it makes it sound like it is a similar kind of thing as the other Transformers in sktime, which I dont think it is.

But it is very similar, it's a direct generalisation! A transformer produces values for each input row. The pairwise transformer produces values for each pair of `X` input row and `X` input row. Strictly speaking (scitype model wise), you would have to allow multiple values to be produced and the result to be a 3D data frame, and the pairwise transforme to be fittable, but those are rare cases.

A triple-wise transformer would create values for each triple of rows of the inputs etc.

sktime/dists_kernels/scipy_dist.py
fkiraly: why that? This can be a kernel, too.

sktime/dists_kernels/scipy_dist.py
fkiraly: yes, but my problem is not that I think it's inaccurate, just that it's redundant.

sktime/dists_kernels/compose_tab_to_panel.py
fkiraly: I think I already won without participating.

sktime/dists_kernels/compose_tab_to_panel.py
fkiraly: well, but then @mloning would say, it should be at the end, to be in line with `sklearn` convention where things like `Classifier` etc come at the end. I don't like that since then the different scitypes don't come in blocks when ordering by alphabet, but I think it's better to conform with `sklearn` convention than inventing sth new.

sktime/dists_kernels/scipy_dist.py
Lovkush-A: It's in a class called PairwiseScipyDist... I assume it only computes distances. If it can calculate kernels too, then maybe the class should be renamed.

Lovkush-A:
> it's a direct generalisation!

* The fact it is a generalisation is the issue; by generalising you have created a distinctly new kind of thing. (It is a similar reason I think 'semigroup' is a bad name; semigroups aren't groups, but a generalisation of them.)

> A transformer produces values for each input row.

* For me, this is a key feature of a transformer. Each data entry goes from some old representation to some new representation.

* Lastly, the individual functions (e.g. euclidean distance metric) that are applied are not referred to as transformations; this is an indication that transformations are not the right word/concept. More explicitly, if you have two points in euclidean space, you'd say a rotation of the space is a transformation, but you wouldn't say that calculating the distance between them was a transformation.

sktime/dists_kernels/scipy_dist.py
Lovkush-A: Pairwise tells us we are operating on two list-like objects, and calculating for every pair between those two lists.

Without pairwise, I think `ScipyDist` best describes something like the euclidean distance function, which does a single distance calculation.

sktime/dists_kernels/compose_tab_to_panel.py
Lovkush-A: I like 'ordering by alphabet' as a criteria for how things should be named! Makes good sense.

But in this particular instance, I do not have strong feelings on whether Pairwise goes at the front or the end. I prefer front just because thats how its used in mathematical phrases

sktime/dists_kernels/compose_tab_to_panel.py
Lovkush-A: You first need a competitor name, which must be a pun containing 'pun', before you can even claim to have won the contest

fkiraly: fine. How about "puncipal data scientist"?

Also, one of the greatest python packages was created when a larger number of telepaths met to play computer games: they formed a psychic LAN.

sktime/registry/_base_classes.py
```diff
@@ -77,6 +81,16 @@
     ("regressor", BaseRegressor, "time series regressor"),
     ("forecaster", BaseForecaster, "forecaster"),
     ("transformer", BaseTransformer, "time series transformer"),
+    (
+        "transformer-pairwise-tab",
+        BasePairwiseTransformer,
+        "pairwise transformer for tabular data, distance or kernel",
+    ),
+    (
+        "transformer-pairwise-ts",
```
chrisholder, line 90: Why is this 'transformer-pairwise-ts'. Wouldn't something like 'transformer-pairwise-pan' be more inline with the other you've named 'transformer-pairwise-tab'

sktime/registry/_base_classes.py
```diff
@@ -77,6 +81,16 @@
...
+        BasePairwiseTransformerPanel,
+        "pairwise transformer for tabular data, distance or kernel",
```
chrisholder, line 92: Update this to "pairwise transformer for panel data, distance or kernel"

fkiraly:
>It is a similar reason I think 'semigroup' is a bad name; semigroups aren't groups, but a generalisation of them.

I don't see this, @Lovkush-A, why is there a problem? Every group is also a semigroup, that's how it intuitively should be.

Like every vehicle that has two tires also has one tire (proof: choose one of the two, by assumption it's a tire on the vehicle). Similarly, an object that has 1 groupness is also an object that has 1/2 groupness (semi = 1/2 or partly, in Greek).

>For me, this is a key feature of a transformer. Each data entry goes from some old representation to some new representation.

Yes. And in the pairwise transformer, each pair of data entries goes from some old representation to some new representation.

>Lastly, the individual functions (e.g. euclidean distance metric) that are applied are not referred to as transformations; this is an indication that transformations are not the right word/concept.

But I refer to them as transformations. Pre-`sklearn`, you would probably also have hard times finding a source that refers to transformations as such.

Lovkush-A:
> > Lastly, the individual functions (e.g. euclidean distance metric) that are applied are not referred to as transformations
> 
> But I refer to them as transformations. Pre-`sklearn`, you would probably also have hard times finding a source that refers to transformations as such.

what do you mean by this?
* Did sklearn change how the word 'transformation' was used, and if so in what way?
* Do you consider yourself in the majority or minority when you refer to things like euclidean metrics as transformations?

------
Anyway, I think we just have different prior perceptions of the word 'transformer' and the grammar of 'PairwiseTransformer'. Ideally, we would find out what other people's intuitions are about  the name 'PairwiseTransformers', but in the meantime, we might as well continue with your naming. It is easy to change later if other developers/users find it confusing.

fkiraly:
>Ideally, we would find out what other people's intuitions are about the name 'PairwiseTransformers', but in the meantime, we might as well continue with your naming. It is easy to change later if other developers/users find it confusing.

Well, it's the name for a base class, which most users won't ever see, or only rarely. Concrete classes would be called `ScipyDistance` or `DtwKernel` or similar.

sktime/dists_kernels/scipy_dist.py
```diff
@@ -0,0 +1,80 @@
...
+        p = self.p
+        metric = self.metric
+
+        if isinstance(X, pd.DataFrame):
+            X = X.select_dtypes("number").to_numpy(dtype="float")
+
+        if isinstance(X2, pd.DataFrame):
+            X2 = X2.select_dtypes("number").to_numpy(dtype="float")
+
+        distmat = cdist(XA=X, XB=X2, metric=metric, p=p)
```
chrisholder, line 78: So potententially I have something wrong locally but when I try run this code passing the parameters p=p to cdist it causes a TypeError. Which is interesting because it defaults to 2 (which is the valueu being passed as p set by your default parameter in the constructor). So I don't know why this is caused. Removing the p parameter it runs fine (even thought cdist defaults p to 2).

fkiraly: strange, let me try this locally again. What is the message in your `TypeError`?

chrisholder: So something that I found that doesn't seem intentional from cdist is that passing p throws an error using any other metric than "minkowski". Even passing p=None throws an error. I would assume this is a bug because surely you just ignore the parameter when you're not using minkowski

fkiraly: yes, there was a bug which I fixed if you didn't pass `X2`.
It was in the base class `transform` input logic:

```python
    if X2 is None:
        X = X2
```
(old state)

d'oh

sktime/registry/_base_classes.py
fkiraly: fixed

sktime/registry/_base_classes.py
fkiraly: fixed

sktime/dists_kernels/scipy_dist.py
chrisholder: O good spot I missed that too 😂

sktime/dists_kernels/scipy_dist.py
fkiraly: ah, of course - in the concrete class. fixed.

sktime/dists_kernels/scipy_dist.py
fkiraly: why `;`?

sktime/dists_kernels/scipy_dist.py
fkiraly: >Pairwise tells us we are operating on two list-like objects, and calculating for every pair between those two lists.

by the same logic you would call a `sklearn` transformer not a transformer but an element-wise transformer, because it calculates the transformation individually for each instance in the data frame or array.

sktime/dists_kernels/_base.py
fkiraly: hm, undecided here. @chrisholder, any strong opinions?

sktime/dists_kernels/compose_tab_to_panel.py
fkiraly: >I like 'ordering by alphabet' as a criteria for how things should be named! Makes good sense.

... was this sarcasm or genuine? Hard to tell from just the writing.

sktime/dists_kernels/scipy_dist.py
fkiraly: is this your problem? If not, please paste code

Lovkush-A:
Btw, @chrisholder, please may you describe the (precise, if possible) sequence of steps/commands you used to contribute to this PR? I have worked something out by googling around (but it feels convoluated), so knowing what you do will either show me a better way or reassure me.

sktime/dists_kernels/compose_tab_to_panel.py
Lovkush-A: This was genuine! I already do this - naming things so that sorting by alphabet groups things sensibly -  in two or three large-ish folders on my laptop.

sktime/dists_kernels/scipy_dist.py
Lovkush-A: > by the same logic you would call a `sklearn`an element-wise transformer

I see the logic, but the element-wise transformer is well-established as the default or base concept.  The question we should be asking ourselves is: if a user saw the class `ScipyDist` for the first time, would they predict it does calculations elementwise or pairwise? If you find *anybody* whose first thought is that it does pairwise calculations, I will be very surprised.

sktime/dists_kernels/scipy_dist.py
Lovkush-A: Its not only the `;`, it is line 37 too

sktime/dists_kernels/scipy_dist.py
Lovkush-A: Interestingly, it is curious how we are on opposites sides of the same argument.

* I prefer `BasePairwiseTransformerTabular` and `BasePairwiseTransformerPanel`, whereas you prefer having `BasePairwiseTransformer`
* In the hypothetical scenario where we created both pairwise and element-wise wrappers for scipy dist, I would prefer `PairwiseScipyDist` and `ScipyDist` whereas you (appear to) prefer `PairwiseScipyDist` and `ElementwiseScipyDist`

I am comfortable being on different sides, because I think the element-wise calculations are a well-established norm and base, whereas I don't think any norms are established for BasePairwiseTransformers.

chrisholder:
@Lovkush-A Well I haven't contributed to it yet but currently im working in my fork. Fetched the branch and just committing on the branch. Then when it's ready I pr into the branch.

So in terms of steps just 

In your fork

git fetch upstream trafo-distances
git checkout trafo-distances

do your changes and commit them to your fork

then pr from your forks trafo-distances branch into the main repos trafo-distances branch.

fkiraly:
well, my preferred process would be a *pull request* to this branch, not direct commits, so I have an opportunity to review.

The workflow for that would be this:

1. clone the repo
2. create a branch based on `trafo-distances`, e.g., `trafo-distances-plus-tests` or similar
3. work on that branch
4. when done or ready to show, push to fork (if not core dev) or `sktime` (if core dev)
5. make a PR from `trafo-distances-plus-tests` into `trafo-distances`
6. I review, we discuss, we merge into `trafo-distances`
7. then we make a PR into `main`, and things proceed as usual

fkiraly:
(but I don't mind direct commits either in this case)

chrisholder:
O sorry if that wasnt clear that is how I was doing it @fkiraly

Lovkush-A:
Thanks!

@fkiraly Direct pushes into 'trafo-distances' would only be possible if you gave write permission, right?


@chrisholder. did you create a fork of Franz's fork of sktime, or did you only need a fork of sktime?

> then merge my trafo-distances into alan turing institutes sktimes branch trafo-distances.

Does the branch trafo-distances exist in sktime repo? I thought it was only in franz's fork.

chrisholder:
Just a regular fork follow the steps https://www.sktime.org/en/latest/contributing.html#id9 to get your fork. Then when you need to update or grab the latest version of main you do

`git fetch upstream`

and if you want a specific branch

`git fetch upstream <branch-name>`

fkiraly:
>@fkiraly Direct pushes into 'trafo-distances' would only be possible if you gave write permission, right?

Yes, core devs have those permissions.

chrisholder:
Sorry just to be clear on that point while you cant directly pr into branches and make it go in straight away, you can pr and wait for it to be approved just like you can for pring into main. So just pr into the branch and wait for Franz to approve it.

sktime/dists_kernels/scipy_dist.py
chrisholder: Sorry @fkiraly just got around to testing again. Same issue the problem is coming from cdist itself when you pass p as a parameter when it isnt used (i.e. for any other metric not 'minkowski').

Just to be more explicit with the error so you can hopefully recreate and see what I mean. Here is (and sorry for no syntax highlighting this is done in just terminal to get it in complete isolation), this is what im on about

image

As you can see the function was working fine with euclidean until I added "p" as a parameter where it threw an error as euclidean doesn't use p. Therefore if you're not using "minkowski" as the metric you can't pass p as a parameter or an error is thrown.

My current work around is this:

```
def _transform(self, X, X2=None):
    p = self.p
    metric = self.metric
    
      X = X.select_dtypes("number").to_numpy(dtype="float")
    
    if isinstance(X2, pd.DataFrame):
      X2 = X2.select_dtypes("number").to_numpy(dtype="float")
    
    if metric is "minkowski":
      distmat = cdist(XA=X, XB=X2, metric=metric, p=p)
    else:
      distmat = cdist(XA=X, XB=X2, metric=metric)
    
    return distmat
```

fkiraly: this is resolved via @chrisholder's PR

sktime/dists_kernels/compose_tab_to_panel.py
fkiraly: further discussed in comment below

sktime/dists_kernels/_base.py
fkiraly: further discussed in comment below

fkiraly:
@chrisholder and @Lovkush-A, you need to re-review since changes were made by @chrisholder and merged into my PR

sktime/dists_kernels/_base.py
```diff
@@ -0,0 +1,251 @@
...
+
+    Returns
+    -------
+    X:
+        Checked and converted to List of pd.Dataframe if one
+        of the other formats
+
+    """
+
+    def arr_check(arr):
```
Lovkush-A, line 129: Minor comment:
seems odd to have function defined within function. any particular reason not to have def arr_check as a separate helper function?

sktime/dists_kernels/_base.py
```diff
@@ -0,0 +1,251 @@
...
+            default X2 = X
+
+        Returns
+        -------
+        distmat: np.array of shape [n, m]
+            (i,j)-th entry contains distance/kernel between X[i] and X2[j]
+
+        Writes to self
+        --------------
+        symmetric: bool = True if X2 was not passed, False if X2 was passed
```
Lovkush-A, line 190: Major comment:
* same as above

sktime/dists_kernels/_base.py
```diff
@@ -0,0 +1,251 @@
...
```
Lovkush-A, line 215: major comment:
* same as above

sktime/dists_kernels/_base.py
```diff
@@ -0,0 +1,251 @@
...
+            for use to make internal calculations efficient, e.g., in _transform
+        """
+
+        X = _pairwise_panel_x_check(X)
+
+        if X2 is None:
+            X2 = X
+            self.symmetric = True
+        else:
+            X2 = _pairwise_panel_x_check(X2)
```
Lovkush-A, line 225: missing `self.symmetric = ...` in the else loop

chrisholder (dismissed):
Happy with all this.

sktime/dists_kernels/compose_tab_to_panel.py
```diff
@@ -0,0 +1,77 @@
...
+__author__ = ["fkiraly"]
+
+import numpy as np
+
+from sktime.dists_kernels._base import BasePairwiseTransformerPanel
+
+
+class AggrDist(BasePairwiseTransformerPanel):
+    """
+    pairwise distance based on aggregation function applied to sample-wise distance
```
Lovkush-A, line 15: Minor (potentially major?) comment:
I don't know what this sentence means. Could be me (I am not used to thinking about panel data) or could be sign docstring needs to be clearer

sktime/dists_kernels/compose_tab_to_panel.py
```diff
@@ -0,0 +1,77 @@
...
+        example: AggrDist(ScipyDist()) is mean Euclidean distance between series
```
Lovkush-A, line 16: Potentially major but I don't know:
* Does `SciPyDist()` default to Euclidean distance? Or is this a typo? Or something else?

sktime/dists_kernels/compose_tab_to_panel.py
```diff
@@ -0,0 +1,77 @@
...
+
+    Components
+    ----------
+    transformer: pairwise transformer of BaseTrafoPw scitype (tabular pairwise)
```
Lovkush-A, line 20: Minor:
```suggestion
    transformer: pairwise transformer of BaseTransformerPairwise scitype (tabular pairwise)
```

sktime/dists_kernels/compose_tab_to_panel.py
```diff
@@ -0,0 +1,77 @@
...
+    ----------------
+
+    aggfunc: aggregation function 2D np.array -> float
+        default = np.mean
+    """
+
+    def __init__(
+        self,
+        transformer,
+        aggfunc=None,
```
Lovkush-A, line 32: Minor comment:
What is the reason reason why default is None instead of np.mean?

sktime/dists_kernels/compose_tab_to_panel.py
```diff
@@ -0,0 +1,77 @@
...
+            aggfunc = np.mean
+
+        distmat = np.zeros((n, m), dtype="float")
+
+        for i in range(n):
+            for j in range(m):
+
+                if symmetric and j < i:
+                    distmat[i, j] = distmat[j, i]
+                elif aggfunc is not None:
```
Lovkush-A, line 74:
```suggestion
                else:
```
I think checking that aggfunc is not None is redundant?

sktime/dists_kernels/compose_tab_to_panel.py
Lovkush-A: Minor:
maybe call this `transformer` something like `tabular_transformer` or `transformer_tabular`?

sktime/dists_kernels/scipy_dist.py
```diff
@@ -0,0 +1,84 @@
...
+__author__ = ["fkiraly"]
+
+import pandas as pd
+
+from scipy.spatial.distance import cdist
+
+from sktime.dists_kernels._base import BasePairwiseTransformer
+
+
+class ScipyDist(BasePairwiseTransformer):
```
Lovkush-A, line 16: minor:
I'd like to have other people's perspective on the name ScipyDist. To me, this sounds like an elementwise transformer to me, rather than a pairwise transformer

sktime/dists_kernels/tests/test_all_dist_kernels.py
```diff
@@ -0,0 +1,135 @@
+# -*- coding: utf-8 -*-
+import numpy as np
+import pytest
+
+from sktime.dists_kernels._base import BasePairwiseTransformer
+from sktime.utils import all_estimators
+from sktime.utils._testing.panel import make_transformer_problem
+from sktime.tests._config import ESTIMATOR_TEST_PARAMS
+
+PAIRWISE_TRANSFORMERS_TAB = all_estimators(
```
Lovkush-A, line 10: Minor comment:
I notice that `_TAB` has been included here, which is contrary to the conventions used in _base

sktime/dists_kernels/tests/test_all_dist_kernels.py
```diff
@@ -0,0 +1,135 @@
...
+        X1_list_df, X2_list_df, pairwise_transformers_pan, test_params
+    )
+
+    # Test numpy of dataframes
+    _general_pairwise_transformer_tests(
+        X1_num_df, X2_num_df, pairwise_transformers_pan, test_params
+    )
+
+
+def _general_pairwise_transformer_tests(x, y, pairwise_transformers_tab, kwargs=None):
```
Lovkush-A, line 99:
```suggestion
def _general_pairwise_transformer_tests(x, y, pairwise_transformers, kwargs=None):
```
Looks like this general set of tests is for both tabular and panel transformers

sktime/dists_kernels/tests/test_all_dist_kernels.py
```diff
@@ -0,0 +1,135 @@
...
+    if kwargs is not None:
+        transformer = pairwise_transformers_tab(**kwargs)
+    else:
+        transformer: BasePairwiseTransformer = pairwise_transformers_tab()
```
Lovkush-A, line 103: Minor question:
Does adding the type hint change how the test runs, or is it more for readability?

sktime/dists_kernels/tests/test_all_dist_kernels.py
```diff
@@ -0,0 +1,135 @@
...
+
+    assert transformer.symmetric is False, (
+        f"Symmetric is set to wrong value for {pairwise_transformers_tab} "
+        f"when both X1 and X2 passed"
+    )
+    assert isinstance(
+        transformation, np.ndarray
+    ), f"Shape of matrix returned is wrong for {pairwise_transformers_tab}"
+    assert transformation.shape == (
+        4,
```
Lovkush-A, line 116: minor comment:
I think the numbers 4 and 5 should not be hard-coded, but instead be set via global variables.

sktime/dists_kernels/tests/test_compose_tab_to_panel.py
```diff
@@ -0,0 +1,81 @@
...
+)
+X2_list_df = make_transformer_problem(
+    n_instances=5, n_columns=5, n_timepoints=5, random_state=2, return_numpy=False
+)
+
+X1_num_df = np.array(X1_list_df)
+X2_num_df = np.array(X2_list_df)
+
+
+def test_aggr_():
```
Lovkush-A, line 44:
```suggestion
def test_aggr():
```
Unless there was a reason for the underscore...

sktime/registry/_base_classes.py
```diff
@@ -77,6 +81,16 @@
     ("regressor", BaseRegressor, "time series regressor"),
     ("forecaster", BaseForecaster, "forecaster"),
     ("transformer", BaseTransformer, "time series transformer"),
+    (
+        "transformer-pairwise-tabular",
```
Lovkush-A, line 85: Minor comment:
Inconsistency in tag and name of Base Class. Tag has tabular at the end, but BaseClass does not

Lovkush-A (commented):
Mostly minor comments.

* Only 'major' issue is the symmetry flag
* I like how tests have been written! I'll be sure to refer back to these tests when I need to write my own in the future.

sktime/dists_kernels/_base.py
chrisholder: So this is more my style of programming than anything so happy to get feedback on what people think about it. The reason it's defined within the scope is because I feel as a stand alone function it's not very useful and only when combined with the rest of the checks going on here does it have value. The reason it is a function then rather than just lines of code is because it is called when certain conditions and the conditions can't ever occur together and so rather than duplicate code I made it a function.

sktime/dists_kernels/_base.py
chrisholder: Symmetric is set to false by default in the constructor so don't think it is needed here as well unless you think it should be more explicit.

sktime/dists_kernels/compose_tab_to_panel.py
chrisholder: I agree this is a left over from previous versions where aggfunc could have been None but now a check is done higher up to make sure it can't be None, so this should be else.

sktime/dists_kernels/tests/test_all_dist_kernels.py
chrisholder: Sorry thats just me being inconsistent I normally type a lot of my Python but try to avoid typing things in tests since you can get some bad interaction between it and Pytest and so this shouldn't be typed.

sktime/dists_kernels/tests/test_all_dist_kernels.py
chrisholder: I don't know what the best practise is for this kind of thing so Ill agree with what anyone else thinks about this :joy:

sktime/dists_kernels/tests/test_compose_tab_to_panel.py
chrisholder: :joy: good spot

TonyBagnall (commented):
I'm not keen on the name dist_kernels, we should just pick one. Maybe a vote?

fkiraly:
>I'm not keen on the name dist_kernels, we should just pick one. Maybe a vote?

I don't mind the name, but how would it be clear that both distances and kernels go in there? People wouldn't expect either if we talk about "pairwise transformers".

fkiraly:
But if we call the folder "avote" as per your suggestion, @TonyBagnall, that would also be non-obvious?

TonyBagnall:
> > I'm not keen on the name dist_kernels, we should just pick one. Maybe a vote?
> 
> I don't mind the name, but how would it be clear that both distances and kernels go in there? People wouldn't expect either if we talk about "pairwise transformers".
I agree pairwise transformers is not a good name. I always thought distances and kernels were synonymous, and the differences were based on the origin of the research, but I am probably wrong and would appreciate enlightenment. If there is a significant difference I would have them in different packages. Otherwise I would prefer we pick one and stick with it. dist_kernels is suggesting to things, which is poor design imo. I also still dont like the abbreviated naming conventions used here which convey no meaning.

TonyBagnall:
> But if we call the folder "avote" as per your suggestion, @TonyBagnall, that would also be non-obvious?

that would be for the ensemble module ....

fkiraly:
how about `dists_n_kernels` or `dists_n_krnls`? Serious suggestions (no, really!)

sktime/dists_kernels/_base.py
fkiraly: I think it's fine to define a function inside a function if you use it multiple times there, but nowhere else.

sktime/dists_kernels/_base.py
```diff
@@ -0,0 +1,251 @@
...
+            default X2 = X
+
+        Returns
+        -------
+        distmat: np.array of shape [n, m]
+            (i,j)-th entry contains distance/kernel between X.iloc[i] and X2.iloc[j]
+
+        Writes to self
+        --------------
+        symmetric: bool = True if X2 was not passed, False if X2 was passed
```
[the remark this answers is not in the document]
fkiraly: yes, I think that's a better convention.

Because `symmetric` should actually be a tag that encodes the property d(x,y) = d(y,x).
Then, if both X=X2 and d is symmetric, the algorithm for filling with i<j should be used.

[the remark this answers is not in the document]
fkiraly: I've changed this now to the "correct" algorithm, including a renaming of the variable `X_equals_X2`.
That required introducing a tag - hope no one minds...

sktime/dists_kernels/_base.py
fkiraly: resolved above

sktime/dists_kernels/_base.py
fkiraly: resolved above

sktime/dists_kernels/_base.py
fkiraly: Yes, but added it nonetheleess for clarity

sktime/dists_kernels/compose_tab_to_panel.py
```diff
@@ -17,21 +17,30 @@ class AggrDist(BasePairwiseTransformerPanel):
...
     ----------
-    transformer: pairwise transformer of BaseTrafoPw scitype (tabular pairwise)
+    transformer: pairwise transformer of BasePairwiseTransformer scitype
 
     Hyper-parameters
     ----------------
-
     aggfunc: aggregation function 2D np.array -> float
         default = np.mean
+    aggfunc_symm: bool - whether aggregation function is symmetric
```
Lovkush-A, line 26: * What does it mean for aggfunc to be symmetric?
* Maybe this was meant to be `transformer_symm`, because it is the transformer that creates the matrix, not the aggfunc. (I think)
* Also, as general good practice, name of boolean variable should make it clear it is a boolean variable. something like `is_aggfunc_symm` or `aggfunc_is_symm`

sktime/dists_kernels/compose_tab_to_panel.py
```diff
@@ -58,18 +67,21 @@ def _transform(self, X, X2=None):
...
+        X_equals_X2 = self.X_equals_X2
 
         aggfunc = self.aggfunc
-        if aggfunc is None:
-            aggfunc = np.mean
+        aggfunc_symm = self.aggfunc_symm
+        transformer_symm = self.transformer._all_tags()["symmetric"]
+
+        # whether we know that resulting matrix must be symmetric
+        all_symm = aggfunc_symm and transformer_symm and X_equals_X2
```
Lovkush-A, line 77: Minor:
The logic here is non-trivial (but I think it is correct). I don't know how, but might be worth documenting this logic somewhere.

sktime/dists_kernels/_base.py
```diff
@@ -0,0 +1,261 @@
...
+        """
+
+        X = check_series(X)
+
+        if X2 is None:
+            X2 = X
+            self.X_equals_X2 = True
+        else:
+            X2 = check_series(X2)
+            self.X_equals_X2 = False
```
Lovkush-A, line 87: I think X can be equal to X2 if X2 is passed, e.g. user may do `.transform(X,X)`

I recommend changing to something like `self.X_equals_X2 = (X == X2)`. (I don't think this is correct syntax)

sktime/dists_kernels/_base.py
```diff
@@ -0,0 +1,261 @@
...
+        """
+
+        X = _pairwise_panel_x_check(X)
+
+        if X2 is None:
+            X2 = X
+            self.X_equals_X2 = True
+        else:
+            X2 = _pairwise_panel_x_check(X2)
+            self.X_equals_X2 = False
```
Lovkush-A, line 235: same as above, i think X can be equal to X2 if X2 is passed.
if you make the change, then update the docstring too

TonyBagnall:
> how about `dists_n_kernels` or `dists_n_krnls`? Serious suggestions (no, really!)

I would vote to keep it called distances, but would accept kernels. distances_and_kernels if you can provide a clear distinction between the two, although then I would separate them into distances and kernels

sktime/dists_kernels/scipy_dist.py
```diff
@@ -0,0 +1,88 @@
...
+            if 'intersect', distance is computed on columns occurring both in X and X2,
+                other columns are discarded; column ordering in X2 is copied from X
+            if 'force-align', raises an error if the set of columns in X, X2 differs;
+                column ordering in X2 is copied from X
+            if 'none', X and X2 are passed through unmodified (no columns are aligned)
+                note: this will potentially align "non-matching" columns
+    """
+
+    _tags = {
+        "symmetric": True,  # all the distances are symmetric
```
Lovkush-A, line 43: Major:
I am not sure about this. There is high chance you are getting mixed up between symmetry of elementwise transformer and symmetry of pairwise transformer. the distance function is symmetric, but that is different from pairwise transformer being symmetric

* in base class, you say symmetric tag means: `# is the transformer symmetric, i.e., t(x,y)=t(y,x) always`
* However, trivially, all pairwise transformer are not symmetric. Just let len(x) != len(y), and the resulting dist_matrix have different shapes
* maybe symmetric condition should be something like `t(x,y) = transpose(t(y,x))`? (and that definition of aggfunc being symmetric is that aggfunc(M) = aggfunc(transpose(M)) )

I think the only way to properly resolve this is to get out some pen and paper, and actually write out explicitly some definitions and some proofs of relevant claims (main claim being that if various boolean tags/variables are true, then the final panel pairwise transformer is symmetirc)

Lovkush-A (changes requested):

Lovkush-A:
> > how about `dists_n_kernels` or `dists_n_krnls`? Serious suggestions (no, really!)
> 
> distances_and_kernels if you can provide a clear distinction between the two

I think we should create a "swear jar" for franz, in which every time he makes an abbreviation that somebody else finds unclear, he has to put some money into it. We'll be able to fund new contributors in no time! 😆

Lovkush-A:
Suggestion, but probably better for future PR, as this pr is already getting dragged out (admittedly by me, but for good reason I hope):

Current design: in `compose_tab_to_panel` composers that create panel pairwise transformers from table pairwise transformers. (Minor note, is this a composition or a reduction?)

Alternative design: split this into two steps:
* a composer/reducer that creates panel *elementwise* transformers from table *pairwise* transformers (i.e. just stick an aggfunc on the end of the pairwise transformer)
* another function/class that creates panel pairwise transformers from panel elementwise transformers

If you agree with this alternative design, then best to create a new issue, rather than doing it in this pr.

fkiraly:
>Alternative design: split this into two steps:

@Lovkush-A, I don't understand what you are suggesting here.
Can you give an example, maybe with mathematical formulae, what the three objects would be?

Lovkush-A:
> Can you give an example, maybe with mathematical formulae, what the three objects would be?

Maybe the middle step should be elementwise *tabular* transformer, not elementwise panel transformer... lets see.
-----
Pairwise tabular transformer, f say. Takes as input two dataframes X1 and Y1 and outputs a matrix of size (len(X1), len(Y1)). E.g. f(X1,Y1) = ScipyDist(X1,Y1)

Gets converted to elementwise transformer, g say. Takes as input two dataframes X1 and Y1 and outputs a single number. g(X1,Y1) would equal aggfunc(f(X1,Y1))

Gets converted to pairwise panel transformer, h say. Takes as input two sequences of dataframes X=(X1,...,Xm) and Y=(Y1,...,Yn) and outputs a matrix of size (m,n). h(X,Y)_ij would equal g(Xi, Yj).
-----
Again, want to emphasise I am not not sure if this is a good idea, and if it is, would be better to create a new issue/PR for it.

fkiraly:
@chrisholder, do you mind giving this another review, after I´ve addressed some of the issues around `symmetric`?

fkiraly:
>Gets converted to elementwise transformer, g say. Takes as input two dataframes X1 and Y1 and outputs a single number. g(X1,Y1) would equal aggfunc(f(X1,Y1)).

@Lovkush-A, that's not a good idea, since your "elementwise panel transformer" encapsulates the same scientific object as the "pairwise panel transformer" without any semantic or computational benefit.

The reason is similar to the old story why you want distances and kernels to be functions `(n x r) matrix x (m x r) matrix -> (m x n) matrix` rather than `float x float -> float`. The main reason here is that inputting samples allows vectorization techniques to be applied. The "pairwise panel transformer" is just the vectorized version of the "elementwise panel transformer" (interface-wise), hence the latter is superfluous.

sktime/dists_kernels/scipy_dist.py
fkiraly: @Lovkush-A, ok, I think I see the point of confusion here.

All transformers in the `dists_kernels` module are meant to be element-wise, but vectorized.

I.e., they all model functions `d: R^d x R^d -> R` (in the tabular case) or functions `d: series x series -> R` (in the panel case.

The machine type implements a vectorized version of this, i.e., `d: R^(nxd) x R^(mxd) -> R^(mxn)` resp. `d: series^n x series^m -> R^(mxn)`, this is precisely also the convention from `scipy`, or from `scikit-learn` (kernels for GP).

fkiraly:
>I think we should create a "swear jar" for franz, in which every time he makes an abbreviation that somebody else finds unclear, he has to put some money into it. We'll be able to fund new contributors in no time!

Only if we also create a "variable rename jar" for you, @Lovkush-A 😉

chrisholder:
Hi @fkiraly been bit busy but im going to take a look at this now ive already spotted some minor things but ill change them and push to the branch directly

fkiraly:
>Hi @fkiraly been bit busy but im going to take a look at this now ive already spotted some minor things but ill change them and push to the branch directly

fine with me, @chrisholder

chrisholder:
@fkiraly I've resolved all the conversations I thought I could, could you take a look at the final 4?

chrisholder (approved):

sktime/dists_kernels/scipy_dist.py
fkiraly: this is a "won´t fix" since transformers are always vectorized.

sktime/dists_kernels/compose_tab_to_panel.py
fkiraly: `aggfunc_symm` encodes knowledge about `aggfunc` which is in general not queriable from `aggfunc`, so I think it's sensible to pass it here?

I've changed the name of it to `aggfunc_is_symm`.

sktime/dists_kernels/compose_tab_to_panel.py
fkiraly: I added a comment for this

sktime/dists_kernels/compose_tab_to_panel.py
fkiraly: I updated the docstring

chrisholder:
Ill give this another review and then hopefully this will be ready to go in

sktime/dists_kernels/compose_tab_to_panel.py
Lovkush-A: @fkiraly Just to confirm, the resolution to this conversation is that this is not an issue and does not require change? (which, to be clear, I am ok with. but just checking that was your intention)

sktime/dists_kernels/tests/test_all_dist_kernels.py
Lovkush-A: @chrisholder I see you have updated tests to remove the _TAB. I was actually hoping the conventions in the tests would be adopted in base! 😅  But being consistent is better in my mind than having different conventions.

sktime/dists_kernels/tests/test_all_dist_kernels.py
Lovkush-A: @chrisholder. I notice you didn't change name of variable here. I am fine either way, but just want to check what the reason is that you prefer the name with _tab than without tab here?

sktime/dists_kernels/compose_tab_to_panel.py
Lovkush-A: @fkiraly

> `aggfunc_symm` encodes knowledge about `aggfunc` which is in general not queriable from `aggfunc`, so I think it's sensible to pass it here?

Sure, but that was not my question. My question is what does it even mean for an aggfunc to be symmetric?

Based on later discussions, I think it means that aggfunc(M) = aggfunc(transpose(M)). Whatever it is, I think it should be clearly stated in docstring because it is non-obvious. (e.g. symmetric might mean that the function is invariant under any permutation of entries of the matrix while preserving the shape of the matrix)

Lovkush-A (dismissed):
@chrisholder @fkiraly. Thanks for all the updates and responding to my many comments.

I am happy for this to be merged, but I think there are some issues that are to be resolved in the future, and I want to note them down before they are forgotten:

* Main one is documentation. There are various subtleties which ought to be explicitly explained somewhere. Unfortunately, I do not know best place to do this. Here are things that I think should be explained:
  * Exactly what kind of thing PairwiseTransformer is, and having consistent ways of talking about it.
    * E.g. in docstring we say transformer is symmetric if`t(x,y)=t(y,x)` but `t`, `y` and `x` are not exactly the same as the PairwiseTransformer object and the possible inputs to a pairwise transformer.
    * A good chunk of my comments/questions above were based on a misunderstanding of what PairwiseTransformers are. I thought the "Pairwise" part (i.e. that it takes two *lists* of inputs and it applies a binary operation *pairwise*) was the crucial part of this new base class. However,  my current understanding is that the crucial new things is that it is a *binary* operation rather than a unary one, and the *pairwise* just corresponds to how the vectorization is done.
    * I think (but am not sure) that the fundamental thing we want is something like a `BinaryTransformer`, and the PairwiseTransformer is a subclass with the pairwise form of vectorization. I think this because it is feasible there are other transformations on two lists of series that are not pairwise.  E.g. in unary transformers, most transformers are elementwise (e.g. StandardScaler applies same function to each element), but some are not elementwise (e.g. calculating series of first differences). Again, not I am not sure about this.
  * Why the `all_sym` condition is a sufficient condition. (Note that the comment explaining that it is a sufficient condition is useful, because it makes it clear that maybe it is symmetric even if all_sym is False)
  * Document why aggfunc default is done in `transform` rather than `init`. (Note, I currently think it ought to be done in init, but likely because I do not understand the downside of doing so)
  * ~~Document what `aggfunc` symmetric means.~~ EDIT: been added to docstring.
* in _base.py, might be worth factoring out the checking that X equals X2 into a helper function?
* worth thinking about whether we want to have "Tabular" at the end of the tabular transformers. Originally Chris added `tab` in names of variables in the tests and then removed it to be consistent with the names of the classes. I personally think the tests were more readable with the `tab` included, than without them included.
* There were couple of conversations marked above as resolved without any changes. I commented there just to check that was intentional

Finally, might be worth thinking about/encouraging a norm of having smaller PRs in the future. This PR was pretty big with many parallel discussions going on that are easy to lose track of. E.g. in this instance, the task of dealing with symmetry could have been done in a separate PR.

sktime/dists_kernels/compose_tab_to_panel.py
fkiraly: it says in the docstring, default='euclidean', so I thought it was not an issue.

sktime/dists_kernels/tests/test_all_dist_kernels.py
fkiraly: hm, I don't think it would be a big issue either way. We just need to be consistent, which it now is.

sktime/dists_kernels/tests/test_all_dist_kernels.py
fkiraly: I guess consistency. Shorter is better if self-explanatory.

sktime/dists_kernels/compose_tab_to_panel.py
fkiraly: >Sure, but that was not my question. My question is what does it even mean for an aggfunc to be symmetric?

Gives the same result if you transpose the matrix, indeed.

fkiraly: I added a comment

sktime/dists_kernels/compose_tab_to_panel.py
Lovkush-A: Just for clarity, my (minor) concern is that this default setting is only explicit in the docstring for ScipyDist, not in the docstring for AggrDist. My preference would be something like `example: AggrDist(ScipyDist(metric = 'euclidean')) is mean Euclidean...`

Lovkush-A (approved):

sktime/dists_kernels/tests/test_all_dist_kernels.py
chrisholder: Funny enough I thought I did change the name here hence I marked it as resolved :joy: - Just checked I forgot this file had two tests in it, I changed it for the first one and not the second one. Ill change the seocnd one.

chrisholder:
So I resolved the rename comments but don't really want to push because then all the tests will have to rerun. Shall I just hold out so this can be merged?

fkiraly:
>So I resolved the rename comments but don't really want to push because then all the tests will have to rerun. Shall I just hold out so this can be merged?

Which rename comments? I thought I had resolved them already.

Issue #1032, opened 2021-06-23
Refactor CI/CD testing


mloning:
The automated unit tests would benefit from:
* Cleaning and restructuring 
* Migrating appveyor Windows builds to GitHub Actions
* Separating standard tests that are run on every PR and accuracy tests that are run periodically (cron jobs)
* Combine basic API tests on fitted objects to avoid having to call `fit` multiple times
* Making them importable from other packages (see https://github.com/pytest-dev/pytest/issues/421)
* Making them run on objects rather than classes (see #1027)

TonyBagnall:
I would like a mapping overview of all the testing that goes on in sktime, then work towards a simple guide to tests, where to put them, what to test for each module etc etc. Improving classifier testing (removing excluded etc) is a mid term goal for me, but its more likely to happen sooner if its easier and intuitive to do

fkiraly:
Agreed - I've put a simple testing guide in the extension guidelines, at the end of the forecaster tutorial

fkiraly:
And appveyor and azure build take ages (2-3 hours??), speeding up is indeed a great suggestion.

Lovkush-A:
I recently saw this video giving whirlwind tour of unittesting using UnitTests. 

[Ned Batchelder: Getting Started Testing - PyCon 2014 - YouTube](https://www.youtube.com/watch?v=FxSsnHeWQBY) 

My summary of main ideas. Might be useful:

* First chunk (~20 mins) is basics of unittests.
* Goes a tiny bit under-the-hood of UnitTests
* Big tip/opinion: treat unittesting the same way as you do rest of software engineering. should refactor it, should write helper functions, if appropriate should have tests for tests
* Idea of mock objects, to isolate what a test is testing, and to remove slow parts of tests (if appropriate). but big downside is that mock objects have to be manualyl updated if there are any interface changes
* use coverage tools to check which lines have/have not been tested
* things to read up on in your own time: dependency injection, ddt (data driven tests), jenkins and travis for continuous testing (I think this is for software in deployment?), integration tests, load testing

Lovkush-A:
Been skimming through some of Ted Kaminski's essays on software engineering, and in this one https://www.tedinski.com/2018/05/01/designing-imperative-code-with-properties-in-mind.html they argue for 'property testing' and recommend the package [Hypotheis](https://hypothesis.readthedocs.io/en/latest/).

[...]

Issue #1027, opened 2021-06-23
test_estimator or a variant should work with object (not classes)


fkiraly:
For easy testing and debugging, it would be great if `test_estimator` or a variant thereof could be extended to work with object instances, not classes; alternatively, classes with default parameter settings not read from a file, but given as an argument.

This is for:
* making it easier for local (non-contributing) extenders to write their own tests
* making new estimators easier to debug, e.g., by making it easier to vary the parameters in the test
* to enable extension and manual testing without having to clone the repository (with a simple workflow that could be added to the tutorials)

fkiraly:
@mloning, @aiwalter, @TonyBagnall - thoughts?

mloning:
Closed in favour of #1032

fkiraly:
I think this is separate - this is a minor change.

Refactoring the entirety of tests is a big change.

We don't need to do a massive refactoring to allow users to run the *current* `test_estimator`.

PR #1072, opened 2021-06-28
estimator tests to allow objects, not just classes

fkiraly:
This PR addresses #1027 and part of #1032.
The post-PR state allows testing of concrete objects, not just classes, using the current test suite.

This is achieved by modifying the `estimator_checks._construct_instance` utility:

* pre-PR, this always assumes the argument `Estimator` is a class, and creates an object after lookup of default parameters from the default parameter registry, using those parameters. This object is then returned.
* post-PR, it first checks whether the argument is a class or an object. If an object, it just returns the argument; if a class, it proceeds as pre-PR.

Since `_construct_instance` is called at the start of every estimator related test, this change enables all those tests to take objects as well as classes.

For the user, this allows convenient command line testing of newly written estimators without first having to deposit default parameters in the default parameter registry, or changing the registry file to test with different parameter settings.


mloning:
Would you want to run your estimator object through `check_estimator` function? I think that would still break as at least one of the other functions assumes the input to be a class (see `check_inheritance`), not sure if there are more.

fkiraly:
>Would you want to run your estimator object through check_estimator function?

Good idea - I think I have now tracked down all functions that were class (not object) related that needed to turn objects back into classes.

fkiraly:
this is ready for review again

mloning (commented):
In general, I do not think that adding these conversions in every test isn't the right way of changing the tests. I'd prefer separating tests that run on classes and tests that run on objects. 

But wouldn't veto the PR if you think this needs to go in now.

mloning:
There was a typo: I do not think that adding the object/class conversions at the beginning is the right way of handling this, but don't have a strong preference.

mloning:
The better way may be to replace the name of the required hyper-parameters in the `_required_parameters` class attribute with the actual hyper-parameters used in testing (as you suggested somewhere else).

mloning:
Can this be closed?

mloning:
Closed due to inactivity.

PR #1077, opened 2021-06-29
Improvement to Multivariate-Detrending

SveaMeyer13:
#### Reference Issues/PRs
Follow up on #1042 

Make a deep copy of input DataFrame in order to no change input of transform/inverse-transform 

No.


TonyBagnall (approved):
adds a deep copy, seems fine to me

Issue #1073, opened 2021-06-28
unit tests for estimator method side effects


fkiraly:
Currently, it seems we have no tests that check estimator methods for side effects on the arguments?
(see also #1042 - before I thought we had)

This should be relatively easy to add - checking whether passing arguments to `fit` etc changes the argument.
The only complication is that some methods have to be called before others, e.g., `fit` before `predict`.

fkiraly:
Being worked on here #1078

mloning:
It's interesting that scikit-learn does not test for this but instead exposes this through `copy` hyper-parameters in estimators. This may be a sensible option, especially for larger data sets and more complex models in which now all components would store a copy of the data if I understand this correctly. 

#1078 still seems like the next logical step at this point. If we want to allow for side effects for better memory efficiency, we could then modify the tests that checks if a `copy=False` hyper-parameter has been set.

fkiraly:
>I'm not sure though if eventually we want to allow for side effects again on the input data

If at all, we really need to think this through very carefully...

PR #1078, opened 2021-06-29
unit test for absence of side effects in estimator methods

fkiraly:
This implements #1073, a test for estimator methods not having side effects.

New method is `check_methods_have_no_side_effects` in `estimator_checks`.


sktime/utils/_testing/estimator_checks.py
```diff
@@ -430,6 +476,38 @@ def check_methods_do_not_change_state(Estimator):
...
+    estimator = _construct_instance(Estimator)
+
+    set_random_state(estimator)
+
+    # Fit for the first time
+    fit_args = _make_args(estimator, "fit")
+    old_fit_args = deepcopy(fit_args)
+    estimator.fit(*fit_args)
+
+    assert dict_equals(
```
mloning, line 494: `args` are tuples here, does the built-in tuple comparison not work, e.g. `a == b`?

fkiraly: yes, just noticed that, see commit below

fkiraly: no, comparison does not work as soon as you have types inside that use custom equality checks which are not defaulted to by `==`

fkiraly: for instance `np.array` does this element-wise, and `pandas` types return errors so you have to use `equals`

fkiraly:
looks like we caught four estimators that have side effects! This is what fishing must feel like, watching the tests complete one after another and fails popping up only on rare occasion.

It's `HampelFilter`, `Tabularizer`, `TimeSeriesKMeans`, and `TimeSeriesKMedoids`... @aiwalter, @chrisholder?

fkiraly:
Looks like we have one estimator that causes side effects, the `HampelFilter` - @aiwalter?

fkiraly:
So, @mloning ... the fact that the tests fail seems to have nothing to do with that the PR is faulty, but in fact means that it works? What do we do?

mloning:
@fkiraly it's only the `HampelFilter` yes? So let's fix that first before merging this one.

fkiraly:
yes, @aiwalter is working on it in #1078.

Philosophical question: does it make sense to test the tests? I.e., having a negative and positive case where you know the test should pass or fail? E.g., if the test has complex logic that is specific to the test, such as testing for equality and computing equality is non-trivial. Or does that mean you want to test the equality utility instead?

sktime/utils/_testing/estimator_checks.py
```diff
@@ -97,6 +98,96 @@ def yield_estimator_checks(exclude=None):
...
+                return False
+            if not np.array_equal(x, y, equal_nan=True):
+                return False
+        elif xi != yi:
+            return False
+
+    return True
+
+
+def dict_equals(x, y):
```
mloning, line 146: I think this can go

sktime/utils/_testing/estimator_checks.py
```diff
@@ -97,6 +98,96 @@ def yield_estimator_checks(exclude=None):
         yield check
 
 
+def tuple_equals(x, y):
```
mloning, line 101: Regarding tests for tests, we may want one for this function

sktime/utils/_testing/estimator_checks.py
fkiraly: well, currently not used, but we may need it for kwargs at some point, so I wouldn't want to keep it in a text file until then

mloning: Let's not add functionality to the main source code which isn't used, I have a separate folder for code that's related to sktime but currently not used, or perhaps add it to `contrib`?

sktime/utils/_testing/estimator_checks.py
fkiraly: which folder do tests for tests go in?

mloning: I would move the function to `sktime/utils/_testing` and then add tests in `sktime/utils/_testing/tests`

sktime/utils/_testing/estimator_checks.py
fkiraly: We might like to use it for new tests in the near future, e.g., kwargs related tests (same as current but use kwargs) or tag related tests (e.g., does it have side effects on `_tags`?) so I'd prefer to leave it in. It doesn't hurt anyone there, and it wouldn't be the precedent (see e.g., `_get_name` in the same folder). If it's really a blocker for you, I'm happy to delete it and add it back later, but it will be more difficult to find again.

fkiraly: where is your separate folder if I were to put it there?

mloning: Yes please remove it, if `_get_name` isn't used we can remove it too. My folder is just local, for smaller things I posted them on gists or anywhere else really but keeping them in the main code base isn't a good idea.

SveaMeyer13:
Shouldn't your tests also throw an error for the multivariate detrender? (this currently is still introducing side effects)

fkiraly:
not sure, @SveaMeyer13 - is the detrender already in `main`? If yes, it probably should.

fkiraly:
@mloning, please make changes as you see fit and merge when you see fit.

sktime/utils/_testing/estimator_checks.py
fkiraly: I'm confused, @mloning - you tell me to move things to a separate folder, but that folder does not exist or is local on your computer (?).

So, would you mind making the changes yourself? It sounds like it's just a small thing and it's faster if you do it. I'm happy with a complete delete too.

SveaMeyer13:
> not sure, @SveaMeyer13 - is the detrender already in `main`? If yes, it probably should.

Until half an hour ago it was in `main` with side effects that were not caught. And now that #1077 is merged there are no longer any side effects.

fkiraly:
well, @SveaMeyer13, I've merged the pre-#1077 merge state into this branch to test whether the test catches the side effects - we'll see... once the checks are finished sometime early 2022, I suppose.

fkiraly:
@SveaMeyer13, it doesn't seem to capture it - not sure why, it should.
Do you have a good idea?

sktime/utils/_testing/estimator_checks.py
```diff
@@ -430,6 +521,36 @@ def check_methods_do_not_change_state(Estimator):
...
+        Estimator = type(Estimator)
+
+    estimator = _construct_instance(Estimator)
+
+    set_random_state(estimator)
+
+    # Fit for the first time
+    fit_args = _make_args(estimator, "fit")
+    old_fit_args = deepcopy(fit_args)
+    estimator.fit(*fit_args)
```
SveaMeyer13, line 537: @fkiraly, I think the reason why the detrender was not caught is here:  `_make_args(estimator, "fit")` creates a `pd.Series`. But as side effects are only introduced in the multivariate case, they won't be caught.

fkiraly: ah! Well observed, I didn't notice this.

So, is there a potential action on adding tests for multivariate arguments to `fit`?
Perhaps that makes sense only once we've settled on a "can handle multivariate" tag, because only if the estimator claims it can an error should be thrown.

SveaMeyer13: In theory, you could already add this to the tests, as we have the `univariate-only tag` -  At least I would assume that each transformer should be able to handle multivariate data, unless it carries the `univariate-only` tag.

fkiraly: ahhhhh but that is not the case, that would be too easy, @SveaMeyer13!
This is exactly what I thought, but according to one of @mloning and @aiwalter (I forgot which) the tag is actually meant to signify whether the exogeneous variables `X` are used or ignored - not whether `y` can be a data frame.

sktime/utils/_testing/estimator_checks.py
mloning: Yes just delete it! Sorry for the confusion.

fkiraly:
thanks @thayeylolu for looking into this!
As discussed, you would help a lot by creating the test for `tuple_equal` that @mloning requests above.
Please make a PR into this branch.

fkiraly:
@chrisholder, for some reason this would break with the new distances, since making arguments for distances' `transform` was not implemented. I quickly cooked up some arguments, but shouldn't this have already been dealt with in your tests, @chrisholder? I'm a bit surprised that (i) it wasn't, and that (ii) it didn't create any problems in the distances PR itself. Any good explanations?

sktime/utils/_testing/estimator_checks.py
fkiraly: I moved the utility to the indicated folder, but *didn't* delete the function. Instead, I refactored it into a single `deep_equals` which also allows to recurse over lists/tuples/dicts of lists/tuples/dicts.

Why: in the panel setting, in the distance module, and in tuning (parameter sets), we have cases where arguments are lists or dicts, so for the comparison of the entire argument set we get a tuple of lists or tuple of dict (of numpy etc).

sktime/utils/_testing/estimator_checks.py
fkiraly: resolved below since superseded

thayeylolu:
> thanks @thayeylolu for looking into this!
> As discussed, you would help a lot by creating the test for `tuple_equal` that @mloning requests above.
> Please make a PR into this branch.

Hi @fkiraly. `Tuple Equals` does not exists in the code base.It has been updated.  Do you still want me to work on it?

fkiraly:
@thayeylolu, thanks, but I think I´m fine with thie PR.
Can you please look at #1083 instead and ensure the checks pass (or prepare that they pass after #980 is merged in case it has to do with the data frame return type).

mloning (approved):
Haven't looked at the code in much depth (still not a fan of the `deep_equals` function) but unless someone else wants to take a closer look, I think we can merge this. 

I'm not sure though if eventually we want to allow for side effects again on the input data (see my comment in #1073)

fkiraly:
>still not a fan of the `deep_equals` function

How exactly would you then check for equality of input tuple/dict across heterogeneous input signatures?
The only alternative I see is write about the same amount of code, less general and less re-usable, and more strongly coupled to the individual test.

PR #1084, opened 2021-06-30
Dictionary based improvements

MatthewMiddlehurst:
Updates the parameter selection method used by TDE, with this and the SFA changes TDE is no longer significantly worse than Java alternative.

Removed BOP from contrib, originally put there by me. Its broken and im not going to put any time into fixing it. The algorithm is rather dated anyway.

A number of SFA transform changes:
- DFT alternative to FFT which no longer lowers the word length when window_size < word_length
- Altered lower bounding stuff in dft/mft to be consistent with the Java version. Input would be good for this.
- More numba
- Included option for a numba Dict as output. For some reason this does not work with multivariate TDE and produces a bunch of KeyErrors so im not using it for now. Dont see much reason to remove it as on option though, it worked fine for the univariate tests.
- No longer a cap on word_length and alphabet_size. Words/bigrams which use <= 64 bits will still use numpy/numba functions.

Open to suggestions, there is quite a bit going on in the SFA class.

Docs still need updating a bit, will open up the PR for now though.


sktime/transformations/panel/dictionary_based/_sfa.py
```diff
@@ -518,8 +617,8 @@ def _mft(self, series, transformed=None, stds=None):
             self.inverse_sqrt_win_size,
         )
 
-        # lower bounding
-        transformed[:, 1::2] = transformed[:, 1::2] * -1
+        if not self.lower_bounding:
+            transformed[:, 1::2] = transformed[:, 1::2] * -1  # lower bounding
```
MatthewMiddlehurst, line 621: So in our Java version of BOSS/TDE we dont do this, but for the WEASEL version we do. Like this lower_bounding == False mimics WEASEL, while lower_bounding == True is BOSS/TDE im pretty sure.

Probably need more input on what would be best or how much an impact this has.
(Also applies to DFT)

MatthewMiddlehurst:
Current state of Java v Python for accuracy. On ~90 UCR datasets with 30 resamples.

Capture

MatthewMiddlehurst:
Cant request a review, but input from @patrickzib would be helpful.

TonyBagnall (dismissed):
it looks good to me, but there is a lot in there, cant honestly say I have checked it all

sktime/transformations/panel/dictionary_based/_sfa.py
patrickzib: @MatthewMiddlehurst Oh, really cool! I will try to give the code a review. But it might take till the next week.

To answer this question:
SFA can be used as a dimensionality reduction technique for indexing and the distance on two SFA words lower bounds the Euclidean distance (there is an analogy for SAX).

The lower_bounding check comes from this background. There is also a division by the length of the subsequence somewhere in the code. 

When solely looking at the multiplication by -1, it doesn't make a difference for classification, I believe.

MatthewMiddlehurst: @patrickzib Thanks, would be great if you could take a look. No rush on this.

There is also the inverse_sqrt_win_size affected by the setting. Individually it didnt make a lot of difference to TDE at least, but there were a few minor things that stacked up. Let me know if you want anything changed, SFA is quite widely used so I dont want to step on anyones feet :)

There are a couple of things im not so happy with i.e. the numerous if/else for different data types, but different functions are required for uncapped word_length and alphabet size unfortunetly.

patrickzib: @MatthewMiddlehurst 

Hard to tell: Indexing is not the focus of sktime. Thus, it might be safe to comment these.

Have you tried to do a critical difference plot with and without these lines of code? That would be interesting.

sktime/classification/dictionary_based/_tde.py
```diff
@@ -269,33 +284,39 @@ def fit(self, X, y):
...
 
-            tde.accuracy = self._individual_train_acc(
+            if self.save_train_predictions:
+                tde._train_predictions = np.zeros(self.n_instances - subsample_size)
+
+            tde._accuracy = self._individual_train_acc(
                 tde,
                 y_subsample,
                 subsample_size,
                 -999999 if num_classifiers < self.max_ensemble_size else lowest_acc,
```
patrickzib, line 294: numpy.NINF instead of -99999?

sktime/transformations/panel/dictionary_based/_sfa.py
```diff
@@ -233,25 +268,44 @@ def transform(self, X, y=None, supplied_dft=None):
 
         return bags
 
-    def _transform_case(self, X, i, supplied_dft):
+    def _transform_case(self, X, supplied_dft=None):
         if supplied_dft is None:
```
patrickzib, line 272: I am not sure, if I understand correctly, but I suppose you either choose between MFT and DFT? Wouldn't it be better to let the code decide, if the window_size too small compared to the word_length to fall back to DFT?

sktime/transformations/panel/dictionary_based/_sfa.py
```diff
@@ -233,25 +268,44 @@ def transform(self, X, y=None, supplied_dft=None):
...
+            word_raw = (
+                SFA._create_word(
+                    dfts[window],
+                    self.word_length,
+                    self.alphabet_size,
+                    self.breakpoints,
+                    self.letter_bits,
+                )
+                if self.word_bits <= 64
+                else self._create_word_large(
```
patrickzib, line 306: Why does it have to be split between create_word and create_word_large? Is _create_word_large not possible with numba ?

patrickzib (commented):
imho it would be nice to automatically switch between MFT and DFT if the window_size and word_length require to do so. This way it would add functionality to existing codes.

sktime/transformations/panel/dictionary_based/_sfa.py
MatthewMiddlehurst: As far as I am aware max number of bits for numpy/numba ints is 64. The functions could be identical, but turning it into a numba function will convert the input/output to a 64 bit int (or crash if it fails to do so).

Makes it rather ugly, but we know how many bits we need beforehand and can use the fast functions for existing classifiers.

sktime/transformations/panel/dictionary_based/_sfa.py
MatthewMiddlehurst: supplied_dft is a multivariate TDE thing, I use the DFT from breakpoint creation to determine if a dimension is worth keeping or not and dont want to redo the processing. I kept it generally in transform because there may be some other use cases.

The DFT option is just for the initial DFT/FFT in the MFT and breakpoints, MFT is still used in the transform for both. I will edit it so the it switches between DFT/FFT depending on the word_length and window_size as suggested.

sktime/classification/dictionary_based/_tde.py
MatthewMiddlehurst: Yeah, that makes more sense :)

MatthewMiddlehurst: ah nvm it doesnt like `required_correct = int(lowest_acc * train_size)` with inf, im just going to go with 0 since its not as weird as whats there currently and works just as well.

sktime/transformations/panel/dictionary_based/_sfa.py
MatthewMiddlehurst: @patrickzib I did this, and the difference was minor for sure, I stopped TDE from doing it because it was very slightly worse and slightly slower. Any difference was definitely not significant from this alone.

Would be good to keep the functionality and sktime is more than just classification, I’m pretty sure the way it is set up now makes sense, but good to check.

TonyBagnall:
hi @patrickzib is there anything else you would recommend changing?

sktime/transformations/panel/dictionary_based/_sfa.py
patrickzib: I see :) 

just one minor suggestion: Maybe it would be more consistent with the sktime / sklearn to have a parameter in the constructor like `keep_dft=True` and store the dft within the the `fit` method if `keep_dft=True` and reuse it in the `transform` method?

sktime/transformations/panel/dictionary_based/_sfa.py
patrickzib: Ah, ok, so the method switches between a return type with 32 and 64 bits, I see :)

patrickzib:
@TonyBagnall @MatthewMiddlehurst  I am having trouble importing this branch into PyCharm with the following error:

`Cannot find reference 'NumbaTypeSafetyWarning' in '__init__.py | __init__.py'`

It might be a PyCharm thing or maybe due to outdated packages? Still, it is akward. Do you have any ideas?

sktime/transformations/panel/dictionary_based/_sfa.py
MatthewMiddlehurst: The 64 bits for the numba functions and unlimited for the pure python one, but essentially yeah.

MatthewMiddlehurst:
Thats annoying, may be the numba version you are using. There is no harm in removing them it just causes some print-outs when using a typed dict. numba defaults to 64 bit ints even if its going to be a massive waste of memory and then throws out warnings when its converted to a smaller int. 

Will try setting the types for the numba functions and removing them, which should be have been done either way tbh.

sktime/transformations/panel/dictionary_based/_sfa.py
patrickzib: One suggestion for readability and reducing the overhead:

Maybe replace it by:
```
bigram = (
                        self._create_bigram_words(
                        word_raw, words[window - self.window_size])
                    )
```

and introduce a wrapper function:

 ```
   def _create_bigram_words(self, word_raw, word):
        return SFA._create_bigram_word(
                    word,
                    word_raw,
                    self.word_bits,
                ) if self.max_bits <= 64 else self._create_bigram_word_large(
                    word,
                    word_raw,
                )
```

This way, it can also be used further down the code for skip-grams:

                     ```
   skip_gram = (
                            self._create_bigram_words(
                                word_raw, words[window - s * self.window_size])
                        )
```

sktime/transformations/panel/dictionary_based/_sfa.py
patrickzib: I am confused :) there is this switch alread:

```
        if self.save_binning_dft:
            self.binning_dft = dft
```

So, this is still not, what you are actually intending to do? :)

sktime/transformations/panel/dictionary_based/_sfa.py
```diff
@@ -163,7 +159,9 @@ def __init__(
         self.skip_grams = skip_grams
 
         self.return_pandas_data_series = return_pandas_data_series
-        self.fourier_transform = fourier_transform
+        self.fourier_transform = (
+            fourier_transform if word_length < window_size - offset else "dft"
```
patrickzib, line 163: I really like this change, thank you :)

Just one minor thing: currently `self.fourier_transform` is a string with a check: `self.fourier_transform == "dft"`

But as this is always a boolean condition with True, False, it would be better to use it as a boolean? 

Maybe something like:
`self.use_fallback_dft=True | False`

sktime/transformations/panel/dictionary_based/_sfa.py
MatthewMiddlehurst: Yeah, there is a parameter for saving the binning DFT in `fit`, but the `transform` part is kept general so it can be used with any DFT. 

The only usage of that is mine however. I dont mind changing it so that if you save the binning DFT, it is assumed you are going to use it in `transform` and removing the current `transform` parameter.

TonyBagnall:
I wasnt meaning to rush you @patrickzib, just doing some house keeping. This branch runs for me with numba 0.52.0

sktime/transformations/panel/dictionary_based/_sfa.py
```diff
@@ -459,51 +536,107 @@ def _discrete_fourier_transform(self, series):
...
+        start_offset = 2 if self.norm else 0
+        length = self.dft_length + start_offset + self.dft_length % 2
+        end = max(1, len(series) - self.window_size + 1)
+
+        phis = SFA._get_phis(self.window_size, length)
+        stds = SFA._calc_incremental_mean_std(series, end, self.window_size)
+        transformed = np.zeros((end, length))
+
+        # first run with dft
+        if self.fourier_transform == "dft":
```
patrickzib, line 622: Not sure, if this is correct?

The first run of the mft is always the fft. and only for the subsequent runs it should use the mft-trick.

Howerver, given this switch, it will either always use the fft or never?

TonyBagnall:
image

sktime/transformations/panel/dictionary_based/_sfa.py
patrickzib: A similar trick can be used in `_shorten_case` to increase readability.

patrickzib:
> I wasnt meaning to rush you @patrickzib, just doing some house keeping. This branch runs for me with numba 0.52.0

That is just fine, it was on my agenda for today, anyways :)

patrickzib:
> Thats annoying, may be the numba version you are using. There is no harm in removing them it just causes some print-outs when using a typed dict. numba defaults to 64 bit ints even if its going to be a massive waste of memory and then throws out warnings when its converted to a smaller int.
> 
> Will try setting the types for the numba functions and removing them, which should be have been done either way tbh.

It seems to be a PyCharm thing in fact.
https://stackoverflow.com/questions/23248017/cannot-find-reference-xxx-in-init-py-python-pycharm

I do not quite understand where to add the NumbaTypeSafetyWarning in the ___init___.py as it is suggested in the solutions. 

However, changing the code to:

```
# noinspection PyUnresolvedReferences
from numba import njit, types, NumbaTypeSafetyWarning

```

Also works for PyCharm, if that is ok with you, too?

sktime/transformations/panel/dictionary_based/_sfa.py
patrickzib: Sorry for my stupid question: is there any other DFT (other than the MFT for sliding windows)?

MatthewMiddlehurst: No, the option essentailly just replaces the MFT stage of the transform with the set of pre-calculated DFT's (in my case the disjoint windows from fit).

sktime/transformations/panel/dictionary_based/_sfa.py
MatthewMiddlehurst: Good idea.

MatthewMiddlehurst:
That works for now, suppressing warnings is not ideal either way so I will work towards fixing the cause. I dont think Markus would like either :)

sktime/transformations/panel/dictionary_based/_sfa.py
patrickzib: In that case I would opt for using the `if self.save_binning_dft:` code. But I will not fight for it ;)

MatthewMiddlehurst: Will go with that then, I kept it general for a potential use case which probably doesn’t exist.

sktime/transformations/panel/dictionary_based/_sfa.py
MatthewMiddlehurst: Im a little confused at which part you are talking about. The the first run is always the DFT/FFT and subsequent ones use the iterate MFT function regardless of the setting.

sktime/transformations/panel/dictionary_based/_sfa.py
MatthewMiddlehurst: From the results I have run the numba DFT seems faster than the numpy FFT anyway, but I expected some pushback with fully replacing it. Just keeping one would solve some of the issues.

If not that's a more sensible way of handling it.

sktime/transformations/panel/dictionary_based/_sfa.py
patrickzib: Sorry, you are absolutely right :) My comment was wrong. There is still the `SFA._iterate_mft` further down, that does exactly this. Sorry.

So, one uses the explicit implementation while the other one used the rfft.

But inverse_sqrt_win_size is also applied some lines further down in 635?

MatthewMiddlehurst: DFT parameter `apply_normalising_factor=False` prevents the application twice.
Output required is slightly different for binning and mft so there is also `cut_start_if_norm=False` for DFT.

sktime/transformations/panel/dictionary_based/_sfa.py
MatthewMiddlehurst: Just renamed and refactored to use booleans for now.

sktime/transformations/panel/dictionary_based/_sfa.py
patrickzib: Ah, ok, that's fine. Missed that part.

But maybe you could shorten the     `def _discrete_fourier_transform(` function using:

        if cut_start_if_norm:
            c = int(start / 2)
        else:
            c = 0
            start = 0

and only write the transformation code once?

sktime/transformations/panel/dictionary_based/_sfa.py
patrickzib: > From the results I have run the numba DFT seems faster than the numpy FFT anyway, but I expected some pushback with fully replacing it. Just keeping one would solve some of the issues.
> 
> If not that's a more sensible way of handling it.

I does makes sense, when using the first few coefficients of the DFT, as only `output_length` iterations have to be performed, and the complexity is O(output_length * n).

It might be different, when output_length becomes large, as for example with the supervised transform in WEASEL, as in that case `output_length` might become larger than log n, and thus the numpy fft is faster with O( n log n)

patrickzib: So, as a future extension, one might use the numba DFT when `anova=False` and the numpy DFT when `anova=True`

patrickzib:
@MatthewMiddlehurst Thank you very much! How do I resolve the unresolved conversations? I do not find the option here in github :)

MatthewMiddlehurst:
Does it not let you? I just have a button to do so under the reply box for each. A bit odd if it wont let the person who created them do so.

Capture

patrickzib:
> Does it not let you? I just have a button to do so under the reply box for each. A bit odd if it wont let the person who created them do it.
> 
> Capture

Funny :) No, I do not have the option. But all my concerns have been addressed 👍

TonyBagnall (approved):
I'm counting approval from Patrick as a vote so will merge this

PR #1091, opened 2021-07-01
Grid/random search forecaster using dynamic tags to check for fh

fkiraly:
This is a refactor of the below refactor.

Parts of this have now been merged as part of other PR. So in effect, this PR only makes one major change:

* in concrete grid search classes, overrides the `"requires-fh-in-fit"` tag in the constructor of the grid search algorithms with the value of the single component (the tuned estimator), using `clone_tags`.

A minor change is docstring linting in `_tune`.

---

This is a follow-on from #1088 attempting to address the incompatibility of the grid search refactor due to its "dynamic tag like" behaviour pre-refactor.

Apologies from working with the tags here, but this naturally emerged in trying to address issues and finish the forecaster refactor.

The solution I've adopted makes the following changes to the grid search algorithms:
* in concrete grid search classes, overrides the `"requires-fh-in-fit"` tag in the constructor of the grid search algorithms with the value of the single component (the tuned estimator)
* in `BaseGridSearch`, completes the refactor by moving all algorithmic logic to the `_fit`, `_predict` etc interface. This includes removing the override to the `cutoff` getter.

The first change, the override, is achived by new dynamic tag handling (object, not class) methods in the `BaseEstimator` class: getters `get_tags`, `get_tag` and setters `set_tags` and `mirror_tags`. These make use of dynamic tags that can be set in construction and override class tags and their inheritance.

The one other change is replacing the tag access in `_set_fh` by `get_tag` so it can see the dynamic tag override.


fkiraly:
FYI @GuzalBulatova

fkiraly:
@GuzalBulatova, @mloning, @TonyBagnall, if you remember we went back and forth about `_update` etc in the grid search.
This is now almost identical with the middle stage (except for the dynamic tag on fh use), please have a look whether it makes sense.

PR #1094, opened 2021-07-02
Fix seeding issue in Minirocket Classifier

### LINKED ISSUE #1041: random_state not working for MiniRocketMultivariate
Hi, 

It seems like the random_state argument is not working for MiniRocketMultivariate().

Under _fit_multi(), num_channels_per_combination is randomly generated but the seed is not set to ensure reproducibility.

Lovkush-A:
#### Reference Issues/PRs
Fixes #1041 

Added the following to the top of the method that was missing it
```python
if seed is not None:
        np.random.seed(seed)
```

No

Does this fix the issue in #1041?


fkiraly (approved):
looks like a simple fix, all good

mloning (commented):
It seems like `np.random.seed` is deprecated and no longer best practice, see https://numpy.org/doc/stable/reference/random/generated/numpy.random.seed.html. Also see `check_random_state` from scikit-learn.

fkiraly:
I see - I wasn't aware.

Do we have specs about handling random seeds anywhere? This is not included in our interface specifications or extension guidelines.

Do you have any "undocumented specs", @mloning, that you have been using? Or is this sth we need to develop?

mloning:
We usually follow scikit-learn's `check_random_state` but there were some issues in ROCKET due to numba.

TonyBagnall (approved):

TonyBagnall:
oops, I approved before seeing the comments above, sorry

Lovkush-A:
I think it should be OK to merge this change. There is already plans for bigger refactoring of rocket classifiers, and dealing with random_state could be part of that.

My main intention was to deal with #1041, and then continue working on refactoring in #1095

angus924:
> We usually follow scikit-learn's `check_random_state` but there were also some issues in ROCKET due to numba, not sure if they apply here, @angus924 knows more.

@Lovkush-A @mloning

I think the proposed fix is the best we can do for the moment, given the restrictions with numba.

mloning:
Now merged 🎉

PR #1099, opened 2021-07-03
Suggestions for PR #891 - BaseObject with dynamic tag handling functionality and dedicated get_tag interface point

fkiraly:
Suggested additions for PR #891. Can't directly PR into the 891 PR apparently since it's coming from a fork (can I?), but it's meant as a PR to that.

This adds dynamic tag functionality to the `BaseObject` class.
Methods added: getters `get_tags`, `get_tag` and setters `set_tags` and `mirror_tags`. These make use of dynamic tags that can be set in construction of any and override class tags and their inheritance (read by `_all_tags`). The methods `get_tags` and `get_tag` are meant to be the primary interface point for tags, replacing current `_all_tags` or external `has_tag` (from testing module).

This should be fully downwards compatible, as the old interface points (`all_tags`, `has_tag`) have not been removed.

Besides, I've cleaned up the merge conflicts from `main` and added docstrings at the module level.

To add more detail on how the tag related methods work:

post-PR, tags would work like this:

* each class will have static tags, that's the class variable `_tags: dict`. Individual static tags are considered to override each other as per normal nested inheritance. E.g., if a parent class and the class have different values for a static tag, the younger class' tag value counts.
* each object can, in addition, have dynamic tags, in the object variable `_tags_dynamic: dict`. These are expected to be initialized in the constructor, for instance in cases where behaviour described by tags depends on components or hyper-parameters. Dynamic tags are always considered to override the static tags - e.g., if the object's class or a parent class have a static tag of the same name, the dynamic tag's value overrides any static tag's value.

The new methods make use of this inheritance logic:

* `get_tags()` collects all static and dynamic tags and their values and returns them as a `dict` (keys = tag names, values = tag values). It overrides values precisely as described above, in the return dictionary.
* `get_tag(tag_name, tag_value_default=None)` retrieves the value of an individual tag with name `tag_name`, and optionally allows to specify a default value if the tag is not found. The inheritance/override logic is the same as in `get_tags`.
* `set_tags(dict)` sets dynamic tags according to keys (=tag names) and values (= tag values) in the argument dictionary. It is expected to be used in constructors typically.
* `mirror_tags(estimator, tag_set=None)` is a shorthand version of `set_tags`, where it sets tags according to their `get_tag` values in `estimator`. The tags that are mirrored/copied in this way can be restricted by specifying a list of tag names in `tag_set`, otherwise all tags are copied over. The mirroring is by-reference and not by-value (if the type allows this), but tags are not expected to be mutated.


fkiraly:
The dynamic tags would be very useful in resolving some inconsistencies in cases where behaviour (and thus the "true" tag) depends on components or parameter settings, e.g., determining the `requires-fh-in-fit` tag of composite forecasters based on the component forecasters. See, e.g., #1091.

mloning:
Could you explain what the different methods do (here or in a call)?

fkiraly:
>Could you explain what the different methods do (here or in a call)? Does this allow to have both static and dynamic tags at the same time?

Yes, and yes - I'll update this PR, and can write a STEP if you think it's substantial enough a change.

fkiraly:
updated, see above

mloning:
Okay, thanks, let's discuss in the next technical call! I'm generally in favour of adding dynamic tags.

fkiraly:
alright - as discussed, not in the next release though

fkiraly:
picking up from #981 now

fkiraly:
superseded by the now merged #1134

PR #1100, opened 2021-07-03
Clustering extension templates, docstrings & get_fitted_params

fkiraly:
@chrisholder, for clustering:

* I've cleaned up the docstrings in the base class
* I've added an extension template

I also noticed that you didn't have a `get_fitted_params` in the base class, I added that.

### DIFF
--- a/sktime/clustering/base/base.py
+++ b/sktime/clustering/base/base.py
@@ -61,7 +61,7 @@ class BaseClusterer(BaseEstimator):
 
         Parameters
         ----------
-        X: 3D np.array with shape (n_instances, n_columns, n_timepoints)
+        X: 2D np.array with shape (n_instances, n_timepoints)
            or pd.DataFrame in nested format
             panel of time series to cluster
 
--- a/sktime/clustering/base/base.py
+++ b/sktime/clustering/base/base.py
@@ -32,24 +32,19 @@ class BaseClusterer(BaseEstimator):
     def __init__(self):
         super(BaseClusterer, self).__init__()
 
-    def fit(self, X: NumpyOrDF, y: NumpyOrDF = None):
+    def fit(self, X: NumpyOrDF):
         """
-        Method that is used to fit the clustering algorithm
-        on the dataset X
+        Fit the clustering algorithm on the dataset X
 
         Parameters
         ----------
-        X: Numpy array or Dataframe
-            sktime data_frame or numpy array to train the model on
-
-        y: Numpy array of Dataframe, default = None
-            sktime data_frame or numpy array that is the labels for training.
-            Unlikely to be used for clustering but kept for consistency
+        X: 2D np.array with shape (n_instances, n_timepoints)
+           or pd.DataFrame in nested format
+            panel of univariate time series to train the clustering model on
 
         Returns
         -------
-        self
-            Fitted estimator
+        reference to self
         """
         if isinstance(X, pd.DataFrame):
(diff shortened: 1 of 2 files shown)


fkiraly:
PS, @chrisholder: since you're now a core developer (congrats), you could directly approve and merge this.
But, please make sure that the tests pass and nothing else breaks that you can think of.

chrisholder:
Good stuff @fkiraly.

> One potentially contentious change: I've removed the y from fitting. I don't think we need that, and you don't use it?
> If the y is there, it's time series classification. Please revert that if you disagree.

 I believe orginally we didnt have the y, but it was a comment from @mloning suggesting for consistency we should still have it in there. I am completely happy to go either way on this, so I'll let you and @mloning decide what is best here.

Just for extra context Sklearn also still has y for its clustering algorithms for consistency. This is  their fit method for their dbscan

Screenshot 2021-07-03 at 19 07 02

fkiraly:
I see - I added back `y` in the base class.

fkiraly:
added you as a reviewer, @chrisholder

fkiraly:
say, is `y` supposed to be an 1D array, or a 2D array with `.shape[1]` equals 1?

chrisholder:
> say, is `y` supposed to be an 1D array, or a 2D array with `.shape[1]` equals 1?

y : ndarray of shape (n_samples,)
            Target values. Will be cast to X's dtype if necessary.

Is usually what y is

chrisholder:
> say, is `y` supposed to be an 1D array, or a 2D array with `.shape[1]` equals 1?

So I had to check this as well since I've never thought about it but I looked through sklearns documentation and found:

'y: array-like of shape (n_samples,) or (n_samples, n_targets) Target values. Will be cast to X’s dtype if necessary

fkiraly:
yes, but I mean what is it in *your* design? I.e., what do we have to write in the doc string so it's accurate?

chrisholder:
Well my design ignores it completly so we will put the same as sklearn 

```
y : Ignored
            Not used, present here for API consistency by convention.
```

fkiraly:
I mean the returned `y`

chrisholder:
O I see sorry I was focused on fit :joy:

y on predict is just a 1d array and it has to be a row rather than a column i.e. it doesnt have .shape[1]

fkiraly:
ok, looks like that was already correct, then. Would you mind to review, then? It's templates and docstrings for the module you wrote.

sktime/clustering/base/base.py
```diff
@@ -60,81 +57,120 @@ def fit(self, X: NumpyOrDF, y: NumpyOrDF = None):
...
+            core logic
+
+        returns
+        ----------
+        param_dict: dictionary of fitted parameters
+        """
+        raise NotImplementedError
+
 
 class BaseClusterCenterInitializer:
```
chrisholder, line 175: Is this here for any reason or did you want me to write add templates for CenterInitializers, ClusterAverage, and ClusterApproximate?

fkiraly: well, I thought there are two main reasons to do clustering:

* be able to make individual cluster assignments, on existing or new data
* look at model outputs of the clustering

For instance, in mixture clustering, you may want to look at the centroids and the individual cluster's parameters (mean and precision matrix).

That's why I thought `get_fitted_params` should be in the base class, but not every method needs to implement it.
We currently have it in the forecaster too, and not every forecaster implements it either.

chrisholder (approved):

PR #1107, opened 2021-07-04
fix fh bug in predict_in_sample of ARIMA models

### LINKED ISSUE #1106: error in _predict_in_sample of ARIMA models
**Describe the bug**

I try to get all in_sample forecasts of fitted models, but error raised.

```python
ValueError: In-sample predictions undefined for start=0 when d=1
```

**To Reproduce**

```python

y = load_airline()
model = AutoARIMA()
model.fit(y)
fh = ForecastingHorizon(y.index, is_relative=False)
model.predict(fh=fh)
```

**Additional context**

I checked out the source code, the error raised because `_pmdarima.py` use all absolute `fh`, e.g. `start=0, end=143` for the airplane series. But `ARIMA.predict_in_sample()` can't be `start=0` if there is difference `d`

**Versions**

same as latest `main`

AngelPone:
#### Reference Issues/PRs

Fixes #1106 

check order `d` of `_PmdArimaAdapter._forecaster_model_` (i.e. fitted ARIMA model) and get in-sample forecasts of start from `d` instead of `0` when `start=0`, and append zeros to the `pred` and `pred_interval`

no

##### For all contributions

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -143,7 +143,6 @@ Thanks goes to these wonderful people ([emoji key](https://allcontributors.org/d
     <td align="center"><a href="https://github.com/vedazeren"><img src="https://avatars3.githubusercontent.com/u/63582874?v=4?s=100" width="100px;" alt=""/><br /><sub><b>vedazeren</b></sub></a><br /><a href="https://github.com/alan-turing-institute/sktime/commits?author=vedazeren" title="Code">💻</a> <a href="https://github.com/alan-turing-institute/sktime/commits?author=vedazeren" title="Tests">⚠️</a></td>
     <td align="center"><a href="https://github.com/vincent-nich12"><img src="https://avatars3.githubusercontent.com/u/36476633?v=4?s=100" width="100px;" alt=""/><br /><sub><b>vincent-nich12</b></sub></a><br /><a href="https://github.com/alan-turing-institute/sktime/commits?author=vincent-nich12" title="Code">💻</a></td>
     <td align="center"><a href="https://github.com/vollmersj"><img src="https://avatars2.githubusercontent.com/u/12613127?v=4?s=100" width="100px;" alt=""/><br /><sub><b>vollmersj</b></sub></a><br /><a href="https://github.com/alan-turing-institute/sktime/commits?author=vollmersj" title="Documentation">📖</a></td>
-    <td align="center"><a href="https://github.com/AngelPone"><img src="https://avatars.githubusercontent.com/u/32930283?v=4&s=100" width="100px;" alt=""><br /><sub><b>Bohan Zhang</b></sub></a><br /><a href="https://github.com/alan-turing-institute/sktime/commits?author=AngelPone" title="Code">💻</a></td>
   </tr>
 </table>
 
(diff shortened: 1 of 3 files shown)


fkiraly:
Hm, I see - is your suggestion to pad with zeroes at the time points where predictions cannot be obtained, can you confirm?

fkiraly:
Question: would the observed values not be a better choice in this case?

AngelPone:
> Hm, I see - is your suggestion to pad with zeroes at the time points where predictions cannot be obtained, can you confirm?

yes



> Question: would the observed values not be a better choice in this case?

Yes, I'll modify the behavior

fkiraly:
would you mind linting your contribution properly?
We use `black`, you can do this manually (e.g., look at linter output), or have a look at the contribution guidelines for setting up precommits.

CONTRIBUTORS.md
```diff
@@ -143,6 +143,7 @@ Thanks goes to these wonderful people ([emoji key](https://allcontributors.org/d
...
     <td align="center"><a href="https://github.com/vincent-nich12"><img src="https://avatars3.githubusercontent.com/u/36476633?v=4?s=100" width="100px;" alt=""/><br /><sub><b>vincent-nich12</b></sub></a><br /><a href="https://github.com/alan-turing-institute/sktime/commits?author=vincent-nich12" title="Code">💻</a></td>
     <td align="center"><a href="https://github.com/vollmersj"><img src="https://avatars2.githubusercontent.com/u/12613127?v=4?s=100" width="100px;" alt=""/><br /><sub><b>vollmersj</b></sub></a><br /><a href="https://github.com/alan-turing-institute/sktime/commits?author=vollmersj" title="Documentation">📖</a></td>
+    <td align="center"><a href="https://github.com/AngelPone"><img src="https://avatars.githubusercontent.com/u/32930283?v=4&s=100" width="100px;" alt=""><br /><sub><b>Bohan Zhang</b></sub></a><br /><a href="https://github.com/alan-turing-institute/sktime/commits?author=AngelPone" title="Code">💻</a></td>
```
mloning, line 146: Please change the `.all-contributorsrc` instead of this file.

mloning (commented):
Just a small comment, I'll leave the review of the other code changes to @fkiraly if you have time?

mloning (commented):
I'm not sure if padding predictions with actual data is a good idea, wouldn't that make performance estimates unreliable? How do these models usually produce in-sample predictions for the first few observations given that they are autoregressive?

AngelPone:
@mloning  agreed. I checked the [issue](https://github.com/alkaline-ml/pmdarima/issues/286) in `pmdarima`, and this should not be a problem of `sktime`.  But I still think the package should be able to provide in-sample forecasts for all observations, as implemented `arima` in R.

close the pull request and the issue.

mloning:
@AngelPone I agree that we should provide in-sample forecasts, I'm not sure how this is done though, `pmdarima` seems like the right place to implement that, but why is the linked issue closed?

AngelPone:
I closed it because `pmdarima` can't give forecasts when `start<d`(thus `sktime` can't too), and it should not be an issue in `sktime`. Besides, I have no idea how to solve it if padding with actual data isn't the right way.

Issue #1076, opened 2021-06-29
Forecasters to have a default ForecastingHorizon of fh=[1] if fh is not given (None)


aiwalter:
**Is your feature request related to a problem? Please describe.**
I would like to propose to have a default 1-step ahead `fh` for all forecasters. Motivation is, that 1-step ahead forecast is often used and a default would simplify the interface and UX.

**Describe the solution you'd like**
We could just catch cases where `fh is None` and set it to `fh=[1]`. We have to see how this works with absolute `fh`.
```python
forecaster = NaiveForecaster()
forecaster.fit(y)
forecaster.predict()
```
___
Martin Walter <martin_friedrich.walter@daimler.com>, Mercedes-Benz AG on behalf of Daimler TSS GmbH.
https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md

fkiraly:
Agreed, this is a very sensible default.

The only thing we should think about what would happen if the index of `y` is not regular, and not float/integer but some date index. Error message? Sth else?

mloning:
Agreed, this makes sense. 

One potential source of confusion may be the forecasters that require `fh` in `fit` if users are not aware of this. In that case, passing no `fh` in `fit` would result in `fh=1` during fitting and if a users passes some other `fh` in `predict` it won't work.  Not sure how to best handle this.

aiwalter:
the index type is already checked somewhere when fitting the model I think?

aiwalter:
@mloning we could handle this by just setting the default `fh=[1]` in `predict()` and not before? Not sure also if the default should be given in the function already like
```python
def predict(fh=[1], X=None):
    sth = "sth"
    return y_pred
```
or if we should still have `fh=None` and then have a separate check inside `predict()` to set default `fh`?

TonyBagnall:
having just done some forecasting from completely cold, I agree this is a good idea, it took me a bit  of time to figure out the horizon

mloning:
@aiwalter the problem with that is that if people specify a different horizon in `_fit` for the forecasters that need the horizon, having a default in `_predict` which may be different from what was passed in `_fit` may lead to errors and more confusion. 

If we just want to provide a default for the forecasters that do not require `fh` in `fit`, it needs to happen in `_set_fh` I think. Alternatively, we can set the default in `fit` for all forecasters, but then users may be confused if they want to pass it `predict`. This turns out to be a bit trickier than I thought ...

[...]

PR #1109, opened 2021-07-04
Forecasting: base/template docstring fixes, added fit_predict method

fkiraly:
Some small changes to the forecasting base class and template:

* updates to docstrings, clarifications
* added `fit_predict` method which does what one would "obviously" expect, fitting & forecast in one step.

### DIFF
--- a/sktime/forecasting/base/_base.py
+++ b/sktime/forecasting/base/_base.py
@@ -158,9 +158,9 @@ class BaseForecaster(BaseEstimator):
 
         return self._predict(self.fh, X, return_pred_int=return_pred_int, alpha=alpha)
 
-    def fit_predict(self, y, X=None, fh=None,
-                    return_pred_int=False, alpha=DEFAULT_ALPHA
-                    ):
+    def fit_predict(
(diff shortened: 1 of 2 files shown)


sktime/forecasting/base/_base.py
```diff
@@ -161,6 +158,38 @@ def predict(self, fh=None, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
...
+        Returns
+        -------
+        y_pred : pd.Series
+            Point predictions
+        y_pred_int : pd.DataFrame - only if return_pred_int=True
+            Prediction intervals
+        """
+
+        if fh is None:
+            fh = y.index
```
mloning, line 187:
```suggestion
```

mloning (commented):
Looks all good to me apart from the default `fh` behaviour, did you leave that in unintentionally or is this a special case where the default would make sense?

fkiraly:
I accepted the suggestion on th `fh` default. Let's not get ahead of the resolution of our discussion in #1076.

fkiraly:
merged since @mloning was happy except for one change request which was accepted

PR #1118, opened 2021-07-07
Fix ForecastingPipeline import

mloning:
#### Reference Issues/PRs
#967 

#### What does this implement/fix? Explain your changes.
* Fixes import error from merging into main without first updating from main

PR #1119, opened 2021-07-07
post-fix merge error in #967

fkiraly:
This is a post-fix to #1118 to make this consistent with changes introduced in #1068.

I.e., removes the new instance of `_has_tag` from #967 in favour of `_all_tags`, as done in #1068 with the other instance in the same class.


mloning:
`_has_tag` is used elsewhere in the library too, that's why I thought it may be best to refactor that everywhere once we agree on the new interface points for the tag system. 

The PR breaks because of the default setting if the tag is missing.

fkiraly:
>The PR breaks because of the default setting if the tag is missing.

Yes, we didn't fix it for transformers...

fkiraly:
so let's just do this later.

fkiraly:
at least I was wise enough to merge your PR and not mine (it wasn't tested).

fkiraly:
no longer needed since replaced by `get_tags` since then.

PR #1127, opened 2021-07-10
Add ExponentTransformer and SqrtTransformer

### LINKED ISSUE #1126: Square Root Transformer
Square root transformations are often referenced as a potential transformation in forecasting, including in Practical TImeseries Analysis by Aileen Nielsen and in Rob Hyndman's forecasting packages in R. 

This can easily be implemented as a series-to-series transformer.

The alternative is to not include this functionality. Users could transform their series prior to use with a `sktime` forecaster. However, packages in other languages include the functionality and it is referenced by relevant texts so including it allows us maintain functionality that users may commonly expect.

**Additional context**
N/A

RNKuhns:
#### Reference Issues/PRs

Resolves #1126.

Implements square root transformation as a series-to-series transformer.

No

#### Any other comments?

The implementation is simple, so the review should hopefully be straight-forward. Note that I'm relying on the test_all_transformations.py that checks equality between original data and transform->inverse_transform.

Note that it is a minor point, but what are your thoughts on the name? For some reason `SqrtTransformer` looks weird with "qrt" lowercase, but at the same time `SQRTTransformer` does not seem right either and `SquareRootTransformer` is long given that "sqrt" is a common abbreviation (and used by NumPy) for square root).

##### For all contributions

##### For new estimators


fkiraly:
this breaks since all estimators should write parameters to self in init

fkiraly:
Quick question: do you want to have a `SqrtTransformer`, or rather a `PowerTransformer`, where you can specify the exponent? `SqrtTransformer` could be a shorthand, of course.

You also often see people squaring residuals, so a `SquareTransformer` might be useful.

RNKuhns:
That's a good question. I just got back and saw the error (I only ran the transformer tests locally prior to heading out). 

I am thinking PowerTransformer makes sense. I also want to add an "offset" parameter that is added prior to transformation so that square root can be applied to data with negative values too. That deals with the lack of hyperparameters either way.

Question is whether to make offset a boolean and to automatically offset if negative value in the input series or to make that a user supplied int/float (e.g. user has to decide what value to supply).

Third option (my preference at the moment) is to accept "auto" (it is automatically determined) or a int/float. 

@fkiraly I've got two quick questions for you:

- What's your preference on the offset options I described?
- When you say shorthand for SqrtTransformer, are you suggesting we have a separate class that implements the the common use (e.g.: SqrtTransformer and SquareTransformer by inheriting from PowerTransformer and fixing the power hyperparameter?)

fkiraly:
>What's your preference on the offset options I described?

my preference would actually be to take the power of the absolute value by default.
An offset can also be added, why not - the default would be zero then?

>When you say shorthand for SqrtTransformer, are you suggesting we have a separate class that implements the the common use (e.g.: SqrtTransformer and SquareTransformer by inheriting from PowerTransformer and fixing the power hyperparameter?)

Yes, though I'm not sure whether that's introducing too many classes.

RNKuhns:
@fkiraly I've updated to have `PowerTransformer` and `SqrtTransformer`. But I did not implement the absolute value functionality. The reason is b/c I'm thinking that the absolute value transformation would mean that would make it difficult to use in a forecasting pipeline and recover forecasts in the original units (since a clear inverse of absolute value transformation doesn't jump out to me).

aiwalter:
Probably not best idea to use `PowerTransformer` as a name, bcz there is already `sklearn.preprocessing.PowerTransformer`, so it would be a name conflict. Maybe we can just use `Power` instead? :) Bcz also all our other transformers dont actually have "transformer" in the class name, right?

fkiraly:
>Probably not best idea to use PowerTransformer as a name, bcz there is already sklearn.preprocessing.PowerTransformer, so it would be a name conflict. 

Oh, of course, the boxcox one. Hm not sure about the naming. Some transformers end in `Transformer`, others don't.

>But I did not implement the absolute value functionality. The reason is b/c I'm thinking that the absolute value transformation would mean that would make it difficult to use in a forecasting pipeline and recover forecasts in the original units (since a clear inverse of absolute value transformation doesn't jump out to me).

Regarding that, if you don't do this, you may end up with non-obvious errors if you have negative values.

Another alternative would be to implement the function $x\mapsto \sgn(x) |x-a|^\alpha$, i.e., for offset $a=0$, it's the $\alpha$-th power if $x$ is positive, otherwise is minus the $\alpha$-th power. For other offsets, the function is shifted.

RNKuhns:
> Probably not best idea to use `PowerTransformer` as a name, bcz there is already `sklearn.preprocessing.PowerTransformer`, so it would be a name conflict. Maybe we can just use `Power` instead? :) Bcz also all our other transformers dont actually have "transformer" in the class name, right?

Good catch @aiwalter. I'm going to switch it to `ExponentTransformer()`.

RNKuhns:
> Regarding that, if you don't do this, you may end up with non-obvious errors if you have negative values.

> Another alternative would be to implement the function $x\mapsto \sgn(x) |x-a|^\alpha$, i.e., for offset $a=0$, it's the $\alpha$-th power if $x$ is positive, otherwise is minus the $\alpha$-th power. For other offsets, the function is shifted.

@fkiraly I've currently got it set up so that when offset="auto" (the default) the code determines if the there are any negative values in the input series and if calculates an "offset value" as the absolute value of the most negative value and adds it to each element prior to applying the exponent. 

The `transform` is then $(X + _offset_value)^{power}$ assuming that _offset_value is determined as I described above. Then the `inverse_tranform` is $(X)^{1/power} - _offset_value$. 

In the case of your version of the transformation, I don't know that we actually would need the offset. I'm only including it as a means to ensure non-negative values, so the calculations don't result in NaNs like you mentioned. Setting the offset aside in your proposed transformation, we'd have: $sign(x)|x|^{\alpha}$ as the `transform` and I'm assuming the `inverse_transform` would be $sign(x)|x|^{1/\alpha}$. 

Both transformations accomplish the goal (having the ability to do square root transformation along with more general exponent transformation). In each case, if the series is all positive, the result is the same. The approaches differ in how they treat series with negative values. Since both are reversible and forecasting pipelines are most interested in the forecasts in the original units, they both work in that regard too. Based on that I don't have any strong feeling one way or the other. 

@aiwalter do you have any preference?

mloning (approved):
Looks all good to me @RNKuhns

mloning:
Now merged @RNKuhns! 🎉

PR #1130, opened 2021-07-12
Fix naive model with last strategy for cases with trailing NaN values

### LINKED ISSUE #918: [BUG] Naive model with last strategy is sensitive to trailing NaN values
**Describe the bug**

The naive model with "last" strategy really picks the last value, even if that's a NaN value. Can we make this more robust, similar to how the "mean" strategy actually does a `nanmean` under the hood? I could offer a PR to make it pick the last non-NaN value, taking into account the requested seasonal periodicity.

**To Reproduce**

```python
import numpy as np
import pandas as pd

data = pd.Series(
    index=pd.date_range(
        "2021-06-01 00:00", periods=12, freq="1h", closed="left"
    ),
    data=[1, 2, 3, 1, 2, 3, np.nan, 2, 3, 1, 2, 3]
)
train_data = data[:7]
test_data = data[7:]

print(train_data)
# 2021-06-01 00:00:00    1.0
# 2021-06-01 01:00:00    2.0
# 2021-06-01 02:00:00    3.0
# 2021-06-01 03:00:00    1.0
# 2021-06-01 04:00:00    2.0
# 2021-06-01 05:00:00    3.0
# 2021-06-01 06:00:00    NaN
# Freq: H, dtype: float64

fh = ForecastingHorizon(test_data.index, is_relative=False)
model = NaiveForecaster(strategy="last", sp=3)
model.fit(train_data)
forecast_data = model.predict(fh)

print(forecast_data)
# 2021-06-01 07:00:00    2.0
# 2021-06-01 08:00:00    3.0
# 2021-06-01 09:00:00    NaN  # Let's make this a 1
# 2021-06-01 10:00:00    2.0
# 2021-06-01 11:00:00    3.0
# dtype: float64
```

**Expected behavior**

**Additional context**

**Versions**
<details>

</details>

Flix6x:
#### Reference Issues/PRs

Fixes #918.

The logic for the `"last"` strategy is now much more similar to the logic for the `"mean"` strategy. I refactored some of the logic to internal util functions so that it's clear that both strategies transform the `last_window` according to the seasonal periodicity in the same way, and also tile the results in the same way. The only difference is now that the `"last"` strategy selects the last non-NaN value from each seasonally periodic row, while the `"mean"` strategy computes the mean of said row.

Just to note, to me it would make sense if the `"drift"` strategy would use the same util functions to:
- make it robust against NaN values, both at the end of the time series and at the start, and
- let it account for seasonal periodicity.
I think this would be a separate issue, but just to be clear: I am not considering working on it, unless someone points me to a use case for this particular strategy.

No.

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/forecasting/naive.py
+++ b/sktime/forecasting/naive.py
@@ -85,25 +85,8 @@ class NaiveForecaster(_BaseWindowForecaster):
 
         n_timepoints = y.shape[0]
 
-        if self.strategy == "last":
-            if self.sp == 1:
-                if self.window_length is not None:
-                    warn(
-                        "For the `last` strategy, "
-                        "the `window_length` value will be ignored if `sp` "
-                        "== 1."
-                    )
-                self.window_length_ = 1
-
-            else:
-                self.sp_ = check_sp(self.sp)
-
-                # window length we need for forecasts is just the
-                # length of seasonal periodicity
-                self.window_length_ = self.sp_
-
-        elif self.strategy == "mean":
-            # check window length is greater than sp for seasonal mean
+        if self.strategy in ("last", "mean"):
+            # check window length is greater than sp for seasonal mean or seasonal last
             if self.window_length is not None and self.sp != 1:
                 if self.window_length < self.sp:
                     raise ValueError(
@@ -114,7 +97,7 @@ class NaiveForecaster(_BaseWindowForecaster):
             self.window_length_ = check_window_length(self.window_length, n_timepoints)
             self.sp_ = check_sp(self.sp)
 
-            #  if not given, set default window length for the mean strategy
+            #  if not given, set default window length
             if self.window_length is None:
                 self.window_length_ = len(y)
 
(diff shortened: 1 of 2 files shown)


fkiraly (changes requested):
To me, this looks very sensible! Thanks for the change.
I think this makes the method a lot more robust.

Now, I have to admit, I wanted to review this earlier, but I had to take some time for this since the docstrings aren't too great - otherwise I would have done with this much earlier.

May I hence kindly ask:
* in the class docstring, explain better what is meant with "last window"
* in the class docstring, explain how nans are now handled
* add docstrings to the new and old utility functions, you will do future developers a big favour...
* add docstrings to the new tests - what does this tests for, under which conditions are what errors raised?

Another point: should the tag `"handles-missing-values"` now  be switched to `True`? The default is `False` if it's not set.

fkiraly:
@Flix6x, are you still working on this?

Flix6x:
> @Flix6x, are you still working on this?

Little busy, but I plan to get back to this on Wednesday evening or Monday evening. Your documentation requests seem reasonable. I'd need to find out how I could set that tag, though. Is that on a class level or strategy level? If it's on a class level, I don't think the drift strategy is robust against missing values, so then it should still be `False`, but maybe made explicit and with a corresponding `# todo: switch to True if GH<number of issue that addresses robustness against nan values for the "drift" strategy> is fixed`?

fkiraly:
>If it's on a class level, I don't think the drift strategy is robust against missing values, so then it should still be False

on the class level, it should be the "worst possible value". If you want, you can do it on the object/strategy level by using `set_tags`.

Flix6x:
> If you want, you can do it on the object/strategy level by using `set_tags`.

Thanks for the hint. I did just that.

> add docstrings to the new tests - what does this tests for, under which conditions are what errors raised?

I added slightly more detail to the docstring and an inline comment. It's already pretty detailed compared to other tests.

sktime/forecasting/naive.py
```diff
@@ -4,7 +4,7 @@
 """Implements simple forecasts based on naive assumptions."""
 
 __all__ = ["NaiveForecaster"]
-__author__ = ["Markus Löning", "Piyush Gade"]
+__author__ = ["Markus Löning", "Piyush Gade", "Felix Claessen"]
```
fkiraly, line 7: this should be GitHub IDs, would zou kindly change that? Markus is mloning

sktime/forecasting/naive.py
```diff
@@ -60,14 +73,22 @@ class NaiveForecaster(_BaseWindowForecaster):
...
     def __init__(self, strategy="last", window_length=None, sp=1):
         super(NaiveForecaster, self).__init__()
         self.strategy = strategy
         self.sp = sp
         self.window_length = window_length
 
+        # Override tag for handling missing data
+        # todo: remove if GH1367 is fixed
+        if self.strategy in ("last", "mean"):
+            self.set_tags(**{"handles-missing-data": True})
```
fkiraly, line 90: you could also write, perhaps more easy to read: `self.set_tags(handles-missing-data=True)`

sktime/forecasting/tests/test_naive.py
```diff
@@ -195,35 +203,43 @@ def test_strategy_mean_seasonal_additional_combinations(n, window_length, sp):
...
-    # let's forecast the next <2 x period> hours with a periodicity of <sp> hours
+    # forecast the next <(n-1) x window_length> hours with periodicity of <sp> hours
     fh = ForecastingHorizon(test_data.index, is_relative=False)
-    model = NaiveForecaster(strategy="mean", sp=sp)
+    model = NaiveForecaster(strategy=strategy, sp=sp)
     model.fit(train_data)
     forecast_data = model.predict(fh)
 
+    # Make sure that the model (object) reports that it handles missing data
+    assert model.get_tag("handles-missing-data") is True
```
fkiraly, line 241: the `is True` is not needed here

fkiraly (dismissed):
looks good from my side, I now understand what this is doing - I added minor comments, neither a blocker.

fkiraly:
but, obviously, you also need to fix:

* doc quality checks
* should be up-to-date with main

(can't be merged without this)

sktime/forecasting/naive.py
Flix6x: I don't think that is possible since the name would be interpreted with two minus symbols.

fkiraly: oh, ouch - we should have thought of this in naming tags.
In fact, tags strings should be valid variable names.

mloning (approved):
Looks all good to me - can we merge @fkiraly?

fkiraly (approved):
yes, all fine - doc checks pass now.

mloning:
Thanks @Flix6x and sorry again for the very long delay on this one - now it's merged 🎉

PR #1134, opened 2021-07-13
BaseObject and rework of tags system, including dynamic tags

fkiraly:
Rework of PR #1099 which is in turn a rework of PR #891 (by @RNKuhns), based on discussion recorded in #981 on tags.

Adds a `BaseObject` class which adds dynamic tag functionality. `BaseObject` contains `BaseEstimator` functionality sans `fit` state change related interfaces.
The new functionality replaces the inconsistent mixture of old interface points (`all_tags`, `has_tag`) in all downstream dependencies.

post-PR, tags work like this:
* each class will have "class tags", stored in the content of the class variable `_tags: dict`. Individual class tags are considered to override each other as per normal nested inheritance. E.g., if a parent class and the class have different values for a class tag, the younger class' class tag value counts.
* each object can, in addition, have dynamic tags, in the object variable `_tags_dynamic: dict`. These are expected to be initialized in the constructor, for instance in cases where behaviour described by tags depends on components or hyper-parameters. Dynamic tags are always considered to override the class tags - e.g., if the object's class or a parent class have a static tag of the same name, the dynamic tag's value overrides any class tag's value.

The new tag related methods make use of this inheritance logic are as follows.

class methods:
* `get_class_tags()` collects all class tags and their values and returns them as a `dict` (keys = tag names, values = tag values). It overrides values precisely as described above, in the return dictionary. This replaces the old `_all_tags`.
* `get_class_tag(tag_name, tag_value_default=None)` retrieves the value of an individual tag with name `tag_name`, and optionally allows to specify a default value if the tag is not found. The inheritance/override logic is the same as in `get_class_tags`.

object methods:
* `get_tags()` collects all class tags and dynamic tags and their values and returns them as a `dict` (keys = tag names, values = tag values). It overrides values precisely as described above, in the return dictionary.
* `get_tag(tag_name, tag_value_default=None)` retrieves the value of an individual tag with name `tag_name`, and optionally allows to specify a default value if the tag is not found. The inheritance/override logic is the same as in `get_tags`.
* `set_tags(**tag_dict)` sets dynamic tags according to keys (=tag names) and values (= tag values) in the argument dictionary. It is expected to be used in constructors typically.
* `mirror_tags(estimator, tag_set=None)` is a shorthand version of `set_tags`, where it sets tags according to their `get_tag` values in `estimator`. The tags that are mirrored/copied in this way can be restricted by specifying a list of tag names in `tag_set`, otherwise all tags are copied over. The mirroring is by-reference and not by-value (if the type allows this), but tags are not expected to be mutated.


sktime/base/_base.py
```diff
@@ -14,14 +54,160 @@
...
+        -------
+        tag_value : value of the tag tag_name in self if found
+                    if tag is not found, returns tag_value_default
+        """
+        collected_tags = cls.get_class_tags()
+
+        if tag_name in collected_tags.keys():
+            return collected_tags[tag_name]
+        else:
+            return tag_value_default
```
mloning, line 110:
```suggestion
        return collected_tags.get(tag_name, tag_value_default)
```

sktime/base/_base.py
```diff
@@ -14,14 +54,160 @@
...
+    def get_tags(self):
+        """Get tags from estimator class and dynamic tag overrides.
+
+        Returns
+        -------
+        collected_tags : dictionary of tag names : tag values
+            collected from _tags class attribute via nested inheritance
+            then any overrides and new tags from _tags_dynamic object attribute
+        """
+        collected_tags = type(self).get_class_tags().copy()
```
mloning, line 121:
```suggestion
        collected_tags = self.get_class_tags()
```
Does this not work?

sktime/base/_base.py
```diff
@@ -14,14 +54,160 @@
...
 
-class BaseEstimator(_BaseEstimator):
+class BaseObject(_BaseEstimator):
+    """Base class for parametric objects with tags sktime.
+
+    Extends scikit-learn's BaseEstimator to include sktime interface for tags.
+    """
+
+    def __init__(self):
+        self._tags_dynamic = dict()
```
mloning, line 64: If we name the others "class tags", we should name these "object tags" 
```suggestion
        self._object_tags = dict()
```

sktime/base/_base.py
```diff
@@ -14,14 +54,160 @@
...
+
+        if hasattr(self, "_tags_dynamic"):
+            collected_tags.update(self._tags_dynamic)
+
+        return collected_tags
+
+    def get_tag(self, tag_name, tag_value_default=None):
+        """Get tag value from estimator class and dynamic tag overrides.
+
+        Arguments
```
mloning, line 131: Here and throughout
```suggestion
        Parameters
```

sktime/base/_base.py
```diff
@@ -14,14 +54,160 @@
...
+        -------
+        tag_value : value of the tag tag_name in self if found
+                    if tag is not found, returns tag_value_default
+        """
+        collected_tags = self.get_tags()
+
+        if tag_name in collected_tags.keys():
+            return collected_tags[tag_name]
+        else:
+            return tag_value_default
```
mloning, line 146: Same as above:
```suggestion
        return collected_tags.get(tag_name, tag_value_default)
```

sktime/base/_base.py
```diff
@@ -14,14 +54,160 @@
...
+
+        Returns
+        -------
+        reference to self
+
+        State change
+        ------------
+        sets tag values in tag_dict as dynamic tags in self
+        """
+        self._tags_dynamic.update(tag_dict.copy())
```
mloning, line 163: Do we need to copy the dict here?
```suggestion
        self._tags_dynamic.update(tag_dict)
```

sktime/base/_base.py
```diff
@@ -14,14 +54,160 @@
...
+
+        return self
+
+    def mirror_tags(self, estimator, tag_set=None):
```
mloning, line 167:
```suggestion
    def mirror_tags(self, estimator, tag_names=None):
```

sktime/base/_base.py
```diff
@@ -14,14 +54,160 @@
...
+
+        Returns
+        -------
+        reference to self
+
+        State change
+        ------------
+        sets tag values in tag_set from estimator as dynamic tags in self
+        """
+        tags_est = estimator.get_tags().copy()
```
mloning, line 184: Why all the copies here? 
```suggestion
        tags_est = estimator.get_tags()
```

sktime/base/_base.py
mloning: I was thinking that perhaps `clone_tags` will also be more intuitive given that we have `clone` in scikit-learn already. Now that I understand what it does I also like the mirror metaphor but it wasn't clear to me intuitively whereas clone tags would have been I believe. 
```suggestion
    def clone_tags(self, estimator, tag_set=None):
```

mloning (commented):
Had a first look at it, haven't reviewed all of it yet.

A few additional comments, perhaps for later implementation/PR:
* in `set_tags` we should check that the passed tags are valid tags for the given estimator

fkiraly:
>in set_tags we should check that the passed tags are valid tags for the given estimator

I'd prefer not to, since this couples the `BaseObject` to the registry.
I also think we want to give the extender - or even the user - the opportunity to set their own tags easily, if they just work locally and hack around.

sktime/base/_base.py
fkiraly: `dict` are mutable, this is defensive programming to avoid side effects that ultimately could even impact the class dictionary! I.e., an unsuspecting user or extendar could mutate one return dictionary and break tag lookup mechanics for all of sktime.

sktime/base/_base.py
fkiraly: oh yes, stupid me

sktime/base/_base.py
fkiraly: hm, I'd prefer not to, conceptually I still feel it's more dynamic overrides of class tags rather than tags that are specifically object tags, because conceptually the class corresponds to the algorithm.

I prefer to think about the dynamic tags as "conditional class tags", i.e., specifying a class tag which is a function of component class tags.

sktime/base/_base.py
fkiraly: yes, indeed, since it uses `sklearn` terminology

sktime/base/_base.py
mloning: but that should be taken care of in `get_tags`, which it is I believe (we create a new dict every time we call the method)

fkiraly: well, but you also want to remove the `copy` there, so you would need at least one.

In any case, I think it's needed in:
* `get_class_tags`
* `get_tags`
* at the start of `set_tags`
* in `clone_tags`

to avoid the various entry points for side effects due to mutability of `dict`

sktime/base/_base.py
mloning: I think the object/class or static/dynamic distinction makes sense, we can explain details in the docs, but when we call one "class_tags" and the other "dynamic_tags" their relation isn't even clear

sktime/base/_base.py
mloning: inside `get_class_tags` we create a new dict, so I'd say no need for copies, if you're worried about side effects, we should test for them

fkiraly: yes, that´s true. But we´d still need the one in `get_tags`, right?

sktime/base/_base.py
fkiraly: the reference to dynamic only happens internally, users never see this.

The general terminology is "tags" and "class tags". Introducing any dichotomy will make the user think there are two sets of tags. Instead they should call `get_tags` primarily, this has *no qualifier* and will retrieve all tags (both object/class, or both static/dynamic).

So I think the current naming makes sense, except perhaps the internal variables which neither users nor extenders should ever access, only core developers working specifically with `BaseObject`.

sktime/base/_base.py
fkiraly: yes, that works too

fkiraly: changed

sktime/base/_base.py
fkiraly: not sure whether `update` is safe

fkiraly:
have addressed change requests, except one:

I´d rather not rename `get_tags` to `get_object_tags`, because it retrieves all tags (not just object/dynamic ones).

sktime/base/_base.py
mloning: I know, but for core devs it would still make the relation clearer between the two, but it's a minor point

sktime/base/_base.py
mloning: `update` mutates the dict on which you call `update`, not the one is which passed

sktime/base/_base.py
```diff
@@ -14,14 +54,154 @@
...
+    def get_tags(self):
+        """Get tags from estimator class and dynamic tag overrides.
+
+        Returns
+        -------
+        collected_tags : dictionary of tag names : tag values
+            collected from _tags class attribute via nested inheritance
+            then any overrides and new tags from _tags_dynamic object attribute
+        """
+        collected_tags = self.get_class_tags().copy()
```
mloning, line 118: Why do we need the copy here? `get_class_tags` always returns a fresh dictionary, if there are any issues with side effects, they should be taken care of inside `get_class_tags` (but don't think there are)
```suggestion
        collected_tags = self.get_class_tags()
```

sktime/base/_base.py
fkiraly: but by reference - now I think we want to `deepcopy` actually

fkiraly:
so, I've thought about the `copy` issue.
I think my concerns about possible side effects are valid - mutability is an issue, but operations that pass reference rather than value are also an entry point for potential side effects. This is a potential issue since tags are no longer just boolean, so let's be defensive here.

I've now added a `deepcopy` in these places: at the end of `get_tags`, `get_class_tags`, and at the start of `set_tags` and `clone_tags`. This should make sense (because it "breaks" any references).

sktime/base/_base.py
fkiraly: I have no strong feeling, but I think it's clearer that way - implicit dichotomies are strong in (mis)guiding perception, even (and especially) if they are fake

fkiraly:
@mloning, can you please have a look? #980 is conditional on this, on which other stuff is conditional.
Would appreciate if you help unblocking the bottleneck.

fkiraly:
(sorry for being pushy, but I like to believe it's for the best)

fkiraly:
from discussion: value range checks should be done by object level tests
this will need additional convention in registry on items with discrete value space, e.g., which strings are admissible

sktime/base/_base.py
```diff
@@ -1,27 +1,209 @@
...
+        -------
+        collected_tags : dictionary of tag names : tag values
+            collected from _tags class attribute via nested inheritance
+            NOT overridden by dynamic tags set by set_tags or mirror_tags
+        """
+        collected_tags = dict()
+
+        # We exclude the last two parent classes: sklearn.base.BaseEstimator and
+        # the basic Python object.
+        for parent_class in reversed(inspect.getmro(cls)[:-2]):
```
mloning, line 83: Do we need to change that here since we add another parent class, assuming that the BaseObject doesn't carry tags?
```suggestion
        for parent_class in reversed(inspect.getmro(cls)[:-3]):
```

sktime/classification/base.py
```diff
@@ -77,7 +77,7 @@ def fit(self, X, y):
         sets is_fitted flag to true
         """
 
-        coerce_to_numpy = self._all_tags()["coerce-X-to-numpy"]
+        coerce_to_numpy = self.get_class_tag("coerce-X-to-numpy", False)
```
mloning, line 80: as discussed, here and throughout, I'd remove the setting of default values here and instead do that in the BaseClassifier (or corresponding base class)
```suggestion
        coerce_to_numpy = self.get_class_tag("coerce-X-to-numpy")
```

mloning (commented):
Testing is still missing, everything else looks good, only a few minor comments.

sktime/base/_base.py
fkiraly: I wouldn't say that `BaseObject` will never carry tags - where else would you put defaults for tags that all objects have?

sktime/classification/base.py
fkiraly: ultimately, yes, very much agreed. But, I think the tag defaults are consistently set only in forecasting (where I've removed the defaults from `get_` etc). I wouldn't dare remove the defaults from classification until we've completed the interface refactor there...

sktime/base/_base.py
fkiraly: resolved in discussion (slightly paranoid use of `deepcopy` it is for now, we can revisit later)

fkiraly:
>Testing is still missing

No longer!

sktime/base/_base.py
mloning: okay makes sense

sktime/classification/base.py
mloning: ok

sktime/utils/_testing/deep_equals.py
```diff
@@ -0,0 +1,131 @@
...
+__author__ = ["fkiraly"]
+
+__all__ = ["deep_equals"]
+
+
+import numpy as np
+import pandas as pd
+
+
+def deep_equals(x, y):
```
mloning, line 20: as before, I still think this is overcomplicated and the costs here outweigh the benefits of having arbitrary tag values, for boolean and string tags there exist well-tested inbuilt comparison functionality

sktime/base/tests/test_base.py
```diff
@@ -0,0 +1,148 @@
...
+    AssertError if inheritance logic in get_class_tags is incorrect
+    """
+    child_tags = FIXTURE_CLASSCHILD.get_class_tags()
+
+    msg = "Inheritance logic in BaseObject.get_class_tags is incorrect"
+
+    assert deep_equals(child_tags, FIXTURE_CLASSCHILD_TAGS), msg
+
+
+def test_get_class_tag():
```
mloning, line 61: could we make test names more descriptive, what exactly does this test?

sktime/base/tests/test_base.py
```diff
@@ -0,0 +1,148 @@
...
+
+# Fixture class for testing tag system, child overrides tags
+class FixtureClassChild(FixtureClassParent):
+
+    _tags = {"A": 42, 3: "E"}
+
+
+FIXTURE_CLASSCHILD = FixtureClassChild
+
+FIXTURE_CLASSCHILD_TAGS = {"A": 42, "B": 2, "C": [], 3: "E"}
```
mloning, line 38: Quick question, why do we test for lists and integers in tag values?

mloning (commented):
A few quick comments

sktime/base/tests/test_base.py
fkiraly: I thought bool, int, and string are fine. I replaced the list.

sktime/base/tests/test_base.py
fkiraly: `test_get_class_tag_core_logic_for_correctness` is not very descriptive, is it?

Not sure what would be a better alternative. `get_tag` has one purpose and one purpose only, similar to the others - so isn't it pretty clear what is being tested?

sktime/utils/_testing/deep_equals.py
fkiraly: removed, resolved.

sktime/base/tests/test_base.py
```diff
@@ -0,0 +1,148 @@
...
+__author__ = ["fkiraly"]
+
+from copy import deepcopy
+
+from sktime.base import BaseObject
+from sktime.utils._testing.deep_equals import deep_equals
+
+
+# Fixture class for testing tag system
+class FixtureClassParent(BaseObject):
```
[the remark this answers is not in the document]
fkiraly: hm yes, but there are no side effects in the tests, because no state change is tested?

The one case where it may make sense is for `set_tags` but there is no reuse issue since there is only one test?

sktime/utils/_testing/deep_equals.py
```diff
@@ -0,0 +1,131 @@
...
+    bool - True if x and y are equal in value
+        x and y do not need to be equal in reference
+    """
+
+    if type(x) != type(y):
+        return False
+
+    # we now know all types are the same
+    # so now we compare values
+    if type(x) in [pd.DataFrame, pd.Series]:
```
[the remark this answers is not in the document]
fkiraly: removed, so no longer relevant

fkiraly:
tests run locally, builds were ok before merging doc related PR, so it's fine to merge

PR #1137, opened 2021-07-14
change to classification experiments

TonyBagnall:
remove the need for cython to run classification_experiments, set up example problem from datasets/data/UnitTest


TonyBagnall:
This is contrib only so will push it through if that's ok @mloning

TonyBagnall:
this has already passed, I will force this in too to try stop the deadlock (or maybe livelock?) we seem to be getting on PRs

PR #1138, opened 2021-07-14
Add estimator overview to docs

mloning:
#### Reference issue/PR
Supersedes #930 

#### What does this PR implement? 
Building on #930, it:
* adds a non-searchable estimator overview table in markdown 
* the table is generated automatically every time the docs are build based on the source code
* minor updates to the docs deployment

#### Things to improve
* make the table searchable/queriable
* link author names to their GitHub profiles encouraging people to use GitHub handles rather than names in `__author__`


fkiraly (dismissed):
approving this on the basis of the python code (data frame creation).

I don't know how to test the table generation, so this may need review by a different person.

mloning:
Note to self: We need to remove/integrate `ESTIMATOR_OVERVIEW.md`

docs/source/conf.py
```diff
@@ -209,13 +213,91 @@ def find_source():
...
+            + '.html">'
+            + modname
+            + "</a>"
+        )
+
+        df = df.append(
+            pd.Series([modname, algorithm_type, author_info], index=COLNAMES),
+            ignore_index=True,
+        )
+    # creates a table in html format
```
afzal442, line 287:
```suggestion
    # creates a table in markdown format
```

afzal442 (approved):
LGTM!

afzal442:
Another point to ask is, could we customize the size of the table in markdown format? It looks like a bit larger expanded to the right side.

docs/source/conf.py
mloning: Thanks!

mloning (commented):
Self-reviewed - looks all good to me.

mloning:
Now merged - thanks again @afzal442 @astrojuanlu 🎉

Issue #1136, opened 2021-07-14
ForecastingGridSearchCV to optionally return also best n forecaster


aiwalter:
**Is your feature request related to a problem? Please describe.**
Currently, `ForecastingGridSearchCV` can only return `best_forecaster_`, however I would like to also get best `n` forecasters in order to create an `EnsembleForecaster` with them. Ensembling would probably especially work well in combination with using `MultiplexForecaster` to find the models which to use inside the ensemble.

**Describe the solution you'd like**
We could add a param like `return_n_forecaster=1` and if `>1` we could return a list `best_n_forecasters_`. The information of best `n` forecaster can be easily extracted from `results`, like done here for `best_forecaster_`:
https://github.com/alan-turing-institute/sktime/blob/7343dc36566935b42167878c45d570099af96dab/sktime/forecasting/model_selection/_tune.py#L292-L296

**Describe alternatives you've considered**
I could write a custom function to get this forecasters without contributing it to `sktime`. However I would need to be able to access `results` which is a `pd.DataFrame`.
___
Martin Walter <martin_friedrich.walter@daimler.com>, Mercedes-Benz AG on behalf of Daimler TSS GmbH.
https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md

mloning:
Sounds like a reasonable extension to me. But can you not get them from `gscv.cv_results_` already? 

If we want to add this, I would:
* call the kwarg `return_n_best_forecasters=1`
* add object variable `n_best_forecasters_` which is a list of length `(f"rank_{i}", <forecaster>)` tuples (for i = 1 ... n), that way they can be passed directly into an ensemble class

aiwalter:
Ok, thx for quick feedback. Ah yes, using `gscv.cv_results_` would already work also probably. I will implement this then as you have proposed.

aiwalter:
closed by #1139

PR #1139, opened 2021-07-15
Added n_best_forecasters to grid searches

aiwalter:
#### Reference Issues/PRs
#1136 

ForecastingGridSearchCV to optionally return also best n forecaster.

Exemplary output:
```
> gscv.n_best_forecasters_

[('1', NaiveForecaster()),
 ('2', NaiveForecaster(strategy='drift')),
 ('3', NaiveForecaster(strategy='mean'))]
```
##### For all contributions

##### For new estimators
___


RNKuhns:
@aiwalter this will be a nice addition. I've got a minor suggestionthat I think will be easy to incorporate and will add a bit of extra functionality. 

Would it be possible to return the scoring associated with the `n` best forecasters in an easy to access manner (to avoid lookup from results)?

I'm thinking of use-case where we add weights to the `EnsembleForecaster`. Inverse error (e.g. inverse mean squared error) weights are one strategy for weighting the predictions from an ensemble of forecasters. 

One option would be:

```python
 gscv.n_best_forecasters_

{"forecasters": 
    [
        ('1', NaiveForecaster()),
        ('2', NaiveForecaster(strategy='drift')),
        ('3', NaiveForecaster(strategy='mean'))
    ],
  "scoring":[16.5, 16.8, 17]
}
````

Other options would be to add another attribute

```
gscv.n_best_scores_

[16.5, 16.8, 17]
```
We could then make an adjustment to `EnsembleForecaster` so its API and example usage was similar to (obviously the weight part would be a little different depending on which option above was used to make the `n` best scores easy to access):

```python
forecasters = gscv.n_best_forecasters_["forecasters"]
# Calculate inverse of weights so forecasteres with higher values of scoring get less weight
weights = [ 1 / score for score in gscv.n_best_forecasters["scoring"]]
# Note that we'd need to add weights to the EnsembleForecaster API
ensemble = EnsembleForecaster(forecasters, n_jobs=None, aggfunc='mean', weights=weights)
```

I'll make a PR to update `EnsembleForecaster` to accept weights.

aiwalter:
@RNKuhns interesting idea. I am thinking about what an alternative solution for an interface between `ForecastingGridSearchCV` and `EnsembleForecaster` could look like. Regarding this:

> ```python
> {"forecasters": 
>     [
>         ('1', NaiveForecaster()),
>         ('2', NaiveForecaster(strategy='drift')),
>         ('3', NaiveForecaster(strategy='mean'))
>     ],
>   "scoring":[16.5, 16.8, 17]
> }
> ```
> 

I think this interface with a `dict` is not so intuitive as the solution with just a list of tuples. I am thinking about a solution where we dont need to expose anything additionally from grid search so all this would be handled under the hood. Fot this I have two options in mind:
1)  We enhave the `EnsembleForecaster` so that it can also receive a grid search as input and pick all stuff needed to do the ensemble you proposed with the weights. Then, we would even not need to expose  ` gscv.n_best_forecasters_` actually, because we could implement all this in `EnsembleForecaster` and it could look as follows:

```python
EnsembleForecaster(forecasters=None, grid_search=ForecastingGridSearchCV(bla, bla), weights=my_weigths)
```
I think this looks however a bit ugly with having a param `grid_search` bcz normally we just have one param `forecaster` or `forecasters`. So thats why I came up this second solution:

2)  We could create a new ensemble class that inherits from `EnsembleForecaster`, it would be sth like `GridSearchEnsembleForecaster` or `AutoEnsembleForecaster`.
```python
AutoEnsembleForecaster(forecaster=ForecastingGridSearchCV(myForecaster, myCV), n_forecaster=3, weights=None)
```
where `forecaster` has to be a grid search. We could also have a param for giving information of doing inverse error weights, not sure how to best handle this.
Then this class would have the logic under the hood to take all required forecasters and their error from `gscv.results_`.  Not sure is this is overhead to create a separate class for that? We could implement more other sophisticated ensemble stuff also in this class later maybe. Ideally this class would find best `n_forecaster` and best `weights` automatically. For this I have some stuff with `MultiplexForecaster` in mind where we could find the best ensemble automatically, but this would be also computationally expensive probably.

Whats your opinion @RNKuhns

RNKuhns:
@aiwalter totally agree on the dicts not being super intuitive. I've also been looking into the OnlineEnsemblers but I don't have a whole lot of familiarity with them. They definately are related to some of what I think we are both after but I also think not exactly it. 

I see a couple use-cases that I've found myself in the past: 

1. Routinely make forecasts from a couple different models and predictions and historical true values are stored. So it is possible to calculate how accurately their forecasts have been. Interested in a (potentially weighted) ensemble of those models going forward, but want to compare whether using weights based on past performance is better than uniform weights moving forward.
2. Pick a model to use going forward and as part of the evaluation, want to compare individual models with a (potentially weighted) ensemble of those models

In case 1, a straight-forward update to `ForecastingGridSearchCV` to expose the `n_best_forecasters_` seems fine and it would be easy to add weights a resulting ensemble based on someone's historical (or apriori) weights or the results of `ForecastingGridSearchCV` if we add a `weights` hyper-parameter to `EnsembleForecaster` and an additional `n_best_scores_` attribute to `ForecastingGridSearchCV`.

In case 2, we probably need something that evaluates the individual models and also optionally their ensemble (maybe by supplying an ensemble strategy like weights or one of those hedging algorithms. Either way, it seems like it would be most efficient to tune over the models and their ensembles at the same time). I think that might be more of an extension of the `MultiplexForecaster` you've worked on  than adding the CV functionality built-into `EnsembleForecaster`. I'm thinking something like `MultiplexForecaster` taking multiple models and then optionally an ensemble strategy (good starting point and eventually we could even generalize to multiple ensemble strategies, like compare to EnsembleForecaster and StackingForecaster with a couple different stacking regressors). That seems like it would be pretty neat functionality to me. 

So based on this, I'd err on the straight-forward side with this addition and add the functionality you propose here with the addition of the additional `n_best_scores_` attribute added to the loop. Then I can make the PR on the `EnsembleForecaster` and if you think my proposal about "case 2" and `MultiplexForecaster` sounds interesting we can think about how to best accomplish that together.

What are your thoughts?

fkiraly (approved):
Yes, why not, if you think it's useful.

I would recommend to write a test though for this, i.e., that after fitting the content of `return_n_best_forecasters` is in the expected format.

aiwalter:
hm there is an issue however regarding testing I think, bcz we cant just use `temporal_train_test_split` to test if the forecasters are in correct order bcz the `ForecastingGridSearchCV` ueses cross-validation and the error on test data can be different - and so the order. So I am not sure how to best test this. Bcz the test should probably always have a different logic than the implementation itself, otherwise a test brings no benefit 🤔

fkiraly:
@aiwalter, I just wanted to merge, but then I read your last post. Not sure whether I understand this - what is the issue you are referring to? Under which conditions does it surface? Does it require action in this PR?

aiwalter:
> what is the issue you are referring to? Under which conditions does it surface? Does it require action in this PR?

@fkiraly I talked about unit tests. I think this PR is fine now, I also added `n_best_scores_` as this might be handy to calculate custom weights based on the error. If its ok for, we can merge this?

RNKuhns (commented):
@aiwalter things look generally good to me. Only a very minor question on the code. Is there a reason your converting the rank (in your (rank, forecaster) tuples) into a string?

Saw there are some docstring issues, but those should be easy to cleanup (looks like they are easy things like missing periods at end of lines or some spacing things). Also can you make sure the docstring headings (e.g. Parameters, Examples, Notes, Returns, etc) have the `s` on the end. 

If they don't match what shows up in [numpydoc sections](https://numpydoc.readthedocs.io/en/latest/format.html#sections), then when we generate the Sphinx docs we will get lots of warnings.

aiwalter:
@RNKuhns thx for review! Sure, let me fix the docstring test also.

The reason for having `name` as a `str` is that in `EnsembleForeacster` we require the tuples `(name, forecaster)` to have a `str` name. So its a convention I think, not sure if it would break sth without `str` but I would follow this convention.
https://github.com/alan-turing-institute/sktime/blob/2f111fa755545c75f832d6f3c42e854502095b75/sktime/forecasting/compose/_ensemble.py#L24-L25

and also here:
https://github.com/alan-turing-institute/sktime/blob/2f111fa755545c75f832d6f3c42e854502095b75/sktime/forecasting/compose/_pipeline.py#L268-L269

sktime/forecasting/model_selection/_tune.py
```diff
@@ -363,9 +389,14 @@ class ForecastingGridSearchCV(BaseGridSearch):
         Time (seconds) to refit the best forecaster
     scorer_ : function
         Function used to score model
+    n_best_forecasters_: list of tuples ("rank", <forecaster>)
+        The "rank" is in relation to best_forecaster_
+    n_best_scores_: list of float
+        The scores of n_best_forecasters_ sorted from best to worst
+        score of forecasters
 
     Example
```
RNKuhns, line 398: Need to rename to Examples (and add the extra _ on the next line), per NumPy docs convention. This doesn't fail `pydocstyle` checks (it is looking for formatting primarily), but will generate warnings in the Sphinx doc build, which makes it tough to see what's going on there.

sktime/forecasting/compose/_multiplexer.py
```diff
@@ -6,12 +6,13 @@
 from sktime.forecasting.base._base import DEFAULT_ALPHA
 from sklearn.base import clone
 
-__author__ = ["Kutay Koralturk"]
+__author__ = ["Kutay Koralturk", "Martin Walter"]
 __all__ = ["MultiplexForecaster"]
 
 
 class MultiplexForecaster(_HeterogenousEnsembleForecaster):
```
RNKuhns, line 13: Can you fix the Example section header in the class doc? It should be Examples with correct number of "_" underneath.

sktime/forecasting/model_selection/_tune.py
RNKuhns: Can you verify that any other occurrence of "Examples" section in this code file is also adjusted (if applicable)?

RNKuhns (commented):
Super minor changes to the formatting of the docstring section headers to show "Examples" not "Example"

sktime/forecasting/compose/_multiplexer.py
aiwalter: sure, I am wondering why `pydocstyle` did not catch it

aiwalter:
@RNKuhns @fkiraly @mloning pls approve or review if possible. I would like to merge this soon :)

RNKuhns (approved):
@aiwalter this looks good to me; thanks for the docstring updates! Sorry for taking so long to review.

mloning (approved):
Looks all good to me!

PR #1151, opened 2021-07-17
Add annotation ext template

mloning:
#### Reference Issues/PRs
#1021 

* adds an extension template for series annotators


mloning:
@fkiraly what do you think about moving some of the text at the top of the extension templates into the `extension_templates/README.md` file? It seems to apply to all templates. I mean this bit: 

> Purpose of this implementation template:
>      quick implementation of new estimators following the template
>      NOT a concrete class to import! This is NOT a base class or concrete class!
>      This is to be used as a "fill-in" coding template.
> 
>  How to use this implementation template to implement a new estimator:
>  - make a copy of the template in a suitable location, give it a descriptive name.
>  - work through all the "todo" comments below
>  - fill in code for mandatory methods, and optionally for optional methods
>  - you can add more private methods, but do not override BaseEstimator's private methods
>      an easy way to be safe is to prefix your methods with "_custom"
>  - change docstrings for functions and the file
>  - ensure interface compatibility by testing forecasting/tests/test_all_forecasters
>          and forecasting/tests/test_sktime_forecasters
>  - once complete: use as a local library, or contribute to sktime via PR

fkiraly:
@mloning, I would advocate *copying* rather than moving, and this removing from the extensiont template.

I prefer the idea to have "all information in one place", i.e., you pick the template and you concentrate only on one source of information.

mloning:
Okay, I just thought it would avoid duplication of information and we wouldn't have to make changes in multiple places

fkiraly:
>Okay, I just thought it would avoid duplication of information and we wouldn't have to make changes in multiple places, so have the copy-pasteable templates + documentation on how to use them

I think in terms of user experience design, the "one focal point" principle beats the "avoid duplication" principle here, but this trade-off is also not a no brainer.

extension_templates/annotation.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+class MySeriesAnnotator(BaseSeriesAnnotator):
+    """Custom series annotator.
+
+    todo: write docstring, describing your custom forecaster
+
+    Parameters
+    ----------
+    fmt : str {"dense", "sparse"}, optional (default="dense")
```
fkiraly, line 48: I'd make `"sparse"` the default - e.g., annotating *only* the outliers

fkiraly:
otherwise looks good to me

fkiraly (approved):
looks good, no comments except the default setting (see above), but also no too strong opinion about that.

mloning:
The current default in the code is "dense" but wouldn't mind changing that. I'll merge this now as a pure contribution to the docs without code changes.

PR #1153, opened 2021-07-17
Add JavaScript and HTML to filter estimator overview

astrojuanlu:
Hello from SciPy 2021 sprints! 👋🏽 Continuation of #1138, gets the filtering working both locally and on Read the Docs 🎉  Thanks @mloning for the guidance!

### DIFF
--- a/docs/source/_static/js/dynamic_table.js
+++ b/docs/source/_static/js/dynamic_table.js
@@ -1,0 +1,40 @@
+$(document).ready(function () {
+  $("#myInput").on("keyup", function () {
+	var value = $(this).val().toLowerCase();
+	$("tr").filter(function () {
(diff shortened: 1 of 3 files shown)


mloning:
Thanks @astrojuanlu! @afzal442 looks like we have found a way how to include the searchable version on readthedocs!

mloning (approved):
It works on readthedocs as intended: https://sktime--1153.org.readthedocs.build/en/1153/estimator_overview.html

afzal442:
Thanks Juan Luis. Amazing to see that. 
I was really worried about why that external page was missing. So my intent was some way to make it searchable dynamically on top of the table. And you made it here. A big shout out to you and Marcus for this setup.

PR #1157, opened 2021-07-17
[wip] test refactor

### LINKED ISSUE #1097: tests refactor: scitype abstraction and dispatch
Reviewing the test suite, I am noticing that the addition of every base class requires addition of another "if/else" statement in a number of places, e.g., `estimator_check._make_args`.

There is dispatching on two things:

* base classes or scientific types
* class methods

This should be refactored to be more extensible, possibly by abstracting around scientific types.

What I would do is implement dispatch mechanisms, based on scitype. Potentially using the registry to implement checks "is this of scitype xyz" where dispatch is on the shorthand string rather than an `isinstance` call.

As a poor man's dispatch mechanism, we can use the dict tuple lookup from, see, e.g., #1061.

D3f0:
#### Reference Issues/PRs

Fixes #1097

Move from if/else to dictionary.

No

sktime/utils/_testing/estimator_checks.py

#### Any other comments?
Need guidance on type based parametrization.

##### For all contributions

##### For new estimators

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1090,6 +1090,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "D3f0",
+      "name": "Nahuel Defossé",
+      "avatar_url": "https://github.com/D3f0.png",
+      "profile": "https://www.linkedin.com/in/nahueldefosse/",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "projectName": "sktime",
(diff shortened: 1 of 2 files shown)


fkiraly:
Nice contribution, thanks!

Unfortuantely, the tests seem to fail, not sure what causes this - seems like something in the logic is broken, e.g., a reference related to one of the types. Could you have another look at this, @D3f0?

mloning:
I'll close this for now, but it's linked in #1505 so that we can reuse code from here.

Issue #958, opened 2021-06-13
forecasting metrics (non-aggregate) should be able to return loss samples


fkiraly:
Update: @RNKuhns has offboarded.

We´re also currently building probabilistic metrics, and we should ensure that the point prediction metrics are interface compatible (e.g., loss samples). I will look into this and take over @RNKuhns´ PRs.
---

For our future roadmap to include advanced benchmarking and testing capabilities, I think forecasting metrics will need to be able to return metrics of losses (in case the metric is some average).

For example, the commonly used Diebold-Mariano test to compare predictions losses of two predictions vs ground truth requires losses by forecast time point, not just an average.

This is perhaps easier to attach to the class version of the metrics (as a method) than to the function version?

Would be interested to hear your thoughts, @RNKuhns, @mloning, @aiwalter.

As a side point, not all metrics will have this, but they could return jackknife pseudo-samples by default, which would more easily give access to algorithms like jackknife and jackknife-plus estimates.

RNKuhns:
@fkiraly this is something that @mloning and I talked about with the forecasting metrics interface. An option we considered was to have a function like `absolute_error(..., aggregation_type='mean')` instead of `mean_absolute_error(...)`.

The `aggregation_type` kwarg (or whatever we would have called it) could take values like: mean, median, max, etc and return either the metric aggregated as specificied (e.g. mean absolute error) or potentially un-aggregated (if users supplied `aggreagtation_type="samples"`. 

I believe that @mloning and I were worried about this type of interface being an issue for end users.

If we wanted to just go down the road that @fkiraly suggests with a method in the metric classes to return the metric applied to each sample (without aggregation) it probably wouldn't be much of a lift and we could avoid changing the user interface for the metric functions. 

The way I coded things for several of the metric groups (e.g. percentage error metrics), I already have a non-public function to apply the metric to each sample. The mean metric function uses this and then does the aggregation For example, I use `_percentage_error(...)` in multiple functions to calculate the non-aggregate part of the percentage error metrics). 

It would be easy to use this for any of the percentage error metrics to add a `samples` method (or whatever we want to name it):

```python

class MeanAbsolutePercentageError(_PercentageMetricFunctionWrapper):
    ...
    def samples(y_true, y_pred, ...)
        return _percentage_error(y_true, y_pred, ...)
```

[...]

PR #1163, opened 2021-07-17
[ENH] Add functionality to evaluate forecasting performance metric on samples

RNKuhns:
#### Reference Issues/PRs

This is a draft of the functionality discussed in #958. 

This adds the ability for forecasting performance_metric classes to be able to evaluate the underlying function (without aggregation) on error samples (e.g. each `y_true - y_pred`). 

Signature of the method is `score_samples(y_true, y_pred, **kwargs)`. This implements an interface that builds on the `__call__` signature of the metrics (`y_train` or `y_pred_benchmark` are passed as kwargs) and the discussion in #958. Note that I'm open to having a different name than `score_samples`. I also built on the forecasting module refactor and have the primary logic in `_score_samples`. 

When implementing I considered the the ability to leverage the same code as much as possible when calculating the metrics on a sample or the aggregate metrics. To do this I did a mild refactor of `_functions.py` to make the functionality that is needed when evaluating samples and the aggregate metrics portable (e.g. the sample based code is mostly in non-public functions used by the aggregate functions and `_score_samples`). 

Note that I've set it up so that the BaseForecastingMetric will evaluate the metric functions on each sample using np.vectorize. This will only be used when creating a metric via `make_forecasting_metric`. This is not efficient, but it allows the user to only pass the aggregate metric function instead of having to write and pass two functions to `make_forecasting_metric` (i.e.: aggregate metric function and a sample function). I sided with making it easier on users relative to performance in that case. 

For all the built-in metrics we offer out of the box, they are using individualized implementations that are more efficient. 

 
No new dependencies.

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/performance_metrics/base/_base.py
+++ b/sktime/performance_metrics/base/_base.py
@@ -20,7 +20,7 @@ class BaseMetric(BaseObject):
         function's name is used.
     """
 
-    def __init__(self, func, sample_func=None, name=None):
+    def __init__(self, func, name=None):
         self._func = func
         self.name = name if name is not None else func.__name__
 
--- a/sktime/performance_metrics/base/_base.py
+++ b/sktime/performance_metrics/base/_base.py
@@ -11,13 +11,19 @@ from sktime.base import BaseObject
 class BaseMetric(BaseObject):
     """Base class for defining metrics in sktime.
 
-    Extends sktime BaseObject.
+    Parameters
+    ----------
+    func : function
+        Function that implements the aggregate metric.
+    name : str, default=None
+        Optional name to pass to metric's name attribute. If None, then the
+        function's name is used.
     """
 
-    def __init__(self, func, name=None):
+    def __init__(self, func, sample_func=None, name=None):
         self._func = func
         self.name = name if name is not None else func.__name__
 
     def __call__(self, y_true, y_pred, **kwargs):
         """Calculate metric value using underlying metric function."""
-        NotImplementedError("abstract method")
+        raise NotImplementedError("abstract method")
--- a/sktime/performance_metrics/forecasting/_classes.py
+++ b/sktime/performance_metrics/forecasting/_classes.py
@@ -232,7 +232,9 @@ class _PercentageErrorMixin:
 
     def _score_samples(self, y_true, y_pred, multioutput, **kwargs):
         """Apply metric's underlying loss function to each sample."""
-        scored_samples = _absolute_percentage_error(y_true, y_pred, self.symmetric)
+        scored_samples = _absolute_percentage_error(
+            y_true, y_pred, symmetric=self.symmetric
+        )
         scored_samples = np.average(scored_samples, weights=multioutput)
         return scored_samples
 
--- a/sktime/performance_metrics/forecasting/_classes.py
+++ b/sktime/performance_metrics/forecasting/_classes.py
@@ -1,4 +1,8 @@
 # -*- coding: utf-8 -*-
+import numpy as np
+import pandas as pd
+from sklearn.metrics._regression import _check_reg_targets
+
 from sktime.performance_metrics.base import BaseMetric
 from sktime.performance_metrics.forecasting._functions import (
     relative_loss,
@@ -19,6 +23,15 @@ from sktime.performance_metrics.forecasting._functions import (
     median_relative_absolute_error,
     geometric_mean_relative_absolute_error,
     geometric_mean_relative_squared_error,
+    _absolute_percentage_error,
+    _squared_percentage_error,
+    _squared_error,
+    _absolute_error,
+    _asymmetric_error,
+    _relative_error,
+    _get_kwarg,
+    _check_y_train,
+    EPS,
 )
 
 __author__ = ["Markus Löning", "Tomasz Chodakowski", "Ryan Kuhns"]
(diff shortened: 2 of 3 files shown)


RNKuhns:
@fkiraly I'm hoping to get your initial take on my decisions related to the API interface and implementation. 

Once that all looks good, I can add unit tests, and extend the documentation.

fkiraly:
Great idea and good work, @RNKuhns!

I've looked at it a few days ago and needed some time to think, sorry for the time this took me.

Generelly, I like the idea very much.
The one thing that I don't like is that we pass the functions in the class constructors, because not all metrics will have a "by-sample-function".

Here is what I would suggest instead:

* change the `BaseMetric` so that only parameters, but not logic is passed in the constructor
* have two main interface points `score` and `score_samples`
* I would suggest also introducing `_score` etc to put checks in `score`, to prepare for multivariate forecasts.
* introduce a "metric factory" that can take `func` and `name` if you like, which creates classes inheriting from `BaseMetric`
* the shorthand functions call the class, rather than the other way round
* introduce a tag `capability:score_samples` that indicates whether the metric can `score_samples`.

Here's how I think the methods should work:
* `__call__` calls `score`, it's just a shorthand; has the same signature as now
* `score` takes care of input checks/conversions, calls `_evaluate`; has the same signature as `__call__` currently
* `_score` contains all the logic, has the same signature as `evaluate`
* `score_samples` has the same signature as yours, and returns individual samples, as currently. In case the metric is not an average, it returns bootstrap pseudo-samples. It calls `_score_by_sample`.
* `_score_by_sample` has the same signature as `score_samples`, and contains all the logic.
* only the `BaseMetric` implements `score` and `score_samples`, for the generic plumbing
* individual metrics implement `_score` and `_score_samples`, but not `score` and `score_samples`.

What do you think? FYI @mloning?

RNKuhns:
@fkiraly a few quick thoughts below, then more specific comments.  

- score_func was not supposed to be in the constructor  of `BaseMetric` -- I had removed its usage but it was not deleted there. I've deleted that now, so that's not an issue.

- There is a downside to having the `score` and `score_by_samples` only defined at the `BaseMetric` level and it is that the documentation becomes less specific and possibly confusing. All forecasting metrics accept `Y_True` and `Y_Pred`. Some accept `Y_Train` and some accept `Y_Pred_Benchmark`. Including the documentation of `Y_Train` and `Y_Pred_Benchmark` in `score_by_samples` and `score` may be confusing since they aren't always needed to be passed. In addition, the docstring for metrics that require `Y_Train` or `Y_Pred_Benchmark` would still suggest they are optional b/c it is the same docstring used by all methods. Worse would be omitting `Y_Train` or `Y_Pred_Benchmark`(since then the docstring for metrics that use them wouldn't note they are needed). So that the documentation is useful to end users, I think at a minimum we'd need to have these common methods redefined in classes that the metrics that use these arguments would inherit from. This way we'd not have the need to define in individual metrics, but also have clear documentation on what args are needed in what metrics.

- Looking at `BaseMetric` it will be hard to use the same `BaseMetric` forecasting and other use cases like clustering or classification. Because even if the call signatures are the same, the docstrings will need to be different (`y_true` may exists in forecasting and classification but in forecasting the number of rows corresponds to forecasting horizon, but in classification is is number of observations/instances). Likewise, clustering metrics may not have `y_true` at all. Should we not use `BaseMetric` and instead focsus on `BaseForecastingMetric` and later `BaseClusteringMetric` and `BaseClassificationMetric`? 

My thoughts on the other specific suggestions are below:
> - change the BaseMetric so that only parameters, but not logic is passed in the constructor - in particular no func or score_func
> - introduce a "metric factory" that can take func and name if you like, which creates classes inheriting from BaseMetric

Per above, `score_func` is not passed. Changing logic so `func` is not passed would be a refactor of all the existing classes. I can do that, but first I think we should consider whether to have `BaseMetric` at all and just have `BaseForecastingMetric`. In terms of the "metric factory", pending the proposed refactor is this just a factory class (that inherits from `BaseMetric` or `BaseForecastingMetric`) that accepts `func` and name in constructor and then uses `func` in `_score`?

> - have two main interface points score and score_samples
> - I would suggest also introducing _score etc to put checks in score, to prepare for multivariate forecasts.

Adding `score` is straight-forward. But do we want to have two methods for returning the aggregate score (`__call__ `and `score`)? Is there a case where someone would want to use `score` instead of the callability of the class? 

> - the shorthand functions call the class, rather than the other way round

Do you mean have the functions in `_functions.py` call the classes as opposed to the other way around? I think in general we need not have a corresponding function in `_functions.py` for every class. Having those was mainly to follow how `scikit-learn` provided both. This would involve a relative major moving of code between the two files. I think spending times on the documentation and user guide would be more useful to end users, but we could consider this later. 

> - introduce a tag capability:score_samples that indicates whether the metric can score_samples.

I'm down for that. Beyond just using the tag in the classes, do I need to register it anywhere?

mloning:
Hi @RNKuhns and @fkiraly

* I like the idea of having `score` and `_score` to take care of the input validation, but I'm not sure if we need to add the `score` method for this and cannot just rely on `__call__`  (avoiding having two methods for the same thing)
* I agree that the docs are important here, is there perhaps a simple way to override the docs without overriding the method? 
* I prefer to keep functions and have classes call them internally - that's how most major packages seem to implement it and defining metrics fundamentally as functions seems intuitive to me, we then add classes to better handle additional use cases like tuning 
* I agree that refactoring isn't a priority now compared to docs
* Yes, adding tags sounds like a good idea

RNKuhns:
@mloning and @fkiraly I've been looking into the ability to update the docstrings without overriding the method. I haven't found anything that has worked yet. Any ideas on your end?

RNKuhns:
@fkiraly and @mloning I can wrap this up once #1472 is merged. What's the consensus on the `score` method?

fkiraly:
@RNKuhns has disengaged with this.

Supserseded by PR https://github.com/alan-turing-institute/sktime/pull/2502 and the design within it.

PR #1164, opened 2021-07-18
add examples in docstrings in classification

ltoniazzi:
#### Reference Issues/PRs

Fixes partially  #932, #1147.

#### What does this implement/fix? Explain your changes.

Added examples in docstrings in `classification/dictionary_based`.

#### What should a reviewer concentrate their feedback on?

For some examples the `pytest` was a bit slow on my machine (took 54 seconds to test `_weasel.py`). Is that ok?

#### Any other comments?

Let me known if the current PR is ok before merging. Then I can add more examples for the other classifiers by next Saturday.

##### For all contributions
- [x] I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/main/.all-contributorsrc).
- [ ] Optionally, I've updated sktime's [CODEOWNERS](https://github.com/alan-turing-institute/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
- [ ] I've added unit tests and made sure they pass locally.

##### For new estimators
- [ ] I've added the estimator to the online documentation.
- [ ] I've updated the existing example notebooks or provided a new one to showcase how my estimator works.

Issue #1148, opened 2021-07-17
[DOC] Completing estimator class docstrings


fkiraly:
Every estimator class should have a complete docstring.
This should be worked on one-by-one, and feel free to complete only individual rubrics if it's unclear what to fill in for the others.

A good estimator docstring should include rubrics:
* one-liner description (top), start capitalized, end with `.`
* description paragraph - what is the algorithm?
* `Components` block - only if there are estimator components. The list of components should be identical with constructor arguments that are estimators (inheriting from `BaseClassifier`, `BaseForecaster`, etc).
* `Parameters` block - individual parameters listed with `param_name: type, explanation`, explanation should include value/structure convention if expectation is more specific than just stating the type, e.g., `n: int, integer between 0 and 42`.
The list of `Parameters` should be identical with constructor arguments that are not estimators.
* `Attributes` block - these are the most important attributes of object instances which are not parameters or components. It should include attributes that correspond to the "fitted model".
* `Notes` - details, formulae, academic references
* `Example` - self-contained example on `sktime` internal toy data that runs

For formatting, we use the [numpy style](https://numpydoc.readthedocs.io/en/latest/format.html), though note that the rubrics are slightly different (because we are dealing with algorithms/estimators).
Also look at the [extension templates](https://github.com/alan-turing-institute/sktime/tree/main/extension_templates) for the algorithm scitype for a "fill-in template" that algorithm implementers are using (or should be using).

Here's an example of a good class docstring:

```python
class BOSSEnsemble(BaseClassifier):
    """Ensemble of bag of Symbolic Fourier Approximation Symbols (BOSS).

    Implementation of BOSS Ensemble from Schäfer (2015). [1]_

    Overview: Input "n" series of length "m" and BOSS performs a grid search over
    a set of parameter values, evaluating each with a LOOCV. It then retains
    all ensemble members within 92% of the best by default for use in the ensmeble.
    There are three primary parameters:
        - alpha: alphabet size
        - w: window length
        - l: word length.

    For any combination, a single BOSS slides a window length "w" along the
    series. The w length window is shortened to an "l" length word through
    taking a Fourier transform and keeping the first l/2 complex coefficients.
    These "l" coefficients are then discretized into alpha possible values,
    to form a word length "l". A histogram of words for each
    series is formed and stored.

    Fit involves finding "n" histograms.

    Predict uses 1 nearest neighbor with a bespoke BOSS distance function.

    Parameters
    ----------
[...]

PR #1166, opened 2021-07-18
Refine the Docstrings for BOSS Classifiers

RNKuhns:
#### Reference Issues/PRs

This is a portion of #1148 and related to #1164 

Refines the docstrings for the BOSS classifiers to make them consistently match our preferred format and render better  in our API reference via Sphinx. 

None

Whether the docstrings conform to our preferred format (See #1148) and importantly whether the information in the docstrings are correct. 

Also we may prefer the classifier docstring examples from #1164. May want to take a quick look there and see if that looks reasonable and we can merge this after that (assuming that looks good and is a quick merge). 

#### Any other comments?

I've got a few placeholders that are requesting information to finish populating the docstring from the reviewers. 

##### For all contributions

##### For new estimators

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1091,6 +1091,15 @@
         "code"
       ]
     }
+    {
+      "login": "ltoniazzi",
+      "name": "Lorenzo Toniazzi",
+      "avatar_url": "https://avatars.githubusercontent.com/u/61414566",
+      "profile": "https://github.com/ltoniazzi",
+      "contributions": [
+        "code"
+      ]
+    }
   ],
   "projectName": "sktime",
   "projectOwner": "alan-turing-institute",
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -28,6 +28,7 @@
         "question",
         "review",
         "talk",
+        "test",
         "tutorial"
       ]
     },
--- a/sktime/base/__init__.py
+++ b/sktime/base/__init__.py
@@ -2,11 +2,14 @@
 # -*- coding: utf-8 -*-
 # copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
 
-__author__ = ["Markus Löning"]
+"""Base classes for defining estimators and other objects in sktime."""
+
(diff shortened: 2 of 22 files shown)


RNKuhns:
@TonyBagnall or @MatthewMiddlehurst can you take a look at this and make sure the info in the docstrings is factually correct? 

Also, you'll notice in `IndividualBOSS` that I've put in a placeholders in a few spots where I need some information to complete the docstrings. If you can post a comment with that, I'll add an additional commit to include it (we clearly want to get rid of the placeholder info before merging).

TonyBagnall (dismissed):
look fine to me, fair play for your attention to detail!

TonyBagnall:
norm : bool, default = False
whether to normalise the series prior to word extraction
    alphabet_size : default = 4
number of possible values for any letter in a word
    save_words : bool, default = True
not sure actually, will let matt answer that!

MatthewMiddlehurst (commented):
I like this, I noticed some of the pages were not some great when googling them (i.e. https://www.sktime.org/en/latest/api_reference/modules/auto_generated/sktime.classification.dictionary_based.BOSSEnsemble.html), so it would be good to see if this makes an improvement so it can be applied elsewhere.

I have left a few comments for the missing parts, your examples will conflict with a previous PR into main, but it should not be that difficult to resolve. The examples seem to be the reason for the test failures also?

RNKuhns:
@MatthewMiddlehurst can you take one more quick look. I've incorporated your comments and the examples from main. 

I've noticed (under details in the docs/readthedocs.org:sktime CI check section above) that the See Also section isn't rendering exactly how I expected. In the performance_metrics API reference that is showing up with links to the other metrics listed there, but that isn't working here. 

Other than any comments from you, I'll figure out how to fix the issue in See Also ahead of merging this. 

Then we can start to roll this format out to all the classifiers.

MatthewMiddlehurst (approved):

TonyBagnall (approved):
looks good, thanks for doing this

Issue #1167, opened 2021-07-19
Derive ForecastingHorizon from DatetimeIndex in predict method


Flix6x:
**Is your feature request related to a problem? Please describe.**
Basically just a simplification request: converting from DatetimeIndex to ForecastingHorizon seems to me like a step that the `BaseForecaster` should be able to handle by itself. So instead of:

```
model = NaiveForecaster()
model.fit(train_data)
fh = ForecastingHorizon(test_data.index, is_relative=False)
forecast_data = model.predict(fh)
```

I would suggest users should be able to do:

```
model = NaiveForecaster()
model.fit(train_data)
forecast_data = model.predict(test_data.index)
```
which would be very elegant imo. It also removes the need to import `ForecastingHorizon`, which shortens user code and—and here's my selfish reason—possibly avoids adding sktime as an explicit dependency in libraries that want to be sktime compatible but do not want to depend on it.

**Describe the solution you'd like**
Perhaps the `BaseForecaster.predict` method can check whether its `fh` argument is a DatetimeIndex instance and run the necessary conversion itself?

fkiraly:
Hm, could you kindly specify an example explicitly, with how it fails currently and what you want to happen instead?

Personally I think it should just forecast whatever index you pass, as long as it is compatible with the index in `fit`.

Currently, I think index inputs are being automatically wrapped by `ForecastingHorizon`, so if you are getting problems, it would mean it can't deal with `DatetimeIndex`?

>possibly avoids adding sktime as an explicit dependency in libraries that want to be sktime compatible but do not want to depend on it.

Yes, I think this is in line with `sktime` philosophy - extensible means it should be easy to be compatible without coupling, e.g., by using the extension templates and specifications.

Also, we're trying to move the "heavy" dependencies out of `sktime` in companion packages on mid-term, in case `prophet` and its friends worry you here...

Flix6x:
Happy to oblige. Explicit example (from a different issue I worked on) below.

This one runs fine:

```
import numpy as np
import pandas as pd
from sktime.forecasting.base import ForecastingHorizon
from sktime.forecasting.naive import NaiveForecaster

data = pd.Series(
    index=pd.date_range(
        "2021-06-01 00:00", periods=12, freq="1h", closed="left"
    ),
    data=[1, 2, 3, 1, 2, 3, np.nan, 2, 3, 1, 2, 3]
)
train_data = data[:7]
test_data = data[7:]

print(train_data)
[...]

PR #1169, opened 2021-07-19
ForecastingHorizon is_relative detection on construction from index type

fkiraly:
This PR makes minor changes to `ForecastingHorizon` construction behaviour, to address the user frustration highlighted by @Flix6x in #1167. It also fixes all the docstrings.

Previously, regarding `ForecastingHorizon` construction behaviour:
* the default `is_relative` in construction of `ForecastingHorizon` was `True`, which caused an error to be thrown when constructing a `ForecastingHorizon` with supported absolute index types that were not also relative types, like `pd.DatetimeIndex`.
* when an index of such a type was passed to the `fh` argument of `fit` or `predict`, the same error would be thrown, since all indices were converted to `ForeceastingHorizon` with the default setting `is_relative=True`.

The new construction behaviour is as follows:
* the constructor allows the `is_relative` argument to be `None` - in this case the value of the `is_relative` argument is inferred: if the index is a supported relative index, it's `is_relative=True`. Otherwise, if it's a supported absolute index, it's `is_relative=False`.
* This covers the case of `pd.DatetimeIndex`, which now can be passed to `fit` and `predict` without the estimator complaining.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1091,6 +1091,15 @@
         "code"
       ]
     }
+    {
+      "login": "ltoniazzi",
+      "name": "Lorenzo Toniazzi",
+      "avatar_url": "https://avatars.githubusercontent.com/u/61414566",
+      "profile": "https://github.com/ltoniazzi",
+      "contributions": [
+        "code"
+      ]
+    }
   ],
   "projectName": "sktime",
   "projectOwner": "alan-turing-institute",
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -28,6 +28,7 @@
         "question",
         "review",
         "talk",
+        "test",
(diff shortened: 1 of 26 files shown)


mloning (dismissed):
Makes sense to me. Note that some index types can be either relative or absolute, depending on interpretation (e.g. an integer index). Currently, integer indices would be treated as relative due to the ordering of the if-conditions which makes sense to me. Should we still clarify that in the docstring?

Otherwise good to be merged!

fkiraly:
>Should we still clarify that in the docstring?

I think the docstring already explains the if/else condition precisely, no? Look at `Parameters` -> `is_relative`

mloning:
Oh yes missed the "if not relative" - all good to merge!

PR #1170, opened 2021-07-19
STSF test fix

### LINKED ISSUE #1028: [BUG] sporadic bug in time series forest? manylinux builds sporadically failing
The `manylinux` builds are sporadically failing with a `numpy` casting error (dimension 1 vs 2) in `SupervisedTimeSeriesForest`.

I suspect this is a sporadic bug due to randomization and incomplete seed-setting in the classifier, e.g., a partition becomes size 1 or 0 or similar.

I haven't made a screenshot of the error because the build restarted, but we should watch out for it appearing again.

MatthewMiddlehurst:
Fixes #1028, no exception after 10,000 runs of the test.

Found the cause of the exception (rare missing classes due to bagging) and fixed it. Cant find any issues with the random number generation after a quick scan.


TonyBagnall (approved):
Continual rebuild!

TonyBagnall:
I'm forcing this in, it has passed the tests, its a localised bug fix involving just two files and it has been hanging around for a while

PR #1179, opened 2021-07-20
[bugfix] `set_tags` call in `BaseObject.clone_tags` used incorrect signature

fkiraly:
This is a bugfix to `BaseObject.clone_tags`.

`set_tags` was called incorrectly from `clone_tags` without the kwargs symbol.

### DIFF
--- a/sktime/base/_base.py
+++ b/sktime/base/_base.py
@@ -190,7 +190,7 @@ class BaseObject(_BaseEstimator):
 
         update_dict = {key: tags_est[key] for key in tag_names}
 
-        self.set_tags(update_dict)
+        self.set_tags(**update_dict)
 
         return self
 


RNKuhns:
@fkiraly the checks are still running, but the update looks pretty clear cut for unpacking the dictionary correctly. I'll approve once it is done.

Minor suggestion (and probably not really worth the CI running again!), but when `if tag_names is None` is True it seems like you could just do `update_dict = tags_est` and in the "else" portion, you could probably just go straight to the dictionary comprehension instead of looping to filter the `tag_names` then looping the ones that survived the filtering.

 
```python
        if tag_names is None:
            update_dict = tags_est # tag_names = tags_est.keys()
        else:
            # if tag_set is passed, intersect keys with tags in estimator
            if not isinstance(tag_names, list):
                tag_names = [tag_names]
            update_dict = {key: tags_est[key] for key in tag_names if key in tags_est.keys()}
            # tag_names = [key for key in tag_names if key in tags_est.keys()]

        # update_dict = {key: tags_est[key] for key in tag_names}
```

RNKuhns (approved):

Issue #1063, opened 2021-06-27
[ENH] time series classification: shorthands for common sktime pipelines including TsfreshClassifier


fkiraly:
One of our most requested/used time series classifiers is the pipeline of `tsfresh` feature extraction and then an `sklearn` classifier. 
- [x] There should be a shorthand `TsfreshClassifier` for that.

We should create a module with simple  "feature extraction based" TSC strategies like this, e.g., `feature_extr_based`.

Other shorthands that would be great to have:
- [ ]  `TabularizerClassifier`, using the tabluarizer; a bonus feature would be using its potential "good first issue" extension to be created, the binner #242
- [x] `MatrixProfileClassifier` using the matrix profile
- [x] `SignatureClassifier` - using signature - this already exists in @jambo6's work, so just needs to be moved into the same "TSC type" sub-folder
- [ ] `SevenNumberClassifier` - using the seven-number-summary of the series (quartiles, mean, variance). A nice feature would be the ability to specify which of these to use, or more "sample summaries" like kurtosis or other percentiles; this should not use any sequential features, only features where the order does not matter.

All of these will have parameters: `estimator` - a `sklearn` classifier; and any parameters that come from the feature extractor, without additional nesting level (that you would get from `Pipeline`).

TonyBagnall:
Sure, although this is a little completist, the reason for having RocketClassifier etc is that there are results published with a specific set up which we want to reproduce. TSfresh classifier is not at all competitive according to Markus's experiments. MPClassifier is kind of ok (according to my own paper!) and I look forward to testing out the SignatureClassifier in the near future. We call the SevenNumberClassifier the summary stats classifier, but that is not informative enough. If there is not a published version to copy, what would we default the classifier to? Random Forest with 500 trees I guess, although CAWPE would be good from our perspective.

TonyBagnall:
oh, and a version of the SevenNumberClassifier is the first default classifier I ever tried on TSC in 2012 :)

fkiraly:
>If there is not a published version to copy, what would we default the classifier to? Random Forest with 500 trees I guess, although CAWPE would be good from our perspective.

Yes, I'd either: use `sklearn` `RandomForestClassifier` with default settings as the `estimator` default, or set no default - the user has to choose a classifier.

> TSfresh classifier is not at all competitive according to Markus's experiments.

I'd invoke here "relevance justifies inclusion" rather than "performance justifies inclusion".
[...]

PR #1180, opened 2021-07-20
Transform classifiers

MatthewMiddlehurst:
Implements some of the classifiers from #1063.

- TSFreshClassifier
- MatrixProfileClassifier

Refactors SignatureClassifier and Catch22ForestClassifier into a new feature_based package alongside the two new pipeline classifiers.

Quite a lot of documentation changes that were required due to the new tests also.


MatthewMiddlehurst:
doc-quality test seems to be failing because it cant check the old location of some refactored files?

fkiraly:
Hm, shouldn't @jambo6 be automatically be added as Pr reviewer when it is about signature methods? Why did that not happen?

MatthewMiddlehurst:
This also happened with #1084 where Patrick was not notified as far as I am aware, I also could not manually add him as a reviewer.

Do people have to be added to the repository to be notified and leave reviews maybe?

mloning (commented):
Hi @MatthewMiddlehurst, for the deprecations, we haven't really agreed on the details in terms of how long we keep them in and so on, but for a simple example how to raise a deprecation warning see: https://github.com/alan-turing-institute/sktime/blob/d296230155d871fe66a973cab942e20f63173516/sktime/forecasting/compose/_reduce.py#L764

TonyBagnall (commented):
I dont think you should default to random forest for Matrix Profile, since MP is a series to series conversion, I think default to rocket, or 1-NN as I did in the paper.  I might call it pipeline classifiers rather than feature_based? But then you could argue STC goes in there, The best name I have thought of so fat is unsupervised transforms, but its not very informative.

MatthewMiddlehurst:
The deprecated decorator only works for functions, so I just put it on the \_\_init__ for both classes. I'm sure that's the intended usage.

Changed the MatrixProfileClassifier base estimator. The matrix profile I used is in the panel transformation package, through now that i look there is a second matrix profile in series transformations. Not sure what's up with that.

TonyBagnall (approved):
this is a refactoring to reduce proliferation of subpackages prior to a new release of hive-cote

MatthewMiddlehurst:
The remaining doc-quality test failures are from deprecated or deleted files. This should be fine to go in if there are no changes or objections.

Issue #1171, opened 2021-07-20
[BUG] load_japanese_vowels() returns the same dataset for training and testing


off99555:
**Describe the bug**
`load_japanese_vowels` function returns the same dataset when `split="train"` and `split="test"`.
This causes the test accuracy to be 1.0 in the [Rocket multivariate demo](https://www.sktime.org/en/latest/examples/rocket.html#3.5-Classify-the-Test-Data)

**To Reproduce**

Just run this code and if there's no assertion error then my point is valid:

```python
from sktime.datasets.base import load_japanese_vowels  # multivariate dataset

X_train, y_train = load_japanese_vowels(split="train", return_X_y=True)
X_test, y_test = load_japanese_vowels(split="test", return_X_y=True)
for i in range(len(X_train)):
    for j in range(X_train.shape[1]):
        assert np.all(X_train.iloc[i,j] == X_test.iloc[i,j])
```

**Expected behavior**
Assertion fails because at least one of the series must not be equal. In fact, all the series must not be equal! So it needs to fail as early as in the first iteration.

**Versions**
<details>

System:
    python: 3.8.5 (default, Sep  3 2020, 21:29:08) [MSC v.1916 64 bit (AMD64)]
executable: C:\Users\off99\anaconda3\python.exe
   machine: Windows-10-10.0.19041-SP0

Python dependencies:
          pip: 20.2.4
   setuptools: 52.0.0.post20210125
      sklearn: 0.24.1
       sktime: 0.7.0
  statsmodels: 0.12.2
        numpy: 1.19.5
        scipy: 1.6.2
       Cython: 0.29.23
       pandas: 1.2.4
   matplotlib: 3.3.4
       joblib: 1.0.1
        numba: 0.51.2
     pmdarima: None
      tsfresh: None

</details>

mloning:
Thanks @off99555 - it looks like the loading functionality is correct but the train and test files [here](https://github.com/alan-turing-institute/sktime/tree/main/sktime/datasets/data/JapaneseVowels) are the same. Would you like to open a PR based on the data on timeseriesclassification.com?

off99555:
@mloning Yes. Done that. Please check associated PR.

TonyBagnall:
thanks for that, we will need to update the unit tests for other classifiers before we can merge your pull request @MatthewMiddlehurst could you take a look please?

PR #1182, opened 2021-07-21
Fix test set for Japanese Vowels

off99555:
#### Reference Issues/PRs
#1171 

Fix test set, previously test set was the same as training set.

No

Check that the test set is now different than training set.


TonyBagnall:
Japanese vowels is used in the tests so we will need to update them. @MatthewMiddlehurst I think its your tests failing, could you update them for the correct Japanese vowel data?

MatthewMiddlehurst:
I'm pretty sure `test_muse.py` is the only test using the dataset. Seems to be crashing now that the datasets are different, not an issue with changing accuracy.

On my branch changing the dataset to BasicMotions also causes it to crash.

`dictionary_based_classification.ipynb` also uses it, so may want to rerun that as well.

TonyBagnall:
hi, so to fix this I suggest either work out the correct output for test_muse.py and update that file in this PR, or if you prefer remove test_muse.py on this PR and we can put it back for you once your PR is in.

MatthewMiddlehurst:
Alternatively change the dataset to `load_basic_motions`, you will have to lower the indices size to i.e. `indices = np.random.RandomState(0).permutation(20)` though.

JapaneseVowels is an unequal length dataset according to the test set (but not the train, the headers now mismatch a bit fyi), can WEASEL/MUSE handle that?

TonyBagnall:
That's a good idea, also need to flag this issue of tags on the data repo

MatthewMiddlehurst:
The train set on the website has a similar header to the test set from this PR, the current train set in sktime is a bit off though. Not really sure why they are different.

MatthewMiddlehurst:
current sktime header:
header1

new header:
header2

ok yeah this is causing an issue

fkiraly:
It seemed like this was close to complete, except some minor code formatting issues - are you still working on this @off99555?

off99555:
@fkiraly No, someone please commit to this PR. I only thought that just uploading the dataset is enough. If I have to modify anything in the file, probably someone else who understand the process should do it.

fkiraly:
@MatthewMiddlehurst, do you want to take this over then? I think the file has been fixed?

MatthewMiddlehurst:
@fkiraly Me or Tony will sort this at some point, turns out theres a bit more than just replacing the duplicate file. Files from the TSC website and those in the repo are a bit different and the swap causes a test to break.

MatthewMiddlehurst:
Both dataset files have been replaced with the unequal length versions from timeseriesclassification.com in #1278.

PR #1082, opened 2021-06-30
Add ColumnEnsemble Forecaster

GuzalBulatova:
#### Reference Issues/PRs
References #1081. See also #923 and #854.

#### What does this implement/fix? Explain your changes.
Implements ColumnEnsebleForecaster.

#### Does your contribution introduce a new dependency? If yes, which one?
No.

##### For all contributions
- [x] I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/main/.all-contributorsrc).
- [x] Optionally, I've updated sktime's [CODEOWNERS](https://github.com/alan-turing-institute/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
- [x] I've added unit tests and made sure they pass locally.

PR #1190, opened 2021-07-23
Extend aggregation functionality in EnsembleForecaster

### LINKED ISSUE #1162: Add weights hyperparemeter to EnsembleForecaster API
EnsembleForecaster does not currently accept weights. With #1136  and #1139 being worked on it would be nice to optionally allow users to pass weights to `EnsembleForecaster`. 

This would give users an out-of-the-box way to use ainverse error based weighting strategy in forecasting ensembles. Users could also supply other types of weights (information criterion, etc).  

An update to the `EnsembleForecaster` API to allow weights to be passed (number weights should equal number of forecasters passed). 

This would require altering the `__init__` to store the weights, adding a check (in `_fit`?), and then updating `_predict` to use versions of the `aggfunc` that allow weights (for examples of this see use of _weighted_percentile, _weighted_geometric_mean, and np.average in performance_metrics\_functions.py).

N/A

**Additional context**
N/A

GuzalBulatova:
#### Reference Issues/PRs
Fixes #1162.

- Makes aggregation functionality more modular
- Allows passing weights as hyperparameter to aggregation functions.

No

#### PR checklist


sktime/forecasting/compose/_ensemble.py
```diff
@@ -108,20 +109,43 @@ def _predict(self, fh, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
...
-
-        if self.aggfunc == "median":
-            return y_pred.median(axis=1)
-        elif self.aggfunc == "min":
-            return y_pred.min(axis=1)
-        elif self.aggfunc == "max":
-            return y_pred.max(axis=1)
-        else:
-            return y_pred.mean(axis=1)
+    def _check_aggfunc(self):
```
mloning, line 119: I'd make this a stand-alone function rather than a class method

sktime/forecasting/compose/_ensemble.py
```diff
@@ -108,20 +109,43 @@ def _predict(self, fh, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
...
+            "average": np.average,
+            "min": np.min,
+            "max": np.max,
+        }
+        if self.aggfunc not in valid_aggfuncs.keys():
+            raise ValueError("Aggregation function %s not recognized." % self.aggfunc)
+        return valid_aggfuncs[self.aggfunc]
+
+
+def _aggregate(y, aggfunc, X=None):
```
mloning, line 132:
```suggestion
def _aggregate(y, aggfunc):
```
unless there are aggregations that depend on X?

sktime/forecasting/compose/_ensemble.py
```diff
@@ -108,20 +109,43 @@ def _predict(self, fh, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
...
+        Aggregation function used for transformation.
+    X : pd.DataFrame, optional (default=None)
+        Exogenous data used in transformation.
+
+    Returns
+    -------
+    column_ensemble: pd.Series
+        Transformed univariate series.
+    """
+    column_ensemble = y.apply(func=aggfunc, axis=1)
```
mloning, line 149:
```suggestion
    y_agg = y.apply(func=aggfunc, axis=1)
```
ensemble usually refers to the algorithm/model rather than aggregated data

sktime/forecasting/compose/_ensemble.py
```diff
@@ -108,20 +109,43 @@ def _predict(self, fh, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
...
+
+    return pd.Series(column_ensemble, index=y.index)
```
mloning, line 151: this function now seems to do two things, it aggregates the data and it sets the index - I'd set the index outside of the function in the `_predict` function

sktime/forecasting/compose/_ensemble.py
mloning: calling `apply` may also be slower than calling the functions directly

mloning (commented):
Hi @GuzalBulatova, I had a quick look, just a few minor comments, otherwise looks really good already!

RNKuhns:
@GuzalBulatova in terms of the weighting parameter I think we can add weighting to all the existing aggregation functions. See the use of `from sklearn.utils.stats import _weighted_percentile` in `sktime\sktime\performance_metrics\forecasting\_functions.py` (for example in one of the functions with median in the name). We can definitely use this for weighted medians, but possibly weighted minimums and maximums (would that just be the 0th and 100th percentiles?). 

I also think if we wanted we could add the geometric mean to this aggregation function list (see `_weighted_geometric_mean` in same file and the regular geometric mean function in `scipy.stats`). 

I also think it would be useful to add "trimmed means" to the aggregation function list. We could accomplish this by writing a helper function that first uses [scipy.stats.trimboth](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.trimboth.html#scipy.stats.trimboth) and then uses np.average when there are weights and [scipy.stats.trim_mean](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.trim_mean.html) when there aren't. Note that this would require adding an additional hyper-paremeter to control the trimming that would only be used when trimmed mean was selected. 

What are your thoughts?

GuzalBulatova:
Thank you very much @mloning I'll make suggested changes in the next commit!

@RNKuhns , thanks, I agree with you, I'll try to do this iteratively. I'll add weights to all the functions and also add weighted geometric mean first. 👍

GuzalBulatova:
@RNKuhns , @aiwalter , @mloning I need your advice on making the _aggregate function more readable. Is there a more concise way to write it?

RNKuhns:
@GuzalBulatova I've got a couple suggestions. 

First, good news and scipy's gmean function accepts weights -- I believe it didn't when I was using it in performance metrics at first. So we can use that instead of my non-public function. 

I also think we can remove the error message about weights and min/max. The 0th percentile of a list of numbers should be the minimum and the 100th percentile should be the maximum. So we should be able to call `_weighted_percentile` accordingly. 

1. Remove "average" as an aggregation choice. I think we can still call that mean.
2. Determine what to do if `weights is None` versus when `weights is not None`
    - First option: default to equal weights ( 1/ # of forecasters) and use weighted agg funcs always (except min/max)
    - Second option: check if weights is None and if so, use non-weighted aggfuncs

I think we go with the 2nd route personally. This would involve a minor refactoring of your code to adjust `check_aggfunc` and `_aggregate`, and also creating some quick wrappers around the `_weighted_percentile` that calculates our percentiles of interest (e.g. 0 for min, 100 for max, 50 for median) and also has the axis paramter, so we can use the same call in `_aggregate` for all the functions. 

Note that `_weighted_median`, `_weighted_min`, `_weighted_max` are my proposed names for those wrapper functions. 


```python
def _check_aggfunc(aggfunc, weighted=False):
    _weighted = 'weighted' if weighted == True else 'unweighted'
    valid_aggfuncs = {
        "mean": {'unweighted':np.mean, 'weighted':np.average},
        "median": {'unweighted':np.median, 'weighted':_weighted_median},
        "min": {'unweighted':np.min, 'weighted':_weighted_min},
        "max": {'unweighted':np.max, 'weighted':_weighted_max},
        "geomean": {'unweighted': scipy.stats.gmean, 'weighted':_scipy.stats.gmean},
    }
    if aggfunc not in valid_aggfuncs.keys():
        raise ValueError("Aggregation function %s not recognized." % aggfunc)
    return valid_aggfuncs[aggfunc][_weighted]

def _aggregate(y, aggfunc, weights):
    """Apply aggregation function by row.
    Parameters
    ----------
    y : pd.DataFrame
        Multivariate series to transform.
    aggfunc : str
        Aggregation function used for transformation.
    Returns
    -------
    y_agg: pd.Series
        Transformed univariate series.
    """
    if weights is None:
        aggfunc = _check_aggfunc(aggfunc, weighted=False)
        y_agg = aggfunc(y, axis=1)
    else:
        aggfunc = _check_aggfunc(aggfunc, weighted=True)
        y_agg = aggfunc(y, axis=1, weights=weights)

    return pd.Series(y_agg, index=y.index)
```

RNKuhns:
@GuzalBulatova two tweaks to the non-public wrappers on the weighted functions and a quick tweak to the tests of the function and I think this is good to go. 

For each of the non-public wrappers around the weighted percentile (e.g. _weighted_median(y, weights)), we probably want the call signature to be the same as the parameters we are using in the `NumPy`, `SciPy` functions that accept weights. 

Something like:

```python
def _weighted_median(y, axis=1, weights=weights):
    ...
```

Other tweak to the functions is instead of of using the `NumPy` version of the function to reduce dimension before applying `_weighted_percentile()`, I think you want to depending on the supplied axis use `pandas` functionality to either apply these wrapper functions to each row or each column. 

You'll need to either use something like `pandas.apply`, which lets you specify the axis (so you can pass your function's axis parameter directly to it) or use `pandas.iterrows/pandas.itertuples` or `pandas.iteritems` depending on what is passed to the axis parameter. 

On the testing part, I think we should make sure there are a pair of test cases of the weighted functionality. 1 that takes a weighted median. Another that takes a weighted median.

RNKuhns:
@GuzalBulatova noticed that my above comment on testing listed weighted median twice. To clarify, I actually meant a test case that does weighted mean, one that does the weighted median, and one that tests weighted geometric mean. 

That should confirm the underlying weighting functionality is all working as expected. Let me know if you want any help on setting up test cases (we'll need to setup an example where we know what the answer will be).

RNKuhns:
@GuzalBulatova I've got a minor change for you. Your module docstring is just the license information. 

Per [NumPy doc style](https://numpydoc.readthedocs.io/en/latest/format.html#documenting-modules), licensing info does not belong in module docstring. Can you switch that back to a comment and include a short blurb as the module docstring. 

Something to the effect of module implementing ensemble forecasters that create, optionally weighted, combinations of the predictions from underlying forecasts.

Also any idea what is up with the build-and-test errors? Is that maybe related to the change to where "_update" is located? Doesn't jump out to me why the errors would be generated in the aggregation code.

mloning (dismissed):
This looks great to me! Very clean and readable code. Happy for this to be merged. 

The code-quality errors should go away once we merge #1082 if I understood this correctly @GuzalBulatova?

GuzalBulatova:
@mloning thank you for the review! Yes, they should be, still I'd like to run the tests again after merging #1082 to main.

aiwalter (dismissed):

aiwalter (dismissed):
@mloning pls approve again :)

mloning (approved):

mloning (approved):

aiwalter (approved):

Issue #1192, opened 2021-07-23
Resolve remaining bug-interactions of multivariate module with update/predict and pred_int interfaces


fkiraly:
Testing with concrete multivariate forecasters highlighted some interactions of the multivariate interface with update/predict and the prediction intervals interface:

- [ ] `test_predict_pred_interval` - needs to be turned off if forecaster produces no predction intervals. There should probably be a tag for this rather than turning them off manually for individual estimators. Addressed in PR #1195 which introduces the tag, and tests for either error raised, or correct return of pred intervals, depending on tag value.
- [ ] `test_update_predict_predicted index` - in `update_predict`, formatting of the multi-cutoff thing is not implemented for data frame returns (or numpy) and will break in that case
- [ ] `check_update_predict_predicted_index` does not just check index, but also checks that data frames have 2 columns or more (which is off-topic), so this needs to be removed and possibly replaced by output tests. #1193
- [ ] `_format_moving_cutoff_predictions` breaks if given a list of data frames, which happens for a multivariate forecaster #1213

This was found by testing @GuzalBulatova's new multivariate column forecaster with the new framework.

fkiraly:
No.2 is related to #984 - there are some methods in the base class which are imo unnecessary

Temporary workaround is to turn off prediction intervals using the new tag in #1195 for multivariate forecasters.

GuzalBulatova:
This is closed by #1193, #1195 and #1213, correct?

fkiraly:
Yes, I think so, thanks for following up on this.

PR #1193, opened 2021-07-23
removed format check from index test

fkiraly:
This addresses problem no.3 in #1192 - removing a data frame format check from the test on "indices are equal" which is off-topic in the test, and causes all combined uni/multivariate forecasters (with `scitype:y = both`) to fail.


fkiraly:
@GuzalBulatova, would suggest you try merge into your PR (or a new branch based on your PR) and test if this resolves class no.3 of test failures.

thayeylolu (commented):
This looks good to me

GuzalBulatova (approved):
Looks all good to me, thanks Franz!

mloning (approved):

PR #1195, opened 2021-07-23
prediction interval NotImplementedError moved to BaseForecaster

fkiraly:
This addresses problem no.1 in #1192, by:

* moving `NotImplementedError` raised for prediction intervals to `BaseForecaster`, dependent on a new boolean capability tag `"capability:pred_int"`
* adding the tag with value `True` to all forecasters which have the capability to return prediction intervals; added it in `clone_tags` of tuning
* updating the test `test_predict_pred_interval` so it checks both for whether prediction intervals are returned, or `NotImplementedError` is raised, explicitly dependent on `"capability:pred_int"`


fkiraly:
@TonyBagnall, FYI, pinged you for an example of how I think "capability tags" would be used and implemented, in ref to our discussion in #981.

TonyBagnall:
it seems reasonable, but I still think we will end up checking in both fit() and _fit() unless the data characteristics are stored. So lets say we have missing values, and a classifier can handle this internally. in fit() we check_X, requires a pass through the whole data set to check, but its all fine so it passes. We then call _fit(). How does _fit know that the data has or doesnt have missing values? We could store somewhere that it contains missing values as a result of the call in _fit, but that seems a bit tangled and brittle imo.  

So really my point is about the interaction between  _fit(), fit(), checkX and using the information from checkX. It would all be solved if the data were tagged ....

fkiraly:
@GuzalBulatova, can you please check whether this resolves the problem you raised?

GuzalBulatova (dismissed):
Looks all good, thanks! Now `test_predict_pred_interval` doesn't fail for a forecaster that doesn't produce prediction intervals.

sktime/forecasting/base/_base.py
```diff
@@ -198,9 +199,15 @@ def predict(self, fh=None, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
         y_pred_int : pd.DataFrame - only if return_pred_int=True
             Prediction intervals
         """
+        # handle inputs
         self.check_is_fitted()
         self._set_fh(fh)
 
+        if return_pred_int and not self.get_tag("capability:pred_int"):
+            raise NotImplementedError(
+                "estimator cannot handle pred. intervals, set return_pred_int=False"
```
mloning, line 208:
```suggestion
                f"Prediction intervals for {self.__class__.__name__} are not yet implemented. Please set return_pred_int=False"
```

sktime/forecasting/croston.py
```diff
@@ -113,7 +115,7 @@ def _predict(
         forecast : pd.series
                    predicted forecasts
         """
-        if return_pred_int or X is not None:
+        if return_pred_int:
```
mloning, line 118: Shouldn't it be the other way around? Keeping  
```suggestion
        if X is not None:
```

mloning (commented):
Hi @fkiraly, two minor comments, everything else looks good to me.

sktime/forecasting/croston.py
fkiraly: actually, the entire "if X is not none" stuff must go.
If a forecaster ignores X, that's fine, but it shouldn't raise an error.

fkiraly:
I changed your error message to just say "not implemented" instead of "not yet implemented".
There seems to be a misunderstanding somewhere that all forecasters will eventually produce prediction intervals, and all forecasters will eventually do something useful with exogeneous variables.
I don't think this is a sensible goal. Some algorithms simply just don't do that, and adding the above would make it a completely different algorithm.
This also has implications regarding the logic, and the meaning of the tags, so might be worth keeping in mind?

mloning:
Oh yes sure that makes sense

sktime/forecasting/base/_base.py
```diff
@@ -198,9 +199,16 @@ def predict(self, fh=None, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
         y_pred_int : pd.DataFrame - only if return_pred_int=True
             Prediction intervals
         """
+        # handle inputs
         self.check_is_fitted()
         self._set_fh(fh)
 
+        if return_pred_int and not self.get_tag("capability:pred_int"):
+            raise NotImplementedError(
+                f"{self.__class__.__name__} does not have the capability to return "
```
mloning, line 208: This may be misleading too now, since some estimators have the capability in principle but it's simply not implemented yet. We may also want to default to bootstrapping or estimator agnostic way of generating prediction intervals similar to R's `forecast` package.

sktime/forecasting/tests/test_all_forecasters.py
```diff
@@ -228,15 +228,29 @@ def _check_pred_ints(pred_ints: list, y_train: pd.Series, y_pred: pd.Series, fh)
...
-    try:
+
+    if f.get_tag("capability:pred_int"):
         y_pred, pred_ints = f.predict(return_pred_int=True, alpha=alpha)
         _check_pred_ints(pred_ints, y_train, y_pred, fh)
-
-    except NotImplementedError:
-        pass
+    else:
+        pytest.raises(NotImplementedError, f.predict, return_pred_int=True, alpha=alpha)
```
mloning, line 253: Could you use the `match` argument here to make sure it's not an NotImplementedError from somewhere else?
```suggestion
        with pytest.raises(NotImplementedError, match="...")
            f.predict, return_pred_int=True, alpha=alpha
```

sktime/forecasting/base/_base.py
fkiraly: >This may be misleading too now, since some estimators have the capability in principle but it's simply not implemented yet. 

Imo, there is nothing misleading since the error message refers to the `sktime` estimator right now, and not an ideal version of it which may or may not have the capability and on which we may or may not disagree what the natural "extension" to the capability is. I.e., my error message is factual and descriptive of the is-state, whereas the reference point for yours is unclear.

I don't mind putting your error message there since it's a very minor thing and most users won't see it. Do you have a strong opinion on this?

sktime/forecasting/tests/test_all_forecasters.py
fkiraly: well, we can use the match argument only once we've agreed on the error message. Shall we agree on that one first?

fkiraly: I changed the match to "prediction intervals" since I feel this should be specific to any error message we end up with...

sktime/forecasting/base/_base.py
fkiraly: after some thinking, I've updated the message to sth that should satisfy both our requirements?

The message refers to the is-state, and encourages users to open an issue if they think the capability should be there.

How about this?

mloning: The confusion comes from conflating things which are simply not implemented and capabilities of an estimator.  Capability is not a common concept that users will be familiar with. When I hear it, I think of what an estimator can do in principle, not about what has or has not been implemented. 

But let's merge this as it is.

mloning (approved):

PR #1196, opened 2021-07-24
Registry lookup - all_estimators refactor, and new all_tags

fkiraly:
This PR consolidates registry lookup in the `registry` module:

* `all_estimators` is moved to the `registry`, with all dependencies referencing the new location
* `all_estimators` has a new argument `filter_tags`, which allows to filter by tags (incl scitypes, capabilities, etc); order of arguments has been changed so filtering is first, output format args last
* new lookup function `all_tags`, which allows to lookup tags, possibly filtered by estimator type
* both lookups have a new argument `as_dataframe`, which allows for pretty printing in jupyter notebooks, useful for the standard user archetype


TonyBagnall (approved):
this all looks fine to  me

PR #1197, opened 2021-07-24
bugfix: checking of input types in plotting

fkiraly:
This fixes an error that made its way onto main, via change of input checks:

* `plot_series` and `plot_correlations` now allow arbitrary input series via conversion
* the second invalid input example in `test_plot_series_with_unequal_index_type_raises_error` has been changed from one-column `pd.DataFrame` to a two-colum `pd.DataFrame`


fkiraly:
@mloning ?

PR #1198, opened 2021-07-24
bugfix: wilcoxon expected result changed

fkiraly:
The test `test_evaluator.test_wilcoxon` was failing, since the expected and found result swiched columns.

No changes have been made to related functionality, so perhaps it is due to a change in `stats.wilcoxon` signature?
Or, do I have a version inconsistency in my local version? (let's see what the remote tests say).

Either way, this PR fixes the expected results to the new results.

### DIFF
--- a/sktime/benchmarking/tests/test_evaluator.py
+++ b/sktime/benchmarking/tests/test_evaluator.py
@@ -187,7 +187,7 @@ def test_plots():
 def test_wilcoxon():
     evaluator, metrics_by_strategy = evaluator_setup(score_function=accuracy_score)
     results = evaluator.wilcoxon_test().iloc[:, 2:].values
-    expected = np.array([[0.5, 0], [0.5, 0], [0.5, 0]])
+    expected = np.array([[0, 0.5], [0, 0.5], [0, 0.5]])
     assert np.array_equal(results, expected)
 
 


fkiraly:
@ViktorKaz, do you know what is happening to the wilcoxon test?

fkiraly:
very very strange - locally the test runs, and the result of wilcoxon is as this fix implies.
On main, my local test breaks, but the remote test runs.

I think I´m satisfying all version requirements, so not sure what is going on.

Any help, @mloning?

ViktorKaz:
@fkiraly `test_wilcoxon` fails locally for me after I cloned your pull request. 

The returned result from `evaluator.wilcoxon_test().iloc[:, 2:].values` is:
```Python
array([[0.5, 0. ],
       [0.5, 0. ],
       [0.5, 0. ]]
```
Maybe not very helpful but based on a quick review it looks to me that there was nothing wrong with the wilcoxon method. I am not sure why the columns are switched for you. It might indeed be due to difference of the version of `scipy`. I have scipy==1.6.3 installed in my virtual environment.

fkiraly:
for really strange reasons, it seems only local.
After reinstalling `sktime` with current requirements, it's fine again, so I assume it must have been some weird interaction of unsupported package versions, and not a change in the actual.

This might still be worth investigating, but the PR is not needed.

PR #1201, opened 2021-07-24
Data types module, collating conversions, mtype inference, checks, register

fkiraly:
This PR collates data type and conversion related concerns in its own module, `datatypes`.

The functionality moved to `datatypes`:

* series conversion functionality formerly in `forecasting.base.convertIO`
* panel conversion functionality formerly in `utils.data_processing._panel`

All references to the old locations have been replaced by references to the new locations.

New features enabled by this move:

* there is just a single `convert` function, for both `Series` and `Panel` scitype
* nested, extensible module structure which is by scitype, e.g., `panel` and `series`

New functionality, partially implemented for extension:

* register of mtypes with explanation, to be complemented by docs
* central fixture generation function `get_example` which produces fixtures of the same "scientific content" but a specific mtype, for external tests, and bulk testing of conversion or mtype inference functionality


fkiraly:
I guess I win the "maximize number of codeowners who have to review" game with this one...

TonyBagnall (approved):
I've discussed this with my group. This is definitely something that needs doing, and whilst the style is not mine, it fine. 

1) I would ask you have some explanation of machine types, scitypes, even if just linked from the docstrings. 

2)I'm also not 100% sure it needs a full  root level module, rather than being in data  (scikit style) or utilities (my preference). I also am trying to think of another name, I dont like datatypes. 

3) I dont  like the mix of underscore and camel case naming conventions, 

That all said, in order to not hold up transformers, I'm fine with it going in

fkiraly:
>I would ask you have some explanation of machine types, scitypes, even if just linked from the docstrings.

Yes, I wanted to write a doc page.

>I'm also not 100% sure it needs a full root level module, rather than being in data (scikit style) or utilities (my preference). I also am trying to think of another name, I dont like datatypes.

Neither am I, I just moved different things from different internal places in one place. Once we come up witih something better, we can deprecate this. Also, "mtypes" or "sciptypes" would be too cryptic as a name...

>I dont like the mix of underscore and camel case naming conventions,

My linter didn't complain...

PR #1205, opened 2021-07-25
Update documentation backend and reduce warnings in doc creation (#1199)

mloning:
#### Reference issues/PR
Fixes #1181 
Fixes #1230 
Partly addresses #1152 

#### What does this implement/fix? Explain your changes.
* replaces `readthedocs-theme` with `pydata-sphinx-theme`
* restructures content in line with new theme
* adds new getting started page
* fixes autodoc display of summaries of classes and functions
* fixes sphinx build and warnings
* fixes docstrings and numpydoc warnings
* adds tutorial notebook thumbnail gallery
* adds links to Twitter, Discord and GitHub
* fixes wrong auto-generation directory in sphinx build
* updates version of sphinx and related packages
* integrates `myst-parser` replacing `m2r2`
* update `.gitignore` to exclude auto-generated documentation files
* adds glossary
*  removes full module path from page titles in API reference
* removes module names from table of content in API reference using more readable names (e.g. instead of "sktime.forecasting" we now simply use "Forecasting")
* adds page for adding estimators using extension templates
* adds page for enhancement proposals

#### Does your contribution introduce a new dependency? If yes, which one?
For generating docs only: 
* `pydata-sphinx-theme` (replaces `readthedocs` theme)
* `myst-parser` (replaces `m2r2`)
* `sphinx-gallery` 

#### What should a reviewer concentrate their feedback on?
Rendered website from ReadTheDocs CI/CD (click details below in CI/CD checks)

PR #1206, opened 2021-07-25
Add myst

mloning:
There are lots of commits from the previous branch, I only added the last 2 of them, you can see the changes in the files changed tab.

#### Reference Issues/PRs
Suggestions for #1205


RNKuhns (commented):
Changes to the docs look good to me. I like the way "About" looks like the other pages (landing page).

Issue #961, opened 2021-06-14
Implement generic TrendForecaster


fkiraly:
This should be a forecaster that wraps a `sklearn` regressor.

The forecaster fits a regression model to the training series `y` where the values are considered `sklearn` outputs, and indices `.index` are considered `sklearn` features/predictors. When asked for a forecast, the horizon is considered as `sklearn` feature/predictor, and the `sklearn` output are the values.

This does not seem to exist currently, there is only `PolynomialTrendForecaster`.

In terms of sensibility, one would expect this to give reasonable performance only if the regressor is capable of extrapolation, e.g., low-complexity ones such as `LinearRegressor` (but not a random forest etc).

mloning:
I think this is covered the `PolynomialTrendForecaster` which by default uses linear regression but does accept other regressors too.

fkiraly:
ah, indeed. I didn´t read the docs carefullly. Changing the title hence

mloning:
I think `PolynomialTrendForecaster` still makes sense as it allows to generate polynomial features first if desired, perhaps mostly an issue around documentation? But happy to rename it to `TrendForecaster`, may be clearer and it's shorter.

fkiraly:
hm, actually tending towards a split now. `PolynomialTrendForecaster` just takes a degree, and `TrendForecaster` just takes a regressor. It seems like there are two fundamentally different regressors in one - if you pass a regressor, you ignore the degree; if you pass a degree, you ignore the regressor.

mloning:
The current implementation doesn't ignore the degree if you pass a regressor. Perhaps we could deprecate the `PolynomialTrendForecaster` and simplify it into a `TrendForecaster` as the polynomial features can always be added by passing a pipeline as the regressor. Either way, I don't have a strong preference on this one.

fkiraly:
Hm, I see - so it´s equivalent to a pipeline of polynomial features, then applying the regressor, right?

In that case, the name seems to be misleading, since polynomial trend is a standing term that implies polynomial regression (linear in the monomials of certain degrees).

How about making a `TrendForecaster` that doesn't have the polynomial features at all, and leaving the `PolynomialTrendForecaster` as-is, as a shorthand where only in rare cases you would switch out the regressor?

fkiraly:
This should have been closed by #1209.

PR #1209, opened 2021-07-26
Adding TrendForecaster in forecasting/trend.py

tensorflow-as-tf:
#961 

It makes a new class named TrendForecaster that doesn't have the polynomial features at, while leaving the PolynomialTrendForecaster unchanged.

The contribution doesn't introduce a new dependency.

Feedback from the reviewer would be very much appreciated. They should look at code quality as well and comment on the readability of line 70 that creates a new variable ("regressor").

##### For new estimators


sktime/forecasting/trend.py
```diff
@@ -16,9 +16,103 @@
...
+            The forecast horizon with the steps ahead to predict.
+        X : pd.DataFrame, optional (default=None)
+            Exogenous variables are ignored
+
+        Returns
+        -------
+        self : returns an instance of self.
+        """
+        # for default regressor, set fit_intercept=False
+        regressor = self.regressor or LinearRegression(fit_intercept=False)
```
fkiraly, line 69: I'd make `fit_intercept=True` here.

sktime/forecasting/trend.py
```diff
@@ -16,9 +16,103 @@
...
+
+        # make pipeline with regressor
+        self.regressor_ = make_pipeline(regressor)
```
fkiraly, line 72: why do you need that? There's just one component

sktime/forecasting/trend.py
```diff
@@ -16,9 +16,103 @@
...
-    Forecast time series data with a polynomial trend.
+
+    _tags = {
+        "univariate-only": True,
+        "requires-fh-in-fit": False,
+        "handles-missing-data": False,
+    }
+
+    def __init__(self, regressor=None):
+        self.regressor = regressor
```
fkiraly, line 48: I think this should be a `clone`, i.e., use `sklearn` `clone` to make a copy to `regressor_`. No need to have two variables, `regressor` and `regressor_`

sktime/forecasting/trend.py
```diff
@@ -16,9 +16,103 @@
...
+        """
+        # for default regressor, set fit_intercept=False
+        regressor = self.regressor or LinearRegression(fit_intercept=False)
+
+        # make pipeline with regressor
+        self.regressor_ = make_pipeline(regressor)
+
+        # transform data
+        n_timepoints = _get_duration(self._y.index, coerce_to_int=True) + 1
+        X = np.arange(n_timepoints).reshape(-1, 1)
```
fkiraly, line 76: here, I'd take into account the case where time points are not equally spaced. The trend strategy still works in that case, you just have to be careful with reformatting.

fkiraly (changes requested):
Good contribution, almost ready to go!

I made some comments above for change:

* there should just be a single `regressor_` attribute
* `regressor` should be cloned before assigning in `__init__`
* `make_pipeline` seems unnecessary
* the time points should be converted so the trend strategy still works for unequally spaced time index (if you know how to do it)

sktime/forecasting/trend.py
mloning: Just to clarify, cloning happens in `_fit`, the `__init__` method just remembers the input.

sktime/forecasting/trend.py
```diff
@@ -1,24 +1,117 @@
 #!/usr/bin/env python3 -u
 # -*- coding: utf-8 -*-
-# copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+"""copyright: sktime developers, BSD-3-Clause License (see LICENSE file)."""
 
-__author__ = ["Markus Löning"]
-__all__ = ["PolynomialTrendForecaster"]
+__author__ = ["Anthony Jancso", "mloning"]
```
fkiraly, line 5: this should be your GitHub ID

fkiraly (approved):
Looks good, except that the author field should contain GitHub name.

mloning:
Great work @tensorflow-as-tf! Could we also add this to the online docs? It should go into `docs/api_reference/forecasting.rst`.

tensorflow-as-tf:
I added the function to the online docs in `docs/api_reference/forecasting.rst` in PR #1292

PR #1216, opened 2021-07-27
checks sctitype tag type for y multivariate and univariate

thayeylolu:
if the estimator has the tag scitype:y = multivariate, the test example should have 2 columns
https://github.com/alan-turing-institute/sktime/issues/1211


fkiraly:
This is great!

This shoudl address the issue, would you mind also updating the other tests in the same way?

fkiraly:
From your post on discord, you listed the tests as:

test_update_predict_single
test_predict_time_index_in_sample_full
test_no_fh_opt
test_fh_in_fit_opt
test_fh_in_predict_opt

fkiraly:
Small mistake: you also need to account for the `both` tag - I think then, the test should try both.

thayeylolu:
Thank you for noticing @fkiraly

PR #1236, opened 2021-08-01
Format setup files

TonyBagnall:
Demonstrates setup fails the tests


mloning:
If you fix the first line: `"""Install script for sktime"""` by adding a period to the end it should work.

TonyBagnall:
ah it was me, I changed a string it was not actually asked to change. May has well push this in now if its ok with you @mloning ?

mloning (approved):

PR #1237, opened 2021-08-01
[DOC] CC groups and minor governance clarifications

fkiraly:
This proposed change to the governance document contains a major proposal and minor changes.
As per the current document, this requires CC approval and is for first reading at CC during the 2021-08-02 meeting.

The major change is the introduction of "CC groups", with mechamisms to prevent imbalance between them, looking at a potential strategic roadmap where we expand the CC for broader and/or diverse representation.
The proposed groups are:

* junior career data scientists
* mid/senior academics
* mid/senior industry and public sector scientists

The changes aim at preventing power imbalance and adversarial takeover scenarios. Common political forces this tries to keep in check:

* takeover by an industry stakeholder or stakeholder group; or the "soft" version of this which is shift to a pure delivery focus
* takeover by academic special interest groups led by a few senior academics; the "soft" version of this is shift to a pure blue sky research focus
* exploitation of junior career individuals, especially junior academics in precarious positions

The main mechanism is:
* defining these groups formally
* weighting votes when they run too far out of balance, with a target balance of 1/3 each
* reducing the risk of group X candidates not supported by group X being voted in by group Y, prevented by changes to the appointment process that require group X approval explicitly

Minor, unrelated changes add default paragraphs for roles, clarifications, and moving topical content to those paragraphs:

* "end of tenure" paragraphs - defining resignation and automatic end of tenure
* "eligibility" paragraphs - who can assume role XYZ?


TonyBagnall (dismissed):

TonyBagnall:
I would like to cut down the number of open PRs. How about we have a recommendation that people should only have one open PR at a time? At least can only have one referred to CC at a time?

TonyBagnall:
I mean obvs small fixes are fine

mloning:
I like the idea but it also further complicate the governance document. Making sure it's concise and easily understandable is crucial I think. Happy to discuss this further.

TonyBagnall:
currently we have  67 open PRs
@fkiraly has 12
@mloning has 6
@thayeylolu has 6
@aiwalter has 4 
@satya-pattnaik has 4

If devs prioritised just one, we can focus on clearing this backlog. As it is I find it all a bit overwhelming, since who knows how they all interact? they

TonyBagnall:
I think rather than governance, I would like a general stated soft policy "devs are encouraged to have only one open PR at any one time"

fkiraly:
>I think rather than governance, I would like a general stated soft policy "devs are encouraged to have only one open PR at any one time"

That's a different topic, and I think indeed a policy matter rather than a terms of reference matter 😄

mloning:
Just to avoid confusion, my above comment referred to this PR rather than your idea Tony, but I guess the same principle applies. I'm not convinced that writing down more rules is a good way forward.

fkiraly:
>I like the idea but it also further complicate the governance document. Making sure it's concise and easily understandable is crucial I think. Happy to discuss this further.

Any suggestion how you would simplify this and still keep the spirit of it?

GOVERNANCE.rst
```diff
@@ -251,27 +265,91 @@ Rights and responsibilities
...
+stay open for 5 days excluding weekends. Approval of appointment requires:
 
 * a 2/3 majority of all cast votes, and
-* a simple majority approval of all the current CC members.
+* a simple majority approval of all the current CC members
+* a simple majority approval of all current CC members from the same CC group
 
-The vote will take place in private communication channels and will be
-anonymous.
+if there is at least on CC member of the same CC group.
```
GuzalBulatova, line 328: at least **one** CC member

GOVERNANCE.rst
```diff
@@ -251,27 +265,91 @@ Rights and responsibilities
...
+
+If there are no CC members from the same CC group, the approval condition changes to:
+
+* a 1/2 majority of all cast votes, and
+* a simple majority approval of all core developers who would be in the same CC group
+  if they were on the CC.
+
+The core developer vote takes place in private communication channels, visible only to core developers, and isanonymous.
```
GuzalBulatova, line 336: Maybe "The core developer vote is anonymous and takes place..."

fkiraly:
We had resolved to re-discuss this proposal once there was again a CC member from the early career group on the CC. This condition is now satisfied with @GuzalBulatova on the CC.

However, since I have made this proposal one year ago, with wisdom (or at least experience) grown, I no longer thing this is a good idea.

Why: I think what sounded like a good idea in the first instance would have severe, unintended secondary effects.

Namely, two main points:

* we want `sktime` to be a diverse and open community. The two "senior" groups, i.e., senior academics and industry/public sector data scientists with reports, are not diverse at the current time and day. Giving them a fixed 2/3 of vote weight would cement that lack of diversity in `sktime`.
* As the system currently is, senior academics are culturally incompatible with `sktime` intended governance principles. Selection in the publish-or-perish age leaves mostly academics in place who lead their research groups like a cross-over of mob boss and robber baron, and the system selects for people who are good at exploiting talent, rather than having talent. Again, these are not cultural patterns we want to import from the broken academic landscape.

In fact, we should perhaps even give a minimum vote weight to the early career persona, but that's another discussion.

Either way, I think this was a stupid idea, and I vote down my own proposal. It is good we have participative decision making, and I would like to thank all people who cautioned the CC from adopting this too prematurely.

Issue #1146, opened 2021-07-17
TSC refactor


fkiraly:
Time series classifiers should be refactored according to the [new extension template](https://github.com/alan-turing-institute/sktime/blob/main/extension_templates/classification.py).

If you would like to work on this: *please post here* and which classifier you are taking on, to avoid duplication of work.
A core dev will then add your name to the classifier under refactor and/or link the PR.

Checklist of classifiers:

- [ ] `BaseColumnEnsembleClassifier`
- [ ] `ComposableTimeSeriesForestClassifier`
- [x] `BOSSEnsemble`, `IndividualBOSS` https://github.com/alan-turing-institute/sktime/pull/1544
- [x] `ContractableBOSS` https://github.com/alan-turing-institute/sktime/pull/1295
- [x] `MUSE` https://github.com/alan-turing-institute/sktime/pull/1359
- [x] `TemporalDictionaryEnsemble`, `IndividualTDE` https://github.com/alan-turing-institute/sktime/pull/1357
- [x] `WEASEL` https://github.com/alan-turing-institute/sktime/pull/1544
- [x] `MUSE` https://github.com/alan-turing-institute/sktime/pull/1544
- [x] `ElasticEnsemble` #1584
- [x] `ProximityForest`, `ProximityStump`, `ProximityTree` #1584
- [x] `ShapeDTW` https://github.com/alan-turing-institute/sktime/pull/1554
- [ ] ~~`KNeighborsTimeSeriesClassifier`~~ (inherits from _KNeighborsClassifier and calls a version of self._fit in its current fit)
- [x] `Catch22Classifier` #1487
- [x] `MatrixProfileClassifier` https://github.com/alan-turing-institute/sktime/pull/1545
- [x] `SignatureClassifier` https://github.com/alan-turing-institute/sktime/pull/1545
- [x] `TSFreshClassifier` #1473 
- [x] `HIVECOTEV1`  #1504
- [x] `CanonicalIntervalForest` https://github.com/alan-turing-institute/sktime/pull/1269
- [x] `DrCIF` https://github.com/alan-turing-institute/sktime/pull/1269
- [ ] ~~`TimeSeriesForestClassifier`~~ (bit complicated, involves the series_as_features base class)
- [x] `RandomIntervalSpectralForest` #1583
- [x] `SupervisedTimeSeriesForest` #1583
- [x] `Arsenal` https://github.com/alan-turing-institute/sktime/pull/1305
[...]

PR #1239, opened 2021-08-01
Time series classifiers refactor/rocket classifier

victordremov:
#### Reference Issues/PRs

#1146 TSC refactor for `ROCKETEstimator`

Refactor `ROCKETEstimator` class according to the [new extension template](https://github.com/alan-turing-institute/sktime/blob/main/extension_templates/classification.py).
- fit and predict turned private
- `capabilities` class attribute renamed to `_tags` class attribute
- capabilities names prefixed with `capability:` prefix
- new tags added to ESTIMATOR_TAG_REGISTER

No

- `capability:missing_values` and `capability:multivariate` are intentionally kept distinct from `handles-missing-values` and `univariate-only` to stick to `capability:` prefix convention at least until refactoring ends. It looks like it will be easier merge them after refactoring and rename all occurences at once.
- I do not know meaning for `train_estimate` and `contractable` tags, so I need help to fill them

##### For all contributions


sktime/classification/kernel_based/_rocket_classifier.py
```diff
@@ -49,13 +47,12 @@ class ROCKETClassifier(BaseClassifier):
     tsml/classifiers/shapelet_based/ROCKETClassifier.java
     """
 
-    # Capability tags
-    capabilities = {
-        "multivariate": True,
-        "unequal_length": False,
-        "missing_values": False,
-        "train_estimate": False,
-        "contractable": False,
```
victordremov, line 58: Hello, Matthew,

Could you kindly help me to understand what `contractable` and `train_estimate` capabilities mean? I am refactoring ROCKERClassifier and I am not sure what they mean. @MatthewMiddlehurst

MatthewMiddlehurst: `contractable` The user can set a time limit for the fit process. i.e. in cBOSS instead of searching a set number of parameter sets, keep searching parameter sets until time runs out.

`train_estimate` The classifier can generate a set of estimated probabilites for the training dataset. i.e. in cBOSS again, cross-validation is used to determine how useful an ensemble member is. We can reuse this if an estimate of performance is required on the training set. See also: out-of-bag estimates for forest classifiers.

fkiraly:
@victordremov, @MatthewMiddlehurst, could you kindly use the `_tags` system?
Then questions like this wouldn't arise.

In the `_tags` syste, we have the capability to attach explanations, e.g., if you use `registry.all_tags`.
Of course, you need to deposit the plain English description in `registry._tags` first for this to work.

Would you mind doing/updating that too, @MatthewMiddlehurst, @victordremov?

fkiraly:
oh, I see you're already doing that - good work, thanks! Please ignore my comment.

sktime/classification/base.py
```diff
@@ -150,7 +150,11 @@ def predict_proba(self, X):
         -------
         y : array-like, shape =  [n_instances, n_classes] - predictive pmf
         """
-        raise NotImplementedError("abstract method")
+        coerce_to_numpy = self.get_class_tag("coerce-X-to-numpy", False)
```
fkiraly, line 153: I think you should be using `get_tag` here - `get_tag` does the same but also allows dynamic overrides.

sktime/classification/base.py
```diff
@@ -219,3 +223,19 @@ def _predict(self, X):
         y = self.label_encoder.inverse_transform(predictions)
 
         return y
+
+    def _predict_proba(self, X):
```
fkiraly, line 227: 👍

sktime/classification/kernel_based/_rocket_classifier.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 """RandOm Convolutional KErnel Transform (ROCKET)."""
 
-__author__ = "Matthew Middlehurst"
+__author__ = ["Matthew Middlehurst", "Viktor Dremov"]
```
fkiraly, line 4: the strings should be GitHub IDs.
I know it's inconsistent currently, but later we'll want to be using GitHub IDs.

sktime/registry/_tags.py
```diff
@@ -124,6 +124,42 @@
...
         "which machine type(s) is the internal _fit/_predict able to deal with?",
     ),
+    # "capability:multivariate" is the inverse of "univariate-only" tag.
+    # They are kept distinct intentionally for easier TSC refactoring.
+    # Will be merged after refactor completion.
+    (
+        "capability:multivariate",
+        "classifier",
+        "bool",
+        "can classifier use exogeneous data (X)?",
```
fkiraly, line 134: that's not exactly true - it should mean "can classifier classify time series with 2 or more variables?"

sktime/registry/_tags.py
```diff
@@ -124,6 +124,42 @@
...
+    ),
+    (
+        "capability:unequal_length",
+        "classifier",
+        "bool",
+        "can classifier handle unequal length data?",
```
fkiraly, line 140: better: "unequal length time series"

sktime/classification/kernel_based/_rocket_classifier.py
```diff
@@ -109,7 +104,7 @@ def fit(self, X, y):
         self._is_fitted = True
```
fkiraly, line 104: this should also go, it's in the base class too

sktime/classification/kernel_based/_rocket_classifier.py
```diff
@@ -49,13 +47,12 @@ class ROCKETClassifier(BaseClassifier):
...
-        "missing_values": False,
-        "train_estimate": False,
-        "contractable": False,
+    _tags = {
+        "capability:multivariate": True,
+        "capability:unequal_length": False,
+        "capability:missing_values": False,
+        "capability:train_estimate": False,
+        "capability:contractable": False,
     }
```
fkiraly, line 56: I'd also put these in the base class with the following defaults:
```python
        "capability:multivariate": False,
        "capability:unequal_length": False,
        "capability:missing_values": False,
        "capability:train_estimate": False,
        "capability:contractable": False,
```
So here you would only have to put

```python
    _tags = {
        "capability:multivariate": True
    }
```

because that's the only one that would differ from the default - but no harm having all of them in the concrete class.

fkiraly (changes requested):
Great work, @victordremov!

Change request with only very small things:

* `self._is_fitted = True` line from the class should also go
* `__author__` field should be populated with GitHub ID, the "real name" goes in the contributors file (so can be linked there via GitHub ID)
* some of the tags are not correctly described in the registry, I suggested plain English descriptions

Not a blocker if we don't do it, but possibly useful to do: putting defaults for the capability tags int he base classifier.

sktime/classification/kernel_based/_rocket_classifier.py
fkiraly: so for the descriptions, how about this:

contractable - `does the estimator support time limits on execution, aka contract time setting?`
train_estimate - `can the classifier produce out-of-bag predictions on the training set?`

Might be worth also adding an explicit mention to the method/argument if that's easy. What do you think, @MatthewMiddlehurst, on this as description strings?

sktime/classification/kernel_based/_rocket_classifier.py
victordremov: I don't really like such implicit defaults, custom tag inheritance + defaults make things difficult. When all tags are listed in one class, it is okay, I can get it from the code. But if some tags are in base class, some in subclass - the only method to get tag values is to instantiate object and inspect it in runtime

sktime/classification/kernel_based/_rocket_classifier.py
victordremov: looks like contractable is not self-explanatory enough, if nobody is against it, I would like to rename it into `capability:set-max-fit-time` after TSC refactoring completion and add TODO comment about it

sktime/classification/base.py
victordremov: done 83a9faa1

sktime/classification/kernel_based/_rocket_classifier.py
victordremov: 96c657bb

sktime/registry/_tags.py
victordremov: bdd99a4a

sktime/registry/_tags.py
victordremov: a0083bd9

sktime/classification/kernel_based/_rocket_classifier.py
victordremov: ac47ae9c

sktime/classification/kernel_based/_rocket_classifier.py
victordremov: For `contractable`, though, I think is better to emphasize that only train time can be limited, not predict, so `does the estimator support limiting max fit time?`
cbf10004

victordremov: I like `can the classifier produce out-of-bag predictions on the training set?` description for `train_estimate`
e94bec0a

victordremov:
Hello, Franz,
Thank you very much for you review! I have gone through your suggestions, please look review the code again.

Note that I did not add default _tags values for BaseEstimator, I have wrote my reasons why we should not do it in https://github.com/alan-turing-institute/sktime/pull/1239#discussion_r684405424 cc @fkiraly

TonyBagnall:
Hi, we call it contractable in all our papers and want to keep the name, sorry

sktime/classification/kernel_based/_rocket_classifier.py
fkiraly: >I don't really like such implicit defaults, custom tag inheritance + defaults make things difficult. When all tags are listed in one class, it is okay, I can get it from the code. But if some tags are in base class, some in subclass - the only method to get tag values is to instantiate object and inspect it in runtime

I get your point, and agree with it.

fkiraly: In fact, my opinion (in previous core developer discussions) was this:

* have default tags in the base class (not `BaseEstimator` but of the scitype, e.g., `BaseClassifier`), so we have "safe defaults" in case someone forgets
* also ask extenders to fill in all the tags in any concrete subclass, as a kind of "poka yoke" which makes it less likely that people forget to enter tags, and since it's easier to look at the code and see all the tags.

The consensus we ended up with though, was:

* have defaults in base classes
* do *not* ask to enter the tags which are equal to defaults

I don't like that entirely, but that's what we arrived at.

fkiraly: in any case, I wouldn't see it as a PR blocker if you fill in all the tags in the subclass.

TonyBagnall: well I was not in favour of custom tag inheritance, hence why I set up capabilities as a much simpler system. Personally I much prefer boolean tags to freeform strings. Sorry it be awkward, but these changes are going to effect every single classifier, so I would like a proper discussion of this rather than just slipping it through with a single classifier. We can refactor later if it is collectively decided to be desirable, but for now I would like a simple porting over to the _tags, and defaults in the base class as Franz suggests.  contractable can change to train-time-contractable if you want.

TonyBagnall:
> * `__author__` field should be populated with GitHub ID, the "real name" goes in the contributors file (so can be linked there via GitHub ID)

is this policy? It is not done that way in most of the code. If so I'll enforce it, but was not aware

> * some of the tags are not correctly described in the registry, I suggested plain English descriptions

I can refine this once its in if you want.

sktime/classification/kernel_based/_rocket_classifier.py
MatthewMiddlehurst: Worth noting that while out-of-bag predictions are the best example, they aren't the only method of generating train set predictions. Maybe replace out-of-bag with bespoke?

Contracting description is good.

sktime/classification/kernel_based/_rocket_classifier.py
TonyBagnall: > In fact, my opinion (in previous core developer discussions) was this:
> 
> * have default tags in the base class (not `BaseEstimator` but of the scitype, e.g., `BaseClassifier`), so we have "safe defaults" in case someone forgets
> * also ask extenders to fill in all the tags in any concrete subclass, as a kind of "poka yoke" which makes it less likely that people forget to enter tags, and since it's easier to look at the code and see all the tags.
> 
> The consensus we ended up with though, was:
> 
> * have defaults in base classes
> * do _not_ ask to enter the tags which are equal to defaults
> 
> I don't like that entirely, but that's what we arrived at.

Oh ok. Having no defaults and also not insisting on assigning values will increase the chances of run time errors. My view would be to have scitype defaults but not insist on subclasses assigning a value, I view that as bloaty.  One point of inheritance is to provide the default behaviour or explicitly specialise. But as I say, this PR does not need to decide the whole issue, I suggest we just get this in with the minimal refactoring and then consider from there.

sktime/classification/kernel_based/_rocket_classifier.py
victordremov: 1. If I understood Matthew right, `train_estimate` description should be generalized to cover other methods to estimate classifier performance. Changed it to `can the classifier estimate its performance on the training set?`. If there are better suggestions, will gladly change it. 4cbd4398
2. It looks like `contract time setting` is a synonym term for `contractable`. If so, better add it to description. `contract time setting, i.e. does the estimator support limiting max fit time?` e84da651

sktime/classification/kernel_based/_rocket_classifier.py
victordremov: I see there is still no agreement about defaults and overrides, and I agree with Tony we should keep things simple until there is some decision on this question.

So for now lets only rename capabilities to tags with `capability:` prefix, as it is very simple for contributors. If later we decide to add tags defaults to base class and enforce to remove from concrete classes tags that have default values, we can do it in automated way at any stage of refactoring or after it.

victordremov: I understand there is already an agreement to have default tag values, but it is not still clear where to keep them, in `BaseEstimator` or `BaseClassifier` and should concrete class list all tag values or only distinct from defaults. I guess when this decision is made, it should be put down in TSC template there https://github.com/alan-turing-institute/sktime/blob/84c4ff004b7e6735becc0e63f105709e2e0b3e8a/extension_templates/classification.py#L70

For now, lets just convert capabilities into tags.

victordremov:
> Hi, we call it contractable in all our papers and want to keep the name, sorry

removed `TODO: rename tag` comment e84da651

victordremov:
Hello, @fkiraly @TonyBagnall @MatthewMiddlehurst,

Fixed all the issues and waiting for your formal review.

sktime/classification/base.py
MatthewMiddlehurst: predict_proba seems to still use `get_class_tag`. Very minor but a bit strange to use one in fit/predict and another in predict_proba unless theres a reason.

fkiraly:
>is this policy? It is not done that way in most of the code. If so I'll enforce it, but was not aware

@TonyBagnall, I thought so? We really need that code style guide written, come doc sprint.

fkiraly (dismissed):
can be merged from my side, nice work!

Thanks, @victordremov!

MatthewMiddlehurst (dismissed):
From an earlier comment: `predict_proba` seems to still use `get_class_tag`. Very minor but a bit strange to use one in `fit`/`predict` and another in `predict_proba` unless theres a reason.

Other than that looks fine.

fkiraly:
yes, agreed with @MatthewMiddlehurst 

There are now some new conflicts in the contributors file due to merging other branches, so if you could resolve those, plus the `get_class_tag` in `update`, we could merge right after that.

MatthewMiddlehurst (approved):

dremovv:
> predict_proba seems to still use get_class_tag. Very minor but a bit strange to use one in fit/predict and another in predict_proba unless theres a reason.
@MatthewMiddlehurst Thank you for pointing this out. I was sure I have changed get_class_tag to get_tag in all 3 methods: fit, predict and predict_proba, but now I can see I have done only 2.

MatthewMiddlehurst:
@dremovv I pushed some commits earlier which fixed the issue and the conflict, so no need for any action. There are a couple of high priority PRs that will go in before this one it seems, but I would like it get this in soon as it will impact some of my stuff.

Thanks for contributing.

fkiraly (approved):
great, good to go!

PR #1242, opened 2021-08-02
Update feature_request.md

aiwalter:
Added `[FEATURE]` as start of title for all feature request issues. So all issues following the issue template would have either `[FEATURE], [DOC] or [BUG]` and we can easily see issues that were opened without templates and request user to use the templates. 

Alternatively we could use `[ENH]` as a abbreviation which is e.g. used by `pandas`.


mloning:
Yes let's use three-letter abbreviations, i.e. `[ENH]`, we already use `[DOC]`

mloning (approved):

TonyBagnall (approved):

PR #1250, opened 2021-08-03
Added Sktime Forecaster Based Outlier Detector

satya-pattnaik:
#### Reference Issues/PRs

Extending #1021 

As discussed with @mloning , this is an Annotator based on forecasts from Sktime Forecasters.
An Outlier Detector which detects those time points as outliers for which the values
fall outside of the Uncertainty Intervals of forecasts generated from
a Sktime Forecaster.
The Sktime Forecaster is passed as an argument.

This method is inspired from :
[alibi-detect ProphetOD](https://docs.seldon.io/projects/alibi-detect/en/latest/methods/prophet.html)

No

- Do we need a better name for the class.
- The structure of the implementation - the parameters to constructor/fit/predict?
- Any other issues you can spot if any.
#### Any other comments?

After first feedback I will add this to Unit tests.

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/annotation/adapters/_sktime_od.py
+++ b/sktime/annotation/adapters/_sktime_od.py
@@ -1,0 +1,110 @@
+# -*- coding: utf-8 -*-
+from sktime.annotation.base._base import BaseSeriesAnnotator
+import pandas as pd
+import numpy as np
+from sklearn import clone
+from sktime.forecasting.base import ForecastingHorizon
+
+__author__ = ["satya-pattnaik", "mloning"]
+
+
+class SktimeODAnnotator(BaseSeriesAnnotator):
+    """An Outlier Detector which detects those time points as outliers for which the values
+    fall outside of the Uncertainty Intervals of forecasts generated from
+    a Sktime Forecaster.
+    The Sktime Forecaster is passed as an argument.
+
+    This method is inspired from :
+    https://docs.seldon.io/projects/alibi-detect/en/latest/methods/prophet.html
+    Parameters
+    ----------
+    estimator : a Sktime Forecaster object
+    fmt : str {"dense", "sparse"}, optional (default="dense")
+        Annotation output format:
+        * If "sparse", a sub-series of labels for only the outliers in X is returned,
+        * If "dense", a series of labels for all values in X is returned.
+    labels : str {"indicator", "score"}, optional (default="indicator")
+        Annotation output labels:
+        * If "indicator", returned values are boolean, indicating whether a value is an
+        outlier,
+        * If "score", returned values are floats, giving the outlier score.
+    alpha : float, Significance level of the Uncertainty Interval to be used,
+            for detecting the Outliers,(default=0.05)
+    """
+
+    def __init__(self, estimator, fmt="dense", labels="indicator", alpha=0.05):
+        self.estimator = estimator  # Sktime forecaster
+        self.alpha = alpha
+        super(SktimeODAnnotator, self).__init__(fmt=fmt, labels=labels)
+
+    def _fit(self, X, Y=None):
+        """Fit to training data.
+
+        core logic
+
+        Parameters
+        ----------
+        X : pd.DataFrame
+            training data to fit model to, time series
+        Y : pd.Series, optional
+            ground truth annotations for training if annotator is supervised
+        Returns
+        -------
+        self : returns a reference to self
+
+        State change
+        ------------
+        creates fitted model (attributes ending in "_")
+        """
+        X_t = X.copy()
+        self.estimator_ = clone(self.estimator)
+        self.estimator_.fit(X_t)
+
+        return self
+
+    def _predict(self, X):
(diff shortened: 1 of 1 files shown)


fkiraly:
@mloning, @satya-pattnaik, interesting - quick question though: should this perhaps also (not instead) be a transformer, which computes something like the residual to a predictive percentile, then you could apply cusum or another detector on top of that?

satya-pattnaik:
> should this perhaps also (not instead) be a transformer

Yes, We can have it, I guess. Once this code refines out, I will make a similar transformer too.

sktime/annotation/adapters/_sktime_od.py
```diff
@@ -0,0 +1,110 @@
...
+from sktime.annotation.base._base import BaseSeriesAnnotator
+import pandas as pd
+import numpy as np
+from sklearn import clone
+from sktime.forecasting.base import ForecastingHorizon
+
+__author__ = ["satya-pattnaik", "mloning"]
+
+
+class SktimeODAnnotator(BaseSeriesAnnotator):
```
mloning, line 11: We need to come up with a better name I think, how about simply `OutlierDetector`? I don't have a good idea on how to capture the fact that a forecaster is used internally. Any other suggestions?

sktime/annotation/adapters/_sktime_od.py
```diff
@@ -0,0 +1,110 @@
...
+            ground truth annotations for training if annotator is supervised
+        Returns
+        -------
+        self : returns a reference to self
+
+        State change
+        ------------
+        creates fitted model (attributes ending in "_")
+        """
+        X_t = X.copy()
```
mloning, line 59: Copying can go here, that should be taken care of inside the forecaster
```suggestion
```

sktime/annotation/adapters/_sktime_od.py
```diff
@@ -0,0 +1,110 @@
...
+        Parameters
+        ----------
+        X : pd.DataFrame - data to annotate, time series
+
+        Returns
+        -------
+        Y : pd.Series - annotations for sequence X
+            exact format depends on annotation type
+        """
+        X_t = X.copy()
```
mloning, line 77: Same as above
```suggestion
```

sktime/annotation/adapters/_sktime_od.py
```diff
@@ -0,0 +1,110 @@
...
+        Returns
+        -------
+        self : returns a reference to self
+
+        State change
+        ------------
+        creates fitted model (attributes ending in "_")
+        """
+        X_t = X.copy()
+        self.estimator_ = clone(self.estimator)
```
mloning, line 60: We need to check that the forecaster can actually generate prediction intervals. I think @fkiraly added a tag for that, something like `capability:pred_int`. So we should check if `estimator.get_tag("capability:pred_int")` evaluates to True.

sktime/annotation/adapters/_sktime_od.py
```diff
@@ -0,0 +1,110 @@
...
+        )
+        uncertainty_intervals["is_outlier"] = (
+            uncertainty_intervals["score"] > 0.0
+        ).astype(int)
+        if labels == "score":
+            Y_val_np = uncertainty_intervals["score"]
+        elif labels == "indicator":
+            Y_val_np = uncertainty_intervals["is_outlier"]
+
+        if fmt == "dense":
```
mloning, line 104: We're seeing the same issue here again with the formatting. Perhaps this is something to briefly discuss in the next developer meeting.

mloning (commented):
Hi @satya-pattnaik - just a few quick comments! Looks good already!

sktime/annotation/adapters/_sktime_od.py
satya-pattnaik: Any suggestions on a better name @fkiraly ?

sktime/annotation/adapters/_sktime_od.py
satya-pattnaik: I don't see this tag in some of the forecasters' code. May be you I am looking in the wrong place.
So do I need to write something like this?

```
if estimator.get_tag("capability:pred_int"):
    Execute the code
else:
    raise NotImplementedError

```

mloning: Yes, or simply have a `_check_estimator` stand-alone function which raises the NotImplementedError. @fkiraly could you help with the tags here?

satya-pattnaik: Screenshot 2021-09-01 at 5 31 44 PM

This is what I get for ARIMA, I don't see anything for prediction(my version is ```0.7.0```). Or I am looking at the wrong thing? @mloning @fkiraly

mloning: When I'm using the latest version from `main`, this gives me the tags as expected:
```python

forecaster = ARIMA()
forecaster.get_tags()
```

PR #1255, opened 2021-08-04
Add 3.9 support for linux/osx

freddyaboulton:

Adds python 3.9 to the list of python versions to build and test against for osx and linux.

No new dependencies but numpy and numba in the build requirements have been upgraded to the minimal version that supports python 3.9. 

##### For all contributions

##### For new estimators

### DIFF
--- a/build_tools/requirements.txt
+++ b/build_tools/requirements.txt
@@ -3,7 +3,7 @@ esig==0.9.7
 fbprophet>=0.7.1
 hcrystalball>=0.1.9
 numba>=0.53
-numpy==1.19
+numpy==1.19.3
 pandas==1.1.5
 pmdarima>=1.8.0,!=1.8.1
 pyod>=0.8.0
--- a/build_tools/requirements.txt
+++ b/build_tools/requirements.txt
@@ -2,7 +2,7 @@ cython>=0.29.0
 esig==0.9.7
 fbprophet>=0.7.1
 hcrystalball>=0.1.9
-numba>=0.50,<0.53
+numba>=0.53
 numpy==1.19
 pandas==1.1.5
 pmdarima>=1.8.0,!=1.8.1
--- a/.binder/requirements.txt
+++ b/.binder/requirements.txt
@@ -4,7 +4,7 @@ fbprophet>=0.7.1
 hcrystalball>=0.1.9
 matplotlib
 notebook
-numba>=0.50,<0.53
+numba>=0.53
 numpy==1.19
 pandas==1.1.5
 pmdarima>=1.8.0,!=1.8.1
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -71,7 +71,7 @@ jobs:
     strategy:
       matrix:
         os: [macos-latest]
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.6, 3.7, 3.8, 3.9]
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -71,7 +71,7 @@ jobs:
     strategy:
       matrix:
         os: [macos-latest]
-        python-version: [3.8, 3.9]
+        python-version: [3.6, 3.7, 3.8, 3.9]
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -71,7 +71,7 @@ jobs:
     strategy:
       matrix:
         os: [macos-latest]
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.8, 3.9]
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python
--- a/.github/workflows/upload-to-pypi.yml
+++ b/.github/workflows/upload-to-pypi.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         os: [ macos-latest ]
-        python-version: [ 3.6, 3.7, 3.8 ]
+        python-version: [ 3.6, 3.7, 3.8, 3.9 ]
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -6,7 +6,7 @@
 
 variables:
   REQUIREMENTS: build_tools/requirements.txt
-  INCLUDED_VERSIONS: "cp36-cp36m, cp37-cp37m, cp38-cp38"
+  INCLUDED_VERSIONS: "cp36-cp36m, cp37-cp37m, cp38-cp38, cp39-cp39"
 
 trigger:
   branches:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -77,38 +77,38 @@ stages:
               PYTHON_ARCHITECTURE: x86
 
 
-  - stage: 'deploy'
-    dependsOn: 'build'
-    condition: and(succeeded('build'), startsWith(variables['Build.SourceBranch'], 'refs/tags/v'))
-    jobs:
-      - job: 'upload_to_pypi'
-        pool:
-          vmImage: 'ubuntu-latest'
-        steps:
-          - task: DownloadPipelineArtifact@2
-            displayName: 'Collect wheels'
-            inputs:
-              source: 'current'
-              tags: '^v[0-9]+\.[0-9]+\.[0-9]+'
-              patterns: 'wheels_*/*manylinux*.whl'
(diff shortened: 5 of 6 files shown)


build_tools/requirements.txt
```diff
@@ -2,8 +2,8 @@ cython>=0.29.0
 esig==0.9.7
 fbprophet>=0.7.1
 hcrystalball>=0.1.9
-numba>=0.50,<0.53
-numpy==1.19
+numba>=0.53
```
freddyaboulton, line 5: These are the minimal versions that support 3.9 that are needed for the build to pass. I can put an upper bound on numba if needed?

fkiraly: hm, @mloning, why did we have those weird bounds on `numba` in the first place?
I'm sure there was a good reason, but I wasn't involved in setting this up.

fkiraly: Better safe than sorry, let's check this twice.

fkiraly (dismissed):
Happy with this and thanks.

@mloning, can you look at the `numba` bounds and/or comment in general?

fkiraly:
so, now it appears there are conflicts in the requirements since the binder fix in #1265 - do you mind updating this, @freddyaboulton?

build_tools/requirements.txt
mloning: There may have been some incompatibility - I'm not exactly sure either.

mloning (commented):
Thanks @freddyaboulton - looks all good to me! Excited to get Python 3.9 support!

mloning (approved):

mloning (approved):
I resolved the merge conflicts.

fkiraly (approved):
alright, all ready to go then

freddyaboulton:
Thank you so much @mloning and @fkiraly !

fkiraly:
>Thank you so much @mloning and @fkiraly !

No, thanks to *you*, @freddyaboulton!

fkiraly:
so... it seems like there were some interactions with the recent binder fix?
Perhaps it's good if @corvusrabus could also have a quick look at this PR, just to avoid interactions.

freddyaboulton:
@fkiraly Yes, there were conflicts with the binder requirements and the build requirements that were causing pip to timeout because it could not resolve them. I think the root cause was that the binder requirements listed `prophet` but the build requirements listed `fbprophet`.

fkiraly:
hmmm, but `prophet`/`fbprophet` was precisely one of the reasons why binder was failing originally - have you checked that this change does not break the binder again?

freddyaboulton:
I built the binder examples [here](https://hub.gke2.mybinder.org/user/freddyaboulton-sktime-zvs26hwg/tree/examples) and it built successfully as far as I can tell.

@fkiraly is verifying and I'm posting here so that others can too.

fkiraly (approved):
yes, looks like it works - thanks!

Some remarks:
* Stupid question - why does it look like 3.9 takes 25-50% longer to build than the other versions?
* the `all-contributors` emoji should also include `infra` for CI/CD, no?

fkiraly:
sorry, @freddyaboulton, it seems like the manylinux builds are failing now.

I do not understand the reason for that, but it looks like it happened after I fixed the binder and `test_examples` which broke probably due to a `numba` version upgrade in #1343, which in turn was made possible by the removal of upper version bound in #1266 (FYI @corvusrabus).

I'm not sure what's going on but it's a `numba` error that's coming up, so I guess it's related...
Would be much appreciated if you could have another look.

freddyaboulton:
@fkiraly @mloning I set the numba version to 0.53 and that fixed the manylinux build errors from the new release. 

The error message is `RuntimeError: module compiled against API version 0xe but this version of numpy is 0xd`. Looking at the numpy C API version [table](https://github.com/numpy/numpy/blob/main/numpy/core/setup_common.py#L35), it looks like numba is compiled against numpy 1.20 and above but we're installing numpy 1.19.

freddyaboulton:
@fkiraly @mloning I think this is good for review again after the numba 0.54 release broke manylinux. Right now I'm limiting numba to 0.53 (supports 3.6 - 3.9) after trying two different solutions:

1. Updating the minimum numpy version to 1.19.5 and setting `numba>=0.53`
2. Using numpy 1.20 , setting `numba>=0.53` and dropping python 3.6 (because `np` 1.20 does not support 3.6).

Both approaches still cause failures unfortunately. I don't think this PR should be blocked by the numba 0.54 release. Limiting numba to 0.53 seems to be the right path forward until we figure out what's happening (could be a numba bug for all we know).

fkiraly:
so it's `numba` 0.54 causing all this?
Yes, it's broken so many things apparently, let's leave it limited for now.

What I find strange is that the second approach causes failure still - what is the error message?
Also, if it weren't causing issues, what would we do to keep supporting 3.6 while upgrading numpy?
Would we need to have two different dependency conditions, with 3.6 remaining on `numba` 0.53?

fkiraly (approved):
In any case I think this now takes care of 3.9 support! Great!

freddyaboulton:
@fkiraly Yes, it's numba that's causing this mayhem hehe.

The second approach fails because `pmdarima` has a [hard pin](https://github.com/alkaline-ml/pmdarima/blob/master/requirements.txt#L3) on `numpy` 1.19.0.

Yes if upgrading numba would mean dropping 3.6 support, we would need to install different versions of the dependencies for different python versions. Let's not spend too much energy thinking about how that would look like just yet - I have hope that this numba issue will  be resolved in a new release. I will file an issue in `numba` to get their thoughts. 

Relatedly, do you know the maintainers of `pmdarima`. I think we should talk to them about being a bit more permissive with their numpy dependency?

.appveyor.yml
```diff
@@ -65,7 +65,7 @@ install:
 
   # Install requirements from inside conda environment
   - cmd: activate testenv
-  - cmd: conda install -c conda-forge "matplotlib>=3.3.2" "seaborn>=0.11.0" "tsfresh>=0.17.0" "hcrystalball>=0.1.9" "stumpy>=1.5.1" "fbprophet>=0.7.1" "numba>=0.50,<0.53" "numpy>=1.19" "pandas>=1.1.0,<1.2" "statsmodels>=0.12.1" "scikit-learn>=0.23.0" "scikit-posthocs>=0.6.5" "pystan>=2.14,<3.0"
+  - cmd: conda install -c conda-forge "matplotlib>=3.3.2" "seaborn>=0.11.0" "tsfresh>=0.17.0" "hcrystalball>=0.1.9" "stumpy>=1.5.1" "fbprophet>=0.7.1" "numba>=0.53" "numpy>=1.19.3" "pandas>=1.1.0,<1.2" "statsmodels>=0.12.1" "scikit-learn>=0.23.0" "scikit-posthocs>=0.6.5" "pystan>=2.14,<3.0"
```
fkiraly, line 68: just to be sure before we merge: is this removal ot `numba` bounds not going to cause us trouble down the line?

setup.py
```diff
@@ -21,11 +21,11 @@
 
 MIN_PYTHON_VERSION = "3.6"
 MIN_REQUIREMENTS = {
-    "numpy": "1.19.0",
+    "numpy": "1.19.3",
     "pandas": "1.1.0",
     "scikit-learn": "0.24.0",
     "statsmodels": "0.12.1",
-    "numba": "0.50",
+    "numba": "0.53",
```
fkiraly, line 28: why is the change of `MIN_REQUIREMENTS` needed?

fkiraly:
sorry, I got a bit paranoid about `numba` - can you quickly let me know about my questions on appveyor and min-requirements?

fkiraly:
>Relatedly, do you know the maintainers of `pmdarima`? I think we should talk to them about being a bit more permissive with their numpy dependency?

We occasionally chat, albeit usually when some dependencies break and/or go out of sync...

So I'm sure they would be thankful for some help around the topic.

.appveyor.yml
freddyaboulton: Agreed - We should cap numba here too. Thank you!

setup.py
freddyaboulton: There are no pre-built 3.9 wheels for 0.50 - 0.52: https://pypi.org/project/numba/0.52.0/#files. Pip uses `install_requires` to determine which dependency versions are installed when users pip install sktime, so this notes that numba 0.53 should be installed

.appveyor.yml
freddyaboulton: Done.

setup.py
fkiraly: ok, got it - i.e., to simultaneously satsify 3.6-3.9 requirements.

.appveyor.yml
fkiraly: alright. Guess paranoia is occasionally helpful.

fkiraly:
@mloning, can you have a look before we merge? Since it affects CI/CD and dependency version requirements.

fkiraly (approved):
Still happy.

PR #1256, opened 2021-08-05
[DOC] docstring fix for distances/series extension templates

fkiraly:
I just noticed that the `_transform` docstrings were switched around between distances/kernels on tabular data and those on time series.

Fixed that, together with some other, minor things.


TonyBagnall (approved):
could you expand the names? MyTrafoPwPanel looks like a bunch of random letters to me, I cant even work out what it is short for!

PR #1259, opened 2021-08-06
Fix all_estimators when extras are missing.

xloem:
#### Reference Issues/PRs

Did not open an issue for this.

sktime.registry.all_estimators() crashes if the user does not have all the extra modules installed.
This change instead outputs warnings for them, allowing the user to continue.

No.

This is intended to be a clear, simple, quick-to-merge contribution that makes the code easier
to use.  Please let me know if that is not the case.

##### For all contributions
- n/a Optionally, I've updated sktime's [CODEOWNERS](https://github.com/alan-turing-institute/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
- n/a I've added unit tests and made sure they pass locally.

##### For new estimators
- n/a I've added the estimator to the online documentation.
- n/a I've updated the existing example notebooks or provided a new one to showcase how my estimator works.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1175,6 +1175,14 @@
       ]
     },
     {
+      "login": "xloem",
+      "name": "patiently pending world peace",
+      "profile": "https://github.com/xloem",
+      "contributions": [
+        "code"
+      ]
+    },
+    {
       "login": "AreloTanoh",
       "name": "Arelo Tanoh",
       "avatar_url": "https://avatars.githubusercontent.com/AreloTanoh",
--- a/sktime/registry/_lookup.py
+++ b/sktime/registry/_lookup.py
@@ -153,7 +153,9 @@ def all_estimators(
                 ]
                 all_estimators.extend(estimators)
             except ModuleNotFoundError as e:
+                # Skip modules with import errors, but notify the user
                 warnings.warn(str(e), ImportWarning)
+                warnings.warn("Error importing {module_name}. Try: pip install sktime[all_extras]", ImportWarning)
 
     # Drop duplicates
     all_estimators = set(all_estimators)
--- a/sktime/registry/_lookup.py
+++ b/sktime/registry/_lookup.py
@@ -153,9 +153,10 @@ def all_estimators(
                 ]
                 all_estimators.extend(estimators)
             except ModuleNotFoundError as e:
-                # Skip modules with import errors, but notify the user
+                # Skip missing soft dependencies
+                if "soft dependency" not in str(e):
+                    raise e
                 warnings.warn(str(e), ImportWarning)
-                warnings.warn("Error importing {module_name}. Try: pip install sktime[all_extras]", ImportWarning)
 
     # Drop duplicates
     all_estimators = set(all_estimators)
--- a/sktime/registry/_lookup.py
+++ b/sktime/registry/_lookup.py
@@ -134,20 +134,26 @@ def all_estimators(
     # packages
     with warnings.catch_warnings():
         warnings.simplefilter("ignore", category=FutureWarning)
+        warnings.simplefilter("module", category=ImportWarning)
         for _, module_name, _ in pkgutil.walk_packages(path=[ROOT], prefix="sktime."):
 
             # Filter modules
             if _is_ignored_module(module_name) or _is_private_module(module_name):
                 continue
 
-            module = import_module(module_name)
-            classes = inspect.getmembers(module, inspect.isclass)
+            try:
+                module = import_module(module_name)
+                classes = inspect.getmembers(module, inspect.isclass)
 
-            # Filter classes
-            estimators = [
-                (name, klass) for name, klass in classes if _is_estimator(name, klass)
-            ]
-            all_estimators.extend(estimators)
+                # Filter classes
+                estimators = [
+                    (name, klass)
+                    for name, klass in classes
+                    if _is_estimator(name, klass)
+                ]
+                all_estimators.extend(estimators)
+            except ModuleNotFoundError as e:
+                warnings.warn(str(e), ImportWarning)
 
     # Drop duplicates
     all_estimators = set(all_estimators)


fkiraly:
Good catch, excellent!

Quick questions:

* if you can't load a module you still look at the `module` variable and get classes from it. Would this not (a) crash if the first module in the list can't be imported, and (b) look at the `module` variable from the previous module it could import, otherwise? Line 149 in the new version of the code. This looks like an unintended bug to me, so you may want to move lines 149 to 155 in the try block?
* Can you think of a way to still get the classes and catch the module requirements, even if imports fail? Not sure how to do this, so not a necessary change.

fkiraly (changes requested):
(see above)

fkiraly:
>Can you think of a way to still get the classes and catch the module requirements, even if imports fail? Not sure how to do this, so not a necessary change.

Perhaps: move soft dependency imports *inside* the classes? @mloning, what do you think?

xloem:
> Good catch, excellent!
> 
> Quick questions:
> 
> * if you can't load a module you still look at the `module` variable and get classes from it. Would this not (a) crash if the first module in the list can't be imported, and (b) look at the `module` variable from the previous module it could import, otherwise? Line 149 in the new version of the code. This looks like an unintended bug to me, so you may want to move lines 149 to 155 in the try block?

Did you see the `continue` statement in the except block, line 148 of the new code?  No code is executed for imports that fail.

> * Can you think of a way to still get the classes and catch the module requirements, even if imports fail? Not sure how to do this, so not a necessary change.

- import extra modules upon construction as you suggest.  pros: flexible.  cons: takes review of multiple estimator classes
- parse files without importing them, e.g. using `ast`.  pros: small.  cons: obscures estimators from user.
- maintain a registry of estimator names.  pros: user-friendly.  cons: takes ongoing maintenance

Should changes to this file be ported upstream to sklearn?

fkiraly:
>Did you see the continue statement in the except block, line 148 of the new code? No code is executed for imports that fail.

That wasn't exactly the case - the variable `module` is still accessed when imports fail, but it contains the value from the previous loop. If there is no previous loop, this would break. No?
Anyway, I think it's fixed now.

>Can you think of a way to still get the classes and catch the module requirements, even if imports fail? Not sure how to do this, so not a necessary change.

Thanks for your thoughts!

I'd prefer option no.1, even though it may take some work to get there. It would make `sktime` less reliant on soft dependencies.

No.2 is elegant, I wasn't aware, thanks.

No.3 I'd rather not, this unnecessarily introduces maintenance burden. We've already tried to do this one, and people just didn't keep it up to date.

>Should changes to this file be ported upstream to sklearn?

I can't speak for `sklearn`. I can imagine they will probably say "no", since they have no external dependencies when it comes to estimators - all `sklearn` estimators are implemented there directly, as far as I know.

.all-contributorsrc
```diff
@@ -1143,7 +1143,15 @@
       "contributions": [
         "doc"
       ]
-    }
+    },
+    {
+      "login": "xloem",
+      "name": "patiently pending world peace",
```
fkiraly, line 1149: ... interesting name :-)
I wouldn't mind it if that's what you want

xloem: My experience is that it is unwise to share our names until things improve, but I don't want to push my views on others.  Happy to change it if you'd like.

fkiraly: no, no, all fine - being anonymous is fine, and world peace is a goal I can also identify with.

sktime/registry/_lookup.py
```diff
@@ -134,13 +134,18 @@ def _is_estimator(name, klass):
...
             # Filter modules
             if _is_ignored_module(module_name) or _is_private_module(module_name):
                 continue
 
-            module = import_module(module_name)
+            try:
+                module = import_module(module_name)
+            except ModuleNotFoundError as e:
+                warnings.warn(str(e), ImportWarning)
+                continue
```
xloem: @fkiraly it doesn't matter, but this line is where access to `module` was prevented.  The effect of a `continue` statement is to suddenly jump to the next loop iteration, skipping the rest of the loop body below.

xloem:
Thanks for your review.

> I'd prefer option no.1, even though it may take some work to get there. It would make `sktime` less reliant on soft dependencies.

Do these changes still seem good to merge, or would you like to focus on No.1 (moving the optional imports) to solve this problem?

> No.3 I'd rather not, this unnecessarily introduces maintenance burden. We've already tried to do this one, and people just didn't keep it up to date.

Noting that No.3 (a list of all estimators) could also be generated automatically on a machine or github vm that has all the extra dependencies, using the current approach.

sktime/registry/_lookup.py
fkiraly: ah, I missed that, got it!

fkiraly:
>Do these changes still seem good to merge, or would you like to focus on No.1 (moving the optional imports) to solve this problem?

I think it's good to merge since it's an improvement already, and no need for more extensive changes.
But let's also give other core developers the opportunity to review.

sktime/registry/_lookup.py
fkiraly: >@fkiraly it doesn't matter,

yes, you were right. I missed the `continue`, so this is not-an-issue

fkiraly:
>Noting that No.3 (a list of all estimators) could also be generated automatically on a machine or github vm that has all the extra dependencies, using the current approach.

Hmmmm - indeed! We could do this as part of the release cycle! Nice idea!
How would this be robustly implemented, I wonder?

fkiraly (approved):
Happy with this, nice and compact fix that solves a potentially annoying problem.

sktime/registry/_lookup.py
```diff
@@ -134,20 +134,26 @@ def _is_estimator(name, klass):
...
+                classes = inspect.getmembers(module, inspect.isclass)
+
+                # Filter classes
+                estimators = [
+                    (name, klass)
+                    for name, klass in classes
+                    if _is_estimator(name, klass)
+                ]
+                all_estimators.extend(estimators)
+            except ModuleNotFoundError as e:
```
mloning, line 155: Could we be more specific on the exception catching? Since we're importing modules here, there may be any number of reasons why we see an `ModuleNotFoundError`. I would add an additional condition based on the error message.

mloning (commented):
Thanks @xloem - this is very useful!

sktime/registry/_lookup.py
xloem: What kind of condition/reason do you imagine?  Does `ModuleNotFoundError` happen for reasons other than extra modules being nonpresent?  I always pause a little when acting on error messages because I don't know how non-english operating systems are handled by python ...

fkiraly:
@xloem, would you kindly mind fixing the merge conflict in `.all_contributors`?

xloem:
Thanks.  Rebased on main.

sktime/registry/_lookup.py
fkiraly: yes, I also don't see what we would change/improve here, perhaps you can be more specific what you mean, @mloning?

Maybe @mloning means that the warning should be more specific to `all_estimators` rather than just a generic import error, so sth like "warning, estimators interfacing the following packages are not listed as they are not installed: [list of packages]"?

xloem: Could also just add a comment making the purpose of the exception more clear.  Not sure where to put the trade-off between user friendliness and clarity for debugging.

fkiraly: sure, I guess sth on what's trying to load the module and how to resolve it would help

mloning: @xloem yes I also looked up what the best way is to handle this. I thought that it may be good to check for the specific error message that we expect from missing soft dependencies rather than the generic `ModuleNotFoundError`. I'm not sure about other possible reasons, perhaps if installations go wrong, missing compiled files, accidentally deleted files? Would the error message raised by sktime when a soft dependency is missing be language dependent? You can find it here: https://github.com/alan-turing-institute/sktime/blob/722a3705a14c9b164cb548a160db4d24b2fd037f/sktime/utils/validation/_dependencies.py#L25

xloem: Incredibly relevant code.  Do you know if that function is called for every import here with a soft dependency?  Or would it be appropriate to call it manually in this function?

mloning: We make sure that missing soft dependencies raise a more helpful error message and not just a generic module not found error. So I don't think we need to call it here, that should be taken care of in the respective modules. But I think we should check the expected error message when catching the exception.

xloem: I'm testing this out a little towards implementing it.  I actually found there's a lot of use in catching errors _other_ than `ModuleNotFoundError`.  For example, on my system right now `ROCKETClassifier` is raising an `UnsupportedError` inside numba when imported, preventing enumeration of estimators.

Although `UnsupportedError` isn't specifically the case of a soft dependency missing, `all_estimators` is still most useful to a user if it can skip estimators that fail to import and return to the user a list of ones that presently work.

What are your thoughts on going the other way, catching `Exception` instead of `ModuleNotFoundError`, and displaying a cause-agnostic warning when an exception is caught?

mloning: Thanks for adding the message based condition. 

Not sure about the other idea. For example, `numba` is a core dependency, so if that cannot be imported without errors I think I'd prefer raising the exception rather than a warning. It may be more confusing for users to see a warning there which makes it unclear whether this is something that needs fixing or not, compared to an error.

fkiraly: hm, what's the logic that you would suggest, @mloning?

xloem: I'm not quite setup to retest, but I think other parts of `numba` may have been working for me; and I think the other solution gives the user more flexibility.  Still, what you say sounds ok.

fkiraly:
@mloning, this is still waiting for you to comment on the resolution of your issue.

fkiraly:
also, sorry to say this, @xloem - there is a new merge conflict with `.all_contributors`... fortunately @mloning is back from holiday and should be able to continue and potentially resolve the discussion on `numba`.

xloem:
I've rebased on main again!

Regarding `numba`, it comes down to me just having a different opinion here, that this code is most useful to users of the library if all failures are soft.  I, personally, would only make errors fatal if none of the estimators could load.  But it's your project, not mine, and I've implemented the approach mloning described, and with it the improved understanding of the roles of parts of the project that it demonstrates.

mloning (approved):
Hi @xloem, sorry for the delay, I was away for a few days. Looks good to me now as it is!

mloning:
Now merged - thanks again @xloem! 🎉

PR #1260, opened 2021-08-06
Classification experiments

TonyBagnall:
1. refactors run_experiments into the benchmarking module
2. adds unit tests for run_clustering and run_classification experiments


MatthewMiddlehurst (commented):
Unsure about limiting the classifiers that can be run in experiments, I think it would be better to add the file to the soft dependencies list if its causing tests to fail.

Any reason for keeping the contrib experiments file? Seems to just be the loading demo.

Removing the following from the new file will impact out cluster runs maybe?
```
os.environ["MKL_NUM_THREADS"] = "1"  # must be done before numpy import!!
os.environ["NUMEXPR_NUM_THREADS"] = "1"  # must be done before numpy import!!
os.environ["OMP_NUM_THREADS"] = "1"  # must be done before numpy import!!
```

TonyBagnall:
> Unsure about limiting the classifiers that can be run in experiments, I think it would be better to add the file to the soft dependencies list if its causing tests to fail.

I dont know how to do that, agree it would be better though. Can you do it? 

> Any reason for keeping the contrib experiments file? Seems to just be the loading demo.

the idea is experiments are run through the contrib file, so that the weird os stuff below is decoupled from the experimental code, and to avoid having a main method in experiments, which is frownedupon

> Removing the following from the new file will impact out cluster runs maybe?
> 
> ```
> os.environ["MKL_NUM_THREADS"] = "1"  # must be done before numpy import!!
> os.environ["NUMEXPR_NUM_THREADS"] = "1"  # must be done before numpy import!!
> os.environ["OMP_NUM_THREADS"] = "1"  # must be done before numpy import!!
> ```

TonyBagnall:
I'm also thinking that the whole set_classifier model should be phased out, its a bit hacky. tsml is moving towards a Producer functional interface method. I'm thinking most will use run_classifier rather than load_and_run, and handle the set up externally (in the contrib file for example)

MatthewMiddlehurst (approved):
See latest commit. Left the clustering stuff alone.

I removed the format option for now. Feel free to add it back in, but currently it just calls load_ts no matter what the input is.

MatthewMiddlehurst (approved):
One of my changes broke the test, oops.

Also worth noting that there is not an option to use pre-defined resamples currently.

PR #1262, opened 2021-08-07
Troubleshooting for C compiler after pytest failed

tensorflow-as-tf:
This Pull Request aims to update the installation guide of C compilers for Mac OSX. Many contributors came across the problem that this troubleshooting addresses with their solution.


docs/source/installation.rst
```diff
@@ -283,6 +283,17 @@ Then you need to set the following environment variables:
 
 Finally you can build the package using the standard command.
 
+Release versions - troubleshooting on Mac OSX
```
fkiraly, line 286: are you sure this is for release versions (which you install from `pip` or `conda` and not just for development versions which you install from a local clone?

I think you mean development version, since you actually install `pip -e`. Right?

fkiraly (changes requested):
Looks good, thanks!

But it should be "development versions", no?

PR #1266, opened 2021-08-08
Fixed Binder Dockerfile

### LINKED ISSUE #662: Binder build fails

### LINKED ISSUE #807: [BUG] binder notebook environment build breaks over scikit-posthocs unable to load pyqt5 in setup.py
**Describe the bug**

When running the sktime binder, the environment setup breaks over scikit-posthocs unable to load pyqt5 in setup.py.
This is reproducible using multiple browsers and computers (all running Windows).

Have not investigated what causes the issue, but it´s possibly an incomplete requirements file somewhere in the dependency graph.

See printout here.

image

image

**To Reproduce**

1. Go to sktime binder
2. watch the "starting prompt"

**Expected behavior**

no error occurs, binder loads as expected

corvusrabus:
#### Reference Issues/PRs
Fixes #807
Fixes #662 

The Binder Dockerfile stopped working at some point. I changed the requirements files for the Docker file and it works now. The Docker file "compiles" and I ran the first example notebook in it. 

I changed the base Docker image used because the one that is in the current Dockerfile is just an arbitrary one copied from the Binder docs. I think that the Jupyter Notebook Python 3.8 is a reasonable base image because it already contains some of the needed libraries and Python 3.8 is the newest version that can currently be used with Numba (according to my understanding).

This is not an optimal solution yet because something is broken with the way Prophet is being built but this can only be changed once the Prophet developers fix it. I know it is weird to have two requirement files but Prophet doesn't build without having some libraries installed first.

No

##### For all contributions


fkiraly (changes requested):
Very nice, thanks!!

Didn't think I'd ever see the day where someone figures out what's wrong with our binder!
So it was `prophet` all along? Go figure.

You forgot to add yourself to `.all-contributorsrc`, though.

fkiraly:
>This is not an optimal solution yet because something is broken with the way Prophet is being built but this can only be changed once the Prophet developers fix it. 

But it's great because it works!
Which means it's no longer pessimal, and that's very much enough, thanks.

fkiraly (changes requested):
Hmmmm - it seems that binder works, but it somehow broke CI/CD (see test_examples below and the inability to run `jupyter`) - could you kindly check what is happening there?

corvusrabus:
Yes I noticed that the Github workflows use the binder requirements file to test the example files. I changed the workflow to include the second requirements file I added. I hope it passes the test now

fkiraly (approved):
looks good to me now!

Would be great if you could have a look, @mloning, since you set up the binder originally

fkiraly:
so, it runs but there's stil the problem of not being able to build a `prophet` wheel (so it defaults to a legacy setup):

image

... any good ideas? All that red output may look scary for the typical user...

mloning:
Thanks @corvusrabus for the fix - great work!

PR #1277, opened 2021-08-11
Critical Difference Diagrams

### LINKED ISSUE #1186: Critical difference diagrams
as part of the development of analysis tools, it would be nice to be implement a way of generating critical difference diagrams for comparing multiple classifiers over multiple data sets of the form

image

solid bars represent cliques, within which it is assumed there is no signficant difference between classifiers. The test of difference can be the post hoc test suggested in the original paper 
https://www.jmlr.org/papers/v7/demsar06a.html
or the pairwise test used in tsml 
https://github.com/uea-machine-learning/tsml/tree/master/src/main/matlab
or if you really want to go to town, port in the Bayesian tests
https://github.com/janezd/baycomp/blob/master/tests/test_multiple.py

SveaMeyer13:
Fixes #1186

Adds functionality to plot a critical difference diagram.

As discussed the entire calculation of critical differences and cliques as well as the plotting happens within one standalone function. 

Coding style checks will fail due to hard-coded qalpha values.


sktime/contrib/critical_difference.py
```diff
@@ -0,0 +1,600 @@
...
+      number of datasets classified per strategy
+    ranked_data : np.array (shape: n_strategies * n_datasets)
+      rank of strategy on dataset
+
+    Returns
+    -------
+    is_significant : bool
+      Indicates whether strategies differ significantly in terms of performance
+      (according to Friedman test).
+    """
```
fkiraly, line 35: I'd include a reference to which algorithm you are implementing, e.g., which formula for the approximation of the null distribution

sktime/contrib/critical_difference.py
```diff
@@ -0,0 +1,600 @@
+# -*- coding: utf-8 -*-
+"""Standalone function to plot critical difference diagram."""
+
+__author__ = ["Svea Meyer"]
```
fkiraly, line 4: should be github user name

sktime/contrib/critical_difference.py
```diff
@@ -0,0 +1,600 @@
...
+
+_check_soft_dependencies("matplotlib")
+import matplotlib.pyplot as plt
+
+
+def _check_friedman(n_strategies, n_datasets, ranked_data, alpha):
+    """
+    Check whether Friedman test is significant.
+
+    Larger parts of code copied from scipy.
```
fkiraly, line 19: if it's copied - why can't you use `scipy` directly, leaving the logic in `scipy`?

TonyBagnall (approved):
Looks good, it's in contrib

sktime/contrib/critical_difference.py
```diff
@@ -0,0 +1,600 @@
...
+    ) / c
+    p = distributions.chi2.sf(chisq, n_strategies - 1)
+    if p < alpha:
+        is_significant = True
+    else:
+        is_significant = False
+    return is_significant
+
+
+def plot_critical_difference(
```
fkiraly, line 64: this is ok-ish, though I think what the function body should be doing is only plotting.
There is some testing logic in there and clique computation, which I'd suggest to move out to separate functions.
Similarly, one could think about factoring out the raw plotting (only positions, cliques etc passed), and have a main function which does everything together - that would be cleaner and follow the "separation of concerns" pattern better.

fkiraly:
(but I would not consider the structural comments a blocker)

PR #1278, opened 2021-08-11
Data writing updates and JapaneseVowels dataset fix

### LINKED ISSUE #1171: [BUG] load_japanese_vowels() returns the same dataset for training and testing

MatthewMiddlehurst:
**Data writing update:**

Implements a new function which takes a time series dataset and tabular output transformation, saves the results of the transform to the Weka .arff format.
  
Also made some changes to the dataframe_to_ts writer currently there. The primary .ts writing function now takes a ndarray and the dataframe function converts to ndarray. The current version can have some strange outputs, lots of trailing 0s and some lines fully in scientific format while others are not. A lot cleaner just to write the contents of a numpy array.
  
Timestamps aren't currently supported and whether the data is univariate or multivariate can be determined from the data, so those parameters have been removed.

**JapaneseVowels fix:**

Fixes #1171.
See dicussion of previous PR for the issue (#1182).

Replaces the sktime JapaneseVowels datasets with the versions currently uploaded to the [timeseriesclassification.com](timeseriesclassification.com) archive. The sktime ones were padded with 0s, while the new versions are unequal length.

Replaces JapaneseVowels with BasicMotions for the tests that used it previously, as they cant handle unequal length it seems and change breaks them.

Thanks to @off99555 for the initial PR and issue.


TonyBagnall (approved):
All looks good to me

PR #1285, opened 2021-08-12
Add binder badge to README

mloning:
Related to #1266


fkiraly (approved):
I was actually looking for that, could't find it

PR #1286, opened 2021-08-12
checks invalid X type for y multivariate and univariate

thayeylolu:
if the estimator has the tag scitype:y = multivariate, the test example should have 2 columns and should have 1 column if univariate.

### DIFF
--- a/sktime/forecasting/tests/test_all_forecasters.py
+++ b/sktime/forecasting/tests/test_all_forecasters.py
@@ -125,13 +125,13 @@ def test_X_invalid_type_raises_error(Forecaster, X):
     f = _construct_instance(Forecaster)
 
     if f.get_tag("scitype:y") == "univariate" or f.get_tag("scitype:y") == "both":
-        y = _make_series(n_columns=1)
+        y_train = _make_series(n_columns=1)
 
     elif f.get_tag("scitype:y") == "multivariate" or f.get_tag("scitype:y") == "both":
-        y = _make_series(n_columns=2)
+        y_train = _make_series(n_columns=2)
     try:
         with pytest.raises(TypeError, match=r"type"):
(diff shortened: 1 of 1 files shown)


thayeylolu:
Reopen

sktime/forecasting/tests/test_all_forecasters.py
```diff
@@ -123,6 +123,12 @@ def test_y_invalid_type_raises_error(Forecaster, y):
 def test_X_invalid_type_raises_error(Forecaster, X):
     """Test that invalid X input types raise error."""
     f = _construct_instance(Forecaster)
+
+    if f.get_tag("scitype:y") == "univariate" or f.get_tag("scitype:y") == "both":
+        y_train = _make_series(n_columns=1)
+
+    elif f.get_tag("scitype:y") == "multivariate" or f.get_tag("scitype:y") == "both":
```
fkiraly, line 130: when we reach this line, it can't be that `f.get_tag("scitype:y") == "both"` because that already is checked in the first line.
So it's not wrong to have it in the condition, but redundant, since the second block will only be entered if `f.get_tag("scitype:y") == "multivariate"`. I would simplify this.

fkiraly (changes requested):
Makes sense and is in-principle correct, but your case distinction has redundant checks so is confusing to read.

sktime/forecasting/tests/test_all_forecasters.py
thayeylolu: Thank you Franz. I will edit it

fkiraly (approved):
I think this one can go in now, no redundant conditions anymore.

Note that I don't think this needs to run both cases if the condition is `"both"`, since it's catching `X` that we're testing, not `y`.

PR #1288, opened 2021-08-13
Checks if predicted time index with an X matches forecasting horizon for multivariate and univariate y

thayeylolu:
Fixing the `fix_test_predict_time_index_with_X` function present in `test_all_forecasters.py` so  it Checks if predicted time index matches forecasting horizon for multivariate and univariate y


thayeylolu:
I am closing this pr because it has been merged to #1301

PR #1292, opened 2021-08-13
Adding TrendForecaster API reference in doc

tensorflow-as-tf:


RNKuhns:
@mloning and @tensorflow-as-tf, do we want to actually have different subsections for the trend forecaster and polynomial trend forecaster?

If I was looking for information, I'd probably want to know what are the trend forecasting methods that sktime has and it would make sense to have them both under the same "Trend" heading in that case.

RNKuhns (changes requested):
I don't think that we should have a separate heading for PolynomialTrendForecaster and TrendForecaster.

I think these are clearly related and should show up under the same heading.

mloning:
Seems like this has been added already.

PR #1295, opened 2021-08-14
Tsc_Refactoring _cboss

BINAYKUMAR943:
#### Reference Issues/PRs
#1146 TSC Refactoring for _cboss

all the public methods are turned into private
input and output checks are removed from the private version
all the test cases are passing
name of capabilities dictionary is changed to _tags
new tags are added to _tag registry

No
##### For all contributions


sktime/classification/dictionary_based/_cboss.py
```diff
@@ -267,7 +264,7 @@ def fit(self, X, y):
         self._is_fitted = True
```
fkiraly, line 264: this line should also be removed

sktime/registry/_tags.py
```diff
@@ -125,11 +125,41 @@
...
+        "bool",
+        "Does classifier support limiting of max build time?",
+    ),
+   (
         "capability:pred_int",
         "forecaster",
         "bool",
         "is the forecaster capable of returning prediction intervals in predict?",
-    ),
+    )
```
fkiraly, line 162: I think there should be a comma, or the linter will be unhappy

BINAYKUMAR943: > I think there should be a comma, or the linter will be unhappy

Indeed! It was failing in the linter check due to it. I have added a comma at the end.

sktime/classification/dictionary_based/_cboss.py
BINAYKUMAR943: Thanks for reviewing. I have made this change in new commit.

> this line should also be removed

fkiraly:
alright, all good!

Now, I'm planning to merge #1239 first (next couple hours once tests pass), which will lead to some conflicts in the tag descriptions.
You could prevent that by just copying the tags from #1239, or resolve the conflicts later.

fkiraly:
so, it doesn't show the `_tags` file as conflicted, but I think it will lead to duplication if you leave it as-is - kindly have a look and check that each tag is contained only once.

fkiraly:
(I could also do this, it's very quick, but I don't want to push directly to your PR in case you're also on it - let me know)

BINAYKUMAR943:
> #1239 is merged now
> 
> Note it doesn't show the `_tags` file as conflicted, but I think it will lead to duplication if you leave it as-is - kindly have a look and check that each tag is contained only once.

thanks for the suggestion!  I have removed tags added by me as they were already added in #1239.

TonyBagnall (approved):

PR #1296, opened 2021-08-15
governance: public list of roles on sktime

fkiraly:
based on some discussions, it's not easy to find out who does what on `sktime` - for instance, CC membership is on the CC repo, CoC membership is in the CoC, and core contributors are listed nowhere in particular (which makes that information not findable).

This PR remedies the problem of findability by adding a `ROLES.md` to the repository root, as a single source of truth.
I've filled it in to the best of my current knowledge.


RNKuhns:
@fkiraly I think it also makes sense to list current Summer program participants (and eventually have a past Summer program participants section). What do you think? 

Markus and I had also been thinking it would be good to add this as a sub-file in the "About" section in our online docs. One benefit in the short-run is during events like a code or the upcoming docsprint we can let people know about that section of the documentation so they can look up relevant people to ask questions.

fkiraly:
>Markus and I had also been thinking it would be good to add this as a sub-file in the "About" section in our online docs. One benefit in the short-run is during events like a code or the upcoming docsprint we can let people know about that section of the documentation so they can look up relevant people to ask questions.

Yes, sure! You can easily embed `md` there, so would be fine with this - I'm not familiar enough with the new webpage back-end though to do that in this PR.

fkiraly:
>@fkiraly I think it also makes sense to list current Summer program participants (and eventually have a past Summer program participants section). What do you think?

That would make sense, my problem is that I don't have the full list of mentees. Someone else could add it?

mloning:
I'd prefer to declutter the root directory and instead add this to the website as describe by @RNKuhns above, see https://github.com/alan-turing-institute/sktime/issues/1204 and https://github.com/alan-turing-institute/sktime/issues/1159.

For the website, `rst` is preferred, `md` doesn't support cross-references.

fkiraly:
>For the website, rst is preferred, md doesn't support cross-references.

I don't mind either way, I just prefer having it above not having it in any form.

We should also think about findability - it should be a one-click user journey from the landing page rather than having to click through through the docs.

What would the location and "click distance" be for your proposed location?

mloning:
Closed in favour of #1328

Issue #1299, opened 2021-08-16
Implement modular ThetaForecaster


GuzalBulatova:
**References re-design of Theta Forecaster #854**

Implements ThetaLinesTransformer #923
Implements ColumnEnsembleForecaster #1082

**Describe the solution you'd like**

Implement modular `Theta Forecaster` that will
1. Decompose original data (univariate series) into theta lines (implementation of `ThetaLinesTransformer`, returns transformed time series in DataFrame);
2. Forecast each of these theta lines with different forecasters (implementation of `ColumnEnsembleForecaster`, returns predicted DataFrame)
3. Then aggregate the forecasts into one resulting prediction (return univariate pd.Series)

Users should be able to pass theta values for the transformation step, forecasters and aggregation function as parameters.

**Related issues:**
Poor Theta model predictions #421
Implementation of AutoTheta Forecaster #738

Existing implementations:
in sktime `sktime/forecasting/theta.py` (theta parameter is assumed)

fkiraly:
done in #1300

PR #1300, opened 2021-08-16
[ENH] modular/configurable Theta forecaster

GuzalBulatova:
#### Reference Issues/PRs

References #1299. See also #1082 #923 and #854.

Implements modular ThetaForecaster.

No.

This is the first draft, so any. functionality suggestions are very welcome. The idea is to start with "hard-coded" autogenerated `self.forecasters`, get this working and then update so that they can be passed by users in the constructor.

#### PR checklist


mloning:
So from our discussion just now, this works for me: 
```python
from sktime.forecasting.compose import EnsembleForecaster, ColumnEnsembleForecaster
from sktime.transformations.series.theta import ThetaLinesTransformer
from sktime.forecasting.exp_smoothing import ExponentialSmoothing

y = load_airline()

theta_values = (0, 2)
forecasters = []
for i, theta in enumerate(theta_values):
    if theta == 0:
        name = "trend" + str(i)
        forecaster = (name, PolynomialTrendForecaster(), i)
    else:
        name = "ses" + str(i)
        forecaster = (name, ExponentialSmoothing(), i)
    forecasters.append(forecaster)

    ("transformer", ThetaLinesTransformer(theta=theta_values)),
    ("forecaster", ColumnEnsembleForecaster(forecasters=forecasters)),
])
pipe.fit(y)

y_pred = pipe.steps_[-1][-1].predict(fh=[1, 2]).mean(axis=1)
y_pred
```

sktime/forecasting/theta_new.py
```diff
@@ -0,0 +1,82 @@
...
+        self.aggfunc = aggfunc
+        self.forecasters = forecasters
+        self.weights = weights
+        self.theta_values = theta_values
+        super(ThetaNewForecaster, self).__init__(forecasters=self.forecasters)
+
+    def _fit(self, y, X=None, fh=None):
+        # TODO TransformedTargetForecaster should be able to work with multivariate data
+
+        forecasters = []
```
miraep8, line 40: I have 2 small suggestions here! :)  

First - I am wondering if it could make sense to first see if `self.forecaster`'s is `None` here, and if it is one could input that into the `ColumnEnsembleForecaster` and if not could populate it based on the theta values as you do here (this suggestion is just based on what you mentioned earlier that it would be nice to allow the user more control over what forecasters are used and their values!)

Second - I know in `MultiplexForecaster` there was this convention of storing modified forecasters in `self._forecasters` rather than `self.forecasters`.  I was just thinking there might be value in doing that here as well? (ie this way you could store the forecasters you generate somewhere without modifying the passed attribute!) 

Also - some handy functionality buried within `HeterogenousEnsembleForecaster` that might be useful to you are some of the functions for taking a list of estimators and returning a list of (name, estimator) tuples - I think the signature is `self._get_estimator_tuples(estimators)`! :) So you could so something like:

``` 
def map_theta(theta):
   if theta == 0:
       return PolynomialTrendForecaster()
  return ExponentialSmoothing()
self._forecasters = self._get_estimator_tuples(list(map(map_theta, self.theta_values))
```
Of course it doesn't really do anything that different from what you are doing, and it is not blocking, just learned about it recently myself and thought it could be nice to take advantage of some of the functionality in `HeterogenousEnsembleForecaster`!

sktime/forecasting/theta_new.py
```diff
@@ -0,0 +1,82 @@
...
+        #         ]
+        #     )
+
+        self._pipe.fit(y=y, X=X, fh=fh)
+        return self
+
+    def _predict(self, fh, X=None, return_pred_int=False):
+        # Call predict on the forecaster directly, not on the pipeline
+        # because of output conversion
+        Y_pred = self._pipe.steps_[-1][-1].predict(fh, X)
```
miraep8, line 76: I think I am just a bit confused here, but I don't understand entirely why you don't want to transform `X` before passing it into the forecaster as part of `predict`?  But again, I suspect this is probably because I just don't understand what `ThetaLinesTransformer` is doing...   Do you think it could be worth adding more detail to the comment?  Or is it perhaps something obvious? (in which case its probably fine!)

sktime/forecasting/theta_new.py
```diff
@@ -0,0 +1,82 @@
...
+from sktime.forecasting.base._meta import _HeterogenousEnsembleForecaster
+from sktime.forecasting.compose import ColumnEnsembleForecaster
+from sktime.forecasting.compose._ensemble import _aggregate
+from sktime.forecasting.compose._pipeline import TransformedTargetForecaster
+from sktime.forecasting.exp_smoothing import ExponentialSmoothing
+from sktime.forecasting.trend import PolynomialTrendForecaster
+from sktime.transformations.series.theta import ThetaLinesTransformer
+
+
+class ThetaNewForecaster(_HeterogenousEnsembleForecaster):
```
miraep8, line 18: One small thought here - I *think* that there seems to be a movement away from using `HeterogenousEnsembleForecaster` and instead just using it's parent `HeterogenousMetaEstimator` instead!  I think you could probably do that here as well (though I am sorry if I am wrong and breaks something!). :)

miraep8:
Overall I think it seems like a sensible change :)  Would be happy to take another look when you are ready to merge!  I really like the idea though + imagine it would be useful!  (Would be really cool to see if it improves `ThetaForecaster`'s performance to have these things tied together!).

fkiraly:
@GuzalBulatova, this is not the most recent code yet that we wrote yesterday, is it?

sktime/forecasting/theta_new.py
GuzalBulatova: Hi Mirae! Thanks for you feedback! 
Those are very valid points.
I think the first one should be now there. I'll also add forecasters check so that user can't pass more/less estimators than required, but if they pass None then it's populated based on theta values like you suggested :)
The second point I'll make sure to check a bit later and come back once it's done.

sktime/forecasting/theta_new.py
GuzalBulatova: Ah yes, correct, this forecaster doesn't work with exogenous values. Actually a great point about adding a comment on X in the class docstring, I'll do that! 👍

sktime/forecasting/theta_new.py
GuzalBulatova: We inherit from Forecaster here because we need the get and set params methods. We do eventually overwrite them so maybe makes sense to inherit from MetaEstimator instead actually. @fkiraly what do you think?

fkiraly:
the mac issue should be resolved if you pull from main.

It was coming from other estimators that have been removed from testing

fkiraly:
I see you already updated from main and uncommented the excepted estimators.

I would recommend not to do investigation of the bug in this PR, do that separately (cf your own advice on keeping different ideas in diffent PR 😄 )

GuzalBulatova:
Hi @fkiraly , I just decided to quickly do it here and not open a new pr, apparently I changed the python version tag in a wrong place. I'll undo these changes of course. Thank you for advice!

fkiraly:
>apparently I changed the python version tag in a wrong place. I'll undo these changes of course. Thank you for advice!

I think you accidentally commented out some lines in `_config`, which is why the tests are failing.

fkiraly:
@GuzalBulatova, what is going on with this?

This looks done, but fails because:
* not updated to `main`
* changes to `_config`

sktime/forecasting/theta_new.py
fkiraly: hm, not sure what the consequences would be of that. Arer we using any of the forecaster specific methods?

fkiraly (changes requested):
I think we are almost there!

May I request the following changes

* move the forecaster into `theta.py`, this is also a theta based forecaster
* rename the forecaster to sth more descriptive, e.g., `FlexibleThetaForecaster` or `ThetaModularForecaster`
* remove commented out code, uncomment the commented out test cases or replace them by sth similar
* write a proper docstring for the class

fkiraly:
Surprised that it is failing, was almost going to merge. What exactly did you change since it passed all the tests?

fkiraly (changes requested):
Thanks, that works!

* the `set_params` overrides etc were from a time where `__init__` would not be called automatically. This has changed in one of the more recent versions, so we can remove all the overrides imo. I made as suggested update to your PR which simplifies things a lot? Kindly check whether this is ok, otherwise you can revert.
* please document the parameters in the docstring
* I would suggest you implement a test that checks whether predictions align with the "standard" theta forecaster

GuzalBulatova:
@fkiraly thanks for your changes, I think they make sense. 
I've made the requested changes. 
As for the test for comparing existing and modular forecaster - they are very different for when y isn't a logarithm of load_airline. I'm investigating that now.

fkiraly (approved):
Thanks, I consider the points addressed.

Left some comments on improving the docstring, but these are not blocking.

sktime/forecasting/theta.py
```diff
@@ -262,3 +268,160 @@ def _zscore(level: float, two_tailed: bool = True) -> float:
...
+    Modularized implementation of Theta method as defined in [1]_ (TODO: add the
+    auto-theta method as described in [2]_).
+
+    Overview: Input :term:`univariate series <Univariate time series>` of length
+    "n" and decompose with :class:`ThetaLinesTransformer
+    <sktime.transformations.series.theta>` by modifying the local curvature of
+    the time series using Theta-coefficient values - `theta_values` parameter.
+    Thansformation gives a pd.DataFrame of shape `len(input series) * len(theta)`.
+
+    The resulting transformed series (Theta-lines) are extrapolated separately.
```
fkiraly, line 285: I think it remains unclear what "theta-lines" are. It might help to explain this in math, or below ...

sktime/forecasting/theta.py
```diff
@@ -262,3 +268,160 @@ def _zscore(level: float, two_tailed: bool = True) -> float:
...
+    The forecasts are then aggregated into one prediction - aunivariate series,
+    of `len(fh)`.
+
+    Parameters
+    ----------
+    forecasters: list of tuples (str, estimator, int or pd.index), default=None
```
fkiraly, line 291: ... you could explain better that `forecasters` and `theta_values` are "paired" with each other.

sktime/forecasting/theta.py
```diff
@@ -262,3 +268,160 @@ def _zscore(level: float, two_tailed: bool = True) -> float:
...
+        Forecasters to apply to each Theta-line. If None, will apply
+        PolynomialTrendForecaster (linear regression) the the Theta-lines
+        where theta_value equals 0, and ExponentialSmoothing - where theta_value
+        is different from 0.
+    theta_values: sequence of float, default=(0,2)
+        Theta-coefficients to use in transformation.
+    aggfunc: str, default="mean"
+        Aggregation function to apply to results of theta-lines predictions
```
fkiraly, line 299: I think it would be better to say precisely how aggregation is taking place (across different theta lines, for given time stamps and hierarchy indices if present; possibly, with math, and in the main body)

sktime/forecasting/theta.py
```diff
@@ -262,3 +268,160 @@ def _zscore(level: float, two_tailed: bool = True) -> float:
...
+        (multivariate) in order to get resulting prediction (univariate).
+    weights: lsit of floats, default=None
```
fkiraly, line 301: typo - list
should also include constraint on length, and explanation how it interacts with `aggfunc`.

sktime/forecasting/theta.py
```diff
@@ -262,3 +268,160 @@ def _zscore(level: float, two_tailed: bool = True) -> float:
...
+    Parameters
+    ----------
+    forecasters: list of tuples (str, estimator, int or pd.index), default=None
+        Forecasters to apply to each Theta-line. If None, will apply
+        PolynomialTrendForecaster (linear regression) the the Theta-lines
+        where theta_value equals 0, and ExponentialSmoothing - where theta_value
+        is different from 0.
+    theta_values: sequence of float, default=(0,2)
+        Theta-coefficients to use in transformation.
+    aggfunc: str, default="mean"
```
fkiraly, line 298: should include an explanation of what are valid strings, and where the function is then obtained from, what its signature is, and its relation to `weights`

sktime/forecasting/theta.py
```diff
@@ -262,3 +268,160 @@ def _zscore(level: float, two_tailed: bool = True) -> float:
...
+    See Also
+    --------
+    ThetaForecaster, ThetaLinesTransformer
+
+    Examples
+    --------
+    >>> from sktime.datasets import load_airline
+    >>> from sktime.forecasting.theta import ThetaModularForecaster
+    >>> y = load_airline()
+    >>> forecaster = ThetaModularForecaster()
```
fkiraly, line 323: I think for the example it would be more instructive to show a non-default case

sktime/forecasting/theta.py
```diff
@@ -262,3 +268,160 @@ def _zscore(level: float, two_tailed: bool = True) -> float:
...
+            Parameters to create testing instances of the class.
+            Each dict are parameters to construct an "interesting" test
+            instance, i.e., `MyClass(**params)` or `MyClass(**params[i])`
+            creates a valid test instance. `create_test_instance` uses the first
+            (or only) dictionary in `params`.
+        """
+        # imports
+        from sktime.forecasting.naive import NaiveForecaster
+
+        params0 = {
```
fkiraly, line 419: I would recommend, also test the `weights` parameter

sktime/forecasting/theta.py
GuzalBulatova: There's more on that in ThetaLinesTransformer and in the original papers referenced below. Should I still include explanation of the math here?

fkiraly: yes, I would suggest that.

An alternative may be linking to the transformer, but that might be more difficult to read.

fkiraly:
your doctest is failing. Newlines in an expression should start with `...`, not `>>>`.

fkiraly (changes requested):
looks all good now, except the failing doctest

fkiraly (approved):
I've fixed the failing doctest, so it's good to go now

PR #1301, opened 2021-08-16
Tests failing for VAR in test_all_forecasters.py now  accommodates multivariate and both estimators

thayeylolu:
The changes made to the `test_all_forecasters.py` was to ensure VAR model pass all test in it. It also ensures that estimators that handles univariate and multivariate forecasters pass. Please seee #1303

PR #1302, opened 2021-08-16
Check correct time index of update-predict when y is univariate, both and multivariate

thayeylolu:
Check correct time index of update-predict when y is univariate, both and multivariate


thayeylolu:
I am closing this pr because it has been merged to #1301

Issue #1303, opened 2021-08-16
[BUG] Tests and checks failing for Multivariate Forecasters that cannot take univariate series


thayeylolu:
**Describe the bug**

Tests and checks failing for Multivariate Forecasters  Using Vector Regression as an Example.

The following are the test failing for VAR model which is a multivariate forecasters. I have created PRs for some
 **NOTE** 
The individual branches in section A have been merged into one branch `test_score` #1301 

### SECTION A
filename: `test_all_forecasters.py`
- test_update_predict_single #1302
- test_score #1301
- test_predict_pred_interval #1298
- test_X_invalid_type_raises_error #1286
- test_predict_time_index_with_X #1288
- test_get_fitted_params #1216
- test_predict_time_index #1287 
- test_update_predict_predicted_index #1307
- test_update_predict_predicted_index_update_params #1307
 
### SECTION B
filename: `test_sktimeforecasters.py`
- test_oh_setting #1306
- test_no_fh_opt #1306
- test_fh_in_fit_opt #1306
- test_fh_in_predict_opt #1306
- test_same_fh_in_fit_and_predict_opt #1306

**Expected behavior**

It is expected that all forecasters including multivariate forecasters should not fail tests present in the `test_all_forecasters.py`. 

**Additional context**

NOTE: this is only for VAR model. Not all the tests are in `all_forecasters.py`

fkiraly:
Note: these are checks that fail if the forecaster is multivariate, but does *not* accept univariate input.

This is a strange edge case that can be found for instance in `statsmodels` VAR, which doesn't take one-column data frames.

The gist is that the tests need to replace all test cases which currently use univariate series with bi- or higher variate ones.

fkiraly:
To summarize the "end point" logic we agreed on in our meeting today, @thayeylolu:

* if the forecaster is "univariate" (in the `scitype:y` tag), the current test logic should be run
* if the forecaster is "multivariate", generation of univariate examples should be replaced by bivariate examples, and any related logic adapted
* if the forecaster is "both", *both* of the above should be tested, i.e., running current tests with univariate examples, and then *also* (as in: in addition) running the tests with bivariate examples

fkiraly:
This was fixed with the linked PR and the multivariate upgrade in 0.8.0.

PR #1083, opened 2021-06-30
Add wrapper for statsmodels VAR 

thayeylolu:
#### Reference Issues/PRs
Issue https://github.com/alan-turing-institute/sktime/issues/929

#### What does this implement/fix? Explain your changes.
This PR aims to wrap the Vector Auto Regressor in Statmodels

#### Does your contribution introduce a new dependency? If yes, which one?
none

task

- [x] Wrap VAR
- [x]  Accepts only DataFrames
- [x] Write Test
- [ ] Write a test with hyper parameters

PR #1304, opened 2021-08-16
Check _y and cutoff is None after construction for univariate, both and multivariate y

thayeylolu:
This pull request is a result of the issue #1303 and PR #1083
Check _y and cutoff is None after construction for univariate, both and multivariate y


thayeylolu:
Made a new branch fixing all test cases

PR #1306, opened 2021-08-16
Fix test sktimeforecasters for univariate, both and multivariate y

thayeylolu:
#1303 
#1083


sktime/forecasting/tests/test_sktime_forecasters.py
```diff
@@ -42,6 +42,20 @@
     if issubclass(forecaster, _BaseWindowForecaster)
 ]
 
+
+# helper function
+def _get_n_columns(tag):
```
mloning, line 47: Let's move this to `https://github.com/alan-turing-institute/sktime/blob/main/sktime/utils/_testing/forecasting.py` and re-use it in both files.

mloning (commented):
As discussed, just a quick comment, I'll later take a closer look, but already looks good!

mloning (approved):
Thanks @thayeylolu - looks great! As discussed, I think we can ignore the failing doc tests here.

Issue #1245, opened 2021-08-02
[DOC] Docstrings issues for upcoming Docsprint


RNKuhns:
#### Describe the issue linked to the documentation

This will be used to track issues related to docstring improvements needed as part of the upcoming docsprint. 

Expectations about docstrings in sktime are summarized in #1297, and include the following: 

- Follow NumPy docstring conventions so they render in sktime's updated documentation website (https://www.sktime.org/en/latest/) and the documentation guide portion (https://www.sktime.org/en/latest/developer_guide/documentation.html)
- Ensuring all parameters and attributes (classes) are documented completely and consistantly
- Including links to the relevant topics in the online documentation's glossary and user guide's in the extended summary
- Including an `Examples` section that demonstrates at least basic functionality in all public code artifacts
- Adding a `See Also` section that references related sktime code artifacts as applicable
- Including citations to relevant sources in a `References` section

Module level docstrings should generally follow the format used by `scikit-learn`. For example, see [column transformer](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/compose/_column_transformer.py) or [compose](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/compose/__init__.py) or [ensemble](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/ensemble/__init__.py). Additional, but brief, information could also be included in some cases, but check with a core developer.

Note: when updating docstrings prioritization should be given to code artifacts that are part of the public API, followed by heavily used non-public code objects. 

#### Suggest a potential alternative/fix
The goal of the docstring section of the docsprint is to reform all docstrings in the following modules to our preferred format:
   - [ ] annotation
   - [ ] base
   - [ ] classification
   - [ ] clustering
   - [ ] distances
   - [ ] forecasting
   - [ ] performance_metrics
   - [ ] regression
   - [ ] transformations
        - [ ] Boxcox -- Ryan
        - [ ] LogTransformer -- Ryan
        - [ ] ExponentTransformer -- Ryan
        - [ ] SqrtTransformer -- Ryan
        - [ ] ThetaLines -- Guzal
        - [ ] Columnwise -- Guzal
        - [ ] TabularAdapter -- Guzal
        - [ ] Detrender -- Svea
        - [ ] Deseasonalizer -- Svea
        - [ ] ConditionalDeasonalizer -- Svea

GuzalBulatova:
Working on 
[] transformations
     [] Theta -- Guzal

[...]

PR #1322, opened 2021-08-17
Update ColumnwiseTransformer and TabularToSeriesAdaptor docstrings

GuzalBulatova:
#### Reference Issues/PRs
References #1245.

Updates the docstrings for `TabularToSeriesAdaptor` and `ColumnwiseTransformer` to satisfy the documentation requirements.

No

Are there other sections that can be added?

### DIFF
--- a/docs/source/glossary.rst
+++ b/docs/source/glossary.rst
@@ -67,7 +67,7 @@ sktime.
         In ``sktime`` reduction is used to allow one learning task to be adapted as a solution for an alternative task.
 
     Variable
-        Refers to some measurement of in terest. Variables may be cross-sectional (e.g. time-invariant measurements like a patient's place of birth) or
+        Refers to some measurement of interest. Variables may be cross-sectional (e.g. time-invariant measurements like a patient's place of birth) or
         :term:`time series`.
 
     Timepoint
--- a/docs/source/glossary.rst
+++ b/docs/source/glossary.rst
@@ -80,6 +80,6 @@ sktime.
         depending on the discipline and context.
 
     Tabular
-        Is a setting where each :term:`timepoint` of the :term:`univariate time series` being measured for each instance are treated as feature and
-        stored as a primitive data type in the DataFrame’s cells. E.g., there are N instances of time series and each has T timepoints, this would
-        yield a pandas DataFrame with shape (N, T), so N columns, T rows.
+        Is a setting where each :term:`timepoint` of the :term:`univariate time series` being measured for each instance are treated as features and
+        stored as a primitive data type in the DataFrame’s cells. E.g., there are N :term:`instances <instance>` of time series and each has T
+        :term:`timepoints <timepoint>`, this would yield a pandas DataFrame with shape (T, N): T rows, N columns.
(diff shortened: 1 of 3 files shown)


RNKuhns (commented):
This is a very minor point, in the TabularToSeriesAdaptor it could be good to link the terms "tabular" and "series" to the glossary. 

I'd change the phrase series to time series and link that to the term that already exists. Might have to add the term "tabular", but I think that will be helpful to people not familiar with our terminology.

docs/source/glossary.rst
```diff
@@ -67,14 +67,19 @@ sktime.
...
 
     Instance
         A member of the set of entities being studied and which an ML practitioner wishes to generalize. For example,
         patients, chemical process runs, machines, countries, etc. May also be referred to as samples, examples, observations or records
         depending on the discipline and context.
+
+    Tabular
+        Is a setting where each :term:`timepoint` of the :term:`univariate time series` being measured for each instance are treated as feature and
+        stored as a primitive data type in the DataFrame’s cells. E.g., there are N instances of time series and each has T timepoints, this would
+        yield a pandas DataFrame with shape (N, T), so N columns, T rows.
```
RNKuhns, line 85: (N, T) would be N rows and T rows.

docs/source/glossary.rst
```diff
@@ -67,14 +67,19 @@ sktime.
...
+        The point in time that an observation is made. A time point may represent an exact point in time (a timestamp),
         a timeperiod (e.g. minutes, hours or days), or simply an index indicating the position of an observation in the sequence of values.
 
     Instance
         A member of the set of entities being studied and which an ML practitioner wishes to generalize. For example,
         patients, chemical process runs, machines, countries, etc. May also be referred to as samples, examples, observations or records
         depending on the discipline and context.
+
+    Tabular
+        Is a setting where each :term:`timepoint` of the :term:`univariate time series` being measured for each instance are treated as feature and
```
RNKuhns, line 83: should be "treated as features" not "treated as feature"

docs/source/glossary.rst
```diff
@@ -67,14 +67,19 @@ sktime.
...
+        stored as a primitive data type in the DataFrame’s cells. E.g., there are N instances of time series and each has T timepoints, this would
```
RNKuhns, line 84: Include a cross-reference to the glossary terms "instance" and "timepoints". 

E.g. there are N :term:`instances <instance>` of the time series and each has T :term:`timepoints <timepoint>`

RNKuhns:
@GuzalBulatova I've added a few minor suggestions for the glossary term you added. 

Also, thanks for catching the typos in the existing glossary terms!

docs/source/glossary.rst
GuzalBulatova: Thanks Ryan! If I understand this correctly we store data in DataFrame in the way that: columns correspond to variables and rows are timepoints at which the values of the variables measured. 
So in this case the following is correct: 
`there are N instances of time series and each has T timepoints, this would yield a pandas DataFrame with shape (T, N): T rows, N columns.`

RNKuhns: I think thought we are storing with N instances in the rows and T timepoints in the column. In tabular (i.e. cross-sectional) format used in scikit-learn and elsewhere, the rows usually correspond to instances. In our tabular reduction we are treating each timepoint as a independent feature (ignoring the time connections inherit in much timeseries data). 

So based on that I think it would be: 

E.g., there are N instances of time series and each has T timepoints yielding a pandas DataFrame with shape (N, T), so N rows and T columns. 

You just had rows and columns switched at the very end.

fkiraly (approved):
looks good to me, thanks!

Issue #1202, opened 2021-07-25
Seven/five number summary transformer


fkiraly:
It appears we don't have a seven/five number summary transformer?

This should be implemented.

Also see discussion from #1063.

fkiraly:
This has been done by @RNKuhns in #1329.

PR #1329, opened 2021-08-18
Add summary transformer

RNKuhns:

While cleaning up docstrings I noticed that `sktime.transformations.series.summarize.py` currently includes a MeanTransformer that returns the mean of a time series. 

I've added a `SummaryTransformer` that provides additional summary functionality (wrapping pandas summary methods). 
`MeanTransformer` is then a special case that inherits directly from `SummaryTransformer`. 

I've also added these transformers to the API reference (they had not previously been there). 

No

Does it make sense to continue to have the `MeanTransformer` as shorthand for `SummaryTransformer(summary_function="mean")`?

It does not appear to be used elsewhere in ``sktime``, so I'd propose we deprecate it and remove it in a future release. This will give us fewer transformers to maintain.

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/transformations/series/summarize.py
+++ b/sktime/transformations/series/summarize.py
@@ -27,7 +27,7 @@ ALLOWED_SUM_FUNCS = (
 )
 
 
-def _series_summary(series_or_df, summary_function="mean"):
+def _series_summary(series_or_df, summary_function="mean", axis=0):
     """Calculate summary value of a time series.
 
     For multivariate data the summary is applied to each column.
@@ -40,6 +40,8 @@ def _series_summary(series_or_df, summary_function="mean"):
         One of pandas summary functions ("mean", "min", "max", "sum",
         "skew", "kurtosis", "var", "std", "sem", "nunique", "count") that is used
         to summarize each column of the dataset.
+    axis : int, default=0
+        Axis to apply the summary function along. Only used for multivariate data.
 
     Raises
     ------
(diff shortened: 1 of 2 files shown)


sktime/transformations/series/summarize.py
```diff
@@ -3,38 +3,219 @@
...
+class SummaryTransformer(_SeriesToPrimitivesTransformer):
+    """Calculate summary value of a time series.
+
+    Calculates scalar value when applied to a :term:`univariate time series`.
+    If input is :term:`multivariate time series` then Scalar values for each
+    column and returned as elements of a pd.Series.
+
+    Parameters
+    ----------
+    summary_function : str, default="mean"
```
fkiraly, line 120: I think this should allow to choose multiple, or all. Otherwise it's clunky to use.

fkiraly (changes requested):
Great addition, much needed!

The one major comment concerns how you specify the summary.
We should be able to choose any sub-set of the strings, and the result is then multiple summaries (primitives).

MatthewMiddlehurst:
Good addition I agree. Should help towards #1202 and #1063. 

As Franz commented, being able to input a list of strings would be an important capability for the transformer.

RNKuhns:
@MatthewMiddlehurst and @fkiraly that should be a relatively easy addition, since pandas can accomodate that through the "agg" method. 

Just to confirm, in the multiple input series, do we want the aggregate output (e.g. mean, median, etc) as rows with the multiple input series representing columns or do we want the series to be the rows and the aggregate output to to be the columns?

MatthewMiddlehurst:
Univariate case is pretty simple, have rows as cases and columns as transformed attributes i.e.

|                  | mean  | median |
| ----------- | ------------- | ------------- |
**Case 1** | mean 1  | median 1 |
**Case 2** | mean 2  | median 2 |
... | ...  | ... |
**Case _n_** | mean  _n_  | median  _n_ |

For multivariate series, just append on the values for each dimension I guess. 

|                  | mean_d1  | mean_d2 | median_d1  | median_d2 |
| ----------- | ------------- | ------------- | ------------- | ------------- |
**Case 1** | mean_d1 1  | mean_d2 1  | median_d1 1 | median_d2 1 |
**Case 2** | mean_d1 2  | mean_d2 2  | median_d1 2 | median_d2 2 |
... | ...  | ... | ...  | ... |
**Case _n_** | mean_d1  _n_  | mean_d2 _n_  | median_d1  _n_ | median_d2 _n_ |

fkiraly:
agreed - though I'd use double-underscores, that will comply with common conventions on converting strings into multi-indices

fkiraly:
also, there´s a conflict in the doc source

RNKuhns:
@fkiraly I've made the changes we talked about. But I'm running into an error standard transformer tests because the `_SeriesToPrimitivesTransformer` expects the output to be a primative or NumPy array of primatives. 

Per our conversation I've got the output as a `pd.DataFrame` regardless of case.

fkiraly:
Hm, how about adding `pd.DataFrame` to the types it can be, in the test?

RNKuhns:
@fkiraly I'm up for that and indicate the potential changes to the tests below. But doesn't this violate the potential output types of the `_SeriesToPrimitivesTransformer` class where the transform indicating a primative or np.ndarray of primitives is returned.

There is logic based on the `SeriesToPrimitivesTransformer` in `sktime\transformations\panel` and we probably need to make sure that works with this too. 

I guess from my perspective, I still have a couple questions about what the transformations module will look like post refactor. Will we still have separate `panel` and `series` submodules? If so, is a transformer in the `series` submodule supposed to only work on univariate and multivariate data that is non-panel? 

Are we relying on a composite class to apply `series` transformers to each instance (row) of a panel in all cases or do we expect some transformers to live in both places (e.g. SummaryTransformer in `sktime.transformations.panel` and sktime.transformations.series`). It seems to me that most transformers could apply to univariate, multivariate and panel data depending on the strategy you use to apply them (e.g. apply univariate transformer to each column in multivariate data, or each instance and each column in panel data, etc). 

I'm just trying to wrap my head around the end state, so I can figure out how to proceed with this. 

**Checks that would need to be updated**
First I'd need to add `pd.DataFrame` to list of primitive types in univariate check.
```python
def check_series_to_primitive_transform_univariate(Estimator, **kwargs):
    out = _construct_fit_transform(Estimator, **kwargs)
    assert isinstance(out, (int, np.integer, float, np.floating, str)) # Would need to pd.DataFrame here
```

Likewise, the following would need to be changed in the spot listed in the comments:

```python
def check_series_to_primitive_transform_multivariate(Estimator):
    n_columns = 3
    if Estimator.get_class_tag("univariate-only", False):
        _check_raises_error(Estimator, n_columns=n_columns)
    else:
        out = _construct_fit_transform(Estimator, n_columns=n_columns)
        assert isinstance(out, (pd.Series, np.ndarray)) # Add pd.DataFrame here
        assert out.shape == (n_columns,)
```

fkiraly:
@RNKuhns, yes, I'd still be supportive of that change.

RNKuhns:
@fkiraly I'm down to add that change. The other thing I want to clarify is where this should go. As coded, this is the series-to-primitives transformer and under `sktime.transformations.series`. 

For the panel-to-primitives version, do I need to add a panel compatible version (super easy using group by functionality in pandas) in `sktime.transformations.panel`?

Issue #1342, opened 2021-08-21
[BUG] test_examples checks are failing, binder is also failing


fkiraly:
I noticed the `test_examples` checks are failing with

`module compiled against API version 0xe but this version of numpy is 0xd`

This looks like a `numpy` compatibility issue - it's mysterious since I thought the changes in the last week's PR merges were largely documentation related and did not touch CI/CD.

@freddyaboulton, @corvusrabus, @mloning, @aiwalter, any idea what is going on there?

fkiraly:
looks like binder *also* broke - FYI, @freddyaboulton, @corvusrabus - might be a `numba` update, since setting `numba==0.53` fixes the binder.

fkiraly:
merged PR #1343 which fixes this, though we still want to understand what happened - probably a new `numba` version broke things?

PR #1343, opened 2021-08-21
bugfix - broken binder and test_examples check

fkiraly:
An attempt to address #1342 by reverting `numba` and `numpy` version bounds to pre #1266 values


fkiraly:
Looks like it fixes both the checks and the binder, so I'll emergency-merge.

Issue #1268, opened 2021-08-09
forecasting: selecting exogeneous variables


fkiraly:
I think there should be wrappers that allow to select which exogeneous variables to use, which is useful in cases where forecasters behave differently if they get exogeneous variabes or not. Notably, a large number of "garbage variables" can hamper performance, so selection of exogeneous variables may be helpful, and enables automated selection, too.
 For instance, useful wrappers could be:

* `IgnoreExogeneous` which blocks any `X` from being passed.
* `SelectExogeneous` which takes a list of column names, and only passes those ones through.

The `SelectExogeneous` can also be used in combination with tuning for automated feature selection.

RNKuhns:
@fkiraly would the idea also be to eventually add functionality for automated exogenous variable selection methods?

aiwalter:
I just wanted to open an issue when I saw this one :) so I rather write my idea here:
___

**Is your feature request related to a problem? Please describe.**
I would like to add a transformer that can selection features based on feature based on feature importance values of a meta-model/regressor.

Related: #1268
**Describe the solution you'd like**
```python
SelectExogeneous(method="feature_importances", regressor=DecisionTreeRegressor(), n_features=12, max_features=None, random_state=None)
```
where `regressor` is any `sklearn-like` tree-based model with an attribute `feature_importances_`. The regressor would be fitted on the `y` as endog and `X` as exog. The regressor could even be a `GridSearchCV` if user want to tune the meta-model.

Other methods that come to my mind:

- Random sampling method where features are sampled randomly. 

But I would start with the proposed solution.
**Additional context**
- pipelining
- tree-based models
- panel/global forecasting

@fkiraly do you agree that this class should be implemented as a `_SeriesToSeriesTransformer`? Bcz it will be part of tuning so it could be tuned in a `ForecastingPipeline`

aiwalter:
I have already started implementing a transformer, but now I faced issues with the transformer interface. So the issue is, that the regressor would need `y` and `X` so also the transformer needs both then. But there is no such case in the transformer interface where `y` and `X` is given and `Xt` is returned. See STEP [here ](https://github.com/sktime/enhancement-proposals/blob/main/steps/01_forecasting_api/forecasting_pipeline.md) where we wrote down the cases.

But the order is switched. So the transformer would get the data as follows:
```python
FeatureSelection(...).fit_transform(Z=X, X=y)
```
[...]

PR #1347, opened 2021-08-22
Added FeatureSelection transformer

aiwalter:
Added FeatureSelection transformer

#### Reference Issues/PRs
#1268 

##### For all contributions

##### For new estimators


aiwalter:
Todo: add method `ignore` to set `Z` to `None`

fkiraly:
I think @ViktorKaz already has this in his PR on the #856 , so perhaps good to align?

aiwalter:
Oh didnt see, but seem my implementation is a superset as it has more methods/algos. @ViktorKaz whats your opinion? Not sure as you implemented a `_PanelToPanelTransformer`?

fkiraly:
oh, I see - you do automated feature selection, so it's not just a subsetter. Please ignore my comment then.

ViktorKaz:
@fkiraly @aiwalter the transformer that I implemented is quite simple, it just selects columns based on the column indices. From what I quickly saw `sktime.transformations.series.feature_selection.FeatureSelection` can also select features by importance or at random which is nice and in this respect is better than my transformer.

I implemented a `_PanelToPanelTransformer` because I needed a transformer that takes a pandas dataframe as input and returns a pandas series. From memory, I initially tried implementing `_PanelToSeriesTransformer` (I suppose this is the correct type) but ran into some issues with the unit tests.

Based on a quick look at  `sktime.transformations.series.feature_selection.FeatureSelection` I don't understand why `Z` can be a pd.Series. If the input has only one column (feature) what is there to select?

aiwalter:
okay, so then we can have both of this transformers if I understand it correctly. Pls feel free to propose other feature selection methods I could add to this transformer.

@ViktorKaz I accept only `pd.DataFrame`, was already in the code but wrong in docstring, updated it now. Thx!

aiwalter:
@fkiraly I think we need another tag `multivariate-only` as equivalence to `univariate-only`? Tests are failing as the `FeatureSelection` transformer can only deal with `pd.DataFrame`. Other solution is that in case it gets a `pd.Series` it would just pass this through. What do you think?

aiwalter:
@fkiraly I am struggling a bit with the tests as I have a line `Z = check_series(Z, enforce_multivariate=True)`, so some tests are failing here. I am wondering if that should actually be ignored then by the tests? I though tests accept some exceptions?

fkiraly:
I didn't touch *transformers* in relation to the univariate/multivariate discussion.
Probably we need to make case distinctions, and possibly even introduce a tag for transformers that require multivariate input?

If that's the "deeper problem" here, would you mind adding this as part of the PR?

Any opinions, @mloning, @GuzalBulatova, @thayeylolu?

aiwalter:
can I use the `scitype:y": "multivariate` to enforce multivariate? Or do we need a `multivariate-only` tag?

fkiraly (changes requested):
Ok, here's my opinion on the test logic.

I think it's in general a good idea to change the tests so they run with multivariate series, and it's much needed. However, I think the way you are doing it is not good.

What I think is not a good idea:
* you are modifying the signature of all tests, to pass some side information in (`n_columns`). That's quite some heavy change there in the architecture.
* You are, in fact, passing the same information twice - once as the `scitype:y` tag (of the `Estimator` argument), and once as the `n_columns` argument. That's redundant. Why don't you parse the tags just in the sub-tests where it is needed, instead of parsing it in the "master test" and then passing it on? This is how it is done for controlling test behaviour in other cases, too.
* the tag `scitype:y` also seems confusing, because there's no `y` in the transformers? I suggest to use a tag, but not this name for it.

My proposed change to this would be:
* leave the test signatures as they were, only taking the estimator
* in the tests where you need `n_columns` to be determined by the estimator, use the tag.
* Stick to the general pattern that's currently used, if you need to control test behaviour, use the tags. I don't think there's a case where we can't do this, right?
* minor note, if you introduce a new tag or use it for an an estimator scitype it hasn't been used before, you ought to update the `_tag` registry

aiwalter:
> Why don't you parse the tags just in the sub-tests where it is needed, instead of parsing it in the "master test" and then passing it on? This is how it is done for controlling test behavior in other cases, too.

I did this in order to avoid copy and pasting if statements like `if Estimator.get_class_tag("scitype:y") == "multivariate":`, bcz this is what I would need to do for almost any test then.

> the tag scitype:y also seems confusing, because there's no y in the transformers? I suggest to use a tag, but not this name for it.

Ok but then I need to introduce a ` scitype:Z` tag

fkiraly:
>I did this in order to avoid copy and pasting if statements like if Estimator.get_class_tag("scitype:y") == "multivariate":, bcz this is what I would need to do for almost any test then.

In `test_all_transformers`, it's two out of nine, in `estimator_checks` it's about half, that's less than the majority, not "almost any".

But I get what you are trying to do, I would suggest one of the two:

* do it in `_get_args`, that way you don't change the general test signature but only how the args are created
* I anticipate that we may want to check multiple args, e.g., `n_columns=1` and `n_columns=2` depending on the various tags. You could do a bit of a refactor by creating a `_yield_args`, and then looping over all scenarios, though that's perhaps too much of a refactor. 

>Ok but then I need to introduce a `scitype:Z` tag

That would be fine with me.

aiwalter:
but why `_get_args`? I guess you mean `_make_args`?

fkiraly:
>but why `_get_args`? I guess you mean `_make_args`?

yes, that was a typo - I mean `_make_args`.

fkiraly:
Ok, I've thought about this some more.

I think we should apply the general principle that we have in our governance docs: if there's a major refactor of APIs, there ought to be a STEP.

Here's what I think:
* changing signature of tests is a major change - who knows what this might break down the line, so it needs a description of what is changing in the tests API, and @mloning also needs to have a look at that. I would also do this in a separate PR.
* Changing the *logic* of individual tests is not a major change though, an I'd say it's fine as long as it's properly documented what, say, `_make_args` does.

sktime/forecasting/base/_meta.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, forecasters, n_jobs=None):
 
     def _check_regressor(self):
         if self.regressor is None:
-            self.regressor_ = GradientBoostingRegressor(max_depth=5)
+            self.regressor_ = GradientBoostingRegressor(max_depth=5, random_state=1)
```
mloning, line 35: Shouldn't we pass a random state here which a user can define in the constructor?
```suggestion
            self.regressor_ = GradientBoostingRegressor(max_depth=5, random_state=self.random_state)
```

sktime/transformations/series/feature_selection.py
```diff
@@ -0,0 +1,187 @@
...
+        # multivariate Z
+        if not isinstance(Z, pd.Series):
+            if self.method == "feature-importances":
+                self._check_regressor()
+                self._check_n_columns(Z)
+                X = check_series(X)
+                # fit regressor with Z as exog data and X as endog data (target)
+                self.regressor_.fit(X=Z, y=X)
+                # create dict with columns name (key) and feauter importance (value)
+                d = dict(zip(Z.columns, self.regressor_.feature_importances_))
```
mloning, line 121: What happens if `feature_importances_` isn't an attribute of the `self.regressor_`? I saw that you describe that in the docstring which is great but perhaps we could check for that an raise an error message?

mloning (commented):
Hi @aiwalter had a quick look, just two minor comments.

sktime/forecasting/base/_meta.py
aiwalter: I was not sure on this, I thought if user want to set seed, a `regressor` can be given with the respective params. But let me add a param `random_state ` as a shorthand

sktime/transformations/series/feature_selection.py
aiwalter: yes, let me add a check for this to raise a proper exception message

aiwalter:
@mloning thx for review. Did some changes and would appreciate to have it in next release

mloning (dismissed):
Thanks @aiwalter, looks good to me now.

aiwalter:
@fkiraly pls review

sktime/forecasting/trend.py
```diff
@@ -207,6 +207,9 @@ def _predict(self, fh=None, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
         y_pred_int : pd.DataFrame
             Prediction intervals for the forecast
         """
+        if return_pred_int:
```
fkiraly, line 210: please revert this - it has nothing to do with the PR, and the `NotImplementedError` is dealt with by the base class

sktime/tests/_config.py
```diff
@@ -332,6 +333,7 @@
     Imputer: {"method": "mean"},
     HampelFilter: {"window_length": 3},
     OptionalPassthrough: {"transformer": BoxCoxTransformer(), "passthrough": True},
+    FeatureSelection: {"method": "all"},
```
fkiraly, line 336: please add this to `get_test_params` instead, new policy on where the defaults are 😄

fkiraly: tests use `get_test_params` which (for downwards compatibility) looks in `_config`, but the primary location for the parameters is in `get_test_params` now

fkiraly (changes requested):
I think this is great and almost ready.

But, there are some minor things which I think should be fixed before merging:

* adding superfluous tag handling in `TrendForecaster` which is out of scope and a duplication
* default test params should move in the designated method from `_config`

sktime/forecasting/trend.py
aiwalter: I think I can remove this `raise` completely then, rather then reverting?

aiwalter: ok I think this is resolved now as I removed it completely, which seems to be the `main` state as well.

aiwalter:
@Aparna-Sakshi I have a quick question regarding your PR #1361. My tests are failing, seems bcz I have different params in `get_test_params` than in the default `init`. But this should in my opinion be allowed? Pls correct me if this issue is not coming from the `get_test_params`

also tagging @fkiraly

fkiraly:
I think this is perhaps a logic error in the test `check_constructor`?
Or, it is an undocumented requirement for the test instance?
Namely, the test assumes that the first test instance's parameters are equal to the defaults, otherwise it raises an error.

This logic did not change when we made the interface change - you can check by reverting to an earlier version and making the same change in `_config` .

fkiraly:
My suggestion would be: for this PR, be compliant with the usual test logic where it is assumed that the `_config` parameters do not overwrite any defaults.

In a separate PR, we update the test logic to:
* allow overrides of default parameter values
* loop through all elements of `get_test_params` if it is a list.

fkiraly:
@aiwalter, actually, it was a bug caused by a remaining reference to the `_config`, in the failing test.

This is now fixed in #1406, have a look.

aiwalter:
Okay thx. Then let me try again with the new way

fkiraly:
While you are at it:

I started refactoring the tests since I thought it might be good to have multiple parameter combinations tested per estimator, and that was not possible with the rigid, non-pytest way the tests and fixtures were looped over.

I've fixed this now and would appreciate your review of #1407.

aiwalter:
@fkiraly should `get_test_params` not be private `_get_test_params`?

mloning:
+1 for making them private

sktime/transformations/series/feature_selection.py
```diff
@@ -0,0 +1,196 @@
...
+        super(FeatureSelection, self).__init__()
+
+    def fit(self, Z, X=None):
+        """Fit the transformation on input series `Z`.
+
+        Parameters
+        ----------
+        Z : pd.DataFrame
+            A time series to apply the transformation on.
+
```
mloning, line 101: docstring for X is missing

sktime/transformations/series/feature_selection.py
```diff
@@ -0,0 +1,196 @@
...
+                    raise AttributeError("Parameter columns must be given.")
+                self.columns_ = self.columns
+            elif self.method == "none":
+                self.columns_ = None
+            elif self.method == "all":
+                self.columns_ = list(Z.columns)
+            else:
+                raise ValueError("Incorrect method given. Try another method.")
+
+        self._is_fitted = True
```
mloning, line 149: This is handled in the base class
```suggestion

```

sktime/transformations/series/feature_selection.py
```diff
@@ -0,0 +1,196 @@
...
+        Returns
+        -------
+        self
+        """
+        Z = check_series(Z)
+        self.n_columns_ = self.n_columns
+        self.feature_importances_ = None
+
+        # multivariate Z
+        if not isinstance(Z, pd.Series):
```
mloning, line 111: for readability it may be easier to read:
```suggestion
        if isinstance(Z, pd.Series):
            return self
        else:
```

sktime/transformations/series/feature_selection.py
```diff
@@ -0,0 +1,196 @@
...
+            else:
+                Zt = Z
+        return Zt
+
+    def _check_n_columns(self, Z):
+        if not isinstance(self.n_columns_, int):
+            self.n_columns_ = int(math.ceil(Z.shape[1] / 2))
+
+    @classmethod
+    def get_test_params(cls):
```
mloning, line 187: Shouldn't this be implemented in the base class? The test parameters should then be a class variable similar to the tags.

mloning (commented):
Hi @aiwalter, I had a quick look, just some minor comments.

sktime/transformations/series/feature_selection.py
aiwalter: I couldnt find how to actually implement this. It says its in the extension template but there I cant find it.

fkiraly: that's because we separated the refactor (moving logic) from adding the documentation in the extension templates.

The extension template PR is still a draft and under review here:
https://github.com/alan-turing-institute/sktime/pull/1395

The docstring should say though what it should do:
return either a dict of params or a list of dict; in either case, the dicts must be suitable for passing to the init, i.e., `__init__(**mydict)` must result in a sensible test instance for the class.

fkiraly: FYI @Aparna-Sakshi

fkiraly: (would be nice to finish the extension template PR 😃 )

fkiraly: besides, you can still use `_config` as before if it would be a blocker for this PR

aiwalter: yes will just use `_config` to get this PR done :)

fkiraly: Well, if your line in `_config` looks like this:
`FeatureSelectionTransformer: {"foo": "bar"}`

then the content of `get_test_params` should look like this:
`return {"foo": "bar"}`

Should be straightforward...

fkiraly:
>Shouldn't this be implemented in the base class? The test parameters should then be a class variable similar to the tags.

@mloning, why should it be in the base class? The parameters will be different per concrete class.

>The test parameters should then be a class variable similar to the tags.

I disagree with the design, since that does not allow you to have code logic that determines multiple parameter settings, instead you have to write everything down manually.

mloning:
> I disagree with the design, since that does not allow you to have code logic that determines multiple parameter settings, instead you have to write everything down manually.

Why, you can always make test params a list and iterate over it? Implementing the method with full docstring if you only return some values seems like a lot of duplication.

fkiraly:
>Why, you can always make test params a list and iterate over it? Implementing the method with full docstring if you only return some values seems like a lot of duplication.

Sorry, I think you have not understood and/or I have not properly communicated what I mean.

I am referring to the case where it ismuch  easier not to paste the list into the method, as a list, but instead code that generates the list. Let's say you want to use a 5 x 5 x 5 test grid of parameters. Would you want to paste a list of length 125, typing it out, or write two lines of code that loop through the set?

Further, making it a method, you can inside call `get_test_params` or `create_test_instance` of any estimator components, thus modularizing your tests and ensuring isolation of test content to the class - e.g., if the default of the component changes, the default of the composite changes with it, ensuring that you can localize bugs to the class, which would not be the case in your solution.

mloning (approved):
Looks all good to me. We can update the test params in a separate PR I think.

mloning:
> Sorry, I think you have not understood and/or I have not properly communicated what I mean.
> 
> I am referring to the case where it ismuch easier not to paste the list into the method, as a list, but instead code that generates the list. Let's say you want to use a 5 x 5 x 5 test grid of parameters. Would you want to paste a list of length 125, typing it out, or write two lines of code that loop through the set? And, which of the two options is easier to read, understand, and debug?
> 
> Further, making it a method, you can inside call `get_test_params` or `create_test_instance` of any estimator components, thus modularizing your tests and ensuring isolation of test content to the class - e.g., if the default of the component changes, the default of the composite changes with it, ensuring that you can localize bugs to the class, which would not be the case in your solution.

Is there a separate issue for this to discuss this in more detail?

fkiraly:
>@fkiraly Is there a separate issue for this to discuss this in more detail?

Yes, of course:
https://github.com/alan-turing-institute/sktime/issues/1098

The generic issue was brought up on July 3, and the step exists since late August

fkiraly (approved):
Yes. looks good to me too now.

PR #1349, opened 2021-08-23
Updates ColumnEnsembleForecaster with "copy forecaster for all columns" functionality

fkiraly:
This extends `ColumnEnsembleForecaster` so the same forecaster is cloned across all columns if not a list of forecasters is passed, but just a single forecaster, as the `forecasters` parameter.

This is achieved by a change in the `_check_forecasters` that is called at the start of `_fit`, which, at the very start, replaced a single forecaster with clones per column in the old expected format.

We ought to merge this quickly, since the current signature of `ColumnEnsembleForecaster` prevents upgrading the forecaster tests to the multivariate case, which in turn is blocking VAR.

FYI @thayeylolu.

### DIFF
--- a/sktime/tests/_config.py
+++ b/sktime/tests/_config.py
@@ -159,7 +159,6 @@ TIME_SERIES_CLASSIFIERS = [
     ("tsf2", TIME_SERIES_CLASSIFIER),
 ]
 FORECASTER = ExponentialSmoothing()
-COLUMN_ENSEMBLE_FORECASTER = [("naive", NaiveForecaster(), 0)]
 FORECASTERS = [("ses1", FORECASTER), ("ses2", FORECASTER)]
 STEPS_y = [
     ("transformer", Detrender(ThetaForecaster())),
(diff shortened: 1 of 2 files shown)


fkiraly:
@aiwalter, would you mind giving this a quick review? It would resolve some blockers for @thayeylolu

sktime/tests/_config.py
```diff
@@ -170,7 +169,7 @@
     ("forecaster", NaiveForecaster()),
 ]
 ESTIMATOR_TEST_PARAMS = {
-    ColumnEnsembleForecaster: {"forecasters": COLUMN_ENSEMBLE_FORECASTER},
+    ColumnEnsembleForecaster: {"forecasters": NaiveForecaster()},
```
aiwalter, line 172:
```suggestion
    ColumnEnsembleForecaster: {"forecasters": FORECASTER},
```

aiwalter (approved):

fkiraly:
ah, got a bug in there, didn´t notice in the local tests. Should be fixed now ... I hope ...

aiwalter:
ok, no worries

fkiraly:
should be resolved now

aiwalter (approved):

thayeylolu (commented):
Looks great. would merge with the two branch (1083 and 1301)

fkiraly:
@thayeylolu, it's merged, so you can use it now.

Issue #1346, opened 2021-08-22
[ENH] ClaSP (Time Series Segmentation paper in CIKM '21)


patrickzib:
**Is your feature request related to a problem? Please describe.**

I would like to add ClaSP (CIKM '21), a state-of-the-art method for time series segmentation (similar to FLOSS) and a SeriesToSeriesTransform to sktime.

**Describe the solution you'd like**

ClaSP create a single profile per time series. However, it would be nice to also return some additional information computed along ClaSP: 

Thus, I am currently struggling with the definition of:
`_SeriesToSeriesTransformer`

As a reference, I had a look at matrix_profile, which returns:
`return pd.Series(Xt[:, 0])`

Sorry, if I am missing out on something: Is there any means to return a tuple from transform?
`return profile, k_nn_distances, ...`

fkiraly:
Great idea, though it should be an annotator, not a transformer - segmentation is a case of annotation..

have a look at the extension template here:
https://github.com/alan-turing-institute/sktime/blob/main/extension_templates/annotation.py

If you want to return more than the segmentation, you can use `get_fitted_params` where you can return anything you'd like.
I've noticed it's not in the annotation template, but the format is as in the [forecasting template](https://github.com/alan-turing-institute/sktime/blob/main/extension_templates/forecasting.py), it simply should return a dict and can return anything you'd like.

For distances, you may want to consider implementing a separate distance transformer, using the [time series distance/kernel extension template](https://github.com/alan-turing-institute/sktime/blob/main/extension_templates/dist_kern_panel.py).

patrickzib:
Thank you very much! That was very helpful!

I am now trying to provide a Jupyter-Notebook alongside the code. However, when I import sktime packages, these are imported from my Anaconda sktime installation. Is there any way to import from my developer version instead?

patrickzib:
@fkiraly This seems to be a workaround for my problem. 

```
import sys
sys.path.insert(0,"..")
```

Is there any better way?

fkiraly:
yes, `sys.path.insert` and `sys.path.append` works - the other option is `pip install -e [dir location]`

Also see here: https://www.sktime.org/en/latest/installation.html
(section: development versions)

patrickzib:
@fkiraly 
```
have a look at the extension template here:
https://github.com/alan-turing-institute/sktime/blob/main/extension_templates/annotation.py

[...]

PR #1352, opened 2021-08-24
Add Clasp for time series segmentation (CIKM'21 publication)

patrickzib:
#### Reference Issues/PRs
#1346 

This request implements ClaSP, a novel time series segmentation method published at CIKM '21. 

The segmentation is implemented as an annotator, and the profile in ClaSP is implemented as a SeriesToSeriesTransformer.


patrickzib:
We would like to add a novel time series segmentation algorithm (ClaSP). 

I did the following in this pull request:
1. Introduce a ClaSP BaseSeriesAnnotator for segmentation
2. Introduce a ClaSP SeriesToSeriesTransformer for creating our profile
2. Add two datasets for segmentation based on the UCR archive
3. Introduce a new tag 'fit-in-predict', as ClaSP performs segmentation on each single time series and does not train on a whole dataset.
4. Add a Use Case (JupterNotebook)
5. Add a Test Case 

Did I miss on something? :)

@TonyBagnall @MatthewMiddlehurst @mloning can you review the code, if everything is ok?

Thank you :)

patrickzib:
> Thinking about it more carefully, we could do it as follows:
> 
> * let's target the end state where we have a "composable" ClaSP, and a "default" ClaSP that reproduces results in your paper and is optimized for the choices in your paper.

We should keep Arik Ermshaus (co-author) in the loop for the composable ClaSP. He did his master thesis on this topic. 

> * this PR is about the "default" ClaSP, i.e., with knn inside. Let's target getting the "default" one merged first, then add a "composable" ClaSP in a separate PR which can use random forest etc. I'll review in detail with the first target in mind, specifically.
> * Then we can take on more segmenters, and aim being able to run the benchmarking experiment in your paper.
> 
> What do you think?

Sounds great to me :) There are some research issues Arik faced when working on the  composable ClaSP:

1. It can be slow to fit and predict the same model n times, which we avoid by the relabelling of the k-NN in the "default" ClaSP. Arik did some research on this in his Master thesis using either sampling the profile but also using other cross-validation schemes to avoid fitting the classifier too often.
2. We still lack a proper benchmark that shows benefits of using different classifiers in ClaSP. Mainly as the scores on the FLOSS benchmark are very low already for the "default" ClaSP but also for FLOSS. 

> For the reducer:
> 
> * I don't think we should inherit from the forecasting reducer, I would just make it a new class, an annotator
> * the annotator takes two estimator arguments in the init, namely the classifier used in the prediction, and the transformer to create the labels. These are then used inside the algorithm.

Great, I think that is what I did. Can you check? :)


> We could also set up a meeting to consolidate the segmentation wishlist into a roadmap.

Sure, maybe we should plan a zoom-meeting with Arik to discuss things. I do not want to post everything on our research plan in this PR though - but we may share it personally but also join efforts, if you like.

TonyBagnall (changes requested):
sorry to be that guy, but the comments are not in the format we are trying to standardize to. Look at TDE for the format of parameter comments https://github.com/alan-turing-institute/sktime/blob/main/sktime/classification/dictionary_based/_tde.py

patrickzib:
> sorry to be that guy, but the comments are not in the format we are trying to standardize to. Look at TDE for the format of parameter comments https://github.com/alan-turing-institute/sktime/blob/main/sktime/classification/dictionary_based/_tde.py

Sure :) Fixed the comments.

patrickzib:
> sorry to be that guy, but the comments are not in the format we are trying to standardize to. Look at TDE for the format of parameter comments https://github.com/alan-turing-institute/sktime/blob/main/sktime/classification/dictionary_based/_tde.py

sorry, forgot to close? fixed :)

patrickzib:
@TonyBagnall not sure: how do I change your change request to resolved on github?

MatthewMiddlehurst:
Don't think you can unless there are specific lines highlighted, Tony just has to re-review. 

Even if there were specific changes, we had some trouble on the SFA pr, think you may have to be added to the repo to do so. Similar to the codeowners stuff, where you can have your name on the file, but it won't actually notify some people even if their file is changed.

TonyBagnall (approved):
sorry patrick, all looks good now

fkiraly (changes requested):
Thanks for a nice contribution!

I think we need to do some housekeeping before merging this though.

I have a few major comments:

* the algorithm, as currently presented, is not a segmentation algorithm but a change-point detection algorithm. The reason is that it does not output segments (in any encoding), and/or segment labels, but point annotations - the change points.
* hence, I would not call it segmentation, but change point detection. And/or explain how it is closely related to segmentation (unlabelled, non-overlapping, continuous).
* The notebook should explain change point detection in general - i.e., we have a series, we want to identify change points. 
* the notebook contains a lot of custom functionality for plotting, which (a) is great and nice but (b) does not belong in a notebook, but a utility folder, e.g., `sktime/annotation/plotting`. I think we should provide the functions as generic plotting functionality fore change point detection.
* I would suggest to produce sub-plot functions, i.e., one for the top plot and one for the bottom plots, which can be used for any change point results (not just from clasp). Adopt the matplotlib signature `ax = plot_sth_sth(ax, input, input2, params)`. These are then imported in the notebook and used.

fkiraly:
(@TonyBagnall, kindly don't merge until we have resolved this)

fkiraly:
@patrickzib, perhaps it's easiest to clarify the "target state" in a quick call?

As I said, this is a very valuable contribution, there's just a little bit of cleaning to do, i.e., where code goes, what type the algorithm has, etc.

fkiraly:
& apologies for the time it took to get to this review, it's a lot of new stuff in a new area (annotation), also conceptually.

patrickzib:
@fkiraly no worries and thanks for your comments. I will adress these asap. Currently my daughter is ill and I have to look after her. :/

I thought about your comment regarding Segmentation and Change Point Detection. 

Our competitor Floss defines (semantic) segmentation as:

`Definition 4: Given a time series T, monitoring some aspect of a system S, infer the boundaries b between changes of state.`

In analog we define it as:

`Definition 4. A segmentation of a TS 𝑇 into 𝐶 + 1 segments is an ordered sequence of change points (or splits) 𝑡𝑖1 ,...,𝑡𝑖𝐶 with 1 < 𝑖1 < · · · < 𝑖𝐶 < 𝑛. `

As such, I dont want to call ClaSP an outlier detector. As both Floss and ClaSP inferr ordered lists of change points (CPs). From CPs we can easily get the self similar segments

Thus, maybe we should add another method
`predict_and_segment()` which not only returns change points but segments? And keep predict to return CPs.

what do you think?

fkiraly:
@patrickzib, regarding the "type of algorithm" discussion:

* you mention "outlier detector" - I didn't claim clasp is that, and it wouldn't make sense to call it as such, since outliers are "points different from their context" which the points you identify are not.
* I see that the base class, for some reason, always refers to "outliers" - this is wrong and needs to be changed, since annotations need not be outliers. Not sure where this is coming from, but it explains why you refer to an "outlier detector". We need to change this.
* in the `sktime` design, a segmentation formally corresponds to a `pd.IntervalIndex`, which is a format that records segment starts/ends. Segments could be overlapping and not cover the full space.
* On the scientific level, I think it is conflation to *define* a segmentation as a sequence of change points! You can *obtain* a segmentation by considering subsequent change points as start/end, but it's not the same, since a segmentation can be overlapping and non-covering.
* On the code level in `sktime`, it's determined by the return type: change points are a simple `pd.Series` sequence (with change points denoted by index occurrence or value indicator). Segmentations are `pd.IntervalIndex`. We can call it a segmentation algorithm only with the right return type.

Reading the above, I think clasp can be used *both* for segmentation and change point detection, but the return types differ. I think clasp should be able to return either, in the specified format.

We have not designed the base class for algorithms that can do multiple things, but here's what I recommend to address this:

* You add an `__init__` argument `task` which can be either `"segmentation"` or `"change-points"`. If `"segmentation", `predict` give a segmentation return; if `"change-points"`, `predict` gives a change point return.
* The specification for these is as follows (look at the base class, `BaseSeriesAnnotator`): if `fmt` is dense, the return index is exactly as the input index of the series in `predict`. For `segmentation`, each segment should have a separate label (categorical). For `change-points`, the value should be bool (change-point yes/no) if `labels=indicator` and float (change-point score) if `labels=score`.
* if `fmt` is sparse and `task="segmentation`, the return should be a `pd.IntervalIndex` of the segments. If `task="change-points`, the return should be a `pd.Series`, with indices being only the change points. If `labels=indicator`, the values are all `True`; if `labels=score`, it should be the change-point scores.
* I'll refactor the base class and create an extension template that better reflects what to do if an algorithm can do multiple things, and to remove the confusing references to "outlier detection".

At least that's how it currently is, in a combination of the segmentation/change-point specification.

The module is still experimental, so if you have better suggestions, it's negotiable.

patrickzib:
@fkiraly sorry, I meant ClaSP is not `Change Point Detection` but wrote `Anomaly Detection`. Sorry for mixing terms, but I am a bit distracted as mentioned above. :)

I feel like it is _not_ a good idea to mix both topics CPD [2] and Segmentation [3]. As I see it _scientifically_, `CPD` is a different research area, which is mostly about hypothesis testing. The problem with using Segmentation algorithms for CPD, to my understanding, mostly boils down to detecting stationary and non stationary signals.

FLOSS has the following paragraph on this:

> 2.3 What FLOSS is not
> Even within the narrow context of time series analytics, the term segmentation is overloaded; thus, it is necessary to explicitly explain some tasks we are not addressing. Change point detection is a method for detecting various changes in statistical prop- erties of time series, such as the mean, variance or spectral density. A helpful review of the literature on this problem is surveyed in detail in a recent paper (Aminikhanghahi and Cook 2017). In contrast to change point detection, we are interested in regimens which are defined by changes in the shapes of the time series subsequences, which can change without any obvious effect on the statistical properties. Consider the following pathological example. Suppose we took an hour of an normal electrocardiogram, and appended to it a reversed copy of itself (to be clear, the discrete analogue of this is the production of the palindrome..beatbeatbeattaebtaebtaeb..). While such a time series would have a visually obvious (indeed, jarring) transition at the halfway point, virtu- ally all change point algorithms that we are aware of would ignore this transition, as the features they consider (mean, standard deviation, zero-crossings, autocorrelation etc.) are invariant to the direction of time. Clearly, one can also create pathological datasets that would stymie our proposed algorithm but be trivial for most change detec- tion algorithms. In other words, they are only superficially similar tasks that do not
> directly inform each other.

Scientifically, `Segmentation` is a different topic, which derives `Change Points`, but is not suitable for `CPD` - for example for stationary vs non stationary TS, for which both ClaSP and FLOSS fail miserably [1]. My colleague does research on bridging that gap, though.

From a `sktime` framework point of view, it might be tempting to mix both, but then you might also be tempted to use a `regressor`, discretize its output, and call it a `classifier` - even if this is a bit over-simplified ;)


I think, `pd.IntervalIndex` is a great idea and we can use a dense or sparse representation. 

A categorical label for a segmentation is also a nice idea - finding repeating segments is currently a hot topic, and both ClaSP and FLOSS would simply assign a new label for each segment rather than detecting something like: A B C A A B C A A, with Segments A B C

CPD Benchmarks:
[1] Turing Change Point Detection Benchmark: https://github.com/alan-turing-institute/TCPDBench
[2] A Survey of Methods for Time Series Change Point Detection https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5464762/
[3] FLOSS: https://link.springer.com/content/pdf/10.1007/s10618-018-0589-3.pdf

fkiraly:
>As I see it _scientifically_, `CPD` is a different research area, which is mostly about hypothesis testing. The problem with using Segmentation algorithms for CPD, to my understanding, mostly boils down to detecting CP in non stationary signals.

To me, it sounds a bit like a contradiction, since further above you say:

> Definition 4. A segmentation of a TS 𝑇 into 𝐶 + 1 segments is an ordered sequence of change points (or splits) 𝑡𝑖1 ,...,𝑡𝑖𝐶 with 1 < 𝑖1 < · · · < 𝑖𝐶 < 𝑛. 

You *define* segmentation as producing change points.

To resolve this apparnet contradiction: do I understand correctly that for you there's a semantic difference between "change point detection" and the task of producing "change points", which is the main point of contention? 

I think it's easily resolved then, let's just give "change point" returns or "segmentation" returns, and avoid the word "detection" (which is no where in the design anyway), let's use "change point annotation".

>From a sktime framework point of view, it might be tempting to mix both, but then you might also be tempted to use a regressor, discretize its output, and call it a classifier - even if this is a bit over-simplified ;)

But that's exactly what I want to avoid, to mix up different tasks by incorrect interface choice. A regressor with discretized output can be *considered* both a classifier and regressor, but which of the two it is in a scikit-learn sense is a matter of return type and estimator type specification (e.g., mixin inheritance and scoring).

From a pure `sktime` perspective the task is defined by "type" of the estimator, in this case the output type:

* if it's a segmentation type, it's a segmenter
* if it's a sequence type, it's a point annotator (e.g., change point annotator)

Clasp does both, as per the definition you gave above - so let's implement the two interfaces properly.

fkiraly:
PS: I do have comments about the "Turing Change Point Detection Benchmark", but I'll refrain from making them here.

patrickzib:
> To me, it sounds a bit like a contradiction, since further above you say:
The term segmentation and change point detection are definitely overloaded by different meanings :/

> To resolve this apparnet contradiction: do I understand correctly that for you there's a semantic difference between "change point detection" and the task of producing "change points", which is the main point of contention?

Let me try to phrase it like this: 

Change Points in CPD are a result of hypothesis testing as a result of changes in _statistical_ properties. Change points in Segmentation are a result of regime changes, which are defined by changes in the _shapes_ of the time series subsequence.

Maybe my concern is unjustified: You may find benchmarks for either of them. I want to avoid someone applying ClaSP to the CPD tasks.

> I think it's easily resolved then, let's just give "change point" returns or "segmentation" returns, and avoid the word "detection" (which is no where in the design anyway), let's use "change point annotation".

ok

> From a pure `sktime` perspective the task is defined by "type" of the estimator, in this case the output type:
> 
> * if it's a segmentation type, it's a segmenter
> * if it's a sequence type, it's a point annotator (e.g., change point annotator)

> Clasp does both, as per the definition you gave above - so let's implement the two interfaces properly.

Ok. 

I'd rather not implement it as suggested though. Having different return types in the same class based on the input type, is really confusing. And if we have 2 tasks (segmentation and regime change points), and 4 options (dense, sparse, indicator and score), there will be 8 different return types. 

From a framework point of view, I do not like mixing all the options in one return type based on the input parameter. I would rather have a single return type. 

It might be pd.Series or pd.IntervalIndex. Yet again, IntervalIndex does not comply with all your options.

I am fine with `{"dense", "sparse"}`, which could map to returning a segmentation or offsets of the change points.

Also `{"indicator", "score"}` are rather confusing if you have only one return type, as we switch the type of the returned series. "Indicator" would be the label in my understanding while "score" would be the ClaSP scores.

fkiraly:
>From a framework point of view, I do not like mixing all the options in one return type based on the input parameter. I would rather have a single return type.

Yes, I see what you mean. I don't like it entirely either.
We're still early enough in the annotation module that we can make minor to medium interface changes.

How would you go about it instead? `predict` and `predict_score`, while removing the `label` parameter?

patrickzib:
@fkiraly sorry for my late reply!

> > From a framework point of view, I do not like mixing all the options in one return type based on the input parameter. I would rather have a single return type.
> 
> Yes, I see what you mean. I don't like it entirely either.
> We're still early enough in the annotation module that we can make minor to medium interface changes.
> 
> How would you go about it instead? `predict` and `predict_score`, while removing the `label` parameter?

Yes, that would be great. In sklearn for the `Predictor` there are two methods to be implemended [1]:

`predict()`

and 

`predict_proba()`

It would be great to do a similar thing for annotation. I am fine with `predict_score` or simply stick to `predict_proba` and removing the `label`.


[1] https://scikit-learn.org/stable/developers/develop.html

patrickzib:
> I have a few major comments:
> 
> * the algorithm, as currently presented, is not a segmentation algorithm but a change-point detection algorithm. The reason is that it does not output segments (in any encoding), and/or segment labels, but point annotations - the change points.

I have changed the return types according to our discussions. The algorithm will now return:

- Segmentation for fmt=='dense' : pd.IntervalSeries
- Change Points for fmt=='sparse' : pd.Series

Also ClaSP will now support two methods:

- predict() : returns the segmentation (dense) or the change points (fmt) found
- predict_probability() : returns the full profile (dense) or the scores associated with the cpts (fmt).

> * The notebook should explain change point detection in general - i.e., we have a series, we want to identify change points.

I have added a paragraph on this.

> * the notebook contains a lot of custom functionality for plotting, which (a) is great and nice but (b) does not belong in a notebook, but a utility folder, e.g., `sktime/annotation/plotting`. I think we should provide the functions as generic plotting functionality fore change point detection.

Sure! I have reworked the notebook, also introducing a plotting-utility class

> * I would suggest to produce sub-plot functions, i.e., one for the top plot and one for the bottom plots, which can be used for any change point results (not just from clasp). Adopt the matplotlib signature `ax = plot_sth_sth(ax, input, input2, params)`. These are then imported in the notebook and used.

See above :)

patrickzib:
@fkiraly What is your opinion on this? Are you ok with the changes and added methods, and would you apply those changes to the BaseSeriesAnnotator? :)

fkiraly:
Apologies for the wait - this is a complex PR and I needed some time available to go through it carefully. Comments will follow.

fkiraly (changes requested):
@patrickzib, thanks again for the nice contribution!

Apologies for the long time, it's a big pull request.

Regarding our discussion: we've agreed to make some interface changes. To avoid back-and-forth here, I'd suggest you make some changes to base class and docstrings to make this work, and then I look over it in a subsequent PR and ensure everything is coherent.

For this PR, I have comments in three areas.

Clasp:
* only use `numpydoc` compatible section headers or it will mess up our auto-doc-generation. `Overview` is not one of them, just remove it. This is a merge blocker.
* Please add docstrings to the private functions `_fit`, etc. You can copy them from the extension template, though you may want to update them.
* should `find_dominant_window_sizes` be private?
* please use `lower_snake_case` naming vor variables, exception are `X` or `y` from sklearn. So please rename `TS` to `time_series`, or rename it to `X` (or `y`).
* `fit_predict` is currently missing in the base class, so please move it there. Then your concrete class can inherit the expected behaviour.
* the base class currently does not have a `predict_scores`. Do you want to add this, or should I? At the same time, you may want to update the docstrings.

Clasp notebook:
* this should use `sktime` compatible time series formats - i.e., either a `pd.Series`, or a `pd.DataFrame` with no index name.
* either rename the notebook to `segmentation_with_clasp`, or re-work this so it's a generic segmentation notebook. E.g., at the start explain the general task, show clasp as an example rather than the only segmenter (even though it currently is), and show general plotting utilities assuming it's not necessarily clasp.
* the notebook should also showcase the output format of segmentations provided, so the user understands what they get. See, e.g., the forecasting notebook.

Interpolation notebook:
* the `interpolation.ipynb` notebook is really nice, since it demonstrates TSC pipelines. However, it seems unrelated to the rest of this PR. Would you mind moving it to a different PR?
* something to consider: currently, the TSC notebook does *not* explain how to use pipelines, or the interpolator/binner in general. Why don't you add it there, it would be a great section? (if you do it, in a separate PR, please)
* it's not a blocker though - we can add it as a loose notebook as-is, but I feel it should be higher-priority content. No one really looks at the loose notebooks, but the numbered ones get 100s ov views a day.

Minor things:
* the author field should be populated with GitHub user IDs, not full names. This is for linking in the database.

fkiraly:
(apologies, I looked at an older version of the notebook)

patrickzib:
Hi @fkiraly,

thank you for the Feedback! I will look into it.

Just regarding the Interpolation Notebook. In fact, I didnt do any changes to it. I guess, I ran it once as a reference, causing a change in the jupyter notebook version of it and somehow accidentally added it to a commit. I dont know how to remove it from this PR though? Do you know how to remove single filed? Or maybe, as the change is just this one version line, is it ok to keep it like it is?

patrickzib:
Hi @fkiraly,

> @patrickzib, thanks again for the nice contribution!

:)

> Clasp:
> 
> * only use `numpydoc` compatible section headers or it will mess up our auto-doc-generation. `Overview` is not one of them, just remove it. This is a merge blocker.
> * Please add docstrings to the private functions `_fit`, etc. You can copy them from the extension template, though you may want to update them.
> * should `find_dominant_window_sizes` be private?
> * please use `lower_snake_case` naming vor variables, exception are `X` or `y` from sklearn. So please rename `TS` to `time_series`, or rename it to `X` (or `y`).
> * `fit_predict` is currently missing in the base class, so please move it there. Then your concrete class can inherit the expected behaviour.
> * the base class currently does not have a `predict_scores`. Do you want to add this, or should I? At the same time, you may want to update the docstrings.

I have addressed all of these, adding new methods and changing doc-strings. Please check - especially for the changes to the Base-Class. However, I did not rewrite your initialization header and the parameters involved. Maybe you can do so in a future PR? I do not want to interfer with it.


> Clasp notebook:
> 
> * this should use `sktime` compatible time series formats - i.e., either a `pd.Series`, or a `pd.DataFrame` with no index name.
> * either rename the notebook to `segmentation_with_clasp`, or re-work this so it's a generic segmentation notebook. E.g., at the start explain the general task, show clasp as an example rather than the only segmenter (even though it currently is), and show general plotting utilities assuming it's not necessarily clasp.
> * the notebook should also showcase the output format of segmentations provided, so the user understands what they get. See, e.g., the forecasting notebook.

I have reworked the notebook, adding an introduction and also how to use the IntervalSeries. Please have a look.

> Interpolation notebook:

Sorry, that accidentally came into the pull request as I had used it as a reference. I have now succeeded to remove it from the pull request.

> Minor things:
> 
> * the author field should be populated with GitHub user IDs, not full names. This is for linking in the database.
> * the conflict in `_data_io` should be resolved

Sure, I changed my name to the ID, but I cannot do so for Arik, as he has not yet contributed to sktime.  Thus, I will leave his name as it is for now.

ermshaua:
Hi @fkiraly,

I am Patrick's co-author for ClaSP. I just changed my name to my GitHub username in the source files. Can you approve the workflow?

Cheers, Arik

fkiraly:
>Sorry, that accidentally came into the pull request as I had used it as a reference. I have now succeeded to remove it from the pull request.

Yes, but it's really nice!
I think we should add this!

patrickzib:
> > Sorry, that accidentally came into the pull request as I had used it as a reference. I have now succeeded to remove it from the pull request.
> 
> Yes, but it's really nice! I think we should add this! (different PR though)

It was just a misunderstanding: the notebook is part of sktime, I ran it for reference, which changed it due to flake, etc, and then accidentally added it to a commit. This resulted in the notebook to appear in this PR, though I have performed no changes to it.

fkiraly:
>t was just a misunderstanding: the notebook is part of sktime

Oh ... well, it's still nice, but at least we don't need to add it them :smiley:

ermshaua:
Hi @fkiraly,

I added myself to the contributors file. Can you approve the workflow once again?

Cheers, Arik

patrickzib:
@fkiraly two things I forgot to mention in my last commit :)

- `find_dominant_window_sizes(ts)`: yes, this is a public utility function, as it is called e.g. in the notebook. 

- The notebook and ClaSP use `pd.Series`. However, the formatting of `pd.Series` is not pretty in Jupyter Notebooks, thus it is wrapped by a `pd.DataFrame` for displaying purposes only. I wrote a comment above the call in the notebook.

fkiraly:
Apologies for the wait - to speed things up, I've resolved some merge conflicts outside the remit of the PR, in the `_config` and test files.

PR #1354, opened 2021-08-26
check input to split function is dataframe or series 

thayeylolu:
I have made changes to the sktime/forecasting/model_selection/_split.py file in lines 99- 103 . this changes now accept DataFrames.


fkiraly:
this is now fixed with other things in #1456

PR #1356, opened 2021-08-27
Add date time dummy extractor

danbartl:
#### Reference Issues/PRs
Closes https://github.com/alan-turing-institute/sktime/issues/1273

This is a first draft.

Introduce method to generate dummy calender features
Will automatically generate flexible number of features based on the frequency of the series (provided as argument) and the complexity argument (ranges from 0 to 2) to represent few, many or all features.

No

Scope of extractors, logic of complexity arguments

#### Any other comments?
Could not finalize Fourier Extractor before my vacation (in 2 weeks will continue on that).

##### For all contributions

##### For new estimators


danbartl:
> Hi @danbartl, thanks for the PR! I had a quick look and left some comments, I hope they're helpful.
> 
> FYI There is a nice function here that does what we want: https://github.com/fastai/fastai/blob/master/fastai/tabular/core.py#L24
> 

Are Fourier features meant to be part of this transformer? I think I'd prefer to have a separate `FourierFeatures` transformer.

Those will be separate, will be next implementation.


https://github.com/fastai/fastai/blob/master/fastai/tabular/core.py#L24

This is quite similar, but does not cover all the datetimefeatures, also not the logic for base frequency and feature_scope arguments.

PR #1358, opened 2021-08-28
changed forecasting univariate-only tag to ignores-exogeneous-X

### LINKED ISSUE #1267: forecasting: rename misleading `univariate_only` tag to `uses_exogeneous`, correct instances
Currently, the tag that indicates whether a forecaster makes use of exogeneous data or just ignores them is called `univariate_only`. This is misleading, since `y` could be multivariate, even though no exogeneous data are used - also, exogeneous data can be univariate if it hast just one column.

So, the tag should be renamed to sth like `uses_exogeneous`.

I also noticed it's not correct in a number of places, possibly due to extenders being confused by the name.
For instance, `pmdarima`'s AutoARIMA can make use of exogeneous variables, but the tag says it can't.

fkiraly:
This PR renames the `univariate-only` tag to `ignores-exogeneous-X` for forecasters. This closes #1267, see for discussion there why we have decided to do this (in last week's core dev meeting) - its name implied an "obvious" meaning that was wrong.

The `univariate-only` tag still exists in transformers, where it means what the name implies.

It also exists in metrics, where I don't know what it means - perhaps @RNKuhns can comment?

Furtner, I renamed entries in some author fields to GitHub identifiers.


mloning (approved):

RNKuhns:
@fkiraly I had originally (mistakenly) taken "univariate-only" to represent that we were dealing with the case where the endogenous data was a univariate timeseries (one that optionally could be modeled using exogenous data). 

But it sounds like it meant that exogenous data was not handled. 

In terms of performance metrics, all of the metrics I included can handle multiple series as inputs.

PR #1359, opened 2021-08-29
Tsc refactoring _muse classifier

BINAYKUMAR943:
#### Reference Issues/PRs
#1146 TSC Refactoring for _muse classifier

all the public methods are turned into private
input and output checks are removed from the private version
all the test cases are passing
name of capabilities dictionary is changed to _tags
new tags are added to _tag registry

No

##### For all contributions


fkiraly:
@BINAYKUMAR943, great contribution, thanks!

To explain the error you are getting: your new tag `coerce_to_pandas` makes sense, but any new tags also need to be added to the `_tags` registry in the `registry` module, or the tests will complain.

May I also make an alternative suggestion, since there are now two tags about coercion which may make handling a bit difficult?

If you look at the base class for the forecasting module, the conversions/coercions are best done in the base class and handled by the `convert` functions. In forecasting, we have a single tag for each input that describes what it is coerced to, e.g., `X_inner_type` which can be `pd.DataFrame` or `np.ndarray`.

Now that the `datatypes` module is available, I recommend you have a look at the `AA_datatypes_and_datasets` notebook in `examples`, and get familiar with the different `Panel` mtypes; once you have seen how `convert_to` can be used as a replacement to earlier "coerce_to" functionality, you can put it in the base class, where the coercion behaviour is controlled by an `X_inner_type` tag (like it's done in forecasting).

BINAYKUMAR943:
> @BINAYKUMAR943, great contribution, thanks!
> 
> To explain the error you are getting: your new tag `coerce_to_pandas` makes sense, but any new tags also need to be added to the `_tags` registry in the `registry` module, or the tests will complain.
> 
> May I also make an alternative suggestion, since there are now two tags about coercion which may make handling a bit difficult?
> 
> If you look at the base class for the forecasting module (`forecasting/base/_base.py`), the conversions/coercions are best done in the base class and handled by the `convert_to` functions. In forecasting, we have a single tag for each input that describes what it is coerced to, e.g., `X_inner_type` which can be `pd.DataFrame` or `np.ndarray`.
> 
> Now that the `datatypes` module is available, I recommend you have a look at the `AA_datatypes_and_datasets` notebook in `examples`, and get familiar with the different `Panel` mtypes; once you have seen how `convert_to` can be used as a replacement to earlier "coerce_to" functionality, you can put it in the base class, where the coercion behaviour is controlled by an `X_inner_type` tag (like it's done in forecasting).
> 
> Of course, if you think this is too involved, we can also do it later.

Thank you so much for your quick review! As of now, I have added coerce_to_pandas in the _tags registry, but I am looking into your suggestion and trying to implement the same.

fkiraly:
ok - let me know when you think it's ready to review.

fkiraly (changes requested):
I think we ought to call it `coerce_X_to_pandas` to be clearer on this, otherwise looks ready to go.

BINAYKUMAR943:
> I think we ought to call it `coerce_X_to_pandas` to be clearer on this, otherwise looks ready to go.

Thanks for reviewing! I have made the required change.

fkiraly (changes requested):
Looks good, and would merge if it were not for the following issue:
it seems you are making large changes to `CONTRIBUTORS.md`, some formatting perhaps, and difficult to review because it's a lot. Can you make sure that you only change the part where you add yourself?

While you're at it, there's a typo in the new tag description (stray period).

BINAYKUMAR943:
> Looks good, and would merge if it were not for the following issue:
> it seems you are making large changes to `CONTRIBUTORS.md`, some formatting perhaps, and difficult to review because it's a lot. Can you make sure that you only change the part where you add yourself?
> 
> While you're at it, there's a typo in the new tag description (stray period).

 Hi @fkiraly, thanks for your comment!. `CONTRIBUTORS.md` file has names in alphabetical order. So, I put my name accordingly, and since one row can have only seven names, I shifted others according to that. Should I just add my name at the end?

Issue #1098, opened 2021-07-03
tests refactor: move default parameter settings into concrete classes


fkiraly:
The framework has been refactored as descriged below.

Now, all there is to do is to move the defaults from the `tests/_config` file into the individual estimator's `get_testing_params`.

PRs are appreciated - if you work on this, please post below which estimators you are helping with refactoring to avoid duplication.

---

We should refactor the default settings in `_config`.

Currently, `_config` has two things related to default settings:

* default "inhabitants" of key scitypes, within and outside `sktime`
* default parameter settings for each individual concrete estimator

I would go and locate these within the boundaries of their "natural concern":

* default parameter settings as an inspectable attribute of the class it is a default for. This should support a single setting, or multiple settings. I would make a method `get_testing_param_defaults` or similar. Imports of components etc should happen inside this method, to avoid strong coupling of modules and circular imports.
* default inhabitants of scitypes should move to the registry. I would suggest a method `get_example(scitype : str)`, which implements an extensible dispatch mechanusm similar to #1097.

I think this would declutter the logic a lot, and make extending `sktime` much easier - since it reduces the "points of extension" by one. Defaeult parameters would no longer have to be entered in `_config`, but become part of the list of things to implement, in the extension template.

Aparna-Sakshi:
@fkiraly I want to work on this issue. From what I understood, instead of setting the parameters every time inside _config, one could use a function like get_testing_param_defaults to set the default parameters. Could you please guide me as to how I should approach this task as I am new to sktime?

fkiraly:
Yes, very much agreed! You can have the issue, no one is working on this.

I'll try post some more details as a STEP today evening.

Aparna-Sakshi:
@fkiraly Thanks a lot! I have also sent you a dm in discord.

fkiraly:
I created a STEP here, have a look:
https://github.com/sktime/enhancement-proposals/pull/14

fkiraly:
closing this in favour of #1678 which has a list of estimators

PR #1361, opened 2021-08-30
Tests refactor: estimator test parameters with the estimator

Aparna-Sakshi:
#### Reference Issues/PRs
Fixes issue #1098 

In this PR two functions get_test_params and create_test_instance were added to sktime/base/_base.py. Using these functions instance of Estimator can be created with default parameters.

#### Any other comments?
This is task 1 as mentioned in the STEP in https://github.com/sktime/enhancement-proposals/pull/14


fkiraly (dismissed):
Great work, this looks good!

Finally we can remove that difficult to understand step from extending sktime, now adding an estimator takes truly just one file.

sktime/base/_base.py
```diff
@@ -208,6 +208,69 @@ def clone_tags(self, estimator, tag_names=None):
...
 
+    @classmethod
+    def get_test_params(cls):
+        """Get default parameters of the estimator.
+
+        Returns
+        -------
+        params : dict or list of dict, default = {}
+            Default parameters related to the estimator class
+
```
fkiraly, line 219: minor comment: there shouldn't be any of these superfluous spaces here

fkiraly (approved):
Looks good to me now, I'd merge if the tests pass.

mloning:
Thanks @Aparna-Sakshi for the PR, much appreciated! 

I'm a bit late to this, but what's your view on making the test related methods `get_test_params` and `create_test_instances` private rather than public?

fkiraly:
Hm, why would these be private? They are called by other methods and other objects.

Isn't the general idea that private methods of a class are only accessed from inside the class?

mloning:
For me, private variables are part of the internal API as opposed to the public, user-facing API. It's just to keep the API that users see as simple as possible.

fkiraly:
Hm, I disagree with that convention - private methods of classes should be part of the internal class API, i.e., the boundary I'd draw is the class, not the user/code interface.

mloning:
Either way, I'd like to keep the user-facing API as simple as possible

fkiraly:
I think there's a case for exposing these to the user, e.g., to create a quick default for playing with, or for use in the users' testing pipelines (for applications, say).

You can always de-emphasize less crucial interface points in tutorials - similar to `clone_tags` which is also not private.

Aparna-Sakshi:
> I think there's a case for exposing these to the user, e.g., to create a quick default for playing with, or for use in the users' testing pipelines (for applications, say).
> 
> You can always de-emphasize less crucial interface points in tutorials - similar to `clone_tags` which is also not private.

I agree with this. Also, the create_test_instance method is being called inside _construct_instance method(sktime/base/_base.py) which cannot be possible if these methods are made private.

mloning:
> I agree with this. Also, the create_test_instance method is being called inside _construct_instance method(sktime/base/_base.py) which cannot be possible if these methods are made private.

Okay! But note that it would be possible. Python isn't strict about private and public methods.

mloning:
> You can always de-emphasize less crucial interface points in tutorials - similar to `clone_tags` which is also not private.

Yes my preference would also be to make that private, since it's only used internally and not by users. But no strong preference here.

Aparna-Sakshi:
> > I agree with this. Also, the create_test_instance method is being called inside _construct_instance method(sktime/base/_base.py) which cannot be possible if these methods are made private.
> 
> Okay! But note that it would be possible. Python isn't strict about private and public methods.

@mloning Oh yeah, that's true. Renaming these methods by adding a single '_' wouldn't be a problem. Thanks for pointing that out.

fkiraly:
>Python isn't strict about private and public methods.

Yes, but it's still bad style to denote methods that are used like a public method with the private convention...

mloning:
> > Python isn't strict about private and public methods.
> 
> Yes, but it's still bad style to denote methods that are used like a public method with the private convention...
> 
> The general convention here being that underscore methods are exactly those which are not meant to be called from the outside _of the class_.

I don't agree, I think you can find that in many places in the PyData ecosystem. It's just a different convention, focused on separating internals from user-facing code. Anyways, didn't mean to start a big discussion about this.

fkiraly:
@mloning, then you disagree with the python documentation proper, see here:
https://docs.python.org/3/tutorial/classes.html#tut-private

>“Private” instance variables that cannot be accessed except from inside an object don’t exist in Python. However, there is a convention that is followed by most Python code: a name prefixed with an underscore (e.g. _spam) should be treated as a non-public part of the API

I would read this as follows: the author already *assumes* that everyone agrees with the definition of private which is not being accessible from outside the object scope. They explain that in python everything is accessible but the convention for "variables that cannot be accessed except from inside an object" is the underscore convention.

So, I'd really prefer not to deviate from widely accepted onventions in object oriented programming, or the python specific conventions around it, for that matter.

If a variable is accessed by other objects/functions/workflows, it should not be private.

fkiraly:
From what you say, it seems that the divergence of opinions may stem from the definition of "cannot be accessed" - you purely consider access by the user (via a console/notebook interface). This is *not* the usual definition in object oriented programming. If that were the definition, then everything apart from the user interface layer would always be private, which doesn't make much sense.

mloning:
@fkiraly I think we're going a bit astray here. As said, I'm aware of the general convention. My point was that I'd prefer not to expose methods in the API which users don't need and may not understand. One way of doing that is using the leading underscore convention. Other package do that too (e.g. `_get_tags` in scikit-learn). But if you're feeling strongly about this, I don't mind keeping it as it is.

fkiraly:
We disagree on two points then:

* that the user never wants to use `get_test_params`, you can well imagine an experienced user wants this
* that some formally public functions (which this is since it's called from outside the class) should follow the private naming convention based on whether the user interacts with it

PR #1370, opened 2021-08-31
bugfix - incorrect/missing weighted geometric mean in forecasting ensemble

fkiraly:
There was a bug in the ensemble forecasters `AutoEnsembleForecaster` and `EnsembleForecaster` where the weighted version of geometric mean was not properly interfaced, instead an unweighted version was provided which would break.

I've fixed this by using `scipy.stats.mstats.gmean`.

### DIFF
--- a/sktime/forecasting/compose/_ensemble.py
+++ b/sktime/forecasting/compose/_ensemble.py
@@ -13,6 +13,7 @@ __all__ = ["EnsembleForecaster", "AutoEnsembleForecaster"]
 import numpy as np
 import pandas as pd
 from scipy.stats import gmean
+from scipy.stats.mstats import gmean as _weighted_gmean
 from sklearn.pipeline import Pipeline
 from sklearn.utils.stats import _weighted_percentile
 
@@ -309,7 +310,7 @@ def _check_aggfunc(aggfunc, weighted=False):
         "median": {"unweighted": np.median, "weighted": _weighted_median},
         "min": {"unweighted": np.min, "weighted": _weighted_min},
         "max": {"unweighted": np.max, "weighted": _weighted_max},
-        "gmean": {"unweighted": gmean, "weighted": gmean},
+        "gmean": {"unweighted": gmean, "weighted": _weighted_gmean},
     }
     if aggfunc not in valid_aggfuncs.keys():
         raise ValueError("Aggregation function %s not recognized." % aggfunc)


RNKuhns:
@fkiraly good catch! We missed the scipy versioning issue in our implementation. 

In terms of the custom function, tere is one in `sktime.performance_metrics._functions.py` that is based on [wikipedia entry](https://en.wikipedia.org/wiki/Weighted_geometric_mean). In your implementation, is the use of `np.exp` and `np.log` flipped?

I've copied the implementation in `sktime.performance_metrics._functions.py` below for convenience. If that implementation is correct, should we move it somewhere (utils?) and have it imported into both places from there? Might be nice to have a single place for these odds-and-ends statistical functions that pop up from time to time.

```python
def _weighted_geometric_mean(x, sample_weight=None, axis=None):
    """Calculate weighted version of geometric mean.

    Parameters
    ----------
    array : np.ndarray
        Values to take the weighted geometric mean of.
    sample_weight: np.ndarray
        Weights for each value in `array`. Must be same shape as `array` or
        of shape `(array.shape[0],)`.

    Returns
    -------
    geometric_mean : float
        Weighted geometric mean
    """
    check_consistent_length(x, sample_weight)
    return np.exp(
        np.sum(sample_weight * np.log(x), axis=axis) / np.sum(sample_weight, axis=axis)
    )
```

sktime/forecasting/compose/_ensemble.py
```diff
@@ -316,14 +316,19 @@ def _aggregate(y, aggfunc, weights):
     return pd.Series(y_agg, index=y.index)
 
 
+def _weighted_gmean(a, axis=0, weights=None):
+    avg = np.exp(a)
+    return np.log(np.average(avg, axis=axis, weights=weights))
```
aiwalter, line 321: just a question out of interest, why are we doing `exp` and `log`?

aiwalter (approved):

sktime/forecasting/compose/_ensemble.py
RNKuhns: @aiwalter I may be mis-remembering from when I did this in `performance_metrics` but I believe it is to avoid the cumulative product getting too large.

RNKuhns:
@fkiraly have you had a chance to look at my comment on this one? Seems like an easy one for us to merge and help lower the number of PRs.

Plus it would be good to fix the bug for users!

sktime/forecasting/compose/_ensemble.py
fkiraly: yes, exactly - this has improved numerical behaviour (I think)

fkiraly:
@RNKuhns, apologies, I did not see your comment.

Yes, @RNKuhns, it looks like the `log` and `exp` are flipped in my implementation, so it's wrong.

Would it be easy to use your implementation instead? Feel free to close this PR and/or commit to it.

sktime/forecasting/compose/_ensemble.py
fkiraly: but... it's the wrong way round (see below)

RNKuhns:
@fkiraly I'm fine using my implementation -- I'll push to the `ensemble-gmean-weighted` branch to do this and also move my implementation to `sktime.utils` since it isn't clear to look for it in `performance_metrics` if we need it in another place again later.

fkiraly (commented):
yes, much better.
I would approve, though I can't review my own pull requests...

fkiraly:
the tests no longer pass, @RNKuhns? The function does not recognize the `weights` argument?

Issue #1364, opened 2021-08-30
[ENH] forecasting: ensure compositors (pipeline, tuning) work for multivariate forecasters


fkiraly:
From the Aug 27 developer meeting.

We need to ensure that the following compositors work for multivariate forecasters, in descending order of priority
- [x] pipeline compositors, `TransformedTargetForecaster`, `ForecastingPipeline` #1376 
- [x] tuning compositors, `ForecastingGridSearchCV`, `ForecastingRandomizedSearchCV` #1376 
- [ ] reducers, all in the `_reduce` module, see #1038
- [ ] multiplexing, `MultiplexForecaster`
- [ ] ensembles, `EnsembleForecaster`, `AutoEnsembleForecaster`
- [ ] stacking, `StackingForecaster`

Some of these might be slightly challenging, but still good first issues.
As usual, we use this issue to track who is working on what, please indicate your pick before starting work on it.

aiwalter:
I added this also to the wish list, but lets keep track here then. I will pick the `TransformedTargetForecaster` as a start :). After my test refactor contained in #1347 is merged.

aiwalter:
can we add `NaiveForecaster` to this list? It could internally just apply a `ColumnEnsembleForecaster`

aiwalter:
What about `ForecastingPipeline`? Do we need to adjust it? Not sure if there is a use-case where `y` and `X` are both `pd.DataFrame`?

fkiraly:
Yes, I think that´s not an uncommon use case.

Btw, it looks like `TransformedTargetForecaster` works for multivairate out of the box?

mloning:
> Btw, it looks like `TransformedTargetForecaster` works for multivairate out of the box?

No, still only works for univariate input data.

aiwalter:
I have a PR open for `TransformedTargetForecaster` and `BaseGridSearch`, see #1376 . Need to wait merging #1301

PR #1376, opened 2021-09-02
Added multivariate TransformedTargetForecaster

aiwalter:
#### Reference Issues/PRs
#1364 
#220 

Accept `isinstance(y, pd.DataFrame)` in `TransformedTargetForecaster`

#### Any other comments?
Waiting for #1301 to be merged, tests will fail before.

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/forecasting/compose/_pipeline.py
+++ b/sktime/forecasting/compose/_pipeline.py
@@ -74,18 +74,21 @@ class _Pipeline(
         """Return the length of the Pipeline."""
         return len(self.steps)
 
-    def _get_inverse_transform(self, y):
+    def _get_inverse_transform(self, y, X):
         """Iterate over transformers.
 
         Inverse transform y (used for y_pred and pred_int)
 
         Parameters
         ----------
-        y : pd.Series
+        y : pd.Series, pd.DataFrame
+            Target series
+        X : pd.Series, pd.DataFrame
+            Exogenous series.
 
         Returns
         -------
-        y : pd.Series
+        y : pd.Series, pd.DataFrame
             Inverse transformed y
         """
         for _, _, transformer in self._iter_transformers(reverse=True):
@@ -93,7 +96,7 @@ class _Pipeline(
             # is not wanted ur meaningful (e.g. Imputer, HampelFilter)
             skip_trafo = transformer.get_tag("skip-inverse-transform", False)
             if not skip_trafo:
-                y = transformer.inverse_transform(y)
+                y = transformer.inverse_transform(y, X)
         return y
 
     @property
@@ -315,7 +318,9 @@ class TransformedTargetForecaster(_Pipeline, _SeriesToSeriesTransformer):
 
     _required_parameters = ["steps"]
     _tags = {
-        "univariate-only": True,
+        "scitype:y": "both",
+        "univariate-only": False,
+        "y_inner_mtype": ["pd.Series", "pd.DataFrame"],
         "requires-fh-in-fit": False,
         "handles-missing-data": False,
     }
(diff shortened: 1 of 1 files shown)


sktime/forecasting/base/_base.py
```diff
@@ -576,24 +575,6 @@ def get_fitted_params(self):
         """
         raise NotImplementedError("abstract method")
 
-    def _set_y_X(self, y, X=None, enforce_index_type=None):
```
fkiraly, line 579: I assume this is removed because nothing calls it anymore?

sktime/forecasting/base/adapters/_fbprophet.py
```diff
@@ -43,7 +44,7 @@ def _fit(self, y, X=None, fh=None, **fit_params):
         """
         self._instantiate_model()
         self._check_changepoints()
-        self._set_y_X(y, X, enforce_index_type=pd.DatetimeIndex)
+        y, X = check_y_X(y, X, enforce_index_type=pd.DatetimeIndex)
```
fkiraly, line 47: if you really insist to make this change to the fbprophet adapter, it should be controlled by the tag `enforce-index-type`

sktime/forecasting/compose/_column_ensemble.py
```diff
@@ -60,6 +60,7 @@ def __init__(self, forecasters):
         # set requires-fh-in-fit depending on forecasters
         if isinstance(forecasters, BaseForecaster):
             self.clone_tags(forecasters, "requires-fh-in-fit")
+            self.clone_tags(forecasters, "scitype:y")
         else:
             forecaster_requires_fh_in_fit = (
```
fkiraly, line 65: this will break if mistakenly `forecasters` is not of the expected type without informative error message. I would check validity in the init, which is the only way to prevent this that I can see.

sktime/forecasting/compose/_pipeline.py
```diff
@@ -166,7 +169,8 @@ class ForecastingPipeline(_Pipeline):
 
     _required_parameters = ["steps"]
     _tags = {
-        "univariate-only": False,
+        "scitype:y": "both",
+        "y_inner_mtype": ["pd.Series", "pd.DataFrame"],
```
fkiraly, line 173: hm, due to the duplication: should these new tags not move to `_Pipeline`? Perhaps it's safer here though, so perhaps not.

fkiraly (approved):
Really great, all seems to work!

Some minor (non-blocking) comments:

* the change in fbprophet interface should be controlled by a tag
* other minor comments
* if you are fiddling around with `_config`, would it be perhaps a good time to move the default parameters into `@classmethod` `get_test_params`? See draft extension template here #1395 (the functionality is already "in").

sktime/forecasting/base/adapters/_fbprophet.py
aiwalter: I did this in order deprecate the `_set_y_X` as it was only used there and there is basically no additional functionality. Aso the index type was already enforced. But I can add a tag there, yes.

fkiraly:
@aiwalter, I think this should go in with the release.

You need to resolve the conflict before we can do this, and there´s some minor things.

aiwalter:
@fkiraly yes but I want to fix first what you commented. Also I need first to add multivariate `NaiveForecaster` to make sure all automated tests are running well. So I will do this by end of this week

mloning:
Note to self: This depends on #1401 to ensure all tests run for the multivariate setting.

aiwalter:
@fkiraly @mloning pls review :)

aiwalter:
waiting for hotfix:
Fixed typo `y:scitype`, renamed to `scitype:y`

fkiraly:
@aiwalter, I'll approve if @mloning is happy how you handled `_NaiveForecaster`.

aiwalter:
Can we get this in @fkiraly @mloning

PR #1383, opened 2021-09-03
Refactor TSF, RISF and TSF classifiers

kejsitake:
#### Reference Issues/PRs

Partially handles #1146. TSC refactoring for RandomIntervalSpectralForest, SupervisedTimeSeriesForest and TimeSeriesForestClassifier

- Turned the public methods into private
- Changed the name of the capabilities dictionary to _tags.
- Removed the checks from the private version
- Checked that all tags already were already in the _tag registry

No

Please look at the changes and let me know if I should make any other edits.
#### Any other comments?

No.


fkiraly:
(please ignore the failing manylinux, it's a known issue)

fkiraly (changes requested):
Looks good!

Before we merge, would you kindly look into the "intermediate base classes" these algorithms are inheriting from, i.e., `BaseTimeSeriesForest` and `ForestClassifier`? I think there's a `fit` there that still needs looking at.

fkiraly:
Actually, RISF also fails the tests - can you check why?

kejsitake:
Hi! Thank you for taking a look at this! I also made changes to `BaseTimeSeriesForest ` classifier. From what I understand `ForestClassifier` is inherited from sklearn. I have been trying to understand why the test for RISF is failing, but the errors seems to point to some validation function in sklearn and I am not sure how to debug this further.

kejsitake:
Hi! I changed the RISF classifier to inherit only from the `BaseClassifier` instead of also `ForestClassifier` and declared an instance of the `ForestClassifier` as an object of the class. The test that was failing before: 02_classification_univariate.ipynb passes, but I am not sure if this is the right way to solve this problem. Happy to take another look too.

PR #1395, opened 2021-09-06
Tests Refactor: Adding get_test_params method to extension template

Aparna-Sakshi:
#### Reference Issues/PRs
Fixes issue #1098 

In this PR, get_test_params has been added to extension template for concrete classes

#### Any other comments?
This is task 2 as mentioned in the STEP in https://github.com/sktime/enhancement-proposals/pull/14


fkiraly (changes requested):
This already looks good!
You should let the core devs know when you think it's ready for review.

Almost ready to merge, but a few minor changes I'd like to see in-text.

extension_templates/annotation.py
```diff
@@ -30,6 +30,9 @@
     fitted state flag       - is_fitted (property)
     fitted state inspection - check_is_fitted()
 
+Testing
```
fkiraly, line 33: the other rubrics end in a colon, this should be consistent

extension_templates/annotation.py
```diff
@@ -30,6 +30,9 @@
...
+    get default parameters for test instance - get_test_params()
```
fkiraly, line 34: `create_test_instance` should also be mentioned, even if extenders don't have to implement it

extension_templates/annotation.py
```diff
@@ -179,3 +182,28 @@ def _update(self, X, Y=None):
...
+    @classmethod
+    def get_test_params(cls):
+        """Get default parameters of the estimator.
+
+        Returns
+        -------
+        params : dict or list of dict, default = {}
+            Default parameters related to the estimator class
+        """
+        params = super().get_test_params()
```
fkiraly, line 196: I would not implement any default logic here - leave it empty, this is not a base class but a "fill in" template.

extension_templates/annotation.py
```diff
@@ -179,3 +182,28 @@ def _update(self, X, Y=None):
...
+
+        # todo: set the default parameters for the estimators
```
fkiraly, line 198: it should be better explained how `create_test_instance` uses the parameters, and what form the dictionary needs to take (needs to result in a constructed instance when called `MyClass(**params)`)

Aparna-Sakshi: @fkiraly Could you please review it now? I have only updated for annotation.py. If it is fine then I'll update for remaining also.

Aparna-Sakshi:
@fkiraly Could you please review it now? I have only updated for annotation.py. If it is fine then I'll update for remaining also.

extension_templates/annotation.py
```diff
@@ -30,6 +30,10 @@
     fitted state flag       - is_fitted (property)
     fitted state inspection - check_is_fitted()
 
+Testing:
+    get default parameters for test instance - get_test_params()
```
fkiraly, line 34: I'd say `test instance(s)` since it can be one or multiple

extension_templates/annotation.py
fkiraly: ok - I think you want to write more precisely what the return is - "default parameters related to the estimator class" is too vague - related how? Suggestion, write sth close to what I wrote, i.e., that `MyClass(**params)` gets a test case, where `params` is...

fkiraly: also, `create_test_instance` should be mentioned

fkiraly: here's how I would write the docstring:

```
        """Return testing parameter settings for the estimator.

        Returns
        -------
        params : dict or list of dict, default = {}
            Testing parameters related to the estimator class.
            Each dict are parameters to construct an "interesting" test instance,
            that is, `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
            `create_test_instance` uses the first (or only) dictionary in `params`
        """
```

extension_templates/annotation.py
```diff
@@ -179,3 +183,46 @@ def _update(self, X, Y=None):
...
+        # example 1: specify params as dictionary
+        # any number of params can be specified
+        # params={"est":value0, "parama":value1, "paramb":value2}
+        #
+        # example 2: specify params as list of dictionary
+        # note: Only first dictionary will be used by create_test_instance
+        # list of dictionary can be of any length
+        # params=[{"parama":value1,paramb:value2},{"parama":value3,"paramb":value4}]
+        #
+        # This method would be called by create_test_instance.
```
fkiraly, line 210: this is too verbose and duplicates content - can you shorten it?

Aparna-Sakshi:
Thanks a lot for your feedback. I'll update the template based on it.
Should I describe create_test_instance inside the docstring of get_test_params or in the comments(starting with #)?

fkiraly:
>Should I describe create_test_instance inside the docstring of get_test_params or in the comments(starting with #)?

I think it needs to be described:
* in `BaseObject`, where it already has a docstring
* by reference, in appropriate places of the extension template (like in my suggested docstring)

extension_templates/annotation.py
Aparna-Sakshi: I have shortened this part, could you please review it?

extension_templates/annotation.py
Aparna-Sakshi: I have updated the docstring to as you have written in annotation.py and _base.py.

Aparna-Sakshi:
@fkiraly could you please review it now.

extension_templates/annotation.py
```diff
@@ -30,6 +30,10 @@
     fitted state flag       - is_fitted (property)
     fitted state inspection - check_is_fitted()
 
+Testing:
+    get default parameters for test instance(s) - get_test_params()
+    create an instance of estimator class       - create_test_instance()
```
fkiraly, line 35: would suggest instead "create a test instance of the estimator class"

extension_templates/annotation.py
```diff
@@ -179,3 +183,30 @@ def _update(self, X, Y=None):
...
+
+    # todo: return default parameters, so that a test instance can be created
+    @classmethod
+    def get_test_params(cls):
+        """Return testing parameter settings for the estimator.
+
+        Returns
+        -------
+        params : dict or list of dict, default = {}
+            Testing parameters related to the estimator class.
```
fkiraly, line 195: change this to sth less vague, not "related to".

Suggestion: "parameters to create testing instances of the class"

fkiraly (changes requested):
Great, thanks!

I would suggest minor changes to the description, otherwise fine.

Aparna-Sakshi:
@fkiraly I was wondering about another thing, why should create_test_instance return only the first element in case of a list of dictionaries? Something like create_test_instance(cls, index=0) can be done and cls(**params[i]) can be returned. That way users wouldn't be restricted to the default first param. Could you please share your thoughts on this?

fkiraly:
>That way users wouldn't be restricted to the default first param. Could you please share your thoughts on this?

Yes, I agree that what you are suggesting is a good idea.
I did consider this, and a variant where you can get an iterator that produces instances.

My reasoning behind the current `create_test_instance` signature is:

* there should be an "easy" way to get one test instance, that's `MyClass.create_test_instance()`.
* if multiple instances are needed, the user could call `MyClass(**MyClass.get_test_params()[i])` - this is a bit clunky, but would be done in testing only mostly.
* I agree that `MyClass.create_test_instance(index=i)` is easier to read - the problem is though if you want to loop over all test instances, you don't know *how many* test instances there are. That could be solved by `MyClass.create_test_instances()` being an iterable, but then creating a single test instance (key use case) would be `MyClass.create_test_instances()[0]` which is counter-intuitive.
* you could add both `create_test_instance` and `create_test_instances`, but that's too many too similar methods for my test.
* finally, one of the "python zen" (google this) design rules is to prefer one obvious way to do things above multiple ways to do the same thing. You can justify separate existence of `get_test_params` and `create_test_instance` through the separate intention, even though you can get the second by  `MyClass(**MyClass.get_test_params()[0])`, but more methods go too far against this for my taste.

Having thought about your proposal, I'm not opposed - how about the following change?

`create_test_instance(which = 0)`, where the `which` argument can be an integer, or the string "all". If it's "all", what is returned is an iterable that contains `[MyClass(**params) for params in MyClass.get_test_params()]`? (if there are multiple parameter combinations)

(let's not implement this yet, and if at all, it should be in a separate PR anyway)

Aparna-Sakshi:
> > That way users wouldn't be restricted to the default first param. Could you please share your thoughts on this?
> 
> Yes, I agree that what you are suggesting is a good idea.
> I did consider this, and a variant where you can get an iterator that produces instances.
> 
> My reasoning behind the current `create_test_instance` signature is:
> 
> * there should be an "easy" way to get one test instance, that's `MyClass.create_test_instance()`.
> * if multiple instances are needed, the user could call `MyClass(**MyClass.get_test_params()[i])` - this is a bit clunky, but would be done in testing only mostly.
> * I agree that `MyClass.create_test_instance(index=i)` is easier to read - the problem is though if you want to loop over all test instances, you don't know _how many_ test instances there are. That could be solved by `MyClass.create_test_instances()` being an iterable, but then creating a single test instance (key use case) would be `MyClass.create_test_instances()[0]` which is counter-intuitive.
> * you could add both `create_test_instance` and `create_test_instances`, but that's too many too similar methods for my taste.
> * finally, one of the "python zen" (google this) design rules is to prefer one obvious way to do things above multiple ways to do the same thing. You can justify separate existence of `get_test_params` and `create_test_instance` through the separate intention, even though you can get the second by  `MyClass(**MyClass.get_test_params()[0])`, but more methods go too far against this for my taste.
> 
> Having thought about your proposal, I'm not opposed - how about the following change?
> 
> `create_test_instance(which = 0)`, where the `which` argument can be an integer, or the string "all". If it's "all", what is returned is an iterable that contains `[MyClass(**params) for params in MyClass.get_test_params()]`? (if there are multiple parameter combinations)
> 
> (let's not implement this yet, and if at all, it should be in a separate PR anyway)

Ok, now everything makes more sense. Thanks for sharing it. The idea of using "all" is nice, I will work on this in later PRs

extension_templates/annotation.py
```diff
@@ -179,3 +183,30 @@ def _update(self, X, Y=None):
...
+    # todo: return default parameters, so that a test instance can be created
+    @classmethod
+    def get_test_params(cls):
+        """Return testing parameter settings for the estimator.
+
+        Returns
+        -------
+        params : dict or list of dict, default = {}
+            Parameters to create testing instances of the class
+            Each dict are parameters to construct an "interesting" test instance, (i.e),
```
fkiraly, line 196: remove brackets around `i.e.`, line should be 
`Each dict are parameters to construct an "interesting" test instance, i.e.,`

extension_templates/annotation.py
```diff
@@ -179,3 +183,30 @@ def _update(self, X, Y=None):
...
+            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+            `create_test_instance` uses the first (or only) dictionary in `params`
+        """
+
+        # todo: set the testing parameters for the estimators
+        # Testing parameters can be dictionary or list of dictionaries
+        #
+        # example 1: specify params as dictionary
+        # any number of params can be specified
+        # params={"est":value0, "parama":value1, "paramb":value2}
```
fkiraly, line 206: please adhere to linting rules in the example code , should be `params = {"est": value0, ` and so on

fkiraly (changes requested):
Great!

Really minor changes requested only on formatting (but since it's the extension guidelines which is a frequent user entry point, let's be meticulous).

extension_templates/annotation.py
Aparna-Sakshi: @fkiraly I have made following changes
Added space around = and a space after : in dictionary wherever the code has been written in comments
Is there any other changes apart from these?

fkiraly: hm, I can't see these changes in here - have you pushed them?

Aparna-Sakshi: I have pushed them now. @fkiraly could you please have a look?

fkiraly (dismissed):
Great, thanks!
All minor issues now resolved as far as I see.

If you can check, I think in the `_base` doc string, you didn't remove the brackets around the "i.e."

Aparna-Sakshi:
> Great, thanks!
> All minor issues now resolved as far as I see.
> 
> If you can check, I think in the `_base` doc string, you didn't remove the brackets around the "i.e."

Oh Sorry! I missed it. I have fixed it now.

fkiraly (approved):
All good now, thanks!

PR #1396, opened 2021-09-08
[MNT] Update codecov config

mloning:
Excludes some files from code coverage metrics.


aiwalter (approved):

mloning:
I'm not really sure if that makes a difference. Doesn't seem like it.

mloning:
The files were actually already excluded by the `.coveragerc` file which is used when generating the coverage reports in `.github/workflows/build-and-test.yml`.

PR #1407, opened 2021-09-11
Tests refactor: using `pytest_generate_tests` instead of loops

fkiraly:
This PR refactors the `test_all_estimators` tests to use `pytest` native functionality instead of hand-crafted loops.

Main changes:

* `check_sthsth` are now `test_sthsth` and moved to `test_all_estimators`
* the dispatch of test parameters is now taken care of by `pytest`'s `pytest_generate_tests`
* the tests have been split into "class" and "estimator" tests. "class" tests use `estimator_class` fixtures, "instance" tests use `estimator_instance` fixtures. The latter are where previously `_construct_instance` was invoked; the former where not (or where previously they were but unnecessarily so).
* the fixtures for individual tests are created in `pytest_generate_tests`; skipped tests using `EXCLUDED_TESTS` are also handled there. This makes the manual dispatch mechanisms unnecessary.
* `_construct_instance` dependency is now removed in favour of `create_test_instance`  or `get_test_params` class methods being directly invoked.


fkiraly:
This all seems to work, but locally `test_fit_does_not_overwrite_hyper_params` fails on `FeatureUnion`, namely the `transformer_list` parameter. I don't think the refactor introduced this issue with `FeatureUnion`, but why was it not detected previously?

fkiraly:
@aiwalter, can you give this a review?

mloning:
I like this, using fixture is indeed cleaner! A question is how we can still make the standard estimator tests exportable so that they can be used in other packages. Not having a `check_estimator` function seems to make this harder. Any idea?

fkiraly:
>Not having a check_estimator function seems to make this harder. Any idea?

You can still export them, I don't see any problem?

fkiraly:
what exactly would the workflow be you are referring to?

Currently, there is not a single interface point either, since, for instance, a forecaster would have to call tests in `tests/test_all_estimators`, `utils/tests/test_all_estimators`, and `forecasting/tests/test_all_forecasters`.

We can work towards a "one-call-comprehensive-test" separately, but in the current main this does not exist either, right?

mloning:
How would you import and run these tests for an estimator defined in a companion package?

fkiraly:
>How would you import and run these tests for an estimator defined in a companion package?

You can't do this easily, but this is imo a false dichotomy because at current state you can't do that either - due to the hard-coded reliance on the `_config` file, for instance.

As said, I'm all for implementing a simple `test_custom_estimator` interface that the user can call, but *currently this does not exist either*, so you can't fault this PR for not introducing it.

fkiraly:
In case you're just asking how I'd do it given the new set-up:

I would use similar functionality to how we scrape all estimators in order to get all tests.

In an end state, the fixtures would be called `estimator_instance`, `forecaster_instance`, etc.
Using introspection and the estimator's scitype, I'd identify the fixtures relevant for the estimator (e.g., `estimator_instance`, `estimator_class`, `forecaster_class`, `forecaster_instance` for a forecaster) and then run those tests.

This assumes an end state where any dependency on `_config` is also removed, e.g., test parameters moved to `get_test_params`.

mloning:
> As said, I'm all for implementing a simple `test_custom_estimator` interface that the user can call, but _currently this does not exist either_, so you can't fault this PR for not introducing it.

@fkiraly I'm not faulting this PR, I've asked a question about it. My worry is that the changes in this PR make this even harder to implement. I believe with the ongoing refactoring of the `_config.py` file we would have that functionality.

fkiraly:
>I believe with the ongoing refactoring of the _config.py file we would have that functionality.

Yes/no.
Yes, because you remove the dependency on `_config.py`.
No, because the tests are still in multiple locations and you would need to know where they are.

I think either way we would still need to build a single-interface dispatch mechanism by scitype, to the various tests.

Doing it by fixture name seems elegant to me.

fkiraly:
We could also look at test classes and using inheritance, i.e., one test class per scitype and then importing?

Anyway, @mloning, have you made up your mind whether we want to move to this?
I really think it's using `pytest` more cleanly than the current state.

fkiraly:
@mloning, have you made up your mind?

You said

>I like this, using fixture is indeed cleaner

fkiraly:
@mloning, please don't ignore this PR.

I've now added a function `tests.estimator_checks.check_estimator` that showcases how you would use the new architecture for manual testing.

fkiraly:
>>the dispatch of test parameters is now taken care of by pytest's pytest_generate_tests

>I have no opinion about this, I do not really know what benefits it brings, but can go with it.

It's the `pytest` official way that is one of the `pytest` package's interface points.
This PR replaces a custom implementation of fixture dispatch by a test suite following the `pytest` specification.
The benefit is adopting the widely used `pytest` interface over reimplementing parts of it in a custom way.

PR #1401, opened 2021-09-10
Added multivariate NaiveForecaster

aiwalter:
#### Reference Issues/PRs
Related #1399 

#### What does this implement/fix? Explain your changes.
Made `NaiveForecaster` multivariate by means of `ColumnsEnsembleForecaster`

also fixes a bug in `OnlineEnsembleForecaster` which was not completely refactored and overrode `update_predict` with an old version

PR #1410, opened 2021-09-12
Forecasting return type update: series in, series out

fkiraly:
This PR implements the change request from the last core developer meeting:

forecasters' `predict` will now return the forecast series in the `mtype` that is idential to the last `mtype` seen for `y`, in `fit` or `update`.

The PR also includes some clean-up of `check_is` in the `datatypes` module which is not necessary for the new feature but improves code quality. The utility `check_is` is called in the check that is changed.

### DIFF
--- a/sktime/forecasting/base/_base.py
+++ b/sktime/forecasting/base/_base.py
@@ -138,7 +138,7 @@ class BaseForecaster(BaseEstimator):
 
         y = check_series(y, **check_y_args, var_name="y")
 
-        self.y_mtype = mtype(y)
+        self._y_mtype_last_seen = mtype(y)
         # end checking y
 
         # checking X
@@ -250,7 +250,7 @@ class BaseForecaster(BaseEstimator):
         # convert to output mtype, identical with last y mtype seen
         y_out = convert_to(
             y_pred,
-            self.y_mtype,
+            self._y_mtype_last_seen,
             as_scitype="Series",
             store=self.converter_store_y,
         )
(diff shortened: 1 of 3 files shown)


sktime/forecasting/base/_base.py
```diff
@@ -137,6 +137,8 @@ def fit(self, y, X=None, fh=None):
         }
 
         y = check_series(y, **check_y_args, var_name="y")
+
+        self.y_mtype = mtype(y)
```
mloning, line 141: let's make this a private variable
```suggestion
        self._y_mtype = mtype(y)
```

mloning: I thought the converter store could handle this?

fkiraly: no, currently it only stores information from lossy conversions such as data frame column names - not the original mtype.
It might be an idea to change the behaviour that it even remembers the original mtype, but what would the syntax be?

fkiraly: Anyway, *this* is a prime case for a private variable.

I've changed it to private, and to be clearer on the content, to `self._y_mtype_last_seen`

fkiraly:
So, @aiwalter, any comments? Yesterday, it sounded you wanted this done quickly, but haven't heard anything from you...

aiwalter:
@fkiraly yes thx, I already merged it into my PR #1401 .

aiwalter (dismissed):

Issue #984, opened 2021-06-19
forecasting interval and probabilistic prediction interface refactor/re-design


fkiraly:
Let's think about a re-factoring of the forecasting interface for interval and probabilistic predictions.

I think we should have:

* `predict_interval`, this should take `alpha` as in `predict` and return quantiles; one should be able to pass lists of `alpha` too (if the forecaster supports it). If a list is passed, we need to think about the return object - perhaps a data frame, with nested column index if we are in the multivariate case? Or, we could annotate columns `[variablename]__[statistic]`, e.g., `temperature__P95` using the reserved double-underscore.
* `predict_var`, this should return predictive variance (not quantiles) - note that that's what comes out Gaussian process type regressors primarily.
* `predict_proba`, this returns a `skpro`-like distribution object. May requires work on `skpro`, or could use `tensorflow-probability` distributions.

For the first two, we could alternatively think about return flags and return objects in `predict`, but that just seems very messy to me.

Any opinions? @aiwalter, @mloning, @RNKuhns

mloning:
I agree that we need to refactor this. 

The main reason for combining point predictions and prediction intervals was that intervals often rely on point predictions, so we would avoid computing point predictions twice. Any solution to that problem? Storing `y_pred` for a given cutoff and forecasting horizon in the forecaster like statsmodels doesn't seem right.

fkiraly:
Looking through the base class, I think there are a number of unnecessary (or refactor worthy) methods in there:

* `compute_pred_int`, seems to duplicate the current interface point in `predict`, used only by `theta`
* `_compute_pred_int`, called by `compute_pred_int` only
* `_get_pred_int`, used only by `_tbats` - move to `_tbats`?
* `_compute_pred_err`, called by `compute_pred_err`, used for override in some descendants

fkiraly:
>The main reason for combining point predictions and prediction intervals was that intervals often rely on point predictions, so we would avoid computing point predictions twice. Any solution to that problem? 

Two solutions I can see:

* caching the prediction it's called the first time
* the "interval" methods also return point predictions when called, so if you want both you just call the "interval" method

I think I prefer the second one.

SveaMeyer13:
FYI: @mloning and I discussed that I could work on this. To get familiar with the current design I started by reviewing #775.

fkiraly:
I made a STEP here, please review:
https://github.com/sktime/enhancement-proposals/pull/15

fkiraly:
[...]

PR #1421, opened 2021-09-15
First step or prediction interval refactor for forecasters

SveaMeyer13:
#### Reference Issues/PRs
Start working on #984

Started with the refactor of the base class. Following STEP 12 and discussion in #984

* add todo's for what needs to be deprecated/where deprecation warnings will be needed
* have everything that is supposed to be deprecated in `predict()` in an additional else statement
* add abstract method with signature `predict_interval(self, fh=None, X=None, alpha=Union[float, np.ndarray[float, dim=1]]) -> pd.DataFrame` 
* make `predict()` call `predict_interval()` unless the boolean flag of `old_predict_interval_logic` is set to true
* move methods related to prediction intervals that are specific to one particular forecaster to that forecaster, for now -> We can further refactor them there
	* more specifically: move `_get_pred_int` to `_tbats`; move `compute_pred_int` to `theta`, completly delete `_get_pred_int` and directly call `_compute_pred_err()` instead. 

No.

#### What should the reviewers concentrate their feedback on
* I think I followed precisely what is suggested in STEP 12, except for this one: 
       * STEP 12 suggests moving `_get_pred_int` to theta, but I think it should be safe to completely delete it.
       
#### Any other comments?
- `predict()` now contains code duplicates but I thought it would be good for readability to have everything that is going to be deprecated in one else statement such that it can easily be deleted.
- As the next step I think we need to follow through with these changes in the inheriting classes before we can add the deprecation warning. But before implementing `_predict_interval()` in the inheriting classes, we should probably decide how the user can decide on the strategy for computing prediction intervals (either `PredIntervalAdder` or just as arguments in the constructor, see discussion about this in #984). 

pinging @fkiraly and @mloning

### DIFF
--- a/sktime/forecasting/base/_base.py
+++ b/sktime/forecasting/base/_base.py
@@ -27,7 +27,7 @@ copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
 """
 
 
-__author__ = ["Markus Löning", "@big-o", "fkiraly"]
+__author__ = ["mloning", "@big-o", "fkiraly"]
 __all__ = ["BaseForecaster"]
 
 from sktime.base import BaseEstimator
@@ -182,7 +182,14 @@ class BaseForecaster(BaseEstimator):
 
         return self
 
-    def predict(self, fh=None, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
+    def predict(
+        self,
+        fh=None,
+        X=None,
+        return_pred_int=False,
+        alpha=DEFAULT_ALPHA,
+        old_predict_interval_logic=True,
+    ):
         """Forecast time series at future horizon.
 
         Parameters
@@ -227,42 +234,71 @@ class BaseForecaster(BaseEstimator):
             store=None,
         )
 
-        # this should be here, but it breaks the ARIMA forecasters
-        #  that is because check_alpha converts to list, but ARIMA forecaster
-        #  doesn't do the check, and needs it as a float or it breaks
-        # todo: needs fixing in ARIMA and AutoARIMA
-        # alpha = check_alpha(alpha)
+        # this is how it is supposed to be after the refactor is complete and in place
+        if not return_pred_int:
+            y_pred = self._predict(
+                self.fh,
+                X=X_inner,
+                alpha=alpha,
+            )
 
-        y_pred = self._predict(
-            self.fh,
-            X=X_inner,
-            return_pred_int=return_pred_int,
-            alpha=alpha,
-        )
-
-        # todo: clean this up, predictive intervals should be returned by other method
-        if return_pred_int:
-            pred_int = y_pred[1]
-            y_pred = y_pred[0]
-
-        # convert to default output type, dependent on scitype
-        scitype_y = self.get_tag("scitype:y")
-        to_dict = {
-            "univariate": "pd.Series",
-            "multivariate": "pd.DataFrame",
-            "both": "pd.DataFrame",
-        }
-        y_out = convert_to(
-            y_pred,
-            to_dict[scitype_y],
-            as_scitype="Series",
-            store=self.converter_store_y,
-        )
-
-        if return_pred_int:
-            return (y_out, pred_int)
-        else:
+            # convert to default output type, dependent on scitype
+            scitype_y = self.get_tag("scitype:y")
+            to_dict = {
+                "univariate": "pd.Series",
+                "multivariate": "pd.DataFrame",
+                "both": "pd.DataFrame",
+            }
+            y_out = convert_to(
+                y_pred,
+                to_dict[scitype_y],
+                as_scitype="Series",
+                store=self.converter_store_y,
+            )
             return y_out
+
+        # keep following code for downward compatibility,
+        # todo: can be deleted once refactor is completed and effective
+        else:
+            # todo add deprecation waring once refactor
+            #  is also established in inheriting classes
+            # warn("return_pred_int in predict() is deprecated;
+            # use predict_interval().", DeprecationWarning)
+
+            # this should be here, but it breaks the ARIMA forecasters
+            #  that is because check_alpha converts to list, but ARIMA forecaster
+            #  doesn't do the check, and needs it as a float or it breaks
+            # todo: needs fixing in ARIMA and AutoARIMA
+            # alpha = check_alpha(alpha)
+
+            y_pred = self._predict(
+                self.fh,
+                X=X_inner,
+                return_pred_int=return_pred_int,
+                alpha=alpha,
+            )
+            if old_predict_interval_logic:
+                pred_int = y_pred[1]
+                y_pred = y_pred[0]
+            else:
+                y_pred = y_pred[0]
+                pred_int = self.predict_interval(fh=fh, X=X, alpha=alpha)
+
+            # convert to default output type, dependent on scitype
+            scitype_y = self.get_tag("scitype:y")
+            to_dict = {
+                "univariate": "pd.Series",
+                "multivariate": "pd.DataFrame",
+                "both": "pd.DataFrame",
+            }
+            y_out = convert_to(
+                y_pred,
+                to_dict[scitype_y],
+                as_scitype="Series",
+                store=self.converter_store_y,
+            )
+
+            return (y_out, pred_int)
 
     def fit_predict(
         self, y, X=None, fh=None, return_pred_int=False, alpha=DEFAULT_ALPHA
@@ -292,7 +328,7 @@ class BaseForecaster(BaseEstimator):
 
         return self._predict(fh=fh, X=X, return_pred_int=return_pred_int, alpha=alpha)
 
-    def compute_pred_int(self, y_pred, alpha=DEFAULT_ALPHA):
+    def predict_interval(self, fh=None, X=None, alpha=DEFAULT_ALPHA):
         """
         Compute/return prediction intervals for a forecast.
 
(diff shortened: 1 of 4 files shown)


fkiraly:
the failure in 3.9 is related to the sporadic bug https://github.com/alan-turing-institute/sktime/issues/1194 which *no one has fixed yet*

sktime/forecasting/base/_base.py
```diff
@@ -27,7 +27,7 @@ class name: BaseForecaster
 """
 
 
-__author__ = ["Markus Löning", "@big-o", "fkiraly"]
+__author__ = ["mloning", "@big-o", "fkiraly"]
```
fkiraly, line 30: while we're at it, I think the "at" from "big-o" should also go

sktime/forecasting/base/_base.py
```diff
@@ -182,7 +182,14 @@ def fit(self, y, X=None, fh=None):
...
         return self
 
-    def predict(self, fh=None, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
+    def predict(
+        self,
+        fh=None,
+        X=None,
+        return_pred_int=False,
+        alpha=DEFAULT_ALPHA,
+        old_predict_interval_logic=True,
```
fkiraly, line 191: I think this should be a tag, not an argument - it indicates, for any concrete forecaster, whether the logic for prediction intervals from `_predict` has been moved to `_predict_interval`.

fkiraly (changes requested):
Great start, really nice!
My comments have to do with a divergence from the design, on how downwards compatibility is implemented:

* I think the refactored logic is not quite as it should be: the idea is to remove the arguments `return_predict_int` and `alpha` from `_predict` - with prediction interval logic moving into `_predict_interval`. This needs to happen individually for each *concrete* forecaster.
* whether this move has happened, is indicated in each *concrete* forecaster by the `old_predict_logic` tag.
* in `_predict` of the base class, I would *completely remove* the arguments. They need to be deprecated in the concrete classes, one-by-one. I would go about it by writing a "refactor manual", i.e., for each concrete class, what do contributors have to do? The purpose of this would be to open an issue like #1146 to tick off the individual forceasters changed.
* `predict_interval` should also check `fh` and `X`, not just `alpha` - just like `predict` does. Now one could simply copy this over from `predict`, but copy-pasting code would go against the DRY principle. The general recommendation is to move the checks into functions, so you may have to think carefully about what to do here (and/or looking at `predict_update_single` etc)

SveaMeyer13:
Thanks for your review @fkiraly. Now I get what `old_predict_logic` is supposed to account for and I will update this PR accordingly. 

My initial train of thought for it was that currently the return type of `_predict(return_pred_int=True)[1]` differs from the one that is suggested in the STEP for `predict_intervals()`.  
( According to the docstring it is currently as follows (at least for theta, the other docstrings are vague): "A table of upper and lower bounds for each point prediction in `y_pred`. If `alpha` was iterable, then `intervals` will be a list of such tables.")

If we now change everything to the new return type as suggested in the STEP that might break the user's code for further processing those prediction intervals. So in my head, the flag was for returning everything either the old or the new way. But I now see how the processing logic and the actual return type are two pairs of shoes. 

Should we still have an option to return prediction intervals in the format in which they were returned before (with a deprecation warning)? Or should we entirely stick to the old format for now? Or is it just fine to have a new return format for prediction intervals?

fkiraly:
>So in my head, the flag was for returning everything either the old or the new way. But I now see how the processing logic and the actual return type are two pairs of shoes.

Ah, that's a good point!
Indeed, I was more concerned with the processing logic than the return type, and you've uncovered an inconsistency in my proposal for the deprecation logic - that direction of thinking that led to your deviation from the proposal in teh STEP is reasonable. I overlooked that we were also changing the return format (multiindex, column names).

More precisely, I agree with your reasoning on requirements during the deprecation cycle: a valid call with the old interface and parameters should return the old output, or at least a very similar output - not sure how to best ensure this. Perhaps by having some "old format" functionality in `predict_interval` that we will remove, while keeping the logic as much inside `predict_interval` as possible?

Which options that are reasonably low-complexity low-maintenance can you think of?

SveaMeyer13:
> `predict_interval` should also check `fh` and `X`, not just `alpha` - just like `predict` does. 
Now one could simply copy this over from `predict`, but copy-pasting code would go against the DRY principle. 
The general recommendation is to move the checks into functions, so you may have to think carefully about what to do here (and/or looking at `predict_update_single` etc)

For `X` I would use the checks you provide in PR #1436 once that is merged and for `fh`, `is_fitted`, and `alpha` I created a helper function - but I'm not sure if it is actually needed - so happy to remove it again if you don't like it. 

> Which options that are reasonably low-complexity low-maintenance can you think of?

Is think the best would be to have a function that converts the new output type into the old one. Like this, we could have `predict_interval()` only return the new type and wouldn't have to touch it again. And for now, we could also offer the old return type by converting to it in `predict()`. I made a stub for this. But I didn't want to write the converter function (new return type to old return type) as I wasn't sure if we already agreed on doing it the way it is proposed in the STEP.

fkiraly:
>Is think the best would be to have a function that converts the new output type into the old one. 

Excellent idea!

The control flow meanders a little, but the "downwards compatibliity" bit sits on top and we can just pick it off in deprecation.
We can also put the deprecation message inside the converter!

sktime/forecasting/base/_base.py
```diff
@@ -210,8 +213,7 @@ def predict(
             Prediction intervals
         """
         # handle inputs
-        self.check_is_fitted()
-        self._set_fh(fh)
+        self.perform_checks(to_be_checked={"is_fitted", "fh"}, fh=fh)
```
mloning, line 216: Do we need to wrap this into a method? I think I'd prefer to just see the lines of code here, unless perhaps this is re-used somewhere else?

sktime/forecasting/base/_base.py
```diff
@@ -229,36 +236,59 @@ def predict(self, fh=None, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
...
-            as_scitype="Series",
-            store=self.converter_store_y,
-        )
+        # keep following code for downward compatibility,
+        # todo: can be deleted once refactor is completed and effective
+        else:
+            if not self.get_tag("old_predict_interval_logic"):
+                warn(
+                    "return_pred_int in predict() is deprecated;"
+                    "use predict_interval().",
```
mloning, line 261:
```suggestion
                    "please use predict_interval() instead to generate prediction intervals.",
```

sktime/forecasting/base/_base.py
```diff
@@ -914,6 +912,7 @@ def _update_predict_single(
         fh,
         X=None,
         update_params=True,
+        # todo will be deprecated after refactor:
```
mloning, line 915: Can we add a `TODO: remove in v0.9.0` comment to the lines that we want to remove? Perhaps also add that info to the deprecation messages.

mloning (commented):
Thanks @SveaMeyer13 - I had a quick look, still need to get to grips with this a bit more, just a few questions at this point.

fkiraly (changes requested):
From discussion yesterday:

* this is really great, thanks!
* we agreed to merge #1436 first, which should solve the copy/pasting issue in the input conversions, checks and coercions; this will lead to some merge conflicts to resolve
* we should also add a test each time a new key interface point is added, thanks for catching that, @mloning
* we agreed to change the return type of `predict_interval` to `alpha` and `1-alpha` columns (no 0.5), please also PR the STEP for this

mloning:
One addition to the notes above: If alpha=None, we agreed to return results for `alpha=[0.95, 0.8, 0.5]`, where `0.5` needs to handled as a special case, as we do not want to repeat the values for upper and lower bounds.

fkiraly:
> One addition to the notes above: If alpha=None, we agreed to return results for `alpha=[0.95, 0.8, 0.5]` following the forecast package in R, where `0.5` needs to handled as a special case, as we do not want to repeat the values for upper and lower bounds.

I thought it should be `alpha=[0.05, 0.2, 0.5, 0.8, 0.95]`, which is a special case of the current specification where values are not "mirrored" if they are passed in a list. If al ist is passed, then predictive quantiles for exactly those values in the list are reported.

That's at least what is in the STEP and what we agreed to, since we agreed not to change the STEP in places other than what happens if the passed value is `float` or `None`.

fkiraly:
@mloning, if we still disagree, could you kindly write a spec or docstring what you think it should be? E.g., do we disagree what is returned in the case where `alpha` is a `list`?

mloning:
Yes, sorry, that's what I mean: 

* if `alpha=None`, `alpha` is internally taken as `[0.05, 0.2, 0.5, 0.8, 0.95]` (wasn't clear above)
* if alpha is a single float, we return `alpha` and `1-alpha` (from core dev voting)
* if alpha is a list of floats, we return `alpha` and `1-alpha` for each alpha in the list (to me, this follows intuitively from the single float case)

fkiraly:
@mloning, you have not fully specified the return format - which *ordering* would the values be in, given input? Also, how do you treat the case where `0.5` is contained in the list of `alpha`?

fkiraly:
PS: as you can hear from my question, I prefer the variant that makes it easier for the user to understand the precise relation between input and ouput. "it's the same alphas" is easier to understand than "it's the ordered sequence of unique alphas based on the collection of `alpha` and `1-alpha` for all `alpha` in the list".

But I'd first like to understand the precise alternative proposal.

fkiraly:
@sveameyer13, #1436 is in, and we've also aligned on the prediction interval signature (on the slack).

So all outstanding questions and conditionalities should be resolved now.

I suggest updating the STEP with the agreed-upon interface, too.

PR #1423, opened 2021-09-16
Datatypes enum

chrisholder:

This is a simple pr with the goal of making the datatypes modules easier to use. Currently the datatypes module requires str inputs for different datatypes. For example if I want to convert a given object to a numpy I would do something like:

convert_to(ts_obj, 'numpy3D')

This requires an extact string match which normally will require the developer to look up the possible strings in a notebook or at the documentation. This extra action of always having to look up the conversion can be solved very simply by allowing the functions in datatypes to accept enums like so:

convert_to(ts_obj, PanelMtype.NUMPY3D) 

This has the benefit of the developer not needing to know the exact string and in addition allows for an ide to autocomplete and suggest other potential datatypes. 

In addition with these enums I hope to implement a new testing utils that allows us to test over every potential type of ts input sktime supports and enums will make this much easier to achieve. 

No

While this isn't in stone and is currently only planned for use in this module, I've added a base registry enum class so other modules can potentially use enums as their registry I don't know if this is a good idea or where I did put this base class is an appropirate place for this to live.

##### For all contributions

##### For new estimators


fkiraly:
Really great idea in general!

As discussed earlier, my main thoughts are around how we avoid multiplication of "sources of truth" a for the conversion strings/enums and increasing the extension burden, i.e., through necessity to update the registry in multiple places in parallel when we introduce a new mtype.

Though, it seems like this PR breaks some things?
Do you mind looking into it?

chrisholder:
Fixing them now it is to do with me not checking for None - Added some extra tests to make sure this can't happen in the future as well since the modules tests didn't flag this to me

chrisholder:
> As discussed earlier, my main thoughts are around how we avoid multiplication of "sources of truth" a for the conversion strings/enums and increasing the extension burden, i.e., through necessity to update the registry in multiple places in parallel when we introduce a new mtype.

This pr I made it so the registry reads from the enum so there is only one place you need to add a new mtype (or even scitype if there is ever a need for that)

sktime/datatypes/_convert.py
```diff
@@ -68,29 +68,37 @@
...
 
 # pool convert_dict-s and infer_mtype_dict-s
 convert_dict = dict()
 convert_dict.update(convert_dict_Series)
 convert_dict.update(convert_dict_Panel)
 
 
-def convert(obj, from_type: str, to_type: str, as_scitype: str = None, store=None):
+def convert(
+    obj, from_type: Mtypes, to_type: Mtypes, as_scitype: Scitypes = None, store=None
```
fkiraly, line 89: I'd call it `Mtype` or `Scitype`, that makes more sense in the type hint?

sktime/datatypes/_convert.py
```diff
@@ -128,16 +143,22 @@ def convert(obj, from_type: str, to_type: str, as_scitype: str = None, store=Non
 
 
 # conversion based on queriable type to specified target
-def convert_to(obj, to_type: str, as_scitype: str = None, store=None):
+def convert_to(
+    obj,
+    to_type: Union[Mtypes, List[str], List[Mtypes]],
```
fkiraly, line 148: `List[str]` is unnecessary, since `Mtype` can also be a `str`, no?

fkiraly (changes requested):
Interesting idea!
Also, I appreciate that this is downwards compatible, great!

Not directly a change request, since I'm not sure how to address this, but two problems that I think need to be addressed or at least discussed before we merge:

* there is a difference in capitalization between the enum keys and the enum values. This can easily become confusing.
* Naive use of the enums will result in something like `convert_to(x, SeriesMtype.PD_DATAFRAME)`, this is quite clunky and seems confusing. Any way to improve this? I was thinking about `convert_to(x, Series.df)`, but shouldn't `Series` be the type appearing in `fit`/`predict` etc which would give a clash.

sktime/datatypes/_registry.py
```diff
@@ -51,32 +58,38 @@
...
+
+    SERIES = ("Series", "uni- or multivariate time series")
+    PANEL = ("Panel", "panel of uni- or multivariate time series")
 
 
-def mtype_to_scitype(mtype: str):
+SCITYPE_REGISTER = [tuple(scitype) for scitype in Scitype]
+
+
+def mtype_to_scitype(mtype: Union[PanelMtype, SeriesMtype, str]) -> str:
```
fkiraly, line 75: should this not just be `Mtype`?

sktime/datatypes/_convert.py
chrisholder: So I called it with the plural because the type itself is multiple things
```
Mtypes = Union[str, SeriesMtype, PanelMtype]
Scitypes = Union[str, Scitype]
```
Happy to change it to your suggestion however.

sktime/datatypes/_convert.py
chrisholder: Good spot

sktime/datatypes/_registry.py
chrisholder: Unfortunatly I can't import the types into this due to circular importing.

fkiraly: Argh - is there any way to resolve this, e.g., by splitting files?

That is, is the circularity coming through object imports, or just via module imports?

sktime/datatypes/_convert.py
fkiraly: aren't most types inhabited by multipe values? It's called `float` and not `floats`, `string` and not `strings`?

sktime/datatypes/_registry.py
chrisholder: Module imports since the types are based off the enums defined in that file. I can move it but it was where you orginally defined mtype_to_scitpye where would you want me to put it instead?

chrisholder:
> Interesting idea!
> Also, I appreciate that this is downwards compatible, great!
> 
> Not directly a change request, since I'm not sure how to address this, but two problems that I think need to be addressed or at least discussed before we merge:
> 
> * there is a difference in capitalization between the enum keys and the enum values. This can easily become confusing.
> * Naive use of the enums will result in something like `convert_to(x, SeriesMtype.PD_DATAFRAME)`, this is quite clunky and seems confusing. Any way to improve this? I was thinking about `convert_to(x, Series.df)`, but shouldn't `Series` be the type appearing in `fit`/`predict` etc which would give a clash.

So I've made some changes based on your feedback to the enums names. Not sure if they are perfect so now you would do 

convert_to(x, SeriesMtype.pd_dataframe);

I can rename the SeriesMtype to just 'Series' but am concerned this could further confuse the scitype stuff. Another thought I had could be access it through a chain of enums i.e.

convert_to(x, Scitype.Series.pd_dataframe)
convert_to(x, Scitpye.Panel.nested_univariate)

sktime/datatypes/_convert.py
chrisholder: that is a very good point :joy:

fkiraly:
@chrisholder, I think this is excellent still, in terms of ideas.

However, I think we really need to be clear as to the intended target interface, since the devil is in the details when it comes to useability.

Would you kindly be able to write a short summary or STEP that explains the "end state" ways to address the types and conversions? Including "old" and "new" way, to compare? And, intended usage?

We need to ensure:

* consistency of naming to avoid user confusion
* immediacy/simplicity of the interface
* we're close to "one obvious way" as per python zen (or at least, one quick way and one very configurable way)

fkiraly:
Also, @MatthewMiddlehurst - could you look at this from the perspective of consistency with your typing extension, please?

chrisholder:
Sorry this is late had something come up. Ill now outline the reasoning for the design in this pr and the problem it solves:

**Motivation**
For data containers we want to define ridgid categories that are distinct from one another. Due to the complex nature of time series, there are many ways the same timeseries could be represented. As such we don't want to restrict how users can store and pass data to various sktime utilities and models. Therefore instead of having one data container as many packages do (normally a class called TimeSeries), Sktime allows for a large list of different time series representations. This approach is a better user experience as it doesn't place a requirement on the user loading their time series data into a pre defined data container just for Sktime and additionally is more inline with sklearns philosophy.

However, this presents an issue for internal development which is: a model or utility designed for Sktime must be able to support numerous potential formats. @fkiraly solved this problem by creating the datatypes module that provides easy to use utility functions that allow for the conversion of one datatype to another. This allows the developer to call convert to a format of their choosing and the assume that format through the rest of their code. An example of this is given below:

```
def fit(X, Y):
   """
   Parameters
   ------------
   X: Mtypes
      Any valid sktime mtype
   Y: Mtypes
      Any valid sktime mtype
   """
   X = convert_to(X, 'numpy3D')
   Y = convert_to(Y, 'numpy3D')
   # Can now assume im working with a numpy3D here on 
```
In the example above X and Y could be any of the sktime valid mtypes, however, in two lines we ensured both X and Y were in the 'numpy3D' format which we can now work with for the rest of the model.

This works very well. However, one thing to notice is when choosing the format to convert to (in the example above converting to 'numpy3D'), we had to pass the desired convert 'to_type' as a string. This is because all of the Sktime mtypes are essentially just different formatted pd.Dataframe, pd.Series and np.ndarrays. As such it doesn't make a lot of sense defining a class for each of them as these would only be there to serve as named typing (would also require extra overhead of initialising and util methods etc.).

Defing the different mtypes as strings presents two main issues:
1) Errors will happen at runtime rather than compile time. Strings are validated as the code in running at the time they are passed to the convert functions. This leads to a slower and worse developer experience as errors occur at runtime rather than compile time. Additionally IDE's won't be able to pick up on the errors and highlight them (again because they are runtime errors).
2) As an extension of issue number 1, as the errors are at runtime, this requires the developer to be aware of the potential mtypes strings and know their exact spellings and capitlisation which is highly error prone and will result in most developers having to keep pulling up the documentation (or notebook) to check.
2.5) A more personal issue using strings doesn't allow for proper typing.

**Solution**
With these two (and a half) issues in mind this pr seeks to fix these two issues by defining the mtypes (and scitypes) of sktime as enums. As stated it doesn't make a lot of sense having a class for each mtype as they don't need all that boiler plate just for a named typing. An enum on the other hands doesn't require any extra boiler plate for new mtypes and scitypes but still has the benefits of typing and class (enum) specific logic. Additionally this instantiates the enums in the eyes of the compiler which means errors can be picked up at compile time and additionally allow IDEs to autosuggest potential mtypes and highlight errors. 

For example one of the greatest benefits of enum is my IDE can autosuggest potential enums thereby meaning I don't have to refer to the documentation everytime. If I put in my IDE:

`PanelMtype.`

then as soon as I put the '.' my IDE autosuggests all of the potential Panel options which greatly improves my efficiency and reduces errors.

Furthermore, as you can give the enums values (such as strings) this means the enums can be resolved to the strings and used in the convert function without the logic needed to be rewritten. This is achieved by overloading the enums str() method allowing for code like this:

```
if mtype is not None:
            mtype = str(mtype)

if scitype is not None:
           scitype = str(scitype)
```
which means only the first few lines of conert and convert_to were needed to be touched to support enums. This means non of the core logic has been touched.

Additionally using enums doesn't require anymore complexity when it comes to registering new mtypes and scitypes. Currently the datatypes module registers valid mtypes and scitypes using an array of tuples as shown below:

```
MTYPE_REGISTER_SERIES = [
    ("pd.Series", "Series", "pd.Series representation of a univariate series"),
    (
        "pd.DataFrame",
        "Series",
        "pd.DataFrame representation of a uni- or multivariate series",
    ),
    (
        "np.ndarray",
        "Series",
        "2D numpy.ndarray with rows=samples, cols=variables, index=integers",
    ),
]
```

The first value in the tuple is the string name, the second value is the scitype and the third value in the description. Therefore if I want to add a new mtype I would add a new tuple to this list. These are then read out to a list that is checked at runtime for valid types. This is currently done as follows:
```
MTYPE_REGISTER_SERIES = [tuple(mtype) for mtype in SeriesMtype]

MTYPE_LIST_SERIES = pd.DataFrame(MTYPE_REGISTER_SERIES)[0].values
```

With this in mind of how the module currently works, the goal when developing the enums for this module was to continue this simplicity and ease of adding new mtypes and scitypes with the developer not needing to perform any extra steps.

This has been achieved by creating the new BaseRegistryEnum, which allows enums that inherit it to essentially appear and work as a list of tuples (as the previous registries did). However, these enums will have all the benefits of enums outlined above. The new enums and how they will replace the existing registeries is shown below.
```
class SeriesMtype(BaseRegistryEnum):
    """Enum class for series mtypes."""

    pd_series = (
        "pd.Series",
        "pd.Series representation of a univariate series",
        "Series",
    )
    pd_dataframe = (
        "pd.DataFrame",
        "pd.DataFrame representation of a uni- or multivariate series",
        "Series",
    )
    np_array = (
        "np.ndarray",
        "2D numpy.ndarray with rows=samples, cols=variables, index=integers",
        "Series",
    )
```
Therefore if I wanted to add a new mtype here you would create a new enum value and pass it a tuple in the same format as I would if I was using the old registries. Overall this means the simplicity and ease of the previous registries it kept but their potential power is enhanced greatly by using enums. The enums after creation are then read to the exact same format as the previous registers were before using similar code:

`MTYPE_LIST_SERIES = pd.DataFrame(MTYPE_REGISTER_SERIES)[0].values`

Therefore the same result has been achieved with the same amount of code but the benefits of enums described above are now also felt.

Finally using enums allows for typing further ensuring maintainability of the code by allowing the creation of the following types:
```
Mtype = Union[str, SeriesMtype, PanelMtype]
SciType = Union[str, Scitype]
```
This allows for the convert functions to defined more clearly as shown:
```
def convert(
    obj, from_type: Mtype, to_type: Mtype, as_scitype: SciType = None, store=None
):
```
Overall this pr introduces enums WITHOUT changing any of the core logic of the datatypes module and allows strings to be continued to be used if the developer wishes. This PR enhances the datatypes module and greatly improves developer experience.

fkiraly:
Thanks!

Yes, this looks quite convincing.

Can I ask some questions:

* IMPORTANT: can you outline the steps that a user would do if they were to add a new mtype; and where they add a new scitype? I.e., where would things have to be added? What probably doesn't change is the logic added to convert dicts, example dicts, and check dicts; but how does it change on the registry level? Currently you'd add one or more entries to the register.
* can you be explicit which existing objects are *removed*, *replaced*, or *duplicated*? I'm uncomfortable with the fact that we may be duplicating the register and there is not a clear, unique source of truth.
* I'm uncomfortable with the fact that a single registry has been split in two, by the third field (`Panel` or `Series`), so the information which scitype an mtype belongs to is no longer easy to obtain using the new structure, e.g., as in the function `mtype_to_scitype`.
* The consequene of this split also seems to be that a new enum needs to be added per scitype. Is there a good alternative way to keep this in one place? One that leads to a call `Mtype.Series.pd_dataframe` perhaps?
* is there a good way we can use the scitypes themselves as type annotations? Following @ABostrom's suggestion of type aliases.

A line of thought: say we move the entire registry to an enum, and *remove* (or deprecate), not *duplicate* the registry within two enums, `Mtype`, and `Scitype`. How would this look like? Would an access like `Mtype.Series.pd_dataframe` be possible?

sktime/datatypes/_registry.py
fkiraly: really good question - perhaps the enums should move somewhere, e.g., `types`, which contains the enums and any logic that allows shorthand use as a `typing` type?

fkiraly:
?

chrisholder:
Just an fyi with the changes to the forecasting module all it is is where you had the convert module things using "pd.Series", "np.ndarray" and "pd.Dataframe", I've just refactored them to instead use Datatypes.Series.pd_series, Datatypes.Series.np_array and Datatypes.Series.pd_dataframe. No logic has been touched and at runtime these are resolved back to the same strings.

fkiraly:
@chrisholder, could we keep these open? I might be using this someday!

fkiraly:
I'm reopening as a draft to have this as a basis for potential m/scitype type hinting and design discussion. No expectation for @chrisholder to work on this.

PR #1431, opened 2021-09-17
Release v0.8.0

fkiraly:
0.8.0 changelog

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 > A unified interface for machine learning with time series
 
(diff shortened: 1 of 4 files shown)


mloning (approved):

mloning:
Thanks @fkiraly! Feel free to merge whenever you think it's good to go. We can ignore the CI checks I think. 

Going forward, if we re-write the changelog before making a new release, my preference would be to update the changelog incrementally in each PR rather than have a big rewrite at the end.

PR #1311, opened 2021-08-16
forecasting: input checks in update_predict

fkiraly:
`update_predict` was missing input checks on `y` and `X` compliant with multivariate forecasting, this has been added

PR #1436, opened 2021-09-19
Update refactor: forecaster x/y checks

fkiraly:
This PR is a first step towards addressing #982, and continues #1311. It should also help @viktorkaz with the `update_predict` issue in #856, and help @SveaMeyer13 to avoid boilerplate code in #1421.

What this does:

* unify the input checks - move repetitive boilerplate code for input checks and conversions to a new `_check_X_y` function which is now called in `fit`, `predict`, `fit_predict`, `update`, `update_predict`, `update_predict_single`.
* plus a shorthand `_check_X` which is the same for only `X`
* due to #982 not done, the input checks were inconsistent in the `update_predict` family of methods. This has now been fixed.
* `_update_y_X` has been updated to work with `np.ndarray` type input as well.
* `converter_store_y` has been renamed to `_converter_store_y` since it is a private variable
* `update_predict_single` was using an argument named `y_new` instead of `y` - for consistency, `y` has been introduced as an alias of `y_new` and a deprecation note added.
* substantial update to the docstrings to make assumptions and consequences much more precise


fkiraly:
FYI @SveaMeyer13 - this PR also addresses the repetition of input checks/conversions we were discussing in #1421.

I.e., if you write `predict_interval` etc, you can re-use `_check_X_y` and the surrounding few lines from `predict`, instead of having to copy-paste a dozen or so lines.

sktime/forecasting/base/_base.py
```diff
@@ -592,14 +502,69 @@ def _set_y_X(self, y, X=None, enforce_index_type=None):
...
+
+        # convert X and y to a supported internal mtype
+        #  it X/y mtype is already supported, no conversion takes place
+        #  if X/y is None, then no conversion takes place (returns None)
+        y_inner_mtype = self.get_tag("y_inner_mtype")
+        y_inner = convert_to(
+            y,
+            to_type=y_inner_mtype,
+            as_scitype="Series",  # we are dealing with series
+            store=self.converter_store_y,
```
mloning, line 548: As discussed, let's make this private
```suggestion
            store=self._converter_store_y,
```

sktime/forecasting/base/_base.py
```diff
@@ -216,18 +173,8 @@ def predict(self, fh=None, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
...
-        # convert X if needed
-        X_inner_mtype = self.get_tag("X_inner_mtype")
-        X_inner = convert_to(
-            X,
-            to_type=X_inner_mtype,
-            as_scitype="Series",  # we are dealing with series
-            store=None,
-        )
+        # input check and conversion for X
+        X_inner, _ = self._check_X_y(X=X)
```
mloning, line 177: Would it make sense to have a separate `_check_X` method, that could be then be called inside `_check_X_y` and would avoid the somewhat awkward second return object. With the new probabilistic prediction methods there will be a number of methods that only need to check X. 
```suggestion
        X_inner = self._check_X(X)
```

mloning (commented):
Generally looks good and in favour of getting this in, two comments below.

sktime/forecasting/base/_base.py
fkiraly: how about a simple shorthand instead, i.e., `_check_X()` is `_check_X_y()[0]`?

In terms of flow logic, that's much simpler than your proposed "splitting up" of `_check_X_y`, and easier to understand (since the code in `check_X_y` is ordered by concerns across X/y, whereas the splitting would create two functions in which some concerns are contained in `_check_X_y` and some in `check_X`).

mloning: I find it unintuitive to see `check_X_y` in methods that do not take in `y`. So I would prefer splitting the methods, I think it should be straightforward, but I may be missing something.

sktime/forecasting/base/_base.py
fkiraly: done

fkiraly:
@mloning, I think I've addressed your comments.

sktime/forecasting/base/_base.py
```diff
@@ -266,26 +229,64 @@ def fit_predict(
...
-        y_pred : pd.Series
-            Point predictions
+        y_pred : pd.Series, pd.DataFrame, or np.ndarray (1D or 2D)
+            Point forecasts at fh, with same index as fh
+            y_pred has same type as y
         y_pred_int : pd.DataFrame - only if return_pred_int=True
+            in this case, return is 2-tuple (otherwise a single y_pred)
             Prediction intervals
+
+        State change
```
mloning, line 258: These are not supported headers for the numpy docstring conversation. Could you move them under "Notes"? Same for "Writes to self" (see below)

sktime/forecasting/base/_base.py
```diff
@@ -266,26 +229,64 @@ def fit_predict(
...
+        ------------
+        Changes state to "fitted".
+
+        Writes to self
```
mloning, line 262: same here

sktime/forecasting/base/_base.py
```diff
@@ -342,78 +343,40 @@ def update(self, y, X=None, update_params=True):
...
         Returns
         -------
         self : reference to self
 
-        Notes
-        -----
-        Update self._y and self._X with `y` and `X`, respectively.
-        Updates  self._cutoff to last index seen in `y`. If update_params=True,
-        updates fitted model that updates attributes ending in "_".
+        Writes to self
```
mloning, line 363: same here

sktime/forecasting/base/_base.py
```diff
@@ -431,19 +394,37 @@ def update_predict(
...
-            Point predictions
-        y_pred_int : pd.DataFrame
+        y_pred : pd.Series, pd.DataFrame, or np.ndarray (1D or 2D)
+            Point forecasts at fh, with same index as fh
+            y_pred has same type as y
+        y_pred_int : pd.DataFrame - only if return_pred_int=True
+            in this case, return is 2-tuple (otherwise a single y_pred)
             Prediction intervals
+
+        Writes to self
```
mloning, line 422: same here

sktime/forecasting/base/_base.py
```diff
@@ -518,17 +491,39 @@ def update_predict_single(
...
         -------
-        y_pred : pd.Series
-            Point predictions
+        y_pred : pd.Series, pd.DataFrame, or np.ndarray (1D or 2D)
+            Point forecasts at fh, with same index as fh
+            y_pred has same type as y
         pred_ints : pd.DataFrame
             Prediction intervals
+
+        Writes to self
```
mloning, line 500: same here

sktime/forecasting/base/_base.py
```diff
@@ -591,30 +592,156 @@ def _set_y_X(self, y, X=None, enforce_index_type=None):
...
+        ------
+        TypeError if y or X is not one of the permissible Series mtypes
+        TypeError if y is not compatible with self.get_tag("scitype:y")
+            if tag value is "univariate", y must be univariate
+            if tag value is "multivariate", y must be bi- or higher-variate
+            if tag vaule is "both", y can be either
+        TypeError if self.get_tag("X-y-must-have-same-index") is True
+            and the index set of X is not a super-set of the index set of y
+
+        Writes to self
```
mloning, line 624: same here

sktime/forecasting/base/_base.py
```diff
@@ -591,30 +592,156 @@ def _set_y_X(self, y, X=None, enforce_index_type=None):
...
         Parameters
         ----------
-        y : pd.Series or pd.DataFrame
+        y : pd.Series, pd.DataFrame, or nd.nparray (1D or 2D)
             Endogenous time series
-        X : pd.DataFrame, optional (default=None)
+        X : pd.DataFrame or 2D np.ndarray, optional (default=None)
             Exogenous time series
+
+        Reads from self
```
mloning, line 703: same here

sktime/forecasting/base/_base.py
```diff
@@ -591,30 +592,156 @@ def _set_y_X(self, y, X=None, enforce_index_type=None):
...
+        ---------------
+        _y : only if exists, then assumed same type as y and same cols
+        _X : only if exists, then assumed same type as X and same cols
+            these assumptions should be guaranteed by calls
+
+        Writes to self
```
mloning, line 709: same here

sktime/forecasting/base/_base.py
```diff
@@ -886,11 +1014,10 @@ def _update(self, y, X=None, update_params=True):
         y_pred_int : pd.DataFrame - only if return_pred_int=True
             Prediction intervals
 
-        Notes
-        -----
-        Update self._y and self._X with `y` and `X`, respectively.
-        Updates  self._cutoff to last index seen in `y`. If update_params=True,
-        updates fitted model that updates attributes ending in "_".
+        Writes to self
```
mloning, line 1017: same here

sktime/forecasting/base/_base.py
```diff
@@ -591,30 +592,156 @@ def _set_y_X(self, y, X=None, enforce_index_type=None):
...
                 self._X = X.combine_first(self._X)
 
     def _update_y_X(self, y, X=None, enforce_index_type=None):
-        """Update training data.
+        """Update internal memory of seen training data.
 
         Parameters
         ----------
-        y : pd.Series or pd.DataFrame
+        y : pd.Series, pd.DataFrame, or nd.nparray (1D or 2D)
```
mloning, line 698: Does that mean we now accept numpy arrays internally?

sktime/forecasting/base/_base.py
fkiraly: you mean "convention"?
I don't think we settled on a docstring convention, and I think it's important to write the guarantees/assumes here out explicitly, on what the function does.

the `numpy` convention, imo, is mostly for pure functions without side effects and/or being class/object methods.
For class methods, you should document guarantees/assumes for side effects that are specific to object variables.

I don't think "notes" is an appropriate section here.

Can we leave this in - since I think it's important for users and developers - and have a separate discussion about our docstring sections?

sktime/forecasting/base/_base.py
fkiraly: yes, since 0.7.something, same version that added `pd.DataFrame` support.
The docstrings have just not been updated - in fact, many of them still referred to `pd.Series` only.

fkiraly: I would recommend to update our tests too, to test for inputs of all the types and different number of columns.
I would, however, strongly recommend to adopt the improved fixtures logic in #1407 - this will make it much easier to create X/y input fixtures dependent on estimator tags.

fkiraly: (I be strongly against adding more nested loops to the current test logic, that architecture seems quite brittle to me)

sktime/forecasting/base/_base.py
mloning: It generates lots of errors in the docs generation, which makes it really hard to debug it. We've had this discussion during the doc sprint and agreed to rename these sections. So the easiest is to add this under "Notes" as it previously was.

sktime/forecasting/base/_base.py
mloning: Oh I must have missed that, I thought we had agreed to internally only support `pd.DataFrame` or `pd.Series`. Could you point me to the PR where this was added?

sktime/forecasting/base/_base.py
fkiraly: Please trust me on software engineering standards for docstrings - in this case I'm pretty sure that the numpy standard is inadequate when it comes to classes and class/object methods.

The "notes" section is for implementation details or background theory.
What I'm talking about is about assumptions/guarantees that are required to document by ["design by contract"](https://en.wikipedia.org/wiki/Design_by_contract) best pracatice.

fkiraly: >It generates lots of errors in the docs generation, which makes it really hard to debug it.

Where exactly can I see these errors?

>We've had this discussion during the doc sprint and agreed to rename these sections. 

What I remember is that we agreed to be compliant with the current testing suite (which it doesn't bother?)

As a compromise, how about intead of moving to "notes", we move to "parameters" and "returns", aligning "parameters" with "assumptions" and "returns" with "guarantees"? With explanation if the write/read is with the class.

fkiraly: What I want to avoid is that assumptions and side effects are not documented or not documented effectively (e.g., in a section where no one expects or reads them).

sktime/forecasting/base/_base.py
fkiraly: If you'd like my view on the history here:

an early discussion on an "is" state where we are simultaneously using `np.array` and the two `pandas` types for time series is in STEP05, here:
https://github.com/sktime/enhancement-proposals/blob/main/steps/05_scitype_based_IO_checks/step.md

We never had proper "support" for `np.array` since we did not have proper API documentation (including assumes/promises on input/output types) - but as the STEP explains, there has been inconsistent use or "support" throughout the codebase at least since late 2020, probably since 2019.

I think we can't localize this to a PR, but it's more an "organically grown" thing where different contributors would make different assumptions about the target types to support.

The first PR which properly documents the different concrete data containers in use is my PR #1232, and I think it is really the first proper attempt at consolidating the situation. This was done based on sifting through the codebase "as-is", and uncovered that we were using 3 mtypes for individual time series, and 7 mtypes for time series panels.

The observation - made before me documenting the "is" state - is what motivated the creation of the `datatypes` module.

fkiraly: at the "end" of this history:

in this PR, besides refactoring the checking logic, I'm trying to move the docstrings to professional grade quality, including clear statements on assumes/guarantees.

As a reviewer, @mloning, I think you shouldn't fault this PR for any property of the "is" state that is not changed by the PR, e.g., the possibility to pass `numpy` objects. You can, of course, ask for making less promises in the docstrings, or discuss formatting issues.

fkiraly:
@SveaMeyer13, @aiwalter, FYI - this PR is relevant for your downstream issues, but it looks like this PR may get delayed or blocked mainly over a discussion over docstrings and/or aspects of the history and roadmap of `sktime` designs to which the docstrings are only peripherally related in my opinion.

Would appreciate your comments here.

I've invested some time in trying to make the docstrings professional grade and precise (while, frankly, expecting no contention about them), but I'd also be happy to revert to the old docstrings where applicable if the changes are controversial.

I acknowledge the issue with the numpy style and made a constructive suggestion to move guarantees/assumes in the parameters/results section, which I'm also happy to implement.

sktime/forecasting/base/_base.py
mloning: As said, we've had this discussion before and decided to only use the sections supported by the convention. @RNKuhns spent a lot of time cleaning up the docstrings to make them more compliant. You can see the errors that are generated here: https://readthedocs.org/projects/sktime/builds/14821966/. If we want to add other sections, we should perhaps try to extend the NumPy docstring convention or somehow make sure that we're not generating all these errors.

fkiraly: Thanks for pointing me towards the errors - why are these not displayed in the checks? Also, can you tell me what the general process is to see these?

As said above, I'm happy to accommodate by moving the reads/writes to the parameters and results section. What do you think about this alternative suggestion?

fkiraly:
@mloning, don't you want to reply to my proposal regarding the docstrings?
@SveaMeyer13 is waiting for this PR.

sktime/forecasting/base/_base.py
mloning: Don't think that the "Parameters" or "Returns" sections are intended for that info (see https://numpydoc.readthedocs.io/en/latest/format.html).

mloning:
@fkiraly replied now (see above)

sktime/forecasting/base/_base.py
fkiraly: @mloning, can you point me to a resource that explains how I can see the build log from readthedocs?
When I click "details", it directs me to a mockup of the post-PR actual documentation, not to the logs.

fkiraly: >Don't think that the "Parameters" or "Returns" sections are intended for that info

Yes, I know - but I think it's more important to have crucial information in the docstring than adherence to a particular documentation style if it doesn't accommodate for full specification of interface contracts.

How about "extended description"? That's a free text section.

fkiraly: I mean "extended summary", it sounds like anything can go in there related to functionality, and it would be a good fit for contracts?

fkiraly: resolved, formatting changed, numpydoc no longer raises "unknown section" errors

sktime/forecasting/base/_base.py
fkiraly: resolved, formatting changed, numpydoc no longer raises "unknown section" errors

sktime/forecasting/base/_base.py
fkiraly: resolved, formatting changed, numpydoc no longer raises "unknown section" errors

sktime/forecasting/base/_base.py
fkiraly: resolved, formatting changed, numpydoc no longer raises "unknown section" errors

sktime/forecasting/base/_base.py
fkiraly: resolved, formatting changed, numpydoc no longer raises "unknown section" errors

sktime/forecasting/base/_base.py
fkiraly: resolved, formatting changed, numpydoc no longer raises "unknown section" errors

sktime/forecasting/base/_base.py
fkiraly: resolved, formatting changed, numpydoc no longer raises "unknown section" errors

sktime/forecasting/base/_base.py
fkiraly: resolved, formatting changed, numpydoc no longer raises "unknown section" errors

sktime/forecasting/base/_base.py
fkiraly: resolved, formatting changed, numpydoc no longer raises "unknown section" errors

fkiraly:
The outstanding item has been resolved - @mloning's request to reformat the new docstring so they are `numpydoc` compliant.

Since there were no comments about content, code, or functionality for 9 days, I'll merge this directly, in order to unblock @SveaMeyer13 with her work on prediction interval refactcor.

If docstrings further need to be worked on, anyone is free to do so in another PR dedicated to docstrings.

Issue #1441, opened 2021-09-21
[BUG] forecasting in sample and out of sample with prediction intervals in ARIMA/AutoARIMA


IlyasMoutawwakil:
> **Describe the bug**

There's a small problem with ARIMA/AutoARIMA when forecasting in sample and out of sample with prediction intervals. In the _pmdarima adapter, when `return_pred_int` is False, `_predict_in_sample` returns a pd.Series but when it's True, it returns a tuple (pd.Series, pd.DataFrame) which raises the append error.

> **To Reproduce**

```python
from sktime.forecasting.arima import ARIMA
import pandas as pd
import numpy as np

y_train = pd.Series(np.arange(1,21), index=pd.date_range(start='1/1/2021', end='20/1/2021'))

arima_forecaster = ARIMA(order=(1, 1, 1), suppress_warnings=True)
arima_forecaster.fit(y_train)
arima_forecaster.predict(np.array([-18, -14, -10, 5]), return_pred_int=True)
```

> **Expected behavior**

To return predictions with prediction intervals

> **Additional context**

Traceback :
```
AttributeError                            Traceback (most recent call last)
<ipython-input-10-d49673c473fc> in <module>()
----> 1 arima_forecaster.predict(np.array([-18, -14, -10, 5]), return_pred_int=True)

1 frames
/usr/local/lib/python3.7/dist-packages/sktime/forecasting/base/_base.py in predict(self, fh, X, return_pred_int, alpha)
    240             X=X_inner,
    241             return_pred_int=return_pred_int,
--> 242             alpha=alpha,
    243         )
    244 

/usr/local/lib/python3.7/dist-packages/sktime/forecasting/base/adapters/_pmdarima.py in _predict(self, fh, X, return_pred_int, alpha)
     69             y_ins = self._predict_in_sample(fh_ins, **kwargs)
     70             y_oos = self._predict_fixed_cutoff(fh_oos, **kwargs)
---> 71             return y_ins.append(y_oos)
     72 
     73     def _predict_in_sample(

AttributeError: 'tuple' object has no attribute 'append'
```

> **Versions**

<details>

I've run this one on a clean colab notebook after a `!pip install sktime[all_extras]`

</details>

mloning:
Thanks for raising the issue @IlyasMoutawwakil - we're currently working on refactoring the interface for probabilistic forecasting interval x(see #1421). Help would be much appreciated for fixing the individual forecasters once #1421 is merged.

[...]

PR #1442, opened 2021-09-21
Pandas.NA for unpredictible insample forecasts

### LINKED ISSUE #859: ARIMA and AutoARIMA should return pandas.NA when value cannot be predicted, e.g., first index in-sample for start=0, d=1
Update 2021-09-15 - re-opened since a recurring issue.

Not a bug, but an interface convention - currently ARIMA, AutoARIMA and other learners fail when asked for a prediction that is undefined, e.g., difference parameter d=1 and first in-sample index (start=0).

It would be more sensible if `pandas.NA` is returned instead for such indices, perhaps a warning should be raised too that `pandas.NA` are produced.

In first instance, we should update ARIMA and AutoARIMA with this convention.

---

**Describe the bug**

When using `Imputer` with `forecaster = AutoARIMA()`  the error "ValueError: In-sample predictions undefined for start=0 when d=1" is raised.

```
<ipython-input-11-2d05f4822fd6> in <module>()
      4 
      5 transformer = Imputer(method="forecaster", forecaster=AutoARIMA()).fit(Z=df.value)
----> 6 y_hat = transformer.transform(Z=df.value)

/usr/local/lib/python3.7/dist-packages/sktime/transformations/series/impute.py in transform(self, Z, X)
    102             z_pred = forecaster.fit(
    103                 z.fillna(method="ffill").fillna(method="backfill")
--> 104             ).predict(fh=fh_ins)
    105             # fill with trend values
    106             z = z.fillna(value=z_pred)

/usr/local/lib/python3.7/dist-packages/sktime/forecasting/base/_sktime.py in predict(self, fh, X, return_pred_int, alpha)
    226         self.check_is_fitted()
    227         self._set_fh(fh)
--> 228         return self._predict(self.fh, X, return_pred_int=return_pred_int, alpha=alpha)
    229 
    230     def compute_pred_int(self, y_pred, alpha=DEFAULT_ALPHA):

/usr/local/lib/python3.7/dist-packages/sktime/forecasting/base/adapters/_pmdarima.py in _predict(self, fh, X, return_pred_int, alpha)
     58         # all values are in-sample
     59         elif fh.is_all_in_sample(self.cutoff):
---> 60             return self._predict_in_sample(fh_ins, **kwargs)
     61 
     62         # both in-sample and out-of-sample values

/usr/local/lib/python3.7/dist-packages/sktime/forecasting/base/adapters/_pmdarima.py in _predict_in_sample(self, fh, X, return_pred_int, alpha)
     80             X=X,
     81             return_conf_int=return_pred_int,
---> 82             alpha=alpha,
     83         )
     84 

/usr/local/lib/python3.7/dist-packages/pmdarima/utils/metaestimators.py in <lambda>(*args, **kwargs)
     51 
     52         # lambda, but not partial, allows help() to work with update_wrapper
---> 53         out = (lambda *args, **kwargs: self.fn(obj, *args, **kwargs))
     54         # update the docstring of the returned function
     55         update_wrapper(out, self.fn)

/usr/local/lib/python3.7/dist-packages/pmdarima/arima/auto.py in predict_in_sample(self, X, start, end, dynamic, return_conf_int, alpha, typ, **kwargs)
    242             return_conf_int=return_conf_int,
    243             alpha=alpha,
--> 244             typ=typ,
    245         )
    246 

/usr/local/lib/python3.7/dist-packages/pmdarima/arima/arima.py in predict_in_sample(self, X, start, end, dynamic, return_conf_int, alpha, **kwargs)
    584         if start is not None and start < d:
    585             raise ValueError("In-sample predictions undefined for start={0} "
--> 586                              "when d={1}".format(start, d))
    587 
    588         # if we fit with exog, make sure one was passed:

ValueError: In-sample predictions undefined for start=0 when d=1
```

**To Reproduce**

Data:
[df.csv](https://github.com/alan-turing-institute/sktime/files/6426660/df.csv) 

```python
import pandas as pd

df = pd.read_csv('df.csv')  
df['timestamp'] = pd.DatetimeIndex(df['timestamp']).to_period('D')
df = df.set_index('timestamp')

transformer = Imputer(method="forecaster", forecaster=AutoARIMA()).fit(Z=df.value)
y_hat = transformer.transform(Z=df.value)
```

**Expected behavior**

`transformer.transform(Z)` should return `Z` with NAn values replaced by AutoArima predictions.

**Additional context**

I think the problem is in `/usr/local/lib/python3.7/dist-packages/sktime/transformations/series/impute.py` in `transform(self, Z, X)`, line 100, because all the points are being passed to `fh_ins` instead of just the points with NaN.

```python
100             fh_ins = -np.arange(len(z))
101             # fill NaN before fitting with ffill and backfill (heuristic)
102             z_pred = forecaster.fit(
103                 z.fillna(method="ffill").fillna(method="backfill")
104             ).predict(fh=fh_ins)
105             # fill with trend values
106             z = z.fillna(value=z_pred)

```

**Versions**
<details>

```
System:
    python: 3.7.10 (default, Feb 20 2021, 21:17:23)  [GCC 7.5.0]
executable: /usr/bin/python3
   machine: Linux-4.19.112+-x86_64-with-Ubuntu-18.04-bionic

Python dependencies:
          pip: 19.3.1
   setuptools: 56.0.0
      sklearn: 0.24.2
       sktime: 0.6.0
  statsmodels: 0.12.2
        numpy: 1.19.5
        scipy: 1.4.1
       Cython: 0.29.22
       pandas: 1.1.5
   matplotlib: 3.2.2
       joblib: 1.0.1
        numba: 0.51.2
     pmdarima: 1.8.2
      tsfresh: None
```

</details>

IlyasMoutawwakil:
> #### Reference Issues/PRs
Fixes #859 and #1441 

When training an AutoArima you don't know what differencing order it'll land at so if data changes at some point you won't be able to do in sample forecasts for some points (earlier to diff order). This implementation makes predictions possible (without errors) if they're ealier to differencing order.
It also fixes a small problem with appending out of sample to in sample predictions when intervals are to be returned.

> #### Does your contribution introduce a new dependency? If yes, which one?

Nope

> #### What should a reviewer concentrate their feedback on?

if the code is clear enough, if the comments should be left.

> #### PR checklist

> ##### For all contributions

> ##### For new estimators


sktime/forecasting/base/adapters/_pmdarima.py
```diff
@@ -89,17 +111,43 @@ def _predict_in_sample(
...
+            fh_idx = fh_idx[fh_idx >= self.order[1]] - self.order[1]
+
         if return_pred_int:
             # unpack and format results
             y_pred, pred_int = result
-            y_pred = pd.Series(y_pred[fh_idx], index=fh_abs)
-            pred_int = pd.DataFrame(
-                pred_int[fh_idx, :], index=fh_abs, columns=["lower", "upper"]
-            )
+            if return_pred_na:
```
fkiraly, line 124: it would be easier to read to keep the old code, and then have one block `if return_pred_na` which uses `np.concatenate` to pre-pend `y_pred = np.concatenate(stuff, y_pred)`

fkiraly (approved):
Seems to work, as intended!

While we wait for a potential second reviewer, you could:

* add docstrings to the class and private functions, while we're here - you've understood what they do, so let's add full docstrings. Not a blocker though.
* I left a comment on the decision/branching at the end, the alternative structure might be easier to read but it's not a blocker.

sktime/forecasting/base/adapters/_pmdarima.py
IlyasMoutawwakil: Yes sounds cleaner and more understandable. I will do that.

IlyasMoutawwakil:
@fkiraly 
There's still one thing that doesn't make sense (or just seems weird) about the way pmdarima is interfaced. 
So we are allowing seperated steps forecasting (like `fh=[-10, -5, 3]`) but in pmdarima, we actually make a forecast from -10 to 3. The documentation says:

```
X: pd.DataFrame, optional (default=None)
Exogeneous data
```
But it doesn't explain that X should have all exogenous values from relative step -10 to 3.

fkiraly:
@IlyasMoutawwakil, that's an interesting point - there's actually two related things here:

* telling the user about the requirement to pass all indices between start/end
* representing the requirement.
* whether we can get around the behaviour

For the first thing, we are using tags, which you can list by `registry.all_tags`.
We cannot use the `_fit` etc docstring directly, since the user facing docstring is the `fit` in the base class - but we probably should state assumptions in the docstrings, related to the tags. I've opened an issue here #1451.

Regarding tags:
For forecasters, we have `"X-y-must-have-same-index"` which covers the case where `X` indices need to contain those in `y` (being forecast or observed).
This does not cover the above case, as you state. So I wonder how we should indicate this as a tag, or whether we should change the meaning of the existing tag. Or, otherwise, how we tell the user.

Any good ideas, @IlyasMoutawwakil ?

fkiraly:
@IlyasMoutawwakil, I think we should discuss the index issue separately - perhaps open an issue for it? It's a valuable observation imo, we should think about how we communicate the requirement on the inputs.

sktime/forecasting/base/adapters/_pmdarima.py
fkiraly: this isn't done yet, is it?

IlyasMoutawwakil: I hope this is what you had in mind ^^

sktime/forecasting/base/adapters/_pmdarima.py
```diff
@@ -66,9 +66,18 @@ def _predict(self, fh, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
 
         # both in-sample and out-of-sample values
         else:
-            y_ins = self._predict_in_sample(fh_ins, **kwargs)
-            y_oos = self._predict_fixed_cutoff(fh_oos, **kwargs)
-            return y_ins.append(y_oos)
+            if return_pred_int:
```
fkiraly, line 69: what are you doing here? This does not look correct.
Sorry for not noticing earlier.

The return type of `self._predict_in_sample(fh_ins, **kwargs)` does not depend on the value of `return_pred_it` - this will break if it's `True`!

It's also unrelated to the change you wanted to make - I'd recommend to simply revert this.

sktime/forecasting/base/adapters/_pmdarima.py
fkiraly: no, not quite - I was asking you to reduce nesting layer and repetition of `if`-s, or this will be hard to read and debug/extend.

In pseudo-code:

1. compute `y_pred`
2. if `return_pred_na`, adapt the `fh_` variables
3. if `return_pred_int`, compute prediction intervals (without NAs)
4. if `return_pred_na`, prepend NAs to `y_pred` and `pred_int` (if exists)
5. return `y_pred` and possibly `pred_int` (if `return_pred_int`)

Maybe it's also possible to reduce the `if return_pred_na` blocks to a single one?

fkiraly: for instance, initializing `pd.Series` and `pd.DataFrame` with all-NA at the start, and only filling in the "valid" predictions? That would avoid index magic and prepending etc.

sktime/forecasting/base/adapters/_pmdarima.py
IlyasMoutawwakil: I tried solving this issue #1441  at the same time since I can't do tests properly with it raising an error in every attempt that tuple append.

fkiraly: ahhhhhh - that's why!

I guess it's fine as a temporary, hacky fix then, but it should not stay this way for too long.

Would you be sticking around to fix it completely, after #1421 is merged?
Might be good to have your eyes on this since you've seen the internals.

sktime/forecasting/base/adapters/_pmdarima.py
IlyasMoutawwakil: Actually it is possible since the NA series are made with length fh_na which can be zero, I can always prepend it without verifying the condition.

fkiraly: great!
looking forward to it.

IlyasMoutawwakil: I guess It looks cleaner now, thou Ihad to rearrange the order of some stuff for less code and parameters.

fkiraly: Hm yes - though the tests fail now...

The tests that fail check whether the indices that are being predicted are indeed those that predictions were asked for. Probably some minor bug in there?

IlyasMoutawwakil: ooh it's not actually that, it's just that I did my tests only with ARIMA while AutoARIMA hides its order under another attribute layer (`self._forecaster.model_.order`). I'll fix it right away.

fkiraly (approved):
Yes this looks good now - clear logic, good to read, and should also fix the #1441 bug.

Thanks!

Issue #929, opened 2021-06-04
[ENH] VAR and VECM models


GitHunter0:
@fkiraly update: almost done, all that is missing is `predict_interval` for VECM.

---

@fkiraly's summary of the discussion:

We should interface VAR and VECM from `statsmodels`.

- [x] VAR #1083, #1491
- [x] VECM #1141, #2829

---

Have you considered porting `Vector Autoregressions (VAR)` and `Vector Error Correction Models (VECM)` from [statsmodels](https://www.statsmodels.org/dev/vector_ar.html#)? This would be a great addition. Thank you

fkiraly:
we first need to consolidate the multivariate forecasting interface, see #919 and #228

fkiraly:
but yes, important next step - join the dev days if interested #863

GitHunter0:
I'm very glad to know, thank you for the feedback

fkiraly:
As discussed in the latest mentoring meeting @thayeylolu will be working on this.

The interface will assume `pd.DataFrame` as input to `_fit`, and as return from `_predict`, no other types for the moment.

This means it should eventually be compatible irrespective of whichever of the two proposed designs #980 and #1074 we go with.

FYI, @mloning.

fkiraly:
Other implementation notes

* we should start by writing a short user journey design
* the statsmodels adapter may be a good way to go, but needs investigating
* tests may also need looking at, e.g., multivariate input generation

thayeylolu:
User Journey Design

```python
class VectorAutoRegression(_StatsModelsAdapter)
```
- The tag`univariate-only` is set to false
- Constructor's parameter `dates` as used by statsmodel's [var](https://www.statsmodels.org/dev/generated/statsmodels.tsa.vector_ar.var_model.VAR.html#statsmodels.tsa.vector_ar.var_model.VAR)
- expected input is `pd.DataFrame`

fkiraly:
this is not a complete user journey design - this should include specification, and usage (e.g., fitting, forecasting) in key use cases.

I also don't think that data should ever be passed to a constructor - look at the forecasting extension template, constructor arguments are reserved to be hyper-parameters and components (if any)

thayeylolu:
Hi @fkiraly , I have updated the User Journey

fkiraly:
just confirming what I said verbally earlier, the user journey should probably look like this:
```params
model = VAR(**params)
model.fit(y=y, fh=fh, X=X)
[...]

PR #1449, opened 2021-09-24
Hotfix tag scitype:y typo

aiwalter:
Fixed typo `y:scitype`, renamed to `scitype:y`

#### Reference Issues/PRs
<!--
related #929 #1083 #1401


fkiraly:
@aiwalter, ouch, this is very painful - no one noticed this?

Why were the tests not failing with the wrong tag name?

fkiraly:
This makes me think: should `get_tag` throw an error if it tries to retrieve a non-existing tag, instead of `None`?

aiwalter:
I think it can still return `None`, but only if the tag exists at all. If I give `myWrongTag` as a tag, it should through an error in my opinion

fkiraly:
Yes, agree with the user expectation and default behaviour, @aiwalter.

When I implemented this, I was mimicking the previous default behaviour which was using `dict.get`, which leads to unexpected results if the tag does not exist (e.g., due to typo), as you say.

Here's my alternative suggestion:
https://github.com/alan-turing-institute/sktime/pull/1450

sktime/utils/_testing/forecasting.py
```diff
@@ -29,6 +29,8 @@ def _get_n_columns(tag):
         n_columns_list = [2]
     elif tag == "both":
         n_columns_list = [1, 2]
+    else:
+        n_columns_list = [1]
```
fkiraly, line 33: I think else it should raise an error, since this is an unexpected input!

fkiraly:
@aiwalter, @mloning, I have identified the (or a) source of the failing tests.

The reason is that `clone` does not clone dynamic tags set properly.
This results in some copies having the `"univariate"` tag set that the estimator inherits by default.

There are multiple potential solutions for this:

* ensure that dynamic tags are set in `_fit` and not in `__init__` - this may be confusing to extenders
* write our own, sktime version of `clone` which adds the line of copying over the `_tags_dynamic` dict
* introduce a new interface point for implementers, a `init_dynamic_tags` or similar, which is called at the start of `fit`.
* introduce a new interface point for intialization of tags *and* components, `init_tags_components`, which initializes dynamic tags and component estimators, and is called at the end of the base `__init__`.

I would be in favour of no.2 and would be against no.1. I would prefer no.4 above no.3.

fkiraly:
@aiwalter, I found another issue after fixing the above one.

Namely, it comes from `evaluate`, which is called in grid search but only lets `pd.Series` through.
However, in the multivariate case it may be called with a `pd.DataFrame`.

So, it means, currently the grid search algorithms are *not* capable of dealing with multivariate data, and the tag should always say `"univariate"`.

aiwalter:
I think its contained also in my PR #1376

fkiraly:
Yes, @aiwalter, but I'd say, let's fix the issues first before we add more stuff.
Proposed merge sequence is #1436, #1454, #1455, #1456, #1449, #1450, also see discussion on slack.

aiwalter:
Thx @fkiraly , will look into it tomorrow

fkiraly:
Hm, looking into #1376, it appears you already discovered two, the issue with the cv splitter and the univariate tag.

Yes, I'd say, let's fix these separately and ensure we have a "clean slate" with tests fixed before we add more content - just to be sure we don't introduce any follow-on bugs.

fkiraly:
For instance, I'm worried that your #1376 doesn't fix the `NaiveForecaster` issue of the missing `update`, but the tests still seem to pass?

aiwalter:
lets get this merged?

fkiraly (approved):
Yes, this should fix all the test issues now.

Looks like we got it fixed!
Great, thanks!

PR #1450, opened 2021-09-24
Error handling for `get_tag`

fkiraly:
This adds the possibility to raise an error when `get_tag` is used.

An error will always be raised if the function is used with parameter defaults and the tag with the specified `tag_name` does not exist, but this can be overridden by setting the new `raise_error` parameter to `False`.

This should hopefully prevent the human error that caused the bug in #1449 for the future.

FYI @aiwalter, @thayeylolu, @mloning 

This PR also contains a second test for `get_tag` to cover the behaviour if `raise_error=True`, the old test covers the behaviour if `raise_errpr=False` with minor changes.

The change has caught the following bugs which are also fixed:

* `"enforce_index_type"` tag was sometimes written `"enforce-index-type"`, this was changed to `"enforce_index_type"` consistently
* transformer base class did not have any default tag values set, which caused framework to look for non-existent tags - added a set of defaults
* `"coerce-X-to-pandas"` was called in some classifiers but not set in the tine series classification base class

### DIFF
--- a/sktime/transformations/base.py
+++ b/sktime/transformations/base.py
@@ -36,6 +36,18 @@ Panel = Union[pd.DataFrame, np.ndarray]  # 3d or nested array
 class BaseTransformer(BaseEstimator):
     """Transformer base class"""
 
+    # default tag values - these typically make the "safest" assumption
+    _tags = {
+        "univariate-only": False,  # can the transformer handle multivariate X?
+        "handles-missing-data": False,  # can estimator handle missing data?
+        "X-y-must-have-same-index": False,  # can estimator handle different X/y index?
+        "enforce_index_type": None,  # index type that needs to be enforced in X/y
+        "fit-in-transform": True,  # is fit empty and can be skipped? Yes = True
+        "transform-returns-same-time-index": False,
+        # does transform return have the same time index as input X
+        "skip-inverse-transform": False,  # is inverse-transform skipped when called?
+    }
+
     def __init__(self):
         super(BaseTransformer, self).__init__()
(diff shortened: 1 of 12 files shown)


fkiraly:
@aiwalter, do you understand why this is *not* failling the tests you fixed in #1449? I'm surprised that it doesn't.

fkiraly:
found it, forgot the `raise` in front of `ValueError`.

sktime/transformations/base.py
```diff
@@ -34,14 +34,25 @@
...
 class BaseTransformer(BaseEstimator):
-    """Transformer base class"""
+    """Transformer base class."""
+
+    # default tag values - these typically make the "safest" assumption
+    _tags = {
+        "univariate-only": False,  # can the transformer handle multivariate X?
+        "handles-missing-data": False,  # can estimator handle missing data?
+        "X-y-must-have-same-index": False,  # can estimator handle different X/y index?
+        "enforce_index_type": None,  # index type that needs to be enforced in X/y
```
aiwalter, line 44: Just a detail, I think `enforce-index-type` would fit better with other tag names instead underscore?

aiwalter (approved):

sktime/transformations/base.py
fkiraly: well, we're changing generally to underscore so we can use tags as kwarg names - that's not possible if a `-` appears in it. There were two inconsistent spellings so I went with the "target" spelling.

aiwalter (approved):

PR #1454, opened 2021-09-25
sktime clean clone method

fkiraly:
Update: This actually does not seem to have been a problem with #1449. We can keep using `sklearn` clone.
Will close once it is confirmed that we can fix #1449 without.

---

This PR adds a `clone` method to `BaseObject`, which extends `sklearn` clone to deal with dynamic tags properly.

This is one potential way to address the issue in #1449, in which behaviour controlling tags inside grid/random search estimators are not cloned properly, and lead to unexpected error messages when multivariate forecasters are cloned.

As a proof-of-concept, the `sklearn` `clone`-s in the tuning module are also replaced by the new `BaseObject.clone`.

An added advantage is our new ability to modify the behaviour of `clone` in the `BaseObject`, in a modular fashion.

### DIFF
--- a/sktime/base/_base.py
+++ b/sktime/base/_base.py
@@ -274,6 +274,22 @@ class BaseObject(_BaseEstimator):
 
         return cls(**params)
 
+    def clone(self):
+        """Create and return deep clone of self.
+
+        Returns
+        -------
+        self_clone : deep clone of self
+            self_clone is identical to sklearn's clone(self), plus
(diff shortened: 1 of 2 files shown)


fkiraly:
(happy to add tests if we decide to go with this solution)

fkiraly:
we don't need this, the problem this was trying to solve isn't actually one (cloning with dynamic tags works fine)

PR #1460, opened 2021-09-26
datatypes.convert to infer indirect (chained) conversions when direct conversions are not implemented

chrisholder:
#### Reference Issues/PRs

Requires PR #1423 

The current convert module only allows direct conversions. For example currently if I have a nested_univ and want to make it into a numpy_flat I am unable to do this. This is because there is no direct conversion between the two implemented. However, there is an indirect conversion that can be done which is nested_univ -> numpy3D -> numpy_flat. This pr has updated the convert module to allow these indirect paths to be found and used. Therefore this pr greatly improves the coverage of conversions.

It does this by using an adjacency matrix which defines which conversion are avaliable and then using dijsktras optimal path to find an indirect path.

In addition using this methods the weights (or length of the paths) can be adjusted to only allow lossy or lossless conversions. This is taken into account and when the user wants a lossless conversion to be done paths that would lead to loss are not considered. However, when the user is fine with loss then all paths are considered.

Requires PR #1423 as this contains the enums needed to find paths.

##### For all contributions

##### For new estimators


fkiraly:
Really nice, thanks a lot! Much needed.

Before discuss this though, let's focus on #1423 and get this merged - that would require hammering out the last areas of design questions and making potential changes there.

So I'd recommend to focus on #1423 design write-up and a round (perhaps targeted meetings) with @MatthewMiddlehurst (because of the link to the proposed typing in #974, #973).

fkiraly:
@chrisholder, could we keep these open? I might be using this someday!

PR #1463, opened 2021-09-27
Added STL transformation

eyalshafran:
I added an STL transformation. The STL transformation can either be used as a standalone transformation to remove seasonality or it can be used as part of a forecasting pipeline. 

For example:
import numpy as np

y = load_airline()
transformer = STL(sp=12)
y_hat = transformer.fit_transform(y)

                            ("deseasonalize", STL(sp=12, robust=True, seasonal=13, seasonal_deg=0)),
                            ("forecast", PolynomialTrendForecaster())])
forecaster.fit(y)
y_pred = forecaster.predict(fh=np.arange(1,13))

### DIFF
--- a/sktime/transformations/series/detrend/_deseasonalize.py
+++ b/sktime/transformations/series/detrend/_deseasonalize.py
@@ -3,21 +3,27 @@
 # copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
 """Implements transformations to deseasonalize a timeseries."""
 
-__author__ = ["mloning"]
+__author__ = ["mloning", "eyalshafran"]
 __all__ = [
     "Deseasonalizer",
     "ConditionalDeseasonalizer",
+    "STL"
 ]
 
 import numpy as np
+import pandas as pd
 from statsmodels.tsa.seasonal import seasonal_decompose
+from statsmodels.tsa._stl import STL as _STL
 
 from sktime.transformations.base import _SeriesToSeriesTransformer
 from sktime.utils.datetime import _get_duration
 from sktime.utils.datetime import _get_freq
+from sktime.utils.datetime import _coerce_duration_to_int
 from sktime.utils.seasonality import autocorrelation_seasonality_test
 from sktime.utils.validation.forecasting import check_sp
 from sktime.utils.validation.series import check_series
+from sktime.forecasting.base._fh import _check_cutoff
+from sktime.forecasting.base._fh import  _coerce_to_period
 
 
 class Deseasonalizer(_SeriesToSeriesTransformer):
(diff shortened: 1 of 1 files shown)


eyalshafran:
I fixed a couple of bugs and the style issues. Let me know if there is anything else I need to do.

sktime/transformations/series/detrend/_deseasonalize.py
```diff
@@ -316,3 +322,223 @@ def fit(self, Z, X=None):
...
+        return y - self.seasonal
+
+    def _inverse_transform(self, y):
+        seasonal_index = self._align_seasonal_index(y)
+        return y + self.seasonal[seasonal_index]
+
+    def transform(self, Z, X=None):
+        """Transform data.
+        Returns a transformed version of y.
+        Parameters
```
aiwalter, line 503:
```suggestion

        Parameters
```

sktime/transformations/series/detrend/_deseasonalize.py
```diff
@@ -316,3 +322,223 @@ def fit(self, Z, X=None):
...
+        ----------
+        y : pd.Series
+        X : pd.DataFrame
+        Returns
```
aiwalter, line 507:
```suggestion

        Returns
```

sktime/transformations/series/detrend/_deseasonalize.py
```diff
@@ -316,3 +322,223 @@ def fit(self, Z, X=None):
...
+    >>> forecaster.fit(y)
+    >>> y_pred = forecaster.predict(fh=np.arange(1,13))
+    """
+
+    _tags = {"transform-returns-same-time-index": True, "univariate-only": True}
+
+    def __init__(self, sp=1, seasonal=7, trend=None, low_pass=None,
+                 seasonal_deg=1, trend_deg=1, low_pass_deg=1,
+                 robust=False, seasonal_jump=1, trend_jump=1, low_pass_jump=1):
+        self.sp = check_sp(sp)
```
aiwalter, line 397: I htink its best to do that check in `fit()`

sktime/transformations/series/detrend/_deseasonalize.py
```diff
@@ -316,3 +322,223 @@ def fit(self, Z, X=None):
...
+        Returns a transformed version of y.
+        Parameters
+        ----------
+        y : pd.Series
+        X : pd.DataFrame
+        Returns
+        -------
+        yt : pd.Series
+            Transformed time series.
+        """
```
aiwalter, line 527:
```suggestion
        """Inverse transform data.

        Returns a transformed version of y.

        Parameters
        ----------
        y : pd.Series
        X : pd.DataFrame

        Returns
        -------
        yt : pd.Series
            Transformed time series.
        """
```

sktime/transformations/series/detrend/_deseasonalize.py
```diff
@@ -316,3 +322,223 @@ def fit(self, Z, X=None):
...
+        """Update fitted parameters.
+        Parameters
+        ----------
+        y : pd.Series
+        X : pd.DataFrame
+        update_params : bool, default=False
+        Returns
+        -------
+        self : an instance of self
+        """
```
aiwalter, line 542:
```suggestion
        """Update fitted parameters.

        Parameters
        ----------
        y : pd.Series
        X : pd.DataFrame
        update_params : bool, default=False

        Returns
        -------
        self : an instance of self
        """
```

sktime/transformations/series/detrend/_deseasonalize.py
```diff
@@ -316,3 +322,223 @@ def fit(self, Z, X=None):
...
+
+    def fit(self, Z, X=None):
+        """Fit to data.
+        Parameters
+        ----------
+        Z : pd.Series
+        X : pd.DataFrame
+        Returns
+        -------
+        self : an instance of self
```
aiwalter, line 464:
```suggestion
    def fit(self, Z, X=None):
        """Fit to data.

        Parameters
        ----------
        Z : pd.Series
        X : pd.DataFrame

        Returns
        -------
        self : an instance of self
```

sktime/transformations/series/detrend/_deseasonalize.py
```diff
@@ -316,3 +322,223 @@ def fit(self, Z, X=None):
...
+    >>> import numpy as np
+    >>> y = load_airline()
+    >>> transformer = STL(sp=12)
+    >>> y_hat = transformer.fit_transform(y)
+    >>> forecaster = TransformedTargetForecaster([
+    >>>     ("deseasonalize", STL(sp=12, seasonal=13, seasonal_deg=0)),
+    >>>     ("forecast", PolynomialTrendForecaster())]
+    >>> )
+    >>> forecaster.fit(y)
+    >>> y_pred = forecaster.predict(fh=np.arange(1,13))
```
aiwalter, line 389:
```suggestion
    >>> y_pred = forecaster.predict(fh=[1,2,3])
```

sktime/transformations/series/detrend/_deseasonalize.py
```diff
@@ -316,3 +322,223 @@ def fit(self, Z, X=None):
...
+    ----------
+    seasonal : array of length sp
+        Seasonal components computed in seasonal decomposition.
+
+    Examples
+    --------
+    >>> from sktime.datasets import load_airline
+    >>> from sktime.forecasting.compose import TransformedTargetForecaster
+    >>> from sktime.forecasting.trend import PolynomialTrendForecaster
+    >>> import numpy as np
```
aiwalter, line 380:
```suggestion
```

aiwalter (commented):
@eyalshafran left a few minor comments

aiwalter:
@eyalshafran pls make sure, that you have `pre-commit` installed, how to do this you will see on the website

eyalshafran:
> related #782

I think this is a better solution than the previous one because it separates the STL deseasonalizing from the forecasting.  So you can use any sktime forecasting method after the STL transformation. When using STLForecasting from statsmodels you are forced to use a forecasting algorithm that works with statsmodels.

sktime/transformations/series/detrend/_deseasonalize.py
eyalshafran: I just followed the Deseasonalizer code for this one. I can go either way but I think it makes sense to check the seasonality parameter when initializing the algorithm and not wait for the fit. I'm happy to move it to fit.

RNKuhns:
> I think this is a better solution than the previous one because it separates the STL deseasonalizing from the forecasting. So you can use any sktime forecasting method after the STL transformation. When using STLForecasting from statsmodels you are forced to use a forecasting algorithm that works with statsmodels.

The implementation of a `STLForecaster` should happen in a separate PR, but I think that we still probably want a sktime `STLForecaster` class. Under the hood it should probably use our STLTransformer (that interfaces Statsmodels stl decomposition functionality) to go from a series to a multivariate series (Trend, Seasonality, Residual). Then we should provide the ability to specify the forecaster for each of these components independently (I believe R's `Fable` lets you do this), then we should use the `ColumnEnsembler` to add these back together.

Interface could be something like this:

```python
stlf = STLForecaster(
    ..., 
    trend_forecaster=SomeSKTIMEForecaster(),
    seasonal_forecaster=SomeOtherSKTIMEForecaster(), 
    residual_forecaster=YetAnotherSKTIMEForecaster(),
    ...
)
```
Note that whenever a user doesn't specify the `trend_forecaster` it should probably default to either a `NaiveForecaster` or `TrendForecaster`, while the `NaiveForecaster` is probably a reasonable default if a user doesn't specify the seasonal_forecaster (assume seasonality is unchanged).

eyalshafran:
@fkiraly @RNKuhns 
I think the name I gave to transformation is confusing and maybe STL should be reserved for STL decomposition. Maybe a different name like STLF can be used instead. 

Below are a few resources that explain STL:
STLForecaster from statsmodels:
https://www.statsmodels.org/stable/generated/statsmodels.tsa.forecasting.stl.STLForecast.html
“ Model-based forecasting using STL to remove seasonality

Forecasts are produced by first subtracting the seasonality estimated using STL, then forecasting the deseasonalized data using a time-series model, for example, ARIMA.”

STLF from forecast in R:
https://pkg.robjhyndman.com/forecast/reference/forecast.stl.html
“ Forecasts of STL objects are obtained by applying a non-seasonal forecasting method to the seasonally adjusted data and re-seasonalizing using the last year of the seasonal component.”

Explanation in Hyndman’s book:
https://otexts.com/fpp2/forecasting-decomposition.html
“ To forecast a decomposed time series, we forecast the seasonal component,  
 , and the seasonally adjusted component  
 , separately. It is usually assumed that the seasonal component is unchanging, or changing extremely slowly, so it is forecast by simply taking the last year of the estimated component. In other words, a seasonal naïve method is used for the seasonal component.

To forecast the seasonally adjusted component, any non-seasonal forecasting method may be used. For example, a random walk with drift model, or Holt’s method (discussed in the next chapter), or a non-seasonal ARIMA model (discussed in Chapter 8), may be used.”

The transformation I wrote can be used as follows to be an STLForecaster:
`stlf= TransformedTargetForecaster([
                            ("deseasonalize", STL(sp=12, robust=True, seasonal=13, seasonal_deg=0)),
                            ("forecast", `PolynomialTrendForecaster())])`

 The advantages of writing the STLForecaster like this is that it is easy to chain more transformations before the STL (for example a log transformation) or even chain multiple STL transformation with different sp in order to remove multiple seasonalities

fkiraly:
@eyalshafran, thanks for clarifying.

Yes, what I also thought:
* there are two concepts here, STL transform (your `STL`, perhaps better called `STLTransformer` or similar?) and STL forecaster (`STLForecaster`?).
* the STL transform would deaseasonlize/detrend the series, and possess an inverse transform that adds those components back, then can
* the STL forecaster applies the STL transform by prepending in a forecaster pipeline, to any forecaster. So it has as hyper-parameters exactly those of the STL transform, plus a forecaster.

So your *intention* looks very sensible, if I understand correctly it's focused on the `STLTransformer`.

@RNKuhns, perhaps you can explain: I don't fully understand what precisely the *request* is in your GitHub change request.
Are you trying to say we don't want an `STLTransformer` but only an `STLForecaster`?
Or, did you think @eyalshafran's *intention* is to implement the `STLForecaster` and you think the implementation (which is a transformer) does not match the intention?

In the first case, I personally don't see why we wouldn't like to have both as a shorthand.
In the second case, I think @eyalshafran wanted to implement the transformer, not the forecaster?

Issue #1478, opened 2021-10-02
[BUG] dependency bug in tsfresh causing tests to fail


fkiraly:
Suggested fix:

- [x] set an upper bound on the last `statsmodels` version that still worked in `setup.py`, see #1478
- [x] same for the binder/build requirements
- [ ] wait for `tsfresh` to resolve the issue
- [ ] remove the version bound on `statsmodels` and test whether things run again

---

The CI tests are breaking, due to a `tsfresh` import now fails, which is required for `TSfreshClassifier`.

Error as below.

Looks like this is coming from an interface change in `statsmodels`, as a `tsfresh` dependency?
Where `tsfresh` is calling `stattools.acf` which no longer has an `unbiased` argument in 0.13.0, but used to have?

Not sure how to best fix this, it seems it needs to be fixed in `tsfresh`? Can someone check whether the diagnosis is correct?

```
070     >>> from sktime.classification.feature_based import TSFreshClassifier
071     >>> from sktime.datasets import load_italy_power_demand
072     >>> X_train, y_train = load_italy_power_demand(split="train", return_X_y=True)
073     >>> X_test, y_test = load_italy_power_demand(split="test", return_X_y=True)
074     >>> clf = TSFreshClassifier()
075     >>> clf.fit(X_train, y_train)
UNEXPECTED EXCEPTION: TypeError("acf() got an unexpected keyword argument 'unbiased'")
multiprocessing.pool.RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/opt/_internal/cpython-3.7.12/lib/python3.7/multiprocessing/pool.py", line 121, in worker
    result = (True, func(*args, **kwds))
  File "/opt/_internal/cpython-3.7.12/lib/python3.7/site-packages/tsfresh/utilities/distribution.py", line 43, in _function_with_partly_reduce
    results = list(itertools.chain.from_iterable(results))
  File "/opt/_internal/cpython-3.7.12/lib/python3.7/site-packages/tsfresh/utilities/distribution.py", line 42, in <genexpr>
    results = (map_function(chunk, **kwargs) for chunk in chunk_list)
  File "/opt/_internal/cpython-3.7.12/lib/python3.7/site-packages/tsfresh/feature_extraction/extraction.py", line 337, in _do_extraction_on_chunk
    return list(_f())
  File "/opt/_internal/cpython-3.7.12/lib/python3.7/site-packages/tsfresh/feature_extraction/extraction.py", line 317, in _f
[...]

PR #1479, opened 2021-10-02
1478 statsmodels max_requirements

fkiraly:
This is a temporary fix for #1478, what looks like a bug introduced with `statsmodels` 0.13.0 through an interface hange that has not yet propagated to `tsfresh`.

Besides setting a max `statsmodels` version of 0.12.1 (the latest version before 0.13.0), it adds a `MAX_REQUIREMENTS` list to the `setup.py`.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -65,7 +65,7 @@ install:
 
   # Install requirements from inside conda environment
   - cmd: activate testenv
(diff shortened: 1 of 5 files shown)


aiwalter (approved):

fkiraly:
@aiwalter, have you checked whether this prevents the error?

fkiraly:
this doesn't fix it, see #1478.

aiwalter:
we also need to change `hard_dependencies.txt`

aiwalter:
and appveyor yaml

fkiraly:
why is there not a *single source of truth*

aiwalter (approved):

fkiraly:
seems like that solve it ... for now

PR #1481, opened 2021-10-02
1478 remove tsfresh from tests

fkiraly:
Another attempt at a temporary fix for the `tsfresh` issue - simply removing from tests.

Minor note, we should be using `pytest.xfail` for this and not custom test logic, @mloning ... but I've just added it in the place that we are currently using.

### DIFF
--- a/sktime/tests/_config.py
+++ b/sktime/tests/_config.py
@@ -110,6 +110,9 @@ EXCLUDE_ESTIMATORS = [
     "ProximityForest",
     "ProximityStump",
     "ProximityTree",
+    # temporary due to tsfresh related bug?
+    "TSfreshClassifier",
+    "TSFreshFeatureExtractor",
 ]
 
 


aiwalter:
I think this is not enough as the issue comes from an example in a notebook:
```
2021-10-02T17:49:33.0088878Z ------------------
2021-10-02T17:49:33.0089321Z classifier = make_pipeline(
2021-10-02T17:49:33.0090133Z     TSFreshFeatureExtractor(show_warnings=False), RandomForestClassifier()
2021-10-02T17:49:33.0090847Z )
2021-10-02T17:49:33.0091261Z classifier.fit(X_train, y_train)
2021-10-02T17:49:33.0091795Z classifier.score(X_test, y_test)
2021-10-02T17:49:33.0092528Z ------------------
2021-10-02T17:49:33.0092790Z 
```

fkiraly:
yes, solved now via #1479, temporarily.

PR #1485, opened 2021-10-04
Add simple reduction strategy from to forecasting to regression

### LINKED ISSUE #1280: [ENH] forecasting: "naive supervised prediction" reducer
We should have a shorthand for the reducer where `y` is simply predicted from `X`, possibly including the index.

This is generally not a good forecaster in isolation, but useful as a component for instance in #1279, in anomaly detection or other annotation tasks.

It's also a special case of the direct regression strategy with exogeneous variables, where there are zero lags.

Question for @ltsaprounis, @angelpone, @mloning: in the composition of this and #1279, could we use the "reconciler" somehow?

AreloTanoh:
#### Reference Issues/PRs
Resolves #1280 

Entering new forecasting algorithm to reduce.py

Currently work in porgress

##### For all contributions

##### For new estimators


AreloTanoh:
@mloning please leave any comments you may have once you approve the changes

mloning (commented):
Hi @AreloTanoh, yes this is the right first step. I'll check back later when you've made more changes to the algorithm. 

Also make sure you set up the pre-commit checks as described in the contributing guide: https://www.sktime.org/en/latest/get_involved/contributing.html

sktime/forecasting/compose/_reduce.py
```diff
@@ -890,3 +939,28 @@ def _get_forecaster(scitype, strategy):
         },
     }
     return registry[scitype][strategy]
+
+
+class MyForecastingAlgorithm(BaseForecaster):
```
mloning, line 944: This algorithm shouldn't be added to sktime, so we need to remove it from here

sktime/forecasting/compose/_reduce.py
```diff
@@ -442,10 +443,39 @@ def _predict_last_window(
...
+    def _fit(self, y, X=None, fh=None):
+        X = X.to_numpy()
+        y = y.to_numpy().reshape(-1, 1)
+        self.regressor_ = clone(self.regressor)
+        self.regressor_.fit(X, y)
+
+        return self
+
+    def _predict(self, fh, X=None, return_pred_int=False, alpha=None):
+        assert not return_pred_int
```
mloning, line 466: Instead of having an assert statement here, we probably want to raise a more informative error message:
```python
if return_pred_int:
    raise NotImplementedError(f"{self.__class__.__name__} does not support prediction intervals.")
```

sktime/forecasting/compose/_reduce.py
```diff
@@ -640,6 +670,24 @@ class DirRecTabularRegressionForecaster(_DirRecReducer):
     _estimator_scitype = "tabular-regressor"
 
 
+class SimpleTabularRegressionForecaster(_SimpleReducer):
+    """Recursive reduction from forecasting to tabular regression.
```
mloning, line 674: the docstring needs to be updated

sktime/forecasting/compose/_reduce.py
```diff
@@ -442,10 +443,39 @@ def _predict_last_window(
         return y_pred[fh_idx]
 
 
+class _SimpleReducer(_Reducer):
+    strategy = "simple"
+    _tags = {
```
mloning, line 448: the tags need to be updated to (see the extension template)

mloning (commented):
Hi @AreloTanoh, just a few comments, happy to discuss them in our next meeting!

fkiraly:
abandoned

PR #1489, opened 2021-10-05
Added prediction intervals

eyalshafran:
I added the option to output prediction intervals for AutoETS. Prediction intervals are based on statsmodels implementation. 

No

### DIFF
--- a/sktime/forecasting/ets.py
+++ b/sktime/forecasting/ets.py
@@ -11,7 +11,7 @@ from itertools import product
 import numpy as np
 from joblib import Parallel, delayed
 from statsmodels.tsa.exponential_smoothing.ets import ETSModel as _ETSModel
-
+from sktime.forecasting.base._base import DEFAULT_ALPHA
 from sktime.forecasting.base.adapters import _StatsModelsAdapter
 
 
@@ -167,6 +167,12 @@ class AutoETS(_StatsModelsAdapter):
     """
 
     _fitted_param_names = ("aic", "aicc", "bic", "hqic")
+    _tags = {
+        "ignores-exogeneous-X": True,
+        "capability:pred_int": True,
+        "requires-fh-in-fit": False,
+        "handles-missing-data": False,
+    }
 
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


eyalshafran:
It seems to fail a few tests. I'll make some changes to verify it passes the tests.

aiwalter:
@eyalshafran what do you think about adding this to `_StatsModelsAdapter` directly? Then we could use it in all children I think?

fkiraly:
If it's not too much of a hassle, would you mind implementing a function `predict_quantiles` and call that within your implementation? It should take a vector of `alpha` and return a data frame with predictive quantiles, column indices being the `alpha`.

This is because we are currently refactoring the prediction interval interface, and this will be the place where the logic would sit in the end state, see #1421 - FYI @SveaMeyer13.

fkiraly:
>@fkiraly I can add predict_quantiles.

Thanks! Let's do it for AutoETS only then - make sure that the PR name matches the content then.

>So I currently do not see an easy way of implementing the prediction intervals in the `_StatsModelsAdapter`. What do you think?

Hm, ok, it looks indeed that the return objects have inconsistent interfaces, so I'd say let's start with AutoETS.

But not all is lost: do the differently named functions perhaps have the same input and output specification (e.g., return type format)?

Then we could implement a generic `predict_quantiles` inside `_StatsModelsAdapter`, which references an abstract `pred_int_function_name` attribute (string) which is added by the respective subclass.
(if a subclass does not implement it, just keep the `capability:pred_int` tag at a `False` value for the subclass, and the user query will be captured before it reaches the implementation)

eyalshafran:
I still need to add the predict_quantiles. I just wanted to fix the test failures first.

I do need clarification on the alpha vector for quantiles. When using alpha for prediction intervals that typically means prediction interval coverage. So an alpha of 0.05 means a 95% prediction interval. 
From statsmodels:
"alpha: The significance level for the prediction interval. Default is 0.05, that is, a 95% prediction interval"
So alpha=0.05 means 2.5 - 97.5 quantiles. Alpha = 0.5 means 25%-75% quantiles.
It seems to me that the alpha in prediction intervals and predict quantiles do not mean the same thing.

eyalshafran:
@SveaMeyer13 - I didn't make sure that it can accept a list of alphas. Thank you for pointing it out.

PR #1490, opened 2021-10-05
ShapeletTransformClassifier rework and documentation for HC2

MatthewMiddlehurst:
todo


TonyBagnall:
I'll look in detail next week, but all I'll say at this point is wow, fantastic job. Honestly, I could not be more impressed in the way you have as sorted this out. Great job, you have done it so much better than I could have!

PR #1493, opened 2021-10-07
[BUG] #1469: stripping names of index X and y 

### LINKED ISSUE #1469: [BUG] ValueError when fitting Prophet and X.index.name=='ds'
**Describe the bug**
The following error is raised when trying to fit Prophet using X where the name of X's DateTimeIndex is named "ds":
ValueError: 'ds' is both an index level and a column label, which is ambiguous.

It happens in the _merge_X method in the _ProphetAdapter, line 193, because after running line 191, X contains a column and an index called "ds".

A suggested solution would be to add this oneliner before line 189 of .forecasting.base.adapters._fbprophet.py:
```python
X.index.name = "index"
```
Note: This assumes that X never has a MultiIndex. Not sure if that ever happens though..

This way, the user is still allowed to call X's index "ds". Since "ds" is the standard name that Prophet users use for their DateTime column, this is the most intuitive and least restrictive solution.

**To Reproduce**
```python
import pandas as pd
from sktime.forecasting.model_selection import ForecastingGridSearchCV

y_train = load_airline().to_timestamp(freq='M')

# Commenting out the following line makes everything run smoothly
y_train.index.name = 'ds'

X_train = pd.DataFrame({'x1':y_train+200, 'x2':y_train+100})

cv = ExpandingWindowSplitter(
    step_length=12, fh=[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12], initial_window=72
)

param_grid = {  
    'changepoint_prior_scale': [0.001, 0.01, 0.05],
    'seasonality_prior_scale': [0.01, 0.1],
    'growth': ['linear'] 
}
forecaster = Prophet(yearly_seasonality=False, weekly_seasonality=False, daily_seasonality=False)
forecaster.fit(y_train, X_train)
```

**Expected behavior**
No error

**Additional context**

**Versions**
<details>

System:
    python: 3.9.7 | packaged by conda-forge | (default, Sep 23 2021, 07:24:41) [MSC v.1916 64 bit (AMD64)]
executable: C:\Users\............\env\python.exe
   machine: Windows-10-10.0.19043-SP0

Python dependencies:
          pip: 21.2.4
   setuptools: 58.1.0
      sklearn: 1.0
       sktime: 0.7.0
  statsmodels: 0.12.2
        numpy: 1.21.2
        scipy: 1.7.1
       Cython: 0.29.24
       pandas: 1.3.3
   matplotlib: 3.4.3
       joblib: 1.0.1
        numba: 0.53.1
     pmdarima: 1.8.3
      tsfresh: None
</details>

boukepostma:
#### Reference Issues/PRs
<!--
Fixes #1469 

Removes names of the index of X and y when those are not a np.array

No

##### For all contributions

### DIFF
--- a/sktime/forecasting/tests/test_all_forecasters.py
+++ b/sktime/forecasting/tests/test_all_forecasters.py
@@ -26,26 +26,30 @@ import pandas as pd
 import pytest
 
 from sktime.exceptions import NotFittedError
-from sktime.forecasting.model_selection import SlidingWindowSplitter
-from sktime.forecasting.model_selection import temporal_train_test_split
-from sktime.forecasting.tests._config import TEST_ALPHAS
-from sktime.forecasting.tests._config import TEST_FHS
-from sktime.forecasting.tests._config import TEST_OOS_FHS
-from sktime.forecasting.tests._config import TEST_STEP_LENGTHS
-from sktime.forecasting.tests._config import TEST_WINDOW_LENGTHS
-from sktime.forecasting.tests._config import VALID_INDEX_FH_COMBINATIONS
-from sktime.performance_metrics.forecasting import (
-    mean_absolute_percentage_error,
+from sktime.forecasting.model_selection import (
+    SlidingWindowSplitter,
+    temporal_train_test_split,
 )
+from sktime.forecasting.tests._config import (
+    TEST_ALPHAS,
+    TEST_FHS,
+    TEST_OOS_FHS,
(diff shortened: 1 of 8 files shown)


fkiraly:
the precommit hooks would like you to order the imports alphabetically in the file you are changing

boukepostma:
Thanks for the feedback! I pushed the change in dependencies

PR #1465, opened 2021-09-29
[MNT] Added pre-commit hook to sort imports

aiwalter:

PR #1494, opened 2021-10-07
Sort all imports

aiwalter:
#### Reference Issues/PRs
#1465 

Sort all imports, to avoid issues with already existing PRs.
I did:
```
pre-commit run isort --all-files
git add .
git commit -m "Sorted imports in all files" --no-verify
git push origin HEAD
```


aiwalter:
@mloning yes, but seems the sorting is an issue for already open PRs as locally it runs just on the newly changed files but on pipeline it runs on all changed files of a PR. So @fkiraly asked for a solution for it. In my opinion we can also go with the workaround of just doing any minor change locally in all changed files (e.g. randomly cahnge order of some imports) in order to trigger the sorting to all changed files. Then we dont need this PR. For all new PRs, there is no issue with sorting.

aiwalter:
@mloning yes agreed. Thx! I think that is what we were missing :) Will close this PR.

fkiraly:
The "fix" is not impacting new contributors! It's putting burden on reviewers to point this out, and on contributors to complicate their set-up.

I would be happy with exactly two solutions:

* apply the ordering to all existing files, so it won't pop up in PR of new contributors who make small changes
* test failure should not be triggered by unchanged lines, only by new lines

I feel relatively strongly about this.

PR #1498, opened 2021-10-08
[ENH] Adding FunctionTransformer

boukepostma:
Adds an sktime-compatible adjustment of sklearn's FunctionTransformer. This class allows the user to easily create custom transformers that are different from the ones already included in sktime (e.g. SqrtTransformer, ExponentTransformer).

No

* Docstring: this is a slightly adjusted version of the docstring in sklearn's FunctionTransformer. Should I give credits to or cite the original authors and if so, how? 
* `def _check_inverse_transform()`: this works but there may be more reliable ways to check whether `func` and `inverse_func` are the inverse. Note: comparing with `==` does not work due to numeric errors.

#### Any other comments?
Looking forward to receiving your feedback!

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/transformations/series/func_transform.py
+++ b/sktime/transformations/series/func_transform.py
@@ -71,6 +71,11 @@ class FunctionTransformer(_SeriesToSeriesTransformer):
            [1.09861229, 1.38629436]])
     """
 
+    _tags = {
+        "handles-missing-data": True,
+        "fit-in-transform": False,
+    }
+
     def __init__(
         self,
         func=None,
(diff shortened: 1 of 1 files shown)


fkiraly:
Huh, we didn't already have this?
@aiwalter, @mloning, did we not already have this somewhere (under a different name)?

fkiraly:
:Not sure which error you are referring to wrt docstrings. Should I remove some spaces after the 0. in the array?

Go here to see CI test failures for the various builds:

image

aiwalter:
@boukepostma would you mind to add `FunctionTransformer` also to docs? You can search for other transformers in the `.rst` files how to do it.

fkiraly:
@RNKuhns, I think this very much makes sense as part of the public API, a general-purpose transformer creator.

PR #1502, opened 2021-10-08
[ENH] GroupedForecaster

boukepostma:
#### What does this implement/fix? Explain your changes.
Let's say we want to forecast daily worldwide sales for company X. We have daily sales data per country where X operates: our dataset contains three columns: `date`, `sales`, and `country`. Now we could consider three modelling options:
1. Aggregate all daily sales and estimate with a single model
2. Cluster countries with similar patterns or characteristics, aggregate per cluster and model per cluster
3. Model all countries separately

Currently, `sktime` only supports option 1. For 2 and 3, it is still necessary to write a custom framework to try different aggregations of X and y and compare performance of strategy 1,2, and 3. In this draft PR I'd like to propose a meta-estimator that removes the need for such custom frameworks.

This PR basically imitates the `GroupedEstimator` meta-estimator from the `scikit-lego` such that it is compatible with `sktime`. For further understanding of the approach I'd recommend to take one minute to read through `scikit-lego`'s documentation:
https://scikit-lego.readthedocs.io/en/latest/meta.html#Grouped-Prediction

#### Any other comments?
Very interested in your feedback!

#### PR checklist
- [ ] Finish `GroupedForecaster`
- [ ] Writing tests for `GroupedForecaster`
- [ ] Adjustment of `forecasting.model_evaluation.evaluate` to support aggregation of y (and X)
- [ ] Adjustment of check_y_X to support duplicate datetimeindex when `isinstance(forecaster, GroupedForecaster)`

##### For all contributions
- [ ] I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/main/.all-contributorsrc).
- [ ] Optionally, I've updated sktime's [CODEOWNERS](https://github.com/alan-turing-institute/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
- [ ] I've added unit tests and made sure they pass locally.

##### For new estimators
- [ ] I've added the estimator to the online documentation.
- [ ] I've updated the existing example notebooks or provided a new one to showcase how my estimator works.

PR #1506, opened 2021-10-12
Add badge to track pypi downloads to README

RNKuhns:

This adds the [PePy](https://pepy.tech/) badge to track pypi downloads. 

No.

- Are we fine indicating our downloads?
- Do we want to show monthly (as opposed cumulative or weekly downloads)? Alternatively we could include each.
-  Are we okay with the coloring or would we prefer it to have the orange color instead of blue (it is semi-customizable)?

#### Any other comments?

When I was looking at the package referenced in #1502, I noticed that it included this badge in that package's README. These statistics are also useful in terms of the discussion about tracking metrics we had as part of the recent roadmap meeting.

##### For all contributions

##### For new estimators


fkiraly:
ca 800k, that's only pypi and not conda

mloning:
> Also, the concept of "inflated" only makes sense if you have an "actual target" in mind. Only people who are notebook users? Why would that be a more sensible target than one that also includes production pipelines?

I don't think we care about downloads per se, the target is users or usage. When I look at a new package, the number of downloads is not something I often look at, because I don't think it doesn't say much about how widely it is used (see [blog post](https://blog.dask.org/2020/01/14/estimating-users)). 

The comparison is also only useful if you have a point of reference. We have 116K monthly PyPI downloads. I have no idea what that means. How many monthly downloads does statsmodels have? Or scikit-learn? I have no idea, so the number doesn't tell me much. 

I don't have a strong objection against including it in the README, I just think for most users and developers the number isn't very useful. Once we track it over time and compare it against other packages in something like a roadmap progress report, it may become more useful.

PR #1509, opened 2021-10-12
Add statistical_tests module

### LINKED ISSUE #1175: Add Module for Statistical Tests
Statistical tests have common use-cases in timeseries analysis, including inspect the properties of a timeseries (e.g. stationarity testing, checking normality) to guide modeling decisions and also to evaluate model output (including evaluating quality of forecasts). 

Adding a interface for statistical tests will allow Sktime to add relevant functionality in this area. In addition to the tests themselves, this will help enable conditional transformers (think differencing a series if it is non-stationary or using a BoxCox transform based on results of normality or conditional variance test) and post-hoc forecast evaluation/benchmarking (Diebold-Mariano and other tests of one set of forecasts against another).

A interface and module for statistical tests in Sktime. The module's `base` would include a class that will be the basis for all tests. 

My thoughts on the interface are generally:

1. Tests will be estimators (they need to be fitted)
2. Instead of `transform` or `predict` they should have a `report` method that returns test results
    - Reported results should be standardized regardless of test and I think they should be p-value, test statistic, and whether null was rejected (all tests would have hyper-parameter `report_detail` that defaults to True and reports all three items. But if it is set to False only whether the null was rejected would be reported)
3. Plant to follow logic of Forecaster refactor and keep the code in non-public methods standardized across sub-classes

Proposed `BaseStatisticalTest` is presented below. 

```python
class BaseStatisticalTest(BaseEstimator):
   
    def __init__(
        test_hyper_parameteres..., 
        hypothesis="two-sided",
        report_detail=True
    ):
        ...
        self.p_value = None
        self.test_statisic = None
        self.reject_null = None
        
    def _fit(Y, X=None):
        """ Logic to fit each test."""
        
    def fit(Y, X=None):
        """Would remain same in each Test's class."""
        
        return self._fit(Y, X=X)
        
    def _report()
        """Logic to return the information to from report.
        
        Returns
        --------
        Plan is to return just boolean reject_null if hyper-parameter `report_detail=False`
        Otherwise will report the following:
        p_value : float or None
            P-value associated with statistical test. If no p-value is 
            available for a test then will return None.
        test_statistic : float or None
            Test statistic from the statistical test. If no test 
            statistic is available for a test then will return None.
        reject_null : bool
            Whether the Test's Null Hypothesis was rejected.
        """
        ...
        if report_detail:
            return p_value, test_statistic, reject_null
        else:
            return reject_null
            
    def report():
        """Would be same for every test."""
        self.check_is_fitted()
        ...
        return self._report()

    def fit_report(Y, X=None):
        """Would be same for every test."""
        return self.fit(Y, X=X).report()
        
    def print_results():
        """Pretty printing the test hyperparameters, timeseries being tested and results."""
        ...
        return None
        
    def results_to_pandas():
        """Output results to a pd.DataFrame.
        
        Useful when you want to apply a test to many series and capture 
        the results.
        
        Returns
        -------
        results_df : pd.DataFrame
            DataFrame containing results in standardized format.
        """
        
        return results_df
        
    def results_to_excel(...):
        """Output results to an excel file.
        
        Useful when you want to apply a test to many series and capture 
        the results and store the results on disk incrementally (potentially
        if applying to many series and you want to ensure results are saved even
        if workflow gets stopped (so you don't have to start at beginning).
        
        Returns
        -------
        None
        """
        self.results_to_pandas().to_excel(...)
        return None
        
    def results_to_csv(...):
        """Output results to an csv file.
        
        Useful when you want to apply a test to many series and capture 
        the results and store the results on disk incrementally (potentially
        if applying to many series and you want to ensure results are saved even
        if workflow gets stopped (so you don't have to start at beginning).
        
        Returns
        -------
        None
        """
        self.results_to_pandas().to_csv(...)
        return None
```

Note that I'm open on design details, particularly the naming conventions (I don't have strong feelings about use of report, results, or something else) and likewise if we want to call this something other than statistical tests that is fine too). 

The main outstanding questions (other then general feedback) involve around the interface for accepting different types of input that works across a range of tests. 

This needs to cover:
1. Univariate diagnostic tests of timeseries "properties" (e.g. normality, stationarity, auto-correlation, etc)
2. Multivariate diagnostic tests (e.g. Granger causality or cointegration)
3. Panel diagnostic tests (panel extension of stationarity tests, etc)
4.  Post-hoc tests of one set of forecasts (whether they be univariate, multivariate or potentially panel) against another ("Y_Other")

Initial thoughts to solve this would be for `fit` to accept either a pd.Series, pd.DataFrame or NumPy array "Y" and optionally accept exogenous data "X" (some tests will use this others won't) and determine how to proceed based on the type of tests.

- For "univariate" diagnostic tests, if a pd.Series is received the test would be done on the series (optionally using X if it is needed). If a pd.DataFrame is received the test would be applied to each series in the DataFrame. 
    - Should we return an error if panel data is received? I'm leaning this way for now but later we could run test for each series for each panel member.
- For a "multivariate" diagnostic test if a pd.Series (or single series DataFrame or NumPy array) then an appropriate message is raised; otherwise tests uses multivariate data from the DataFrame or NumPy array. 
- Panel tests would likewise raise error if appropriate panel data format not received and otherwise run as expected.
- In all cases, we'd have error checking to make sure the X data is of the correct dimension to be used with the Y data

This leaves the last piece, which is how to accept the data for post-hoc tests. Note that these tests often can be applied to univariate data, while an extension allows them to be applied to multivariate data. I'd propose we don't want separate classes based on that distinction. Instead, I propose the following logic:
1. Optionally pass  "Y_Other" in `fit` (kind of like how we handle y_train in performance metrics). If Y_other is received we check its dimension against Y and assume we are doing a post-hoc comparison of Y against "Y_Other"
2. If "Y_Other" is not passed and a pd.Series is received then raise an error (you'd have nothing to compare series against)
3. If "Y_Other" is not passed and Y is a pd.DataFrame then make assumptions about its structure and proceed with test (e.g. if it has 2 columns test column 1 against column 2).

Note that I will edit this comment later to add a list of tests that can be interfaced (primarily from statsmodels) and a set of tests we'd need to write ourselves. 

Plan would be to chunk this out in phases: 
1. Decide on framework and implement BaseStatisticalTest and unit tests
2. Have issue with checklist of good first issues for interfacing tests in Statsmodels (and possibly elsewhere if we can avoid adding un-needed additional dependencies)
3. Create issue with checklist of tests we need to code ourselves (as of now these are mostly post-hoc tests and some boutique extensions of diagnostic tests)

An alternative I've considered is to import and use tests from other packages (Statsmodels) when available. But there are tests not in Statsmodels that we should add (post-hoc forecast evaluation ones in particular). Having a common interface that can be used to adapt Statsmodels tests to our format and also be used for our own statistical tests seems like the way to go to me for uniformity.

RNKuhns:
#### Reference Issues/PRs

Fixes #1175

This adds a module containing class based statistical tests. All statistical tests will inherit from `BaseStatisticalTest` and have a common interface (See #1175 for discussion on interface).

Sub-modules are delineated based on the scientific usage of the test (e.g. diagnostic vs. post-hoc).

No new dependencies have been added.

This remains a work in progress, but any initial comments can focus on the structure of the module, as well as the implementation of `BaseStatisticalTest`. Note that the input type checking and related logic still needs to be added. 

#### Any other comments?

This is a work in progress. I'll be adding additional code to the base class and common unit tests (to verify statistical test interface compatibility).

Once this is completed I'll open up a new issue with a list of diagnostic tests and a list of post-hoc tests to be implemented.

##### For all contributions

##### For new estimators

### DIFF
--- a/docs/source/api_reference.rst
+++ b/docs/source/api_reference.rst
@@ -23,6 +23,7 @@ For a scientific manual, see the :ref:`user_guide`.
     api_reference/series_as_features
     api_reference/transformations
     api_reference/performance_metrics
+    api_reference/statistical_tests
     api_reference/datasets
     api_reference/utils
     api_reference/exceptions
--- a/docs/source/api_reference/statistical_tests.rst
+++ b/docs/source/api_reference/statistical_tests.rst
@@ -1,0 +1,26 @@
+
+.. _statistical_test_ref:
+
+Statistical Tests
+=================
+
+The :mod:`sktime.statistical_tests` module contains statistical tests for diagnosing timeseries properties and
+comparing the performance of sktime estimators.
+
+.. automodule:: sktime.statistical_tests
+    :no-members:
+    :no-inherited-members:
+
+Base
+----
+
+.. currentmodule:: sktime.statistical_tests.base
+
+Classes
+~~~~~~~
+
+.. autosummary::
+    :toctree: auto_generated/
+    :template: class.rst
+
+    BaseStatisticalTest
--- a/docs/source/user_guide/statistical_tests.rst
+++ b/docs/source/user_guide/statistical_tests.rst
@@ -1,0 +1,19 @@
+.. _user_guide_performance_metrics:
+
+Statistical Tests
+=================
+
+.. note::
+
+    The user guide is under development. We have created a basic
+    structure and are looking for contributions to develop the user guide
+    further. For more details, please go to issue `#361 <https://github
+    .com/alan-turing-institute/sktime/issues/361>`_ on GitHub.
+
+
+Diagnostic
+----------
+
+
+Post-hoc
+--------
--- a/sktime/registry/_base_classes.py
+++ b/sktime/registry/_base_classes.py
@@ -1,4 +1,6 @@
 # -*- coding: utf-8 -*-
+# !/usr/bin/env python3 -u
+# copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
 """Register of estimator base classes corresponding to sktime scitypes.
 
 This module exports the following:
@@ -50,13 +52,9 @@ TRANSFORMER_MIXIN_LIST - list of string
 
 TRANSFORMER_MIXIN_LOOKUP - dictionary
     keys/entries are 0/1-th entries of TRANSFORMER_MIXIN_REGISTER
-
-
 """
 
 __author__ = ["fkiraly"]
-
-import pandas as pd
 
 from sktime.annotation.base import BaseSeriesAnnotator
 from sktime.classification.base import BaseClassifier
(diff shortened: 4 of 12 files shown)


RNKuhns:
@fkiraly and @mloning can you take a fast look and give me some quick feedback while I work on the unit testing and datatype logic?

fkiraly:
Abandoned and superseded by the parameter estimator module - which follows similar ideas but uses `get_fitted_params` instead of `report`.

Issue #1397, opened 2021-09-09
[ENH] all dashes in tags should be replaced by underscores


fkiraly:
In tags, all dashes should be replaced by underscores.

The reason is that in a vanilla call to `set_tags`, one would write `set_tags(tag_name=tag_value)`, but this will break if `tag_name` has a symbol in it that is not permitted in a python variable name (for instance, a dash).

For each tag, the changes need to be done:
* in `registry/_tags` 
* all estimators where it is set
* any other logic that accesses the tag, including: concrete estimator inits (dynamic tags), base class plumbing, and the tests.

vscode "find in files" allows to highlight all occurrences.

If someone would like to work on this, please make yourself known here and pick one tag first, to keep changes small and avoid duplication.

PuneetDabral:
sir can i do my first contribution here ..can you please give me chance 
(i am new in open source i need all of your support so i can motivate my self)

mloning:
Hi @PuneetDabral, yes, no-one else is working on this as far as I can see so please give it a go! Let us know if you get stuck.

anurag629:
Hello, i am new to open source contribution. I want to contribute. Please guide me .....

fkiraly:
go to `registry/_tags` and pick one with a dash, then say here which one you picked; then make a PR in which these are renamed.

PuneetDabral:
ok i will clone this

PR #1510, opened 2021-10-12
changed dashes in every "tag_name" to underscores

tensorflow-as-tf:
This PR aims to address #1397.

All dashes in tags are replaced by underscores so that a vanilla call to `set_tags` with `tag_name=tag_value` will not break as `tag_name` does not contain any dash symbol.


TonyBagnall:
hi, I know its a small change but I'm worried this may cause conflicts with other ongoing PRs thus causing others grief. Might it be better to do it module by module (forecasting as one PR, then transformations, etc)?

TonyBagnall:
hi, I think the code has gone beyond this, it is probably worth doing this more incrementally when the classification refactor is complete. Can I close this PR?

fkiraly:
@TonyBagnall, not sure - pending fixes in the merge conflicts, might be worth doing this before the release?

fkiraly:
out of sync since 1 y old, closing. Let´s try the renaming separately.

PR #1517, opened 2021-10-13
Classifier base class

TonyBagnall:
Some changes to the base class for classifiers and the tags

1. Remove classification from the tags "X-y-must-have-same-index" and "enforce_index_type". They are not used by any classifiers
2. Add the capability tags to the base class.  think  capabilities should be checked against data in the check_X functionality. 
some not dependent on the data could be move down, but will be defined so may as well have them in the base imo
3. Changes in comments. Note it claims to allow 2D arrays, which is not true, but I left them because it will soon be true. 
4. remove unnecessary argument to get_tag
coerce_to_numpy = self.get_tag("coerce-X-to-numpy")
4. removed the classifier_list
its not complete and probably should not be here, I imagine I put it there. 
5. Remove default _predict behaviour, add default _predict_proba behaviour. The enhancement template states predict *must* be implemented. I think that is correct

note there will be a few more refinements, but this will allow us to blast through the fit/_fit refactor

### DIFF
--- a/sktime/classification/base.py
+++ b/sktime/classification/base.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
 """
-Base class template for time series classifier.
+Abstract base class for time series classifiers.
 
     class name: BaseClassifier
 
@@ -33,7 +33,7 @@ from sktime.utils.validation.panel import check_X, check_X_y
 
 
 class BaseClassifier(BaseEstimator):
-    """Base time series classifier template class.
+    """Abstract base class for time series classifiers.
 
     The base classifier specifies the methods and method signatures that all
     classifiers have to implement.


fkiraly:
quick question, while you're at it:

do you want to add the planned support for the various `Panel` formats?

All it would take is adding one or two `convert` from `datatypes` to the functions.

TonyBagnall:
> quick question, while you're at it:
> 
> do you want to add the planned support for the various `Panel` formats?
> 
> All it would take is adding one or two `convert` from `datatypes` to the functions.

Not on this PR, I want to finish refactor to fit first then look at checking and transforming data, one thing at a time

TonyBagnall:
> * docstrings of `BaseObject.get_tag` are changed from sth that's correct imo to sth that's incorrect? This is a very small change, but a blocker.

yes, fixed, my misunderstanding (see above)

> * I think the default behaviour in `_predict` should remain? Perhaps with a check whether an infinite look occurs (in the public part of the function) that raises an error.

I dont. Your extension guidelines say that _predict *must* be implemented. I see no value in this default behaviour, it will allow the user to run code which should raise an error, if, say, they implement _predicted instead of _predict

> * Happy with the changes to the `BaseClassifier`, but I think we need to add a section to the docstring that explains the state change caused by `fit` etc. See `BaseForecaster`. Good to have but not a blocker.

yes I agree, on the next PR for base when I change the checking/recasting/capabilities, I want to push through the fit refactor first.

fkiraly:
>I think the default behaviour in _predict should remain? Perhaps with a check whether an infinite look occurs (in the public part of the function) that raises an error.

@TonyBagnall, generally please ask for re-review if changes have been requested and don't merge.

The above has not been addressed, it's an interface breaking change, so it should not have been merged.

It breaks the interface for anyone who has locally implemented a classifier and has only implemented `_predict_proba`, you are breaking the extension contract.

Please unmerge or introduce deprecation logic.

I don't think we should remove the "create `_predict` from `_predict_proba`" logic since this is standard in scikit-learn.

PR #1319, opened 2021-08-17
Replace Appveyor CI for Windows builds with GitHub actions

mloning:
#### Reference Issues/PRs
Partly fixes https://github.com/alan-turing-institute/sktime/issues/1032

#### What does this implement/fix? Explain your changes.
* Adds GitHub Action CI scripts for Windows building and testing
* Removes appveyor CI

PR #1542, opened 2021-10-19
Pin pandas at 1.1.5 to fix windows tests in github actions 

freddyaboulton:
#### Reference Issues/PRs

Unbreaks #1319 

Adds back the <1.20 pin to the pandas version. The problem was that pip installing pandas>1.1.5 after the conda installation caused numpy 1.19.5 to be installed. I suspect that since `sklearn` and `scipy` were installed with conda along side numpy version 1.21.2, the fact that pip uninstalled the conda installed numpy and installed a different version caused incompatibility between `sklearn`, `scipy` and `numpy`.

The CI should be green once we merge in the latest main since the failing tests are for manylinux 2010 (already deprecated) and tsfresh components (removed from tests cause they caused problems).

The only issue is that the gh windows workers do not have openmp installed so I disabled it for now. 

##### For all contributions

##### For new estimators

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -118,7 +118,7 @@ jobs:
       run: python -c "import sys; print(sys.version)"
     - name: Install dependencies
       run: |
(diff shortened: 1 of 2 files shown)


TonyBagnall:
I would approve to give it a go but the tests are failing?

PR #1543, opened 2021-10-20
remove MrSEQL

TonyBagnall:
removing MrSEQL in order to purge cython. We will look to bring it back using numba


MatthewMiddlehurst:
Would just have to put decorator on the _init_ then actually remove it on a later version im pretty sure. Though I thought we wanted to get rid of cython soonish as it was causing issues, not in 3 months or whatever.

PR #1545, opened 2021-10-20
Feature based classification package refactor

MatthewMiddlehurst:
Refactor along the lines of #1146 for the distance_based package (Catch22Classifier, MatrixProfileClassifier, TSFreshClassifier and SignatureClassifier). Reworks the tests for these classifiers to speed them up.

Removes old deprecated signature and catch22 files.


sktime/classification/feature_based/_catch22_classifier.py
```diff
@@ -135,32 +135,34 @@ def _fit(self, X, y):
...
         ----------
-        X : pd.DataFrame of shape (n_instances, n_dims)
+        X : 3D np.array of shape = [n_instances, n_dimensions, series_length]
+            The data to make predictions for.
 
         Returns
         -------
-        preds : np.ndarray of shape (n, 1)
-            Predicted class.
+        y : array-like, shape = [n_instances]
```
chrisholder, line 147: Same here array-like isnt a type

sktime/classification/feature_based/_catch22_classifier.py
```diff
@@ -135,32 +135,34 @@ def _fit(self, X, y):
...
         ----------
-        X : pd.DataFrame of shape (n_instances, n_dims)
+        X : 3D np.array of shape = [n_instances, n_dimensions, series_length]
+            The data to make predict probabilities for.
 
         Returns
         -------
-        predicted_probs : array of shape (n_instances, n_classes)
-            Predicted probability of each class.
+        y : array-like, shape = [n_instances, n_classes_]
```
chrisholder, line 164: array-like isn't a type

sktime/classification/feature_based/tests/test_tsfresh_classifier.py
```diff
@@ -209,30 +173,28 @@ def test_tsfresh_classifier_on_basic_motions():
 #     print(']')
 #
 # if __name__ == "__main__":
```
chrisholder, line 175: Is this staying to help test in the future and so is commented out just so the tests dont run on it, or is it a forgot to remove?

sktime/classification/feature_based/_catch22_classifier.py
```diff
@@ -96,27 +92,31 @@ def __init__(
...
     def _fit(self, X, y):
-        """Fit an estimator using transformed data from the Catch22 transformer.
+        """Fit a pipeline on cases (X,y), where y is the target variable.
 
         Parameters
         ----------
-        X : nested pandas DataFrame of shape [n_instances, n_dims]
-            Nested dataframe with univariate time-series in cells.
-        y : array-like, shape = [n_instances] The class labels.
+        X : 3D np.array of shape = [n_instances, n_dimensions, series_length]
```
chrisholder, line 103: Technically the type is np.ndarray but everyone knows what np.array is so isn't an issue.

chrisholder (changes requested):
Essentially just add a type to the 'array-like' (assuming it's a 1d np.ndarray). I also raised some questions about commented out code/variable names which could be just me missing the reasoning behind it.

sktime/classification/feature_based/_catch22_classifier.py
```diff
@@ -38,7 +37,7 @@ class Catch22Classifier(BaseClassifier):
 
     Attributes
     ----------
-    n_classes : int
+    n_classes_ : int
```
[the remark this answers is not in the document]
TonyBagnall: I think we use the convention that variables set in fit have underscore suffix?

sktime/classification/feature_based/_catch22_classifier.py
TonyBagnall: so we have used array-like historically. I think it needs to be np.array at the moment, I dont think conversion to numpy from df is actually implemented (I'm putting it in on another branch)

TonyBagnall (approved):

MatthewMiddlehurst:
Thanks, Chris.

We use array-like elsewhere, and the acceptable types seem to be changing, so I would rather leave it for now. If it was public facing perhaps it would be more important to change, but it's a private method also.

The test comments are there so it's easy to remake the test arrays if the results legitimately change. Could be cleaner, but that's also how it is everywhere else so would rather wait for an overhaul of tests.

TonyBagnall (approved):

chrisholder (approved):

PR #1548, opened 2021-10-22
Deprecate various

TonyBagnall:
### Reference Issues/PRs
Deprecates MrSEQL cython version in lines with 
https://github.com/alan-turing-institute/sktime/issues/1547
https://github.com/alan-turing-institute/sktime/issues/1538

Deprecation for MrSEQL prior to the removal of cython

is this sufficient to indicate deprecation?


MatthewMiddlehurst:
Tests can be excluded in tests/_config.py by adding them to EXCLUDED_TESTS.

Is the decorator necessary? Could we not just print a warning in the init or at the top of the function or similar?

TonyBagnall:
We could just leave the comment, it's my preference, but would like consensus on it. I will ask core devs

mloning:
You don't have to use the decorator, you can just raise a FutureWarning in the method/function, would that help?

TonyBagnall:
> You don't have to use the decorator, you can just raise a FutureWarning in the method/function, would that help?

I'll try it, thanks, lets see if it still screw the tests

TonyBagnall:
raising a future warning or using warn() doesnt work, the cython breaks

MatthewMiddlehurst:
Are there only a few tests failing? Should be kept in mind that the current exclusion list for ALL tests is currently:

EXCLUDE_ESTIMATORS = [
    "ShapeDTW",
    "ElasticEnsemble",
    "ProximityForest",
    "ProximityStump",
    "ProximityTree",
]

so distances arent really in the greatest shape either way

TonyBagnall:
That explains why the clustering ones are the only fails!

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

TonyBagnall:
seems warn() works and passes the test, could you review @MatthewMiddlehurst and @mloning?

MatthewMiddlehurst (approved):
Not sure about todos in the header comments and it doesn't seem the data_io functions have been moved yet.

Nothing that would block this though.

sktime/utils/data_io.py
```diff
@@ -55,6 +59,10 @@ class values.
         all time-series and (if relevant) a column "class_vals" the
         associated class values.
     """
+    warn(
```
mloning, line 62: can we make this a future warning `warn("msg", FutureWarning)`

mloning: and the other warn call below

mloning (commented):
Just one minor comment, everything else looks good

MatthewMiddlehurst (approved):

Issue #813, opened 2021-04-15
MiniRocket as an Option for ROCKETClassifier


angus924:
**Is your feature request related to a problem? Please describe.**
NA

**Describe the solution you'd like**
I thought I'd ask first, before submitting a PR—@MatthewMiddlehurst because it's your code, @kachayev and @RavenRudi because you are working on related PRs—would it be helpful to add [MiniRocket](https://github.com/alan-turing-institute/sktime/blob/main/sktime/transformations/panel/rocket/_minirocket.py) as an option for ROCKETClassifier?

If so, I'd suggest making MiniRocket the default, because it's so much faster than Rocket, but this is only a suggestion.

**Describe alternatives you've considered**
Alternatively, of course, we could have a separate MINIROCKETClassifier.  This would essentially just duplicate all the ROCKETClassifier code, but perhaps it would be clearer to do it this way.

**Additional context**
NA

kachayev:
I think it would make sense to have optional `transformation` parameter for `ROCKETClassifier` class that could be set to one of the following: `Rocket`, `MiniRocket` or `MiniRocketMultivariate`. Though I would keep  `Rocket` default for backwards compatibility.

RavenRudi:
Hi,

I agree with @kachayev. That is probably the best option. What I did in my PR, is refactoring duplicate code of the `ROCKETClassifier` and `ROCKETRegressor` into `series_as_features/base/estimators/shapelet_based/_rocket_estimator.py`. That's were the `transformation` parameter could be added.

MatthewMiddlehurst:
Having it as a parameter would be best. No opinion on the best default value.

mloning:
I think it would be nice to develop a more unified interface for using Rocket/MiniRocket at the level of the RocketClassifier/RocketRegressor. Adding a `transformation` or `method` argument seems fine, with possible values being "minirocket" or "rocket". I would not add `MiniRocketMultivariate". I prefer multivariate data being handled automatically when given in fit/predict without requiring users to set hyper-parameters. If multivariate data is not supported, we should raise an error.

angus924:
Thanks very much everyone.

TonyBagnall:
hi, just bumping this, does this still need doing? If so, I may find someone to do so, but dont want to duplicate what others have done.

MatthewMiddlehurst:
Related to #1095.

TonyBagnall:
closing this issue, direct it to #1095

PR #1567, opened 2021-10-30
Integrate multi rocket

### LINKED ISSUE #1558: [ENH] Integrate MultiRocket
Include MultiRocket (https://github.com/ChangWeiTan/MultiRocket) as a classification algorithm. 

MultiRocket can be integrated analogously to Rocket or MiniRocket. In my test data sets, the algorithm showed very good results. I have already redesigned the code of @ChangWeiTan accordingly so that there is a transformer and a classifier analogously to Rocket. @ChangWeiTan could possibly say whether integration in sktime is also in his sense. After that, I would upload the code accordingly as a pull request. Since I'm not familiar with the licenses, I don't know if I can just use the code.

fstinner:
#### Reference Issues/PRs
Fixes #1558. This PR integrates MultiRocket and MultiRocketMultivariate as transformers and MultiRocketClassifier as well as MultiRocketMultivariateClassifier as classifiers.

I have restructured @ChangWeiTan MultiRocket (https://github.com/ChangWeiTan/MultiRocket) for sktime interfaces. I redesigned the source code into a transformer and a classifier. In the original source code, there was also a part of the preprocessing (transformer) in the classifier itself. The number of features here refers to the total number of features. In the original source code, the total number of features is only used in the classifier. This is analogous to MiniRocket, but there the number of kernels=number of features. Therefore, I have chosen the number of features as input for the classifier and the transformer. The total number of kernels is then calculated internally (num_kernels_total). num_features=50_000 corresponds approximate to MiniRocket (num_features=10_000) with the corresponding 4 additional features per kernel.

No, no further dependencies.

The reviewer should check if the interfaces are correct and if the variables are named correctly.

##### For all contributions

##### For new estimators
MultiRocket works similar to Rocket or MiniRocket. Therefore no new examples are required.


MatthewMiddlehurst:
I thought we had another issue relating to mini-rocket similar to this, but it seems it was closed #813. 

While the transformer contributions are great, I don't think having a new classifier for each is necessary (especially multivariate versions).

Like Markus said in the above issue, having "rocket", "minirocket" and "multirocket" would be good as a parameter for ROCKETClassifier, while the multivarite versions should be handled automatically based on the data input. 

Maybe just MultiRocketClassifier and ROCKETClassifier (is it ROCKET or Rocket?) would be acceptable, since the parameters seem the be quite different.

MatthewMiddlehurst:
Also it appears you have been hit by the [code quality checks](https://dev.azure.com/mloning/sktime/_build/results?buildId=6608&view=logs&jobId=594eef4e-0c7a-52cb-8378-dee4e90c28fa&j=594eef4e-0c7a-52cb-8378-dee4e90c28fa&t=4479d0ca-b591-5791-6ab3-63abe80fc9c7). Setting up the dev enviroment and [pre-commit](https://www.sktime.org/en/v0.5.1/contributing.html#id12) is a good way for dealing with that.

TonyBagnall:
I'm in favour of a single configurable ROCKETClassifier

fstinner:
I can understand that the multitude of classifiers is too complicated. Linter was installed but didn't seem to show anything for me. The erroneous imports came from a short-term copy of the source code.

The question is whether to use the central parameter "num_features" or "num_kernels" for all Rockets for one "ROCKETClassifier" or "RocketClassifier". An example: num_kernels=10_000 means for Rocket num_features=20_000, for MiniRocket num_features=10_000, for MultiRocket num_features=50_000. For consistent use as a classifier, one would just have to decide whether to use num_features or num_kernels consistently. I find num_features more intuitive since it corresponds to the actual output of the transformer.

MatthewMiddlehurst:
My preference is `num_kernels`, it’s what is compared in the paper, is more descriptive of what’s actually going on and lets you set a sensible default for all of them (or is the default used in mini and multi rockets different than 10k?). 

Also, with `num_features` you have to deal with a bunch of verification. i.e. what if the user inputs an odd number of features for ROCKET or the number of features isn’t divisible by 5 for MultiRocket.

Are `max_dilations_per_kernel` and `n_features_per_kernel ` differences from the ROCKET algorithm for MultiRocket or just differences in implementation?

fstinner:
> My preference is `num_kernels`, it’s what is compared in the paper, is more descriptive of what’s actually going on and lets you set a sensible default for all of them (or is the default used in mini and multi rockets different than 10k?).

The number of kernels is the same in each case. MultiRocket is in principle MiniRocket with 4 additional features per kernel. I will change the input parameter to num_kernels accordingly.

> Also, with `num_features` you have to deal with a bunch of verification. i.e. what if the user inputs an odd number of features for ROCKET or the number of features isn’t divisible by 5 for MultiRocket.

In fact, the `num_features` is also only approximated, since it is the nearest multiple of 4*84=336 < 50,000

> Are `max_dilations_per_kernel` and `n_features_per_kernel ` differences from the ROCKET algorithm for MultiRocket or just differences in implementation?

`max_dilations_per_kernel` and `n_features_per_kernel` are analogous to Rocket. The question is whether the use of the API offers an advantage. In my analyses, I could see that the default variant was not always the best solution. Thus, controlling the parameters offers the advantage of tuning the algorithm. But the `num_kernels` is of course the most decisive factor.

Another thing I noticed: `normalise` is spelled differently than `normalize` in sklearn (https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.normalize.html) (British/Australian? English vs American English). Which language version should be used consistently?

TonyBagnall:
@fstinner thanks for this. I'm happy to take advise from anyone at Monash on naming.

ChangWeiTan:
Hi all, 

I thought about this a lot when I was implementing MultiRocket. In the end, I prefer `num_features` because that is more intuitive to compare with all the Rockets in the paper - 10k features is 10k features for all of them but 10k kernels is not the same as 10k kernels for all the Rockets. In the transformer itself, we can just calculate `num_kernels` for each Rocket (`num_kernels=num_features` for MiniRocket, `num_kernels=num_features/2` for Rocket, and `num_kernels=int(num_features/2/4)` for MultiRocket). Note that however the actual `num_features` might be slightly less than the given `num_features` as it is the nearest multiple of 84 less than `num_kernels`, calculated from the given `num_features`. 

But if we don't care about the comparisons and how the actual `num_features` might be different from what is given, `num_kernels` is more user friendly. I am happy with that too.

Hope it makes sense.

MatthewMiddlehurst:
I have sorted the code quality issues. The test you implemented seems to be failing, however.

Happy to take input regarding the naming of inputs and the transformer itself.

fstinner:
The naming itself makes sense (`multirocket` instead of `multi_rocket`). With the number of features I am still unsure whether the num_kernels is actually understandable for outsiders (possibly even non-scientists). Ultimately, the num_features can be supplemented by MultiRocket with the help of additional features. In Rocket, 2*num_kernels of features are also extracted. This makes the comparison of MiniRocket and Rocket unfair. In `MiniRocket(num_kernels=20_000)`, `Rocket(num_kernels=10_000)`, `MultiRocket(num_kernels=2_500)` the num_features are the same. One could also add more features to each kernel and these could also minimally improve the comparison depending on the dataset. In the end, however, only the num_features are inflated. That is in my estimation also the core statement that @ChangWeiTan wanted to make (feel free to correct me). 

I have adjusted the transformer. It was still at `num_kernels=10_000` unfortunately. I have supplemented it on `num_kernels=6_250`. Thus this corresponds also to the approach of @ChangWeiTan. With me the tests ran locally again.

MatthewMiddlehurst:
I don't think the number of features extracted per kernel being different is so important, they are different algorithms. Comparing the algorithms with number of kernels is just as valid IMO. I would be more confused in the case where I explicitly set the number of feature output and then receive a different amount. 

The output features per kernel should be easily accessible information from the class docs. Especially when you have `n_features_per_kernel`, I don't feel like it would be confusing to anyone who reads the docstrings. Though, to be fair, the `num_features` issue above could also be solved by just explaining it in the docs.

At the end of the day though, it isn't my algorithm. I'm fine with being outvoted and going with `num_features` as long as the 3 are consistent.

TonyBagnall (approved):
hi, I think I agree with Matt re number of kernels/features, in that getting back more features than I set the argument to would seem confusing. However, as Matt said, its your algorithm so go with what you think best.

TonyBagnall:
thanks again for the contribution. Re the testing, could you use UnitTest instead of GunPoint? Our tests take forever, so we are trying to speed things up.

ChangWeiTan:
Hi, I went with `num_features` mainly for the sake of the paper. I agree that it will be confusing to get different number of features from what was inputted. So `num_kernels` makes more sense for packages and users. I think more experienced users would be able to work out the number of features per kernel; less experienced users can refer to the docs or paper if they are really interested in the process of getting the number of features. So, I'll go with `num_kernels`.

TonyBagnall (commented):
hi @fstinner in light of conversation, could you change num_features to num_kernels then this can go in? Really pleased to get this contribution, thanks.

fstinner:
In the code `num_kernels` is already implemented. I do not object to the unified representation using `num_kernels`. It is definitely unambiguous. It is not my algorithm and it is mainly a style issue in the end. I just made minor changes in the documentation trying to explain the difference between `num_kernels` and the calculation of `number of features`. Happy to check if the explanation can be understood that way.

@TonyBagnall For the tests, I am happy to replace `load_gunpoint` with `UnitTest`. Unfortunately, I didn`t find an example here in the `transformer` folder or `datasets`, so I`ll gladly replace that if someone can give me an example.

MatthewMiddlehurst:
@fstinner its mainly the classification algorithms which use it. Just replacing `load_gunpoint` with `load_unit_test` is fine or we can do it.

Will give this a proper read through again at some point this weekend or Monday, but looks good. Thanks for contributing.

MatthewMiddlehurst (approved):
The only concerns I have are the necessity of the large indices arrays being static and the datasets used for testing, but neither of those will block it from going in.

The arrays seem like a bit of an unnecessary optimisation, if numba is the issue it could just be an input? If there is another reason it fine if it stays though. As mentioned earlier the unit_test dataset is a lot quicker for testing, Rocket is fast already, but good to have some uniformity with this.

TonyBagnall (approved):
looks good

TonyBagnall:
Appveyor is broken, but this passed before, we have two approvals and is ready to go in

Issue #1565, opened 2021-10-30
[DOC] Add toggle button to code examples to remove ">>>", "..." and output 


RNKuhns:
#### Describe the issue linked to the documentation

The way code examples must be formatted in code-blocks is not conducive to copying the examples into a code editor to try out. Scikit-learn has a toggle button on their code examples to remove the ">>>", "...", output and other items that are needed in the code-block, but make it difficult to copy the code directly.  We should implement something similar.

#### Suggest a potential alternative/fix

We need to add javascript to every page that lets users hide the items described above so that the example code is easier to copy. We can easily adapt Scikit-learn's javascript blurb that does this. Then we need to make sure it is included on every page using a layout.html template.

TonyBagnall:
assume we can close this @RNKuhns ?

PR #1572, opened 2021-10-30
Add toggle button to make examples easy to copy

RNKuhns:
#### Reference Issues/PRs

#1565 

It adds  a layout.html template to extend the pydata-sphinx-theme when building our docs. We need to run a javascript script after the content of pages is loaded. This script will identify example code blocks (e.g. blocks with python highlighting) and add a toggle button so we can turn on/off the text that makes the examples difficult to copy (e.g. >>>, ...,  example output, etc). 

It also adds css styling for the new button. Changes to the conf.py should just be to add the css styling file.

##### For all contributions

##### For new estimators

### DIFF
--- a/docs/source/_static/css/custom.css
+++ b/docs/source/_static/css/custom.css
@@ -1,0 +1,26 @@
+  /* example_toggle that hides >>>, ..., etc */
+.example_toggle {
+    cursor: pointer;
+    position: absolute;
+    top: 0px;
+    right: 0px;
+    border: 1px solid rgb(221, 221, 221);
+    color: rgb(221, 221, 221);
+    font-family: monospace;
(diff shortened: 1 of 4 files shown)


RNKuhns:
@aiwalter and @mloning we should go through and make sure it seems like this is working on every page after the docs build (it appears to work locally). 

Note that there is likely a way to have the javascript script in a separate file (in _static/js/) but I've had trouble getting that to load into the layout.html template update and work correctly, so I went with what is working.

TonyBagnall (approved):
A nice improvement +1 from me

aiwalter (approved):
Great thx

mloning (approved):
Nice work @RNKuhns

PR #1573, opened 2021-10-31
Governance: eligibility and end of tenure clarification

fkiraly:
This PR makes minior changes to the governance document. It adds, for all roles:

* an "eligibility" section, which explicitly states conditions on role eligibility
* an "end of tenure" section, which describes how the tenure ends

In most cases, this has been already part of the respective sections, and just moved to a new section and/or explicitly clarified.

### DIFF
--- a/docs/source/governance.rst
+++ b/docs/source/governance.rst
@@ -159,6 +159,10 @@ Rights and responsibilities
    * - Nomination
      - They can nominate new core developers, CoC committee members and CC members.
 
+Eligibility
+^^^^^^^^^^^
+
+Anyone is eligible to be a core developer.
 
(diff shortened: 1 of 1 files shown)


TonyBagnall (approved):

fkiraly:
Just FYI, I should have said this before this was merged:

before, it said that "one of the CC members will have a tie breaking privilege", but it didn't say who that was.

I've changed it to the CC member with shortest tenure, that would be you, @aiwalter.

If this is contentious, we can unmerge and remove the change, but note that it wouldn't specify who exactly has the tie breaking privilege as of now.

TonyBagnall:
The voting rule seemed reasonable to me

Issue #1547, opened 2021-10-22
[ENH] Replace Cython version of MrSEQL with numba version


TonyBagnall:
**Is your feature request related to a problem? Please describe.**
the way we are currently using cython causes us no end of problems, and we think the simplest solution is simply to use numba instead. See
https://github.com/alan-turing-institute/sktime/issues/1538

apart from distances, MrSEQL is the only algorithm to use cython
https://github.com/alan-turing-institute/sktime/tree/main/sktime/classification/shapelet_based/mrseql

I will depreciate this version, and we would like to replace it with a numba version.

**Describe the solution you'd like**

A numba based implementation of MrSEQL that produces results that are not significantly different to the current version.

TonyBagnall:
rolled into #1538

Issue #1538, opened 2021-10-18
[ENH] remove all use of cython, use numba instead


TonyBagnall:
cython is a problem with installation. Whether we have just done it wrong is a moot point. numba is a so much clearer and lightweight, if there is no performance hit then its should be used instead of cython. This effects

1. distances: need a numba version of them all
2. knn classifier
3. EE classifier
4. MrSEQL classifiier

first three should be easy to sort, last we will need to talk developer

TonyBagnall:
this has progressed and been split into four issues

PR #1574, opened 2021-11-01
Numba distance module

chrisholder:
#### Reference Issues/PRs

 #1538 #1547 

This pr takes all the current distance metrics in sktime and converts them to numba. The distances can be used in the exact same way they were before (params are consistent). Furthermore, a multitude of bugs/errors were found and fixed.

In addition there are 3 new features that come along with this. The first is pairwise distance. This allows you to perform a distance computation across a dataset of timeseries or between two timeseries with massive performance gains.

The second (a much demanded request) both distance and pairwise distance now support 1d and 2d arrays.

The third new feature is the ability to create a numba compiled function for your specific use case via a factory method. This then allows you to embed your created function directly into another numba function  (you can't call non numba compiled functions from inside another numba meaning you have to exit effectivly removing any potential performance gains). This will allow for many other modules to more easily be converted to numba (i.e. classification and clustering) and find massive performance gains.

Finally there were no docstrings or tests for the previous distances and this pr adds a huge amount of testing and validation and well as docstrings to everything.

There is also custom logic to create LowerBounding matrices compatiable with the module.

No

There is a lot stuff here (sorry). Normally i'd split this up but the intention of the pr was to swap the old distances out all at the same time and so we never have to touch cython again. Most of the 'new' stuff is really docstrings and tests that the previous distances were severily lacking (there were no docstrings or tests at all).

When looking at the code there are going to be a couple of things that will appear a 'weird' way of doing something. 90% of the time it's will have something to do with working around a restriction with numba so please ask about anything you see.

##### For all contributions

##### For new estimators


chrisholder:
I forgot to add there are 3 more distances coming Ill finish them tomorrow.

chrisholder:
As an addition because someone is going to bring up the location of it being in sktime/dist-kernels/numba/distance/... I put it in here just because @fkiraly suggested it would be a good location, but we need to have a discussion about the 'dist-kernels' name and just renaming this whole directory to 'distances' now we're replacing the cython stuff. But ill let you disucss I have no preference.

TonyBagnall (commented):
hi chris,

awesome. Couple of points
1. I think this should be in distances. My requests to change dists_kernels name were ignored, but I just dont like it. Refactoring takes an age these days, so better to put it where it belongs now, in distances directory
2. The dtw_based directory is not required. ERP and LCSS are not dtw based, and generally there is too much nesting of directories. Jusat put them directly in distances. Also the name distance is redundant in the private files
distances/_euclidean
distances/_dtw
etc. Similar redundancy in the private function names, 
3. They do not take 1D arrays, which you said they would. This is essential. I could do without all the callable stuff, I just wanted 1D distance functions.

TonyBagnall:
also we are using github IDs not full names **__author__ = ["Chris Holder"]**

TonyBagnall:
update CODEOWNERS

TonyBagnall:
I think these should all be accessed via an init rather than directly?
image

TonyBagnall:
this test failing is something we see all the time when refactoring, I'm not sure if its a bug or not, maybe we should raise an issue? @MatthewMiddlehurst has the same on the distance_based  refactor. 
image

TonyBagnall:
thats a strange error!

TonyBagnall:
image

TonyBagnall:
great, simple use case of full DTW seems to work fine, I will run it through the lot later to compare to cython and tsml. Questions to address prior to approval are
1) setting parameters (how do you do it and is it equivalent)
2) tuning parameters in the way done in knn (is it still fast)

TonyBagnall:
I think we will get this lot in then raise an issues for MSM, KNN and PF.

chrisholder:
The results I know will differ from the previous knn are for:

ddtw - changed the derivative algorithm to match the original papers implementation
wdtw - this I may have implemented wrong but I know it gives a different answer
wddtw - as derivative algorithm changed will give different (and depending on if wdtw is correct
edr - this is new (I accidently implemented this one as I missread erp 😂. erp is also implemented as well)

fkiraly (approved):
Looks good to me - clean and well-structured imo.
Implements the changes we discussed earlier verbally.

fkiraly:
I'm going to merge this to unlock the CI/CD refactor, I assume it's "done" and there are no objections?

TonyBagnall:
hi @fkiraly there is still work in progress, but it can go in and we can refine in a new PR

TonyBagnall:
lol, ok then, 2 mins to answer!

fkiraly:
whoops, sorry - I thought this was done (according to statements of @chrisholder).
I thought worst case we can unmerge, but if you're fine to have increments in another PR, I would prefer that too since the "big chunk of work" is now in.

TonyBagnall:
Yes it all good, just a bit surprising :)

PR #1579, opened 2021-11-03
Implement predict_quantiles and predict_intervals in theta forecasters, tests in base class

kejsitake:
#### Reference Issues/PRs

Related to https://github.com/alan-turing-institute/sktime/issues/984

Implements a function to test predict_quantiles

No

Implemented these checks:
- [X] check if the multi-indexed column names are the same as the expected ones
- [X] check if the row indexes or the time frame are the right ones 
- [X] check that the predicted quantiles are strictly monotonically increasing

Modified extension docs to account for base forecasters class:
- [X] removed _compute_pred_int because according to the STEP 13 that should only be in theta
- [X] added _predict_interval function
- [X] added _predict_quantiles
- [X] added capability:pred_int to the tags

Changes in the Theta classifier:
- [X] Implemented predict_quantiles using compute_pred_int
- [X] Implemented predict_interval

### DIFF
--- a/extension_templates/forecasting.py
+++ b/extension_templates/forecasting.py
@@ -223,6 +223,7 @@ class MyForecaster(BaseForecaster):
         # implement here
         # IMPORTANT: avoid side effects to y, X, fh
 
+
     def _predict_quantiles(self, fh, X=None, alpha=0.5):
         """
         Compute/return prediction quantiles for a forecast.
--- a/extension_templates/forecasting.py
+++ b/extension_templates/forecasting.py
@@ -219,7 +219,6 @@ class MyForecaster(BaseForecaster):
         """
         self.update(y, X, update_params=update_params)
         return self.predict(fh, X)
-
         # implement here
         # IMPORTANT: avoid side effects to y, X, fh
 
--- a/extension_templates/forecasting.py
+++ b/extension_templates/forecasting.py
@@ -222,7 +222,6 @@ class MyForecaster(BaseForecaster):
         # implement here
         # IMPORTANT: avoid side effects to y, X, fh
 
-
     def _predict_quantiles(self, fh, X=None, alpha=0.5):
         """
         Compute/return prediction quantiles for a forecast.
--- a/sktime/forecasting/tests/test_all_forecasters.py
+++ b/sktime/forecasting/tests/test_all_forecasters.py
@@ -309,13 +309,14 @@ def test_predict_pred_interval(Forecaster, fh, alpha):
         f = Forecaster.create_test_instance()
         y_train = _make_series(n_columns=n_columns)
         f.fit(y_train, fh=fh)
+        y_pred = f.predict(alpha = alpha)
         if f.get_tag("capability:pred_int"):
-            y_pred, pred_ints = f.predict(return_pred_int=True, alpha=alpha)
+            pred_ints = f.predict_interval(fh)
             _check_pred_ints(pred_ints, y_train, y_pred, fh)
 
         else:
             with pytest.raises(NotImplementedError, match="prediction intervals"):
-                f.predict(return_pred_int=True, alpha=alpha)
+                f.predict(alpha=alpha)
 
 
 def _check_predict_quantiles(pred_quantiles: list, y_train: pd.Series, fh, alpha):
@@ -374,7 +375,7 @@ def test_predict_quantiles(Forecaster, fh, alpha):
                 f.predict_quantiles(fh=fh, alpha=TEST_ALPHAS)
         else:
             quantiles = f.predict_quantiles(fh=fh, alpha=alpha)
-            _check_predict_quantiles(quantiles, y_train, alpha)
+            _check_predict_quantiles(quantiles, y_train, fh, alpha)
 
 
 @pytest.mark.parametrize("Forecaster", FORECASTERS)
--- a/sktime/forecasting/tests/test_all_forecasters.py
+++ b/sktime/forecasting/tests/test_all_forecasters.py
@@ -15,6 +15,7 @@ __all__ = [
     "test_y_multivariate_raises_error",
     "test_get_fitted_params",
     "test_predict_time_index_in_sample_full",
+    "test_predict_pred_interval",
     "test_update_predict_single",
     "test_y_invalid_type_raises_error",
     "test_predict_time_index_with_X",
@@ -303,11 +304,11 @@ def test_predict_pred_interval(Forecaster, fh, alpha):
     AssertionError - if Forecaster test instance does not have "capability:pred_int"
             and no NotImplementedError is raised when asking predict for pred.int
     """
-    f = Forecaster.create_test_instance()
+    f = _construct_instance(Forecaster)
     n_columns_list = _get_n_columns(f.get_tag("scitype:y"))
 
     for n_columns in n_columns_list:
-        f = Forecaster.create_test_instance()
+        f = _construct_instance(Forecaster)
         y_train = _make_series(n_columns=n_columns)
         f.fit(y_train, fh=fh)
         y_pred = f.predict(alpha=alpha)
--- a/sktime/forecasting/tests/test_all_forecasters.py
+++ b/sktime/forecasting/tests/test_all_forecasters.py
@@ -295,7 +295,7 @@ def test_predict_pred_interval(Forecaster, fh, alpha):
     ---------
     Forecaster: BaseEstimator class descendant, forecaster to test
     fh: ForecastingHorizon, fh at which to test prediction
-    alpha: float, alpha at which to make prediction intervals
+    alpha: float, coverage at which to make prediction intervals
 
     Raises
     ------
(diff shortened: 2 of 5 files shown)


sktime/forecasting/tests/test_all_forecasters.py
```diff
@@ -286,11 +287,56 @@ def test_predict_pred_interval(Forecaster, fh, alpha):
             with pytest.raises(NotImplementedError, match="prediction intervals"):
                 f.predict(return_pred_int=True, alpha=alpha)
 
+def _check_predict_quantiles(pred_quantiles: list, y_train: pd.Series, fh, alpha):
+    # alpha can be a float or a list
+    # if a list, it has to be sorted
+    # check if alpha is increasing and in the given interval
+    if isinstance(alpha, list):
```
fkiraly, line 294: hm, this looks more like a test that the `BaseForecaster` should carry out? `alpha` is an input, and we know that the `alpha` in the test are fine? 

I suggest:
* move this to the `BaseForecaster` as an input check for the `alpha`; extend this that it checks that `alpha` is a `float`, or list of `float`.
* add a test which checks that `BaseForecaster` raises an error for "wrong" `alpha`.

sktime/forecasting/tests/test_all_forecasters.py
```diff
@@ -286,11 +287,56 @@ def test_predict_pred_interval(Forecaster, fh, alpha):
...
+    _assert_correct_pred_time_index(pred_quantiles.index.levels[0], y_train.index[-1], fh)
+
+    # multiply variables with all alpha values
+    expected = pd.MultiIndex.from_product(y_train.columns, alpha)
+    assert all(expected == pred_quantiles.columns.to_flat_index())
+
+    # check if values are monotonically increasing
+    for var in pred_quantiles.columns.levels[0]:
+        for index in range(len(pred_quantiles.index)):
+            assert pred_quantiles[var].iloc[index].is_monotonic_increasing
```
fkiraly, line 316: this doesn't always give the expected result, since the user could pass `alpha` in the list in a non-monotonic order? An easy way around this is to retrieve the alpha, and sort the columns before checking.

sktime/forecasting/tests/test_all_forecasters.py
fkiraly: isn't there already a `check_alpha`, btw?

fkiraly:
@kejsitake, it seems your failures are related to linting, i.e., code formatting conventions that `sktime` enforces.
For instance, use of double-quote instead of quote in strings.

You can check what is going wrong by clicking on "details" next to the failing test, or locally in your dev IDE if you have linting and testing set up there. Let us know in the community collab session if your local set-up is not fully working, this can be fiddly, especially on Windows.

kejsitake:
I finally figured out how to run pre-commit locally. However, the PR is not ready for review yet, just commiting my changes.

extension_templates/forecasting.py
```diff
@@ -90,6 +90,7 @@ class MyForecaster(BaseForecaster):
         "requires-fh-in-fit": True,  # is forecasting horizon already required in fit?
         "X-y-must-have-same-index": True,  # can estimator handle different X/y index?
         "enforce_index_type": None,  # index type that needs to be enforced in X/y
+        "capability:pred_int": False,  # is forecaster probabilistic?
```
fkiraly, line 93: maybe a better description: "does forecaster implement `predict_quantiles`?"

Also consider renaming this to `"capability:predict_quantiles"`, but perhaps it's best to leave it like this since people may know the tag.

extension_templates/forecasting.py
```diff
@@ -238,19 +239,62 @@ def _update_predict_single(
 
     # todo: consider implementing this, optional
     # if not implementing, delete the method
-    def _compute_pred_int(self, alphas):
-        """Calculate the prediction errors for each point.
+    def _predict_interval(self, fh, X, coverage):
```
fkiraly, line 242: right idea, although I think it may be confusing. In the extension template I'd only include `_predict_quantiles`, because `_predict_interval` automatically calls that. I'd add an explanation that the user can implement `_predict_interval` as well if there is a faster variant, but I think it is so rare that it shouldn't be in the template as a "fill in" function, but only mentioned in the body of `_predict_quantiles`.

extension_templates/forecasting.py
```diff
@@ -238,19 +239,62 @@ def _update_predict_single(
 
     # todo: consider implementing this, optional
     # if not implementing, delete the method
-    def _compute_pred_int(self, alphas):
```
fkiraly, line 241: important: we also need to remove the "old" prediction interval related arguments from `_predict`, and similarly from the docstring preamble at the top (`return_pred_int` etc)

sktime/forecasting/tests/test_all_forecasters.py
```diff
@@ -278,19 +279,61 @@ def test_predict_pred_interval(Forecaster, fh, alpha):
...
-
+        y_pred = f.predict(alpha=alpha)
+        if not f._has_predict_quantiles_been_refactored():
+            with pytest.raises(NotImplementedError):
+                pred_ints = f.predict_interval(fh)
+                _check_pred_ints(pred_ints, y_train, y_pred, fh)
         else:
-            with pytest.raises(NotImplementedError, match="prediction intervals"):
-                f.predict(return_pred_int=True, alpha=alpha)
+            f.predict(alpha=alpha)
```
fkiraly, line 288: I'm confused about this line - should this not be `f.predict_interval`?

fkiraly: we also need to do something to check the return. It should be in the expected format.

sktime/forecasting/tests/test_all_forecasters.py
```diff
@@ -278,19 +279,61 @@ def test_predict_pred_interval(Forecaster, fh, alpha):
...
+
+
+def _check_predict_quantiles(pred_quantiles: list, y_train: pd.Series, fh, alpha):
```
fkiraly, line 291: we should probably also test this for multivariate `y`, too

fkiraly: you do this below, but I think the type hints are inaccurate

sktime/forecasting/theta.py
```diff
@@ -162,9 +162,9 @@ def _predict(self, fh, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
         if self.deseasonalize:
             y_pred = self.deseasonalizer_.inverse_transform(y_pred)
 
-        if return_pred_int:
-            pred_int = self.compute_pred_int(y_pred=y_pred, alpha=alpha)
-            return y_pred, pred_int
+        # if return_pred_int:
```
fkiraly, line 165: the `return_pred_int` and `alpha` arguments should also be removed

fkiraly (changes requested):
Thanks, nice work!

First of all, I'd suggest to move the work on the extension template into a separate PR if it's not too much of a hassle. This is independent, so could be independently reviewed on. I feel we're almost there with that.

On the tests:
* tests look good in general!
* I think there's a problem with the logic around handling whether the method has been refactored already or not. I think methods that haven't been refactored will currently break if their `predict_interval` or `predict_quantiles` are run? The tests don't account for that (though I may be mistaken). Perhaps have another look at what happens for refactored, and non-refactored estimators?

On theta:
* I think this is very convoluted - methods calling methods calling methods unnevessarily, i.e., `predict_quantiles` calls `compute_pred_int` which calls `_compute_pred_err`. I think out of the three items, there is still most work needed. What I suggest below will involve deleting a lot of lines of code, so the work is in carefully thinking what calls what and cutting down unnecessary layers.
* Suggested strategy: just kill `compute_pred_int` and `_compute_pred_err` entirely, and move the real logic into `predict_quantiles`. It will be much nicer (and less convoluted).
* `_predict_interval` is not needed at all, once you've implemented `predict_quantiles`.

kejsitake:
Thank you for the comments @fkiraly.

I believe I have solved most of them. I noticed that one of the tests fails. Upon looking more at the issue, this is happening because there is still a test `test_predict_pred_interval` for the old format of `pred_ints`, which expects the columns of the returned dataframe to be `lower` and `upper`.  But, the new predict_interval calls `predict_quantiles` and returns a dataframe in the new format(with the multi index), causing the test to fail. I am not sure if it is safe to remove this test, or we should keep it at this time, and I should instead look into convert the format to the old one.

kejsitake:
I thought some more about this. In my opinion the test for the old format of the output can be removed. I wasn't sure about the format that predict_intervals should return. Currently it returns the same format dataframe as predict_quantiles. I thought about renaming the first level to "Intervals" instead of "Quantiles", but that didn't seem correct either. I didn't find a specification about this on the STEP either, but please let me know if I am missing anything.

fkiraly:
> But, the new predict_interval calls predict_quantiles and returns a dataframe in the new format(with the multi index), causing the test to fail. I am not sure if it is safe to remove this test, or we should keep it at this time, and I should instead look into convert the format to the old one.

The story is, `predict` should still return the *old* format correctly until we have deprecated it completely.
For this, the function `_convert_new_to_old_pred_int` has been written, it should convert the new format to the old format.

So, the test should pass if `_predict_quantiles` is implemented correctly and `_convert_new_to_old_pred_int` is correct, which may not be the case (I would recommend to theck).

Due to the deprecation, we shouldn't remove the "old" test yet - we should remove it when we deprecate the old prediction interval interface.

fkiraly (changes requested):
Yes, the earlier issues seem to have been resolved.

Major comments:
* you accidentally deleted the `forecaster` extension template? This should be reverted.
* the test for the old prediction intervals should pass, it should work out-of-the box via `_convert_new_to_old_pred_int`. Could you kindly check why it doesn't? Could be the implementation, could be the conversion. May be worth asking @SveaMeyer13 who I believe wrote the conversion, if the test failure can be localized to there. But it could also be that you are returning the wrong format in `_predict_quantiles`, or sth with the base class.

kejsitake:
Just summarizing my latest changes on this PR to make it easier to review:

- I reverted the forecaster extension template, which I had removed by mistake when splitting this PR.
- I added the conversion to the old format and the old predict interval tests now pass
- I fixed a bug related to rounding of c during division by converting it to a float.

All the tests seem to pass successfully now.

fkiraly (dismissed):
Looks good, issues seem to have been addressed.

Could be merged if it weren't up to the merge conflict which should be resolved.

sktime/forecasting/tests/test_theta.py
```diff
@@ -1,4 +1,9 @@
 # -*- coding: utf-8 -*-
+"""Tests for ThetaForecaster.
+
+# copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+"""
+
 __author__ = ["@big-o"]
```
fkiraly, line 7: kindly add yourself, too to this

fkiraly:
also, could you kindly comment on the open discussions, how they have been resolved and click "reolve"? So unresolved conversations = 0.

fkiraly (approved):
looks good now, thanks!

PR #1584, opened 2021-11-03
Distance based classification package refactor

MatthewMiddlehurst:
Refactor along the lines of #1146 for the distance_based package (PF, EE, and ShapeDTW). Adds tests for these classifiers as well.

I have left KNeighborsTimeSeriesClassifier mostly alone. It requires changes beyond what I am willing to look into for this PR. In its current fit it calls a different version of self._fit inherited from sklearn NeighborsBase.


TonyBagnall:
so is this failure due to a CI bug? I seem to remember seeing this kind of failure before when refactoring. Its saying it cannot see KNeighborsTimeSeriesClassifier, but it is in init 
image

TonyBagnall:
I had hoped that the bug would be resolved by rerunning the tests, but it isnt. I think this is a problem with the testing, not your code.

TonyBagnall:
I am going to try recreate this bug on a branch, its annoying

TonyBagnall:
something going on with the stump. I actually think we should make "ProximityStump" and "ProximityTree" private. I see no reason to use them independently of ProximityTree.

TonyBagnall:
error is 
    value = getattr(self, key)
AttributeError: 'ProximityStump' object has no attribute 'get_exemplars'

TonyBagnall:
and there is even an issue, google object has no attribute 'get_exemplars' gives this, lol  https://github.com/alan-turing-institute/sktime/issues/598

chrisholder (approved):

TonyBagnall (approved):

PR #1595, opened 2021-11-08
fix issue with update_y_X when we refit forecaster

### LINKED ISSUE #1593: [BUG] evaluate raises KeyError for AutoETS and CutoffSplitter / SlidingWindowSplitter - issue with BaseForecaster._update_y_X
**Describe the bug**
evaluate function raises an invalid error for the CutoffSplitter and SlidingWindowSplitter. 
ExpandingWindowSplitter works fine.

The error initially appears to be in the _predict method of ets.
But, a few debug prints later I managed to pinpoint the issue in the [_update_y_X](https://github.com/alan-turing-institute/sktime/blob/922cd71d0d82d849025a080be826cf1c3c4777e5/sktime/forecasting/base/_base.py#L804) method of the BaseForecaster class.

If the forecaster was fitted before and is now refitted, forecaster._y is not changed based on the new y_train and this causes a difference in the absolute horizon of y and the prediction output. This explains why we don't get the error in the ExpandingWindowSplitter as in that case forecaster._y[0] is the same for all splits.

**To Reproduce**
```python
import numpy as np
import pandas as pd
from sktime.forecasting.model_selection import CutoffSplitter
from sktime.performance_metrics.forecasting import MeanSquaredScaledError

synthetic_data = pd.Series(
    10 + np.random.lognormal(5, 1, 365),
    index = pd.PeriodIndex(pd.date_range(start="2020-01-01", freq="D", periods=365), freq="D"), 
)

custom_cutoffs = np.array([
    len(synthetic_data[synthetic_data.index < date]) 
    for date in 
    [
        "2020-07-01",
        "2020-08-01", 
        "2020-09-01", 
        "2020-10-01", 
        "2020-11-01", 
        "2020-12-01",
    ]
])

# create the CV object
CV = CutoffSplitter(
    # custom cuttoffs
    cutoffs=custom_cutoffs, 
    # 3 step ahead forecast
    fh=np.arange(3)+1,
    # 5 years of training data
    window_length=6*30,
)

evaluate(
    forecaster=AutoETS(),
    y=synthetic_data,
    cv=CV,
    strategy="refit",
    return_data=True,
    scoring=MeanSquaredScaledError(square_root=True),
)

>>>  KeyError: "None of [PeriodIndex(['2020-08-02', '2020-08-03', '2020-08-04'], dtype='period[D]')] are in the [index]"
```

**Expected behavior**
evaluate should run smoothly for all cv strategies.
BaseForecaster._y should be updated based on the latest training data. 
I suggest we remove the combine_first logic from  [_update_y_X](https://github.com/alan-turing-institute/sktime/blob/922cd71d0d82d849025a080be826cf1c3c4777e5/sktime/forecasting/base/_base.py#L804), but I'm not sure if it serves a different purpose elsewhere.

Otherwise a deepcopy(forecaster) in evaluate will solve the problem too.

ltsaprounis:
#### Reference Issues/PRs
Fixes #1593 

When re-fitting (as opposed to updating) forecasting models self._y will be updated to the latest training data used.
This fix adds an additional check in update_y_X. If the forecaster is fitted it means that this is an update call (in fit the first line of code should be self._is_fitted = False). If it's fitted we use the combine first, if it's not fitted we overwrite _y with y. 

- [x] Does combine_first have a different purpose that this fix will break (maybe when update is used).   (It does) 

##### For all contributions


ltsaprounis:
The new test fails on ColumnEnsembleForecaster because f._y and y_train[3:] have different shapes. 
(47, 1) and (47,) respectively. Should I fix this in the test and make it flexible for this type of shape issues OR fix it in the ColumnEnsembleForecaster instead?

ltsaprounis:
Went for the "make the test flexible on this type of shape differences" option.

fkiraly (changes requested):
Yes, I think this is an actual issue.

This should be quickly fixed short-term, and we want to have a more principled fix long-term.

For the short-term the idea is good, but few issues to fix before merge:

* the proposed fix works only for `pandas`, not for `numpy`.
* lines 806ff already check whether `_y` exists or not. The existing check is problematic however, since `_y` can exist already in the case `fit` is called a second time. So instead of adding another `if` in 814, I would simply suggest to add a third condition `or not self._is_fitted` in 806. This would also make the fix work for `numpy`.
* the test is good, as a short-term fix. Long-term, we may want to test whether calling `fit` twice, with different arguments, results in the second call equivalent to the first call. The current `test_fit_idempotent` does not capture this because both `fit` calls use the same arguments, so the problem that `_y` is not overwritten is not caught, because it would be overwritten by the same argument if it were.

ltsaprounis:
> * the proposed fix works only for `pandas`, not for `numpy`.
> * lines 806ff already check whether `_y` exists or not. The existing check is problematic however, since `_y` can exist already in the case `fit` is called a second time. So instead of adding another `if` in 814, I would simply suggest to add a third condition `or not self._is_fitted` in 806. This would also make the fix work for `numpy`.

Thanks - that makes sense, the previous one was lousy. 
As you said - the test will be redundant when  #1613 is fixed, but useful in the meantime.

fkiraly (approved):
Thanks for the update, @ltsaprounis!

As you are aware, there is some intersection with #1613, we need to untangle this in #1613. I would appreciate if you watch #1613 and help untangle merge conflicts and updates, where applicable. For instance, I'd like to avoid the `[3:]` and instead use `_make_series` twice, but we can solve it in #1613.

PR #1600, opened 2021-11-09
Fixes #1234

Carlosbogo:
Removes unused _assertions method in signatures module.

### DIFF
--- a/sktime/transformations/panel/signature_based/_signature_method.py
+++ b/sktime/transformations/panel/signature_based/_signature_method.py
@@ -1,14 +1,15 @@
 # -*- coding: utf-8 -*-
 from sklearn.pipeline import Pipeline
+
 from sktime.transformations.base import _PanelToTabularTransformer
-from sktime.transformations.panel.signature_based._compute import (
-    _WindowSignatureTransform,
-)
 from sktime.transformations.panel.signature_based._augmentations import (
     _make_augmentation_pipeline,
 )
 from sktime.transformations.panel.signature_based._checks import (
     _handle_sktime_signatures,
+)
+from sktime.transformations.panel.signature_based._compute import (
+    _WindowSignatureTransform,
 )
 
 
@@ -64,7 +65,7 @@ class SignatureTransformer(_PanelToTabularTransformer):
         self.setup_feature_pipeline()
 
     def setup_feature_pipeline(self):
-        """Sets up the signature method as an sklearn pipeline."""
+        """Set up the signature method as an sklearn pipeline."""
         augmentation_step = _make_augmentation_pipeline(self.augmentation_list)
         transform_step = _WindowSignatureTransform(
             window_name=self.window_name,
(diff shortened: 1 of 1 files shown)


fkiraly:
Thanks a lot, @Carlosbogo!

Now you still need to pass the code formatting and linting checks (click "details" on the failing checks to see what goes wrong)

Carlosbogo:
Hi, I'm trying to solve the problems mentioned in the checks, but I'm not sure about some of them:

The test "build-and-test / code-quality (pull_request)" is giving the errors:

```
sktime/transformations/panel/signature_based/_signature_method.py:68 in public method `setup_feature_pipeline`:
        D401: First line should be in imperative mood (perhaps 'Set', not 'Sets')
sktime/transformations/panel/signature_based/_signature_method.py:89 in public method `fit`:
        D102: Missing docstring in public method
sktime/transformations/panel/signature_based/_signature_method.py:95 in public method `transform`:
        D102: Missing docstring in public method
```

Those are all things I haven't changed and I don't know whether I did something wrong or there are problems with the changes I made that affect those sections.

chrisholder:
Hi @Carlosbogo thanks for contributing! So these errors are coming from docstrings of the other methods in the file. Even though you haven't touched these methods, recently we've updated our docstring checks and it would appear this file hasn't been updated since, and therefore you're getting error flags. So lets work through them:

```
sktime/transformations/panel/signature_based/_signature_method.py:68 in public method `setup_feature_pipeline`:
        D401: First line should be in imperative mood (perhaps 'Set', not 'Sets')
```

This error is about the first line of a method docstring. The first line should be an imperative sentence meaning explain in one what it does exactly (without plurals). If we check the line the error is referring to (68) we find this:

```
    def setup_feature_pipeline(self):
        """Sets up the signature method as an sklearn pipeline."""
```
As you can see the words 'Sets' is used which is a plural. We can fix this by removing the plural like so:

```
    def setup_feature_pipeline(self):
        """Set up the signature method as an sklearn pipeline."""
```

The other two errors are complaining about missing docstrings on public methods (meaning their names don't start with _):

```
sktime/transformations/panel/signature_based/_signature_method.py:89 in public method `fit`:
        D102: Missing docstring in public method
sktime/transformations/panel/signature_based/_signature_method.py:95 in public method `transform`:
        D102: Missing docstring in public method
```
These errors are saying no docstring exists for either of the two public methods 'fit' (line 89) and 'transform' (line 95). 

```
@_handle_sktime_signatures(check_fitted=False)
def fit(self, data, labels=None):
    self.signature_method.fit(data, labels)
    self._is_fitted = True
    return self

@_handle_sktime_signatures(check_fitted=True)
def transform(self, data, labels=None):
    return self.signature_method.transform(data)
```

As you can see these are two public methods (they don't start with _) and therefore they need docstrings and don't currently have them. Im not 100% familar with everything going on in the transformer module or specifically here but we add a docstring that looks something like this (do change anything you see wrong here I've just guessed at the parameters types and what the method does):


```
@_handle_sktime_signatures(check_fitted=False)
def fit(self, data, labels=None):
    """Fit to data, then transform it.

    Parameters
    ----------
    data: pd.Dataframe or np.ndarray (3d array)
        Data to transform.
    labels: np.ndarray (1d array) or pd.series or list
        Labels for the data.

    Returns
    -------
    pd.Dataframe or np.ndarray or pd.series
        Transformed data.
    """
    self.signature_method.fit(data, labels)
    self._is_fitted = True
    return self


@_handle_sktime_signatures(check_fitted=True)
def transform(self, data, labels=None):
    """Transform the class from the signature method.

    Parameters
    ----------
    data: pd.Dataframe or np.ndarray (3d array)
        Data to transform.
    labels: np.ndarray (1d array) or pd.series or list
        Labels for the data.

    Returns
    -------
    pd.Dataframe or np.ndarray or pd.series
        Transformed data.
    """
    return self.signature_method.transform(data)
```

I've completly guessed at the parameters types and the return type. Please check they are correct. The above is just intended as an example for docstring commenting here. @fkiraly can check the types?

As a final note you'll find with sktime a lot of your errors in prs (especially early on) are going to be failing to conform to our docstring best practises. However if you follow the contributers guide and install pre-commit here: https://www.sktime.org/en/v0.5.1/contributing.html#:~:text=common%20coding%20style.-,Code%20quality%20checks,-To%20check%20if it will save you a lot of grief. Essentially using pre-commit will lint and check your docstrings everytime you commit anything. This means you can fix the issues as you write them rather than all at the end when you pr.

fkiraly (approved):
Thanks, straightforward.

Thanks for the docstrings too.
I think they are not entirely accurate, currently signature cannot handle `pandas`.
That's a mistake with the transformer (it should be compliant with the interface you specified), and it's addressed in the transformers refactor #1365 though, so we can leave it like this temporarily.

PR #1604, opened 2021-11-11
convert 2D to 3D rather than throw exception

TonyBagnall:
#### Reference Issues/PRs

https://github.com/alan-turing-institute/sktime/issues/1596

This is a very simple fix to allow 2D numpy input to represent univariate, equal length time series, where each row is a series. If a 2D is passed, it is just converted in check_X to 3D. All the classifiers that only accept univariate equal length simply convert any input into a 2D numpy anyway, so this is the easiest way to allow for 2D input without having to change a large number of classifiers. I believe the conversion is a constant time operation i.e. it does not clone the data, but just changes the iteration (although feel free to disabuse me). I had meant to bring this in as part of a bigger change to the base classifier, https://github.com/alan-turing-institute/sktime/pull/1594 but I have run into some testing errors that are beyond my competence to solve, at least without pain. So this is a compromise. 

No.

Assuming the testing doesn't break, the issue is really whether we want to accept 2D input for univariate, equal length series, rather than insist on 3D. I feel quite strongly we should, it greatly simplifies the getting started use case, but if there are general objections we can ditch the plan.

### DIFF
--- a/sktime/utils/validation/panel.py
+++ b/sktime/utils/validation/panel.py
@@ -12,9 +12,11 @@ import numpy as np
 import pandas as pd
 from sklearn.utils.validation import check_consistent_length
 
-from sktime.datatypes._panel._convert import from_3d_numpy_to_nested
-from sktime.datatypes._panel._convert import from_nested_to_3d_numpy
 from sktime.datatypes._panel._check import is_nested_dataframe
+from sktime.datatypes._panel._convert import (
+    from_3d_numpy_to_nested,
+    from_nested_to_3d_numpy,
+)
 
 VALID_X_TYPES = (pd.DataFrame, np.ndarray)  # nested pd.DataFrame and 3d np.array
 VALID_Y_TYPES = (pd.Series, np.ndarray)  # 1-d vector
--- a/sktime/utils/validation/panel.py
+++ b/sktime/utils/validation/panel.py
@@ -76,7 +76,7 @@ def check_X(
             X = X.reshape(X.shape[0], 1, X.shape[1])
         elif X.ndim == 1 or X.ndim > 3:
             raise ValueError(
-                f"If passed as a np.array, X must be a2 or 3-dimensional "
+                f"If passed as a np.array, X must be a 2 or 3-dimensional "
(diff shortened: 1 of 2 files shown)


sktime/utils/validation/panel.py
```diff
@@ -70,9 +72,11 @@ def check_X(
     # check first if we have the right number of dimensions, otherwise we
     # may not be able to get the shape of the second dimension below
     if isinstance(X, np.ndarray):
-        if not X.ndim == 3:
+        if X.ndim == 2:
+            X = X.reshape(X.shape[0], 1, X.shape[1])
+        elif X.ndim == 1 or X.ndim > 3:
             raise ValueError(
-                f"If passed as a np.array, X must be a 3-dimensional "
+                f"If passed as a np.array, X must be a2 or 3-dimensional "
```
chrisholder, line 79: If passed as a np.array, X must be a 2-dimensional or 3-dimensional array

sktime/utils/validation/panel.py
```diff
@@ -70,9 +72,11 @@ def check_X(
     # check first if we have the right number of dimensions, otherwise we
     # may not be able to get the shape of the second dimension below
     if isinstance(X, np.ndarray):
-        if not X.ndim == 3:
+        if X.ndim == 2:
```
chrisholder, line 75: If we're reshaping 2d arrays now doesn't it just make sense to reshape 1d arrays as well?

i.e.
if X.ndim == 1:
    X = X.reshape(X.shape[0], 1, 1)

TonyBagnall: hmmm, maybe, that would make sense for predict and predict_proba, but not for fit, doesnt make sense to build a classifier on one case. But lets stick to this for now :)

chrisholder (approved):

MatthewMiddlehurst (approved):
There is potential for confusion with the single series 2d numpy array, but I believe they can co-exist for classification at least. 

There arent really any cases where you would want to input a single series for fit, could potentially cause some problems if we decide to start accepting single series for predict.

PR #1610, opened 2021-11-11
load from UCR fix

### LINKED ISSUE #1603: [BUG] problem with strings input for target variable
**Describe the bug**
using the sklearn train test split breaks HC2, because the numba code expects np.str not Series strings.
 
**To Reproduce**
    # breaks
    X, y = load_UCR_UEA_dataset("ItalyPowerDemand", return_X_y=True)
    X_train, X_test, y_train, y_test = train_test_split(X, y)
    hc2 = HIVECOTEV2()
    hc2.fit(X_train, y_train)

**Expected behaviour**

it should work with train_test_split. The simplest fix for this is to test for Series strings in the base classifier and always convert to numpy.str. I'll work this into https://github.com/alan-turing-institute/sktime/pull/1594

TonyBagnall:
#### Reference Issues/PRs
Fixes #1603 

A few minor fixes
1. corrects a bug in contracted rotation forest: this was only a problem with small contracts and should not effect testing.
2. casts the target variable to numpy strings rather than Series strings when loading the concatenated train and test using load_UCR_UEA_dataset (#1603). This is needed to fix the problem in STC with numba, and it is more consistent, since these are returned when either train or test is loaded
3. removes the deprecation warning is data.io. tbh there is a lot in here that could be improved, so I think a gradual move over, with duplication is preferable
4. change the default return_X_y in load_UCR_UEA_dataset to True. I think its a legacy of Weka conversion to append the target variable, This could conceivably break something somewhere, although I cannot think of a reason to do it and the variable can still be set to False so its a very edgy edge case.

No

I think (4) is the most controversial?

### DIFF
--- a/sktime/contrib/vector_classifiers/_rotation_forest.py
+++ b/sktime/contrib/vector_classifiers/_rotation_forest.py
@@ -229,8 +229,8 @@ class RotationForest(BaseEstimator):
                 estimators, pcas, groups, transformed_data = zip(*fit)
 
                 self.estimators_ += estimators
-                self._pcas = pcas
-                self._groups = groups
+                self._pcas += pcas
+                self._groups += groups
                 self.transformed_data += transformed_data
 
                 self._n_estimators += self._n_jobs
--- a/sktime/datasets/_data_io.py
+++ b/sktime/datasets/_data_io.py
@@ -122,7 +122,7 @@ def _list_downloaded_datasets(extract_path):
     return datasets
 
 
-def load_UCR_UEA_dataset(name, split=None, return_X_y=False, extract_path=None):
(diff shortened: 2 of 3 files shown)


MatthewMiddlehurst:
Unless there are no plans to move any of the data_io functions in the next few months, it may be best to keep them deprecated?
If it's undecided whether they are going to be moved at all that's fine.

I support (4), but am aware that it may impact anyone using the current default as its public. I doubt it would be a significant fix for anyone who does so though.

TonyBagnall:
Yeah good point, I'll put them back in

MatthewMiddlehurst (approved):

chrisholder (approved):

PR #1615, opened 2021-11-12
Ets strictly pos mul models

### LINKED ISSUE #1387: [ENH] AutoETS to ignore multiplicative components when time series is not strictly positive
If a time series is not strictly positive, ETS models (from statsmodels) with multiplicative components will raise the following error:
```
ValueError: endog must be strictly positive when usingmultiplicative error, trend or seasonal components.
```

Currently AutoETS will raise the same error because it tries all the available ETS models. Ideally AutoETS will skip the models that result in this error and will try all the other ones. 

One solution would be to add a try/except in the [_fit](https://github.com/alan-turing-institute/sktime/blob/2d1eaa22854beecdbe63fcd076aa486140f35193/sktime/forecasting/ets.py#L278) method. The other solution I can think of is to check whether the time series is strictly positive and if it's not remove the multiplicative components from [_iter](https://github.com/alan-turing-institute/sktime/blob/2d1eaa22854beecdbe63fcd076aa486140f35193/sktime/forecasting/ets.py#L258).

**The following code reproduces the error:** 
```python
import pandas as pd
import numpy as np

ts = pd.Series(
    10*np.sin(np.array(range(0,264))/10)+10, 
    pd.date_range("2017-01-01", periods=264,  freq="W")
)
ts[ts<1]=0
forecaster = AutoETS(auto=True, sp=52, n_jobs=-1, ignore_inf_ic=True)
plot_series(ts)
forecaster.fit(ts)
```
image

chernika158:
#### Reference Issues/PRs

Fixes #1387

If a time series is not strictly positive, ETS models with multiplicative components will raise the error. I've add conditions to omit multiplicative models for not strictly positive models.

No

##### For all contributions

##### For new estimators

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -986,6 +986,15 @@
       ]
     },
     {
+      "login": "chernika158",
+      "name": "Galina Chernikova",
+      "avatar_url": "https://avatars.githubusercontent.com/u/43787741?s=400&v=4",
+      "profile": "https://github.com/chernika158",
+      "contributions": [
+        "code"
+      ]
+    },
+    {
       "login": "GuzalBulatova",
       "name": "Guzal Bulatova",
       "avatar_url": "https://avatars.githubusercontent.com/GuzalBulatova",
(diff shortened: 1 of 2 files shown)


fkiraly (approved):
makes sense - thanks a lot for your contribution, @chernika158!

Issue #1613, opened 2021-11-12
[ENH] test for two `fit` calls with different argument


fkiraly:
The current test `test_fit_idempotent` did not catch the bug in https://github.com/alan-turing-institute/sktime/pull/1595 because the test checks two calls to `fit` with the same arguments.
This fails to catch the case where, for instance, the resulting estimator behaves like after the first, but not the *second* `fit` call.

A better test would call `fit` twice, but with different arguments.
The resulting estimator should behave like the estimator fitted only by the *second* `fit` call.

This should be a new test, or a change to `test_fit_idempotent`.

Carlosbogo:
Hi, I would be glad to help with this. Is there anyone already working on it?

fkiraly:
I don't think so, please feel free to go ahead! You should be able to use the `_make_args` helper functions to get this done easily.

@ltsaprounis, FYI

PR #1618, opened 2021-11-15
Solves #1613

Carlosbogo:
Changed `test_fit_idempotent` so that the second call to the `fit` method is done using different arguments.

### DIFF
--- a/sktime/tests/test_all_estimators.py
+++ b/sktime/tests/test_all_estimators.py
@@ -381,7 +381,8 @@ def test_fit_idempotent(estimator_instance):
 
     # Fit again
     set_random_state(estimator)
-    estimator.fit(*fit_args)
+    new_fit_args = _make_args(estimator, "fit")
+    estimator.fit(*new_fit_args)
 
     for method in NON_STATE_CHANGING_METHODS:
         if hasattr(estimator, method):


fkiraly:
@Carlosbogo, thanks!

Question, why is this failing? Because some existing estimators are not compliant, or is there a problem with the test?

fkiraly (changes requested):
Change request:
* we should have two tests: once fitting twice with same arguments, once fitting twice with different arguments
* Using `_make_args` is good, I would avoid using subsetting that @ltsaprounis suggets (since it assumes a specific format of the return of `_make_args` which we can't always guarantee)
* however, if we use `_make_args` twice, we must somehow guarantee that the second call results in *different* results. I don't think that is currently guaranteed! `new_fit_args`, in many cases, may just be the same thing. I would recommend, think about how difference can be guaranteed (perhaps just asking for different sized arguments) and add a comment to that effect in the test that explains why this is different.

ltsaprounis:
> Question, why is this failing? Because some existing estimators are not compliant, or is there a problem with the test?

Existing operators are definitely not compliant!  
So if the test is correct it should fail here. 

#1595 fixes one of the problems (the one with self._y), but I'm not sure if that's the only problem.

fkiraly:
@Carlosbogo, @ltsaprounis, I've now merged #1595.
Can we ensure there are, at the end, two separate tests (one for "same args" and one for "different args")?
It may also be that some estimators will fail - once we are sure the tests are correct, we need to list these failing estimators in an issue, and only merge the tests once those estimators pass the new tests.

ltsaprounis:
@Carlosbogo - update the branch and merge main here. 
There's a small chance that the only issue was with _y and the tests pass now.

fkiraly:
I've updated from `main` to see what happens.

PR #1619, opened 2021-11-15
Experiments tidy

TonyBagnall:

simple change to benchmarking/experiments.py to move set classifier and set clusterer into contrib. The problem they introduce (apart from ugly code and general poor usability) is to require all soft dependencies to run experiments.py. This is a pain. Now the actual classifier/clusterer has to be passed, and the string name is only used to set the results location.

### DIFF
--- a/sktime/benchmarking/experiments.py
+++ b/sktime/benchmarking/experiments.py
@@ -135,10 +135,10 @@ def run_clustering_experiment(
 def load_and_run_clustering_experiment(
     problem_path,
     results_path,
-    cls_name,
     dataset,
     clusterer,
     resample_id=0,
+    cls_name=None,
     overwrite=False,
     format=".ts",
     train_file=False,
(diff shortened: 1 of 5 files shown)


chrisholder (dismissed):

MatthewMiddlehurst:
Good changes, pretty sure only we use the set_classifier functionality for cluster stuff but good to have it available in contrib for others.

With an actual classifier now required, I think cls_name could be changed to an optional parameter. It is still useful to have, but maybe set the default to none and use `type(transform).__name__` if nothing is provided.

TonyBagnall:
> Good changes, pretty sure only we use the set_classifier functionality for cluster stuff but good to have it available in contrib for others.
> 
> With an actual classifier now required, I think cls_name could be changed to an optional parameter. It is still useful to have, but maybe set the default to none and use `type(classifier).__name__` if nothing is provided.

good idea

MatthewMiddlehurst (dismissed):

TonyBagnall:
suspect apveyor is broken again

Issue #1563, opened 2021-10-29
[BUG] Numba needs NumPy 1.20 or less, but `sktime` installation does not ensure this on all operating systems


srinuti:
**Describe the bug**

**To Reproduce**

---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-63-9c74fc495c79> in <module>
----> 1 all_estimators("forecaster", as_dataframe=True)

C:\ProgramData\Anaconda3\envs\tensorflow_env\lib\site-packages\sktime\registry\_lookup.py in all_estimators(estimator_types, filter_tags, exclude_estimators, return_names, as_dataframe)
    140 
    141             try:
--> 142                 module = import_module(module_name)
    143                 classes = inspect.getmembers(module, inspect.isclass)
    144 

C:\ProgramData\Anaconda3\envs\tensorflow_env\lib\importlib\__init__.py in import_module(name, package)
    125                 break
    126             level += 1
--> 127     return _bootstrap._gcd_import(name[level:], package, level)
    128 
    129 

C:\ProgramData\Anaconda3\envs\tensorflow_env\lib\importlib\_bootstrap.py in _gcd_import(name, package, level)

C:\ProgramData\Anaconda3\envs\tensorflow_env\lib\importlib\_bootstrap.py in _find_and_load(name, import_)

C:\ProgramData\Anaconda3\envs\tensorflow_env\lib\importlib\_bootstrap.py in _find_and_load_unlocked(name, import_)

C:\ProgramData\Anaconda3\envs\tensorflow_env\lib\importlib\_bootstrap.py in _load_unlocked(spec)

C:\ProgramData\Anaconda3\envs\tensorflow_env\lib\importlib\_bootstrap_external.py in exec_module(self, module)

C:\ProgramData\Anaconda3\envs\tensorflow_env\lib\importlib\_bootstrap.py in _call_with_frames_removed(f, *args, **kwds)

C:\ProgramData\Anaconda3\envs\tensorflow_env\lib\site-packages\sktime\annotation\clasp.py in <module>
     25 import pandas as pd
     26 
---> 27 from sktime.transformations.series.clasp import ClaSPTransformer
     28 from sktime.utils.validation.series import check_series
     29 

C:\ProgramData\Anaconda3\envs\tensorflow_env\lib\site-packages\sktime\transformations\series\clasp.py in <module>
     22 # import numpy.fft as fft
     23 import pandas as pd
---> 24 from numba import njit
[...]

PR #1623, opened 2021-11-18
Numba numpy

TonyBagnall:
#### Reference Issues/PRs
fixes  https://github.com/alan-turing-institute/sktime/issues/1563
using the six file change given in 
https://github.com/alan-turing-institute/sktime/issues/1482

note that several errors are still raised on the install related to numpy versions, e.g. 

image

we have tested it on two different windows machines, it all seems to work on code that does not need the soft dependencies. 

This also relates to 
https://github.com/alan-turing-institute/sktime/issues/1372
and the soft dependencies, as there are problems with fbprophet,

pip install -e .[all_extras]

does not work on its own. Jay will post instructions below for dealing with that. I can now run pytest locally!

We have not added esig to soft deps, but can do so if its wanted, 

https://github.com/alan-turing-institute/sktime/issues/1373

Fixes a clash between numpy and numba by simply fixing numpy to 1.20.0 using method described in #1563  

conda create --name sktime python=3.8
conda activate sktime
pip install -e .

now installs a working environment for the hard dependencies.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -65,11 +65,7 @@ install:
 
   # Install requirements from inside conda environment
   - cmd: activate testenv
-  - cmd: conda install -c conda-forge "matplotlib>=3.3.2" "seaborn>=0.11.0"
-  - "tsfresh>=0.17.0" "hcrystalball>=0.1.9" "stumpy>=1.5.1"
-  - "numba==0.53" "numpy==1.20.3" "pandas>=1.1.0,<1.2" "statsmodels==0.12.1"
-  - "scikit-learn>=0.23.0" "scikit-posthocs>=0.6.5" "pystan>=2.14,<3.0"
-  - "deprecated>=1.2.13" "fbprophet>=0.7.1"
+  - cmd: conda install -c conda-forge "matplotlib>=3.3.2" "seaborn>=0.11.0" "tsfresh>=0.17.0" "hcrystalball>=0.1.9" "stumpy>=1.5.1" "numba==0.53" "numpy==1.20.3" "pandas>=1.1.0,<1.2" "statsmodels==0.12.1" "scikit-learn>=0.23.0" "scikit-posthocs>=0.6.5" "pystan>=2.14,<3.0" "deprecated>=1.2.13" "fbprophet>=0.7.1"
 
   - cmd: pip install -r %REQUIREMENTS%
 
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -66,7 +66,10 @@ install:
   # Install requirements from inside conda environment
   - cmd: activate testenv
   - cmd: conda install -c conda-forge "matplotlib>=3.3.2" "seaborn>=0.11.0"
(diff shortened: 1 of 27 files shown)


TonyBagnall:
ah pycharm has inserted  line breaks ffs

TonyBagnall:
so AppVeyor.yml did not like me fixing numpy == 1.19.3. I have reverted to >=1.19.3 in the .yml file, not sure if that will effect the rest. Could try <=1.19.3

jasonlines:
to confirm, I've used this branch and installed with the following instructions:

```
conda create --name sktime python=3.8
conda activate sktime
pip install -e .[all_extras]
pip install pytest
pip install esig
conda install -c conda-forge fbprophet
pip install scipy --force-reinstall
pip install numpy==1.19.3 --force-reinstal
```

doing the above installed everything successfully for me, including running tests with Numba and cythonised code. A couple of notes, however:
- `pip install -e .[all_extras]` generates an error when installing `fbprophet` (as expected), hence the conda install further down the list of instructions 
- `conda install -c conda-forge fbprohpet` does successfully install the package, but it upgrades NumPy above the maximum requirement of Numba (hence why `1.19.3` is installed at the end.. but more on that in a second)
- there is a library issue with SciPy when following these instructions; I'm not quite clear on why that happens, but reinstalling SciPy at the end rectifies this issue (though doing so also upgrades NumPy, hence why the downgrade of NumPy comes last and after)

TonyBagnall:
I thought it was a type, but weirdly --force-reinstal and --force-reinstall both seem to work

.appveyor.yml
```diff
@@ -65,7 +65,7 @@ install:
 
   # Install requirements from inside conda environment
   - cmd: activate testenv
-  - cmd: conda install -c conda-forge "matplotlib>=3.3.2" "seaborn>=0.11.0" "tsfresh>=0.17.0" "hcrystalball>=0.1.9" "stumpy>=1.5.1" "fbprophet>=0.7.1" "numba==0.53" "numpy>=1.19.3" "pandas>=1.1.0,<1.2" "statsmodels==0.12.1" "scikit-learn>=0.23.0" "scikit-posthocs>=0.6.5" "pystan>=2.14,<3.0" "deprecated>=1.2.13"
+  - cmd: conda install -c conda-forge "matplotlib>=3.3.2" "seaborn>=0.11.0" "tsfresh>=0.17.0" "hcrystalball>=0.1.9" "stumpy>=1.5.1" "numba==0.53" "numpy>=1.19.3" "pandas>=1.1.0,<1.2" "statsmodels==0.12.1" "scikit-learn>=0.23.0" "scikit-posthocs>=0.6.5" "pystan>=2.14,<3.0" "deprecated>=1.2.13" "fbprophet>=0.7.1"
```
freddyaboulton, line 68: Maybe we should switch over to `prophet` instead of `fbprophet` ? 

I believe `fbprophet` is quite old and they list pystan >= 2 in their requirements, which will pystan 3 when prophet officially only supports pystan 2. 

I suspect that may be causing some of our woes.

TonyBagnall: If switching tidies up the windows process I would be strongly in favour! fbprophet throws all sorts of errors even when building is successful.

freddyaboulton (approved):
Looks good to me @TonyBagnall . I will file an issue to switch over to prophet to see if that helps get rid of the errors.

freddyaboulton:
Filed here: https://github.com/alan-turing-institute/sktime/issues/1631

PR #1633, opened 2021-11-20
Updated classification base class typing.

chrisholder:

Updated the types in the classification base class for np.array to the type np.ndarray (np.array is a method np.ndarray is the type of a numpy array).

##### For all contributions

##### For new estimators


fkiraly (approved):
Yes, `array` is a function, not a type.
Didn't notice this, but obvious fix.

PR #1637, opened 2021-11-23
MiniRocket and MultiRocket as options for RocketClassifier

MatthewMiddlehurst:
Solves #813.

Adds both MiniRocket and MultiRocket as transformer options for both RocketClassifier and Arsenal.

Changed MiniRocket num_features to num_kernels to match with Rocket and MultiRocket.

Moved the rocket transformer tests into the rocket sub-package.


TonyBagnall (approved):

jasonlines (approved):
Looks good to me :) Code looks great and if you want to be REALLY picky then the find and replace of "ROCKET" with "Rocket" looks like it's also changed the term in the citation within the comments (e.g. in _rocket_classifier.py the paper's title now includes "Rocket" instead of "ROCKET" as published). Approving anyway though as that would be extremely harsh :)

PR #1638, opened 2021-11-23
fix parameter mismatch in ShapeDTW

TonyBagnall:
simple bug fix in parameter names so that ShapeDTW can be removed from the EXCCLUDED_ESTIMATORS list

### DIFF
--- a/sktime/classification/distance_based/_shape_dtw.py
+++ b/sktime/classification/distance_based/_shape_dtw.py
@@ -114,13 +114,13 @@ class ShapeDTW(BaseClassifier):
 
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


TonyBagnall:
not as simple as I thought, and as no doubt vincent found when implementing. The problem is that metric_params defaults to None, but then if not set it is always made into an empty dictionary in fit. The solution could be to wrap a lot of the code around if not None,  or just to reset to None in a nasty hack. Obvs went for the latter

TonyBagnall:
this is so tiny, I'll just put it in otherwise it will all get log jammed

PR #1646, opened 2021-11-25
lower bound pandas to 1.2.0

TonyBagnall:
#### Reference Issues/PRs
resolves difference in testing described in 
https://github.com/alan-turing-institute/sktime/issues/1642
(but not the actual bug, see https://github.com/alan-turing-institute/sktime/pull/1644 for that)

and hopefully fixes https://github.com/alan-turing-institute/sktime/issues/1645

uses this 
https://github.com/alan-turing-institute/sktime/issues/1482

this fixes pandas to be at least 1.2. It was fixed to 1.1.5 in testing, and this causes testing anomalies

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -65,7 +65,7 @@ install:
 
   # Install requirements from inside conda environment
   - cmd: activate testenv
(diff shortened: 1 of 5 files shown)


fkiraly:
hmmm, as discussed on slack, this would lose us python 3.6 support, since 1.2.0 no longer supports 3.6?
I don't think we should lose python 3.6.

PR #1648, opened 2021-11-25
[BUG] Fixes to registry look-up, test suite for registry look-up

### LINKED ISSUE #1647: [BUG] all_tags returns empty list with estimator type
**Describe the bug**
`all_tags` returns empty list when passing estimator type. I suspect it was just copied over from `all_estimators` and can just be removed for now. Otherwise the docstring needs to be updated too. 

**To Reproduce**
```python
from sktime.registry import all_tags
all_tags(estimator_types="forecaster")
```

**Expected behavior**
Get list of tags applicable for given estimator type.

fkiraly:
This PR fixes #1647 and introduces tests for `all_estimators` and `all_tags`.

It further fixes the following, additional bugs discovered in the process:
* the `symmetric` tag had a non-existent scitype in it
* `_check_estimator_types` had side effects on its argument if it was a list

This PR also introduces tests for `all_estimators`, `all_tags`:
* return object being of correct types
* at least one element being returned in the list, when subsetting by scitypes


mloning (approved):
The bug in #1647 seems to be fixed. I've not checked the tests. I added a missing init file, so that the tests are being run on the CI.

Issue #1649, opened 2021-11-26
[BUG] forecaster fit/predict interface


mloning:
**Describe the bug**
I think the current behaviour of the forecasting API isn't very consistent. My preferred specification is described here: https://github.com/sktime/community-org/blob/main/meetups/previous_meetings/20210909-meetup.md

**To Reproduce**
```python
import numpy as np
from sktime.forecasting.exp_smoothing import ExponentialSmoothing
from sktime.forecasting.naive import NaiveForecaster
from sktime.forecasting.model_evaluation import evaluate
from sktime.forecasting.model_selection import SlidingWindowSplitter
from sktime.utils.plotting import plot_series

# fails with: TypeError: No valid mtype could be identified
y = np.random.normal(size=20)
forecaster = ExponentialSmoothing()
forecaster.fit(y)

# works but predict returns np.ndarray
y = np.random.normal(size=(20, 1))
forecaster = NaiveForecaster()
forecaster.fit(y)
y_pred = forecaster.predict(fh=1)
type(y_pred)
# numpy.ndarray

# but then other functionality breaks
# fails with TypeError
plot_series(y_pred)

# fails with TypeError: No valid mtype could be identified
y = np.random.normal(size=20)
forecaster = NaiveForecaster()
forecaster.fit(y)

# fails with ValueError: Data must be 1-dimensional
y = np.random.normal(size=(20, 1))
forecaster = ExponentialSmoothing()
forecaster.fit(y)

# fails with ValueError: Index data must be 1-dimensional
y = np.random.normal(size=(20, 1))
forecaster = NaiveForecaster()
cv = SlidingWindowSplitter(fh=1)
evaluate(forecaster, cv, y)
```

**Expected behavior**
Consistent handling of numpy arrays. User-friendly error messages. Preferably also consistent output types.

fkiraly:
@mloning, some of this was one of the things fixed in the transformers refactor, #1365.
Originally, we were going to allow only 2D numpy, but decided to relax to 1D numpy
With it merged, the checks for 1D now work properly, here;s the relevant bullet point:

> relaxing np.ndarray mtype for Series so 1D numpy arrays are also allowed (and considered univariate series or vector). This is required since some tests use univariate numpy arrays as label vectors.

On top of this, there was another genuine bug (thanks for reporting!), namely:
[...]

PR #1650, opened 2021-11-26
[BUG] Fix `convert` and `_split` for Numpy 1D input

fkiraly:
This fixes buggy behaviour of `convert` and `_split` when a numpy 1D input is passed, see #1649.

It ensures that it is correctly coerced to the internally supported numpy 2D mtype.


sktime/datatypes/_series/_convert.py
```diff
@@ -142,9 +142,12 @@ def convert_np_to_MvS_as_Series(obj: np.ndarray, store=None) -> pd.DataFrame:
 def convert_np_to_UvS_as_Series(obj: np.ndarray, store=None) -> pd.Series:
 
     if not isinstance(obj, np.ndarray) and len(obj.shape) < 3:
-        raise TypeError("input must be a np.ndarray of dim 1 or 2")
+        raise TypeError("input must be a one-column np.ndarray of dim 1 or 2")
+
+    if len(obj.shape) == 2 and not obj.shape[1] == 1:
```
mloning, line 147:
```suggestion
    if obj.ndim == 2 and obj.shape[1] != 1:
```

sktime/datatypes/_series/_convert.py
```diff
@@ -142,9 +142,12 @@ def convert_np_to_MvS_as_Series(obj: np.ndarray, store=None) -> pd.DataFrame:
 def convert_np_to_UvS_as_Series(obj: np.ndarray, store=None) -> pd.Series:
 
     if not isinstance(obj, np.ndarray) and len(obj.shape) < 3:
```
mloning, line 144:
```suggestion
    if not isinstance(obj, np.ndarray) and obj.ndim < 3:
```

mloning: Why do we test for `isinstance(obj, np.ndarray)` here but not in the other if-condition below? Perhaps we should have an `isinstance` check at the beginning since that's what the docstring specifies? I'm not sure if this is meant to be a public function, but passing objects without a shape attribute will result in an AttributeError (e.g. passing a list).

I would find this more readable:

```python
if not isinstance(obj, np.ndarray):
    raise ValueError(...)

if obj.ndim > 2:
   raise ValueError(...)

if obj.ndim == 2 and obj.shape[1] > 1:
    raise ValueError(...)
```

fkiraly: This is a typo, the line should say:
`if not isinstance(obj, np.ndarray) or len(obj.shape) > 2:`

fkiraly: if it's `or`, the second statement will only be evaluated if the first statement is false, so it's fine

mloning: I'd prefer the three conditions above, more readable

sktime/datatypes/_series/_convert.py
fkiraly: yes, that's better

mloning (approved):

PR #1653, opened 2021-11-26
Fix `test_wilcoxon` compatibility between pandas versions

### LINKED ISSUE #1651: [BUG] benchmarking.tests.test_evaluator.test_wilcoxon passes/fails with different versions of pandas
**Describe the bug**

thanks to @immentel for identifying this issue previously identified here 
https://github.com/alan-turing-institute/sktime/pull/1198

the problem is caused by changes in pandas between versions 1.1.5 (used for the automatic repo tests) and versions >=1.2.0 (specifically tested with 1.3.4) that results in inconsistency in the ordering of the underlying data structure when directly converted to numpy. Related to https://github.com/alan-turing-institute/sktime/pull/1644 and https://github.com/alan-turing-institute/sktime/pull/1646.

```python
def test_wilcoxon():
    evaluator, metrics_by_strategy = evaluator_setup(score_function=accuracy_score)
    results = evaluator.wilcoxon_test().iloc[:, 2:].values
    # Fails pandas 1.3.4 passes 1.1.5
    expected = np.array([[0.5, 0], [0.5, 0], [0.5, 0]])
    # Works pandas 1.3.4 fails 1.1.5
    expected = np.array([[0, 0.5], [0, 0.5], [0, 0.5]])
    assert np.array_equal(results, expected)
```

*To Reproduce**
pip install pandas==1.1.5
set
expected = np.array([[0, 0.5], [0, 0.5], [0, 0.5]])
or 
pip install pandas==1.3.4
set
expected = np.array([[0.5, 0], [0.5, 0], [0.5, 0]])

**Expected behavior**
the ordering of the expected values need to be consistent. Based on conversations with @fkiraly, says 

the actual function still fulfills its contract, since the column names are correctly attributing the things that are compared, it's just that the data frame that is returned has permuted columns (with permuted correct index!) between versions. But if you convert to values, the information about the column index is removed, which results in a change in the numpy array, so as data frame, the two would be equal (if column names would be kept), but after converting to numpy  they no longer are. It's my suspicion"

I think he is correct, but have not tested it. @fkiraly suggests two possible solutions

results_df = evaluator.wilcoxon_test()
results_df_cols = evaluator.wilcoxon_test().columns
results = results_df[sort(results_df_cols)][:, 2:].values

or

results_df = evaluator.wilcoxon_test()
expected = pd.DataFrame(put the right thing here)
assert results_df.equals(expected)

**Additional context**

**Versions**
<details>

</details>

lmmentel:
#### Reference Issues/PRs

Fixes #1651 
See also: #1198 

Updated asset condition to take into account column reordering. 

no

Sorry for all the additional edits but `pre-commit` made me do it.

##### For all contributions

### DIFF
--- a/sktime/benchmarking/tests/test_evaluator.py
+++ b/sktime/benchmarking/tests/test_evaluator.py
@@ -1,14 +1,18 @@
 # -*- coding: utf-8 -*-
-from sktime.benchmarking.results import RAMResults
-from sktime.benchmarking.evaluation import Evaluator
-from sktime.benchmarking.metrics import PairwiseMetric
-from sklearn.metrics import accuracy_score
-from sktime.series_as_features.model_selection import PresplitFilesCV
+"""Test evaluator."""
+
 import numpy as np
 import pandas as pd
+from sklearn.metrics import accuracy_score
+
+from sktime.benchmarking.evaluation import Evaluator
+from sktime.benchmarking.metrics import PairwiseMetric
+from sktime.benchmarking.results import RAMResults
+from sktime.series_as_features.model_selection import PresplitFilesCV
 
 
 def dummy_results():
+    """Results that are dummy."""
     results = RAMResults()
     results.cv = PresplitFilesCV()
     results.save_predictions(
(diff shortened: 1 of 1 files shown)


fkiraly (approved):
Looks good, thanks!
Waiting for tests before merge.

PR #1656, opened 2021-11-27
Added meta-tuning example to docstring of ForecastingGridSearchCV

aiwalter:
Added meta-tuning example to docstring


fkiraly (changes requested):
Uh-oh, looks like the return is not what is expected. Fix that, I guess?

Might also be worth passing a smaller `fh`, or the printout gets too big.

aiwalter:
yes, I made a smaller `fh` and I forgot the `(...)` in the output placeholder

fkiraly (approved):
Yes, looks right now, thanks.

Issue #1404, opened 2021-09-10
[ENH] AutoEnsembleForeacaster should have a method to use inverse error as weights


aiwalter:
**Describe the solution you'd like**
`AutoEnsembleForeacaster` should be able to automatically calculate internally the error of the forecasters and use the inverse error as weights. Credits to @RNKuhns who came up with this idea :)

There should be a `method="inverse-error"` added.

AngelPone:
@aiwalter are you working on this? If not, I can help. While I suggest  to add this `inverse-error` method into `EnsembleForecaster` instead of `AutoEnsembleForecaster`, such as `EnsembleForecaster(forecasters, weights='inverse-error')`.  If added into `AutoEnsembleForecaster`, `method` and `regressor` may be confused.

aiwalter:
Hi @AngelPone we missed you :) especially for the HTS topic, but glad that you are back again :) I would put it into `AutoEnsembleForecaster`, we might have more such weight selection algos implemented there. Then we would have the regressor method and the inverse-error one. The `EnsembleForecaster` is for me more a vehicle for users to give custom weights.

AngelPone:
@aiwalter agree. I will work on this.

PR #1660, opened 2021-11-28
Add inv_var method into AutoEnsembleForecaster

AngelPone:
#### Reference Issues/PRs

See #1404

Implement `inv_var` method in `AutoEnsembleForecaster` that uses the inverse of variance of in-sample-fit errors as weights.  

There are some points that need to be discussed further:

* if `regressor` and `method` both `None`, which is the default. Now the default is `GradientBoostingRegressor`. 
* If they are provided both, raise a `ValueError`

##### For all contributions
- [ ]


AngelPone:
also, the docstring need to be refined.

sktime/forecasting/compose/_ensemble.py
```diff
@@ -135,38 +139,51 @@ def _fit(self, y, X=None, fh=None):
...
+            self.regressor_.fit(X=X_meta, y=y_test)
+
+            # check if regressor is a sklearn.Pipeline
+            if isinstance(self.regressor_, Pipeline):
+                # extract regressor from pipeline to access its attributes
+                self.weights_ = _get_weights(self.regressor_.steps[-1][1])
+            else:
+                self.weights_ = _get_weights(self.regressor_)
+            self._fit_forecasters(forecasters, y, X, fh)
+        elif self.method == "inv_var":
```
aiwalter, line 173: I would rename it to `inverse-variance`, pls use minus not underscore

sktime/forecasting/compose/_ensemble.py
```diff
@@ -135,38 +139,51 @@ def _fit(self, y, X=None, fh=None):
...
+            # get in-sample forecasts
+            self._fit_forecasters(forecasters, y, X, fh)
+            fh_insample = ForecastingHorizon(y.index, is_relative=False)
+            inv_var = np.array(
+                [
+                    1 / np.var(y - fit)
```
aiwalter, line 179: `fit` could be renamed to `y_pred_insample`? Or what else is it?

sktime/forecasting/compose/_ensemble.py
```diff
@@ -135,38 +139,51 @@ def _fit(self, y, X=None, fh=None):
...
+                    for fit in self._predict_forecasters(fh_insample, X)
+                ]
+            )
+            # standardize the inverse variance
+            self.weights_ = list(inv_var / np.sum(inv_var))
+        else:
+            raise NotImplementedError()
```
aiwalter, line 186:
```suggestion
            raise NotImplementedError(f"Given method {self.method} does not exist, please provide valid method parameter")
```

sktime/forecasting/compose/_ensemble.py
```diff
@@ -114,6 +115,9 @@ def __init__(
             forecasters=forecasters,
             n_jobs=n_jobs,
         )
+        if method and regressor:
+            raise ValueError("method and regressor can not be used together!")
+        self.method = method
```
aiwalter, line 120: I think we need a default for `method` which is not None and also a `method="feature-importances"`.

sktime/forecasting/compose/_ensemble.py
```diff
@@ -135,38 +139,51 @@ def _fit(self, y, X=None, fh=None):
...
+            if isinstance(self.regressor_, Pipeline):
+                # extract regressor from pipeline to access its attributes
+                self.weights_ = _get_weights(self.regressor_.steps[-1][1])
+            else:
+                self.weights_ = _get_weights(self.regressor_)
+            self._fit_forecasters(forecasters, y, X, fh)
+        elif self.method == "inv_var":
+            # get in-sample forecasts
+            self._fit_forecasters(forecasters, y, X, fh)
+            fh_insample = ForecastingHorizon(y.index, is_relative=False)
```
aiwalter, line 176: I am not sure if we should do insample predictions here. What about using also the `test_size` param here? We could do also first a train-test-split like in the method with feature-importances. Then we fit the forecasters on the train data and we check the inverse variance for the predictions only?

aiwalter: also I think some forecasters can not completely predict all insample values, they might raise an error.

aiwalter (commented):
@AngelPone I left a few comments. Did you test your method a bit for the error in comparison to the one that is already implemented? Could you find at least an ensemble that is outperforming each single given forecaster? You can find a code example in the description of my previous PR for `AutoEnsembleForecaster` for such a benchmark if you are interested

sktime/forecasting/compose/_ensemble.py
AngelPone: > I am not sure if we should do insample predictions here. What about using also the test_size param here? We could do also first a train-test-split like in the method with feature-importances. Then we fit the forecasters on the train data and we check the inverse variance for the predictions only?

I see your points, we can do that.
What I was considering is that using error variance in test sample may be not stable because of limited observations. Also I didn't see many papers using this strategy and I'm not sure how it works in real-world applications. Besides, the method I implemented is usually taken as a benchmark in the literature and what I thought originally was to implement as in the literature.  And another defect of using test set you need to fit the forecasters twice, which will double computation cost.
The advantage of using your suggested method is that it is more compatible with current class design.
Anyway, you decide it.



> also I think some forecasters can not completely predict all insample values, they might raise an error.

That's true. But I don't think it's a problem that this forecaster need to consider. I think the forecaster used should produce `NA` values instead of raising error when can't give in-sample fit, and `NA` will not affect the implementation now.

sktime/forecasting/compose/_ensemble.py
AngelPone: Do you suggest:
1. if `method` and `regressor` both are not given, we go to `method="feature-importance"` and Boosting regressor.
2. If `method` is given as `feature-importance`, we go to  `method="feature-importance"` and Boosting regressor.
3. If `method` is given as `feature-importance` and regressor is given, we go to `method="feature-importance"` and given regressor.
4. If `method` is given as others, we go to `method`
5. If `method` is given as others and regressor is also given, raise an error.
Maybe a little confusing?

AngelPone:
@aiwalter  I will do that after the code is finished.

sktime/forecasting/compose/_ensemble.py
aiwalter: what is the variance between NaN and a value? I am fine with in-sample approach, but then we might also change the one I implemented and get rid of the train-test-split. We can do that in a separate PR and we would have to deprecate the `test_size` parameter

sktime/forecasting/compose/_ensemble.py
aiwalter: I would say lets default to `method="inverse-variance"` and not None as it will be faster to compute also. For each of the methods, we have to check that all needed params are given or setted (in case of None e.g.).

aiwalter: I think then things are more clear for user and for us to implement?

sktime/forecasting/compose/_ensemble.py
AngelPone: > what is the variance between NaN and a value? 
`NA` will be omitted when compute statistics such as mean, variance, just like missing values.
Let's stick with the behavior now, I will use the testset to compute variance.

aiwalter: yes, I think that is the cleaner and more stable way to calculate the variance :)

fkiraly (dismissed):
Makes sense to me.
Inverse variance weighting is a common method so makes sense to add.

sktime/forecasting/compose/_ensemble.py
```diff
@@ -39,15 +39,23 @@
...
     ----------
     forecasters : list of (str, estimator) tuples
         Estimators to apply to the input series.
+    method : str, optional, default="feature-importance"
+        Strategy used to compute weights. Available choices: `feature-importance`,
+        `inverse-variance`.
+        `feature-importance`: use the `feature_importances_` or `coef_` as optimal
+        weights.
+        `inverse-variance`: use the inverse of forecasting error variance to compute
+        optimal weights, the passed regressor will be omitted.
```
aiwalter, line 58:
```suggestion
    method : str, optional, default="feature-importance"
        Strategy used to compute weights. Available choices:
        - ``feature-importance``: use the ``feature_importances_`` or ``coef_`` from given ``regressor`` as optimal
        weights.
        - ``inverse-variance``: use the inverse variance of the forecasting error (based on the internal train-test-split)
        to compute optimal weights, a given ``regressor`` will be omitted.
```

aiwalter: just pushed this as formatting here might mess up

aiwalter (approved):
I just fixed some details in the docstring. I tested the error of `method="inverse_variance"` on `load_airline` and `load_shampoo` and compared it with `method="feature-importance"`. The last was always better, also for different `test_size` values. So I think its good to keep it as default for now, until we have another method that might outperform.

fkiraly (approved):
Looks reasonable to me - inverse variance is not an uncommon technique, and the addition does not change the default behaviour.

PR #1620, opened 2021-11-16
CI/CD refactor and switch to github actions

lmmentel:
#### Reference Issues/PRs

- Fixes #1482 
- Fixes #1728 
- Fixes #1803
- Fixes #1832

See also #1150 
Refactoring of github actions workflows started in #1321 

Currently blocked by: 
- [x] #1644
- [x] #1661
- [x] #1662 
- [x] #1792
- [x] #1747
- [x] #1783
- [x] #1803
- [x] update installation instructions in [docs/source/installation.rst](docs/source/installation.rst)

#### What does this implement/fix? Explain your changes.

This is a rather substantial effort involving changes to how `sktime` is build and distributed. The main changes are:

-  Add [pyproject.toml](https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/) (see [PEP 621](https://www.python.org/dev/peps/pep-0621/) and [PEP 633](https://www.python.org/dev/peps/pep-0633/)]. The intention behind is to collect all package matadata and dependencies in one place.
- Add sets of optional dependencies:
    - `all_extras` all "soft" dependencies
    - `dev` packages useful for development and testing
    - `binder` packages required for binder and running notebook examples
    - `docs` packages required to build the documentation
- Add section to build time dependencies.
- Refactor existing CI/CD pipelines to use the new build configuration.
- Introduced a new CI/CD pipelines for running tests across win, mac, linux and py37, py38, py39 with github actions.
- Removed the obsolete requirements files.
- Include wheels build for macosx on arm64.

**Removed**

- ❌ Drop support for python3.6
- ❌ Drop appveyor and azure pipelines

**High level diagram for the CI/CD workflows**

sktime-cicd

#### Does your contribution introduce a new dependency? If yes, which one?

It does not introduce any new runtime/install dependencies but it could introduce an optional development depencency/build system such as [poetry](https://python-poetry.org/) to simplify development/build/distribution.

#### What should a reviewer concentrate their feedback on?

This PR introduces changed to the CI/CD pipelines that would hopefully allow to drop ``appveyor`` and ``azure pipelines`` in favor of ``github actions`` which would simplify the current build and maintenance process.

##### For all contributions
- [x] I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/main/.all-contributorsrc).
- [x] Optionally, I've updated sktime's [CODEOWNERS](https://github.com/alan-turing-institute/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
[...]

PR #1661, opened 2021-11-28
[BUG] Bugfix alignment tests

fkiraly:
This comes indirectly from @lmmentel's work on PR #1620.

Four tests related to the `Alignment` scitype were failing for him in the `datatypes` module.

The reason for that was that no fixtures for alignments were defined, so the tests for mtype checks and conversions related to alignments (data containers of the `Alignment` mtype) were failing.

The error messages were informative, so it was clear what to do:
* add the missing fixtures
* escape the tests assuming mtypes to be unique for alignments

What is, however, very troubling that:
* the remote CI did *not* detect this failure
* I not detect this failure when locally testing the alignment module. I do not understand why I didn't.

Currently - in a newly set up environment - the tests *are* failing locally.

Hence, there may be additional issues about our CI which this might uncover.


lmmentel:
Would it possible to merge this one? Since tests are failing without it, it's hard to test the CI workflow in #1620.

fkiraly:
a core developer not identical to me needs to review it first...
the list of current core devs is in the "roles" doc.

lmmentel:
Would you please advise on what would be the preferred action to take here? I found the list of 10+ core devs but I'm a bit reluctant to call out everyone to help merge this. I appreciate your help!

fkiraly:
> Would you please advise on what would be the preferred action to take here? I found the list of 10+ core devs but I'm a bit reluctant to call out everyone to help merge this. I appreciate your help!

Google their home addresses and throw bricks through windows until someone reviews this. That's what I'm doing usually.

Jokes aside, I think this is best resolved in the Friday stand-up (1pm discord), asking core devs directly.
I've already posted in the core dev channel and people seem busy.

lmmentel:
I like the escalation process here! I'm not a brick thrower myself but I'll find a way.

I'm happy to wait until Friday since there is no real urgency.

mloning:
It's possible that the tests weren't picked up by the CI because the tests folders don't have init files.

mloning (commented):
I added the init file to the test folder.

sktime/datatypes/_alignment/_examples.py
```diff
@@ -0,0 +1,28 @@
...
+elements are data objects, considered examples for the mtype
+    all examples with same index are considered "same" on scitype content
+    if None, indicates that representation is not possible
+"""
+
+import pandas as pd
+
+example_dict = dict()
+
+###
```
chrisholder, line 19: ?

chrisholder (approved):

fkiraly:
>I added the init file to the test folder.

Thanks, that must be why the test eloped detection on remote.

sktime/datatypes/_alignment/_examples.py
fkiraly: separator for different fixtures/examples.
There's just one in the file so it might be confusing.

As just discussed on slack, would leave it in for consistency.

PR #1663, opened 2021-11-28
transformers refactor: summarizer and base class

fkiraly:
This refactors the `SummaryTransformer` to use the new transformer template.
Some minor changes there - `transform` was still present and overwrote `_transform`, and the "inner type" tags were set incorrectly.

In addition, this fixes two bugs:
* in the base template, appears when a `Panel` is passed to a `Series`-wise transform that outputs `Primitives` - in this case, thet different rows were not concatenated properly (some index issues).
* the base class now enforces that `fit` is called before `transform`, even if the `fit-in-transform` flag as True (which means `fit` is empty and can be skipped, but `fit` still needs to be called). As a shorthand for "transform right away", `fit_transform` should be called, not `transform`.


sktime/tests/test_all_estimators.py
```diff
@@ -348,9 +348,14 @@ def test_fit_returns_self(estimator_instance):
 def test_raises_not_fitted_error(estimator_instance):
     """Check that we raise appropriate error for unfitted estimators."""
     estimator = estimator_instance
+
     # call methods without prior fitting and check that they raise our
     # NotFittedError
     for method in NON_STATE_CHANGING_METHODS:
+        # if has fit-in-transform tag (empty fit), no error needs to be raised
+        if method == "transform" and estimator.get_tag("fit-in-transform"):
```
mloning, line 356: I think we still want the workflow to first call `fit` and then `transform`, even if `fit` is empty. Changing that seems like a big risk.

fkiraly: can you give me then the definition of the tag `"fit-in-transform"`? Please give a precise spec. What does the tag tell us about a transformer?

mloning: I'm not sure, perhaps we have to come up with a precise definition. I think it tries to capture the fact that in some transformers, `transform` will change parameters, something that usually is only happening in `fit`.

fkiraly: That's not what I thought - I thought it just meant that `fit` was empty.
Who came up with it? Was it you, @aiwalter?

fkiraly: I would be in favour of all methods requiring either a call to `fit` or `fit_transform`.
But is this your understanding, i.e., `transform` should *always* raise an error, even if `fit` is empty?

mloning: I think it was me but doesn't really matter who came up with it. So I don't think we should move aways from always requiring fit to be called first, that would be a big interface change.

fkiraly: fine with me - I thought you wanted it to work it this way.
I'll just switch it out to "`fit` must *always* be called even if empty". Would be great if you could confirm if that's what you want.

fkiraly: changed it, if it's right now in your opinion, please resolve

mloning: yes that's fine, can't really comment on the other changes

fkiraly: ?

MatthewMiddlehurst (approved):
Just tested the SummaryTransformer for panel data portion, it all works fine with univariate data. With an input of shape (20, 1, 24) I get an output of shape (20, 7) and I can build a sklearn classifier with no issue.

Multivariate data seems to have an issue, however. With an input of (10, 6, 100) I get an output of shape (60, 7). The conversion process has given me 50 extra cases, I would expect (10, 42).

If we need this in to be on-track for release, I'm not going to push that the above be sorted now, as you could probably reshape the above output for the correct format (even though that would be annoying). Will approve as the other changes are good and let you decide if more should be added to the PR.

fkiraly:
> Multivariate data seems to have an issue, however. With an input of (10, 6, 100) I get an output of shape (60, 7). The conversion process has given me 50 extra cases, I would expect (10, 42).

What should the result be in the multivariate case, in your opinion? Would you expect multi-index across rows, or across columns? Or, something else?

fkiraly:
@MatthewMiddlehurst, I opened an issue here, I think it's primarily a design question:
https://github.com/alan-turing-institute/sktime/issues/1719

FYI, the vectorization over the variables isn't coming from me, i.e., not automated by the base class.

If you didn't intend it, it's probably coming from apply-like `pandas` commands which also happen to work seamlessly in the multivariate case, without you taking it into account.

PR #1664, opened 2021-11-29
Distance metric refinement

chrisholder:

TLDR:
Improves the speed of the metrics and fixes a bug that occurred when using the numba fastmath flag and bounding matrix parameters. In addition the lower_bounding parameter has been removed and instead the lower_bounding technique to use is inferred based on the other parameters passed.

Full version:
Previously the distances weren't being cache'd as efficiently as possible and so the cache flag has been added to the factory methods to allow them to be cache'd for repeat calls.

In addition there was a bug caused by using the fastmath numba flag where values were being inserted into the wrong indexes of the bounding matrix (still not sure why this is caused). As such for now the fastmath flag has been removed.

Finally the lower_bounding parameter felt redundent as it could simply be inferred by which parameters are being passed and therefore the parameter was removed.

##### For all contributions

##### For new estimators


fkiraly (changes requested):
I think this is ok if we can get it in for the next release.
If the release happens before the merge, I would have major reservations because this would result in interface changes, so can we get this done quickly?

Besides that, I only have one minor comment: please remove any commented out code.

TonyBagnall:
> I think this is ok if we can get it in for the next release. If the release happens before the merge, I would have major reservations because this would result in interface changes, so can we get this done quickly?
> 
> Besides that, I only have one minor comment: please remove any commented out code.

hi franz, what interface changes are there? All of thes numba distances are new, and all the existing code is still there

TonyBagnall (dismissed):
I have tested this for correctness and speed on dtw. Any further problems can be resolved by latter PRs, but I am confident this is now all correct.

TonyBagnall (approved):

Issue #1635, opened 2021-11-23
[DOC] documentation & developer instructions (Windows experience) fixing umbrella issue


fkiraly-shell:
I've been trying to set up a sktime developer environment in my day job role, we have a standard hardware set-up that uses Windows laptops.
Unlike the unix/mac experience which seems smooth, I truly must say that the Windows experience of the developer instructions is quite a bit atrocious, even mere installation can be confusing.
We've done user testing with 5 people, and no one managed to complete the `sktime` developer set-up without any hassle.

So, we've decided to help by fixing this, with @eenticott-shell, @amrith-shell, by:

- [x] updating and validating the installation instructions in windows. Related issue: #1372, #1203
- [ ] completing and re-organizing the developer guide (currently mostly empty!) and the "how to contribute" page. Related issue: #1448, #1203

amrith-shell:
I will pick up the 2nd task -  completing and re-organizing the developer guide (currently mostly empty!) and the "how to contribute" page.

fkiraly:
Great (as discussed) - @amrith-shell, it would be nice if you could kindly outline what in your opinion should go where, i.e., a sketch of end state. Happy to help with that tomorrow, although I assume you might already have better ideas since you worked through this in the user journey.

TonyBagnall:
@jasonlines has a version of this too, good to compare notes

jasonlines:
yep, went through the pain of getting it up and running on windows last week so uninstalled everything to do it again yesterday from first principles and document it. I've taken screenshots throughout and the plan is to put it together in a doc/guide - happy to put it together as a notebook or similar initially and go from there?

fkiraly:
great! Let's do this together then

eenticott-shell:
I have some instructions written up in markdown that I was able to use for a successful install on a new windows pc. It will be interesting to see if we have the same path.

jasonlines:
ok great, sounds good. I'll aim to get mine into a markdown by tomorrow lunchtime (hopefully the close of play today though) and we can compare notes before going from there

jasonlines:
It took longer than planned but please see #1640 for my first pass at this

amrith-shell:
In terms of the 2nd task of re-organizing the documentation/developer guide, I took the approach of categorizing various sections of the documentation and created a hierarchy/ decision tree of the documentation structure. Please see my initial draft.

If you have any comments or suggestions please do let  me know.

[...]

Issue #1372, opened 2021-09-01
[DOC] installation instructions general and windows, updates/clarifications


fkiraly:
This issue is about reproducing and potentially documenting installation of the developer version that seems to deviate from current installation instructions; and reproducing whether the current instructions, as they are written down, work properly (or not).

---
Via an environment mishap, I was forced to re-install my `sktime` development version on windows.

The vanilla workflow described in the installation instructions did not work - I had to do this (on the newest Windows version, with Anaconda):

first navigate to local `sktime` clone root folder, then:

```
conda create --name sktime python=3.8
conda activate sktime
conda install brotlipy
pip install -e .[all-extras]
pip install pytest
pip install esig
conda install -c conda-forge fbprophet
pip install numpy==1.19.5 --force-reinstall
```

Some takeaways, I think we need to update the install instructions:
* `fbprophet` causes problems, it needs to be singled out in a separate installation step. The instructions should be explicit, e.g., about installing it first separately, or later (which then makes the force-install `numpy` line necessary)
* setup.py extras do not include esig, which causes breakage in the dev version
* `pip install -e .[all-extras]` is actually what the developer user wants, but that line actually doesn't appear in the install instructions
* and what's that about `brotlipy`, not sure - I did this after following a random fix for an error message I got

fkiraly:
Would also be good if someone could test setting up environment on windows under the current installation instructions.

I feel this also has to do with the binder fix of @corvusrabus which required a two-step approach, and possibly with work done by @freddyaboulton - perhaps you something wise to say about the above?

freddyaboulton:
@fkiraly Regarding prophet, I don't think we need a two-step approach. I think we need to switch over to the `prophet` package as opposed to `fbprophet`. I think the problem is that `fbprophet` is quite old and they list `pystan>=2.14` as a requirement. This causes `pystan` 3.2.0 to be installed which is actually not compatible with prophet! Prophet only supports pystan 2. `prophet` has the correct pin of `pystan~=2.19.1.1`. 

I modified the setup.py to list prophet instead of prophet and then changed `fbprophet.py` to import `prophet` instead of `fbprophet` and I think everything is working ok. 

Can you give that a shot? I guess the alternative is to list `pystan` 2.19 as a requirement but I think just using `prophet` is cleaner.

Not sure what happened with `brotlipy`. I agree we should also list `esig` in `all_extras`. You can't even run the unit tests without it.

fkiraly:
[...]

PR #1672, opened 2021-12-01
Update windows install doc

eenticott-shell:
Updated the Windows build requirements and installation instructions to be up to date with latest Visual Studio code. Also updated instructions to ensure that the conda environment set up works from a fresh install to get an editable developer version.

#### Reference Issues/PRs
Contributes to part 1 of #1635 

It updates the Windows section of the development installation guide. Specifically it updates the instructions for installing Visual Studio Code for 2022 and clearly shows which components are needed. It also updates the sequence of commands to obtain a working conda environment for sktime development stating how to resolve some common dependency issues as well as some ways that users can check that their installation has worked.

No

The clarity of the new instructions. Is there anything else that should be addressed? Does the image I've added in the Build Tools for Visual Studio Tools section load correctly?

I have kept as much of the old instructions as I could, for example attempting to install via `python setup.py install` outside an environment, however I couldn't get this to work myself. Should I remove this and only have instructions for an anaconda environment? Thus introducing a new requirement for windows installation, or keep it as is, as it may work for some people.
 
#### Any other _comments?_
I plan to make further changes to the order of the installation instructions once the restructuring in #1635 part 2 has taken place. This PR is simply to update the windows instructions so that they should work.


docs/source/installation.rst
```diff
@@ -374,19 +390,29 @@ And build sktime from this environment:
...
+1. Open up an anaconda terminal
+2. Navigate to your local sktime folder :code:`cd sktime`
+3. Create new environment: :code:`conda create -n sktime-dev python=3.8`
+4. Activate environment: :code:`conda activate sktime-dev`
+5. Install required packages:
+    1. :code:`pip install numpy=1.19.3`
+    2. :code:`pip install cython`
+    3. :code:`pip install -r build_tools/requirements.txt`
+    4. If **fbprophet** fails to install:
+      1. :code:`conda install -c anaconda ephem`
```
niekvanderlaan, line 408: manually installing ephem was not needed in my case. Have you tried without?

eenticott-shell: It does seem to work without, so I will remove this line

fkiraly: well, that line was ephemeral...

docs/source/installation.rst
```diff
@@ -350,20 +350,36 @@ The building command depends on the architecture of the Python interpreter,
...
 
 For 64-bit Python, configure the build environment with:
 
 .. code-block:: bash
 
     SET DISTUTILS_USE_SDK=1
-    "C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Auxiliary\Build\vcvarsall.bat" x64
+    "C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Auxiliary\Build\vcvarsall.bat" x64
+
+For 32-bit Python, configure the build environment with:
```
fkiraly, line 378: the code block for 64-bit and for 32-bit python seems to be exactly identical. Is that a mistake?
I suppose so, either you don't need to make a case distinction, or there is a typo, perhaps in relation to the `x64` at the end.

fkiraly: should be `x86`, right?

docs/source/installation.rst
```diff
@@ -350,20 +350,36 @@ The building command depends on the architecture of the Python interpreter,
...
+2. Scrolls down to "Tools for Visual Studio 2022" and click the dropdown,
+3. Download and install "Build Tools for Visual Studio 2022"
+4. Open the Visual Studio installer
+5. Tick "Desktop Development with C++"
+6. In the panel to the write make sure "MSVC v143 VS 2022 C++ x64/x86 build tools (Latest)" is selected
+7. Click install
+
+.. images:: images/visual_installer_selection.png
 
 For 64-bit Python, configure the build environment with:
```
fkiraly, line 371: may need some more instructions that this should be typed in command line or power shell. The current formulation "configure the build environment" is vague on what exactly to do. It sounds like an instruction, but in fact it's a summary, that ambiguity should be resolved.

docs/source/installation.rst
```diff
@@ -374,19 +390,29 @@ And build sktime from this environment:
...
+4. Activate environment: :code:`conda activate sktime-dev`
+5. Install required packages:
+    1. :code:`pip install numpy=1.19.3`
+    2. :code:`pip install cython`
+    3. :code:`pip install -r build_tools/requirements.txt`
+    4. If **fbprophet** fails to install:
+      1. :code:`conda install -c anaconda ephem`
+      2. :code:`conda install -c conda-forge install -c pystan`
+      3. :code:`conda install -c conda-forge install -c fbprophet`
+      4. Verify all requirements are satisfied by running :code:`pip install -r build_tools/requirements.txt` with no errors
```
fkiraly, line 411: what to do if the requirements fail to verify? Try again from where, how often? Bug report on sktime? Ask for help on gitter?

fkiraly: same below. Would be helpful if there was general advice, perhaps at the start, what to do if it doesn't work. I guess try again in a clean env, and if it doesn't work multiple times, ask for help?

docs/source/installation.rst
```diff
@@ -350,20 +350,36 @@ The building command depends on the architecture of the Python interpreter,
 The above commands assume that you have the Python installation folder in your
 PATH environment variable.
```
fkiraly, line 351: maybe add a link how to modify Windows env variables (this is a very common topic, so perhaps somewhere on Microsoft webpage), and what a common path could look like (note this may differ in admin and in restricted user install)

docs/source/installation.rst
```diff
@@ -374,19 +390,29 @@ And build sktime from this environment:
...
-1. Install Anaconda
-2. Create new environment: :code:`conda create -n sktime-dev python=3.8`
-3. Activate environment: :code:`conda activate sktime-dev`
-4. Install NumPy (pinned to 1.19.3) from pip: :code:`pip install numpy==1.19.3`
-5. Install requirements: :code:`pip install -r build_tools/requirements.txt`
-6. Follow the instructions above to point to "vcvarsall.bat"
-7. Run :code:`pip install --verbose --no-build-isolation --editable .`
+1. Open up an anaconda terminal
+2. Navigate to your local sktime folder :code:`cd sktime`
+3. Create new environment: :code:`conda create -n sktime-dev python=3.8`
```
fkiraly, line 401: -> "create new environment with python 3.8"?

fkiraly: maybe also add a warning, that this only works for the first time if such an environment does not already exist. Second time round, you have to nuke the existing environment first.

docs/source/installation.rst
```diff
@@ -374,19 +390,29 @@ And build sktime from this environment:
 
     python setup.py install
 
-Replace ``x64`` by ``x86`` to build for 32-bit Python.
-
 Some users have experienced issues when installing NumPy, particularly version 1.19.4. Note that a recent Windows update may affect compilation using Visual Studio (see `Windows update issue <https://developercommunity.visualstudio.com/content/problem/1207405/fmod-after-an-update-to-windows-2004-is-causing-a.html>`_).
 
+Setting up a Conda environment
+""""""""""""""""""""""""""""""
 If you run into a problem installing the development version and are using Anaconda, try:
```
fkiraly, line 397: is this section really only "if you run into a problem"? It looks like this condition should be removed, since it appears like it's the standard workflow, and not conditional "ran into problem" coming from another section.

fkiraly (changes requested):
Nice, thanks!

I think this should clear up a lot.

I made some comments that I'd say minor/non-blocker. There may be typos in there though and minor improvements.

Another comment: if you compare with my own working commands in #1372, yours did not require to install `brotlipy` or `esig`. Do we want to include these, and `ephem`, as possible "people have reported this" in a troubleshooting section?

docs/source/installation.rst
eenticott-shell: yes it should

docs/source/installation.rst
eenticott-shell: added this

docs/source/installation.rst
eenticott-shell: Yes, I wasn't sure how to handle this, so I left the old instruction up which just runs `python setup.py install` to install sktime. I did this as I wanted to avoid having introducing a dependency on anaconda for the windows installation. I have now changed this so these instructions are the default but added some preamble saying these instructions are for anaconda but the process will be similiar for any virtual environment manager.

docs/source/installation.rst
eenticott-shell: done this

docs/source/installation.rst
fkiraly: makes sense.

fkiraly (approved):
looks good now, thanks!

PR #1677, opened 2021-12-02
Added trend_ and resid_ attributes to STLTransformer. Bugfix.

aiwalter:
Added trend_ and resid_ attributes to STLTransformer . Refactor.


aiwalter:
@fkiraly @mloning this `STLTransformer` can not transform on data that is not equal to the data given in `fit()`. Is just returns NaN which is obviousl when looking in to this:

https://github.com/alan-turing-institute/sktime/blob/18cf7953014876153fba2d0a3db32be51ba4f69d/sktime/transformations/series/detrend/_deseasonalize.py#L515-L516

I think we should at least raise an error here? @eyalshafran I htink you contributed this, whats your opinion here?

eyalshafran:
I can’t think of a use case that would require STL transformation on data different then the fit data. If there are no use cases then we should at least raise an error.

aiwalter:
@eyalshafran well currently the `STLTransformer` is returning the deseasonalized data. Other outputs like `resid_` and `trend_`, I just added as attributes. So deseasonalizing in pipeline will still require `Deseasonalizer` as it can transform unseen data, and thus inverse transform `y_pred` in the pipeline. So I am going to add an `NotImplemented` error probably then for `STLTransformer` if index in `transform()` or `inverse_transform()` is different to the one given in `fit()`

eyalshafran:
@aiwalter - those changes make sense to me.

The deseasonalizer is better for transforming on unseen data since it assumes the seasonality does not change from one period to another. That is not the case with STL - the seasonality can slowly change over time. Therefore STL should probably not be used in unseen data.

RNKuhns:
@aiwalter and @eyalshafran the points about not applying to unseen data make sense. 

And @eyalshafran's point about slowly changing seasonality is spot on. 

To use this in a pipeline for forecasting we need a STLForecaster. 

The STLForecaster can use the STL transformer for decomposition and then let people specify how they want to forecast the components (e.g. seasonality, trend, resid). Finally it can add the three components back together.

aiwalter:
@RNKuhns I have it locally already an STLForecaster. Will push it soon. You can be my co-author and reviewer :) there are just some interface question, like should we give STLTransformer also as a param so user can set the paramd there.

fkiraly (changes requested):
Comments/change requests:

* why can `STLtransformer` not be applied out-out-sample? Trend and seasonal component both exist on the real line, you can hence subtract them arbitrarily.
* instead of storing the components internally, I'd return them from `transform`, and control by parameters like `return_components` whether only the filtered series is returned, or also the components. The return would have to be a 3-column `pd.DataFrame` then, even if a univariate series is pasesd. Would this suggestion make sense?
* in line with the new transformer interface, no `transform` should be present in `STLTransformer`, so perhaps you can try remove it while we are working on it?
* you are changing the `sp` parameter from 1 to 2, why? That would change behaviour of default instances across versions.

eyalshafran:
@fkiraly Regarding the first bullet point - STL is locally fitted so the trend and seasonality directly relate to the fitted data. I can’t think of a reason to arbitrary subtract the trend or seasonality. I think that would lead to unwanted results. At least that is my understanding.

RNKuhns:
@aiwalter shoot me a message on slack when you are ready for me to take a look (so I see it--I always miss the GitHub notifications). Happy to think through parameter choices and all that. 

I like Franz's suggestion that you have a parameter to determine if you return the adjusted series or the three components. 

I agree with @eyalshafran that I think because it is a local transformation you couldn't apply it to unseen data (it would need the same transform).

ltsaprounis:
STL is very popular for deseasonalising before forecasting. The fact that it allows for the seasonal component to change is a feature not a bug. Here is a reference from the 3rd edition of Forecasting Principles and Practice: [13.1  Weekly, daily and sub-daily data](https://otexts.com/fpp3/weekly.html#weekly). 

In the STL example in that chapter, STL will return the seasonality component forecast in a seasonal naive style (i.e. for any horizon step it will return the seasonal component value on `horizon step - sp`). In cases where the seasonality changes over time, this will be better than the Deseasonalizer that uses moving averages [(statsmodels ref)](https://www.statsmodels.org/stable/generated/statsmodels.tsa.seasonal.seasonal_decompose.html) and the seasonality is always the same.

I guess that this will be possible using  @aiwalter's `STLForecaster` but something like an `STLDeseasonalizer` that does this only would be a good feature to have.

@HYang1996 - Can you shed some light on how this is done in fable and/or forecast in R?

RNKuhns:
@ltsaprounis I believe what @aiwalter and I have been talking about will cover that typical use case with some additional flexibility. 

It looks like in Hyndman's textbook the decomp_model first applies the decomposition to return the components and then is using ETS along with Naive re-seasonalization. 

I think we should be able to specify how to forecast each of the three components (trend/cycle, seasonality and residual). A sensible default for the seasonality is naive forecaster.

HYang1996:
@ltsaprounis I've never looked into this in fable or forecast. If you're looking for their implementation then you may check it [here](https://github.com/tidyverts/fable/blob/86ce90ecf550ecfb5ba7bd5d04b4c8f731eae3f4/R/stl.R#L16). Otherwise @mloning probably knows much more than I do.

eyalshafran:
@ltsaprounis the STLtransformer piped together with a forecaster (see documentation of STLtransformer) works the same as R implementation.

ltsaprounis:
> @ltsaprounis the STLtransformer piped together with a forecaster (see documentation of STLtransformer) works the same as R STLF implementation.

@eyalshafran - Yes, that's the what the current implementation does, but this PR is changing this behaviour.

> I think we should be able to specify how to forecast each of the three components (trend/cycle, seasonality and residual). A sensible default for the seasonality is naive forecaster.

@RNKuhns - Cool, that makes sense.

fkiraly:
>I agree with @eyalshafran that I think because it is a local transformation you couldn't apply it to unseen data (it would need the same transform).

@eyalshafran, @rnkuhns, why can't we apply the transformation to unseen data?

The *trend* component is simply a linear fit. It results in a linear function f: R-> R.
The *seasonal* component is on a periodic grid, it's a function g: aZ -> R where a is some constant.
Assuming the test data has indices that "extend the pattern", at an index t and observation y(t) you can just compute the STL transformation as y(t) - f(t) - g(t).

Or am I missing things here?

eyalshafran:
@fkiraly 
From a mathematical stand point here is the original paper - http://www.nniiem.ru/file/news/2016/stl-statistical-model.pdf

From a practical perspective the STL implementation in statsmodels does not return anything that would allow us to compute the trend and seasonality for unseen data.
I also don’t understand why someone would want to do that. What is the use case?

Going back to the reason I wrote this transformation - sktime didn’t have an stl forecaster and I needed one the works in a similar way to STLForecaster in statsmodels and STLF in R. So I wrote this transformation so stl can be used as a pipeline with any forecasting algorithm from sktime. I’ve been using this transformation with ETS forecaster in my production code for the past few months and it is my best preforming forecasting algorithm out of all the sktime forecasters.

mloning:
I agree, that the trend and seasonal component can be in principle be extrapolated into the future (or past) for a given time index. We do this in the `Detrender` for example if I remember correctly.

fkiraly:
>From a practical perspective the STL implementation in statsmodels does not return anything that would allow us to compute the trend and seasonality for unseen data.

I disagree, see above. Seasonality is periodic, and an "abstract" trend is simply a forecaster/predictor which can be extrapolated in-principle.

On a side note: what is implemented in the various STL methods internally is not what is in the paper, when it comes to the trend computation? There is smoothing (LOESS) and filtering in the paper, but our default is just a linear trend, if I'm not mistaken?

> I also don’t understand why someone would want to do that. What is the use case?

I don't think it's a good principle to restrict functionality just based on us individually thinking "no one wants this".
The general rule in a widely used package such as `sktime` is "someone wants this as long as it's not totally wrong".

aiwalter:
@fkiraly :

> The return would have to be a 3-column pd.DataFrame then, even if a univariate series is pasesd. Would this suggestion make sense?

this would be an interface change which I would like to avoid. We could rather have a param that specifies what is returned. Either deseasonalized (currently), detrended or 
"deresidualized" (removed residuals). What do you think?

> you are changing the sp parameter from 1 to 2, why? That would change behaviour of default instances across versions.

Yes, bcz `statsmodels.STL` accepts only ` sp>=2`. The workaround that was implemented to accept `sp=1` has just returned the identity transformation which is useless and I guess there is a reason why `statsmodels` also excluded this case.

@ltsaprounis 

> @eyalshafran - Yes, that's the what the current implementation does, but this PR is changing this behaviour.

Which behavior is changed in your opinion?

eyalshafran:
@fkiraly - The seasonality is not periodic. It can change over time in a nonlinear fashion. Just look at the extracted seasonality figures in the paper to see that. You can look at the cycle-trend figures in the paper to get a better sense of how the change in seasonality is calculated - you take all of the data point for each period (for example all the data points in Jan.) and you run loess regression to smooth the scattered data. This is your Jan. seasonality trend. There are a bunch of smoothing operations after that step. There is no analytical solution. 

@mloning - yes. In theory the data can be extrapolated. This should be done carefully. For in sample points it should be extrapolated per period separately (similar to how the cycle-trend is computed in the paper). The extrapolated results for time index outside the range of the data used for fitting would be unreliable. Future values should be computed by seasonal naive (which is what the inverse transform does and I guess it can also be used for transform). Not sure what the proper method would be for past data is.

sktime/transformations/series/detrend/_deseasonalize.py
```diff
@@ -482,47 +432,60 @@ def fit(self, Z, X=None):
...
+                """
+                STLTransformer is only a descriptive trasnformer and
+                can only transform data that was given in fit().
+                Please use Deseasonalizer or Detrender."""
+            )
         return y - self.seasonal_
 
     def _inverse_transform(self, y):
-        seasonal_index = self._align_seasonal_index(y)
-        return y + self.seasonal_[seasonal_index]
```
ltsaprounis, line 520: > Which behavior is changed in your opinion?

@aiwalter, @eyalshafran - By removing `_align_seasonal_index` here we won't be able to use the transformer as a deseasonaliser, which was possible in v.0.8.1. Am I missing something here?

aiwalter: it was not possible before, it has just returned NaN for outsample transformations. At least this is what I got.

RNKuhns:
I wanted to clarify a couple of items.

First, we should be clear in delineating between what a decomposition should return and the different ways we might use the results. Decomposition methods are actually univariate-to-multivariate transformers at their core. The goal is take a single input series and decompose it into `n` components. STL is a specific case of this, where a series of locally estimated scatterplot smoothing (LOESS) passes are used to decompose the input series into 3 components (note there are extensions that are designed to return multiple seasonal patterns, but I'm talking about the classical case): the trend `T` (which is really the trend and non-seasonal cycle), seasonality `S` and the irregular (residual) `e`.

So if X is the input series we have decomposed the series so X = T + S + e. 

In terms of use cases:

From a straight univariate-to-univariate transformation perspective:
1. Returning deseasonalized data (X-S) -- you can use the results of the decomposition as a deasosonalizer
2. Returning seasonal component (S) -- you might want to examine how it changes over time or just inspect the pattern
2. Returning data without trend/cycle component (I don't think this is common and I don't think we need to worry about it personally; X - T or S + e)
3. Returning residual component (e or X - T - S) -- you might want to see if these are white noise.

There is a separate use-case for forecasting. You often see this being done by returning the deseasonalized data (use case 1 above), applying a forecasting model to that data and then re-seasonalizing it by adding the seasonal naive forecast of the seasonal component. However, this embeds a series of assumptions and it is quite easy to generalize from this. 

- First, forecasting the seasonal component using the seasonal naive method ignores potential patterns in how the seasonal component is evolving. Modeling this could improve the forecasts by better predicting the changing seasonal patter (this is important in longer-run forecasts and we don't want to restrict our implementation to just support seasonal naive). 
- Second (and less important) is that it lumps the irregular component in with the trend/cycle component. In most cases this is probably fine. A more general case is forecasting the trend/cycle, seasonal and irregular components separately. If the irregular component happens to be white-noise this doesn't matter. Even if the irregular component is not white-noise, many forecasting approaches used to forecast the trend/cycle will handle the additional noise from the irregular component.

 Based on this, we probably want the STLForecaster to allow users to specify a forecaster to apply to the combined trend/cycle + irregular components (e.g. de-seasonalized data) and a separate forecaster (default to seasonal naive) to the seasonal component. The STLForecaster would internally handle the decomposition and apply the forecasters to the correct components and add them back together.

Based on the potential uses cases, I think we are really looking at 3 separate things:

1. A univariate-to-multivariate STLDecomposer transformer (STL being just one implementation of this type of decomposer, in the future we'll probably have others including the multiple seasonal pattern STL that Hyndman has written about)
2. A STL based de-seasonalizer (this is what already exists) that is kept with the other de-seasonalizers in the transformation package (and makes use of the univariate-multivariate STLDecomposer)
3. A STL based forecaster that decomposes the series, forecasts the components separately and adds them back together (This will make use of the STLDecomposer as well)

sktime/transformations/series/detrend/_deseasonalize.py
aiwalter: you can still use it as deseasonalizer, but only for the data that was given in `fit()`

eyalshafran:
@RNKuhns 
For forecasting we might want to have a general forecaster called DecompositionForecaster that takes in the following inputs:
1. Deseasonalizer - for example STL or seasonal_decomposition
2. Trend - forecaster to predict trend
3. Seasonal - forecaster to predict seasonal
4. Residual - forecaster to predict residuals

STLFotecaster can then be a private case of the DecompositionForecaster

RNKuhns:
> @RNKuhns For forecasting we might want to have a general forecaster called DecompositionForecaster that takes in the following inputs:
> 
> 1. Deseasonalizer - for example STL or seasonal_decomposition
> 2. Trend - forecaster to predict trend
> 3. Seasonal - forecaster to predict seasonal
> 4. Residual - forecaster to predict residuals
> 
> STLFotecaster can then be a private case of the DecompositionForecaster

I like this, in general, but just to clarify -- the first input should not be a "Deseasonalizer" it should be a "decomposer" that returns the components of the series (whereas I'd expect a deseasonalizer to return a single series that has had the seasonal component removed).

sktime/transformations/series/detrend/_deseasonalize.py
eyalshafran: We can have transform behave like the inverse_transform which would fix the example above. 

One other issue is that the statsmodels implementation does not work for data with missing data points or partially works for data with NaN values. I should have checked for that in the original code before using a numpy array to calculate STL.

ltsaprounis: @aiwalter you're right, the following code snippet returns NaNs for the version in main. The strange thing is that it works fine as a deseasonalizer for the 0.8.1. release.

```python
import numpy as np
from matplotlib import pyplot as plt
from sktime.transformations.series.detrend import STLTransformer, Detrender

y = load_airline()

    ("stl", STLTransformer(sp=12)),
    ("detrend", Detrender()),
    ("ets", AutoETS(trend=None, seasonal=None, error="add"))
])
forecaster.fit(y[:-12])
pred = forecaster.predict(fh=np.arange(12))

plot_series(y, pred)
plt.show()
```

0.8.1 result: 
image

aiwalter: do you know what has changed?

aiwalter:
> @ltsaprounis the STLtransformer piped together with a forecaster (see documentation of STLtransformer) works the same as R STLF implementation.

this did not work, as outsample transformation of `STLTransformer` did nor work and just returned NaN. I removed therefore the docstring example. Please correct me if I am wrong.

aiwalter:
I think returning a `pd.DataFrame` with three columns is not intuitive and an interface change. I would prefer to keep the other components as attributes or we can have a param. which of the attributes should be removed from `y`. Otherwise is there a concrete change request for this PR? We might get this in, if anyone else want to enhance it, we can have another PR. I need this PR for #1696

aiwalter:
This PR is not changing the current `STLTransformer`, but just adding exceptions and attributes. So I see not a big argument against it :)

fkiraly:
note to myself - need to have a look at this @fkiraly

aiwalter:
@fkiraly @ltsaprounis @RNKuhns @eyalshafran pls provide feedback, otherwise I assume we can go ahead with this PR. I need it for `STLForecaster` #1696 to be completed. Its a bit annoying, first many comments and now nobody is reacting anymore 😅

fkiraly (changes requested):
Ok, writing up what I've said above as a change request:

* can you kindly answer my question whether this transformer implements what is in the paper you linked (with LOESS)? If not, what reference for "STL" are you implementing? I think we need clarified which algorithm this specifically corresponds to.
* I do think the components should be returned by `transform`, otherwise users can't easily chain extraction of STL components for use in forecasters. This is possible only if it's a transformer output. It's also not an interface change (because `transform` returns `Series` in the standard series-to-series transformer). 
* It should be possible to apply `transform` to data not equal in `fit`. @mloning also agreed above that this is possible. I have reservations against merging it this way, since it may also indicate that fitting and application are not properly separated. At least I'd like to have that discussion resolved, since currently it just looks like opposite assertions.

fkiraly:
@eyalshafran, @RNKuhns, would your forecaster not just be an instance of a pipeline, where first you run a transformer that decomposes the series, and then you apply a forecaster?

aiwalter:
thx for review @fkiraly 

> can you kindly answer my question whether this transformer implements what is in the paper you linked (with LOESS)? If not, what reference for "STL" are you implementing? I think we need clarified which algorithm this specifically corresponds to.

This transformer implements a wrapper to `statsmodels.STL` and is not implementing anything else. I actually did not add this transformer to `sktime`, I just added the missing attributes and fixed some silent bugs. So I did not want to trigger again the whole STL method discussion with this PR :) @eyalshafran has added `STLTransformer` initially, so he might answer here. 

> I do think the components should be returned by transform, otherwise users can't easily chain extraction of STL components for use in forecasters. This is possible only if it's a transformer output. It's also not an interface change (because transform returns Series in the standard series-to-series transformer).

How should this work? the `statsmodels.STL` is returning 3 series, which I store now separately in self as attributes. If we return them all together as a ´pd.DataFrame´, this would be an interface change, as so far if we give a `Series` in, we get a `Series` out, and not a 3 column `DataFrame`. I dont get how you want to chain sth together, bcz that is why we need a separate `STLForecaster` that does the addition of the 3 forecated components, otherwise you would just have a vector forecast of the 3 components separately, but not a final `Series` prediction. Maybe things get more clear when you look into #1696

sktime/transformations/series/detrend/_deseasonalize.py
ltsaprounis: @aiwalter - sorry for taking ages to reply. Found the "issue". It's line 511, added in #1611: 
```python
self.seasonal_ = pd.Series(self.seasonal_, index=Z.index)
```

If you comment this out in the code that's in main, then the code snippet above works.

ltsaprounis:
I think @RNKuhns proposal makes sense and this transformer will be on the backend of all those applications of STL (forecaster and deseasonaliser). At the end of the day, STL is a decomposition that yields 3 series out of 1 (seasonality, trend, residuals). 

From my side the confusion is coming from the initial implementation of the `STLTransformer` that came with the 0.8.1 release. This was behaving like a deseasonalzer, which allowed to use with future data and as part of a `TransformedTargetForecaster`. An example of how I used it is here: [03-single-ts-forecasting.ipynb](https://github.com/ltsaprounis/dsf-ts-forecasting/blob/main/notebooks/03-single-ts-forecasting.ipynb) (scroll down to the "Forecasting Pipelines with sktime" section). 

In #1611 we added this code in line 511:
```python
self.seasonal_ = pd.Series(self.seasonal_, index=Z.index)
```
This caused problems with applying the fitted transform on future data. In that version it was applying seasonal naive to the seasonality series from the STL decomposition.

> would your forecaster not just be an instance of a pipeline, where first you run a transformer that decomposes the series, and then you apply a forecaster?  

You would need to apply a forecaster on 3 series (trend, seasonality and residuals) and then sum them. I don't think that's straightforward with the pipelines we have.

Maybe the best option is to implement the version in this PR and name it STLDecomposer as it's a prerequisite for the other classes in @RNKuhns suggestion. The decomposer should not be able to transform unseen data, just like in statsmodels. After we implement `STLForecaster` and `STLDeseasonalizer` we can then do something like `STLTransformer = STLDeseasonalizer` for backward compatibility with 0.8.1.

aiwalter:
@ltsaprounis i dont think that my previous PR has introduced any issue bcz I just changed the type of `seasonal_` to be `pd.Series` instead `np.array`. I have just tested `0.8.1` and I got an error on this:
```python
from sktime.transformations.series.detrend import STLTransformer

y = load_airline()
y_train, y_test = temporal_train_test_split(y, test_size=12)
stl = STLTransformer(sp=12, trend=13, seasonal=13)
stl.fit(y_train)
stl.transform(y_test)
```
```
<ipython-input-8-b369ca6fc20b> in <module>
      1 stl = STLTransformer(sp=12, trend=13, seasonal=13)
      2 stl.fit(y_train)
----> 3 stl.transform(y_test)

~\AppData\Local\Programs\Python\Miniconda3\envs\playground\lib\site-packages\sktime\transformations\series\detrend\_deseasonalize.py in transform(self, Z, X)
    535         self.check_is_fitted()
    536         z = check_series(Z, enforce_univariate=True)
--> 537         return self._transform(z)
    538 
    539     def inverse_transform(self, Z, X=None):

~\AppData\Local\Programs\Python\Miniconda3\envs\playground\lib\site-packages\sktime\transformations\series\detrend\_deseasonalize.py in _transform(self, y)
    512 
    513     def _transform(self, y):
--> 514         return y - self.seasonal_
    515 
    516     def _inverse_transform(self, y):

~\AppData\Local\Programs\Python\Miniconda3\envs\playground\lib\site-packages\pandas\core\ops\common.py in new_method(self, other)
     63         other = item_from_zerodim(other)
     64 
---> 65         return method(self, other)
     66 
     67     return new_method

~\AppData\Local\Programs\Python\Miniconda3\envs\playground\lib\site-packages\pandas\core\arraylike.py in __sub__(self, other)
     95     @unpack_zerodim_and_defer("__sub__")
     96     def __sub__(self, other):
---> 97         return self._arith_method(other, operator.sub)
     98 
     99     @unpack_zerodim_and_defer("__rsub__")

~\AppData\Local\Programs\Python\Miniconda3\envs\playground\lib\site-packages\pandas\core\series.py in _arith_method(self, other, op)
   4996         lvalues = extract_array(self, extract_numpy=True)
   4997         rvalues = extract_array(other, extract_numpy=True)
-> 4998         result = ops.arithmetic_op(lvalues, rvalues, op)
   4999 
   5000         return self._construct_result(result, name=res_name)

~\AppData\Local\Programs\Python\Miniconda3\envs\playground\lib\site-packages\pandas\core\ops\array_ops.py in arithmetic_op(left, right, op)
    187     else:
    188         with np.errstate(all="ignore"):
--> 189             res_values = _na_arithmetic_op(lvalues, rvalues, op)
    190 
    191     return res_values

~\AppData\Local\Programs\Python\Miniconda3\envs\playground\lib\site-packages\pandas\core\ops\array_ops.py in _na_arithmetic_op(left, right, op, is_cmp)
    140 
    141     try:
--> 142         result = expressions.evaluate(op, left, right)
    143     except TypeError:
    144         if is_cmp:

~\AppData\Local\Programs\Python\Miniconda3\envs\playground\lib\site-packages\pandas\core\computation\expressions.py in evaluate(op, a, b, use_numexpr)
    233         if use_numexpr:
    234             # error: "None" not callable
--> 235             return _evaluate(op, op_str, a, b)  # type: ignore[misc]
    236     return _evaluate_standard(op, op_str, a, b)
    237 

~\AppData\Local\Programs\Python\Miniconda3\envs\playground\lib\site-packages\pandas\core\computation\expressions.py in _evaluate_standard(op, op_str, a, b)
     67         _store_test_result(False)
     68     with np.errstate(all="ignore"):
---> 69         return op(a, b)
     70 
     71 

ValueError: operands could not be broadcast together with shapes (12,) (132,) 
```

aiwalter:
@ltsaprounis what should an `STLDecomposer` return? I think lets keep this PR but add a param where you can select between `method="deseasonalize"` and `method="detrend"`. The extrapolation would be a separate PR. It didnt work at any point in time for `STLTransformer` so I dont see this on my plate to fix it in this PR. I just added missing exceptions.

aiwalter (commented):
I discussed with @ltsaprounis how we will extrapoplat for `trasnform()` and `inverse_transform()`, we will just use a `NaiveForecaster(method"last")` for deseasonalizing and a `NaiveForecaster(method="drift")` for detrending. Further there will be another param `method` as descibed in above comment. @ltsaprounis will implement it before next release, so we can go ahead with this PR and merge it for now.

ltsaprounis:
Yes - we agreed that it's best to merge this with `transform` and `inverse_transform` as not implemented for unseen data to allow for the `STLForecaster` and fix the bugs that this PR is resolving. I'll then make another issue and implement these in a similar way to the `Deseasonalizer` and `Detrender` in another PR.

ltsaprounis (approved):

fkiraly (changes requested):
Ok, I've looked at this, I think we need a few changes still.

To comment on the three open points:
* regarding "what is STL". Ok, I'm fine with adopting the position that we just wrap `statsmodels` STL. But: then this needs to be written clearly in the docstring. For instance, at the top, write "wraps `statsmodels` STL". In the references/notes, put the reference to the statsmodels docs. The user needs to know what they get when they use this algorithm.
* regarding "returning series as attributes". It's not an interface change to return a `pd.DataFrame`. A `Series` (abstract type) can be `pd.DataFrame` or `pd.Series`. It's not a change to the transformer interface either. I'd suggest, add an argument `return_components`. If `False` (default), return only the current return. If `True`, return a `pd.DataFrame` with three columns. Transformers like this are fine and compliant with our current interface to the best of my knowledge.
* regarding "don't know how to extrapolate". If the transformer fits and transforms the same data, then don't raise a custom exception, but make it a `fit-in-transform`, and put the logic in `transform`. Ignore any data in `fit`. Perhaps also explain in the docstring that this transformer has that property. The reason for this is that all different interfaces and behaviours should be clear at base class level (and captured in tags where possible), different transformers shouldn't introduce different error messages etc, or you violate the unified interface pattern.

aiwalter:
> But: then this needs to be written clearly in the docstring. For instance, at the top, write "wraps statsmodels STL"

I already mentioned int in the docstring. Just added the reference as well.

> It's not an interface change to return a pd.DataFrame

For me it is, the case that we give a `pd.Series` in and we get a `pd.DataFrame` out did not exist so far. Therefore it is an interface change. Also the fact that we currently have "what goes in comes out" is not compliant with that.

>  different transformers shouldn't introduce different error messages etc, or you violate the unified interface pattern.

As descibed above, @ltsaprounis will remove that exceptions by implenting the extrapolation. So I dont see a need now to change anything regarding fit/transform.

fkiraly:
> For me it is, the case that we give a pd.Series in and we get a pd.DataFrame out did not exist so far. Therefore it is an interface change. Also the fact that we currently have "what goes in comes out" is not compliant with that.

That's not right in my opinion.
Transformers can always give back series with more variables, or a data frame with multiple columns even if the imput is a `pd.Series`. Look at `SummaryTransformer`, for instance. I think you're mixing this up with forecasters.

>As descibed above, @ltsaprounis will remove that exceptions by implenting the extrapolation. So I dont see a need now to change anything regarding fit/transform.

No exceptions should be raised - can we not agree on an empty `fit` instead?

fkiraly:
> Also the fact that we currently have "what goes in comes out" is not compliant with that.

In the current base transformer design, if the input is `pd.Series` but the output is multivariate, `pd.DataFrame` is returned.

fkiraly:
@aiwalter, all, no need to close this - I'm happy to make the suggested changes if you don't have time to.
We need to comply with the base interface, that's all!

fkiraly:
@aiwalter, @eyalshafran, @ltsaprounis, kindly have a look here:
https://github.com/alan-turing-institute/sktime/pull/1773

where I've attempted to implement my suggestions.

fkiraly:
@aiwalter, I've merged this now, so it's in. We can discuss additional changes on top of this.

Issue #1194, opened 2021-07-23
[BUG] Fix timing bug in `test_evaluate_common_configs`


fkiraly:
Strange fail in `test_evaluate_common_configs`, from #1191 but unrelated to the PR.
Opening an issue for investigating it. I think this relates to code of @Aiwalter and @RNKuhns.

See details for error message:
<details>
```
=================================== FAILURES ===================================
__ test_evaluate_common_configs[scoring0-update-5-10-0-SlidingWindowSplitter] __

CV = <class 'sktime.forecasting.model_selection._split.SlidingWindowSplitter'>
fh = 0, window_length = 10, step_length = 5, strategy = 'update'
scoring = MeanAbsolutePercentageError(multioutput='uniform_average', symmetric=True)

    @pytest.mark.parametrize("CV", [SlidingWindowSplitter, ExpandingWindowSplitter])
    @pytest.mark.parametrize("fh", TEST_FHS)
    @pytest.mark.parametrize("window_length", [7, 10])
    @pytest.mark.parametrize("step_length", TEST_STEP_LENGTHS)
    @pytest.mark.parametrize("strategy", ["refit", "update"])
    @pytest.mark.parametrize(
        "scoring",
        [
            MeanAbsolutePercentageError(symmetric=True),
            MeanAbsoluteScaledError(),
        ],
    )
    def test_evaluate_common_configs(CV, fh, window_length, step_length, strategy, scoring):
        y = make_forecasting_problem(n_timepoints=30, index_type="int")
        forecaster = NaiveForecaster()
        cv = CV(fh, window_length, step_length=step_length)
    
        out = evaluate(
            forecaster=forecaster, y=y, cv=cv, strategy=strategy, scoring=scoring
        )
>       _check_evaluate_output(out, cv, y, scoring)

CV         = <class 'sktime.forecasting.model_selection._split.SlidingWindowSplitter'>
cv         = SlidingWindowSplitter(fh=0, step_length=5)
fh         = 0
forecaster = NaiveForecaster(sp=1, strategy='last', window_length=None)
out        =    cutoff  fit_time  ...  pred_time  test_MeanAbsolutePercentageError
0    12.0  0.000662  ...   0.010524             ...                   0.181488
4    32.0  0.003227  ...   0.009938                          0.874410

[5 rows x 5 columns]
scoring    = MeanAbsolutePercentageError(multioutput='uniform_average', symmetric=True)
step_length = 5
strategy   = 'update'
window_length = 10
[...]

PR #1680, opened 2021-12-02
Replace time.time with time.perf_counter

mloning:
#### Reference Issues/PRs
Fixes https://github.com/alan-turing-institute/sktime/issues/1194

* replaces `time.time` with `time.perf_counter`

### DIFF
--- a/sktime/forecasting/model_evaluation/_functions.py
+++ b/sktime/forecasting/model_evaluation/_functions.py
@@ -104,19 +104,19 @@ def evaluate(
         fh = ForecastingHorizon(y_test.index, is_relative=False)
 
         # fit/update
-        start_fit = time.time()
+        start_fit = time.perf_counter()
         if i == 0 or strategy == "refit":
             forecaster = clone(forecaster)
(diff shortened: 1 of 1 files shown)


fkiraly (changes requested):
@RNKuhns, @mloning, can you kindly explain why this fixes #1194?
I don't see the link to the bug, and looking at your discussion in #1194 isn't very concrete there.

mloning:
See https://github.com/alan-turing-institute/sktime/issues/1194#issuecomment-986061801

fkiraly:
No changes, just the explanation why this fixes the bug.
Typical sequence is reproduce, understand, fix. I need to check that I've understood the fix.

mloning:
Sure, I've added a short explanation on the issue. A "comment" review would have been better in this case I guess.

fkiraly (approved):
Got it, explanation makes sense.

fkiraly:
> A "comment" review would have been better in this case I guess.

Well, a "bugfix" is changes to the code and sufficient evidence that it's fixed, and I've asked for the latter.

PR #1681, opened 2021-12-03
Fix ColumnwiseTransformer example

mloning:
* fix example of ColumnwiseTransformer


fkiraly (approved):
looks good, minor change which makes the example clearer.

It's not a bugfix though, right?

PR #1695, opened 2021-12-04
[FIX] Add missing init files

mloning:
#### Reference Issues/PRs
Related to #1661, #1699

* add missing init files identified via: `find sktime -type d '!' -exec test -e "{}/__init__.py" ";" -not -path "**/__pycache__" -not -path "**/datasets/data*" -print`

### DIFF
--- a/sktime/alignment/tests/__init__.py
+++ b/sktime/alignment/tests/__init__.py
@@ -1,0 +1,2 @@
+# -*- coding: utf-8 -*-
+"""Tests."""
--- a/sktime/base/tests/__init__.py
+++ b/sktime/base/tests/__init__.py
@@ -1,0 +1,2 @@
+# -*- coding: utf-8 -*-
+"""Tests."""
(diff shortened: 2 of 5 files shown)


chrisholder (dismissed):

mloning:
This is failing as expected because these tests weren't picked up previously.

mloning:
Should be good to go now since #1661 has been merged

fkiraly:
> `find sktime -type d '!' -exec test -e "{}/__init__.py" ";" -not -path "**/__pycache__" -not -path "**/datasets/data*" -print`

Excellent! Could we also add this as a CI check? I.e., that there are no missing init files? A check based on this line should work?
FYI @lmmentel

fkiraly (approved):
Looks good - should merge once tests pass.

mloning:
Can we wait before we merge to see if #1699 fails without the init files?

fkiraly:
of course

fkiraly:
merged, since #1699 failed as expected.

PR #1699, opened 2021-12-04
[MNT] Fail CI on missing init files

mloning:
#### Reference Issues/PRs
#1695 

Adds check to fail unit testing if init files are missing in modules


fkiraly (approved):
Looks good, although the error message would be more helpful if you tell the developer *which* init file is missing *where*.

mloning:
Yes just added this now, let's see if this actually fails before we merge #1695

mloning:
Ok, 4ad9060 now fails with the expected error message:

Screenshot 2021-12-05 at 11 25 00

So after merging #1695 and syncing with main, this should be good to go.

fkiraly:
#1695 is now merged, the init files are in.
After update from main, this should no longer fail.

fkiraly (approved):
It indeed does no longer fail, so ready for merge.

PR #1706, opened 2021-12-06
[ENH] transformer base class to allow multivariate output if input is always univariate

### LINKED ISSUE #1697: [BUG] transformer multi-column pd.DataFrame output conversion fails
**Describe the bug**
It seem like series-to-series transformers cannot return multiple columns when passing a pd.Series as input. 

**To Reproduce**
```python
from sktime.transformations.base import BaseTransformer
import numpy as np
import pandas as pd

class MyTransformer(BaseTransformer):
    _tags = {
        "scitype:transform-output": "Series",
        "scitype:transform-labels": "None",
        "scitype:instancewise": True,
        "univariate-only": False,
        "handles-missing-data": True,
        "X_inner_mtype": "pd.DataFrame",
        "y_inner_mtype": "None",
        "X-y-must-have-same-index": True,
        "enforce_index_type": None,
        "fit-in-transform": False,
        "transform-returns-same-time-index": True,
        "skip-inverse-transform": True,
    }

    def _fit(self, X, y=None):
        return self
    
        return pd.concat([X, X], axis=1)
    
transformer = MyTransformer()
transformer.fit_transform(pd.DataFrame(np.random.normal(size=(10,)))) # works
transformer.fit_transform(pd.Series(np.random.normal(size=(10,))))  # fails
```
> ValueError: input must be univariate pd.DataFrame, with one column

**Additional context**
This is relevant for "featurizers" as in #1693.

**Version**
main at commit 8ca23bd85e60d

fkiraly:
Fixes #1697 by introducing the following base class behaviour: returning `pd.DataFrame` output if transform output is multivariate but input was `pd.Series`.

### DIFF
--- a/sktime/transformations/base.py
+++ b/sktime/transformations/base.py
@@ -474,7 +474,7 @@ class BaseTransformer(BaseEstimator):
             check_res = check_is_mtype(
                 ["pd.DataFrame", "pd.Series", "np.ndarray"], return_metadata=True
             )
-            if not check_res["is_univariate"] and X_input_mtype == "pd.Series":
+            if not check_res[2]["is_univariate"] and X_input_mtype == "pd.Series":
                 X_output_mtype = "pd.DataFrame"
             else:
                 X_output_mtype = X_input_mtype
--- a/sktime/transformations/base.py
+++ b/sktime/transformations/base.py
@@ -471,10 +471,10 @@ class BaseTransformer(BaseEstimator):
         if output_scitype == "Series":
             # if the transformer outputs multivariate series,
             #   we cannot convert back to pd.Series, do pd.DataFrame instead then
-            check_res = check_is_mtype(
+            _, _, metadata = check_is_mtype(
                 ["pd.DataFrame", "pd.Series", "np.ndarray"], return_metadata=True
             )
-            if not check_res[2]["is_univariate"] and X_input_mtype == "pd.Series":
+            if not metadata["is_univariate"] and X_input_mtype == "pd.Series":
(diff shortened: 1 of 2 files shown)


sktime/transformations/base.py
```diff
@@ -469,9 +469,18 @@ def transform(self, X, y=None, Z=None):
             Xt = convert_Panel_to_Series(Xt)
 
         if output_scitype == "Series":
+            # if the transformer outputs multivariate series,
+            #   we cannot convert back to pd.Series, do pd.DataFrame instead then
+            check_res = check_is_mtype(
+                ["pd.DataFrame", "pd.Series", "np.ndarray"], return_metadata=True
+            )
+            if not check_res[2]["is_univariate"] and X_input_mtype == "pd.Series":
```
mloning, line 477: this may make it a bit more readable
```suggestion
            _, _, metadata = check_is_mtype(
                ["pd.DataFrame", "pd.Series", "np.ndarray"], return_metadata=True
            )
            if not metadata["is_univariate"] and X_input_mtype == "pd.Series":
```

mloning (approved):
Haven't tested it, but changes look fine.

sktime/transformations/base.py
fkiraly: I've thought about it, but the first option is more stable against addition of further return args.
Though that may be too defensive, I agree your variant is more readable. Will change to it.

fkiraly:
appveyor timing out, but otherwise ok.
Side note: suggested code change by @mloning had a typo/bug. Fixed that but "morally" accepted the code change (i.e., the correct version of it).

PR #1709, opened 2021-12-06
[BUG] CIF Bugfix

### LINKED ISSUE #1690: [BUG] CanonicalIntervalForest breaks with CIT tree
**Describe the bug**
When running the below code snippet, I get: 

> `AttributeError: '_TreeNode' object has no attribute '_predict_proba_cif'`

**To Reproduce**
```python
import numpy as np
from sklearn.model_selection import train_test_split

from sktime.classification.compose import ColumnEnsembleClassifier
from sktime.classification.interval_based import CanonicalIntervalForest
from sktime.datasets import load_basic_motions
from sktime.transformations.panel.compose import ColumnConcatenator

X, y = load_basic_motions(return_X_y=True)
X_train, X_test, y_train, y_test = train_test_split(X, y, random_state=42)

steps = [
    ("concatenate", ColumnConcatenator()),
    ("classify", CanonicalIntervalForest(n_estimators=100, base_estimator="CIT")),
]
clf = Pipeline(steps)
clf.fit(X_train, y_train)
clf.score(X_test, y_test)```
```

**Additional context**
There's a [question on TSF on StackOverflow](https://stackoverflow.com/questions/70212188/how-to-know-from-which-interval-of-the-input-the-features-used-in-sktimes-times) in which I wanted to show CIF as an alternative but got the above error.

MatthewMiddlehurst:
Fixes #1690 and adds a test to prevent similar issues.


TonyBagnall:
does it need check_X_y? That should be handled in the BaseClassifier?

MatthewMiddlehurst:
It extends sklearn BaseEstimator, because it is not a time series classifier. This and RotationForest currently only accept 2d numpy arrays (or 3d numpy arrays with 1 dimension).

I'm not even sure if they are visible publicly, they are in contrib and the files are private.

TonyBagnall:
> It extends sklearn BaseEstimator, because it is not a time series classifier. This and RotationForest currently only accept 2d numpy arrays (or 3d numpy arrays with 1 dimension).
> 
> I'm not even sure if they are visible publicly, they are in contrib and the files are private.

ah ok, no problem. As an aside, I think rotation forest could come out of contrib, just not sure where to put it!

TonyBagnall (approved):

fkiraly (dismissed):
Some questions, could be resolved by discussion:

* as said, this does not conform with the TSC template but is in fact an sklearn classifier. You say it is not a TSC. But how does this fit with the docstring which says it *is* TSC?
* is adding a test which tests statistical performance a good idea? If yes, why do we add this only for CIT an not for estimators in general? What is specific about CIT?

MatthewMiddlehurst:
> as said, this does not conform with the TSC template but is in fact an sklearn classifier. You say it is not a TSC. But how does this fit with the docstring which says it is TSC?

Which part of the docstring indicates it is a time series classifier? I think it is fine, but if it is confusing from a different perspective, I can change it. The original name for the tree from its paper was Time Series Tree, but I renamed it to its current name as the previous was pretty general and confusing.

It is pretty much only meant as a base classifier for two of our interval_based ensembles, which is why it is private and relegated to contrib.

> is adding a test which tests statistical performance a good idea? If yes, why do we add this only for CIT an not for estimators in general? What is specific about CIT?

Fair enough. Don't really need to compare accuracies to determine whether the CIF specific functions work. Removed it.

TonyBagnall:
you can remove return_X_y=True from the loading. We will need to update this when check_X get deprecated, probably 0.11 now. The comments seem fine, We need to resolve what to do with both this and RotationForest, they should not stay in contrib, but we can do that in another PR, I'll raise an issue

MatthewMiddlehurst:
I think return_X_y is going to be similar for all tests, will deal with it in the classification test refactor we talked about in the pipeline PR.

@fkiraly Are you fine with this PR? If the docs are still an issue, I can make some changes for parts you suggest are unclear.

TonyBagnall (approved):

chrisholder (approved):

chrisholder (approved):

PR #1711, opened 2021-12-06
Fixes prophet bug concerning the internal change of exogenous X

kejsitake:
#### Reference Issues/PRs

Fixes https://github.com/alan-turing-institute/sktime/issues/1673

- Reproduced the error described in  https://github.com/alan-turing-institute/sktime/issues/1673
- Creates a copy of X in _Prophet_Adapter before changes are made to it.
- Verified that the code snippet int he issue document a bug doesn't throw the error after this change.
- Checks if X contained a column ds were already implemented in _merge_X, so I didn't implement a new one

No

### DIFF
--- a/sktime/forecasting/base/adapters/_fbprophet.py
+++ b/sktime/forecasting/base/adapters/_fbprophet.py
@@ -42,8 +42,6 @@ class _ProphetAdapter(BaseForecaster):
         -------
         self : returns an instance of self.
         """
-        if X:
-            X = X.copy()
         self._instantiate_model()
         self._check_changepoints()
         y, X = check_y_X(y, X, enforce_index_type=pd.DatetimeIndex)
@@ -65,6 +63,7 @@ class _ProphetAdapter(BaseForecaster):
 
         # Add regressor (multivariate)
         if X is not None:
+            X = X.copy()
             df, X = _merge_X(df, X)
             for col in X.columns:
                 self._forecaster.add_regressor(col)
@@ -103,8 +102,6 @@ class _ProphetAdapter(BaseForecaster):
         Exception
             Error when merging data
         """
-        if X:
-            X = X.copy()
         self._update_X(X, enforce_index_type=pd.DatetimeIndex)
 
         fh = self.fh.to_absolute(cutoff=self.cutoff).to_pandas()
@@ -114,6 +111,7 @@ class _ProphetAdapter(BaseForecaster):
 
         # Merge X with df (of created future DatetimeIndex values)
         if X is not None:
+            X = X.copy()
             df, X = _merge_X(df, X)
 
         # don't compute confidence intervals if not asked for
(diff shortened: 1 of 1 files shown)


aiwalter (commented):
There are already below 2 `if X is not None`. Can we include it there? @fkiraly should we even put it into base class and also copy `y`?

fkiraly (changes requested):
@kejsitake, thanks!

This is in-principle a fix, but looking at the code I noticed there is another problem. If the series passed *does* have the `ds` column, the forecaster raises an error, which it shouldn't. Also, the `X.copy()` are hard to keep track of.

I would address this as follows:
* write a functions `df_to_prophet` which takes `X` and does not have any side effects on it.
* `df_to_prophet` moves index of `X` to `ds` column. If a `ds` column is already present, it renames it to a name that doesn't exist yet in the columns. For instance, the longest column name plus an underscore or similar.
* the function is applied whenever `X` is passed.

aiwalter (approved):

aiwalter:
I think I would not rename a column, thats not intuitive to me and unexpected. Especially bcz that ds column probably contains an index already, which will then raise an error but for a different column name?

fkiraly:
> I think I would not rename a column, thats not intuitive to me and unexpected. Especially bcz that ds column probably contains an index already, which will then raise an error but for a different column name?

Can you elaborate?

I thought prophet wants the index in the `ds` column, not in the index.
All `sktime` forecasters work fine if a column is already named `ds` (a variable column). To make prophet work fine in that case too, an existing `ds` has to be renamed to make place for the index which is renamed to `ds`.

Otherwise the prophet forecasters fails when a `ds` column is already present?

fkiraly:
> Especially bcz that ds column probably contains an index already, which will then raise an error but for a different column name?

Could you explain: what would raise an error if `ds` is there already?

fkiraly:
signing off for today - and I'd be fine with the fix as is, @kejsitake, @aiwalter.
It fixes the bug that was reported, although not the problem what happens if `X` already has a `ds` column.
I'd say to take the pressure off the 0.9.0, let's just merge it like this, and open an issue that captures the problem if `ds` is already a column in `X`.

kejsitake:
I created another bug for the issue that remains. I can work on it, but are not very sure with which approach we want to move forward with.

aiwalter:
thx @kejsitake ! :) renaming an existing columns `ds` to sth e.g. `my_ds` could make issues when `ds` already contains an index. `fbprophet` users might do this by accident as they are used to this format. So that means we would have two index columns then and I guess `Prophet` would raise an error (and should) that there is a non-numeric column. This column has now name `my_ds` and users does not understand where this is coming from. So if we are already checking if `ds` exists and raise an error, I dont see a need for a change.

fkiraly:
>I created another bug for the issue that remains. I can work on it, but are not very sure with which approach we want to move forward with.

Yes, I think that's the best thing. Let's merge the fix for *this bug*, and have a separate discussion how we fix the remaining issue - it's closely related but distinct.

fkiraly (approved):
bug should be fixed now; will have to fix and refactor further (see discussion above) but it's a separate PR

fkiraly:
appveyor timed out but didn't fail

PR #1721, opened 2021-12-08
[ENH] New transformation based pipeline classifiers

### LINKED ISSUE #1063: [ENH] time series classification: shorthands for common sktime pipelines including TsfreshClassifier

MatthewMiddlehurst:
Implements new feature based pipelines for summary statistics, random intervals (with a new transformer to generate them) and a TSFresh based pipeline from an upcoming workshop paper by me and @TonyBagnall. We can probably close #1063 after this unless there are any objections.

There is quite a lot going on here, but the majority of new lines and file changes are just tests.

No

The new estimators, they are relatively simple and most are basic transformer > classifier pipelines.
Files of interest are:

sktime/classification/feature_based/_fresh_prince.py
sktime/classification/feature_based/_random_interval_classifier.py
sktime/classification/feature_based/_summary_classifier.py
sktime/transformations/panel/random_intervals.py


TonyBagnall (approved):
this looks fine, and we can work it into the new classification notebook I'm planning with pure pipeline examples and the static implementations. Needs to wait until after the release of 0.9 though, which should happen tonight (weds 8/12/21)

fkiraly (changes requested):
Question: why are we introducing n classifiers following the pattern [transformer][classifier], but not a generic pipeline object that implements this pattern?

Makes sense to have shorthands, but it would be more robust if the shorthand is a shorthand for the composable pipeline?

MatthewMiddlehurst:
Is the sklearn Pipeline not a suitable generic object that follows this pattern? Unsure if a middle step is necessary. 

```
fresh_prince = FreshPRINCE()
```
Is a shorthand for:

```
fresh_prince = Pipeline([
    ("transform", TSFreshFeatureExtractor(default_fc_parameters="comprehensive")),
    ("classifier", RotationForest(n_estimators=200)),
])
```

We could have a TransformClassifier:

```
fresh_prince = TransformClassifier(
    transform=TSFreshFeatureExtractor(default_fc_parameters="comprehensive"),
    classifier=RotationForest(n_estimators=200),
)
```

But I don't think it would add much value beyond people being unfamiliar with pipelines, possibly. I think in the upcoming example rework for classification, more focus will be put on them.

It should be pretty easy to put together, however. I don't mind doing so for this PR if people think it is a good addition.

fkiraly:
@MatthewMiddlehurst, that's almost what I mean.

Though what I mean is one of the following two

(a) a class factory along the lines of

```python
FreshPRINCE = PipelineFactory([
    ("transform", TSFreshFeatureExtractor(default_fc_parameters="comprehensive")),
    ("classifier", RotationForest(n_estimators=200)),
])
```

or (b) internally you make use on `sklearn` `Pipeline`, instead of re-implementing and copy-pasting the pipeline logic three times, e.g.,

```python
def __init__(stuff):
    more_stuff
    self.pipeline_ = Pipeline([
        ("transform", TSFreshFeatureExtractor(default_fc_parameters="comprehensive")),
        ("classifier", RotationForest(n_estimators=200)),
    ])

def _fit(self, X, y):
    foo
    bar
    self.pipeline_.fit(X, y)
    return self
````
or similar.

**Comment by fkiraly** [comment]
looking at this again, though - I wonder whether it's possible to avoid the boilerplate code in there. 

So I'd downgrade my above comments to more something for consideration, not a blocker.

File: sktime/classification/feature_based/tests/test_random_interval_classifier.py [hunk 1]
```diff
@@ -0,0 +1,220 @@
...
+            0.2,
+            0.8,
+            0.0,
+            0.0,
+        ],
+    ]
+)
+
+
+# def print_array(array):
```
fkiraly, line 174: please remove commented out code

sktime/classification/feature_based/tests/test_summary_classifier.py
```diff
@@ -0,0 +1,197 @@
...
+            0.1,
+            0.9,
+            0.0,
+            0.0,
+        ],
+    ]
+)
+
+
+# def print_array(array):
```
fkiraly, line 161: please remove commented out code

fkiraly (changes requested):
my requests would be:
* please remove dead/commented code
* consider the "boilerplate" issue - if you can argue that there are no DRYer alternatives, I'm happy

MatthewMiddlehurst:
Pipelines get a bit restrictive when you want to add additional functionality. Or in some cases just would not work, i.e. for the summary classifier the default multivariate output would not be compatible for any sklearn classifier.

For sure, some of them could be Pipeline wrappers though.

sktime/classification/feature_based/tests/test_random_interval_classifier.py
MatthewMiddlehurst: That is the code for generating the results we compare against, left in with the reasoning that it would be very annoying to have to re-implement it every time the results have to be redone after a legitimate change.

I'm against removing it completely, but we could consider formalising it a bit more, i.e. relocating the print function to utils and cleaning up the main portion a bit.

TonyBagnall:
> Question: why are we introducing n classifiers following the pattern [transformer][classifier], but not a generic pipeline object that implements this pattern?
> 
> Makes sense to have shorthands, but it would be more robust if the shorthand is a shorthand for the composable pipeline?

we already have a pipeline, or do you mean a wrapper for a pipeline where transformer and classifier are specified in constructor? Something like a PipelineClassifier? Seems a bit superfluous and I thought you liked using pipelines directly, but I see no reason why not have one.

I agree its best to not have commented out code. Maybe put it all in a directory in contrib called regenerate tests or somesuch?

MatthewMiddlehurst:
RE: Commented out sections for tests, that is a bigger issue than this PR (same pattern follows for most tests), but I don't mind doing it in a separate PR with all the others.

TonyBagnall (dismissed):

TonyBagnall:
> RE: Commented out sections for tests, that is a bigger issue than this PR (same pattern follows for most tests), but I don't mind doing it in a separate PR with all the others.

lets discuss at the dev meeting, I'll put together an agenda

TonyBagnall (approved):

TonyBagnall (dismissed):

TonyBagnall:
@fkiraly could you rereview in the light of comments above please? ta

TonyBagnall:
we agreed at the dev meeting to centralise the commented out code in a separate PR. @fkiraly does this answer your issues?

fkiraly:
> @fkiraly does this answer your issues?

Yes, happy if there's a separate issue that says what needs be doing.

What did you decide on the boilerplate/DRY issue?

MatthewMiddlehurst:
I still don't really see an issue with the current set-up, issues such as SummaryClassifier multivariate output and extra functionality like train estimates in FreshPrince would not lend well to a generic pipeline.

Im not against having a generic pipeline wrapper for users. Using such a wrapper in the current pipelines would require a bit of unwrapping for the above cases, however.

sktime/classification/feature_based/_fresh_prince.py
```diff
@@ -0,0 +1,206 @@
...
+from sktime.classification.base import BaseClassifier
+from sktime.contrib.vector_classifiers._rotation_forest import RotationForest
+from sktime.transformations.panel.tsfresh import TSFreshFeatureExtractor
+from sktime.utils.validation.panel import check_X_y
+
+
+class FreshPRINCE(BaseClassifier):
+    """Fresh Pipeline with RotatIoN forest Classifier.
+
+    This classifier simply transforms the input data using the TSFresh [1]
```
chrisholder, line 21: I believe you need an _ to reference so it should be:

`the input data using the TSFresh [1]_`

chrisholder (approved):

TonyBagnall:
so to summarise, we thought a wrapper containing a configurable pipeline was a layer of wrapping too far, and it is in fact preferable to encourage users to use a Pipeline, since sklearn users are familiar with them. There is not that much code repetition and not all classifiers could use the base class.

fkiraly (approved):
questions answered, looks fine to me now.

Can you please remove commented out code before you merge?

TonyBagnall:
> questions answered, looks fine to me now.
> 
> Can you please remove commented out code before you merge?

this is deferred to here https://github.com/alan-turing-institute/sktime/issues/1765

PR #1727, opened 2021-12-08
[ENH] Added Featureizer series transformer

### LINKED ISSUE #1443: [ENH] Adding a Featureizer as _SeriesToSeriesTransformer
When doing feature engineering, we want to calculate new features by means of any transformer and add them to existing ones. For this there is a need for a new transformer composition called `Featurizer` which gets any `_SeriesToSeriesTransformer` and appends the transformed data as new columns instead of overwriting the input data. `Featurizer` could either get a single transformer or a list of transformers (advanced implementation), this two cases look as follows.

Related: #1174

A clear and concise description of what you want to happen, ideally taking into consideration the existing toolbox design, classes and methods.

Use case when getting a single transformer:
```python
pipe = ForecastingPipeline(steps=[
    ("imputer", Imputer(method="mean")),
    ("minmaxscaler", Featurizer(TabularToSeriesAdaptor(MinMaxScaler(feature_range=(1,2))))),
    ("boxcox", Featurizer(TabularToSeriesAdaptor(PowerTransformer(method="box-cox")))),
    ("forecaster", AutoARIMA(suppress_warnings=True))])
pipe.fit(y=y_train, X=X_train)
```
In this case, the `boxcox` gets also the newly added features from `minmaxscaler`, whereas below the features are appended in parallel and not in sequence.
Use case when getting a list of transformers:
```python
pipe = ForecastingPipeline(steps=[
    ("imputer", Imputer(method="mean")),
    ("featurizer", Featurizer(
        ("minmaxscaler", TabularToSeriesAdaptor(MinMaxScaler(feature_range=(1,2)))),
        ("boxcox", TabularToSeriesAdaptor(PowerTransformer(method="box-cox"))),
    )),
    ("forecaster", AutoARIMA(suppress_warnings=True))])
pipe.fit(y=y_train, X=X_train)
```
The `Featurizer` would by default add the new columns by means of a suffix which is derived from the name of the transformer. Otherwise there could be a praram `suffix` and `prefix` so that user can also define this optionally.

**Additional context**
Pipelining

Any opinions @RNKuhns @mloning @fkiraly @ViktorKaz

aiwalter:
#### Reference Issues/PRs
closes #1443 

see above issue

##### For all contributions

##### For new estimators

#### to do's:
- [ ] Add tests
- [x] Add check that `X_pred` is either equal to `X_train` or continuous from `X_test` and without gap anywhere.
- [x] Get rid of param `shift`. Maybe add fh as param


fkiraly:
sorry, I wasn't aware that this was there - only saw the link from the recent conversation here #2459

I think this is a nice idea! Though, how would it work technically?
That the output should be moved to `X` is not an intrinsic property of the transformer, but part of the specification given to the pipeline.

How would this transformer solve that? E.g., can you give a code example around it?
Or is there logic missing in the pipeline that needs to interact with this?

aiwalter:
@fkiraly pls give it a review. I will add following before merge:
- docstring example
- adding class to docs

aiwalter:
Ok I need to fix sth in the logic first.

PR #1730, opened 2021-12-08
Add deprecation message to ambiguous arg reference in `ForecastingPipeline`, removed `Z` kwarg

fkiraly:
Update: this is not a bug, as originally thought.

Following some discussion, I've added a deprecation message that was missing in `ForecastingPipeline` and changed the `Z` kwarg to first arg, to prevent the code breaking in 0.10.0.

---

I think the transformers refactor introduced a bug:
in series-to-series transformers, the order of arguments was not changed, but the names were.
So args calls are fine since downwards compatible, but kwargs calls may break.

One reference in the `ForecastingPipeline` was kwargs, which makes pipelines with any transformers that are not yet refactored break.

This PR fixes the issue, by replacing the kwargs reference with an args reference.

I've also searched the repository, there are no other kwargs references to the sub-set of transformers that had `Z` previously as the first argument.

### DIFF
--- a/sktime/forecasting/compose/_pipeline.py
+++ b/sktime/forecasting/compose/_pipeline.py
@@ -203,7 +203,7 @@ class ForecastingPipeline(_Pipeline):
             # transform X
             for step_idx, name, transformer in self._iter_transformers():
                 t = clone(transformer)
-                X = t.fit_transform(X, y)
+                X = t.fit_transform(X=X, y=y)
                 self.steps_[step_idx] = (name, t)
 
         # fit forecaster
@@ -240,6 +240,8 @@ class ForecastingPipeline(_Pipeline):
         if self._X is not None:
             # transform X before doing prediction
             for _, _, transformer in self._iter_transformers():
+                # todo: deprecation in 0.10.0
+                # add kwarg X= after deprecation of old trafo interface
                 X = transformer.transform(X)
 
         return forecaster.predict(fh, X, return_pred_int=return_pred_int, alpha=alpha)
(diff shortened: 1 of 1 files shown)


fkiraly:
FYI @aiwalter, I think I messed up the args in the `ForecastingPipeline` - can you check whether I'm correct with the bug and the fix?

aiwalter (approved):

aiwalter:
was this causing issues? I have just tested `main` with `ForecastingPipeline` and seemed to work fine although the old transformers  dont have `y`, i am wondering why it went without exception. Probably bcz you already changed the `fit_transform()` method also?

fkiraly:
Yes, I'm confused why this was not actually causing an issue.

The tests were usng "old style" transformers, and still didn't break. So maybe there is nothing to be fixed and I'm just imagining a bug?

fkiraly:
In any case, I've extended the testing framework to be able to test through multiple instances of a class, see https://github.com/alan-turing-institute/sktime/pull/1732.

I think this is going to be useful for pipelines where we can have different instance cases, and to achieve better case coverage in terms of class parameters. So far we were restricted to covering a single parameter case which is quite brittle.

fkiraly:
Hm, the tests in #1732 all run through. as well as locally, including pipelines with the `BoxCoxTransformer` inside.
So no problem after all?

aiwalter:
the `BoxCox` you fixed already in another PR right? Lets get this PR in anyway, as at some point in time we will get rid of  the old transformer interface, so pipelines will then already be fine.

fkiraly:
but the point is, they already seem fine and I don't fully understand why they don't break. My mental model of the code must be wrong, and that makes me fear that merging this PR might introduce bugs down the line. I'd rather take a moment and try to thing this situation through.

Perhaps a quick discussion in the dev meeting, with more eyes on it, may help.

fkiraly:
Feels a bit like @fkiraly from back then is starting at me, shaking heads and disappointed that I am so stupid...

fkiraly:
See update to PR message for what I did now.

aiwalter (approved):

PR #1594, opened 2021-11-08
[ENH] Refactor of classifier base class: input checking and conversion

TonyBagnall:
#### Reference Issues/PRs

https://github.com/alan-turing-institute/sktime/issues/1537

#### What does this implement/fix? Explain your changes.
this addresses some problems with check_X_y/check_X and integrates the capabilities tags into the new fit/predict structture. It does this in four stages
1. check the input is a valid data type: function check_classifier_input (utils.validation.panel)  
2. gets the data characteristics: missing values, multivariate and unequal length. function get_data_characteristics  (utils.validation.panel)  
3. checks the classifier can handle the data characteristics. member function check_capabilities (classification.base)
4. performs any conversions from DataFrame to array or vice versa member function convert_input (classification.base)

#### Does your contribution introduce a new dependency? If yes, which one?
no

#### What should a reviewer concentrate their feedback on?

whether to use datatypes conversion or not, whether the y conversions are ok, testing

PR #1745, opened 2021-12-11
[ENH] `Table` scitype and refactor of `convert` module

fkiraly:
This PR refactors the `datatypes` module and adds a `Table` scitype, details below.

This is in anticipation of two things:
* refactoring and extension of transformer tests
* moving conversions in classifiers to `datatypes`, my contribution to https://github.com/alan-turing-institute/sktime/pull/1594

The changes to the `datatypes` module:
* the `test_convert` tests are refactored - the six loops over fixtures (from, to, fixture, etc) are moved to `pytest_test_generate` where it belongs. This should make the code much easier to read.
* similarly, the individual conversions now appear as individual fixture combinations, which should make debugging much easier
* a `Table` scitype is added. The mtypes are `pandas` `DataFrame`, `numpy1D`, and `numpy2D`. This will be useful for checks and conversions as follows: to check the output of transformers with a `Primitives` output - this must be `DataFrame`; to check the output of classifiers - this must be `numpy1D`; for classifiers to accept the other two mtypes in fitting.
* a metadata field `has_nans` which checks whether the container has any nans
* for the `nested_univ` mtype, nan and equal spaced checks use @TonyBagnall's `_nested_dataframe_has_nans` and `_nested_dataframe_has_unequal` from #1594


fkiraly:
@chrisholder, @MatthewMiddlehurst, @TonyBagnall, would appreciate your review, as this will be a dependency of the classifier conversions.

@Lovkush-A, you might also be happy over how the `convert` tests have been refactored.

fkiraly:
quick question, @TonyBagnall, @chrisholder, @matthewmiddlehusrt:
in #1594, do you have a method that checks whether `nested_univ` data containers have nans?
If so, kindly point me to it, since I don't want to re-implement this...

fkiraly:
> If so, kindly point me to it, since I don't want to re-implement this...

Found and added.

TonyBagnall (approved):
Looks ok to me, but @chrisholder is better placed to comment, especially on the NaN checking

fkiraly:
Ultimately, we should whether the `check` functions correctly recover properties of a fixture.

That is, add functionality to the `examples` modules to specify what properties the fixtuers have, and then write a test to check.
Each mtype/property combination should appear (at least) twice, once with the property positive, and once as a negative.

fkiraly:
appveyor timed out, but seems fine. Merging so I can get to work more easily on dependencies.

PR #1748, opened 2021-12-12
[ENH] Testing for metadata returns of `check_is_mtype`

fkiraly:
This module continues the refactor in #1745.

Changes:
* refactor of tests in `datatypes/tests/test_check`
* added tests for metadata inference, i.e., `"is_univariate"`, `"has_nans"`, etc - `test_check_metadata_inference`
* added expected metadata outputs to all examples for `Series`, `Panel` and `Table`
* added more fixtures to cover more of the metadata conditions
* fixed bug - conversions for `Table` were not added to min conversion dictionary
* fixed bug - incorrect `"is_univariate"` inference for `nested_univ` type
* fixed bug - added scitype specification in now ambiguous `mtype` checks in forecasters and transformers (after introduction of  `Table`)

PR #1749, opened 2021-12-12
[ENH] BaseClassifier refactor to use datatypes checks/conversions, on top of PR 1594

fkiraly:
As discussed with @TonyBagnall, this PR upgrades the base classifier class refactor #1594 to use the existing checks and coercions in `datatypes` instead of the new, custom functions written by @TonyBagnall.

It is based on #1745 (merged) and #1748 (not yet merged), which in turn have some routines by @TonyBagnall from #1594 moved into the `datatypes` module (e.g., inferring the data characteristics for certain mtypes).

Note: this is a PR into the other PR, not into `main`.

A lot of files appear changed, that's because #1745 (merged in `main`) and #1748 (not merged in main) are not yet part of #1594. To see the "real" diffs (localized to `BaseClassifier` and small of size), #1748 needs to be merged with `main`, and #1594 needs to be updated.

Reviewers of this PR can hence focus exclusively on changes in `BaseClassifier` and the `base` module it is in.


TonyBagnall (changes requested):
thanks for this franz. You have combined the checking and conversions again, and I want them in separate functions. Personally I do not like the new tagsYou will also need to change all the

TonyBagnall (commented):
this comment is ambiguous to me, not sure what "minor coercions means
input checks and minor coercions on X, y

TonyBagnall:
is the Table "scitype" new? Where are the scitypes formally defined? Cant see it here https://github.com/alan-turing-institute/sktime/blob/main/examples/AA_datatypes_and_datasets.ipynb

TonyBagnall:
The use case for unequal length series I am working with ins nested dataframes, the same as we have used since the initial decision. Is this being deprecated?

TonyBagnall:
So I'm afraid I really just dont like this version. It introduces levels of complexity that require anyone who read this code understand the whole complex nature of the datatypes module, most of which assumes formal definitions which as far as I can see do not exist. What is the definition of the abstract data type Table, Series and Panel? The benefits of ease of extending to new allowed inputs seems too small to merit the whole complex edifice. How often are we going to change data structures? 99% of uses cases for classification are 2D or 3D numpy. Even if we do develop bespoke internal methods for handling it, padding or truncating is still going to be the most popular approach. The requirement to introduce a whole new mtype just to replace my simple but clear converting and checking just seems too much. 

We can discuss further, and maybe I can run some code but it currently doesn't run 
image
 
At the classifier dev meeting I urged all to form and express an opinion and hope others will join the discussion. However, I feel this might have to go to a vote, by which of course I will  stand. My group can obviously disagree with me, (they frequently do, particularly on engineering issues) and the vote can be anonymous. It would probably be better to vote sooner to avoid either of us doing work that is not needed.

fkiraly:
> I'm not sure I get your point about repeating in all classes, this is only in the base class so where is the repetition?

I said, repeating in all *methods* (not classes), i.e., `fit`, `predict`, `predict_proba`.

fkiraly:
> The use case for unequal length series I am working with ins nested dataframes, the same as we have used since the initial decision. Is this being deprecated?

I think there is intention to soft-deprecate, i.e., still support them (currently indefinitely) but make it easier to use more "natural" mtypes, i.e., avoid nesting data frames.

fkiraly:
> You are adding two levels of function call to the design, it is in fact more code not less. I guess I can concede the principle. However, when we start dealing with unequal length series in _fit we will need to store the unequal characteristics in the base class unless we either

I was adding only one, but I think you wanted the second one, i.e., to separate out checking and conversion?

fkiraly:
> 99% of uses cases for classification are 2D or 3D numpy

I don't think all users would agree with you. We want to leave our interfaces open to allow unequal length panels, and unequal sampling rates.
If you listen to user talk around the relevant types, you'll find issues and discussions repeatedly popping up.
https://github.com/alan-turing-institute/sktime/issues/1756
https://github.com/alan-turing-institute/sktime/issues/1737
https://github.com/alan-turing-institute/sktime/issues/1667
https://github.com/alan-turing-institute/sktime/issues/397
https://github.com/alan-turing-institute/sktime/issues/230
https://github.com/alan-turing-institute/sktime/discussions/1592

There's a chicken/egg problem here: if you restrict the template interface that this is never possible, people won't be able implement individual estimators. They also won't make PR to the base class if you're strictly against.

fkiraly:
> Even if we do develop bespoke internal methods for handling it, padding or truncating is still going to be the most popular approach

I think it is a bit a of a stretch, to make this assumption. In any case, at least we should implement these as transformers!

fkiraly:
> The requirement to introduce a whole new mtype just to replace my simple but clear converting and checking just seems too much.

I did this because it appears now in *two* places. In transformers, the `Table` type appears as the output of series-to-primitives and panel-to-primitives transformers. It was only one place, so I "jury-rigged" it by using the `Series` scitype.
Now it appears in a second place, too, that's why I introduced `Table`.

fkiraly:
> It would probably be better to vote sooner to avoid either of us doing work that is not needed.

You're aware you would be voting on duplicating functionality that already exists, within your module. You're voting on the DRY principle basically.

fkiraly:
> How often are we going to change data structures? 

The hope is, rarely - but when we change them, it should be able to easily do this in all modules simultaneously.
Let's say, we adopt `awkward-array` when it's ready. Do we want to add it in one place, in the `datatypes` module, or in every module that has decided to write their own input checking and conversion routines?

TonyBagnall:
> > How often are we going to change data structures?
> 
> The hope is, rarely - but when we change them, it should be able to easily do this in all modules simultaneously. Let's say, we adopt `awkward-array` when it's ready. Do we want to add it in one place, in the `datatypes` module, or, instead, once per module that has decided to write their own input checking and conversion routines?

well that would be at most one file per module, since the whole point of the fit/_fit is to factor that out. It would be literally a 10 min job to change BaseClassifier, BaseClusterer and BaseRegressor. Alternatively, we can move it up in to BaseEstimator. Changing there would be equivalent to changing datatypes. 

Another point is that the nested dataframe/awkward array is an uncommon use case. We are never really going to change from numpy for equal length problems.

TonyBagnall:
> > It would probably be better to vote sooner to avoid either of us doing work that is not needed.
> 
> You're aware you would be voting on duplicating functionality that already exists, to be duplicated within your favourite module. You're voting on the DRY principle basically. A key mission point of `sktime` is interoperability and compatibility between different learning tasks - see our roadmap statement.

which functionality already exists? This is to replace check_X_y. That already exists sure. Your PR brings in huge changes that currently do not exist, such as the Table "scitype"

TonyBagnall:
@fkiraly this fails tests, I cannot review this unless it passes

PR #1752, opened 2021-12-13
Update dataset headers

TonyBagnall:
updates the metadata for some of the classification problems, which were missing the equal length tag


TonyBagnall:
this changes no code and tests pass, I need to put this in,

PR #1757, opened 2021-12-14
data_io -> _data_io

TonyBagnall:
convert data loading of classification problems to return 2D numpy/3D numpy for univariate/multvariate equal length, and dataframes for unequal length


TonyBagnall:
going to try from scratch, import errors doing my head in

Issue #1737, opened 2021-12-10
[ENH] Train/test split and evaluation with disaggregated time series


khrapovs:
**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is.

I have a very specific forecasting use case that I solved elsewhere on my own. But only recently I discovered this fantastic library and obviously regret that it did not happen much earlier. I think that my use case could be a nice contribution.

The problem is the following. Suppose I have disaggregated data `y` and `X`. "Disaggregated" means that the forecasting frequency is, say, weekly, but the raw data is daily. My model extracts exogenous features from the past daily data (`X_train`) and uses an "auxiliary" model to forecast those features (`X_test`) which are in turn used in `.predict` method of the main forecaster. The current limitation of the library is that all the splitters and evaluation function assume that the data is already "regular", in my case weekly. This of course is not a problem if I only want to forecast once. But evaluation with `sktime.forecasting.model_evaluation._functions.evaluate` is impossible.

**Describe the solution you'd like**
A clear and concise description of what you want to happen, ideally taking into consideration the existing toolbox design, classes and methods.

I see two alternative solutions. Either a couple of new splitters alternative to existing ones, or a generalization of existing ones. The core change I need is replacing `np.arange` (e.g. in `train = np.arange(start - window_length, split_point)`) by pandas Series or DataFrame indices that are directly linked to a regular grid. Function `sktime.forecasting.model_evaluation._functions._split` will have to be changed as well by rewriting this arithmetic: `test = np.arange(test[0] - fh.min(), test[-1]) + 1`. At the first glance it seems enough.

Given these changes I can easily imagine a forecasting pipeline where one of the steps would be an auxiliary model forecaster that would generate `X` on-the-fly. Next step would be data aggregation into, say, weekly frequency. And finally a main usual time series model for a regular frequency data.

I would love to contribute the solution myself. But I write this issue first in order to prevent unnecessary work if my solution violates some high level vision of the core developers. Please, let me know how do I proceed. Thanks!

fkiraly:
OK, I've thought about this.

I think that indeed it fits really nicely with the design and the roadmap.

Historically, much of the functionality was written around the assumption of equal indices for all variables. It was a conscious decision by the core developer majority to cut corners there (and I won't say here who had which opinion in the discussion 😄).

[...]

PR #1758, opened 2021-12-14
[ENH] Generalize splitters to accept timedeltas (equally spaced)

khrapovs:
#### Reference Issues/PRs

https://github.com/alan-turing-institute/sktime/issues/1737

- All splitters now accept not only integer, but also `timedelta` or `pd.DateOffset` for `initial_window`, `window_length`, and `step_length` arguments. No regression was introduced. The core changes are localized to `sktime/forecasting/model_selection/_split.py`.
- All existing tests (`sktime/forecasting/model_selection/tests/test_split.py`) are only enriched for additional argument types. Test cases boil down to adding new types in `sktime/forecasting/tests/_config.py`. See variables `TEST_WINDOW_LENGTHS`, `TEST_INITIAL_WINDOW`, and `TEST_STEP_LENGTHS`.
- I could not resist and added type hints in `sktime/forecasting/base/_fh.py` although it is not necessary for the core change. I added type hints to `_split.py` first, but for completeness I needed type hint for forecasting horizon which triggered the change in `_fh.py`.
- `sktime/forecasting/model_evaluation/tests/test_evaluate.py` was only changed to preserve original tests. Other changes are automatic commit hooks. Same for `sktime/forecasting/model_selection/tests/test_tune.py`, `sktime/forecasting/tests/test_naive.py`, and `sktime/forecasting/tests/test_all_forecasters.py`.
- In `sktime/utils/validation/__init__.py` and `sktime/utils/validation/forecasting.py` I have added a few validations for new types.

none

The focus of this PR are the new types of above mentioned arguments. There are relatively a lot of changes, but all of them are triggered by this generalization.

#### Any other comments?

This is the first, but hopefully, logically complete batch of changes. I have a much bigger target in mind (see #1737). Personally, I prefer smaller PRs, hence I would prefer to stop here and address other required changes (like generalization of `ForecastHorizon`) in another PR.

##### For all contributions


fkiraly:
hello @khrapovs, how is it going? Do you need any help or feedback with this?

khrapovs:
Hello, Franz. Happy New Year! I decided to switch off for one week while I am on holidays with my family. Will come back to this issue upcoming week. It goes very well, I do not see any major obstacles. Just need a bit time to generalize a few functions. Eventually I would love to have it scrutinized by someone experienced, but right now my changes are too dirty.

khrapovs:
@fkiraly I think it is time to show this to someone. Of course, it does not implement everything I need for my use case, but this is already logically complete and I would love to proceed with a review here.

fkiraly (commented):
ok, quick question, would you be able to give a three-bullet-point summary of the changes you made so far?
(i.e., which functions, modules, classes are changed how?)

Is it correct that the changes are localized to `_split` and tests, plus adding type hints in every file you saw?

khrapovs:
@fkiraly I have updated the description of my PR at the top as you requested. I hope it is sufficiently clear.

fkiraly:
thanks, will probably have time on Fri or Mon

fkiraly (changes requested):
Excellent work! We've been too closely stuck with the integer indexing, so using timedeltas will be useful for irregularly spaced series!

Looks good and logical on detail inspection.
However, I think it is crucial for review to make clear what the behaviour of the various splitters is in "timedelta" mode - the devil is in the detail.

Would you hence be able to add the following to the docs:
* proper docstrings for key methods, such as `split`, `get_n_splits` and `_split`. They were missing before (not your fault), but now with timedeltas it's crucial to explain well what happens.
* for all splitters, describe in the class docstring what kind of splits these generate, especially if you have the general setting.

Further: would it be easy to make the code robust with respect to unequally sampled time series? Or difficult?

A minor question on the side:
We're looking at extending the forecasting functionality to panel and hierarchical time series. There, we would use `pd.DataFrame` with a `MultiIndex`, with time being the last level. May be better to extend that later, but do you foresee any substantial problems?

sktime/forecasting/base/_fh.py
```diff
@@ -149,7 +155,11 @@ def __new__(cls, values=None, is_relative=None):
             setattr(cls, method, _delegator(method))
         return object.__new__(cls)
 
-    def __init__(self, values=None, is_relative=True):
+    def __init__(
+        self,
+        values: Optional[Union[VALID_FORECASTING_HORIZON_TYPES]] = None,
```
fkiraly, line 160: I think you don´t need the `Optional` if the default is already `None` (then `Optional` is inferred)?

sktime/forecasting/model_evaluation/tests/test_evaluate.py
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3 -u
 # -*- coding: utf-8 -*-
 # copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+"""Testt evaluate."""
```
fkiraly, line 4: Maybe a bit more descriptive? "Tests for model evaluation module", etc

sktime/forecasting/model_selection/_split.py
```diff
@@ -158,42 +175,100 @@ def _get_end(y, fh):
...
-            f"the length of `y`. Found `window_length`={window_length}, `max(fh)`="
-            f"{fh_max}, but len(y)={n_timepoints}."
-        )
-
-    if initial_window is not None:
-        if initial_window + fh_max > n_timepoints:
+    if is_timedelta_or_date_offset(x=window_length):
+        if y.get_loc(min(y[-1], y[0] + window_length)) + fh_max > n_timepoints:
+            raise ValueError(
+                f"The `window_length` and the forecasting horizon are incompatible "
```
fkiraly, line 190: Here, I think it would be good to spell out the precise condition in the error message (which is violated)

sktime/forecasting/model_selection/_split.py
```diff
@@ -158,42 +175,100 @@ def _get_end(y, fh):
...
+                f"with the length of `y`. Found `window_length`={window_length}, "
+                f"`max(fh)`={fh_max}, but len(y)={n_timepoints}."
+            )
+    else:
+        if window_length + fh_max > n_timepoints:
             raise ValueError(
-                f"The `initial_window` and the forecasting horizon are incompatible "
-                f"with the length of `y`. Found `initial_window`={initial_window},"
+                f"The `window_length` and the forecasting horizon are incompatible "
+                f"with the length of `y`. Found `window_length`={window_length}, "
```
fkiraly, line 198: same here

sktime/forecasting/model_selection/_split.py
```diff
@@ -158,42 +175,100 @@ def _get_end(y, fh):
...
                 f"`max(fh)`={fh_max}, but len(y)={n_timepoints}."
             )
 
+    if initial_window is not None:
+        if is_timedelta_or_date_offset(x=initial_window):
+            if y.get_loc(min(y[-1], y[0] + initial_window)) + fh_max > n_timepoints:
+                raise ValueError(
+                    f"The `initial_window` and the forecasting horizon are "
+                    f"incompatible "
+                    f"with the length of `y`. Found `initial_window`={initial_window}, "
```
fkiraly, line 208: same, and below

fkiraly (changes requested):
Thanks for the docstrings!

I think these are currently major issue with the `_split` module, they need to become clearer.

I'd kindly request: in `BaseSplitter`, explain the type of object that this models - a collection of training and test splits. Explain which methods can be used to retrieve them.

In the classes that inherit from `BaseSplitter`, explain in addition what the split precisely will be,  depending on the parameters.
This should not just be examples, but a clear, explicit statement of what the splits are. Optimally, use mathematical or code formalism to give the description.

Otherwise, the changes look ok - but I'd give it another go and compare actual to intention, once you've written clear docstrings that describe what splitters you are *intending* to implement (so I can validate the timedelta case).

sktime/forecasting/base/_fh.py
khrapovs: you are right. I have reread one more time https://docs.python.org/3/library/typing.html#typing.Optional and removed `Optional`

sktime/forecasting/model_evaluation/tests/test_evaluate.py
khrapovs: Now it reads:
```
"""Tests for model evaluation module.

In particular, function `evaluate`, that performs time series
cross-validation, is tested with various configurations for correct output.
"""
```

sktime/forecasting/model_selection/_split.py
khrapovs: Added these two lines:
```
f"It is required that the window length plus maximum forecast horizon "
f"is smaller than the length of the time series `y` itself."
```
I have also extracted common error text into variables which are used twice.

sktime/forecasting/model_selection/_split.py
khrapovs: same here

sktime/forecasting/model_selection/_split.py
khrapovs: Slightly changed wording:
```
"The `initial_window` and `window_length` types are incompatible. "
"They should be either all timedelta or all int."
```

khrapovs:
@fkiraly As requested, I have introduced math formalism in the main section of the `BaseSplitter`. In order to avoid duplication I have referenced its docstring in concrete implementations and added missing details relative to the base. Please, let me know if it is still does not sound sufficiently clear in terms of intentions. Thanks again for a very detailed review!

fkiraly (approved):
@khrapovs, thanks! The changes to the docstring are quite good!
Good for approval, since the requirement was to understand what you were trying to do.

Couple of things could be improved, I leave this to the next PRs:
* explanation is not precise or a bit vagout about that a splitter encodes *multiple splits*, i.e., there is a collection of k, ell. If notation is introduced in the `BaseSplitter` for multiple k, ell, this can be referred to in concrete implementation
* explanation is very verbose, could be shortened
* explanation of methods and variables is a bit scattered, could be more list-like and concise
* you probably want to pass to a framing in terms of intervals, i.e., start/end of the window. In the end state, your observation time stamps may not need to be exactly at the start and end of a window that the splitter parameterizes

sktime/forecasting/model_selection/_split.py
fkiraly: yes, ok. May be improved by ensuring this is clearly linked to functino/method inputs, but good enough.

sktime/forecasting/model_evaluation/tests/test_evaluate.py
fkiraly: ok, thanks!

PR #1761, opened 2021-12-15
Numba based knn

TonyBagnall:
primarily this

1. swaps cython for numba
2. presents a possible solution to the fit/_fit inheritance problem.

(2) may be sub optimal in design, there are alternatives. The problem here is that KNeighborsTimeSeriesClassifier inherits from both sklearn's KNeighborsTimeSeriesClassifier and sktime's BaseClassifier. Both _KNeighborsClassifier and BaseClassifier have methods fit and _fit. If we switch to _fit in  KNeighborsTimeSeriesClassifier then  because of the method resolution order algorithm, it inherits the fit from _KNeighborsClassifier, which is what we dont want. We cannot (its seems) simply switch the inheritance order, since we need to call the _KNeighborsClassifier constructor. This hacky solution overrides fit/predict/predict_proba, but then simply calls the correct method in the BaseClassifier. I can see two alternatives

1) Make KNeighborsTimeSeriesClassifier private and only inherit from _KNeighborsClassifier, wrap it in a new class KNeighborsClassifier that extends BaseClassifier and contains _KNeighborsClassifier 
2) Figure out how to swap the inheritance order and call the constructors, or how to direct super to the correct fit method

I had a first stab at both of these but could not get it to work, feel free to revisit. (1) may seem the best solution, but it has several draw backs
1) It requires more changes to a lot of existing code, which will inevitably cause more problems
2) the _KNeighborsClassifier will not be in the sktime inheritance hierarchy at all. It is really a time series classifier but will present itself as something else.
3) its really no less hacky than this approach, where at least the hackiness is obvious and justifiable. 
4) I'm not sure on the impact on tags for the wrapper and contained object, it might get messy there, especially when moving towards unequal length capability


fkiraly:
is this still a draft, @TonyBagnall, or ready for review?
PR description says draft, but PR says it's not a draft.

TonyBagnall:
@fkiraly feel free to review, looking for opinions on how to do this, preferably with code :)

sktime/classification/distance_based/tests/test_time_series_neighbors.py
```diff
@@ -9,28 +9,19 @@
...
-# Distance measure  LCSSDistance -e "0.05" -ws "3" gets 137 correct out of 175,
-# but one tie, so expect 136 since sktime picks the first
-
+# expected correct on test set using default parameters.
 expected_correct = {
     "euclidean": 19,
     "dtw": 21,
     "wdtw": 21,
     "msm": 20,
-    "erp": 19,
```
chrisholder, line 32: This lower accuracy than before ill make a note and try fix

sktime/classification/distance_based/_time_series_neighbors.py
```diff
@@ -187,42 +136,15 @@ def fit(self, X, y):
...
-            X,
-            y,
-            enforce_univariate=not self._tags["capability:multivariate"],
-            coerce_to_numpy=True,
-        )
         # Transpose to work correctly with distance functions
         X = X.transpose((0, 2, 1))
 
+        if isinstance(self.distance, str):
+            self.metric = distance_factory(X[0], X[0], metric=self.distance, window=10)
```
chrisholder, line 143: Ill refactor this to use params

TonyBagnall:
@fkiraly can you express an opinion as to whether this will do for conforming to fit/_fit? I happy for it to be redone differently at a later date, but if you really dont like this way of doing it I will revert to its current state and just put in the cython version changes without update for fit/_fit

fkiraly (changes requested):
Reviewing this for interface compliance upon request by @TonyBagnall.

Re. `fit`, `predict`, etc - I think this is hacky but in-principle not wrong.
However, I feel we should do a `**kwargs` pass-through, that seems safer in terms of being upwards compatible (e.g., for additional arguments to `fit` etc). That is,

```python
def predict(self, **kwargs):
    return BaseClassifier.predict(self, **kwargs)
```

TonyBagnall:
> Reviewing this for interface compliance upon request by @TonyBagnall.
> 
> Re. `fit`, `predict`, etc - I think this is hacky but in-principle not wrong. However, I feel we should do a `**kwargs` pass-through, that seems safer in terms of being upwards compatible (e.g., for additional arguments to `fit` etc). That is,
> 
> ```python
> def predict(self, **kwargs):
>     return BaseClassifier.predict(self, **kwargs)
> ```

done that, happy if someone wants to adopt a better solution, but this is the simplest way of conforming for the problematic classifiers I feel, especially those no one wants to own.

fkiraly (approved):
Happy with this, interface-wise.

Before you merge, can you check what that new line added to `.gitignore` is, please? What is that about, is it necessary?

TonyBagnall:
its passed apveyor before, Im putting this in before it need another update, been trying for a week!

PR #1769, opened 2021-12-17
Fix bug related to exogenous X already containing a column named ds

kejsitake:
#### Reference Issues/PRs

First attempt at fixing https://github.com/alan-turing-institute/sktime/issues/1712

Implements a fix for when the X contains a variable named "ds", by copying the column to another column and renaming in the case the column is numeric and not an index.
This is done in order to be more consistent with the other forecasters, which don't throw an error if exogenous X contains a column "ds".


fkiraly:
the code by @kejsitake has already made it into `_ProphetAdapter` somehow, it is in `_merge_X`.

So, not sure why this PR is still there, probably a leftover or an earlier version of the PR that was actually merged.

selewaut:
[_merge_X](https://github.com/sktime/sktime/blob/main/sktime/forecasting/base/adapters/_fbprophet.py#L344) has commented out a raise ValueError. Is this ok, or should it be uncommented?

Issue #1105, opened 2021-07-04
Forecaster interface: getting in-sample forecasts, getting residuals


AngelPone:
**Is your feature request related to a problem? Please describe.**
I think there should be a method or property to obtain in-sample fitted value or residuals, which are important in many forecasting methods, such as forecast combination.

**Describe the solution you'd like**

Add a method or property to `BaseForecaster` and it should be implemented for every forecaster.

**Additional context**
A traditional forecast combination method based on variance of in-sample residuals.

fkiraly:
Ok, I think there are two requests in there, and a proposed way of addressing these.

The two requests: residuals and in-sample values.

Regarding residuals, I fully agree that there should be an easy way to return them!
But, I don't think this is a concern of the `BaseForecaster`.

There also already is a way to obtain residuals from a forecaster, it's the `Detrender` - simply detrend with the forecaster you would like to get residuals of - it's a transformer, and the output gives you the residuals.
Perhaps we should say this in the forecasting tutorial, since it's not obvious from the name "detrender"?

Regarding in-sample fitted values: Is this not what you would get when you run `predict` at the values that have already been seen? Simply pass a fixed forecasting horizon with the indices that you've observed.

`forecaster.fit(y).predict(fh=y.index)` or `forecaster.fit(y, fh=y.index).predict()`. Does this work for you?

What I'm wondering, should there perhaps be a `fit_predict(y)`, which, if `fh` is not passed, defaults to in-sample forecasts, i.e., `fh=y.index`?

Feedback on whether it is what you need, or whether it's "obvious" would be appreciated.

fkiraly:
for forecast combination, you may like to look at the `HedgeExpertEnsemble`, it may do what you want, i.e., re-weight depending on errors or forecast residuals.

Would be nice to hear from you whether you think that's what you need or not, and whether it should be made clearer in the tutorial.

RNKuhns:
@fkiraly and @AngelPone outside of the forecast combination case, I could see it being useful to be able to expect the residuals from a model. For example, to see if they have a clear pattern (suggesting the model doesn't fit well) or run a statistical test on them (e.g. are they normally distributed in the case of models like Arima that make this assumption).

Adding a convenience method would be pretty straight-forward -- something like:

```python
class BaseForecaster(BaseEstimator):
    def insample_residuals(self):
        self.check_is_fitted()
        if self.get_tag("requires-fh-in-fit", False):
[...]

PR #1770, opened 2021-12-17
[ENH] Forecaster convenience method to return forecast residuals

fkiraly:
This PR implements a frequently requested convenience method (`predict_residuals`) for forecasters to return forecast residuals. See also #1105.
Tests for `predict_residuals` are also included (checks whether indices are correct).

FYI @ngupta23, @AngelPone, is this what you were looking for?


fkiraly:
(addresses *residuals* - *in-sample* request should be automatic if the forecaster can make in-sample forecasts)

fkiraly:
what we could also do:
* `fit_predict_residuals`
* in `fit_predict` and `fit_predict_residuals`, if no `fh` is provided, then `y.index` is used (so it's in-sample automatically)

sktime/forecasting/base/_base.py
```diff
@@ -631,6 +637,48 @@ def update_predict_single(
...
+        fh : int, list, np.ndarray or ForecastingHorizon
+            Forecasting horizon
+
+        Returns
+        -------
+        y_res : pd.Series, pd.DataFrame, or np.ndarray (1D or 2D)
+            Forecast residuals at fh, with same index as fh
+            y_pred has same type as y passed in fit (most recently)
+        """
+        y_pred = self.predict(fh=fh, X=X)
```
ngupta23, line 670: @fkiraly I do not see the handling for `fh=None` (i.e. return all in-sample residuals). Is that something that will be added later?

ngupta23:
@fkiraly , could we decouple the fit process from the 

> * in `fit_predict` and `fit_predict_residuals`, if no `fh` is provided, then `y.index` is used (so it's in-sample automatically)

@fkiraly , I am not completely following this. Will residual prediction be available as its own method (with an option for in-sample and out of sample residuals? If so, what would the interface look like?

sktime/forecasting/base/_base.py
```diff
@@ -631,6 +637,48 @@ def update_predict_single(
             alpha=alpha,
         )
 
+    def predict_residuals(self, y, X=None, fh=None):
```
ngupta23, line 640: Shouldn't `y` be optional for in-sample predictions?

sktime/forecasting/base/_base.py
fkiraly: we could do this in this PR if you'd prefer that? I wouldn't mind.
Also not sure what the other core devs think what the default behaviour for `fh=None` should be.

sktime/forecasting/base/_base.py
fkiraly: this is not necessarily in-sample, but we could make it that if no `y` is passed, in-sample residuals are computed, and the already seen `_y` are used. If that's convenient?

ngupta23: So just confirming --- 

## For in sample residuals, the call would simply be 

```python
# Option 1: All In-sample residuals
forecaster.predict_residuals()  

# Last "n" (n=4 in this case) in-sample residuals
forecaster.predict_residuals(fh=[0, -1, -2, -3]) 
```

## For out of sample, it would be 

```python
# Option 1: # fh would be same length as Sth
forecaster.predict_residuals(y=Sth)  

# Option 2
# Would non-matching indices (between y and predictions) be ignored or raise error or warning?
forecaster.predict_residuals(y=Sth1, fh=Sth2)  
```

fkiraly: ok, here's with what I'm currently thinking.

if you want in-sample residuals:
```python
# if fh must be passed in fit
forecaster.fit(y, fh=y.index).predict_residuals()
# if fh can be passed in predict
forecaster.fit(y).predict_residuals()
```

for out-of-sample residuals:
```python
# if fh must be passed in fit
forecaster.fit(y, fh=my_fh).predict_residuals(y_true)
# if fh can be passed in predict
forecaster.fit(y).predict_residuals(y_true)
```

The `fh` is inferred in either case, it's the inde of `y` in `fit` (in-sample) or `y_true` (out-of-sample); if no `y` is passed, the `y` in `fit` is used as the argument in `predict_residuals`.

chrisholder (dismissed):

sktime/forecasting/base/_base.py
ngupta23: Hi @fkiraly Sorry for the delayed response. I was out over the holidays. I think your alternate solution is cleaner and we can work with this.

fkiraly: alright!

fkiraly:
@chrisholder, @ngupta23, I wrote some tests.

I also removed the redundant `fh` argument from `predict_residuals` (was not used in the examples).

sktime/forecasting/base/_base.py
fkiraly: I'll write tests then so it's ready to go in.

fkiraly: (done)

sktime/forecasting/base/_base.py
fkiraly: ok, done - see below (if no argument is passed)

chrisholder (approved):

ngupta23 (approved):

PR #1771, opened 2021-12-18
[ENH] Sequential augmenter (as panel transformer) for easy TS data-augmentation

MrPr3ntice:
#### Reference Issues/PRs
New feature (discussion #1592)

By now transformers in sktime do not support basic augmentation techniques like noise adding, flipping, reversing... as well as the sequential (and also randomized)  operation of these transformations. See discussion here: #1592

No new dependencies, I think.

* Does overall structure apply to #1365?
* We used a base class `_BasePanelAugmenter` (child of `BaseTransformer`). Hopefully it complies to https://github.com/alan-turing-institute/sktime/blob/main/extension_templates/transformer.py 
* All actual augmenters are childs of `_BasePanelAugmenter`, as 90% of functionality is shared. Are you happy with this?
* Is our pytest for the existing augmenter sufficient?

#### Any other comments?
Work still in progress, thanks for your feedback!

##### For all contributions

##### For new estimators/transformers
- [ ] I've added the estimator/transformers to the online documentation.


lmmentel:
This is a great feature to add, thanks!

In case it's helpful, let me throw some of the existing time series augmentation packages:

- [deltapy](https://github.com/firmai/deltapy) Tabular Data Augmentation & Feature Engineering.
- [tsaug](https://github.com/arundo/tsaug) A Python package for time series augmentation.
- [time_series_augmentation](https://github.com/uchidalab/time_series_augmentation) An example of time series augmentation methods with Keras.

They follow slightly different designs.

I would be happy to help and get it merged. Would you be open to suggestions?

MrPr3ntice:
> This is a great feature to add, thanks!
> 
> In case it's helpful, let me throw some of the existing time series augmentation packages:
> 
>     * [deltapy](https://github.com/firmai/deltapy) Tabular Data Augmentation & Feature Engineering.
> 
>     * [tsaug](https://github.com/arundo/tsaug) A Python package for time series augmentation.
> 
>     * [time_series_augmentation](https://github.com/uchidalab/time_series_augmentation) An example of time series augmentation methods with Keras.
> 
> 
> They follow slightly different designs.
> 
> I would be happy to help and get it merged. Would you be open to suggestions?

Thanks for your feedback and for the links to existing packages! I'm quite fimiliar with [tsaug](https://github.com/arundo/tsaug) (and slightly with [time_series_augmentation](https://github.com/uchidalab/time_series_augmentation)) and really appreciate its features and structure!
One of the differences to tsaug or time_series_augmentation: We try to provide augmenters, that are (if possible and desired) fittble to the dataset (both _fit-in-transform_ or separate _fit_...), which can be useful e.g. in contrast to the absolute noise adding (compare `tsaug.add_noise()` or `time_series_augmentation.jitter(x, sigma)` -> Noise with **absolute** std will be added in constrast to adding relative noise (SNR) ...). Furthermore, we of course try to provide compatability to `sklearn.pipeline.Pipeline` (and obviously all other sktime dependencies).

Thanks for mentioning  [deltapy](https://github.com/firmai/deltapy), which I did not know so far.

Of course we are open to any of your suggestions (e.g. questions of understanding, implementation of _not-yet-implemented_ augmenters, documentation, help with checks...) - Thanks and loccking forward for your support!

sktime/transformations/panel/augmenter.py
```diff
@@ -0,0 +1,1217 @@
...
+        n_jobs=1,
+    ):
+        # input parameters
+        self.p = p
+        self.param = param
+        self.use_relative_fit = use_relative_fit
+        self.relative_fit_stat_fun = relative_fit_stat_fun
+        self.relative_fit_type = relative_fit_type
+        self.random_state = random_state
+        if excluded_var_indices is None:
```
lmmentel, line 121: If you would like to stay compatible with sklearn design principles (https://scikit-learn.org/stable/developers/develop.html#apis-of-scikit-learn-objects) there should not be any logic in the `__init__` methods beyond setting the attributes.

sktime/transformations/panel/augmenter.py
```diff
@@ -0,0 +1,1217 @@
...
+        if (
+            self.use_relative_fit
+            and self.relative_fit_type == "fit-transform"
+            and not self._is_fittable
+        ):
+            # calculate demanded statistical param over (a concatenation of)
+            # all instances for each variable (like in case of "fit" but
+            # directly on the data to be transformed).
+            self._stats = []
+            for col in range(X.shape[1]):  # loop over demanded variables
```
lmmentel, line 225: Is this block duplicated in `_fit` ?

sktime/transformations/panel/augmenter.py
```diff
@@ -0,0 +1,1217 @@
...
+            False.
+        """
+        list_of_aug_info = []
+        for aug in self.steps:
+            if isinstance(aug, _BasePanelAugmenter):
+                list_of_aug_info.append(aug[1]._last_aug_random_variate)
+        return list_of_aug_info
+
+    @staticmethod
+    def draw_random_samples(
```
lmmentel, line 388: Any reason this should be a part of the class rather than a standalone function?

sktime/transformations/panel/augmenter.py
```diff
@@ -0,0 +1,1217 @@
...
+    -------
+    matplotlib.figure.Figure: A figure with a [n_variables, 2] subplot-grid.
+    """
+    _check_soft_dependencies("matplotlib")
+    import matplotlib.pyplot as plt
+
+    n_vars = X.shape[1]  # get number of variables of X
+    # pick (stratified regarding categorical y) examples from the original input
+    # data
+    X, y, idx = SeqAugPipeline.draw_random_samples(
```
lmmentel, line 514: `SeqAugPipeline` feels a bit unnecessary to call here, maybe a call to a standalone `draw_random_samples` would suffice?

sktime/transformations/panel/augmenter.py
```diff
@@ -0,0 +1,1217 @@
...
+            f"relative_fit_stat_fun={ft.relative_fit_stat_fun.__name__}, "
+            f"relative_fit_type='{ft.relative_fit_type}', "
+            f"random_state={ft.random_state},{nl}"
+            f"excluded_var_indices={ft.excluded_var_indices}, "
+            f"n_jobs={ft.n_jobs}.{nl}"
+            f"Date: {datetime.now().strftime('%Y-%m-%d %H:%M:%S')}"
+        )
+    except Exception as e:
+        # param_str = "Unknown augmenter and parameterization"
+        raise ModuleNotFoundError(e)
```
lmmentel, line 549: Any reason to call `ModuleNotFoundError` here?

sktime/transformations/panel/augmenter.py
MrPr3ntice: Matches with your question above. We first planned to make `draw_random_samples` as indispensable part of the pipeline class (i.e. by calling `transform` or  `fit` on the pipeline random samples are augmented) but this does not comply to a lot of rules of sklearn. So short answer: Yes, I will definitely move  `draw_random_samples` out of the pipeline class.

sktime/transformations/panel/augmenter.py
MrPr3ntice: Good point, it seems that we doubled sklearn's `augmenter.fit_transfrom()` here... I don't remember my intention here - I will check. Do you know the exact definition of the `_tag` `fit-in-transform` of the transformer template? Is it whether `fit_transform()` is **generally** allowed or that **only** `fit_transform()` is allowed (or s.th. completely different).

lmmentel:
Would you consider a slightly different approach to this where the actual augmentation functions are standalone functions that could then be wrapped into transformers? That would make the augmenters a little bit less coupled to the Transformer api and open up possibilities of using the functions outside of Transformer/Pipeline api. 

A good example of similar design is [tsfresh](https://github.com/blue-yonder/tsfresh) consider for example this module https://github.com/blue-yonder/tsfresh/blob/main/tsfresh/feature_extraction/feature_calculators.py where primitive feature extractors are implemented. Then there are multiple wrappers that expose the grouped feature extractors etc.

This would also allow to document the actual functions since currently the docs are buried in the class docs with all the other parameters.

Finally it would reduce the boilerplate code for each of the classes since, in the example below you have 70 lines of code with only **1 line** performing the actual logic:

```python
class InvertAugmenter(_BasePanelAugmenter):
    """Augmenter inverting the time series (i.e. multiply each value with -1).
    This is a sub class inheriting its functionality from the superclass
    `_BasePanelAugmenter`. See there for more details.
    Parameters
    ----------
    p: float, optional (default = 1.0)
        Probability, that a univariate time series (i.e.: defined by (instance,
        variable) of a multivariate panel) is augmented.
        Otherwise the original  time series is kept. In case of p=1.0,
        every  time series in the panel is augmented. Notice, that in case of
        multivariate time series the decision whether a certain variable of
        an instance is augmented is stochastically independent from the
        other variables of the same instance.
    param: any, optional (default = None)
        Ignored, as well as use_use_relative_fit.
    use_relative_fit: bool, optional (default = False)
        Whether to a fit the augmenter regarding a statistical parameter of
        the data. If True, relative_fit_stat_fun specifies the statistical
        parameter and relative_fit_type specifies the fitting reference data.
    relative_fit_stat_fun: a function, optional (default = np.std)
        Specifies the statistical parameter to calculate. Any function
        that satisfies the following syntax is allowed: stat = fun(X:
        pd.Series). The only allowed input is a numeric pd.Series and output
        has to be float or int. The default is the standard deviation np.std.
    relative_fit_type: str, optional (default = "fit")
        "fit": fit demanded statistics with a separate train set.
        "fit-transform": fit statistics just before transformation
        regarding the whole given panel.
        "instance-wise": fit statistics just before transformation
        individually per instance (and variable).
    random_state: int, optional (default = None)
        A random state seed for reproducibility.
    excluded_var_indices: iterable of int optional (default = None)
        Iterable (e.g. tuple or list) of int, containing the indices of those
        variables to exclude from augmentation. Default is None and all
        variables are used.
    n_jobs: int, optional (default = 1)
        Integer specifying the maximum number of concurrently running
        workers on specific panel's instances. If 1 is given, no parallelism is
        used. If set to -1, all CPUs are used. For n_jobs below -1, (n_cpus
        + 1 + n_jobs) are used. For example with n_jobs=-2, all CPUs but one
        are used. Not implemented yet.
    """

    def __init__(
        self,
        p: float = 1.0,
        param=None,
        use_relative_fit=False,
        relative_fit_stat_fun=np.std,
        relative_fit_type="fit",
        random_state=None,
        excluded_var_indices=None,
        n_jobs=1,
    ):
        self._is_fittable = False
        super().__init__(
            p,
            param,
            use_relative_fit,
            relative_fit_stat_fun,
            relative_fit_type,
            random_state,
            excluded_var_indices,
            n_jobs,
        )

    def _univariate_ser_aug_fun(self, X, _, __):
        return X.mul(-1)
```

MrPr3ntice:
I fully agree that this structure is not appropriate right now. Actually the more wrapper-like approach you mentioned has been our approach before but we failed some checks and changed to this not-ideal boilerplate structure just to pass the tests. I will take a look at tsfresh's version of doing the wrapping, but I remember they use standalone functions for their FE. This decision is strongly coupled to the way of documentation - We will give an updated proposal for this soon.

sktime/transformations/panel/augmenter.py
MrPr3ntice: I fixed that.

sktime/transformations/panel/augmenter.py
MrPr3ntice: Fixed that (see above...)

sktime/transformations/panel/augmenter.py
MrPr3ntice: Thanks for the link, That should be possible by moving all checks, etc. to `_fit()` (seems also to be the preferred way in sklearn). I will provide a updated solution soon.

sktime/transformations/panel/augmenter.py
MrPr3ntice: Not really, I fixed that. Notice that `plot_augmentation_example()` is a nice feature for the developement of augmented TS but not mandatory as it provides no essential functionality. Maybe we will move this to the example notebook later.

sktime/transformations/panel/augmenter.py
fkiraly: `fit-in-transform` says whether `fit` is empty and could, in-principle, be ommitted. An example for that would be the transformer that squares all entries, no fitting needed. A counter-example would be centering, i.e., subtracting the mean on the training set.

fkiraly (changes requested):
@MrPr3ntice, really nice contribution, thanks a lot!

May I first confirm whether I understand correctly what's going on on the abstract conceptual level? I'd also appreciate paper links. Is the below correct:

* this PR is about "augmenters". These are used to enrich a training set with more samples. In general, they add samples to a training set, i.e., starting with N labelled series, you arrive at M >= N labelled series. Typically, there's an 1:n correspondence before/after, with one series resulting in n (where n may vary by series) transformed/augmented series in the augmented panel.
* a special sub-case is where you start with a panel of N series and end up with a panel of N series, an individual "augmentation" operation is applied to each of the series. Here, there's a 1:1 correspondence between before/after indices.

Is this correct?

If so, I have some questions/suggestions.

* In the general augmentation (result can have more series than original training set), isn't it important to keep track of the labels? I.e., should the augmentation transformation not return the panel, *and* labels? If so, there's an open design question, i.e., how to transform the labels. A method `transform_labels`? Getting it via attribute, or `get_fitted_params`?
* in the case where you have a 1:1 connection between indices before/after, it looks like you are applying individual transformations to individual series. Is that correct? These transformations look like they are `Series` to `Series`, so should be implemented as such.

More generally, I'd suggest a little different architecture: remove the augmentation boilerplate from the individual transformations, I agree with the philosophy of what @lmmentel suggests. I suggest to leave the augmentation logic in one compositor.

What I think might be an elegant way is to create a single panel-to-panel transformer like 
```python
my_augmenter = PanelAugmenter(
    transformers = [
        ("white_noise", WhiteNoiseAdder()),
        ("reverse", ReverseTransformer()),
    ]
    augment_probabilities = [0.3, 0.3],
)
```
(applying either with probability 0.3, applying at most one; you could think about arguments that control whether these are applied in sequence, and/or independently)
and `WhiteNoiseAdder` or `ReverseTransformer` are series-to-series (ordinary transformers).

Variations on the theme:
* have two classes, one that keeps the number of instances the same with an 1:1 correspondence before/after, and one for more complex behaviour that enlarges the training set.
* or, have a parameter `augmentation_logic` which can be `"replace"` or `"add"`. If you replace, it's 1:1, each time an augmentaiton step is carried out, the series is replaced. If `"add"`, the augmented series is instead appended to the training set as a new series.

What do you think?

PS: I've been cleaning up the transformers base class here:
https://github.com/alan-turing-institute/sktime/pull/1790

You may like to try (in a separate, local branch) whether this causes any issues for you. I hope not, since it's only cleaning up the code and adding functionality for methods that you don't use.

fkiraly:
PPS: for the suggestion of a "compositor", I would recommend you look at `ColumnwiseTransformer` which also has a `transformers` argument - you can follow the implementation almost verbatim to avoid fiddling around with how to check and clone arguments of `__init__` which are estimators/transformers.
For that, I'd recommend to look at the version in PR #1795 which has a *refactored* version compliant with the new 0.9.0 interface in it rather than the current version on `main` which is still following the old one. It passes all the tests and should work (but of course there's never a 100% guarantee). 

For convenience, generic instructions for composite estimators with a `transformers` parameter:
1. in `fit`, ensure you `sktime.clone` all transformers into a new attribute `transformers_`. If you don't do this, the tests will complain since `__init__` arguments are changed by `fit`.
2. To deal with the parameter access, inherit from `_HeterogenousMetaEstimator`, and include the two methods:
```python

    def get_params(self, deep=True):
        """Get parameters of estimator in `transformers_`.

        Parameters
        ----------
        deep : boolean, optional
            If True, will return the parameters for this estimator and
            contained sub-objects that are estimators.

        Returns
        -------
        params : mapping of string to any
            Parameter names mapped to their values.
        """
        return self._get_params("transformers_", deep=deep)

    def set_params(self, **kwargs):
        """Set the parameters of estimator in `transformers_`.

        Valid parameter keys can be listed with ``get_params()``.

        Returns
        -------
        """
        self._set_params("transformers_", **kwargs)
        return self
```
This will automatically call boilerplate code for nested parameter access in the `transformers_` variable you created in 1.

fkiraly:
another thought that I had: as @lmmentel mentions, there's some boilerplate code - some comes from the augmentation logic, some from the fact that we are wrapping one-line functions. There is already a shorthand for wrapping one-line functions, it's the `FunctionTransformer` in `sktime` (similar to the one in `sklearn` but with the series interface).
So you could use that directly in the `PanelAugmenter`, or use it to define shorthands like `ReverseTransformer`.

sktime/transformations/panel/augmenter.py
MrPr3ntice: I fixed that.

iljamaurer:
@lmmentel & @fkiraly 
We are new to contribute to open-source, thank you very much for your valuable feedback, we are trying to catch up. :-) 
We have reviewed all of your suggestions extensively and would like to discuss a possible redesign of the augmentation feature.

To my understanding, we currently have 3 major topics / issues:

1. The augmentors should probably be `Series` to `Series`
2. Lots / duplicates of parameters inside individual augmentors
3. Boilerplate code long docstrings

We took a look at `FunctionTransformer` and we think it would work for the current implemented augmentors. But we have a couple of other augmentors inside our dev-pipeline, and are not sure if this approach will also fit in future and stay consistent with the further development / tests of the `FunctionTransformer` class. 

So, we propose the following redesign of the augmentation module:

Individual augmentors will be moves to `Series` transformers (1.):
```
class ReverseAugmenter(_BaseSeriesAugmenter)
```

Only taking augmentor-specific parameters as inputs - and only if needed (2. & 3.):
```
class ReverseAugmenter(_BasePanelAugmenter):
    """Augmenter reversing the time series.

    Parameters
    ----------
        none
    """
    def __init__(self):
        super().__init__()

class ScaleAugmenter(_BasePanelAugmenter):
    """Augmenter reversing the time series.

    Parameters
    ----------
    method: string 
        'factor' or 'distribution'
    param: dict
        Parameters for selected method, e.g. 
        Scale factor 'factor': {'value': 1.0} or
        Given a distribution, an i.i.d. random variate will be
        drawn for each TS augmentation. If use_relative_fit is True,
        the actual scale factor will be the product of the fitted statistical
        parameter and this value.
        'distribution': {'function': scipy.stats.rv_continuous, 'use_relative_fit':True}
    """
    def __init__(self, method, param):
        self.method = method
        self.param = param
        super().__init__()
```


The previous parameter are either moved to the pipeline / panel (s. below), or to the `_BaseSeriesAugmenter` class with default-values and accessible via `get_params` and `set_params`:
```
class _BaseSeriesAugmenter(_SeriesToSeriesTransformer):
    random_state = 42

    def get_params(self, deep=True):
         ...
    def set_params(self, random_state):
        self.random_state = random_state
```

For just a single time-series, the augmentors can now be called individually:
```
X = pd.Series()
rev_aug = ReverseAugmenter()
rev_aug.fit(X) 
Xt = rev_aug.transform(X)
# Or Xt = fit_transform(X)
```

To perform an augmentation on a dataset, we will implement (or redesign) a `Panel` transformer, as suggested by @fkiraly. Further, reduce the parameters and docstrings (2. & 3.):
```
class PanelAugmenter(_PanelToPanelTransformer)
    """Panel / pipeline augmentation transformer.

    Parameters
    ----------
    transformers: list or dict? To evaluate
       mapping of augmentation to individual time-series instances
    p: list
        Probability, that a univariate time series (i.e.: defined by (instance,
        variable) of a multivariate panel) is augmented.
        Otherwise the original  time series is kept. In case of p=1.0,
        every  time series in the panel is augmented. Notice, that in case of
        multivariate time series the decision whether a certain variable of
        an instance is augmented is stochastically independent from the
        other variables of the same instance.
    excluded_var_indices: iterable of int optional (default = None)
        Iterable (e.g. tuple or list) of int, containing the indices of those
        variables to exclude from augmentation. Default is None and all
        variables are used.
    """

    n_jobs = 1

    def __init__(
        self,
        transformers,
        augment_probabilities,
        excluded_var_indices=[]
    ):
    self.transformers = transformers 
    self.augment_probabilities = augment_probabilities 
    self.excluded_var_indices = excluded_var_indices

    def _fit(self, X, y=None)
        ... 

    def set_params(self, n_jobs):
        self.n_jobs= random_state
```
callable e.g. by
```
my_augmenter = PanelAugmenter(
    transformers = [
        ("white_noise", ReverseAugmenter()),
        ("reverse", ScaleAugmenter(method='factor', param={'value':1.0})),
    ]
    augment_probabilities = [0.3, 0.3],
)
X = pd.DataFrame()
Xt = my_augmenter(X)
```

`use_relative_fit` and `relative_fit_stat_fun` parameter are not longer needed or will be part of the `Series`-augmentors `param` aurguments (if needed). 

We are currently evaluating the use `ColumnwiseTransformer` for `PanelAugmenter`.

In our opinion, this will reduce the boilerplate code and make the module more maintainable(?)

What do you think about this design? Thank you for the review, we are looking forward to your feedback!

@MrPr3ntice you are welcome to complete my post, if I missed something :-)

Again thanks a lot @lmmentel @fkiraly 
- Ilja

lmmentel:
Thanks a lot @iljamaurer and @MrPr3ntice for putting the time to think though the design and share your thoughts!

Although the discussion about class design, inheritance and code architecture so far has been very interesting and I would very much like to jump in with my thoughts right away, I will wait until we have more concrete concerns and less abstract challenges. 

If you are open I would like to propose to turn the attention to actual augmentation functionality first and then developing a wrapper that would expose that functionality as sktime compatible `Transformers` later on. I think that could speed up the progress and have the benefit of us being able to split up the work into smaller chunks.

What I would have in mind specifically is first developing a module (or other collection) of functions doing the heavy lifting in the augmentation. For example your `InvertAugmenter` class could simply be

```python
def invert(X: np.array | pd.Series):
    """Invert a series or array by multiplying it by -1"""
    return X.mul(-1)
```

following the same logic the `WhiteNoiseAugmenter` could be sth like (I renamed some args, sorry)

```python
    def add_white_noise(X: np.array | pd.Series, use_relative_fit: bool, scale: float, stat_param: float):
        """Augmenter adding Gaussian (i.e . white) noise to the time series.
        
       Args:
           X: description
           scale: description
           ....
        """
        n = X.shape[0]  # length of the time series.
        if self.use_relative_fit:
            return X + norm.rvs(loc=0, scale=scale * stat_param, size=n)
        else:
            return X + norm.rvs(loc=0, scale=scale, size=n)
```

Having such a module would simplify the code by a big factor and would make it simple to add any new augmenting functions.

There is already a lot to do with those fundamental functions since it needs to be figured out which of them operate on `pd.Series` vs `pd.DataFrame` vs multidimentional array or any other structure. However once this step is completed we can start thinking about convenient wrappers a compositions of those functions. That should be a little easier since by then we will have concrete function signatures that will inform how to construct those higher level objects.

In terms of process I think developing a module with augmenting function is already very useful on it's own and could possibly be merged as a standalone PR to create some momentum behind this work and not be slowed down by figuring out all the wrapping and class issues which would be a natural second step to this. WDYT @fkiraly ?

I hope this makes sense but I'm happy to elaborate if there are any questions or concern.

fkiraly:
@iljamaurer, I agree 95% with your design! And I agree 95% with @lmmentel's, too.

>To my understanding, we currently have 3 major topics / issues:
>The augmentors should probably be Series to Series

most of them look like they should, yes.

>Lots / duplicates of parameters inside individual augmentors

Not necessarily, or I don't understand what you mean.

>Boilerplate code

Yes

> long docstrings

no! This is not a problem but good. Long docstrings that explain what happens are *good*!

Overall:

I agree with you, @iljamaurer and @lmmentel, that we should have the individual augmentation transforms as transformers.
I'd suggest they inherit directly from `BaseTransformer` and not from a `BaseAugmenter` class - augmentation functionality should be provided by the `PanelAugmenter`.

For example:
```python
class ScaleAugmenter
    """Augmenter scaling the time series.

    Parameters
    ----------
    method: string 
        'factor' or 'distribution'
    param: dict
        Parameters for selected method, e.g. 
        Scale factor 'factor': {'value': 1.0} or
        Given a distribution, an i.i.d. random variate will be
        drawn for each TS augmentation. If use_relative_fit is True,
        the actual scale factor will be the product of the fitted statistical
        parameter and this value.
        'distribution': {'function': scipy.stats.rv_continuous, 'use_relative_fit':True}

    NOTE: needs to be better described what this does. Use formulae of applicable, or cite a section in a paper.
    MINOR NOTE: I'd also set some defaults for method and param. "sensible defaults" make this easier to use.
    """
    _tags = {
        "scitype:transform-output": "Series",
        "scitype:transform-labels": "None",
        "scitype:instancewise": True,
        "handles-missing-data": False,
        "X_inner_mtype": "pd.DataFrame", # or pd.Series, or no.ndarray, depending what you want in _transform
        # can be a list of multiple types, if `_transform` can deal with multiple.
        "X-y-must-have-same-index": False,  # can estimator handle different X/y index?
        "fit-in-transform": True,  # is fit empty and can be skipped? Yes = True
        "transform-returns-same-time-index": False, # probably depends on augmentation?
        # does transform return have the same time index as input X
        "capability:inverse_transform": False, # use only after transformer refactor merged
    }

    def __init__(self, method, param):
        self.method = method
        self.param = param
        super().__init__()

        # stuff happens here
```

Regarding boilerplate:
* you don't need to re-implement `get_params`, `set_params` for the "simple functions"! This comes automatically from `BaseObject` (via `BaseTransformer`). Writing this is generally not needed, unless you have estimators within nested structs in init parameters.
* If the `_tags` turn out to be repetitive across a dozen transformers, you can put the `_tags`, in a class called `_AugmenterTags` (which contains only the tags) and inherit from it.
* you don't need to write a `_fit` if the transformer is not fittable. This is dealt with by the `fit-in-transform` tag, if `True`, nothing happens in `_fit` and the framework recognizes that.

Regarding @lmmentel's suggestion to have separate augmentation functions.
If these are one-liners, may be more hassle than worth, but it's not a bad idea (depends how many there are).

fkiraly:
@iljamaurer, @MrPr3ntice, how is it going?

This would be a nice thing to have in our 0.10.0 release!
What's your time estimate?

iljamaurer:
> @iljamaurer, @MrPr3ntice, how is it going?
> 
> This would be a nice thing to have in our 0.10.0 release! What's your time estimate?

Sorry, I'm pretty busy at work atm. :-( I try to take a couple of hours next weekend and re-design the module like you both suggested (thank you!!!). Afterwards, we can have a look at the new structure. :-)

fkiraly:
thanks! Let us know if you need help, and feel free to join the collaborative sessions for pair programming if you'd like.

MrPr3ntice:
Same for me, I have a deadline this Friday and hope to find time at the weekend to go into the changes and especially enable easy addition of new transformers, as rightly requested by @lmmentel. Can you provide some more info about collaborative coding? How is it organized?
Thanks a lot.

fkiraly:
> Can you provide some more info about collaborative coding? How is it organized?

We meet at 1pm on discord and first give quick updates what we've been working on. Then there's ca 20min to ask questions, or share interesting learnings. You could also ask for more eyes to review code, or advice, or show sth.

Smaller groups can then split off into pair or mob programming, depending on interest.

iljamaurer:
Thanks again for your fantastic feedback @fkiraly and @lmmentel. I redesigned the module based on your comments and in my opinion it definitely makes more sense now:

Every augmenter 

- is a series-to-series transformer with own parameters and just a `_transform` method.
- inherits from `_AugmenterTags`

In my opinion, one of the main problems of the previous version was, that we tried to use random sampling AND the specific augmentation at the same time. Therefore, I created a new `RandomSamplesAugmenter` class. Now, the routines are independent, and we also can combine them more flexible, e.g.

```
Xt_1 = InvertAugmenter().fit_transform(X)
Xt_2= RandomSamplesAugmenter ().fit_transform(Xt_1)
…
```

Nevertheless, there is some heavy lifting left. What I ignored so far: Augmentation based on multiple time series (instances). I think the series augmenter should stay 1:1 and n:1 or n:n should be done in a separate `PanelAugmentor` (pipeline), e.g.:

```
class PanelAugmentor:
	# some configs: which augmeters to use etc.
	# handle probability, that a univariate time series is augmented
	# combine all relevant time-series (X_1, … X_n) instances in a songle time-series X
	X = pd.concat([X_1,…,X_n])
	# calculate parameter based on X (if necessary) e.g. 
	std_scale = np.std(X)
	# pass defined time-series X_i and parameter(s) to defined augmenters
	aug = WhiteNoiseAugmenter(scale = std_scale)
	for xn in [X_1,…,X_n]:
		Xt_i = aug.fit_transform(xn)
```

Please tell me if you agree.

Next steps:

- [ ] Validate current approach @all
- [ ] Review and bug fixing current augmenters @MrPr3ntice  
- [ ] Design and implementation of ` PanelAugmentor `  @iljamaurer 
- [ ] Further development of current augmenters (more parameter, parameter checking, etc.)

Thank you one again, your help is much appreciated! :-)

fkiraly (changes requested):
Excellent, thanks!

First, some of the "easier" things to fix.

The test failures are likely arising from the wrong tags: it seems that, clearly, you want the inner type to be `pd.DataFrame`, but it ends up being `pd.Series`.

Why: you are inheriting from two classes that are basically sets of tag defaults, ` _SeriesToSeriesTransformer` and `_AugmenterTags`. As with `python` inheritance resolution order, the *first* one in the list overrides the subsequent ones, so `_SeriesToSeriesTransformer` overrides the tags of `_AugmenterTags`.

Instead, you want `_AugmenterTags` to take priority, and the tags in  `_SeriesToSeriesTransformer` you don't need because (a) you have your own tag template, and (b) it's legacy. Inherit instead this way: `MyClass(_AugmenterTags, BaseTransformer` and it should work.

Further, I suspect that something isn't quite working with the `random_state` in `RandomSamplesAugmenter`, since the tests that are failing compare two invocations (but I did not look in detail).

fkiraly:
> Please tell me if you agree.

Yes, I do - my suggestion above was also along those lines.
One thing perhaps, the `PanelAugmenter` should be able to take panel transformers as components as well, optimally (although you may want to start with an implementation that takes only "by-series" ones)

iljamaurer:
> Excellent, thanks!
> 
> First, some of the "easier" things to fix.
> 
> The test failures are likely arising from the wrong tags: it seems that, clearly, you want the inner type to be `pd.DataFrame`, but it ends up being `pd.Series`.
> 
> Why: you are inheriting from two classes that are basically sets of tag defaults, ` _SeriesToSeriesTransformer` and `_AugmenterTags`. As with `python` inheritance resolution order, the _first_ one in the list overrides the subsequent ones, so `_SeriesToSeriesTransformer` overrides the tags of `_AugmenterTags`.
> 
> Instead, you want `_AugmenterTags` to take priority, and the tags in `_SeriesToSeriesTransformer` you don't need because (a) you have your own tag template, and (b) it's legacy. Inherit instead this way: `MyClass(_AugmenterTags, BaseTransformer)` and it should work.
> 
> Further, I suspect that something isn't quite working with the `random_state` in `RandomSamplesAugmenter`, since the tests that are failing compare two invocations (but I did not look in detail).

Thank you, you are absolutely right. :-) Now the tests should succeed. But probably we should switch the `X_inner_mtype` to `pd.Series`... I will evaluate this. 

Regarding `RandomSamplesAugmenter`: On my machine (sktime Dockerfile) it passes all tests. Maybe it has something to do with `random.seed` and different OS? I think I also had something in the last build. Anyway, I will have a look at this next weekend (I am always on business trips working days).

- [ ] Evaluate switching  `X_inner_mtype` to `pd.Series`
- [ ] Fix `RandomSamplesAugmenter` unit tests

lmmentel:
Great to hear about the progress @iljamaurer and @MrPr3ntice!

Since you are already splitting the work would it make sense to limit the scope of this PR to series-to-series transformers/augmenters and once we merge start working on extending the functionality towards more complex data structures like panels? My motivation here is to not wait too long before merging the work, since long cycles and big PR feed on each other and make the cycles even longer.

iljamaurer (approved):

iljamaurer (approved):

fkiraly:
@iljamaurer, do you need a review? Guessing that you maybe wanted to request a review instead of approving your PR?

iljamaurer:
> @iljamaurer, do you need a review? Guessing that you maybe wanted to request a review instead of approving your PR?

yes, sorry :-D. I think I found&fixed the issue with `RandomSamplesAugmenter` and all checks are successful at my local machine. Maybe it is worth a review & PR before implementing more stuff, as long as the module is that compact? What do you think?

fkiraly (changes requested):
Excellent, nice work!
Almost ready to go.

Sorry for the delay in review, we were busy with the 0.10.0 release.

May I ask for one change:
could we replace inheriteance from  `_SeriesToSeriesTransformer`, by `BaseTransformer`?
Because it seems you are overriding all tags introduced in `_SeriesToSeriesTransformer` anyway.

iljamaurer:
> Excellent, nice work! Almost ready to go.
> 
> Sorry for the delay in review, we were busy with the 0.10.0 release.
> 
> May I ask for one change: could we replace inheriteance from `_SeriesToSeriesTransformer`, by `BaseTransformer`? Because it seems you are overriding all tags introduced in `_SeriesToSeriesTransformer` anyway.

Thank you, you are right. I just committed the change.  :-) Please tell me, if there is something else to fix before merging.

fkiraly (dismissed):
Looks good to me now.

We can merge this, as it is the first in a sequence of PR.

Things I'd like added in the future:
* full docstrings in *all* classes - all should describe what the transformers do. Optimally, use formal math or algorithmic notation.
* is "invert" the right term in `InvertAugmenter`? Usually, isn't that used for multiplicative inversion, i.e., "one divided by x"? `FlipSignAugmenter` is perhaps better?
* `RandomSamplesAugmenter` docstring does not properly describe what the transformer does, this should be fixed/corrected.

fkiraly:
@MrPr3ntice, @iljamaurer, what are your timelines working on this, in continuation?
I'd merge this only if you are planning to continue working on this - since without the rest, it would be a bit incomplete.
Also, are you on our developer slack? Let's exchange a few messages there on the plan and/or which releases we put this in.

iljamaurer:
> @MrPr3ntice, @iljamaurer, what are your timelines working on this, in continuation? I'd merge this only if you are planning to continue working on this - since without the rest, it would be a bit incomplete. Also, are you on our developer slack? Let's exchange a few messages there on the plan and/or which releases we put this in.

Thank you for your reply and the feedback in the previous post. We will definitely continue to work on the augmenter modules. I just joint the slack group, maybe we can continue there to discuss our contribution to sktime in more detail and – if you allow – to pick your brain how to organize our schedule :-)

fkiraly:
Thanks for updating the docs!

I couldn´t merge this, because the examples were failing in the docstrings.
The reason for this is that the printout must be equal to the actual printout, in the examples, and you were assigning to `Xt` but not printing it. I took the liberty of fixing that, I hope you don't mind.

fkiraly:
still failing because there's a difference between the printout and the chars in the docstring - this must be identical, including newlines and whitespaces. Have to log off for now, but it should be clear what to change?

fkiraly:
ok, I think I fixed it now.

Btw, I noticed, you didn't add a description of your augmenters in the docstrings? That should happen after the first line and say sth like "`transform` reverses the order of a time series `X`, i.e.,` (then add some light math or code formalism)

iljamaurer:
@fkiraly thank you for fixing and merging. I will have a look at the docstrings again on the weekend. Also working on the concept of the panel augmenter. :-)

fkiraly:
> @fkiraly thank you for fixing and merging. I will have a look at the docstrings again on the weekend. Also working on the concept of the panel augmenter. :-)

I´ve fixed the examples, but didn´t merge yet.
Would you mind adding the methodology content in the docstrings before we merge?

iljamaurer:
@fkiraly thank you. :-) I just added some mathematical descriptions to the docstrings. 
@MrPr3ntice & Moritz: feel free to review and adjust my description if needed and add references (I think you or Moritz had some?)

sktime/transformations/series/augmenter.py
```diff
@@ -0,0 +1,200 @@
...
+                "Type of parameter 'scale' must be a float value or a distribution."
+            )
+        return X[0] + norm.rvs(0, scale, size=len(X), random_state=self.random_state)
+
+
+class ReverseAugmenter(_AugmenterTags, BaseTransformer):
+    r"""Augmenter reversing the time series.
+
+    If given time series $X={x_1, x_2, ... , x_n}$, then
+    $X_t={x_n, x_{n-1}, ..., x_2, x_1}.
```
fkiraly, line 87:
```suggestion
    returns $X_t={x_n, x_{n-1}, ..., x_2, x_1}.
```

sktime/transformations/series/augmenter.py
```diff
@@ -0,0 +1,200 @@
...
+            raise TypeError(
+                "Type of parameter 'scale' must be a float value or a distribution."
+            )
+        return X[0] + norm.rvs(0, scale, size=len(X), random_state=self.random_state)
+
+
+class ReverseAugmenter(_AugmenterTags, BaseTransformer):
+    r"""Augmenter reversing the time series.
+
+    If given time series $X={x_1, x_2, ... , x_n}$, then
```
fkiraly, line 86:
```suggestion
    If `transform` is given a time series $X={x_1, x_2, ... , x_n}$, then
```

sktime/transformations/series/augmenter.py
```diff
@@ -0,0 +1,200 @@
...
+        "transform-returns-same-time-index": False,
+        "capability:inverse_transform": False,
+    }
+
+
+class WhiteNoiseAugmenter(_AugmenterTags, BaseTransformer):
+    r"""Augmenter adding Gaussian (i.e. white) noise to the time series.
+
+    If given time series $X={x_1, x_2, ... , x_n}$, then
+    $X_t={x_1+e_1, x_2+e_2, ..., x_n+e_n}$ where $e$ is a value randomly drawn from
```
fkiraly, line 42:
```suggestion
    returns $X_t={x_1+e_1, x_2+e_2, ..., x_n+e_n}$ where $e$ is a value randomly drawn from
```

sktime/transformations/series/augmenter.py
```diff
@@ -0,0 +1,200 @@
...
+        "fit-in-transform": True,
+        "transform-returns-same-time-index": False,
+        "capability:inverse_transform": False,
+    }
+
+
+class WhiteNoiseAugmenter(_AugmenterTags, BaseTransformer):
+    r"""Augmenter adding Gaussian (i.e. white) noise to the time series.
+
+    If given time series $X={x_1, x_2, ... , x_n}$, then
```
fkiraly, line 41:
```suggestion
    If `transform` is given time series $X={x_1, x_2, ... , x_n}$, then
```

sktime/transformations/series/augmenter.py
```diff
@@ -0,0 +1,200 @@
...
+    $X_t={x_1+e_1, x_2+e_2, ..., x_n+e_n}$ where $e$ is a value randomly drawn from
+    $\frac{1}{\sigma\sqrt{2 \pi}}e^{-\frac{1}{2}(\frac{x}{\sigma})}$,
```
fkiraly, line 43:
```suggestion
    the normal distribution with mean 0 and standard deviation ``scale``.
```

sktime/transformations/series/augmenter.py
```diff
@@ -0,0 +1,200 @@
...
+        super().__init__()
+
+    def _transform(self, X, y=None):
+        return X.mul(-1)
+
+
+class RandomSamplesAugmenter(_AugmenterTags, BaseTransformer):
+    r"""Draw random samples form time series.
+
+    If given time series $X \in \mathbb{R}^n$ then $X_t \subseteq X$
```
fkiraly, line 140: hm, this should describe the distribution of the draw. Not sure what that is.

fkiraly (changes requested):
Great, thanks!
Issues to fix:
* I think the familiar "dollar" notation for TeX does not work with rst. Instead of $X$, you need to write :math:`X` etc (with a single diacritic, this is unfortunately not rendered properly, i.e., colon-math-colon-diacritic-X-diacritic)
* you should add your transformers to the API reference. See `docs/source/api_reference` for examples, should be straightforward copy/paste/adapt
* made some minor suggestions related to the docstrings

fkiraly:
Sorry for the delay here ... in order to merge, the math formulae must render properly, or it will shoot the docs apart.
Here´s an example how to use the "math" environment inside a docstring:

image

iljamaurer:
> Sorry for the delay here ... in order to merge, the math formulae must render properly, or it will shoot the docs apart. Here´s an example how to use the "math" environment inside a docstring:
> 
> image
> 
> top one is docstring embed for "single-dollar", bottom one is "double dollar". What you write inside the math-quote is plain TeX.

I am working on in right now ;-) Have some issues with my environment, I will commit as soon as I have tested the docs ;-)

iljamaurer:
> > Sorry for the delay here ... in order to merge, the math formulae must render properly, or it will shoot the docs apart. Here´s an example how to use the "math" environment inside a docstring:
> > image
> > top one is docstring embed for "single-dollar", bottom one is "double dollar". What you write inside the math-quote is plain TeX.
> 
> I am working on in right now ;-) Have some issues with my environment, I will commit as soon as I have tested the docs ;-)

Thanks again for the feedback @fkiraly. I just committed the fixes. But please wait with the merge until tomorrow evening / Monday, @MrPr3ntice will check it first :-)

MrPr3ntice:
> Thanks again for the feedback @fkiraly. I just committed the fixes. But please wait with the merge until tomorrow evening / Monday, @MrPr3ntice will check it first :-)

Thanks @fkiraly, @iljamaurer! I will check the docstrings and maybe add some paper links to the docs until tomorrow morning.

fkiraly (approved):
Looks good now, thanks!

PR #1773, opened 2021-12-18
[ENH] STL transformer refactor based on 1677

fkiraly:
This is refactoring PR #1677 to be compliant with the general transformer interface, following discussion in there.

Changes to the PR:
* removing `fit` to comply with the transformer interface, changing arg names of `_fit`, `_transform`
* when `transform` is called on new indices, the STL transformation is carried out again
* added `return_components` argument - if `True`, the transformer will return the components as part of the transformed series. In PR 1677, they were stored in "fitted model" attributes which were not accessible to the user (not even via the `get_fitted_params` interface).

This also fixes one bug:
* it appears that the *trend* component was not subtracted when transforming the series, originally? Looks like a mistake.


chrisholder (approved):

sktime/transformations/series/detrend/_deseasonalize.py
```diff
@@ -379,15 +382,14 @@ class STLTransformer(_SeriesToSeriesTransformer):
...
         time.
-
-    Attributes
-    ----------
-    trend_ : pd.Series
-        Trend component.
-    seasonal_ : pd.Series
-        Seasonal components.
-    resid_ : pd.Series
-        Residuals component.
```
aiwalter, line 390: why did you remove Attributes section?

sktime/transformations/series/detrend/_deseasonalize.py
```diff
@@ -434,32 +441,21 @@ def __init__(
...
+        self._X_index = X.index
 
         # The statsmodels.tsa.seasonal.STL can only deal with sp >= 2
-        if sp < 2:
+        if self.sp < 2:
             raise ValueError("sp must be positive integer >= 2")
-        self._stl = _STL(
-            z.values,
-            period=sp,
+        self.stl_ = _STL(
```
aiwalter, line 456: this needs then to be added to Attributes section in docstring as well

sktime/transformations/series/detrend/_deseasonalize.py
```diff
@@ -471,36 +467,46 @@ def fit(self, Z, X=None):
             trend_jump=self.trend_jump,
             low_pass_jump=self.low_pass_jump,
         ).fit()
+        self.seasonal_ = pd.Series(self.stl_.seasonal, index=X.index)
+        self.trend_ = pd.Series(self.stl_.trend, index=X.index)
 
-        self.seasonal_ = pd.Series(self._stl.seasonal, index=Z.index)
-        self.resid_ = pd.Series(self._stl.resid, index=Z.index)
```
aiwalter, line 476: why did you remove `resid_` attribute?

sktime/transformations/series/detrend/_deseasonalize.py
```diff
@@ -517,23 +523,49 @@ def transform(self, Z, X=None):
...
-        return self._transform(z)
+        return self._inverse_transform(z)
 
-    def inverse_transform(self, Z, X=None):
-        """Inverse transform data.
+    def _make_return_object(self, X, stl):
 
-        Returns a transformed version of y.
+        # deseasonalize only
+        transformed = pd.Series(X.values - stl.seasonal, index=X.index)
```
aiwalter, line 531: as discussed in my PR #1677 we wanted to make this flexible with another parameter, so user could chose between `method="deseasonalize"` and `method="detrend"`

sktime/transformations/series/detrend/_deseasonalize.py
fkiraly: it's being returned as the columns of the result now

sktime/transformations/series/detrend/_deseasonalize.py
fkiraly: hm, but aren't the components all returned? The user can then choose by choosing a column.

sktime/transformations/series/detrend/_deseasonalize.py
fkiraly: because the attributes got removed.
The design switched to returning them as the `transform` output.

sktime/transformations/series/detrend/_deseasonalize.py
aiwalter: this is not working when you wnat to put the transformer in a pipeline

sktime/transformations/series/detrend/_deseasonalize.py
aiwalter: I am still not a fan of returning all components at same time, what is the use case for this? User could just use the `statsmodels` implementation. We need to make it compliant with pipelining, that is more the use case of the `sktime` transformer in my opinion

fkiraly:
Superseded by https://github.com/alan-turing-institute/sktime/pull/1677

PR #1774, opened 2021-12-18
[DOC] missing get_test_params in transformer extension template

fkiraly:
This PR adds the missing `get_test_params` function in the transformer extension template.

Also cleans up the preambles of the other templates so it's clear what needs to be implemented.


TonyBagnall:
this still branch still has the old classifier tags in it. these have changed to "X_inner_mtype": "numpy3D",

fkiraly:
well, that should be resolved with an update from `main`, doing this now

fkiraly:
done, is it resolved by update from main? If not, then you (or I) need to update the tag names in a separate PR.

fkiraly:
@TonyBagnall, your issue should be fixed, kindly have a look.

extension_templates/transformer.py
```diff
@@ -241,3 +235,42 @@ def _inverse_transform(self, X, y=None):
...
+        """Get fitted parameters.
+
+        Returns
+        -------
+        fitted_params : dict
+        """
+        # implement here
+
+    # todo: return default parameters, so that a test instance can be created
+    @classmethod
```
chrisholder, line 251: Make this static rather than class you don't use the cls instance in any of the examlpes i've seen

chrisholder (changes requested):

chrisholder (approved):

PR #1775, opened 2021-12-18
[BUG] fix fh -> self.fh in `predict_interval` and `predict_quantiles`

fkiraly:
There was a small mistake in `predict_interval` and `predict_quantiles` which would lead to buggy behaviour if `fh` was already passed in `fit`.

The problem was using `fh` (passed in the function) instead of using `self.fh` which contains the `fh` if it was already passed in `fit`.

I admit that this convention seems very counterintuitive and exists more out of historical legacy.
For this to be more consistent, we should perhaps move to a syntax
`fh = self._check_fh(fh)` instead of `self._set_fh(fh)`
(and expect all developers ot know that the latter modifies `self.fh` which must be used instead of `fh` after that)

However, this is not done in this PR, the scope is only to fix the inconsistency in `predict_interval` and `predict_quantiles`.

FYI, @kejsitake, @SveaMeyer13, @mloning, @aiwalter, @eenticott-shell, comments are appreciated (on this PR or the issue with `_set_fh`).

### DIFF
--- a/sktime/forecasting/base/_base.py
+++ b/sktime/forecasting/base/_base.py
@@ -348,15 +348,17 @@ class BaseForecaster(BaseEstimator):
             Row index is fh. Entries are quantile forecasts, for var in col index,
                 at quantile probability in second col index, for the row index.
         """
+        self.check_is_fitted()
+        # input checks
         if alpha is None:
             alpha = [0.05, 0.95]
-
-        self.check_is_fitted()
         self._set_fh(fh)
         alpha = check_alpha(alpha)
+
         # input check and conversion for X
         X_inner = self._check_X(X=X)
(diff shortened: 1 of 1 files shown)


GuzalBulatova:
Looks all good to me! It has a failing test that I can’t see though, AzureDevOps says “the build not found”, any idea what’s going on there?

chrisholder (approved):

fkiraly:
> any idea what’s going on there?

no idea. Azure has been intermittently broken with timeouts and things like this.
Really looking forward to @lmmentel's #1620.

fkiraly:
"build not found” apparently means that the results could not be retrieved?
But the 185m failure time would correspond well to the "time out" hypothesis, so in addition somehow the logs got lost?

PR #1777, opened 2021-12-19
copy functions and tests from utils/data_io to datasets/_data_io

TonyBagnall:
Phase 1 of the dataset input/output redesign. 
Fixes https://github.com/alan-turing-institute/sktime/issues/1762
this is simply copying functions and tests over as an interim step. thats all it is. it will do nothing more, whoever asks. simply copying code prior to deprecation. Part of: https://github.com/alan-turing-institute/sktime/issues/1778


MatthewMiddlehurst (approved):
Are the new tests just duplicates? If so, may want to remove the old test_data_io file.

TonyBagnall:
> Are the new tests just duplicates? If so, may want to remove the old test_data_io file.

tests are duplicates, but thought I would leave everything as is over in utils until it gets canned.

PR #1780, opened 2021-12-19
Update extension template for predict_quantiles

kejsitake:

Updates the extension template to account for changes introducing predict_quantiles.

No

#### Any other comments?

This was included in another PR https://github.com/alan-turing-institute/sktime/pull/1579, but I am separating to make review easier. I have fixed some of the issues mentioned there.

### DIFF
--- a/extension_templates/forecasting.py
+++ b/extension_templates/forecasting.py
@@ -188,16 +188,11 @@ class MyForecaster(BaseForecaster):
             Forecasting horizon
         X : pd.DataFrame, optional (default=None)
             Exogenous time series
-        return_pred_int : bool, optional (default=False)
-            If True, returns prediction intervals for given alpha values.
-        alpha : float or list, optional (default=0.95)
 
         Returns
         -------
         y_pred : pd.Series
             Point predictions
-        y_pred_int : pd.DataFrame - only if return_pred_int=True
-            Prediction intervals
 
         State change
         ------------
@@ -247,10 +242,10 @@ class MyForecaster(BaseForecaster):
         -------
         pred_quantiles : pd.DataFrame
             Column has multi-index: first level is variable name from y in fit,
-                second level being quantile fractions for interval low-high.
-                Quantile fractions are 0.5 - c/2, 0.5 + c/2 for c in coverage.
+                second level being the quantile forecasts for each alpha.
+                Quantile forecasts are calculated for each a in alpha.
             Row index is fh. Entries are quantile forecasts, for var in col index,
-                at quantile probability in second col index, for the row index.
+                at quantile probability in second-level col index, for each row index.
         """
         # implement here
 
--- a/extension_templates/forecasting.py
+++ b/extension_templates/forecasting.py
@@ -20,12 +20,12 @@ How to use this implementation template to implement a new estimator:
 
 Mandatory implements:
     fitting         - _fit(self, y, X=None, fh=None)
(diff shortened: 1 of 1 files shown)


extension_templates/forecasting.py
```diff
@@ -90,8 +90,10 @@ class MyForecaster(BaseForecaster):
         "requires-fh-in-fit": True,  # is forecasting horizon already required in fit?
         "X-y-must-have-same-index": True,  # can estimator handle different X/y index?
         "enforce_index_type": None,  # index type that needs to be enforced in X/y
+        "capability:pred_int": False,  # does forecaster implement predict_quantiles?
+        # in the future can be changed to capability:predict_quantiles
```
fkiraly, line 94: bit unclear what this means to the person who wants to implement the forecaster. Perhaps say sth like "deprecated and will be renamed to `capability:predict_quantiles` in 0.11.0"?

fkiraly (changes requested):
looks good!

Some comments:
* docstring at the top should also be updated
* one comment about deprecation

kejsitake:
Hi! Thank you for the comments. 

I made these changes and also removed the `pred_int` and `alpha` arguments from `_update_predict_single` and `_predict_moving_cutoff`. Please let me know if this isn't correct.

extension_templates/forecasting.py
```diff
@@ -215,42 +212,46 @@ def _update(self, y, X=None, update_params=True):
...
+        If alpha is iterable, multiple quantiles will be calculated.
+
+        Users can implement _predict_interval if calling it makes this faster.
 
         Parameters
         ----------
-        alpha : float or list, optional (default=0.95)
-            A significance level or list of significance levels.
+        fh : int, list, np.array or ForecastingHorizon
+            Forecasting horizon, default = y.index (in-sample forecast)
```
fkiraly, line 241: there is no `y` here, so the default does not seem to make sense - or does it and I'm missing sth?
If this is a typo, please remove the "default" part.

fkiraly (dismissed):
look good now.

I left one comment on the default `fh`, please fix if that's a typo.

kejsitake:
Fixed the typo! Thank you for the comments!

fkiraly (changes requested):
Thanks, looks great!

I have one more request, looking through this.
I don't think we should import from private modules here, i.e., the `DEFAULT_ALPHA`.

How I would avoid it is to set the default in the `_predict_quantiles` to `None` or, perhaps better, `0.5`. In the body of the function, explain that `alpha` is guaranteed to be a float or list of float by `predict_quantiles` calling this function, which is by default the `DEFAULT_ALPHA` coming from `predict_quantiles`.

kejsitake:
I believe I made this changes. I will keep in mind to include `DEFAULT_ALPHA` in the other PR refactoring `predict_quantiles`. Thank you!

fkiraly (changes requested):
Great, thanks!

I noticed another thing, unfortunately, sorry for the bits-and-pieces here.

The docstring of `_predict_quantiles` is incorrect in describign the return, `pred_quantiles`. This is incorrectly referring to coverage, i.e., to the input of `_predict_interval`. Instead, it shoudl be referring to `alpha`.

kejsitake:
I changed the docstring and tried to read through the file for additional issues, could only spot some deprecated args in `_update.` But please let me know if there is anything else.

fkiraly (approved):
yes, looks great now - thanks!

PR #1785, opened 2021-12-21
Classification dataset tidy

TonyBagnall:
#### Reference Issues/PRs
fixes  https://github.com/alan-turing-institute/sktime/issues/1753
contributes to https://github.com/alan-turing-institute/sktime/issues/1778

1. GunPoint data updated so it is 8 dp not 6 dp
2. Headers fixed to include equalLength tag, double check all datasets match our local copies of ArchiveData on the beast. 
3. return_X_y set to default to True for all data loaders. The only occasion it needs to be false is in the benchmarking module
4. single problem data loaders split off into own file
5. docstrings fixed in benchmarking. Most of the changes relate to this, so it seems to touch a lot of files but the whole thing is pretty minor


docs/source/developer_guide/add_dataset.rst
```diff
@@ -7,7 +7,7 @@ Adding a New Dataset
 Follow these steps to add a new dataset to sktime:
 
 *  Include CSV file or other supported format under :code:`sktime/datasets/data/<dataset-name>`
-*  Add :code:`load_<dataset-name>(...)` function in file :code:`sktime/datasets/base.py`
+*  Add :code:`load_<dataset-name>(...)` function in file
+:code:`sktime/datasets/_single_problem_loaders.py`
 *  Add :code:`<dataset-name>` to the list :code:`__all__ = [...]` in file :code:`sktime/datasets/__init__.py`
 *  Add :code:`<dataset-name>` as argument to method :code:`included_datasets = (...` in file :code:`sktime/sktime/datasets/setup.py`
-*  Add :code:`<dataset-name>` to the list of included problems in file :code:`sktime/sktime/datasets/setup.py`
```
MatthewMiddlehurst, line 13: Is this no longer the case? The setup file still exists.

MatthewMiddlehurst (commented):
There is a bit of inconsistency for between datasets currently, with some having arff files and others not. GunPoint seems to have a train arff but no test also?

Don't believe the arff files are necessary at all, TBH.

TonyBagnall:
thanks for the comments matt, most helpful. Im not sure about the git ignore, I thought it didnt go into main. You have to comment it out though in order to update the datasets, so should probably add something like that in the guide.

.gitignore
```diff
@@ -33,7 +33,7 @@ results/
 MANIFEST
 
 #downloaded datasets
-sktime/datasets/data/
+#sktime/datasets/data/
```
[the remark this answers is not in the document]
TonyBagnall: I just forgot to replace the comment, need to uncomment it to change the datasets via git.

MatthewMiddlehurst:
It seems a bit hacky to have datasets we want to store in the repo in the same folder with datasets we don't (any downloaded from `load_UCR_UEA_dataset`). A consequence of how it is set up currently all using the same `_load_dataset` function.

Leads to the situation where we want to include it in the gitignore to stop new ones being added accidently, and also not so we can change the existing ones and add new stuff if we want to. Downloaded datasets should be a put in their own directory imo and gitignore that.

docs/source/developer_guide/add_dataset.rst
TonyBagnall: not sure what happened there! ta

TonyBagnall:
> 



> It seems a bit hacky to have datasets we want to store in the repo in the same folder with datasets we don't (any downloaded from `load_UCR_UEA_dataset`). A consequence of how it is set up currently all using the same `_load_dataset` function.
> 
> Leads to the situation where we want to include it in the gitignore to stop new ones being added accidently, and also not so we can change the existing ones and add new stuff if we want to. Downloaded datasets should be a put in their own directory imo and gitignore that.

yeah I agree, I will add to the list for the next PR on this https://github.com/alan-turing-institute/sktime/issues/1778

chrisholder (dismissed):

sktime/datasets/tsc_dataset_names.py
```diff
@@ -37,6 +37,7 @@
     "AllGestureWiimoteY",
     "AllGestureWiimoteZ",
     "ArrowHead",
+    "AsphaltObstacles",
```
MatthewMiddlehurst, line 40: There are two other Asphalt datasets other than this, may want to remove this for now or add the other new datasets. Not a blocker though.

MatthewMiddlehurst (dismissed):

sktime/datasets/tsc_dataset_names.py
TonyBagnall: I think we need a proper rethink about how to handle all this tbh!

PR #1786, opened 2021-12-21
[ENH] experimental: hierarchical time series scitype

fkiraly:
This PR adds a scitype `Hierarchical` with a single mtype, with 2 or more levels of hierarchy.

The representation is a `pandas.DataFrame` with 3 or more levels, out of which the lowest one must correspond to an `sktime` supported time index type.

The scitype can be used in input checks and conversions.

### DIFF
--- a/sktime/datatypes/_hierarchical/_check.py
+++ b/sktime/datatypes/_hierarchical/_check.py
@@ -27,13 +27,15 @@ msg: str - error message if object is not valid, otherwise None
 metadata: dict - metadata about obj if valid, otherwise None
         returned only if return_metadata is True
     fields:
-        "is_univariate": bool, True iff all series in panel have one variable
+        "is_univariate": bool, True iff all series in hier.panel have one variable
         "is_equally_spaced": bool, True iff all series indices are equally spaced
         "is_equal_length": bool, True iff all series in panel are of equal length
         "is_empty": bool, True iff one or more of the series in the panel are empty
-        "is_one_series": bool, True iff there is only one series in the panel
+        "is_one_series": bool, True iff there is only one series in the hier.panel
+        "is_one_panel": bool, True iff there is only one flat panel in the hier.panel
         "has_nans": bool, True iff the panel contains NaN values
-        "n_instances": int, number of instances in the panel
+        "n_instances": int, number of instances in the hierarchical panel
+        "n_panels": int, number of flat panels in the hierarchical panel
 """
 
 __author__ = ["fkiraly"]
@@ -92,6 +94,7 @@ def check_pdmultiindex_Hierarchical(obj, return_metadata=False, var_name="obj"):
         return _ret(False, msg, None, return_metadata)
 
     inst_inds = obj.index.droplevel(-1).unique()
+    panel_inds = inst_inds.droplevel(-1).unique()
 
     check_res = [
         check_pdDataFrame_Series(obj.loc[i], return_metadata=True) for i in inst_inds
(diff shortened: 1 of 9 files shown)


sktime/datatypes/_hierarchical/_check.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+def _ret(valid, msg, metadata, return_metadata):
+    if return_metadata:
+        return valid, msg, metadata
+    else:
+        return valid
+
+
+def check_pdmultiindex_Hierarchical(obj, return_metadata=False, var_name="obj"):
```
chrisholder, line 81: Don't capatalised Hierarchical

sktime/datatypes/_hierarchical/_check.py
```diff
@@ -0,0 +1,127 @@
...
+    nlevels = obj.index.nlevels
+    if not nlevels > 2:
+        msg = f"{var_name} have a MultiIndex with 3 or more levels, found {nlevels}"
+        return _ret(False, msg, None, return_metadata)
+
+    inst_inds = obj.index.droplevel(-1).unique()
+    panel_inds = inst_inds.droplevel(-1).unique()
+
+    check_res = [
+        check_pdDataFrame_Series(obj.loc[i], return_metadata=True) for i in inst_inds
```
chrisholder, line 100: Where is this check_pdDataFrame_Series please burn it and rename it properly to: check_pd_dataframe_series

sktime/datatypes/_hierarchical/_convert.py
```diff
@@ -0,0 +1,25 @@
...
+#  2nd element = convert to - type
+#  3rd element = considered as this scitype - string
+# elements are conversion functions of machine type (1st) -> 2nd
+
+convert_dict = dict()
+
+
+def convert_identity(obj, store=None):
+
+    return obj
```
chrisholder, line 20: I assume this is intended more logic in the future

sktime/datatypes/_hierarchical/_registry.py
```diff
@@ -0,0 +1,19 @@
...
+    "MTYPE_LIST_HIERARCHICAL",
+]
+
+
+MTYPE_REGISTER_HIERARCHICAL = [
+    (
+        "pd_multiindex_hier",
+        "Hierarchical",
+        "pd.DataFrame with MultiIndex",
+    ),
```
chrisholder, line 16: Is it a good idea to mix dots and underscores with naming (and capitals) I'd suggest all lower and underscores.

chrisholder (changes requested):

sktime/datatypes/_hierarchical/_convert.py
fkiraly: no, not really - this is the "identity conversion" which is needed if you want to convert an mtype to itself. Instead of raising an error, it simply does nothing. This will be called whenever you already pass the "right" mtype to an estimator, which is probably a majority use case for many estimators.

sktime/datatypes/_hierarchical/_registry.py
fkiraly: yes, exactly.

I'd also like to move to all lower case and underscores, but we can't easily change this for the currently existing mtypes/scitypes, it would require deprecation.

But we can changei t for the new ones, that's why I used all lower case and underscore which is inconsistent with `pd-multiindex` (which I'd like to rename to `pd_multiindex` later).

fkiraly: but there wouldn't be an action here, right? the third string is not a name, but an English description of the mtype.

sktime/datatypes/_hierarchical/_check.py
fkiraly: this is not added by the PR, we can do this in a separate PR - let's not widen scope, or I have to go through multiple files.
I can carry out a lower_snake_case renaming later.

sktime/datatypes/_hierarchical/_check.py
fkiraly: changed, now lower_snake_case

chrisholder (approved):

PR #1790, opened 2021-12-22
[ENH] Transformer base class DRY-ing, and `inverse_transform`

fkiraly:
This PR implements two main changes:

* making the transformer base class more DRY, by moving repeated boilerplate code into private convenience methods, roughly in line with @TonyBagnall's refactor of TSC
* adding the base level check/conversion functionality for `inverse_transform` and `update` which were not yet implemented

A few things had also to be changed in the testing suite:
* so far, whether `inverse_transform` was tested was inferred from whether the method was available. This is no longer possible, since `inverse_transform` always exists, following the `fit/_fit` architecture.
Instead, the check now relies on a new capability tag, `"capability:inverse_transform"` which defaults to `False`. In the tests, which function is implemented is seen through the new `_has_capability` utility.
* as estimators can now directly inherit from `BaseTransformer`, this was added to the checks of the `_make_args` family of utilities and a few tests
* this also contains the bugfix in #1789 which helps with avoiding repetitive checks and metadata retrieval.

And one change to the `datatypes` module:
* conversion store of `Series` scitype now also remembers the `index` of `pandas` objects. This is now used in the `transform` and `inverse_transform` where the index needs to be restored after conversion to an internal 2D `numpy`.


sktime/transformations/base.py
```diff
@@ -560,11 +436,387 @@ def fit_transform(self, X, y=None, Z=None):
...
+                subject to sktime mtype format specifications, for further details see
+                    examples/AA_datatypes_and_datasets.ipynb
+        y : Series or Panel, default=None
+            Additional data, e.g., labels for transformation
+        Z : possible alias for X; should not be passed when X is passed
+            alias Z will be deprecated in version 0.10.0
+
+        Returns
+        -------
+        inverse transformed version of X
```
chrisholder, line 465: What type is returned

sktime/transformations/base.py
```diff
@@ -617,6 +869,45 @@ def _transform(self, X, y=None):
...
+        ----------
+        X : Series or Panel of mtype X_inner_mtype
+            if X_inner_mtype is list, _inverse_transform must support all types in it
+            Data to be transformed
+        y : Series or Panel, default=None
+            Additional data, e.g., labels for transformation
+
+        Returns
+        -------
+        inverse transformed version of X
```
chrisholder, line 887: What is the type of the return?

chrisholder (approved):

sktime/transformations/base.py
fkiraly: I added more detail

sktime/transformations/base.py
fkiraly: I added detail and a reference to the extension templates

fkiraly:
@chrisholder, I hope I've addressed the docstring issues? I added more detail, both in the base class and extension template, and referenced the extension template from the base class.

chrisholder (approved):

PR #1792, opened 2021-12-22
[MNT] Refactor legacy test config

lmmentel:

- replace legacy `check_` with `test_` in test configuration
- exclude `test_fit_does_not_overwrite_hyper_params` for `FeatureUnion`
- small refactoring of the tests

No.

Check if all the test are running.

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/tests/test_all_estimators.py
+++ b/sktime/tests/test_all_estimators.py
@@ -62,9 +62,6 @@ def pytest_generate_tests(metafunc):
     """
     # get name of the test
     test_name = metafunc.function.__name__
-    # previously, the tests were called "check_", so temporarily we do this
-    #  (ultimately, entries in the "exclude" list should be renamed)
-    test_name = test_name.replace("test_", "check_")
 
     # tests can be tests for classes or instances
     # tests for classes use estimator_class fixture name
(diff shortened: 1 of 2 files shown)


sktime/tests/test_all_estimators.py
```diff
@@ -168,23 +165,26 @@ def test_estimator_tags(estimator_class):
...
 
     # Usually estimators inherit only from one BaseEstimator type, but in some cases
     # they may be predictor and transformer at the same time (e.g. pipelines)
-    n_base_types = sum(
-        [issubclass(Estimator, cls) for cls in VALID_ESTIMATOR_BASE_TYPES]
+    assert (
+        2
+        >= (
+            sum(
+                [issubclass(estimator_class, cls) for cls in VALID_ESTIMATOR_BASE_TYPES]
```
fkiraly, line 178: I think this change makes the assertion more obscure to understand. Bracket overload...
Would you feel strongly about reverting it to its original state?

fkiraly (approved):
Thanks a lot!

Agree with most of the "side changes", they improve style and readability ... except the one where you introduce 100 brackets. Do you maybe want to revert this? Not feeling strongly though.

sktime/tests/test_all_estimators.py
lmmentel: This was autorefactored by an extension I use. I can revert back if it bother you.

fkiraly (approved):
Thanks!.
Unix test failures do not seem content/change related.

Issue #1778, opened 2021-12-19
[ENH] classification/clustering data loading restructuring


TonyBagnall:
I'm rejigging that data input output for classifiers and will do it in several stages
1. Done
2. Add a new data loader that returns 2D numpy for equal length univariate, 3D numpy for  equal length multivariate, DataFrame for any unequal length problems
https://github.com/alan-turing-institute/sktime/issues/1718
3. Add writers that can write 2d/3d numpy and pandas
https://github.com/alan-turing-institute/sktime/issues/1753
https://github.com/alan-turing-institute/sktime/issues/1756
4. Done
5. Adjust the web downloader to put in a different download location, then gitignore that
6. Adjust all the single problem loaders to use the new loader, fix any resulting notebook issues where pandas are expected, move all the single loaders into a separate file. Possibly also split input/output into different files 
7. Provide a general purpose loader than can handle .ts, .csv, .arff and .txt
8. Update the notebook to demonstrate new usage
  

Done
1. Clone functions currently in utils/data_io.py into datasets/_data_io.py 
https://github.com/alan-turing-institute/sktime/pull/1777
4. Add univariate unequal length dataset (AsbestosObstacles) [actually just used PLAID which is already there], sort out multivariate unequal length to remove nan padding, sort out other data formatting.

fkiraly:
Added a descriptive title.

Re. point 2 and 3, some random thoughts. This could be most easily achieved by:
* having logic that loads to a very general format, e.g., `pd-multiindex`
* having an `mtype` argument in the loader, which determines the output format
* using `convert` at the end of the loader, from the very general format, instead of writing seven or so loaders or pieces of logic

TonyBagnall:
@fkiraly so for the .ts format, the idea is to read the data type from the file header, removing the need to explicitly check or convert. We are starting to look at larger problems (EEG are often 10-100 GB), so we need to minimize unnecessary checking. If the file does not match the header (e.g. unequal length data registered as equal in header) it will throw an exception. For other formats, the idea is to load into dataframe as we always have done, then use check and convert it get it into the best format.

PR #1799, opened 2021-12-24
[ENH] user selected return for classification problems data loading functions 

TonyBagnall:
adjusted now to only offer the option of return type, not determine return type from the data. Part of #1778. 

1. Includes a new function load_from_tsfile that will ultimately replace load_from_tsfile_to_dataframe in all loading helper functions. load_from_tsfile returns 2D/3D numpy for equal length problems, DataFrame for unequal problems
EDIT: the return type is now selected by the user, and the data driven version has gone into contrib. 
2. load_from_tsfile now actually checks for the equalLength flag in the header. 
3. A private method _load_provided_dataset that will replace _load_dataset in the baked in loaders. The new method simply looks in the datasets/data directory for the problem requested and fails if it is not there, rather than attempt to download from tsc.com
6. Adjusts the current _load_dataset to work in local_data directory when it downloads from tsc.com, and change the gitignore to local_data


fkiraly (changes requested):
* there are a lot of changes to tests - these seem unrelated to the topic of this PR. Would you like to move these fixes in a different PR, for quick approval, and making the review of the dataIO stuff easier?
* if you want the return polymorphic, why not put a `convert_to` at the end, and give it `return_mtype` arg? The current behaviour seems very rough and confusing to the user, since the return mtype is now inconsistent and will depend on properties of the dataset (univariate, multivariate, equal length)
* in particular, I'd strongly advise that there is one default `return_mtype`. We've soft-adopted the design principle earlier (core dev vote) that there is one single output mtype.
* a clear strategy would be, always convert it internally first to `pd_multiindex` or `df-list`, as the most general `Panel` `mtype`-s, and then convert it to `return_mtype`.

TonyBagnall:
> * there are a lot of changes to tests - these seem unrelated to the topic of this PR. Would you like to move these fixes in a different PR, for quick approval, and making the review of the dataIO stuff easier?
> * if you want the return polymorphic, why not put a `convert_to` at the end, and give it `return_mtype` arg? The current behaviour seems very rough and confusing to the user, since the return mtype is now inconsistent and will depend on properties of the dataset (univariate, multivariate, equal length)
> * in particular, I'd strongly advise that there is one default `return_mtype`. We've soft-adopted the design principle earlier (core dev vote) that there is one single output mtype.
> * a clear strategy would be, always convert it internally first to `pd_multiindex` or `df-list`, as the most general `Panel` `mtype`-s, and then convert it to `return_mtype`.

1. no I cant separate out the changes to tests, because they are made necessary 
2. the basic logic of classifiers is that univariate equal length goes into 2D, multivariate equal length into 3D and unequal length into DataFrame. This is completely clear to me, not sure why it would confuse a user. It is entirely consistent with the recent base class restructure. 
3.The vote was on the y variable, which is always numpy with us now. 
4. It seems pointless to put data into a more complex data structure then converting it back into a simpler one

TonyBagnall:
ote that I will have do redo the tests as they wil lconflict with https://github.com/alan-turing-institute/sktime/pull/1799, but thats ok, only took 10 mins

TonyBagnall:
and I would add, the whole point is to make the return type dependent on the data type. Thats how classification is now structured. Rather than but everything in nested data frames, we now use the best data structure for the problem. This is how i have always wanted it, but always deferred. Now the base class structure has changed to allow it, I want it this way for the UCR/UEA data sets in .ts format. Users can put their own data in numpy or dataframes, which ever they think is best.

fkiraly:
> the basic logic of classifiers is that univariate equal length goes into 2D, multivariate equal length into 3D and unequal length into DataFrame. This is completely clear to me, not sure why it would confuse a user.

Well, try to explain it to a user and see what they say.
"The function return is 2D numpy, except when it is 3D numpy, or pandas DataFrame. It's 3D numpy when we can't put it into 2D numpy, and pandas when we can't put it into 3D numpy. Of course we could always return pandas and convert it to numpy 3D on request, but we don't do that"
I honestly challenge you to explain this to someone who's not worked on this.

I strongly recommend to have a `return_type` argument that controls what comes out of this? And not do conditional type, but raise error messages if the return can't be represented in a given type.

TonyBagnall:
It's easy to explain:
If it is univariate, equal length, it returns 2d
If it's multivariate, equal length, it returns a 3d
If it's unequal length, if returns data frame
If this is not understood, then they have not understood the nature of the data and task, so maybe they will then try do so. It is not a user driven option for this specific function.

fkiraly:
> If this is not understood, then they have not understood the nature of the data and task, so maybe they will then try do so. It is not a user driven option for this specific function.

The problem with this is that you are loading a file.
You don't know *in advance* which properties the data will have, e.g., univariate, multivariate, etc. You are introducing what is effectively a *non-inspectable condition* in the output type, so from a user inspective the output type will be more or less random.

I'd propose an indicative/non-binding core dev vote after the holidays?

TonyBagnall:
You do not in advance the properties of the data you are loading, thats the whole point of the ts format, the meta data is in the header. The output is not in any way random, it is a property of the dataset, and to think the user will know nothing about the property of the data they are loading implies they know nothing about the data they are loading, even to the point they do not understand it is a classification problem. Any output from this function can be passed to any sktime classifier. I generallly think a user will find it easier to understand the difference between a 2D or 3D numpy than it is to understand the whole scitype mtype definition you have invented.  

So, sure we can have a vote if you want, but what would we be voting on? The very existence of the loading function, its use in the single problem loaders or planned use in general loading (or all of the above)? I can just put it in contrib and limit use to my group if you object this strongly.

fkiraly:
> So, sure we can have a vote if you want, but what would we be voting on?

I'd propose, three options, docstring 1, docstring 2, "neither is good"?

Docstring 1 has the return as you describe. Docstring 2 has a default return of `pd-multiindex`, and an `return_mtype` arg that allows to pick others such as `numpy3D` (with error raised if not possible.

Here's my docstring:
```python
    """Load Panel data from a .ts file into an sktime compatible format.

    Parameters
    ----------
    file_path: str
        File lication of the .ts file to read, full path and file name.
    return_mtype: str (optional), default="pd-multiindex"
        in-memory (mtype) format in which the data is loaded
        mtype strings with explanation are in datatypes.MTYPE_REGISTER

    Returns
    -------
    X : object of type `return_mtype`
        the loaded data

    Raises
    ------
    RuntimeError if no file at location exists
    RuntimeError if file exists but is in incorrect format
    RuntimeError if file exists and is in correct format, but `return_mtype` is too specific.
        example: data has unequal length series, but `return_mtype=numpy3D`
            which can deal only with equal length series
    """

```

TonyBagnall:
Pd-multiindex is not supported by the estimators, so can't be that unless we redesign everything which I don't want to do. Default should be nested data frame as it is now. I don't mind changing it to default to that or switch on user parameter I suppose although it will get messy with exceptions when data doesn't allow the choice.

TonyBagnall:
it needs tidying up, not ready for review, but in principle is structure for load_from_tsfile acceptable @fkiraly? If the user passes an invalid option (say return 2d for multivariate data) would you rather raise an error or just ignore and return the default type? I assume the former

fkiraly:
>Pd-multiindex is not supported by the estimators, so can't be that unless we redesign everything which I don't want to do. 

?? I thought it was, since you've refactored the base class?
I looked at it and there are some minor issues, I'm making a PR with suggested fixes.

>Default should be nested data frame as it is now. I don't mind changing it to default to that or switch on user parameter I suppose although it will get messy with exceptions when data doesn't allow the choice.

I don't mind, but all inputs should be supported.

fkiraly:
@TonyBagnall, I just checked `BaseClassifier`, and I think indeed all inputs are supported after your refactor!
While looking, I found a few bugs and inaccurate docstrings, I've fixed them here:
https://github.com/alan-turing-institute/sktime/pull/1804

sktime/contrib/data_io.py
```diff
@@ -0,0 +1,118 @@
...
+    is_first_case = True
+    instance_list = []
+    class_val_list = []
+    line_num = 0
+    num_dimensions = 0
+    num_cases = 0
+    # equal_length = True
+    with open(full_file_path_and_name, "r", encoding="utf-8") as file:
+        _meta_data = _read_header(file, full_file_path_and_name)
+        for line in file:  # Will this work?
```
chrisholder, line 58: Remove comment :joy:

chrisholder (changes requested):

fkiraly (changes requested):
I'm generally happy with the form of `load_from_tsfile` and the input/output args.

I have one major reservation which I would consider a blocker: you are introducing `return_data_type` strings which are *not* the same strings as we use in the `datatypes` module, this will be confusing to users.

Can you kindly use the same strings?
This could easily done by using a `convert_to` instead of equivalent, custom functionality in lines 385-406 of `_data_io` (plus some error handling, perhaps).
A beneficial side effect would be that the loading functions automatically support all `sktime` mtypes. 
If you don't want to couple the function to the `datatypes` module, you could have a "bare" function which is not coupled to anything, and an enriched function which adds the `convert_to`.

fkiraly:
I think this is still a valuable contribution, @TonyBagnall - no need to close.

Consider handing it over if you don't have the time to work on it.

Or, if my review is a deal breaker, kindly let me know. I thought it was a small but important thing to ask that we are consistent across the package in our naming of data formats.

TonyBagnall:
closed because conflicted, I will revisit at a later date

TonyBagnall:
reopened, as next in line. @fkiraly "Can you kindly use the same strings?" I thought the strings were numpy3D and numpy2D, and I use these (albeit also allowing lower case and np3d as an alternative)
        return_data_type = return_data_type.strip().lower()
        if return_data_type == "numpy2d" or return_data_type == "np2d":

TonyBagnall:
added numpyflat as in input option for 2d numpy

fkiraly (approved):
Yes, these are the right strings now, thanks!

chrisholder (approved):

PR #1816, opened 2021-12-31
Correct the bash error propagation for running notebook examples

lmmentel:

Update the bash script parameters to fail when one of the subcommands fails.
##### For all contributions

##### For new estimators

### DIFF
--- a/build_tools/run_examples.sh
+++ b/build_tools/run_examples.sh
@@ -1,7 +1,8 @@
 #!/bin/bash
 
 # Script to run all example notebooks.
-set -euxo -pipefile
+set -euxo pipefail
+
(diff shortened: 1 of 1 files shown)


fkiraly:
Good, it fails - which proves that your hypothesis about the error collection was correct, sicne `test-examples` is failing with the consequence of the bug fixed in #1813.

I've merged #1813 now. The hypothesis would be further confirmed if fixing the error causes the test failure to go away (or the notebook failing at another point).

fkiraly:
(updating from main now to see)

fkiraly:
ok! Looks like I did guess right about what the bug was. `test-examples` seems to run now, will merge once the other tests pass.

fkiraly (approved):

PR #1806, opened 2021-12-30
[ENH] vectorization/iteration utility for `sktime` time series formats

fkiraly:
This PR introduces a vectorization/iteration utility for the two use cases:
* a `Panel` is passed to a series transformer, which then loops through all the series in the panel. That is, `fit` and `transform` are applied iteratively to all series, the result is then combined back to a panel.
* a `Panel` is passed to a simple forecaster, which loops through all the series in a simple panel or a hierarchical panel. That is, `fit` and `predict` are applied iteratively, and the result is combained back to a panel forecast.

This utility would facilitate to use all current forecasters and transformers in an automatically vectorized fashion, where `fit` etc are automatically cast to the dimensions defined in the panel and hierarchical indices.

The current transformer base class refactor in #1790 does not yet use this utility, but could be easily refactored.
The extension to hierarchical forecasting would also profit from this, the original motivation to have this is that this would constitute a second use case. In consequence, this PR also makes use of PR #1786 (the experimental hierarchical time series scitype).
This is also dependent on #1805 which was uncovered through testing.

In detail, content introduced in this PR:
* the module `datatypes._vectorize` which contains the class `VectorizedDF`. It enables easy looping through slices of `Series` and `Panel` data which are automatically converted to `sktime` formatting conventions of `pandas.DataFrame` based containers.
* a series of tests for `VectorizedDF`, in `datatypes.test.test_vectorize`

PR #1818, opened 2022-01-01
[ENH] Forecaster base class - extension to hierarchical forecasting

fkiraly:
This experimental draft PR extends the `BaseForeaster` with the following functionality:

* support for `Hierarchical` and `Panel` inputs and hence forecasts of that type. This is needed to support panel forecasting, global forecasting, and hierarchical forecasting.
* automated vectorization/looping functionality used in any forecaster which is not natively capable of handling `Panel` or `Hierarchial`. In that case, the forecaster loops over instances and/or hierarchies.

This PR is the first step to implement [STEP 14 for hierarchical/panel/global forecasting](https://github.com/sktime/enhancement-proposals/pull/17).

It relies on the following PR which should be merged first:
* the `Hierarchical` scitype for collections of series with two or more hierarchy levels, from #1786
* the series vectorization/iteration utility from #1806
* cutoff getter/updater utility in #2115

Other minor changes included in this PR:
* adaptations to existing forecaster related tests to be compliant with the new interface.
* a refactoring of the input checking logic in `BaseForecaster`.
* partial input checks are removed from `check_equal_time_index`, in favour of input checks already done by `check_is_scitype` and `check_is_mtype`

Further additions to the test framework will be necessary to consolidate this prototype - for that, we'll need "scenarios", since the current testing framework can handle only a single combination of inputs for `fit` and `predict`, but we need to check whether forecasters can handle `Series`, `Panel`, and `Hierarchical` inputs correctly. I'll work on this separately, since we also need this for the transformer refactor.


fkiraly:
FYI @danbartl - for you to play around with.
I'd expect this to be very buggy, until I've upgraded the testing framework to "scenarios".

I'd suggest the following approach:
* create a new branch with the intent to use this new base class instead of your "hacked" base class (feel free to make a draft PR too, might be useful for me to debug)
* best way is probably: update #1779 to be in line with main; branch off your #1779, in the new branch merge #1818
* try to move to this new base class from your "hacked" base class
* try out manually functionality you'd expect to work and see what happens
* write down any bugs you encounter and/or tests you think would be useful, post them here

danbartl:
@fkiraly 
Some comments regarding the PR, not sure where to put them since they cover chunks that were not edited in this PR:
_base.py:

    def _set_cutoff_from_y(self, y):
Here we need a case distinction or a change, since y.index[-1] does not work. In my PR I changed it to:

            elif y_mtype == "pd-multiindex":
                self._cutoff = _get_time_index(y)[-1]

but just _get_time_index(y)[-1] would also work. I integrated changes to _get_time_index in my PR. Another issue where we need a case distinction in this method is the mtype detection of y, which shoud not be strictly Series:

        if self.get_tag("y_inner_mtype") == "pd-multiindex":
            y_mtype = mtype(y, as_scitype="Panel")
        else:
            y_mtype = mtype(y, as_scitype="Series")

Finally in _sktime.py we need to work around the conversion of Panel data to Series data in the step

y_return = pd.Series(y_pred, index=index):

Here the self._predict_last_window in the RecursiveReducer class already returns Panel data, so no conversion needs to take place, not sure how to cover that.

    def _predict_fixed_cutoff(
        self, fh, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA
    ):
  ...
        if self.get_tag("y_inner_mtype") == "pd-multiindex":
            y_return = y_pred
        else:
            y_return = pd.Series(y_pred, index=index)

        return y_return

danbartl:
@fkiraly Regarding the predict funtion in _base.py I made the following changes we already discussed:

            # convert to output mtype, identical with last y mtype seen
            if self.get_tag("y_inner_mtype") == "pd-multiindex":
                y_out = convert_to(
                    y_pred,
                    self._y_mtype_last_seen,
                    as_scitype="Panel",
                    store=self._converter_store_y,
                )
            else:
                y_out = convert_to(
                    y_pred,
                    self._y_mtype_last_seen,
                    as_scitype="Series",
                    store=self._converter_store_y,
                )

and also  in

fkiraly:
@danbartl, thanks a lot, very useful pointers. Will comment below on how these are addressed.

fkiraly:
>`_set_cutoff_from_y`: Here we need a case distinction or a change, since `y.index[-1]` does not work.

I introduced a case distinction: for `MultiIndex`, we do `self._cutoff = y.index.get_level_values(-1).max()`, the latest time point seen in the new series.

I also simplified logic in how `_X` and `_y` are stored. Now the data is converted to either long `pd.DataFrame` formats or `np.ndarray`, so the treatment can be more explicit and avoids repeated checking via `mtype`. Some case distinction magic was needed for `np.ndarray`, since 3D has time in axis 2, while 2D and 1D have time in axis 0 (this comes from needing to be consistent with the time seires classification module, don't ask).

>Regarding the predict funtion in _base.py I made the following changes we already discussed:

In `predict`, the fix would have been easier:

```python
            y_out = convert_to(
                y_pred,
                self._y_mtype_last_seen,
                store=self._converter_store_y,
            )
```
i.e., just removing the assertion that it is `Series`. The `_y_mtype_last_seen` is enough to determine the output conversion.

Similarly, in `_check_X_y`, the `as_scitype` should be `y_scitype` and `X_scitype`. Setting it to `"Panel"` will break everything in the `"Series"` case.

The `_predict_fixed_cutoff` I'm not sure, could you perhaps explain in more detail what is happening there?

danbartl:
@fkiraly
Thanks the conversion in ``def _check_y_X `` works now.

However, the the new change to  ``def _update_y_X  `` will still result in an attempt to convert to series

``
    def _update_y_X
...
             y = convert_to(
                y,
                to_type=[
                    "np.ndarray",
                    "numpy3D",
                    "pd.Series",
                    "pd.DataFrame",
                    "pd-multiindex",
                    "pd_multiindex_hier",
                ],
            )
``
Particularly in this code segment

``
    if as_scitype is None:
        if isinstance(to_type, str):
            as_scitype = mtype_to_scitype(to_type)
        else:
            as_scitype = mtype_to_scitype(to_type[0])
    elif not isinstance(as_scitype, str):
        raise TypeError("as_scitype must be a str or None")

    from_type = infer_mtype(obj=obj, as_scitype=as_scitype)
``

to_type resolves to ['np.ndarray', 'numpy3D', 'pd.Series', 'pd.DataFrame', 'pd-multiindex', 'pd_multiindex_hier']
and as_scitype = mtype_to_scitype(to_type[0]) will return Series, since mtype_to_scitype for 'np.ndarray resolves to series.

from_type = infer_mtype(obj=obj, as_scitype=as_scitype)

will then run on error since it checks for series type (which is not valid, since we defined y as pd-multiindex dateframe.

Do you have an idea how to fix this?

danbartl:
Another point:
```
    def _set_cutoff_from_y(self, y):
        """Set and update cutoff from series y.

        Parameters
        ----------
        y: pd.DataFrame (long format) or np.array (1D, 2D, 3D)
            Time series from which to infer the cutoff.

        Notes
        -----
        Set self._cutoff to latest index seen in `y`.
        """
        if len(y) > 0:
            if isinstance(y, (pd.Series, pd.DataFrame)):
                if not isinstance(y.index, pd.MultiIndex):
                    # if index is not a multiindex, last index value is latest
                    self._cutoff = y.index[-1]
                else:
                    # otherwise, we need to look in last level, take the max index
                    self._cutoff = y.index.get_level_values(-1).max()
```
will drop the frequency, which we need e.g. for shifting the cutoff by window length. The previous solution     
``
self._cutoff = _get_time_index(y)[-1]
``

together with the suggested change to _get_time_index would retain the index.

```
def _get_time_index(X):
    """Get index of time series data, helper function.
    Parameters
    ----------
    X : pd.DataFrame
    Returns
    -------
    time_index : pandas Index
        Index of time series
    """
    # assumes that all samples share the same the time index, only looks at
    # first row
        if isinstance(X.index, pd.MultiIndex):
            return X.xs(
                X.index.get_level_values("instances")[0], level="instances"
            ).index
        else:
            return _get_index(X.iloc[0, 0])

    elif isinstance(X, pd.Series):
        if isinstance(X.index, pd.MultiIndex):
            return X.xs(
                X.index.get_level_values("instances")[0], level="instances"
            ).index
        else:
            return _get_index(X.iloc[0])

    elif isinstance(X, np.ndarray):
        return _get_index(X)

    else:
        raise ValueError(
            f"X must be a pandas DataFrame or Series, but found: {type(X)}"
        )

```

In general, due to the complexity of accessing the time index, I think reworking all access of index to _get_time_index would have a certain charm I think.

sktime/datatypes/_hierarchical/_check.py
```diff
@@ -78,7 +78,7 @@ def _ret(valid, msg, metadata, return_metadata):
         return valid
 
 
-def check_pdmultiindex_hierarchical(obj, return_metadata=False, var_name="obj"):
+def check_pdmultiindex_Hierarchical(obj, return_metadata=False, var_name="obj"):
```
fkiraly, line 81: note to self: this should be lower_snake_case, should be reverted. Probably a merge accident.

fkiraly:
Thanks for all the help!

I've thought about a few things here in relation to the "machinery" this touches, and I've arrived at the following opinions:

* we probably need to revisit the `convert_to` functionality. I think it should be possible to pass arguments that are multiple scitypes, although I have to work out the exact behaviour that makes sense. The entire module may benefit from closer scrutiny, the arguments `as_scitype` also seem a bit counterintuitive in hindsight.
* Using `_get_time_index` is a good idea. However, it is located in an obscure place. I think it should be moved to a dedicated module in `datatypes`, such as `accessors` (name to be determined).
* The case of `np.ndarray` is not as easy as it looks. Indices are to be interpreted as a `RangeIndex` when it is *first seen*. However, in update, the index should start counting at `cutoff`, not at `0`. This means, there needs to be dedicated functionality that encodes the logic of "interpret the index as this, if data has already been seen". The right location would be that new module.

fkiraly:
PS: we're almost there with the testing framework, so it should be easier for me to detect bugs like the above once it's up and running.

danbartl:
ok sounds very promising, looking forward to that. I tried. 

Unfortunately my other PR on LaggedWIndowSummarizer runs on error 
https://github.com/alan-turing-institute/sktime/pull/1924
basically, same issue with convert_to, for the panel case. I will now continue debugging however for the univariate case.

fkiraly:
> basically, same issue with convert_to, for the panel case. I will now continue debugging however for the univariate case.

Argh - ok, so let me prioritize the `convert_to`, I will do this in a separate PR then so you can use it in both.

fkiraly:
ok, here you go, first stab:
https://github.com/alan-turing-institute/sktime/pull/1929

fkiraly (commented):
reviewed by @danbartl, and falls under "2 weeks" rule for work on hierarchical/probabilistic content. Merging.
(once tests pass on up-to-date `main`)

sktime/datatypes/_hierarchical/_check.py
fkiraly: fixed

Issue #1740, opened 2021-12-10
[ENH] transformers refactor, expand tests to cover auto-vectorization, panel types, multiple input mtypes


fkiraly:
With the transformers refactor done, the test coverage should be expanded to cover:
* auto-vectorization functionality
* input/output scitype cases
* panel transformations, panel types
* multiple input mtypes

fkiraly:
done by migration to `TestAllTransformers`

Issue #1684, opened 2021-12-03
[ENH ] Global/panel forecasting base class & tests refactor - modification route


fkiraly:
@danbartl's alternative proposal to modify existing forecasting base classes & tests to allow quick upgrade to global/panel.

This needs to be presented as PR for discussion and feedback.

Related issue: #1272 (sub-case of panel forecasting).

fkiraly:
Review prototype and near-final implementation is in https://github.com/alan-turing-institute/sktime/pull/1818.

Issue #1683, opened 2021-12-03
[ENH] Global/panel/hierarchical forecasting base class design & first implementation


fkiraly:
We've aligned in the workstream to design a *joint* base class for global, panel, and hierarchical forecasting.
This can be a separate base class or later adapted as an extension to the existing forecasting base class.

A STEP for this is here:
https://github.com/sktime/enhancement-proposals/pull/17

We need to:
- [ ] align on, and merge the STEP
- [ ] implement a prototype (perhaps first without the "specialist" methods, only `fit`/`predict`/`update`) with a mock concrete and rudimentary tests

fkiraly:
I believe @satya-pattnaik was working on a prototype, although I notice we haven't agreed on or merged the STEP.

fkiraly:
in the meeting, @danbartl has proposed an alternate design or process based on gradual change rather than new base class. He'll summarize, issue for that is #1684.

fkiraly:
Ultimately, we went with the "modification" route as described in #1684 - the for review prototype and near-final implementation is in #1818.

Issue #1674, opened 2021-12-01
[ENH] `test_methods_have_no_side_effects` should test for side effects in all key arguments


fkiraly:
See #1673.

The prophet wrapper has side effects on `X`.

This was not discovered by `test_all_estimators/test_methods_have_no_side_effects` since its key dependency `_make_args`, which creates default arguments, does not create non-trivial `X` (this is set to `None`).

`_make_args` should be modified that all arguments of critical functions are properly tested for side effects.

fkiraly:
This is automatically happening since we upgraded to test scenarios in #1819, accidentally this issue did not get linked and closed.

PR #1819, opened 2022-01-02
[ENH] Test scenarios for advanced testing

fkiraly:
This PR introduces a class `TestScenario` which models a script to run methods sequentially with certain arguments and retrieve the outcome, plus a check whether the script is applicable to an object (e.g., estimator).

This addresses a deficiency in the current testing framework, in that unit tests can retrieve only a single argument combination per estimator scitype, via the `_make_fit_args` etc family of functions.

This limitation prevents addressing a few existing and future testing requirements:

* testing two subsequent `fit` calls with different arguments, #1613. The current tests are incapable of passing two different sets of arguments to the same method.
* testing forecasters both with only `y` argument and with `y` and `X` argument #1674. The current tests did not detect side effects on `X` since only `y` were passed, and it is not possible to use two different sets of arguments.
* the transformer refactor which allows all transformers to accept both `Series` and `Panel` inputs, requiring us to test both types of inputs #1740. This would be very hard in the current framework.
* same for extending forecasters to global/hierarchical #1683 (or #1684), which will also see forecasters accepting multiple input scitypes, `Series` and `Panel` and `Hierarchical`.

After this PR, the class would be used to replace existing testing functionality, and write new tests easily.

This PR also contains tests for the `TestScenario` functionality.


fkiraly:
FYI @danbartl, @ltsaprounis, @satya-pattnaik - needed for testing in global forecasting, especially of the new base class. Would appreciate a review.

FYI @GuzalBulatova, @Lovkush-A, since you've reviewed the transformers refactor. This would help testing multiple input types. Would appreciate a look.

FYI @Carlosbogo, @ltsaprounis,  in relation to the "double fit calls".

FYI @aiwalter, @IlyasMoutawwakil since this relates to the prophet side effects.

FYI @mloning who wrote the original testing framework.

chrisholder (approved):

sktime/utils/_testing/scenarios.py
```diff
@@ -0,0 +1,200 @@
...
+        if args is not None:
+            self.args = _check_dict_of_dict(args)
+        else:
+            if not hasattr(self, args):
+                raise RuntimeError(
+                    "args must either be given to __init__ or set in a child class"
+                )
+            _check_dict_of_dict(self.args)
+
+    def run(self, obj, method_sequence=None, arg_sequence=None, return_all=False):
```
Lovkush-A, line 59: not clear to me why we allow overwriting of `method_sequence` and `arg_sequences`, but not `args`. what was the reason for this particular choice?

Note my instinct is that we do not need any over-writing and that we would create a new TestScenario object when we need to change something. but you must have had a reason to do things that way you did!

sktime/utils/_testing/scenarios.py
```diff
@@ -0,0 +1,200 @@
...
+        """Run a call(args) scenario on obj, and retrieve method outputs.
+
+        Runs a sequence of commands
+            res_1 = obj.method_1(**args_1)
+            res_2 = obj.method_2(**args_2)
+            etc, where method_i is method_sequence[i],
+                and args_i is arg_sequence[i]
```
Lovkush-A, line 66: typo, i think
```suggestion
                and args_i is self.args[arg_sequence[i]]
```

sktime/utils/_testing/scenarios.py
```diff
@@ -0,0 +1,200 @@
...
+    Parameters
+    ----------
+    args : dict of dict, default = None
+        dict of argument dicts to be used in methods
+        names for keys need not equal names of methods these are used in
+            but scripted method will look at key with same name as default
+        must be passed to construtor, or set in a child class
+    default_method_sequence : list of str, default = None
+        optional, if given, default method sequence to use in "run"
+        if not provided, "run" must alwayss be passed a method_sequence
```
Lovkush-A, line 30: 1. fixing minor typos.
2. in current way `run` is written, `method_sequence` does not *need* to be provided, as it can copy arg_sequence. either update docstring here or (my preference) simplify the logic a little in `run` by having copying only be one way. Explicitly, in `run`, if `arg_sequence` is not provided, then `arg_sequence` copies `method_sequence`, but not vice versa

```suggestion
    default_method_sequence : list of str, default = None
        optional, if given, default method sequence to use in `run`
        if not provided, `run` must always be passed a method_sequence
```

sktime/utils/_testing/scenarios.py
```diff
@@ -0,0 +1,200 @@
...
+        results: output of the last method call, if return_all = False
+            list of deepcopies of all outputs, if return_all = True
+        """
+        # if both None, fill with defaults if exist
+        if method_sequence is None and arg_sequence is None:
+            method_sequence = getattr(self, "default_method_sequence", None)
+            arg_sequence = getattr(self, "default_arg_sequence", None)
+
+        # if only one is None, fill one with the other
+        # this will also raise an error if both are still None
```
Lovkush-A, line 98: you have not implemented this error yet!
```suggestion
        # this will also raise an error if both are still None
```

sktime/utils/_testing/scenarios.py
```diff
@@ -0,0 +1,200 @@
...
+            else:
+                if isclass(obj):
+                    res = obj(**args)
+                else:
+                    res = type(obj)(**args)
+                obj = res
+            if return_all:
+                results += [deepcopy(res)]
+            else:
+                results = res
```
Lovkush-A, line 130: * adding some white space to separate the two if blocks

```suggestion
            args = self.args[arg_sequence[i]]

            if methodname != "__init__":
                res = getattr(obj, methodname)(**args)
            # if constructor is called, run directly and replace obj
            else:
                if isclass(obj):
                    res = obj(**args)
                else:
                    res = type(obj)(**args)
                obj = res

            if return_all:
                results += [deepcopy(res)]
            else:
                results = res
```

sktime/utils/_testing/scenarios.py
```diff
@@ -0,0 +1,200 @@
...
+
+    Parameters
+    ----------
+    obj : any object, check whether is list of str
+    name : str, default="obj", name of obj to display in error message
+
+    Returns
+    -------
+    obj, unaltered
+
```
Lovkush-A, line 167: why not return None?

sktime/utils/_testing/scenarios.py
```diff
@@ -0,0 +1,200 @@
...
+                    res = type(obj)(**args)
+                obj = res
+            if return_all:
+                results += [deepcopy(res)]
+            else:
+                results = res
+
+        return results
+
+    def is_applicable(self, obj):
```
Lovkush-A, line 134: when would this be called? it is not used in `run`

Lovkush-A (changes requested):
looks good! i like the unit-testing for this class.

* some typos fixes
* various small questions to help me understand

sktime/utils/_testing/scenarios.py
fkiraly: it will raise an error (because it executes line 100 which evaluates to `_check_list_of_str(None)`, but the message is not descriptive. Will update this.

sktime/utils/_testing/scenarios.py
fkiraly: because then I couldn't follow the `object = check_object(object)` pattern that is used throughout sktime

sktime/utils/_testing/scenarios.py
fkiraly: will be used in testing, to check whether the scenario should be skipped or not.

fkiraly: to see where it is going, have a look at the `test-refactor-with-scenarios` branch, file `scenarios_forecasting`

fkiraly: (maybe I should have written a design first here - perhaps I'll do that, but I first need to play around to make the two "sides" of the equation fit, i.e., the current design and the multi-scenario intention)

sktime/utils/_testing/scenarios.py
fkiraly: yes

sktime/utils/_testing/scenarios.py
fkiraly: if you would pass all three, would it not be easier, in terms of writing code to just run the methods explicitly?

The key (intended) benefit of this design is to allow appling the same sequence of input to the same sequence of methods across many estimators.

The ability to choose `method_sequence` is for cases where you may want very similar sequences with the same data, e.g., `fit-predict` vs `fit-predict-update` or `fit-predict-update-predict`.

sktime/utils/_testing/scenarios.py
```diff
@@ -0,0 +1,200 @@
...
+class TestScenario:
+    """Class to run pre-defined method execution scenarios for objects.
+
+    Parameters
+    ----------
+    args : dict of dict, default = None
+        dict of argument dicts to be used in methods
+        names for keys need not equal names of methods these are used in
+            but scripted method will look at key with same name as default
+        must be passed to construtor, or set in a child class
```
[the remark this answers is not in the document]
fkiraly: fixed

sktime/utils/_testing/scenarios.py
fkiraly: done, hope better?

sktime/utils/_testing/scenarios.py
Lovkush-A: The various checking functions in `datatypes` do not have this interface...? Is that an inconsistency or have I missed something?

sktime/utils/_testing/scenarios.py
Lovkush-A: resolving, moving general design discussion into main area, rather than in multiple conversation

Lovkush-A:
Yes, I think some kind of design document would be helpful, as I do not fully understand the current design.
* not clear in what contexts we would want to provide parameters to `run` versus creating new `TestScenario` class with different defaults.
* current design looks like we will end up creating *many* child classes, based on example here: https://github.com/alan-turing-institute/sktime/blob/test-refactor-with-scenarios/sktime/utils/_testing/scenarios_forecasting.py. This does not seem ideal, but maybe all designs will have analagous flaws.
* current design does not make clear how we would test different inputs but with same sequence of methods.
  * E.g. we want to test fit-predict with and without X. Does that mean we create two different TestScenario classes, or, a single class in which args has keys `fit_noX`, `fit_withX`, `predict_noX`, `predict_withX`
  * E.g. how to deal with testing various combinations of arguments, without having to create new class for each combination.

sktime/utils/_testing/scenarios.py
Lovkush-A: yup, looks good

sktime/utils/_testing/scenarios.py
fkiraly: hm, that's true. I was referring to things like `_check_X` or `_check_fh` or `_check_alpha`, which raise errors instead of returning yes/no. That would be consistent with the `check_raise` function.

The functions starting `check_is` have a different signature, returning boolean, message, metadata.

fkiraly:
> E.g. how to deal with testing various combinations of arguments, without having to create new class for each combination.

The plan was to create a new class for each combination - as a light-weight container to contain little more than the input arguments and the sequence in which they are called.

fkiraly:
> current design does not make clear how we would test different inputs but with same sequence of methods

Multiple options:
* just copy/paste the sequence every time you define a class, and change the inputs. That's not much of a copy-paste, one line.
* If it is a complex sequence and comes with tags, one could have an inheritance tree: base class -> class which specifies sequence -> classes which specify the inputs
* the "brute force" alternative (probably the worst option) is specify everything in `run` and let the test take control of method calls and parameters.

Lovkush-A:
> The alternative is that you may have to loop over parameter combinations, then you have to store somehow the possible parameter combinations that you want to loop over. Having the containers/scenarios seems to keep it simple.

Presumably you store the possible combinations in a list, no?

Also, no matter what design we choose, the possible combinations will have to be stored *somewhere*. Currently, we have to create a new class for each combination. I dislike this because:
* it is not DRY.
* it is not easy to quickly check that we have indeed covered all the combinations we want.
* if the number of combinations is "large", then having one class per combination is impractical.
* in my platonic ideal, if I wanted to test on another combination, i should just have to add a single item into a list somewhere, rather than creating a new class.

I have not thought about this as much as you have, but what is bad about something *like* this:
```python
for method_sequence in method_sequences:
  for args in args_for_each_method_sequence[method_sequence]:
    scenario = ForecastingScenario(method_sequence, args)
    for forecaster in forecasters_to_test:
      if scenario.is_applicable(forecaster)
        scenario.run(forecaster)
```

fkiraly:
>but what is bad about something like this

Nothing? My intention was very similar, sth like: 

```python
for method_sequence in method_sequences:
  for scenario in scenarios_for_method_sequence[method_sequence]:
    for forecaster in forecasters_to_test:
      if scenario.is_applicable(forecaster):
        scenario.run(forecaster)
```

Where the first line might not be an explicit loop, but "conversions" of different tests in the first instance, e.g., `test_fit_predict, etc`.

Note there is another potential loop where we test things *about* a given method sequence, e.g., return type of `predict`, or whether `predict` has side effects or not.

Lovkush-A:
> > but what is bad about something like this
> 
> Nothing? My intention was very similar, sth like:

If there is nothing bad about the suggestion I made, then I prefer it, as it avoids having to hard-code many different classes.

Anyway, to summarise:
* overall happy with this.
* my preference is to aim for a design in which we do not need to write lots of Scenario classes, but can instead create the Scenario objects in a loop. However, happy to progress with current design. We can always refactor again if deemed necessary (and by 'we', I of course mean 'you'. :P )
* I think it would be good to make explicit what kind of test coverage we are aiming for, e.g. cover every single path in the 'decision-flow graph', or cover enough paths so that each node in the decision-flow is reached, or ... . This would help in making design choices and determining if our testing is on track or not.
* i have small preference for the copying in `.run` to only be one way (arg sequence can be copied from method sequence, but not the other way around). Makes logic of code easier to follow, makes docstrings easier to write (see unresolved conversation above), follows the principle of 'there should be one obvious way to do things, for an end user of the code'.
* happy with changes in the latest two commits. I suggest including a unit-test with  'return_args=True' and check the output is what we expect.

@chrisholder. I saw you approved this PR earlier. What are your thoughts on the design, if any?

fkiraly:
(still thinking about design, won't merge yet)

fkiraly:
> my preference is to aim for a design in which we do not need to write lots of Scenario classes, but can instead create the Scenario objects in a loop.

How would the "in-the-loop" creation of scenario objects look like?

I estimate we'll write one or few scenario classes per estimator scitype, that corresponds roughly to independent pieces of logic.

fkiraly:
For design exploration sake, I've refactored the `test_all_estimator` tests with scenarios *plus* conditional fixtures #1839 (new pytest plugin) which gives you an easy, modular way to define something quite similar to the loops above (I hope):
https://github.com/alan-turing-institute/sktime/pull/1833
The main problem I have learnt about with a "naive loops" design is that this is not the full thing, since in the tests a lot of other very specific things happen. So it needs to happen via fixtures in my opinion.

Comments appreciated.

sktime/utils/_testing/scenarios.py
fkiraly: fixed, clarified requirement/defaulting behaviour

fkiraly:
Dealing with `inverse_transform`, I discovered we need dynamic creation of scenarios based on `obj` (the estimator) and the `arg`-key to avoid proliferation of scenario classes - since the input to `inverse_transform` depends on a tag of `obj`, otherwise all other things are equal.

I.e., there are two independently varying things in the scenario, so one either has to manually copy-paste to achieve the Cartesian product (four times four), or introduce functionality that determines one of the fields dynamically. This is done by the `get_args(key, obj)` function which scenarios can override, and which defaults to `return self.args[key]`

lmmentel:
Is this complementary to #1839 or an alternative approach?

fkiraly:
> Is this complementary to #1839 or an alternative approach?

Complementary/orthogonal.

Both things come together in https://github.com/alan-turing-institute/sktime/pull/1833, which uses both conditional fixtures and scenarios.

fkiraly:
Hm, interesting side effects - seems to have brought down testimng time on azure consistently to 140min from "runs too long".

chrisholder (approved):

fkiraly:
Approval, and @Lovkush-A's comments are addressed.

This also does not change the tests, just adds utilities, so I'm merging this to simplify the dev/review process around the testing refactor.

PR #1822, opened 2022-01-02
In interval_based_classification example notebook, use multivariate dataset for the multivariate examples

ksachdeva:
Correct typo in the internval_based_classification example notebook. Even for multivariate example, the univariate dataset was being passed

No

##### For all contributions

##### For new estimators


ksachdeva:
> Can you kindly reset your notebook and run it through once?

@fkiraly Reason for this unordered execution of cells is because there are multiple warnings because of some recent deprecations in `sktime`. To squish those warnings I ran some cells twice. 

Examples:

image

image


If you are okay with having those warnings in the output, I can run the notebook again.

The alternative is that I fix only the typos without running the notebook.

Regards
Kapil

fkiraly:
hmm - these are deprecation warnings from `sktime` itself - why don't you just address these while at it?
If it's not too much of a hassle, of course.

For instance, above, `RandomIntervalSpectralForest` should be replaced by `RandomIntervalSpectralEnsemble`.

That should fix the issues with the multi-execution?

fkiraly:
Another thing, you should add yourself to the contributors file.

MatthewMiddlehurst (approved):

Issue #1776, opened 2021-12-18
[ENH] `_set_fh` with side effects should be replaced by `_check_fh` with return


fkiraly:
[let's first have an API discussion before implementing/changing anything]

Currently, our input check interfaces are not consistent in forecasting.

We have `X, y = _check_X_y(X, y)`, `X = _check_X(X)`, `alpha = check_alpha(alpha)`, but for the forecasting horizon we have `_set_fh(fh)` (this returns nothing) and then the user needs to retrieve a potentially changed `fh` from `self.fh` to which `_set_fh` writes.

This is inconsistent and counterintuitive in my opinion, I think the bug addressed in #1775 shows this - @SveaMeyer13 and @kejsitake didn't know and also didn't guess that the above was how `_set_fh` had to be used.
I don't think this is at all their fault, the problem is the inconsistency in the input check and conversion, where `_set_fh` breaks what is otherwise an obvious pattern.

So I think this should be changed to `fh = _check_fh(fh)`, notationally, without changing any logic.
More explicitly, the line `fh = _check_fh(fh)` should replace the line `_set_fh(fh)` and then repeated subsequent uses of `self.fh` in places where `fh` seems more natural.
(it would also not affect users, only developers, since the logic doesn't change)

Any thoughts? @aiwalter, @mloning, @SveaMeyer13, @kejsitake, @eenticott-shell?

kejsitake:
I agree that this is counterintuitive and _check_fh makes more sense.

PR #1823, opened 2022-01-03
[ENH] Change `_set_fh` to a `_check_fh` that returns `self._fh`

fkiraly:
This implements the change suggested in issue #1776:

Main change is changing the `_set_fh(fh)` method to `_check_fh(fh)` which returns `self.fh` (the coerced/validated `ForecastingHorizon`), instead of no return.

Further additions concern docstrings which clarify the functionality of `_check_fh` and the `fh` attribute.


fkiraly:
FYI @kejsitake.

I added you as a potential reviewer, @SveaMeyer13, since this is arising from changes you originally added (see #1776).

chrisholder (approved):
Code looks fine someone who knows what this actually does and impact should also review :joy:

fkiraly:
@aiwalter, any opinions?

fkiraly:
In #1776, @kejsitake said

> I agree that this is counterintuitive and _check_fh makes more sense.

so I'll count this as "someone who knows what this actually does and impact" for the purpose of your statement, @chrisholder.
Though I'll give it a few more days so core devs can have a look if they want.

fkiraly:
@chrisholder, I had to update this since `predict_residuals` was added since the PR opened, which had another reference to `_set_fh`

kejsitake (approved):
This looks good to me, for the following two reasons:
- I tested the existing code and it looks like `set_fh(fh)` modifies `self._fh` without returning the Forecasting Horizon, making it confusing for the user to keep track of which fh to use next(fh passed to the function or self.fh).
- I also think that it is more consistent to have a returning value, given that `check_alpha` has the same behaviour.

Regarding code, I agree with @chrisholder that it looks okay.

Issue #1405, opened 2021-09-10
[BUG] pandas frequency W-WED in `ForecastingHorizon` causes failure in `AutoETS.predict`


asksteph:
**Describe the bug**
 I have a weekly dataset with W-WED frequency that I want to train an AutoETS model on. The predict step throws an error when it does not understand this datetime frequency.

**To Reproduce**

```python
import numpy as np
import pandas as pd

# play example: 2 years of weekly data
train = pd.Series(np.random.uniform(low=2000, high=7000, size=(104,)), index=pd.date_range("2019-01-02", freq="W-WED", periods=104))

# setup AutoETS model
from sktime.forecasting.ets import AutoETS
forecaster = AutoETS(auto=True, sp=52, n_jobs=-1, restrict=True)
forecaster.fit(train)
pred = forecaster.predict(np.arange(1, 27)) # <---- this step throws the error
```

**Expected behavior**
 I expect the model to be trained on the weekly data and to return predictions for the requested forecast horizon.

**Additional context**
 If I set my datetime index to None, it will complain there is no freq supplied. But it does not understand the weekly index, so I am stuck.

**Versions**

System:
    python: 3.7.11 (default, Jul 27 2021, 09:42:29) [MSC v.1916 64 bit (AMD64)]
executable: C:\Users\szipk\anaconda3\envs\py37\python.exe
   machine: Windows-10-10.0.19041-SP0

Python dependencies:
          pip: 21.0.1
   setuptools: 49.6.0.post20200925
      sklearn: 0.24.2
       sktime: 0.7.0
  statsmodels: 0.12.2
        numpy: 1.19.2
        scipy: 1.6.2
       Cython: 0.29.24
       pandas: 1.3.2
   matplotlib: 3.4.2
       joblib: 1.0.1
        numba: 0.53.1
     pmdarima: 1.8.2
      tsfresh: 0.18.0

aiwalter:
Hi @asksteph , I think you cant use `np` inside `predict` as `fh`. Rather use a `list` or `ForecastingHorizon`. Otherwise pls provide error.

asksteph:
I tried several fh conventions from the example in https://github.com/alan-turing-institute/sktime/blob/main/examples/01_forecasting.ipynb. Neither the numpy, list, nor ForecastingHorizon work. I don't think it's an issue with the forecast horizon. 

```
import numpy as np
import pandas as pd

# 2 years of weekly data
[...]

PR #1830, opened 2022-01-05
1523 fixing forecast horizon bug

### LINKED ISSUE #1523: [BUG] `ForecastingHorizon.to_absolute` conversion for `W-XXX` frequencies
`ForecastingHorizon.to_absolute` can produce the wrong day of the week in `to_absolute` conversions.

This seems to occur only for the `W-XXX` frequency, leading to the error in #1405.

Minimally reproducing code:

```python
train = pd.Series(1, index=pd.date_range("2021-10-06", freq="W-WED", periods=3))
fh.to_absolute(train.index[-1])
```

results in `ForecastingHorizon(['2021-10-31', '2021-11-07', '2021-11-14'], dtype='datetime64[ns]', freq='W-SUN', is_relative=False)`.

Expected result is `ForecastingHorizon(['2021-10-27', '2021-11-03', '2021-11-10'], dtype='datetime64[ns]', freq='W-WED', is_relative=False)`.

Further unintuitive behaviour happens if the `fh` has length 2 or 1, in that case the resulting frequency is `None` but should be `'W-WED'` too... I think?

We should:

- [x] fix this
- [x] introduce tests for this specific frequency

eenticott-shell:
#### Reference Issues/PRs
Fixes #1523 

Removes use of `_get_freq` in the `fh.to_absolute()` conversion. `_get_freq` removes anchorings, this was causing issues when users needed to convert a frequency involving an anchoring as the information was lost and the horizon defaults back to "W-SUN" even if another day e.g. "W-WED" was specified.

no

Whether this change impacts other parts of the code. Presumably this functionality of removing anchorings is there for a reason, it just seems unnecessary for the `to_absolute` functionality. 

##### For all contributions


fkiraly-shell:
Hmmm, on first glance the fix seems simple enough.

I've also thought about this what ripple effects this might have, and I wasn't able to come up with anything.
The tests also all seem to be happy.

Just to be sure, would you kindly be able to include tests (add new or include in existing where close enough):
* a test where we convert absolute to relative, then back to absolute - should be the same thing. And, another where we convert relative to absolute to relative. Should again be the same.
* a test which is close to @asksteph's problem in #1405 where an estimator is being fit. Just to ensure this works properly in estimators, too.

All, of course, using frequencies with anchors.

fkiraly (approved):
Yes, looks like this is indeed a fix, the tests also seem to indicate it´s a "safe" change. Thanks!

PR #1831, opened 2022-01-05
DRAFT - [ENH] Naive C removal

fkiraly:
Draft PR which attempts a naive removal of C sources, to test a GitHub actions CI/CD without C sources. FYI @lmmentel.

FYI @TonyBagnall, @chrisholder, @MatthewMiddlehurst, since you're also working on this.
What I did is very hacky, kill MrSEQL and replace distances imports by those that sound like they should work.

Where feature parity wasn't there with `twe_distance`, I've imported another distance under its name as a placeholder.

Note: MrSEQL is scattered throughout the notebooks, I did not touch those (except removing the one called MrSEQL).


fkiraly:
@lmmentel, this is failing on linting and input format expectation for distances, but it no longer uses C, and the issues are localized to individual classifiers.

Using xfail on a small number failures, this should allow to test the GitHub actions workflow on windows.

The assumption would be that @tonybagnall & TSC team will change the failing classes until before the 0.10.0 release.

TonyBagnall:
do you want to close this PR? The work is ongoing and wont be on this branch.

fkiraly:
I know, @TonyBagnall - I do not intend the "proper" work to be here. This is just a hacky patch to allow conditional development, with the intent to switch this to the "real" thing once done.

lmmentel:
The original issue we discussed in solved so I don't believe we need this anymore. If the work to remove C is ongoing somewhere else maybe we can clean this up.

PR #1838, opened 2022-01-05
[ENH] estimator scitype utility

fkiraly:
This PR adds a single utility `scitype` which produces the unique scitype string in `BASE_CLASS_REGISTER` and `BASE_CLASS_SCITYPE_LIST`, given a class or object inheriting from `BaseObject`.

This is useful in refacturing occurrences of this query which have been dealt with manually, for instance when combining a lookup call `all_estimators` or `all_tags` starting at the class, or for a near-future refactoring of `_make_args`.

This PR also introduces a test which links `scitype` to its (quasi) inverse, an existing utility function called `_check_estimator_type`.

### DIFF
--- a/sktime/registry/__init__.py
+++ b/sktime/registry/__init__.py
@@ -1,26 +1,23 @@
 # -*- coding: utf-8 -*-
 """Implements registry for sktime estimator base classes and tags."""
 
-from sktime.registry._tags import (
-    ESTIMATOR_TAG_REGISTER,
-    ESTIMATOR_TAG_LIST,
-    check_tag_is_valid,
-)
-
 from sktime.registry._base_classes import (
-    BASE_CLASS_REGISTER,
     BASE_CLASS_LIST,
     BASE_CLASS_LOOKUP,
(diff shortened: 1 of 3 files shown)


lmmentel:
I tried reviewing the code but I'm not sure I understand the changes and the motivation behind. Maybe someone with more context can have a look.

fkiraly:
> I tried reviewing the code but I'm not sure I understand the changes and the motivation behind. Maybe someone with more context can have a look.

Each estimator has an "estimator type" string associated to it in various parts of the testing machinery. E.g., all children of `BaseForecaster` are linked to the string `"forecaster"` Having this decouples base class imports from the location of the dispatch on the base class, otherwise you´d always have to import all base classes if you want to check what kind of estimator you are looking at, and incrementally add new base classes in 100 locations as they appear.

This can be avoided by having one location where base classes are registered to the strings.
By putting the lookup mechanism in a function, the lookup gets even simpler.

chrisholder (approved):

PR #1839, opened 2022-01-06
[ENH] pytest conditional fixtures

fkiraly:
This PR contains a plugin for `pytest` that allows the user to easily specify conditional/nested fixtures over combinations of variables.

The utility is in `utils._testing._conditional_fixtures`, the function `conditional_fixtures_and_names`. It takes conditional fixture generation functions and allows arbitrary nesting of conditionals (as long as they are non-circular).

As a proof-of-concept of how this simplifies things, I've used it to refactor `test_all_estimators`.

This will allow us to specify conditional fixtures in the future, and `pytest_generate_tests` will always be the same. This will also unlock easy inheritance of tests when wrapped into classes.

Closely related to discussion in https://github.com/alan-turing-institute/sktime/pull/1819, especially the part about the loops at the end. The conditional fixtures utility allows to easily specify the loops, by moving modular conditional functions around.


fkiraly:
FYI @Lovkush-A, related to our design discussion.

I think *this* the "right" form of the loops.

sktime/utils/_testing/_conditional_fixtures.py
```diff
@@ -0,0 +1,181 @@
...
+    Returns
+    -------
+    obj, unaltered
+
+    Raises
+    ------
+    TypeError if obj is not list of str
+    """
+    if not isinstance(obj, list) or not np.all(isinstance(x, str) for x in obj):
+        raise TypeError(f"{obj} must be a list of str")
```
Lovkush-A, line 163: have not yet reviewed the pr, but happened to spot this typo, so might as well add the suggestion.
```suggestion
        raise TypeError(f"{name} must be a list of str")
```

lmmentel:
@fkiraly would you able able to give a more accessible explanation of what this PR is trying to achieve, fix, enhance?

fkiraly:
Sure. From the docstring, an example, which may help:
```
    Example: we want to loop over two fixture variables, "number" and "multiples"
        "number" are integers from 1 to 10,
        "multiples" are multiples of "number" up to "number"-squared
        we then write a generator_dict with two entries
        generator_dict["number"] is a function (test_name, **kwargs) -> list
            that returns [1, 2, ..., 10]
        generator_dict["multiples"] is a function (test_name, number, **kwargs) -> list
            that returns [number, 2* number, ..., number*number]
```

In `pytest`, how would you do that? You would do a `mark.parameterize`, and manually generate the combination of fixtures.
If you have more than two fixture variables, this gets tedious. So we decompose this into conditional fixtuers, and write only the logic that generates fixtures for variable i given fixtures for variables i-1, i-2, ..., 1. The rest is done by the framework (see docstring for the syntax).

chrisholder (approved):

fkiraly:
So, this has an approval, no change requests or rejection after being there for a while.

This PR also does not change any of the tests, just adds a utility, so will cause no fundamental change.
To simplify the develop, review, merge process around the testing, I'm merging this now.

fkiraly:
> @fkiraly would you able able to give a more accessible explanation of what this PR is trying to achieve, fix, enhance?

now in #1922 which describes the "target state", hope that is accessible...

sktime/tests/test_all_estimators.py
```diff
@@ -54,6 +55,61 @@
...
 
+generator_dict = dict()
+
+
+def is_excluded(test_name, est):
+    """Shorthand to check whether test test_name is excluded for estimator est."""
+    return test_name in EXCLUDED_TESTS.get(est.__name__, [])
+
+
+def _generate_estimator_class(test_name, **kwargs):
```
lmmentel, line 66: It seems what this function does is filtering the estimators from ALL_ESTIMATORS, but the name suggests something else. How about renaming it to be more accurate? maybe `filter_estimator_classes`

sktime/tests/test_all_estimators.py
```diff
@@ -54,6 +55,61 @@
...
+    """Return estimator instance fixtures.
+
+    Fixtures parameterized
+    ----------------------
+    estimator_instance: instance of estimator inheriting from BaseObject
+        ranges over all estimator classes not excluded by EXCLUDED_TESTS
+        instances are generated by create_test_instance class method
+    """
+    # we need to exclude instances from classes on the exclude list
+    estimator_classes_to_test = [
```
lmmentel, line 95: this seems duplicated from `_generate_estimator_class` wouldn't it be more convenient o call `_generate_estimator_class` instead? I would also allow to not have `is_excluded` as a separate function since it would be absorbed by `_generate_estimator_class` as a statement.

sktime/utils/_testing/_conditional_fixtures.py
```diff
@@ -0,0 +1,217 @@
...
+class FixtureGenerationError(Exception):
+    """Raised when a fixture fails to generate."""
+
+    def __init__(self, fixture_name=""):
+        self.fixture_name = fixture_name
+        fixture_name = fixture_name + " "
+        super().__init__(f"fixture {fixture_name}failed to generate")
+
+
+def conditional_fixtures_and_names(
```
lmmentel, line 25:
```suggestion
def create_conditional_fixtures(
```

sktime/utils/_testing/_conditional_fixtures.py
```diff
@@ -0,0 +1,217 @@
...
+    This function automatically creates the inputs for pytest.mark.parameterize
+        fixture_param_str = "number,multiples"
+        fixture_prod = [(1, 1), (2, 2), (2, 4), (3, 3), (3, 6), ...]
+        fixture_names = ["1-1", "2-2", "2-4", "3-3", "3-6", ...]
+
+    Parameters
+    ----------
+    test_name : str, name of the test, from pytest_generate_tests
+    fixture_vars : list of str
+        fixture variables to loop over
```
lmmentel, line 51:
```suggestion
        fixture variables
```

sktime/utils/_testing/_conditional_fixtures.py
```diff
@@ -0,0 +1,217 @@
...
+    """Raised when a fixture fails to generate."""
+
+    def __init__(self, fixture_name=""):
+        self.fixture_name = fixture_name
+        fixture_name = fixture_name + " "
+        super().__init__(f"fixture {fixture_name}failed to generate")
+
+
+def conditional_fixtures_and_names(
+    test_name, fixture_vars, generator_dict, fixture_sequence=None
```
lmmentel, line 26: consider adding type annotations
```suggestion
    test_name: str, fixture_vars: List[str], generator_dict: Dict[str: Callable], fixture_sequence: Union[None,List[str]] = None
```
will also need to import 

```
from typing import List, Dict, Callable
```

lmmentel (approved):
I don't really get what's happening in `get_fixtures` or how to test it, I trust that it's doing something useful.

fkiraly:
> I don't really get what's happening in get_fixtures or how to test it, I trust that it's doing something useful.

it's just a wrapper around an invocation of `generator_dict` values (which are functions generating fixture lists and optionally names), and adds names if these are not returned by that function.

I'll add clarification to the docstring.

sktime/tests/test_all_estimators.py
fkiraly: I disagree, since this is a "generic" following a template pattern, where functions are called `_generate_[fixturename]`.
The function happens to filter, yes, but it is concrete behaviour that fills an abstract hole whose name is "generate the fixtures of name `estimator_class`".

sktime/tests/test_all_estimators.py
fkiraly: similar - we need to fill the abstract hole for generating the `estimator_instance` by providing a function `_generate_estimator_instance`. I'd like to have `is_excluded` separate, too, since it wraps the constant we may want to move, or replace by alternative logic.

sktime/utils/_testing/_conditional_fixtures.py
fkiraly: agreed, this is a better name. Replaced throughout.

sktime/utils/_testing/_conditional_fixtures.py
fkiraly: yes, removed and added clarification

sktime/utils/_testing/_conditional_fixtures.py
fkiraly: sure - btw, you don't need to do `Union[None`, or similar to deal with an optional `None` default.

sktime/tests/test_all_estimators.py
fkiraly: apologies, I misunderstood this statement - I thought you suggested to *replace* `_generate_estimator_instance`.
Calling `_generate_estimator_class` is indeed better here.

sktime/tests/test_all_estimators.py
fkiraly: the name describes the *purpose* of the function (that it generates fixtures) rather than its *logic* (i.e., how the fixture list is obtained). In this place, the purpose is more natural to use as a description.

fkiraly: I now added detailed comments above the functions, which should further clarify things and explain the names better

fkiraly:
approvals from @chrisholder, @lmmentel up to non-blocking comments which have been addressed (I hope) - merging

PR #1840, opened 2022-01-06
polymorphic data loader in contrib

TonyBagnall:
moved the polymorphic data loader to contrib due to reviewer objections to its presence in main

### DIFF
--- a/sktime/contrib/data_io.py
+++ b/sktime/contrib/data_io.py
@@ -122,8 +122,8 @@ def load_from_tsfile(
     # equal_length = True
     with open(full_file_path_and_name, "r", encoding="utf-8") as file:
         _meta_data = _read_header(file, full_file_path_and_name)
-        for line in file:  # Will this work?
-            num_cases = num_cases + 1
+        for line in file:
+            num_cases += 1
             line = line.replace("?", replace_missing_vals_with)
             dimensions = line.split(":")
             # If first instance then note the number of dimensions.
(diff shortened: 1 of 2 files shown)


sktime/contrib/data_io.py
```diff
@@ -0,0 +1,185 @@
...
+            num_cases = num_cases + 1
+            line = line.replace("?", replace_missing_vals_with)
+            dimensions = line.split(":")
+            # If first instance then note the number of dimensions.
+            # This must be the same for all cases.
+            if is_first_case:
+                num_dimensions = len(dimensions)
+                if _meta_data["has_class_labels"]:
+                    num_dimensions -= 1
+                for _dim in range(0, num_dimensions):
```
chrisholder, line 135: Remove for loop here just do:

instance_list = [ [] for _ in range(num_dimensions) ]

sktime/contrib/data_io.py
```diff
@@ -0,0 +1,185 @@
...
+    is_first_case = True
+    instance_list = []
+    class_val_list = []
+    line_num = 0
+    num_dimensions = 0
+    num_cases = 0
+    # equal_length = True
+    with open(full_file_path_and_name, "r", encoding="utf-8") as file:
+        _meta_data = _read_header(file, full_file_path_and_name)
+        for line in file:  # Will this work?
```
chrisholder, line 125: Remove comment

sktime/contrib/data_io.py
```diff
@@ -0,0 +1,185 @@
...
+            num_cases = num_cases + 1
```
chrisholder, line 126: num_cases += 1

chrisholder (changes requested):

sktime/contrib/data_io.py
TonyBagnall: does it make any difference? this is in contrib, its the wild west :) Still, have changed it

PR #1844, opened 2022-01-06
[ENH] Extended deep_equals, with precise indication of why equality fails

fkiraly:
This PR extends the `deep_equals` utility used in testing with an optional second return argument, an error message that states at which location precisely two potentially highly nested objects are not equal.

This is useful since printouts in tests may involve `repr`-s of nested estimators or nested dicts, lists, tuples, which are really hard to parse. Already two dicts with 6 elements, where elements may be in different display order, can easily cause headache in reading the diagnostic output from a test failure.

This PR tries to alleviate that tedious, repeated staring at printouts of dicts, arrays, data frames and whatnot.

### DIFF
--- a/sktime/utils/_testing/deep_equals.py
+++ b/sktime/utils/_testing/deep_equals.py
@@ -51,12 +51,14 @@ def deep_equals(x, y, return_msg=False):
                     if class/object, names of attributes and methods are not equal
             .dtype - dtype of pandas or numpy object is not equal
             .index - index of pandas object is not equal
-            .series_equals, .df_equals - .equals call on pandas objects returns False
+            .series_equals, .df_equals, .index_equals - .equals of pd returns False
             [i] - if tuple/list: i-th element not equal
             [key] - if dict: value at key is not equal
             [colname] - if pandas.DataFrame: column with name colname is not equal
             != - call to generic != returns False
     """
+
+    print(type(x))
 
     def ret(is_equal, msg):
         if return_msg:
@@ -103,6 +105,8 @@ def deep_equals(x, y, return_msg=False):
             return ret(True, "")
         else:
             return ret(x.equals(y), f".df_equals, x = {x} != y = {y}")
+    elif isinstance(x, pd.Index):
+        return ret(x.equals(y), f".index_equals, x = {x} != y = {y}")
     elif isinstance(x, np.ndarray):
         if x.dtype != y.dtype:
             return ret(False, f".dtype, x.dtype = {x.dtype} != y.dtype = {y.dtype}")
--- a/sktime/utils/_testing/deep_equals.py
+++ b/sktime/utils/_testing/deep_equals.py
@@ -97,7 +97,7 @@ def deep_equals(x, y, return_msg=False):
         # if columns are equal and at least one is object, recurse over Series
         if sum(x.dtypes == "object") > 0:
             for c in x.columns:
-                is_equal, msg = deep_equals(x[c], y[c])
+                is_equal, msg = deep_equals(x[c], y[c], return_msg=True)
                 if not is_equal:
                     return ret(False, f'["{c}"]' + msg)
             return ret(True, "")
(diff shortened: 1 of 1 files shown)


sktime/utils/_testing/deep_equals.py
```diff
@@ -87,30 +130,47 @@ def _tuple_equals(x, y):
...
+        is_equal, msg = deep_equals(xi, yi, return_msg=True)
+        if not is_equal:
+            return ret(False, f"[{i}]" + msg)
 
-    return True
+    return ret(True, "")
 
 
-def _dict_equals(x, y):
+def _dict_equals(x, y, return_msg=False):
```
lmmentel, line 173: Have you tried https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertDictEqual ?

sktime/utils/_testing/deep_equals.py
```diff
@@ -30,52 +32,93 @@ def deep_equals(x, y):
...
     elif x != y:
-        return False
+        return ret(False, f" !=, {x} != {y}")
 
-    return True
+    return ret(True, "")
 
 
-def _tuple_equals(x, y):
+def _tuple_equals(x, y, return_msg=False):
```
lmmentel, line 121: Have you tried https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertTupleEqual instead?

sktime/utils/_testing/deep_equals.py
```diff
@@ -30,52 +32,93 @@ def deep_equals(x, y):
...
+            .dtype - dtype of pandas or numpy object is not equal
+            .index - index of pandas object is not equal
+            .series_equals, .df_equals - .equals call on pandas objects returns False
+            [i] - if tuple/list: i-th element not equal
+            [key] - if dict: value at key is not equal
+            [colname] - if pandas.DataFrame: column with name colname is not equal
+            != - call to generic != returns False
     """
+
+    def ret(is_equal, msg):
```
lmmentel, line 61: This is a little confusing since it takes two args and an additional one from the closure, any thoughts on refactoring this to be a bit easier to read without jumping back and forth?

sktime/utils/_testing/deep_equals.py
```diff
@@ -87,30 +130,47 @@ def _tuple_equals(x, y):
...
+    is_equal: bool - True if x and y are equal in value
         x and y do not need to be equal in reference
+    msg : str, only returned if return_msg = True
+        indication of what is the reason for not being equal
+            concatenation of the following elements:
+            .len - length is not equal
+            [i] - i-th element not equal
     """
+
+    def ret(is_equal, msg):
```
lmmentel, line 147: This function is already defined in the previous scope.

sktime/utils/_testing/deep_equals.py
fkiraly: that´s not nested, I think? Only looks at the first level. I want to look into data frames and into objects in obj columns, too, since nested data frames occur in the nested_univ mtype representationl

fkiraly: also applies only to dict

fkiraly: to be clearer, it wouldn´t deal with the recursion, since we need to call `deep_equals` from `_dict_equals`

sktime/utils/_testing/deep_equals.py
fkiraly: same, wouldn´t deal with the recursion, since we need to call `deep_equals` from `_dict_equals`

sktime/utils/_testing/deep_equals.py
fkiraly: no good ideas. We could have a third argument, and move it out of the scope, but that would make the return statements hard to read. Any suggestions?

chrisholder (approved):

sktime/utils/_testing/deep_equals.py
fkiraly: yes, not too DRY - although I think it´s easier to read if it´s close to the usage.

PR #1846, opened 2022-01-07
[BUG] bugfixes for various bugs discovered in scenario testing

fkiraly:
Testing using scenarios https://github.com/alan-turing-institute/sktime/pull/1833 has uncovered a number of bugs that went undetected due to limited input case coverage prior. The smaller ones are fixed here.

This does *not* assume the scenario testing PR which is still a draft.

Bugs fixed:
* bug in `StackingForecaster`: passing `X` raised `NotImplementedError`, while instead the argument `X` should have been ignored. In fact, `X` could have been used in the forecaster, which is what happens now.
* bug in `StackingForecaster`, `X` was passed to the `fh` argument of `_predict_forecasters`; `fh` was using `self.fh` and not input to `_fit` and `_predict` (which comes from `self.fh`, but it should be passed/referenced)
* the pipelines inheriting from `_Pipeline` were using a test case pipeline that would break for multivariate input, since the component `BoxCoxTransformer` was univariate. This has been replaced by `ExponentTransformer`
* test example for `FunctionTransformer` was breaking if negative values were passed, since log was the `transform` and exp the `inverse_transform`, which would negative values to nan. The other way round there is no problem, so switched it.
* in `_HeterogeneousEnsembleForecaster`, `return_pred_int` based raises were removed, because all this does is iterate over `fit`s or `predict`s in the ensemble (which can pass the args on 1:1)
* `AlignerDTWfromDist` did not clone parameter `dist_trafo` before fitting, as is required by sklearn interface (`__init__` parameters should be cloned if estimators). Added a `clone`.
* `ThetaLinesTransformer._transform` would break for `X` being `pd.DataFrame`, despite a docstring claim that one-column `pd.DataFrame` were supported. Fixed.

Changes which are not bugfixes:
* changed check upon `"X-y-must-have-same-index"=True` in `BaseForecaster` to fail only when `X.index` does not contain `y.index`, instead requiring those to be equal.

Todo (not fixed):
* pipelines need to infer correctly whether they support univariate data or not. For instance, the `TransformedTargetForecaster` will not support multivariate forecasts if one of the transformers is univariate.

### DIFF
--- a/sktime/alignment/dtw_python.py
+++ b/sktime/alignment/dtw_python.py
@@ -9,6 +9,8 @@ __author__ = ["fkiraly"]
 import numpy as np
 import pandas as pd
 
+from sklearn import clone
+
 from sktime.alignment.base import BaseAligner
 from sktime.utils.validation._dependencies import _check_soft_dependencies
 
@@ -234,6 +236,7 @@ class AlignerDTWfromDist(BaseAligner):
         super(AlignerDTWfromDist, self).__init__()
 
         self.dist_trafo = dist_trafo
+        self.dist_trafo_ = clone(self.dist_trafo)
         self.step_pattern = step_pattern
         self.window_type = window_type
         self.open_begin = open_begin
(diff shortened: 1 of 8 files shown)


fkiraly:
FYI @GuzalBulatova re `ThetaLinesTransformer`

chrisholder (approved):

Issue #1845, opened 2022-01-07
[ENH] Add option to set smoothing parameters for ExponentialSmoothing class based models


ishannangia001:
`smoothing_level`, `smoothing_trend`, and `smoothing_seasonal` should be added as parameters to `ExponentialSmoothing` which wraps `statsmodels` exponential smoothing.

In `statsmodels`, these parameters are passed to `fit`, but in `sktime`, they should be passed to `__init__` together with the other hyper-parameters.

Adding other parameters which are arguments to `fit` in `statsmodels` to `__init__` in `sktime` should also be considered.

---

**PROBLEM** 
I have been using the ExponentialSmoothing class recently. Can't seem to find a way to input/set the smoothing_level, smoothing_trend and the smoothing_seasonal parameters. This functionality is provided by statsmodels and imo should be offered by sktime too.  

**Solution**
Add a way to input smoothing_level, smoothing_trend, and smoothing_seasonal parameters for the ExponentialSmoothing models.

fkiraly:
Hm, could you kindly paste a reference to the parameters you mentioned?

I just looked in statsmodels, and here is the list of parameters I could find.
The list does not include `smoothing_level`, `smoothing_trend`, or `smoothing_seasonal`.

```
        endog,
        trend=None,
        damped_trend=False,
        seasonal=None,
        seasonal_periods=None,
        initialization_method=None,  # Future: 'estimated',
        initial_level=None,
        initial_trend=None,
        initial_seasonal=None,
        use_boxcox=None,
        bounds=None,
        dates=None,
        freq=None,
        missing="none",
```

ishannangia001:
They have been mentioned in their tutorial notebook: https://github.com/statsmodels/statsmodels/blob/main/examples/python/exponential_smoothing.py

These three parameters are called while fitting the model and not while initialising it. The below image shows a snippet from the code in the above link.

![image](https://user-images.githubusercontent.com/96042791/148554304-640caa45-3e79-4378-a978-1759f5b6d098.png)

fkiraly:
Ohhhh, it's passed to `fit` and not to the constructor! Thanks!

fkiraly:
We overlooked these, since they were being passed to `fit`. Bit of a questionable design decision on the side of `statsmodels` to pass some parameters in the constructor, and some in `fit`...

Aaaanyway, I've marked this a good first issue, since it should be only a few lines of code to add/change.

ishannangia001:
[...]

PR #1849, opened 2022-01-07
[ENH] add missing `fit` parameters to `statsmodels` Holt-Winters exponential smoothing interface

fkiraly:
This PR adds parameters of `statsmodels.tsa.holtwinters` in `fit` as loopthrough to `sktime`'s `ExponentialSmoothing` interface.

So far, only the constructor parameters were accessible, not the parameters in `fit`.

Solves #1845.

### DIFF
--- a/sktime/forecasting/exp_smoothing.py
+++ b/sktime/forecasting/exp_smoothing.py
@@ -146,7 +146,7 @@ class ExponentialSmoothing(_StatsModelsAdapter):
         self.initial_level = initial_level
         self.initial_trend = initial_trend
         self.initial_seasonal = initial_seasonal
-        self.initialization_method = initialization_method
+        self.initialization_method = initialization_method,
         self.smoothing_level = smoothing_level,
         self.smoothing_trend = smoothing_trend,
         self.smoothing_seasonal = smoothing_seasonal,
--- a/sktime/forecasting/exp_smoothing.py
+++ b/sktime/forecasting/exp_smoothing.py
@@ -14,6 +14,8 @@ from sktime.forecasting.base.adapters import _StatsModelsAdapter
 class ExponentialSmoothing(_StatsModelsAdapter):
     """Holt-Winters exponential smoothing forecaster.
 
+    Direct interface for `statsmodels.tsa.holtwinters`.
+
     Default settings use simple exponential smoothing without trend and
     seasonality components.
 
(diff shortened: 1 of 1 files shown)


sktime/forecasting/exp_smoothing.py
```diff
@@ -97,6 +147,16 @@ def __init__(
...
+        self.smoothing_seasonal = smoothing_seasonal
+        self.damping_trend = damping_trend
+        self.optimized = optimized
+        self.remove_bias = remove_bias
+        self.start_params = start_params
+        self.method = method
+        self.minimize_kwargs = minimize_kwargs
+        self.use_brute = use_brute
 
         super(ExponentialSmoothing, self).__init__()
```
lmmentel, line 161: is this call necessary when no argument are being passed?

fkiraly: they are being passed later, in `fit`? Or what do you mean?

lmmentel: Sorry, I'm only referring to the call to `super`

fkiraly: > is this call necessary when no argument are being passed?

ah, I see - yes, since it handles setting of the `_is_fitted` flag and, potentially, related concerns.

lmmentel: I can only see `_StatsModelAdapter` setting `self._forecaster = None` and  `self._fitted_forecaster = None`, shouldn't the `_is_fitted` be set in `fit`?

lmmentel, line 161:
```suggestion
        super().__init__()
```

fkiraly: all fittable estimators should set `_is_fitted` to `False` in `__init__`.
In this case, it should happen two levels higher (after two `super` calls), in `BaseEstimator`.

lmmentel: I think there is something to be said about the growing inheritance depth that does not make it straightforward to either review or write code extending functionality. That is probably a broader design discussion, but I want to make a comment here since it a good illustration of that point.

lmmentel (approved):
:+1:

atsangarides:
when pip installing sktime 0.9.0 I still don't see the new params in ExponentialSmoothing
`Downloading sktime-0.9.0-cp38-cp38-macosx_10_15_x86_64.whl 

```
class ExponentialSmoothing(_StatsModelsAdapter):
    """Holt-Winters exponential smoothing forecaster.

    Default settings use simple exponential smoothing without trend and
    seasonality components.

    Parameters
    ----------
    trend : {"add", "mul", "additive", "multiplicative", None}, default=None
        Type of trend component.
    damped_trend : bool, default=False
        Should the trend component be damped.
    seasonal : {"add", "mul", "additive", "multiplicative", None}, default=None
        Type of seasonal component.Takes one of
    sp : int or None, default=None
        The number of seasonal periods to consider.
    initial_level : float or None, default=None
        The alpha value of the simple exponential smoothing, if the value
        is set then this value will be used as the value.
    initial_trend : float or None, default=None
        The beta value of the Holt's trend method, if the value is
        set then this value will be used as the value.
    initial_seasonal : float or None, default=None
        The gamma value of the holt winters seasonal method, if the value
        is set then this value will be used as the value.
    use_boxcox : {True, False, 'log', float}, default=None
        Should the Box-Cox transform be applied to the data first?
        If 'log' then apply the log. If float then use lambda equal to float.
    initialization_method:{'estimated','heuristic','legacy-heuristic','known',None},
        default='estimated'
        Method for initialize the recursions.
        If 'known' initialization is used, then `initial_level` must be
        passed, as well as `initial_trend` and `initial_seasonal` if
        applicable.
        'heuristic' uses a heuristic based on the data to estimate initial
        level, trend, and seasonal state. 'estimated' uses the same heuristic
        as initial guesses, but then estimates the initial states as part of
        the fitting process.```

fkiraly:
>when pip installing sktime 0.9.0 I still don't see the new params in ExponentialSmoothing
`Downloading sktime-0.9.0-cp38-cp38-macosx_10_15_x86_64.whl

@atsangarides, it's not in 0.9.0, only in the current *development branch*.
As such, it is scheduled to appear in the next MINOR release (0.10.0 probably), but won't ever be in 0.9.0.

You can use it right now if you follow the install guide for the most recent development version
https://www.sktime.org/en/stable/installation.html
of course there's always a higher risk for bugs compared to the release.

atsangarides:
> > when pip installing sktime 0.9.0 I still don't see the new params in ExponentialSmoothing
> > `Downloading sktime-0.9.0-cp38-cp38-macosx_10_15_x86_64.whl
> 
> @atsangarides, it's not in 0.9.0, only in the current _development branch_. As such, it is scheduled to appear in the next MINOR release (0.10.0 probably, sometime in the next weeks), but won't ever be in 0.9.0.
> 
> You can use it right now if you follow the install guide for the most recent development version https://www.sktime.org/en/stable/installation.html of course there's always a higher risk for bugs compared to the release.

ah, how silly of me! apologies, and thanks for the quick response!

PR #1851, opened 2022-01-08
[ENH] TSC refactor:  TSF, RSF

### LINKED ISSUE #1589: [BUG] using hasattr on RandomIntervalSpectralForest to test for feature_importances_ causes the program to crash.
**Describe the bug**
When I use ```hasattr``` to ask ```RandomIntervalSpectralForest``` if it has the feature_importances_ attribute it causes my code to crash with the following error:
```NotImplementedError: The impurity-based feature importances of RandomIntervalSpectralForest is currently not supported.```

**To Reproduce**

```python
from sktime.classification.interval_based import RandomIntervalSpectralForest
test = RandomIntervalSpectralForest()
hasattr(test, 'feature_importances_')
```

**Expected behavior**
I expected ```hasattr(test, 'feature_importances_')``` to return ```False```.

**Possible fix**
I think it would be better to remove this method unless it is implemented.

### LINKED ISSUE #1609: [ENH] Refactor RandomIntervalSpectralForest to comply with the new fit/predict/predict_proba model
RandomIntervalSpectralForest is an alternative form of RISE. I have marked it for deprecation, but if someone wants to maintain it could they please adapt to the new model?  It inherits from both ForestClassifier and BaseClassifier which complicates matters. 

could wrap it.

deprecate and stick with the original. RISE is not currently competitive

TonyBagnall:
#### Reference Issues/PRs
fixes #1068 and fixes #1609 
also fixes #1589 

this adds wrapper methods for fit, predict and predict proba with the pattern used in https://github.com/alan-turing-institute/sktime/pull/1761

the problem is these classifiers use multiple inheritance, from BaseClassifier and sklearn classifiers. If we do not override the methods fit and predict then they are inherited from sklearn rather than BaseClassifier. These wrapper methods are the simplest way around this problem. They redirect fit to BaseClassifier fit. See #1761 for a discussion of alternatives. 

Not this also effects the TSF based regressor. This has been adapted to work, but the solution is temporary since  the BaseRegressor class will soon be refactored.

Also based on https://github.com/alan-turing-institute/sktime/issues/1589, the skeleton method feature_importances_ has been removed

### DIFF
--- a/sktime/classification/base.py
+++ b/sktime/classification/base.py
@@ -182,7 +182,6 @@ class BaseClassifier(BaseEstimator):
 
         # boilerplate input checks for predict-like methods
         X = self._check_convert_X_for_predict(X)
-
         return self._predict_proba(X)
 
     def score(self, X, y) -> float:
--- a/sktime/classification/interval_based/_risf.py
+++ b/sktime/classification/interval_based/_risf.py
@@ -124,7 +124,7 @@ class RandomIntervalSpectralForest(ForestClassifier, BaseClassifier):
 
     def predict(self, X, **kwargs) -> np.ndarray:
         """Wrap predict to call BaseClassifier.predict."""
-        BaseClassifier.predict(self, X, **kwargs)
+        return BaseClassifier.predict(self, X, **kwargs)
 
     def predict_proba(self, X, **kwargs) -> np.ndarray:
         """Wrap predict_proba to call BaseClassifier.predict_proba."""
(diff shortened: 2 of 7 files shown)


fkiraly (approved):
Yes, this seems sensible! Finally, I didn't know how to fix this.

One request, would you be able to call the `BaseClassifier` methods via kwargs instead of args, i.e., `X=X, y=y` and so on? That seems "safer". But I wouldn't see it as a blocker if you're against.

TonyBagnall:
> Yes, this seems sensible! Finally, I didn't know how to fix this.
> 
> One request, would you be able to call the `BaseClassifier` methods via kwargs instead of args, i.e., `X=X, y=y` and so on? That seems "safer". But I wouldn't see it as a blocker if you're against.

sure, just pushed the change

fkiraly (approved):

PR #1854, opened 2022-01-08
[ENH] Transformers module full refactor - part II, `panel` module (1st half)

fkiraly:
This PR refactors all transformers in the `transformations` module to the new interface.

This is part II - a sizeable chunk of the `transformations.panel` module.

For all transformers, the following was carried out:
* inheritance parent is now `BaseTransformer`
* adding tags as appropriate
* moving logic to underscore methods `_fit`, `_transform`, and `_inverse_transform` (where applicable)
* removal of check/conversion logic which is already covered by `BaseTransformer`
* linting and cleaning up docstrings

Reviewers should focus on individual transformers and the general pattern.

This PR includes bugfixes and changes to `BaseTransformer` from #1855, these should be reviewed there.

This PR makes the following changes to `BaseTransformer`:
* added `requires_y` tag for transformers that need a `y` passed in `fit`. This is also used in filtering scenarios with `y` being passed.
* raises error if no `y` is passed but `requires_y` tag is `True`
* `y` is converted only if it is not ignored internally (`y_inner_mtype` is not `"None"`), otherwise replaced by `None`
* allowed `y` depend now on type of `X`. If `X` is `Series`, then `y` must be `Series`. If `X` is `Panel`, `y` can be `Panel` or `Table`.

This PR also contains some changes to individual transformers:
* the PCA transformer was not correctly implemented - it returned principal scores (primitives), not the projected principal components (time series). This was fixed, and dependent tests removed.
* tests for PCA duplicated tests in the general test suite. These were removed.
* `Detrender` sets its multivariate tag correctly now, depending on the forecaster that is passed
* simplified `ExponentTransformer` to use `pandas` onboard functionality directly, and removed requirement to run `fit` (no training is required in a simple exponentiation).
* test for `TSFreshFeatureExtractor` no longer creates unused variables

Further changes:
* added a `pandas.Series` typed `Table` mtype, to allow `y` in transformers to be passed as `pd.Series` if a `Table` is meant (as in various tests of former "panel-to-tabular" transformers, such as the `TSFreshRelevantFeatureExtractor`)
* changed transform scenarios so number of time points in `fit` is same as in `transform`
* `y` argument was missing in the `transform` of the scenario `TransformerFitTransformPanelUnivariateWithClassY`, fixed
* `ShapeDTW` called the `SlidingWindowSegmenter` transformer, assuming `numpy3D` input and `nested_univ` output, which was hard-coded in the transformer. As that had changed, `convert` needed to be introduced.
* `ShapeDTW` was also calling legacy conversion functions. Replaced by suitable `convert` calls.
* `_conditional_fixtures` now has a functionality to raise errors instead of just passing them. By default, the fixture generation now raises errors if fixture construction fails.
* removed strange check in `check_series_to_primitive_transform_multivariate` that would assert that univariate series-to-primitives transformers always return data frames with 3 columns. Why would that always be true? Must be leftovers from an incomplete coding session.

Dependence on other PR:
* `test_persistence_via_pickle` now uses scenarios, removed `_make_args` dependency from `test_all_estimators` entirely https://github.com/alan-turing-institute/sktime/pull/2105


fkiraly:
merging after self-review, as discussed in dev meeting.

PR #1855, opened 2022-01-08
[BUG] transformer base class changes and bugfixes

fkiraly:
This PR collects bugfixes and changes to `BaseTransformer` arising from #1854.

Bugfixes:
* the output logic for `Panel` returns of `_transform` and `_inverse_transform` was incorrect. The line that was introduced to deal with multivariate outputs when `Series` are passed accidentally blocked all `Panel` returns. The check is now done only if the return is `Series`.
* when a `Series` is passed to `transform` and the internal `_transform` supports only `Panel`, and the internal mtype is `nested_univ`, the back-conversion function `Panel`->`Series` could not cope, since it was not compatible with the `nested_univ` mtype. An additional `convert_to` fixes this.
* checks and conversion for `convert_Panel_to_Series`, was incorrect for `np.ndarray` input (3D `Panel` -> 2D, `Series`)
* relaxed condition when attempting `convert_panel_to_Series` and passing `y`: the case that `y` is not `None` but we know that it is ignored anyway (`self.get_tag("y_inner_mtype")=None`) no longer throws an error.

### DIFF
--- a/sktime/datatypes/_series_as_panel/_convert.py
+++ b/sktime/datatypes/_series_as_panel/_convert.py
@@ -77,8 +77,7 @@ def convert_Panel_to_Series(obj, store=None):
         obj.index = obj.index.droplevel(level=0)
 
     if isinstance(obj, np.ndarray):
-        shape = obj.shape
-        if not len(shape) == 3 or not shape[0] == 1:
+        if obj.ndim != 3 or obj.shape[0] != 1:
             raise ValueError("if obj is np.ndarray, must be of dim 3, with shape[0]=1")
         obj = np.reshape(obj, (obj.shape[1], obj.shape[2]))
 
--- a/sktime/datatypes/_series_as_panel/_convert.py
+++ b/sktime/datatypes/_series_as_panel/_convert.py
@@ -78,7 +78,7 @@ def convert_Panel_to_Series(obj, store=None):
 
     if isinstance(obj, np.ndarray):
         shape = obj.shape
-        if not len(shape == 3) or not shape[2] == 1:
+        if not len(shape) == 3 or not shape[2] == 1:
             raise ValueError("if obj is np.ndarray, must be of dim 3, with shape[2]=1")
         obj = np.reshape(obj, (obj.shape[0], obj.shape[1]))
 
--- a/sktime/datatypes/_series_as_panel/_convert.py
+++ b/sktime/datatypes/_series_as_panel/_convert.py
@@ -78,8 +78,8 @@ def convert_Panel_to_Series(obj, store=None):
 
     if isinstance(obj, np.ndarray):
         shape = obj.shape
-        if not len(shape) == 3 or not shape[2] == 1:
-            raise ValueError("if obj is np.ndarray, must be of dim 3, with shape[2]=1")
-        obj = np.reshape(obj, (obj.shape[0], obj.shape[1]))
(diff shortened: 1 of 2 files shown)


sktime/datatypes/_series_as_panel/_convert.py
```diff
@@ -78,8 +78,8 @@ def convert_Panel_to_Series(obj, store=None):
 
     if isinstance(obj, np.ndarray):
         shape = obj.shape
-        if not len(shape == 3) or not shape[2] == 1:
-            raise ValueError("if obj is np.ndarray, must be of dim 3, with shape[2]=1")
-        obj = np.reshape(obj, (obj.shape[0], obj.shape[1]))
+        if not len(shape) == 3 or not shape[0] == 1:
```
lmmentel, line 81:
```suggestion
        if obj.ndim != 3 or obj.shape[0] != 1:
```

fkiraly: yes, that's better

sktime/transformations/base.py
```diff
@@ -211,10 +211,11 @@ def fit(self, X, y=None, Z=None):
 
         # 3. internal only has Series but X is Panel: loop over instances
         elif X_input_scitype == "Panel" and "Panel" not in X_inner_scitypes:
-            if y is not None:
+            if y is not None and self.get_tag("y_inner_mtype") != "None":
                 raise ValueError(
-                    "no default behaviour if _fit does not support Panel, "
-                    " but X is Panel and y is not None"
+                    "no default behaviour if _fit does not support Panel and "
+                    'self.get_tag("y_inner_mtype") is not "None",'
```
lmmentel, line 217: If this is intended as a message to the user it might not be as helpful as anticipated. Consider adding a suggestion about what to change.

sktime/transformations/base.py
```diff
@@ -644,10 +645,11 @@ def update(self, X, y=None, Z=None, update_params=True):
...
```
lmmentel, line 651: If this is intended as a message to the user it might not be as helpful as anticipated. Consider adding a suggestion about what to change.

sktime/transformations/base.py
```diff
@@ -671,10 +673,11 @@ def _vectorized_transform(self, X, X_input_mtype=None, y=None, inverse=False):
...
         if X_input_mtype is None:
             X_input_mtype = mtype(X, as_scitype=["Series", "Panel"])
-        if y is not None:
-            ValueError(
-                "no default behaviour if _fit does not support Panel, "
-                " but X is Panel and y is not None"
+        if y is not None and self.get_tag("y_inner_mtype") != "None":
+            raise ValueError(
+                "no default behaviour if _fit does not support Panel and "
+                'self.get_tag("y_inner_mtype") is not "None",'
```
lmmentel, line 679: If this is intended as a message to the user it might not be as helpful as anticipated. Consider adding a suggestion about what to change. BTW this is repeated 3rd time at this point, maybe it could be a custom Exception?

sktime/transformations/base.py
fkiraly: ok, suggested sth better here. Should be more useful to a user who is not a developer.

fkiraly: (same error messages three times, so resolved the others to keep discussion here, understanding that all comments apply to all three instances)

chrisholder (approved):

Issue #1857, opened 2022-01-09
[DOC] Formulae in `distances` module seem incorrect


fkiraly:
The formulae in the `distances` module docstrings seem incorrect:

* `wdtw_distance`, `wddtw_distance`: There is reference to a weight vector `w`, but it is not an input to the formula. How is `w` computed? (or defined?)
* `ddtw_distance`: symbol `x` is unbound on LHS, but does not appear on RHS; `d_q` is referenced in "where" but does not appear at all. Further, is this meant to be the second finite difference, then it's not correct.
* `ddtw_distance`, `wddtw_distance`: the distance itself is never defined.

FYI @TonyBagnall, @chrisholder.

fkiraly:
FYI @Lovkush-A due to pure math background.

PR #1858, opened 2022-01-09
[ENH] Composable distances interface prototype for numba distance module

fkiraly:
This PR contains the new `sktime` native, `numba` based distances wrapped as composable distance classes in `dists_kernels`.

There is one new class, `DtwDist`, which wraps the four dynamic time warping distances in `distances` as `BaseEstimator` compatible objects in `dists_kernels`.

This PR also contains some minor modifications to the panel distance base class, `BasePairwiseTransformerPanel`, and testing:
* now allows arbitrary internal mtypes, since the `sktime` native distances require 3D `numpy`, not `df-list` internal format
* for that, moved `_pairwise_panel_x_check` into the class to be in line with other base classes' input checking and coercion
* `fit` related tests were escaped since before work on `BaseObject`. This has been removed, and remedied by adding the `"fit-in-transform"` tag to all pairwise transformers (set to `True` by default).

I would particularly appreciate a review by @TonyBagnall and @chrisholder.

PS: while doing this, I noted that the math in the `distances` module docstrings seems incorrect or incomplete, I've collated the points in issue #1857.

### DIFF
--- a/sktime/dists_kernels/dtw.py
+++ b/sktime/dists_kernels/dtw.py
@@ -1,0 +1,170 @@
+# -*- coding: utf-8 -*-
+"""BaseEstimator interface to sktime dtw distances in distances module."""
+
+__author__ = ["fkiraly"]
+
+import numpy as np
+
+from typing import Union
+
+from sktime.distances import pairwise_distance
+from sktime.dists_kernels._base import BasePairwiseTransformerPanel
+
+
+class DtwDist(BasePairwiseTransformerPanel):
+    r"""Interface to sktime native dtw distances, with derivative or weighting.
+
+    Interface to simple dynamic time warping (DTW) distance,
+    and the following weighted/derivative versions:
+    WDTW - weighted dynamic tyme warping
+    DDTW - derivative dynamic time warping
+    WDDTW - weighted derivative dynamic time warping
+
+    DTW:
+    Originally proposed in [1]_, DTW computes the distance between two time series by
+    considering their alignments during the calculation. This is done by measuring
+    the pointwise distance (normally using Euclidean) between all elements of the two
+    time series and then using dynamic programming to find the warping path
+    that minimises the total pointwise distance between realigned series.
+
+    DDTW is an adaptation of DTW originally proposed in [2]_. DDTW attempts to
+    improve on dtw by better account for the 'shape' of the time series.
+    This is done by considering y axis data points as higher level features of 'shape'.
+    To do this the first derivative of the sequence is taken, and then using this
+    derived sequence a dtw computation is done.
+
+    WDTW was first proposed in [3]_, it adds a multiplicative weight penalty based on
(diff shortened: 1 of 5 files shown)


TonyBagnall:
you might want to wait until we have finished the distances/cython purge, distances are not finalised, Chris has an ongoing PR on it I think

fkiraly:
> you might want to wait until we have finished the distances/cython purge, distances are not finalised, Chris has an ongoing PR on it I think

This would probably work regardless, it's really just a light wrapper.
I can also remove the wrapper and include only the modifications in the base class and testing.

Which PR are you referring to?

fkiraly:
tagging @chrisholder

sktime/dists_kernels/dtw.py
```diff
@@ -0,0 +1,171 @@
...
+
+        Returns
+        -------
+        distmat: np.array of shape [n, m]
+            (i,j)-th entry contains distance/kernel between X[i] and X2[j]
+        """
+        metric_key = self.metric_key
+        kwargs = self.kwargs
+
+        distmat = pairwise_distance(X, X2, metric=metric_key, **kwargs)
```
lmmentel, line 169: it seems like a lot of boilerplate code to just call one line... any reason to go through the hassle?

fkiraly: the one-line distances do not expose parameters in `sklearn` compatible style. If you want to use them in composition patterns like "k-nearest neighbors with that distance" and tune the distance parameters, you need to have the boilerplate wrap.

One *could* have implemented the distances right away like this, but the TSC crowd prefers the non-`sklearn` compatible interface. Which can be argued for, since it gives you sth that's more lightweight at a substantial cost of interoperability.

fkiraly: e.g., the lightweight functions don't know which inputs are parameters that need be exposed via `sklearn`-style `get_params`, and which ones are data inputs, e.g., `X`, `X2`.

lmmentel: I'm sure I'm missing something here since `pairwise_distance` if from sklearn so it is sklearn compatible. Is there an example workflow that you are trying to enable? That would help to understand where this is going.

fkiraly: no, the `pairwise_distance` is from `sktime`, the `distances` module?
For the bigger picture of how to slot distances in estimators or aligners in distances and all to be composable with exposed parameters via `get_params`/`set_params` see the STEP here: https://github.com/sktime/enhancement-proposals/blob/main/steps/10_alignment_warping_knn/step.md

lmmentel: OK, it seems I misread the imports. BTW sklearn has `pairwise_distances` not pariwise_distance` :sweat_smile: Thanks for the additional context!

chrisholder (approved):

PR #1847, opened 2022-01-07
[ENH] Numba distance parameter tests and optimisations

chrisholder:
#### What does this implement/fix? Explain your changes.

This pr changes the following:
1. Sakoe chiba window is now a float between 0 and 1 which is the % of the window.
2. itakura max slope is now a float between 0 and 1 which is the % of the max slope height
3. Updated comments
4. Numba compiled the boundings.
5. Add tests for numba distances with parameters.
6. Updated the knn to be able to use these custom parameters.

##### For all contributions
- [x] I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/main/.all-contributorsrc).
- [x] Optionally, I've updated sktime's [CODEOWNERS](https://github.com/alan-turing-institute/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
- [x] I've added unit tests and made sure they pass locally.

##### For new estimators
- [x] I've added the estimator to the online documentation.
- [x] I've updated the existing example notebooks or provided a new one to showcase how my estimator works.

PR #1864, opened 2022-01-11
[ENH] Clustering module refactor

chrisholder:
#### Reference Issues/PRs

This pr relies on pr #1847

This pr refactors the clustering module base class and removes the use of cython distance metrics.

The base class refactor is aimed to be essentially the same as classification for consistency.

With the base class and distances now properly supporting multivariate problems the two implemented algorithms (k-means and k-medoids) have been refactored to properly support this. This includes removing cython distances and greatly simplifying the implementation of both algorithms.

In addition huge performance benefits from refactoring the distances can be seen.

This pr is only missing DBA as a mean averaging for k-means but this will be a large change and therefore this pr is split up from it (refactor for DBA coming in the next few days).

##### For all contributions

##### For new estimators


fkiraly (changes requested):
Tests don´t pass due to issues with moved/deleted folders, can you please fix?

Otherwise, can´t see entirely which tests fail/pass.

TonyBagnall:
seems to be complaining about a missing init file in clustering/tests, although I did not think it was an enforced requirement
image

TonyBagnall:
re the tests, I'm not sure you are allowed to import from contrib in the examples, could be the cause? 
image

chrisholder:
Ah notebooks forgot!

TonyBagnall (commented):
best to not test on Beef, better to use UnitTest for univariate and BasicMotions for multivariate

TonyBagnall (commented):
not sure why the test is failing, will take a look. Just a comment on the base class _check_clusterer_input always converts to a numpy. However, if the series is unequal length, and the clusterer has that capability, then it needs to stay in a dataframe

fkiraly (changes requested):
Happy with this in principle. Great!

I´ve one blocking change request
* can you kindly update the extension template so it is in line with the introduction of `X_inner_mtype` and other various changes?

Non-blocking comments:
* the docstrings now seem to restrictive, since in-principle you can pass any `Panel` input to the functions. Perhaps this can be similar to what the classifier base class has?
* are we certain that we do not need deprecation here? Are all changes downwards compatible? I do not know the module well enough to over look this.

TonyBagnall:
good point about the extension template. The doc string changes are mine, because I desire simplicity in description! I want to restrict usage tbh, there are significant complexities coming our way when we start using unequal length, which will be this year. We can always adapt docstrings later at no cost.

TonyBagnall:
I cant see how it needs deprecating, maybe @chrisholder can comment

chrisholder:
The module was never 'public' as it were before so nothing needs to be deprecated - it isnt even included on the sktime docs if you look here: https://www.sktime.org/en/stable/api_reference.html so I don't really see or think anything needs deprecating.

TonyBagnall (changes requested):
we should have a default predict_proba here, as with BaseClassifier

chrisholder:
What do you want that to return? For fuzzy clustering I could see the use but for default non fuzzy based clustering what do you want it to return?

fkiraly:
> we should have a default predict_proba here, as with BaseClassifier
>What do you want that to return?

Probabilistic cluster assignment? That would make sense in clusterers which, for instance, fit a full mixture distribution. Then, the concept of conditional likelihood of cluster membership makes sense (which would be a pmf).

TonyBagnall (dismissed):

TonyBagnall:
actually, on discussion, predict_proba is problematic for some potential clustering algorithms. so we will defer this feature until we actually need it. This is good to go now, last orders @fkiraly, I want to put it in when it passes tests

TonyBagnall (approved):

fkiraly (changes requested):
Thanks!

Sorry to block, the docstrings in the extension template seems to be incorrect. 
X in `_fit` should say that it´s guaranteed to be of `X_inner_mtype`  - if it´s not numpy, the implementation will break despite assurances by the template.

See the `BaseClassifier` for a formulation that people thought wasn´t too obscure (could be obscure if you just talk about mtypes without ref) and also correct.

TonyBagnall:
Oh ok I'll look next week

TonyBagnall:
@fkiraly I've changed the comments in the extension templates, let me know if they are ok now

fkiraly (approved):
ok, I think this is now good enough.
I have some suggestions on how to make the extension templates clearer, but let´s do this in a separate PR.

Further, is `_predict` really mandatory for clusterers?
I don´t think it should be, not all clustering algorithms result in a unique cluster assignment!
Example: probabilistic/likelihood based; hierarchical or dendrogram clustering which results in a tree but not a cluster assignment.

Issue #1687, opened 2021-12-03
[ENH] Forecasters: prediction interval, probabilistic forecasting interface refactor


fkiraly:
We're currently working on a major refactor of the forecasting core interfaces to allow probabilistic and interval predictions.
The design is here:
https://github.com/sktime/enhancement-proposals/blob/main/steps/13_proba_forecasting/step.md

Major work items:

Design:
- [x] STEP (design) #984

Interface/framework refactor:
- [x] framework: `predict_interval` and `predict_quantiles`
- [x] framework: `predict_var`
- [x] framework: `predict_proba` 

Refactoring or adding individual methods:
- [x] `predict_interval` and `predict_quantiles`
    - [x] `ARIMA`,  `AutoARIMA` (inherit from same intermediate adaptor class) #1874
    - [x] `AutoETS` #1874 
    - [x] BATS/TBATS - https://github.com/alan-turing-institute/sktime/pull/1842 by @k1m190r
    - [x] `ThetaForecaster` - https://github.com/alan-turing-institute/sktime/pull/1579
    - [x] `Prophet`
- [x] `predict_var`
    - [ ] `ARIMA`,  `AutoARIMA` (inherit from same intermediate adaptor class)
    - [ ] `AutoETS`
    - [ ] BATS/TBATS
    - [ ] `ThetaForecaster`
    - [ ] `Prophet`
- [x] `predict_proba` 
    - [ ] `ARIMA`,  `AutoARIMA` (inherit from same intermediate adaptor class)
    - [ ] `AutoETS`
    - [ ] BATS/TBATS
    - [ ] `ThetaForecaster`
    - [ ] `Prophet`
 
Capstone activities:
- [x] refactor capstone
    - [x] `predict_interval` and `predict_quantiles`
    - [x] `predict_var`
    - [x] `predict_proba` 
- [x] deprecation complete

kejsitake:
According to my initial checks the forecastors that should have probabilistic capabilities and still need to be refactored are:

- [ ] ARIMA
- [ ] AutoARIMA
- [ ] AutoETS
- [ ] BATS
- [ ] TBATS
- [ ] Theta - I am working on refactoring predict_quantiles and predict_intervals here https://github.com/alan-turing-institute/sktime/pull/1579

kejsitake:
[...]

PR #1865, opened 2022-01-11
[ENH] - Naive variance prediction estimator

IlyasMoutawwakil:
> #### Reference Issues/PRs
#1687 

> #### What does this implement/fix? Explain your changes.

@fkiraly 
As part of the probabilistic forecasting API, this will implement the `_predict_var` method to forecasters

> #### What should a reviewer concentrate their feedback on?

STEP [probabilistic forecasting API](https://github.com/sktime/enhancement-proposals/blob/main/steps/13_proba_forecasting/step.md)

> #### PR checklist

- [x] add `_predict_var` and its corresponding tag to the forecasting extension template.
- [x] add `NaiveVariance` composite class which implements the 'simple strategy' for variance prediction.
- [x] implement `_predict_var` and its corresponding in `NaiveVariance`.
- [x] implement `_predict_quantiles` in `NaiveVariance` where quantiles are inferred from a normal distribution. 
- [x] implement covariance prediction (`cov=True`)

> ##### For all contributions


IlyasMoutawwakil:
@fkiraly I have created the naive variance estimator and tested it with some forecasters but there some cases where it won't work:
- if the forecaster can't do in sample forecasts on all of `self._y.index`, which can be solved (won't be able to compute variance for in sample forecasts)
- if there's not enough training points (`self._y.index[-1]` < `fh[-1]`)

extension_templates/forecasting.py
```diff
@@ -261,6 +264,31 @@ def _predict_quantiles(self, fh, X=None, alpha=0.5):
...
+        Must be run *after* the forecaster has been fitted.
+
+        Parameters
+        ----------
+        fh : int, list, np.array or ForecastingHorizon
+            Forecasting horizon
+        X : pd.DataFrame, optional (default=None)
+            Exogenous time series
+        cov : bool, optional (default=False)
+            Wether to return marginal variances or covariance matrices.
```
fkiraly, line 280: which one is False? Clarify

extension_templates/forecasting.py
```diff
@@ -261,6 +264,31 @@ def _predict_quantiles(self, fh, X=None, alpha=0.5):
         """
         # implement here
 
+    def _predict_var(self, fh, X=None, cov=False):
+        """
+        Compute/return prediction variance for a forecast.
```
fkiraly, line 269: suggest, in formatting and style, try to be consistent with `_predict`

extension_templates/forecasting.py
```diff
@@ -261,6 +264,31 @@ def _predict_quantiles(self, fh, X=None, alpha=0.5):
...
+
+        Must be run *after* the forecaster has been fitted.
+
+        Parameters
+        ----------
+        fh : int, list, np.array or ForecastingHorizon
```
fkiraly, line 275: when it shows up in `_predict_var`, it´s *always* a `ForecastingHorizon` - compare `_predict` docstring

extension_templates/forecasting.py
```diff
@@ -261,6 +264,31 @@ def _predict_quantiles(self, fh, X=None, alpha=0.5):
...
+            Forecasting horizon
+        X : pd.DataFrame, optional (default=None)
+            Exogenous time series
+        cov : bool, optional (default=False)
+            Wether to return marginal variances or covariance matrices.
+
+        Returns
+        -------
+        pred_var : np.ndarray
```
fkiraly, line 284: come to think of it, it should probably be a `DataFrame`, where row indices (and in the case of `cov=True`, column indices) are those of `fh`

sktime/forecasting/naive_variance.py
```diff
@@ -0,0 +1,192 @@
...
+        "ignores-exogeneous-X": True,
+        "capability:pred_int": True,
+        # deprecated and will be renamed to capability:predict_quantiles in 0.11.0
+        "capability:pred_var": True,
+        # deprecated and will be renamed to capability:predict_variance in 0.11.0
+    }
+
+    def __init__(self, forecaster):
+
+        self.forecaster = forecaster
```
fkiraly, line 42: component forecasters should be cloned, or the tests will fail!

sktime/forecasting/naive_variance.py
```diff
@@ -0,0 +1,192 @@
...
+        Returns
+        -------
+        pred_quantiles : pd.DataFrame
+            Column has multi-index: first level is variable name from y in fit,
+                second level being the quantile forecasts for each alpha.
+                Quantile forecasts are calculated for each a in alpha.
+            Row index is fh. Entries are quantile forecasts, for var in col index,
+                at quantile probability in second-level col index, for each row index.
+        """
+        self.check_is_fitted()
```
fkiraly, line 90: not necessary, done in `predict_quantiles`

sktime/forecasting/naive_variance.py
```diff
@@ -0,0 +1,192 @@
...
+        errors = []
+        for a in alpha:
+            z = _zscore(1 - a)
+            errors.append(pred_var * z)
+
+        y_pred = self.predict(fh, X)
+
+        pred_quantiles = pd.DataFrame()
+        for a, error in zip(alpha, errors):
+            if a < 0.5:
```
fkiraly, line 104: this is also not needed? `norm.ppf` comes with the correct sign, automatically.

sktime/forecasting/naive_variance.py
```diff
@@ -0,0 +1,192 @@
...
+
+        Returns
+        -------
+        pred_var : np.ndarray
+            if cov=False,
+                a vector of same length as fh with predictive marginal variances;
+            if cov=True,
+                a square matrix of size len(fh) with predictive covariance matrix.
+        """
+        self.forecaster.check_is_fitted()
```
fkiraly, line 138: the checks are not needed in the private function, they are all in the public function. In this case, you just haven´t implemented it yet (see below)

sktime/forecasting/naive_variance.py
```diff
@@ -0,0 +1,192 @@
...
+
+        fh_ = check_fh(fh)
+
+        n_timepoints = len(self._y)
+        residuals = []
+        for i in range(0, n_timepoints):
```
fkiraly, line 144: instead of doing this manually, you can call `update_predict`

fkiraly (changes requested):
Good start!

Regarding the design:
* you should also implement `predict_var` (without underscore) in the `BaseForecaster`. Boilerplate methods like `check_fh` should go there - see how the other `predict_etc` methods are implemented.
* I think it´s better to return not a `numpy.ndarray`, but a `pd.DataFrame` (so the index is preserved)

I also left comments in the code.

Some bugs not mentioned there:
* your code does not work if the forecaster is multivariate
* I think your `_predict_quantiles` is incorrect
* is your `_predict_var` implementing the formula we discussed? I think not? I also recommend to put the math in the docstring.

sktime/forecasting/naive_variance.py
```diff
@@ -0,0 +1,192 @@
...
+from scipy.stats import norm
+from sklearn.base import clone
+
+from sktime.forecasting.base import BaseForecaster
+from sktime.utils.validation.forecasting import check_alpha, check_fh
+
+
+class NaiveVariance(BaseForecaster):
+    """
+    Compute the prediction variance using the 'simple strategy'.
```
fkiraly, line 19: put the math here, or it´s totally unclear what this implements.

sktime/forecasting/naive_variance.py
```diff
@@ -0,0 +1,192 @@
...
+        n_timepoints = len(self._y)
+        residuals = []
+        for i in range(0, n_timepoints):
+            # a duplicate for rolling window fit/predict
+            forecaster = clone(self.forecaster)
+
+            try:
+                forecaster.fit(self._y[: i + 1])
+                y_hat = forecaster.predict(self._y.index, self._X)
+                residuals.append(y_hat - self._y)
```
fkiraly, line 151: alternatively, you can use `predict_residuals`

fkiraly:
> if the forecaster can't do in sample forecasts on all of self._y.index, which can be solved (won't be able to compute variance for in sample forecasts)

I´m slightly confused - are your predictions not all out of sample, with a rolling window?

sktime/forecasting/naive_variance.py
```diff
@@ -0,0 +1,192 @@
...
+                at quantile probability in second-level col index, for each row index.
+        """
+        self.check_is_fitted()
+        alpha = check_alpha(alpha)
+
+        pred_var = self.predict_var(fh, X)
+
+        errors = []
+        for a in alpha:
+            z = _zscore(1 - a)
```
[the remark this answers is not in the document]
IlyasMoutawwakil: Great! I will use that.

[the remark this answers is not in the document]
fkiraly: perhaps without the "one minus", you need to check which way round it is...

extension_templates/forecasting.py
IlyasMoutawwakil: So no `pd.Series` at all ?

IlyasMoutawwakil:
> I´m slightly confused - are your predictions not all out of sample, with a rolling window?

they are all out of sample now. I actually made it compute in sample as well so that it would be possible to compute in sample variance but I removed it now.

fkiraly:
> they are all out of sample now. I actually made it compute in sample as well so that it would be possible to compute in sample variance but I removed it now.

ok, makes sense - this could be a parameter option of the estimator (in sample or out of sample, say; more generally, which splitter is used how)

Let me know when you would like another review.

extension_templates/forecasting.py
IlyasMoutawwakil: clarrified:
        cov : bool, optional (default=False)
            if True, computes the covariance matrix.
            if False, computes the marginal variance.

extension_templates/forecasting.py
IlyasMoutawwakil: yes it's now `"""Compute/return...`

extension_templates/forecasting.py
IlyasMoutawwakil: Understood, the `predict` method in `BaseForecaster` does the check/traformation of `fh`

extension_templates/forecasting.py
IlyasMoutawwakil: Done, it's `Dataframe` now

sktime/forecasting/naive_variance.py
IlyasMoutawwakil: Makes sense, it's cloned now.

sktime/forecasting/naive_variance.py
IlyasMoutawwakil: makes sense

sktime/forecasting/naive_variance.py
IlyasMoutawwakil: implemented the public function.

sktime/forecasting/naive_variance.py
IlyasMoutawwakil: since each forecaster has a minimum length of training time points, especially when they have a trend (two points) or a seasonality (sometimes it's twice the seasonality)

sktime/forecasting/naive_variance.py
IlyasMoutawwakil: Just found how to use it, nice shortcut

sktime/forecasting/naive_variance.py
IlyasMoutawwakil: I put a description but I'm not sure if what you mean is to put latex formulas

fkiraly:
I see you requested a review - the first thing I note is that tests are failing.
Would you like me to review why the tests are failing?

IlyasMoutawwakil:
> I see you requested a review - the first thing I note is that tests are failing. Would you like me to review why the tests are failing?

Yes, I still don't understand how it's failing if I didn't add any tests for the `NaiveVariance` class

fkiraly:
> Yes, I still don't understand how it's failing if I didn't add any tests for the NaiveVariance class

Ok. The general idea is that `sktime` automatically runs tests to any estimator that is added.
Which tests are run are also automatically determined, by the type of the estimator (in thie case, a forecaster).

You will need to ensure that the tests in `test_all_estimators` and the tests in `test_all_forecasters` pass.
To debug and check what is causing the failures (there are many), it is strongly advised that you get your local developer set up running, and ensure that you are able to run tests locally and see which ones fail.
Consult the developer documentation, or you can also bring it up in one of the collaboration sessions.

fkiraly:
@IlyasMoutawwakil, your test failures appear like they are related to #2118.
So far, the common pattern seems to be that this appears in composite forecasters.

@GuzalBulatova has encountered this recently, and @aiwalter a while ago.
Do we have an understanding what is going on for composite forecasters with the `update_predict_predicted_index`?

sktime/forecasting/naive_variance.py
```diff
@@ -0,0 +1,184 @@
+# -*- coding: utf-8 -*-
```
aiwalter, line 1: can you pls add this to `naive.py` instead of having a new file

sktime/forecasting/naive_variance.py
```diff
@@ -0,0 +1,184 @@
...
+            "handles-missing-data",
+            "y_inner_mtype",
+            "X_inner_mtype",
+            "X-y-must-have-same-index",
+            "enforce_index_type",
+        ]
+        self.clone_tags(self.forecaster, tags_to_clone)
+
+    def _fit(self, y, X=None, fh=None):
+        self.fitted_forecaster = clone(self.forecaster)
```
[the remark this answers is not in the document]
IlyasMoutawwakil: okay will do and leave `self.forecaster` as is (just a holder of the configuration of the internal forecaster) ?

sktime/forecasting/naive_variance.py
IlyasMoutawwakil: it's a separate estimator because it should be able to work with all other forecasters (allowing us to add naive variance strategy to any forecaster).

sktime/forecasting/naive_variance.py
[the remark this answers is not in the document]
aiwalter: yes

IlyasMoutawwakil:
Finally found why predict quantiles and intervals tests were failing. It was because I didn't use `pd.MultiIndex` to index quantiles and instead used plain indexes and columns

sktime/forecasting/naive_variance.py
IlyasMoutawwakil: ah right I see what you mean, I will move it there.

fkiraly:
I made some minor changes to the base class docstrings, extension template, and signature, to ensure everything is coherent after the batch of recent merges.

fkiraly (approved):
Thanks for implementing!

Excellent, does not seem to require any additional changes imo.

I took the liberty of ensuring that this consistently merges with a number of other PR on probabilistic forecasting recently merged. Mostly docstring changes, but I also made sure that the signatures are consistent.

PR #1873, opened 2022-01-13
[ENH] Base classifier update

TonyBagnall:
#### Reference Issues/PRs
Fixes https://github.com/alan-turing-institute/sktime/issues/1708

This PR is conditional on https://github.com/alan-turing-institute/sktime/pull/1864

this ended up changing more files than I wished for because of making the base class access via init rather than directly. This PR does the following

1. make the BaseClassifier implement ABC and makes _fit and _predict abstract. I wanted to do this now just to make sure all classifiers comply to fit/_fit model. 
2. makes the containing file abstract (_base) to conform to other modules. This has touched a fair few files and made this a bigger PR than I wanted, but done it now and ....
3. corrects about 200 docstrings that resulted from this refactor in benchmarking/strategies and proximity forest (a LOT in pf).
4. correct and change the docstrings in BaseClassifier
5. BaseClassifier now stores the characteristics of the input data, so further checks can be made in predict. One reason for this is to allow for early classification capability. predict now checks that the length of the series passed to predict is the same as that passed to fit, unless the classifier has the capability for unequal length or it can make early predictions. 

I have left a copy of base.py in there because MrSEQL was causing problems, need to rebuild the cpp, but equally might want to deprecate base, so not sure. I think I will remove base when MrSEQL goes


fkiraly:
Nice!

Though, if we switch to `abc`, we should do uniformly throughout the package. We also ought to have a design discussion on this.

Also, do you know what happens to methods that are abstract but not mandatory to implement_ That is, how would one best deal with this in `abc`?

TonyBagnall:
> Nice!
> 
> Though, if we switch to `abc`, we should do uniformly throughout the package. We also ought to have a design discussion on this.
> 
> Also, do you know what happens to methods that are abstract but not mandatory to implement? That is, how would one best deal with this in `abc`?

ABC is already used a lot throughout sktime, clustering, benchmarking and distances. By definition, you cannot have an "optional" abstract method. I guess you would use a method stub instead, raising an exception if called. However, imo if its optional and has no sensible default behaviour, it should not be in the base class.

TonyBagnall:
I need to go through and replace

TonyBagnall:
> Nice!
> 
> Though, if we switch to `abc`, we should do uniformly throughout the package. We also ought to have a design discussion on this.
> 
> Also, do you know what happens to methods that are abstract but not mandatory to implement? That is, how would one best deal with this in `abc`?

this is not a change of design, simply a different way of implementing abstract methods. Rather than raise an error, we are using ABC. If you want to discuss how to use ABC elsewhere, fine by me, but its not really needed for this PR

fkiraly (changes requested):
Ok, I think there are two serious blockers:

* the duplication `base`/`_base` which is caused by MrSEQL. Let's park this PR until we have removed MrSEQL and that is no longer necessary. Let's not have hacks around key modules, especially not ones that duplicate files! I'd be really really hesitant moving that state into a release!
* we should discuss use of `abc` at one of the developer meetings. I feel the design point around optional abstract methods unresolved, e.g., `predict_proba`. If we park until the C code is deprecated, we should have ample time.

sktime/classification/_base.py
```diff
@@ -0,0 +1,516 @@
...
+        unequal = not self._X_metadata["is_equal_length"]
+        # Check this classifier can handle these three characteristics
+        self._check_capabilities(missing, multivariate, unequal)
+        # Convert data as dictated by the classifier tag self.get_tag("X_inner_mtype")
+        X = self._convert_X(X)
+        # Store series length for checking in predict
+        if self.get_tag("X_inner_mtype") == "numpy3D":
+            self._series_length = self._find_series_length(X)
+        else:
+            self._series_length = 0
```
fkiraly, line 125: but the series could be of equal length even if it is not of `numpy3D` mtype?

sktime/classification/_base.py
```diff
@@ -0,0 +1,516 @@
...
+        missing = self._X_metadata["has_nans"]
+        multivariate = not self._X_metadata["is_univariate"]
+        unequal = not self._X_metadata["is_equal_length"]
+        # Check this classifier can handle these three characteristics
+        self._check_capabilities(missing, multivariate, unequal)
+        # Convert data as dictated by the classifier tag self.get_tag("X_inner_mtype")
+        X = self._convert_X(X)
+        # Store series length for checking in predict
+        if self.get_tag("X_inner_mtype") == "numpy3D":
+            self._series_length = self._find_series_length(X)
```
fkiraly, line 123: why do we need this? the `_series_length` seems nowhere used in the base class?

sktime/classification/_base.py
```diff
@@ -0,0 +1,516 @@
...
+        if not self.get_tag("capability:unequal_length") and not self.get_tag(
+            "capability:early_prediction"
+        ):
+            # check input is the same length as train data, if using numpy
+            if self.get_tag("X_inner_mtype") == "numpy3D":
+                length = self._find_series_length(X)
+                if length != self._series_length:
+                    raise ValueError(
+                        "Error in predict: input series different length to "
+                        "training series, but tags capability:unequal_length and "
```
fkiraly, line 176: perhaps say which values of the tags mean what?

sktime/classification/_base.py
fkiraly: if it is an important property for `predict` series and `fit` seris to have the same length, should we make this a metadata field that is returned by the checks?

TonyBagnall: > Ok, I think there are two serious blockers:
> 
> * the duplication `base`/`_base` which is caused by MrSEQL. Let's park this PR until we have removed MrSEQL and that is no longer necessary. Let's not have hacks around key modules, especially not ones that duplicate files! I'd be really really hesitant moving that state into a release!

ok, thats fair. Pehaps I should just deprecate MrSEQL? 

> * we should discuss use of `abc` at one of the developer meetings. I feel the design point around optional abstract methods unresolved, e.g., `predict_proba`. If we park until the C code is deprecated, we should have ample time.

I dont agree that this should block the PR. I am not changing existing code structure with this PR, predict_proba is the same is were before. For sure consider other uses/designs for ABC, but please dont shoehorn it onto this PR. ABC is already used in benchmarking, clustering and distances.

TonyBagnall:
I'll revisit this after the CI/CD and version 10 release, since it needs to wait until that is done.

fkiraly:
why close? There is a lot of useful work here, let´s leave it open!

I would also recommend, there are a lot of docstring fixes - would you perhaps make a separate PR out of these, so the changes here are indeed localized to the base class?
We don´t need to wait on the docstring fixes.

PR #1874, opened 2022-01-14
[ENH] Add predict_quantiles to ets, pmdarima adapter

kejsitake:
#### Reference Issues/PRs

Progress on #1687 

- [x] Implements predict_quantiles for ets
- [x] Implements predict_quantiles for pmdarima adapter
- [x] extends pmdarima to support a list of alphas instead of a single alpha


fkiraly:
looks good on superficial view.

May I ask, what was the outcome of your investigation that ETS produced non-symmetric predictive intervals?
(non-symmetric around the mean/point forecast)

kejsitake:
I had misunderstood the output of predict from statsmodels. Once I printed that out, it became clearer and I could spot the mistake. This issue has been solved on this version.

For consistency with tbats and theta, the output of predict-interval is still a two level dataframe, but I can change that if necessary.

fkiraly:
> For consistency with tbats and theta, the output of predict-interval is still a two level dataframe, but I can change that if necessary.

Yes, fully agreed - let's keep it that way, for now. Let's not mix output types, and change it either not at all or consistently.

sktime/forecasting/ets.py
```diff
@@ -447,15 +404,64 @@ def _predict(
...
+            Row index is fh. Entries are quantile forecasts, for var in col index,
+                at quantile probability in second col index, for the row index.
+        """
+        self.check_is_fitted()
+        alpha = check_alpha(alpha)
+        # convert alpha to the one needed for predict intervals
+        alphas = []
+        for a in alpha:
+            if a < 0.5:
+                alphas.append(round(((0.5 - a) * 2), 4))
```
fkiraly, line 440: I think no rounding should take place here

sktime/forecasting/ets.py
```diff
@@ -447,15 +404,64 @@ def _predict(
...
+        valid_indices = fh.to_absolute(self.cutoff).to_pandas()
+
+        start, end = valid_indices[[0, -1]]
+        prediction_results = self._fitted_forecaster.get_prediction(
+            start=start, end=end
+        )
+
+        pred_quantiles = pd.DataFrame()
+        for a, coverage in zip(alpha, alphas):
+            pred_int = prediction_results.pred_int(round(1 - coverage, 4))
```
fkiraly, line 455: or here

sktime/forecasting/ets.py
```diff
@@ -447,15 +404,64 @@ def _predict(
...
+            pred_int.columns = ["lower", "upper"]
+            if a < 0.5:
+                pred_quantiles[a] = pred_int["lower"].loc[valid_indices]
+            else:
+                pred_quantiles[a] = pred_int["upper"].loc[valid_indices]
+        arrays = [len(alpha) * ["Quantiles"], alpha]
+        index = pd.MultiIndex.from_tuples(list(zip(*arrays)))
```
fkiraly, line 462: if you use `from_prod`, zou don't need to use `zip` etc.

sktime/forecasting/base/adapters/_pmdarima.py
```diff
@@ -117,45 +131,162 @@ def _predict_in_sample(
...
+
+        Returns
+        -------
+        quantiles : pd.DataFrame
+            Column has multi-index: first level is variable name from y in fit,
+                second level being the values of alpha passed to the function.
+            Row index is fh. Entries are quantile forecasts, for var in col index,
+                at quantile probability in second col index, for the row index.
+        """
+        self.check_is_fitted()
```
fkiraly, line 228: this line is unnecessary, already checked in the public method

sktime/forecasting/base/adapters/_pmdarima.py
```diff
@@ -117,45 +131,162 @@ def _predict_in_sample(
...
+        alpha = check_alpha(alpha)
+
+        # convert alpha to the one needed for predict intervals
+        alphas = []
```
fkiraly, line 232: perhaps better call this `coverage`

sktime/forecasting/base/adapters/_pmdarima.py
```diff
@@ -117,45 +131,162 @@ def _predict_in_sample(
...
+        if self.fh.is_all_out_of_sample(self.cutoff):
+            y_oos_pred, y_oos_pred_int = self._predict_fixed_cutoff(fh_oos, **kwargs)
+            pred_quantiles = pd.DataFrame()
+            for oos_intervals, a in zip(y_oos_pred_int, alpha):
+                if a < 0.5:
+                    pred_quantiles[a] = oos_intervals["lower"]
+                else:
+                    pred_quantiles[a] = oos_intervals["upper"]
+            arrays = [len(alpha) * ["Quantiles"], alpha]
+            index = pd.MultiIndex.from_tuples(list(zip(*arrays)))
```
fkiraly, line 255: if you use `from_prod`, no need to use `zip`

fkiraly (changes requested):
Nice! This is the last existing probabilistic forecaster, right?

Some minor changes above.

The one that I'd consider is a blocker is the rounding of `alpha` in your `_predict_quantiles`. This is because this violates the interface contract - the floats in the column `MultiIndex` should be identical with the `alpha` passed to `_predict_quantiles`.

sktime/forecasting/base/adapters/_pmdarima.py
kejsitake: Removed

sktime/forecasting/base/adapters/_pmdarima.py
kejsitake: changed in the new verison

sktime/forecasting/base/adapters/_pmdarima.py
kejsitake: edited

kejsitake:
Thanks for the comments! I believe I fixed all of the issues! Yes, I think these are the last two probabilistic forecasters.

fkiraly:
are you certain you have pushed your newest changes to the remote branch?
It does appear there have been no changes since my review, so I suspect you may have accidentally forgotten to do this.

kejsitake:
Hi! I had amended my commit, I don't know if that makes it more difficult to review or was causing the issue, but I can avoid doing this in the future. I also spotted some other small things, so I added a new commit.

In the 4th tab `Files changed`, I can see that `predict_quantiles` doesn't round the alpha anymore and the usage of `pd.multiIndex.from_product`, which was not there in the previous version.

fkiraly (approved):
Looks good now, thanks!

Maybe it doesn´t properly update the displayed changes if you amend a commit, instead of making a new commit?
Anyway, thanks, all fine now.

Issue #1877, opened 2022-01-14
[ENH] Design: default behaviour between `predict_interval`, `predict_quantiles`, and `predict_proba`


fkiraly:
Design - we should define what the default behaviour is, since one of these, when implemented, can be called by the other functions ot provide output.

fkiraly:
Ok, here´s my thoughts, @kejsitake, @k1m190r:

We have three functions that can be implemented or not:
* `_predict_interval`
* `_predict_quantiles`
* `_predict_proba`

Here´s how I think defaulting should go:
* if `_predict_interval` is not implemented but `_predict_quantiles` is, `_predict_interval` should call `_predict_quantiles` at quantiles correponding to upper/lowers for the coverage probability asked for, and convert to interval predictions.
* if `_predict_quantiles` is not implemented but `_predict_interval` is, `_predict_quantiles` should call `_predict_interval` at coverage probabilities correponding to the quantiles asked for, and convert to quantile predictions.
* if neither of `_predict_interval` and `_predict_quantiles` are implemented, these should both call methods of the distribution returned by `_predict_proba` if that is implemented.
* If `_predict_proba` is not implemented but one of `_predict_interval` and `_predict_quantiles` is, it should return a distribution with an `inverse_cdf` method (or the corresponding method in the framework we´ll adopt) that is wrapping `predict_quantiles`, and a corresponding `cdf` (possibly using bisection inversion)
* if none of the methods are implemented, an error is raised.

fkiraly:
There's a fourth method, `_predict_var`, hm.

If `_predict_proba` is implemented and `_predict_var` is not, `_predict_var` should call `variance` of the `_predict_proba` return.
If neither of the two are implemented, `_predict_var` should raise an error.

Otherwise, I don't think there are useful relations that `_predict_var` can use, if `_predict_interval` or `_predict_quantiles` are implemented (there are no general algebraic relations between variance and finite evaluateions of the latter, without any parametric assumptions, I think).

PR #1842, opened 2022-01-06
[ENH] Prediction intervals refactor: BATS/TBATS; bugfix for #1625; base class updates on `predict_quantiles`

k1m190r:
#### Reference Issues/PRs
 Fixes #1625 

#### What does this implement/fix? Explain your changes.
Fixed #1625. 
Implemented TBATS `_predict_quantiles()`.

##### Re-implemented `_has_predict_quantiles_been_refactored()` due to failing tests.

[Previous implementation](https://github.com/alan-turing-institute/sktime/blob/bce815d11fff62b52882f13cd650c88972b08ab6/sktime/forecasting/base/_base.py#L1250):  
```python
def _has_predict_quantiles_been_refactored(self):
    if "_predict_quantiles" in type(self).__dict__.keys():
        return True
    else:
        return False
```
Relied on the expectation that `type(self)` (last class in inheritance chain) would be class implementing `_predict_quantiles()`, and thus `type(self).__dict__.keys()` would have `"_predict_quantiles"`. 

BUT in case of TBATS `type(self)` is [TBATS class](https://github.com/alan-turing-institute/sktime/blob/bce815d11fff62b52882f13cd650c88972b08ab6/sktime/forecasting/tbats.py#L21) which does not re-implement itself but rather inherits implementation from [_TbatsAdapter](https://github.com/alan-turing-institute/sktime/blob/bce815d11fff62b52882f13cd650c88972b08ab6/sktime/forecasting/base/adapters/_tbats.py#L18). Thus `type(self).__dict__.keys()` does **not** have `"_predict_quantiles"`, and returns _undesired_ `False`.

Suggested implementation:
```python
def _has_predict_quantiles_been_refactored(self):
    """Check if specific forecaster implements _predict_quantiles()."""
    base_pq = BaseForecaster._predict_quantiles
    this_pq = self.__class__._predict_quantiles
    # true if self's _predict_quantiles is new implementation
    return base_pq != this_pq

```
[...]

PR #1879, opened 2022-01-15
[ENH] Defaults for `_predict_interval` and `_predict_coverage`

fkiraly:
This PR implements defaults for `_predict_interval` and `_predict_quantiles`, as proposed in #1877 - for discussion (not necessarily as final version). This does not yet contain defaults in terms of `_predict_proba`, since we have not settled on output format yet.

It is partly based on work in @k1m190r's PR #1842 and @kejsitake in #1579, behaviour moved to base class default implementations.

Behaviour implemented:
* if `_predict_interval`  is not implemented but `_predict_quantiles` is, it is called and outputs are reformatted
* if `_predict_quantiles`  is not implemented but `_predict_interval` is, it is called and outputs are reformatted
* the check whether sth is implemented is done by a utility added to `BaseObject`, a method `_has_implementation_of`, which is based on @k1m190r's utility that checks whether the forecaster has been refactored, which was based on @kejsitake's and some discussion between the three of us.

Further, the following forecasters were relying primarily on `_predict_quantiles` but interfacing interval prediction logic. This has been changed by tidying up the interface, implementing a leaner `_predict_interval`, and removing `_predict_quantiles`:
* `_PmdArimaAdapter`
* `_TbatsAdapter`
* `_ProphetAdapter`

Additional, minor changes:
* docstring fixes in `BaseForecaster`
* multiple minor instances of clean-up in the forecasting extension template
* shortened `ThetaForecaster` internals
* fix for bug in `ThetaForecaster`, where intervals were computed for 1-coverage, not coverage. Same in test for `ThetaForecaster`.
* fix for bug #1954 in `_ProphetAdapter`
* renamed `test_predict_pred_interval` to `test_predict_interval` for consistency with `test_predict_quantiles`


fkiraly:
(breaking example due to different format in `_predict_interval`)

kejsitake:
I think that this is a good idea because it reduces code repetition in the inheriting forecasting classes, as this methods don't need to be re-implemented. It also looks like `predict_interval' bypasses the floating issue (discussed in PR https://github.com/alan-turing-institute/sktime/pull/1842) by using the user input-ed alpha.

fkiraly:
ok, so what I´d propose is that we first merge @k1m190r´s PR and then I try to make this work.

How about https://github.com/alan-turing-institute/sktime/pull/1874 ?
Perhaps it´s also good if we finish that first?

fkiraly:
(only tests failing are for `Prophet` proba return - which we had missed in #1687)

fkiraly:
@kejsitake, @k1m190r, I've used the "default logic" to refactor the `_PmdArimaAdapter`, which now relies on the `_predict_interval` and not on the `_predict_quantiles` function. All boilerplate moveing between interval and coverage is now in the defaults. Would be appreciative of having your eyes on this, since both of you have implemented similar code.

fkiraly:
PS: thanks for your earlier work which enabled/inspired this refactor! If you hadn't arrived at the same kind of solution independently, it wouldn't have been so clear what to factor out!

fkiraly:
FYI @SveaMeyer13 as well

sktime/base/_base.py
```diff
@@ -323,6 +323,40 @@ def create_test_instances_and_names(cls):
...
+        bool, whether method has implementation in cls
+            True if cls.method has been overridden at least once in
+                the inheritance tree (according to method resolution order)
+        """
+        # walk through method resolution order and inspect methods
+        #   of classes and direct parents, "adjacent" classes in mro
+        mro = inspect.getmro(cls)
+        # collect all methods that are not none
+        methods = [getattr(c, method, None) for c in mro]
+        methods = [c for c in methods if c is not None]
```
lmmentel, line 348:
```suggestion
        methods = [getattr(c, method, None) for c in mro if c is not None]
```

sktime/base/_base.py
```diff
@@ -323,6 +323,40 @@ def create_test_instances_and_names(cls):
...
+
+        for i in range(len(methods) - 1):
```
lmmentel, line 350: how about using a `return len(set(methods)) >= 2 `  instead of this block?

sktime/base/_base.py
fkiraly: yes, that's better

sktime/base/_base.py
fkiraly: I'm not sure whether that's "safe" since we're operating with pointers here?

extension_templates/forecasting.py
```diff
@@ -260,6 +283,55 @@ def _predict_quantiles(self, fh, X=None, alpha=0.5):
...
+                    in the same order as in input `coverage`.
+                Third level is string "lower" or "upper", for lower/upper interval end.
+            Row index is fh. Entries are forecasts of lower/upper interval end,
+                for var in col index, at nominal coverage in second col index,
+                lower/upper depending on third col index, for the row index.
+                Upper/lower interval end forecasts are equivalent to
+                quantile forecasts at alpha = 0.5 - c/2, 0.5 + c/2 for c in coverage.
+        """
+        # implement here
+        # IMPORTANT: avoid side effects to y, X, fh, coverage
```
aiwalter, line 331: side effects should be solved in parent class, like `X = X.copy()`? before calling the private method of the child?

sktime/forecasting/base/_base.py
```diff
@@ -360,6 +360,13 @@ def predict_quantiles(self, fh=None, X=None, alpha=None):
             Row index is fh. Entries are quantile forecasts, for var in col index,
                 at quantile probability in second col index, for the row index.
         """
+        if not self.get_tag("capability:pred_int"):
```
aiwalter, line 363: is this tag now valid for both `predict_interval` and `predict_quantiles`? So if one of both is implemented as a private method in a child, then this tag is `True`? is this documented anywhere what this tag is doing now?

aiwalter: maybe lets rename the tag so that its clear that its covering both cases? E.g. `pred_int_or_quantiles`

sktime/forecasting/base/_base.py
```diff
@@ -1251,42 +1275,78 @@ def _predict_interval(self, fh, X=None, coverage=0.90):
...
+        implements_proba = self._has_implementation_of("_predict_proba")
+        can_do_proba = implements_interval or implements_quantiles or implements_proba
+
+        if not can_do_proba:
+            raise RuntimeError(
+                f"{self.__class__.__name__} does not implement "
+                "probabilistic forecasting, "
+                'but "capability:predict_int" flag has been set to True incorrectly. '
+                "This is likely a bug, please report, and/or set the flag to False."
+            )
```
aiwalter, line 1305: should this not go into a unit test instead here?

fkiraly:
merged after self-review, as agreed with @TonyBagnall

aiwalter (commented):
left few comments, didnt look into all in detail

sktime/forecasting/base/_base.py
fkiraly: > So if one of both is implemented as a private method in a child, then this tag is True? is this documented anywhere what this tag is doing now?

Agreed to adding that info in the tag. Yes, it now stands for "is one of the interval-like methods implemented".
Regarding renaming, I was going to avoid it this close to a new version since that would be a minor interface change but still one (e.g., people referring to the tag)

sktime/forecasting/base/_base.py
fkiraly: yes, good idea

extension_templates/forecasting.py
fkiraly: this would be a general question about `predict` and all.

I'd say: no, copying around wastes memory - it's perfectly fine passing references if they are not mutated, so we should not enforce a copy.
We have a test which checks whether there are side effects to arguments, `test_methods_have_no_side_effects`

sktime/forecasting/base/_base.py
fkiraly: fixed in https://github.com/alan-turing-institute/sktime/pull/1961

sktime/forecasting/base/_base.py
fkiraly: added in https://github.com/alan-turing-institute/sktime/pull/1961

Issue #1678, opened 2021-12-02
Estimator Test Parameters Refactor


Aparna-Sakshi:
### Aim
To start to move default parameters out of `tests/_config` and into the concrete estimators, by specifying the parameters inside the `get_test_params` method.

This would declutter the logic, and make extending sktime much easier - since it reduces the "points of extension" by one. Default parameters would no longer have to be entered in _config but become part of the list of things to implement, in the extension template.

### Related Issue
#1098 

### Step
[link for step](https://github.com/sktime/enhancement-proposals/blob/main/steps/12_estimator_test_instances/step.md)
This issue is associated with point 3 of the **suggested refactor sequence** given in the above step.

### Contributing Guidelines
If you would like to work on this: please post here and which estimator you are taking on, to avoid duplication of work.
A core dev will then add your name to the estimator under refactoring and/or link the PR.

A general recipe for refactoring is as follows:
- pick up an estimator class from the below checklist of estimators.
- look for parameters of the estimator in `ESTIMATOR_TEST_PARAMS` (in the tests._config file)
- add `get_test_params` method to the corresponding estimator class which returns a dictionary of the above parameters 
- refer to [extension template](https://github.com/alan-turing-institute/sktime/tree/main/extension_templates) to add get_test_params to the corresponding estimator class
- For example, in case of `ColumnEnsembleForecaster` , dictionary of parameters which is `{"forecasters": FORECASTER} `, should be returned from `get_test_params` method. 

Checklist of estimators:

- [x] ColumnEnsembleForecaster
- [x] OnlineEnsembleForecaster
- [x] FeatureUnion
- [x] DirectTabularRegressionForecaster
- [x] MultioutputTabularRegressionForecaster
- [x] RecursiveTabularRegressionForecaster
- [x] DirRecTabularRegressionForecaster
- [x] DirectTimeSeriesRegressionForecaster
- [x] RecursiveTimeSeriesRegressionForecaster
- [x] MultioutputTimeSeriesRegressionForecaster
- [x] DirRecTimeSeriesRegressionForecaster
- [x] TransformedTargetForecaster
- [x] ForecastingPipeline
- [x] EnsembleForecaster
- [x] StackingForecaster
- [x] AutoEnsembleForecaster
- [x] Detrender
- [x] ForecastingGridSearchCV
- [x] ForecastingRandomizedSearchCV
- [x] TabularToSeriesAdaptor
- [x] ColumnEnsembleClassifier
[...]

PR #1902, opened 2022-01-21
[ENH] moving forecaster test params from `_config` into classes - all forecasters excluding reduction

fkiraly:
This PR continues the refactor in #1678 by moving the parameters of most forecasters (all excluding reductions) from the `_config` to the classes.
Also removes the parameters for one distance that was erroneously still there.


lmmentel (approved):
Nice!

Issue #1712, opened 2021-12-07
[BUG] FBProphet causes an issue if exogenous X already contains a column named ds


kejsitake:
**Describe the bug**

As discussed in PR https://github.com/alan-turing-institute/sktime/pull/1711: 
If the series passed does have the` ds `column, the forecaster raises an error, which it shouldn't. 

**To Reproduce**

```python
from sktime.forecasting.fbprophet import Prophet
import pandas as pd
import numpy as np

y_train = pd.Series(np.arange(1,21), index=pd.date_range(start='1/1/2021', end='20/1/2021'))
X_train = pd.DataFrame(data=np.arange(1,21), columns=['ds'], index=pd.date_range(start='1/1/2021', end='20/1/2021'))

arima_forecaster = Prophet()
arima_forecaster.fit_predict(fh=y_train.index, y=y_train, X=X_train)
print(X_train)
```
**Error Thrown**
`ValueError: Column name 'ds' is reserved in fbprophet`

**Expected behavior**

The forecaster should not throw an error in this case

**Additional Context**

In https://github.com/alan-turing-institute/sktime/pull/1711, there were two approaches suggested to solve this issue:
First by @fkiraly 
- write a functions df_to_prophet which takes X and does not have any side effects on it.
- df_to_prophet moves index of X to ds column. If a ds column is already present, it renames it to a name that doesn't exist yet in the columns. For instance, the longest column name plus an underscore or similar.
- the function is applied whenever X is passed.

Second approach would be not renaming the column. I am not sure how to handle the issue in this case.

aiwalter:
for me this is not a bug, as described in #1711

fkiraly:
Welllll, I think it's a bug, since it's non-compliance with the base class interface.

Why would/should one algorithm throw an error when we ask for a forecast, just because it has a column with a special name in it? None of the forecasters do this, and it's not part of the usage contract. All algorithms should be able to produce forecasts in alignment with their abstract capabilities.

aiwalter:
but when we want to copy an existing columns `df` we should then check if this is numeric, bcz. it might contain an index already. If there is an index, we should probably raise the error or overwrite it. Otherwise we will end up with 2 index columns and it will raise an error probably and confuse user.

benHeid:
The following code is working, thus I cannot reproduce the error anymore. It seems that #1910 introduced a fix.

```python
import numpy as np
import pandas as pd
from sktime.forecasting.fbprophet import Prophet

[...]

PR #1910, opened 2022-01-22
Add predict_quantiles to FBprophet

kejsitake:
#### Reference Issues/PRs

See https://github.com/alan-turing-institute/sktime/issues/1687

- Refactors `predict_quantiles` for FBProphet.
- Changes the test to only check if `pred_int['upper'] >= pred_int['lower']`.
- Implements a fix for when the X contains a variable named "ds", by copying the column to another column and renaming in the case the column is numeric and not an index. This is done in order to be more consistent with the other forecasters, which don't throw an error if exogenous X contains a column "ds" related to #1712.


fkiraly:
@kejsitake, might the defaults implemented in https://github.com/alan-turing-institute/sktime/pull/1879 perhaps help with this?
It would allow you to implement `_predict_interval` only.

kejsitake:
> @kejsitake, might the defaults implemented in #1879 perhaps help with this? It would allow you to implement `_predict_interval` only.

Thank you. I will explore implementing just `predict_interval`. I suspect that the issues I am encountering are due to the additional index ("ds") that Prophet generates and requires.

fkiraly:
fron our discussion - looks like this failed `assert np.all(pred_int["upper"] > y_pred)`.
Generally, the predictive mean (`y_pred`) does not need to have a relation with any quantile (`pred_int["upper"]`), so that is not a correct check on the output. So, occasionally it can happen that this fails.

The fact that the tests run locally for you and me, I'd put down to random number generator behaviour.

In short: looks like the test is broken, not the estimator.

I'd just change the two lines 276, 277 of the test in question to
`assert np.all(pred_int["upper"] >= pred_int["lower"])`

fkiraly:
& looks like you already fixed it!

sktime/forecasting/base/adapters/_fbprophet.py
```diff
@@ -76,21 +75,17 @@ def _fit(self, y, X=None, fh=None, **fit_params):
...
+    def _predict(self, fh=None, X=None):
         """Forecast time series at future horizon.
 
         Parameters
         ----------
         fh : array-like
             The forecasters horizon with the steps ahead to to predict.
             Default is
-            one-step ahead forecast, i.e. np.array([1]).
+            one-step ahead forecast, i.e. np.arraygit r([1]).
```
fkiraly, line 86: typo

sktime/forecasting/base/adapters/_fbprophet.py
```diff
@@ -114,19 +109,86 @@ def _predict(self, fh=None, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
...
+
+        Returns
+        -------
+        quantiles : pd.DataFrame
+            Column has multi-index: first level is variable name from y in fit,
+                second level being the values of alpha passed to the function.
+            Row index is fh. Entries are quantile forecasts, for var in col index,
+                at quantile probability in second col index, for the row index.
+        """
+        self.check_is_fitted()
```
fkiraly, line 148: these two checks are already done in the "non-underscore" function, kindly remove

sktime/forecasting/tests/test_all_forecasters.py
```diff
@@ -327,6 +327,7 @@ def test_predict_pred_interval(Forecaster, fh, alpha):
 def _check_predict_quantiles(
     pred_quantiles: pd.DataFrame, y_train: pd.Series, fh, alpha
 ):
+
```
fkiraly, line 330: typo, remove empty line before function docstring

fkiraly:
Quick question: have you checked that it's fine to overwrite the `inverval_width` late? Do the quantiles look sensible?

kejsitake:
Thank you for the comments!

> Quick question: have you checked that it's fine to overwrite the `inverval_width` late? Do the quantiles look sensible, i.e., that you get similar answers when you change it early, compared to changing it late?

I am not sure what do you mean by early and late. Do you mean when creating object vs. changing the private variable? If yes, I tried printing the output and it looked reasonable.

fkiraly:
> Do you mean when creating object vs. changing the private variable? If yes, I tried printing the output and it looked reasonable.

Yes, that's exactly what I meant, early = at construction, late = before invoking prediction. Sorry for being unclear.
If it looked reasonable/similar, probably all is fine...

sktime/forecasting/base/adapters/_fbprophet.py
kejsitake: Removed!

fkiraly (approved):
Thanks, looks good!

Tests ran and passed, I just accidentally restarted them.

PR #1917, opened 2022-01-23
Test PR to check display of CI checks

fkiraly:
DO NOT MERGE

This is a test PR to check whether PR branching after the CI refactor #1620 still display the old checks, in the GitHub web interface.


fkiraly:
yes, the old stuff is still displaying.

What is really strange is that the display, at the start, seems to jump between the old set and the new set of tests.
At the end, both are displayed.

FYI @freddyaboulton, @lmmentel - any idea what is going on?

fkiraly:
& readthedocs do not display as completed, even though it completes.

Issue #1914, opened 2022-01-23
Relax "numpy==1.19.3" dependency 


TKassis:
**Is your feature request related to a problem? Please describe.**
Currently there is a hard dependency for numpy 1.19.3. Is this actually needed or is there room for newer numpy versions?

**Describe the solution you'd like**
Can this be relaxed to 1.20?

lmmentel:
Hi @TKassis, thanks for reporting this. We are currently working on this. There are already similar issues posted so we might close this and give another issue to track instead.

fkiraly:
duplicate of #1794, closing.
Yes, indeed this is sth we're working on, now that the CI/CD is upgraded.

PR #1921, opened 2022-01-23
[MNT] Bump numpy version to >=1.21.0, <1.22

### LINKED ISSUE #1787: StatsModels and Numpy to have a >= version instead of == version
Would it be possible to use StatsModels >= 0.12.1 and Numpy >= 1.19.3 instead of StatsModels == 0.12.1 and Numpy == 1.19.3?

Currently the file main/setup.py uses the same version for StatsModels  and Numpy  in min and max requirements

### LINKED ISSUE #1794: [ENH] Upgrading to numpy 1.20
sktime requires numpy <1.20. This severely limits its use in larger more modern projects. Moving to numpy 1.20 would help allow sktime to be used in more projects without causing dependency issues.

lmmentel:
#### Reference Issues/PRs

- Fixes #1794 (& duplicate #1914)
- Fixes #1787 

Bumping numpy to >=1.21.0,<1.22 

##### For all contributions

##### For new estimators


lmmentel:
Looks like nothing break after bumping numpy :slightly_smiling_face: :rocket:

TonyBagnall (approved):

fkiraly (approved):
Great!

Quick Q, why does nothing break? Because we discontinued support for `python 3.6` (which in turn influences the upper `pandas` bound)?

lmmentel:
> Quick Q, why does nothing break? Because we discontinued support for `python 3.6` (which in turn influences the upper `pandas` bound)?

This is not an easy question to answer, but recent changes in #1620  made it easy to introduce this change also dropping 3.6 made release some of the constraints on package versions.

fkiraly:
So, here's another question, @lmmentel: why a lower bound of 01.21.0? This does not include the old version, and may break things for users.
Would it be problematic to have this at 1.19.3 (the old version)?

lmmentel:
I bumped it since I got a dependabot alert about my other package that there are vulnerabilities in numpy the recommended action is to bump it at least to 1.21, I can decrease the lower bound if you are afraid of users not being able to use sktime. BTW this will only affect people installing sktime 0.10.0 onward so I don't think it's an issue.

fkiraly:
> I bumped it since I got a dependabot alert about my other package

This violates the "Thou shalt have no other packages before sktime" clause in the sktime code of conduct ... or am I confusing this with another literature source? :-)

ilkersigirci:
Is there any specific reason not to include numpy version 1.22 ?

fkiraly:
@ilkersigirci, since this is a general Q, answering it here: https://github.com/alan-turing-institute/sktime/discussions/2037

Issue #1716, opened 2021-12-07
[ENH] Implement ElasticEnsemble with numba distances


TonyBagnall:
adapt EE to use the new numba distances

Issue #1715, opened 2021-12-07
[ENH] update ProximityForest


TonyBagnall:
1. Adapt proximity forest to use numba not cython
2. Address possible outstanding issues
https://github.com/alan-turing-institute/sktime/issues/600
https://github.com/alan-turing-institute/sktime/issues/598
https://github.com/alan-turing-institute/sktime/issues/1114

fkiraly:
I've added a clarification that the ref to closed issues does not mean "resolved".

Issue #1714, opened 2021-12-07
[ENH] numba distances enhancements 


TonyBagnall:
a few things
- [x] 1. numbas version of move-split-merge
- [x] 2. handle variable length? (KNN not enabled to do this, capability andled in base class)
https://github.com/alan-turing-institute/sktime/issues/645
- [x] 3. Check parameters
https://github.com/alan-turing-institute/sktime/issues/589
- [x] 4. Check derivatives
https://github.com/alan-turing-institute/sktime/issues/654
- [x] 5. Confirm unit tests in place (distances all now tested)
https://github.com/alan-turing-institute/sktime/issues/646

TonyBagnall:
branch looking at this https://github.com/alan-turing-institute/sktime/tree/numba_neighbours

TonyBagnall:
ERP and LCSS results are not comparable to tsml. This may be implementation, but they seem worse, so this should be investigated further

TonyBagnall:
resolved by https://github.com/alan-turing-institute/sktime/pull/2268

PR #1926, opened 2022-01-24
[ENH] introduce msm distance and adapt KNN classifier to use it.

TonyBagnall:
Include a numba MSM distance function and adapt KNN to use it.

linked to #1714 #1715 and #1716

still need to sort out the multivariate version using Shifaz's method


jasonlines (dismissed):
Looks good to me - only small additions to the existing code that are within the existing established style. Documentation looks sound too

jasonlines (approved):

PR #1929, opened 2022-01-24
[ENH] update convert_to to allow multiple scitypes among target mtypes

fkiraly:
This PR resolves a problem encountered in both #1818 and #1924 when one would like to convert an input that can be of a union-scitype to multiple target mtypes (also of different scitypes possible), where the conversion should prioritize conversion to an mtype of the *same* scitype.

Currently, the naive input for this will run into an error since if multiple mtypes are provided as possible target `to_list`, and the input is of an mtype not on the `to_list`, then `convert_to` will select the first one on `to_list` to convert to, which may be of a different scitype as the input, and results in an error.

This PR modifies the behaviour that the first mtype *of the same scitype as the input* is converted to, if the input mtype is not on the list of `to_type`-s.


danbartl:
image

I still get an error for the multiindex, specifically here:

image

fkiraly:
Ah, thanks for pointing this out, @danbartl!

This is an interaction between the `Table` scitype and how I implemented inference of the scitype - if multiple scitypes fit, the initial inference would break, because it does not narrow down based on the `to_type` list.

I added that narrowing down, and now the inferred scitype should be unique. Can you try again with this improved version?

fkiraly:
PS: in the original code, it should have worked via specifying `as_scitype = ["Series", "Panel", "Hierarchical"]`, that would do the narrowing down by hand, but it was unclear to the user that this needed be done from the error. So I adapted the code to the user expectation, and this parameter does not need to be specified.

fkiraly:
PPS: I added tests that should catch sth like this from happening.

fkiraly:
Fine to merge without core dev review, as discussed today.
& @danbartl has reviewed this.

Issue #1923, opened 2022-01-23
[ENH] investigate use of abc for base classes, write STEP


fkiraly:
@TonyBagnall recently suggested to use of abc for classifiers in *#1873, #1708), and this was also previously discussed a number of times: #1424, #1708, #510, #420.

I do think we need to work out a STEP for this, it would be a major change to the interface framework.

There are some open questions that I have:
* which classes would inherit from `ABCMeta`?
* how do we deal with optional abstract methods?
* how does this integrate with the tag system in `BaseObject`? This could be updated using `ChainMap`.
* are there any problematic interactions of the public/private nesting pattern `fit`/`_fit`?
* we would still like to have capability tags that tell us whether an optional method is implemented or not, e.g., `predict_proba` or `predict_interval`. How would this interact with `abc`?
* how would downwards compatibility and deprecation work here?
* what about polymorphic estimators, e.g., a pipeline that can act like a transformer or a forecaster? Would `abc` not complain about missing transformer methods if it happens to behave like a forecaster and vice versa?

There's probably more once we have a proper design discussion, this is only off the top of my head.

chrisholder:
Why is this an issue. This is the standard best practice way of doing abstract methods in python why wouldn't we do it.

chrisholder:
To answer your questions more directly though I think you're overcomplicating  a lot of it. Literally it will feel the same as the current raise NotImplementedError except it's standardised and more util methods (and ides can better support auto implement functionality):

- which classes would inherit from ABCMeta?
BaseClusterer, BaseClassifier, BaseForecaster etc. Essentially any estimator/transformer that current uses raise NotImplementedError(). This is because ABC just does what that does.

- how do we deal with optional abstract methods?
Optional abstract methods aren't really abstract method. What I think you are wanting is just method overriding. So add a default behaviour and override it in the parent class when you want to change the default behaviour.

- how does this integrate with the tag system in BaseObject? This could be updated using ChainMap.
This wouldn't touch any of this?

- are there any problematic interactions of the public/private nesting pattern fit/_fit?
No.

- we would still like to have capability tags that tell us whether an optional method is implemented or not, e.g., predict_proba or predict_interval. How would this interact with abc?
It doesn't touch it. It's an abstract method it doesn't touch functionality.

- how would downwards compatibility and deprecation work here?
Abc has been supported since 3.6. Again you don't deprecate anything you're just moving away from raise NotImplementedError().

[...]

PR #1931, opened 2022-01-25
[ENH] make _fit and _predict abstract in BaseClassifier

### LINKED ISSUE #1708: [ENH] Make BaseClassifier abstract with ABC

both _fit and _predict should be properly abstract using ABC, rather than just raising an error. 

This cannot be done until all classifiers conform to _fit/fit model. This is linked here

- #1609
- #1608
- #1607
- #1606

this is complicated by the fact that it seems TimeSeriesForestClassifier is used in annotation, benchmarking, forecasting, series_as_features, transformations and utils.

TonyBagnall:
this PR makes _fit and _predict abstract with ABC, and nothing else.

#### Reference Issues/PRs
fixes #1708

 I do not think this is contentious or requires a core developer meeting to approve, for the following reasons

1. The methods are already conceptually abstract, this just allows us to enforce it, implicitly checking all classifiers properly extend the base class. There is no design change. 
2. Using ABC is a completely standard way of making abstract methods in python.
3. ABC is already used in forecasting, benchmarking, clustering and distances. 

any debate about possible other uses of ABC, or changes in the current design utilising ABC, are, in my opinion, secondary to this PR.

no


chrisholder (approved):

fkiraly (dismissed):
apologies for blocking this.

I feel a little contentious and wouldn´t like to move `ABC` in too quickly in a stable module.
I would like to understand the answers to my questions in https://github.com/alan-turing-institute/sktime/issues/1923 first.

Downwards compatibility is quite important, and I can't fully grasp the consequences currently.

Would you mind, at least, doing it after 0.10.0, where we have more time to deal with unintended consequences if arising?

fkiraly:
Re contention:

> I do not think this is contentious or requires a core developer meeting to approve, for the following reasons

Well, contention is subjective, but also quite clearly defined via people having different opinions...

> The methods are already conceptually abstract, this just allows us to enforce it, implicitly checking all classifiers properly extend the base class. There is no design change.

I do think there is - it's subtle, but it's one: a change of decoration vs default method with error handling.
Does not need to be bad - see #1923, could we discuss these Q please?

> Using ABC is a completely standard way of making abstract methods in python.

Yes, and so is the current, implicit way of defining abstract methods.

> ABC is already used in benchmarking, clustering and distances.

These are unstable/experimental modules, while classification and forecasting are stable, and do not use ABC.
I feel we need to be more careful with changes to the stable modules.
The base class inheritance tree is so fundamental that we really can't take a chance to get this wrong imo.

TonyBagnall:
> Re contention:
> 
> > I do not think this is contentious or requires a core developer meeting to approve, for the following reasons
> 
> Well, contention is subjective, but also quite clearly defined via people having different opinions...
> 
> > The methods are already conceptually abstract, this just allows us to enforce it, implicitly checking all classifiers properly extend the base class. There is no design change.
> 
> I do think there is - it's subtle, but it's one: a change of decoration vs default method with error handling. Does not need to be bad - see #1923, could we discuss these Q please?
> 
> > Using ABC is a completely standard way of making abstract methods in python.
> 
> Yes, and so is the current, implicit way of defining abstract methods.
> 
> > ABC is already used in benchmarking, clustering and distances.
> 
> These are unstable/experimental modules, while classification and forecasting are stable, and do not use ABC. I feel we need to be more careful with changes to the stable modules. The base class inheritance tree is so fundamental that we really can't take a chance to get this wrong imo.

I have honestly nothing else to add, other than I think properly abstract methods are better than skeletons that raise errors. I think 99% of programmers would agree with me. But I guess it is not very important. I'll close this and the related issue, if you change your mind let me know

fkiraly:
> I'll close this and the related issue, if you change your mind let me know

I would very much appreciate to hear what you think about the points I raised in #1923.
I´m not against the idea in general, but would like to have a discussion first.
I've put a couple points in #1923, I hope it isn't too much to ask to engage in discussion about a PR before we merge it?

lmmentel (approved):

fkiraly (approved):
Happy with this after our discussion - but why do the mac tests fail?

TonyBagnall:
> Happy with this after our discussion - but why do the mac tests fail?

I cancelled the tests until some other minor PRs go in, will sort it tomorrow

PR #1934, opened 2022-01-26
Added .tsf loader to data_io.py

### LINKED ISSUE #1933: [ENH] add .tsf loader in sktime
Monash Time Series Forecasting Repository has introduced a new file format for storing time series data for forecasting problems called .tsf, inspired by sktime's .ts format. The repository has a lot of nice/popular benchmark datasets, including Australian tourism datasets and 3/5 of the M competitions.

A .tsf reader is available from the Monash repository team here:
https://github.com/rakshitha123/TSForecasting/blob/master/utils/data_loader.py

We should make this available in sktime!

rakshitha123:
#### Reference Issues/PRs
Fixes #1933 

Added .tsf loader from Monash Forecasting Repository to data_io.py

##### For all contributions


TonyBagnall:
hi, this will be great, thanks. You might want to also consider looking at adapting load_from_tsfile, which gives the user the option of setting the return data type. For classification at least I would like to move away from putting everything in a nested dataframe to encourage more use of numpy arrays.

sktime/datasets/tests/test_data_io.py
```diff
@@ -1080,3 +1081,41 @@ def test_write_dataframe_to_ts_fail(tmp_path):
             path=str(tmp_path),
             problem_name="GunPoint",
         )
+
+
+def test_convert_tsf_to_dataframe():
```
ltsaprounis, line 1086: The name of the test to `test_load_tsf_to_dataframe`

ltsaprounis (changes requested):

sktime/datasets/tests/test_data_io.py
```diff
@@ -1080,3 +1081,41 @@ def test_write_dataframe_to_ts_fail(tmp_path):
...
+    """Test function for loading tsf format."""
+    data_path = MODULE + "/data/UnitTest/UnitTest_Tsf_Loader.tsf"
```
ltsaprounis, line 1088: The tests are failing in CI but pass locally. 
The issue has to do with the path to the test .tsf file. 

**local:**
image
**CI**
`FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Miniconda\\envs\\test\\lib\\site-packages\\sktime\\datasets/data/UnitTest/UnitTest_Tsf_Loader.tsf'`

Try this instead of adding the strings: 
```python
import os

data_path = os.path.join(MODULE, "/data/UnitTest/UnitTest_Tsf_Loader.tsf")
```

ltsaprounis (changes requested):

sktime/datasets/tests/test_data_io.py
rakshitha123: Modified the function name.

sktime/datasets/tests/test_data_io.py
rakshitha123: Modified the data_path variable. The given suggestion was a bit modified and ensured that the test is passed locally.

ltsaprounis:
@rakshitha123  - The tests are still failing for the same reason. I can't figure out what the problem is. I'm calling the cavalry! 

@TonyBagnall / @fkiraly - Do you have any idea why we get the ` FileNotFoundError: [Errno 2] No such file or directory: '/Users/runner/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/sktime/datasets/data/UnitTest/UnitTest_Tsf_Loader.tsf'` error in CI? 

The tests pass locally and the similar test for [load_from_tsfile in line 34](https://github.com/alan-turing-institute/sktime/blob/03f4a92136e4184144bbec75af3918ad7bc44bc5/sktime/datasets/tests/test_data_io.py#L41) works fine in the Github actions...

fkiraly:
> Do you have any idea why we get the `FileNotFoundError: [Errno 2] No such file or directory`

Hm, this is clearly due to a file not being found, and either the file or the directory does not exist.
You're welcome.

Jokes aside, I think the issue might be with you using `sktime.__file__` and not just `__file__` in the `os.path`?
If you replicate 1:1 the other test and use the `MODULE` constant, that should fix it - educated guess.

sktime/datasets/tests/test_data_io.py
ltsaprounis: @fkiraly - have a look here as well. This is identical to the other test 🤯

rakshitha123:
Previously, I used a similar code to [load_from_tsfile in line 34](https://github.com/alan-turing-institute/sktime/blob/03f4a92136e4184144bbec75af3918ad7bc44bc5/sktime/datasets/tests/test_data_io.py#L41) to load the test .tsf file. 

After @ltsaprounis pointed out the test failing issue, I used a similar code to [test_write_dataframe_to_ts_success in line 1047](https://github.com/alan-turing-institute/sktime/blob/03f4a92136e4184144bbec75af3918ad7bc44bc5/sktime/datasets/tests/test_data_io.py#L1047) to load the test .tsf file. 

In both cases, the test was passed locally. However, it was failed in CI. I have no idea about the reason, because the tests which have used similar file loading codes are working. What do you think @fkiraly @TonyBagnall @ltsaprounis?

TonyBagnall:
hi, you will need to add the file you are trying to load to the repo, since it is trying to load a file not there. The guide on how to do that is here 
https://www.sktime.org/en/stable/developer_guide/add_dataset.html

fkiraly:
@rakshitha123, update: today, in the collab session I've tried to run the test locally and it succeeded (windows, 3.9). Bit mysterious to me why it fails - could be what @TonyBagnall says, or it could theoretically be the OS/version combination which is not the ones that are failing (they are all 3.7).

rakshitha123:
Many thanks for the feedback @TonyBagnall @fkiraly. I followed the guide on adding datasets and based on that, officially added the tsf unit testing dataset.

rakshitha123:
Seems like the tests are still failing. Any suggestions @fkiraly @TonyBagnall?

TonyBagnall:
try doing them as separate PRs: one to add the example problem, then once that is in the repo, add the loading function. I suspect that this is the part that is confusing the CI (suspicion based purely on I cannot think what else could be doing it!)

rakshitha123:
Thanks @TonyBagnall. I have now added the .tsf dataset as a separate PR: https://github.com/alan-turing-institute/sktime/pull/1996.

fkiraly:
Hm, doesn't look like adding the data set separately helped...

rakshitha123:
Yes, the tests are still failing @TonyBagnall @fkiraly

fkiraly:
really mysterious to me. I see no reason why it should fail, and it passes locally for me.
I tried changing the path object here: https://github.com/alan-turing-institute/sktime/pull/2015
also does not work.

I've asked the other core devs for help, maybe someone sees sth.

ltsaprounis:
It's time to hit this with the big guns! 
I suggest we try: 
```python
from pathlib import Path

data_path = Path(...)
```

lmmentel:
Could you please try adding `*.tsf` at the end of line 7 in `MANIFEST.in` https://github.com/alan-turing-institute/sktime/blob/b02a98b55c30b9fee3ab82af4c2636b03472022c/MANIFEST.in#L7

fkiraly:
THAT WAS IT! And it explains why full repository searches for the other files did not pick the location up, because it uses the star character.

sktime/datasets/_single_problem_loaders.py
```diff
@@ -863,3 +868,40 @@ def load_macroeconomic():
...
+    frequency: str
+        The frequency of the dataset.
+    forecast_horizon: int
+        The expected forecast horizon of the dataset.
+    contain_missing_values: bool
+        Whether the dataset contains missing values or not.
+    contain_equal_length: bool
+        Whether the series have equal lengths or not.
+    """
+    name = "UnitTest"
```
lmmentel, line 890: it seems that the local variables are not rally necessary here

rakshitha123:
Many thanks for all your suggestions @lmmentel @fkiraly @TonyBagnall @ltsaprounis. Seems like this time it is working. One test has already passed!

sktime/datasets/_single_problem_loaders.py
rakshitha123: Removed the extra local variables.

lmmentel (commented):
:+1:

lmmentel (approved):

lmmentel:
Great work @rakshitha123 !

Issue #1763, opened 2021-12-15
[ENH] Design probabilistic prediction metrics


eenticott-shell:
Currently the implemented metrics all evaluate point estimates, but with probabilistic forecasters we may want to evaluate a full distribution or quantile/interval estimates. We need to design for a new base class for handling probabilistic forecasting metrics.

eenticott-shell:
I have began writing up a proposal here:
[Designing probabilistic prediction metrics.md](https://github.com/alan-turing-institute/sktime/files/7719740/Designing.probabilistic.prediction.metrics.md)

Any feedback would be appreciated.

fkiraly:
Thanks @eenticott-shell, looks neat!
I think the general ideas make all sense, I have some detail comments.

comments in general:
* you didn't specify the format of `y_pred`. I assume that is the output of `predict_proba`? Or of `predict_quantiles`? You need to be more specific here. One or the other? Are there perhaps different kinds of metrics, depending on which input they take?
* per design, the metric wouldn't call the estimator directly, but take an output. There are multiple options here, `predict_proba`, `predict_var`, `predict_interval`. How do we know which output the metric takes?
* the pinball loss you have is for a single quantile prediction, but the output of `predict_quantiles` and `predict_interval` returns multiple quantiles. The usual loss here (see Gneiting paper) is the sum, or mean, over all quantiles appearing in the prediction.
* further, there is a special multi-index format that `predict_quantiles` gives you if you have a multivariate forecast. How should the output of the metric look like in this case? I think there are two options, take the average over variables, or don't (keep multi-index). In `scikit-learn`, whether it's one or the other is controlled by the `multioutput` argument.

For the class:
* most importantly: I'd put parameters like the `alpha` for the pinball loss in the constructor (object parameter). All methods, including `__call__`, should expose the same signature `(y_true, y_pred)`, in line with the base class pattern.
* Style-wise I'd not put logic in `__call__` but have it call an `evaluate` method. In case we need checks etc, one might consider `evaluate`/`_evaluate` split (like `fit`/`_fit`), but that may be overkill.
* often, it is very useful to have not just the aggregate score (over multiple test samples), but a per-sample function, i.e., if you have `n` test points, you get a vector of `n` loss values. This could be `evaluate_sample`. Not all metrics will have this (although you can always provide jackknife pseudo-samples).
* tags make sense, good idea. Perhaps separate with colon, not with hyphen (to be in line with `capability` tags)
[...]

PR #1936, opened 2022-01-26
[ENH] Quantile metric design proposal

eenticott-shell:
#### Reference Issues/PRs
Addresses #1763 

A prototype of a new metric class designed for probabilistic or quantile inputs. The specific implementation is for the pinball loss.

The overall structure of the class, public and private method specification, defaults and handling of input types. Some of this remains to be properly implemented. 

##### For all contributions


fkiraly (changes requested):
I assume this is still work in progress?

In terms of design ideas, looks good for me, but this needs to be cleaned up.

* pinball loss won't work for `multioutput`, will it?
* `I don't think `if y_true >= y_pred` runs in either of the places where you compute the loss. You want to vectorize the computation.
* should `_evaluate` of the pinball loss call a vectorized function `pinball_loss`?
* the base class should do input checks and/or conversions
* you'll need to iterate over multiple variables
* please add docstrings
* please add tests

eenticott-shell:
Closing as addressed by #1981

sktime/performance_metrics/forecasting/probabilistic/_classes.py
```diff
@@ -0,0 +1,177 @@
...
+import numpy as np
+
+# TODO: add formal tests
+import pandas as pd
+from sklearn.metrics._regression import _check_reg_targets
+
+from sktime.performance_metrics.forecasting._classes import _BaseForecastingErrorMetric
+
+
+def _check_y_pred_type(metric_object, y_pred):
```
fkiraly, line 14: good idea - I guess this is where we would put checks and the conversions?

There's currently one conversion - `_convert_pred_interval_to_quantiles` in `forecasting.base._base`, and one check, `_check_predict_quantiles`. There is no check for the new "predict interval" format.

sktime/performance_metrics/forecasting/probabilistic/_classes.py
```diff
@@ -0,0 +1,177 @@
...
+    """Base class for probabilistic forecasting error metrics in sktime.
+
+    Extends sktime's BaseMetric to the forecasting interface. Forecasting error
+    metrics measure the error (loss) between forecasts and true values. Lower
+    values are better.
+    """
+
+    _tags = {"scitype:y_pred": "quantiles"}
+
+    def __init__(self, func, name=None, multioutput="uniform_average"):
```
fkiraly, line 32: what is the "func" being used for, is that really necessary?

sktime/performance_metrics/forecasting/probabilistic/_classes.py
```diff
@@ -0,0 +1,177 @@
...
+    ----------
+    y_true : np.array of shape (fh,) or (fh, n_outputs) \
+             where fh is the forecasting horizon
+        Ground truth (correct) target values.
+
+    y_pred : np.array of shape (fh,) or (fh, n_outputs) \
+             where fh is the forecasting horizon
+        Forecasted alpha quantiles
+
+    alpha : The significance
```
fkiraly, line 154: not a significance here, but the fraction at which the quantile is computed (not sure what the right term is, it's not the quantile itself)

fkiraly: I think it would make sense to allow vectorization of alpha? Then do `np.array` casting?

sktime/performance_metrics/forecasting/probabilistic/_classes.py
```diff
@@ -0,0 +1,177 @@
...
+    """
```
fkiraly, line 155: docstring is missing a `Returns` section

fkiraly (commented):
I think this is good, could be turned into a "non-draft" PR.

I left some comments, mainly about docstrings, vectorization, and parameters.

sktime/performance_metrics/forecasting/probabilistic/_classes.py
eenticott-shell: The `func` contains the underlying logic implementation for the metric. It is necessary as I am inheriting from the previous metric class which included it. It can be useful if you need two different metric classes that use the same function in different ways.

sktime/performance_metrics/forecasting/probabilistic/_classes.py
eenticott-shell: Yes this is mostly a placeholder for now. I'll likely add to this in the next PR (adding interval metrics)

sktime/performance_metrics/forecasting/probabilistic/_classes.py
eenticott-shell: I have added this.

sktime/performance_metrics/forecasting/probabilistic/_classes.py
fkiraly: hm, ok - is there a way to make it optional?

I'm thinking about extenders, who will probably fund it very confusing tha `pinball_loss` is passed in two places, and `func` isn't really used in the child, `PinballLoss`.
For instance, why can `_evaluate` not take the place of `func`? Or the other way round?

sktime/performance_metrics/forecasting/probabilistic/_classes.py
```diff
@@ -0,0 +1,225 @@
...
+    ----------
+    alpha : The significance level required for the pinball loss
+
+    multioutput : string "uniform_average" or "raw_values" determines how multioutput
+    results will be treated
+    """
+
+    def __init__(self, alphas, multioutput="uniform_average"):
+        name = "PinballLoss"
+        # _tags = {"scitype:pred": "quantile"}
```
fkiraly, line 102: this should be without comment, and directly in the class, no?

sktime/performance_metrics/forecasting/probabilistic/_classes.py
```diff
@@ -0,0 +1,225 @@
...
+        func = pinball_loss
+        self.alphas = self._check_alpha(alphas)
+        super().__init__(func=func, name=name, multioutput=multioutput)
+
+    def _check_alpha(self, alpha):
```
fkiraly, line 107: would it make sense to put this in a quantile mixin, or in the base class? Other quantile losses will need this too?

sktime/performance_metrics/forecasting/probabilistic/_classes.py
```diff
@@ -0,0 +1,225 @@
...
+        raise TypeError("Alpha should be a float or numpy array")
+
+    def _check_y_pred_quantiles(self, y_pred):
+        assert isinstance(y_pred, pd.DataFrame)
+        assert y_pred.columns.get_level_values(0)[0] == "Quantiles"
+        pred_alphas = y_pred.columns.get_level_values(1)
+        if not all([alph in pred_alphas for alph in self.alphas]):
+            raise ValueError(f"Predictions must include the {self.alphas} quantile.")
+
+    def evaluate(self, y_true, y_pred, multioutput="uniform_average", **kwargs):
```
fkiraly, line 129: this feels like it shoudl be in a base class

fkiraly (approved):
Great, I think for a first go this is excellent.

I think we still should try refine the extension pattern though - think of a user implementing their own metric. Optimally, they would just implement `_evaluate` and `_evaluate_by_index` while inheriting from a suitable base class. Currently, that extension pattern is not possible.

I would suggest we revisit this and:
* cleaner separation between base class and implementation
* moving checks/conversions between probabilistic return types out
* writing an extension template for probabilistic metrics

I'd suggest we merge, then I take this on since I have a clear idea how these changes would look like, unless you want to look at some of the above.

PR #1938, opened 2022-01-27
DRAFT for diagnosing CI/CD: Revert "Add predict_quantiles to FBprophet"

fkiraly:
Reverting #1910, to diagnose whether it is causing the recent CI/CD congestion.


fkiraly:
Hm, no - so this wasn't probably it.

Issue #1941, opened 2022-01-27
[BUG] Identify cause and fix CI/CD stuck in loop on unix, mac - reproduction & diagnostics appreciated


fkiraly:
Since yesterday evening, CI/CD on `main` and PR branching off it have `test-linux` and `test-mac` getting stuck in a loop, running for hours.
Strangely, the "run test" element does not produce any logs, while running (unlike `test-windows`).

There does not seem to be a clear reason or a clear culprit among PRs.

First I suspected it may be #1910, but reverting still shows the same behaviour, see here: https://github.com/alan-turing-institute/sktime/pull/1938

Other potential culprits are #1907, #1926, #1927 which were merged recently, but it seems with all tests passing.

But, the culprit may not be a PR.

Local runs on windows do not show any problem.

@lmmentel who has a unix set-up said he could reproduce the issue locally, earlier today.

Ideas & reproduction attempts at this issue (linux, mac) are appreciated.

k1m190r:
Local testing under Ubuntu 20 WSL suggests the following:

1. Main culprit is some hard to identify interaction between pytest-xdist (with n>1, e.g. n=2), pytest-cov and joblib. What is clear, this is not memory leak, or some other excessive resource usage, it appears to be "deadlock". 

Rest of commentary focused on the:
[`sktime/tests/test_all_estimators.py::TestAllEstimators::test_multiprocessing_idempotent`
](https://github.com/alan-turing-institute/sktime/blob/be65769f1c7e5536d77025ca91d008f5b1a7177b/sktime/tests/test_all_estimators.py#L975)

Which is currently skipped but was un-skipped for testing.

Hang ups stop if:
- `-n 2` is dropped in the invocation of pytest which is equivalent to "no xdist"
- or `--cov=sktime` is dropped so no coverage report
- or -k "not fresh" which is equivalent to dropping all of the test variants of `{TSFreshClassifier TSFreshFeatureExtractor FreshPRINCE} `

3. Observation (htop with python filter) suggests that at some point when tests hit one of the "fresh" as above (with `--cov=sktime` and `-n 2`) all CPU activity stops -> deadlock. 
 - Sometimes after hours (3+h) of waiting it snaps out, but for most tests I didn't wait beyond ~5 min.

5. For now practical change will be to separate
- `test_all_estimators.py::TestAllEstimators::test_multiprocessing_idempotent` out and test it independent of all the other tests with xdist disabled = remove `-n 2`. 
[...]

PR #1949, opened 2022-01-28
DRAFT for diagnosing CI/CD: smaller set of class tests test_all_estimators, instance tests all removed

fkiraly:
An attempt at diagnosing #1941: after removing all tests in `test_all_estimators`, adds back classes individually.

Aim is to conduct a binary search to identify a precise failure condition. I'll determine the number `i` such that adding back `estimator_classes_to_test[0:i]` does not fail but `estimator_classes_to_test[0:i+1]` does

This must exist because the tests run for `estimator_classes_to_test[0:0]` (empty list, #1946) but fail for `estimator_classes_to_test` (situation on main, and #1946 in earlier stage where only instance tests were removed), for a total of 135 estimator classes in `sktime` currently.

### DIFF
--- a/sktime/tests/test_all_estimators.py
+++ b/sktime/tests/test_all_estimators.py
@@ -84,7 +84,7 @@ def pytest_generate_tests(metafunc):
             est for est in ALL_ESTIMATORS if not is_excluded(est)
         ]
         # parameterize test with the list of classes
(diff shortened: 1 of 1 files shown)


fkiraly:
first try: adding back first 30 estimators

PR #1952, opened 2022-01-28
DRAFT for diagnosing CI/CD: removal of tests from `test_all_estimators`

fkiraly:
An attempt at diagnosing #1941.

This PR removes individual tests from `test_all_estimators`:

* `test_multiprocessing_idempotent`, this could lead to freezes due to parallelization/virtualization issues
* `test_methods_have_no_side_effects`, this could lead to an infinite loop in `deep_equals` if an estimator contains an infinitely looping dict of dict where some pointer accident has happened


fkiraly:
this worked.

PR #1958, opened 2022-01-29
Switch fbprophet to prophet

### LINKED ISSUE #1631: [BUG] Use prophet instead of fbprophet in requirements
**Describe the bug**

Posted in :

> I think we need to switch over to the prophet package as opposed to fbprophet. I think the problem is that fbprophet is quite old and they list pystan>=2.14 as a requirement. This causes pystan 3.2.0 to be installed which is actually not compatible with prophet! Prophet only supports pystan 2. prophet has the correct pin of pystan~=2.19.1.1.

You can repro this like follow:
- Download the fbprophet tar ball [here](https://pypi.org/project/fbprophet/#files)
- Open `requirements.txt`

You should see the following:
```
Cython>=0.22
cmdstanpy==0.9.5
pystan>=2.14
numpy>=1.15.4
pandas>=1.0.4
matplotlib>=2.0.0
LunarCalendar>=0.0.9
convertdate>=2.1.2
holidays>=0.10.2
setuptools-git>=1.2
python-dateutil>=2.8.0
tqdm>=4.36.1
```

`pystan>=2.14` will install the latest pystan ([pystan 3](https://pypi.org/project/pystan/)). However, Prophet does not support [pystan 3](https://github.com/facebook/prophet#installation-in-python)!!

I suspect this is making our build more complicated than it needs to be.

lmmentel:
#### Reference Issues/PRs

Fixes #1631 

##### For all contributions

##### For new estimators


TonyBagnall (approved):
should forecastig/fbprophet.py be refactored to just prophey.py?

fkiraly:
>should forecastig/fbprophet.py be refactored to just prophey.py?

you mean, renamed? Why not `metaprophet`?
The problem would be downwards compatibility - `from forecasting.prophet import stuff` would cease to work

lmmentel:
I didn't want to rename user facing imports since that would be a breaking change and a lot of the code that people have would stop working. It can be done but I think it's a bigger job then. This is contributing more towards maintainability.

fkiraly:
> I didn't want to rename user facing imports since that would be a breaking change and a lot of the code that people have would stop working. It can be done but I think it's a bigger job then. This is contributing more towards maintainability.

Yes, exactly my point. Let's not rename modules or imports will break. Renaming public modules must undergo deprecation.

fkiraly (approved):
Looks fine to me, great!

I'd suggest, let's try to merge #1879 first, so we avoid interactions between the probabilistic interface refactor and the change in prophet. Interface refactor is linked to deprecations and has higher priority imo so should be merged first.

aiwalter (approved):
thx @lmmentel

fkiraly:
updated now with the `_predict_interval` refactor in `main` - once runs, let's merge.

PR #1959, opened 2022-01-29
[ENH] split tests in series_as_features into classification and regression 

TonyBagnall:
Splits the tests previously in the "series_as_features" module into specific classification and regression modules. There is some code duplication, this is the first stage, because once we enforce implementation of _fit and _predict with ABC they are more or less redundant (e.g. checking 3d input repeatedly is really just checking the base class functionality). Will in time merge with test_base I think. 

doing this now to conform with the new testing documentation that states that test_all_classifiers.py should exist in classification/tests


fkiraly (changes requested):
Overall, I think this is good!

Although, I do think that `test_3d_numpy_input` and `test_multivariate_input` - which you have now copied twice - is a duplicate of one or two of the `test_fit` in `test_all_estimators`. (e.g., `test_fit_returns_self`)
It should ultimately go there, in my opinion, instead getting duplicated across modules.
Is this not already tested there, by the way?

The `test_output` tests are fine where they are, in my opinion.

TonyBagnall:
> Overall, I think this is good!
> 
> Although, I do think that `test_3d_numpy_input` and `test_multivariate_input` - which you have now copied twice - is a duplicate of one or two of the `test_fit` in `test_all_estimators`. (e.g., `test_fit_returns_self`) It should ultimately go there, in my opinion, instead getting duplicated across modules. Is this not already tested there, by the way?
> 
> The `test_output` tests are fine where they are, in my opinion.

yes I have simply split into two files, in order to make your testing documentation correct. I will review all tests but I dont want to change too much this close to a release

fkiraly:
> yes I have simply split into two files, in order to make your testing documentation correct. I will review all tests but I dont want to change too much this close to a release

Good point - so split now and refactor later.

Could you still kindly check whether there is a duplication? Deleting tests is easy and non-invasite.

TonyBagnall:
> 

I'll happily delete tests ... I'll see if I have time to check tomorrow, feature freeze is monday?

fkiraly:
yes - if you don't find time btw, I'm fine with merging this on the grounds that this is just "moving tests to the right place", and we can deduplicate later

TonyBagnall:
> yes - if you don't find time btw, I'm fine with merging this on the grounds that this is just "moving tests to the right place", and we can deduplicate later

I've just made one change to enforce y is ndarray, but I think it best I leave them here for now and review when I have more time, I'm confident they can go and agree the input checks should be in test_all_estimators, I just dont want to overlook something because I am doing this in the breaks I'm given between minecraft and mariokart :) After, I will remove redundant tests and combine the rest with test_base to have just one file

chrisholder (approved):

chrisholder:
I think it's fine for now as a last min addition to the release we can have a look at this again in the future.

chrisholder (approved):

fkiraly (changes requested):
Ok, I´ve reviewed this.

If it were just moving tests, I'd be totally fine with the change.

However, the change in effect *removes* some tests, as far as I understand: classifiers and regressors are still tested, but the *panel transformers* are removed from the tests.

Currently, this removes important basic input/output checks for half of the transformers, so I don't think we can merge it like this.

Please correct me if I'm wrong about transformers being no longer tested.

fkiraly:
suggestion: perhaps do what is exactly in here, but don't remove the transformer tests, leave them temporarily in `series_as_features`?

TonyBagnall:
> suggestion: perhaps do what is exactly in here, but don't remove the transformer tests, leave them temporarily in `series_as_features`?

if you want, but the only test in there for transformers is test_3d_numpy_input, which is most likely redundant and will be moved anyway.

TonyBagnall:
> > suggestion: perhaps do what is exactly in here, but don't remove the transformer tests, leave them temporarily in `series_as_features`?
> 
> if you want, but the only test in there for transformers is test_3d_numpy_input, which is most likely redundant and will be moved anyway.

but in fact there is no time, and may as well get this in to this release.

fkiraly:
>if you want, but the only test in there for transformers is test_3d_numpy_input, which is most likely redundant and will be moved anyway.

perhaps, but we still want to keep testing it for the time being?

TonyBagnall:
> > if you want, but the only test in there for transformers is test_3d_numpy_input, which is most likely redundant and will be moved anyway.
> 
> perhaps, but we still want to keep testing it for the time being?

and there was you before telling me to remove them! Its back in

fkiraly:
alright, happy with this.

> if you want, but the only test in there for transformers is `test_3d_numpy_input`, which is most likely redundant and will be moved anyway.

I looked over it again, and you are completely right!
I got confused about the removal, your sentence sounded like it was still there, but I've suggested removal in the first place.

Should we remove it then again in a separate PR?

PR #1960, opened 2022-01-29
Fix issues with predict_quantiles in Theta forecaster

kejsitake:
#### Reference Issues/PRs

Attempt to fix https://github.com/alan-turing-institute/sktime/issues/1896

### DIFF
--- a/sktime/forecasting/theta.py
+++ b/sktime/forecasting/theta.py
@@ -218,20 +218,11 @@ class ThetaForecaster(ExponentialSmoothing):
             self.fh.to_relative(self.cutoff) * self.initial_level_ ** 2 + 1
         )
 
-        errors = []
-        for a in alpha:
-            z = _zscore(1 - a)
-            error = z * sem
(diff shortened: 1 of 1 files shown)


fkiraly (commented):
Thanks!
Coincidentlally, I already did the same here: https://github.com/alan-turing-institute/sktime/pull/1879

What this seems to cause is a failure of the `test_theta`, because actuals are no longer within 10% predictive intervals. What seems like an error, may be caused by what was an accidental flip of the coverage (vs 1 minus coverage) *prior* to this. I.e., if you give it 10% coverage, it computes 90% CI; if you give it 90% coverage, it computes 10% CI.

Do you also find this?

fkiraly:
I've merged #1879 now - seems our solutions are only minimally different (e.g., when the data frame is constructed)

fkiraly:
@kejsitake, should we close this? I´ve made similar changes in 0.10.0.
Unless there sth from this which we need to move over?

fkiraly:
hm... what about the increasing variance? I did not fix that

PR #1961, opened 2022-01-29
[ENH] Additions to PR 1879 (probabilistic prediction defaults) arising from aiwalter's review

fkiraly:
This PR adds changes from the review of @aiwalter on PR #1879 after it had been merged:

* the tag `capability:pred_int` description has been updated to mirror the interface refactor (moving from `pred_int` argument in `predict` to dedicated methods `_predict_interval`, `_predict_quantiles`)
* automated test to check whether `capability:pred_int` is correctly set, against whether the methods `_predict_interval`, `_predict_quantiles` have an implementation
* spotted and fixed a bug that was introduced by accepting @lmmentel's suggestion for `_has_implementation_of`. The bug was only revealed by the test (above) suggested by @aiwalter


sktime/registry/_tags.py
```diff
@@ -184,7 +184,7 @@
         "capability:pred_int",
         "forecaster",
         "bool",
-        "is the forecaster capable of returning prediction intervals in predict?",
+        "does the forecaster implement predict_interval and predict_quantiles?",
```
aiwalter, line 187:
```suggestion
        "does the forecaster implement predict_interval or predict_quantiles?",
```

aiwalter (approved):
one minor suggestion above

sktime/registry/_tags.py
fkiraly: it's always "and" - if the one is implemented, the other one also works. So, no change needed?

aiwalter: yes, but not both need to be implemented to require this tag to be `True`?

fkiraly: yes, depending on how you interpret "implement". I guess the "or" version is less confusing, so changed to that.

Issue #1843, opened 2022-01-06
[ENH] rename `fit-in-predict` and `fit-in-transform` to `fit-is-empty`


fkiraly:
I think we should rename `fit-in-predict` and `fit-in-transform` to `fit-is-empty`.

This has two advantages:
* one tag instead of multiple; no future need to introduce new tags for other estimators such as aligners where there is no `predict` or `transform`, or annotators
* the `fit-is-empty` tag is much clearer and unambiguous in my opinion. Because, what does `fit-in-predict` mean? It could mean, for instance, that `predict` also `fit`-s if `fit` was not called before, it's a statement about `predict`, and not clearly a statement about `fit`.

fkiraly:
Any opinions, @aiwalter, @GuzalBulatova, @TonyBagnall?

GuzalBulatova:
I don't have strong opinion regarding this, to me it does make more sense.
`fit-is-empty` could mean though that this estimator doesn't perform/require fit at all, is it always the case?

fkiraly:
>fit-is-empty could mean though that this estimator doesn't perform/require fit at all, is it always the case?

Currently we require `fit` to be run before `transform` even if `fit` is empty, @mloning was adamant on that.
I understand how it makes sense, since it results in consistent run conditions and behaviour.

The function to call when you want to transform the data in one go is `fit_transform`, that's more python zen with "one and only one way to do things".

PR #1962, opened 2022-01-29
[ENH] tag aliasing mixin for aliasing and deprecation of tags

fkiraly:
This PR introduces a `TagAliaserMixin` with the purpose of allowing aliasing of tags and easy deprecation, in cases where we want to replace one tag by another and keep the old one pointing to the new one for a deprecation period, and show a warning.

Example applications would be:
* https://github.com/alan-turing-institute/sktime/issues/1843
* renaming `capability:pred_int` to `capability:pred_interval_and_quantiles`, from @aiwalter's review of https://github.com/alan-turing-institute/sktime/pull/1879.

For discussion, currently.

As a proof-of-concept, this carries out #1843 partially by adding those tags and using the deprecation mixin to alias.


fkiraly:
note: this currently goes only one direction and many tests are failing: i.e., old tags are translated to new tags.
But this does not translate new tags to old tags yet.

TonyBagnall (approved):
I think I follow and if we need to do this to tidy up the tags, it is most welcome. A few non blocking comments
1. afaik we are moving away from fit-is-empty towards fit_is_empty, it would be good if your example enforced that
2. For comments in tags, e.g. "does fit contain no logic and can be skipped? yes/no - deprecated for 0.12.0", this is a boolean variable, so the values it can take should be True/False not yes/no, which (to me at least implies) it is a string. I would write the comment  "fit contains no logic and can be skipped", True/False
3. TagAliaserMixin is public, should it not be declared in __all__ or be private? Or does the fact that it is static and not for instantiation mean that is not required? 
4. A short "how to deprecate" in the doc string of TagAliaserMixin would be helpful. Is aliasing essential, or can you just deprecate an unused tag?

fkiraly:
>afaik we are moving away from `fit-is-empty` towards `fit_is_empty`, it would be good if your example enforced that

Yes, you are right, @TonyBagnall - renamed to `fit_is_empty`

> I would write the comment "fit contains no logic and can be skipped", True/False

Yes, changed it to make clearer - is this better?

> `TagAliaserMixin` is public, should it not be declared in all or be private? Or does the fact that it is static and not for instantiation mean that is not required?

It's a mixin, and anyway in the base module so I wouldn't worry about that?

> A short "how to deprecate" in the doc string of TagAliaserMixin would be helpful.

Yes, valid point. I've added a docstring and a section in the "deprecation" developer docs.

> Is aliasing essential, or can you just deprecate an unused tag?

As long as it has been in use, it needs to be deprecated. Aliasing is necessary only for renaming. For removal, a warning still needs to be raised during a deprecation period.

Issue #1850, opened 2022-01-07
[ENH] `statsmodels` interfaces may be missing parameters - investigate and/or add them


fkiraly:
In issue #1845, an instance of a `statsmodels` interfacing estimator was discovered which was missing crucial parameters.

The reason was that in `statsmodels`, model parameteres are spread out across constructor (`__init__`), `fit`, and potentially other functions.

Due to this, it would be important to look at other `statsmodels` interfaces to check whether we are missing useful parameters from non-constructor methods.

If we are, a notice here would already be useful. Further, these should be added to the constructor (not `fit` etc) of the corresponding `sktime` interface. A "blueprint" PR to carry out this addition of parameters is #1849 (for Holt-Winters smoothing).

biologioholic:
I would like to handle this issue

aiwalter:
@biologioholic its yours

biologioholic:
> 

Thanks

fkiraly:
@biologioholic, great!
Btw, if you're looking into statsmodels, you could also look at interfacing DynamicFactor, if you want, issue here:
https://github.com/alan-turing-institute/sktime/issues/2059

fkiraly:
Ah, just noticed that somebody has already picked that up.

fkiraly:
@biologioholic, sorry to as, I might be confused.
Where exactly is your draft PR? I haven't seen one.

biologioholic:
> @biologioholic, sorry for this, I might be confused. Where exactly is your draft PR? I haven't seen one. Could you kindly post a link?

Sorry for the confusion. I have closed that draft since it only has ETS. This PR includes other model as well. 

https://github.com/biologioholic/sktime/pull/3#issue-1177143455

fkiraly:
Investigated by @biologioholic - with #2304, all currently interfaced statsmodels models have the full parameter set.

Details from #2304, citing @biologioholic there:

There is no more parameters missing for statsmodel interface.  I realize it's been a lengthy first PR. Really appreciate the patience and input of everyone involved (Next time I'll make a draft PR first 😄). Below is a summary of what happened:

Five statsmodel-based files were reviewed/worked on. These are:

1. var.py - added `freq`, `dates` and `ic` parameters. Added default values for other parameters already in place.
2. trend.py - no change, no missing parameters.
3. theta.py - no change, no missing parameters.
[...]

PR #1963, opened 2022-01-30
[ENH] Added STLForecaster

### LINKED ISSUE #782: Implement STL forecaster

aiwalter:
#### Reference Issues/PRs
closes #782

##### For all contributions

##### For new estimators


aiwalter:
@danbartl : fyi I am working here on `STLForecaster`

aiwalter:
moved `STLForecaster` into `trend.py` as in `compose._stl.py` its causing a cyclic import issue for `NaiveForecaster`, I am not sure why

aiwalter:
note to self: **do not merge before #1990 is not fixed.** Also think about if the circular import issue could be resolved so `STLForecaster` could be imported from `compose`.

fkiraly:
> Also think about if the circular import issue could be resolved so `STLForecaster` could be imported from compose.

Why would `STLForecaster` be imported from `compose`, or where?

aiwalter:
Bcz its a composition of three forecasters

sktime/forecasting/trend.py
```diff
@@ -212,3 +214,280 @@ def _predict(self, fh=None, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
...
+        self.seasonal_ = pd.Series(self.stl_.seasonal, index=y.index)
+        self.resid_ = pd.Series(self.stl_.resid, index=y.index)
+        self.trend_ = pd.Series(self.stl_.trend, index=y.index)
+
+        self.forecaster_seasonal_.update(
+            y=self.seasonal_, X=X, update_params=update_params
+        )
+        self.forecaster_trend_.update(y=self.trend_, X=X, update_params=update_params)
+        self.forecaster_resid_.update(y=self.resid_, X=X, update_params=update_params)
+        return self
```
ltsaprounis, line 493: Isn't `_update` optional for forecasting?

sktime/forecasting/trend.py
```diff
@@ -212,3 +214,280 @@ def _predict(self, fh=None, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
...
+        STL, by default None. If None, then
+        a NaiveForecaster(strategy="last") is used.
+    forecaster_resid : sktime forecaster, optional
+        Forecaster to be fitted on resid_ component of the
+        STL, by default None. If None, then
+        a NaiveForecaster(strategy="mean") is used.
+    sp : int, optional
+        Seasonal period for defaulting forecasters and/or STL.period param,
+        by default None. Can only be used if at least on the forecasters
+        or the stl is None.
```
ltsaprounis, line 246: Should probably be something like: 
```
Length of the seasonal period for STL, by default 2. 
It's also the default seasonal period for the forecasters 
(forecaster_trend, forecaster_seasonal, forecaster_resid) that are None.
```

sktime/forecasting/trend.py
```diff
@@ -212,3 +214,280 @@ def _predict(self, fh=None, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
...
+    period : {int, None}, optional
+        Periodicity of the sequence. If None and endog is a pandas Series or
+        DataFrame, attempts to determine from endog. If endog is a ndarray,
+        period must be provided.
```
ltsaprounis, line 250: I think this needs to be removed. There is no period in `__init__`.

sktime/forecasting/trend.py
aiwalter: there is an issue with default `_update`, it was not working correctly. We need to refactor that anyways. What I do here is excactly refitting again, the `STL` as well as forecasters, its the same behavior as `_fit()`, unless I made a mistake?

aiwalter:
@fkiraly @ltsaprounis pls review/approve

sktime/forecasting/trend.py
ltsaprounis: Ok - in that case replace `update` in lines 488-492 with `fit`. This way it will 100% refit. Now it's refitting STL and updating the season, trend, residuals forecasters instead of refitting these too.

fkiraly (changes requested):
Looks good!

I think, currently, the `STLForecaster` is on the `EXCLUDE_ESTIMATORS` list in `tests._config`?
Can you kindly remove it and run the tests? I think they are not currently being run?

sktime/forecasting/trend.py
aiwalter: I think its fine to `update` the forecasters if as they might have implemented a custom `_update`. The `STL` does not have an `update` so we need to refit it. Actually I think how I have implemented is even better than using the default `_update` from base class (if it would work 😆 )

sktime/forecasting/trend.py
```diff
@@ -212,3 +214,272 @@ def _predict(self, fh=None, X=None, return_pred_int=False, alpha=DEFAULT_ALPHA):
         X_pred = fh.to_numpy().reshape(-1, 1)
         y_pred = self.regressor_.predict(X_pred)
         return pd.Series(y_pred, index=self.fh.to_absolute(self.cutoff))
+
+
+class STLForecaster(BaseForecaster):
+    """Implements STLForecaster based on statsmodels.tsa.seasonal.STL implementation.
```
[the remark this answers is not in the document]
fkiraly: so, sth like "`sktime` interface to `statsmodels.tsa.seasonal.STL`"

fkiraly (changes requested):
Great! Nice addition, tests pass.

May I request one change: generally, in `statsmodels`, some hyper-parameters may be arguments to `fit`, while others are passed in the constructor.
In `sktime`, these should all be passed in the constructor.

See #1850 for a discussion and an earlier instance.

In this specific case, I think there are parameters in `fit` too that the `STLForecaster` should expose - kindly add these (or argue why we shouldn't).

sktime/forecasting/trend.py
ltsaprounis: It might be fine, but I'm not sure. When you refit STL you generate 3 new time series for the same time index as before. So if you then `update` instead of `fit`, you're not really updating in the sense that you got a few more time steps in your time series. You're updating your estimator with a completely new time series, which is why it makes more sense to refit instead (at least to me). What do you think?

fkiraly (approved):
Looks good now!

aiwalter:
@ltsaprounis are you ok with this now? So my update function is basically doing the same as refit, but it will use `_update` of the forecasters if its implemented. I think that is a reasonable behavior.

ltsaprounis:
> @ltsaprounis are you ok with this now? So my update function is basically doing the same as refit, but it will use `_update` of the forecasters if its implemented. I think that is a reasonable behavior.

Yep, it's a small detail anyway. There's no right or wrong.

ltsaprounis (approved):

PR #1968, opened 2022-01-31
Generalize CutoffSplitter to accept non-integer cutoffs and forecasting horizons

khrapovs:
#### Reference Issues/PRs

Next step in #1737.

The main change is to allow datetime `cutoffs` and timedelta `fh` in `CutoffSplitter`. No regressions were introduced. More test cases are added in `sktime.forecasting.model_selection.tests.test_split.test_cutoff_window_splitter`.

More detailed description of changes:
- Reused type checking functions from `sktime/utils/validation/__init__.py` in `sktime.forecasting.base._fh._check_values`.
- Added `pd.TimedeltaIndex` to `RELATIVE_TYPES` in `sktime/forecasting/base/_fh.py` to allow timdelta forecasting horizons.
- Added `pd.TimedeltaIndex` to `VALID_INDEX_TYPES` in `sktime/utils/validation/series.py` to allow timdelta forecasting horizons.
- Added `TEST_CUTOFFS` to `sktime/forecasting/tests/_config.py` so that all splitter test configurations are collected in one place.
- Added `TEST_FHS_TIMEDELTA` to `sktime/forecasting/tests/_config.py`. These are timedelta forecasting horizons used in test cases for `CutoffSplitter`.
- Made `sktime.utils.validation.tests.test_forecasting.test_check_fh_empty_input` more precise by adding `match` argument to `raises`.
- Added a few new type checking functions in `sktime/utils/validation/__init__.py` which are used in splitters and corresponding tests.
- Generalized `sktime.utils.validation.forecasting.check_cutoffs` to allow datetime cutoffs.

None

##### For all contributions

##### For new estimators


sktime/forecasting/base/_fh.py
```diff
@@ -158,7 +170,7 @@ def __new__(
     def __init__(
         self,
         values: Union[VALID_FORECASTING_HORIZON_TYPES] = None,
-        is_relative: bool = True,
+        is_relative: Optional[bool] = True,
     ):
         if is_relative is not None and not isinstance(is_relative, bool):
```
fkiraly, line 175: is the case of `is_relative=None` correctly treated below?

sktime/forecasting/model_selection/_split.py
```diff
@@ -29,10 +29,17 @@
 from sktime.utils.validation import (
     ACCEPTED_WINDOW_LENGTH_TYPES,
     NON_FLOAT_WINDOW_LENGTH_TYPES,
+    array_is_datetime64,
```
fkiraly, line 32: due to repetition - would it make sense to refactor these into a single function, with `datetime64` etc being parameters? Or would that be ugly?

sktime/forecasting/model_selection/_split.py
```diff
@@ -222,6 +229,67 @@ def _check_window_lengths(
                 raise ValueError(error_msg_for_incompatible_types)
 
 
+def _cutoffs_fh_window_length_types_are_supported(
```
fkiraly, line 232: please write a docstring - I know these are private, but for sth like this it is very much needed.
For review, future contributors, or just a future version of yourself...

sktime/forecasting/model_selection/_split.py
```diff
@@ -388,39 +457,59 @@ class CutoffSplitter(BaseSplitter):
     Parameters
     ----------
     cutoffs : np.array or pd.Index
-        cutoff points, positive and integer-index like, usable with pandas
```
fkiraly, line 391: I think it is crucial to describe how the splits are defined, for a given cutoff. Be precise in the docstring.

For instance, is the test set comprised of all data with bigger-equal, or strictly-bigger-equal time index?

fkiraly: another detail is if you pass a `pandas.Series` with integer indices, but the indices are not `RangeIndex(len(y))`. On can imagine two behaviours here, `loc` and `iloc` based. Which one applies? What if the user wants the other one? Would adding a flag with a default make sense (to constructor, perhaps)?

sktime/utils/validation/series.py
```diff
@@ -18,7 +18,13 @@
 
 # We currently support the following types for input data and time index types.
 VALID_DATA_TYPES = (pd.DataFrame, pd.Series, np.ndarray)
-VALID_INDEX_TYPES = (pd.Int64Index, pd.RangeIndex, pd.PeriodIndex, pd.DatetimeIndex)
+VALID_INDEX_TYPES = (
+    pd.Int64Index,
+    pd.RangeIndex,
+    pd.PeriodIndex,
+    pd.DatetimeIndex,
+    pd.TimedeltaIndex,
```
fkiraly, line 26: why do we need to add this here? Should the series index not always be absolute?
Only in the forecasting horizon we want to allow relative ones.

Is there something I`m missing?

sktime/utils/validation/tests/test_forecasting.py
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3 -u
 # -*- coding: utf-8 -*-
 # copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+"""Test forecasting module."""
 
 __author__ = ["Markus Löning"]
```
fkiraly, line 6: while you´re here, kindly change this to GitHub Id (`mloning`)

sktime/utils/validation/forecasting.py
```diff
@@ -25,9 +25,18 @@
 from sklearn.base import clone, is_regressor
 from sklearn.ensemble import GradientBoostingRegressor
 
-from sktime.utils.validation import is_date_offset, is_int, is_timedelta
+from sktime.utils.validation import (
```
fkiraly, line 28: kindly add yourself as an author to files where you make substantial changes

fkiraly (changes requested):
Really nice work!
I suspect the next change will be extending forecasters and forecaster base classes?

Also, looks like a generalization that is conceptually straightforward, but non-trivial coding-wise, yet really well carried out!

Now, if this were somewhere on the periphery of `sktime`, I'd approve (up to usual comments on "please write docstrings" etc).

The `ForecastingHorizon`, however, is "in the heart" of ` sktime` forecasting.

Hence, I think we ought to have examples and/or a mini-tutorial on the new functionality, plus clarification on what precisely the behaviour is, e.g., when is it `loc` and when is it `iloc` referential.

I would suggest a mini-notebook `01b_forecasting_ForecastingHorizon` which explains:
* the basics, using integer/iloc indexing
* including common uses: in forecasting, and in splitters or updates
* the basics, using datetime plus period indexing; integer references in this case
* relative vs absolute; loc vs iloc references (when? how?)
* the two ways of relative indexing for datetime indices (integer and timedelta)

(first couple bullet points can mostly be copied from forecasting notebook)

This notebook would serve as sth between a unit test and an integration test, as well as an extended example for reviewers and developers. It would also serve as a usage design document (which lets us avoid having to draw up a STEP, that's usually necessary for "heartland changes").

What do you think?

sktime/forecasting/base/_fh.py
```diff
@@ -83,22 +90,27 @@ def _check_values(values: Union[VALID_FORECASTING_HORIZON_TYPES]) -> pd.Index:
         Sorted and validated forecasting horizon values.
     """
     # if values are one of the supported pandas index types, we don't have
-    # to do
-    # anything as the forecasting horizon directly wraps the index, note that
-    # isinstance() does not work here, because index types inherit from each
-    # other,
+    # to do anything as the forecasting horizon directly wraps the index,
```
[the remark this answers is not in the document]
khrapovs: no idea. I just realigned the sentence over different lines. not a single word changed

sktime/forecasting/base/_fh.py
khrapovs: sorry, I don't know. I did not touch this part of the logic. is it relevant for my PR?

sktime/forecasting/model_selection/_split.py
khrapovs: well, these type checkers are due to be extended with more and more supported types. for starters I added only `datetime64`, but later it could be also `pd.Timestamp`. so, I would prefer to keep this list hardcoded somewhere and a function that checks an argument belonging to this list.

sktime/forecasting/model_selection/_split.py
khrapovs: added docstrings to all `_check...` functions in `_split.py`

sktime/forecasting/model_selection/_split.py
khrapovs: added the following sentence:
```
Training window's last point is equal to the cutoff,
while test window starts from the next observation in `y`.
```

khrapovs: > another detail is if you pass a `pandas.Series` with integer indices, but the indices are not `RangeIndex(len(y))`. On can imagine two behaviours here, `loc` and `iloc` based. Which one applies? What if the user wants the other one? Would adding a flag with a default make sense (to constructor, perhaps)?

Now I really find this sentence confusing. I have removed "usable with pandas `.iloc[]` or `.loc[]` indexing, respectively" everywhere since neither `loc` nor `iloc` are actually used in splitters to construct windows. A user implicitly specifies what indexing he/she wants by choosing the type of e.g. `cutoffs`. It is our task to figure out how to treat user's choice. Some choices are clearly forbidden (`TypeError`).

sktime/utils/validation/series.py
khrapovs: I added `pd.TimedeltaIndex` to `RELATIVE_TYPES` in `_fh.py`. Otherwise, I hit this exception with my test cases:
```
raise TypeError(f"{type(values)} is not a supported fh index type")
```
If I do not add the same index type to `VALID_INDEX_TYPES` in `series.py`, then I hit this roadblock:
```
assert set(RELATIVE_TYPES).issubset(VALID_INDEX_TYPES)
```
Any suggestions to circumvent this?

sktime/utils/validation/tests/test_forecasting.py
khrapovs: changed

sktime/utils/validation/forecasting.py
khrapovs: added

khrapovs:
@fkiraly Thanks a lot for a very detailed and constructive review!

> I suspect the next change will be extending forecasters and forecaster base classes?

Still not. When I started working on this PR I did not plan to touch `ForecastingHorizon` at all, but I very quickly hit the wall of adding a datetime cutoff and an integer forecasting horizon. Hence, the first unavoidable step generalizing also `ForecastingHorizon` is made. My next step is to generalize all other splitters to accept timdelta `ForecastingHorizon`.

Regarding the notebook. That's a perfectly valid and natural request, but I would insist on postponing advertising this to the users. The main reason is that I have no idea if timedelta horizon will work in forecasting! Right now this extension is only tested for one single splitter.

fkiraly:
> Regarding the notebook. That's a perfectly valid and natural request, but I would insist on postponing advertising this to the users. 

yes, valid point - putting this in the notebook folder would advertise it before it's ready to work with forecasters etc.

But, I think we still need to have examples to test.
One way would be to write "workflow" style tests which do the same, or examples in the docstrings.

khrapovs:
@fkiraly I have added a few tests in `sktime/forecasting/base/tests/test_fh.py` that partially mirror existing tests, but with timedelta horizons. I have tried more test cases, which fail. Fixing those will inevitably lead to a very substantial refactoring of `ForecastingHorizon`. I am very reluctant to do that in this PR...

fkiraly:
> I have tried more test cases, which fail. Fixing those will inevitably lead to a very substantial refactoring of ForecastingHorizon. I am very reluctant to do that in this PR...

Well, that's precisely what I was worried about!
We can't do a partial upgrade if we haven't covered the corners.

Can you perhaps explain which test cases are failing? I.e., for me to judge whether they are related to the upgrade, or to pre-existing issues with `ForecastingHorizon`.

fkiraly:
I also feel this is complex enough for a tech meeting on the topic. Would you have time for one of the dev meetings, Fr 5pm UTC?

fkiraly:
@khrapovs, thanks for this great PR!

> I have added a few tests in sktime/forecasting/base/tests/test_fh.py that partially mirror existing tests, but with timedelta horizons. I have tried more test cases, which fail. Fixing those will inevitably lead to a very substantial refactoring of ForecastingHorizon. I am very reluctant to do that in this PR...

Can you quickly give an overview of the test cases you added?

And can you give an overview of those that fail?

khrapovs:
@fkiraly 

> Can you quickly give an overview of the test cases you added?
> 
> And can you give an overview of those that fail?

Let's come back to this PR once https://github.com/alan-turing-institute/sktime/pull/1981 is merged. In this bug fix i have added a lot more tests and some of them were failing in this PR. So I hope the difference between this branch and master will become smaller after the bug fix merge.

khrapovs:
@fkiraly 

> Can you quickly give an overview of the test cases you added?

- New `test_check_fh_relative_values_input_conversion_to_pandas_index` simply mirrors old `test_check_fh_absolute_values_input_conversion_to_pandas_index` for the new kind of forecasting horizon, `pd.TimedeltaIndex`.
- `FREQUENCY_STRINGS` are broken down into `FIXED_FREQUENCY_STRINGS` (it is possible to `pd.timedelta_range(1, freq=freq, periods=2)`) and `NON_FIXED_FREQUENCY_STRINGS` (`timedelta_range` throws `ValueError: invalid unit abbreviation`).
- New `test_absolute_to_absolute_with_timedelta_horizon` mirrors `test_absolute_to_absolute_with_integer_horizon` for timedelta forecasting horizon.
- New `test_relative_to_relative_with_timedelta_horizon` mirrors `test_relative_to_relative_with_integer_horizon` for timedelta horizon.

> And can you give an overview of those that fail?

New tests mentioned above if I use all frequencies `FREQUENCY_STRINGS`, instead of only `FIXED_FREQUENCY_STRINGS`. This makes some sense since `pd.Period + pd.Timedelta` with those frequencies fails.

sktime/forecasting/base/_fh.py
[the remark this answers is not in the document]
fkiraly: I don't think you did - you deleted part of the sentence?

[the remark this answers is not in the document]
khrapovs: let's make it simpler. I have reverted this comment to its original state

fkiraly (dismissed):
Great work!

After looking through the code and the tests, and playing around a bit, I'm 90% sure that this is solid.
Generally, this is accompanied by clearer code structure and docstrings.

I'd like a second review though, perhaps by someone else who has recently looked into forecasting horizons.

In the meanwhile, one minor point: in the docstrings, there are compatibility conditions between various arguments, e.g., "if `fh` is this, then `y` must have this index" or similar. Not a blocker for merging, but might be useful to add clarification on, before we forget this.

sktime/forecasting/model_selection/_split.py
```diff
@@ -214,12 +252,143 @@ def _check_window_lengths(
...
+    either all inputs are integers, or they are all datetime or timedelta
+
+    Parameters
+    ----------
+    cutoffs : np.array or pd.Index
+        cutoff points, positive and integer- or datetime-index like
+    fh : int, timedelta, list or np.array of ints or timedeltas
+    y : pd.Series, pd.DataFrame, np.ndarray, or pd.Index
+        coerced and checked version of input y
+
```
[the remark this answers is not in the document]
khrapovs: I have edited this docstring. Please, take a look. I actually don't care, at least at this stage, what index does `y` have. For validation depending on the other types I either take the length of `y` or it's last value.

khrapovs:
> PS: can you kindly report the unrelated, probably sporadic failure from the win 3.7 test?

https://github.com/alan-turing-institute/sktime/issues/2066

khrapovs:
@fkiraly 

> In the meanwhile, one minor point: in the docstrings, there are compatibility conditions between various arguments, e.g., "if `fh` is this, then `y` must have this index" or similar. Not a blocker for merging, but might be useful to add clarification on, before we forget this.

Added a couple of lines under parameter description in `CutoffSplitter` docstring and in `_check_cutoffs_fh_y`.

sktime/utils/validation/series.py
fkiraly: no, this makes sense. This includes both absolute and relative, my mistake.

aiwalter:
I have reviewed the `fh` stuff, I cant see anything now that requires changes. @fkiraly @khrapovs

fkiraly:
alright! let´s go when the tests pass!

PR #1970, opened 2022-01-31
[ENH] more testing parameters for `NaiveForecaster`

fkiraly:
This PR adds more testing parameters to `NaiveForecaster`, which now should cover all significant parameter cases.

This follows a report of @aiwalter that `NaiveForecaster` does not work with the parameter setting `sp=2`, with a failure in `_reshape_last_window_for_sp`.

### DIFF
--- a/sktime/forecasting/naive.py
+++ b/sktime/forecasting/naive.py
@@ -379,7 +379,12 @@ class NaiveForecaster(BaseForecaster):
             `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
             `create_test_instance` uses the first (or only) dictionary in `params`
         """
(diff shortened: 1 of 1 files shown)


fkiraly:
@aiwalter, all seems to work, at leaest with the usual tests?
Which parameter settings were you using? Can you post minimally reproducing code?

aiwalter (approved):
@fkiraly thx for looking into it, now I also know how to define a grid in `get_test_params` :) 
We can get this in already, I need to look closer how `STLForecaster` was able to create this exception

fkiraly:
it's not a grid - it's just a collection of five settings.
If you want a grid, you have to create it manually within `get_test_params`.

Issue #1976, opened 2022-02-02
[BUG] bug if `transform` is called twice with a lossy output conversion


fkiraly:
If `transform` is called twice, and the output conversion is lossy, the second `transform` call may error out under certain circumstances.

The reason is that the output `convert_to` of the first `transform` may write to the converter store (`self._converter_store_X`) and the first `convert_to` of the second `transform` may read from it.

This behaviour is erroneous, since the second `transform` call's first `convert_to` may not be applied to a data container of the same structure as the first `transform` call's second `convert_to`.

This may be quite rare, but will always happen for a transformer that creates columns, that internally uses `pd.DataFrame`, and which is being passed a `pd.Series` - such as the case in the prototype of #1924 which unearthed this bug.

FYI @danbartl.

PR #1977, opened 2022-02-02
[ENH] Convert store reset/freeze behaviour & fix of bug 1976

fkiraly:
This PR introduces an argument `store_behaviour` to `convert` and `convert_to` which allows the user to determine the nature of side effects (or their absence) on `store`: 
```
    store_behaviour : str, optional, one of None (default), "reset", "freeze", "update"
        "reset" - store is emptied and then updated from conversion
        "freeze" - store is read-only, may be read/used by conversion but not changed
        "update" - store is updated from conversion and retains previous contents
        None - automatic: if store is empty and not None, "update"; otherwise, "freeze"
```

This is useful as follows: if there is a pair of `convert_to` intended to be "in" and "out" conversion, the first should be "reset" or "update", the second should be "freeze". The default (`None`) behaviour tries to guess the right setting automatically.

This should prevent the bug #1976 which occurs under quite specific conditions, when methods with in/out conversions are executed multiple times.

For good measure, the `convert_to` calls in `BaseForecaster` and `BaseTransformer` have also been hard-coded with their expected `store` side effect behaviour, to avoid further bugs along the lines of #1976.


fkiraly:
FYI @danbartl

sktime/datatypes/_convert.py
```diff
@@ -115,6 +132,16 @@ def convert(obj, from_type: str, to_type: str, as_scitype: str = None, store=Non
         as_scitype = mtype_to_scitype(to_type)
     elif not isinstance(as_scitype, str):
         raise TypeError("as_scitype must be str or None")
+    if store is not None and not isinstance(store, dict):
+        raise TypeError("store must be a dict or None")
+    if store_behaviour is not None and not isinstance(store_behaviour, str):
+        raise TypeError("store_behaviour must be a str or None")
+    elif store_behaviour not in [None, "reset", "freeze", "update"]:
```
aiwalter, line 139: why is this `elif`? Can this not just be `if`?

fkiraly: I guess it could be, does it make a difference?
The first two lines could even be removed.
Although the current way is perhaps more formulaic: first check type, then continuing from that check values, so perhaps easier to read.

fkiraly: I simplified the input check, have a look

aiwalter (approved):

PR #1978, opened 2022-02-03
[BUG] fix import on clustering extension template

schettino72:
Fix broken import. Module was rename from `base.py` to `_base.py`.

##### For all contributions

### DIFF
--- a/extension_templates/clustering.py
+++ b/extension_templates/clustering.py
@@ -31,7 +31,7 @@ copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
 """
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


fkiraly (approved):
Yes, that is currently correct.

`base` should still work though - I think this may be a mistake, since there was no depreation.

fkiraly:
I opened a bug report here #1979 to check whether this is a mistake (I suspect it is) or whether this is intended.

TonyBagnall:
hi, yes it is a bug, chris took base private and uses the init model, thanks for the spot @schettino72

PR #1975, opened 2022-02-02
[BUG] Fixing forecast horizon index bugs

eenticott-shell:
#### Reference Issues/PRs
Fixes #1935. Fixes #1435. Fixes #1876. 

#### What does this implement/fix? Explain your changes.
Within the forecasting horizon methods `to_relative` and `to_absolute_int` coerces absolute and cutoff to the same type to avoid an error that can occur in `pandas` where time differences are multiplied by the frequency multiple (i.e. 2H multiplies the difference by 2). This caused numerous index errors.

 #### Does your contribution introduce a new dependency? If yes, which one?
No

#### What should a reviewer concentrate their feedback on?
Whether forcing the type of the input to change could have unintended consequences.

PR #1981, opened 2022-02-03
[BUG] Fix a bug in PeriodIndex arithmetic

khrapovs:
#### Reference Issues/PRs

Fixes https://github.com/alan-turing-institute/sktime/issues/805. Fixes https://github.com/alan-turing-institute/sktime/issues/1935. Fixes https://github.com/alan-turing-institute/sktime/issues/1435. Fixes https://github.com/alan-turing-institute/sktime/issues/1876.

Convert `PeriodIndex` `absolute` to `DatetimeIndex`, as well as `cutoff` and `start` `Period` to `Timestamp` before taking the difference between the two.

This circumvents the following bug in pandas:
```
periods = pd.period_range(start="2021-01-01", periods=3, freq="2H")
periods - periods[0]
Out: Index([<0 * Hours>, <4 * Hours>, <8 * Hours>], dtype = 'object')
[v - periods[0] for v in periods]
Out: Index([<0 * Hours>, <2 * Hours>, <4 * Hours>], dtype='object')
```

Bug report scripts were added as explicit test cases and should be replaced later with lightweight unit tests.

No.

##### For all contributions


fkiraly:
I think @eenticott-shell was already trying to fix this here: #1975 - assuming it is the same bug?
Perhaps it would be a good thing to check whether it's the same root cause, and, if yes, what broke in #1975 as @eenticott-shell tried to fix it.

fkiraly:
uh-oh... are these the same failures as in #1975?

eenticott-shell:
Thanks for continuing this, it looks like the tests are passing now.

sktime/forecasting/base/_fh.py
```diff
@@ -273,23 +273,23 @@ def to_relative(self, cutoff=None):
...
 
-            if isinstance(absolute, pd.DatetimeIndex):
-                # We cannot use the freq from the the ForecastingHorizon itself (or its
-                # wrapped pd.DatetimeIndex) because it may be none for non-regular
-                # indices, so instead we use the freq of cutoff.
-                freq = _get_freq(cutoff)
+            # We cannot use the freq from the ForecastingHorizon itself (or its
+            # wrapped pd.DatetimeIndex) because it may be none for non-regular
+            # indices, so instead we use the freq of cutoff.
+            freq = _get_freq(cutoff)
```
fkiraly, line 279: does cutoff always have a frequency?

sktime/forecasting/base/_fh.py
```diff
@@ -273,23 +273,23 @@ def to_relative(self, cutoff=None):
...
 
+            if isinstance(absolute, pd.DatetimeIndex):
                 # coerce to pd.Period for reliable arithmetics and computations of
                 # time deltas
                 absolute = _coerce_to_period(absolute, freq)
                 cutoff = _coerce_to_period(cutoff, freq)
 
             # Compute relative values
-            relative = absolute - cutoff
+            relative = pd.Index([date - cutoff for date in absolute])
```
fkiraly, line 288: can we avoid this back/forth conversion?

khrapovs:
@fkiraly @eenticott-shell Finally, all tests pass! I have explicitly added all the bug report examples mentioned in the PR description as test cases. Obviously, at this state this PR is impossible to merge. It is too dirty and looks like a hack. I still can not comprehend the full scale of what is going on with those endless conversions in forecasting horizon. So, I will try to clean it up a bit. Any help is of course highly welcome!

sktime/forecasting/base/_fh.py
khrapovs: I wonder too. There are not enough tests to say for sure. Not all public methods are tested explicitly, but only implicitly through forecasters and transformers.

sktime/forecasting/base/_fh.py
khrapovs: this specific line is exactly the hack we need to circumvent the bug in pandas with difference between `PeriodIndex` and `Period` objects. converting the list to index is necessary as further down the road it is only index, but not a list is expected

sktime/forecasting/base/_fh.py
fkiraly: hm... somewhat troubling.
We may need an explicit design document on the different values that `cutoff` can (and is expected to) have; as well as behaviour of `ForecastingHorizon` around that. There is clearly an interaction with the `BaseForecaster`.

fkiraly:
> Finally, all tests pass! I have explicitly added all the bug report examples mentioned in the PR description as test cases. Obviously, at this state this PR is impossible to merge. It is too dirty and looks like a hack. I still can not comprehend the full scale of what is going on with those endless conversions in forecasting horizon. So, I will try to clean it up a bit. Any help is of course highly welcome!

I would not mind merging this and cleaning up in a separate PR.
I've looked through it and think the flow and logic is preserved, it's not as bad as a hack as you may think.

Of course the flow and logic, and structure, is not perfect, and we should definitely revisit that, and architecture/design, when thinking about time stamp based horizons.

khrapovs:
@fkiraly Ok, let's merge it then. I have cleaned up a bit the code, added more comments about pandas bug, a few TODO lines in the test module. I will come back to this either in https://github.com/alan-turing-institute/sktime/pull/1968, or separately.

sktime/forecasting/base/tests/test_fh.py
```diff
@@ -264,3 +308,113 @@ def test_estimator_fh(freqstr):
...
+    forecaster.fit(y, fh=[1, 2, 3, 4, 5, 6])
+    y_pred = forecaster.predict()
+    pd.testing.assert_index_equal(
+        y_pred.index, pd.period_range("2019-01-02 00:00", periods=6, freq=freq)
+    )
+
+
+# TODO: Replace this long running test with fast unit test
+def test_theta_forecaster():
+    """Test example in notebook.
```
fkiraly, line 360: the notebooks are already all tested as part of the CI/CD, i.e., that they all run without exception.

fkiraly (approved):
I think this is an improvement in cleaning up the code, and also a bugfix for the various bugs linked.

Since you worked on it too, @eenticott-shell, could you kindly have another look before we merge this?

sktime/forecasting/base/tests/test_fh.py
khrapovs: I have added this test only for my convenience. But now that everything is fixed, I have removed it.

eenticott-shell:
Thank you for your work on this @khrapovs. The fix is much cleaner and tidier now. Given that is now passes all the tests hopefully we can be safe to assume that the coercing hasn't caused any other issues.

fkiraly (approved):
Looks good!
Let's just check whether all the tests pass with an update from `main`.

fkiraly:
@eenticott-shell, @khrapovs, quick Q: did any of you report the `pandas` bug to `pandas` yet?
Let me know if not and/or whether you want to do it.

PR #1987, opened 2022-02-05
[ENH] test framework: `pytest` base class for module tests

fkiraly:

---

This experimental PR continues carrying out the test refactor, and turns `test_all_estimators` into multiple test classes.

The expected benefits are:
* easy use of tests in sister packages, extension packages, packages adopting the `sktime`interface, or custom local use cases
* "run tests" one-method call to run all tests for one estimator

In this PR, the pre-PR logic in `test_all_estimators` is moved into two classes:
* `BaseFixtureGenerator`, which takes responsibility for fixture generation
* `TestAllEstimators`, which contains all the tests, and inherits from `BaseFixtureGenerator`

`pytest` automatically discovers the tests in `TestAllEstimators` with fixtures coming from `BaseFixtureGenerator`, so there should be zero change to the test logic. This part is only a structural refactor.

`BaseFixtureGenerator` is designed to provide extension by inheritance, i.e., it is meant to be customized by inheritance overrides, such as for use in module tests in `sktime`, or external packages.

This PR also adds another class, `QuickTester`, which is a mixin that adds the method `run_tests`. This can be invoked on any estimator or estimator instance, and will run all tests in `TestAllEstimators` only for that estimator. In the future, this can be used in any other test class we may write in the future based on the same design.

Minor changes:
* `create_conditional_fixtures_and_names` will ignore fixture variables for which no generators exist (i.e., no keys in the generator dict), instead of erroring out
* fixes bug in `get_fixtures` on return handling in rare case where precisely to fixtures are returned and no names. In this case, the list would be split, instead of adding names, because only `len` and not the type was checked.


fkiraly:

PR #1989, opened 2022-02-05
[BUG] Changes to proximity forest classifiers

### LINKED ISSUE #1715: [ENH] update ProximityForest

TonyBagnall:
this PR is a general update of the proximity forest classifiers, 
ProximityForest
ProximityTree
ProximityStump

It is aimed at updating the code and removing the need for these classifiers to be on the EXCLUDED_CLASSIFIER list. It goes some way to addressing #1715, but not completely. 

I have attempted to do this with the minimal amount of changes. These implementations are complex, and not done in the way I would do it. There are a lot of function pointers and static functions passed pf objects. This PR is NOT a code redesign

changes:
1. fit_idempotent was failing because of the random_state. On the first call, if passed an integer, a new np.random object is created. On the second call, the random_state was not reset, and so the random sequence simply continued from before, which can give different exemplars, and hence different predictions. I've fixed this by storing an integer in a private attribute, and always recreating the random_state on fit using this integer
2. removed all input checks which are no longer necessary
3. added the get_test_params methods for all (see #1678)
4. The function pointer attributes distance_measure, setup_distance_measure and get_distance measure were all set through the constructor, defaulting to None. This broke the hyper parameter tests and stop pickling. Simplest solution is to remove the option to set these through the constructor. There is no single evident use case where they are set differently to the default.  
5. The function pointers are causing pickling to fail. I have not any more time for this, so will attempt to get approval for the current changes


fkiraly (approved):
Looks reasonable.

While we're at it, can you kindly look at #2027, which fixes very similar issues?

fkiraly (changes requested):
Hm, I just wanted to merge this and changed my mind on whether this is ready  to merge.

If arguments get removed, they would need to be deprecated at least.

So we should either:
* add deprecation to the arguments that get removed
* ensure that the distance measure is being used properly. This could be done along the lines of #1998

fkiraly:
do you mind if I make one of these changes, @TonyBagnall?

TonyBagnall:
I'd rather you didnt @fkiraly it is ongoing here https://github.com/alan-turing-institute/sktime/pull/1989

fkiraly:
ok - so, we either need deprecation, or handle the inputs in order to merge this.

Handling inputs could be achieved by copying them to private attributes, then using those internally, or a more serious refactor of the "distances" business.

PR #1992, opened 2022-02-05
[ENH] test framework: test class for forecasting module

fkiraly:
See here for overview of recent test framework related PRs:
https://github.com/alan-turing-institute/sktime/issues/2053

---

This PR refactors the `test_all_forecasters` module to the new test class based design in #1987.
This PR requires and is conditional on #1987, all changes to `test_all_estimators` should be reviewed in #1987.

Besides the refactor, the conditional fixture generation plug-in is used to remove the dozen times repeated boilerplate code to obtain the `n_columns` variable, this is now a conditional fixture instead.

No further refactoring is carried out, although there would be some opportunity based on this (e.g., using scenarios or moving to `test_all_estimators`). Ideas for further refactoring along natural lines are left in the code as todo comments.

PR #1993, opened 2022-02-05
[ENH] one-stop estimator validity checker

fkiraly:
This PR contains a `check_estimator` utility which calls all interface tests in `sktime` relevant to an estimator class or estimator instance. The class or instance in question need not be part of `sktime`.

This replaces the earlier, hacky prototype `check_estimator` which covered only `test_all_estimators`.

The new `check_estimator` uses the `QuickTester` class from #1987 and the forecaster test refactor in #1992 to collect test results via the universal `run_tests` interface.

Currently, only general tests and forecaster tests (if relevant) are called, but the pattern can be extended easily by inheritance from `QuickTester`.

This PR also introduces a *dependency change*:
* `pytest` is added to the `binder` set of dependencies, since the `check_estimator` utility, which is showcased on some notebooks, now depends on `pytest`.

---
Update: with #2142 merged, this PR no longer depends on #1987 and #1992.


fkiraly:
Merging with self-review, after minor adjustments (merge main, linting), as discussed in dev meeting

PR #1998, opened 2022-02-06
[ENH] refactored `KNeighborsTimeSeriesClassifier`

fkiraly:
This PR is a refactor of the `KNeighborsTimeSeriesClassifier`.

It does three things:
* removes the (in my opinion) awkward and hard-to-maintain double inheritance of `KNeighborsTimeSeriesClassifier` from `sklearn`'s knn classifier. Instead adopts a wrapper pattern around the `sklearn` public interface, with much of the code simplified through passing distance matrices directly to `sklearn`.
* extends the interface to allow passing of arbitrary distances that are callables. This supports the composable distance interface from #1858 out-of-the-box (with `sklearn` style parameter inspectable composition/pipelining), and can be used to plug in custom time series distances.
* tests that rely on random seeds are updated with new expected output values

This PR also fixes two unreported bugs:
* the `algorithm` parameter of `KNeighborsTimeSeriesClassifier` was not passed to the `sklearn` `KNeighbors` classifier, the value being passed was always the default `"brute"`. https://github.com/alan-turing-institute/sktime/issues/2774
* corrected docstring in `pairwise_distance` which was missing options

and makes one minor change:
* replaces loops in tests for `KNeighborsTimeSeriesClassifier` by `mark.parametrize`


TonyBagnall (changes requested):
have you done timing tests on this? The reason we stopped importing distances directly was that it does not work properly with numba, it ends up rebuilding the c multiple times. The factory method is a way around that. Also, why have you removed chris as an author?

fkiraly:
> Also, why have you removed chris as an author?

Because I hate him.

No, that was a joke - truth is: I probably copy-pasted code from an earlier version, and I did not notice that his name dropped out. I added him back.

fkiraly:
> have you done timing tests on this?

No. Naively, I'd expect that all numba business remains and should remain in the distances? And there's no benefit from having it living in the class?

fkiraly:
>have you done timing tests on this?

@TonyBagnall, do you perhaps have some code available for timing?
You did that for a couple of your recent papers, so I assume that must exist somewhere?

TonyBagnall:
I can have a look at this PR at the weekend if you want and run some timing experiments

fkiraly:
much appreciated, @tonybagnall!

TonyBagnall:
I've got this branch, I will just post observations here as I play with it. 
First is you dont seem able to pass it a distance function

TonyBagnall:
in terms of design, I much prefer containment too. I am not sure about the removal of the distance factory. In my opinion, it is better to keep the classifier and the distances loosely coupled, so that changes/additions to distances to not also involve changes to the classifier code. Tight coupling like this increases maintenance burden. However, I dont feel that strongly. Main potential blocker is the speed, as when I did it this way we had trouble with repeatedly rebuilding numba distances. I'll set up a timing test now

TonyBagnall:
on and btw algorithm was hard coded to brute for a reason. The other options were not compatible with bespoke distances. This may have changed in scikit. I will look into this if it passes timing tests

TonyBagnall:
ah I see now, you have converted to always use a precomputed distance matrix, and it seems chris has code all set up to do that. Nice.

TonyBagnall:
so this is actually much faster in predict (at small cost in fit), but is it logically correct to do it this way with elastic distances, which are not metrics? I'll look some more, and do some black box accuracy tests over the weekend

fkiraly:
> First is you dont seem able to pass it a distance function

You can pass it a distance function, but it needs to be 3D x 3D -> number currently.
I.e., sth of the signature `pairwise_distance`.
I do think that's the most "logical" interface.

Do you want to be able to pass a distance function 2D x 2D -> number?

I think the first one makes more sense, since that's the vectorized signature.

I also talked to @chrisholder, and I think that it makes most sense that the distances exported by `distances` should be able to deal with 2D *and* 3D input, simultaneously.
This can easily be achieved by:
* references to `pairwise_distance`, instead of directly exporting the 2D x 2D signatured functions
* ensuring tha `pairwise_distance` works with 2D *and* 3D input

fkiraly:
> in terms of design, I much prefer containment too. I am not sure about the removal of the distance factory. In my opinion, it is better to keep the classifier and the distances loosely coupled, so that changes/additions to distances to not also involve changes to the classifier code. Tight coupling like this increases maintenance burden

I do agree with light coupling being better, but I thought my version is much more loosely coupled.
The distance factory and all that stuff should remain in the distance module.
This way, you have the concern "convert 2D to 3D signature" *inside the KNN classifier* where it doesn't really belong imo, that's not a concern of the KNN classifier!

fkiraly:
> Main potential blocker is the speed, as when I did it this way we had trouble with repeatedly rebuilding numba distances. 

Well, if repeated rebuilding is a problem, then you could put the factory/building in the `distance` module, and get the built distance from there.

fkiraly:
> on and btw algorithm was hard coded to brute for a reason. The other options were not compatible with bespoke distances. This may have changed in scikit. I will look into this if it passes timing tests

well, it was not agreeing with the docstring. I implemented what the docstring said.

fkiraly:
> so this is actually much faster in predict (at small cost in fit), but is it logically correct to do it this way with elastic distances, which are not metrics? I'll look some more, and do some black box accuracy tests over the weekend

Why would it be incorrect specifically with elastic distances? Is it not actually a distance?
As long as the distance does not depend on the presence of series other than those where distances are being computed on, it should make no difference?

fkiraly:
> The other options were not compatible with bespoke distances. This may have changed in scikit. I will look into this if it passes timing tests

Sklearn is compatible with bespoke distances, via the "precomputed matrix" interface. You can plug in any distance you like, and that's what this estimator is doing.

TonyBagnall:
yeah just need some time to have a look

TonyBagnall:
> > The other options were not compatible with bespoke distances. This may have changed in scikit. I will look into this if it passes timing tests
> 
> Sklearn is compatible with bespoke distances, via the "precomputed matrix" interface. You can plug in any distance you like, and that's what this estimator is doing.

ys but you cannot use the classifier with this version as you can with the current version. This does not work with yours

TonyBagnall:
so dont get me wrong, if this is correct its great, but I cant see how it can be. I imagine it is using the triangle inequality to speed up the search from the precomputed distance. However, the triangle inequality does not hold for DTW,  dtw(a,c)+dtw(c,b) can be less than dtw(a,b), so this will not necessarily find the nearest neighbour with anything but Euclidean distance.

fkiraly:
> ys but you cannot use the classifier with this version as you can with the current version. This does not work with yours

My classifier wants a distance that can already do 3D x 3D -> matrix of distances

In my opinion, that separates concerns better. No numba magic should happen in the estimator, and the estimator should not depend on numba, since the distance computations are not the knn algorithm's concern - it *uses* the distance.

I've made another pull request which ensures that the distances exported by the `distances` module can cope with 1D, 2D and 3D input:
https://github.com/alan-turing-institute/sktime/pull/2051

TonyBagnall:
I suggest you investigate how sklearn uses the precomputed distances to classify new cases, I suspect it uses speed ups based on the triangle inequality

fkiraly:
> I suggest you investigate how sklearn uses the precomputed distances to classify new cases, I suspect it uses speed ups based on the triangle inequality

That's an interesting topic, though I'm not sure how it would resolve outstanding points of our discussion?

E.g., are you happy or not with "upgrading" the exported distances in `distances` as in #2051, so they can deal with 1D, 2D, *and* 3D input?

TonyBagnall:
I'm formally voting against the PR, so that it cannot be put in under the lazy consensus rule currently in place. My objection is primarily that this is a different algorithm that I believe for some distances will not give the correct nearest neighbour. 

I'm not against it per se. If it works, that would be great, its much faster, although I would probably still like the option of either using a train distance matrix or not (which would be easy with a bespoke implementation of nearest-neighbour). I just have not got the time or inclination to dig into this, I am not researching distance based classification at the moment. For this to progress, I'd like details on how the precomputed matrix is being used with dynamic time warping distance and experiments to verify there is no significant difference to the old version with all distances. It is also not backward compatible, so would need deprecation, unless the distance factory is restored.

fkiraly:
@TonyBagnall, would you be ok to have this in a different class?

TonyBagnall:
@fkiraly I was thinking about this yesterday, I would like to test over the summer whether it makes any difference in terms of accuracy. It could be a separate class, but would you mind waiting? The distances are still not fully stable, and I want to finish the clustering side before coming back to this.

fkiraly:
I would like to call for a vote on this, we should not leave this open for a year.
As per the regulations, vote ends 5pm UTC, July 28.

fkiraly:
Related issues and posts:
https://github.com/alan-turing-institute/sktime/issues/2774, FYI @AurumnPegasus 
https://github.com/alan-turing-institute/sktime/discussions/2714, FYI @sly8787

fkiraly:
I´m in favour

Lovkush-A:
I am somewhat in favour.
- overall ideas and bug fixes sound good. based on comments, tony agrees with this
- I say 'somewhat' because I have not understood the details enough to be confident that everything works as intended.
- I think an issue should be created to address tony's concerns, instead of them being lost. E.g. to carry out the comparisons between old and new version, to investigate if sklearn assumes the triangle inequality.

MatthewMiddlehurst:
This is not a vote either way, but I believe it would better to test both versions before putting this in given the concerns raised.

Even if they do turn out to be different, I don't think having this as an option would be a bad thing, letting people accept an approximation in return for a speed-up.

fkiraly:
>to investigate if sklearn assumes the triangle inequality.

I think this is an incorrect statement, as provided in the argument.

`sklearn` allows to pass any distances or distance matrices, but for some algorithms only certain metrics satisfy all assumptions of the algorithms. Examples are `KDTree` or `BallTree`. It still allows to pass any metrics.
Ironically, the algorithm that is currently hard-wired via the bug, `brute`, allows any pairwise functions, even those that do not satisfy distance metrics.

Even if the bug would not be there, I do not think the argument has any validity, since it does not lead to a meaningful difference between current implementation and new implementation.

That is, the new implementation does not *remove* satisfaction of the triangle equality in equivalent cases where it is satisfied, and does not *add* satisfaction of the triangle equality where it is not.

The new implementation is, simply, mathematically equivalent, as far as I can tell. It is just much shorter.

fkiraly:
>This is not a vote either way, but I believe it would better to test both versions before putting this in given the concerns raised.
>Even if they do turn out to be different, I don't think having this as an option would be a bad thing, letting people accept an approximation in return for a speed-up.

@MatthewMiddlehurst, there is no "approximation" as you claim, to the best of my knowledge.
So the vote is not about trading in speed for approximation, these are just the same algorithms.
If you think otherwise, kindly point out where the approximation happens in your opinion.

To the best of my knowledge: both implementations are algorithmically and mathematically equivalent, save for removing unnecessary algorithmic steps, and a simpler implementation, and a simplified interface.

fkiraly:
So, @Lovkush-A, to summarize.

To my current understanding, there were two concerns raised:

* claim about the triangle inequality causing problems since it is no longer satisfied (? by what?)
* claim about the refactored algorithm being an approximation instead of an "exact" algorithm, providing a speed-up at the cost of not being exact anymore

I do not think the concerns are logically, scientificially valid:

* as I have said above, the triangle equality may or may not be satisfied, but nothing changes in the new implementation as compared to the old one in cases where it is, or isn't.
* the new implementation is not an approximation, it is mathematically identical with the old implementation, and simpler algorithmically (unnecessary steps) and interface wise (cleaner structure)

fkiraly:
There is, in addition, another, possibly secondary problem.

If the present algorithm were used in benchmarks of algorithmic performance or speed, those benchmarks would find for this algorithm, incorrectly, a much less favourable performance, since the brute-force algorithm is hard-wired, instead of the more efficient algorithms for k-nearest neighbor that `sklearn` also provides (e.g., `KDTree` or `BallTree`).

I'm sure this is just an innocent oversight, but it should be taken into consideration if this code was used in any benchmarks.

Lovkush-A:
Even if mathematically/logically the new and old version are the same, still seems useful to do an *empirical* comparison of the two and see if/where their behaviour differs. E.g. it might identify bugs in the new version.  I am proposing creating an issue to do these comparisons though (could be a good first issue??), rather than halting this PR being merged.

What happens if I see (minor) things I want to change, e.g. in docstrings. Does that invalidate the vote?

Lovkush-A:
To better understand the details I read through the code of new version in detail and tried for old version. I am now fully in favour of new version (rather than tentative), given how much easier it is to understand the new version and hence to maintain.  I still think it is worthwhile to have an issue that compares.

(I have several minor suggestions for this - let me know if I should suggest here or in separate issue/pr given the vote).

MatthewMiddlehurst:
I don't believe I claimed anything.

Since no one seems to be volunteering to compare the two, I am currently running it on a subset of the UCR datasets over 10 resamples.

Early results suggest the output is the same, but the new version is slower to train, i.e. 2 seconds vs 113 seconds on the Computers dataset.

This is just default parameters with DTW, are any other settings areas of concern?

fkiraly:
@MatthewMiddlehurst, you mean, the newer version is slower?
That surprises me a bit. Because the full distance matrix is computed both in the new and the old version.

Do you have code to replicate this in python?

MatthewMiddlehurst:
```
from sktime.benchmarking.experiments import load_and_run_classification_experiment

load_and_run_classification_experiment(
     problem_path="D:/CMP Machine Learning/Datasets/UnivariateTS/",
     results_path="D:/CMP Machine Learning/temp/",
     cls_name="KNN2",
     classifier=KNeighborsTimeSeriesClassifier(),
     dataset="Computers",
     resample_id=0,
     build_train=False,
     predefined_resample=False,
)
```

This is the function I ran, the problem path and results path would have to change to where you want to read the data from and write the results to.

The function outputs this file for the new version, where the first 3 items on line 3 are: accuracy, train time (milliseconds), test time (milliseconds).
[testResample0.csv](https://github.com/alan-turing-institute/sktime/files/9173937/testResample0.csv)

The equivalent file for the current version:
[testResample0.csv](https://github.com/alan-turing-institute/sktime/files/9173946/testResample0.csv)

Taking a quick look, it seems distances are always calculated in _fit, but I don't think they are always required? They may be if you wanted to tune DTW or something, but by default or for ED all the heavy lifting is done in _predict I believe. I have not touched the distances module/classifiers at all, so I'm unsure where any fitting or tuning for distances would be done if it's even possible currently 🙂.

fkiraly:
I have written a profiler utility that avoids having to spend too much time in writing files to one´s local hard drive: https://github.com/alan-turing-institute/sktime/pull/3076

I can confirm that the new version takes more time both in fit and predict. I'm not sure what causes it.
Maybe it is a memory issue?

fkiraly:
So, I am not sure about what causes the slowdown in the new version.

I concur with @Lovkush-A though, there is some code in the "old" knn classifier that seems very hacky and brittle, as it faces off to private interfaces in `sklearn` and makes case distinctions based on sklearn version. That is a recipe for disaster...
I did not know of this, when did this get added in?

Example of this:
```python
  if hasattr(check_array, "__wrapped__"):
      temp = check_array.__wrapped__.__code__
      check_array.__wrapped__.__code__ = _check_array_ts.__code__
```

fkiraly:
>let me know if I should suggest here

@Lovkush-A, here is as good as anywhere? We can transplant to issue if it gets too long.

MatthewMiddlehurst:
I don't see an equivalent to `dist_mat = self._distance(X, X)` in the old version `_fit`, and it just uses the input data X for the KNN sklearn `_fit`. I don't see any reason why the sklearn KNN fit would calculate the distances to other train cases, unless they are required for one of the non-default options.

Computers has a train size of 250 and series length of 720, probably accounts for the timing increase.

sktime/classification/distance_based/_time_series_neighbors.py
```diff
@@ -88,6 +63,28 @@ class KNeighborsTimeSeriesClassifier(_KNeighborsClassifier, BaseClassifier):
...
           array of distances, and returns an array of the same shape
           containing the weights.
+    algorithm : str, optional. default = 'brute'
+        search method for neighbours
+        one of {'auto’, 'ball_tree', 'kd_tree', 'brute'}
+    distance : str or callable, optional. default ='dtw'
+        distance measure between time series
+        if str, must be one of the following strings:
+            'euclidean', 'squared', 'dtw', 'ddtw', 'wdtw', 'wddtw',
+            'lcss', 'edr', 'erp', 'msm'
```
Lovkush-A, line 73: - the older version also had something called `twe`. Was it removed intentionally or accidentally?
- the `squared` option is not in the `DISTANCES_SUPPORTED` variable

sktime/classification/distance_based/_time_series_neighbors.py
```diff
@@ -88,6 +63,28 @@ class KNeighborsTimeSeriesClassifier(_KNeighborsClassifier, BaseClassifier):
...
+        this will substitute a hard-coded distance metric from sktime.distances
+        When mpdist is used, the subsequence length (parameter m) must be set
+            Example: knn_mpdist = KNeighborsTimeSeriesClassifier(
+                                metric='mpdist', metric_params={'m':30})
```
Lovkush-A, line 77: - `mpdist` is not in the docstring or the `DISTANCES_SUPPORTED` variable. I do not know what resolution is.
- I think `metric` should be `distance`. (and more metric_params)

sktime/classification/distance_based/_time_series_neighbors.py
```diff
@@ -88,6 +63,28 @@ class KNeighborsTimeSeriesClassifier(_KNeighborsClassifier, BaseClassifier):
...
+        if callable, must be of signature (X: Panel, X2: Panel) -> np.ndarray
+            output must be mxn array if X is Panel of m Series, X2 of n Series
+            if distance_mtype is not set, must be able to take
+                X, X2 which are pd_multiindex and numpy3D mtype
+        can be pairwise panel transformer inheriting from BasePairwiseTransformerPanel
+    distance_params : dict, optional. default = None.
+        dictionary for metric parameters , in case that distane is a str
```
Lovkush-A, line 84: typo

```suggestion
        dictionary for metric parameters , in case that distance is a str
```

sktime/classification/distance_based/_time_series_neighbors.py
```diff
@@ -114,46 +111,68 @@ class KNeighborsTimeSeriesClassifier(_KNeighborsClassifier, BaseClassifier):
...
         self.n_jobs = n_jobs
 
-        super(KNeighborsTimeSeriesClassifier, self).__init__(
+        # translate distance strings into distance callables
+        if isinstance(distance, str) and distance not in DISTANCES_SUPPORTED:
+            raise ValueError(
+                f"Unrecognised distance measure string: {distance}. "
+                f"Allowed values for string codes are: {DISTANCES_SUPPORTED}. "
+                "Alternatively, pass a callable distance measure into the constuctor."
+            )
```
Lovkush-A, line 144: the comment does not describe what is done. adjust or remove comment

sktime/classification/distance_based/_time_series_neighbors.py
```diff
@@ -114,46 +111,68 @@ class KNeighborsTimeSeriesClassifier(_KNeighborsClassifier, BaseClassifier):
...
 
-    def fit(self, X, y, **kwargs):
-        """Override fit is required to sort out the multiple inheritance."""
-        return BaseClassifier.fit(self, X, y, **kwargs)
+        # the distances in sktime.distances want numpy3D
+        #   otherwise all Panel formats are ok
+        if isinstance(self.distance, str):
+            self.set_tags(X_inner_mtype="numpy3D")
+        elif distance_mtype is not None:
+            self.set_tags(X_inner_mtype=distance_mtype)
```
Lovkush-A, line 163: what about case where self.distance is a PairwiseTransformer? The docstring says "only set this [`distance_mtype`] if distance is not BasePairwiseTransformerPanel descendant"

sktime/classification/distance_based/_time_series_neighbors.py
```diff
@@ -225,8 +203,7 @@ def kneighbors(self, X, n_neighbors=None, return_distance=True):
 
         Parameters
         ----------
-        X : sktime-format pandas dataframe with shape([n_cases,n_dimensions]),
-        or numpy ndarray with shape([n_cases,n_readings,n_dimensions])
+        X : sktime-compatible data format, Panel or Series, with n_samples series
         y : {array-like, sparse matrix}
             Target values of shape = [n_samples]
```
Lovkush-A, line 208: I dont think we want any `y` in docstring

sktime/classification/distance_based/_time_series_neighbors.py
```diff
@@ -245,163 +222,40 @@ def kneighbors(self, X, n_neighbors=None, return_distance=True):
...
-                X,
-                self._fit_X,
-                reduce_func=reduce_func,
-                metric=self.effective_metric_,
-                n_jobs=n_jobs,
-                **kwds,
-            )
-        else:
-            raise ValueError("internal: _fit_method not recognized")
+        neigh_ind = self.knn_estimator_.kneighbors(
```
Lovkush-A, line 228: minor: the variable name `neigh_ind` could be confusing, as it could represent the pair `(neigh_dist, neigh_ind)`.  one possible alternative is `kneighbors_output`?

Lovkush-A (changes requested):
Various small suggestions.

One suggestion I couldn't make inline is typo on Line 180, docstring for _fit. the final `(d)` should be `(m)`

fkiraly:
I tried to investigate what causes slower runtime, and one suspicion that I had was that, maybe it is because we precompute the matrix instead of passing the distance as a function.

This PR modifies the current PR so the distance is still passed as a function, rather than being precomputed:
https://github.com/alan-turing-institute/sktime/pull/3078

The runtime seems much, much worse.

sktime/classification/distance_based/_time_series_neighbors.py
fkiraly: how so? I looked at it twice and it seems to match. Please explain. Probably it's sth really silly on my side...

sktime/classification/distance_based/_time_series_neighbors.py
fkiraly: ah, I just copy-pasted that. You are right, though.

sktime/classification/distance_based/_time_series_neighbors.py
fkiraly: I think this was copied from sklearn, fixed

sktime/classification/distance_based/_time_series_neighbors.py
fkiraly: then it will still override `X_inner_mtype`, but the `sktime` transformer will convert it back. If the user insists on an unnecessary conversion, why not.

sktime/classification/distance_based/_time_series_neighbors.py
Lovkush-A: i could very well be mistaken, but I think the code directly below the comment just does input check on `distance`.  the str is converted into a function in the separate `_distance` method?

sktime/classification/distance_based/_time_series_neighbors.py
fkiraly: clarified by using different name and adding comment

sktime/classification/distance_based/_time_series_neighbors.py
Lovkush-A: (One reason I ask is because at the top of the class, the tag is: `"X_inner_mtype": ["pd-multiindex", "numpy3D"]`.  But it looks like that is always over-written by the loop above.)

fkiraly:
>I don't see an equivalent to `dist_mat = self._distance(X, X)` in the old version `_fit`, and it just uses the input data X for the KNN sklearn `_fit`. I don't see any reason why the sklearn KNN fit would calculate the distances to other train cases, unless they are required for one of the non-default options.

Well, I thought if the strategy is `brute`, it still happens under the `sklearn` hood.
But true, the brute force method should not need distances train to train. This is only needed in the kernel versions.

Let's see what happens if we pass a sparse matrix of zeroes.

sktime/classification/distance_based/_time_series_neighbors.py
Lovkush-A: resolved. happy with change

sktime/classification/distance_based/_time_series_neighbors.py
Lovkush-A: resolved-  happy with explanation.

fkiraly:
Hm, it seems to reduce the runtime to that of the "old" classifier - was that it, @MatthewMiddlehurst? Good thinking!

My only worry was that the algorithm inside does sth odd if we give an empty matrix, but it seems that it passes the "expected output classes" tests, too, so the algorithm inside still seems to be the same.

MatthewMiddlehurst:
Good to hear, I don't think it should change the output, but can run it again for peace of mind. 

My only concern would be it would affect the non-brute methods without us knowing, but difficult to tell without looking at the sklearn code or running them all with and without the proper distances and comparing. Maybe worth an issue to check. 

Still, given that it was always brute previously, it's an improvement even without verifying (not even counting the easier to read code). Will compare the results I ran later, but unless something comes up, I am voting in favour.

fkiraly:
>My only concern would be it would affect the non-brute methods without us knowing, but difficult to tell without looking at the sklearn code or running them all with and without the proper distances and comparing. Maybe worth an issue to check.

I already added a line to that effect in the description of the `pass_train_distances` parameter.

In either case, as you say: that no longer concerns the comparison old vs new, because all the old implementation could do was "brute".

sktime/classification/distance_based/_time_series_neighbors.py
fkiraly: ah, you meant *comment* and not *error message*. My bad for misreading what you said. Fixed.

sktime/classification/distance_based/_time_series_neighbors.py
fkiraly: Sadly, there is no ground truth for `DISTANCES_SUPPORTED` in the `distances` module, so I suppose it just has to be updated permanently.

Perhaps we should open an issue.

fkiraly: here: https://github.com/alan-turing-institute/sktime/issues/3080

Lovkush-A: good idea with the issue.

Lovkush-A:
@fkiraly There is one other typo I saw. Line 180, docstring for _fit. the final (d) should be (m)

fkiraly:
>@fkiraly There is one other typo I saw. Line 180, docstring for _fit. the final (d) should be (m)

Fixed - the type statement was also not correct, also fixed

> Also, might be worth adding a small comment about why zero matrix is passed (rather than no matrix at all?). 

Added/explained

Lovkush-A (approved):
happy with all changes and improvements. 👍

fkiraly:
@MatthewMiddlehurst, any ideas why this bug in HIVE-COTE2 would surface multiple times in the CI here?
https://github.com/alan-turing-institute/sktime/issues/3083

I thought HIVE-COTE2 has nothing to do with k neighbors?

Lovkush-A (approved):
happy with changes in most recent commit (new test params and using checks for X in _fit).

fkiraly:
vote has ended - 3 in favour, none against, merging

PR #2015, opened 2022-02-07
Trying alt path for tsf loader

fkiraly:
Trying to help with #1934 - trying different way to define path. Perhaps this helps?

FYI @rakshitha123 

Oddly, tests pass locally for me.


fkiraly:
not needed, wrong approach. solution to the original problem turned out to be different

PR #2016, opened 2022-02-07
[MNT] Remove `pytest-xdist` from mac&linux CI-CD, un-skip `test_multiprocessing_idempotent`

fkiraly:
This PR:
* removes `pytest-xdist` from macos & linux CI/CD 
* makes macos & linux CI/CD conditional on windows CI/CD
* adds back `test_multiprocessing_idempotent`.

Alternative version of #2004 which ensures "fail fast" through windows CI/CD which still uses `pytest-xdist`.

Implements https://github.com/alan-turing-institute/sktime/issues/2012 (also see there for further details)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -87,7 +87,7 @@ jobs:
           cp .coveragerc testdir/
           cp setup.cfg testdir/
           cd testdir/
-          python -m pytest --showlocals --durations=10 --cov-report=xml -n 2 --cov=sktime -v --pyargs sktime
+          python -m pytest --showlocals --durations=10 --cov-report=xml --cov=sktime -v --pyargs sktime
       - name: Display coverage report
         run: ls -l ./testdir/
       - name: Publish code coverage
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -76,7 +76,9 @@ jobs:
           conda install -c conda-forge scipy matplotlib
 
       - name: Install sktime and dependencies
-        run: python -m pip install .[all_extras,dev]
+        run: |
+          python -m pip install .[all_extras,dev]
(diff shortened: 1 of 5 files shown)


.github/workflows/test.yml
```diff
@@ -96,7 +96,7 @@ jobs:
           file: ./testdir/coverage.xml
 
   test-linux:
-    needs: code-quality
+    needs: test-windows
```
lmmentel, line 99: ?

.github/workflows/test.yml
```diff
@@ -128,7 +128,7 @@ jobs:
           file: ./testdir/coverage.xml
 
   test-mac:
-    needs: code-quality
+    needs: test-windows
```
lmmentel, line 131: ?

lmmentel:
you should also remove `pytest-xdist` from `pyproject.toml`

fkiraly:
> you should also remove `pytest-xdist` from `pyproject.toml`

Windows still needs it, so I don´t think it can go? The idea is to speed up windows and succeed or error out quickly.

.github/workflows/test.yml
fkiraly: the idea is that windows runs first, quickly, *with* `pytest-xdist`.

fkiraly:
ok, this is funny - the linux 3.8 and 3.9 are hanging, just like in #1941.

That is, even if we are using a single worker, `pytest-xdist` merely being *installed* causes the CI/CD to hang.

.github/workflows/test.yml
lmmentel: I'm afraid it might not be quicker since you would be running linux and mac after windows which itself takes ~30min so at minimum we are looking at an hour if run sequentially instead of ~40 min.

Issue #1971, opened 2022-02-01
[BUG] sporadic failure in `HIVECOTEV2.fit`


fkiraly:
It seems that `HIVECOTEV2.fit` sporadically fails on the standard test cases.

From this run:
https://github.com/alan-turing-institute/sktime/runs/5018516510

Relevant piece of log is below.

```
___________ test_fit_returns_self[HIVECOTEV2-ClassifierFitPredict] ____________
[gw0] win32 -- Python 3.7.9 C:\Miniconda\envs\test\python.exe

estimator_instance = HIVECOTEV2(arsenal_params={'n_estimators': 2, 'num_kernels': 20},
           drcif_params={'n_estimators': 2}, n_jobs=...eter_samples': 4,
                       'randomly_selected_params': 2},
           time_limit_in_minutes=0, verbose=0)
scenario = ClassifierFitPredict(args={'fit': {'X':                                                var_0
0  0     0.248357
1    -0...836
1     0.178556
2     0.738947
3...}},
                     default_arg_sequence=None, default_method_sequence=None)

    def test_fit_returns_self(estimator_instance, scenario):
        """Check that fit returns self."""
>       fit_return = scenario.run(estimator_instance, method_sequence=["fit"])

estimator_instance = HIVECOTEV2(arsenal_params={'n_estimators': 2, 'num_kernels': 20},
           drcif_params={'n_estimators': 2}, n_jobs=...eter_samples': 4,
                       'randomly_selected_params': 2},
           time_limit_in_minutes=0, verbose=0)
scenario   = ClassifierFitPredict(args={'fit': {'X':                                                var_0
0  0     0.248357
1    -0...836
1     0.178556
2     0.738947
3...}},
                     default_arg_sequence=None, default_method_sequence=None)

C:\Miniconda\envs\test\lib\site-packages\sktime\tests\test_all_estimators.py:481: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
C:\Miniconda\envs\test\lib\site-packages\sktime\utils\_testing\scenarios.py:183: in run
    res = getattr(obj, methodname)(**args)
        arg_sequence = ['fit']
        args       = {'X':                                                var_0
0  0     0.248357
1    -0.069132
2     0.323844
3...
1  0  ...93542
2     100.57...
9  0     0.312834
1    -0.428579
[...]

PR #2020, opened 2022-02-08
HC2 component bugfixes

### LINKED ISSUE #1991: [BUG] Potential bug with TDE
**Describe the bug**

Currently on line 916 of TDE, it attempts to call choice on self.random_state, which is defined as an int.

**To Reproduce**

I noticed this issue when running classification_experiments using TDE and the FingerMovements dataset.

**Additional context**

I think all thats missing is that check_random_state should be called first, then choice.

</details>

MatthewMiddlehurst:
Fixes #1991 and  #1971


TonyBagnall (approved):
I assume the cancelled test is due to the current problems with testing

fkiraly:
> I assume the cancelled test is due to the current problems with testing

We do not have a problem with testing, as far as I know.
Since this got merged, we have timeouts on the linux tests, so our CI/CD is no longer working.

May indicate a memory problem - any case, I'm unmerging this for the time being.

PR #2022, opened 2022-02-08
[ENH] test framework: test class for forecasting module, generators for fixture lists

fkiraly:
See here for overview of recent test framework related PRs:
https://github.com/alan-turing-institute/sktime/issues/2053

---

Version of #1992 with generators rather than lists being returned by the `generator_dict`.
This should declutter memory ... hopefully.

PR #2023, opened 2022-02-08
[ENH] test framework: test classes, generators for fixture lists, pytest-xdist enabled

fkiraly:

This seems to be the fastest combination of conditions, and one where the tests pass!

---

Version of #1992 with generators rather than lists being returned by the `generator_dict`, and `pytest-xdist` added back.
Equivalently, a version of #2022 with `pytest-xdist` added back.

This is to diagnose whether #2022 - which seems to have solved a major memory bottleneck - will now work with `pytest-xdist` 
That is, except for the known bug with `test_multiprocessing_idempotent` #1941; a skip for the test is also introduced in this PR.


fkiraly:
Weird - this is just as fast as #2022, without `pytest-xdist`, which in turn is substantially faster than current CI/CD.

fkiraly:

PR #2027, opened 2022-02-08
[BUG] MiniRocket to comply with sklearn init specification, fix `random_state` modification in `__init__`

### LINKED ISSUE #2026: [BUG] MiniRocket does not comply with sklearn interface: modifies random_state in constructor
`clone(MiniRocket)` errors out with the following:

```
RuntimeError: Cannot clone object MiniRocket(max_dilations_per_kernel=32, n_jobs=1, num_kernels=10000,
random_state=0), as the constructor either does not set or modifies parameter random_state
```

since it modifies the parameter `random_state` in the constructor.

We should probably also have a test for this, or check why current tests do not catch this.

FYI @angus924.

fkiraly:
Fixes #2026


fkiraly:
... and we have a hanging CI/CD here.

lmmentel (approved):
LGTM :+1:

PR #2041, opened 2022-02-10
[DOC] forecaster tutorial: multivariate forecasting, probabilistic forecasting

kejsitake:
#### Reference Issues/PRs

Attempt to fix https://github.com/alan-turing-institute/sktime/issues/1985 and https://github.com/alan-turing-institute/sktime/issues/1446

This PR adds an explanation of the new `predict_quantiles` interface to the forecasting tutorial. It also adds a multivariate example as requested in issue https://github.com/alan-turing-institute/sktime/issues/1447


lmmentel:
@kejsitake could you clear the output cells in the notebook before committing changes? It'll make it easier to review the actual changes :slightly_smiling_face:

kejsitake:
Thank you for the suggestion! I have cleared the cells. I see there is some difference in the notebook metadata causing the conflict(i.e. difference in python version). I am not sure if there is a preference in handling that?

lmmentel:
I think you can update to your latest version in order to resolve.

fkiraly (changes requested):
Thanks for your great work, @kejsitake!

Regarding structure, I think adding the content in sections 1.2.4 and 1.2.5 is the best place, and this is the best sequence (first multivariate, then prediction intervals).

My comments revolve around being more explicit in your text. The picture that I have is being the tour guide, so you should say things like "this is the forecaster", "the input should look like this", etc. I´ve added details in my comments.

examples/01_forecasting.ipynb
```diff
@@ -60,7 +60,8 @@
     "        * [1.2.1 Basic deployment workflow in a nutshell](#section_1_2_1)\n",
     "        * [1.2.2 Forecasters that require the horizon already in `fit`](#section_1_2_2)\n",
     "        * [1.2.3 Forecasters that can make use of exogeneous data](#section_1_2_3)\n",
-    "        * [1.2.4 Prediction intervals](#section_1_2_4)      \n",
+    "        * [1.2.4 Multivariate Forecasters](#section_1_2_4)\n",
+    "        * [1.2.5 Prediction intervals](#section_1_2_5)      \n",
```
fkiraly, line 64: I would also include quantile forecasts, i.e., "prediction intervals and quantile forecasts"

examples/01_forecasting.ipynb
```diff
@@ -613,11 +614,60 @@
...
    "source": [
-    "#### 1.2.4 prediction intervals<a class=\"anchor\" id=\"section_1_2_4\"></a>\n",
+    "#### 1.2.4. multivariate forecasting <a class=\"anchor\" id=\"section_1_2_4\"></a>"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Forecasters can also be used with multivariate datasets. Below is an example using the longley dataset from `sktime.datasets`. "
```
fkiraly, line 624: I would be more verbose about the following things:
* some, not all forecasters are multivariate
* for a multivariate example: show input and output
* say that the workflows are exactly the same, except for certain things that differ
* how to see whether a forecaster is multivariate, how to find multivariate forecasters (optional?)

examples/01_forecasting.ipynb
```diff
@@ -613,11 +614,60 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "#### 1.2.5 prediction intervals <a class=\"anchor\" id=\"section_1_2_5\"></a>\n",
     "\n",
-    "`sktime` provides a unified interface to return prediction interval when forecasting. This is possible directly in the `predict` function, by setting the `return_pred_int` argument to `True`. The `predict` method then returns a second argument, Not all forecasters are capable of returning prediction intervals, in which case an error will be raised.\n",
+    "`sktime` provides a unified interface to return prediction interval when forecasting. This is possible using the `predict_interval` function. Not all forecasters are capable of returning prediction intervals, in which case an error will be raised.\n",
```
fkiraly, line 668: I would recommend to also talk about the following things:
* how to check whether a forecaster can make interval predictions
* the multi-index structure of the return
* that the input minus `coverage` is exactly as for `predict` (and how this relates to the other cases, workflows)
* that `coverage` can be a vector

Further things that I think you should look at:
* the argument is called `coverage`, not `alpha`. Please refer to `coverage` (single number) as "nominal coverage of the prediction interval" and say the interval is symmetric.
* please add `predict_quantiles`, it fits nicely together with `predict_interval`.

kejsitake: - described how to check if a forecaster can make interval predictions
- described multi-index structire
- described how prediction quantiles and intervals are related
- explained that coverage can be a list of floats
- tried to replace alpha with coverage
- added predict_quantiles

fkiraly (approved):
Excellent, really nice work!

(also agree with the change in the `ThetaForecaster` tag)

PR #2048, opened 2022-02-11
[ENH] Tslearn added as soft dependency and used to add new clusterers.

chrisholder:

Tslearn offers functionality for clustering that would take a very long time to implement without it. This pr wraps tslearns clusterer and uses functionality from tslearn in existing sktime clusterer. 

Exact list of what is implemented:
1) New clusterer K-shapes (wrapping tslearn)
2) New clusterer Kernel-K-mean (wrapping tslearn)
3) Improved K-mean but offer barycenter averaging as a new averaging technique.

Tslearn 0.5.2

##### For all contributions

##### For new estimators


lmmentel:
It seems that `tslearn` is now required to build the docs, it that intentional?

sktime/clustering/_k_shapes.py
```diff
@@ -0,0 +1,154 @@
+# -*- coding: utf-8 -*-
+from typing import Union
+
+import numpy as np
+from numpy.random import RandomState
+
+from sktime.clustering._base import BaseClusterer, TimeSeriesInstances
+from sktime.utils.validation._dependencies import _check_soft_dependencies
+
+_check_soft_dependencies("tslearn")
```
lmmentel, line 10: @chrisholder I'm not 100% sure since readthedocs logs are not very transparent but I think might be triggering the error when this module is being imported in  `sktime/clustering/__init__.py`

aiwalter: Maybe doing the ckeck inside class/init?

lmmentel: That seems like a good option @aiwalter! If we want to follow the `sklearn` doctrine it might be better to put in `fit` to avoid side effects but it is also late in the execution sequence. As a user I would like to get the soft dependency info as fast as possible which practically means on importing the class from the module. One way of resolving this might be to update `_check_soft_dependencies` to act differently when in building the docs by using the env var that the rtfd sets during build https://docs.readthedocs.io/en/stable/faq.html#how-do-i-change-behavior-when-building-with-read-the-docs That would allow for documenting all the modules with soft dependencies and not actually require them to build the docs.

aiwalter: It cant be in `fit` as we need to instantiate the dependency usually in `init`, at least in the forecasters

chrisholder: So if Im reading correctly here: for now if I remove _k_shapes from the init, and then longer term we can put in place a solution that will allow us to have soft dependency files inside init.

fkiraly: @lmmentel, would you be so kind and open a "good first issue" that outlines your proposed solution to the soft dependency handling, so someone from the community could pick it up?

lmmentel:
BTW `clustering` module is not currently configured to render api any documentation.

fkiraly:
>BTW clustering module is not currently configured to render api any documentation.

Yes, indeed! opened an issue here: #2077

TonyBagnall (commented):
you might want to make the base class public, now that init cant be used normally. The other base classes are public and likely to stay that way, so it makes sense to conform imo.

TonyBagnall (approved):

sktime/clustering/k_means.py
```diff
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+"""Time series kmeans."""
```
fkiraly, line 2: does this change not need deprecation?

fkiraly (changes requested):
looks good.

One thing, to fix or discuss why not necessary: there are a lot of module/name/location changes in this PR. Does that need deprecation?

fkiraly (approved):
> could we not have more relaxed deprecation rules for what you call "experimental" modules? Otherwise change will be glacial.

Yes, valid point - we´ve agreed earlier (if I remember correctly?) that deprecation policy applies only to the more mature modules.
That resolves my only change/clarification request.

PR #2054, opened 2022-02-12
[ENH] test framework: invert test loop order - estimators first, then tests

fkiraly:

---

This PR inverts the current loop order in the test framework.

Post-PR, the order is *first* loop over estimators, *then* loop over test modules, *then* loop over individual tests.

Pre-PR, the order was *first* loop over test modules, *then* loop over indivdual tests, *then* loop over estimators.

The post-PR order is somewhat similar to the original loop order of @mloning's framework, which was:
*first* loop over test modules, *then* loop over estimators, *then* loop over individual tests.

Implementation-wise:
* there is now just a single "big test" per estimator, `test_estimators_all_tests` in `test_all_estimators`.
* this uses the `check_estimator` utility from PR #1993 to loop over all estimator tests for an individual estimator
* which in turn uses the framework refactor to test classes in PR #1992 and #1987

Hopefully, this will reduce memory usage, since tests are chunked by estimators, and estimator related artefacts persist only for the duration of that chunk, as opposed to throughout the testing.


fkiraly:

PR #2055, opened 2022-02-12
[ENH] test framework: invert test loop order plus generators

fkiraly:
See here for overview of recent test framework related PRs:
https://github.com/alan-turing-institute/sktime/issues/2053

---

Combines modifications to tests in #2054 and #2022

PR #2056, opened 2022-02-12
[ENH] test framework: `check_estimator` test, plus rename `tests` to `_tests`

fkiraly:

---

This PR is #2055, with the one difference that the `tests` folder is renamed to `_tests`.

This ensures that the "long duration" tests on estimators are executed first and get picked up by `pytest-xdist` workers, preventing the behaviour observed that a single worker works through all the long tests and the other one is idle.

There does not seem to be an easy way to ensure that using parameters or otherwise (that I know), e.g., via the `--dist` parameter.


fkiraly:

PR #2065, opened 2022-02-13
[ENH] Univariate time series bootstrapping

ltsaprounis:

Implements the `boxcox -> STL -> remainder moving block bootstraping` algorithm described in [Forecasting Principles and Practice - Chapter 12.5](https://otexts.com/fpp3/bootstrap.html). The transformer takes as input a series and returns a panel of similar time series. The panel can then be used to create bagged forecasts and prediction intervals. Will implement the `BaggedForecaster` after this is merged.

PS: thanks @aiwalter for the STL parameter docstrings, I shamelessly stole these from `STLForecaster` 😄 

1. The transformer takes as input a series and outputs a multi-index Panel mtype. Need a second check on the tags to make sure they're ok. 

#### Example usage

```python
from sktime.transformations.bootstrap import STLBootsrapTransformer

ts = load_airline()

t = STLBootsrapTransformer(10)
df = t.fit_transform(ts)

series_list = []
names = []
for group, series in df.groupby(level=[0], as_index=False):
    series.index = series.index.droplevel(0)
    series_list.append(series)
    names.append(group)

plot_series(
    *series_list,
    labels=names
)
plt.show()
```
image

##### For all contributions
- [X] I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/main/.all-contributorsrc).institute/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.

##### For new estimators


sktime/transformations/series/bootstrapping.py
```diff
@@ -0,0 +1,330 @@
...
+        else:
+            # implement as static method
+            self.block_length_ = (
+                self.block_length
+                if self.block_length is not None
+                else min(self.sp * 2, len(X) - self.sp)
+            )
+
+        # fit boxcox to get lambda and transform X
+        self.BoxCoxTransformer_ = BoxCoxTransformer(
```
aiwalter, line 195: This should be snake case I think

sktime/transformations/series/bootstrapping.py
```diff
@@ -0,0 +1,395 @@
...
+
+
+class UnivariateBootsrappingTransformer(BaseTransformer):
+    """Creates a population of similar time series.
+
+    This method utilises a form of bootstrapping to generate a population of
+    similar time series to the input time series [1]_, [2]_.
+
+    First the observed time series is transformed using a Box-Cox transformation to
+    stabilise the variance. Then it's decomposed to seasonal, trend and remainder
```
aiwalter, line 25:
```suggestion
    stabilise the variance. Then it's decomposed to seasonal, trend and residual
```

sktime/transformations/series/bootstrapping.py
```diff
@@ -0,0 +1,395 @@
...
+    time series, using the STL algorithm from statsmodels [4]_. We then sample blocks
```
aiwalter, line 26:
```suggestion
    time series, using the STL implementation from statsmodels [4]_. We then sample blocks
```

sktime/transformations/series/bootstrapping.py
```diff
@@ -0,0 +1,395 @@
...
+    from the remainder time series using the Moving Block Bootstrapping (MBB) method
+    [3]_ to create synthetic remainder series that mimic the autocorrelation patterns
+    of the observed series. Finally these bootstrapped remainders are added to the
+    season and trend components and we use the inverse Box-Cox transform to return
+    a panel of similar time series.
+
```
aiwalter, line 32: Could you maybe describe here also some examples of good use-cases for this transformer?

sktime/transformations/series/bootstrapping.py
```diff
@@ -0,0 +1,395 @@
...
+    Parameters
+    ----------
+    number_of_new_series : int, optional
```
aiwalter, line 35:
```suggestion
    n_series : int, optional
```

sktime/transformations/series/bootstrapping.py
```diff
@@ -0,0 +1,395 @@
...
+        if series name is None).
+    series_name : str, optional
+        The series name, by default None
+        If provided, the synthetic series names will have the series_name as a
+        prefix followed by an undescore e.g. "<series_name>_synthetic_1". If not
+        provided the series names will no prefix will be present e.g. "synthetic_1".
+    boxcox_bounds : Tuple, optional
+        Lower and upper bounds used to restrict the feasible range
+        when solving for the value of lambda, by default None.
+    boxcox_method : str, optional
```
aiwalter, line 58:
```suggestion
    bounds : Tuple, optional
        Lower and upper bounds used to restrict the feasible range
        when solving for the value of lambda, by default None.
    method : str, optional
```
I would keep original bixcox param names and rather describe in docstring that it goes into boxcox

sktime/transformations/series/bootstrapping.py
```diff
@@ -0,0 +1,395 @@
...
+
+import numpy as np
+import pandas as pd
+from statsmodels.tsa.api import STL as _STL
+
+from sktime.transformations.base import BaseTransformer
+from sktime.transformations.series.boxcox import BoxCoxTransformer
+
+
+class UnivariateBootsrappingTransformer(BaseTransformer):
```
aiwalter, line 18: this is a long class name :) what about just `BootstrapTransformer`? The univariate in the name is at least not needed imho.

sktime/transformations/series/bootstrapping.py
```diff
@@ -0,0 +1,395 @@
...
+    Examples
+    --------
+    >>> from sktime.transformations.series.bootstrapping import (
+    ...     UnivariateBootsrappingTransformer
+    ... )
+    >>> from sktime.datasets import load_airline
+    >>> from sktime.utils.plotting import plot_series
+    >>> ts = load_airline()
+    >>> ubt = UnivariateBootsrappingTransformer(10)
+    >>> df = ubt.fit_transform(ts)
```
aiwalter, line 126:
```suggestion
    >>> y = load_airline()
    >>> transformer = UnivariateBootsrappingTransformer(10)
    >>> y_hat = transformer.fit_transform(y)
```

sktime/transformations/series/bootstrapping.py
```diff
@@ -0,0 +1,395 @@
...
+    >>> series_list = []
+    >>> names = []
+    >>> for group, series in df.groupby(level=[0], as_index=False):
```
aiwalter, line 129:
```suggestion
    >>> for group, series in y_hat.groupby(level=[0], as_index=False):
```

sktime/transformations/series/bootstrapping.py
```diff
@@ -0,0 +1,395 @@
...
+                iterables=[[new_series_id], new_series.index],
+                names=["series_id", "time_index"],
+            )
+
+            df = df.append(pd.DataFrame(data=new_series.values, index=new_df_index))
+
+        return df
+
+    @staticmethod
+    def moving_block_bootstrap(
```
aiwalter, line 325:
```suggestion
    def _moving_block_bootstrap(
```

ltsaprounis: keeping it static, but making it private

sktime/transformations/series/bootstrapping.py
ltsaprounis: changed it to `lambda_bounds` & `lambda_method`

sktime/transformations/series/bootstrapping.py
```diff
@@ -0,0 +1,401 @@
...
+
+from sktime.transformations.base import BaseTransformer
+from sktime.transformations.series.boxcox import BoxCoxTransformer
+
+
+class BootsrappingTransformer(BaseTransformer):
+    """Creates a population of similar time series.
+
+    This method utilises a form of bootstrapping to generate a population of
+    similar time series to the input time series [1]_, [2]_.
```
fkiraly, line 22: the references are textbooks, can you kindly point to the section/chapter this is specifically implementing?

sktime/transformations/series/bootstrapping.py
```diff
@@ -0,0 +1,401 @@
...
+
+    First the observed time series is transformed using a Box-Cox transformation to
+    stabilise the variance. Then it's decomposed to seasonal, trend and residual
+    time series, using the STL implementation from statsmodels [4]_. We then sample
```
fkiraly, line 26: please point to the class/function/method

sktime/transformations/series/bootstrapping.py
```diff
@@ -0,0 +1,401 @@
...
+    blocks from the residuals time series using the Moving Block Bootstrapping (MBB)
+    method [3]_ to create synthetic residuals series that mimic the autocorrelation
+    patterns of the observed series. Finally these bootstrapped residuals are added
+    to the season and trend components and we use the inverse Box-Cox transform to
+    return a panel of similar time series.
+
+    The resulting panel can be used for Bagging forecasts, prediction intervals and
+    data augmentation.
```
fkiraly, line 34: the description should detail the structure of the returned data frame or panel object, since multiple variables are generated. Since the panel object could be `numpy`, the order of columns should also figure in this description.

sktime/transformations/series/bootstrapping.py
```diff
@@ -0,0 +1,401 @@
...
+        in the Box-Cox transformation.
+    seasonal : int, optional
+        Length of the seasonal smoother. Must be an odd integer, and should
+        normally be >= 7, by default 7
+    trend : int, optional
+        Length of the trend smoother, by default None.
+        Must be an odd integer. If not provided uses the smallest odd integer greater
+        than 1.5 * period / (1 - 1.5 / seasonal), following the suggestion in the
+        original implementation.
+    low_pass : int, optional
```
fkiraly, line 73: this should say that this parameter belongs to STL

sktime/transformations/series/bootstrapping.py
```diff
@@ -0,0 +1,401 @@
...
+    blocks from the residuals time series using the Moving Block Bootstrapping (MBB)
+    method [3]_ to create synthetic residuals series that mimic the autocorrelation
+    patterns of the observed series. Finally these bootstrapped residuals are added
+    to the season and trend components and we use the inverse Box-Cox transform to
+    return a panel of similar time series.
+
+    The resulting panel can be used for Bagging forecasts, prediction intervals and
+    data augmentation.
+
+    Parameters
```
fkiraly, line 36: perhaps order parameters by component, and/or say which component they belong to.

sktime/transformations/series/bootstrapping.py
fkiraly, line 22: Hyndman (I assume section 12.5?) does not mention the BoxCox transformer explicitly.
Personally, I would interpret the text that the bootstrap is STL plus the mentioned block bootstrap method.

sktime/transformations/series/bootstrapping.py
```diff
@@ -0,0 +1,401 @@
...
+        The length of the block in the MBB method, by default None.
+        If not provided, the following heuristic is used, the block length will the
+        minimum between 2*sp and len(X) - sp.
+    sampling_replacement: bool, optional
+        Whether the MBB sample is with or without replacement, by default False.
+    return_actual : bool, optional
+        If True the output will contain the actual time series, by default True.
+        The actual time series will be labelled as "<series_name>_actual" (or "actual"
+        if series name is None).
+    series_name : str, optional
```
fkiraly, line 53: should this not be passed as the column name in a single column `pd.DataFrame`, or name of a `pd.Series`, instead of a parameter?

fkiraly (changes requested):
This looks good!

I have questions that should be clarified (relevant? obsolete?) or acted upon (fix):
* a bundle of improvement comments about the docstring
* is the box-cox transform really part of this estimator, according to Hyndman?
* would it be worth implementing the block bootstrap in isolation, too? All the methods are there.
* is the `series` folder the right folder here? Or should it be in a `resample` folder or `series_to_panel` folder, or similar? Together with augmentation techniques?

fkiraly:
PS: just saw that the second ref, Bergmeir 2016, includes boxcox - but it also includes bagging

sktime/transformations/series/bootstrapping.py
ltsaprounis: Yes it's a bit vague in the 3rd edition, but the original reference is Bergmeir 2016. In the second edition of forecasting principles and practice, Hyndman mentions BoxCox explicitly ([fpp 2nd edition chapter 11.4](https://otexts.com/fpp2/bootstrap.html)). In general all the implementations and papers I've seen that utilise this bootstrapping method always have BoxCox (with Guerrero to find lambda) as the first step. 

Conceptually this makes sense. The variance needs to be stable for MBB on the residuals to work. The transformed series has stable variance and when the inverse transform is applied in the end, the synthetic series has the same variance pattern as the original series.

ltsaprounis:
> PS: just saw that the second ref, Bergmeir 2016, includes boxcox - but it also includes bagging

Yes! we need this transformer first and then I'll implement the `BaggingForecaster`. 
`BaggingForecaster` will generate a point forecast for each series (using any forecaster) and we can then use the mean as the point forecast and the quantiles from the simulated series as the prediction intervals. 

I also have my eye on the [`tfp.distributions.Empirical`](https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/Empirical) as the probabilistic interface for this kind of sampling methods. Another method where the same logic can be used is this one: https://otexts.com/fpp3/prediction-intervals.html#prediction-intervals-from-bootstrapped-residuals

ltsaprounis:
> This looks good!
> 
> I have questions that should be clarified (relevant? obsolete?) or acted upon (fix):
> 
> * a bundle of improvement comments about the docstring
> * is the box-cox transform really part of this estimator, according to Hyndman?

I'd say yes! (see detailed explanation in the comment section above)
> * would it be worth implementing the block bootstrap in isolation, too? All the methods are there.

Good idea! will do it in this PR
> * is the `series` folder the right folder here? Or should it be in a `resample` folder or `series_to_panel` folder, or similar? Together with augmentation techniques?

Agreed! What name should we go for? `series_to_panel`?

fkiraly:
> I also have my eye on the [`tfp.distributions.Empirical`](https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/Empirical) as the probabilistic interface for this kind of sampling methods.

Yes, exactly! One of my minor points that tipped my personal scales towards `tfp` was performant support for empirical distribiutions.

sktime/transformations/series/bootstrapping.py
fkiraly: well, you could also say: Bergmeir is a pipeline [boxcox][bootstrap][bagging], and this estimator should be [bootstrap]
Or, you could say, Bergmeir is [boxcox+bootstrap][bagging], and this estimator should be [boxcox+bootstrap].

If there is no computational advantage to put boxcox and the bootstrap together, I´d personally split them up, and then use a pipeline constrution when implementing the full Bergmeir.

But I have no strong opinion on this. Doing the "combined" version just makes the docstrings more complicated, and may result in us having to implement the pure [bootstrap] separately later (time series bootstrap resampling is generally a common component).

ltsaprounis:
Changes to the PR: 
1. Renamed original Transformer to `STLResidualBootstrapTransformer`, added the original `MovingBlockBootstrapTransformer`. 
2. Moved bootstrapping Transformers to `sktime.transformations.bootstrap`
3. Addressed docstring changes from @fkiraly

fkiraly:
uh-oh, tests are failing!

`X length in STLResidualBootsrapTransformer should be greater than block_length`

indicates at parameters that are not applicable to the test data which has time series of length 10.
Probably `block_length` needs to be smaller than 10?

ltsaprounis:
> uh-oh, tests are failing!
> 
> `X length in STLResidualBootsrapTransformer should be greater than block_length`
> 
> indicates at parameters that are not applicable to the test data which has time series of length 10. Probably `block_length` needs to be smaller than 10?
> 
> (I wouldn´t change the default, just the test parameter setting in `get_test_params`)

Yes! thanks for the tip. Took me a while to figure it out.

`test_fit_idempotent` is still failing for the `MovingBlockBootsrapTransformer`, I'm guessing it's because of the random nature of the transform. @fkiraly do you have any idea on how to deal with this?

Will also write a few more tests to make sure all the exceptions are tested.

aiwalter:
without looking into all discussions, why is is named `STLResidualBootsrapTransformer`? `STL` is per definition returning residuals, so we might rather go for `STLBootsrapTransformer`?

ltsaprounis:
> without looking into all discussions, why is is named `STLResidualBootsrapTransformer`? `STL` is per definition returning residuals, so we might rather go for `STLBootsrapTransformer`?

I'll be the first to admit that I'm bad with names 😄 
My thought process was that STL returns trend, seasonal and residual series and the Transform bootstraps the residuals while keeping the other 2 as is. If you think `STLBootstrapTranform` is informative enough I prefer it too as it's a bit shorter.

ltsaprounis:
@fkiraly / @aiwalter 
Changed name to STLForecaster and fixed the `test_fit_idempotent` by adding the `random_state` parameter.
Good to review again!

docs/source/api_reference/transformations.rst
```diff
@@ -360,3 +360,25 @@ FeatureSelection
     :template: class.rst
 
     FeatureSelection
+
+STLResidualBootsrapTransformer
```
fkiraly, line 364: typo, should be Bootstrap

sktime/transformations/bootstrap/__init__.py
```diff
@@ -0,0 +1,9 @@
+# -*- coding: utf-8 -*-
+"""Bootstrapping methods for time series."""
```
fkiraly, line 2: perhaps explain here, and in the docs, that this takes a single series and produces a panel of series?

fkiraly (dismissed):
Looks good to me now!

Let's wait for @aiwalter to check whether his concerns were addressed.

Meanwhile, I've added some minor non-blocking doc comments.

ltsaprounis:
@aiwalter - Should be good to go now, can you have a look and review/approve?

fkiraly:
merging to unblock further work of @ltsaprounis.
No review activity since one week, and all change requests by @aiwalter seemed resolved.
@aiwalter, let us know if you had any major reservations in relation to your change requests, and feel free to unmerge.

PR #2069, opened 2022-02-15
[ENH] Change default of percentage error functions #1986

### LINKED ISSUE #1986: [ENH] MAPE/sMAPE `symmetric` flag default should be switched
Current parameter defaults of MAPE, MdAPE metric classes etc heavily frustrates user expectations, since the default parameter settings actually produce sMAPE and sMdAPE etc (as pe common definition) and not MAPE, MdAPE (as per common definition).

This should be switched around, so default invocation of MAPE metric classes indeed give MAPE (as commonly defined).

The switch is easy - just flip the default to `symmetric=False` - but we will need to add a deprecation message since this is part of a widely used public interface.

See the [documentation on deprecation policy](https://www.sktime.org/en/stable/developer_guide/deprecation.html), we're probably looking at deprecation in 0.11.0, and removal in 0.12.0.

I.e., adding a "will be deprecated" warning now, then a "will be removed" warning in 0.11.0, then removing in 0.12.0.

Good first issue for someone to get involved with `sktime`.

---

from discussions, @k7e7n7t:

https://github.com/alan-turing-institute/sktime/discussions/1508

> I ran mean_absolute_percentage_error in cell 41 of 01-forecasting.ipynb and got 0.145427686270316.
> I then ran MeanAbsolutePercentageError class with the default symmetric = False (which make it asymmetric just like MAPE I assume) but I get  0.13189432350948402 which I would expect from sMAPE, but not with the flag set to False.  If I set symmetric = True it does match.  This seems backwards to me.  What am I missing?

ciaran-g:
Fixes #1986. Introduced deprecation in percentage error functions for v.0.11.0 to precede a change of the default value for `symmetric` from `True` to `False` in v.0.12.0

#### Reference Issues/PRs

Fixes #1986 

Added depreciation notice for changing default in percentage error functions

No

- correct use of the `@deprecated' function

##### For all contributions

##### For new estimators


fkiraly (changes requested):
@ciaran-g, there´s one linting failure coming from unsorted imports, otherwise looks good

fkiraly:
I'm not sure what is happening with all the tests here - some interaction with the deprecation decorator??

ciaran-g:
Hi @fkiraly - just checked the notebook tests. Hopefully the previous commit addressed the problem, but I'm not sure if it's the best way.

The tests were failing because there was an error thrown in the 01_forecasting notebook when calling this model tuning function
```
gscv = ForecastingGridSearchCV(forecaster=pipe, param_grid=param_grid, cv=cv, n_jobs=-1)
gscv.fit(y_train)
```
The deprecation decorator was causing an error with the default parallel backend in joblib. Basically it was failing to pickle the results. Changing the default parallel backend from `backend='loky'` to `backend='threading'` seems to do the trick for me anyway.

fkiraly:
@ciaran-g, how weird - thanks for tracking this down!

Regarding the change in the grid search, we should not just change the backend without warning - this would also be something that requires deprecation!

Instead, I would introduce a parameter for the parallelization back-end, if it does not exist, and change it in the notebook and/or test cases - but not in the default.

That way, we don't risk to break user code that uses instances of the grid search estimator.

ciaran-g:
Good shout Franz, just updated! Introduced the `backend` parameter, changed the call on the notebook, and the ForecastingGridSearchCV docstring example

sktime/forecasting/model_selection/_tune.py
```diff
@@ -336,6 +342,9 @@ class ForecastingGridSearchCV(BaseGridSearch):
     error_score: numeric value or the str 'raise', optional (default=np.nan)
         The test score returned when a forecaster fails to be fitted.
     return_train_score: bool, optional (default=False)
+    backend: str, optional (default=None)
+        Specify the parallelisation backend implementation in joblib, where
+        'loky' is used by default.
```
fkiraly, line 347: confusing - is it None or `"loky"`? Suggest: to avoid confusion, use the "effective" default which is `"loky"`.

fkiraly (changes requested):
with all the problems the decorators cause:

I'd suggest to move away from use of `deprecated` and instead raise a warning inside the function. Then everything should work again...

ciaran-g:
updated @fkiraly! left the `backend` argument open to the user, thought that would be useful anyway

fkiraly (changes requested):
I think this is good - would you mind reverting your changes to the forecaster tutorial?
As far as I can see, no content is changed or added.

fkiraly (approved):
I just did that if you don't mind, and will merge if/once tests pass.

Issue #2064, opened 2022-02-13
[BUG] `SFA` transformer does not comply with transformer interface since it returns unsupported format, list of dict


fkiraly:
Discovered in refactoring the transformers to the new interface. Updated after discussion below.

It would appear that the `SFA` does not comply with any `sktime` transformer interface, because `transform` returns a list of dictionaries.

Based on discussion with @MatthewMiddlehurst, @patrickzib (below and on slack), the list of dict can be read as a table, with list index = row index and key = col index, so it is formally/scientifically a transformer.

The non-compliance comes through `sktime` currently not supporting list of dict as a table format.
Two options would exist in-principle: always convert to a compliant output format (i.e., modify SFA), or introduce list of dict as a compliant format (i.e., modify the datatypes module).

We've reached the conclusion that we should do the second, i.e., that it should be added, since conversions to `pd.DataFrame` etc greately reduce performance of composites, namely classifiers that use the SFA transformer.

MatthewMiddlehurst:
The SFA transformer was one of the first things I contributed to the package, it has been in the package in this state for years at this point.

I brought this up in the refactor PR, but it was not addressed.

fkiraly:
ah, which PR is that?

fkiraly:
also, can you esplain why all the tests were passing? The `SFA` transformer is non-compliant, was not excepted from tests, and also did not raise any test breaks - I don't quite understand that.

fkiraly:
ok, so do I understand that this is really old legacy code then, before transformers had a consolidated interface?
If it is useful to have dict returns from transformers, should that be another return type?

MatthewMiddlehurst:
https://github.com/alan-turing-institute/sktime/pull/1365

We provide an option to return a dataframe instead, which is used for tests. This is not the ideal output datatype, however. The output is very sparse, so dictionaries are ideal.

MatthewMiddlehurst:
There are 200 comments in that PR so it was probably just washed away in the tide of feedback, but having a return types for dictionaries or a similar structure would be nice.

MatthewMiddlehurst:
The issue with the transformer's output is that there are possibly hundreds of thousands of potential attributes depending on the parameters, of which only a few will have non-0 values for each series. 

You could return a dataframe containing only that attributes which appear for all cases, which the previously mentioned option does, but then it becomes very difficult to compare two sets of data separately transformed.

patrickzib:
[...]

PR #2076, opened 2022-02-16
[ENH] new table mtypes: `pd.Series` based, `list` of `dict` (as used in bag of words transformers)

fkiraly:
This PR introduces two new mtypes for the `Table` scitype:
* a `pandas.Series` based mtype, `pd_Series_Table`, which has less conditions on the type than for time series (e.g., no ordered indices needed). This is being used in various components of classifiers, where `y` is a `pandas.Series` but without ordered time index.
* `list_of_dict`, which is a `list` of `dict` with the condition that all entries are primitive (`str`, `int`, or `float`). This is currently returned by bag of words type transformers, see discussion in #2064. Introduction of this type would make those transformers interface compliant.

FYI @patrickzib, @MatthewMiddlehurst

### DIFF
--- a/sktime/datatypes/_table/_check.py
+++ b/sktime/datatypes/_table/_check.py
@@ -36,10 +36,14 @@ __author__ = ["fkiraly"]
 
 __all__ = ["check_dict"]
 
+from multiprocessing.sharedctypes import Value
 import numpy as np
 import pandas as pd
 
 check_dict = dict()
+
+
+PRIMITIVE_TYPES = (float, int, str)
 
 
 def _ret(valid, msg, metadata, return_metadata):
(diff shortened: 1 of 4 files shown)


MatthewMiddlehurst:
Dictionary stuff looks good to me so far. Are there any new tests required for the new datatypes?

Will dictionary output be converted to a suitable format for tests from existing code, or do we still need the DataFrame conversion parameter in SFA?

fkiraly:
>Dictionary stuff looks good to me so far. Are there any new tests required for the new datatypes?

The existing tests in `datatypes` automatically test whether you can convert all fixtures of one type to all fixtures of another type, and whether checking correctly identifies compliant and non-compliant objects.
(for that, all that needs be done is adding the fixtures to `_examples` and the new mtypes to `_registry`)

>Will dictionary output be converted to a suitable format for tests from existing code, or do we still need the DataFrame conversion parameter in SFA?

With this, it would be automatically converted to whatever the input was, once the transformer is refactored
But I recognize it's an unsatisfactory end state.

With the "output conversion" update, we'll probably want to add an `output_mtype` arg to all transformers, so you can selectively ask for an output that's different from the input type - since you want dicts being output in the classifiers, I assume.

See here:
https://github.com/sktime/enhancement-proposals/blob/main/steps/15_output_type_homogeneization/step.md#the-proposed-solution

But that will have to happen in a separate PR, and I propose to leave SFA unrefactored until then, so you can control via its onboard parameter whether you get a dict or not (instead of losing that control).

MatthewMiddlehurst (approved):
Sounds good. I am fine with both these additions. Not that I can see anything looking through, but any potential issues can be dealt with when a classifier is refactored to use them.

patrickzib:
Thank you! :)

PR #2080, opened 2022-02-17
Minor cleanup of WindowSummarizer. Moved to summarize.py

### LINKED ISSUE #2072: [BUG] `WindowSummarizer._transform` has side effects on multivariate arguments
The recently merged `WindowSummarizer` apparently has side effects on the `X` argument of `_transform` when it is a `pandas.DataFrame`, see this run:
https://github.com/alan-turing-institute/sktime/runs/5204384690

This was discovered when `main` was merged into the transformers refactor #1854, which introduces a test suite with a larger set of test cases.
The current tests do not discover the bug (due to the old "only one input combination is tested" issue), but it is still a bug that needs fixing.

To reproduce, run `test_methods_have_no_side_effects` on the test instance of `WindowSummarizer`, and the input scenario `TransformerFitTransformSeriesMultivariate`.

FYI @danbartl. Great transformer btw!

aiwalter:
closes #2072


fkiraly:
@aiwalter, if you're working on this, have you looked at this:
https://github.com/alan-turing-institute/sktime/issues/2072
?

fkiraly (changes requested):
oh, I think we missed to do this before the current release.

Now we would have to deprecate the old location.
Can you add the old file back, and all it has is a single import?

Whenever the file is imported, a deprecation warning should be raised.

Further, @danbartl, can you check whether you are happy with this move?
Do you also want to be a code owner (in CODEOWNERS) since you wrote this?

aiwalter:
Why was it forgotten? Its asigned to the milestone?

danbartl:
Hi yes I would like to be a code owner, and  I am fine with the relocation.

fkiraly:
>Why was it forgotten? Its asigned to the milestone?

Apologies, we're no longer using the milestones for release management, but the project board: https://github.com/alan-turing-institute/sktime/projects/29

fkiraly:
& @danbartl hasn't said that it was fine to move his code, so I probably put this mentally on "wait for @danbartl" and forgot to think of it in the context of release scheduling.

fkiraly (changes requested):
After the discussion, from my side Iwould request the following 3 changes:
* add back old module, which imports new module
* old module raises deprecation message when imported
* add @danbartl to CODEOWNERS of new module

fkiraly:
Included in #2154

PR #2090, opened 2022-02-18
[ENH] Transformer pipeline and dunder method

fkiraly:
This PR introduces a `TransformerPipeline` for use in pipeline building.
Meant as a design study, but can probably already be merged if it works.

Besides being useful in its own, the intention is to use it as an element in a generic pipeline building framework.

The `TransformerPipeline` inherits from `BaseTransformer` and `_HeterogeneousMetaEstimator`, and can take a list of transformers, or a list of `(str, transformer)` pairs. If no names are provided, names are created and then made unique.

The `TransformerPipeline` is automatically invoked when the `*` (i.e., `__mul__` and `__rmul__`) dunder methods are called from any transformer. If multiple `TransformerPipeline`s are multiplied, they are conveniently merged to a single `TransformerPipeline`, instead of creating a pipeline of pipelines.

The idea can be extended later by:
* specific scitypes defining multiplication with transformers, e.g., forecasters or classifiers
* defining the addition `+` dunder methods to be `FeatureUnion` (once that's fixed and somone reviews the transformers refactor!) - that would satisfy the usual distributive property (although not being commutative)


sktime/transformations/compose.py
```diff
@@ -0,0 +1,464 @@
...
+            for instance, `my_trafo1 * my_trafo2 * my_trafo3`
+            will result in the same object as  obtained from the constructor
+            `TransformerPipeline([my_trafo1, my_trafo2, my_trafo3])`
+        magic multiplication can also be used with (str, transformer) pairs,
+            as long as one element in the chain is a transformer
+
+    Parameters
+    ----------
+    transformer : list of sktime transformers, or
+        list of tuples (str, transformer) of sktime transformers
```
aiwalter, line 55: is `List` also allowed in `sklearn.Pipeline`? I thought its required to be `List[Tuple]`? So might stick to what `sklearn` uses?

sktime/transformations/compose.py
```diff
@@ -0,0 +1,464 @@
...
+    `transform` - result is of executing `trafo1.transform`, `trafo2.transform`, etc
+        with `trafo[i].transform` input = output of `trafo[i-1].transform`,
+        and returning the output of `trafoN.transform`
+    `inverse_transform` - result is of executing `trafo[i].inverse_transform`,
+        with `trafo[i].inverse_transform` input = output `trafo[i-1].inverse_transform`,
+        and returning the output of `trafoN.inverse_transform`
+    `update` - changes state by chaining `trafo1.update`, `trafo1.transform`,
+        `trafo2.update`, `trafo2.transform`, ..., `trafoN.update`,
+        where `trafo[i].update` and `trafo[i].transform` receive as input
+            the output of `trafo[i-1].transform`
```
aiwalter, line 36: why is there different indentation? I think here we dont need any indentation?

sktime/transformations/base.py
```diff
@@ -133,6 +133,30 @@ def __init__(self):
 
         super(BaseTransformer, self).__init__()
 
+    def __mul__(self, other):
```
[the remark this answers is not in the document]
fkiraly: sure, will do

sktime/transformations/compose.py
```diff
@@ -0,0 +1,464 @@
...
+            will result in the same object as  obtained from the constructor
+            `TransformerPipeline([my_trafo1, my_trafo2, my_trafo3])`
+        magic multiplication can also be used with (str, transformer) pairs,
+            as long as one element in the chain is a transformer
+
+    Parameters
+    ----------
+    transformer : list of sktime transformers, or
+        list of tuples (str, transformer) of sktime transformers
+    """
```
aiwalter, line 56: There is missing a section `Attributes` where you can add e.g. `transformers_`. I actually would rename this param and attribute to `steps` and `steps_`

sktime/transformations/base.py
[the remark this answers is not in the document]
fkiraly: done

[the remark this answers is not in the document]
fkiraly: Usage is already documented, in the `TransformerPipeline` docstring:

```
    `TransformerPipeline` can also be created by using the magic multiplication
        on any transformer, i.e., any estimator inheriting from `BaseTransformer`
            for instance, `my_trafo1 * my_trafo2 * my_trafo3`
            will result in the same object as  obtained from the constructor
            `TransformerPipeline([my_trafo1, my_trafo2, my_trafo3])`
        magic multiplication can also be used with (str, transformer) pairs,
            as long as one element in the chain is a transformer
```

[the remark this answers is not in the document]
fkiraly: I intend to include this in the transformer tutorial, too

aiwalter (commented):
I looked a bit into. I was thinking why we need a separate class for it and cant integrate it into existing pipeline classes or replace them with sth that is superior in functionality? On the other hand, this class could be used in other pipeline classes

sktime/transformations/compose.py
fkiraly: >why is there different indentation
so you can see where to look for the different methods, each indented paragraph is about a different method, like `fit`, `predict`, etc. Without indentation, it would be less structured and more difficult to read.

sktime/transformations/compose.py
fkiraly: the parameter needs to stay, without underscore - that's what's visible to the user and to the `sklearn` interface.

The underscore version is an attribute, indeed that's an oversight, I'll add that to the docstring.

sktime/transformations/compose.py
aiwalter: yes it says `stepslist of tuple`: https://scikit-learn.org/stable/modules/generated/sklearn.pipeline.Pipeline.html

sktime/transformations/compose.py
fkiraly: added this to the docstring

sktime/transformations/compose.py
fkiraly: it extends the interface. Why always the hassle to name your estimators if you can type it more quickly?
Also, the syntax `mytrafo1 * mytrafo2` would not work without.

fkiraly:
> I looked a bit into. I was thinking why we need a separate class for it and cant integrate it into existing pipeline classes or replace them with sth that is superior in functionality? On the other hand, this class could be used in other pipeline classes

Yes, that was my thinking - have on class by what the "result" is, and then use it in an interpretar pattern, or call it in more powerful pipeline classes.

fkiraly:
other reason, imo we want to chain transformers without having a specialized estimator in the mix. Under that assumption, this piece of functionality must exist somewhere, and it most naturally exists in the transformers module.

sktime/transformations/compose.py
aiwalter: but without giving a name, you cant tune it as you need the name to do `mytrafo__method="trend"`.

fkiraly: it auto-generates names which you can look up in `transformers_`.
In the simplest case, the name is just the class name (if they are unique), otherwise `_0`, `_1` are appended etc.

fkiraly: (& the names are unique for the entire pipeline - if you add more elements, and accidentally create a name clas, it is auto-reolved)

fkiraly:
```
    transformers_ : list of tuples (str, transformer) of sktime transformers
        clones of transformers in `transformers` which are fitted in the pipeline
 -->    is always in (str, transformer) format, even if transformers is just a list
        strings not passed in transformers are unique generated strings
        i-th transformer in `transformers_` is clone of i-th in `transformers`
```

sktime/transformations/base.py
```diff
@@ -134,6 +134,56 @@ def __init__(self):
...
```
aiwalter, line 137: I am wondering why we need this, ans why is it not `__add__` instead? I dont see a concrete benefit and usecase for using a dunder method. So I would prefer to not implement a dunder method.

sktime/transformations/compose.py
```diff
@@ -0,0 +1,502 @@
...
+        i-th transformer in `transformers_` is clone of i-th in `transformers`
+    """
+
+    _required_parameters = ["transformers"]
+
+    # no default tag values - these are set dynamically below
+
+    def __init__(self, transformers):
+
+        self.transformers = transformers
```
aiwalter, line 73: I would still prefer to name it `steps` as in our other pipeline classes and also like in `sklearn`

sktime/transformations/compose.py
```diff
@@ -0,0 +1,502 @@
...
+
+    @property
+    def _transformers(self):
+        return self._get_estimator_tuples(self.transformers, clone_ests=False)
+
+    @_transformers.setter
+    def _transformers(self, value):
+        self.transformers = value
+
+    def __mul__(self, other):
```
aiwalter, line 104: I would remove the two dunder methods

sktime/transformations/compose.py
```diff
@@ -0,0 +1,502 @@
...
+        on any transformer, i.e., any estimator inheriting from `BaseTransformer`
+            for instance, `my_trafo1 * my_trafo2 * my_trafo3`
+            will result in the same object as  obtained from the constructor
+            `TransformerPipeline([my_trafo1, my_trafo2, my_trafo3])`
+        magic multiplication can also be used with (str, transformer) pairs,
+            as long as one element in the chain is a transformer
+
+    Parameters
+    ----------
+    transformers : list of sktime transformers, or
```
aiwalter, line 54: I would prefer to only accept tuple input as in our other pipeline classes and as in `sklearn`. I think now it makes the interface more confusing and its not. Also in the zen of Python, it says:

> There should be one-- and preferably only one --obvious way to do it.

😃

aiwalter (commented):
left some comments

sktime/transformations/base.py
fkiraly: the benefit is that you can write sth like `my_trafo * my_trafo2` and that´s your pipeline.

Also, "add" seems more natural for `FeatureUnion` rather than for composition.
If you use addition for feature union and multiplication for composition, the pipeline satisfies the familiar distributive law.

sktime/transformations/compose.py
fkiraly: why? If one does not like them, one can simply not use them.
I think it does add a lot of user comfort.

sktime/transformations/compose.py
fkiraly: can do that

sktime/transformations/compose.py
fkiraly: 1 Beautiful is better than ugly.
2 Simple is better than complex.
3 Flat is better than nested.
4 Readability counts.
-> compare `my_trafo * my_trafo2` with `TransformerPipeline(steps=[("mytrafo", mytrafo), ("mytrafo2", mytrafo2)])`

sktime/transformations/base.py
fkiraly: > I dont see a concrete benefit and usecase for using a dunder method. 

Simple specification of pipelines, increased readibility of pipeline specifications, user comfort.

fkiraly: see this for a widely popular design in R which is related
https://cran.r-project.org/web/packages/magrittr/vignettes/magrittr.html

aiwalter:
ok I dont mind to approve it if we could change the param name to `steps`. Where is this dunder method described in docs actually? It might be helpful to add an `Examples` section in docstring?

fkiraly:
> Where is this dunder method described in docs actually?

Currently, all functionality should be documented properly in the `TransformerPipeline` docstring, and in the docstrings of `__mul__`, `__rmul__`.
I have also wondered about where it *should* be. In the future, we may want to add this in a "pipelines" tutorial, and in the estimator specific tutorials. 

> It might be helpful to add an Examples section in docstring?

Yes, will add examples, and change the parameter to `steps`.

sktime/transformations/compose.py
```diff
@@ -51,32 +51,80 @@ class TransformerPipeline(BaseTransformer, _HeterogenousMetaEstimator):
...
+    >>> tp = TransformerPipeline(steps = [t1, t2])
+    >>> # unique names are generated for the two components t1 and t2
+    >>> tp.get_params()
+    {'steps': [ExponentTransformer(power=2), ExponentTransformer()],
+    'ExponentTransformer_1': ExponentTransformer(power=2),
+    'ExponentTransformer_2': ExponentTransformer(),
+    'ExponentTransformer_1__offset': 'auto',
+    'ExponentTransformer_1__power': 2,
+    'ExponentTransformer_2__offset': 'auto',
+    'ExponentTransformer_2__power': 0.5}
```
aiwalter, line 84: are the names automatically created in upper camel case? If so, I would suggest to lower, so we would get `exponenttransformer_1`

sktime/transformations/compose.py
```diff
@@ -51,32 +51,80 @@ class TransformerPipeline(BaseTransformer, _HeterogenousMetaEstimator):
...
+    Examples
+    --------
+    >>> # we'll construct a pipeline from 2 transformers below, in three different ways
+    >>> # preparing the transformers
+    >>> from sktime.transformations.series.exponent import ExponentTransformer
+    >>> t1 = ExponentTransformer(power=2)
+    >>> t2 = ExponentTransformer(power=0.5)
+
+    >>> # Example 1: construct without strings
+    >>> tp = TransformerPipeline(steps = [t1, t2])
```
aiwalter, line 75: I would rename `tp` to `pipe` or `pipeline`

sktime/transformations/compose.py
fkiraly: done

sktime/transformations/compose.py
aiwalter: hm, usually names are in lower case 🤔

aiwalter (approved):

Issue #1738, opened 2021-12-10
[ENH] Define and add small dataset used to showcase and test hierarchical forecasting. Add synthetic data generators.


fkiraly:
For development and testing in hierarchical forecasting we'll need:
- [x] at least one data set like airline, but hierarchical (at least two levels of hierarchy)
- [x] a data generator like `_make_series`, but hierarchical, input should be: number variables, number of levels, perhaps also number of units per level

fkiraly:
task 2 is solved by #2093

fkiraly:
task 1 is solved by `_bottom_hier_datagen`

PR #2093, opened 2022-02-19
[ENH] hierarchical mtype generator 

ltsaprounis:
#### Reference Issues/PRs

See also #1738

Implements a synthetic data generator for the hierarchical mtype

#### TO-DOs: 
- [X] Unequal length 
- [X] Docstring
- [ ] Unit tests


fkiraly (changes requested):
Looks great!

Can you write a docstring, please?
(some things I really have to guess...)

danbartl:
Hi thanks looks great and really helpful!

Just as an idea: Do we want the total column h0 to be included, since it is in principle redundant?
```
_make_hierachical(hierarchy_levels = (1, 4))
                            c0
`...`
h0   h1   time                
h0_0 h1_0 2000-01-01  1.000000
          2000-01-02  4.804243
```
 
Also regarding the naming: For multtiindex we have instances and timepoints as name, maybe here have the time column also be named timepoints?

Regarding time series to have unequal time index: Guess it could make sense, since hierarchical reconciliation does not really care about different lengths (as long as fh is the same)

ltsaprounis:
Addressed the comments from @fkiraly and @danbartl. 
Will add some unit tests before requesting the next round of reviews.

fkiraly (approved):
Looks great, very useful, thanks!

Issue #1746, opened 2021-12-12
[ENH] Probability Distributions in probabilistic forecasters


kejsitake:
**Is your feature request related to a problem? Please describe.**

We are currently working on refactoring the forecasting interfaces to allow probabilistic and interval predictions as described here https://github.com/alan-turing-institute/sktime/issues/1687 and a high level summary included here https://github.com/sktime/enhancement-proposals/blob/main/steps/13_proba_forecasting/step.md

**Describe the solution you'd like**

We would like to include a way to return the probability distributions. We briefly discussed using methods from tensorflow, but are open to other suggestions.

**Potential issues**
This would introduce a new dependency.

fkiraly:
The STEP suggests `tensorflow-probability`, but other options are `pytorch` and probabilistic programming languages.

The key requirement is that we don't just want to sample from them, but call functions like `pdf`, `cdf`, etc, and vectorize over them.

kejsitake:
Notes on this topic from meeting with @ltsaprounis @eenticott-shell @fkiraly 

The things we need from the new dependency are:
- symbolic representation of return distributions
- methods pdf, cdf, quantile (of distributions)
- composability of distributions
- optimally: vectorization of pdf, cdf calls and parameter constructor

Options we are looking at are:
- [ ] tensorflow-probability
- [ ] pytorch
- [ ] pyro
- [ ] pymc3

fkiraly:
Done in https://github.com/alan-turing-institute/sktime/pull/2100

fkiraly:
reopening this - `tensorflow-probability` has proven clunky to use, and a heavy dependency (>300MB in the env!) just for users who want probabilistic forecasting.

I think we need to move to a back-end agnostic approach.

fkiraly:
STEP is here:
https://github.com/sktime/enhancement-proposals/pull/31

prototype:
https://github.com/sktime/sktime/pull/4190

fkiraly:
closing in favour of https://github.com/sktime/sktime/issues/1746

PR #2100, opened 2022-02-20
[ENH] `predict_proba` for forecasters, `tensorflow-probability` dependency

fkiraly:
Relates to #1746, proposes interface and technical solution.
Solves interface/framework for `predict_var` and `predict_proba` in #1687.

This PR adds `predict_proba` and `predict_var` interface and default behaviour to `BaseForecaster`:
* `predict_var` returns a `pd.DataFrame` of variance predictions
* `predict_proba` returns a `tfp` `Distribution` with batch indices corresponding to `fh` indices

The default behaviour is as follows:
* `predict_var` first looks if `predict_proba` is implemented, and gets variance there if it is. Otherwise, if intervals/quantiles are implemented, returns variance of a normal with same IQR as the predictions
* `predict_proba` uses `predict` as location and square root of `predict_var` as scale parameter. If `predict_var` is not directly implemented, leverages its default behaviour to get this from intervals/quantiles.

This PR also introduces a soft dependency on `tensorflow` and `tensorflow-probability`, localized to a `dl` dependency set.
This dependency set is added to dependency specifications where necessary; a function `_check_dl_dependencies` to check and raise error if the dependency is not installed is introduced.

Further, this PR turns on a number of tests to work for `predict_var` and `predict_proba`, by adding defaults for these methods to the scenarios.

Depends on the following PR:
* https://github.com/alan-turing-institute/sktime/pull/2105
* https://github.com/alan-turing-institute/sktime/pull/2108


aiwalter:
I think `[all_extras]` should still install all soft-dependencies. We can have the `dl` option as a subset of it. You can look here how this can be done: https://github.com/pandera-dev/pandera/blob/master/setup.py

fkiraly:
>You can look here how this can be done:

I can´t copy this, because we use `pyproject.toml` and not a python file to define the dependency sets?
Or do I misunderstand sth?

I can, of course, just copy the `dl` dependencies into the `all_extras`.

@lmmentel, @freddyaboulton, any opinions on this from the maintenance perspective?

fkiraly:
> I can, of course, just copy the `dl` dependencies into the `all_extras`.

@aiwalter, I just did that.

.github/workflows/test.yml
```diff
@@ -76,7 +76,7 @@ jobs:
           conda install -c conda-forge scipy matplotlib
 
       - name: Install sktime and dependencies
-        run: python -m pip install .[all_extras,dev]
+        run: python -m pip install .[all_extras,dev,dl]
```
aiwalter, line 79:
```suggestion
        run: python -m pip install .[all_extras,dev]
```

.github/workflows/test.yml
```diff
@@ -112,7 +112,7 @@ jobs:
 
       - name: Install sktime and dependencies
         run: |
-          python -m pip install .[all_extras,dev]
+          python -m pip install .[all_extras,dev,dl]
```
aiwalter, line 115:
```suggestion
          python -m pip install .[all_extras,dev]
```

.github/workflows/test.yml
```diff
@@ -157,7 +157,7 @@ jobs:
...
```
aiwalter, line 160:
```suggestion
          python -m pip install .[all_extras,dev]
```

.github/workflows/wheels.yml
```diff
@@ -51,7 +51,7 @@ jobs:
         run: echo "WHEELNAME=$(ls ./wheelhouse/sktime-*none-any.whl)" >> $GITHUB_ENV
 
       - name: Install wheel and extras
-        run: python -m pip install "${{ env.WHEELNAME }}[all_extras,dev]"
+        run: python -m pip install "${{ env.WHEELNAME }}[all_extras,dev,dl]"
```
aiwalter, line 54:
```suggestion
        run: python -m pip install "${{ env.WHEELNAME }}[all_extras,dev]"
```

.github/workflows/wheels.yml
```diff
@@ -111,7 +111,7 @@ jobs:
         run: conda activate test
 
       - name: Install wheel and extras
-        run: python -m pip install "${env:WHEELNAME}[all_extras,dev]"
+        run: python -m pip install "${env:WHEELNAME}[all_extras,dev,dl]"
```
aiwalter, line 114:
```suggestion
        run: python -m pip install "${env:WHEELNAME}[all_extras,dev]"
```

build_tools/azure/build_wheels.sh
```diff
@@ -28,7 +28,7 @@ echo "Included Python versions: " "${PYTHON_VERSIONS[@]}"
 cd /io/ # Change directory
 
 for PYTHON in "${PYTHON_VERSIONS[@]}"; do
-  "${PYTHON}/python" -m pip install ".[all_extras,dev]"
+  "${PYTHON}/python" -m pip install ".[all_extras,dev,dl]"
```
aiwalter, line 31:
```suggestion
  "${PYTHON}/python" -m pip install ".[all_extras,dev]"
```

docs/source/installation.rst
```diff
@@ -219,7 +219,7 @@ In the ``anaconda prompt`` terminal:
 
    If you fail to satisfy all the requirements see the `troubleshooting section <https://www.sktime.org/en/stable/installation.html#release-versions-troubleshooting>`_.
 
-5. Build an editable version of sktime :code:`pip install -e .[all_extras,dev]`
+5. Build an editable version of sktime :code:`pip install -e .[all_extras,dev,dl]`
```
aiwalter, line 222:
```suggestion
5. Build an editable version of sktime :code:`pip install -e .[all_extras,dev]`
```

sktime/forecasting/base/_base.py
```diff
@@ -448,6 +449,111 @@ def predict_interval(
...
+        self.check_is_fitted()
+        # input checks
+        fh = self._check_fh(fh)
+
+        # check and convert X
+        X_inner = self._check_X(X=X)
+
+        pred_var = self._predict_var(fh=fh, X=X_inner)
+
+        return pred_var
```
aiwalter, line 499: maybe stupid question, but how many columns does this `pd.DataFrame` have? Can it also be a ´pd.Series´?

sktime/forecasting/base/_base.py
```diff
@@ -448,6 +449,111 @@ def predict_interval(
 
         return pred_int
 
+    def predict_var(
+        self,
+        fh=None,
+        X=None,
+    ):
+        """Compute/return variance forecasts.
```
aiwalter, line 457: can we explain here a bit better what a variance forecast is about? For me its not so clear yet.

aiwalter (commented):
left few comments

.github/workflows/test.yml
fkiraly: yes, of course.

sktime/forecasting/base/_base.py
fkiraly: it could, but I thought we were moving to "one output type", so it´s currently always a `pd.DataFrame`.

sktime/forecasting/base/_base.py
fkiraly: If you are making predictions of random variables `Y_1, Y_2, ... Y_N`, it is a statement about what the estimator thinks `Var(Y_1), Var(Y_2),  ..., Var(Y_N)` are.

sktime/forecasting/base/_base.py
aiwalter: but we still use `pd.Series` for univariate, so I would expect also a `pd.Series` here if I give a `pd.Series`?

aiwalter: now I am a bit confused, I though we use "what goes in comes out"? So why do we fix here to `pd.DataFrame`? Should it then not also support `numpy` and `pd.Series`?

fkiraly: >now I am a bit confused, I though we use "what goes in comes out"?

We decided we have a target of "one output type", while the status quo is mostly "what goes in comes out".
While existing interfaces may take time to refactor, change, or deprecate, we can easily ensure that new interfaces satisfy the design principle.

aiwalter: can we not use the "dtype machinery" also for `predict_var`? Then we can deprecate all together at some point in time?

sktime/forecasting/base/_base.py
aiwalter: ah I see, maybe just add this to the docstring?

fkiraly: ok, added two lines. Anything more, and we need some groundwork in mathematical notation - I´ll work on the scientific reference.

sktime/forecasting/base/_base.py
fkiraly: yes, we could, and it would be pretty easy...
although the "sister functions" `predict_interval` and `predict_quantiles` return only `pd.DataFrame`, since they need to use a `pd.MultiIndex` to structure the data.
So not only would it be a step back, it might also be confusing due to the inconsistency with the other "probabilistic prediction" functions.

PR #2102, opened 2022-02-20
[ENH] Refinements to the clustering module

### LINKED ISSUE #1857: [DOC] Formulae in `distances` module seem incorrect

TonyBagnall:
Fixes some issues with distances and clustering. Also fixes #1857

1. Bug in the use of the distance factory meant windows were not being used correctly
2. Bug in the difference function meant they were not being correctly calculated
3. Change from the Keogh style slope averaging [1] to simple first order differences
4. Add informative docstrings to weighted dtw as requested in #1857 

Still to do:
1. Fix erp, edp and lcss to gain equivalence with tsml
2. Figure out why msm is producing such poor clusterings
3. Remove the difference based distances, since it is inefficient to take the differences each time, it is better to just difference in the classifier or clusterer and avoid bloat. 
4.  Includes kshape and tslearn. This is being done fully in #2048, will merge once that is in

This will stop being draft when we have completed the extensive benchmarking, but I want to make sure it passes all tests. 

[1] Keogh, E., Pazzani, M.: Derivative dynamic time warping. In: Proc. 1st SDM (2011)


TonyBagnall:
no its getting superceeded by other PRs could you please stop reopening my PRs!

fkiraly:
alright, sorry!
I probably conflated this with #2112 where you said

> if you want to take it over please do, I have other stuff to do,. I an really am not a fan of this code, so knock yourself out

PR #2103, opened 2022-02-20
[ENH] tsf loader mtype fix

ltsaprounis:

Introduces 2 major changes to the `load_tsf_to_dataframe`
1. metadata are returned as a dictionary and as separate values
2. output dataframe can either be the original return type or a multiindex/hierarchical mtype

##### For all contributions


fkiraly (changes requested):
Great, seems to work!

May I make one change request:
the return is a `Panel` data frame of sktime `pd-multiindex` type, right?

So, what I would do, is let `return_type` be any sktime `Panel` mtype string, plus `default_tsf`. This can be achieved by, after `convert_tsf_to_multiindex`, running a `convert(obj, from_type="pd-multiindex", to_type=return_type)`.

in the docstring, you could write
```
   return_type : str - "tsf_default" (default), or valid sktime mtype for `Panel` scitype
        the in-memory format to return the loaded data container as
        valid mtype strings, with explanation, are in datatypes.MTYPE_REGISTER

Do you think this makes sense, for consistency?

Mid-term (after this PR), we could move your converter to `datatypes`.

**Comment by ltsaprounis** [comment]
> the return is a `Panel` data frame of sktime `pd-multiindex` type, right?

It's actually `pd_multiindex_hier` type. I changed the code to allow for all valid panel types as you recommended, but we need to implement the converters from `pd_multiindex_hier` to all the other Panel mtypes. The easiest way will be to take all hierarchy levels and concat them with a `"_"` or `"|"` seperator into a single column and convert it to `pd-multiindex`.

**Comment by fkiraly** [comment]
oh, I see!
In this case, `pd_multiindex_hierarchical` or `pd-multiindex` should be the default return (depending on number of levels) imo.
I thought it would always be panel.

Recommended new docstring & specification then:

```
   return_type : str - "hierarchical" (default), "tsf_default", or valid sktime mtype
        string for in-memory data container format specification of the return
        "hierarchical" = pd.DataFrame of sktime type `pd_multiindex_hier` or `pd-multiindex`
             depending on the number of levels: 2 -> `pd-multiindex; >2 -> `pd_multiindex_hier`
        "tsf_default" = container that faithfully mirrors tsf format, (add more information 
        other valid mtype strings are Panel or Hierarchical mtypes in in datatypes.MTYPE_REGISTER
            if Panel or Hierarchical mtype str is given, a conversion to that mtype will be attempted
        for tutorial and full specifications, see examples/AA_datatypes_and_datasets.ipynb
```

ltsaprounis:
`pd-multiindex` requires that the level names are `instances` and `timepoints`. I've added the necessary code that renames the levels, but I want to challenge the strict naming rule here. I think we should just ensure that the last level is the time dimension and that's it. If we do this `pd_multiindex_hierarchical` will be equivalent to `pd-multiindex`

In the most recent commit here, the user will have to specify the `return_type="pd-multiindex"` to enforce the level names as `instances` and `timepoints`. Otherwise the .tsf file names are preserved.

fkiraly:
>pd-multiindex requires that the level names are instances and timepoints. I've added the necessary code that renames the levels, but I want to challenge the strict naming rule here.

Yes, fully agreed! 
It proved to be a nightmare recently, since it is inconsistent with the hierarchical mtype.
It´s a legacy specification element which in turn comes from the legacy conersion function.

We´ve relaxed that https://github.com/alan-turing-institute/sktime/pull/2262, but I forgot to update the notebook - would appreciate PR if you have the time.

ltsaprounis:
There's a few conversion functions that still require the `instances` and `timepoints` spec. I've made an attempt to relax them in #2384. Lets wait till #2384 passes the tests and gets merged and I'll then adapt the tsf loader as described in your second review below:



> oh, I see! In this case, `pd_multiindex_hierarchical` or `pd-multiindex` should be the default return (depending on number of levels) imo. I thought it would always be panel.
> 
> Recommended new docstring & specification then:
> 
> ```
>    return_type : str - "hierarchical" (default), "tsf_default", or valid sktime mtype
>         string for in-memory data container format specification of the return
>         "hierarchical" = pd.DataFrame of sktime type `pd_multiindex_hier` or `pd-multiindex`
>              depending on the number of levels: 2 -> `pd-multiindex; >2 -> `pd_multiindex_hier`
>         "tsf_default" = container that faithfully mirrors tsf format, (add more information 
>         other valid mtype strings are Panel or Hierarchical mtypes in in datatypes.MTYPE_REGISTER
>             if Panel or Hierarchical mtype str is given, a conversion to that mtype will be attempted
>         for tutorial and full specifications, see examples/AA_datatypes_and_datasets.ipynb
> ```

fkiraly:
> I've made an attempt to relax them in https://github.com/alan-turing-institute/sktime/pull/2384.

Yes, I´ve also tried to do this as part of the "magic failure" PR #2381, but that part should, in isolation, be hopefully tame...

ltsaprounis:
@rakshitha123 - made a few changes in the loader here!

fkiraly:
ready for review, @ltsaprounis ?

ltsaprounis:
> ready for review, @ltsaprounis ?

Yes please!

fkiraly (changes requested):
Looks great!

One request:
in the test `test_load_tsf_to_dataframe`, would you be able to use `check_is_mtype` to assert that the loaded data frame indeed has the right type? Just as another safety layer here.

rakshitha123:
> @rakshitha123 - made a few changes in the loader here!

Looks good!

ltsaprounis:
> One request:
> in the test `test_load_tsf_to_dataframe`, would you be able to use `check_is_mtype` to assert that the loaded data frame indeed has the right type? Just as another safety layer here.

@fkiraly - good catch! the mtype test failed initially due to the returned dtypes. Fixed it now.
Ready for review!

fkiraly (approved):
> the mtype test failed initially due to the returned dtypes. Fixed it now.

I knew it would be useful!
Looks all done to me now!

PR #2105, opened 2022-02-20
[ENH] use of scenarios in `test_persistence_via_pickle`

fkiraly:
This PR refactors `test_persistence_via_pickle` to use scenarios.
This brings it in line with all the other tests in `test_all_estimators`, it seems for some reason `test_persistence_via_pickle` got overlooked.
This also removes the `_make_args` dependency from `test_all_estimators` entirely.


fkiraly:
This was already included with #2142, so is no longer needed.
Must have forgotten and made the same change twice...

fkiraly:
included with merged https://github.com/alan-turing-institute/sktime/pull/1854, thus no longer needed separately.

PR #2108, opened 2022-02-21
[BUG] fixed state change caused by `ThetaForecaster.predict_quantiles`

fkiraly:
Contrary to specification `ThetaForecaster.predict_quantiles` changed object attributes - namely, an attribute `sigma_` was created in `self`. This was probably accidental.

The tests did not pick this up, since much of the generic test suite was deactivated for the new probabilistic forecasting methods.

This is being changed in https://github.com/alan-turing-institute/sktime/pull/2100, which picked up the issue.

### DIFF
--- a/sktime/forecasting/theta.py
+++ b/sktime/forecasting/theta.py
@@ -131,8 +131,12 @@ class ThetaForecaster(ExponentialSmoothing):
         super(ThetaForecaster, self)._fit(y, fh=fh)
         self.initial_level_ = self._fitted_forecaster.params["smoothing_level"]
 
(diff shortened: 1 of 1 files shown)


fkiraly:
> General question, shouldn't we then have a `n_timepoints = len(y)` in `_compute_drift` as well?

Not sure whether I understand - `_compute_drift` is called in `predict` since it needs to access the `fh` which can be passed as late as `predict`. At this time point, the `y` passed in fit is in `self._y`, so seems fine to me?

fkiraly:
bugfix, no review activity since two weeks. Merging.

PR #2112, opened 2022-02-21
[ENH] Proximity forest minor fixes

TonyBagnall:
goes someway to fixing #1715, and has not changed the constructor, so no need for deprecation, but does not solve the pickling issue. Pickling is tricky, because its failing due to two locally defined functions which are hard to untangle without a major restructuring. This is a big job I'm not willing to take on, so this is an interim measure. If anyone wants to take it on, be my guest!


fkiraly:
Do you have a link to a run where the pickling error occurs, or a printout?

sktime/classification/distance_based/_proximity_forest.py
```diff
@@ -1003,6 +1001,23 @@ def _predict_proba(self, X):
...
+        Returns
+        -------
+        params : dict or list of dict, default = {}
+            Parameters to create testing instances of the class
+            Each dict are parameters to construct an "interesting" test instance, i.e.,
+            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+            `create_test_instance` uses the first (or only) dictionary in `params`
+        """
+        params = {
+            "random_state": 0,
```
fkiraly, line 1017: should the `random_state` not be controlled in the tests? E.g., if the tests want to check stochastic behaviour, this would override that. Many (refactored) tests set the `random_state`. More generally, should all tests do that?
Bit off-topic comment, perhaps sth for an issue or discussion.

fkiraly: relevant fact for this PR is, I think none of the other estimators sets `random_state` in `get_test_params`, or only a small minority (consistency issue)

fkiraly (changes requested):
Looks reasonable and straightforward.

I need to request one change, since the change to the constructor makes it non-interface-compliant. The `random_state` is set to `0` by default, which makes a default instantiation of the estimator deterministic. However, default instantiation should be "independent sample", as in `sklearn`.

What should happen is that `random_state` has a `None` default, and behaves randomly in that case.

fkiraly:
Did you close this by accident?
I think this contains useful fixes.

TonyBagnall:
no, I have time for it, if you want to take it over please do, I have other stuff to do,. I an really am not a fan of this code, so knock yourself out

fkiraly (changes requested):
Main change request (blocking) is about `random_state`, default should be `None`.

While you are at it, you could carry out the following, non-blocking refactor steps:

* remove `check_X_y` and `check_X` from inside the inner functinos, this is now all base class business and already done there (more generally and uniformly)
* replace `from_nested_to_2d_array` with the corresponding `datatypes.convert` call, `convert(instance_x, from_type="nested_univ", to_type="numpyflat")`
* alternatively, you could try to use the numpy3D compatible distances directly, and try to use numpy internally, e.g., by setting `X_inner_mtype="numpy3D`. That would be a bigger refactor though.

TonyBagnall:
@fkiraly I cant see check_X and check_X_y anywhere, its not imported into proximity forest. I tried converting inner type to numpy3d, but there are too many uses of dataframe operations to make it completely trivial. I also tried using convert but it doesnt work, it wont create a 2D array from a single instance in a dataframe. I'll leave it as is for now, perhaps we can persuade someone to convert the inner type to numpy as with the rest and sort out the pickling. Would like to get this merged asap 
image

fkiraly (changes requested):
Ok, thanks!

Resolved points:
* it seems `check_X` etc are indeed not used? Not sure why I thought so. Or did you remove them?
* `random_state` looks as it should be.

Some questions:

* you mention the `test_fit_does_not_overwrite_hyper_params` is covered in #2490, but it does not seem to be. `test_fit_does_not_overwrite_hyper_params` is not pickling based, while in #2490 you say the reason for failure is pickling? Could you kindly explain?
* `__init__` does not need a docstring, the parameters should be described in the class docstring. You seem to have both, and the two docstrings are inconsistent.
* re conversion, did you try this, and it did not work?
> replace `from_nested_to_2d_array` with the corresponding `datatypes.convert` call, `convert(instance_x, from_type="nested_univ", to_type="numpyflat")`

TonyBagnall:
>it seems check_X etc are indeed not used? Not sure why I thought so. Or did you remove them?

removed it in the first conversion a loooong time ago, not sure where you got that from, maybe looking at a different PR.


>you mention the test_fit_does_not_overwrite_hyper_params is covered in https://github.com/alan-turing-institute/sktime/issues/2490, but it does not seem to be. test_fit_does_not_overwrite_hyper_params is not pickling based, while in https://github.com/alan-turing-institute/sktime/issues/2490 you say the reason for failure is pickling? Could you kindly explain?

hmm, I cant remember what the problem was there or why this is failing, I will take another look. 

>__init__ does not need a docstring, the parameters should be described in the class docstring. You seem to have both, and the two docstrings are inconsistent.

yep, just legacy docs left there, I've removed them

>re conversion, did you try this, and it did not work?
>replace from_nested_to_2d_array with the corresponding datatypes.convert call, convert(instance_x, from_type="nested_univ", >to_type="numpyflat") 

yes, this code straight swap, 
image

gives this error
image

fkiraly:
ok, indeed it is a pickling problem. Then we have all discussion points resolved now.

fkiraly (dismissed):

fkiraly:
> yes, this code straight swap,
> gives this error

That should 1:1 map to the function `from_nested_to_2d_np_array`. The error you get may come from somewhere else? For instance, why did you change the `return` statement, that does not look like an 1:1 swap.

fkiraly:
I found hte bug related to `convert`: direct conversions were overwritten by indirect conversions
https://github.com/alan-turing-institute/sktime/pull/2517

Although, even if indirect conversions are used, this should pass.
Could you tell me what `instance_a` is, once it reaches the conversion function?

We don´t need to fix it in this PR, but it may indicate there is a bug somewhere.

fkiraly:
pushed the `convert` change, and merged from main, for convenience - feel free to revert if you don´t like.

TonyBagnall:
> pushed the `convert` change, and merged from main, for convenience - feel free to revert if you don´t like.

could you revert this please? You have broken something

fkiraly:
I have reverted it to the pre-`convert` state, but I do not think the problem is on my plate.

The error that you posted to me on slack is coming from infinite recursion depth being the standard parameter of `ProximityForest`.

This PR https://github.com/alan-turing-institute/sktime/pull/2518 has the combination of this PR and the `convert` fix, and all tests run.

fkiraly (approved):
As discussed on slack, this is an improvement - some tests fail and are escaped due to pickling, but that´s better than no tests.

Runtime issues persist, so we should probably open an issue so it´s recorded, but not spend too much time on it anymore right now?

TonyBagnall:
> As discussed on slack, this is an improvement - some tests fail and are escaped due to pickling, but that´s better than no tests.
> 
> Runtime issues persist, so we should probably open an issue so it´s recorded, but not spend too much time on it anymore right now?

yep, if you approve it when clear I will merge it and update the PF issue

fkiraly:
>yep, if you approve it when clear I will merge it and update the PF issue

I've already approved this, see above

Issue #2044, opened 2022-02-11
[ENH] refactor `get_time_index` for hierarchical forecasting


fkiraly:
For hierarchical forecasting, we need to be able to update the `cutoff` in `BaseForecaster` based on `Panel` and `Hierarchical` inputs. The current `cutoff` update logic only deals with the case of `Series`.

For this, the idea to use `get_time_index` or a similar function, and replace the current `cutoff` update logic with a downwards compatible utility.

That utility should be designed and implemented.

PR #2115, opened 2022-02-23
[ENH] cutoff getter for Series, Panel, and Hierarchical mtypes

fkiraly:
This is an alternative approach to the `get_time_index` issue #2044.

Instead of a function that gets the entire time index - which is ambiguous for `Panel` data since different instances may have different index sets - this PR adds a function `get_cutoff` which returns the latest time index observed in the data.

The index is returned as a length 1 `pd.Index`, which can be indexed to get the value.
The function `get_cutoff` also allows to input an existing `cutoff` which shifts the `cutoff` in case a `np.ndarray` format is input.

This PR also adds replaces the logic of `_set_cutoff_from_y` in `BaseForecaster` with `get_cutoff`, which is hence upgraded to `Panel` and `Hierarchical` data container types.

FYI @danbartl, does this work for you?

Update: the original approach is failing tests since somewhere, length zero series `y` are generated as args to `fit` or `update`. I don't understand why this happens, and feel it should probably not. Raised a bug report https://github.com/alan-turing-institute/sktime/issues/2118. For now, it's treated as "no update to `cutoff` happens if length zero `y` is passed".


danbartl:
hi thx,  this works for the base class
But in the reduce function I also need the total length of the time series currently, primarily to check whether the window length exceeds the time series length. Guess I could rework that somehow however. 

Can I read somewhere about those inhomogeneous time index sets and what that means? At least when searching for inhomogeneous in the repo I do not find anything. Does it mean we have multivariate data of different lengths in one data set?

fkiraly:
> Does it mean we have multivariate data of different lengths in one data set?

Different length, or different index set.

Where exactly do you need the entire index set?
Can you post a link to your PR?

I have two questions:
* the "index set" is not fully defined if you have inhomogenous index sets in panels - you have multiple choices (one per series at worst). So, what should it be in that case, or do we not need it then?
* What do we need it for?

PR #2122, opened 2022-02-23
[DOC] enhance distance doc strings

### LINKED ISSUE #1857: [DOC] Formulae in `distances` module seem incorrect

TonyBagnall:
something in differences and clusterers needs further investigation. Keeping this PR because I think it fixes #1857, but I'm sure Franz will have something to say on that. 

Note that, clustering and classifying with derivative distances using these functions is very inefficient, since it will repeatedly just take the same differences. It makes much more sense to just take the differences once during fit and predict, so will move to that soon.


chrisholder (approved):

fkiraly:
@TonyBagnall, sorry, didn´t see this earlier. Changes look good, *but* :-) for the formulae to display correctly in the docs, I think you need to use `:math:` (and then accents) instead of TeX dollar.

TonyBagnall:
yeah I have several other doc strings to adapt with distances, I will iteratively work it out :)

PR #2130, opened 2022-02-24
[ENH] datatypes support for interval and quantile based probabilistic predictions

fkiraly:
This PR introduces the `Proba` scitype to the `datatypes` module.

This covers checks, and conversions between, partial and complete representations of probability distributions.
For now, only prediction intervals and collections of quantile predictions are supported.

Looking forward, this can:
* be used to add input conversions to probabilistic losses/metrics like in #1936 so it can be applied to more input types, e.g., having a pinball loss that can be invoked for *both* interval and quantile predictions
* be used as input checks in probabilistic losses/metrics like in #1936
* be called from default implementations of the probabilistic prediction functions, where defaults leverage implemented other modes of prediction
* be used in tests of `predict_interval` etc to check whether the output is of the right format
* be extended to variance and full distribution predictions as in #2100


fkiraly:
merging due to inactivity & 14 day rule.

PR #2141, opened 2022-02-25
[ENH] test framework: using pytest-xdist with 2 workers

fkiraly:
This PR adds test parallelization via `pytest-xdist`, with two workers.

See https://github.com/alan-turing-institute/sktime/issues/2053 for further experimentation with the test framework.


fkiraly:

Issue #2066, opened 2022-02-15
[BUG] Sporadic test failure in `test_plotting`, of `_tkinter.TclError` on Windows, Python 3.7 CI/CD


khrapovs:
**Describe the bug**

Some tests sometimes fail. Here are a few lines from the latest failure (https://github.com/alan-turing-institute/sktime/runs/5180703321?check_suite_focus=true)
```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <tkinter.Tk object .>, screenName = None, baseName = '__main__'
className = 'matplotlib', useTk = 1, sync = 0, use = None

    def __init__(self, screenName=None, baseName=None, className='Tk',
                 useTk=1, sync=0, use=None):
        """Return a new Toplevel widget on screen SCREENNAME. A new Tcl interpreter will
        be created. BASENAME will be used for the identification of the profile file (see
        readprofile).
        It is constructed from sys.argv[0] without extensions if None is given. CLASSNAME
        is the name of the widget class."""
        self.master = None
        self.children = {}
        self._tkloaded = 0
        # to avoid recursions in the getattr code in case of failure, we
        # ensure that self.tk is always _something_.
        self.tk = None
        if baseName is None:
            import os
            baseName = os.path.basename(sys.argv[0])
            baseName, ext = os.path.splitext(baseName)
            if ext not in ('.py', '.pyc'):
                baseName = baseName + ext
        interactive = 0
>       self.tk = _tkinter.create(screenName, baseName, className, interactive, wantobjects, useTk, sync, use)
E       _tkinter.TclError: invalid command name "tcl_findLibrary"

baseName   = '__main__'
className  = 'matplotlib'
ext        = '.py'
interactive = 0
os         = <module 'os' from 'C:\\Miniconda\\envs\\test\\lib\\os.py'>
screenName = None
self       = <tkinter.Tk object .>
sync       = 0
use        = None
useTk      = 1

C:\Miniconda\envs\test\lib\tkinter\__init__.py:2023: TclError
```

**To Reproduce**

Does not seem possible to reproduce.

**Expected behavior**

CI/CD pipeline passes.

**Versions**
<details>

```
from sktime import show_versions; show_versions()
System:
    python: 3.8.11 (default, Aug  6 2021, 10:28:10)  [GCC 10.3.0]
[...]

PR #2145, opened 2022-02-27
[MNT] remove pytest-xdist to diagnose #2066

fkiraly:
The bug in https://github.com/alan-turing-institute/sktime/issues/2066 may have become more frequent since `pytest-xdist` was added.

This PR checks the hypothesis by removing `pytest-xdist`.

Note to reviewers: `main` currently has #2144 which adds `xfails` to the problematic tests.
This is currently not merged to this PR, and merge will mask overall pass/fails below.


fkiraly:
closing - PR has served its diagnostic purpose

PR #2152, opened 2022-03-01
[BUG]: changed how derivative based distance metric are handled

chrisholder:
#### Reference Issues/PRs

Bug fix for using derivative based distances in kmeans and kmedoids.

Tony adds: I have changed the averaging transformation from happening in an explicit transformer to happening in a method in ddtw_distance. The transposing also happens now in the transformation method. This is a precursor to refactoring distances to avoid the transformation, not to be merged just yet, I am just synching the code

TLDR:
- Derivaitve transformer to take derivative.
- Conditional checks in kmeans and kmedoids for metric. If using a derivative to use the transformer and then pass to dtw distance.

Derivative based distances were performing poorly which we found was due to how the derivative for panel data was being transformed (or not). As such while ddtw 'works' to compute the distance between two time series, however, when using the factory over a data set some steps are skipped (mainly due to efficiency). As such a derviative transformer has been added to take the derivative of panel data and then passed to regular dtw. This should solve the bugs we were having with it and additionally improve performance.

##### For all contributions

##### For new estimators


sktime/transformations/panel/derivative.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+"""Derivative transformer."""
+import numpy as np
+
+from sktime.distances._ddtw import _average_of_slope
+from sktime.transformations.base import BaseTransformer
+
+
+class DerivativeTransformer(BaseTransformer):
```
fkiraly, line 9: I would recommend not to give this name to the transformer - derivative has a common meaning in the time series domain, which is sometimes used interchangeably with the discrete version of it indicated by the `d` in ARIMA.

The formula for that `d`-erivative (or `d`-ifferences) is different from the funny "average of slope" used here, the coefficients are (1, -2, 1) and not (0.25, 0.5, -0.75).

sktime/transformations/panel/tests/test_derivative.py
```diff
@@ -0,0 +1,15 @@
+# -*- coding: utf-8 -*-
+"""Test derivative transformer."""
+import numpy as np
+
+from sktime.distances.tests._utils import create_test_distance_numpy
+from sktime.transformations.panel.derivative import DerivativeTransformer
+
+
+def test_derivative():
```
fkiraly, line 9: is this not already being tested by the generic tests?

fkiraly (changes requested):
Fine in-principle, though I have concerns about the name "derivative" (above), we should change (or discuss why it might be fine).

fkiraly:
PS: which bug does this precisely fix?

sktime/transformations/panel/tests/test_derivative.py
chrisholder: I didn't realise there was generic testing going on in this module Ill remove this then and leave it to that!

PR #2153, opened 2022-03-02
[ENH] DRAFT - improved (?) design for probabilistic error metrics

fkiraly:
Improvement upon the probabilistic error metric design.

Relies on #2130 for input checks and conversions.

Added some comments and todos in the code, it doesn't run since some content is missing in the base class.

FYI @eenticott-shell.


sktime/performance_metrics/forecasting/probabilistic/_classes.py
```diff
@@ -55,25 +59,31 @@ def __call__(self, y_true, y_pred, **kwargs):
...
+        # Input checks and conversions
+        y_true_inner, y_pred_inner, multioutput = self._check_ys(
+            y_true, y_pred, multioutput
+        )
+        # pass to inner function
+        return self._evaluate(y_true_inner, y_pred_inner, multioutput, **kwargs)
 
     def _evaluate(self, y_true, y_pred, multioutput, **kwargs):
+        # todo: should have a default implementation, which is just the mean over
+        #       _evaluate_by_index
```
eenticott-shell, line 73: I have done this, however as the default implementations both end up calling each other, at least one of them needs to be implemented manually to avoid a recursion error (I've added a check for this case.)

sktime/performance_metrics/forecasting/probabilistic/_classes.py
```diff
@@ -85,8 +95,42 @@ def _evaluate_by_index(self, y_true, y_pred, multioutput, **kwargs):
             )
         return out_series
 
+    def _check_alpha(self, alpha):
```
eenticott-shell, line 98: Not sure this is needed anymore. Alphas are now retrieved directly from the y_pred dataframe rather than being user inputs, so there should never be an invalid alpha.

sktime/performance_metrics/forecasting/probabilistic/_classes.py
```diff
@@ -85,8 +95,42 @@ def _evaluate_by_index(self, y_true, y_pred, multioutput, **kwargs):
...
+            if not ((alpha > 0) & (alpha < 1)):
+                raise ValueError("Alpha must be between 0 and 1.")
+            return [alpha]
+        raise TypeError("Alpha should be a float or numpy array")
 
-class PinballLoss(_BaseProbForecastingErrorMetric):
+    def _check_ys(self, y_true, y_pred, multioutput):
+        if multioutput is None:
+            multioutput = self.multioutput
+        _, y_true, y_pred, multioutput = _check_reg_targets(y_true, y_pred, multioutput)
```
eenticott-shell, line 116: This check no longer works. It used to work as I would check it column by column rather than all at once, but this is inefficient and I've now made my own version which implements similar checks.

sktime/performance_metrics/forecasting/probabilistic/_classes.py
```diff
@@ -85,8 +95,42 @@ def _evaluate_by_index(self, y_true, y_pred, multioutput, **kwargs):
...
+        valid, msg, metadata = check_is_scitype(
+            y_pred, scitype="Proba", return_metadata=True, var_name="y_pred"
+        )
+        if not valid:
+            raise TypeError(msg)
+        y_pred_mtype = metadata["mtype"]
+        inner_y_pred_mtype = self.get_tag("scitype:y_pred")
+        y_pred_inner = convert(
```
eenticott-shell, line 124: This is extremely useful. One thing that popped up in testing was that if I am converting quantiles -> intervals and there is a 0.5 quantile, it produces a column for 0 coverage. This is redundant so it may be worth adding a case distinction to not include this column when quantile is 0.5.

eenticott-shell: Also as far as I could work out, the conversions assume that coverage is using symmetric quantiles, i.e. 0.9 coverage gives quantiles for 0.05 and 0.95 but it could correspond to 0.01 and 0.91 quantiles?

eenticott-shell: Is there ever a case where a half interval is useful? Sometimes the conversion from quantiles to intervals can only find a lower bound, is this useful on its own or should  incomplete intervals be removed?

sktime/performance_metrics/forecasting/probabilistic/_classes.py
fkiraly: correct - something similar happens for `predict_interval` etc in the `BaseForecaster`, and in the design here:
https://github.com/alan-turing-institute/sktime/pull/2100
if one is implemented, the others do sth sensible; if none is, an error message is raised. You can use `_has_implementation` if you are inheriting from `BaseObject`

sktime/performance_metrics/forecasting/probabilistic/_classes.py
fkiraly: I guess - and we also do input checks on `y_pred`, which I think also checks the alpha?

sktime/performance_metrics/forecasting/probabilistic/_classes.py
fkiraly: > Is there ever a case where a half interval is useful?

I think the only "frequent" use case is applying a quantile loss to predictive intervals, the other way round does not seem sensible (although I included it for completeness). Similarly, the only conversion that I see being used frequently is from intervals to quantiles, not the other way round.

fkiraly: >Also as far as I could work out, the conversions assume that coverage is using symmetric quantiles, i.e. 0.9 coverage gives quantiles for 0.05 and 0.95 but it could correspond to 0.01 and 0.91 quantiles?

Do they? I thought I changed that in the PR, then you get "half intervals". Only on `main`, symmetric quantiles are still assumed.

fkiraly: > This is redundant so it may be worth adding a case distinction to not include this column when quantile is 0.5.

Well, it´s a degenerate case, and anyway converting from intervals to quantiles is the more frequent case. In that, you usually do not have a "zero width" interval.

fkiraly:
The design is now implemented in #2232

PR #2160, opened 2022-03-03
[DOC] add scitype/mtype register pointers to docstrings in datatypes

fkiraly:
This PR adds pointers to the scitype and mtype registers, which contain the list of valid scitype and mtype strings, to all docstrings in the `datatypes` module that have arguments which need to be valid scitype or mtype strings.

In addition, this PR makes the `scitype` argument of `get_examples` optional, inferring `scitype` from `mtype`.


fkiraly:
merging via 14 day rule

PR #1630, opened 2021-11-19
[ENH] Probability threshold early classifier

MatthewMiddlehurst:
#1616

First draft early classifier aiming deviate from the regular classification API as little as possible.

Contains:
- A new classification tag for early classifiers.
- A Package for early classifiers and early classifier tests
- A probability threshold early classifier, which can make predictions using an incomplete series at certain `classification_points`. A decision on the trustworthiness of these predictions is made in `decide_prediction_safety`, using a prediction probability threshold and optionally a required number of consecutive positive decisions.

Issue #1616, opened 2021-11-12
[ENH] Design and API for Early Classification


MatthewMiddlehurst:
Early classification revolves around allowing classifiers to take incomplete series (i.e, the first half of a series recorded over a day) and decide whether it can make an accurate prediction or not with the presented partial time series. These algorithms are similar enough to the rest of the classification package that they could just reside in a sub-package for now, such as `classification/early_classification`.

While most of the current classification API is suitable for these algorithms, some issues may arise and will require discussion: 
- How do we deal with an algorithm refusing to make a prediction? Both in the sense of what would we actually return (`None` seems reasonable) and how would we go about testing, as returning this is likely to break some of the standard classifier ones.
- How do we verify input when predicting? Some algorithms can only make decisions at specific series lengths (commonly in 5% intervals), do any additions need to be made for this, or should it be left to the individual algorithm? 
- Some algorithms keep data from previous prediction attempts, i.e. a decision at 15% of the series may use information from the decisions at 5% and 10%. Do we require any additional predict inputs to accommodate this? How do we reset for new data when predicting?
- Do we require an extended base class for any of this? Would having it anyway help with testing, or would it be better as a tag?

Good first implementations would be [TEASER](https://link.springer.com/article/10.1007/s10618-020-00690-z) and a simple probability threshold for making the decision.

MatthewMiddlehurst:
@patrickzib has shown interest in the topic from previous discussions, though anyone feel free to contribute and edit the issue if necessary.

MatthewMiddlehurst:
I have put up a draft PR #1630 containing probability threshold early classifier as a test. I have set it up to follow the current classification API as much as possible. We can completely revamp any part of it as we make decisions, but it may require an extended base class to be implemented.

To make sure the current tests pass and that we can use BaseClassifier, a second function is used to determine whether the predictions/probabilities are safe.

> Interesting question! :) I would rather say, one should keep the classifier state-less

Agreed, currently I have set up the decision method to take and return a tuple of values containing the current state. The second option could get very expensive depending on the algorithm.

MatthewMiddlehurst:
Any objections to closing this issue? Any problems with the early classification base class or early classifiers should be raised as a separate issue, IMO.

fkiraly:
no, I thought we arrived at a consolidated design through discussions on the early classification PR.
Perhaps just link those discussions here, then close?

MatthewMiddlehurst:
That's fair.

[...]

PR #2162, opened 2022-03-04
[ENH] TEASER early classification implementation

MatthewMiddlehurst:
Implementation of the TEASER early classification algorithm (@patrickzib)

Related to #1616 and #1630.


sktime/classification/early_classification/_teaser.py
```diff
@@ -0,0 +1,423 @@
...
+        # tune the one class svm gamma
+        best_svm = None
+        best_acc = -1
+        if len(X_svm) > 1:
+            cv_size = min(len(X_svm), 10)
+
+            for gamma in self._svm_gammas:
+                svm = OneClassSVM(gamma=gamma)
+                threads = 1 if m is None else self._threads_to_use
+                svm_preds = cross_val_predict(svm, X_svm, cv=cv_size, n_jobs=threads)
```
patrickzib, line 397: Not sure, but isn't this a GridSearchCV?


ocSVM = OneClassSVM(kernel="rbf", tol=1e-4)
parameters = {'gamma':self.SVM_GAMMAS,
                              'nu': self.SVM_NUS}
clf = GridSearchCV(ocSVM, parameters, scoring="accuracy")
clf.fit(confidences, correctPredictions)
self.masterClf[i] = clf.best_estimator_

sktime/classification/early_classification/_teaser.py
```diff
@@ -0,0 +1,423 @@
...
+            state_info = [(0, 0, 0) for _ in range(n_instances)]
+            finished = [False for _ in range(n_instances)]
+
+            for i in range(len(self._classification_points)):
+                if i == len(self._classification_points) - 1:
+                    decisions = [True for _ in range(n_instances)]
+                elif self._svms[i] is not None:
+                    decisions = self._svms[i].predict(X_svm[i]) == 1
+                else:
+                    decisions = [False for _ in range(n_instances)]
```
patrickzib, line 159: Maybe initialize decisions to False in line 153?
 (instead of having one branch in the if  ... else ...?

sktime/classification/early_classification/_teaser.py
```diff
@@ -0,0 +1,423 @@
...
+            for i in range(len(X)):
+                for n in range(self.n_classes_):
+                    if n != preds[i]:
+                        X_svm[i][self.n_classes_] = min(
+                            X_svm[i][self.n_classes_], X_svm[i][preds[i]] - X_svm[i][n]
+                        )
+
+            decisions = self._svms[idx].predict(X_svm) == 1
+        else:
+            decisions = [False for _ in range(n_instances)]
```
patrickzib, line 314: Same: Maybe initialize decisions to False instead of branching

sktime/classification/early_classification/_teaser.py
```diff
@@ -0,0 +1,423 @@
...
+            )
+            for i in range(len(self._classification_points))
+        )
+
+        self._estimators, self._svms, X_svm, train_preds = zip(*fit)
+
+        # tune consecutive predictions required to best harmonic mean
+        best_hm = -1
+        for g in range(2, min(6, len(self._classification_points))):
+            state_info = [(0, 0, 0) for _ in range(n_instances)]
```
patrickzib, line 150: Not sure, if I understand the state_info correctly. Maybe encapsulate it into some methods and add comments?

I personally would have used lists of dicts() - not claiming this works better

sktime/classification/early_classification/_teaser.py
```diff
@@ -0,0 +1,423 @@
...
+
+        m = getattr(self._estimators[idx], "predict_proba", None)
+        if callable(m):
+            return self._estimators[idx].predict_proba(X)
+        else:
+            dists = np.zeros((X.shape[0], self.n_classes_))
+            preds = self._estimators[idx].predict(X)
+            for i in range(0, X.shape[0]):
+                dists[i, self._class_dictionary[preds[i]]] = 1
+            return dists
```
patrickzib, line 235: Not sure, if I understand correctly. Isnt a dist the inverse of a probability? i.e. higher is worse.

Maybe consider using: CalibratedClassifierCV() ?

patrickzib (changes requested):
Thank you so much. Very cool! I had a brief check. I might continue on monday

sktime/classification/early_classification/_teaser.py
MatthewMiddlehurst: Holdover from tsml I have ingrained, I think. Short for probability distributions. Will rename them to probas.

We didn't have much success with probability calibration when we looked into it with HIVE-COTE. Classifiers which don't produce probability estimates such as ROCKET did not perform better with any of the techniques we tried. 

That sounds like a wrapper for an estimator? The user can pass it as the base estimator for TEASER if required. If you think it improves WEASEL/MUSE then we can use it as a default.

sktime/classification/early_classification/_teaser.py
MatthewMiddlehurst: Don't think it's necessary to create the list twice, setting everything else should not happen often. But it's a minor optimisation either way.

sktime/classification/early_classification/_teaser.py
MatthewMiddlehurst: Same as above.

sktime/classification/early_classification/_teaser.py
MatthewMiddlehurst: It's mostly modelled after the state_info from decide, but I could split it up. Need to store the current time stamp used for earliness, the number of consecutive decisions made and the class prediction. 

Will add comments explaining it a bit more for now.

sktime/classification/early_classification/_teaser.py
```diff
@@ -0,0 +1,423 @@
...
+        )
+        # remove duplicates
+        self._classification_points = list(set(self._classification_points))
+        self._classification_points.sort()
+        # remove classification points that are less than 3
+        self._classification_points = [i for i in self._classification_points if i >= 3]
+        # make sure the full series length is included
+        if self._classification_points[-1] != series_length:
+            self._classification_points.append(series_length)
+        # create dictionary of classification point indicies
```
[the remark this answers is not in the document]
MatthewMiddlehurst: Fixed the typo, stores the index for each time stamp used i.e for the test {6 : 0, 10 : 1, 16 : 2, 24 : 3}.

sktime/classification/early_classification/_teaser.py
```diff
@@ -0,0 +1,423 @@
...
+            Make a prediction using the series length at classification point i.
+            Decide whether the predcition is safe or not using decide_prediction_safety.
+
+    Parameters
+    ----------
+    estimator: sktime classifier, default=None
+        An sktime estimator to be built using the transformed data. Defaults to a
+        WEASEL classifier.
+    classification_points : List or None, default=None
+        List of integer time series thresholds to build classifiers and allow
```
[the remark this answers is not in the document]
MatthewMiddlehurst: Changed to time stamps.

sktime/classification/early_classification/_teaser.py
```diff
@@ -0,0 +1,423 @@
...
+
+        self._classification_points = (
+            copy.deepcopy(self.classification_points)
+            if self.classification_points is not None
+            else [round(series_length / i) for i in range(1, 21)]
+        )
+        # remove duplicates
+        self._classification_points = list(set(self._classification_points))
+        self._classification_points.sort()
+        # remove classification points that are less than 3
```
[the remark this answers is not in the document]
MatthewMiddlehurst: Fixed

sktime/classification/early_classification/_teaser.py
```diff
@@ -0,0 +1,423 @@
...
+        self.n_jobs = n_jobs
+        self.random_state = random_state
+
+        self._estimators = []
+        self._classification_points = []
+        self._svms = []
+        self._consecutive_predictions = 0
+
+        self._svm_gammas = [100, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1.5, 1]
+
```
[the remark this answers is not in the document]
MatthewMiddlehurst: Added this as a parameter, passed to the OneClassSVM. Good catch.

sktime/classification/early_classification/_teaser.py
```diff
@@ -0,0 +1,423 @@
...
+                X_svm.append(train_probs[i])
+
+        # tune the one class svm gamma
+        best_svm = None
+        best_acc = -1
+        if len(X_svm) > 1:
+            cv_size = min(len(X_svm), 10)
+
+            for gamma in self._svm_gammas:
+                svm = OneClassSVM(gamma=gamma)
```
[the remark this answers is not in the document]
MatthewMiddlehurst: Done

sktime/classification/early_classification/_teaser.py
```diff
@@ -0,0 +1,423 @@
...
+                    decisions = self._svms[i].predict(X_svm[i]) == 1
+                else:
+                    decisions = [False for _ in range(n_instances)]
+
+                # record consecutive class decisions
+                state_info = [
+                    (
+                        i,
+                        state_info[n][1] + 1
+                        if decisions[n] and train_preds[i][n] == state_info[i][2]
```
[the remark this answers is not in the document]
MatthewMiddlehurst: Added some comments. A bit annoying as you can't split it up when you create lists like that. May all change if we decide to do state_info a bit differently.

To summarise: Add one to consecutive decisions if positive decision and same class. If there is a different class predicted or negative decision, set to 0 if the decision is negative, 1 if its positive but different class.

sktime/classification/early_classification/_teaser.py
MatthewMiddlehurst: Yeah, that would probably be better. Also saw the tolerance is slightly different to the one you gave, new CV would be:

```
cv_size = min(len(X_svm), 10)
gs = GridSearchCV(
    OneClassSVM(tol=self._svm_tol, nu=self._svm_nu),
    {"gamma": self._svm_gammas},
    scoring="accuracy",
    cv=cv_size,
)
gs.fit(X_svm, np.ones(len(X_svm)))
svm = gs.best_estimator_
```

sktime/classification/early_classification/tests/_test_teaser.py
```diff
@@ -0,0 +1,31 @@
...
+    pt = TEASER(
+        random_state=0,
+        classification_points=[6, 10, 16, 24],
+        estimator=TimeSeriesForestClassifier(n_estimators=10, random_state=0),
+    )
+    pt.fit(X_train, y_train)
+
+    X_test = from_nested_to_3d_numpy(X_test)
+    states = None
+    for i in pt.classification_points:
```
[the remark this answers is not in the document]
MatthewMiddlehurst: You are correct, there currently aren't correctness tests for the early classifiers. I can look into adding them.

We would want to compare both the prediction output and earliness of the decision I would assume.

sktime/classification/early_classification/_teaser.py
patrickzib: Yes, CalibratedClassifierCV  comes from: https://scikit-learn.org/stable/modules/generated/sklearn.calibration.CalibratedClassifierCV.html
It will add the `predict_proba` to a classifier.

So, I am not sure, what should predict_proba() return when the estiamtor has no predict_proba available?

patrickzib:
And a general question: how do I resolve a conversation? I can see that I can delete it. Is it supposed to be like that?

MatthewMiddlehurst:
You should have access to the rep as a core dev? Pretty annoying you cant resolve the ones you created even without access.

TonyBagnall:
@patrickzib I've just given you write access, sorry for the oversight.

sktime/classification/early_classification/_teaser.py
MatthewMiddlehurst: Given we have to train a classifier for each selected time stamp, I think adding a cross-validation for each on top of that is a bit much. Did the Java version return valid probabilities?

In tsml (and currently here) we just return 1 for the predicted class as a default method for predict_proba, i.e. [0, 1, 0, 0] if the 2nd class is predicted in predict.

sktime/classification/early_classification/_teaser.py
patrickzib: Ok, but what about this simplification 

```
if self._svms[i] is not None:
    decisions = self._svms[i].predict(X_svm[i]) == 1
else:
    decisions = [i == len(self._classification_points) - 1 for _ in range(n_instances)]
```

patrickzib: This does not create twice but is still more compact

sktime/classification/early_classification/_teaser.py
patrickzib: Same as above  ;)

sktime/classification/early_classification/_teaser.py
patrickzib: Agreed! 

Hmm, TEASER in Java is based on WEASEL, which returns probabilities. 

I am not pushing this, but maybe it would be nice to separate this:

1. If the classifier has a predict_proba(), return the probabilities
2. If the classifier has only a decision_function(), return distances

I think, sklearn handles it like that: https://github.com/scikit-learn/scikit-learn/blob/7e1e6d09b/sklearn/svm/_base.py#L807

But, if it is done like that in sktime so far, I am not pushing it :)

sktime/classification/early_classification/tests/_test_teaser.py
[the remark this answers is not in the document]
patrickzib: For simplicity, average earliness and accuracy would be fine, I assume?

sktime/classification/early_classification/_teaser.py
patrickzib: Same as above :)

sktime/classification/early_classification/_teaser.py
```diff
@@ -0,0 +1,429 @@
...
+        Tune the number of consecutive safe svm predictions required to consider the
+        prediction safe.
+
+        While a prediction is still deemed unsafe:
+            Make a prediction using the series length at classification point i.
+            Decide whether the predcition is safe or not using decide_prediction_safety.
+
+    Parameters
+    ----------
+    estimator: sktime classifier, default=None
```
patrickzib, line 46: As the classifier is parameterized, it might be worth parameterizing the anomaly detector, too. It is a One-Class-SVM but it doesn't have to be. It might as well be any other method available in sklearn such as: IsolationForests, ...

sktime/classification/early_classification/_teaser.py
```diff
@@ -0,0 +1,429 @@
...
+            )
+            hm = (2 * accuracy * earliness) / (accuracy + earliness)
+
+            if hm > best_hm:
+                best_hm = hm
+                self._consecutive_predictions = g
+
+        return self
+
+    def _predict(self, X):
```
patrickzib, line 221: Not sure: what is the expected output of _predict() in early classification? 

As I understand it right now, it is a list of  probabilities for each offset? Not the overall decision after the full length?

I guess, I would rather model it as the overall decision. And have another method, to return the array of decisions.

patrickzib (commented):
Some thoughts regarding the "framework"-aspect and regarding the interface for early classification

sktime/classification/early_classification/_teaser.py
patrickzib: Thinking about my suggestion, maybe it is not the same as your code ...

sktime/classification/early_classification/_teaser.py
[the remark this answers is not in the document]
MatthewMiddlehurst: Done. Slightly different from what is here, but very similar.

sktime/classification/early_classification/tests/_test_teaser.py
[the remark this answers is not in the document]
MatthewMiddlehurst: Added some correctness tests, just compares probabilities obtained when a decision is deemed safe.

sktime/classification/early_classification/_teaser.py
patrickzib: Agreed

sktime/classification/early_classification/_teaser.py
MatthewMiddlehurst: I'm unsure if any of our classifiers have a decision_function() implementation actually, not entirely sure what it does.

It is done similarly to this currently, not saying thats the best way to go about it however. WEASEL has predict_proba so not a big issue by default, I think if you change it to an estimator without predict_proba that's the user's issue 🙂.

sktime/classification/early_classification/_teaser.py
MatthewMiddlehurst: Done.

sktime/classification/early_classification/_teaser.py
MatthewMiddlehurst: Unfortunately, we are sort of constrained by the sktime classifier interface unless we want to create new stuff specifically for early classification. It definitely could be better, but would require a new base class and set of tests most likely.

_predict() is required and has to return a list of class predictions, currently the only different for early classifiers is that it can return predictions for different series lengths.

_predict_proba() returns a list of probabilties for the input, assuming it is a valid input length. We leave the decision on whether those are safe or not to decide_prediction_safety().

sktime/classification/early_classification/_teaser.py
patrickzib: Rocket with RidgeClassifierCV, for example, has only the decision function... :) 

decision_function() predicts confidence scores  in sklearn - i.e. it simply returns the distance. 

I agree that we do not have to over-design the issue. Is it possible to check, if a classifier passed to TEASER has the predict_proba method as a prerequisite?

patrickzib: So, eventually check if 
m = getattr(self._estimators[idx], "predict_proba", None)
is true and given an error if not.

sktime/classification/early_classification/_teaser.py
patrickzib: Hmm, still the contract of _predict_proba() is not clear to me. 

`idx = self._classification_point_dictionary.get(series_length, -1)`
I am not sure, if I understand the code correctly. It reads to me like `_predict_proba` takes the full length of the time series and returns the probability for that full length - am I correct about that?


I would expect _predict_proba() to probe the time series at the known offsets and return the probability of the time stamp where the earliest decision taken out. And there should be some method to get this time stamp.

MatthewMiddlehurst: If your time stamps are [6, 10, 16, 24], _predict_proba() will accept series of those lengths then return the probability from the corresponding estimator. `idx = self._classification_point_dictionary.get(series_length, -1)` just checks which of [6, 10, 16, 24] the input series length is, or returns -1 if it's none of them so an error can be thrown.

We can't edit _predict_proba() to return null/a decision because of the classifier interface/unit tests I'm pretty sure. That's why it's done as a separate method the user has to call.

The problem with checking all the offsets at once is it requires the full series, and if you require the full series you can't really do early classification in practice.

sktime/classification/early_classification/_teaser.py
MatthewMiddlehurst: Probably a safer option that just returning bad probabilities, I agree. Will include it in the next push.

sktime/classification/early_classification/_teaser.py
patrickzib: I have a different understanding of `_predict_proba()`:

Assuming we have given
(a) time stamps: [6, 10, 16, 24]  and
(b) time series has length 30. 

Then we probe all offsets 6, 10, 16, 24 and return the **first index** where the decision would be taken out. I.e. 3 consecutive positive votes.

If no decision is taken out earlier than the last index (24), we could 
(a) return the probability of the classifier at the last index (ndex=24) or 
(b) simply 0 for all classes.

For that you do not need to return null.

sktime/classification/early_classification/_teaser.py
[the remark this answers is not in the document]
patrickzib: Thank you! Sorry for bothering again :) Should this be used in line 327, too? Or is this code different?

sktime/classification/early_classification/_teaser.py
MatthewMiddlehurst: Are we assuming 30 is the final length of the series? Can you also input a set of series of length 16 and also get probabilities?

If yes: You are doing a lot of repeated work if you want to attempt to make a decision at each successive time point.
If no: You require the full series to make predictions, the classifier is only useful for research purposes.

I may be misunderstanding a bit.

MatthewMiddlehurst: I don't think the current set-up is the end-state we should be aiming for, having it all encapsulated into predict/predict_proba would be best. That would require a little extra time/drive though to set up interfaces and tests. Will add some bits to #1616.

sktime/classification/early_classification/_teaser.py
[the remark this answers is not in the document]
MatthewMiddlehurst: Ah yeah, think it's slightly different but can make it work. Don't worry, not a bother.

sktime/classification/early_classification/_teaser.py
patrickzib: > If yes
Yes :)

> You are doing a lot of repeated work if you want to attempt to make a decision at each successive time point.
Not sure what you mean by that. Yes, if you are assuming a streaming setting, extending the time series and doing predict() after each new bit of information, you will have extra work.

But, there is no point in checking for consecutive votes in fit(), if we only take the last index in predict()...
Thus, imo, there have to be consecutive votes in predict(), too...

MatthewMiddlehurst: The way it is currently set up, the decision-making process is separate from the probability prediction process. All predict/predict_proba does essentially is find the correct WEASEL and make predictions.

`decide_prediction_safety(X, X_probabilities, state_info)` Takes X, the data with a series length contained in [6, 10, 16, 24, 30]. X_probabilities, the result of calling predict_proba() on X. state_info, decision/prediction information from the previous decision made (or None if it's the first/6).

We only have to check the current index, as we save relevant information (the number of consecutive predictions, the last class predicted) in state_info. It is assumed that if you are trying to make a decision on length 16, you have already run it for length 10 and have the state_info.

We probably could put this all in predict_proba, and ideally we would, but currently that would mean we have to repeat predictions of all previous data points and figure out how to return negative decisions (I don't know if we have a test to make sure probabilities sum to 1, for the all 0 suggestion).

MatthewMiddlehurst: so i know i just wrote an essay above, but I'm pretty sure I could make it so that predict_proba() takes an optional state_info i.e. `_predict_proba(X, state_info=None)` and have it only return decisions i.e. (probas, decisions) if a state_info was given or the series length is < the max. Pretty sure that would pass all the tests

MatthewMiddlehurst: would have to write a bit for running on all previous time stamps, however.

patrickzib: > We only have to check the current index, as we save relevant information (the number of consecutive predictions, the last class predicted) in state_info. It is assumed that if you are trying to make a decision on length 16, you have already run it for length 10 and have the state_info.

Sorry again - and glad to help rubber-ducking :)

I do not quite understand the implication of this. `predict()` can be called from outside at any time point by the user.

Thus it would be great to start this process probing all offsets, if there is no `state_info` available, and continue with the next available time stamp if state_info is available :)

MatthewMiddlehurst: Ok, I have redone this and it is all integrated into predict_proba. It is not completely tested yet, will do that tomorrow.

If no state_info is provided, it predicts all time stamps up to the current series_length. If a state info is provided, it uses it and predicts for the current series length only.

Should be noted, it is possible for these to have different outputs currently. Case 1 will stop updating when a decision is made, Case 2 may update its decision even if it considered it safe in a previous call.

patrickzib:
@MatthewMiddlehurst
Is it ok for you, if I push a change into this pull request? Rather than ask you to do some minor changes?

Not sure, if I have the rights to do so, though :)

MatthewMiddlehurst:
Go ahead. You should be able to push to the branch.

Do you want me to add you to the codeowners for the file? Not necessarily an obligation, but will request a review when the file is changed.

sktime/classification/early_classification/_teaser.py
patrickzib: Thank you. I will check the code. I think, passing one `state_info=state_info` was missing, which I fixed. 

I would like to go over the code. Still trying to understand it :)

sktime/classification/early_classification/_teaser.py
```diff
@@ -0,0 +1,551 @@
...
+        # positive decisions made, and 3. the prediction made
+        state_info = [(0, 0, 0) for _ in range(n_instances)]
+        # stores whether we have made a final decision on a prediction, if true
+        # state info wont be edited in later time stamps
+        finished = [False for _ in range(n_instances)]
+
+        for i in range(n_timestamps):
+            if i == len(self._classification_points) - 1:
+                decisions = [True for _ in range(n_instances)]
+            elif self._one_class_classifiers[i] is not None:
```
patrickzib, line 505: I am not sure, if I understand correctly:  The `predictions` are still made in case of `finished == True` but the `state_info` is not updated anymore.

But, we might as well stop predictions?

Thus, shouldn`t it be?

```
                offsets = np.argwhere(finished==False).flatten()
                decisions_subset = self._one_class_classifiers[i].predict(X_oc[i][offsets]) == 1
                decisions = np.ones(n_instances, dtype=bool)
                decisions[offsets] = decisions_subset
```

This will only do predictions for offsets that are not finished.

MatthewMiddlehurst: Yeah, if that works it's better a way of going about it.

Similar could be done for the actual predictions in predict_proba, but unsure if we would be able to reuse the method from _fit.

MatthewMiddlehurst:
Made a push, main change is I reverted how to assign the default estimator/one-class classifier to how it was previously. We can't change the parameter fields (i.e. self.estimator from None to WEASEL), it has to be done locally or in an alternate field (self._estimator).

patrickzib:
> Made a push, main change is I reverted how to assign the default estimator/one-class classifier to how it was previously. We can't change the parameter fields (i.e. self.estimator from None to WEASEL), it has to be done locally or in an alternate field (self._estimator).

Ah ok, I wasn't aware of this :) 

The incremental state-info is still on my list as a TODO.

MatthewMiddlehurst:
I forgot it was a test to be fair. Been doing it like that as a default for long enough I forgot why!

sktime/classification/early_classification/_teaser.py
```diff
@@ -0,0 +1,564 @@
...
+            )
+
+            # if we have the full series, always decide True
+            if idx == len(self._classification_points) - 1:
+                decisions = np.ones(n_instances, dtype=bool)
+            else:
+                decisions = np.array(
+                    [
+                        True
+                        if state_info[i][1] >= self._consecutive_predictions
```
patrickzib, line 369: I am currently refactoring the code. And this is likely wrong? this should be new_state_info[i][1]?

patrickzib (changes requested):
Please check :)

sktime/classification/early_classification/_teaser.py
MatthewMiddlehurst: yes

patrickzib: Thank you :)

sktime/classification/early_classification/_teaser.py
patrickzib: I have refactored the code

patrickzib:
@MatthewMiddlehurst I have refactored the `_predict_proba()` code a little. It will now always do the predictions for all missing indices. This may be just one index or the whole bunch of indices. 

Not sure, how we should continue with the reviews, as I did some changes now? 

As a future PR a jupyter notebook might be nice for early classification.

patrickzib (approved):

patrickzib:
@MatthewMiddlehurst 

Ahh, and one question: right now, there will always be a label predicted. This is intended, right? A different option would be to return a default undecided label such as `-1`, if the decision has not be carried out. Same as `DBSCAN` which returns `-1` for non-cluster elements (noise).

sktime/classification/early_classification/_teaser.py
```diff
@@ -0,0 +1,560 @@
...
+
+        return self._predict_proba(X, state_info=state_info)
+
+    def _predict_proba(self, X, state_info=None):
+        n_instances, _, series_length = X.shape
+
+        # maybe use the largest index that is smaller than the series length
+        next_idx = self._get_next_idx(series_length)
+
+        if next_idx == -1:
```
MatthewMiddlehurst, line 296: I'm unsure why this is required instead of the dictionary. Is next_idx supposed to mean the index of the timestamp after the current input?

I don't think it can be -1 also.

sktime/classification/early_classification/_teaser.py
```diff
@@ -0,0 +1,560 @@
...
+
+        # record consecutive class decisions
+        state_info = np.array(
+            [
+                self._update_state_info(
+                    decision_needed, estimator_preds, state_info, i, idx
+                )
+                for i in range(n_instances)
+            ]
+        )
```
MatthewMiddlehurst, line 508: Do we want to update state_info for cases we have already made a decision on? I feel like we lose useful information doing so.

sktime/classification/early_classification/_teaser.py
```diff
@@ -0,0 +1,560 @@
...
+                i,
+            )
+            for i in range(len(self._classification_points))
+        )
+
+        self._estimators, self._one_class_classifiers, X_oc, train_preds = zip(*fit)
+
+        # tune consecutive predictions required to best harmonic mean
+        best_hm = -1
+        for g in range(2, min(6, len(self._classification_points) + 1)):
```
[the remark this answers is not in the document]
patrickzib: I wasn't sure about that part. i had added it when debugging. But maybe you are right.

sktime/classification/early_classification/_teaser.py
patrickzib: Oh, you are right: The problem is, that if we add 
`if finished[i] else state_info[i]`

we get the error:

> ValueError: All state_info input instances must be from the previous classification point series length.

sktime/classification/early_classification/_teaser.py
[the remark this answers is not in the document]
patrickzib: Reverted to old state

sktime/classification/early_classification/_teaser.py
patrickzib: Hmm, I was wrong about that. We do not get the error

sktime/classification/early_classification/_teaser.py
MatthewMiddlehurst: I think just go with the full series length. self.classification_points is not required as an input, it defaults to Nones which gives <=20 values depending on the input series length.

self._classification_points removes any duplicates, any time stamps below a minimum length and adds the final series length, so they may not match up well either.

sktime/classification/early_classification/_teaser.py
```diff
@@ -0,0 +1,563 @@
...
+            if offset <= series_length:
+                next_idx = idx
+        return next_idx
+
+    def _fit_estimator(self, X, y, i):
+        rs = 255 if self.random_state == 0 else self.random_state
+        rs = None if self.random_state is None else rs * 37 * (i + 1)
+        rng = check_random_state(rs)
+
+        default = MUSE() if X.shape[1] > 1 else WEASEL()
```
patrickzib, line 354: There is a problem with the default classifier:

> raise ValueError("Base estimator must have a predict_proba method.")
>      E           ValueError: Base estimator must have a predict_proba method.

patrickzib (changes requested):

sktime/classification/early_classification/_teaser.py
```diff
@@ -0,0 +1,563 @@
...
+        predictions at. Early predictions must have a series length that matches a value
+        in the _classification_points List. Duplicate values will be removed, and the
+        full series length will be appeneded if not present.
+        If None, will use 20 thresholds linearly spaces from 0 to the series length.
+    n_jobs : int, default=1
+        The number of jobs to run in parallel for both `fit` and `predict`.
+        ``-1`` means using all processors.
+    random_state : int or None, default=None
+        Seed for random number generation.
+    return_safety_decisions : bool, default=True
```
patrickzib, line 66: I just noticed, that `score()` has to enforce `return_safety_decisions` to be set to false. Otherwise it will run into an error.

How can that best be done?

patrickzib (changes requested):

sktime/classification/early_classification/_teaser.py
patrickzib: Also it is unclear to me how to derive the actual earliness ...

patrickzib: And how would I get the accuracy and earliness after training? Maybe store it as a local variable in fit? Or does that change the classifier again, leading to an error?

MatthewMiddlehurst: It would need a custom score function.

I calculated earliness using state_info and the _classification_points array before. Could add another column containing the input series length so it is easier for others. 

Do you mean the accuracy and earliness from the _fit parameter search? could just store them in i.e. self._fit_accuracy and self._fit_earliness.

patrickzib: >  accuracy and earliness from the _fit parameter search?

I wondered about both - i.e., from score and fit :) 

> could just store them in i.e. self._fit_accuracy and self._fit_earliness.

I see - so simply store these values after training in these variables?

> It would need a custom score function.

Could you give pointers or help with the implementation?

MatthewMiddlehurst: I can try do the score function. 

I think you would have to return them with hm since its in its own functions, then store the best ones yeah.

sktime/classification/early_classification/_teaser.py
MatthewMiddlehurst: Never seen this before but changed the  _fit check, also accepts None.

patrickzib: There is another potential problem with using the default classifier. Now we run into:

> E               ValueError: Found array with 0 feature(s) (shape=(20, 0)) while a minimum of 1 is required.

When running, i.e.:

```
teaser = TEASER(
        random_state=0,
        classification_points=[6, 10, 16, 24],
        return_safety_decisions=False,  # must be enforced?!
    )
```

I.e. if the classificaiton_point is smaller than the smallest weasel window size...

sktime/classification/early_classification/_teaser.py
patrickzib: Cool, thank you! Finally: how might we also get the earliness in score()?

MatthewMiddlehurst: New score is basic, could probably edit it to return earliness/hm but at least it works for now

sktime/classification/early_classification/_teaser.py
MatthewMiddlehurst: I have set a min length of 8 if WEASEL/MUSE is used.

patrickzib: Yes, noticed this issue at the same time :)

sktime/classification/early_classification/_teaser.py
MatthewMiddlehurst: I don't know what limitations there are with it really, never edited it. Maybe you can just calculate the earliness from state_info and return it along with accuracy?

patrickzib: I have modified the score-method to return acc, earliness and hm. Not sure, if we have to consider the decision-state to compute these values though. Right now, I simply take the last ones as accepted by TEASER.

patrickzib (dismissed):

MatthewMiddlehurst:
Found an issue with predict_proba, last_idx was set incorrectly, I think. predict_proba would update state_info twice each call for the first test where we only move up a single time stamp. Trying to fix now.

MatthewMiddlehurst:
Unsure if we need another reviewer, I don't think so though. Pretty sure we can check off #1616 after this 🙂

patrickzib:
> Unsure if we need another reviewer, I don't think so though. Pretty sure we can check off #1616 after this 🙂

Just one thing: You might want to run the jupyter notebook. The code you have added to the end of the notebook did not run, yet.

MatthewMiddlehurst:
Oh, I also added some stuff to the notebook and added references to the class file if you want to take a look at those again before any merge.

patrickzib:
> Oh, I also added some stuff to the notebook and added references to the class file if you want to take a look at those again before any merge.

Yes, thank you. The jupyter notebook might need some explanation on the code. Currently there is not even an output for the last cell.

patrickzib:
> > Oh, I also added some stuff to the notebook and added references to the class file if you want to take a look at those again before any merge.
> 
> Yes, thank you. The jupyter notebook might need some explanation on the code. Currently there is not even an output for the last cell.

I will work on that now

MatthewMiddlehurst:
Ok, will leave it then. It's most likely a bit confusing as is I agree.

patrickzib:
> Ok, will leave it then. It's most likely a bit confusing as is I agree.

There is a bug that comes up with the notebook:

```
X = arrow_test_X[:, :, :100]
probas, decisions, states = teaser.predict_proba(X)
idx = np.where(np.array(probas) >= 0)
print(probas)
print(idx)
```

This code always returns -1 als probas for all classes...

sktime/classification/early_classification/_teaser.py
```diff
@@ -362,7 +371,7 @@ def _predict_proba(self, X, state_info=None):
         probas = np.array(
             [
                 probas[new_state_info[i][0] - (last_idx + 1)][i]
-                if accept_decision[i]
+                if accept_decision[i] and new_state_info[i][0] == last_idx + 1
```
patrickzib, line 374: I think, this should be:


```
probas = np.array(
[
probas[new_state_info[i, 0] - (last_idx + 1)][i]
if accept_decision[i] and new_state_info[i, 0] <= next_idx
else [-1 for _ in range(self.n_classes_)]
for i in range(n_instances)
]
)
```

patrickzib (changes requested):
A bug for getting probas

sktime/classification/early_classification/_teaser.py
MatthewMiddlehurst: I'm struggling to reproduce this actually, moved all the notebook code into a test, and it runs fine. Don't get any error raised in the notebook.

We don't have any probabilities to return below (last_idx + 1) I'm pretty sure, so we have to return [-1] for anything that has already made a decision.

patrickzib: Hmm, ...

This test:
```
X = arrow_test_X[:, :, :100]
probas, decisions, states = teaser.predict_proba(X)
idx = np.where(np.array(probas) >= 0)
print(probas)
print(idx)
```

Returns always -1 for all probabilities. When you look into the decisions however, there are a couple of "true" decisions.

patrickzib: This fix addresses a couple of things:

```
probas = np.array(
[
probas[new_state_info[i, 0] - (last_idx + 1)][i]
if accept_decision[i] and new_state_info[i, 0] <= next_idx
else [-1 for _ in range(self.n_classes_)]
for i in range(n_instances)
]
```

- Use `next_idx` instead of last+1. The test may do a couple of runs but not just one if the length is larger than the subsequent _classification_points
- use new_state_info[i, 0] instead of new_state_info[i][0] - pretty sure these are not the same for 2d data

MatthewMiddlehurst: I think we are testing different things. The previous method broke a single run, I think this will break multiple consecutive runs. 

if we are on timestamp 5 and we already made a decision in the call at time stamp 3, we dont have any probabilities to return for the case that already made a decision.

```
        probas = np.array(
            [
                probas[new_state_info[i][0] - (last_idx + 1)][i]
                if accept_decision[i] and new_state_info[i][0] >= last_idx + 1
                else [-1 for _ in range(self.n_classes_)]
                for i in range(n_instances)
            ]
        )
```

changing to >= should work? for the test above at least.

patrickzib: > if we are on timestamp 5 and we already made a decision in the call at time stamp 3, we dont have any probabilities to return for the case that already made a decision.

I do not understand.

I am on your notebook. You wrote an example how to use TEASER with partial data. And when you run this example (regardless of the length you give it), it will always return -1 as all probabilities. Even when the decisions were taken out. 

And the reason is: when calling predict_proba with a TS length, that is larger than the first classification point it will do multiple predictions (not just one) for all offsets that are smaller than the current length of the time series.

patrickzib: But yet:

This
`                if accept_decision[i] and new_state_info[i, 0] <= next_idx`
seems to give the same result as:
`                if accept_decision[i] and new_state_info[i][0] >= last_idx + 1`

I personally find the first one more intuitive :)

patrickzib: The second example in the notebook however needs a fix, too.

```
probas = np.array(
    372     [
    373         # probas[max(0, new_state_info[i, 0] - (last_idx + 1))][i]
--> 374         probas[new_state_info[i, 0] - (last_idx + 1)][i]
    375         if accept_decision[i] and new_state_info[i, 0] <= next_idx
    376         # if accept_decision[i] and new_state_info[i][0] >= last_idx + 1
    377         else [-1 for _ in range(self.n_classes_)]
    378         for i in range(n_instances)
    379     ]
    380 )
    382 return (probas, accept_decision, new_state_info)

IndexError: tuple index out of range
```

This might be fixed by:
`probas[max(0, new_state_info[i, 0] - (last_idx + 1))][i]`

Not sure, if this is the best fix?

MatthewMiddlehurst:
```
def test_bug():
    # Imports used in this notebook
    import numpy as np

    from sktime.classification.early_classification._teaser import TEASER
    from sktime.classification.interval_based import TimeSeriesForestClassifier
    from sktime.datasets import load_arrow_head

    # Load default train/test splits from sktime/datasets/data
    arrow_train_X, arrow_train_y = load_arrow_head(split="train", return_type="numpy3d")
    arrow_test_X, arrow_test_y = load_arrow_head(split="test", return_type="numpy3d")

    teaser = TEASER(
        random_state=0,
        classification_points=[25, 50, 75, 100, 125, 150, 175, 200, 251],
        estimator=TimeSeriesForestClassifier(n_estimators=10, random_state=0),
    )
    teaser.fit(arrow_train_X, arrow_train_y)

    X = arrow_test_X[:, :, :25]
    probas, decisions, states = teaser.predict_proba(X, state_info=None)

    print(probas)
    print(decisions)

    test_points = [50, 75, 100, 125, 150, 175, 200, 251]
    final_probas = np.zeros((arrow_test_X.shape[0], np.unique(arrow_test_y).shape[0]))
    final_decisions = np.zeros(arrow_test_X.shape[0])

    for i in test_points:
        X = arrow_test_X[:, :, :i]
        probas, decisions, states = teaser.predict_proba(X, state_info=states)

        print(probas)
        print(decisions)

        for n in range(arrow_test_X.shape[0]):
            if decisions[n] and final_decisions[n] == 0:
                final_probas[n] = probas[n]
                final_decisions[n] = i
```

This currently gives the output i would expect. When a decision becomes true for the first time, probabilities are returned, else -1.

that is with `if accept_decision[i] and new_state_info[i][0] >= last_idx + 1`

with `if accept_decision[i] and new_state_info[i, 0] <= next_idx` i get 

```
.0 = <range_iterator object at 0x00000155667F5C30>

        [
>           probas[new_state_info[i][0] - (last_idx + 1)][i]
            if accept_decision[i] and new_state_info[i, 0] <= next_idx
            else [-1 for _ in range(self.n_classes_)]
            for i in range(n_instances)
        ]
    )
E   IndexError: tuple index out of range
```

MatthewMiddlehurst: because if we are doing them 1 by 1, we can only return probabilities for decisions made in that predict_proba call, not previous ones (old state_infos wont be updated, and will try to get probas that dont exist)

patrickzib: Isn`t it possible with:

```
 probas = np.array(
            [
                probas[max(0, new_state_info[i, 0] - (last_idx + 1))][i]
                if accept_decision[i] and new_state_info[i, 0] <= next_idx
                else [-1 for _ in range(self.n_classes_)]
                for i in range(n_instances)
            ]
        )
```

i.e. if the probas aren't updated, we simply take the only one (first) entry?

patrickzib: At least your test runs with that addition.

patrickzib: The above code also works with the notebook with multiple calls.

MatthewMiddlehurst:
```
 probas = np.array(
            [
                probas[max(0, new_state_info[i, 0] - (last_idx + 1))][i]
                if accept_decision[i]
                else [-1 for _ in range(self.n_classes_)]
                for i in range(n_instances)
            ]
        )
```

dont think we need the and with that. Yeah should work, only difference is whether we want to return probabilities for decisions we have already made previously

patrickzib: I would rather keep that information, otherwise we cannot compute earliness and accuracy easily without merging

patrickzib: Are you ok, if I push your last code and the notebook?

MatthewMiddlehurst: Yeah go ahead. 

```
        probas = np.array(
            [
                probas[
                    new_state_info[i][0] - (last_idx + 1)
                    if new_state_info[i][0] - (last_idx + 1) >= 0
                    else len(probas) - 1
                ][i]
                if accept_decision[i]
                else [-1 for _ in range(self.n_classes_)]
                for i in range(n_instances)
            ]
        )
```

This is uglier but will return the last set of probas instead of the first (in theory more accurate?) i'm not gonna push that it be used though :)

patrickzib: Is there any difference?

Quickly tried the following with your test:

```
probas2 = np.array(
            [
                probas[max(0, new_state_info[i, 0] - (last_idx + 1))][i]
                if accept_decision[i]
                else [-1 for _ in range(self.n_classes_)]
                for i in range(n_instances)
            ]
        )


probas = np.array(
            [
                probas[
                    new_state_info[i, 0] - (last_idx + 1)
                    if new_state_info[i, 0] - (last_idx + 1) >= 0
                    else len(probas) - 1
                ][i]
                if accept_decision[i]
                else [-1 for _ in range(self.n_classes_)]
                for i in range(n_instances)
            ]
        )

print (probas != probas2)
```

Always false

MatthewMiddlehurst: Only if you predict steps larger than 1 i.e. [10, 20, 30, 40, 50, 60] going from 10 to 30 then 50. It is pretty minor.

if you are going in steps of 1 then len(probas) - 1 always == 0, and if you are doing it all at once it doesn't really matter.

patrickzib: I see. I am using 
```
        probas = np.array(
            [
                probas[max(len(probas) - 1, new_state_info[i, 0] - (last_idx + 1))][i]
                if accept_decision[i]
                else [-1 for _ in range(self.n_classes_)]
                for i in range(n_instances)
            ]
        )
```

then. This surely will not harm

MatthewMiddlehurst: that will always return the last probability right?

MatthewMiddlehurst: Just go with 0 if it is getting complicated, it doesn't really matter if we return the probabilities from time stamp 40 or 50 if we already returned a decision telling them to use timestamp 30.

patrickzib (approved):

MatthewMiddlehurst:
I don't get a vote, but I approve the changes made by Patrick 🙂. I don't believe we need any additional reviews to merge, but if anyone else wants to add a review, they are welcome.

MatthewMiddlehurst:
I will update the other eTSC classifier in another PR, and possibly add some more docs to both.

PR #2164, opened 2022-03-04
[ENH] Classifier pipeline and dunder method

fkiraly:
This PR introduces an `sktime` compatible transformer/classifier pipeline, and extends the design study in #2090.
It also contains minor additions to the `_HeterogenousMetaEstimator`, and common pipeline related utility functions moved there.

As a proof of concept, this demonstrates how estimators of different type can be used in a dunder method based pipeline in a way that is hopefully very user friendly.

The `ClassifierPipeline` inherits from `BaseClassifier` and `_HeterogeneousMetaEstimator`, and takes a list of transformers, or string-transformer pairs, and a classifier. Internally, it uses the `TransformerPipeline`

The `ClassifierPipeline` and `BaseClassifier` now implement the `rmul` right multiplication dunder method, which implements the "obvious" functionality of calling `ClassifierPipeline` and/or `TransformerPipeline` concatenation whenever a `BaseTransformer` child is encountered as `other`.

This allows easy creation of transformer/classifier pipelines by commands such as
```python
my_trafo1 * my_trafo2 * my_clf
```
where dispatch and class selection happens "under the hood" of the dunder methods.

This PR is also an illustration to [earlier discussion here](https://github.com/alan-turing-institute/sktime/pull/2090#issuecomment-1053536808) with @aiwalter on implementing an interpreter pattern, as a reminder (first comment is by @aiwalter, reply is by me):

>> I looked a bit into. I was thinking why we need a separate class for it and cant integrate it into existing pipeline classes or replace them with sth that is superior in functionality? On the other hand, this class could be used in other pipeline classes

> Yes, that was my thinking - have one class by what the "result" is, and then use it in an interpreter pattern, or call it in more powerful pipeline classes.

> For instance, if you write `my_est1 * my_est2`, the scitype of the result will depend on the scitype of `my_est1` and `my_est2`. E.g., is it a transformer or a forecaster or a classifier? Transformers can "connect" with transformers using the `TransformerPipeline`, whereas the specialized scitypes will have mechanics to connect with individual transformers and/or the `TransformerPipeline`.

> That would replace a huge tree of dispatch and case distinctions with functionality localized to interactions of the specialized classes with transformers and transformer pipelines.


fkiraly:
Hope you like this, @TonyBagnall - I started with classification and not forecasting, since in forecasting, you have *three* different ways you could connect transformers to a forecaster, and one of these three ways has two "directions".
I wanted to see how it works out for a simpler type of pipeline first.

TonyBagnall:
I didn't realise python had this form of operator overloading, nor that it had its own terminology, I had to google it, so forgive my ignorance. If there is a use case for this, it all looks fine. I'm guessing it is for the case when sklearn pipelines wont work (e.g. nested dataframe input). I'm a bit confused by the test function, which is due to my lack of experience with python operator overloading I'm sure. 

    t12c_1 = t1 * (t2 * c)
    t12c_2 = (t1 * t2) * c
    t12c_3 = t1 * t2 * c

The use of brackets seems to imply enforced precedents, but then t12c_1 doesnt make sense to me, it seems to be composing "transform with t1, fit classifier c, then transform the output of (t1,c) with t2" That doesnt make sense to me, I thought the classifier had to be the last operator. t12c_2 and t12_c3 seem the same?

    y_pred = c.fit(X, y).predict(X_test)

    _assert_array_almost_equal(y_pred, t12c_1.fit(X, y).predict(X_test))
    _assert_array_almost_equal(y_pred, t12c_2.fit(X, y).predict(X_test))
    _assert_array_almost_equal(y_pred, t12c_3.fit(X, y).predict(X_test))

the fact these are the same is an artifact of the transforms selected, rather than testing a general property. I guess thats fine, but it did confuse me, as it is not a general property to enforce on the output of a composed classifier. It would be good to test it on baked in datasets with different input data types for correctness? It would be nice to see the output were identical for multivariate data (e.g. BasicMotions) when passed as a 3D numpy and a nested dataframe.

TonyBagnall (approved):

fkiraly:
> The use of brackets seems to imply enforced precedents, but then t12c_1 doesnt make sense to me, it seems to be composing "transform with t1, fit classifier c, then transform the output of (t1,c) with t2" That doesnt make sense to me, I thought the classifier had to be the last operator. t12c_2 and t12_c3 seem the same?

Not quite, in the algebra defined by this design, the *left* estimator is always applied first, then the *right* one.
So, `t1 * (t2 * c)` translates to, "create the classifier that first applies transformation `t1`, then applies classifier `t2 * c` to the transformed output (both in `fit` and `predict`)", where the classifier spec `t2 * c` translates to "create the classifier that first applies transformation `t2`, then applies classifier `c` to the transformed output (both in `fit` and `predict`)".

This is just to test that different bracketing all leads to the same result, namely the classifier that first applies `t1`, then `t2`, then `c`.

fkiraly:
PS: this can't be merged, since there is an open discussion with @aiwalter on the "transformers" pipeline in #2090, which must be resolved first.

fkiraly:
> the fact these are the same is an artifact of the transforms selected, rather than testing a general property.

Yes, correct - I was unable to come up with something generic quickly, assuming that as soon as the generic compatibility fails, this specific compatibility is also very likely (albeit not 100% guararnteed) to fail.

fkiraly:
> It would be good to test it on baked in datasets with different input data types for correctness? It would be nice to see the output were identical for multivariate data (e.g. BasicMotions) when passed as a 3D numpy and a nested dataframe.

I think that is actually tested, and extensively, since the same kinds of pipelines is created via `get_test_params`.
Although the pipeline is not constructed using the overloaded operators.

fkiraly:
> I didn't realise python had this form of operator overloading, nor that it had its own terminology, I had to google it, so forgive my ignorance.

On, Comet! On, Cupid! On, Dunder and Blitzen!

... it's obvious what it means, isn't it?
(it's obvious that this was sarcasm, isn't it?)

TonyBagnall:
> > The use of brackets seems to imply enforced precedents, but then t12c_1 doesnt make sense to me, it seems to be composing "transform with t1, fit classifier c, then transform the output of (t1,c) with t2" That doesnt make sense to me, I thought the classifier had to be the last operator. t12c_2 and t12_c3 seem the same?
> 
> Not quite, in the algebra defined by this design, the _left_ estimator is always applied first, then the _right_ one. So, `t1 * (t2 * c)` translates to, "create the classifier that first applies transformation `t1`, then applies classifier `t2 * c` to the transformed output (both in `fit` and `predict`)", where the classifier spec `t2 * c` translates to "create the classifier that first applies transformation `t2`, then applies classifier `c` to the transformed output (both in `fit` and `predict`)".
> 
> This is just to test that different bracketing all leads to the same result, namely the classifier that first applies `t1`, then `t2`, then `c`. (depending on the bracketing, `__mul__` or `__lmul__` may be called, and it may be called in different sequence).

ah of course. I wonder on the need for the bracketing functionality at all? I'm trying to think of a scenario  where I would want it

fkiraly:
>ah of course. I wonder on the need for the bracketing functionality at all? I'm trying to think of a scenario where I would want it

I don't think a user really needs it, but I think we ought to test it, to cover all the dunder methods, and to avoid unexpected surprises...

TonyBagnall:
I was just wondering what functionality this offers over sklearn's Pipeline. What is the use case not already covered by Pipeline? And as for the brackets,  I understand testing it if its there, but not sure why its there in the first place. Have you got a use case for the brackets?

fkiraly:
> I was just wondering what functionality this offers over sklearn's Pipeline. What is the use case not already covered by Pipeline? 

If you use sklearn's `Pipeline`, you no longer have an `sktime` classifier as a result. Further composition, e.g., the popular `GridSearchCV` on top of it, can hence fail; or, passing data containers that are not 2D numpy arrays can go wrong.

>And as for the brackets, I understand testing it if its there, but not sure why its there in the first place. Have you got a use case for the brackets?

I don't think that users would use brackets. They are there in the tests, because the different bracketing allows to test `__rmul__` with `other` being a "standard" transformer (if the brackets are on the right), and `other` being a `TransformerPipeliene` (if the brackets are on the left), this covers the two principal cases inside `__rmul__`.

fkiraly:
(merging as approved earlier, and more recent changes were just consolidation with `main`)

Issue #1390, opened 2021-09-05
[ENH] classifier family as estimator tag


fkiraly:
I think the classifier family (distance based etc) whould also be a classifier tag.

That way, we can look up classifiers using the tag with `all_estimators`.

Opinions or preferences here, @tonybagnall?

TonyBagnall:
Yes @fkiraly i think that's a good idea, I'll do it in my base class refactor

MatthewMiddlehurst:
We now have a classifier_type tag, and relevant classifiers have been updated to use it.

PR #2165, opened 2022-03-04
[ENH] Introduce classifier_type tag

MatthewMiddlehurst:
Adds a new tag for classifier type and adds it to most classifiers (#1390). Classifiers missing the tag either don't have a type (early classification, composable) or are not refactored/lack tags (ShapeDTW and TSF).

~~Removes the capability:missing_values tag, replaces any usage with handles-missing-data. As noted by the comment currently there, capability:missing_values is essentially a duplicate of the latter. The inconsistent usage of _ and - is a bit jarring, but that's an issue for another time, I guess.~~

There were a few commented out tags, I assume for annotation. Is there any reason these should not be removed or uncommented?


fkiraly (changes requested):
> Removes the capability:missing_values tag, replaces any usage with handles-missing-data. As noted by the comment currently there, capability:missing_values is essentially a duplicate of the latter. The inconsistent usage of _ and - is a bit jarring, but that's an issue for another time, I guess.

If we´re already doing this, do we want to change this in a separate PR; and everything to `handles_missing_data`?

Also, I think this needs deprecation, since it is a tag that users will likely have referred to in their code. I´ve tried to write a deprecation utility here https://github.com/alan-turing-institute/sktime/pull/1962 (not entirely complete, feel free to take it over)

MatthewMiddlehurst:
If removing this is a larger job, then we can leave it to a separate PR. 

Deprecation seems a little overkill, no classifier actually uses the tag, only the base class, dummy classes for testing and templates. Keeping it in just leaves time for the issue to grow larger IMO.

fkiraly:
> Deprecation seems a little overkill, no classifier actually uses the tag, only the base class, dummy classes for testing and templates. Keeping it in just leaves time for the issue to grow larger IMO.

Deprecation is not just about internal and developer use, imo, but also - and especially - about direct use by users and dependent packages. The tag might be common in use, since it´s about missing data.

MatthewMiddlehurst:
We'll leave removing it for another PR then.

fkiraly:
>We'll leave removing it for another PR then.

Ok - would you be so kind and move it to another PR?
So I can work on the deprecation utility and then move it in once that's merged.
I suggest a move to a name with only underscores, but we may like to keep the "capability" grouping.

patrickzib:
Looks fine to me. 

One suggestion: what about adding a tag that defines, if the classifier returns a probability of a decision function?

patrickzib (approved):
I am fine with the change. See my comment about decision_function() vs predict_proba()

TonyBagnall (approved):
all looks fine to me

TonyBagnall (approved):

Issue #2132, opened 2022-02-24
[ENH] add tests for classifier compatibility with scikit model selection


TonyBagnall:
we should automatically test that all classifiers (and subsequently, clusterers) are completely compatible with scikit learn model selection and evaluation tools, and document recommended parameter/hyper parameter candidates for tuning. It would also be interesting to find out if tuning actually improves and o the core classifiers, I suspect it does not.  

https://scikit-learn.org/stable/model_selection.html

fkiraly:
Related: #2116, I don´t think they currently are compatible?

MatthewMiddlehurst:
We have some tests (not all classifiers, but the interface) for classification. I don't think there is anything for clustering?

Issue #2116, opened 2022-02-23
[ENH] grid/random search tuning for time series classification


fkiraly:
We should ensure one of two things:
* that grid/random search tuning is straighforward for TSC with `sklearn` and onboard tools.
* that a grid/random search tuning wrapper is written for TSC, similar to `ForecastingGridSearch`

In either case, this should be added to the TSC tutorial.

@TonyBagnall, @MatthewMiddlehurst, @ABostrom, @chrisholder, am I right in assuming this does not exist yet?
If it does, kindly correct me, and I would appreciate pointers (and @jmwhyte possibly too).

jmwhyte:
@fkiraly @TonyBagnall @MatthewMiddlehurst @ABostrom @chrisholder I will be very happy to have pointers, and ideally, a clear tutorial on the hyper-parameter tuning and cross-validation tasks in TSC.

MatthewMiddlehurst:
Ideally, they should just work with the sklearn functions? There isn't any fundamental difference between cross-validating/tuning time series data and regular data.

Are there any cases where this does not work currently that you are aware of? We use this sklearn function fine when running our experiments.

`train_probs = cross_val_predict(classifier, X_train, y=y_train, cv=cv_size, method="predict_proba")`

MatthewMiddlehurst:
There is still a classification notebook. 

https://github.com/alan-turing-institute/sktime/blob/main/examples/02_classification.ipynb

TonyBagnall:
all classifiers should work with scikit cross validation, although admittedly I have not fully tested this. I  plan to greatly expand the classification notebook in due course, but I can add a section on this sooner than the whole thing. Structurally, it will be identical to scikit learn, but it would be good to present what parameters there are for each algorithm.

TonyBagnall:
dont really see the need for a wrapper, unless you mean a TunedClassifier wrapper such as that we use in tsml https://github.com/uea-machine-learning/tsml/blob/master/src/main/java/machine_learning/classifiers/tuned/TunedClassifier.java

fkiraly:
> all classifiers should work with scikit cross validation, although admittedly I have not fully tested this. I plan to greatly expand the classification notebook in due course, but I can add a section on this sooner than the whole thing. Structurally, it will be identical to scikit learn, but it would be good to present what parameters there are for each algorithm.

Are we absolutely certain you can use `sklearn` grid search with `sktime` time series classifiers, and it doesn't jumble the different series together into train/test set?

[...]

PR #2180, opened 2022-03-08
[ENH] sklearn model selection tests for classification

MatthewMiddlehurst:
Adds TSC tests to ensure compatibility with sklearn model selection, cross-validation and certain composite classifiers with the sktime panel data formats and classifier framework (#2132 & #2116).

Includes some other small test clean-ups. `self._estimator_type = "classifier"` is required for `CalibratedClassifierCV`.

As of creation, the only test I think fails is the Pipeline with DataFrame input.


TonyBagnall:
great stuff, I started on this but got side tracked. Is it the case that anything that implements the abstract base class must work correctly with scikit learn? We could just create a dummy class to test if this is the case? I think the classifier tests that already exist need looking at, the input compatibility tests are I think redundant because it is all handled in the abstract base class.

MatthewMiddlehurst:
Anything implementing the base class should work, yes. Some of them could be replaced with a dummy classifier, but for things like grid search it needs a bit more than the ones already implemented.

I think the pipeline test fail is on our end, not sklearns. This fails, but replacing the Differencer with PCA passes.
```
    Pipeline(
        [
            ("transform", Differencer()),
            ("clf", CanonicalIntervalForest.create_test_instance()),
        ]
    ),
```

TonyBagnall (approved):
we might look at using a DummyClassifier to do the tests with the minimal interface implementation and a dummy parameter, but can leave that for a later date

PR #2182, opened 2022-03-09
[DOC] updated soft dependency docs with two tier check

fkiraly:
This updates the docs on soft dependency handling with the workflow implied by the fix in https://github.com/alan-turing-institute/sktime/pull/2170.

The main change is explaining that the check needs to be in two places, at the top of the module (warning), and in the estimator init (error).

Other, minor changes:
* restructuring of the "soft dependency" section, hopefully better to follow now
* removal of duplicate "deprecation" section, newer version is on the "deprecation" page

### DIFF
--- a/docs/source/developer_guide/coding_standards.rst
+++ b/docs/source/developer_guide/coding_standards.rst
@@ -54,29 +54,29 @@ other packages as soft dependencies when feasible.
 Soft dependencies in sktime should usually be restricted to estimators.
 To add an estimator with a soft dependency, ensure the following:
 
-* imports of the soft dependency only happen inside the estimator, e.g., in ``_fit``
-or ``__init__`` methods of the estimator.
-* Errors and warnings, with informative instructions on how to install the soft dependency,
-are raised through ``_check_soft_dependencies``
-`here <https://github.com/alan-turing-institute/sktime/blob/main/sktime/utils/validation/_dependencies.py>`__.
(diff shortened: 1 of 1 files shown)


docs/source/developer_guide/coding_standards.rst
```diff
@@ -51,22 +51,32 @@ other packages as soft dependencies when feasible.
...
-This is handled through our ``_check_soft_dependencies`` defined
-`here <https://github.com/alan-turing-institute/sktime/blob/main/sktime/utils/validation/_dependencies.py>`__.
-
-We use continuous integration tests to check if all soft
-dependencies are properly isolated to specific modules.
-If you add a new soft dependency, make sure to add it
-`here <https://github.com/alan-turing-institute/sktime/blob/main/build_tools/azure/check_soft_dependencies.py>`__
-together with the module that depends on it.
+Soft dependencies in sktime should usually be restricted to estimators.
+To add an estimator with a soft dependency, ensure the following:
```
aiwalter, line 55:
```suggestion
Soft dependencies in sktime should usually be restricted to estimators. Only soft-dependencies
with permissive FOSS licenses can be added as soft-dependencies.
To add an estimator with a soft dependency, ensure the following:
```

aiwalter:
where is that deprecation part now living? Was that copy pasted what you have removed?

aiwalter (approved):

PR #2186, opened 2022-03-09
[DOC] New estimator implementation guide

fkiraly:
This PR adds a new estimator implementation guide for developers and contributors.

The current "adding estimator" guide just displayed the forecaster extension template.

This PR contains a complete rewrite and a step-by-step guide, which should hopefully be much more helpful to new contributors.


fkiraly:
> should this be called add_forecasters rather than add_estimators, or are you envisaging other sections?

No? This is for all estimators! Forecasters are just used as the occasional example.

TonyBagnall (changes requested):
under "What is my scitype?" could you first explain or link to an explanation of what a scitype is? It is not a standard term,  and I do not think "Usually, the scitype of a given estimator is directly determined by what the estimator does." is sufficient. Just a sentence such as "scitype is a term we use to characterise the type of problem an estimator is designed to address" then a link to a more detailed explanation, or sth like that. Otherwise people will just google scitype and all they will find is a link to mlj. 
image

Also, could you list the existing templates under What are ``sktime`` extension templates?

fkiraly:
@TonyBagnall, added clarification on this - is this better?

TonyBagnall (approved):
looks good

PR #2191, opened 2022-03-10
[DOC] Replaced youtube link with recent PyData Global

aiwalter:


fkiraly (commented):
can we somehow also count the views on the old video?

aiwalter:
what do you mean with count?

fkiraly:
> what do you mean with count?

View count, which for the old video is at 80k or so, and for the new video at 2k?

aiwalter:
yes, but who cares about that if its old content? Also the new one was PyData Global and the other one "only" PyData Amsterdam.

fkiraly:
>yes, but who cares about that if its old content?

Well, if there´s a way to link the up-to-date content and cumulative view counts on the sktime channel, instead of an individual video? Because the alternative would reset the view count shown to zero each time we update the basic tutorial.

fkiraly:
sth like this, random google?
https://github.com/ntec-io/Youtube-Channel-Badge

fkiraly (approved):
(happy with this)

PR #2198, opened 2022-03-12
Resid docs

### LINKED ISSUE #2128: [DOC] add residual prediction convenience function `predict_residuals` to forecasting tutorial
As the title says, there should be a short section on predicting residuals `predict_residuals` to the forecasting tutorial - both about getting all the in-sample residuals (default arguments) and getting specific residuals out-of-sample or in-sample.

varuniyer:
#### Reference Issues/PRs
Fixes #2128.

This adds a step in the forecasting tutorial regarding the `predict_residuals` function.

No

The step 6 of the forecasting tutorial mentioning `predict_residuals` should be looked at.

#### Any other comments?
No

##### For all contributions


fkiraly (changes requested):
Thanks, @varuniyer, for helping with the tutorial!

I would not suggest to add this to the basic workflow, though. Reason being, `predict` is sth one always wants, whereas `predict_residuals` is more of a convenience function.

instead of a "step 6", it might make sense to add this in a section 1.6, hence?
"Convenience methods: scoring and residual prediction" (which also explains `score`)

varuniyer:
Thanks for the feedback. May latest commit moves this to a newly created section (1.6). I also added brief information along with an example of how to use the score function.

fkiraly:
abandoned

fkiraly:
@varuniyer, please ensure the linting passes.
Are you still working on this?

varuniyer:
I'm not currently working on it.

PR #2199, opened 2022-03-12
updated the environments with python version for sktime, added python 3.9

Vasudeva-bit:
#### Reference Issues/PRs

Not related to any fixe/issue, just updating the environments for python versions (sktime), added 3.9

just updating compatible python versions for sktime
No dependency is introduced

cross-check that sktime works with python 3.9

##### For all contributions

### DIFF
--- a/docs/source/get_started.rst
+++ b/docs/source/get_started.rst
@@ -11,7 +11,7 @@ Installation
 
 ``sktime`` currently supports:
 
-* environments with python version 3.6, 3.7, 3.8, or 3.9.
+* environments with python version 3.7, 3.8, or 3.9.
 * operating systems Mac OS X, Unix-like OS, Windows 8.1 and higher
 * installation via ``PyPi`` or ``conda``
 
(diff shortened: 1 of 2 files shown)


docs/source/get_started.rst
```diff
@@ -11,7 +11,7 @@ Installation
 
 ``sktime`` currently supports:
 
-* environments with python version 3.6, 3.7, or 3.8.
+* environments with python version 3.6, 3.7, 3.8, or 3.9.
```
fkiraly, line 14: thanks!
3.6 is no longer supported, so that should actually be removed.
Thanks for spotting!

fkiraly (changes requested):
Thanks for this, this was indeed a typo!

The "hierarchical" function has already been renamed, so once you update from `main`, it shoudl be in your PR.

Re support, 3.6 is no longer supported, that should be removed, too.

docs/source/get_started.rst
Vasudeva-bit: oh okay, I will do this small fix!

fkiraly (approved):
Thanks for updating this!

PR #2208, opened 2022-03-12
[ENH] sklearn typing util

fkiraly:
This PR contains the following typing utilities related to sklearn:
* a function `is_sklearn_estimator` which returns True/False whether the argument is an `sklearn` estimator.
* a function `sklearn_scitype` which returns a string (`"classifier"`, `"regressor"` etc) denoting the scitype of the estimator.
* four shorthances `is_sklearn_classifier`, for `is_sklearn_estimator` logical-and `sklearn_scitype` equal to certain scitype.

This can be used in type checking, as `sklearn_scitype` is closer to the use of `type`, and avoids `sklearn` imports of mixin classes.

Note: the logic is intentionally different from `sklearn.base.is_classifier` etc, as that assumes certain attributes to be set. The functions above do *not* assume that the argument is an `sklearn` estimator, and they check sub-class membership only.

### DIFF
--- a/sktime/utils/sklearn.py
+++ b/sktime/utils/sklearn.py
@@ -65,6 +65,6 @@ def sklearn_scitype(obj, var_name="obj"):
     if issubclass(obj, sklearn_mixins):
         for mx in sklearn_mixins:
             if issubclass(obj, mx):
-                return mixin_to_scitype(mx)
+                return mixin_to_scitype[mx]
     else:
         return "estimator"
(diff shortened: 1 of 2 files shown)


TonyBagnall (approved):
looks good, my only query would be the file name sklearn.py, which could be confusing. sklearn_tests ot sklearn_types maybe? Also might be worth testing on KNeighborsTimeSeriesClassifier which is both an sklearn estimator and an sktime classifier

fkiraly:
> Also might be worth testing on KNeighborsTimeSeriesClassifier which is both an sklearn estimator and an sktime classifier

If it's both, the sktime type takes prioritiy (`is_sklearn_estimator` returns no) which imo makes sense.

fkiraly:
> `sklearn_tests` ot `sklearn_types` maybe? 

Well, it's "utilities regarding sklearn", so `utils.sklearn` should be clear enough?
I assume it's not a blocker since you approved.

PR #2209, opened 2022-03-12
[ENH] Tabular transformer adaptor "fit in transform" parameter

fkiraly:
This PR adds a boolean parameter `fit_in_transform` to `TabularToSeriesAdaptor`.
This lets a user choose whether `fit` is mapped on `fit`, or whether it is executed in `transform`.
The default setting defaults to `False`, which is the current behaviour, ensuring downwards compatibility.

The reason for that is as follows:
* when used in forecasting or on a single series, the most common use is usually `fit_in_transform=False`.
* when used in TSC or TSR, on a panel, the most common (intended) use is usually `fit_in_transform=True`, since the test indices/instances are usually considered distinct from the training indices/instances. Otherwise, the automatic vectorization from the base class will consider train and test indices identical, and refuse to `transform` if number or set of indices is different between train/test.

Minor changes:
* docstring for `TabularToSeriesAdaptor` has been extended with a more formal description to clarify behaviour.

### DIFF
--- a/sktime/transformations/series/adapt.py
+++ b/sktime/transformations/series/adapt.py
@@ -18,6 +18,46 @@ class TabularToSeriesAdaptor(BaseTransformer):
     to :term:`series <Time series>`, but only works with transformations that
     do not require multiple :term:`instances <instance>` for fitting.
 
+    The adaptor behaves as follows.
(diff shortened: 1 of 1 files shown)


fkiraly:
no review after 14 days, merge

PR #2210, opened 2022-03-13
[ENH] dunder pipelines sklearn estimator support

fkiraly:
This PR extends pipeline composition dunders of classifier and transformer pipelines such that `sklearn` transformers can be directly passes as parts of the chain, e.g.,

`my_trafo * StandardScaler() * my_clf` is possible.

The behaviour in this case is:
* first, the `sklearn` transformer is wrapped in `TabularToSeriesAdapter`
* then, the expression is parsed with the wrapped transformer (now a `sktime` transformer) instead of the original one

This should make pipeline construction easier - although the default setting of `TabularToSeriesAdaptor` can cause problems if too naively applied in TSC.

Conditional on the following PR:
* #2208, which is used for checking whether `other` is an `sklearn` transformer


fkiraly:
14 days rule merge

Issue #1888, opened 2022-01-17
[ENH] weighted average of distances & distance composites


fkiraly:
There should be a wrapper class `CompositeDistance` (or similar) that allows to specify, as parameters:

* an iterable of distances, or just a single distance
* a way to aggregate, e.g., `"mean"` or `"median"` (string) or a `numpy` function
* an iterable of weights (optional)
* an iterable of variables, or lists of variables (optional)

If iterables are passed, they must be of same length.

The distance computed is the aggregate/average of: the i-th distance (the one distance, if not an iterable), computed on the i-th variable (set; all variables if none provided), multiplied by the i-th weight (1 if not provided).

abheesht17:
Hello, @fkiraly! I would like to take this up.

I'll describe what you've said above, and you can tell me when I've understood it correctly. 

I'll be given a couple of dataframes. For example,

```
>>> a
     x1    x2    x3
0  0.34  0.87  0.30
1  0.45  0.56  0.67
2  0.12  0.32  0.31
>>> b
     x1    x2    x3
0  0.98  0.79  0.03
1  0.56  0.69  0.76
2  0.25  0.25  0.12
```

For distances, I'll be given something like "SquaredDistance" and "EuclideanDistance".

Assume
```
distances = ["SquaredDistance", "EuclideanDistance"]
weights = [0.4, 0.6]
variables = [("x1", "x2"), "x3"]
```

Essentially, this is what I'll have to compute:

```
composite_distance = 0.4 * squared_distance(a["x1", "x2"], b["x1", "x2"]) + 0.6 * euclidean_distance(a["x3"], b["x3"]).
```

Is this interpretation correct?

fkiraly:
Yes, or almost, depending on what you were thinking.
The last line would eventually be computed, but as follows.

First, you write sth
`composite_distance = CompositeDistance(distances= [SquaredDistance, EuclideanDistance], weights = [0.4, 0.6], variables=[["x1", "x2"], "x3")`

and then you can call

`composite_distance(a, b)`, this computes the value in the last line of yours.

fkiraly:
The extension template to follow would (probably) be this:
https://github.com/alan-turing-institute/sktime/blob/main/extension_templates/dist_kern_panel.py

abheesht17:
Great! Sounds good. Will open a PR soon :)

fkiraly:
implemented as `CombinedDistance` in #3264, forgot to close this.

PR #2212, opened 2022-03-13
Add Class for Composite Distance

abheesht17:

Resolves https://github.com/alan-turing-institute/sktime/issues/1888

N.A.

@fkiraly:
 
I'm a bit confused about how I should organise the code for this. There are two scenarios I can think of:
 
1. I thought we had to do something like this:
Make a class `CompositeDistance`, subclass of `NumbaDistance`. However, since `NumbaDistance` is an abstract class, I will have to implement `_distance_factory(x: np.ndarray, y: np.ndarray, ...)`. The issue with this is that I won't be passing numpy arrays, I will pass a dictionary (for example, `{"x1": np.array([1,2,3]), "x2": np.array([4,5,6]}`). So, I'm not sure if I should inherit from `NumbaDistance`. I can change the definition for `_distance_factory` in `NumbaDistance` to `_distance_factory(x: Union[np.ndarray, Dict[str, np.ndarray]], y: Union[np.ndarray, Dict[str, np.ndarray]])` and proceed, but not sure if I should do this. 
2. This is what I've done:
    Make a class `Composite Distance`, with `__call__`. Example usage:

```
>>> import numpy as np 
>>> from sktime.distances.composite import CompositeDistance
>>> cd = CompositeDistance(metrics=["euclidean", "squared"], weights=[0.4,0.6], variables=[["x1", "x2"], ["x3"]])
>>> d1 = {"x1": np.array([1,2]), "x2": np.array([3,4]), "x3": np.array([5,6])} 
>>> d2 = {"x1": np.array([2,3]), "x2": np.array([4,5]), "x3": np.array([6,7])} 
>>> cd(d1, d2)
2.331370849898476
```

#### Any other comments?
I haven't added doc-strings and unit tests yet. Let me know if what I've done is the correct way, and I'll add doc-strings and UTs.

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/distances/composite.py
+++ b/sktime/distances/composite.py
@@ -23,14 +23,17 @@ class CompositeDistance:
                 NumbaDistance,
             ]
         ],
-        weights: List[List],
         variables: List[str],
+        weights: List[List] = None,
     ):
         self.metrics = metrics
-        self.weights = weights
         self.variables = variables
+        self.weights = weights
 
     def __call__(self, x: Dict[str, np.ndarray], y: Dict[str, np.ndarray]) -> float:
+        if self.weights is None:
+            self.weights = [1] * len(self.metrics)
+
         composite_distance_val = 0
         for metric, weight, variable in zip(self.metrics, self.weights, self.variables):
             for var in variable:
--- a/sktime/distances/composite.py
+++ b/sktime/distances/composite.py
@@ -1,0 +1,41 @@
+# -*- coding: utf-8 -*-
+from typing import Callable, Dict, List, Union
+
+import numpy as np
+
+from sktime.distances._distance import _METRIC_INFOS
+from sktime.distances._numba_utils import to_numba_timeseries
+from sktime.distances._resolve_metric import _resolve_metric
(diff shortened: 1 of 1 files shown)


fkiraly (changes requested):
Thank you for your work!

The main issue with your distance is that you did not follow the extension template that I linked in #1888 - for instance, you should be inheriting from one of the base classes.
You should be implementing `_transform` and not override `__call__`.

abheesht17:
> Thank you for your work!
> 
> The main issue with your distance is that you did not follow the extension template that I linked in #1888 - for instance, you should be inheriting from one of the base classes. You should be implementing `_transform` and not override `__call__`.

Ah, I see. I was trying to make it a subclass of the `NumbaDistance` class, but couldn't find a way since `_distance_factory` takes numpy arrays, and not dictionaries. Great, I'll go through the extension template and update the PR. Thanks!

fkiraly:
@abheesht17, are you still working on this?

fkiraly:
closed due to inactivity

PR #2219, opened 2022-03-13
[ENH] Upgrade of `BaseTransformer` to use vectorization utility, hierarchical mtype compatibility

fkiraly:
This PR upgrades `BaseTransformer` to use the `VectorizedDF` utility (that `BaseForecaster` uses) instead of bespoke vectorization.
As a side benefit, this upgrades `BaseTransformer` to be compatible with `Hierarchical` inputs.

The PR also adds:
* a `test_base` test suite which targets testing of the boilerplate functionality in `BaseTransformer`. Currently, this uses actual transformers for the test, but these could be replaced by a mock transformer.

Depends on a couple bugfixes and changes, which should be reviewed in other PR:
* #2216
* #2217
* #2220
* relaxation of `pd-multiindex` mtype to remove index name requirement, same as in https://github.com/alan-turing-institute/sktime/pull/2188


lmmentel:
Let's not rush this for the release since there a lot of changes here.

lmmentel (approved):
:+1:

PR #2223, opened 2022-03-14
[MNT] test parameter refactor: forecasting reducers and ColumnEnsembleClassifier

fkiraly:
This refactors the test parameter generation of the eight forecasting reducers and `ColumnEnsembleClassifier`, in partial completion of https://github.com/alan-turing-institute/sktime/issues/1678

The case of the eight reducers is a bit special, as all eight reducers inherit from an intermediate base class for reducers. There are only two cases of parameters, depending on which substring appears in the class name.
This is resolved by having a single `get_test_params` at the intermediate base class and a case distinction based on class name.

While we're there, the `window_length` parameter has been reduced to 3 to allow quicker testing, and later reduction of the series lengths in the test suite.


fkiraly:
14 day rule, merging.

Issue #2215, opened 2022-03-13
[ENH] ForecastingHorizon support in deep_equals


ltsaprounis:
**Is your feature request related to a problem? Please describe.**
Currently `sktime.utils._testing.deep_equals.deep_equals` doesn't support the `ForecastingHorizon` type.

**Example**
```python
from sktime.utils._testing.deep_equals import deep_equals
from sktime.forecasting.base import ForecastingHorizon
fh_1 = ForecastingHorizon(values=[1,2,3], is_relative=True)
fh_2 = ForecastingHorizon(values=[1,2,3], is_relative=True)
deep_equals(fh_1, fh_2)
# Output: ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
# Expected Output: True
```

**Describe the solution you'd like**
We need to add an additional `if` block for `ForecastingHorizon` equality, similar to `numpy.ndarray`, `pandas.Series` etc. 
e.g. https://github.com/alan-turing-institute/sktime/blob/6cc0d41e778f9ed61a5cd8b24a21be680e23dd6f/sktime/utils/_testing/deep_equals.py#L108

**Describe alternatives you've considered**
One alternative fix would be changing `__eq__` in `ForecastingHorizon`. 
This feels a bit more risky.

**Additional context**
Discovered this while writing the tests for #2197.
Related to #2214.

fkiraly:
hm... why the `ForecastingHorizon`'s own `__eq__` result in "truth valus is ambiguous" do you understand that?

ltsaprounis:
It behaves like numpy arrays, pandas Series, Dataframes etc. In most cases this is a feature not a bug. Unfortunately, I don't know enough about the `ForecastingHorizon` to say what the case is here.

fkiraly:
Hm, how about just adding another case, at the end, pseudocode:
``` 
    elif it´s a ForecastingHorizon, then:
    return deep_equals(fh1.values, fh2.values) and fh1.is_relative == fh2.is_relative
    plus deal with the informative return message
```

PR #2197, opened 2022-03-12
[ENH] Mock estimators and mock estimator generators for testing

ltsaprounis:
#### Reference Issues/PRs

Related to #2091. Starting with a Univariate Forecaster.

#### What does this implement/fix? Explain your changes.

Adding a Mock Forecaster for use in debugging and testing. 
The Forecaster will log all the methods called and their parameters.
[Edit]: I added a new function `make_mock_estimator` that makes any estimator log all the methods that are called along with their inputs.

**Example:**
```python
from sktime.datasets import load_airline
from sktime.utils.estimators import MockUnivariateForecaster
forecaster = MockUnivariateForecaster()
y = load_airline()
forecaster.fit(y.iloc[:5])
y_pred = forecaster.predict(fh=[1,2,3])
y_pred_quantiles = forecaster.predict_quantiles(fh=[1,2,3], alpha=[0.1, 0.9])
forecaster.update(y.iloc[5:10])
pprint(forecaster.log)
```

Output: 
```python
[('_fit',
  {'X': None,
   'fh': None,
   'y': Period
1949-01    112.0
1949-02    118.0
1949-03    132.0
1949-04    129.0
1949-05    121.0
Freq: M, Name: Number of airline passengers, dtype: float64}),
 ('_predict',
  {'X': None,
   'fh': ForecastingHorizon([1, 2, 3], dtype='int64', is_relative=True)}),
 ('_predict_quantiles',
  {'X': None,
   'alpha': [0.1, 0.9],
   'fh': ForecastingHorizon([1, 2, 3], dtype='int64', is_relative=True)}),
 ('_update',
  {'X': None,
   'update_params': True,
   'y': Period
1949-06    135.0
1949-07    148.0
1949-08    148.0
1949-09    136.0
1949-10    119.0
Freq: M, Name: Number of airline passengers, dtype: float64})]
```

#### What should a reviewer concentrate their feedback on?

- The tests for the `make_mock_estimator` are a bit lazy. Suggestions on how to make them better are welcome! But bear in mind we don't want to make them too dependent on the private interface etc.
- Is Mock the right naming here? Maybe we can call it `LoggingForecaster` / `make_logging_estimator`

##### For all contributions
- [x] I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/main/.all-contributorsrc).
- [x] I've added unit tests and made sure they pass locally.

PR #2225, opened 2022-03-15
[ENH] extendin `deep_equals` for `ForecastingHorizon`

fkiraly:
Solves issue #2215, ensures that `deep_equals` returns the correct output if the nested structure contaians `ForecastingHorizon` somewhere.

The utility uses recursion, making use of the fact that two `ForecastingHorizon` are equal iff their `_value` (`pandas` or `numpy` container) and `is_relative` (`bool`) attributes are equal.

The extension avoids importing the `ForecastingHorizon`, ensuring that the `deep_equals` utility does not have an `sktime` dependency.
(for reviewers: that's why `type(x).__name__` is used and not `isinstance`)

### DIFF
--- a/sktime/utils/_testing/deep_equals.py
+++ b/sktime/utils/_testing/deep_equals.py
@@ -15,8 +15,6 @@ __all__ = ["deep_equals"]
 import numpy as np
 import pandas as pd
 
-# from sktime.base import BaseObject
-
 
 def deep_equals(x, y, return_msg=False):
     """Test two objects for equality in value.
--- a/sktime/utils/_testing/deep_equals.py
+++ b/sktime/utils/_testing/deep_equals.py
@@ -114,6 +114,8 @@ def deep_equals(x, y, return_msg=False):
         return ret(*_tuple_equals(x, y, return_msg=True))
     elif isinstance(x, dict):
         return ret(*_dict_equals(x, y, return_msg=True))
+    elif type(x).__name__ == "ForecastingHorizon":
+        return ret(*_fh_equals(x, y, return_msg=True))
     elif x != y:
         return ret(False, f" !=, {x} != {y}")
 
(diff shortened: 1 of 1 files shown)


fkiraly:
FYI @ltsaprounis, this is what you needed, right?

Turns out I also needed it here #2207 😄

sktime/utils/_testing/deep_equals.py
```diff
@@ -114,6 +112,8 @@ def ret(is_equal, msg):
         return ret(*_tuple_equals(x, y, return_msg=True))
     elif isinstance(x, dict):
         return ret(*_dict_equals(x, y, return_msg=True))
+    elif type(x).__name__ == "ForecastingHorizon":
```
ltsaprounis, line 115: Why not do `isinstance(x, ForecastingHorizon)` ?

ltsaprounis:
Thanks @fkiraly! this is really good and I need it for #2197!
Left one minor comment/question.

sktime/utils/_testing/deep_equals.py
fkiraly: I wrote that in the PR message 😄 

> The extension avoids importing the `ForecastingHorizon`, ensuring that the `deep_equals` utility does not have an `sktime` dependency.
> (for reviewers: that's why `type(x).__name__` is used and not `isinstance`)

ltsaprounis: > I wrote that in the PR message
🤦‍♂️ I must have run straight into the code, it was a bit late at night!

OK, I get it now. But why is it a problem to create an sktime dependency? 
The only use case I can think of is if someone copies the code.

ltsaprounis (approved):
Looks good to me! 

> lowest coupling would be achieved, btw, by overriding `__eq__` of the `ForecastingHorizon`, but I'd rather not touch it as I'm unsure about the consequences.

I agree with that! so, I suggest we treat this as a quick fix and take some time to address `ForecastingHorizon.__eq__` properly in a separate PR.

PR #2226, opened 2022-03-15
[ENH] unit tests for `deep_equals` utility

fkiraly:
This PR adds unit tests for the `deep_equals` utility which were not present so far.
Includes #2555 (`ForecastingHorizon` compatibility).


fkiraly:
Can you also kindly look at this, @ltsaprounis?
Introduces tests to cover `deep_equals`, the utility was not tested before.

PR #2229, opened 2022-03-15
[MNT] Faster docstring examples - `ForecastingGridSearchCV`, `MultiplexForecaster`

fkiraly:
This PR speeds up the two slowest docstring examples by reducing the size of data sets in those examples:
`ForecastingGridSearchCV` and `MultiplexForecaster`. This saves ca 60-90sec in testing (according to CI/CD and local timings).

### DIFF
--- a/sktime/forecasting/compose/_multiplexer.py
+++ b/sktime/forecasting/compose/_multiplexer.py
@@ -64,14 +64,14 @@ class MultiplexForecaster(_HeterogenousEnsembleForecaster):
     ...     ForecastingGridSearchCV,
     ...     ExpandingWindowSplitter,
     ...     load_airline)
-    >>> y = load_airline()
+    >>> y = load_airline()[0:24]
     >>> forecaster = MultiplexForecaster(forecasters=[
     ...     ("ets", AutoETS()),
     ...     ("arima", AutoARIMA(suppress_warnings=True, seasonal=False)),
     ...     ("naive", NaiveForecaster())])
     >>> cv = ExpandingWindowSplitter(
     ...     start_with_window=True,
-    ...     step_length=24)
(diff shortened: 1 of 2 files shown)


aiwalter:
I would rather remove the `AutoARIMA` and maybe replace with another fast model such as `STLForecaster`. Maybe also make `step_length` bigger. Just trimming the data looks a bit odd. Other solution might be to use `load_shampoo_sales` which has length 36

fkiraly:
> I would rather remove

I don't think that would make a difference, but if you don't like it, feel free to replace it with anything that takes only seconds to compute at most. Feel free to commit to this PR.

sktime/forecasting/compose/_multiplexer.py
```diff
@@ -64,14 +64,14 @@ class MultiplexForecaster(_HeterogenousEnsembleForecaster):
     ...     ForecastingGridSearchCV,
     ...     ExpandingWindowSplitter,
     ...     load_airline)
-    >>> y = load_airline()
+    >>> y = load_airline()[0:24]
```
aiwalter, line 67:
```suggestion
    >>> y = load_shampoo_sales()
```

sktime/forecasting/compose/_multiplexer.py
```diff
@@ -64,14 +64,14 @@ class MultiplexForecaster(_HeterogenousEnsembleForecaster):
     ...     ForecastingGridSearchCV,
     ...     ExpandingWindowSplitter,
     ...     load_airline)
```
aiwalter, line 66:
```suggestion
    ...     load_shampoo_sales)
```

aiwalter (approved):

PR #1810, opened 2021-12-31
[BUG] fix `FeatureUnion` test failure mk4

fkiraly:
Another attempted fix for #1662 which PR #1665, #1701 did not fix. Hopefully the definitive fix now.

This fix completely rewrites the `FeatureUnion` instead of inheriting from the `sklearn` one.
With the new transformers refactor, we can simply use `pandas` functionality to merge/concatenate data frames internally, while the framework does the rest.

Minor complication:
With the transformer refactor, the "row-wise transformer" have become obsolete, since vectorization (to panel) is performed automatically whenever such a data container is encountered.
Some tests not directly related to the `FeatureUnion` but also using `FeatureUnion` were still using them, so I `xfailed` them for the time being, until the `RowTransformer`s are entirely removed, see issue #2179.

PR #2231, opened 2022-03-15
[ENH] `FeatureUnion` refactor - moved to `transformations`, tags, dunder method

fkiraly:
This PR contains a refactor of `FeatureUnion`:
* move to `transformations.compose`, deprecate `series_as_feature` occurrence
* common refactor with `TransformerPipeliene`,  joint and duplicate methods moved to `_HeterogeneousEnsembleEstimator`
* tag handling
* `+` dunder method for `FeatureUnion`

Depends on PR #1810 which is included.


lmmentel:
Nice work! Are you running the test that was originally failing for `FutureUnion` ?

fkiraly:
> Nice work! Are you running the test that was originally failing for FutureUnion ?

Of course :-)

TonyBagnall (approved):
personally I would like to deprecate the whole module series_as_features, and this is a step to that. 

Maybe rename anytagis? Had trouble parsing that name, was thinking of haggis or tardis. maybe has_tag? but I leave it

fkiraly:
> personally I would like to deprecate the whole module series_as_features, and this is a step to that.

Completely agreed, @TonyBagnall! It's a bit solitary and the "concern" it addresses is not entirely clear.
It does have some useful stuff, the other content being the splitters, which we should find a new home for first, though.

PR #2232, opened 2022-03-16
[ENH] Improve proba metrics

eenticott-shell:
#### Reference Issues/PRs
Continues #2153.

Fixes functionality and general tidyup of changes made in #2153.

No


eenticott-shell:
@fkiraly I still need to improve the tests for the proba metrics. Are there any forecasters that implement `predict_quantile` or `predict_interval` for a multioutput case?

fkiraly:
>@fkiraly I still need to improve the tests for the proba metrics. Are there any forecasters that implement predict_quantile or predict_interval for a multioutput case?

Currently none, I think.

fkiraly (approved):
Yes, this is what I had in mind in #2153, thanks!

Aware that this is not fully done, I'm merging this for the 0.11.0 release on Monday.
Based on our discussion last week, changes to the metrics will be downwards compatible, e.g., the addition of the `alpha` argument. Whereas if we leave it as currently, we'll have the problem of deprecation which we will avoid this way.

Further changes can be added in a separate PR.

PR #2235, opened 2022-03-16
[ENH] Early classification base class

MatthewMiddlehurst:
Implementation of a base class and converted tests for early classification https://github.com/alan-turing-institute/sktime/issues/1616. Refactors TEASER to use the new base class.

The new class is very similar to the classification base class, except predict/predict_proba returns 3 ndarrays instead of 1. The first is the already present predictions/probabilities, the second is a set of decisions on whether the returned predictions are safe to use, and the third is a set of state information for each case for use in later decisions and putting together eTSC statistics.

An optional state_info input for the predict/predict_proba is also required.


fkiraly:
Hm - I think this might be too similar for a separate base class.

If the difference is really just what you say, I would handle this with a tag that describes the expected output type of `predict`.
This can be an `early_classification` tag or similar.

MatthewMiddlehurst:
We have an early classification tag. The issue with having it as a BaseClassifier is we would have to overload `predict`, `predict_proba` and `score` for every classifier, as well as `_predict` and `_predict_proba`. This is because we require an optional state_info input for the methods.

While there are similarities (`fit` is exactly the same, I just call the BaseClassifier methods and tests), I feel having these different output and input types conjoined to the regular BaseClassifier may be more trouble than it's worth. These early classifiers are not necessarily something you would want to pick out from a list of BaseClassifiers without knowing what they do, either.

I'm fine with looking into better integrating it into BaseClassifier rather than separating it if people think that is better. Pros and cons for both.

fkiraly:
@MatthewMiddlehurst, can you kindly expand on what that `state_info` is?
I've read the docstring and haven't really grasped it.
Do you have a paper or a scientific reference, and can point to what mathematical concept that `state_info` maps to?

MatthewMiddlehurst:
It depends on the classifier. It is just a record of information from a decision on whether a prediction is safe to make at the given series length. It is used to aid future decision-making at larger series lengths.

fkiraly:
> It depends on the classifier. It is just a record of information from a decision on whether a prediction is safe to make at the given series length. It is used to aid future decision-making at larger series lengths.

Could you provide a reference?

I do not get why it is an input to `predict` rather than a return.

MatthewMiddlehurst:
A reference to what? This is an implementation detail, not anything algorithmic.

It is both. `state_info` is input into `predict` to help make a decision, then predict returns a new `state_info`.

MatthewMiddlehurst:
We have an issue about API (#1616) and an example notebook (https://github.com/alan-turing-institute/sktime/blob/main/examples/early_classification.ipynb) if there is some confusion about the eTSC problem.

MatthewMiddlehurst:
https://link.springer.com/article/10.1007/s10618-020-00690-z Is a link to the TEASER algorithm by our friend @patrickzib, the implementation of which I alter in this PR to use the new `BaseEarlyClassifier`. It requires that _g_ consecutive positive decisions for the same class on whether a prediction is safe or not be made until a final positive decision on prediction safety is returned. We need a way to tell the algorithm how many consecutive decisions have been made so far.

Enter, `state_info`, which in the case of our TEASER implementation is a (n_instances, 4) ndarray of values. Each row contains the following: The time stamp index (the algorithm creates _x_ classifiers at _x_ different series lengths, this records which one we are making a decision for), the number of consecutive decisions made, the predicted class for these consecutive decisions and the series length on the input. This lets us record consecutive decisions, and verify how much time has passed/how many new values in the series there are since the last input.

An alternative to `state_info` is just storing the information in a class field. We decided against this as a) it alters the state of the classifier in `predict`/`predict_proba` and b) it can be easily overwritten, and requires the same instances to be input consecutively.

patrickzib:
@fkiraly consider a streaming scenario where new data comes in from time to time and new predictions need to be done every x-time stamps. 

State Info is a way to capture the last state (predictions) and use it for subsequent predictions on the stream without recomputing everything from scratch. And at the same time keep the estimator class stateless.

Thus, in predict() for time stamp t+1 it uses the predictions made at time stamp <=t to update its decisions.

i did not reference to any state info in my paper though, as I did not implement a streaming scenario. This is rather an implemention detail.

MatthewMiddlehurst:
Included the class in SCITYPES_WITHOUT_TAGS to fix one of the failing tests. I have no idea why estimator_scitype = ['series-annotator', 'clusterer'] is failing the same test, though.

fkiraly:
> Included the class in SCITYPES_WITHOUT_TAGS to fix one of the failing tests. I have no idea why estimator_scitype = ['series-annotator', 'clusterer'] is failing the same test, though.

Can you please elaborate? This looks like a hack, so we should understand what's going on here.

MatthewMiddlehurst:
I was hoping you would understand. Why does me adding ("early_classifier", BaseEarlyClassifier, "early time series classifier") to BASE_CLASS_REGISTER cause two (previously, one with the base class added to an array) test_all_tags tests to fail. The remaining one seems to have nothing to do with the new base class also?

I was under the assumption that that the scitype/mtype stuff was to deal with data formats? There are other base classes in the same array SCITYPES_WITHOUT_TAGS = ["series-annotator", "clusterer", "early_classifier"].

MatthewMiddlehurst:
readthedocs also seems to be failing due to a dependency `ModuleNotFoundError: No module named 'esig'`? Does not seem like any other PR's are having this issue.

fkiraly:
>Does not seem like any other PR's are having this issue.

Yes, seems specific to what you are doing.
You seem to be triggering a soft dependency somewhere by an import (docs should not require soft dependency).

fkiraly:
> I was hoping you would understand. Why does me adding ("early_classifier", BaseEarlyClassifier, "early time series classifier") to BASE_CLASS_REGISTER cause two (previously, one with the base class added to an array) test_all_tags tests to fail. The remaining one seems to have nothing to do with the new base class also?

No failures should be triggered through thsi addition imo, what is failing here? Can you print or point to the errors?

fkiraly:
> @fkiraly consider a streaming scenario where new data comes in from time to time and new predictions need to be done every x-time stamps.

> State Info is a way to capture the last state (predictions) and use it for subsequent predictions on the stream without recomputing everything from scratch. And at the same time keep the estimator class stateless.

> Thus, in predict() for time stamp t+1 it uses the predictions made at time stamp <=t to update its decisions.

I see - I would then go for another design, @patrickzib, @MatthewMiddlehurst.
To me, it seems more matural to make the classifier stateful, with internal state remembering which predictions it has already made. The `state_info` should be part of the state, anything else seems unnatural for an object oriented design.

MatthewMiddlehurst:
> >  Does not seem like any other PR's are having this issue.
> 
> Yes, seems specific to what you are doing.
> You seem to be triggering a soft dependency somewhere by an import (docs should not require soft dependency).

Strange, I thought esig was only used in the signature classifier/transformer. This pr does not touch either. I may have unintentionally touched some file used in creating the docs? 

> No failures should be triggered through thsi addition imo, what is failing here? Can you print or point to the errors?

This is the only failing test currently.

```
 [gw1] win32 -- Python 3.8.5 C:\Miniconda\envs\test\python.exe

estimator_scitype = ['series-annotator', 'clusterer']

    @pytest.mark.parametrize("estimator_scitype", estimator_scitype_fixture)
    def test_all_tags(estimator_scitype):
        """Check that all_tags return argument has correct type."""
        tags = all_tags(estimator_types=estimator_scitype)
        assert isinstance(tags, list)
    
        # there should be at least one tag returned
        # exception: scitypes which we know don't have tags associated
        if estimator_scitype not in SCITYPES_WITHOUT_TAGS:
>           assert len(tags) > 0
E           assert 0 > 0
E            +  where 0 = len([])

estimator_scitype = ['series-annotator', 'clusterer']
tags       = []

C:\Miniconda\envs\test\lib\site-packages\sktime\registry\tests\test_lookup.py:102: AssertionError
```

I am not really familiar with this code, the double estimator_scitype seems to be coming from this portion of `test_lookup.py`

```
# selected examples of "search for two types at once to avoid quadratic scaling"
double_estimator_scitypes = [[b[i], b[(i + 3) % n]] for i in range(n)]
# fixtures search by individual scitypes, "None", and some pairs
estimator_scitype_fixture = [None] + BASE_CLASS_SCITYPE_LIST + double_estimator_scitypes
```

fkiraly:
> This is the only failing test currently.

Ah. This checks whether there are any tags for the scitype.
It says that currently there are not, which typically is an error.

Is that intended? I.e., do you really not want to use tags for early classifiers?
Or do you want to use the classifier tags for the early classifiers as well?

If so, you have to add `"early_classifier"` in the tag registry for every tag that you want to use for early classifiers.

MatthewMiddlehurst:
Ahh, I understand. Yeah, I can add it to the relevant tags if this has a chance of going through. Thanks.

`estimator_scitype = ['series-annotator', 'clusterer']` failing happens regardless of whether we tell the test to skip `"early_classifier"` or not, however.

fkiraly:
Re "going through", yes, looks like it's almost there in terms of extension.
Just to mention briefly two points:
* we have feature freeze until after 0.11.0 and priority is getting the board items in.
* I'd also like to discuss the design - passing what is called a "state" around as inputs/outputs of functions instead of it being actually the state of an object is not a good idea imo. It should be part of the object's state.

fkiraly (changes requested):
as stated above, looks good, but I have a major concern with the design. The "state" should not be passed around as function arguments, but be part of the object's actual state, imo.

MatthewMiddlehurst:
> To me, it seems more matural to make the classifier stateful, with internal state remembering which predictions it has already made. The state_info should be part of the state, anything else seems unnatural for an object oriented design.

Not discounting this as it is a valid option for sure, and we initially considered it. I thought having a "stateful" classifier which changes based off predictions would have more pushback, however. Definetly open to discuss and make changes.

Main issue is how would we go about matching input cases to the internally stored state_info? If we could create somthing like a dictionary of references that would be fine, but can you do that with rows of a 3d numpy array?

fkiraly:
> I thought having a "stateful" classifier which changes based off predictions would have more pushback, however.

Well, we're using object oriented patterns, and statefulness is natural to that.

>Main issue is how would we go about matching input cases to the internally stored state_info? If we could create somthing like a dictionary of references that would be fine, but can you do that with rows of a 3d numpy array?

Not sure what the requirement is - would that not be a 2d array, or perhaps better `pd.DataFrame`, where you have as indices the instance index and the time index (when you attempted classification)?

MatthewMiddlehurst:
Could possibly convert each input row using [.tobytes](https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.ndarray.tobytes.html)() and use that as a key for a dictionary. I don't think that would that be horribly slow?

patrickzib:
> Could possibly convert each input row using [.tobytes](https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.ndarray.tobytes.html)() and use that as a key for a dictionary. I don't think that would that be horribly slow?

I feel like that is a bit of over-engineering?

Wouldn`t a simple boolean work as a parameter to predict() or likewise a function like reset_internal_state()?

MatthewMiddlehurst:
> I feel like that is a bit of over-engineering?
> 
> Wouldn`t a simple boolean work as a parameter to predict() or likewise a function like reset_internal_state()?

Perhaps. In theory, it should just be something like:

`state_info = np.array([self.state_info[X[i].tobytes()] for i in n_instances)])`

and then you have the state_info for each input case. If there is no state_info we just create a new one up to the current series length. Should be similar to what we have now, but with no input? Can create a method to return the state info in the same way if its required for earliness.

Maybe I am missing something and the above is not feasible. Just having a reset function is a lot simpler.

patrickzib:
> Maybe I am missing something and the above is not feasible. Just having a reset function is a lot simpler.


Or extend the interface:

`predict()` you always start from scratch and set internal state to the new object.

`update()` you use the internal state to update predictions.

This will also split the streaming and the batch use case clearly.

MatthewMiddlehurst:
> If you are returned a positive decision there is no need to submit the same case again, so the input array size may progressively decrease.

I'm ok with splitting it, but the above remains unsolved. There is no way to tell which cases have been removed if we just store a numpy array.
If we say you have to always input the same number of instances, we are doing unnecessary work for decisions that have already been made (previously you could just exclude these cases and their state_info). 
If we change it so that we no longer make any predictions for cases we have already made a decision on, we have to return None or [-1, -1, -1, ...] for those cases.

Something has to change with the way we currently do it, unless we want to just make predictions whether we have to or not. These are streaming only problems, but I feel streaming is the main use case for these classifiers.

I'm fine with any of the suggestions in the above comments without any of these changes to be honest, but I'm pro-avoiding unnecessary computation (or options to at least 🙂). These are probably small things, but I would like to implement a few more algorithms at a later date, so having an agreed upon framework now is helpful for that.

MatthewMiddlehurst:
True. That makes it so the user has to remove cases which have had decisions made, but that's fine IMO. 

Will put together the split `predict()` and `update()` version at some point.

sktime/classification/early_classification/tests/_test_teaser.py
```diff
@@ -66,15 +73,22 @@ def test_teaser_with_different_decision_maker():
...
+        inv_dec = np.invert(decisions)
+        X_test = X_test[inv_dec, :, :]
+        ec_idx = ec_idx[inv_dec]
 
-        for n in range(10):
-            if decisions[n] and final_decisions[n] == 0:
-                final_probas[n] = probas[n]
-                final_decisions[n] = i
+        X = X_test[:, :, :i]
+        probas, decisions, states = teaser.update_predict_proba(X)
```
patrickzib, line 86: Does it still need state-info to be returned? I think, decisions and probas are the two essential ones?

sktime/classification/early_classification/tests/_test_teaser.py
```diff
@@ -35,15 +35,22 @@ def test_teaser_on_unit_test_data():
...
+        ec_idx = ec_idx[inv_dec]
+
+        X = X_test[:, :, :i]
+        probas, decisions, states = teaser.update_predict_proba(X)
 
-        for n in range(10):
-            if decisions[n] and final_decisions[n] == 0:
-                final_probas[n] = probas[n]
-                final_decisions[n] = i
+        for n in range(len(X)):
```
patrickzib, line 50: isn't this the same as?

`final_decisions = decisions | final_decisions`

Using boolean or between two numpy arrays?

patrickzib: and

`final_probas[decisions] = probas`

sktime/classification/early_classification/tests/_test_teaser.py
```diff
@@ -35,15 +35,22 @@ def test_teaser_on_unit_test_data():
...
-    states = None
+    X_test = X_test[indices, :, :]
+
+    ec_idx = np.arange(0, 10)
+    decisions = np.zeros(len(X_test), dtype=bool)
     for i in teaser.classification_points:
-        X = X_test[indices, :, :i]
-        probas, decisions, states = teaser.predict_proba(X, state_info=states)
+        inv_dec = np.invert(decisions)
+        X_test = X_test[inv_dec, :, :]
```
patrickzib, line 44: It might be cool to have a helper function for this in the base class? 

It should filter X based on the decisions made.

patrickzib (commented):
Some initial thoughts

sktime/classification/early_classification/tests/_test_teaser.py
MatthewMiddlehurst: I'm a bit unsure which bit of code you are referring to, sorry.

If its:
```
       for n in range(len(X)):
            if decisions[n] and final_decisions[ec_idx[n]] == 0:
                final_probas[ec_idx[n]] = probas[n]
                final_decisions[ec_idx[n]] = i
```
then the size of decisions is decreasing as cases are successfully classified. The index of decisions will match the input X, but not necessarily the full dataset.

sktime/classification/early_classification/tests/_test_teaser.py
patrickzib: Ok

sktime/classification/early_classification/tests/_test_teaser.py
patrickzib: Ok, I agree my short version does not help. 

But still: the code is 13 Lines now for the streaming use case. 

```    
    ec_idx = np.arange(0, 10)
    decisions = np.zeros(len(X_test), dtype=bool)
    for i in teaser.classification_points:
        inv_dec = np.invert(decisions)
        X_test = X_test[inv_dec, :, :]
        ec_idx = ec_idx[inv_dec]

        X = X_test[:, :, :i]
        probas, decisions, states = teaser.update_predict_proba(X)

        for n in range(len(X)):
            if decisions[n] and final_decisions[ec_idx[n]] == 0:
                final_probas[ec_idx[n]] = probas[n]
                final_decisions[ec_idx[n]] = i
```

And you have to repeat it every time, you do this. I think, we should think about something much more compact.

MatthewMiddlehurst: It is sort of a consequence of us doing it backwards, we already have the full series and need to shorten it instead of the array growing over time so we can't just overwrite X_test. A lot of it is required to collate the results in the correct order, so we can compare later also.

I agree it is not great, if anyone has a suggestion on how to do it better that would be good. I will try to think of a better way. Probably fine for testing, could be better for the example notebook.

patrickzib: as a start, what about something like this:


 ```
   for i in teaser.classification_points:
        X = X_test[:, :, :i]
        probas, decisions, states, open_idx, final_idx = teaser.update_predict_proba(X, X_idx)

        final_probas[final_idx] = probas
        final_decisions[final_idx] = decisions

        X_test = X_test[open_idx]
        X_idx = open_idx
```

I.e. pass the indices of the current TS to proba as an optional parameter and in this case return the `open_idx` for ts where no decision was taken and  `final_idx` for ts where decision was taken?

File: sktime/classification/early_classification/tests/_test_teaser.py (hunk 3 above)
Comment (MatthewMiddlehurst) [reply sktime/classification/early_classification/tests/_test_teaser.py:44]: Put in a few helper methods for filtering the data in the next commit.

File: sktime/registry/tests/test_lookup.py [hunk 4]
```diff
@@ -98,7 +98,10 @@ def test_all_tags(estimator_scitype):
 
     # there should be at least one tag returned
     # exception: scitypes which we know don't have tags associated
-    if estimator_scitype not in SCITYPES_WITHOUT_TAGS:
+    if estimator_scitype not in SCITYPES_WITHOUT_TAGS or (
+        isinstance(estimator_scitype, list)
+        and all(i not in SCITYPES_WITHOUT_TAGS for i in estimator_scitype)
+    ):
```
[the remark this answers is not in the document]
fkiraly: wait, why can the scitype of an estimator be a list here?
Should it not always be a string, can you kindly explain how we could arrive at a list under your PR?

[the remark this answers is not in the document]
MatthewMiddlehurst: It could be a List prior to my PR, this change is to stop it failing due to the test taking a list of scitypes all of which don't have tags.

The tests checks everything in ` estimator_scitype_fixture`, which is:
```
# selected examples of "search for two types at once to avoid quadratic scaling"
double_estimator_scitypes = [[b[i], b[(i + 3) % n]] for i in range(n)]
# fixtures search by individual scitypes, "None", and some pairs
estimator_scitype_fixture = [None] + BASE_CLASS_SCITYPE_LIST + double_estimator_scitypes
```

I think previously, when it was testing `double_estimator_scitypes` at least one would have tags. The inclusion of the early classifier base class has shifted the list a bit, allowing for the problem double scitype to appear in `double_estimator_scitypes`. At least that is what I have gathered from taking a look and attempting to fix it.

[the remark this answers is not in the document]
fkiraly: ah, I see - because there could now be two scitypes that have no tags.
But, I thought, the early classifiers *do* have tags? So they should not be in the list withou tags?

Anyway, I agree that the test needs fixing, although my approach would be
```
est_list = estimator_scitype if isinstance(estimator_scitype, list) else [estimator_scitype]
if set(est_list).issubset(SCITYPES_WITHOUT_TAGS):
```
which seems less brittle to me

sktime/registry/_tags.py
```diff
@@ -237,6 +237,7 @@
         "bool",
         "contract time setting, does the estimator support limiting max fit time?",
     ),
+    # todo, deprecate. We have base class for this now.
```
fkiraly, line 240: add version please

sktime/registry/tests/test_lookup.py
```diff
@@ -98,7 +98,13 @@ def test_all_tags(estimator_scitype):
 
     # there should be at least one tag returned
     # exception: scitypes which we know don't have tags associated
-    if estimator_scitype not in SCITYPES_WITHOUT_TAGS:
+    if (
```
fkiraly, line 101: see below, I would do this differently. Agree that this needs fixing.

sktime/utils/_testing/estimator_checks.py
```diff
@@ -134,7 +135,7 @@ def _make_fit_args(estimator, **kwargs):
     elif isinstance(estimator, BaseSeriesAnnotator):
         X = make_annotation_problem(**kwargs)
         return (X,)
-    elif isinstance(estimator, BaseClassifier):
+    elif isinstance(estimator, (BaseClassifier, BaseEarlyClassifier)):
```
fkiraly, line 138: I believe this will never be called, so you don´t need to add it here.
You can test it by raising a `RuntimeError` if you arrive here with a `BaseEarlyClassifier`.

`_make_fit_args` is on its way to removal, it is only coupled with exotic scitypes anymore and transformers (until someone reviews my refactor PR 😄 )

sktime/classification/base.py
```diff
@@ -61,7 +61,6 @@ class BaseClassifier(BaseEstimator, ABC):
         "capability:missing_values": False,
         "capability:train_estimate": False,
         "capability:contractable": False,
-        "capability:early_prediction": False,
```
fkiraly, line 64: I guess this is safe without deprecation, as no classifier was using it? Is that correct?

fkiraly (changes requested):
Ok, I've looked at the new design.
I think this is great, going far in the right direction!

I have a few change requests related to this.

* major request (for discussion whether sensible). Now that `state_info` is part of the state, I would turn it from a return of all the functions into an attribute. I.e., remove it from the return of `predict`, `update`, `god`, `the_world` and `everything`, and turn it into an `@attribute` that can be read but not set from outside (only by the methods), or simply a public attribute (non underscore).
* docstrings need improvement: `decisions` and `state_info` need to document what the i-(j-)th entry is, and to which indices of other objects it corresponds.
* docstrings need improvement: is there an assumption that certain dimensions in the panel stay the same? If yes, all docstrings except `fit` should say sth like "dimension ... must be equal to that of `X` seen in `fit`".
* minor, non-blocking thing: I think it's important that docstrings also include the `pd-multiindex` type for `X`, i.e., `pd.DataFrame` with 2-level multiindex, and higher up than the nested data container type.

fkiraly:
Quick quesiton: do coordinates change only from "unsafe" to "safe", or can they become "unsafe" again after having been "safe"? If the former, might be worth considering the decision as an attribute, too.

sktime/registry/tests/test_lookup.py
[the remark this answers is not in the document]
MatthewMiddlehurst: I will try that.

Early classifiers do use tags, clusterers and I think series-annotators do not. Due to the way the doubles are selected, it so just happens that they weren't tested together previously.

sktime/registry/_tags.py
MatthewMiddlehurst: Do we have the capability to deprecate tags now? I was waiting for that, and just left a note so I didn't forget.

sktime/classification/base.py
MatthewMiddlehurst: Yes should be safe, deprecate the tag itself just in case I think.

fkiraly: ok, makes sense. If it was not used for anything so far.

sktime/registry/tests/test_lookup.py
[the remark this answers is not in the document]
fkiraly: hm, that's not true?
I just looked into clusterers and randomly found `TimeSeriesKShapes` an `TimeSeriesKernelKMeans` using `capability:multivariate`

sktime/registry/_tags.py
fkiraly: yes! See the developer guide, deprecation

fkiraly: https://www.sktime.org/en/stable/developer_guide/deprecation.html#deprecating-tags

sktime/registry/tests/test_lookup.py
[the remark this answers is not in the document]
MatthewMiddlehurst: I don't think the registry has clusterers listed as using it, for `capability:multivariate` only ["classifier", "early_classifier"] are listed. Clusterer is part of SCITYPES_WITHOUT_TAGS, current contents are: ["series-annotator", "clusterer"].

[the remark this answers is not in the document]
fkiraly: well, that looks like an oversight then by whoever added that tag!
I would recommend to open an issue, or just fix it here, quickly.

patrickzib:
@MatthewMiddlehurst I was on vacation. Will look into it the coming next week again. Thank you for all the work :)

sktime/registry/tests/test_lookup.py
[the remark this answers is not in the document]
MatthewMiddlehurst: I would create an issue. Not really relevant to this PR, and may be a larger issue with how clustering interacts with tags currently.

MatthewMiddlehurst:
@patrickzib Hope you had a good time!

I think we should be close to getting this completed and merged.

MatthewMiddlehurst:
> major request (for discussion whether sensible). Now that state_info is part of the state, I would turn it from a return of all the functions into an attribute. I.e., remove it from the return of predict, update, god, the_world and everything, and turn it into an @attribute that can be read but not set from outside (only by the methods), or simply a public attribute (non underscore).

I am not against this, but currently the `state_info` returned and the `state_info` stored in the class could be different after the `update_predict`. If we input 10 cases and make a "safe" decision on 3, the updated `state_info` would be returned for all 10 instances, while we would only store the info for the remaining 7 undecided cases. Should be able to remove cases from the stored `state_info` at the start of the method instead of at the end to remedy this though. 

Is @attribute a tag you would put over the field in init? If its public the name would just be self.state_info as the trailing _ is only for fit, correct?

> Quick quesiton: do coordinates change only from "unsafe" to "safe", or can they become "unsafe" again after having been "safe"? If the former, might be worth considering the decision as an attribute, too.

Our current set up wouldn't allow for making a decision on "safe" cases. There is little reason to do so other than requiring multiple consecutive "safe" decisions to make a final decision, but that is handled internally for the algorithms.

sktime/registry/tests/test_lookup.py
MatthewMiddlehurst: Changed the if: to your suggested version above.

sktime/registry/_tags.py
MatthewMiddlehurst: Added the tag to the deprecation dictionaries, I assume that's all that is required for now?

sktime/utils/_testing/estimator_checks.py
MatthewMiddlehurst: Removed it. We will figure out if it was required if the tests fail, I guess!

fkiraly:
> If its public the name would just be self.state_info as the trailing _ is only for fit, correct?

I would make it part of the interface - it´s not something that is being fitted but a statement by the model, and sth every early classifier must have, no?

> If we input 10 cases and make a "safe" decision on 3, the updated state_info would be returned for all 10 instances, while we would only store the info for the remaining 7 undecided cases. Should be able to remove cases from the stored state_info at the start of the method instead of at the end to remedy this though.

I do not understand what you are saying here, since I am missing reference to the formal objects here.
Can you please say this in formal language with a defined reference frame?

MatthewMiddlehurst:
`state_info` is now a public attribute, `predict`/`update `methods no longer return it. Included some of the docstring improvements suggested.

Regarding the suggestions for method parameter docstrings, these are the same as the `BaseClassifier` ones. I think it would be better to keep it consistent for now, but I agree they could all be improved.

fkiraly (changes requested):
Looks good to me now!
Change requests are mostly small things, but blockers since "hard" interface requirement related.

* the base class module docstring needs to include the `state_info` at the top
* an extension template should be written
* add the test class to the `testclass_dict` in `check_estimator`. Non-blocking, but would be nice: in a separate PR, add a paragraph in "test framework" developer guide that this is a place that needs to be looked at when writing a new module.
* kindly reset the notebook before committing

MatthewMiddlehurst:
The latest commit includes an extension template and the other requested changes.

fkiraly (dismissed):
Thanks!

In the extension template, could you kindly add in instructions what the underscore methods can/should read and write to? See the forecasting template.

I think this is needed since it is non-obvious how the `state_info` should be handled.

MatthewMiddlehurst:
Ok, added a bit more instruction for the predict/update docs in the template and base class regarding state_info.

patrickzib:
@MatthewMiddlehurst thank you very much for all the changes.

I am ok with the code. The streaming use case became much better, though it is still not super intuitive to use. But I guess we are as close as possible now. :) 

Some minors:

(a) Is it possible to merge the two calls of

`open_idx, final_idx = teaser.split_indices(open_idx, decisions)`

and

`X_test = teaser.filter_X(X_test, decisions)`

to something like

`X_test , open_idx, final_idx = teaser.split_indices_and_filter(X_test , open_idx, decisions)`


(b) In the jupyter notebook there is a spelling mistake:

...and pass **theSe** to TEASER whenever new data is available...



But I am not fighting for these. I am fine to merge.

fkiraly:
I also think this is almost there. Though, could we hold this off until 0.12.0?
I.e., until after 0.11.3 next week which has bugfixes to existing functionality.

Just to keep "big new features" for 0.12.0.

MatthewMiddlehurst:
No problem holding this off. I will have another look through and see if I can fix the issues Patrick brought up.

MatthewMiddlehurst:
Are there any more issues with this PR or are we ok to merge?

patrickzib (approved):
Good to go!

PR #2249, opened 2022-03-18
[MNT] dockerized test environment

lmmentel:
##### For all contributions

##### For new estimators


lmmentel:
@chrisholder would you mind giving this a look?

fkiraly (changes requested):
Nice!

My main comment is about the docs:
* the docker testing should not be presented as the only way to test locally.
* docker testing should be presented together with the other options for local testing, and an explanation when to use what. E.g., not after the "extending the framework" section where it does not belong?

fkiraly (changes requested):
Great, I think the docker testing is all very useful and nice!

I have some requests about documentation:

* I think the content is in the wrong place, the `testing_framework` page is only about the python code.
* There does not seem to be a clear place where to add the docs, but the current place for local testing related content is in `continuous_integration`.
* Now, local testing is not the same as continuous integration. May I hence suggest to refactor the documentation into another page, "local testing", and move both the new content and the section "unit testing" in that new page? I would give it three sections, one with `check_estimator`, one with `pytest`, and one for docker.

If you think the documentation stuff is too much work, you could factor it out into another PR and get this in while we work on documentation separately?

PR #2250, opened 2022-03-18
[ENH] renamed `fit-in-transform` and `fit-in-predict` to `fit_is_empty`

fkiraly:
This PR renames the tags `fit-in-transform` and `fit-in-predict` to `fit_is_empty`.

It uses deprecation (with target of rename in 0.12.0) functionality in #1962 to carry out #1843.
I have separated to PR to ensure that the deprecation functionality is not drowning in the tag changes from dozens of files.
(along the principle that a PR should optimally introduce a localized change, in terms of files or in terms of content)


fkiraly:
@TonyBagnall, I've done the renaming of tags in this PR.

Issue #2181, opened 2022-03-08
[ENH] Add interface to nixtla AutoARIMA


ngupta23:
**Is your feature request related to a problem? Please describe.**
AutoARIMA from statsmodels is quite slow. 

**Describe the solution you'd like**
A faster alternate was recently released by [Nixtla](https://github.com/Nixtla/statsforecast). Add this as an extension to `sktime`

**Describe alternatives you've considered**
None

**Additional context**
Currently it has a copyleft license. This needs to be resolved in order to be added. See [discussion here].(https://www.linkedin.com/feed/update/urn:li:activity:6902012650709204992?commentUrn=urn%3Ali%3Acomment%3A%28activity%3A6902012650709204992%2C6902192208720617472%29)

![image](https://user-images.githubusercontent.com/33585645/157331210-88311192-21e2-4a9a-80cd-7503e7c86306.png)

ngupta23:
Tracking the license issue here: https://github.com/Nixtla/statsforecast/issues/62

aiwalter:
thx for oopneing this issue @ngupta23

markvrma:
I would like to work on this.

fkiraly:
I believe @fede (one of the developers) is already working on this - but perhaps he'd like some assistance?
@fede, how is it going?

FedericoGarza:
Hi @fkiraly @markvrma! I've opened a PR: #2251. If you have comments on the PR @markvrma, they are welcome. :)

fkiraly:
Merged, hence done.

PR #2251, opened 2022-03-19
[ENH] Add AutoARIMA from StatsForecast 

FedericoGarza:
#### Reference Issues/PRs

#2181 

I've added `AutoARIMA` from `StatsForecast`.

Yes, [`StatsForecast`](https://github.com/Nixtla/statsforecast). I've included it as a soft dependency.

##### For all contributions

##### For new estimators


fkiraly (approved):
Looks good to me, excellent!
Not much iteration needed on this PR :smile:

Non-blocking comment, I wonder how much of the boilerplate can be avoided, or how much is duplicated between `pmdarima` and `statsforecast`.

aiwalter:
Hm now we have 2 classes with same name. I think that is not ideal solution, have never seen a package that has sth like this

FedericoGarza:
Hi @aiwalter ! I could change the name if necessary :)

aiwalter:
@FedericoGarza did you look into the exisitng `AutoARIMA` and whether we could have both in this class with 2 different backends?

aiwalter:
@fkiraly @lmmentel I think we should not release before this here is resolved

fkiraly:
ok, let´s move the release by 1 day then. I feel this needs discussion at the CC

fkiraly:
(sorry for merging, @aiwalter, I assumed tacit approval)

lmmentel:
Wouldn't a simple solution be to add a prefix to the class name e..g `StatsForecastAutoARIMA`?

aiwalter:
yes that is one solution, there was a more general discussion on Slack about this topic. In my opinion it should not go in as it is now at least

FedericoGarza:
Hey @aiwalter @fkiraly @lmmentel! I've created a PR renaming the class as suggested  (#2272).  Let me know what you think😉

Issue #2255, opened 2022-03-19
[ENH] Refactor multiplication of classifier tests


fkiraly:
Looking through the tests, I found that every classifier has a test file, and in it there is `test_[classifiername]_on_unit_test_data` and `test_[classifiername]_on_basic_motions`. This is highly redundant and should be moved to a single file, using `pytest.mark.parametrize` and `all_estimators`.

MatthewMiddlehurst:
On the issue of different parameter config for tests with expected results, we could just add an optional parameter to the test instance and get parameter methods. i.e.

```
    @classmethod
    def get_test_params(cls, parameter_set="default"):
```

```
    @classmethod
    def create_test_instance(cls, test_params="default"):
```

fkiraly:
Hm, not a bad idea, @MatthewMiddlehurst.
Although I worry it will be more difficult on implementers.

How would the implementation pattern look like?
Changing from a list to a dict?

MatthewMiddlehurst:
I'm not really sure what the List is used for currently, what I had in mind was just an if statement really, i.e.

```
    @classmethod
    def get_test_params(cls, parameter_set="default"):
        if parameter_set == "default": 
            return {"n_parameter_samples": 5}
        elif parameter_set == "result_comparison":
            return {"n_parameter_samples": 25}
        else:
            raise ValueError("Invalid test parameter set.")
```

Returning a dictionary could also work, like you said? i.e.

```
    @classmethod
    def get_test_params(cls):
        params = {}
        params["default"] = {"n_parameter_samples": 5}
        params["result_comparison"] = {"n_parameter_samples": 25}
        return params
```

fkiraly:
ah! I see.
Your variant is much more downwards compatible, since you can simply ignore the `parameter_set` argument.

I'd be happy with you adding the `parameter_set` parameter in `BaseObject` with a default of `"default"`, and having two sets for classifiers (this must be explained in the docstrings of course, so you may want to override the default in `BaseClassifier` with a docstring that explains `"result_comparison"`).

MatthewMiddlehurst:
All the classifiers have been updated. The remaining ones are moving to a different base class and set of tests.

fkiraly:
that´s the thresholder and teaser, right?
Should there then an issue for moving them, and this can close?

MatthewMiddlehurst:
I was planning to update them both after the base class is in, shouldn't need a new issue.

PR #2256, opened 2022-03-19
[ENH] Skipping basic motions in classifier tests

fkiraly:
This PR skips all tests of classifiers the "basic motions" dataset.
For validity checking, this is redundant with the "unit test" dataset, and incurs a large amount of compute due to the size of the "basic motions" dataset.

### DIFF
--- a/sktime/classification/compose/_column_ensemble.py
+++ b/sktime/classification/compose/_column_ensemble.py
@@ -223,9 +223,9 @@ class ColumnEnsembleClassifier(BaseColumnEnsembleClassifier):
     Examples
     --------
     >>> from sktime.classification.interval_based import DrCIF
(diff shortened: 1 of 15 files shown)


MatthewMiddlehurst:
Is this still required? We will be refactoring all this soon (#2255), and the tests still seem to take an hour even with them commented.

I dont think it is, but I will approve if you think so as I dont feel strongly either way,

MatthewMiddlehurst:
I can only see a single test there, and that could be changed to be <5 seconds pretty easily. The majority are TestAllForecasters::test_update_predict_predicted_index.

fkiraly:
ok, @MatthewMiddlehurst - closing this then.

PR #2257, opened 2022-03-19
[ENH] refactoring `test_all_classifiers` to test class architecture

fkiraly:
This PR changes `test_all_classifiers` to the test class architecture.

The main change is moving the tests into a class inheriting from `BaseFixtureGenerator` and `QuickTester`.

Changes that deserve special note:
* `test_3d_numpy_input` was removed - this is redundant with `test_methods_do_not_change_state` since that already runs the methods with 3D `numpy` input (using the scenario `ClassifierFitPredict`)
* 100-times copy-pasted tests on unit test and basic motions data were moved to the test suite. This has been carried out for only two instances in this PR since it's tedious. Opened #2255 for doing the rest.
* added multivariate scenario for classifiers. This allows to move the positive case from `test_multivariate_input`, because `test_all_estimators` covers that. Hence, now only the negative (exception raising) case is tested as `test_multivariate_input_exception`.


fkiraly:
All tests pass, except those for some on unit test or basic motions.
Not sure why, I copied the expected outputs 1:1 and thought I set the same random seed.

fkiraly:
@MatthewMiddlehurst , @TonyBagnall, would appreciate if you could take a look - why are the outputs discrepant?

MatthewMiddlehurst:
If the new test is using parameters from `get_test_params` or `config.py`, then the parameters used to produce the results are probably different. Having very quick parameters for regular tests such as `n_estimators=2` does not create good results to compare against.

fkiraly:
> Having very quick parameters for regular tests such as `n_estimators=2` does not create good results to compare against.

What is the purpose of the test, @MatthewMiddlehurst?
If it is stability of the output (e.g., ensuring the algorithm stays the same), either way should be sufficient?

MatthewMiddlehurst:
You could significantly change portions of an implementation, and base estimators could still output the same prediction. You are more likely to record any change with more base estimators.

fkiraly:
> You could significantly change portions of an implementation, and base estimators could still output the same prediction. You are more likely to record any change with more base estimators.

I don't think so - if the estimator is being changed, I don't think it matters what `n_estimators` is in order to pick up that it has changed at all.

MatthewMiddlehurst:
We are limited to a single small dataset with 2 classes and are judging whether it has changed using predicted probabilities, it is entirely possible for the algorithm to change and the results to stay the same. If you are tossing a coin, you are more likely to see heads appear by flipping it 5 times instead of 2.

Whether that is worth it is another discussion. A lot of these take a second or less to run currently (for unit_test at least), so not much gain in lowering the values.

Overall consolidating these tests into a single place is a good idea. We had redoing/cleaning up the tests on a todo list somewhere, but never got around to doing so 🙂.

fkiraly:
>  it is entirely possible for the algorithm to change and the results to stay the same. If you are tossing a coin, you are more likely to see heads appear by flipping it 5 times instead of 2.

Yes, "entirely possible", but I the probability seems extremely low, especially since we are looking at floating point numbers in the predicted probabilities, where every change up until the 2nd digit is detected.
It's also "entirely possible" that you don't detect a change in classifier if you fit and predict on 1.000.000 data points.

> A lot of these take a second or less to run currently

That is not actually true ?!?
The `load_basic_motions` tests are among the most time consuming ones in the suite, with individual iterations of this eating up to a minute or more.

fkiraly:
>We had redoing/cleaning up the tests on a todo list somewhere, but never got around to doing so 🙂.

Would appreciate help with this.
Given the refactor being done, the rest is very formulaic - just move the expected probabilities over in the register.

MatthewMiddlehurst:
> Yes, "entirely possible", but I the probability seems extremely low, especially since we are looking at floating point numbers in the predicted probabilities, where every change up until the 2nd digit is detected.
> It's also "entirely possible" that you don't detect a change in classifier if you fit and predict on 1.000.000 data points.

Given that most of the classifiers are ensembles which use votes to determine probabilities, your possible class probability with 2 base estimators are 0, 0.5 and 1. We can't eliminate the possibility entirely, only mitigate the chance of it happening. 

> A lot of these take a second or less to run currently (for unit_test at least), 

Convenient omission, I support finding an alternate dataset for basic_motions though. It was chosen due to it being a smallish multivariate dataset that is equal length. Like unit_test is derived from the Chinatown dataset, could possibly make another smaller version for basic_motions?

fkiraly:
> Given that most of the classifiers are ensembles which use votes to determine probabilities, your possible class probability with 2 base estimators are 0, 0.5 and 1. We can't eliminate the possibility entirely, only mitigate the chance of it happening.

I'm slightly lost at what we are discussing or what you are arguing for, can you kindly elaborate?
The confusion is possibly entirely on my side.

Bigger data sets? I just translated 1:1 the tests there were for basic motions and unit test data.

MatthewMiddlehurst:
1. The regular test parameters are not sufficient when comparing to recorded results, most of them do not have a wide range of possible outputs due to being ensemble votes. Changing the current ones used is not worth the time gain, except for:

2. basic_motions tests for multivariate capable classifiers, which do take a long time, and we should either look at the parameters used or select a new dataset to reduce that time.

MatthewMiddlehurst:
```
Would appreciate help with this.
Given the refactor being done, the rest is very formulaic - just move the expected probabilities over in the register.
```

I don't mind going through these at some point when this is merged.

fkiraly:
>I don't mind going through these at some point when this is merged.

Well, there are four examples which are necessary to ensure the class works - we need to fix those, and can do the rest later. Do you have a quick way to get these?

TonyBagnall:
these are meant to be correctness tests, to stop inadvertent changes of the algorithm logic. As matt says, they are very small for speed reasons, so cannot capture all changes, that may alter the algorithm, but are still indicative (e.g. if the test is changed in a PR then the algorithm has definitely been altered). I didnt think the current set up is that slow. Of the ten slowest, three are classification. PF tests are the slowest because it has to build all numba functions, and the two column ensemble ones colud probably be improved, that code is not mine. 

 Ultimately, I would like correctness tests to run only if the related code is changed, but this is also tricky due to interdependence.  The end state I am looking for is a test harness only triggered on certain conditions which checks multiple results again published reference results. I dont think I will get there quickly, I have some students working on related aspects.

Re this PR, I am not going to have time to usefully contribute, I am personally neutral on the need to do this now, but recognise the usefulness of getting it done.

fkiraly:
>I didnt think the current set up is that slow.

The basic motions tests cost ca 30sec depending on classifier...

>Re this PR, I am not going to have time to usefully contribute, I am personally neutral on the need to do this now, but recognise the usefulness of getting it done.

@TonyBagnall, I think there's a concrete need within the next month or so: the ability of `tslearn` to reuse our test suite.
Without the classifier test suite refactored, they can't use `check_estimator`, as it does not see the classifier tests.

fkiraly:
@MatthewMiddlehurst:

>The regular test parameters are not sufficient when comparing to recorded results, most of them do not have a wide range of possible outputs due to being ensemble votes. Changing the current ones used is not worth the time gain, except for:

I didn't change any of the test parameter settings, these are as they were. I only refactored the structure of the tests.
You can verify by looking at the changes, no changes to estimators there or to `_config` where the parameters sit.

>basic_motions tests for multivariate capable classifiers, which do take a long time, and we should either look at the parameters used or select a new dataset to reduce that time.

Yes.

MatthewMiddlehurst:
> I didn't change any of the test parameter settings, these are as they were. I only refactored the structure of the tests.
> You can verify by looking at the changes, no changes to estimators there or to _config where the parameters sit.

If the tests are using parameters from `_config`, then the tests classifier parameters have been changed. If it is using the parameters from the individual classifier test files, I don't know why you are getting errors, I can take a look sometime next week.

fkiraly:
> If the tests are using parameters from _config, then the tests classifier parameters have been changed. If it is using the parameters from the individual classifier test files, I don't know why you are getting errors, I can take a look sometime next week.

Are you saying you have two sets of parameters, one from the individual test files, one in `_config`??

MatthewMiddlehurst:
> If the new test is using parameters from get_test_params or config.py, then the parameters used to produce the results are probably different

Yes. See previous comments for why.

> The regular test parameters are not sufficient when comparing to recorded results, most of them do not have a wide range of possible outputs due to being ensemble votes.

fkiraly:
>>Are you saying you have two sets of parameters, one from the individual test files, one in `_config`??

> Yes. See previous comments for why.

😱 when did that happen, this is all over the place

So, I would suggest:
* let's not discuss the "why". I do have comments about that, but they are in the wrong place: I think we should really just restrict to structural refactor with this PR.
* from an architecture perspective, test param settings etc should be in *one* place. I hence suggest to make the first param setting the one that is used for these tests. That would ensure we have 1:1 before/after refactor

MatthewMiddlehurst:
> 😱 when did that happen, this is all over the place

When was the first test file pushed 🙂. These tests are ancient, I'm pretty sure (the format at least).

> from an architecture perspective, test param settings etc should be in one place. I hence suggest to make the first param setting the one that is used for these tests. That would ensure we have 1:1 before/after refactor

The "first param setting" being those in the original test file, not config.py? If so, regular tests may be a bit slower until it is sorted. Not an objection, just noting that they were split in the first place because extra work was not required when there are no results to compare against.

fkiraly:
>When was the first test file pushed 🙂. These tests are ancient, I'm pretty sure (the format at least).

Looks like early Miocene, but I can't get a clear radioisotope reading.

>The "first param setting" being those in the original test file, not config.py? If so, regular tests may be a bit slower until it is sorted. Not an objection, just noting that they were split in the first place because extra work was not required when there are no results to compare against.

Well, we could make it the second setting then?

fkiraly:
@MatthewMiddlehurst, @TonyBagnall, I'm just skipping the tests in the class until we have resolved on how we want to transfer them. The tests are now placeholders, so feel free to play around with it as you see fit. I'm in favour of moving everything in one place, removing the 100-fold boilerplate and only storing the different pieces of config (param settings, expected outputs).

Would be nice if this could be merged then, and we leave the discussion on refactoring the unit and basic motion tests to #2257.

MatthewMiddlehurst (changes requested):
Since the tests are now being skipped, I think we should add back the old tests files temporarily. I will move discussion on how best to refactor these to the issue, as you said.

The rest of the changes are good. Moving our stuff over to this framework and centralising the current tests is an improvement.

fkiraly:
> Since the tests are now being skipped, I think we should add back the old tests files temporarily. I will move discussion on how best to refactor these to the issue, as you said.

Ah, yes, I forgot that. Added those two files back.

PR #2259, opened 2022-03-20
[BUG] `TSInterpolator` and `nested_univ` check fix

fkiraly:
This PR resolved a bug detected through the refactor of the transformers tests in #2522:
* the `nested_univ` checking logic allowed entries of the `pd.DataFrame` to be `np.ndarray`, in contrast to the specification. This has been restricted to `pd.Series`, as specified.
* `TSInterpolator` returned a `pd.DataFrame` filled with `np.ndarray`, this has been changed to a `nested_univ` compliant return, i.e., `pd.DataFrame` filled with `pd.Series`

The fix seems to have uncovered that `PlateauFinder` was broken, see #2263, which is hence excluded from the tests.


fkiraly:
Further, `sp=2` is also not allowed. Is this correct, or should it be strictly bigger than 1 (not 2)?

PR #2260, opened 2022-03-20
[BUG] fixing `STLBootstrapTransformer` error message and docstrings

fkiraly:
This PR fixes an error message and docstrings in `STLBootstrapTransformer`.

@ltsaprounis, can you kindly check whether the error message regarding `sp` is correct now?

### DIFF
--- a/sktime/transformations/bootstrap/_mbb.py
+++ b/sktime/transformations/bootstrap/_mbb.py
@@ -248,9 +248,14 @@ class STLBootstrapTransformer(BaseTransformer):
         -------
         self: reference to self
         """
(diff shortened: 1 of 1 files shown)


fkiraly:
Further, `sp=2` is also not allowed. Is this correct, or should it be strictly bigger than 1 (not 2)?

ltsaprounis:
Thanks @fkiraly! and yes `sp` should be just strictly greater than 1!
While you're at it do you mind changing it to `sp > 1` ?

ltsaprounis (dismissed):

ltsaprounis (approved):

PR #2262, opened 2022-03-20
[ENH] Relaxing `pd-multiindex` mtype to allow string instance index

fkiraly:
This PR relaxes the `pd-multiindex` mtype to allow arbitrary instance indices.

`STLBootstrapTransformer` was (sensibly) returning non-integer instance indices, this relaxation of the mtype now makes that interface compliant.


ltsaprounis:
Is this the right place to also enforce the convention that the time index will be the last level of the multiindex?

fkiraly:
> Is this the right place to also enforce the convention that the time index will be the last level of the multiindex?

Yes, that is enforced through the call to `check_pddataframe_series` which checks whether the time index is one of the supported indices. That way, there´s only one place for that check.

ltsaprounis:
Ah yes that makes sense! I confused it with the hierarchical...

ltsaprounis (approved):

PR #2264, opened 2022-03-20
[ENH] test parameters for `Differencer`

fkiraly:
This PR adds test parameters to the `Differencer` transform.

It ensures that tests are always run with `drop_na=False`, which ensures that `inverse_transform` is inverse to `transform` (otherwise the first observation goes missing).


fkiraly:
included in #2252

PR #2268, opened 2022-03-21
[ENH, BUG] Distance refactor and bug fix

TonyBagnall:
A range of internal changes to distances module and how they are used. The primary change is in the internal representation. From the beginning, we assumed an instance was of shape (m,d), where m is the series length and d is the number of dimensions. The reason for this was because thats how others did it, and there were some native python operations that were more efficient. However, it was later decided to use and (n,d,m) throughout sktime for numpy representations (n is number of cases) . This meant any estimator using the distances would need to transform internally, since it is not possible to know if data is passed in (d,m) or (m,d) shape. This has caused errors in code (e.g. KNN was broken for a while in 2021 because the transform was removed) and its generally far too brittle. This PR converts all distances to expect input series to be shape (d,m). Given it is all compiled to C, there will be no performance hit, its all just loops in the end.  This means we can remove all transposing from distance clusterers and classifiers. 

This PR has got a bit too big, apologies. A few other things includes
1. Classifiers and clusterers using difference series now have option to taking the difference just once. direct calls to ddtw and wddtw still take the difference of the series, but kmeans takes the difference once and switches from ddtw/wddtw to dtw/wdtw
This has been done by introducing another function average_of_slope_transform in ddtw rather than as a transformer. We could make it a transformer if thought desirable  
2. Add doc strings to describe distances. Let me know if any of these are unclear. I could have added more mathematical description for some. However, I have written the maths in sphinx, but I currently have no way of locally debugging it, so not sure what it looks like. I've put references for each.
3. Add correctness tests for distances. These should be restructured to conform to standard parameterised approach, but would rather raise an issue and do that after the thing is in, would be a good first issue
4. Single problem loads unit_test and basic_motions now have option to return numpy, but no other changes in the problem loaders, so hope this is not controversial
5. fixes a bug in EDR and ERP, both of which were not using the correct data in calculations due to a typo in a nested local function
6. fixed a bug in numba conversion that was still going to (m,d) on edge cases
7. fixed pairwise_distance to remove transpose
8. remove all transposing from distance based classifiers and clusterers


sktime/classification/distance_based/_proximity_forest.py
```diff
@@ -155,9 +155,6 @@ def numba_wrapper(distance_measure):
     def distance(instance_a, instance_b, **params):
         instance_a = from_nested_to_2d_array(instance_a, return_numpy=True)
         instance_b = from_nested_to_2d_array(instance_b, return_numpy=True)
```
fkiraly, line 157: would you mind using `convert(instance, from_type="nested_univ", to_type="numpyflat")` instead of `from_nested_to_2d_array`?

Also, are we sure you would like to pass a 2D array and not a 2D array here?

sktime/classification/distance_based/_time_series_neighbors.py
```diff
@@ -141,7 +141,7 @@ def _fit(self, X, y):
             Target values of shape = [n_samples]
         """
         # Transpose to work correctly with distance functions
-        X = X.transpose((0, 2, 1))
+        # X = X.transpose((0, 2, 1))
```
fkiraly, line 144: remove comments

sktime/classification/distance_based/tests/test_time_series_neighbors.py
```diff
@@ -6,11 +6,11 @@
 from sktime.datasets import load_unit_test
 
 distance_functions = [
-    "euclidean",
+    # "euclidean",
```
fkiraly, line 9: why are we testing less?

sktime/classification/distance_based/tests/test_time_series_neighbors.py
```diff
@@ -28,8 +28,8 @@
 def test_knn_on_unit_test():
```
fkiraly, line 28: this will likely clash with the refactor of @MatthewMiddlehurst - this is now part of the classifier test class. Should be moved there and/or uniformized.

sktime/distances/base/_base.py
```diff
@@ -73,12 +73,12 @@ def distance_factory(
 
         no_python_callable = self._distance_factory(x, y, **kwargs)
 
-        if not hasattr(no_python_callable, "signatures"):
-            raise RuntimeError(
-                "The distance metric specified could not be no_python"
-                "compiled. Try again and if the problem persists raise"
-                "an issue."
-            )
+        # if not hasattr(no_python_callable, "signatures"):
```
fkiraly, line 76: remove comments. Also, what are we removing here, and why?

sktime/classification/distance_based/_proximity_forest.py
TonyBagnall: so proximity forest is not part of the PR and wont be changed here, there is a PR for PF, coulkd you ask there instead?

sktime/distances/base/_base.py
TonyBagnall: removed, @chrisholder what was this for?

sktime/classification/distance_based/tests/test_time_series_neighbors.py
TonyBagnall: we can resolve any clashes once one of the PRs are in, it will be simple

sktime/classification/distance_based/_proximity_forest.py
fkiraly: where is it?

TonyBagnall: https://github.com/alan-turing-institute/sktime/pull/2112

fkiraly: that is abandoned, per your statement

TonyBagnall: its not abandoned, just delayed

lmmentel:
:open_mouth: this is a massive PR

TonyBagnall:
> 😮 this is a massive PR

yes sorry about that, it touches a lot of files, the distances module is still experimental. I think we need to enforce [n_cases][n_channels][n_timepoints] throughout sktime, and that has got caught up in a lot of chris's changes/enhancements. I will wait to merge this until we have finished the associated paper though, just so we have some ground truth results beyond tiny problems.

MatthewMiddlehurst (approved):

sktime/distances/_resolve_metric.py
```diff
@@ -22,9 +22,9 @@ def _resolve_metric(
     metric: str or Callable or NumbaDistance
         The distance metric to use.
     x: np.ndarray (2d array)
-        First timeseries.
+        First time eries.
```
MatthewMiddlehurst, line 25: time eries

TonyBagnall:
note to self, wait until after Franz does pydata thing to merge this

fkiraly:
> note to self, wait until after Franz does pydata thing to merge this

Indeed, the pydata thing is done now 😄

PR #2269, opened 2022-03-21
[ENH] Imputer improved - correct docstrings, leakage free mean/median imputation

fkiraly:
This PR addresses the concern raised in https://github.com/alan-turing-institute/sktime/discussions/2267 about information leakage on the test set when using the `Imputer` transformer in a forecasting pipeline.

Changes in this PR:
* improved docstrings, some corrections, and a highlight which methods are prone to information leakage
* adds two new methods, variants of `mean` and `median`, which use the mean/median from fit
* changes `get_test_params` so all admissible method strings are tested. This should be a "fast" estimator so no huge overhead in terms of compute (hopefully)
* replaced `Z` by `X` in `_transform` and named the output `Xt`, for better readability throughout

Also fixes a bug in the example in `ForecastingGridSearchCV` which used an unknown method `"last"` for the imputer. This never raised an error, since the data in the example did not have nan in it (and error handling was less stringent than in this PR).

### DIFF
--- a/sktime/forecasting/model_selection/_tune.py
+++ b/sktime/forecasting/model_selection/_tune.py
@@ -420,7 +420,7 @@ class ForecastingGridSearchCV(BaseGridSearch):
     ...         "forecaster": [ThetaForecaster(sp=12)],
     ...     },
     ...     {
-    ...         "imputer__method": ["mean", "last"],
+    ...         "imputer__method": ["mean", "median"],
     ...         "forecaster": [ExponentialSmoothing(sp=12)],
     ...         "forecaster__trend": ["add", "mul"],
     ...     },
--- a/sktime/transformations/series/impute.py
+++ b/sktime/transformations/series/impute.py
@@ -23,6 +23,13 @@ class Imputer(BaseTransformer):
     The Imputer transforms input series by replacing missing values according
     to an imputation strategy specified by `method`.
 
+    IMPORTANT NOTE: this transformer may lead to information leakage on the test set
+        for some choices of method when used for forecasting.
+        It is safe for time series classification, or in a panel setting where entire
+        series are either in the training or in the test set.
+        Leakage-free methods in all settings are:
+            "constant", "mean_fit", "median_fit"
+
     Parameters
     ----------
     method : str, default="drift"
--- a/sktime/transformations/series/impute.py
+++ b/sktime/transformations/series/impute.py
@@ -93,11 +93,11 @@ class Imputer(BaseTransformer):
     }
 
     ALLOWED_METHODS = [
+        "mean",
         "drift",
         "linear",
         "nearest",
         "constant",
-        "mean",
         "mean_fit",
         "median",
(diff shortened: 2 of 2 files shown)


aiwalter (commented):
I would rename the methods to use `-` instead `_` in name. In other classes we also do so, e.g. here: https://www.sktime.org/en/stable/api_reference/auto_generated/sktime.forecasting.compose.AutoEnsembleForecaster.html

Now as you added this PR, I am thinking about if we should also have the option to select the fit and the transform data to be used as training data. So actually there are 3 possible data sets an algo can be fitted on:
- fit
- transform
- fit + transform

This does not matter for all implemented methods, but it matters for more than median and mean. It also matters for  "linear" and "forecaster". What do you think? I think what you did is already the right direction, but then at least we should then also have "linear_fit" and "forecaster_fit", no? Or we add a param like "fit_data"  which can be a values out of ["fit", "transform", "fit_transform"]?

fkiraly:
> What do you think?

I think there's not a difference in terms of implementation? 
Because there's in reality two cases: you want to fit on sth that's different from transform. Or you want to fit on the same thing that you transform on.

What you call the X-es (fit/transform etc) does not matter, or how you mentally label them?

aiwalter:
I think we can achieve this by putting all into `fit`, I would have expected it there actually. Idk why I implemented it in `transform`. But I would prefer this since then we dont need to add any params. I can also work on it if you want

fkiraly:
> I think we can achieve this by putting all into fit, I would have expected it there actually.

Fit doesn't return anything, so that wouldn't work?

aiwalter:
We need to save some results like mean in `self`.

aiwalter:
Methods like `ffill` can just go in `transform`

fkiraly:
Hm, I disagree with all being in `fit`, because for many of the methods we don't need it in `fit`?
There are genuine cases where we want to impute on the test batch, in TSC.

If we move everything to `fit`, that will no longer be possible.

Further, that change would not be downwards compatible.

aiwalter:
I think downward compatibility should not be an argument for continuing sth in a wrong way. When you want to impute only on a test data, you can just give the test data to `Imputer.fit_transform()`. I dont like the adding of this params, the slim fix would be to move those methods that are actually doing a fit into `fit()`. As said, I can work on it if you dont want to. Can we add a deprecation just now before the release to warn user about the affected methods?

fkiraly:
>When you want to impute only on a test data, you can just give the test data to `Imputer.fit_transform()`.

Imo, incorrect argument, since that would not work *in a pipeline*.

fkiraly:
> I dont like the adding of this params, the slim fix would be to move those methods that are actually doing a fit into fit().

I thought I did that? There is a case for having summaries computed on the `transform` bit.

fkiraly:
Superseded by https://github.com/alan-turing-institute/sktime/pull/2382

PR #2278, opened 2022-03-22
Plateau max length

srggrs:
this PR extend the `PlateauFinder` by adding an optional `max_length` parameter for searching plateaus between a minimum and maximum length.

No

Looking at the `/sktime/sktime/transformations/panel/summarize/_extract.py`

##### For all contributions
- [ x ] I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/main/.all-contributorsrc).
- [ x ] Optionally, I've updated sktime's [CODEOWNERS](https://github.com/alan-turing-institute/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
- [ x ] I've added unit tests and made sure they pass locally.

##### For new estimators

### DIFF
--- a/sktime/transformations/panel/summarize/_extract.py
+++ b/sktime/transformations/panel/summarize/_extract.py
@@ -31,9 +31,14 @@ class PlateauFinder(_PanelToPanelTransformer):
 
     _tags = {"fit-in-transform": True, "univariate-only": True}
 
-    def __init__(self, value=np.nan, min_length=2):
+    def __init__(self, value=np.nan, min_length=2, max_length=None):
         self.value = value
         self.min_length = min_length
+        self.max_length = max_length
+        # param checks
+        if min_length > max_length:
+            raise ValueError(f'Max length {max_length} less than Min length {min_length}')
+
         super(PlateauFinder, self).__init__()
 
     def transform(self, X, y=None):
@@ -81,8 +86,12 @@ class PlateauFinder(_PanelToPanelTransformer):
             lengths = ends - starts
 
             # filter out single points
-            starts = starts[lengths >= self.min_length]
-            lengths = lengths[lengths >= self.min_length]
+            selection = lengths >= self.min_length
+            if self.max_length is not None:
+                selection = selection & lengths <= self.max_length
+
+            starts = starts[selection]
+            lengths = lengths[selection]
 
             self._starts.append(starts)
             self._lengths.append(lengths)
--- a/sktime/transformations/panel/summarize/_extract.py
+++ b/sktime/transformations/panel/summarize/_extract.py
@@ -36,8 +36,10 @@ class PlateauFinder(_PanelToPanelTransformer):
         self.min_length = min_length
         self.max_length = max_length
         # param checks
-        if min_length > max_length:
-            raise ValueError(f'Max length {max_length} less than Min length {min_length}')
+        if max_length is not None and min_length > max_length:
+            raise ValueError(
+                f'Max length {max_length} less than Min length {min_length}'
+            )
 
         super(PlateauFinder, self).__init__()
 
--- a/sktime/transformations/panel/summarize/_extract.py
+++ b/sktime/transformations/panel/summarize/_extract.py
@@ -90,7 +90,7 @@ class PlateauFinder(_PanelToPanelTransformer):
             # filter out single points
             selection = lengths >= self.min_length
             if self.max_length is not None:
-                selection = selection & lengths <= self.max_length
+                selection = selection & (lengths <= self.max_length)
 
             starts = starts[selection]
             lengths = lengths[selection]
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -184,6 +184,7 @@ Thanks goes to these wonderful people ([emoji key](https://allcontributors.org/d
     <td align="center"><a href="https://github.com/vincent-nich12"><img src="https://avatars3.githubusercontent.com/u/36476633?v=4?s=100" width="100px;" alt=""/><br /><sub><b>vincent-nich12</b></sub></a><br /><a href="https://github.com/alan-turing-institute/sktime/commits?author=vincent-nich12" title="Code">💻</a></td>
     <td align="center"><a href="https://github.com/vollmersj"><img src="https://avatars2.githubusercontent.com/u/12613127?v=4?s=100" width="100px;" alt=""/><br /><sub><b>vollmersj</b></sub></a><br /><a href="https://github.com/alan-turing-institute/sktime/commits?author=vollmersj" title="Documentation">📖</a></td>
     <td align="center"><a href="https://github.com/xiaobenbenecho"><img src="https://avatars.githubusercontent.com/u/17461849?v=4?s=100" width="100px;" alt=""/><br /><sub><b>xiaobenbenecho</b></sub></a><br /><a href="https://github.com/alan-turing-institute/sktime/commits?author=xiaobenbenecho" title="Code">💻</a></td>
+    <td align="center"><a href="https://github.com/srggrs"><img src="https://avatars.githubusercontent.com/u/34258464?v=4" width="100px;" alt=""/><br /><sub><b>serg</b></sub></a><br /><a href="https://github.com/alan-turing-institute/sktime/commits?author=srggrs" title="Code">💻</a></td>
   </tr>
 </table>
 
--- a/sktime/transformations/panel/summarize/tests/test_PlateauFinder.py
+++ b/sktime/transformations/panel/summarize/tests/test_PlateauFinder.py
@@ -52,3 +52,54 @@ def test_PlateauFinder(value):
     for i in range(n_samples):
         np.testing.assert_array_equal(actual_starts[i], expected_starts[i])
         np.testing.assert_array_equal(actual_lengths[i], expected_lengths[i])
+
+
+@pytest.mark.parametrize("value", [np.nan, -10, 10, -0.5, 0.5])
+def test_PlateauFinderMaxLength(value):
+    # generate test data
+    value = np.nan
+    X = pd.DataFrame(
+        pd.Series(
+            [
+                pd.Series([value, 1, 2, 3, value, 2, 2, 3]),  # nan at start
+                pd.Series([value, value, 3, 3, value, 2, 2, 3]),
+                pd.Series([0, value, value, value, value, value, 2, value]),
+                # nan at end
+                pd.Series([0, value, value, 3, 4, 5, value, value]),
+                pd.Series([2, value, value, value, 2, value, 3, 1]),
+                pd.Series([0, value, value, 3, value, value, 2, 0]),
+                pd.Series([0, 0, value, value, value, value, 2, value]),
+            ]
+        )
+    )
+    n_samples = X.shape[0]
+
+    t = PlateauFinder(value=value, min_length=2, max_length=3)
+    Xt = t.fit_transform(X)
+
+    actual_starts = Xt.iloc[:, 0]
+    actual_lengths = Xt.iloc[:, 1]
+
+    expected_starts = [
+        np.array([]),
+        np.array([0]),
+        np.array([]),
+        np.array([1, 6]),
+        np.array([1]),
+        np.array([1, 4]),
+        np.array([]),
+    ]
+    expected_lengths = [
+        np.array([]),
+        np.array([2]),
+        np.array([]),
+        np.array([2, 2]),
+        np.array([3]),
+        np.array([2, 2]),
+        np.array([]),
+    ]
+
+    # compare results
+    for i in range(n_samples):
+        np.testing.assert_array_equal(actual_starts[i], expected_starts[i])
+        np.testing.assert_array_equal(actual_lengths[i], expected_lengths[i])


ltsaprounis:
@srggrs - the code quality checks are failing. These are usually simple formatting rules (e.g. a line should not exceed 88 characters in length). To avoid any surprises from this, try configuring the [pre-commit hooks](https://pre-commit.com). To do this just follow these steps:
1. `pip install --editable .[dev]`
2. `pre-commit install`

In the next commit you try, you will be able to see all the checks and their results/recommendations. If any of them fails, you'll need to fix the issue and then `git add` and `git commit` again.

srggrs:
I understand but my changes respect the code quality (see below):
```bash
pre-commit run --files sktime/transformations/panel/summarize/_extract.py
```
output:
```bash
Check for added large files..............................................Passed
Check for case conflicts.................................................Passed
Check for merge conflicts................................................Passed
Check for broken symlinks............................(no files to check)Skipped
Check Yaml...........................................(no files to check)Skipped
Debug Statements (Python)................................................Passed
Fix End of Files.........................................................Passed
Fix python encoding pragma...............................................Passed
Fix requirements.txt.................................(no files to check)Skipped
Trim Trailing Whitespace.................................................Passed
isort....................................................................Passed
black....................................................................Passed
flake8...................................................................Passed
nbqa-black...........................................(no files to check)Skipped
nbqa-isort...........................................(no files to check)Skipped
nbqa-flake8..........................................(no files to check)Skipped
pydocstyle...............................................................Passed
shellcheck...........................................(no files to check)Skipped
```
So the code quality check fails for other files that I haven't modified or am I wrong?

fkiraly:
check here for a summary of what's failing:
image

mantmic:
I've added missing docstrings to the tests if you're happy to run the workflows

fkiraly:
@mantmic, I think there's one still missing?

mantmic:
Ah cheers Franz, my mistake. Should be sorted now.

fkiraly (changes requested):
Change requests:
* Please do not edit the `contributors.md` directly.
Contributors should be adde to `all-contributorsrc`
* the new argument `max_length` must be described in the docstring, and how it impacts the output or logic

fkiraly:
Great, thanks! Looks good, I have a few change requests above.

fkiraly:
FYI, I just merged a larger refactor of sktime transformers which enables hierarchical functionality for them, including for PlateauFinder, I believe. I recommend to update to the latest main.

mantmic:
Sure thing, I've
- Merged main into branch
- Added docstring for max_length parameter
- Removed new entry on contributors.md
- Added contributors to all-contributorsrc

fkiraly (changes requested):
Thanks!

Almost there.
I think the docstring needs to be clearer about:
* what the transformer does, in general. I.e., given a series, what is being extracted precisely? Which format are "starting indices" and "lengths" precisely returned in? This is not your fault that it´s not there, but it is a condition for the next item.
* in the context of that description, it should be said what min/max-length args do. These are just working as a filter, is that correct?

srggrs:
Thx @fkiraly, will change the documentation asap. 2 more questions:
1. Should we change the online documentation as well? Or that is autogenerated from the python doc?
2. Should we add ourself in the [`__author__`](https://github.com/Gridsight/sktime/blob/plateau-max-length/sktime/transformations/panel/summarize/_extract.py#L5) at the beginning of `_extract.py`, as well?

Thx again!

fkiraly:
> Should we change the online documentation as well? Or that is autogenerated from the python doc?

It is auto-generated (as long as the estimator is linked there already, which it is afaik)

> Should we add ourself in the [__author__](https://github.com/Gridsight/sktime/blob/plateau-max-length/sktime/transformations/panel/summarize/_extract.py#L5) at the beginning of _extract.py, as well?

Yes, since you made a substantial change and this algorithm is not "owned".

srggrs:
@fkiraly I added some extra documentation with an example. I can make things more obvious (e.g. adding input and outputs dataframe like these).
|   | 0                                       |
|--:|:----------------------------------------|
| 0 | 0 NaN 1 NaN 2 3 ... 157 NaN 158 2 159 3 |
| 1 | 0 NaN 1 NaN 2 NaN ... 157 2 158 2 159 2 |
| 2 | 0 2 1 2 2 3 ... 157 3 158 2 159 3       |

sktime/transformations/panel/summarize/_extract.py
```diff
@@ -17,18 +22,53 @@
...
+    >>> from sktime.transformations.panel.summarize import PlateauFinder
+    >>> # generate some data
+    >>> X = pd.DataFrame(
+            pd.Series([
+                pd.Series([-1, -1, 3, 3, -1, 2, 2, 3]),
+                pd.Series([0, -1, -1, -1, -1, -1, 2, -1]),
+                pd.Series([2, -1, -1, -1, 2, -1, 3, 1]),
+                pd.Series([1, -1, -1, 3, -1, -1, 2, 0]),
+            ])
+        )
```
ltsaprounis, line 68: Tests are failing because of this example. If you add `...` in all indented rows it should work. 
e.g. [example syntax for doc examples](https://github.com/alan-turing-institute/sktime/blob/3194e3ee961cdb47bac3b487a779582b134eab68/sktime/transformations/bootstrap/_mbb.py#L154) 
To test the examples locally you can run the following command in your terminal: 
`python -m doctest -o=ELLIPSIS -v sktime/transformations/panel/summarize/_extract.py`

fkiraly: > To test the examples locally

Oh, I did not know that, @ltsaprounis!
This would be useful in our "adding a new estimator" or "testing" dev documentation.

srggrs: thx! fixed in [a4d2eae](https://github.com/alan-turing-institute/sktime/pull/2278/commits/a4d2eae6a1db0f959c6d821e2c9864d8d71da507)

ltsaprounis: Yep! that's done

ltsaprounis: Yep! that's done

sktime/transformations/panel/summarize/_extract.py
```diff
@@ -1,8 +1,13 @@
 # -*- coding: utf-8 -*-
-"""Sequence feature extraction transformers."""
+"""Sequence feature extraction transformers.
+
+These transformers expect a dataframe of series, usually univariate and
```
fkiraly, line 4: please remove these three lines - this is true for all transformers

sktime/transformations/panel/summarize/_extract.py
```diff
@@ -17,18 +22,53 @@
 class PlateauFinder(BaseTransformer):
     """Plateau finder transformer.
 
-    Transformer that finds segments of the same given value, plateau in
-    the time series, and
-    returns the starting indices and lengths.
+    Transformer that finds segments of the same given value, aka plateau(s) in
+    the time series, and returns the starting indices and lengths. It expects a
+    dataframe with one column of nested Series and it extracts the plateau for
```
fkiraly, line 27: the format assumption is not correct - you can pass non-nested data frames to the transformers, too.
I assume the features are computed by series. Can you simply detail what happens for a single series, e.g., a single `pd.Series` object (not a nested data frame)? That would be easier to understand.

It should work out of the box.

sktime/transformations/panel/summarize/_extract.py
```diff
@@ -17,18 +22,53 @@
...
+    each series. The output is a dataframe with two columns of nested lists,
+    where the first column has the lists with the starting indices of the
+    plateau(s), and the second column has the lists of the length of the
+    plateau(s). If no flate segements are found, the output is a dataframe of
+    nested empty lists.
+
+    In this context the `value` parameter tells the transformer to look for flat
+    segments with a specific value, the `min_lenght` tells to filter
```
fkiraly, line 35: filter out, or filter in?

fkiraly, line 35: is it always plateaus of that given value?

fkiraly (changes requested):
Ok, thanks!

I think this is almost ready to go! I've left some comments around the docstrings.

sktime/transformations/panel/summarize/_extract.py
srggrs: yep. If the value is not given it is set to `NaN`

sktime/transformations/panel/summarize/_extract.py
srggrs: done in cd28ff93

sktime/transformations/panel/summarize/_extract.py
srggrs: changed in cd28ff93

sktime/transformations/panel/summarize/_extract.py
srggrs: mmm not sure what you mean... If the you pass a dataframe like this (where the timeseries is spread into multiple columns ) is not going to work:
|    |   0 |   1 |   2 |   3 |   4 |   5 |   6 |   7 |
|---:|----:|----:|----:|----:|----:|----:|----:|----:|
|  0 |  -1 |  -1 |   3 |   3 |  -1 |   2 |   2 |   3 |
|  1 |   0 |  -1 |  -1 |  -1 |  -1 |  -1 |   2 |  -1 |
|  2 |   2 |  -1 |  -1 |  -1 |   2 |  -1 |   3 |   1 |
|  3 |   1 |  -1 |  -1 |   3 |  -1 |  -1 |   2 |   0 |
Output:
```bash
X must be in an sktime compatible format, of scitype Series, Panel or Hierarchical, for instance
a pandas.DataFrame with sktime compatible time indices, or with MultiIndex and lowest level
a sktime compatible time index. allowed compatible mtype format specifications
are: ['pd.Series', 'pd.DataFrame', 'np.ndarray', 'nested_univ', 'numpy3D', 'pd-multiindex',
'df-list', 'pd_multiindex_hier'] See the data format tutorial examples/AA_datatypes_and_datasets.ipynb
```

fkiraly: I mean:

* I think the plateaufinder transform is applied per series, in a panel
* can you please write in the docstring what precisely happens, for a series `y(t_0), ..., y(t_k)`?

srggrs: I'm not sure what type the plateau finder will accept but I will check now. I suppose you are talking about the "columns" of this dataframe (the one you call "panel"):
image
While the plateau example that is already in the code base [here](https://github.com/alan-turing-institute/sktime/blob/main/examples/plateau_finder.ipynb), I think, use a type of `nested_univ`. And that is the one I based the docstring on.

I will write in the docstring of what happens to a series of timepoints for one "instance", then will test a couple of more input data the `PlateauFinder` can work with. How does that sound?

srggrs: I can confirm that the same example data, reshaped as panel with multindex like this below are not working in the when calling `fit_transform` over it.
image
Output error:
```bash
X must be in an sktime compatible format, of scitype Series, Panel or Hierarchical, for instance
a pandas.DataFrame with sktime compatible time indices, or with MultiIndex and lowest level
a sktime compatible time index. allowed compatible mtype format specifications
are: ['pd.Series', 'pd.DataFrame', 'np.ndarray', 'nested_univ', 'numpy3D', 'pd-multiindex',
'df-list', 'pd_multiindex_hier'] See the data format tutorial
examples/AA_datatypes_and_datasets.ipynb
```
Neither this works:
image
Error:
```bash
X must be univariate but is not
```
So I assume the `PlateauFinder` wants the data shapes as `nested_univ` with a single column....?

ltsaprounis:
Looks like the only problem is a docstring. Here is a link to the line in the logs:
https://github.com/alan-turing-institute/sktime/runs/6135840487?check_suite_focus=true#step:6:73

srggrs:
> Looks like the only problem is a docstring. Here is a link to the line in the logs: https://github.com/alan-turing-institute/sktime/runs/6135840487?check_suite_focus=true#step:6:73

done in [db64b14](https://github.com/alan-turing-institute/sktime/pull/2278/commits/db64b14f39530589ab8e1f774e0773c632590fcb)

sktime/transformations/panel/summarize/_extract.py
fkiraly: can you paste some code please?

fkiraly:
@srggrs, thanks for this!

You mentioned above that the plateau finder breaks for "panel" dataframe input, could you kindly provide a reproducible code snippet where this breaks? I do not hink it should break, it sounds like a bug.

srggrs:
> @srggrs, thanks for this!
> 
> You mentioned above that the plateau finder breaks for "panel" dataframe input, could you kindly provide a reproducible code snippet where this breaks? I do not hink it should break, it sounds like a bug.

@fkiraly have look at this (need to have `pandas>=1.3`)
```python
import pandas as pd
from sktime.transformations.panel.summarize import PlateauFinder

# initialise df as type "nested_univ" as per docstring
X = pd.DataFrame(
     pd.Series([
         pd.Series([-1, -1, 3, 3, -1, 2, 2, 3]),
         pd.Series([0, -1, -1, -1, -1, -1, 2, -1]),
         pd.Series([2, -1, -1, -1, 2, -1, 3, 1]),
         pd.Series([1, -1, -1, 3, -1, -1, 2, 0]),
     ])
    )
# make df to type "panel"
X.index.name = 'instances'
X = (
    X.assign(timepoints=X[0].apply(
        lambda row: pd.Series([row.index.values]))
    )
    .explode([0, 'timepoints'])
    .set_index('timepoints', append=True)
)

# apply plateau to df
t = PlateauFinder(value=-1)
Xt = t.fit_transform(X)
```

srggrs:
Any news about this?

fkiraly:
Ah, sorry, I thought I had replied to this.

The failure is due to `X` not being in an `sktime` supported format - it is not in fact compliant with the `nested_univ` specification. If you print `X`, you already see there are no `pd.Series` *inside*  the `pd.DataFrame`. A more formal way would be to do this:

```python
from sktime.datatypes import check_raise

check_raise(X, mtype="nested_univ")
```

(like in the data format tutorial) and it tells you the same.

fkiraly:
This should work though, the input is in the `numpy3D` format.
However, it fails, so somewhere there might be a bug?

```python
import numpy as np
from sktime.transformations.panel.summarize import PlateauFinder

# initialise df as type "nested_univ" as per docstring
X = np.array(
    [[[-1, -1, 3, 3, -1, 2, 2, 3]],
     [[0, -1, -1, -1, -1, -1, 2, -1]],
     [[2, -1, -1, -1, 2, -1, 3, 1]],
     [[1, -1, -1, 3, -1, -1, 2, 0]],
     ]
    )

# apply plateau to df
t = PlateauFinder(value=-1)
Xt = t.fit_transform(X)
```

fkiraly:
in terms of how to get this PR in, I think we still neened to work on two small things. The "main work" is already done.

* ensure that the docstring precisely says what comes out of `transform`, in relation to the input of `transform`, possibly also with an example. This can still be improved.
* ensure we understand what is happening with the bug. It may be the return is not compliant.

I understand that the docstring was incomplete and confusing already before you started working on this - but let's fix that while we are at it, especially since you are adding complexity to it.
Similarly, the bug may not be in `PlateauFinder`, but we should investigate whether it is and then ensure the right person (you or a core developer familiar with the root cause) is on it.

fkiraly:
@srggrs, are you still working on this? Almost there!

srggrs:
@fkiraly sorry I been giving priority to other things, and I am not working on this atm. Also it seems the core function to calculate the plateau is quite fast, while the other functions (I suspect the check `_check_X_y` or the vectorization in the `fit` method of the `BaseTransformer` class are quite slow in comparison)

fkiraly:
>@fkiraly sorry I been giving priority to other things, and I am not working on this atm. 

No problem. As long as you are aware that we probably won't merge if it is not described what the change precisely does - users need to be able to read up on it.

> Also it seems the core function to calculate the plateau is quite fast, while the other functions (I suspect the check `_check_X_y` or the vectorization in the `fit` method of the `BaseTransformer` class are quite slow in comparison)

Hm, that may well be - how do you infer this?

PR #2224, opened 2022-03-15
Dummy pr test hierarchical forecast

ltsaprounis:
Dummy PR to test issue #2194
Added some logging to see whether the issue comes from the pandas dtypes in `_make_hierarchical`
Tests pass locally on Mac & python 3.8

PR #2279, opened 2022-03-22
Fix windows bug with index freq in VetorizedDF.__getitem__

### LINKED ISSUE #2194: [BUG] odd failure in `fh._coerce_to_period` that occurs only on windows, in context of hierarchical forecast
When making hierarchical forecasts on default output of `_make_hierarchical`, `fh.coerce_to_period` fails, on a windows run:
https://github.com/alan-turing-institute/sktime/runs/5503723513

This is quite strange, and I cannot reproduce the behaviour locally on my windows machine (python 3.8 or 3.9).

It looks like `freq` is not correctly preserved in subsetting from a hierarchical multi-index ... but only under specific conditions on a windows system???

Expected behaviour: the test does not fail, just like on linux/max and as on my local system.

ltsaprounis:
#### Reference Issues/PRs

Fixes #2194 

After a lot of debugging in #2224, I found that the tests are failing for Windows because the series obtained when looping through `VetorizedDF` have frequency `None` in Windows and `"D"` for both Linux and Mac. The fix is to  add a new function that uses `pd.infer_freq` to update the frequency of each series in `VetorizedDF.__getitem__`.

##### For all contributions

### DIFF
--- a/sktime/datatypes/_vectorize.py
+++ b/sktime/datatypes/_vectorize.py
@@ -4,7 +4,6 @@
 
 Contains VectorizedDF class.
 """
-
 import pandas as pd
 
 from sktime.datatypes._check import check_is_scitype, mtype
@@ -150,6 +149,7 @@ class VectorizedDF:
         X = self.X_multiindex
         ind = self.get_iter_indices()[i]
         item = X.loc[ind]
+        item = _enforce_index_freq(item)
         # pd-multiindex type (Panel case) expects these index names:
         if self.iterate_as == "Panel":
             item.index.set_names(["instances", "timepoints"], inplace=True)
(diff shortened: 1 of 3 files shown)


fkiraly (approved):
How odd!

I guess this is fine as a workaround, but there must be an underlying `pandas` bug, I suspect.
Have you been able to identify what that is? We should probably report it.

ltsaprounis:
> How odd!
> 
> I guess this is fine as a workaround, but there must be an underlying `pandas` bug, I suspect. Have you been able to identify what that is? We should probably report it.

Not really. Will throw it in the pandas gitter at some point soon to see if they get what's wrong. We're also using an earlier version of pandas so it might be fixed in the later versions.

ltsaprounis:
@fkiraly FYI - question in the [pandas gitter](https://gitter.im/pydata/pandas#)
image

PR #2282, opened 2022-03-23
[ENH] ARCH model (arch package) as forecaster

### LINKED ISSUE #2173: [ENH] Add adapter for `arch` package.

Financial modeling usually deals with modeling the volatility of time series data, ARCH, GARCH, etc. are common models used for this purpose. 

Add support for https://github.com/bashtage/arch (as an adapter) in sktime. This will be very useful for folks in the financial industry.
This is detailed in this video as well for reference: https://www.youtube.com/watch?v=74rDhJexmTg.

None available at this time.

**Additional context**

- Refer to the video above to see how these models are used.
- Further examples can be found in the repo's example folder: https://github.com/bashtage/arch/tree/main/examples

Vasudeva-bit:
#### Reference Issues/PRs

Implements `arch` estimator, Fixes #2173

A new estimator is added i.e., the `ARCH` model to forecast the volatility. Using the `arch` package, an `arch` estimator is implemented.

Yes, the `arch` package is the soft dependency introduced for implementing the `arch` estimator.

Review the implementation of the `arch` model in sktime and also the updated doc, API reference, other files w.r.t to `arch` estimator.
#### Any other comments?

As far as I know, the `arch` package is not supporting the `multivariate` arch models, so need to work on that.! The `update` functionality is implicitly available in the `fit` method of the `arch` model (so, skipped its implementation). Methods such as `predict_intervals` or `predict_quantiles` are yet to be implemented as the `arch` package doesn't support (studying this closely). 

Upcoming updates include:
1. Adding the estimator to online documentation.
2. Add an example notebook to showcase how the `arch` estimator works.
##### For all contributions

- I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/main/.all-contributorsrc).

- I've updated sktime's [CODEOWNERS](https://github.com/alan-turing-institute/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.


sktime/forecasting/arch.py
```diff
@@ -0,0 +1,307 @@
...
+
+from sktime.forecasting.base import BaseForecaster
+from sktime.utils.validation._dependencies import _check_soft_dependencies
+
+_check_soft_dependencies("arch", severity="warning")
+
+class ARCH(BaseForecaster):
+    """Generalized Autoregressive Conditional Heteroskedasticity (GARCH) model for forecasting 
+    votility in high frequency time series data.
+
```
fkiraly, line 18: should say explicitly which package and model therein it interfaces

sktime/forecasting/arch.py
```diff
@@ -0,0 +1,307 @@
...
+    Implements Generalized Autoregressive Conditional Heteroskedasticity (GARCH) models introduced 
+    and implemented in [1]_, [2]_, and [3]_ to forecast the volatility in high frequency time series data.
+    
+    ARCH models are a popular class of volatility models that use observed values of returns or residuals 
+    as volatility shocks. A basic GARCH model is specified as
+        rt = μ+ϵt
```
fkiraly, line 24: if you use math, kindly use the math environment in rst.
Also kindly ensure that all symbols are properly defined, if you use math, and clearly linked to the inputs y, X

sktime/forecasting/arch.py
```diff
@@ -0,0 +1,307 @@
...
+        σ^2t = ω+αϵ^2t−1+βσ^2t−1
+    
+    A complete ARCH model is divided into three components:
+        a mean model, e.g., a constant mean or an ARX;
+        a volatility process, e.g., a GARCH or an EGARCH process; and
+        a distribution for the standardized residuals.
+
+    Hyper-parameters
+    ----------------
+    y : {np.ndarray, Series, None}
```
fkiraly, line 35: y and x are not hyperparameters

Also, kindly adhere to numpydoc, this should be called "Parameters"

sktime/forecasting/arch.py
```diff
@@ -0,0 +1,307 @@
...
+    ...    hold_back=None, 
+    ...    rescale=None)
+    >>> forecaster.fit(y)
+    ARCH(...)
+    >>> y_pred = forecaster.predict(fh=[1,2,3])
+    """
+
+    _tags = {
+        "scitype:y": "univariate", 
+        "y_inner_mtype": ("np.ndarray", "pd.Series"),
```
fkiraly, line 111: this might have to be a list

sktime/forecasting/arch.py
```diff
@@ -0,0 +1,307 @@
...
+        self.q = q
+        self.power = power
+        self.dist = dist
+        self.hold_back = hold_back
+        self.rescale = rescale
+    
+        super(ARCH, self).__init__()
+
+
+    def _fit(self, y, X=None, **fit_params):
```
fkiraly, line 146: please use the correct signature, `def _fit(self, y, X=None, fh=None):`

Kindly refer to the extension templates

sktime/forecasting/arch.py
```diff
@@ -0,0 +1,307 @@
...
+        """Fit the training data to the estimator
+
+        Parameters
+        ----------
+        y : guaranteed to be of a type in self.get_tag("y_inner_mtype")
```
fkiraly, line 151: replace this by the assumption in this special case. Was that numpy and DataFrame?

sktime/forecasting/arch.py
```diff
@@ -0,0 +1,307 @@
...
+            since model-specific non-linear transformations are applied to
+            value before computing the variance recursions.
+
+        Returns
+        -------
+        self : returns an instance of self
+        """
+        from arch import arch_model as _ARCH
+
+        self.y = y
```
fkiraly, line 198: you don´t need to write y, X to self. You don´t use it later, do you?

fkiraly (changes requested):
Thanks, great work!

Some change requests:
* the linter is failing, ensure you fix the formatting errors. Look for "precommits" in the developer guide, and/or click on "details" below.
* you should fully comply with the extension template, there are a number of small deviations (but they are blocking since we need interface compatibility)
* the ARCH model is only "interesting" if you also return sigma-t, i.e., the heteroskedastic variance forecasts. It might be easiest to return them using the `predict_var`(iance) interface. This is currently not in the extension template, but is in the base class. For convenience, signature is here:

```python
    def _predict_var(
        self,
        fh=None,
        X=None,
    ):
        """Compute/return variance forecasts.

        private _predict_proba containing the core logic, called from predict_var

        Parameters
        ----------
        fh : guaranteed to be ForecastingHorizon
            The forecasting horizon with the steps ahead to to predict.
        X : optional (default=None)
            guaranteed to be of a type in self.get_tag("X_inner_mtype")
            Exogeneous time series to predict from.

        Returns
        -------
        pred_var : pd.DataFrame
            Column names are exactly those of `y` passed in `fit`/`update`.
                For nameless formats, column index will be a RangeIndex.
            Row index is fh. Entries are variance forecasts, for var in col index.
        """
```

sktime/forecasting/arch.py
Vasudeva-bit: Oh okay, understood.

sktime/forecasting/arch.py
Vasudeva-bit: Yeah, I shouldn't have overlooked this part.

sktime/forecasting/arch.py
Vasudeva-bit: I knew that this should be a list but, I thought the `y_inner_mtype` could be a tuple as it is immutable, so I preferred tuple. Anyway, I can change this.

sktime/forecasting/arch.py
Vasudeva-bit: Actually no, will change this.

sktime/forecasting/arch.py
Vasudeva-bit: Hmm yeah, will do this as well.

sktime/forecasting/arch.py
Vasudeva-bit: will correct the signature.

Vasudeva-bit:
I shall work on the changes requested, the `predict_var` method. I think I need to do a close review before a PR. Thank you very much @fkiraly.

sktime/forecasting/arch.py
fkiraly: well, it might work as a tuple, but it´s not the official design so might run into trouble somewhere.

Btw I think you are right, design-wise, tuples would have probably been better.
But interface compliance wise, it needs to be list...

sktime/forecasting/arch.py
fkiraly: btw, are you certain that the `arch` package  can deal with both kinds of input, i.e., numpy and pandas?

fkiraly:
one more thing, that´s non-blocking: I think `arch` does `update` via its `fit`, so `update` of the sktime needs estimator needs to call the `arch` interface's `fit` with the update setting?

sktime/forecasting/arch.py
Vasudeva-bit: Got it.

Vasudeva-bit:
Yeah, the `arch` interface provides the `update` functionality in the `fit` method, so we can use that by specifying the `update_freq` parameter (like you said).

sktime/forecasting/arch.py
Vasudeva-bit: As specified in the [`arch` documentation](https://arch.readthedocs.io/en/latest/univariate/introduction.html), it can deal with both kinds of input. Anyway, I will verify this before the next PR.

fkiraly: hm, ok!
Then the question would be, does it work natively, or are they doing an internal conversion?
If there's an `arch` internal conversion to one of the two, I think it would be better to keep all conversions in `sktime` and convert only to the "real" internal type. On the other hand, if there's implementations for both formats without conversions within `arch`, then that's the right setting.

Vasudeva-bit: I am not sure with the exact implementation in the `arch` package, I will verify that now.

fkiraly: (if it´s difficult to find out, I don´t mind if you leave it as is, as long as it works)

Vasudeva-bit:
Please guide me with this: 
Shall I add an `update` method in the `arch` model to implement update functionality by invoking the `fit` method with the mandatory `update_freq` parameter or is it okay to leave it as an option in the `fit` method only?

fkiraly:
> Please guide me with this: 
> Shall I add an `update` method in the `arch` model to implement update functionality by invoking the `fit` method with the mandatory `update_freq` parameter or is it okay to leave it as an option in the `fit` method only?

No, kindly leave the `arch` model unchanged.

The idea is as follows:
* when the `fit` of the `sktime` estimator is called, it should call `arch` `fit`, without update
* when the `update` of the `sktime` estimator is called, it should call update functionality of `arch`, if possible. I have not looked at it, but that would probably be a suitable call to `fit` (of `arch`) with update related parameters?

In particular, no non-conformant args should be added to `sktime` `fit`, any update related functionality should be called from the `sktime.update` interface point.

It may be confusing that `arch` does update in its `fit`, but nevertheless it should be invoked from `sktime` `update` and not `sktime` `fit`.

Does this make sense?

fkiraly:
PS: kindly address the linting failures - you can check what needs to be fixed by going on details here:
image

Vasudeva-bit:
Yeah, I will not change the arch model but add an `update` (which is not currently existing) to invoke the `fit` method of the `arch` model internally with the `update_freq` parameter. 
Btw, I could fix the linting issues, anyway will again crosscheck that after adding methods like `_predict_var` and `update`.

sktime/forecasting/arch.py
Vasudeva-bit: Actually no, I will change that.

Vasudeva-bit:
I will clear the linting issues following the details.

sktime/forecasting/arch.py
fkiraly: I would say "Directly interfaces ARCH models from python package `arch`"

sktime/forecasting/arch.py
```diff
@@ -0,0 +1,319 @@
...
+        self.q = q
+        self.power = power
+        self.dist = dist
+        self.hold_back = hold_back
+        self.rescale = rescale
+        self._forecaster = None
+
+        super(ARCH, self).__init__()
+
+    def _fit(self, y, X=None, fh=None, **fit_params):
```
fkiraly, line 144: the methods should not have parameters `**fit_params` etc, only the specified args

Vasudeva-bit: Oh okay, but the `fit` method of the `arch` model takes some other parameters as well. What can I do about that?

Vasudeva-bit:
I have cleared all the linting issues as shown in the details but still, the code couldn't pass the code quality test. I hope you can help me with the approach!

fkiraly:
>I have cleared all the linting issues as shown in the details but still, the code couldn't pass the code quality test. I hope you can help me with the approach!

I think you haven't read all the sections, there is content further down in the log, for instance here:

image

There is also more further down from this (not in the screenshot), a green "plus" without anything after it means you need to add a newline.

Vasudeva-bit:
Yes, there is something more I need to do, it's completely my mistake.

sktime/forecasting/arch.py
Vasudeva-bit: Thanks for letting me know about linting issues, may I know how I can handle multiple parameters of the `fit` method in the `arch` model. (currently **fit_params is used)

sktime/forecasting/arch.py
Vasudeva-bit: Okay, your option is more clear so will update it as "Directly interfaces ARCH models from python package `arch`"

Vasudeva-bit:
Thanks for letting me know about linting issues, may I know how I can handle multiple parameters of the `fit` method in the `arch` model. (currently `**fit_params` is used)

fkiraly:
> may I know how I can handle multiple parameters of the fit method in the arch model

If you have more parameters than just the data and the forecasting horizon, you need to do as follows:

* pass the parameter to the constructor
* write it to `self` in the constructor
* in `fit`, read it from `self` and pass it to the `fit` method.

This allows to expose all parameter choices via the scikit-learn compatible interface.

For an example of this, you can look at `smoothing_level` in `ExponentialSmoothing`.

sktime/forecasting/arch.py
```diff
@@ -0,0 +1,405 @@
...
+
+    Parameters
+    ----------
+    y : {np.ndarray, Series, None}
+        The dependent variable
+    x : {np.ndarray, DataFrame}, optional
+        Exogenous regressors.  Ignored if model does not permit exogenous
+        regressors.
+
+    Hyper-parameters
```
aiwalter, line 43: there should only be a sectiion ´Parameters´

sktime/forecasting/arch.py
```diff
@@ -0,0 +1,405 @@
...
+    rescale : bool
+        Flag indicating whether to automatically rescale data if the scale
+        of the data is likely to produce convergence issues when estimating
+        model parameters. If False, the model is estimated on the data without
+        transformation.  If True, than y is rescaled and the new scale is
+        reported in the estimation results.
+
+    See Also
+    --------
+    Autoregressive Integrated Moving Average (ARIMA) models
```
aiwalter, line 81: this should only link class names from `sttime`,

aiwalter, line 81:
```suggestion
    ARIMA
    AutoARIMA
    StatsForecastAutoARIMA
```

sktime/forecasting/arch.py
```diff
@@ -0,0 +1,405 @@
...
+    ...    lags=0,
+    ...    vol="Garch",
+    ...    p=1,
+    ...    o=0,
+    ...    q=1,
+    ...    power=2.0,
+    ...    dist="Normal",
+    ...    hold_back=None,
+    ...    rescale=None,
+    ...):
```
aiwalter, line 111: please make this shorter, maybe even default all params is possible

sktime/forecasting/arch.py
```diff
@@ -0,0 +1,405 @@
...
+        "cov_type",
+        "show_warning",
+        "first_obs",
+        "last_obs",
+        "tol",
+        "options",
+        "backcast",
+    )
+
+    def __init__(
```
aiwalter, line 138: is there a param equivalent to `sp` from other models, which descibed the seasonal periodicity

sktime/forecasting/arch.py
```diff
@@ -0,0 +1,405 @@
...
+            Last observation to use when estimating model
+        tol : float, optional
+            Tolerance for termination.
+        options : dict, optional
+            Options to pass to 'scipy.optimize.minimize'.  Valid entries
+            include 'ftol', 'eps', 'disp', and 'maxiter'.
+        backcast : {float, np.ndarray}, optional
+            Value to use as backcast. Should be measure '\sigma^2_0'
+            since model-specific non-linear transformations are applied to
+            value before computing the variance recursions.
```
aiwalter, line 243: why this here?

sktime/forecasting/arch.py
```diff
@@ -0,0 +1,405 @@
...
+            optional. If not provided, a warning is raised about the future change in
+            the default which will occur after September 2021.
+
+        Returns
+        -------
+        y_pred : pd.Series
+            Point predictions
+        """
+        if fh is not None:
+            self.fh = fh
```
aiwalter, line 331: I think this is not needed and done in base class, but you might just check that. There is `check_fh()` called I think

fkiraly:
> Btw I used **fit_params in the _fit method after looking at the code of adapters such as _pmdarima.py or _fbprophet.py.

Indeed! That looks like an oversight, from legacy (old) code when we hadn't settled yet on the interface patterns.
Have fixed this here: https://github.com/alan-turing-institute/sktime/pull/2343

sktime/forecasting/arch.py
fkiraly: yes, that's the reason why readthedocs is failing. These need to be class names, and the reference section is auto-generated.

sktime/forecasting/arch.py
```diff
@@ -0,0 +1,405 @@
...
+        options=None,
+        backcast=None,
+        params=None,
+        start=None,
+        align=None,
+        method="analytic",
+        simulations=None,
+        rng=None,
+        random_state=None,
+        reindex=None,
```
aiwalter, line 167: what does `reindex` do? I think we might drop this param if its operating on index

aiwalter (changes requested):
left few comments

sktime/forecasting/arch.py
Vasudeva-bit: Oh okay, I will change accordingly.

sktime/forecasting/arch.py
Vasudeva-bit: Hmm okay, I will put everything under `parameters`.

sktime/forecasting/arch.py
Vasudeva-bit: I don't think there is any parameter in the `arch` interface that's equivalent to seasonal periodicity (`sp`).

sktime/forecasting/arch.py
Vasudeva-bit: okay, will replace it with the `check_fit` method.

Vasudeva-bit:
Thank you @aiwalter. I will do the requested changes.

sktime/forecasting/arch.py
Vasudeva-bit: Okay, I shall make all the parameters default.

sktime/forecasting/arch.py
```diff
@@ -0,0 +1,405 @@
...
+            Custom random number generator to use in simulation-based forecasts. Must
+            produce random samples using the syntax rng(size) where size the 2-element
+            tuple (simulations, horizon).
+        random_state : RandomState, optional
+            NumPy RandomState instance to use when method is 'bootstrap'
+        reindex : bool, optional
+            Whether to reindex the forecasts to have the same dimension as the series
+            being forecast. Prior to 4.18 this was the default. As of 4.19 this is now
+            optional. If not provided, a warning is raised about the future change in
+            the default which will occur after September 2021.
```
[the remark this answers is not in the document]
Vasudeva-bit: Do I need to replace these details of parameters (parameters of `_predict()` method) and put them in the constructor or do I need to remove them without replacement?

sktime/forecasting/arch.py
fkiraly: no, kindly do not do anything of that kind at all. Input checks are carried out by the base class.
All the inner methods should implement is model logic.

Vasudeva-bit: Oh, I got what you meant.

fkiraly: what happens if you pass the horizon both in `fit` and `predict`, to `arch`?

Vasudeva-bit: Very sorry, excuse me. Only predict method takes horizon param. The issue is resolved.

sktime/forecasting/arch.py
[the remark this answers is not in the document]
aiwalter: model related params schould only be exposed in constructor. Where you use them internally is not relevant. I think that docstring was lost here?

fkiraly:
> How can I address the failure of test Install and test / test-windows (3.8) (pull_request) ?

Generally, you should be able to run the tests locally, either via `pytest` or `check_estimators`, and then see which tests are failing.
In your case, `_predict` always breaks - you should *not* be trying to write to `fh`. If you read the docstring, all you should do is write model parameters. Do not do any input checks or try to replicate logic around `fh`.

sktime/forecasting/arch.py
```diff
@@ -0,0 +1,405 @@
...
+            since model-specific non-linear transformations are applied to
+            value before computing the variance recursions.
+
+        Returns
+        -------
+        self : returns an instance of self
+        """
+        from arch import arch_model as _ARCH
+
+        self.fh = fh
```
aiwalter, line 251:
```suggestion
```

Vasudeva-bit:
Yes, but I have addressed issues such as side effects to `fh`, and updated the `See Also` section with suggested changes. I will go through the logs once again.

Vasudeva-bit:
I didn't push because I thought of addressing the rest of the failed tests like `Install and test / test-windows (3.7) (pull_request)`.

aiwalter:
seems the example code in docstring is failing, can you try to run it to see what is wrong there?

Vasudeva-bit:
Okay, I will do that.

Vasudeva-bit:
I tried testing the example code and there is a weird case when the `params` parameter of the `arch` `forecast` method is `None`, an error is raised. I couldn't understand the exact cause because when I tried testing the `forecast` function directly with the `None` value to the `params` parameter, the code worked perfectly well. I am still trying to resolve it.

fkiraly:
I think it´s not just the example code - the estimator `ARCH` is failing a number of tests, check out the logs for what is going wrong.

Vasudeva-bit:
Yes, that's right, I understood it from the logs, I will try to fix those.

Vasudeva-bit:
I had some exams last week so could not spare enough time working on the arch estimator. I am still working on fixing the issues and I hope I can spend a good amount of time on this task from now. Will reach out to you if I am stuck!

Vasudeva-bit:
Hello @fkiraly 
I made the editable install successfully! The error `horizon must be integer >= 1` is because the `fh` is internally converted to class `ForecastingHorizon`. Since `horizon` in the `arch` package is not compatible with class type `ForecastingHorizon` it raised a general error. Using `fh[0]` to get the value for `fh` solves the issue, hope this is a good move. Kindly let me know the alternatives if any.

fkiraly:
> Kindly let me know the alternatives if any.

`fh[0]` will only give you the first horizon value, I don´t think that´s a good idea. You want to make predictions at all horizon values.

Vasudeva-bit:
Okay, I got it. btw the `arch` `forecast` method is compatible with only int type value for `fh` forecasting horizon. In order to forecast at all values, I think the `forecast` method needs to be invoked multiple times.

fkiraly:
>btw the arch forecast method is compatible with only int type value for fh forecasting horizon. 

You can convert back/forth between date if you need to, one of the methods of `ForecastingHorizon` should do this (also see the other estimators)

> In order to forecast at all values, I think the forecast method needs to be invoked multiple times.

Why not just a single time up to the maximum horizon index?

Vasudeva-bit:
Yes, forecasting up to the maximum index works. I am also going through methods of class `ForecastingHorizon`.

fkiraly:
@Vasudeva-bit, are you still working on this?

Vasudeva-bit:
I am working on fixing errors, and hope to resolve them soon

sktime/forecasting/arch.py
```diff
@@ -0,0 +1,392 @@
...
+            single column/variable
+        X : optional (default=None)
+            guaranteed to be of a type in np.ndarray or pd.DataFrame
+            Exogeneous time series to fit to.
+
+        Returns
+        -------
+        self : returns an instance of self
+        """
+        from arch import arch_model as _ARCH
```
lbventura, line 276: Should this import occur inside the _fit function?

fkiraly: no, because it is a soft dependency.

All soft dependency imports should be isolated in a way like this, so importing the module does not error out under the minimal dependency set (which excludes soft dependencies).

All estimators should also be aware of the soft dependencies in them, via the `python_dependencies` tag.

Issue #2157, opened 2022-03-02
[ENH] Hierarchical reconcilers for point forecasts 


ciaran-g:
Some initial thoughts, first two will be a good place to start

 - [x] Bottom up (aggregate levels ignored)
 - [x] Top down (historical proportions)
 - [x] Top down (forecast proportions)
 - [x] Unweighted OLS combination
 - [x] Weighted LS combination
 - [x] Trace minimisation (MinT) (various options for covariance matrix , including things like regularisation)
 - [x] Middle out (?) 

Think it might be best for now to focus on spatial hierarchies? i.e. get these up and running and worry about grouped or temporal hierarchies afterwards?

More info [here](https://github.com/robjhyndman/reconciliation_review_talk)

---

@fkiraly update: there are now two estimators, `Reconciler` (transformer) and `ReconcilerForecaster` (forecaster) where the two remaining reconciling algorithms could/should be added.
Since this prescribes a good templated structure, I think it's a good first issue (for individuals familiar with, or interested in, the reconciliation algorithms)

fkiraly:
Thanks for the list!

What would be extremely useful is a quick analysis which of these reconciliation strategies can be mapped on transformers (e.g., apply after forecasting, use only forecasts to fit and reconcile), and which need to be wrappers/compositors (e.g., needs the original or historical data, needs to call methods of the forecaster as part of the algorithm).

ciaran-g:
Good shout Franz - started working on the bottom-up and OLS if you want to have a [look](https://github.com/ciaran-g/sktime/blob/reconciler-bottom-up/examples/hierarchical_play.ipynb)

Transformers:
 - bottom up
 - OLS
 - Weighted LS (structural)

Wrappers/compositors:
- Weighted LS (variance) - uses residuals
- MinT (sample) - uses residuals
- MinT (shrinkage) - uses residuals
- Top down - historic proportions

fkiraly:
linking this here, some open work items related to the above:
https://github.com/alan-turing-institute/sktime/issues/2416

fkiraly:
@ciaran-g, just to check - these are all implemented now, right?

ciaran-g:
@fkiraly the `middle_out` and `top_down_historic proportions` methods have not been implemented. I'm not currently working on these at the minute though if you have a plan in mind to get them implemented

fkiraly:
ok, we'll leave this open then - I think it might be a good first issue since it could be added into an existing estimator.

[...]

PR #2287, opened 2022-03-23
[ENH] Implement reconcilers for hierarchical predictions - transformers

ciaran-g:
#### Reference Issues/PRs

This PR adds a reconciler transformer, which takes hierarchical predictions and processes them so they sum appropriately. Three methods are implemented here out of the list in issue #2157
 - bottom up 
 - ols
 - wls_var

Hierarchical reconciler which depends only on the structure of the hierarchy. The three methods are detailed [here](https://otexts.com/fpp3/hierarchical.html) and the code is constructed similar to the methodology as noted in these [slides](https://github.com/robjhyndman/reconciliation_review_talk/blob/main/10years_reconciliation.pdf). The only difference between the three methods is the construction of the "G" matrix.

No


ciaran-g:
just pushed a few checks on the index and method @fkiraly

fkiraly (changes requested):
Thanks for this, good start!

My main comment is that this needs to work for an arbitrary number of hierarchy levels. Currently I think it works only for precisely three.
It should work for two, four, and even one (in that case the series is returned unchanged).

For that, you also should set the tag
`"X_inner_mtype": ["pd.DataFrame", "pd-multiindex", "pd_multiindex_hier"]`
that way you can assume you are being passed a `pd.DataFrame` with one, two, three, or more levels.

fkiraly:
PS: I took the liberty to update your branch from `main` and fix your tags (except `X_inner_mtype`)

ciaran-g:
Closing for now - it should now work with hierarchy levels greater than or equal to two tho 👍

fkiraly:
? why are you closing?
This has not been merged yet afaik?

fkiraly:
We should try to get this merged, or am I misunderstanding sth?

ciaran-g:
lol no, that's my bad! I was just closing until the we sorted the problem with the tests

fkiraly:
currently it's linting - and we should leave it open.
We can merge the type check updates into it and then the tests should run.

PR #2288, opened 2022-03-23
Classification test parameter refactor

MatthewMiddlehurst:
Part of the Estimator Test Parameters Refactor #1678 leading into #2255.

Adds `get_test_params` methods for all relevant classifiers which did not have one, and removes the parameters from `_config.py`. While going through these, changed some to use faster parameters where I thought it was applicable (mostly "n_estimators" 3 > 2). If this ends up causing issues, I will revert it.


fkiraly (approved):
Looks good, thanks for this tedious work!

PR #2292, opened 2022-03-24
[ENH] Aggregator as a transformer

ciaran-g:
#### Reference Issues/PRs

This PR adds a aggregator transformer, which takes observations at the bottom level of the hierarchy and sums them according to the index provided. This is linked to PR #2287 and issue #2157

A transformer that takes hierarchical observations at the bottom level of the hierarchy only, and sums them appropriately using the index. An identifier is placed in the index for aggregated nodes called `__total`. The reconciler implemented in #2287 use this identifier, so probably this is a better one to focus on first.

Also I noticed for some reason, if you pass an a non-multiindex type input to the transformer it doesn't raise an error like a guess it should given the tags filled in at the top of the script? I tried to include a test for this as well, but it only seems to work correctly when you call the hidden _transform function 

No


sktime/transformations/hierarchical/aggregate.py
```diff
@@ -0,0 +1,181 @@
...
+        "scitype:instancewise": True,  # is this an instance-wise transform?
+        "X_inner_mtype": "pd_multiindex_hier",
+        "y_inner_mtype": "None",  # which mtypes do _fit/_predict support for y?
+        "capability:inverse_transform": False,  # does transformer have inverse
+        "skip-inverse-transform": True,  # is inverse-transform skipped when called?
+        "univariate-only": False,  # can the transformer handle multivariate X?
+        "handles-missing-data": False,  # can estimator handle missing data?
+        "X-y-must-have-same-index": False,  # can estimator handle different X/y index?
+        "enforce_index_type": True,  # index type that needs to be enforced in X/y
+        "fit-in-transform": True,  # is fit empty and can be skipped? Yes = True
```
ciaran-g, line 43: should be `"fit_is_empty"`

fkiraly:
It was `enforce_index_type`, that should be an index type (not boolean).

fkiraly:
pushed a fix

fkiraly:
Hm, @ciaran-g, my fix repairs the tag issue.

Now, tests seem to be failing on `unique` applied to the multiindex level which has `__total` - I do not yet see why this should be failing, since manual treatment does not crash.

Would you mind investigating what happens in `convert_output` so that the tests are failing? I'll also look at it later.

ciaran-g:
Thanks for that Franz, just updated the branch as well to fix an issue with the `flatten_single_levels` argument

I think the test errors are to do with the aggregated index character I've added, i.e. `__total`. By the looks of it some of the tests sort the index, but a mix of integers and strings on the index causes this error to be raised

```
TypeError: '<' not supported between instances of 'str' and 'int'
```

fkiraly:
> By the looks of it some of the tests sort the index, but a mix of integers and strings on the index causes this error to be raised

I looked into this, and am getting confused - the `sort` seems to be called inside `unique`, so I tried to reproduce it with a manually created index set that has numbers and strings. That works fine.

My suspicion is, whatever you do, it leaves the `dtype` as `int`, while it should turn it to `object`. Since it is `int`, something internal tries to do `sort`. I have not had time to check that suspicion, but perhaps that gives you one more avenue to investigate.

fkiraly:
Solution works - though, is there a way to avoid casting to `str`? Could we instead cast to the `object `dtype` somehow?

fkiraly (approved):
Tests pass now, with the fix to the `check_pdmultiindex_panel` that allows object dtype.

PR #2295, opened 2022-03-24
Classification expected output test updates

MatthewMiddlehurst:
Adds a new parameter to create_test_instance() and get_test_params() for selecting alternate parameter test sets. A majority of the changes are adding test_params="default" to existing get_test_params() methods, and moving over classifier parameters. See #2255.

Removes all basic_motions and unit_test results comparisons from individual test files. Most of the files have been removed as a result. All the expected results have been moved into a single file (_expected_outputs.py), and if a set of expected results for a classifier are present, a test is run comparing against them using a "results_comparison" set of parameters from get_test_params(). Fixed these tests so they work as intended.


fkiraly:
so, if you're already doing this, would you perhaps mind moving it all into the `TestAllClassifiers` class? Using, possibly, external configs for the chosen parameters and the expected results?

MatthewMiddlehurst:
I don't mind moving it all (this is prep for that, really), but I think it would be good to discuss how best to deliver the configs. 

For now, I could just create another file and draw them from that, or in the issue suggested adding an optional parameter to the get_params method for alternate test sets.

fkiraly (changes requested):
Huge pile of work, thanks!

May I kindly ask for the following:
* I think we ought to use the same argument name in `get_test_params` and in `create_test_instance`.
* I think if a string is passed that is not `"default"`, no error should be raised. I.e., all strings default to the `"default"` parameters instead of carrying out a "is this defined" check. That way, if not available, "special" parameter settings will use the default. (otherwise, the test suite will fail if we query special parameters; and it would not be queriable whether special parameters are available)
* consequently, no change should be made to those estimators where both the "special" and the "default" set can be equal, or where no "special" set is defined.

sktime/transformations/series/summarize.py
```diff
@@ -289,9 +289,15 @@ def _transform(self, X, y=None):
...
 
     @classmethod
-    def get_test_params(cls):
+    def get_test_params(cls, parameter_set="default"):
         """Return testing parameter settings for the estimator.
 
+        Parameters
+        ----------
+        parameter_set : str, default="default"
+            Name of the set of test parameters to return. The method must be overridden
```
fkiraly, line 298: the docstring is not clear. What does this mean "must be overridden"?

Suggest the following string instead
```Name of the set of test parameters to return, for use in tests. If no special parameters are defined for a string, will always return the `"default"` set.```

MatthewMiddlehurst:
> I think we ought to use the same argument name in get_test_params and in create_test_instance.

Done

> I think if a string is passed that is not "default", no error should be raised. I.e., all strings default to the "default" parameters instead of carrying out a "is this defined" check. That way, if not available, "special" parameter settings will use the default. (otherwise, the test suite will fail if we query special parameters; and it would not be queriable whether special parameters are available)

I have removed the "default" check in the base class. If I am understanding this correct, no method should return any errors, and just return the default parameters if the input is not available?

I think this has the potential to cause some confusion, and a simple misspell could give you the wrong set of parameters without you knowing unless you compare them against the default. I think cleanly displaying that the requested set doesn't exist is better, but I don't feel strongly enough to argue for it if you think this is necessary.

> consequently, no change should be made to those estimators where both the "special" and the "default" set can be equal, or where no "special" set is defined.

Reverted the ones which had the same output for both default and expected. Only 4 classifiers, other estimators only have the header/docs changes.

sktime/transformations/series/summarize.py
MatthewMiddlehurst: Done, but may change again. See part 2 of below comment.

fkiraly:
> I have removed the "default" check in the base class. If I am understanding this correct, no method should return any errors, and just return the default parameters if the input is not available?

Yes - anything else would require us to change *every single instance* of `get_test_params` and add 100s of copies of boilerplate key checking logic. That's not easily extensible and would cause a substantial change to the extender contract.
Due to this, I think this is really out of the question without major deprecation happening.

Whereas, if the extender contract does not change, we have no such problem.

fkiraly:
PS: I think you should also describe in the extension template how and when to add this second, optional set of parameters.

MatthewMiddlehurst:
Ok, I have removed all error raising and updated the templates.

fkiraly (changes requested):
Thanks!

Change requests:
* `BaseObject` has the new parameter, but not the docstring
* to an implementer, I think it is still not clear what they should do. How should the two sets `"default"` and `"results_comparison"` be chosen? Do they need to implement both? This is unclear to me, and I think the implementer will also be confused.
* I'd suggest to add an explanation in the extension templates that the typical implementer does not need to provide `parameter_set`.

MatthewMiddlehurst:
Added the BaseObject parameter docs and a bit more to the extension templates.

fkiraly:
@MatthewMiddlehurst, may I make some changes to the docs and extension templates?
I think the instructions could be clearer and more actionable.

Just to speed this up, you can revert my commit if you don't like it.

MatthewMiddlehurst:
Sure go ahead.

fkiraly:
Done, kindly check whether happy.

I also added the argument to `create_test_instances_and_names` where it makes sense for consistency.

fkiraly (changes requested):
Sorry, I thought we were done, but I noticed one more thing while resolving a merge clash:

We need some kind of deprecation here.
Why: `create_test_instance` is passing the new parameter to `get_test_params`, so it will break with estimators that do not have it.
The error happens in testing/validation utilities like `check_estimator`.

I see three options:
* downwards compatible behaviour in `create_test_instance`, which checks if `get_test_params` has the new argument of not. If not, does not pass it on. This is the end state.
* The above, but with a deprecation message, and the end state is current logic.
* not changing `create_test_instance`

fkiraly:
*actually*, we could just let this be the end state?

MatthewMiddlehurst:
The two create_test methods are the only places which use the optional argument.

Do you mean just go with option 1 from your previous comment and remove the deprecation warning? I don't mind that either.

fkiraly:
> Do you mean just go with option 1 from your previous comment and remove the deprecation warning? I don't mind that either.

Both fine with me - let´s go with your preference.

fkiraly (approved):
Ok, happy with that reasoning & end state.

Issue #2291, opened 2022-03-24
[BUG] `update` of forecasting pipelines does not transform model


atsangarides:
Update by @fkiraly: the root cause of the below is, probably, forecasting pipelines not performing transform in the `update` method. Changed the title to reflect this.

---

**Describe the bug**

`update_predict_single` no longer refits the model in v0.10.1

**To Reproduce**

```python
forecast_horizon=24
cov=0.95
preds, conf_ints = model.update_predict_single(
    y=y_new, X=None, 
    fh=np.arange(1, forecast_horizon+1),
    update_params=True,
    return_pred_int=True, alpha=1-cov
)
```

**Expected behavior**

The model should be refitted when setting `update_params` to True. It seems like the property class `_y` attribute gets updated with the new observations, but the model is not refitted, and predictions returned are just longer forecasts of the originally trained model.

**Additional context**

**Versions**
<details>

System:
    python: 3.8.11 (default, Jan 25 2022, 23:19:55)  [Clang 13.0.0 (clang-1300.0.29.3)]
executable: /Users/tsangis/Projects/uk-ds-day-ahead-prices-forecast/venv/bin/python
   machine: macOS-12.0.1-x86_64-i386-64bit

Python dependencies:
          pip: 22.0.3
   setuptools: 58.0.4
      sklearn: 1.0.2
       sktime: 0.10.1
  statsmodels: 0.13.1
        numpy: 1.21.1
        scipy: 1.7.3
       pandas: 1.4.1
   matplotlib: 3.5.1
       joblib: 1.1.0
        numba: 0.55.1
     pmdarima: 1.8.4
      tsfresh: None

</details>

fkiraly:
>`update_predict_single` no longer refits the model in v0.10.1

Thanks for reporting, @atsangarides!

I briefly looked into the base class, and it does seem that `update_predict_single` calls `update` by default, which calls the inner `_update`. The latter has as default "refitting on the entire data seen so far".

So from a framework perspective, it seems that refitting/updating is hardcoded and also done as a default (if all works as intended).

What could be the case, is that individual models override this incorrectly and don't update despite the interface looking like it would.

So, could you kindly paste some reproducible code, and let us know whether you think it's a particular model, or whether it's a general problem? If you think it's a general problem, can you explain why you think it is?

fkiraly:
[...]

PR #2296, opened 2022-03-24
[BUG] fixing forecasting pipelines `update`

fkiraly:
This PR (hopefully) fixes the bug reported in #2291, and a parallel bug in `ForecastingPipeline`.

This is based on the assumption that the bug in #2291 is caused by incorrect implementation of `update` in `TransformedTargetForecaster`, where `transform` of the transformers in the pipeline were not being called, and all `update` were called on the input data.

Depends on #2300 which fixes an issue arising from current tests causing empty `y` being passed to `update` when `update_predict` is called within those tests, which is encountered by the fix.


fkiraly:
Not sure why it doesn't show here properly, but the tests are passing:
https://github.com/alan-turing-institute/sktime/actions/runs/2039968292

lmmentel (approved):

PR #2298, opened 2022-03-24
[ENH] test generation error to raise and not return

fkiraly:
This PR changes the `raise_exception` flag in test fixture generation to True, such that test generation errors are raised (and the test suite not run) if there is a fixture generation error in test collection.

This should take off testing compute load from tests where the collection already crashes, and produce informative errors and tracebacks in those cases.


TonyBagnall (dismissed):

fkiraly:
(merged after tests passed already & fix of independent CI failure)

PR #2300, opened 2022-03-25
[ENH] treating empty y in forecaster update

fkiraly:
This PR allows empty and length 0 `y` to be passed to `update` and `update_predict_single`, in which case no updated is carried out.

This is in order to treat an edge case where a splitter object was passed to `update_predict` which resulted in an empty `y` being passed to `update`, in consequence causing potential issues with test failures in `update_predict_predicted_cutoff` or `test_update_predict_predicted_index`, due to the splitter in the test causing length 0 series to appear via `_predict_moving_cutoff`.

I think the entire update logic is a bit messy and needs to be cleaned up, but this should at least be a symptomatic fix until that happens.


fkiraly:
merged as part of #2296

PR #2302, opened 2022-03-25
removed for loop, changed arguments of convert()

### LINKED ISSUE #1856: [ENH] refactor panel transformers that are incorrectly labelled to series transformers
A number of transformers are in fact transformers that act independently on individual series, and have some inefficienet vectorization (using "nested data frames" inside. This inefficient vectorization shoudl be removed, and the transformers moved to the `transformations.series` module, with deprecation.

One of two things should happen:
* inefficient/loop vectorization should be replaced with efficient vectorization; `X_inner_mtype` should be changed to something more efficient, e.g., `pd-multiindex` (a `Panel` mtype)
* the vectorization should be removed, `X_inner_mtype` should be replaced with `pd.DataFrame` (a `Series` mtype). This will simplify the transformer and confer automated vectorization from the base class.

The following transformers have this issue:
- [ ] `DWTTransformer`
- [ ] `SlopeTransformer`

Until #1854 is merged, work on this should be based on #1854.

markvrma:
#### Reference Issues/PRs

Fixes #1856 

- Made changes to `_tags`, removed `for loop` to remove inefficient vectorization.
No

- [ ] running `pytest` on `test_dwt.py` gives tabularization errors. I'm not sure how to handle this. Out of 11 tests, 6 give the same error
```
================================================================== short test summary info ==================================================================
FAILED test_dwt.py::test_output_of_transformer - ValueError: Tabularization failed, it's possible that not all series were of equal length
FAILED test_dwt.py::test_no_levels_does_no_change - ValueError: Tabularization failed, it's possible that not all series were of equal length
FAILED test_dwt.py::test_output_dimensions[2-12] - ValueError: Tabularization failed, it's possible that not all series were of equal length
FAILED test_dwt.py::test_output_dimensions[3-11] - ValueError: Tabularization failed, it's possible that not all series were of equal length
FAILED test_dwt.py::test_output_dimensions[4-12] - ValueError: Tabularization failed, it's possible that not all series were of equal length
FAILED test_dwt.py::test_dwt_performs_correcly_along_each_dim - ValueError: Tabularization failed, it's possible that not all series were of equal length
```
- Will it be right to use [padder.py](https://github.com/alan-turing-institute/sktime/blob/main/sktime/transformations/panel/padder.py) to make all series of equal length? 
     
- [ ] I haven't moved the file to `series` folder yet, once the error is fixed, it can be moved along with the `test_dwt`.
#### Any other comments?
- I'm not sure if my approach is correct, please let me know if there's a better way to approach this. 
- Fixed a typo in `_convert.py`
- I have commented out code for my reference rather than deleting them, once the error is fixed I'll delete them.
- I haven't moved the file to `series` folder yet, once the error is fixed, it can be moved along with the `test_dwt`.

### DIFF
--- a/sktime/transformations/panel/dwt.py
+++ b/sktime/transformations/panel/dwt.py
@@ -23,12 +23,12 @@ class DWTTransformer(BaseTransformer):
     """
 
     _tags = {
-        "scitype:transform-input": "Series",
+        "scitype:transform-input": "Panel",
         # what is the scitype of X: Series, or Panel
         "scitype:transform-output": "Series",
         # what scitype is returned: Primitives, Series, Panel
-        "scitype:instancewise": False,  # is this an instance-wise transform?
-        "X_inner_mtype": "nested_univ",  # which mtypes do _fit/_predict support for X?
+        "scitype:instancewise": True,  # is this an instance-wise transform?
+        "X_inner_mtype": "pd-multiindex",  # which mtypes do _fit/_predict support for X?
         "y_inner_mtype": "None",  # which mtypes do _fit/_predict support for X?
         "fit-in-transform": True,
     }
@@ -62,27 +62,27 @@ class DWTTransformer(BaseTransformer):
         col_names = X.columns
 
         Xt = pd.DataFrame()
-        for x in col_names:
-            # Convert one of the columns in the dataframe to numpy array
-            arr = convert(
-                pd.DataFrame(X[x]),
+        # for x in col_names:
+        # Convert one of the columns in the dataframe to numpy array
+        arr = convert(
+                pd.DataFrame(X),
                 from_type="nested_univ",
                 to_type="numpyflat",
-                as_scitype="Panel",
+                # as_scitype="Panel",
             )
 
-            transformedData = self._extract_wavelet_coefficients(arr)
+        transformedData = self._extract_wavelet_coefficients(arr)
 
-            # Convert to a numpy array
-            transformedData = np.asarray(transformedData)
+         # Convert to a numpy array
+        transformedData = np.asarray(transformedData)
 
-            # Add it to the dataframe
-            colToAdd = []
-            for i in range(len(transformedData)):
+          # Add it to the dataframe
+        colToAdd = []
+        for i in range(len(transformedData)):
                 inst = transformedData[i]
                 colToAdd.append(pd.Series(inst))
 
-            Xt[x] = colToAdd
+        Xt = colToAdd
 
         return Xt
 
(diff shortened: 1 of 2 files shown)


fkiraly:
@markvrma, thanks for your contribution!

I´m not quite sure what you are trying to do, would you mind explaining?

markvrma:
The intent was to try and apply the 1st method as mentioned in the #1856. 

- modified the `_tags`, `"X_inner_mtype": "pd-multiindex"`
- removed the `for loop` to eliminate inefficient vectorisation (though I'm not sure if I should've done that)
- I could not find a way to convert `from_type = pd-multiindex` in the `convert` function to a `1D array` / `numpyflat` hence I stuck with `nested_univ`

fkiraly (changes requested):
I see, thanks for clarifying!

My recommendation would be:
* use the `numpy3D` internally, i.e., `X_inner_mtype="numpy3D"`, and use `numpy` vectorization and reshape.
* ensure you also return a `numpy3D` array.

That´s because the `_extract_wavelet_coefficients` utility seems to be using `numpy` arrays internally.
Would that work?

markvrma:
> * ensure you also return a `numpy3D` array.

 Will it be a problem since tests in `test_dwt.py` rely on `_transform` returning a nested pandas DataFrame?

fkiraly:
>Will it be a problem since tests in test_dwt.py rely on _transform returning a nested pandas DataFrame?

No, because it tests `transform` and not `_transform`, and the former should convert back to a nested pandas df if one is passed.

markvrma:
- converts from `numpy3D` to `numpyflat` back to `numpy3D`
- `check_if_dataframes_are_equal` throws AssertionError (fails 3 tests)
   - Upon closer inspection( printing `df1` and `df2`  in the  `check_if_dataframes_are_equal` function ) it looks like the column names of `df1` and `df2` are different and for the 3rd test there is only 1 column in `df1` while `df2` has 2 columns

sktime/transformations/panel/dwt.py
```diff
@@ -23,12 +24,12 @@ class DWTTransformer(BaseTransformer):
     """
 
     _tags = {
-        "scitype:transform-input": "Series",
+        "scitype:transform-input": "Panel",
```
fkiraly, line 27: this should be `Series`

fkiraly:
>  3rd test there is only 1 column in df1 while df2 has 2 columns

Yes, you removed the loop over columns. I still think you may need to loop over the variables (2nd index in the numpy 3D container)

PR #2307, opened 2022-03-26
[ENH] escape predict_proba in multiprocessing-idempotent test

fkiraly:
This PR escapes `predict_proba` in the multiprocessing-idempotent test which is currently skipped.

The test currently breaks on `predict_proba` since the comparison utility at the end cannot deal with tfp distributions.

This leads to `check_estimator` reporting the test as broken, since there it is not skipped.


fkiraly:
no longer a problem, comparison is fine now

Issue #2202, opened 2022-03-12
Improved error messaging in transformers (and likely other places)


ngupta23:
https://github.com/alan-turing-institute/sktime/blob/87bdf36dbc0990f29942eb6f7fa56a8e6c5fa7b7/sktime/transformations/base.py#L313

This is the current error message that I get. I suspect that the list concatenation is not happening as intended. It would be nice to improve the error message by handling the list of allowed values appropriately.

```python
ValueError: ["X must be a pandas.Series, found <class 'NoneType'>", "X must be a pandas.DataFrame, found <class 'NoneType'>", "X must be a numpy.ndarray, found <class 'NoneType'>", "X must be a pd.DataFrame, found <class 'NoneType'>", "X must be a numpy.ndarray, found <class 'NoneType'>", "X must be a pd.DataFrame, found <class 'NoneType'>", "X must be list of pd.DataFrame, found <class 'NoneType'>"]
```

May need to be modified here: 

https://github.com/alan-turing-institute/sktime/blob/87bdf36dbc0990f29942eb6f7fa56a8e6c5fa7b7/sktime/datatypes/_check.py#L170

fkiraly:
hm, what are you trying to do, can you post code?

ngupta23:
Honestly, I don't recall now. I think X was None and it gave that message. I think it is just a print issue. It is not concatenating the allowed types correctly.

fkiraly:
ah, I see. No, it was not a print issue, the wrong error message was returned - for some reason, I had it return an internal error message instead of the correct one (like in forecasting). The correct message should be
```
            msg = (
                "X must be in an sktime compatible format, "
                "of scitype Series, Panel or Hierarchical, "
                "for instance a pandas.DataFrame with sktime compatible time indices, "
                "or with MultiIndex and lowest level a sktime compatible time index. "
                "See the forecasting tutorial examples/01_forecasting.ipynb, or"
                " the data format tutorial examples/AA_datatypes_and_datasets.ipynb"
            )
```

Do you think this is clear?

ngupta23:
Thanks @fkiraly . That should be fine. It would be nice to also print what type of X was passed also. It will help with debugging. Thanks!

fkiraly:
>It would be nice to also print what type of X was passed also. It will help with debugging. 

Well, that might be impossible - the user can pass whatever `X` they like - or they can pass a `pd.DataFrame` which does not comply with the specification.

How about extending the error message:
```
[...]

PR #2314, opened 2022-03-27
[ENH] Improved docstrings in forecasters, improved error message in forecasters when receiving an incompatible input

fkiraly:
This PR makes three changes:
* addresses #2202 by producing a clearer error message and instructions for diagnosis if an incompatible input is provided to an sktime forecaster.
* updates docstrings in `BaseForecaster` to the current (target) state of hierarchical and probabilistic forecasting, will be correct after #2355 is merged
* adds a utility `scitype_to_mtype` which allows to list all mtypes for a given scitype - this is used in the error message.
It also replaces other ad-hoc instances of the same functionality with the new public utility.


fkiraly:
due to your earler comments on docstrings, would appreciate your eyes on this, @GuzalBulatova & @Lovkush-A

Lovkush-A:
had a brief skim and changes look good

Issue #2313, opened 2022-03-27
[ENH] correct capability tags in classifier pipelines, forecasting pipelines, etc


fkiraly:
The following case in pipeline building is currently not handled correctly:

Suppose we build a pipeline with transformers that ensure the output is equally spaced and has no missing data (e.g., imputer, and padder).
Then, the resulting pipeline should know that it can deal with missing and unequally spaced data, even if the last components cannot.

Currently, this would result in an error being thrown by the pipeline constructed.

The deeper problem is that there is no way to query the information that the output of a transformer is ensured to be equally spaced, or without missing data. The natural solution would be introducing tags for that, e.g., `ensures_output:equally_spaced` or similar.

fkiraly:
same for forecasting pipelines, #2137

Issue #2243, opened 2022-03-18
[ENH] ensure that pipelines and compositors work for hierarchical forecasters and transformers


fkiraly:
With forecasters and transformers interface upgraded to hierarchical, we should ensure that key compositors and pipelines work for hierarchical data:

- [x] transformer pipelines https://github.com/sktime/sktime/pull/2318
- [x] forecaster pipelines https://github.com/sktime/sktime/pull/2318
- [x] performance metrics https://github.com/sktime/sktime/pull/2601
- [x] evaluation https://github.com/sktime/sktime/pull/3542
- [x] grid search & tuning https://github.com/sktime/sktime/pull/3548

fkiraly:
transformer and forecasting pipelines are in 0.11.2:
https://github.com/alan-turing-institute/sktime/pull/2318

fkiraly:
Done, last item was https://github.com/sktime/sktime/pull/3548

Issue #2088, opened 2022-02-18
[ENH] Ensure or add probabilistic prediction functionality in key compositions (tuning, pipelines ensembles, ...)


fkiraly:

fkiraly:
> Do we want to change the interface to use feature names in all cases?

I thought about this and have no 100% answer. My trouble is that in the univariate case, for common user journeys, there are no column names - e.g., if the input is a `pd.Series`. If the column is then called "0", it might be confusing for users.

Do you have any good suggestions?

kejsitake:
I see how it could be confusing in the univariate case if there is no column names. But i wonder if it is difficult in the multivariate case to know which column belongs to which feature. I wonder if it it bad design to use the names of features only on the multivariate case, even though this would complicate the tests?

kejsitake:
I looked into documentation to come up with a list of multivariate forecasters where predict_interval still needs to be refactored and came up with the following:

- [ ] MultiplexForecaster
- [ ] EnsembleForecaster
- [ ] TransformedTargetForecaster
- [ ] BaseGridSearch
- [ ] ColumnEnsembleForecaster https://github.com/alan-turing-institute/sktime/pull/2081

I am a little bit unsure about the Reducers as they don't have a "scitype-y" tag.

fkiraly:
> I am a little bit unsure about the Reducers as they don't have a "scitype-y" tag.

Hm, that seems like an oversight - although I'm not sure off the top of my head which value would be correct.

fkiraly:
> I looked into documentation and come up with the following list of forecasters where predict_interval still needs to be refactored:

Thanks! I added that list to the issue description.

fkiraly:
>Alternatively, what about adding a keyword feature before the feature name e.g. if the column name is '0', we could use 'feature_0' and so on to make it more clear and consistent in both uni and multivariate cases?

I've been thinking about this, perhaps a `RangeIndex` for columns is the best solution? That would keep '0', but not as a string.

I lean this direction now since I've been running into case distinctions in the `predict_etc` defaults, and I also noticed, what do we do if there is a variable called `"Coverage"` or `"Quantiles"` etc? That would spell the same trouble as the `"ds"` reserved variable name in prophet...

fkiraly:
> I looked into documentation and come up with the following list of forecasters where predict_interval still needs to be refactored:

Is `ForecastingPipeline` missing?

kejsitake:
[...]

PR #2318, opened 2022-03-27
[ENH] adding compatibility with probabilistic and hierarchical forecasts to `ForecastingPipeline` and `TransformedTargetForecaster` 

fkiraly:
This PR adds compatibility of `ForecastingPipeline` and `TransformedTargetForecaster` with:
* the probabilistic forecasting interface points `predict_interval` etc
* hierarchical and panel inputs for hierarchical, panel, global forecasting

This PR alsp:
* fixes `_has_capability` for the case where tags can be dynamic (such as here)
* adds test parameters to `ForecastingPipeline` and `TransformedTargetForecaster` that cover the probabilistic case
* fixes one missing argument of `_predict_proba` in the extension template.
* adds a `_transform` to the `ForecastingPipeline`.
* fixes a broken error message in the `test_pred_int_tag` test
* fixes `BaseForecaster` which was missing the `marginal` arg in `predict_proba`.

Addresses parts of #2088, #2243, #2313.

Contains the fixes in https://github.com/alan-turing-institute/sktime/pull/2322 which are necessary for tests not to break, since `_has_capability` has changed and that uncovers the issues with `inverse_transform` etc in #2322.


ngupta23:
@franz - Also addresses https://github.com/alan-turing-institute/sktime/issues/2203?

fkiraly:
> @franz - Also addresses https://github.com/alan-turing-institute/sktime/issues/2203?

It's only the underscore transform, you can't use this as a transformer yet - is that what you want?

ngupta23:
This fix works now. Thanks @fkiraly for the quick fix.

image

fkiraly:
note to self: there may be still problems with multivariate transformers here - the transformer is directly applied to the return object, which may have different number of columns, and the transformer may complain.

Instead, `_get_inverse_transform` should be looped over, and applied to the collection of columns with the same quantile/coverage value.

extension_templates/forecasting.py
```diff
@@ -384,7 +384,7 @@ def _predict_var(self, fh, X=None, cov=False):
     #   implementation which uses variance or quantiles under normal assumption
     #
     # if not implementing, delete _predict_proba
-    def _predict_proba(self, fh, X):
+    def _predict_proba(self, fh, X, marginal=True):
         """Compute/return fully probabilistic forecasts.
 
         private _predict_proba containing the core logic, called from predict_proba
```
GuzalBulatova, line 390: Not in the current PR, but could you kindly update docstr, the lines below are unclear to me and aren't according to docstr standard. `X: one of self.get_tag("X_inner_mtype"), optional, default=None`, `marginal: bool, optional, default=True`. What does "guaranteed to be ForecastingHorizon" mean?

ngupta23:
> It's only the underscore transform, you can't use this as a transformer yet - is that what you want?

Yes, correct. Ability to transform X is what I wanted.

extension_templates/forecasting.py
fkiraly: hm, can you be precise what should be improved with the docstring? I agree that it might be obscure and should be improved, but is there an additional problem?
Will fix in separate PR.

> "guaranteed to be ForecastingHorizon"

The underscore function is a function that is called only by other code, never by the user.
"guaranteed" in such a context means: those other functions only call it with certain inputs (instance of `ForecastingHorizon`), and that is guaranteed by the architecture, logic, tests, etc.

fkiraly:
> Yes, correct. Ability to transform X is what I wanted.

Let's leave the other issue open for a moment, I'll see whether we can do this by inheriting from `BaseTransformer`, once this PR is merged.

ngupta23:
@aiwalter @GuzalBulatova Would appreciate it if you could complete this review quickly. With the deprecation of `return_pred_int` in 0.11.0, a few things are breaking in `pycaret` so a release of this PR's update in 0.11.1 would be much appreciated. Thanks for your help with this!

extension_templates/forecasting.py
fkiraly: I've updated the docstrings in this PR which also updates the error message:
https://github.com/alan-turing-institute/sktime/pull/2314

I hope this is clearer?

I've separated PR to keep the different changes compartmentalized.

fkiraly:
Merging as discussed with @aiwalter and @TonyBagnall

PR #2320, opened 2022-03-27
[ENH] `AutoETS` prediction intervals simplification

fkiraly:
This PR simplifies the probabilistic prediction functionality in `AutoETS`.

Previously, `_predict_quantiles` was implemented, but `AutoETS` returns intervals. Hence `_predict_quantiles` was removed and replaced by a more direct interface through `_predict_interval`.

By discussion below, the interface was previously correct, most probably, so this is merely a simplification of the logic.

### DIFF
--- a/sktime/forecasting/base/_base.py
+++ b/sktime/forecasting/base/_base.py
@@ -1529,7 +1529,7 @@ class BaseForecaster(BaseEstimator):
             # variable names (unique, in same order)
             var_names = idx.get_level_values(0).unique()
             # if was univariate & unnamed variable, replace default
-            if var_names == ["Quantiles"]:
+            if len(var_names) == 1 and var_names == ["Quantiles"]:
                 var_names = ["Coverage"]
             # idx returned by _predict_interval should be
             #   3-level MultiIndex with variable names, coverage, lower/upper
@@ -1603,7 +1603,7 @@ class BaseForecaster(BaseEstimator):
             # variable names (unique, in same order)
             var_names = idx.get_level_values(0).unique()
             # if was univariate & unnamed variable, replace default
-            if var_names == ["Coverage"]:
+            if len(var_names) == 1 and var_names == ["Coverage"]:
                 var_names = ["Quantiles"]
             # idx returned by _predict_quantiles should be
             #   is 2-level MultiIndex with variable names, alpha
--- a/sktime/forecasting/ets.py
+++ b/sktime/forecasting/ets.py
@@ -448,11 +448,11 @@ class AutoETS(_StatsModelsAdapter):
         )
 
         pred_int = pd.DataFrame()
-        for coverage in coverage:
-            pred_statsmodels = prediction_results.pred_int(1 - coverage)
+        for c in coverage:
+            pred_statsmodels = prediction_results.pred_int(1 - c)
             pred_statsmodels.columns = ["lower", "upper"]
-            pred_int[(coverage, "lower")] = pred_statsmodels["lower"].loc[valid_indices]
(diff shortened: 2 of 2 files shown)


ngupta23:
Hi @fkiraly , I tested these changes with the fix branch, but I still see a difference in the results

image

ngupta23:
Actually, there seems to be some randomness to this process.  A second and third run of the same code gives different results:

image

fkiraly:
@ngupta23, I was already going crazy, all seemed fine...

It does not seem possible to set a random state in there, does it?

ngupta23:
@fkiraly there is no option to set any seed. Also I think the point forecasts are reproducible. The randomness is in the intervals and quantiles.

fkiraly:
So, all looks plausible then!

Since the logic in the conversion intervals/quantiles is identical (per inheritance) with that used in another case where we know it works (ARIMA et al), this is actually, most likely, not-a-bug.

Hence, this PR is a clean-up and not a bugfix.

ngupta23:
@fkiraly - Here is another example. Asking for the same interval multiple times results in different results. Could it be that the prediction is changing some internal state in the model?

Why do you think this is not a bug? Should it not be reproducible?

```python
print(forecaster.predict_quantiles(alpha=[0.1, 0.1, 0.1, 0.5, 0.5, 0.9, 0.9, 0.9]))
```

image

fkiraly:
> Why do you think this is not a bug? Should it not be reproducible?

Only for a fixed random state, which this model does not have.

Many models are non-deterministic, e.g., as they do some random sampling inside, especially when it comes to estimating variation. Having some intrinsic randomness there is not unusual and expected.

The fluctuations seem large for this particular model, and this may go together with sub-optimal performance in predicting uncertainty. But, I currently see no reason to believe that it would be "wrong" in the sense of correctly interfacing whatever the original `AutoETS` does.

ngupta23:
> Many models are non-deterministic, e.g., as they do some random sampling inside, especially when it comes to estimating variation. Having some intrinsic randomness there is not unusual and expected.

@fkiraly You are right, it looks like in this case, the intervals are based on simulations and those simulations are not exact or repeatable. This model has multiplicative components and hence exact method is not available

https://www.statsmodels.org/stable/_modules/statsmodels/tsa/exponential_smoothing/ets.html

```python
if method is None:
    exact_available = ["ANN", "AAN", "AAdN", "ANA", "AAA", "AAdA"]
    if results.model.short_name in exact_available:
        method = "exact"
    else:
        method = "simulated"
self.method = method
```

However, they can be made repeatable by passing a `random_state` inside the sktime ets.py `_predict_interval` method. What do you think - maybe open an enhancement request for this?  

```python
def _predict_interval(self, fh, X=None, coverage=None):
    valid_indices = fh.to_absolute(self.cutoff).to_pandas()

   start, end = valid_indices[[0, -1]]
   # Maybe add random_state in get_predictions (and accept random_state in the AutoETS initialization)?
   # https://www.statsmodels.org/stable/_modules/statsmodels/tsa/exponential_smoothing/ets.html
   prediction_results = self._fitted_forecaster.get_prediction(
       start=start, end=end
   )
```

fkiraly:
> @fkiraly You are right, it looks like in this case, the intervals are based on simulations and those simulations are not exact or repeatable. This model has multiplicative components and hence exact method is not available

I also looked into it - it seems like the `simulate` method has a `random_state` that can be set?
But it does not seem like it can be easily called from the model prediction?

ngupta23:
> But it does not seem like it can be easily called from the model prediction

I think the simulate kwargs can be passed to `get_prediction` method. So should be possible - right?

https://www.statsmodels.org/stable/_modules/statsmodels/tsa/exponential_smoothing/ets.html

image

PR #2322, opened 2022-03-27
[ENH][BUG] fixes and implementations of missing `inverse_transform` in compositions

fkiraly:
This PR contains an upgrade of `_has_capability` that works with classes as well as estimators.

The upgrade uncovered a number of estimators that did not have `inverse_transform` implemented, or bugs in it:
* bug: in the `TransformerPipeline`, inverse transforms should be applied backwards, but they were applied forward. This was fixed
* bug: in the `TransformerPipeline`, if all transformers had empty fit, `fit` should have been called in `inverse_transform`. This was fixed.
* `FeatureUnion` had `inverse_transform` tag that said it had one, but it is in fact unclear what the `inverse_transform` should be. Naively concatenating inverses is not inverse to `transform`. Hence, the `capability:inverse_transform` tag is now always `False`.

Depends on #2324 due to the column naming issue with `FeatureUnion`.


fkiraly:
(doc failure: docs are ok and unchanged, but readthedocs seems to be down)

fkiraly:
Merged as discussed with @aiwalter and @TonyBagnall.

PR #2325, opened 2022-03-27
[ENH] memory for `WindowSummarizer` to enable `transform` windows to reach into the `fit` time period

fkiraly:
This PR is a proof-of-concept which gives `WindowSummarizer` a memory of its training data, avoiding the problem identified by @grzegorzrut in https://github.com/alan-turing-institute/sktime/discussions/2309, and based on @grzegorzrut's workaround.

This initial attempt might be inefficient since it also summarizes windows which do not result in a summary returned by `transform`, but should show how the current logic can easily be adapted.
It could be improved along the lines of @grzegorzrut's suggestion to remember only the latest `min_window` observations, since nothing older will participate in a window looked at in `transform` - although older windows could be queried if the `transform` input is in the past of the current latest observation.

I've also added an `_update` method, which simply updates the memory.


fkiraly:
comments by @grzegorzrut, from #2309:

1. `pd.concat([self._X_memory, X]) ` - this assumes that indices in X are not the same as in `self._X_memory`.
Should the code try to deduplicate records by default or raise a flag that some index values are duplicated?

2. I'm not sure if this is a problem, should data be sorted on the index? This is connected with the 'issue' above. Both points are a bit artificial, I guess.

3. Should there be a check on the completeness of an index?
For instance, if `self._X_memory` has [1940,1941] in the index and X has  [1943,1944] - should the updated memory have
[1940,1941,1942, 1943, 1944] or throw an error?

4. In principle we can focus only on `self._target_cols` in the memory - the rest is not relevant for the window functions.
This would require to have 
`self._X_memory = X[self._target_cols]` 
after the definition of `self._target_cols` .

5. I was not able to confirm that the code integrates well with other classes like `ForecastingPipeline`. Wrote about it on slack.

fkiraly:
>1. `pd.concat([self._X_memory, X]) ` - this assumes that indices in X are not the same as in `self._X_memory`.
Should the code try to deduplicate records by default or raise a flag that some index values are duplicated?

Non-duplicate indices are not explicitly part of our data specification, but possibly they should be.
Will think about this.

> 2. I'm not sure if this is a problem, should data be sorted on the index? This is connected with the 'issue' above. Both points are a bit artificial, I guess.

Yes, monotonic sorting *is* part of our data specification, you are right. The proposed implementation will cause problems if the transform is called on indices already seen, or in the range already seen.

> 3. Should there be a check on the completeness of an index?
> For instance, if `self._X_memory` has [1940,1941] in the index and X has  [1943,1944] - should the updated memory have
[1940,1941,1942, 1943, 1944] or throw an error?

I don't think this is strictly required - the references in the `WindowSummarizer` are integer location differences, so if a year is skipped, it is skipped, and that is as documented (even if slightly counterintuitive).

Side note: would be interesting to extend `WindowSummarizer` to `timedelta` input, btw. Nice project if you'd like?

> 4. In principle we can focus only on `self._target_cols` in the memory - the rest is not relevant for the window functions.
This would require to have 
`self._X_memory = X[self._target_cols]` 
after the definition of `self._target_cols` .

Yes, this is indeed an improvement.

> 5. I was not able to confirm that the code integrates well with other classes like `ForecastingPipeline`. Wrote about it on slack.

Seems to work!
Just for completeness, you posted this code on slack:

```python
from sktime.datasets import load_airline, load_longley
import pandas as pd
y, X = load_longley()
kwargs = {
    "lag_config": {
        "mean": ["mean", [[3, 1], [4, 1]]],
    }
}
ws = WindowSummarizer(**kwargs, n_jobs=1, target_cols=["POP", "TOTEMP"])
Z_train = pd.concat([X_train,y_train],axis=1)
Z_test = pd.concat([X_test,y_test],axis=1)
ws.fit(Z_train,y_train)
ws.transform(Z_test)
```

For me, that works and returns this table, which I guess is the expected one:
image

GrzegorzRut:
> > 3. Should there be a check on the completeness of an index?
> >    For instance, if `self._X_memory` has [1940,1941] in the index and X has  [1943,1944] - should the updated memory have
> >    [1940,1941,1942, 1943, 1944] or throw an error?
> 
> I don't think this is strictly required - the references in the `WindowSummarizer` are integer location differences, so if a year is skipped, it is skipped, and that is as documented (even if slightly counterintuitive).
> 
> Side note: would be interesting to extend `WindowSummarizer` to `timedelta` input, btw. Nice project if you'd like?

Sure! I'm not sure how much time I'd be able to dedicate to this but I'd be happy to help.

> For me, that works and returns this table, which I guess is the expected one: image

Yup, looks fine!

fkiraly:
> `pd.concat([self._X_memory, X])` - this assumes that indices in `X` are not the same as in `self._X_memory`.

This has now been replaced by `combine_first` which should address the bug.

I won't deal with `target_cols` in this PR, since it's a reference anyway, so doesn't consume additional memory (and subsetting increases complexity of the code). Is this reasoning correct?

GrzegorzRut:
Ah, this was a very important change (`combine_first`). Thank you for catching this.

PR #2326, opened 2022-03-27
[ENH] require uniqueness from multiple alpha/coverage in interval/quantile forecasts

fkiraly:
This PR adds an additional assumption to `predict_interval` and `predict_quantiles`: that values in potential lists `alpha` and `coverage` should be unique.

Related to discussion in https://github.com/alan-turing-institute/sktime/pull/2320.

Reason for adding these assumptions is that not all implemented methods may work if non-unique values are passed, e.g., if indexing of columns happens.

A check for uniqueness has been added to the utility `check_alpha`; docstrings, comments, and error messages have been improved.


ltsaprounis (approved):
looks good!

PR #2328, opened 2022-03-28
[ENH] Prohibit incompatible splitter parameters

khrapovs:
#### Reference Issues/PRs

Next small step in #1737.

My grand plan is to generalize splitters to accept more argument types in splitters. Along the way I realized that there are too many possible combinations of input parameters to treat at the same time. So, a better strategy would be to generalize in very small steps. In this PR I simply raise a `TypeError` if input arguments are either not all ints (as before I started working on #1737), or not all timedelta/dateoffset. I do this check immediately at the initialization of splitter objects. Tests are added to check that the exception is risen.

Note that I have also replaced `TEST_WINDOW_LENGTHS` (includes ints and timedeltas) with `TEST_WINDOW_LENGTHS_INT` (only ints) in `sktime.forecasting.tests.test_all_forecasters.TestAllForecasters.test_update_predict_predicted_index` to restore the original set of tests existed before the work on #1737 has started.

No

##### For all contributions


khrapovs:
@fkiraly @aiwalter Would you mind taking a look?

fkiraly (changes requested):
@khrapovs, thanks for this, and sorry for the delay in response.

High-level, I think the intended end state is sensible.

I was thinking about this, and also with the usual point in mind, do we need to do any deprecation, if yes where.
As far as I can see there is no deprecation needed, since we are downwards compatible, right?

Change request is, I would make one major change suggestion, to structure/design.
This is to ensure maintainability and understandability of the code.

Instead of having utility functions with long names and difficult to understand input/output relation, I'd suggest to have utilities of the kind `is_iloc_like` and `is_time_like` or similar, and perhaps a `index_type` utility that returns `"time"` or `"iloc"`. The first one check whether you have time stamps, time deltas, etc; the second checks whether you have integers or integer difference indices.

The checks would then be easy, you check whether all arguments have the same `index_type`. This also opens up later extensions for conversions.

Btw, @lmmentel made me aware of a nice package that might help here (you may like to look at it), `visions` for declarative typing and type checking.

Please let me know if you think my request for a partial re-design is sensible or you'd rather not.

khrapovs:
@fkiraly Thanks, as usual, for a very insightful review! I have done the following refactoring.

- New functions in `sktime.utils.validation.__init__`: `is_iloc_like`, `is_time_like`, `all_inputs_are_iloc_like`, and `all_inputs_are_time_like`.
- Instead of 6 new very specific utility functions in `sktime.forecasting.model_selection._split` I have now only 2 very generic ones: `_inputs_are_supported` and `_check_inputs_for_compatibility`.

Personally, I love it! :D

I looked into `versions` library. It is very nice and very good to know. But I am not sure it is applicable here. At the very least for the reason you once expressed yourself: we don't want to introduce new dependencies without a very good reason.

sktime/utils/validation/__init__.py
```diff
@@ -83,6 +83,36 @@ def array_is_timedelta_or_date_offset(x) -> bool:
     return all([is_timedelta_or_date_offset(value) for value in x])
 
 
+def is_iloc_like(x) -> bool:
+    """Check if input is .iloc friendly."""
+    try:
+        iter(x)
```
fkiraly, line 89: I first thought this is really strange, but a quick investigation seems to confirm that this is one of the most reliable ways to check whether an object is iterable...

Either way, this occurs two times, so I´d recommend to factor this check out in a function `is_iterable`.

fkiraly:
Hm, looks like the suggestion has reduced the amount of code quite a bit?
Just a quick question before deeper review: is that all explained by "the right abstraction", or did you also remove some logic or functionality?

sktime/utils/validation/__init__.py
khrapovs: done, pushed

khrapovs:
"if you torture the data long enough it will confess". I guess the same applies to refactoring :)))))

sktime/forecasting/model_selection/tests/test_split.py
```diff
@@ -149,18 +146,25 @@ def _check_cv(cv, y, allow_empty_window=False):
...
-    cv = SingleWindowSplitter(fh=fh, window_length=window_length)
-    train_windows, test_windows, cutoffs, n_splits = _check_cv(cv, y)
-
-    train_window = train_windows[0]
-    test_window = test_windows[0]
-    assert n_splits == 1
-    assert train_window.shape[0] == _coerce_duration_to_int(
-        duration=window_length, freq="D"
+    fh_horizon = (
+        ForecastingHorizon(fh) if not isinstance(fh, ForecastingHorizon) else fh
```
fkiraly, line 150: I do not think the test should do this, but the splitter should.
In fact, it is doing this in `get_fh` (but not on construction). Perhaps at some point this should be changed to an attribute, like in `BaseForecaster`.

fkiraly (dismissed):
Great!
I think this is ok to go, below are non-blocking suggestions.

I just looked at the tests, and I have noticed at a few things:
* there is coercion to `ForecastingHorizon` in the test. That should not happen in the test, but in the splitter (where it already does happen?). See my comment above.
* there is now a lot of redundant testing that error is raised if the splitter is constructed with incompatible input combinations, for each splitter, and this is tested together with all kinds of other stuff. Is there a way to make the test logic easier, i.e., to loop over splitters and test the error handling for incompatible inputs, and only that?

sktime/forecasting/model_selection/tests/test_split.py
khrapovs: Replaced all instances of
```
    fh_horizon = (
        ForecastingHorizon(fh) if not isinstance(fh, ForecastingHorizon) else fh
    )
```
with
```
check_fh(fh)
```
My bad. Forgot about this small utility function.

fkiraly: Still, are we sure we want checks in the tests?
I think the *objects* should be doing this, and not the *tests*.

khrapovs:
@fkiraly 

> * there is coercion to `ForecastingHorizon` in the test. That should not happen in the test, but in the splitter (where it already does happen?). See my comment above.

For some, probably good, reason incoming `fh` is assigned to `self.fh` "as is" without conversion. Function `check_fh` is then used much later at some other various points in time. Since this PR is primarily not about `ForecastingHorizon` I would prefer to mess with it here as little as possible.

> * there is now a lot of redundant testing that error is raised if the splitter is constructed with incompatible input combinations, for each splitter, and this is tested together with all kinds of other stuff. Is there a way to make the test logic easier, i.e., to loop over splitters and test the error handling for incompatible inputs, and only that?

That's a very good point and I also scratched my head about it. Let me present my argument for the current state. In order to remove `if...else` branching from the tests I have to leave only supported combination of parameters, say only integer and only time-like combinations. This would substantially complicate the usage of now straightforward `@pytest.mark.parametrize` decorators. Creating a list of incompatible parameters is also not so straightforward as all possible permutations of some lists, especially where we talk about 4 arguments at the same time. Finally, these tests are super fast. Does this sound reasonable?

fkiraly:
> Since this PR is primarily not about ForecastingHorizon I would prefer to mess with it here as little as possible.

Not what I mean. Preveiously, the test would pass `fh` to the test logic unaltered. Now, you run it through the `check_fh` or `ForecastingHorizon`, which means only `ForecatsingHorizon` type `fh` are passed to the test logic, and `int` etc is passed no more. This results in the test having less coverage, namely for the case where you pass list of int etc directly.

> Does this sound reasonable?

For the tests, ok - not 100% happy, but I also cannot spell out an easier solution right now.

khrapovs:
Wait a sec. In the tests I pass "raw" `fh` to splitters. `check_fh` is only used to branch between test success/fail tests. Besides, these are unit tests for splitters, not for forecasting horizon.

fkiraly:
> Wait a sec. In the tests I pass "raw" `fh` to splitters. `check_fh` is only used to branch between test success/fail tests. Besides, these are unit tests for splitters, not for forecasting horizon.

If I may rephrase, are you saying there is an argument why the additional introduction of `check_fh` does not reduce test coverage?

Can I ask conversely: if that is the case, why do we introduce `check_fh` at the start of the test at all, instead of just looping over `ForecastingHorizon` typed `fh`?

khrapovs:
This is very embarrassing, but it turns out that after my last refactoring effort `check_fh` is no longer needed in `_inputs_are_supported`. I have removed `check_fh` from `_inputs_are_supported` in testsand in `_fh.py` itself. Sorry... I hope it brings your happiness level to 100% now :D

fkiraly (approved):
Yes, looks good now!
Happiness at 120% and rising.

PR #2333, opened 2022-03-28
[ENH] Allow `pd.Timedelta` values in `ForecastingHorizon`

khrapovs:
#### Reference Issues/PRs

One more contribution to #1737.

In this PR I added `TEST_FHS_TIMEDELTA` test cases to  `sktime.forecasting.base.tests.test_fh.test_fh`. In order to pass the tests two changes were required in `ForecastingHorizon` itself: in `sktime.forecasting.base._fh.ForecastingHorizon._is_in_sample` and in `sktime.forecasting.base._fh.ForecastingHorizon._is_out_of_sample`. Both methods perform comparison with zero which in case of `pd.Timdelta` values should actually be `pd.Timedelta(0)`.

Note that `sktime.forecasting.base._fh.ForecastingHorizon.to_indexer` returns `None` in the new timedelta case. I have no idea what this method should return in this case. Any suggestions are very much welcome.

In addition, I have silenced the pre-commit error from (described here: https://github.com/alan-turing-institute/sktime/issues/2338) by using `# noqa: B019` next to the `@lru_cache` decorator.

No

##### For all contributions


khrapovs:
@fkiraly @aiwalter Would you mind taking a look?

sktime/forecasting/base/_fh.py
```diff
@@ -466,7 +471,12 @@ def is_all_in_sample(self, cutoff=None):
     def _is_out_of_sample(self, cutoff=None):
         """Get index location of out-of-sample values."""
         # return ~self._in_sample_idx(cutoff)
-        return self.to_relative(cutoff).to_pandas() > 0
```
fkiraly, line 469: redundant - should be simply `return not _is_in_sample(self, cutoff=cutoff)`, no?

sktime/forecasting/base/_fh.py
```diff
@@ -504,7 +514,11 @@ def to_indexer(self, cutoff=None, from_cutoff=True):
             Indexer.
         """
         if from_cutoff:
-            return self.to_relative(cutoff).to_pandas() - 1
+            try:
```
fkiraly, line 517: should be timedelta based index, no?

sktime/forecasting/base/_fh.py
khrapovs: correct! but it is not `not`, but rather `np.logical_not`. too bad that there are no type suggestions, otherwise I would have spent much less time to figure out why a simple `not` does not work...

sktime/forecasting/base/_fh.py
khrapovs: The more I think about it, the less sense this method (`to_indexer`) makes to me. The whole reason I want to extend to timdelta/dateoffset is to break the link between time series index and forecasting horizon. Let's say we have a time series with the following date index: `[Jan 1, Jan 2, Jan 2, Jan 5]`. The integer index here is `[0, 1, 2, 3]`. But if I start thinking in relative time-based terms, say `fh = 1 day`, and a `cutoff = Jan 2`, then indexing into the this time series is nothing but defining a time threshold and simply filtering out either values before or after this threshold, say `Jan 3`, or `y[y.index < cutoff + fh]`. In such arithmetic integer index is just not very helpful.

sktime/forecasting/base/_fh.py
fkiraly: ah... sorry to hear 😄

sktime/forecasting/base/_fh.py
fkiraly: I don't think filtering should play a role here, since in the deployment setting, the time stamps for which you forecast have not been yet observed, hence do not exist in the form of a `pd.DataFrame` with data in it, a-priori.

khrapovs: exactly! I have the same argument in my head, just in different words, that blocks me from writing anything in this method. do you have a clear picture how this method should be used in case of timedelta horizons?

fkiraly: hmm, what needs to be done/decided to resolve this? I'm slightly lost where this was going.

khrapovs: I have already asked @aiwalter for his opinion on this matter. To me it looks like a completely new issue, implement `to_indexer` for timedelta horizons, which still does not make any sense to me. How about `NotImplementedError` for the time being and allowing me to move forward with my main idea?

fkiraly: hm, yes.

Then, let's do as follows:
* kindly open an issue on this identified item
* here, kindly summarize your argument why we can skip this, and how we ensure downwards compatibility in the matter

khrapovs: - https://github.com/alan-turing-institute/sktime/issues/2376
- In this PR the core change is the comparison with `pd.Timedelta(0)` in `sktime.forecasting.base._fh.ForecastingHorizon._is_in_sample`. This comparison allows me to move forward with generalizing splitters in `sktime.forecasting.model_selection._split` to work with non-integer windows, cutoffs, and forecasting horizons. `to_indexer` has nothing do with these changes. Hence, I would prefer to stay focused on my original plan here https://github.com/alan-turing-institute/sktime/issues/1737.

fkiraly: this is the one, #2377, right?

khrapovs: Correct, the new issue is #2376

Issue #2332, opened 2022-03-28
[ENH] Replace deprecated `pandas.Int64Index` with `pandas.Index`


khrapovs:
**Is your feature request related to a problem? Please describe.**

`pandas.Int64Index` is deprecated since pandas 1.4.0 and will be removed in 2.0. https://pandas.pydata.org/docs/reference/api/pandas.Int64Index.html

Now we get the warning:
```
FutureWarning: pandas.Int64Index is deprecated and will be removed from pandas in a future version. Use pandas.Index with the appropriate dtype instead
```

**Describe the solution you'd like**

Replace with pandas.Index. https://pandas.pydata.org/docs/reference/api/pandas.Index.html

fkiraly:
Duplicate of #2049, closing #2049 in favour of this since it is better described here.

An attempt was already made by @gcchablani https://github.com/alan-turing-institute/sktime/pull/2082, it is probably abandoned.
Naive, ulk replace does not work, since we had used the type in type checking as well as type casting and type enforcement, replacing by `Index` is too permissive.

fkiraly:
... @aiwalter, maybe not a good first issue? As said, bulk replace does *not* work, and one needs to understand various intricate pieces of logic to get this fixed.

PR #2339, opened 2022-03-29
[ENH] Partially replace `pd.Int64Index` with `pd.Index`

khrapovs:
#### Reference Issues/PRs

Partially addresses #2332.

I have replaced `pd.Int64Index` with `pd.Index` only in obvious and harmless places. More complicated like `VALID_INDEX_TYPES` in `sktime.utils.validation.series` are left for another ambitious PR.

No

##### For all contributions


khrapovs:
> not _entirely_ harmless, is it? smile image

OMG, what did I got myself into! now I see why others abandoned it :D

fkiraly (changes requested):
Take that, `Int64Index`!

In-principle, this look ok, *but* - we need to deprecate!

Why: suppose some poor user's highly valuable deployed pipelines are relying entirely on `Int64Index` for some reason. This PR would completely break it, wouldn't it?

Breakage occurs wherever we do input checks and the index is no longer allowed.
In those cases, we need to raise a warning (deprecation since 0.12, removal in 0.13), while being defensive in the sense of making sktime robust against the `pandas` deprecation...

In the many cases where we *create* an index, there is no deprecation necessary.

khrapovs:
@fkiraly Yes, you are right. _But_, right now there is no way to say whether a user created the index using `pd.Index` or `pd.Int64Index`:
```
type(pd.Index([1, -1]))
<class 'pandas.core.indexes.numeric.Int64Index'>
type(pd.Int64Index([1, -1]))
<input>:1: FutureWarning: pandas.Int64Index is deprecated and will be removed from pandas in a future version. Use pandas.Index with the appropriate dtype instead.
<class 'pandas.core.indexes.numeric.Int64Index'>
```

Additionally, from https://pandas.pydata.org/docs/reference/api/pandas.Int64Index.html:
> Deprecated since version 1.4.0: In pandas v2.0 Int64Index will be removed and NumericIndex used instead. Int64Index will remain fully functional for the duration of pandas 1.x.

But `NumericIndex` does not even exist yet!

fkiraly:
@khrapows, I see. That makes things difficult.
Is there any way to not make *us* break our interface contracts?
How about replacing checks for index *types* by uses of the `pandas` method `Index.is_integer`?

khrapovs:
very good suggestion! I completely overlooked that such a method exists. replaced all usages of `is_numeric` with `is_integer`. thanks!

fkiraly (changes requested):
Nice!
I think this could almost go in, but I'm concerned about the maintainability of the type check, where we now repeatedly do sth like `isinstance(obj, VALID_TYPES) or (isinstance(obj, Index) and obj.is_integer())`.

"Why" is hard to understand to someone who hasn't followed our discussion, and it's also repeated in multiple places across the codebase now.

I would hence recommend to bundle this in one function, perhaps
`is_valid_fh_index(obj, type="any")`, where `type` can be `"relative"` or `"absolute"` - what do you think?

khrapovs:
@fkiraly 

Very good! Added two new functions:
- `sktime.utils.validation.series.is_in_valid_index_types`
- `sktime.utils.validation.series.is_integer_index`

The check for `RELATIVE_TYPES` or `ABSOLUTE_TYPES` happens only a couple of times in the single file `_fh.py`. A combined check for either valid index or integer index only happens just two times. I do not think it makes sense to over-engineer for such rare use cases. But if you insist of a function with `type` argument, I would be happy to write a couple more lines of code.

fkiraly:
> But if you insist of a function with type argument, I would be happy to write a couple more lines of code.

No, I do not insist. I just wanted to bundle the logic in one place, which you did.

Once the tests pass, I'm happy to approve.

fkiraly (changes requested):
Looking at it again, this is not entirely what I meant - I meant we should bundle the logic "check in relative/absolute types plus check is integer index" in one function.

That's because previously it was one line, since the "check is integer index" was, previously, "check for certain type".

So, possibly, one function `is_relative_fh_type` which does `isin RELATIVE_TYPES or is_integer_index` and `is_absolute_fh_type` which does `isin ABSOULTE_TYPES or is_integer_index`?

khrapovs:
pushed two new functions: `functions is_relative_fh_type` and `and is_absolute_fh_type`

fkiraly:
😱 
image
what is this now - seems unrelated?

Maybe let's record this as a sporadic failure?

khrapovs:
> scream image what is this now - seems unrelated?
> 
> Maybe let's record this as a sporadic failure?

Reported here: https://github.com/alan-turing-institute/sktime/issues/2368

sktime/forecasting/base/_fh.py
```diff
@@ -95,19 +107,20 @@ def _check_values(values: Union[VALID_FORECASTING_HORIZON_TYPES]) -> pd.Index:
     # isinstance() does not work here, because index types inherit from each
     # other,
     # hence we check for type equality here
-    if type(values) in VALID_INDEX_TYPES:
+    if is_in_valid_index_types(values) or is_integer_index(values):
```
fkiraly, line 110: this should be `is_relative_fh_type or is_absolute_fh_type` imo, no?

sktime/forecasting/base/tests/test_fh.py
```diff
@@ -152,7 +153,7 @@ def test_check_fh_values_duplicate_input_values(arg):
 def test_check_fh_absolute_values_input_conversion_to_pandas_index(arg):
     """Test conversion of absolute horizons to pandas index."""
     output = ForecastingHorizon(arg, is_relative=False).to_pandas()
-    assert type(output) in VALID_INDEX_TYPES
+    assert is_in_valid_index_types(output) or is_integer_index(output)
```
fkiraly, line 156: `is_relative_index_type or is_absolute_index_type`

sktime/utils/validation/series.py
```diff
@@ -27,6 +27,11 @@
 )
 
 
+def is_in_valid_index_types(x) -> bool:
+    """Check that the input type belongs to the valid index types."""
+    return isinstance(x, VALID_INDEX_TYPES)
```
fkiraly, line 32: I would actually change this to `isinstance(x, VALID_INDEX_TYPES) or is_integer_index(x)` and then use it everywhere

fkiraly (changes requested):
Ok, one last change request:
* change `is_in_valid_index_types` as suggested to include the `is_integer_index` check
* move it to `_fh.py`, next to the two other functions
* then use it in the two other places where you check that manually

sktime/forecasting/base/_fh.py
fkiraly: better probably `is_in_valid_index_types` once it´s changed (see below)

sktime/forecasting/base/tests/test_fh.py
fkiraly: better probably `is_in_valid_index_types` once it´s changed (see below)

fkiraly:
(I'm not insistent about my idea `is_valid_fh_index(obj, type="any")`, where `type` can be `"relative"` or `"absolute"` - also happy with three loose functions - but just wanted to bring it up again in case it seems more natural now)

sktime/forecasting/base/_fh.py
khrapovs: done

sktime/forecasting/base/tests/test_fh.py
khrapovs: done

sktime/utils/validation/series.py
khrapovs: done

khrapovs:
> * change `is_in_valid_index_types` as suggested to include the `is_integer_index` check

done

> * move it to `_fh.py`, next to the two other functions

`is_in_valid_index_types` is very generic and does not have much connection to `ForecastingHorizon` other than it is used once to check fh values. I would prefer to keep it next to where `VALID_INDEX_TYPES` are defined. This function is used much more frequently by other modules.

> * then use it in the two other places where you check that manually

done

> (I'm not insistent about my idea `is_valid_fh_index(obj, type="any")`, where `type` can be `"relative"` or `"absolute"` - also happy with three loose functions - but just wanted to bring it up again in case it seems more natural now)

Yes, I remember about this proposal and even now would prefer to keep the functions as they are now.

fkiraly (approved):
I'd be happy to approve now, although I leave the merge decision to you after considering my non-blocking comment below.

>`is_in_valid_index_types` is very generic and does not have much connection to `ForecastingHorizon` other than it is used once to check fh values. I would prefer to keep it next to where `VALID_INDEX_TYPES` are defined. This function is used much more frequently by other modules.

Ih that case, I would move *the other way round*, i.e., rename `is_relative_fh_type` to `is_in_valid_relative_index_types` or `is_in_valid_index_types_relative`, and put the three functions together into the `validation` module because they are so very similar.

khrapovs:
> Ih that case, I would move _the other way round_, i.e., rename `is_relative_fh_type` to `is_in_valid_relative_index_types` or `is_in_valid_index_types_relative`, and put the three functions together into the `validation` module because they are so very similar.

Very good. I moved not only `is_in_valid_relative_index_types` and `is_in_valid_absolute_index_types` to `utils/validation/series.py`, but also `RELATIVE_INDEX_TYPES` and `ABSOLUTE_INDEX_TYPES` next to `VALID_INDEX_TYPES` in the same module. I like that!

fkiraly (changes requested):
Great!

I see that, in this newest version, you changed `INDEX_TYPE_LOOKUP` too, but did not address the consequences of that change.

I think we should address those ramifications, before we merge.

There are a number of places in which a check `in INDEX_TYPE_LOOKUP.get(index_type)` occurs, this should be using `is_in_valid_index_types,` no?

khrapovs:
> I see that, in this newest version, you changed `INDEX_TYPE_LOOKUP` too, but did not address the consequences of that change.
> 
> I think we should address those ramifications, before we merge.
> 
> There are a number of places in which a check `in INDEX_TYPE_LOOKUP.get(index_type)` occurs, this should be using `is_in_valid_index_types,` no?

Correct. Thanks for noticing. I have used `is_integer_index` in two places of `test_fh` where `assert isinstance` was testing index type. I have double checked other usages of `INDEX_TYPE_LOOKUP` and do not see anything I should change there.

sktime/forecasting/base/tests/test_fh.py
```diff
@@ -49,7 +53,10 @@ def test_fh(index_type, fh_type, is_relative, steps):
     """Testing ForecastingHorizon conversions."""
     # generate data
     y = make_forecasting_problem(index_type=index_type)
-    assert isinstance(y.index, INDEX_TYPE_LOOKUP.get(index_type))
+    if index_type == "int":
```
fkiraly, line 56: confused. Instead of this if/else construction, why do you not just write 
`assert is_in_valid_index_types(y.index)`?

fkiraly (changes requested):
See above. Should it not be much shorter than the if/else?
Happy to approve if there is a good reason not to.

sktime/forecasting/base/tests/test_fh.py
khrapovs:
```
assert isinstance(y.index, INDEX_TYPE_LOOKUP.get(index_type))
```
This line checks for a very specific index type, not just the set of valid indices. Since the index that corresponds to `int` key in `INDEX_TYPE_LOOKUP` is very generic `pd.Index` (instead of `pd.Int64Index`) I have to check that it is indeed integer index. Hence, this ugly `if...else`.

fkiraly: oh, I seel. How about giving `is_in_valid_index_types` an optional second argument, from the lookup dictionary then? Then it checks only for that string.

khrapovs: Sorry, but at this point I will start to resist. This proposal moves complexity from the test to the source code of the library. We already have a few specific and super simple functions to check for specific index types: `is_integer_index`, `is_in_valid_relative_index_types`, and `is_in_valid_absolute_index_types`. Refactoring those once more to something with optional arguments just to accommodate one single test does not sound very attractive to me.

fkiraly: > from the test to the source code

Ah, that´s a valid point!
It´s a *test*, not the library code!
Ok, I´m fine with the status quo then.

fkiraly (approved):
Discussion revealed a good reason to leave as is, which resolves the last change request.

Issue #2059, opened 2022-02-12
[ENH] Adding DynamicFactor from statsmodels as a forecaster


aiwalter:
**Is your feature request related to a problem? Please describe.**
Adding `DynamicFactor` from `statsmodels` as a forecaster. 

https://www.statsmodels.org/devel/generated/statsmodels.tsa.statespace.dynamic_factor.DynamicFactor.html
**Describe the solution you'd like**
Using the class `_StatsModelsAdapter ` to add this forecaster, should be straight forward and equivalent to other forecasters that already inherit from this class.

Ris-Bali:
@aiwalter  Can I work on this ? .  It should have similar structure to `https://github.com/alan-turing-institute/sktime/blob/main/sktime/forecasting/exp_smoothing.py` which also inherits from _StatsModelsAdapter right ?

aiwalter:
@Ris-Bali sure, yes exactly. Also it might help to look into the enhancement template: https://github.com/alan-turing-institute/sktime/blob/main/extension_templates/forecasting.py

fkiraly:
@Ris-Bali, the latest sktime docs include detail instructions on how to add estimators:
https://www.sktime.org/en/latest/developer_guide/add_estimators.html

fkiraly:
was finished by @lbventura in https://github.com/sktime/sktime/pull/2859 but we forgot to close the issue

PR #2349, opened 2022-03-30
[ENH] Added DynamicFactor model

Ris-Bali:
#### Reference Issues/PRs
Fixes issue #2059 

This PR implements Dynamic Factor Forcaster

No

- [x] Have implemented the fit method of the forcaster.
- 

##### For all contributions

##### For new estimators


fkiraly:
In case you are wondering what is breaking: you have not defined a `get_test_params` function that returns test parameter settings.

Ris-Bali:
Its still breaking I don't know maybe my implementation is wrong , is it because I have not specified the parameters ?

sktime/forecasting/dynamicfactor.py
```diff
@@ -0,0 +1,411 @@
...
+
+        Returns
+        -------
+        params :dict or list of dict , default = {}
+            arameters to create testing instances of the class
+            Each dict are parameters to construct an "interesting" test instance, i.e.,
+            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+            `create_test_instance` uses the first (or only) dictionary in `params
+        """
+        return super().get_test_params()
```
aiwalter, line 411: why are you calling `super` here?

aiwalter, line 411:
```suggestion
        return {"k_factors": 1, "factor_order": 1}
```

aiwalter (changes requested):
I found the reason for test failure

sktime/forecasting/dynamicfactor.py
fkiraly: if you wanted to do this, you can simply delete the function. "inheritance" ensures that, if a function is not implemented, the parent's function is called.

Although you probably don't want to leave it empty?

Ris-Bali: Yes I didn't want to leave it empty ; also I thought the get_test_params method is defined in statsmodeladapter class so it would return parameters for any statsmodel forcaster but now I discovered the method is  from baseobject class in sktime/base/_base.py . I was not aware I needed to return specific params . Thanks for helping me out .

Ris-Bali:
> I found the reason for test failure

Yes I'll correct it as soon as possible . Thanks

aiwalter:
@Ris-Bali I think we should actually default also for the two params `k_factors `and `factor_order `to 1 in the `init`. Whats your take @fkiraly ? In other models afaik we also default always to sth.

fkiraly:
yes, these sound like sensible defaults. That would make the hidden factor series a univariate AR(1), right?

The other possible default could be `k_factors=0`, but that would remove factors and result in a completely cross-sectional least squares fit if I am not mistaken, so it seems like a worse choice.

Ris-Bali:
I still can't find the reason for these unsuccessful checks would appreciate some help . Thank You

fkiraly:
You might have set the tags incorrectly.

The `"scitype:y"` tag should indicate whether the estimator can deal with multivariate data. If you choose to allow multivariate, the `y_inner_mtype` cannot be `pd.Series` because that is always univariate.

Ris-Bali:
> You might have set the tags incorrectly.
> 
> The `"scitype:y"` tag should indicate whether the estimator can deal with multivariate data. If you choose to allow multivariate, the `y_inner_mtype` cannot be `pd.Series` because that is always univariate.
> 
> This is inconsistent - so you need to check whether the estimator can deal with multivariate data, and change either the `y_inner_mtype` tag or the `scitype:y` tag.

Thanks . I'll make the changes as soon as possible.

fkiraly:
> Thanks . I'll make the changes as soon as possible.

Great!
To make your life easier, I recommend you use `check_estimator` to test your estimator locally, that allows you to debug it more easily.

sktime/forecasting/dynamicfactor.py
```diff
@@ -0,0 +1,411 @@
...
+        self.return_params = return_params
+        self.optim_score = optim_score
+        self.optim_complex_step = optim_complex_step
+        self.optim_hessian = optim_hessian
+        self.flags = flags
+        self.low_memory = low_memory
+
+        super(DynamicFactor, self).__init__()
+
+    def _fit_forcaster(self, y, X=None, fh=None):
```
aiwalter, line 198:
```suggestion
    def _fit_forcaster(self, y, X=None):
```

sktime/forecasting/dynamicfactor.py
```diff
@@ -0,0 +1,411 @@
...
+    References
+    ----------
+    [1] Lütkepohl, Helmut. 2007. New Introduction to Multiple Time Series Analysis.
+    Berlin: Springer.
+
+    """
+
+    _tags = {
+        "scitype:y": "multivariate",
+        "ignores-exogeneous-X": True,
```
aiwalter, line 135:
```suggestion
        "ignores-exogeneous-X": False,
```
If I am not wrong, this model can use `X`? https://www.statsmodels.org/dev/generated/statsmodels.tsa.statespace.dynamic_factor.DynamicFactor.html

Ris-Bali: Yes  it can ; that was the change I was going to push .

Ris-Bali: Pushed this change . Thanks

Ris-Bali:
Are these tests failing because of something wrong with the tags ?? I checked the tests but couldn't make much out of them .

fkiraly:
> Are these tests failing because of something wrong with the tags ?? I checked the tests but couldn't make much out of them .

I don't think so - my interpretation is that the `_fit` method of your estimator does not run (for any input).

Ris-Bali:
@fkiraly the fit method runs now I tried check_estimator again it fails for predict_residuals and predict time_index
image
should these tests run as there is no fh in my model in fit .

fkiraly:
@Ris-Bali, great! It looks like you have solved most of the issues!

The remaining ones look like interface compatibility issues around one topic: if you forecast, then the forecast indices should be those of the time points for which you make the forecasts. In your case, you probably start counting at `0` again, instead kindly make sure you have the right indices in the return object.

fkiraly:
superseded by https://github.com/alan-turing-institute/sktime/pull/2859

PR #2359, opened 2022-03-31
[ENH] scenarios for hierarchical forecasting and tests for probabilistic forecast methods

fkiraly:
This PR adds scenarios for hierarchical/panel/global forecasting and tests for the vectorization in the probabilistic prediction methods.

Includes #2355 which this is testing.


fkiraly:
Merging, 7 day rule

PR #2360, opened 2022-03-31
[ENH] Upgrade _has_capability to work for both classes and objects

fkiraly:
This PR contains an upgrade of `_has_capability` that works with classes as well as estimator instances.


fkiraly:
duplicate of https://github.com/alan-turing-institute/sktime/pull/2322

PR #2362, opened 2022-03-31
[BUG] fix `_update` default for late `fh` pass case

fkiraly:
Fixes https://github.com/alan-turing-institute/sktime/issues/2361:

If no `fh` was passed in `fit`, and `update` was called with the default `_update`, it would break as it tried to read `self.fh` which would raise an error. Instead, `self_fh` should be read, which does not raise an error.

### DIFF
--- a/sktime/forecasting/base/_base.py
+++ b/sktime/forecasting/base/_base.py
@@ -1420,29 +1420,36 @@ class BaseForecaster(BaseEstimator):
     def _update(self, y, X=None, update_params=True):
         """Update time series to incremental training data.
 
+        private _update containing the core logic, called from update
+
+        State required:
+            Requires state to be "fitted".
+
+        Accesses in self:
+            Fitted model attributes ending in "_"
+            self.cutoff
+
         Writes to self:
-            If update_params=True,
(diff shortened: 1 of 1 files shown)


Lovkush-A:
Looks fine to me. Do we need a unittest to make sure this bug does not occur again?

Lovkush-A (changes requested):
Just reading base class to check what self._fh is. I noticed the docstring for _update is inaccurate. It does not have entry for update_params when it needs one, and it has entry for fh when fh is not a parameter.

fkiraly:
Yes, there is more wrong with the docstrings, too.
I went over all of them and updated them here:
https://github.com/alan-turing-institute/sktime/pull/2314

I updated `_update` here - it was correct in the extension template, but not in the base class.

Lovkush-A (approved):
looks fine to me. replacing self.fh with self._fh should prevent inapropriate raising errors when self._fh is None.

fkiraly:
passed already before change to docstrings only; readthedocs errors out with 0sec, that´s most likely a connection error?

PR #2364, opened 2022-04-01
[ENH] Extract cached `ForecastingHorizon` methods to functions and avoid B019 error

### LINKED ISSUE #2338: [ENH] @lru_cache in ForecastingHorizon is an error for `flake8-bugbear`

Starting from version 22.3.20 of `flake8-bugbear` (https://github.com/PyCQA/flake8-bugbear/releases/tag/22.3.20) `@lru_cache` decorator in `sktime.forecasting.base._fh.ForecastingHorizon.to_relative` and `sktime.forecasting.base._fh.ForecastingHorizon.to_absolute` raises a pre-commit error. Adding `# noqa: B019` silences the warning, but I do not think this is a very sustainable solution.

Is `@lru_cache` strictly necessary? Is the speed gain so dramatic that it overweights the dangers suggested by flake8? I do not have much experience with caching, so I have no good solution to propose.

khrapovs:
#### Reference Issues/PRs

closes #2338 

Extract `ForecastingHorizon.to_relative` and `ForecastingHorizon.to_absolute` to `_to_relative` and `_to_absolute`, respectively, to allow caching without B019 error.

No

##### For all contributions

### DIFF
--- a/sktime/forecasting/base/_fh.py
+++ b/sktime/forecasting/base/_fh.py
@@ -452,8 +452,9 @@ class ForecastingHorizon:
         return f"{class_name}({pandas_repr}, is_relative={self.is_relative})"
 
 
-# This function is a direct replacement of the `ForecastingHorizon.to_relative` method
-# to avoid B019 error from flake8-bugbear.
+# This function needs to be outside ForecastingHorizon
+# since the lru_cache decorator has known, problematic interactions
+# with object methods, see B019 error of flake8-bugbear for a detail explanation.
 # See more here: https://github.com/alan-turing-institute/sktime/issues/2338
 # We cache the results from `to_relative()` and `to_absolute()` calls to speed up
 # computations, as these are the basic methods and often required internally when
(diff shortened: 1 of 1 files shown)


fkiraly (approved):
Looks good, thanks!

Non-blocking comments:
* might be worth to add to the comments preceding the new, losse functions, a refrence to the B019, or someone might get the idea to put them back
* you remember the `pandas` bug? I think we should maybe change the comments there from "replace when fixed" to "replace when we upgrade our lower pandas bound to a version where this is fixed"? Or we´ll have the bug back on the lower end of our supported versions...

khrapovs:
@fkiraly Addressed both of your suggestions.

fkiraly:
> @fkiraly Addressed both of your suggestions.

Thanks!

May I ask for reformulation of this:

> This function is a direct replacement of the `ForecastingHorizon.to_relative` method to avoid B019 error from flake8-bugbear.

The problem with this is that it's obsccure.
You need to imagine someone who has not been part of our conversation.
Hence they will not get the reference "replacement" which is referring to the earlier state, since they will see only the current state (in the future).

Would suggest to change to 

"This function needs to be outside `ForecastingHorizon` since the `lru_cache` decorator has known, problematic interactions with object methods, see B019 error of `flake8-bugbear`"

khrapovs:
@fkiraly I change the wording as you proposed

PR #2366, opened 2022-04-02
[ENH] `BaseObject.is_composite` utility, relax errors in `BaseClassifier` input checks to warnings for composites

fkiraly:
This PR:
* introduces an `is_composite` method to `BaseObject` which checks whether `self` is a composite object
* relaxes errors in `BaseClassifier` input checks on missing value and unequal length capabilities to warnings, it `self` is a composite
* the message is extended to indicate which class (e.g., out of many in a pipeline) has caused the warning, and which problems there are with the input (not just the first problem)

Further, this adapts and fixes `test_check_capabilities`.
* it needs to be adapted since some errors have changed to warnings
* it needed to be fixed, since the `with pytest.raises` (or `pytest.warns`) should have been applied to each case, not to all of them in once (that checks only that there is at least one occurrence, not per case!)

Why the change in case of composites:
In pipelines and composite, these tags need to be inferred.
This can be done, but is is general complicated. Too strict warnings may lock users out of using certain pipelines - currently, for instance, pipelining a padder or truncator with a classifier that cannot handle unequal length.

Also, there are edge cases where "no missing values" cannot be fully guaranteed as it might depend on inputs, e.g., `SqrtTransformer` or binners. Still, users may want to construct these pipelines and not be locked out of using them at all.
The alternative option would be to give the pipelines the most permissive tags, but that might also cause problems since they may be inaccurate.


MatthewMiddlehurst:
Good fix for the test, and the more descriptive error messages are nice.

I disagree that we should be raising warnings for missing and unequal length cases instead of errors, however. If these errors are being raised, then surely the input data is unequal length or has missing values? If the pipeline someone creates is feeding invalid data into an estimator, they should be locked out of it, imo.

For something like `SqrtTransformer` > classifier, surely there should be a middle step if there is a risk of inputting missing values into an estimator which can't handle them. i.e. `SqrtTransformer` > some missing value replacement transformer > classifier.

fkiraly:
> I disagree that we should be raising warnings for missing and unequal length cases instead of errors, however.

Imo we should leave this to the user, especially since there is no good and complete way to infer the tag/properties in even basic pipelines.
For instance, given the current tags, how would you infer it for the pipeline made out of padder and any classifier that cannot handle unequal length data.

Just go and test it, it breaks because of this check which is too strong!

fkiraly:
The corresponding design principle is "don't nanny the user".
Provide meaningful warnings, but don't prevent them to do sth, just because you can't imagine certain sensible ways in which to combine the components of your package.

MatthewMiddlehurst:
Inputting unequal length series into a classifier which can't handle unequal length series is not sensible.

```

from sktime.classification.interval_based import CanonicalIntervalForest
from sktime.datasets import load_plaid
from sktime.transformations.panel.padder import PaddingTransformer

if __name__ == "__main__":
    train_X, train_y = load_plaid(split="train")
    test_X, test_y = load_plaid(split="test")

    clf = Pipeline(
        [
            ("transform", PaddingTransformer()),
            ("clf", CanonicalIntervalForest.create_test_instance()),
        ]
    )

    clf.fit(train_X, train_y)
    print(clf.score(test_X, test_y))
```
I can run this fine? Does this fail with the sktime classifier pipeline?

fkiraly:
> I can run this fine? Does this fail with the sktime classifier pipeline?

Yes, because it assigns the tags.

Note: the `sklearn` pipeline cannot be used anymore in further `sktime` compositions.

fkiraly:
> Inputting unequal length series into a classifier which can't handle unequal length series is not sensible.

That's not what I'm saying or attempting.
I'm saying there are cases where there is no sensible way to set the tag, because the classifier could or could not handle unequal length series data, *depending on parameter settings and the data input*.

fkiraly:
> The corresponding design principle is "don't nanny the user".

The `sklearn` `Pipeline` is another example of this, btw! We can use it with 3D arrays and it does not complain. If it would check conformance with the input, we would be unable to.

MatthewMiddlehurst:
This seems like an issue with the pipeline and not the capabilities method? With this change if you were to try build any classifier with unequal length data if would throw a warning in its capability checking. 

> Note: the sklearn pipeline cannot be used anymore in further sktime compositions.

Unsure what you mean by this.

fkiraly:
> This seems like an issue with the pipeline and not the capabilities method? With this change if you were to try build any classifier with unequal length data if would throw a warning in its capability checking.

Ok, let me try to set out the argument and split it up.
Can I first gauge your agreement with the following:

* every classifier should have the three capability tags set (missing data, unequal length, multivariate)
* a classifier pipeline should behave like a classifier

MatthewMiddlehurst:
Yes

fkiraly:
Both, ok?

So, logically, it follows that pipelines [transformer][classifier] must also have the three tags and set them.
Now, in your opinion, what should the logic be that sets, say, the "unequal length" tag in a [transformer][classifier] pipeline, and how would it work concretely in the following three cases, where we assume that [classifier] cannot handle unequal length data:

* transformer = padder (parameter = total length)
* transformer = identity
* transformer = binner (parameter = bin width, frequency)

MatthewMiddlehurst:
- "capability:unequal_length": True
- "capability:unequal_length": True
- "capability:unequal_length": True

I will be honest, I'm currently not really interested in how best to dynamically change the tags in the classifier pipeline to ensure it matches the capabilities of all its components. I'm against changing the capability checking for every single classifier as a way of handling it.

If I build TSF with unequal length data, I should get an error, not a warning that I'm about to get an error.

fkiraly:
To confirm, you want the pipeline [identity][classifier] to have the tag `"capability:unequal_length": True`, even though it is identical with the original classifier which has "capability:unequal_length": False?

fkiraly:
> I'm currently not really interested in how best to dynamically change the tags in the classifier pipeline to ensure it matches the capabilities of all its components

If you do not deal with this, you potentially get:
* user frustration since certain pipelines that should be possible to build cannot be built
* logic errors, inconsistencies, or failures in the test suite. E.g., when testing pipelines for handling their input checks.

MatthewMiddlehurst:
Again, this seems like an issue with the pipeline and we are accomodating for it by making our overall capability checking worse.

Are the dynamic tags necessary? Why not just set it all to true? What is the benefit of changing `capability:unequal_length` to true/false after the pipeline has been created?

If it is only for testing maybe just set it up with a configuration which can handle these cases by default and test others configs in bespoke tests if required.

MatthewMiddlehurst:
Can we override _check_capabilities in the pipeline so only that gives warnings maybe?

fkiraly:
>Are the dynamic tags necessary? Why not just set it all to true?

The "all true" or "all false" logic has the following major problems in pipelines:
* "all true": pipelines like [identity][classifier] will let unsuitable inputs through - user frustration, and it's the same kind of thing you want to avoid with the check logic in the first place, by setting the tag to `False` sometimes. Why would the reasoning be different here, just because it is a pipeline?
* "all true": testing. Generic tests that rely on the tag to pass tests and then check for "error raised" or "classifier runs" find the wrong behaviour. Yes, this can be "hacked" in the tests, but it would require an entire layer of coding this for *every parameter setting*, so an overhead of logic and a burden on the extender.
* "all false": pipelines that should run do not, high user frustration. For instance, passing unequal length to [padder][classifier]
* "all false": same problem with the tests.

> Can we override _check_capabilities in the pipeline so only that gives warnings maybe?

That would solve the user frustration from preventing users from running a pipeline, but it would not solve the testing issue.
I'd be happy to do that in all composites as a temporary measure though, would this be a good compromise?

fkiraly:
e.g., a tag `is_composite` and moving to warnings whenever composite?

MatthewMiddlehurst:
I was thinking have a separate `def _check_capabilities(self, missing, multivariate, unequal)` implementation in `ClassifierPipeline` which raises warnings instead of errors. That version would be used instead of the base class version for `ClassifierPipeline`, and we could test for both `UserWarning` and `ValueError` in the `test_check_capabilities` test. 

There may be some repeat if it was extended to other composites i.e. `ColumnEnsembleClassifier`, however. A composite base inheriting from `BaseClassifer` could solve that but may be a bit overkill.

If that does not work, an `is_composite` tag would also work as you suggest. I can see the value of the dynamic tags, just would like to avoid impacting the whole package to get them in. While the pipeline should function the same as other classifiers, it does have some cases like this which set it apart.

fkiraly:
ok, just to be concrete, I would propose to implement an `is_composite` method in `BaseObject`, which returns `True` if the object has been constructed with component `BaseObject`s. In `_check_capabilities` the default behaviour for composites (returning `True`) would be warnings, otherwise exceptions.

MatthewMiddlehurst:
That sounds fine to me.

fkiraly:
Ok, I'll make it so.

I've removed the changes to `BaseClassifier` from this PR hence:
https://github.com/alan-turing-institute/sktime/pull/2367

This now only includes the better tag inference, would appreciate you having a look at it.
It should fix the issue with [padder][classifier] for the sktime pipeline.

fkiraly:
ok, @MatthewMiddlehurst, done - have a look!

MatthewMiddlehurst (approved):
Yep, that looks good. The contents will be useful if any similar issues relating to composite estimators arise in the future.

Thanks for the work.

PR #2367, opened 2022-04-02
[ENH] Capability inference for transformer and classifier pipelines

fkiraly:
This PR adds:
* inference of `TransformerPipeline` and `ClassifierPipeline` on "missing value handling" and "unequal length handling" tags
* two tags which indicate, for transformers, whether they *remove* the complication of missing values or unequal length from input data. This is necessary to allow correct inference in case of padder/truncator like transformers and imputers.
* tests for the tag inference capability.
* tags on unequal length and missing value removal for (as far as I can tell) all refactored transformers with these properties


fkiraly:
Removed the changes to base classifier input checks since they seem controversial.

MatthewMiddlehurst (dismissed):
Have not really looked at the pipelines in full, but don't see any issues with these changes for allowing dynamic tags.

Multiple comments seem to have "iff" where I assume it should be "if". 

Not necessarily for this PR, but transformers seem to have a large amount of tags now which could get confusing. Are all of these necessary?

fkiraly:
> Multiple comments seem to have "iff" where I assume it should be "if".

No, I think it should be iff, that's formal jargon for saying "if and only if"
https://en.wikipedia.org/wiki/If_and_only_if

Q, which of these did you mean:

* you did not know the formal definition and thought it was a type in English
* you knew the formal definition but would contest that the condition is necessary or that it is sufficient

fkiraly:
("iff vs if" was not blocking, so merging)

PR #2369, opened 2022-04-02
[ENH] Transformers module full refactor - part IV, `panel` module (3rd batch)

fkiraly:
This PR refactors more transformers in the `transformations` module to the new interface, and continues #2253.
This is part IV - working on the `transformations.panel` module.

For all transformers in this PR, the following was carried out:
* inheritance parent is now `BaseTransformer`
* adding tags as appropriate
* moving logic to underscore methods `_fit`, `_transform`, and `_inverse_transform` (where applicable)
* removal of check/conversion logic which is already covered by `BaseTransformer`
* linting and cleaning up docstrings

Reviewers should focus on individual transformers and the general pattern.

This PR also includes:
* a fix for https://github.com/alan-turing-institute/sktime/issues/2374, by removing the transposition of numpy arrays throughout the signature transformer module and in `SignatureClassifier`


fkiraly:
ready for review. Removed some transformers from the refactor since they pose problems.

sktime/transformations/panel/signature_based/_compute.py
```diff
@@ -51,13 +62,11 @@ def __init__(
...
+
+        import esig
 
-    def transform(self, data, y=None):
-        # Input checks
-        self.check_is_fitted()
+        data = np.swapaxes(X, 1, 2)
 
         # Path rescaling
         if self.rescaling == "pre":
```
jambo6, line 73: Line 94-96 requires changing to
```
signature = np.stack(
    [transform(x[window.start : window.end].T) for x in data]
).reshape(data.shape[0], -1)
```

fkiraly: ok, did that - let´s see what happens

fkiraly:
> Sorry my bad, just realised this should be transform(x[:, window.start:window.end].T) or transform(x.T[:, window.start:window.end])

Which one now, these are not the same

jambo6:
No its changed from `[window.start:window:end]` -> `[:, window.start:window.end]` i.e. extract from the second dim. This is the same as doing the transpose first and extracting from the first dim so either of the suggestions are fine.

fkiraly:
I see, you are right about the current version.
Still, your two proposed alternative versions will have a different effect, since selecting i-th column then transposing, or transposing, then selecting i-th column are not the same. Transposing, then selecting the i-th column is the same as selecting the i-th *row*, then transposing.

jambo6:
Yeah you're right, I meant`transform(x.T[window.start:window.end])` for the second option

fkiraly:
alright! correct now? 😄

fkiraly:
ok, @jambo6, I feel your advice might have been incorrect - the tests with expected results are now failing, but were passing before. I'm hence reverting to what it was.

jambo6:
Ok having looked at the tests I think this is incorrect - which is obviously a shocker from the POV of my tests. 

The tests pass because one checks the outputs global window, and provided `window.end > n_channels` then `x[window.start:window.end]` just grabs all the data even though its accessing the incorrect dimension.

Another test checks the dimensions of the output of the dyadic window are correct, which is lazy and I should make it check the values. Problem is by doing `transform(x[window.start:window.end])` then the number of features are the same even though the windowing operation is taken along the incorrect dimension.

I'm really confused as to why the checks would fail when you do the suggested changes though. From looking at the code it seems that it must be `transform(x[:, window.start:window.end].T)`.

What is the best way to sort this out? Shall I clone this branch and modify + add a new test, or add a bugfix to this when added to main?

fkiraly:
> What is the best way to sort this out? Shall I clone this branch and modify + add a new test, or add a bugfix to this when added to main?

What this PR does is leave functionality intact to the best of my knowledge, so the issue must also be on main, albeit in different form (e.g., without transposes in certain places).

So, I would suggest, you could branch off this, since we´ll probably merge from a refactor standpoint.

You could also branch off main and see whether it´s an issue there, though that will have ripple effects through the change in the transposition in case this PR is merged. So one of us may then have to go through the code and change it in accordance to whatever gets merged first.

In either case I would suggest a separate PR, since we are working on content now, which is beyond the scope of an interface refactor.

fkiraly:
7 day rule, merging

PR #2370, opened 2022-04-02
[ENH] Deprecation of row transformers

fkiraly:
Addresses https://github.com/alan-turing-institute/sktime/issues/2179

This PR deprecates the row transformers, i.e., `SeriesToSeriesRowTransformers` and `SeriesToPanelRowTransformers`.

The original reason to have these was to allow application of series transformers to panel data. This is functionality which is now integrated into the base class, i.e., `fit` and `transform` automatically vectorize (see https://github.com/alan-turing-institute/sktime/pull/2219).

For the deprecation period, `make_row_transformer` will return the transformer passed to it (since it already has the functionality it was meant to attach), and the row transformers will simply patch through all transformer methods.

Relies on https://github.com/alan-turing-institute/sktime/pull/3087 to ensure panel compatibility of `sklearn` `FunctionTransformer`, by closing the gap between "stateless" scopes.


aiwalter (approved):

fkiraly:
@aiwalter, after removing a million minor inconsistencies in the transformer interfaces, we can now finally deprecate the row transformers!

The only remaining fix is https://github.com/alan-turing-institute/sktime/pull/3087, then we can deprecate the row transformers, then remove them in 0.15.

Issue #2377, opened 2022-04-03
[MNT] remove minute-long tests and docstring examples


fkiraly:
A small number of estimators' tests contribute to >10 minutes of test time:
* proximity sth and HIVE COTE in classification
* facebook prophet
* statsforecast AutoARIMA

We should remove these entirely and/or find much faster test parameters (smaller data, hyperparameter choices).

FYI @MatthewMiddlehurst, @TonyBagnall, @aiwalter, @FedericoGarza:
Culprits are below:

![image](https://user-images.githubusercontent.com/7985502/161435173-3554ff8e-1893-4d10-a337-2cc5dda62b6a.png)

fkiraly:
good first issue since it requires mainly "playing around" with data/parameters and can be solved by anyone who is able to run tests (see https://www.sktime.org/en/stable/developer_guide/continuous_integration.html developer instructions)

PR #2378, opened 2022-04-03
[ENH] Faster classifier example parameters

MatthewMiddlehurst:
Alters classifier example parameters so they don't take as much time (#2377).


fkiraly:
@MatthewMiddlehurst, these choices are failing?

MatthewMiddlehurst:
Remaining tests are stalled/failing only due to the overload of testing going on currently.

For some of these, we are reaching the limit of how much time usage we can reduce by tweaking parameters. I have a hunch a bunch of time is being added from having to compile the numba functions some use, i.e. catch22 related classifiers and I think distances now.

fkiraly:
Thanks! So now it´s mainly Prophet...

fkiraly (approved):
Looks good, thanks!

Btw, in `ClassifierPipeline`, would it be nicer to write
`pipeline = PCATransformer() * TimeSeriesForestClassifier(n_estimators=5)`
?

MatthewMiddlehurst:
I personally prefer how it is currently, but won't object to you changing it before merging if you think it would look better.

fkiraly:
added it as alternative example

TonyBagnall (approved):

PR #2379, opened 2022-04-03
[ENH] refactored `ColumnConcatenator`, rewrite using `pd-multiindex` inner type

fkiraly:
This PR is a rewrite of the `ColumnConcatenator`, using `pd-multiindex`.
Originally in https://github.com/alan-turing-institute/sktime/pull/2369.

The conversion issues are fixed now, should all be good to go.

Relies on the fix to data loaders to ensure unique input index:
https://github.com/alan-turing-institute/sktime/pull/3031

---

from first post:

Unfortunately, this has dredged up all kinds of issues with type casting and conversion, so I am isolating it to one PR and removing it from #2369.

The failures in this refactor are linked to the failures in the conversions here: https://github.com/alan-turing-institute/sktime/pull/2375, since in the tests we make conversions between the nested data frame format and the `pd-multiindex` format.


fkiraly:
Test failure is due to #1290 - the old `ColumnConcatenator` with its byzantine code was apparently able to cope with repeated instance indices, whereas the `pandas` native functionality breaks. Fix in #3029.

PR #2382, opened 2022-04-03
[BUG] Fixed fit method of Imputer

aiwalter:
#### Reference Issues/PRs
closes #2267 and provides alternative solution to PR #2269

Mover methods that actually fit sth into `_fit`.


fkiraly:
@aiwalter, I get the idea, but I think this approach has some significnat drawbacks:

* not downwards compatible. So this would need deprecation, since it changes the behaviours that users know, for the `mean` etc options. Over and directly after the deprecation period, it would cause user frustration.
* it does not provide an option where transformation parameters are compited in `transform`, which is a sensible setting for time series classification - that option would be removed.

aiwalter:
> it does not provide an option where transformation parameters are compited in transform, which is a sensible setting for time series classification - that option would be removed.

In this case you would just do `fit_transform` right? So its like you fit on the transform data

> not downwards compatible. So this would need deprecation, since it changes the behaviours that users know, for the mean etc options. Over and directly after the deprecation period, it would cause user frustration.

Well the interface is downwards compatible. Not sure if we need to deprecate also things that are not interface related. And also we could deprecated it if we decide to do so, it wont be a stopper for this PR

fkiraly:
> In this case you would just do `fit_transform` right? So its like you fit on the transform data

You can not "just do" `fit_tranform` if it is part of a pipeline, it does whatever it is called with, and that is `fit `in `fit` and `transform` in `transform`.

> Well the interface is downwards compatible. 

No, it´s not, since you get different, potentially breaking behaviour.

Anything that can break user code - interfae *and* functionality change - should be deprecated.

aiwalter:
about which pipeline are you talking? then it would be more clear to understand this. Well then let me make a separate PR to add deprecation, this PR here would then go in after next release

fkiraly:
> about which pipeline are you talking? 

Any, in which transformers are being used. For instance the `TransformedTargetForecaster` that you wrote.

> Well then let me make a separate PR to add deprecation, this PR here would then go in after next release

I disagree with the end state here, sorry.
I think we should not change the functionality of these keys at all but add new keys.

I suggest a core dev discussion with cons/pros spelled out as we usually do in such a case.

fkiraly (changes requested):
Could you kindly adopt some things from https://github.com/alan-turing-institute/sktime/pull/2382:

* the fix to `ForecastingGridSearchCV` which currently looks for an undefined imputation method
* docstring clarifications for the methods, especially `random` and `forecaster`
* input check in `__init__` and `fit`
* rename of returned `X` to `Xt`
* move read of `self.method` to the start of `_check_method`
* `get_test_params` should check all methods, not just one

Additional request:

* `__init__` should set `fit_is_empty` to `False` or to `True`, dependent on `method`

aiwalter:
> Could you kindly adopt some things from https://github.com/alan-turing-institute/sktime/pull/2382:

This is 2382? Did you do a typo?

fkiraly:
ah yes, I meant #2269 (my alternative PR, the parts that do not touch the interface and design but things like docstrings etc)

aiwalter:
> __init__ should set fit_is_empty to False or to True, dependent on method

I dont get this, bcz the `fit` is not empty, so why should the tag say so in some cases?

fkiraly:
> I dont get this, bcz the fit is not empty, so why should the tag say so in some cases?

Because depending on the parameter, `fit` actually does nothing?

aiwalter:
closes #2696

fkiraly (approved):
Excellent, thanks!

PR #2234, opened 2022-03-16
[ENH] Proba metric grid search integration

eenticott-shell:
#### Reference Issues/PRs
Addresses #2233. Branches from #2232.

#### What does this implement/fix? Explain your changes.
Initial plan for grid search implementation.

#### Does your contribution introduce a new dependency? If yes, which one?
No

PR #2383, opened 2022-04-03
[ENH] New prob metrics

eenticott-shell:
#### Reference Issues/PRs
Expands on #2234

Improves prob metric tests and adds interval metrics.

No

#### Any other comments?
Also refactors prob metric class, changes responsibility of dealing with score_average and multioutput from `_evaluate` to `evaluate`


eenticott-shell:
Draft as I need to lint and finish implementing decided changes.

fkiraly:
Great, thanks!
Linting is failing, but only minor stuff.

eenticott-shell:
Reduced scope of PR, rework of tests will be done in separate PR. This is now only for adding the interval metrics and reworking how multioutputs/scores are handled.

fkiraly:
Can you kindly summarize the nature of your changes in the PR first post?

I would also appreciate advice and comments on this PR:
https://github.com/alan-turing-institute/sktime/pull/2502

we should make sure that the final interfaces are type consistent:
https://github.com/alan-turing-institute/sktime/pull/2502

eenticott-shell:
I found the cause of the unexpected quantile multiindex output. When using the `ColumnEnsembleForecaster` it adds another level to the multindex with the variable name so you end up with a 3 level multindex rather than 2. I don't think this is a bug as this would be the expected output of a `ColumnEnsembleForecaster` prediction. But it does cause the output to be incompatible with my metrics. The solution is to run `df.droplevel(level=1)` but I don't know if this should be done inside or outside of the `ColumnEnsembleForecaster`, what are your thoughts @fkiraly?

fkiraly:
@eenticott-shell, thanks for tracking this down - could you be so kind and be precise which functions output a 3-level multiindex, and which functions expect a 2-level multiindex? And where the clash comes from, i.e., when substituting what precisely in which function?

Re metrics, I thought the public boilerplate part was taking care of the multiindices, converting them either way to whatever is internally used.

fkiraly:
Issue with the `ColumnEnsembleForecaster` was local to the forecaster and has been resolved here: https://github.com/alan-turing-institute/sktime/pull/2512

Is this ready for review again?

eenticott-shell:
@fkiraly yes this is ready for review again

eenticott:
Or maybe not, for some reason merging it with main has caused the tests to fail.

fkiraly:
> Or maybe not, for some reason merging it with main has caused the tests to fail.

Odd - it seems to fail your own tests though?

fkiraly (changes requested):
Look ok, thanks!

But, can you explain why you are importing `FALSE` and `TRUE` from `pickle`?

For the tags, you should be using `python` native `False`, `True`.

eenticott-shell:
I use native `True`/`False` everywhere apart from the tags. I think I'd seen tags done with the pickle `TRUE`/`FALSE` somewhere else and assumed it was required. I will change this

fkiraly:
> I think I'd seen tags done with the pickle `TRUE`/`FALSE` somewhere else and assumed it was required. 

I didn't find any other occurrence - where was that?

fkiraly (approved):
looks ok for a start.

Since this is work in progress, I'd kindly ask you to adhere to variable naming conventions at some case: only classes should be `UpperCamelCase` variable names (except `X` etc for data) should be `lower_snake_case`.

eenticott-shell:
I will update naming conventions in #2683. Can we merge this one now? I don't have the option

PR #2384, opened 2022-04-03
relax name rules for multiindex

ltsaprounis:
Relaxing `pd-multiindex` specification for level names (`instances` and `timepoints`)
The only rule should be that the time index is the last level.
This will make the `pd-multiindex` equivalent to `Hierarchical` with one hierarchy level.


ltsaprounis:
@fkiraly - tests passed! shall we?

fkiraly (approved):
Ah yes - looks like an oversight.
Thanks!

We can merge once (of ... if) the tests pass.

PR #2389, opened 2022-04-04
[ENH] upgrade codecov uploader and cleanup coverage reporting

tarpas:

* addresses the deprecation of Codecov Bash Uploader: https://docs.codecov.com/docs/about-the-codecov-bash-uploader#deprecation-notice-and-schedule
* fixes the partial misconfiguration of codecov uploader https://github.com/alan-turing-institute/sktime/runs/5817665318?check_suite_focus=true#step:9:37
* fixes covergepy warnings: https://github.com/alan-turing-institute/sktime/runs/5807142605?check_suite_focus=true#step:7:12
* `make test` by default doesn't measure coverage but allows it via PYTESTOPTIONS variable
* uses `pytest -n auto`, which seems to work reliably. GHA MacOS runners have 3 core CPUs available, so that should speed up these builds

Check if code coverage got uploaded as it should. See if there are unwanted differences in the build or tests in general.

##### For all contributions

##### For new estimators

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,7 +3,6 @@ show_missing = True
 
 [run]
 source = sktime
-include = */sktime/*
 omit =
     */setup.py
     */sktime/__init__.py
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -88,12 +88,9 @@ jobs:
           cp setup.cfg testdir/
           cd testdir/
           python -m pytest --showlocals --durations=10 --cov-report=xml --cov=sktime -v -n 2 --pyargs sktime
-      - name: Display coverage report
-        run: ls -l ./testdir/
+
       - name: Publish code coverage
-        uses: codecov/codecov-action@v1
-        with:
-          file: ./testdir/coverage.xml
+        uses: codecov/codecov-action@v2
 
   test-linux:
     needs: code-quality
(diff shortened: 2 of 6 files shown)


tarpas:
I'm working on adding https://github.com/tarpas/pytest-testmon to sktime. In my tests I had great results with it, congrats for the nice test suite! 
This PR is a pre-requisite. testmon doesn't support --pyargs :/ 
Adding testmon will come in a second PR. I have limited capacity to work on this so I wanted to let you know about my work, so that we don't duplicate effort .
@fkiraly @lmmentel

tarpas:
> Previously we've had issues with pytest picking up source from `./sktime` rather than from the `sktime` package installed at `PYTHONPATH`. That was when we had compiled extensions present but I'm not sure if we don't need to a bit of caution here.
Hi Lukasz, I saw you did the testdir, that's why I notified you. You're right, ./sktime and --pyargs sktime are different. ./sktime picks up around 500 tests more than --pyargs sktime. I admit in CI it makes more sense to test with --pyargs. 
I'm trying to fix this PR so that it uses --pyargs. 
But unfortunately testmon doesn't support --pyargs yet. I was testing with `pytest --testmon ./sktime` and it was doing great. It was selecting very small subset of tests and hitting the same test failures as the original build without testmon. If no new problems arise, do you think you could live with testing ./sktime until testmon supports it? As said the build results were equivalent. @lmmentel @fkiraly 

It would be beneficial for local development experience to actually fix the tests so that ./sktime and --pyargs sktime pick the same tests. 

Also some base import always takes 2 minutes, whenever you need to execute even just one simplest test. I assume it's a bug and could be fixed with not much effort.

tarpas:
@lmmentel @fkiraly @TonyBagnall could you please have a look? The changes are fairly straightforward and seem to work fine.

lmmentel (approved):
:+1:

lmmentel:
Thanks @tarpas!

PR #2392, opened 2022-04-05
[ENH] Test SingleWindowSplitter with Timedelta forecasting horizon

khrapovs:
#### Reference Issues/PRs

Next step in #1737 

- Test SingleWindowSplitter with Timedelta forecasting horizon. That is, add `TEST_FHS_TIMEDELTA` to tests cases in `sktime.forecasting.model_selection.tests.test_split.test_single_window_splitter` and in `sktime.forecasting.model_selection.tests.test_split.test_single_window_splitter_default_window_length`.

`_get_end` function now returns the value which can be directly used with either `.loc` or `iloc` indexing. For that the value is `-2` relative to the original version. This change makes the function more intuitive. `_get_end` function is used in:
- `sktime.forecasting.model_selection._split.BaseWindowSplitter._split`. I added 2 to offset the change.
- `sktime.forecasting.model_selection._split.BaseWindowSplitter.get_cutoffs`. I added 2 to offset the change.
- `sktime.forecasting.model_selection._split.SingleWindowSplitter._split`. I added 1 directly where it is used (now `end = _get_end(y, fh)`) and one more 1 in calculation of integer `start` (now `start = end - window_length + 1`).
- `sktime.forecasting.model_selection._split.SingleWindowSplitter.get_cutoffs`. I have added 2 (now `end = _get_end(y, fh)`).

No external changes were made. Tests are the same plus added more test cases for Timedelta forecasting horizons.

No.

##### For all contributions


sktime/forecasting/model_selection/_split.py
```diff
@@ -186,12 +190,24 @@ def _get_end(y: ACCEPTED_Y_TYPES, fh: ForecastingHorizon) -> int:
     # For purely in-sample forecasting horizons, the last split point is the end of the
     # training data.
     if fh.is_all_in_sample():
-        end = n_timepoints + 1
+        if array_is_int(fh):
+            end = n_timepoints - 1
```
fkiraly, line 194: wait, why is it "minus 1" here if it is "plus one" in the old version??

sktime/forecasting/model_selection/_split.py
```diff
@@ -186,12 +190,24 @@ def _get_end(y: ACCEPTED_Y_TYPES, fh: ForecastingHorizon) -> int:
...
+                end = y.get_loc(y[-1])
+            else:
+                end = y.index.get_loc(y.index[-1])
 
     # Otherwise, the last point must ensure that the last horizon is within the data.
     else:
         fh_max = fh[-1]
-        end = n_timepoints - fh_max + 1
+        if is_int(fh_max):
+            end = n_timepoints - fh_max - 1
```
fkiraly, line 205: same

fkiraly (changes requested):
quick, possibly stupid question: there seems to be an offset of 2 between "old version with integer indices" and "new version in the case of integer indices". Where is that coming from?

sktime/forecasting/model_selection/_split.py
khrapovs: I have been struggling with this part a lot. The "old version" was totally confusing to me. So I have changed to to what you see now with the following comment in the docstring:
```
For a time series `y`, `y.iloc[end]` will give
the last value of the training window,
while `y.index[end]` will provide the last index of the training window.
```
Let me know if you disagree, that this interpretation/intention of `_get_end` is much more understandable.

In addition, of course, I have compensated this change in `_get_end` output by adding 2 accordingly downstream.

fkiraly (changes requested):
Ok, I see. There seems to have been a subtantial interface change, at least internally.

Are the external interfaces still respected?
Or do we also have an external interface change that would need deprecation?

This is very hard to review due to the "plus 2" change.
Can you please write more details in the PR summary about the end state and the nature of the intenal and external changes in the splitters?

sktime/forecasting/model_selection/_split.py
fkiraly: why "plus two", though, can you please explain that? Still don't understand.

khrapovs: The original formula was `end = n_timepoints + 1`. Now it is `end = n_timepoints - 1`. The difference is 2. Let's reverse the question: do you understand the original formula? I do not.

sktime/forecasting/model_selection/_split.py
khrapovs: see above

khrapovs:
> Ok, I see. There seems to have been a subtantial interface change, at least internally.

In my opinion to the best.

> Are the external interfaces still respected? Or do we also have an external interface change that would need deprecation?

If you take a look at the diff between this branch and master, you would see that absolutely nothing has changed in the signatures of the existing functions and classes. No deprecations, nothing. `SingleWindowSplitter` is adjusted to accept Timedelta forecasting horizons and I have added tests for that.

> This is very hard to review due to the "plus 2" change. Can you please write more details in the PR summary about the end state and the nature of the intenal and external changes in the splitters?

`_get_end` function is used in:
- `sktime.forecasting.model_selection._split.BaseWindowSplitter._split`. I added 2 to offset the change.
- `sktime.forecasting.model_selection._split.BaseWindowSplitter.get_cutoffs`. I added 2 to offset the change.
- `sktime.forecasting.model_selection._split.SingleWindowSplitter._split`. I added 1 directly where it is used (now `end = _get_end(y, fh)`) and one more 1 in calculation of integer `start` (now `start = end - window_length + 1`).
- `sktime.forecasting.model_selection._split.SingleWindowSplitter.get_cutoffs`. I have added 2 (now `end = _get_end(y, fh)`).

No external changes were made. Tests are the same plus added more test cases for Timedelta forecasting horizons.

fkiraly (changes requested):
@khrapovs, the interface changes look good!
And, I want to say, in general your work has been great and is much appreciated!

Now, to the content discussion: we can't just change a "plus 1" to a "minus 1" and hope for the best (this is how it looks from my perspective, maybe it's different from yours). I need to understand what is happening here, for the review.

> The original formula was `end = n_timepoints + 1`. Now it is `end = n_timepoints - 1`. The difference is 2. Let's reverse the question: do you understand the original formula? I do not.

I need to understand this better before I can approve.
I also do not understand the "plus one" here either, but at least you need to be able to explain, may I ask you to do this:

* why do you change it to "minus one"? I.e., what makes you think "minus one" is correct and "plus one" is wrong, or at least worse?
* how are we sure that we are either not changing any behaviour, or correcting behaviour that was previously wrong? This requires explanation.

fkiraly:
> If you take a look at the diff between this branch and master, you would see that absolutely nothing has changed in the signatures of the existing functions and classes.

I know - what worries me is that we have a change in logic that is not captured by tests for consistent signature. I.e., we are possibly making a change in the math that turns "right" into "wrong" or "wrong" into "right". Do we know whether there is a change, and if yes, which way it goes?

khrapovs:
> Now, to the content discussion: we can't just change a "plus 1" to a "minus 1" and hope for the best (this is how it looks from my perspective, maybe it's different from yours). I need to understand what is happening here, for the review.

What do you mean "hope for the best"? There are 2556 test cases. Try to change `-1` to `0` or `+1` and you will see if it is for the best ;)

> I need to understand this better before I can approve. I also do not understand the "plus one" here either, but at least you need to be able to explain, may I ask you to do this:
> * why do you change it to "minus one"? I.e., what makes you think "minus one" is correct and "plus one" is wrong, or at least worse?

I have rewritten the docstring of `_get_end` once again in an attempt to make it even more clear what I want from this function.

> * how are we sure that we are either not changing any behaviour, or correcting behaviour that was previously wrong? This requires explanation.

Previous behavior was not wrong, it was not intuitive. I (hopefully) made the output of `_get_end` much more intuitive since now it can be interpreted in terms of standard `.loc` or `.iloc` indexing. To preserve the original behavior I have offset the change in every place where the function was used. Correctness of the change is insured by thousands of test cases that break if I change the output of `_get_end`.

fkiraly:
> What do you mean "hope for the best"? There are 2556 test cases. Try to change -1 to 0 or +1 and you will see if it is for the best ;)

You're a little like my quantum physics professor, when he thought someone asked a question of which the answer should be obvious, he would glare menacingly and/or indicate that he would hit you on the head with one of his heavy Landau books.

That may or may not be a didactial approach, but for a PR review you need to explain what you are doing and why - I review a lot of PR a day and also do other stuff, so it would be nice if you could make my life easier here by telling me what explains the change in the plusminus 1.

Pretty please?

khrapovs:
@fkiraly 

I think I have exhausted my arguments, but let me try again.

There is a function `_get_end` which was returning something I could not understand. To make it intuitive I have adjusted the output by `-2`. To explain the output of the function I have added a few sentences in the docstring. In short, the output now can be used in `.iloc` directly without subtracting anything, as in the master version. In every place, where the function was used I have added `+2` to offset the change in the output. Since this function is not tested directly, but only implicitly through splitter outputs, I have tried to pass the tests with `-1` or `0` adjustment in the `_get_end` function. I did not count exactly, but most of the tests fail, hence I did not change the function with a hope that it will miraculously work out.

Please, let me know if it is still not clear, or I am answering not the question you are asking.
Thanks.

khrapovs:
I am an economist by training. In my profession Mas-Colell would probably be an adequate replacement for Landau in this case :D

fkiraly:
> Please, let me know if it is still not clear, or I am answering not the question you are asking.

Yes, this is clear, exactly whhat I was looking for, thanks!
Let me take some time to go through the code with that information, if you don't mind.

fkiraly:
> I am an economist by training. In my profession Mas-Colell would probably be an adequate replacement for Landau in this case

One of the many advantages of virtual collaboration is that you can't hit me with a book that easily.

fkiraly (approved):
Alright, makes sense to me now. Thanks for the explanation.

khrapovs:
Thanks a lot!

PR #2394, opened 2022-04-05
[ENH] replace `np.arange` by `np.arghwere` in splitters to enable time based indexing and selection

khrapovs:
#### Reference Issues/PRs

Next step in #1737

Replace `np.arange` usage in cross-validation splitters with `np.argwhere`. This potentially allows passing non-equally spaced time series.

In addition I could not resist not to clean up a bit `sktime.forecasting.model_selection._split.BaseWindowSplitter._split`. There I have extracted a separate method `sktime.forecasting.model_selection._split.BaseWindowSplitter._split_for_initial_window` to make code a bit more readable.

No

##### For all contributions


fkiraly:
This looks good to me.
Before we merge, would you kindly be able to look at `get_window` in `datatypes` - is that a duplication, is it useful, or not?

khrapovs:
> This looks good to me. Before we merge, would you kindly be able to look at `get_window` in `datatypes` - is that a duplication, is it useful, or not?

`sktime.datatypes._utilities.get_window` definitely has a flavor of train/test split. It is used only in `sktime.forecasting.stream._update.UpdateRefitsEvery._update`. Could be an interesting/challenging task to replace it with some intermediate function from splitters.

fkiraly:
> sktime.datatypes._utilities.get_window definitely has a flavor of train/test split. It is used only in sktime.forecasting.stream._update.UpdateRefitsEvery._update. Could be an interesting/challenging task to replace it with some intermediate function from splitters.

So, question: `get_window` afaik works on the panel and hierarchical data types as well.
Do the splitters also work on hierarchical data?

khrapovs:
Honestly, I don't know. Is it relevant for the current PR?

fkiraly:
> Honestly, I don't know. Is it relevant for the current PR?

Yes, due to potential duplication in a core concern, for the PR.
For the start, I just want to understand whether there is a duplication (it seems yes?), what the extent of is, and how easy it is to resolve. If we conclude it's easy and there is substantial duplication, it should be resolved. If it is substantial but not easy, we can merge but should open an issue that summarizes the refactor entry points. If it is not substantial, then the topic would turn out as not relevant and we can merge as well.

khrapovs:
> > Honestly, I don't know. Is it relevant for the current PR?
> 
> Yes, due to potential duplication in a core concern, for the PR. For the start, I just want to understand whether there is a duplication (it seems yes?), what the extent of is, and how easy it is to resolve. If we conclude it's easy and there is substantial duplication, it should be resolved. If it is substantial but not easy, we can merge but should open an issue that summarizes the refactor entry points. If it is not substantial, then the topic would turn out as not relevant and we can merge as well.

IMHO, the effort is substantial and has nothing to do with this particular PR. Above mentioned function and splitters existed long before I started working on spliiter generalization. If there is a side effect of what I do in terms of generalization, I would be happy to use it to address the issue I just created: https://github.com/alan-turing-institute/sktime/issues/2472. Otherwise, please, let me focus on my main target ;)

fkiraly:
ok, could you kindly explain why you think the refactor effort would be substantial?

khrapovs:
> ok, could you kindly explain why you think the refactor effort would be substantial? I.e., using `get_window`, compared to adding part of the existing functionality again in your separate extension?

It will be substantial for me personally since I have not seen this part of the code (`get_window` and the class that uses it). Even from description and signature I can see that this function does not directly map to anything existing in `_split.py`. The only similarity I see is that it does some kind of splitting. And of course, the fact that it handles panels, and existing splitters probably not, makes it even more complicated.

Once again, IMHO this specific PR has nothing to do with `get_window`. Please, correct me if I am wrong.

sktime/forecasting/model_selection/_split.py
```diff
@@ -521,6 +522,17 @@ def get_fh(self) -> ForecastingHorizon:
...
+    @staticmethod
+    def _get_train_window(
+        y: ACCEPTED_Y_TYPES, train_start: int, split_point: int
+    ) -> np.ndarray:
+        if (split_point <= len(y)) & (split_point > 0):
+            return np.argwhere(
+                (y >= y[max(train_start, 0)]) & (y <= y[min(split_point, len(y)) - 1])
+            ).flatten()
+        else:
+            return np.array([], dtype=np.int)
```
fkiraly, line 534: is this correct? This does not look right:

* should it not be empty only if the split point is in the past of the `train_start`, otherwise non-empty?
* it also seems this is correct only for an index? Namely, if `y` is not an index, you'll slice to the series where the *values* are between the starting and end values, not the *index* being between the start/end index!
* this also looks like sth that could 1:1 be replaced by `get_window`, if applied to index
* given that it seems incorrect and identical to `get_window`, consider replacing?

fkiraly (changes requested):
Sorry for the delay in my response.

I am worried about the correctness of the methods added, especially `_get_train_window`.

I also find it a hard time to review this, since there are no docstring whatsoever. I know this is not your fault, but it does not help me as a reviewer - in fact, it is downright confusing.

Hence, would you mind adding docstrings with input/output description (to the methods changed, especially the private ones)? And, could you kindly point out what the change would be to the previous specification?

My main issue is with the `y` - is this an index only, or can this also be the time series? What are we splitting here, only the index, or also the time series? It looks a bit ambiguous from the code, and the devil is in the details...

Sorry to ask for this, but I feel otherwise I can't reliably review this.

sktime/forecasting/model_selection/_split.py
fkiraly: (depends on the specification of course, so I may be wrong - docstrings would be appreciated)

khrapovs: - changed to
```
if split_point > max(0, train_start):
```
- added docstring. `y` is always an index here due to uniform usage of `_check_y` function across all splitters.
- `get_window` is not applicable here. This method takes start/end indices, not window length.

khrapovs:
@fkiraly 

- added dosctring to `_split_for_initial_window` and `_get_train_window`.
- The main change is summarized in the PR description: "Replace `np.arange` usage in cross-validation splitters with `np.argwhere`. This potentially allows passing non-equally spaced time series." This is the only major change in this PR. The rest is just cosmetics.

sktime/forecasting/model_selection/_split.py
```diff
@@ -521,6 +522,35 @@ def get_fh(self) -> ForecastingHorizon:
...
+    def _get_train_window(
+        y: ACCEPTED_Y_TYPES, train_start: int, split_point: int
+    ) -> np.ndarray:
+        """Get train window.
+
+        For formal definition of the train window see docstring of the `BaseSplitter`
+
+        Parameters
+        ----------
+        y : pd.Index
```
fkiraly, line 535: ok, thanks!

Now here is what confused me. The code looks like it's written for `pd.Index`.
But, the type hint says that `y` can be `ACCEPTED_Y_TYPES = Union[pd.Series, pd.DataFrame, np.ndarray, pd.Index]`.
How does this fit together, and with the docstring? Is it correct that we also want to be able to pass all this stuff, or do we really only want an index?

sktime/forecasting/model_selection/_split.py
```diff
@@ -702,6 +707,49 @@ def _split(self, y: Optional[ACCEPTED_Y_TYPES]) -> SPLIT_GENERATOR_TYPE:
...
             yield train, test
 
+    def _split_for_initial_window(
+        self, y: Optional[ACCEPTED_Y_TYPES]
+    ) -> SPLIT_ARRAY_TYPE:
+        """Get train/test splits for non-empty initial window.
+
+        Parameters
+        ----------
+        y : pd.Index
```
fkiraly, line 717: same as above.

fkiraly (changes requested):
Thanks for the docstrings!
This is becoming a lot clearer.

I am now able to localize my concern/confusion, see above. Is `y` only an index? Or can it be sth else, incl the full series?

sktime/forecasting/model_selection/_split.py
fkiraly: > This method takes start/end indices, not window length.

end = start + length?

The real "burden" is with treating the multiple possible data format types, not in the selection, is it?

khrapovs: Look, I will not be using this function. At least not in this PR. It operates with window distances, while `_get_train_window` operates with exact points in the index. Distances are computed outside. Just leave it, please.

sktime/forecasting/model_selection/_split.py
khrapovs: I went through all functions/methods that use `y` and replaced type suggestions and corrected docstrings.

sktime/forecasting/model_selection/_split.py
fkiraly: ok, sure

sktime/forecasting/model_selection/_split.py
fkiraly: ah, that makes more sense!

So if `y` is not an index, it gets the index from `y`?

khrapovs: I have renamed `_check_y` to more clear `_get_index_of_time_series`. This function is used in every public method of splitters before going to any internal processing.

fkiraly: I see. But *this* is a straightforward duplication then! Namely, of `datatypes.utilities.get_time_index`.

khrapovs: this change is straightforward. replaced `_get_index_of_time_series` with `sktime.datatypes._utilities.get_index_for_series`

fkiraly: ok, I think with me now better understanding what is going on (thanks for clarifying and docstrings!), and with the duplication removed, I think we are good now! Give me some time to think about it still, but I feel we're ready.

fkiraly (approved):
Ok, I think this is good now. Thanks for all the hard work (in this hard-to-navigate part of sktime)!

PR #2395, opened 2022-04-05
[ENH] `NaiveVariance`: verbose arg and extended docstring

fkiraly:
This small PR adds two things to `NaiveVariance`:

* a `verbose` arg which is by default `False`, suppressing the many warnings that are raised on toy data
* an explanation in the docstring which methods are added and how the wrapped estimator relates to the composite


fkiraly:
rapid merge as discussed with CC

PR #2404, opened 2022-04-07
[ENH] post-processing in `TransformedTargetForecaster`, dunder method for (transformed `y`) forecasting pipelines

fkiraly:
This PR:
* adds the capability to `TransformedTargetForecaster` to postprocess the forecasts with transformers.
* adds dunder `*` feature to construct forecasting pipelines with automatic name generation, similar to classifier and transformer pipeline

This allows, for instance, reconciliation of results, and quick constructions of reconcilers, e.g., `aggregate * my_forecaster * my_reconciler`.

In detail makes the following changes to `TransformedTargetForecaster` and `BaseForecaster`:
* in line with `ClassifierPipeline` and `TransformerPipeline`, allows list of estimators as input, and name/estimator tuples with non-unique names. In either case, unique names are automatically generated.
* adds dunder method `*` to forecasters via `BaseForecaster`. Any forecaster, when multiplied with transformers, creates a non-nested `TransformedTargetForecaster`. Earlier transformers preprocess and inverse transform; later transformers postprocess.

Relies on PR:
* https://github.com/alan-turing-institute/sktime/pull/2406


aiwalter:
would be nice to have an example in the docstring

fkiraly:
> would be nice to have an example in the docstring

There is now :-)

Issue #2357, opened 2022-03-31
[ENH] statsmodels AutoReg and SARIMAX


fkiraly:
We should interface `AutoReg` and `SARIMAX` from `statsmodels` as forecasters.

- [x] `SARIMAX`
- [ ] `AutoReg`

SARIMAX is currently not available in `sktime`, so having it would be nice.

`AutoReg` would be good to use in `sktime` `VAR` (which interfaces `statsmodels` `Var`) in a case where the `y` passed is univariate.
Currently `VAR` erros out when a univariate `y` is passed; mathematically, AR is the univariate special case of VAR, so it would be natural to use the AR implementation from the same package instead of returning an error.

Should be an easy, good first issue task with these instructions:
https://www.sktime.org/en/stable/developer_guide/add_estimators.html

For `AutoReg`, I would recommend implementation as a separate estimator; in a second step, we can make `Var` default to it for univariate `y`.

biologioholic:
😄 @fkiraly Looks interesting! Sure, I will take on this work

ryali1:
I'd like to get started on adding SARIMAX functionality to sktime if you haven't already worked on that part @biologioholic! Please let me know, and I can get started right away.

biologioholic:
Hello @ryali1, I'm already half way through Sarimax. Debugging atm

ryali1:
Hi @biologioholic, that sounds good, appreciate your response! Have you worked on AutoReg already? If not, maybe I could get started on that. I'm still familiarizing myself with contributing to sktime, so maybe we could collaborate on this, or I could come to you with questions about how to properly test/debug the code I write with the sktime framework?

biologioholic:
Sure, you can go ahead and do the AutoReg. You can look at sktime/forecasting for some interface examples

https://github.com/alan-turing-institute/sktime/tree/main/sktime/forecasting

ryali1:
Great! I'll get started on the AutoReg functionality. I'll start with creating a separate estimator as @fkiraly mentioned and will get feedback on it before working on step 2 (getting VAR to default to it).

fkiraly:
excellent, thanks!

TNTran92:
@fkiraly SARIMAX.py is in testing phase. I have test_sarimax.py in  sktime/forecasting/tests and it works locally. How can I include it in pytest so that it becomes a part of check_estimator?

TNTran92:
@fkiraly Draft PR submitted 😊

[...]

PR #2407, opened 2022-04-07
[ENH] Added AutoReg Interface for sktime

ryali1:
#### Reference Issues/PRs
Reference Issue  #2357  

- Implemented AutoReg in sktime/forecasting
- Implemented test_autoreg in sktime/forecasting/tests

No, there are no new dependencies.

- Ran simple tests to ensure sktime AutoReg agrees with results of statsmodel AutoReg, but need to test with exogenous inputs 

- I didn't run this through the pytest suite, since I wasn't sure how to. Any relevant feedback or guidance on this will be helpful!

#### Any other comments?

- This is my first PR so any relevant feedback or comments are appreciated!

##### For all contributions
- [ *] I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/main/.all-contributorsrc).
- [ *] Optionally, I've updated sktime's [CODEOWNERS](https://github.com/alan-turing-institute/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
- [* ] I've added unit tests and made sure they pass locally. (I ran this through a test script I created)
- [* ] The PR title starts with either [ENH], [DOC] or [BUG] indicating wether the PR topic is related to enhancement, documentation or bug

##### For new estimators


fkiraly:
> I didn't run this through the pytest suite, since I wasn't sure how to. Any relevant feedback or guidance on this will be helpful!

see here: https://www.sktime.org/en/stable/developer_guide/add_estimators.html#how-to-test-interface-conformance

The easiest way might be to use the `check_estimator` utility and run it. `pytest` with `-k` arg is another option to run tests specifically for the new estimator.

I've also approved a round of CI tests above - click on "details" for any failing test to see what's wrong.
In this case, linting (code formatting) tests fail, these need to be resolved first before the others will run here on remote. It's mainly things such as newlines, whitespaces, and missing docstrings.

ryali1:
Hi Dr. @fkiraly, I installed pre-commit and made sure all the tests ran locally on my local machine. Hope this works now!

fkiraly:
> Just to confirm, the pytests I run locally through "check_estimator()" would be the same as the CI/CD tests run above right?

Yes, that's usually the case. Sometimes there are more tests, e.g., if you are using an intermediate base class, or add your own tests.

ryali1:
Hi Dr. @fkiraly, the test: "test_predict_time_index_with_X" had a number of scenarios which failed when I used the default predict statsmodels _predict function. 

While trying to debug this, I found that line 81: "y_pred = self._fitted_forecaster.predict(start=start, end=end, exog=X)" was being called during each of the test runs, which made me think that there may have been something wrong in the way I defined the exogenous array parameter. I tried to redefine the exogenous variable in AutoReg's init function in every possible way, but I found that each of the same tests were still failing. 

This is why I made a slight change to the _predict function to only use the "start" and "end" inputs instead of the "exog=X" input. By making this change, all of the test scenarios ran successfully. 

I may definitely be missing something here since I'm still getting familiarized with the forecasting interface and working with adaptors, so please let me know if you think I should try anything else, or if this change in the _predict function was the right thing to do! 

I've included 3 screenshots below: (1) The error messages I get when running the default _predict function, (2) the default _predict function, and (3) the changed _predict function that the current AutoReg implementation is calling to get successful test results.

FailedTests
StatsModelsDefault_Predict
NewAutoReg_Predict

fkiraly:
@ryali1, sorry if I'm confused, but it doesn't look like any of your tests are failing?

fkiraly (changes requested):
Thanks, this looks good!

One thing worries me, namely the treatment of the exogeneous `X`. As of now, it is not used as an argument of `_predict`. But, in accordance with the `sktime` interface, the "future" part of exogeneous data may only be passed there. Is that not a problem?

ryali1:
Dr @fkiraly, the tests were failing when I used the default _predict method found in the statsmodels adaptor. The code I committed uses an editted _predict function that doesn't include exog (which is why I think all the tests are passing). I didn't explicitly test for exog, so it might not work with exogenous inputs yet. I'll dig deeper today to see how to get this implementation to work with exog inputs.

fkiraly:
>I'll dig deeper today to see how to get this implementation to work with exog inputs.

That would be great! Let´s try to get it to a state where we can also forecast with exogeneous input.

fkiraly:
are you still working on this?

If you do not know how to get exogeneous input to work properly, turn the "exogeneous" tag off please and write in the issue that this still needs to be done.

ryali1:
Hi Dr. @fkiraly, my apologies on the delay on this. I'm fully recovered from COVID now and added support for exog variables for in sample and out sample forecasting for autoreg. Please let me know if this looks alright or if there's any other changes you'd like to see on it and I'll work on it right away!

Thanks!

Issue #2368, opened 2022-04-02
[BUG] Seemingly sporadic test failure on Mac Python 3.7


khrapovs:
**Describe the bug**

Here is the failed run: https://github.com/alan-turing-institute/sktime/runs/5788191259?check_suite_focus=true

```
=================================== FAILURES ===================================
_____________________ benchmarking/tests/test_evaluator.py _____________________
[gw0] darwin -- Python 3.7.12 /Users/runner/hostedtoolcache/Python/3.7.12/x64/bin/python
worker 'gw0' crashed while running 'benchmarking/tests/test_evaluator.py::test_plots'
=============================== warnings summary ===============================
../../../../hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/coverage/inorout.py:472
../../../../hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/coverage/inorout.py:472
../../../../hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/coverage/inorout.py:472
../../../../hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/coverage/inorout.py:472
  /Users/runner/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/coverage/inorout.py:472: CoverageWarning: --include is ignored because --source is set (include-ignored)
    self.warn("--include is ignored because --source is set", slug="include-ignored")

benchmarking/tests/test_experiments.py::test_run_clustering_experiment
  /Users/runner/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/tslearn/utils/utils.py:156: DeprecationWarning: `np.float` is a deprecated alias for the builtin `float`. To silence this warning, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here.
  Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
    def to_time_series_dataset(dataset, dtype=numpy.float):

benchmarking/tests/test_experiments.py::test_run_clustering_experiment
[...]

PR #2410, opened 2022-04-08
[ENH] add new argument return_tags to all_estimators

### LINKED ISSUE #2387: [ENH] list tag values in `all_estimators` return table
`all_estimators` currently returns tuples/columns (classname, class), but it would be useful to users, e.g., for listing, or for display in the sktime docs, if additional columns return values of certain tags.

I would propose to add an additional `return_tags` argument which can be a str or list ot str of tag names (see `all_tags` for valid tags). Default None which retrieves no tags in addition.

The utility would then return the value of those tags in a tuple (classname, class, tagvalue1, tagvalue2, etc), with tagvalue being empty (None) value if the estimator does not have the tag.

miraep8:
#### Reference Issues/PRs

Closes #2387

I have added a new argument 'return_tags' to the registry function all_estimators().  'return_tags' is an optional argument, either a str or list of str of tag names the user wants returned for all estimators returned.  If provided all_estimators will return each estimator's value for each tag in 'return_tags' or else None if the estimator doesn't have that tag.  I also added a new test to ensure all_estimators works if 'return_tags' is passed as a str or list of str. (the default - None is tested in other tests!)

No

Does this address the issue suggested?  Any test cases I should add?  Any other feedback?  (This is my first Pull Request so please let me know if I should be doing something differently!) 

#### Any other comments?

Thank you for reviewing my PR!  Excited to be making my first contribution to sktime, hope this small change is useful! :)

##### For all contributions


sktime/registry/_lookup.py
```diff
@@ -77,6 +78,12 @@ def all_estimators(
     as_dataframe: bool, optional (default=False)
                 if False, return is as described below;
                 if True, return is converted into a DataFrame for pretty display
+    return_tags: str or list of str, optional (default=None)
```
fkiraly, line 81: the docstring should say more clearly what happens in the two cases `as_dataframe=True` and `as_dataframe=False`.

sktime/registry/_lookup.py
```diff
@@ -90,6 +97,11 @@ def all_estimators(
             in alphabetical order of class name, where
             ``name`` is the estimator class name as string
             ``class`` is the actual class
+        if return_tags are passed:
```
fkiraly, line 100: It looks like this docstring was incomplete!
Before you worked on it already.

Would you mind fixing this, while you´re on it?
The "returns"  docstring needs to say what the output is, depending on `return_names` and `as_dataframe`. Might be a little fiddly, apologies, but once that´s done it´s more natural to integrate the effect of `return_tags`.

Currently it´s referring to "columns" which have no clear reference frame, because there is no dataframe that has previously been mentioned (it appears only in the case `as_dataframe=True` which is not documented!)

fkiraly (changes requested):
Nice work, very useful contribution for users!

Some change requests:
* the docstring of `all_estimators` is incomplete, and already had problems before you started working on it. Unfortunately complicates things a little, but would you mind fixing this? More details above. I consider this important/blocking, since this could be a user confusion point.
* optional, non-blocking: you wrote input checks for `return_tags`, which coerce to list of str and check str or list of str. That would also be useful for `estimator_types`.
* optional, non-blocking: there is some retrieval functionality in `all_estimators` that could now be replaced by the utility `registry._scitypes.scitype`, specifically the logic in lines starting `def _is_in_estimator_types` up until `if _is_in_estimator_types`. Would appreciate if you could have a look whether this makes sense and would get shorter or less import heavy.

miraep8:
Thank you for the prompt review!  All your suggestions make sense to me and I will update the code later today!   :)

sktime/registry/_lookup.py
miraep8: Hey!  I just updated the docstring.  I hope both the explanation of the new functionality, as well as the fleshed out returns section of the docstring makes it easier for users to understand what is going on.

sktime/registry/_lookup.py
miraep8: Hey, I fleshed out the as_dataframe section of the docstring in more depth.  Let me know if the improvements are satisfactory.

miraep8:
Hey!  Just added some new commits that I hope address the issues you raised. 

1 - I reformatted the docstring a little to (hopefully) address the pre-existing issues. 
2 - I agree!  I broke out the str/list of str check into a little function called ``_check_list_of_str_or_error()``

A few other small changes:

- added some negative tests for ``return_tags`` as well - eg, ensuring that ``all_estimators`` also breaks as expected if passed bad inputs for return_tags
- shortened some of the if statements for improved readability.  eg if ``estimator_types`` is not None -> if ``estimator_types`` 

I looked into the suggestion you made regarding ``registry._scitypes.scitype``.  At first glance - it didn't strike me off hand as a more readable replacement, but I am open to returning to this suggestion again/considering it in more depth (potentially via another pull request, where we could make such replacements more systematically?)

sktime/registry/_lookup.py
fkiraly: yes, excellent! I was wondering how to make the different interacting conditions clear here, introducing the "attributes returned" makes it a lot clearer. Return section is great.

Minor thing: instead of pandas dataframe, write `pandas.DataFrame`

fkiraly (changes requested):
Looks great, and thanks for the docs and the tests!

One thing to fix:
Linting breaks, mostly since it wants some docstrings, and the first line of docstrings formatted in pydocstyle/numpydoc convention.

sktime/registry/_lookup.py
miraep8: In the process of fixing these/style issues now!  Will add a new comment when I am ready for review!

miraep8:
Both pydocstyle and flake8 seem to be happy when I run them locally now!

sktime/registry/_lookup.py
miraep8: Thanks!  Ready for review, though no rush :)

miraep8:
I apologize, pydocstyle is not in my usual workflow, so may need to poke around a bit to figure out why its not complaining locally, but is complaining here.  Just a quick check while I am working on that though - is there a better way to locally the code-quality tests then what I have been doing?  Figured it might help reduce iterations/potentially might be a useful note for other first time contributors?  For reference - been calling black flake8 and pydocstyle on all files I changed.

fkiraly:
> For reference - been calling black flake8 and pydocstyle on all files I changed.

If you are using vscode, you can go to "preferences" and search for black, pydocstyle etc, and enable linting with a checkmark. Then all the issues will show up in the "problem" pane.

Further, you can enable precommits, which will automatically fix some of the issues and complain about others.

Might be worth being more explicit in the developer documentation about this, btw.

miraep8:
Hmmm, seems like the test-windows (3.7) test just timed out/was canceled?  From what I can tell from the details it doesn't necessarily seem like it stalled out on something I added, but not sure why this build would take much longer than other tested.   I will take a look at it more later - but if anyone has any insights/ideas that would be appreciated as well. :)

miraep8:
Can I possibly request that we run the tests again and just see if the time out issue is repeatable?

fkiraly:
@lmmentel or @k1m190r, can you perhaps make any sense of the stuck windows run logs?

miraep8:
Well at least this time around it seems to have cleared the test-windows (3.7) test without any chances to the code! :) 🎉Perhaps it is the randomness that is sometimes introducing different run times.  Not immediately clear why this would only be an issue for windows (3.7) though!

miraep8:
Oh no.  I accidentally made it want to run the workflow again by merging main into the branch.  It did clear all the test though!

miraep8:
I think these issues should be resolved now

fkiraly (approved):
Excellent, thank you!

Let's hope windows 3.7 doesn't hang this time...

miraep8:
windows 3.7 seems to have worked, although - I think GitHub actions must have hit max number of concurrent Mac OS jobs within this repo and it put those jobs on hold. Is there anyway you can assign them to a new runner on your end?

fkiraly:
> windows 3.7 seems to have worked, although - I think GitHub actions must have hit max number of concurrent Mac OS jobs within this repo and it put those jobs on hold. Is there anyway you can assign them to a new runner on your end?

There is a smaller number of macos runners, so we decided to run them only if all other tests have run, to avoid them being blocked by PR that end up breaking all OS.

miraep8:
Makes sense! I was only asking because I checked back a few times after all the other tests had finished and it looked like they were still stuck in the queue, so I was just wondering if there was something else holding them back. :)

lmmentel (approved):

miraep8:
Haha, now seems like the test-mac (3.8) is failing?  The issue seems related to the issues brought up in #2368 ?  Thoughts @fkiraly @khrapovs ? Also - any merit in going ahead and merging this PR? :)

fkiraly:
@lmmentel probably restarted the tests since he noticed it was #2368

lmmentel:
Yup I managed to restart only the failed job for macOS py3.7 and this time it passed :slightly_smiling_face:

miraep8:
Thank you @lmmentel !! :)  Thanks all!

PR #2414, opened 2022-04-08
[BUG] fix soft dependency handling for `esig` imports

fkiraly:
This PR fixes too restrictive soft dependency handling for `esig` imports.

These have been moved into functions/classes so loading is possible without having `esig` installed.


lmmentel (approved):
:+1:

PR #2422, opened 2022-04-09
[MNT] Make the contrib module private

MatthewMiddlehurst:
`contrib` > `_contrib`. Suggestion by @aiwalter in Slack. There are some duplicate files in contrib and the package proper, which makes things a bit confusing.

It is exempt from most tests, and in the words of @fkiraly: "It’s a folder where any content can change without any notice". So it shouldn't really be public facing either way.

Possible issues:
- There may be a few ancient files which get hit by the linting checks
- There are a couple of items which see use in the package proper, i.e. `RotationForest` and `ContinuousIntervalTree`


MatthewMiddlehurst:
`ts_distance_measures.py` requires changes for the code quality test. This file has not been touched since 08/2019, ignore it or get rid of it @TonyBagnall?

TonyBagnall:
unless @jasonlines wants to keep them, I think just remove it. jay is off, so I think delete it and we can put it back if he wants them, but I cant imagine he will

TonyBagnall (approved):

fkiraly:
hmmm - do we want a deprecation message for the old location, and a warning in the new location, at import, that anything in the module may change without deprecation message?

TonyBagnall:
oh sorry, jsut saw your message @fkiraly , I dont think we need deprecation, this is lower than experimental imo!

fkiraly:
fine - as long as you talk to the pitchfork mob of angry users whose private method calls are going to break 😁

PR #2423, opened 2022-04-09
[BUG] fixing proba predict methods of forecasting tuning estimators

fkiraly:
This PR fixes a bug in the grid search estimators, which prevented from calling probabilistic prediction methods directly on the composite.

The bug is resolved by simply patching through all the methods using the existing delegator decorator.


fkiraly:
FYI @eenticott-shell if you are reading this 😁

PR #2424, opened 2022-04-09
[ENH] WindowSummarizer with time windows

GrzegorzRut:
#### Reference Issues/PRs
An attempt to enhance the WindowSummarizer, as discussed in [#2309](https://github.com/alan-turing-institute/sktime/discussions/2309)

##### For all contributions
- [x ] I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/main/.all-contributorsrc).

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1538,7 +1538,16 @@
         "code",
 	"example"
       ]
-    }
+    },
+    {
+      "login": "GrzegorzRut",
+      "name": "Grzegorz Rut",
+      "avatar_url": "https://avatars.githubusercontent.com/u/39675431?v=4",
+      "profile": "https://github.com/GrzegorzRut/",
+      "contributions": [
+        "ideas"
+      ]
+    },
   ],
   "projectName": "sktime",
   "projectOwner": "alan-turing-institute",


GrzegorzRut:
Hi!
This is an approach to summarize data with the datetime index. I think that the main issue with this task is the way `.shift` changes the index values. If, for instance, some dataframe has 'MS' frequency and we'd shift the data by 15 days, we'd normally get only nans:

```
import pandas as pd

y = load_airline()
y = pd.DataFrame(y)

transformed = y.shift(15,'D').rolling('30D').mean().rename(columns= {x:x+'_tr' for x in y.columns})
transformed.merge(y, left_index=True,right_index = True,how='right')
```
Screenshot from 2022-04-10 13-09-16

My work around temporarily changes the frequency of the dataframe (in the above example, it'd be changed to 'D') before `.rolling` is applied (see: `_calc_freq_data ` function). After this is done, the frequency is reverted back to the original one.
It is assumed that the format of lag input will be the same as in the rolling function (either an int or a string like '15D')
This approach has some limitations - I needed to limit the available frequency options to constant ones (days, hours, minutes, etc.)

The update affects the `_fit` method in the `WindowSummarizer` - for the time being I've commented out the `truncate_start` attribute as it's functionality here is limited (not sure if it is very useful after  #2325).  
Also, the `func_dict` had to be slightly altered as well - one step is added to the lag in `_window_feature`, not outside of it. 

Here's a small test:

```
import pandas as pd
y = load_airline()
y = pd.DataFrame(y)

kwargs = {
    "lag_feature": {
        "lag": ['30D'],
        "mean": [['15D','60D']],
    }
}
transformer = WindowSummarizer(**kwargs)
transformer.fit(y)
transformer.transform(pd.DataFrame(y)).head()
```

Screenshot from 2022-04-10 13-30-16

GrzegorzRut:
> NICE, thanks!
> 
> Before I dive into this, just wanted to mention an "obvious to see" issue:
> 
> you probably branched off an earlier version of `summarize.py`, and now it is conflicting with the version on `main`. You'll need to resolve these merge conflicts sooner or later (better sooner) - let us know if you need help with that.

many thanks, i think i was able to solve the conflicts

fkiraly:
alright! Linting is next. Need any help with that?
(click on "details" next to the failed check to see what's going on. Installing `pre-commit` locally will also help you with this in the future, see dev docs)

GrzegorzRut:
> Linting

will ask for help if I fail to do it myself, thank you:-)

fkiraly:
Indeed, linting passes 😄 
Let's see the rest...

GrzegorzRut:
> Indeed, linting passes 😄 Let's see the rest...

Ok, needs more work;-P.

fkiraly:
You can use `check_estimator` to run generic tests on the estimator from your python console, for quick debugging. Let me know if you feel you get stuck and need another pair of eyes.

GrzegorzRut:
> You can use `check_estimator` to run generic tests on the estimator from your python console, for quick debugging. Let me know if you feel you get stuck and need another pair of eyes.

Many thanks, will try to look into it as soon as I can.

GrzegorzRut:
The previous version of  WindowSummarizer with timedeltas was about temporal change of time series' frequencies ( to make sure that, after applying .shift, the new time series will also have the old indices).
This approach will work only if the index has a periodic index and I'm not sure if it's not too strong of a requirement.

The new version is much simpler - it's about adding empty rows with a reversed lag to the object in the _window_feature function. This will guarantee that the new table will have all the original indices.

fkiraly:
ah, that sounds good, and should work for irregular time indices, too, right?

GrzegorzRut:
> 

yup

Argh, I don't understand why the code quality check passed locally, if it failed here!

fkiraly:
> Argh, I don't understand why the code quality check passed locally, if it failed here!

Most probably your linting or precommit config is not 1:1 the same as sktime CI-CD?

If they aren´t there could be a bug with the config or the doc instructions

fkiraly:
(but it seems to run now)

fkiraly:
I think the tests are failing due to a bug unrelated to your work?
https://github.com/alan-turing-institute/sktime/issues/2628

FYI @danbartl

fkiraly (changes requested):
Really nice, thanks!

The tests are unfortunately failing, does not seem to be your fault (?) but still needs to be fixed. See #2628. Not sure what to do there, perhaps a quick chat with @danbartl helps to see what needs to be done.

Blocking issues with the docstring:
* it needs to be described that you can use integers *and* time offsets.
* from looking at your code, you can also use `pandas.Timedelta`, is that correct? That's great, but it should be mentioned in the docstring too. Best doing this in a list of three things.

GrzegorzRut:
> Really nice, thanks!
> 
> The tests are unfortunately failing, does not seem to be your fault (?) but still needs to be fixed. See #2628. Not sure what to do there, perhaps a quick chat with @danbartl helps to see what needs to be done.
> 
> Blocking issues with the docstring:
> 
>     * it needs to be described that you can use integers _and_ time offsets.
> 
>     * from looking at your code, you can also use `pandas.Timedelta`, is that correct? That's great, but it should be mentioned in the docstring too. Best doing this in a list of three things.

pd.TimedeltaIndex wasn't working but it's fixed now. I've also enabled back truncate_start (only for integer windows).

I added some lines to the docstring - do you think it's not enough? As you will see below, there is a small difference between integer and time windows. Perhaps this is something I should've explained in the docstring but I'm not sure how 

One issue that's a bit problematic is the forced lag in the integer lag specification [here](https://github.com/alan-turing-institute/sktime/blob/4d874c1c20a94d9006604a3916b6b434750b4735/sktime/transformations/series/summarize.py#L284)
It does make sense to look only at previous records but this will not generalize well with timedelta. It would be awkward to have a lag '5D' and add '1D' behind the curtain. That's why the line of code is slightly changed [here](https://github.com/GrzegorzRut/sktime/blob/4ca1f8b5fabf88752ac9d1ff26b6dc952b1da5d9/sktime/transformations/series/summarize.py#L335) and [here](https://github.com/GrzegorzRut/sktime/blob/4ca1f8b5fabf88752ac9d1ff26b6dc952b1da5d9/sktime/transformations/series/summarize.py#L602).

Many thanks for your help!

sktime/transformations/series/summarize.py
```diff
@@ -32,14 +32,31 @@ class WindowSummarizer(BaseTransformer):
...
+
+        The first element of the `window` is the `lag`, which specifies how far back
+        in the past the window will start. The second element, `window length`,
+        is the length of the window across which to apply the function.
+
+        For ease of notation, for the key "lag", only a single integer or offset
         specifying the `lag` argument will be provided.
 
-        Please see blow a graphical representation of the logic using the following
+        The offsets work only with datasets with time index (either pd.DateimeIndex,
```
fkiraly, line 47: the above text is better! But this section is unclear. I suggest to make it much clearer that you are talking about a sub-case.

E.g., "For offsets of string format..."

sktime/transformations/series/summarize.py
```diff
@@ -32,14 +32,31 @@ class WindowSummarizer(BaseTransformer):
...
         Dictionary specifying as key the type of function to be used and as value
         the argument `window`.
-        For all keys other than `lag`, the argument `window` is a length 2 list
-        containing the integer `lag`, which specifies how far back
-        in the past the window will start, and the integer `window length`,
-        which will give the length of the window across which to apply the function.
-        For ease of notation, for the key "lag", only a single integer
+
+        For all keys other than `lag`, the argument `window` is a length 2 list.
+        The window elements can be either integers indicate the number of
```
fkiraly, line 37: Suggestion:

```
Either elements of the window list can be one of:

* an integer, in which case it will indicate a difference in number of observations
* a string specifying a time period, see below - only for data `X` with `pd.DatatimeIndex` or `pd.TimedeltaIndex`
* an time difference object of type `pd.Timedelta`

Comment (fkiraly) [reply sktime/transformations/series/summarize.py:37]: (this is better structured and makes the options clearer)

File: sktime/transformations/series/summarize.py [hunk 3]
```diff
@@ -32,14 +32,31 @@ class WindowSummarizer(BaseTransformer):
...
+        * S - second,
+        * L - milisecond,
+        * U - microsecond,
+        * N - nanosecond.
+        For instance, the following offesets would be acceptable: '1D', '2H', '30S'.
+
+        Please see below a graphical representation of the logic using the following
         symbols:
 
         ``z`` = time stamp that the window is summarized *to*.
```
fkiraly, line 62: random note: should this not be `window_length-1` rather than `1-window_length`?

GrzegorzRut: I think it should be `1-window_length`. [This](https://github.com/alan-turing-institute/sktime/blob/4d874c1c20a94d9006604a3916b6b434750b4735/sktime/transformations/series/summarize.py#L46) could cover negative lags (not sure how they are used though).

sktime/transformations/series/summarize.py
GrzegorzRut: agreed!

fkiraly:
@grzegorzrut, aftere @danbartl is now back and has finished his refactor, I've tried to merge the two development branches. The problem with `truncate_start` should be resolved, as @danbartl has removed the requirement that the transformer has that param.

I'm not sure whether all has been merged correctly, so would appreciate if you could have a look.

GrzegorzRut (commented):
there is an issue with this approach that I was not able to solve - 
it either is going to introduce values that are not seen in the integer version of the algo or will cut some of the expected output (I can't find a solution between these two extremes)

i do not recommend merging this

```python
import pandas as pd

index = pd.date_range("2000-01-01", "2000-01-05", freq="D")
length = len(index)
y = pd.DataFrame(range(length), index=index)
kwargs = {
    "lag_feature": {
        "lag": [ 1, "1D"],
        "mean": [
            [0, "36H"],
            [0, int(1.5)],

            [0, "48H"],
            [0, 2],

            ['1D', "48H"],
            [1, 2],
        ],
    }
}
transformer = WindowSummarizer(**kwargs, n_jobs=1)
print(transformer.fit_transform(y).to_markdown())
```
|                     |   0_lag_1 |   0_lag_1D |   0_mean_0_36H |   0_mean_0_1 |   0_mean_0_48H |   0_mean_0_2 |   0_mean_1D_48H |   0_mean_1_2 |
|:--------------------|----------:|-----------:|---------------:|-------------:|---------------:|-------------:|----------------:|-------------:|
| 2000-01-01 00:00:00 |       nan |        nan |          nan   |            0 |          nan   |        nan   |           nan   |        nan   |
| 2000-01-02 00:00:00 |         0 |          0 |          nan   |            1 |          nan   |          0.5 |           nan   |        nan   |
| 2000-01-03 00:00:00 |         1 |          1 |            1.5 |            2 |            1.5 |          1.5 |           nan   |          0.5 |
| 2000-01-04 00:00:00 |         2 |          2 |            2.5 |            3 |            2.5 |          2.5 |             1.5 |          1.5 |
| 2000-01-05 00:00:00 |         3 |          3 |            3.5 |            4 |            3.5 |          3.5 |             2.5 |          2.5 |


i'll try to commit the latest version of what I have (it is still going to be having these issues but fixes some other things) in the nearest future

fkiraly:
> it either is going to introduce values that are not seen in the integer version of the algo or will cut some of the expected output (I can't find a solution between these two extremes)

Hm, I think I had the same problem when I wrote the `Lag` transformer.
(if I understand correctly, this is about handing the NA, right? where we have a choice to cut, extend, etc - if not, please correct me)

Could you look at that? Do you think we can handle this similarly, i.e., with an `index_out` argument, that allows the user to choose?

GrzegorzRut:
> Hm, I think I had the same problem when I wrote the `Lag` transformer. (if I understand correctly, this is about handing the NA, right? where we have a choice to cut, extend, etc - if not, please correct me)

Yup, it's connected with NAs (0_mean_1D_48H and 0_mean_1_2 are good examples of this).
I've commented out some [code](https://github.com/GrzegorzRut/sktime/blob/e809d878ee436e7604e02be9b4175131faccefa4/sktime/transformations/series/summarize.py#L673) that tries to deal with the issue (but currently it's either too aggressive or defensive).

> Could you look at that? Do you think we can handle this similarly, i.e., with an `index_out` argument, that allows the user to choose?
> 
> (and we set the default to whatever is consistent with the integer version)

I'll try to look at this during the weekend but I don't have much time for this.

fkiraly:
> I'll try to look at this during the weekend but I don't have much time for this.

What we could do instead, you make one defensive and one aggressive version (you ahve both, right)? And we could then provide that as an option.

GrzegorzRut:
> > I'll try to look at this during the weekend but I don't have much time for this.
> 
> What we could do instead, you make one defensive and one aggressive version (you ahve both, right)? And we could then provide that as an option.

there is one more thing i'd like to try - if it fails, I'll provide these 2 options

fkiraly:
how are things going, @GrzegorzRut, have you tried?

GrzegorzRut:
> how are things going, @GrzegorzRut, have you tried?

didn't work on it for the last couple of weeks, I'm sorry

fkiraly:
no worries, we're all busy in our day job...

PR #2432, opened 2022-04-10
[BUG] temp workaround for unnamed levels in hierarchical X passed to aggregator

fkiraly:
The `Aggregator` incorrectly assumed that all levels of the hierarchy were named.

This made it near impossible to use it in certain relevant pipelines.

This PR introduces a workaround which names the levels at the start and reverts to the original names at the end.
It should eventually be replaced by sth less hacky, but will have to do for pydata...


fkiraly:
FYI @ciaran-g

fkiraly:
as discussed in core dev channel, emergency merge due to bug in 0.11.1 release and pydata conf

Issue #2385, opened 2022-04-03
[BUG] disable very aggressive citation message from `dtw-python`


fkiraly:
Every time one imports the `dtw-python` module, even when attempting to do so silently and with warning suppressed, the following is printed

```
Importing the dtw module. When using in academic works please cite:
  T. Giorgino. Computing and Visualizing Dynamic Time Warping Alignments in R: The dtw Package.
  J. Stat. Soft., doi:10.18637/jss.v031.i07.
```

Besides being annoying, it is quite aggressive and pops up on various utilities that are unrelated to the estimator specifically.

We should find out how to get rid of this.
May or may not be easy to do...

TonyBagnall:
Where is this package used?

fkiraly:
the alignment module. It is one of the most sequence alignment packages in python.

Really great package, but we don't need to show the citation message every time we do a random unrelated thing with `sktime`...

KatieBuc:
@fkiraly curious what you think about the proposed solution in the latest pull request.

Here's the culprit: https://github.com/DynamicTimeWarping/dtw-python/blob/master/dtw/__init__.py

fkiraly:
> @fkiraly curious what you think about the proposed solution in the latest pull request.

I think it works, partially - I guess the solution is as aggressive as the message itself, but perhaps that cannot be avoided.
Partially, becausae it does not address the `import dtw` statement later on. Would you mind also adding the same wrapper there? I.e., in the two `_fit` methods?

Might also be worth making a PR to `dtw-python` someday to simply remove that message? But that's a different story.

fkiraly:
@KatieBuc, I'm on main and still getting this message - whatever you did might not have killed it completely?

![image](https://user-images.githubusercontent.com/7985502/162615205-9e9dd25b-1506-46e1-8653-3bccaeec3ae0.png)

Re-opening this issue, hence...

KatieBuc:
Ah, I see @fkiraly 

Line 147 there's import_module which will need to have the same overwritten sys.stdout as in _check_soft_dependencies. Will work on it, and check elsewhere.

KatieBuc:
![image](https://user-images.githubusercontent.com/34525024/162656900-bd7ecd1b-3aeb-4c47-80ed-18b0c0d73684.png)

registry/_lookup.py *** created PR for fix
utils/_maint/_show_versions.py (doesn't have dtw in deps list, not necessary to change)
[...]

PR #2439, opened 2022-04-11
Katie

KatieBuc:
#### Reference Issues/PRs
BUG] disable very aggressive citation message from dtw-python #2385

Importing the dtw module has a forced print message. Everywhere that calls import_module:
  registry/_lookup.py *** fixing now
  utils/_maint/_show_versions.py (doesn't have dtw in deps list, not necessary to change)
  utils/validation/_dependencies.py (_check_soft_dependencies changed in previous commit d6fb526)

no

I wasn't able to recreate the problem from the below, which I find a bit strange
from sktime.registry import all_estimators
all_estimators('forecaster')

However, testing this works with the changes
all_estimators('forecaster', suppress_import_stdout=True)

### DIFF
--- a/sktime/alignment/dtw_python.py
+++ b/sktime/alignment/dtw_python.py
@@ -13,7 +13,7 @@ from sklearn import clone
 from sktime.alignment.base import BaseAligner
 from sktime.utils.validation._dependencies import _check_soft_dependencies
 
-_check_soft_dependencies("dtw", severity="warning")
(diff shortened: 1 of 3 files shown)


fkiraly (changes requested):
looks good and this seems to solve it for me now.

one request, could we make the default in the `all_estimators` to be `True`?

In terms of user expectation, no printouts whatsoever is what is strongly expected here, imo.

Issue #2388, opened 2022-04-04
Add `random_state` to `statsmodel` adapter


ngupta23:
Requesting the addition of a `random_state` argument to the statsmodels adapter and passing it appropriately to underlying non-deterministic models such as ETS so that the predictions can be made reproducible. Currently, multiple runs result in different interval results for the same quantiles.

More on this below:

> Many statsmodels models are non-deterministic, e.g., as they do some random sampling inside, especially when it comes to estimating variation. Having some intrinsic randomness there is not unusual and expected.

@fkiraly You are right, it looks like in this case, the intervals are based on simulations and those simulations are not exact or repeatable. This model has multiplicative components and hence exact method is not available

https://www.statsmodels.org/stable/_modules/statsmodels/tsa/exponential_smoothing/ets.html

```python
if method is None:
    exact_available = ["ANN", "AAN", "AAdN", "ANA", "AAA", "AAdA"]
    if results.model.short_name in exact_available:
        method = "exact"
    else:
        method = "simulated"
self.method = method
```

However, they can be made repeatable by passing a `random_state` inside the sktime ets.py `_predict_interval` method. What do you think - maybe open an enhancement request for this?  

```python
def _predict_interval(self, fh, X=None, coverage=None):
    valid_indices = fh.to_absolute(self.cutoff).to_pandas()

   start, end = valid_indices[[0, -1]]
   # Maybe add random_state in get_predictions (and accept random_state in the AutoETS initialization)?
   # https://www.statsmodels.org/stable/_modules/statsmodels/tsa/exponential_smoothing/ets.html
   prediction_results = self._fitted_forecaster.get_prediction(
       start=start, end=end
   )
```

_Originally posted by @ngupta23 in https://github.com/alan-turing-institute/sktime/issues/2320#issuecomment-1080029680_

ngupta23:
For example, in ETS, the random state can be passed to the `get_prediction` method that is called from within sktime to make the prediction intervals deterministic.

https://github.com/alan-turing-institute/sktime/pull/2320#issuecomment-1087330315

fkiraly:
agreed! If `random_state` is not already passed properly, it should be.

Ris-Bali:
Do we need to pass the random_state argument to the constructor of Statsmodels adapter ? and then pass that argument to the appropriate methods(methods with get_prediction) of forcasters that interface from statsnmodels ?

Ris-Bali:
Okay  understood ,can I work on this ?

[...]

Issue #2315, opened 2022-03-27
[ENH] AutoETS: heavy random fluctuations in interval predictions - investigate random seed and correctness


ngupta23:
Update by @fkiraly: the below is not a bug, but a consequence of what seems like quite serious random variation in the prediction intervals and quantiles produced by `AutoETS`.

It may be useful to investigate whether:
* a random seed can be set in `AutoETS` to allow deterministic comparisons
* whether the prediction intervals should be something different (e.g., variation predicted added to variation in re-sampling the prediction)
* whether this fluctuation implies anything about the correctness of `AutoETS`

---

**Describe the bug**

Followup from ##2275: AutoETS behavior still seems inconsistent in the 0.11.0 release (ARIMA and Theta has to be fixed now).

**To Reproduce**

```python
from sktime.datasets import load_airline
from sktime.forecasting.ets import AutoETS
y = load_airline()

forecaster = AutoETS(auto=True, n_jobs=-1, sp=12)
forecaster.fit(y, fh=[1,2,3])

# return_pred_int and alpha are disabled in sktime 0.11.0
# forecaster.predict(return_pred_int=True, alpha=0.05)
forecaster.predict()

forecaster.predict_quantiles(alpha=[0.025, 0.05, 0.1, 0.5, 0.9, 0.95, 0.975])

forecaster.predict_interval(coverage=[0.8, 0.9, 0.95])
```

![image](https://user-images.githubusercontent.com/33585645/160263672-3deed4ef-41cb-423a-9218-cab02994230b.png)

**Expected behavior**
Would have expected the point forecast to match 50% quantile and coverage = 80% to match quantiles [0.1, 0.9] and so on. This seems to be the case in ARIMA and ThetaForecaster but not in AutoETS.

**Versions**
<details>

System:
    python: 3.7.13 (default, Mar 16 2022, 17:37:17)  [GCC 7.5.0]
executable: /usr/bin/python3
   machine: Linux-5.4.144+-x86_64-with-Ubuntu-18.04-bionic

Python dependencies:
          pip: 21.1.3
   setuptools: 57.4.0
      sklearn: 1.0.2
       sktime: 0.11.0
  statsmodels: 0.13.2
        numpy: 1.21.5
        scipy: 1.4.1
       pandas: 1.3.5
   matplotlib: 3.2.2
       joblib: 1.1.0
        numba: 0.55.1
     pmdarima: 1.8.5
      tsfresh: None

</details>

[...]

PR #2440, opened 2022-04-11
[ENH] Add random_state to statsmodels adapter 

Ris-Bali:
#### Reference Issues/PRs
#2388 

Adds random_state to statsnmodels so that it can be called internally by any model.

No

- [x] Have passed random_state to statsmodeladapter
- [x]  Passed it as an argument to get_predictions method. 

##### For all contributions

##### For new estimators


fkiraly (changes requested):
This covers only the adapter.
The individual classes that inherit from it do not set the `random_state` to anything, so this does not address the issue fully.

fkiraly (changes requested):
Looks good for a start!

That is, `AutoETS` looks good now. I think this needs to be passed on in the same way in the constructors, for all other models interfaced from `statsmodels`.

Ris-Bali:
@fkiraly I have added the parameter to all the models.

sktime/forecasting/ets.py
```diff
@@ -444,7 +450,7 @@ def _predict_interval(self, fh, X=None, coverage=None):
 
         start, end = valid_indices[[0, -1]]
         prediction_results = self._fitted_forecaster.get_prediction(
-            start=start, end=end
+            start=start, end=end, random_state=self.random_state
```
aiwalter, line 453: is this only used here this `get_prediction` method?

Ris-Bali: Yes , actually in PR #2315 ets needs a random_state to be set so as to avoid fluctuations in internal predictions random_state can be set only in the simulate method get_prediction can pass parameters to the simulate method through (**simulate kwargs)
image

aiwalter: are we passing that then also to `simulate`?

Ris-Bali: Not separately ; no

aiwalter: ah `get_prediction` is a `statsmodels` method, so then I think its fine

aiwalter (approved):

fkiraly (changes requested):
Great, this looks good now!

Could we add a test that loops through all the affected estimators, runs fit/predict twice on the same data, and checks that the predictions are identical?

Ris-Bali:
@fkiraly I checked with the AutoETS model (the model in which the problem existed) and I'm still getting the randomness in the results as in #2320 
image
I set random_state =0 and still, there is some randomness in the results.

aiwalter:
might be the `**kwargs` where you have to pass it also?

aiwalter:
oh thos are prediction intervals that you showed, that might not be related to the estimator. Can you pls add a test for testing the predictions for random? You would have to fit and predict the same estimator twice and make `assert_frame_equal` or `assert_series_equal`

Ris-Bali:
I tried to loop through the models by creating a models list and iterating through it but I got an error unknown keyword argument random_state error thus I did not loop through the models .

fkiraly:
are you using the `all_estimators` utility?

You may want to check whether `random_state` is in `get_params().keys()`

Ris-Bali:
Is this implementation okay or should I make some changes?

Ris-Bali:
@fkiraly , @aiwalter  Sorry for disturbing ; but could you please review the changes I have made ?

sktime/forecasting/base/tests/randomtest.py
```diff
@@ -0,0 +1,44 @@
...
+
+fh = np.arange(1, 5)
+
+y = load_airline()
+y_1 = make_forecasting_problem(n_columns=3)
+
+
+def test_random_state():
+    """Function for testing models with random_state parameter."""
+    models = []
```
fkiraly, line 25: use pytest.mark.parametrize instead

Ris-Bali:
@fkiraly I think I made the suggested changes . Pls review them .

fkiraly (approved):
Looks good!

Minor, non-blocking things, you can do them in a separate PR:

* round brackets in line 25 of tests are not needed
* might be better to use `model.create_test_instance` instead of `model()`, that ensures that the instances run quickly

fkiraly:
we need the tests ro pass before merge of course

PR #2447, opened 2022-04-11
[ENH] First transfer of deep learning classifiers and regressors from sktime-dl

TonyBagnall:
this is my first go at this and open to discussion as to structure. Once we have agreed how to do CNN, the rest of the algorithms will come over quickly. However, there are features here some may dislike, so want to out there early. Points to note
1. Introduces a whole new module called networks. This contains the network structures that are shared by both classifiers and regressors. This is necessary because it was agreed long ago to split the algorithms into regressors/classifiers (rather than have one class with both capabilities, which was my preferred structure)
2. classifiers go in classification/deep_learning, and there is an abstract class there that inherits from BaseClassifier. This intermediate class is there because we can define default DL classifier predict_proba behaviour, and we want an extra abstract method.

In terms of the implementation, this is not yet correct for CNN, and there is not yet any testing. Following issues

1. Input X: sktime uses (n,d,m) but it seems keras assumes (n,m,d), easy to fix but I need to confirm this, but its a bit annyoing
2. Input Y: I'm hoping we dont need the label encoding, and can just use the one hot encoder if required, again needs resolving with base class. The issue here is whether keras assumes classes are mapped onto 0,...,n_classes-1 or not. 
3. The validation data. This is trickier, since this is redefining _fit from the base, which is undesirable. I would like input/opinions on preferred way of setting validation data in training. 
4. I need to sort out the random state, but that should be easy
5. Need to check it works with soft dependency set up

### DIFF
--- a/sktime/classification/base.py
+++ b/sktime/classification/base.py
@@ -41,12 +41,12 @@ class BaseClassifier(BaseEstimator, ABC):
     """Abstract base class for time series classifiers.
 
     The base classifier specifies the methods and method signatures that all
-    classifiers have to implement. Attributes with a underscore suffix are set in the
+    classifiers have to implement. Attributes with an underscore suffix are set in the
     method fit.
 
     Parameters
     ----------
-    classes_            : ndarray of class labels, possibly strings
+    classes_            : array of class labels, possibly strings
     n_classes_          : integer, number of classes (length of classes_)
     fit_time_           : integer, time (in milliseconds) for fit to run.
     _class_dictionary   : dictionary mapping classes_ onto integers 0...n_classes_-1.
@@ -298,6 +298,7 @@ class BaseClassifier(BaseEstimator, ABC):
             for specifications, see examples/AA_datatypes_and_datasets.ipynb
         y : 1D np.array of int, of shape [n_instances] - class labels for fitting
             indices correspond to instance indices in X
+
         Returns
         -------
         self :
--- a/sktime/classification/base.py
+++ b/sktime/classification/base.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-# copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
 """
 Abstract base class for time series classifiers.
 
@@ -82,7 +81,8 @@ class BaseClassifier(BaseEstimator, ABC):
     def __rmul__(self, other):
         """Magic * method, return concatenated ClassifierPipeline, transformers on left.
 
-        Implemented for `other` being a transformer, otherwise returns `NotImplemented`.
+        Overloaded multiplication operation for classifiers. Implemented for `other`
+        being a transformer, otherwise returns `NotImplemented`.
 
         Parameters
         ----------
--- a/sktime/classification/deep_learning/base.py
+++ b/sktime/classification/deep_learning/base.py
@@ -1,0 +1,113 @@
+# -*- coding: utf-8 -*-
+"""
+Abstract base class for the Keras neural network classifiers.
+
+The reason for this class between BaseClassifier and deep_learning classifiers is
+because we can generalise tags, _predict and _predict_proba
+"""
+__author__ = ["James-Large", "ABostrom", "TonyBagnall"]
+__all__ = ["BaseDeepClassifier"]
+
+import numpy as np
+from abc import abstractmethod
+from sklearn.preprocessing import LabelEncoder
+from sklearn.preprocessing import OneHotEncoder
+from sktime.classification.base import BaseClassifier
+
+
+class BaseDeepClassifier(BaseClassifier):
+    """Abstract base class for deep learning time series classifiers.
+
+    The base classifier provides a deep learning default method for
+    _predict and _predict_proba, and provides a new abstract method for building a
+    model.
+
+    Parameters
+    ----------
+    batch_size : int, default = 40
+        training batch size for the model
+    Arguments
+    ---------
+    self.model = None
+
+    """
+    _tags = {"X_inner_mtype": "numpy3D"}
+
+    def __init__(self, batch_size=40):
+        self.batch_size = batch_size
+        self.model_ = None
+
+    @abstractmethod
+    def build_model(self, input_shape, n_classes, **kwargs):
+        """
+        Construct a compiled, un-trained, keras model that is ready for
+        training
+
+        Parameters
+        ----------
+        input_shape : tuple
+            The shape of the data fed into the input layer
+        n_classes: int
+            The number of classes, which shall become the size of the output
+            layer
+
+        Returns
+        -------
+        output : a compiled Keras Model
+        """
+        ...
+
+    @abstractmethod
+    def _fit(self, X, y):
+        """Pass through abstract method from BaseClassifier."""
+        ...
+
+    def _predict(self, X, **kwargs):
+        probs = _predict_proba(self, X,**kwargs)
+        return probs # Max
+
+    def _predict_proba(self, **kwargs):
+        """
+        Find probability estimates for each class for all cases in X.
+        Parameters
+        ----------
+        X : a nested pd.Dataframe, or (if input_checks=False) array-like of
+        shape = (n_instances, series_length, n_dimensions)
+            The training input samples. If a 2D array-like is passed,
+            n_dimensions is assumed to be 1.
+        input_checks: boolean
+            whether to check the X parameter
+        Returns
+        -------
+        output : array of shape = [n_instances, n_classes] of probabilities
+        """
+
+        probs = self.model_.predict(X, self.batch_size, **kwargs)
+
+        # check if binary classification
+        if probs.shape[1] == 1:
+            # first column is probability of class 0 and second is of class 1
+            probs = np.hstack([1 - probs, probs])
+
+        return probs
+
+    def convert_y(self, y, label_encoder=None, onehot_encoder=None):
+        if (label_encoder is None) and (onehot_encoder is None):
+            # make the encoders and store in self
+            self.label_encoder = LabelEncoder()
+            self.onehot_encoder = OneHotEncoder(sparse=False,
+                                                categories="auto")
+            # categories='auto' to get rid of FutureWarning
+
+            y = self.label_encoder.fit_transform(y)
+            self.classes_ = self.label_encoder.classes_
+            self.n_classes_ = len(self.classes_)
+
+            y = y.reshape(len(y), 1)
+            y = self.onehot_encoder.fit_transform(y)
+        else:
+            y = label_encoder.fit_transform(y)
+            y = y.reshape(len(y), 1)
+            y = onehot_encoder.fit_transform(y)
+
+        return y
--- a/sktime/classification/deep_learning/_cnn.py
+++ b/sktime/classification/deep_learning/_cnn.py
@@ -4,36 +4,38 @@
 __author__ = ["JamesLarge", "TonyBagnall"]
 __all__ = ["CNNClassifier"]
 
-from sktime.classification.deep_learning import BaseDeepClassifier
+from sktime.classification.deep_learning.base import BaseDeepClassifier
 from sktime.networks import CNNNetwork
 
 
-from sktime_dl.utils import check_and_clean_data, \
-    check_and_clean_validation_data
+#from sktime_dl.utils import check_and_clean_data, \
+#    check_and_clean_validation_data
 from sklearn.utils import check_random_state
 from tensorflow import keras
 
 
 class CNNClassifier(BaseDeepClassifier, CNNNetwork):
-    """Time Convolutional Neural Network (CNN).
+    """Time Convolutional Neural Network (CNN), as described in [1].
 
     Parameters
     ----------
-    nb_epochs: int, the number of epochs to train the model
-    batch_size: int, the number of samples per gradient update.
-    kernel_size: int, specifying the length of the 1D convolution
-     window
-    avg_pool_size: int, size of the average pooling windows
-    n_conv_layers: int, the number of convolutional plus average
-     pooling layers
-    filter_sizes: int, array of shape = (nb_conv_layers)
-    callbacks: list of tf.keras.callbacks.Callback objects
-    random_state: int, or sklearn Random.state
-    verbose: boolean, whether to output extra information
-    model_name: string, the name of this model for printing and
-    file writing purposes
-    model_save_directory: string, if not None; location to save
-    the trained keras model in hdf5 format
+    should inherited fields be listed here?
+    n_epochs       : int, default = 2000
+        the number of epochs to train the model
+    batch_size      : int, default = 16
+        the number of samples per gradient update.
+    kernel_size     : int, default = 7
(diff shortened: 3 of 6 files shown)


sktime/classification/base.py
```diff
@@ -42,12 +41,12 @@ class BaseClassifier(BaseEstimator, ABC):
...
 
     The base classifier specifies the methods and method signatures that all
-    classifiers have to implement. Attributes with a underscore suffix are set in the
+    classifiers have to implement. Attributes with an underscore suffix are set in the
     method fit.
 
     Parameters
     ----------
-    classes_            : ndarray of class labels, possibly strings
+    classes_            : array of class labels, possibly strings
```
fkiraly, line 49: should be `np.ndarray` if that's the class/type, `array` is not a class/type

sktime/classification/deep_learning/base.py
```diff
@@ -0,0 +1,119 @@
...
+            The shape of the data fed into the input layer
+        n_classes: int
+            The number of classes, which shall become the size of the output
+            layer
+
+        Returns
+        -------
+        output : a compiled Keras Model
+        """
+        ...
```
fkiraly, line 64: is sth missing here?

sktime/classification/deep_learning/base.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def _predict(self, X, **kwargs):
```
fkiraly, line 66: I think we have no `kwargs` in the functions? Or do we want them?

sktime/classification/deep_learning/base.py
```diff
@@ -0,0 +1,119 @@
...
+        probs = self.model_.predict(X, self.batch_size, **kwargs)
+
+        # check if binary classification
+        if probs.shape[1] == 1:
+            # first column is probability of class 0 and second is of class 1
+            probs = np.hstack([1 - probs, probs])
+
+        return probs
+
+    def convert_y(self, y, label_encoder=None, onehot_encoder=None):
```
fkiraly, line 100: maybe call this `convert_y_to_keras` or similar?
to avoid confusion and accidental overrides

sktime/classification/deep_learning/cnn.py
```diff
@@ -0,0 +1,143 @@
...
+    def __init__(
+        self,
+        n_epochs=2000,
+        batch_size=16,
+        kernel_size=7,
+        avg_pool_size=3,
+        n_conv_layers=2,
+        callbacks=None,
+        verbose=False,
+    ):
```
fkiraly, line 60: there should be a `_check_soft_dependencies` here too, at error severity level

sktime/classification/deep_learning/cnn.py
```diff
@@ -0,0 +1,143 @@
...
+        """
+        if self.callbacks is None:
+            self._callbacks = []
+        y_onehot = self.convert_y(y)
+        check_random_state(self.random_state)
+
+        # Transpose to conform to Keras input style.
+        X = X.transpose(0, 2, 1)
+        self.input_shape = X.shape[1:]
+        self.model_ = self.build_model(self.input_shape, self.n_classes_)
```
fkiraly, line 132: this looks like much of it will be the same - should this be boilerplate that is in the intermediate base class?
Can you perhaps get rid of `_fit` entirely?

fkiraly (commented):
had a quick look - I think, high-level, code structure makes sense.

Having the "intermediate base class" or "adapter base class" is the right idea - I'd encourage to think about how as much as possible keras boilerplate can go there. I feel some of what is in `_fit` and perhaps `build_model` could move there.

TonyBagnall:
> had a quick look - I think, high-level, code structure makes sense.
> 
> Having the "intermediate base class" or "adapter base class" is the right idea - I'd encourage to think about how as much as possible keras boilerplate can go there. I feel some of what is in `_fit` and perhaps `build_model` could move there.

I agree, but I think I want to wait until I do the subsequent ones until I see how much can be moved up to the parent class, the commonality will be clearer then

fkiraly:
> I agree, but I think I want to wait until I do the subsequent ones until I see how much can be moved up to the parent class, the commonality will be clearer then

Sure - I assume you want to refactor then as you develop, before this PR is final?

sktime/classification/deep_learning/base.py
TonyBagnall: I've removed the text output : it returns a keras Model object,

sktime/classification/deep_learning/base.py
TonyBagnall: dont need kwargs, I've removed them, not used

sktime/classification/deep_learning/cnn.py
TonyBagnall: I've raised the severity to error, but was then wondering whether I could I not just do it in the base classes rather than in every individual file, I can try it out with the second one

sktime/classification/deep_learning/cnn.py
TonyBagnall: possibly, I'll look through on the next PR to see if it could be moved up to the base.

TonyBagnall:
I'd like to complete the port over several PRs to avoid one massive PR that drags on for months. I think this can go in as the first of several, I will do the regression equivalent next with a refactor of the regressor base class to fit/_fit etc

fkiraly (approved):
OK, makes sense.

Some remarks, blocking until resolved or discussed:

* The parameters of `CNNNetwork` cannot be accessed from `CNNClassifier` via the `sklearn` interface. Should this not be a component, i.e., a constructor arg? And should `BaseDeepNetwork` not inherit from `BaseObject` so these parameters can be accessed?
* a name `CNNClassifier` is pretty generic. There is not going to be only a single CNN classifier, is there? Perhaps a better name would be appropriate?
* the entire conversion business at the start of `_fit` looks *again* like conversion boilerplate that you intend to multiply across all classes inheriting from `BaseDeepClassifier`. This looks like a prime case for an mtype that applies to the `y`. You know, `y_inner_mtype = "keras_labels"` or similar.

fkiraly (changes requested):
mis-clicked, the above should have been "change request"

fkiraly:
> so you mean containment rather than inheritance? If not, could you mock me an example of what you mean? I'm not sure that I really want to build an interface between sklearn and keras, I'm sure such things already exist, but this is is just the first vanilla port from sktime-dl to sktime. Happy for BaseDeepNetwork to inherit from our BaseObject,

Would be a simple change imo (may be too naive though). Idea is, make the following two changes:

1. let `BaseDeepNetwork` inherit from `BaseObject` (I see you have already done this)

2. change `CNNClassifier` as follows:

```python

class CNNClassifier(BaseDeepClassifier):

    def __init__(self, cnn_network, n_epochs=2000, etcetera)
        self.cnn_network = self.cnn_network
        self.cnn_network_ = clone(self.cnn_network)   # use this clone internally

    ...
line 86:
        input_layer, output_layer = self.cnn_network_.build_network(input_shape, **kwargs)
```

fkiraly:
> I agree it can all move up to the base classifier if appropriate, but I need to check it is the same across all 13 classifiers before doing so.

Ok, I'm happy to do it as a refactor and not up-front if it is not clear that it is this precise boilerplate that will be used 13 times later. I consider this point addressed by your comment.

fkiraly:
> this is a "vanilla" CNNNetwork, used in the most basic way, and it was called just CNN in Fawaz et al., whose naming conventions we have adopted.

Ok, fair - if it is common terminology in the field of time series classification, happy to keep it this way. Point addressed.

TonyBagnall:
so to be clear, as the base model for all regressors and classifiers, you prefer containment to inheritance? I'm ok with that, but I want confirmation because I could have sworn you said previously that you liked it as it is.  Also I dont get this

fkiraly:
> so to be clear, as the base model for all regressors and classifiers, you prefer containment to inheritance?

Can you clarify this, please? I do not understand your question, sorry.

I thought the base model for all classifiers was `BaseClassifier`, and the base model for all neural network classifiers was `BaseDeepClassifier`, neither of which would be impacted by this?

>did you mean `self.cnn_network = cnn_network`

yes, of course - silly mistake, sorry.

> Will `sklearn.clone` work with something that is not an sktime estimator?

You mean sklearn estimator, right?

No, it won't work with sth that's not an sklearn estimator, but `BaseObject` *is* an sklearn estimator!

fkiraly:
> you prefer containment to inheritance?

Generally, yes,
https://en.wikipedia.org/wiki/Composition_over_inheritance
and especially if it looks like "putting a model/algorithm together from multiple pieces" (which the above looks like to me, but I'd appreciate if you'd help me understand the above point better).

TonyBagnall:
> > so to be clear, as the base model for all regressors and classifiers, you prefer containment to inheritance?
> 
> Can you clarify this, please? I do not understand your question, sorry.

currently, sktime-dl uses multiple inheritance
image

I asked, probably on slack now I think about it, for expressions of preference as to inheritance vs composition/aggregation. I'll see if containment raises any issues, but I'm not sure it sure be configurable via the init arguments, the CNNClassifier should always use a CNNNetwork, and the parameters of that network should be parameters of the constructor to CNNClassifier, not the network itself. I think there should be an argument for activation="sigmoid",  loss="mean_squared_error", and metrics=["accuracy"] in the constructor though

TonyBagnall:
> 
first test for inheritance vs containment is whether the relationship is "IS-A" or "HAS-A". A CNNclassifier could be considered to both be a network and have a network. In this situation, I would always prefer containment, specially when multiple inheritance is involved. However, my other null position is if it aint broke, dont fix it, especially as my understanding of keras is minimal. Given it was already implemented with multiple inheritance, I started off with that position. But given your preference, I'll try switch it to containment, and if it doesnt raise any problems with keras, proceed with that

TonyBagnall:
switched to containment, seems ok, although there is some conflict between strings and keras, so its still hard coded as sigmoid and precommit has weirdly reformatted the input strings 
image

I also need to look at seeding, although this will always be a problem, I'm not sure you can make GPU deterministic, I'll look. I think this PR is a way from going in, I want to reproduce the published results and get it running on the GPU server before putting it in

TonyBagnall:
weird this, it works with activation = "sigmoid" but not this 
image

fkiraly:
>  the CNNClassifier should always use a CNNNetwork, and the parameters of that network should be parameters of the constructor to CNNClassifier

Well, but you could come up with another architecture of the network, no? So it would not always be a CNNNetwork, but sth that shares interface with it.

fkiraly:
> weird this, it works with activation = "sigmoid" but not this

What do you mean by that?

fkiraly:
you also want to remove the tuple brackets from around `loss` and `metric`, this was probably added by precommit after you left a comma there by accident

TonyBagnall:
I have recreated the CNN results to my satisfaction from here, https://raw.githubusercontent.com/hfawaz/dl-4-tsc/master/results/results-ucr-128.csv so this is ready to go in as the first version. I will then take a deeper look at the rest to see if we can generalise more code into the base

sktime/networks/cnn.py
```diff
@@ -0,0 +1,105 @@
+# -*- coding: utf-8 -*-
+"""Time Convolutional Neural Network (CNN) (minus the final output layer)."""
+
+__author__ = ["James-Large, Withington, TonyBagnall"]
+
+from sktime.networks.base import BaseDeepNetwork
+from sktime.utils.validation._dependencies import _check_soft_dependencies
+
+_check_soft_dependencies("tensorflow", severity="error")
```
fkiraly, line 9: this should be a warning, and inside the `__init__` with an error.
See the "soft dependencies" docs.

Also, it should be `_check_dl_dependencies`.

fkiraly (changes requested):
Looks ok, but you need to adhere to the dependency handling via checks, see docs for that.

TonyBagnall:
> Looks ok, but you need to adhere to the dependency handling via checks, see docs for that.

tried _check_dl_dependencies but it doesn't have a severity argument, so reverted back to _check_soft_dependencies. I'm not sure why there is a separate method for dl dependencies, will surely just be duplication of code, and I know how much you hate that

fkiraly:
>tried `_check_dl_dependencies` but it doesn't have a severity argument

now it does: https://github.com/alan-turing-institute/sktime/pull/2627

> so reverted back to `_check_soft_dependencies`.

would in-principle be fine by me

> I'm not sure why there is a separate method for dl dependencies, 

more informative (deep learning specific) error message and developer comfort for an anticipated frequent developer case.

> will surely just be duplication of code, and I know how much you hate tha

Agree, could easily be refactored.
Although in this case I would put user and developer comfort above "neat code".

TonyBagnall:
@fkiraly I believe you have someone working on the porting of sktime-dl, which is great. I would like to get this PR in first in order to fix the structure. I've lost track of what you wanted changing here, are you still blocking it, and if so, what do you want changing?

fkiraly:
@TonyBagnall, what have you done/decided about the dependency handling?

Also, it's not "me having someone work" on DL. `sktime`, as a programme & community, is hosting summer projects, one of which is working on this, yes. You are more than welcome to engage with the summer mentees, the meetings are on our calendar, and as CC member you also have a full view of the process (see "updates on GSoC" items).

TonyBagnall:
> @TonyBagnall, what have you done/decided about the dependency handling?
> 
> Also, it's not "me having someone work" on DL. `sktime`, as a programme & community, is hosting summer projects, one of which is working on this, yes. You are more than welcome to engage with the summer mentees, the meetings are on our calendar, and as CC member you also have a full view of the process (see "updates on GSoC" items).

ok, soz my wording offends you, not sure what you mean by "decided about the dependency handling", could you clarify?

fkiraly:
> ok, soz my wording offends you, 

I'm not offended, I just wanted to disagree with the (possibly unintentionally) possessive framing.

> not sure what you mean by "decided about the dependency handling", could you clarify?

Refers to my change request, in which I said:

> Looks ok, but you need to adhere to the dependency handling via checks, see docs for that.

More precisely, I was asking above, for:

* adding informative warning and error messages for developers using the DL functionality
* removing any errors that would be triggered by module import

and with my more recent question, I have asked you for explaining whether and how this has been addressed.

Addressing could mean (a) agreeing with my request and taking some action, or (b) disagreeing and providing a statement why you think you do not need to do it, or (c) thinking something different needs to be done, stating the alternative/refined need and explaining the action taken.

TonyBagnall:
I have switched to your _check_dl_dependencies given you went to the trouble of updating it. I think it already complied with the docs on new soft dependencies (warn/error, imports in functions)

TonyBagnall:
@fkiraly  I would like this in before the next release, could you unblock or give specific changes you would like?

xxl4tomxu98:
is this resolved? e.g. am I able to install the sktime[all_extra] and then all soft dependency are satisfied? including tensorflow?

xxl4tomxu98:
The problem with sktime-dl is that I am using ARM based Mac
 and it won't work when I try to do "conda install sktime-dl"

PR #2454, opened 2022-04-13
[ENH] Add prediction intervals for `UnobservedComponets` forecaster

juanitorduz:
#### Reference Issues/PRs

Fixes https://github.com/alan-turing-institute/sktime/issues/2441

Add prediction intervals for the `UnobservedComponents` model wrapper from `statsmodels`.

No

The functionality is just taken from `statsmodels` itself as described in https://github.com/alan-turing-institute/sktime/issues/2441 so its mainly that the implementation works as expected. This PR takes as a base the existing implementation for the `ETS`  forecaster, for comparison  see the method [`_predict_interval`](https://github.com/alan-turing-institute/sktime/blob/main/sktime/forecasting/ets.py#L407).

##### For all contributions


sktime/forecasting/structural.py
```diff
@@ -319,6 +325,59 @@ def _fit_forecaster(self, y, X=None):
             low_memory=self.low_memory,
         )
 
+    def _predict_interval(self, fh, X=None, coverage=0.90):
```
fkiraly, line 328: according to the extension template, this should be
`    def _predict_quantiles(self, fh, X=None, alpha=None):`

fkiraly (approved):
looks good!

Left one comment, it shouldn´t make any difference, but it´s better to stick 1:1 to the extension template in general.

sktime/forecasting/structural.py
juanitorduz: thanks for the input! I changed the implementation in https://github.com/alan-turing-institute/sktime/pull/2454/commits/5d069bb1d0b39c3c593c9dba06321f4d21e7f419. Now I am writing some unit-test because I think the current implementation handles `exog` variables correctly (that is why I this one still a draft 😅 ).

I am also refactoring the tests for this estimators because I have gained a bit more experience writing nicer  pytests :)

fkiraly: great!

Although, now they fail...

juanitorduz:
I added the the possibility for `exog` input + tests in https://github.com/alan-turing-institute/sktime/pull/2454/commits/a7b4e3d2b95e703b2487fdb1ed62d6f0560523ad Do you think more tests are needed?

fkiraly:
> Do you think more tests are needed?

No, looks quite exhaustive to me.

juanitorduz:
mmm in https://github.com/alan-turing-institute/sktime/pull/2454/commits/fec1d7f870b6ff75d708e171d86f90ad97bfe914 with the coverage implementation all tests passes whereas with the quantile one the tests are failing ... which could mean I am probably using the wrong quantile format? Maybe alpha -> alpha/2 right?

fkiraly:
I am getting really confused about what you have been doing in the last commits.
Everything was running fine!

From my view, you should implement (as you have) `_predict_interval`, because you are interfacing an algorithm that produces prediction intervals, as simple as that.

`_predict_quantiles` is easily inferred by default logic, you do not need to implement that if you have implemented `_predict_interval`. It also has a different return type, but you are still using the `_predict_interval` return type.

As said, I'm confused, because your earlier version was fine, and now you seem to be implementing the wrong function with the wrong interface - but just hours after you've implemented the right function with the right interface and all tests passed??

I hope I did not do anything to confuse you here? Please let me know if I can help clarify.

fkiraly:
PS: regarding quantiles vs coverage - your latest post seems to be one of the "frequent confusions" mentioned in the pydata Berlin 2022 notebook no.2 (you were there in the talk).

I would recommend, just do what you did - `coverage` should be `1-alpha` (the `alpha` in `statsmodels`), I think that was correct, your first thought was exactly right. Now you seem to try change it to something that is not quite right...

Perhaps the confusion comes from there being an `alpha` in `statsmodels` and an `alpha` in `_predict_quantiles`? These are not the same, and for the moment I would advise just to forget about them...

juanitorduz:
Thanks for your feedback! I think I got confused with your comment
> according to the extension template, this should be
> `def _predict_quantiles(self, fh, X=None, alpha=None):` 

And I thought I had to implement the `_predict_quantiles` method instead the `_predict_interval` 🙈 ! Sorry about the confusion. I will revert the changes! 🙏 !

juanitorduz:
@fkiraly I fixed the implementation and seems all test are passing (I checked locally `test_all_estimators` 😄

fkiraly:
> Thanks for your feedback! I think I got confused with your comment

Ohhhhhhh, really sorry! It was me, after all, creating the confusion here.
What I meant to copy was the line with the `_predict_interval` function signature, I wanted to say that the `coverage` parameter should be `None`, not a particular value. I accidentally copied the line with `_predict_quantiles` function signature.

fkiraly:
So, (cough,) getting back to whre we were, I think the `coverage` parameter default should be `None`, in line with the extension template specification. Otherwise, we are good to go.

juanitorduz:
I added the changes and the test are passing ✅

fkiraly (approved):
Indeed, looks good now!

Apologies again for causing confusion in the review.

Issue #2455, opened 2022-04-13
[BUG] ForecastingPipeline cannot be constructed due to erroneous type check


fkiraly:
The following `ForecastingPipeline` cannot be constructed due to what looks like an erroneous type check.

Expected behaviour is that it constructs without error.
Works on 0.11.0, but not on 0.11.2.

```python
from sktime.forecasting.compose import ForecastingPipeline, TransformedTargetForecaster
from sktime.transformations.series.compose import OptionalPassthrough
from sktime.transformations.series.boxcox import LogTransformer
from sktime.forecasting.fbprophet import Prophet
from sktime.transformations.series.detrend import Detrender
from sktime.forecasting.ets import AutoETS

pipe_prophet = ForecastingPipeline(steps=[
        ("logX", OptionalPassthrough(LogTransformer())),
        ("detrenderX", OptionalPassthrough(Detrender(forecaster = AutoETS()))),
        ("prophetforecaster", TransformedTargetForecaster(steps = [
            ("log", OptionalPassthrough(LogTransformer())),
            ("prophet", Prophet(yearly_seasonality='auto'))
        ]))  
])
```

fkiraly:
The cause of this is that `TransformedTargetForecaster` has, in theory, *two* scitypes, per the inheritance check.

These are not correctly returned by the `scitype` utility which has a bug.
The check in ForecastingPipeline` also erroneously assumes that an estimator can have only one scitype.

ngupta23:
FYI...I am getting this in pycaret as well after upgrading to 0.11.2

```python
from sktime.forecasting.naive import NaiveForecaster
from sktime.forecasting.compose import ForecastingPipeline, TransformedTargetForecaster

forecaster = TransformedTargetForecaster(
    steps=[
        ("model", NaiveForecaster(strategy="drift"))
    ]
)

pipe = ForecastingPipeline(
    steps=[("forecaster", forecaster)]
)
```

```python-traceback
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
[<ipython-input-15-5a138d05a5e2>](https://localhost:8080/#) in <module>()
     10 
     11 pipe = ForecastingPipeline(
---> 12     steps=[("forecaster", forecaster)]
     13 )
     14 # pipe.fit(y_train)

1 frames
[...]

PR #2456, opened 2022-04-13
[BUG] Fix type checking error due to pipeline type polymorphism when constructing nested pipelines 

fkiraly:
This PR fixes three interacting bugs, causing #2455:

* an error in the `scitype` utility where None would be returned instead of a list of scitype in a case where the passed estimator was type polymorphic (had multiple scitypes). This is fixed.
* an error in the forecasting pipelines constructor check that would consider another forecasting pipeline as type-less and hence raise an error, because forecasting pipelines were expected to be constructed with transformers or forecasters. This is fixed by changing the default behaviour of `scitype` to what was assumed (just one type is returned).
* Since the `scitype` would come back empty for the `TransformedTargetForecaster`, no scenarios would be found in testing, because retrieval was based on the `scitype` utility. Now, with the change, scenarios are retrieved, and more tests are run, which (without fix) would fail, since tests for transformers are also run due to the polymorphism. This is remedied by removing transformer inheritance from `TransformedTargetForecaster` for now.

This PR also adds:
* a test for constructing a nested pipeline to ensure the functionality/ability of nesting pipelines does not get broken in the future.
* functionality to the `scitypes` function that forces it to return only a single scitype. Ties are resolved by the order in which the scitypes appear in the register, which prefers forecasters above transformers. This is enabled by default, as it was assumed by all references.

### DIFF
--- a/sktime/forecasting/compose/_pipeline.py
+++ b/sktime/forecasting/compose/_pipeline.py
@@ -24,7 +24,7 @@ class _Pipeline(
 
     def _get_pipeline_scitypes(self, estimators):
         """Get list of scityes (str) from names/estimator list."""
-        return [scitype(x[1], coerce_to_list=True) for x in estimators]
+        return [scitype(x[1]) for x in estimators]
 
     def _get_forecaster_index(self, estimators):
         """Get the index of the first forecaster in the list."""
--- a/sktime/forecasting/compose/_pipeline.py
+++ b/sktime/forecasting/compose/_pipeline.py
@@ -24,6 +24,7 @@ class _Pipeline(
 
     def _get_pipeline_scitypes(self, estimators):
         """Get list of scityes (str) from names/estimator list."""
+
         def resolve_scitype(x):
             """Return forecaster, then transformer, then None, from scitype list."""
             if "forecaster" in x:
@@ -32,6 +33,7 @@ class _Pipeline(
                 return "transformer"
             else:
                 return None
+
         # first we get lists
         scitypes = [scitype(x[1], coerce_to_list=True) for x in estimators]
         # then we default to one scitype, forecaster if contained, else transformer
--- a/sktime/forecasting/compose/_pipeline.py
+++ b/sktime/forecasting/compose/_pipeline.py
@@ -585,8 +585,9 @@ class TransformedTargetForecaster(_Pipeline):
     ----------
     steps : list of sktime transformers and forecasters, or
         list of tuples (str, estimator) of sktime transformers or forecasters
-            must contain exactly one transformer
-        these are "blueprint" transformers, states do not change when `fit` is called
+            the list must contain exactly one forecaster
+        these are "blueprint" transformers resp forecasters,
+            forecaster/transformer states do not change when `fit` is called
 
     Attributes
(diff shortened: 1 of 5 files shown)


fkiraly:
@aiwalter, would you be available to review? This affects nesting of pipelines.

aiwalter (commented):
`    Parameters
    ----------
    steps : list of sktime transformers and forecasters, or
        list of tuples (str, estimator) of sktime transformers or forecasters
            must contain exactly one transformer
        these are "blueprint" transformers, states do not change when `fit` is called``

I think this must be "must exactly contain one forecaster"?

sktime/forecasting/compose/tests/test_pipeline.py
```diff
@@ -76,3 +75,34 @@ def test_skip_inverse_transform():
...
+    pipe = ForecastingPipeline(
+        steps=[
+            ("logX", OptionalPassthrough(LogTransformer())),
+            ("detrenderX", OptionalPassthrough(Detrender(forecaster=AutoETS()))),
+            (
+                "etsforecaster",
+                TransformedTargetForecaster(
+                    steps=[
+                        ("log", OptionalPassthrough(LogTransformer())),
+                        ("autoETS", AutoETS()),
```
aiwalter, line 99: I think the `OptionalPassthrough` is rudimentary here, so would remove it

fkiraly:
> I think this must be "must exactly contain one forecaster"?

Ah, yes, of course. What a silly mistake.

sktime/forecasting/compose/tests/test_pipeline.py
fkiraly: I wanted composites inside the pipeline, to make sure that nesting works

fkiraly: in this case, more nesting is better, even if it overcomplicates the pipeline

sktime/forecasting/compose/_pipeline.py
```diff
@@ -536,7 +535,7 @@ def _transform(self, X=None, y=None):
 #     return Zt
 
 
-class TransformedTargetForecaster(_Pipeline, _SeriesToSeriesTransformer):
+class TransformedTargetForecaster(_Pipeline):
```
aiwalter, line 538: dont we need to deprecate the transformer interface and point to the `TransformerPipeline`?

fkiraly: summary of our discussion on slack: the `transform` method is still there and functioning as before, so no deprecation is necessary. We may want to deprecate this in the future, and indeed replace by `TransformerPipeline`.

sktime/utils/_testing/scenarios_getter.py
```diff
@@ -41,10 +41,10 @@ def retrieve_scenarios(obj, filter_tags=None):
 
     Parameters
     ----------
-    obj : class or object, or string.
+    obj : class or object, or string, or list of str.
```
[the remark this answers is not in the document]
fkiraly: correct, class is not the same as object (instance of class), as far as I know, and here you can pass both

PR #2463, opened 2022-04-13
[ENH] remove error message on exogeneous X from DirRec reducer

fkiraly:
The `_DirRecReducer` did not comply with the generic forecaster interface, where exogeneous `X` should not raise an error (in the worst case it should be ignored). The reducer did raise an error.

This PR does not add use of exogeneous `X`, it simply makes the reducer interface compliant, by ignoring `X` instead of raising an error.

Surfaced via widening tests here: https://github.com/alan-turing-institute/sktime/pull/2462


fkiraly:
(approved as part of #2462)

PR #2466, opened 2022-04-14
[ENH] refactor `_predict_moving_cutoff` and bugfix, outer `update_predict_single` should be called

fkiraly:
This PR carries out a small refactor on `_predict_moving_cutoff`:

* removal of code blocks conditional on `return_pred_int=True`, these were never reached
* changing `_update_predict_single` call to `update_predict_single`, this was a bug, since this causes the cutoff to not be updated in `_update_predict`

### DIFF
--- a/extension_templates/forecasting.py
+++ b/extension_templates/forecasting.py
@@ -442,13 +442,7 @@ class MyForecaster(BaseForecaster):
 
     # todo: consider implementing this, optional
     # if not implementing, delete the method
-    def _predict_moving_cutoff(
-        self,
-        y,
-        cv,
-        X=None,
-        update_params=True,
-    ):
+    def _predict_moving_cutoff(self, y, cv, X=None, update_params=True):
         """Make single-step or multi-step moving cutoff predictions.
 
         Parameters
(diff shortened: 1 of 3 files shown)


fkiraly:
@aiwalter, @GuzalBulatova, I think I finally found out what was causing all those `test_update_predict_predicted_index` failures with the repeated time index!

The default implementation of `_predict_moving_cutoff` was calling the private `_update_predict` instead of the public `update_predict`, which would cause to not update the cutoff, and hence resulting in repeated index and/or forecasts all made for the same time point.

Would appreciate your review on this.

fkiraly:
PS: discovered by debugging this PR https://github.com/alan-turing-institute/sktime/pull/2458, a refactor of a composite estimator. It seems, in hindsight, that this would always break `test_update_predict_predicted_index` of composites...

aiwalter:
oh nice! Then I can try to get rid of `_update` in the `STLForecaster` and see if your solution works

aiwalter (approved):

fkiraly:
> oh nice! Then I can try to get rid of `_update` in the `STLForecaster` and see if your solution works

Your optimism is inspiring, as always.

PR #2471, opened 2022-04-14
[ENH] Fix _update from STLForecaster to handle update_params=False

aiwalter:
##### For all contributions

##### For new estimators


aiwalter:
@fkiraly seems the PR #2466 did not resolve the issue?

fkiraly:
This should do it!
https://github.com/alan-turing-institute/sktime/pull/2557

extension_templates/forecasting.py
```diff
@@ -22,9 +22,12 @@
 Mandatory implements:
     fitting         - _fit(self, y, X=None, fh=None)
     forecasting     - _predict(self, fh=None, X=None)
+    updating        - _update(self, y, X=None, update_params=True).
+                    - Only mandatory to for compositions, see e.g.  STLForecaster
```
fkiraly, line 26: I think this is misleading, people would ready this as "not mandatory if not a composition", or not know what a composition is. So I would recommend to remove this.
I would add explanatory text in the body instead, about `update_params=False`.

sktime/forecasting/trend.py
```diff
@@ -452,28 +452,42 @@ def _update(self, y, X=None, update_params=True):
...
+                seasonal_deg=self.seasonal_deg,
+                trend_deg=self.trend_deg,
+                low_pass_deg=self.low_pass_deg,
+                robust=self.robust,
+                seasonal_jump=self.seasonal_jump,
+                trend_jump=self.trend_jump,
+                low_pass_jump=self.low_pass_jump,
+            ).fit(inner_iter=self.inner_iter, outer_iter=self.outer_iter)
+
+            self.seasonal_ = pd.Series(self._stl.seasonal, index=y.index)
```
fkiraly, line 480: seems incorrect to me, because `y` is only the new data, not all data

fkiraly (changes requested):
Yes, this is the right fix imo!

I would kindly ask you to change the comments in the extension template - details like that should not be in the table of contents, but in the explanation of what to fill in in `_update`.

fkiraly:
this can't be replaced by the default because we have three components, which need to be updated by specific `y` (not by the one passed to `update`.

PR #2468, opened 2022-04-14
[BUG] remove `alpha` arg from `_boxcox`, remove private method dependencies, ensure scipy 1.8.0 compatibility

fkiraly:
This PR fixes https://github.com/alan-turing-institute/sktime/issues/1994, what I think is the last issue that prevents sktime from being compatible with `scipy` 1.8.0.

The fix is to remove the otherwise unused `alpha` arg from `_boxcox`, which allows to remove the `_boxcox_conf_interval dependency` entirely

PR #2474, opened 2022-04-14
[MNT] loosen strict upper bound on `scipy` to 1.9.0

fkiraly:
This PR loosens the strict upper bound on the `scipy` dependency to `scipy<1.9.0`.

This requires https://github.com/alan-turing-institute/sktime/pull/2468 which should be the last issue preventing the bump.


lmmentel:
It seems it has some of the same changes as #2468 ?

lmmentel:
Whenever we tinker with deps we also need to bump make the corresponding changes in the conda recipe https://github.com/conda-forge/sktime-feedstock. Would be great to have some automation around this :grin:

fkiraly:
> It seems it has some of the same changes as https://github.com/alan-turing-institute/sktime/pull/2468 ?

As I said, it relies on the other PR which fixes the bugs/problems that prevented us from loosening the bounds in the first place. Otherwise this would simply break `main`.

fkiraly:
seems to work, @lmmentel

fkiraly:
> Would be great to have some automation around this

Should we open an issue?

fkiraly:
@lmmentel, are you happy or do you have any requests?

lmmentel (approved):

aiwalter (approved):

PR #2476, opened 2022-04-14
Remove prob integration notebook 

eenticott-shell:
The probabilistic grid integration notebook is outdated and the important points regarding how it can be used are covered in the notebook used for pydata Berlin. https://github.com/sktime/sktime-tutorial-pydata-berlin-2022/blob/main/notebooks/2_proba_forecasting.ipynb


fkiraly:
also, it looks like it ended up in the wrong place 😄

fkiraly (approved):
Straightforward - notebook was added by accident and should not be there.

Issue #2433, opened 2022-04-10
[BUG] Aggregator incorrectly assumes that all hierarchy levels are named


fkiraly:
The `Aggregator` incorrectly assumes that all hierarchy levels are named, in contrast to the specification of `pd_multiindex_hier` which does not. The estimator breaks when receiving input where at least one level is not named.

#2432 proposes a temp workaround, but this should be:

* cleaned up so the inner logic of `Aggregator` no longer assumes it
* tested, so we ensure in all estimators that no level names are assumed. This is possibly a longer project since the `Panel` mtype `pd-multiindex-hier` until recently did assume the same.

FYI @ciaran-g

fkiraly:
might be a good first issue since the problem is very localized

ciaran-g:
I'm happy to make these changes 👍

benHeid:
This bug does not exist anymore. Sems to be fixed by #2479. Also the code generated by Claude Code is not able to reproduce the error anymore:

```python
import numpy as np
import pandas as pd
from sktime.transformations.hierarchical.aggregate import Aggregator

# Synthetic 2-level hierarchy + time index, with the second hierarchy level
# deliberately left unnamed, as allowed by the pd_multiindex_hier mtype spec.
dates = pd.date_range("2020-01-01", periods=3, freq="D")
idx = pd.MultiIndex.from_product(
    [["A", "B"], ["x", "y"], dates],
    names=["level_0", None, "time"],
)
df = pd.DataFrame({"value": np.arange(len(idx), dtype=float)}, index=idx)

print("Input index names:", df.index.names)
print("Input shape:", df.shape)

agg = Aggregator()
try:
    out = agg.fit_transform(df)
except Exception as e:
    raise AssertionError(
        f"Aggregator raised an exception on input with an unnamed hierarchy "
        f"level (index names={df.index.names!r}): {e!r}"
    ) from e

print("Output shape:", out.shape)
print("Output index names:", out.index.names)

assert out.shape[0] > df.shape[0], (
    "Aggregator did not add any aggregated '__total' rows to the output"
)

n_levels = out.index.nlevels
# rows where every non-time level equals '__total' are the fully-aggregated totals
mask = np.ones(len(out), dtype=bool)
for i in range(n_levels - 1):
    mask &= (out.index.get_level_values(i) == "__total")
total_rows = out.loc[mask].sort_index()

assert not total_rows.empty, "No fully-aggregated '__total' rows found in output"

expected_totals = (
[...]

PR #2479, opened 2022-04-14
Aggregator remove index naming

ciaran-g:
#### Reference Issues/PRs

This PR removes the index naming requirement in the code completely, after the workaround by @fkiraly. This is linked to issue #2433

Changes to the code to allow for non-named indexes and a bug fix as well for flattening single indexes in complicated hierarchies. I didn't implement the second bullet in #2433, but can give it a go with a pointer to start me off?

No

### DIFF
--- a/sktime/transformations/hierarchical/aggregate.py
+++ b/sktime/transformations/hierarchical/aggregate.py
@@ -71,11 +71,6 @@ class Aggregator(BaseTransformer):
         -------
         df_out : multi-indexed pd.DataFrame of Panel mtype pd_multiindex
         """
-        X = X.copy()
-        X_orig_idx_names = X.index.names
-        new_idx_names = [str(x) for x in range(len(X_orig_idx_names))]
-        X.index.names = new_idx_names
-
         if X.index.nlevels == 1:
             warn(
                 "Aggregator is intended for use with X.index.nlevels > 1. "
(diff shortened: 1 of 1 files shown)


fkiraly (changes requested):
This looks good!

Just to make sure, would you be so kind and add some tests to this, where you `fit_transform` with or without index naming, and then check that the result has the right index and columns?

ciaran-g:
@fkiraly just three tests

- the output of `fit_transform()` with named and unnamed indexes are equal
- `fit_transform()` fails when `__total` is already present in the index
- the `flatten_single_levels` argument for a fixed example returns the right number of unique indexes

fkiraly:
> `fit_transform()` fails when `__total` is already present in the index

I'm wondering about the behaviour here - what I would do if `__total` is already in the index is to recompute/overwrite, instead of raising an error?

fkiraly:
More generally, how about an `inverse_transform` that *removes* all levels that contain `__total`?

```python
for i in range(X.index.nlevels-1):
    X = X.drop(index=`__total`, level = i)
return X
```
(done)

ciaran-g:
Included the inverse transform now 👍 

Also, I made some minor changes to the panel `convert_to()` functions, because some of the inverse transformation tests were failing. The `"__total"` in the index was being overwritten to a number during conversion to `nested-univ` and so the transform -> inverse transform test wasn't returning the same object. Should be fine now (hopefully)...

```
from sktime.utils._testing.scenarios_getter import retrieve_scenarios
from sktime.datatypes import convert_to

# this was the test that failing
test = retrieve_scenarios(Aggregator)[4]
X = test.args["transform"]["X"]
agg = Aggregator()
Xt = test.run(agg, method_sequence=["fit", "transform"])

X_long = convert_to(X, to_type="pd-multiindex")
Xt_long = agg.fit_transform(X_long)

# this is the problem, "__total" needs to be tracked to this index
convert_to(Xt_long, to_type="nested_univ")
```

sktime/datatypes/_panel/_convert.py
```diff
@@ -994,9 +994,9 @@ def from_dflist_to_multiindex(obj, store=None):
 def from_multiindex_to_dflist(obj, store=None):
 
     instance_index = obj.index.levels[0]
-    n = len(instance_index)
+    # n = len(instance_index)
```
fkiraly, line 997: please remove commented out code

fkiraly:
> in the index was being overwritten to a number during conversion to nested-univ

Good find! Indeed, the conversion was incorrectly assuming integer indices.
In case it is not too much of a hassle, would it be possible to include one example in `_panel._examples` to prevent this from happening in the future? Perhaps worth doing this in a separate PR as it may dredge up further bugs.
Perhaps one example that, in its `pd-multiindex` incarnation, has no index level names, and the time index is datetime?

sktime/transformations/hierarchical/aggregate.py
```diff
@@ -41,8 +42,8 @@ class Aggregator(BaseTransformer):
             "pd_multiindex_hier",
         ],
         "y_inner_mtype": "None",  # which mtypes do _fit/_predict support for y?
-        "capability:inverse_transform": False,  # does transformer have inverse
-        "skip-inverse-transform": True,  # is inverse-transform skipped when called?
+        "capability:inverse_transform": True,  # does transformer have inverse
+        "skip-inverse-transform": False,  # is inverse-transform skipped when called?
```
fkiraly, line 46: I would leave the `skip-inverse-transform` as `True`.

Why: otherwise a `TransformedTargetPipeline` with the aggregator at the start will no longer produce forecasts of totals.

sktime/transformations/hierarchical/aggregate.py
```diff
@@ -63,97 +64,105 @@ def _transform(self, X, y=None):
 
         Parameters
         ----------
-        X : Panel of mtype pd_multiindex_hier
-            Data to be transformed
+        X : Series or Panel of mtype X_inner_mtype
```
fkiraly, line 67: I guess you want to say `X : pd.DataFrame`, with any number of row levels and not nested?

fkiraly (changes requested):
Great!

My only blocking change would be the `skip-inverse-transform` tag, I would leave this as `True`.

I also left some other comments.

ciaran-g:
Should be good to go now @fkiraly

fkiraly:
@ciaran-g, would you mind quickly checking whether https://github.com/sktime/sktime-tutorial-pydata-berlin-2022/blob/main/notebooks/3_hierarchical_global_forecasting.ipynb still runs with the changes?

ciaran-g:
all good 👍

Issue #2478, opened 2022-04-14
[ENH] `Differencer` NA handling parameter


fkiraly:
The first entry is being eaten by the `Differencer` in its current standard setting, which may cause user frustration, especially when combined with a pipeline (which is its "typical use"), see e.g., here: https://github.com/alan-turing-institute/sktime/issues/2452

We should add an NA handling parameter setting and make the default to fill in sth for the first value, e.g., a difference from an earlier seen training set, or the value 0 or sth.

fkiraly:
FYI @anthonygiorgio97

fkiraly:
solved by https://github.com/alan-turing-institute/sktime/pull/2487 and already released in 0.12.0

PR #2487, opened 2022-04-15
[ENH] Differencer NA handling - "fill zero" parameter

fkiraly:
From a pair programming session with @lbventur (at the community meetup).

This PR solves #2478, it upgrades `Differencer` with NA handling parameter `na_handling`, which allows, besides the two current behaviours, a third behaviour `fill_zero` that fills zero.

This parameter replaces the boolean `drop_na` parameter, and adds deprecation with replacement at 0.13.0, and move to a `fill_zero` default.

Why is the `fill_zero` parameter a better default? It results in a "same index" and "no NA" condition at the end, which is the most compatible out of the alternatives when used in a pipeline. See also discussion (and user frustration) here: https://github.com/alan-turing-institute/sktime/issues/2452 and here: https://github.com/alan-turing-institute/sktime/issues/2451

### DIFF
--- a/sktime/transformations/series/difference.py
+++ b/sktime/transformations/series/difference.py
@@ -135,7 +135,7 @@ class Differencer(BaseTransformer):
         # note: internally, we will use self._na_handling
         #   because we must never change input param saves for sklearn compatibility
         #   and because we need to "translate" the old "drop_na" arg
-        #   this could, in certain cases, overwrite the self. parameger
+        #   this could, in certain cases, overwrite the self. parameter
         #   and cause sklearn compatibility issues due to the point mentioned
 
         translate_arg = {True: "drop_na", False: "keep_na"}
--- a/sktime/transformations/series/difference.py
+++ b/sktime/transformations/series/difference.py
@@ -60,14 +60,20 @@ def _inverse_diff(Z, lag):
     return Z
 
 
+# understand syntax
+# design - new parameters? for handling input part
+#   there is already drop_na, but there isn't a "fill" option
+#   padding, or fill (zeros?)
+# perhaps bonus: ensure this also works out of the box for hierarchical (using pandas?)
+
+
 class Differencer(BaseTransformer):
     """Apply iterative differences to a timeseries.
 
     The transformation works for univariate and multivariate timeseries. However,
     the multivariate case applies the same differencing to every series.
 
-    Difference transformations are applied at the specified lags in the order
-    provided.
+    Difference transformations are applied at the specified lags in the order provided.
 
     For example, given a timeseries with monthly periodicity, using lags=[1, 12]
     corresponds to applying a standard first difference to handle trend, and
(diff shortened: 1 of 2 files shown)


sktime/transformations/series/difference.py
```diff
@@ -121,16 +126,51 @@ class Differencer(BaseTransformer):
...
+        #   this could, in certain cases, overwrite the self. parameter
+        #   and cause sklearn compatibility issues due to the point mentioned
+
+        translate_arg = {True: "drop_na", False: "keep_na"}
+
+        if drop_na is not None:
+            warn(
+                f"the drop_na parameter is deprecated and will be removed in 0.13.0, "
+                f'use na_handling="{translate_arg[drop_na]}" instead'
+            )
```
aiwalter, line 147: should we not use `DeprecationWarning`? We might have to add that to our deprecation doc as well

fkiraly: yes, you are right. Might not be too clear in the deprecation section what one needs to do if the decorator cannot be used.

fkiraly: added the `DeprecationWarning`

sktime/transformations/series/difference.py
```diff
@@ -213,8 +254,19 @@ def _transform(self, X, y=None):
...
+
+        na_handling = self._na_handling
+        if na_handling == "drop_na":
             Xt = Xt.iloc[self._cumulative_lags[-1] :]
+        elif na_handling == "fill_zero":
+            Xt.iloc[: self._cumulative_lags[-1]] = 0
+        elif na_handling == "keep_na":
+            pass
+        else:
+            raise RuntimeError(
```
aiwalter, line 266: hm is that really a `RuntimeError`? Maybe just `AttributeError`?

fkiraly: well, it's an unexpected condition, just a very defensive error.
If it's reached, it's unclear what it would be imo.

chrisholder: This is a ValueError not RuntimeError. From the docs a value error is:
'Raised when an operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as IndexError'.

chrisholder (approved):

sktime/transformations/series/difference.py
lmmentel: I agree with @chrisholder

sktime/transformations/series/difference.py
```diff
@@ -121,16 +126,52 @@ class Differencer(BaseTransformer):
...
+        self.na_handling = self._check_na_handling(na_handling)
+        # note: internally, we will use self._na_handling
+        #   because we must never change input param saves for sklearn compatibility
+        #   and because we need to "translate" the old "drop_na" arg
+        #   this could, in certain cases, overwrite the self. parameter
+        #   and cause sklearn compatibility issues due to the point mentioned
+
+        translate_arg = {True: "drop_na", False: "keep_na"}
+
+        if drop_na is not None:
```
lmmentel, line 143: following sklrean we should not have any logic in `__init__` just argument assignment, all logic should happen in fit

sktime/transformations/series/difference.py
fkiraly: ok, changing to ValueError

sktime/transformations/series/difference.py
fkiraly: I will open an issue that highlights the problem

fkiraly: @lmmentel, issue opened here:
https://github.com/alan-turing-institute/sktime/issues/2573

I interpret the fact that you did not make a change request as that this is non-blocking.

PR #2489, opened 2022-04-15
[BUG] ensure row index names are preserved in hierarchical forecasting when vectorizing

fkiraly:
Fixes https://github.com/alan-turing-institute/sktime/issues/2434.

There were two reasons for the issue:

* row index name overwrite was turned off in the `BaseForecaster` class, although this was available as an option in `reconstruct`
* `reconstruct` did not overwrite the index when asked to do so, since it did not assign the result of a (non-inplace) `set_names` to the index


ltsaprounis (approved):
Nice!

PR #2491, opened 2022-04-16
[ENH] Distance alignment paths

chrisholder:
#### Reference Issues/PRs

Implements distance paths to distances.

This pr adds functionality to allow for the distance path to be retrieved along with the distances. Additionally it adds some
plotting to contrib.

The pr adds an addition distance_path callable where a tuple is returned containing (path, distance).

No

##### For all contributions

##### For new estimators


fkiraly:
@chrisholder, is a "distance path" the same thing as an alignment?

chrisholder:
Yea it is @fkiraly but it's done for every distance we have so your wrapper can go around it if you want

chrisholder:
So for example for dtw the path is given in white here:
dtw_path_through_costmatrix

chrisholder:
So alignment or path Im happy to rename it or change ti to whatever you think is best!

fkiraly:
Should we call it alignment path for consistency?

Also, can we expose these as alignment algorithms?

chrisholder:
We can expose them however you like it works the same way as distances. 

Any of these work

```

from sktime.distances import distance_path, distance_path_factory, dtw_path

x = np.array([[1], [2], [3], [4]])
y = np.array([[5], [6], [7], [8]])
path, dist = distance_path(x, y, metric='dtw')

callable = distance_path_factory(x, y, metric='dtw)
path, dist = callable(x, y)

path, dist = dtw_path(x, y)

```

fkiraly:
sure, works for me.
I´d call this an "alignment path".

TonyBagnall (commented):
distance_paths: "Dtw cost matrix to find dtw path through." is it just dtw?

TonyBagnall (commented):
image
?

TonyBagnall (commented):
could you add some explanation for what resolve_metric does? There is also a block of commented out code

chrisholder:
Made all these changes ready for re-review @fkiraly @TonyBagnall

sktime/distances/_distance_alignment_paths.py
```diff
@@ -0,0 +1,118 @@
...
+                alignment.append((i - 1, j - 1))
+            elif score == 1:
+                alignment.append((i - 1, j))
+            else:
+                alignment.append((i, j - 1))
+    return alignment[::-1]
+
+
+@njit(cache=True)
+def compute_lcss_return_path(
```
TonyBagnall, line 68: is this just for LCSS, or will it work for ERD and EPR also? Could you expand the comment to explain the difference to compute_min_return_path?

sktime/distances/_edr.py
```diff
@@ -112,19 +212,17 @@ def _edr_cost_matrix(
     bounding_matrix: np.ndarray,
     epsilon: float,
 ):
-    """Compute the edr cost matrix between two time series.
+    """Compute the edr cost matrix between two timeseries.
```
TonyBagnall, line 215: time series not timeseries ... or if you really dont like time series, use time-series, but there is no word timeseries

sktime/distances/_numba_utils.py
```diff
@@ -135,10 +137,10 @@ def to_numba_pairwise_timeseries(x: np.ndarray) -> np.ndarray:
     num_dims = _x.ndim
     if num_dims == 1:
         shape = _x.shape
-        _x = np.reshape(_x, (1, 1, shape[0]))
+        _x = np.reshape(_x, (shape[0], 1, 1))
```
TonyBagnall, line 140: I think the old version of this is correct? This is converting it to (m,1,1), should it not be (1,1,m)?

sktime/distances/_numba_utils.py
```diff
@@ -135,10 +137,10 @@ def to_numba_pairwise_timeseries(x: np.ndarray) -> np.ndarray:
...
     elif num_dims == 2:
         shape = _x.shape
-        _x = np.reshape(_x, (1, shape[1], shape[0]))
+        _x = np.reshape(_x, (shape[0], 1, shape[1]))
```
TonyBagnall, line 143: if num_dims == 2 are we assuming this is (n,m), (d,m) or (m,d)? this seems to convert (n,m) to (n,1,m) and (d,m) to  (d,1,m)?

TonyBagnall (changes requested):
I dont think to_numba_pairwise_timeseries is correct in dimensions, it seems to be converting to (m,d). Could you clarify in the doc string the exact assumptions about the input dimensions for this function?

sktime/distances/_numba_utils.py
chrisholder: I've merged this function into _make_3d_series and better documented why it is reshaping in this way.

TonyBagnall (approved):

PR #2495, opened 2022-04-16
[ENH] nonparallel WindowSummarizer with timedelta 

GrzegorzRut:
same as #2424 but with disabled parallel computation


fkiraly:
Looks like this prevents the issues from happening in the other PR!
Interesting, so it has to do with the parallelization.

There are two genuine test failures now.

GrzegorzRut:
Work continued in [#2424](https://github.com/alan-turing-institute/sktime/pull/2424)

PR #2497, opened 2022-04-17
[ENH] speed up Facebook Prophet tests

fkiraly:
This PR tries to speed up Facebook `Prophet` tests.

Changes in the test settings:
* reduced number of "uncertainty samples" from 1000 to 10


aiwalter (approved):
yes its faster

PR #2500, opened 2022-04-18
[ENH] metrics rework part II - metrics internal interface refactor

fkiraly:
This PR is a refactor of the current (point forecasting) metrics interface.

* introduces a new intermediate base class `BaseForecastingErrorMetric`, with changes as below. Concrete metrics now inherit from this base class.
* adds `evaluate` and `evaluate_by_index` methods, localizes the metric specific logic in `_evaluate`, and adds conversion (to internal `pandas.DataFrame`) in the boilerplate layer between public and inner methods
* the existing `__call__` dunder functionality remains unchanged from the public side, but defaults to `evaluate` on the inner side
* the logic of passing `multioutput` through the init and then passing it to the metric remains unchanged
* a currently unused parameter `multilevel` is introduced for averaging behaviour over hierarchies
* a further intermediate class `BaseForecastingErrorMetricFunc` is introduced, for interfacing the `numpy` based metrics. This will be the target of a combination of further refactor and change, in which the highly repetitive underscore mixins will be folded into this single class.
* concrete metrics remain unchanged, except for inheriting from `BaseForecastingErrorMetricFunc` throughout.
* mixins are changed from overriding `__call__` to overriding `_evaluate`, in line with the move above of specific logic into the inner method `_evaluate`
* a deprecation message is introduced for the `func` and `name` args in `BaseMetric` - these should not be constructor parameters, at base level, but handled by the intermediate class `BaseForecastingErrorMetricFunc`
* the `greater_is_better` attribute has been moved to a tag `lower_is_better` (flipped since lower is almost always better). The old attribute still works, and reads the tag directly. If the old `greater_is_better` attribute is called, a deprecation warning is raised.

This PR uses the tests in https://github.com/alan-turing-institute/sktime/pull/2496.
The tests certify that the public interface has not changed.
The tests are further extended to ensure that the pre-existing `__call__` dunder interface is identical to the (non-dunder) `evaluate` interface.

Also fixed a bug in a docstring example:
* one of the inputs of `MeanSquaredScaledError` was a list (an `np.array` was missing). This is outside the specified inputs, and the old interface did not pick this up since the `sklearn` checker did some conversion. Now an error is raised instead for non-supported input types, like lists of integers, which surfaced the problem.


fkiraly:
FYI @RNKuhns if you are still following sktime

sktime/performance_metrics/forecasting/_classes.py
```diff
@@ -68,45 +73,351 @@ class _BaseForecastingErrorMetric(BaseMetric):
...
+        return y_true, y_pred_inner, multioutput, multilevel
+
+
+class BaseForecastingErrorMetricFunc(BaseForecastingErrorMetric):
+    """Adapter for numpy metrics."""
+
+    def _evaluate(self, y_true, y_pred, **kwargs):
+        """Evaluate the desired metric on given inputs."""
+        multioutput = self.multioutput
+        res = self.func(y_true=y_true, y_pred=y_pred, multioutput=multioutput, **kwargs)
```
eenticott-shell, line 419: Is this the only time the deprecated `func` is ever used?

eenticott-shell (commented):
I think the rework makes sense and this allows for greater flexibility in the `metric` class. Also brings it more in line with the structure that I have used in probabilistic metrics which will clear up confusion. Only a few minor comments about potential additional options/missing docstring.

sktime/performance_metrics/forecasting/_classes.py
```diff
@@ -68,45 +73,351 @@ class _BaseForecastingErrorMetric(BaseMetric):
...
+            (n_outputs,), default='uniform_average'
+        Defines whether and how to aggregate metric for across variables.
+        If 'uniform_average' (default), errors are mean-averaged across variables.
+        If array-like, errors are weighted averaged across variables, values as weights.
+        If 'raw_values', does not average errors across variables, columns are retained.
+
+    multilevel : {'raw_values', 'uniform_average'}
+        Defines how to aggregate metric for hierarchical data (with levels).
+        If 'uniform_average' (default), errors are mean-averaged across levels.
+        If 'raw_values', does not average errors across levels, hierarchy is retained.
```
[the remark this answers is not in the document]
fkiraly: yes, I thought about that... I think this is something for part V or part VI 😄

sktime/performance_metrics/forecasting/_classes.py
```diff
@@ -68,45 +73,351 @@ class _BaseForecastingErrorMetric(BaseMetric):
...
+            warn(
+                "y_pred and y_true do not have the same column index. "
+                "This may indicate incorrect objects passed to the metric. "
+                "Indices of y_true will be used for y_pred."
+            )
+            y_pred = y_pred.copy()
+            y_pred.columns = y_true.columns
+
+        # check multioutput arg
+        allowed_multioutput_str = ("raw_values", "uniform_average", "variance_weighted")
```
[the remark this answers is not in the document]
fkiraly: yes, indeed, good spot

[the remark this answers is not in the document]
fkiraly: hm, it´s not actually allowed internally. I should probably remove it here instead

sktime/performance_metrics/base/_base.py
```diff
@@ -15,14 +16,53 @@ class BaseMetric(BaseObject):
...
-        func,
+        func=None,
         name=None,
     ):
+
+        if func is not None or name is not None:
+            warn(
+                "func and name parameters of BaseMetric are deprecated from 0.12.0"
+                "and will be removed in 0.13.0"
+            )
```
ltsaprounis, line 30: Shouldn't this have the deprecation warning class in it? 
(Same for all other deprecation warnings in the PR)

```python 
from warnings import warn, DeprecationWarning

warn(
    "func and name parameters of BaseMetric are deprecated from 0.12.0"
    "and will be removed in 0.13.0", 
    DeprecationWarning
)
```

fkiraly: yes, it should!

fkiraly: I always forget that...

Issue #2443, opened 2022-04-11
[ENH] Discretizer transformer for post-processing


aiwalter:
**Is your feature request related to a problem? Please describe.**
In real world, many business cases need a discrete forecast and no float numbers, but all of  `sktime` forecasters we have return float predictions.

**Describe the solution you'd like**
Implment a transformer `Discretizer` with a param to round up or round down. 

**Describe alternatives you've considered**

**Additional context**
Post-processing

fkiraly:
sounds like a good first issue?

fkiraly:
Should we also add a parameter "round to next multiple of", or a parameter that contains a finite set of values to round to?

fkiraly:
Might also be worth allowing to encode intervals somehow, e.g., round to the next points contained in a `pd.IntervalIndex`

AJarman:
Hi @fkiraly i'd like to pick this up please - it looks like it would be just implementing the `BaseTransformer._fit()` method?
My suggested implementation is
- `BaseTransformer`  child class
- _fit() method to round to integers (initial requirement for discrete values) as default.
- Constructor /`__init__` taking optional arguments : 
  - Round to dp `round_to_dp:Optional[int]=None`
  - Round to multiple of: `round_to_multiple:Optional[Union[float,int]]=None` 
  - List of finite values to round to `round_to_list:Optional[Union[float,int]]=None`
 
I am a little confused by the `pd.IntervalIndex` but we can clear that up later 👍

fkiraly:
@AJarman, sure!

I'm not sure though why you want to implement the `_fit` method. Should the rounding not happen in the `_transform` method? There's nothing to fit as far as I can see.

Probably just a typo of yours? Otherwise your explanation looks good.

Re `pd.IntervalIndex`, just ignore that.

AJarman:
Hello,
Yes I have started the implementation, I meant `_transform` of course 😄  that was a strange typo.

I am assuming that we don't want to store the old (untransformed/rounded) values and return them in `_inverse_transform`?
My understanding is that this would be the last step in a `TransformedTargetForecaster` pipeline and therefore the only method used is actually `_transform()`?

fkiraly:
> I am assuming that we don't want to store the old (untransformed/rounded) values and return them in `_inverse_transform`?

That wouldn't work, since in general we'd be seeing new indices.

[...]

PR #2501, opened 2022-04-18
[ENH] Rounding forecasts / Discretizer

AJarman:
##### For all contributions

##### For new estimators


sktime/transformations/series/rounding.py
```diff
@@ -0,0 +1,303 @@
...
+                type(i) for i in self.round_to_list if not isinstance(
+                    i, (int, float))]
+            raise ValueError(
+                f"""
+                round_to_list: {self.round_to_list}, not all elements are integer or float.
+                {wrong_types} were found in the list.
+                """)
+
+    def _fit(self, X, y=None):
+        return super()._fit()
```
fkiraly, line 158: unnecessary, and will likely break, since `_fit` has no args. Just use inheritance.

sktime/transformations/series/rounding.py
```diff
@@ -0,0 +1,303 @@
...
+        "scitype:transform-input": "Series",
+        # what is the scitype of X: Series, or Panel
+        "scitype:transform-output": "Series",
+        # what scitype is returned: Primitives, Series, Panel
+        "scitype:transform-labels": "None",
+        # what is the scitype of y: None (not needed), Primitives, Series, Panel
+        "scitype:instancewise": True,  # is this an instance-wise transform?
+        "capability:inverse_transform": False,  # can the transformer inverse transform?
+        "univariate-only": True,  # can the transformer handle multivariate X?
+        "X_inner_mtype": "pd.Series",  # which mtypes do _fit/_predict support for X?
```
fkiraly, line 56: use `pd.DataFrame`

sktime/transformations/series/rounding.py
```diff
@@ -0,0 +1,303 @@
...
+    _tags = {
+        "scitype:transform-input": "Series",
+        # what is the scitype of X: Series, or Panel
+        "scitype:transform-output": "Series",
+        # what scitype is returned: Primitives, Series, Panel
+        "scitype:transform-labels": "None",
+        # what is the scitype of y: None (not needed), Primitives, Series, Panel
+        "scitype:instancewise": True,  # is this an instance-wise transform?
+        "capability:inverse_transform": False,  # can the transformer inverse transform?
+        "univariate-only": True,  # can the transformer handle multivariate X?
```
fkiraly, line 55: why that? Rounding works for multivariate, too

sktime/transformations/series/rounding.py
AJarman: I had actually just left these in as the linter was complaining about abstract classes... sorry about that.

sktime/transformations/series/rounding.py
AJarman: Altered as suggested.

AJarman:
@fkiraly I will aim to complete the unit tests tomorrow or Wednesday.
My only query is about the dataframe compatibility with the function `round_to_nearest_value()` : should we add the ability to optionally pass a string-keyed dict to `round_to_list` for multivariate uses as below?
`{
"columnname1":[1,2,3,5,8,13],
"columnname2":[1,2,5,13],
}`
please can you link this PR to #2443 


Many Thanks for quick review and feedback.

fkiraly:
> should we add the ability to optionally pass a string-keyed dict to

sounds like a good idea to me

fkiraly:
@AJarman, I fixed the linting issues, but the discretizer is failing tests.
Would appreciate if you debug, by running this:

```python
from sktime.transformations.series.rounding import Discretizer

check_estimator(Discretizer)
```

fkiraly:
I'd say, let's keep this open, so we don't lose the partial work.
Perhaps someone else wants to finish it.

PR #2505, opened 2022-04-19
[ENH] new `_make_panel` utility (separate from `_make_panel_X`) with arbitrary return mtype

fkiraly:
This PR refactors the `_testing.panel` module, with the aim to bring the data generation in line with series and hierarchical, and slowly phase out `nested_univ` mtype to replace by the `pd-multiindex` mtype.

The PR introduces a `_make_panel` utility which has an `return_mtype` arg.
The logic from `_make_panel_X` is moved to  `_make_panel`, with the old interface retained.

Relies on the following PR that fixes some test breaks:
* https://github.com/alan-turing-institute/sktime/pull/2506


fkiraly:
(approved as part of #2508)

PR #2506, opened 2022-04-19
[ENH] Legacy test refactor - move `test_data_processing`, mtype handling in `test_classifier_output`

fkiraly:
This PR makes changes to legacy tests:

* moves `test_data_processing` into `datatypes` since it tests only functions in there (not in `datasets`)
* changes the logic in `test_classifier_output` that obtains number of instances in `X_new` - to make this compatible with arbitrary mtype, uses the `check_is_scitype` function


fkiraly:
(approved as part of #2508)

PR #2508, opened 2022-04-19
[ENH] Tests for multiple classifier input mtypes

fkiraly:
This PR introduces a test which passes the four main input mtypes plus `numpyflat` to classifiers (one that uses `numpy3D` internally, one that uses `nested_univ` internally).

This checks `BaseClassifier` functionality, integrated with the boilerplate and input conversions and checks.

Relies on https://github.com/alan-turing-institute/sktime/pull/2505 to easily generate panel data of different mtypes.
The only change above #2505 is that to `classification.tests.test_base`.


sktime/classification/tests/test_base.py
```diff
@@ -281,3 +283,21 @@ def _create_unequal_length_nested_dataframe(cases=5, dimensions=1, length=10):
         testy["dimension_" + str(i + 1)] = instance_list
 
     return testy
+
+
+MTYPES = ["numpy3D", "pd-multiindex", "df-list", "numpy3D"]
```
TonyBagnall, line 288: numpy3D listed twice

TonyBagnall (changes requested):
I've always thought the making of example data is hugely convoluted, things being pulled in and out of pandas unnecessarily, at least this hides a lot of that. It doesnt seem to be testing 2d numpy though, perhaps thats just a typo wher enumpy3D is repeated?

sktime/classification/tests/test_base.py
fkiraly: yes, that was a typo, should have been `"numpyflat"`.

TonyBagnall (approved):

PR #2512, opened 2022-04-20
[BUG] Fix output to ColumnEnsembleForecaster with `pred_quantile`/`pred_interval`

### LINKED ISSUE #2511: [BUG] `ColumnEnsembleForecaster` adds extra level to probabilistic predictions output
**Describe the bug**
When using the `predict_quantile` or `predict_interval` methods on a `ColumnEnsembleForecaster` the resulting output contains an extra irrelevant level of just `"quantile"` or `"interval"` respectively. 

**To Reproduce**

```python
from sktime.forecasting.compose import ColumnEnsembleForecaster

import numpy as np

y = _make_series()
f = ColumnEnsembleForecaster(ThetaForecaster())
fh = np.arange(5)
f.fit(y)
f.predict_quantiles(fh)
```

**Expected behavior**
A dataframe of two levels, top level names variables and second level names quantiles, instead a redundant middle level is included just saying "quantile".

eenticott-shell:
#### Reference Issues/PRs
Fixes #2511.

Removes unnecessary additional levels in output to `pred_quantile` and `pred_interval`


fkiraly:
I would recommend, finish the tests, and then merge the test PR into this PR.
So we ensure that the bug picked up by the new tests is removed.

fkiraly:
> Where would be the best place to do this?

Ideally, `forecasting.tests.test_all_forecasters`.
I think there already exist two tests, `test_predict_interval`, and `test_predict_quantiles`, you may have to look at these to check why these did not pick up the issue with the `ColumnEnsembleForecaster`.

You may also like to consider a refactor, using `check_is_mtype` with the `Proba` mtypes (this should give you a full format check already, but the tests were written before that existed).

GuzalBulatova (commented):
Logic seems in place, but I would vote for adding tests for this particular behaviour. Great idea to add tests for pred_quantile/pred_interval in all forecasters. Could start with unit testing ColumnEnsemble.

eenticott-shell:
It turned out that all the correct testing infrastructure was in place, but there was no test parameters for the column ensemble that included probabilistic metrics so it was missed by the tests. I have added a new case for the column ensemble, we will need to ensure that all composite forecasters are tested with probabilistic inputs, currently pipeline is the only other one that has prob methods implemented and that already included a prob case.

fkiraly (approved):
Looks indeed like a fix, and tests now catch it.

I will merge once #2552 is resolved.

PR #2513, opened 2022-04-20
[BUG] fixed `ColumnEnsembleClassifier` handling of unequal lenght data

fkiraly:
This PR fixes two bugs in ColumnEnsembleClassifier` related to this report https://github.com/alan-turing-institute/sktime/issues/2312:

* tags were incorrectly set - `ColumnEnsembleClassifier` always thought it could not handle unequal length data, even if components supported that. This is now correctly inferred from the internal tags.
* the internal mtype was set to `numpy3D` which could handle only equal length data by definition of this data type. This was changed to the `pandas.DataFrame` based data types.

### DIFF
--- a/sktime/classification/compose/_column_ensemble.py
+++ b/sktime/classification/compose/_column_ensemble.py
@@ -4,7 +4,7 @@
 Builds classifiers on each dimension (column) independently.
 """
 
-__author__ = ["Aaron Bostrom"]
+__author__ = ["abostrom"]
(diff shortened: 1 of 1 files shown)


fkiraly:
FYI @TonyBagnall, @MatthewMiddlehurst, would appreciate a review.

TonyBagnall (approved):

TonyBagnall:
although I did wonder what an anytagis is ...
_anytagis_then_set

PR #2519, opened 2022-04-21
[ENH] cleaning up `_panel._convert` module

fkiraly:
This PR does some minor cleaning in the `_panel._convert` module which had much copy-pasted legacy code:

* removal of commented out code
* removal of heavy input checks (this should be done by the `check` functions if needed)
* added module docstring

It also removes redundant and outdated parts in the `loading_data` tutorial, which was importing a lot of now private functionality of the `datatypes` module. This is replaced by references to the datatypes tutorial which explains the multiple data formats.

### DIFF
--- a/examples/loading_data.ipynb
+++ b/examples/loading_data.ipynb
@@ -4,45 +4,20 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Loading and working with data in sktime\n",
+    "# Loading data into sktime\n",
     "\n",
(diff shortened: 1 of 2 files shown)


TonyBagnall (approved):

TonyBagnall (approved):
definitely an improvement, but would it be possible to show an example of loading from .ts/arff to numpy in addition to Loading from .ts file to pandas DataFrame? I think it helps remove a whole layer of complexity to new users and is much more scikit-like

fkiraly:
> efinitely an improvement, but would it be possible to show an example of loading from .ts/arff to numpy in addition to Loading from .ts file to pandas DataFrame?

Yes, I thought about the same!

I think the optimal end state would be a version of the loaders where you have an `mtype` argument and the `convert` business is hidden unde the hood. Then show one more cell where you show what the output format is.

I would have done that if I have had enough time (probably a fiddly 30-60minutes?), so instead I´ll open an issue to track this, how about that?

fkiraly:
wait, this already exists

fkiraly:
ok, added the conversion functionality.

PR #2520, opened 2022-04-21
[ENH] Added tests for MultiplexForecaster

miraep8:
#### Reference Issues/PRs

As mentioned in #2514 wrote some tests to see if `MultiplexForecaster` performs as expected.  

Essentially - run `ForecastingGridSearchCV`  on `MultiplexForecaster` and verify that the best forecaster from the grid search is also the one we select if we simply evaluated all the forecasters separately. 

*Note - I use a very small data set and only compare have 2 forecasters in `MultiplexForecaster`.  I did this to not make the test too slow - but it is not the most thorough test.  Hopefully it would still help catch some obvious issues.*

No

Any other obvious tests that should be added? Should we add more forecasters?  Does this test make sense to you? 

#### Any other comments?

@aiwalter I went ahead and added you as an author to this file.  I figured since you had a hand in co-creating `MultiplexForecaster` in the first place and also weighed in on #2514 it seemed reasonable.  However - please let me know if you would prefer if I removed you. 

#### PR checklist

### DIFF
--- a/sktime/forecasting/compose/tests/test_multiplex.py
+++ b/sktime/forecasting/compose/tests/test_multiplex.py
@@ -1,0 +1,60 @@
+#!/usr/bin/env python3 -u
+# -*- coding: utf-8 -*-
+# copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+"""Tests for MultiplexForecaster and associated dunders."""
+
+__author__ = ["aiwalter", "miraep8"]
+
+from sktime.datasets import load_shampoo_sales
+from sktime.forecasting.all import (
+    AutoETS,
+    ExpandingWindowSplitter,
+    ForecastingGridSearchCV,
+    MultiplexForecaster,
+    NaiveForecaster,
+)
+from sktime.forecasting.model_evaluation import evaluate
+from sktime.utils.validation.forecasting import check_scoring
+
+
+def _score_forecasters(forecasters, cv, y):
+    """Will evaluate all the forecasters on y and return the name of best."""
+    scoring = check_scoring(None)
+    scoring_name = f"test_{scoring.name}"
+    score = None
+    for name, forecaster in forecasters:
+        results = evaluate(forecaster, cv, y)
+        results = results.mean()
+        new_score = float(results[scoring_name])
+        if not score or new_score < score:
+            score = new_score
+            best_name = name
+    return best_name
+
+
+def test_multiplex():
+    """Test results of MultiplexForecaster.
+
+    Because MultiplexForecaster should essentially just be a framework for
+    comparing different models/selecting which model does best, we can check
+    that it performs as expected.
+    """
+    y = load_shampoo_sales()
+    forecasters = [
+        ("ets", AutoETS()),
+        ("naive", NaiveForecaster()),
+    ]
+    multiplex_forecaster = MultiplexForecaster(forecasters=forecasters)
+    forecaster_names = multiplex_forecaster.get_forecaster_names()
+    # check that get_forcaster_names performs as expected:
+    assert forecaster_names == [name for name, _ in forecasters]
+    cv = ExpandingWindowSplitter(start_with_window=True, step_length=12)
+    gscv = ForecastingGridSearchCV(
+        cv=cv,
+        param_grid={"selected_forecaster": forecaster_names},
+        forecaster=multiplex_forecaster,
+    )
+    gscv.fit(y)
+    gscv_best_name = gscv.best_forecaster_.selected_forecaster
+    best_name = _score_forecasters(forecasters, cv, y)
+    assert gscv_best_name == best_name
--- a/sktime/forecasting/compose/tests/test_multiplex.py
+++ b/sktime/forecasting/compose/tests/test_multiplex.py
@@ -3,18 +3,20 @@
 # copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
 """Tests for MultiplexForecaster and associated dunders."""
 
-__author__ = ["aiwalter", "miraep8"]
+__author__ = ["miraep8"]
+
+from sklearn.base import clone
 
 from sktime.datasets import load_shampoo_sales
-from sktime.forecasting.all import (
-    AutoETS,
+from sktime.forecasting.compose import MultiplexForecaster
+from sktime.forecasting.ets import AutoETS
+from sktime.forecasting.model_evaluation import evaluate
+from sktime.forecasting.model_selection import (
     ExpandingWindowSplitter,
     ForecastingGridSearchCV,
-    MultiplexForecaster,
-    NaiveForecaster,
-    ThetaForecaster,
 )
-from sktime.forecasting.model_evaluation import evaluate
+from sktime.forecasting.naive import NaiveForecaster
+from sktime.forecasting.theta import ThetaForecaster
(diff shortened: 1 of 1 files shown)


fkiraly (changes requested):
Thanks for this, @miraep8! This was indeed missing!

I would suggest is to add another, more confined check for consistency.
Because your test is already testing consistency of the multiplexer, integrated with grid search, i.e., not just of the multiplexer.

The more elementary consistency check would be: to check whether the `MultiplexForecaster` behaves exactly as the selected forecaster. I.e., calling `fit` then `predict` etc on one or more scenarios, on the `MultiplexForecaster`,  and on the wrapped forecaster, and comparing results to be equal. Caveats are randomness, i.e., a deterministic forecaster should be chosen or a random seed set.

fkiraly:
& I´m looking forward to use these tests to check correctness of https://github.com/alan-turing-institute/sktime/pull/2458.

I somehow overlooked there were no estimator specific tests for the multiplex forecaster, good catch!

miraep8:
Done! Since it seems there is some work currently being done on adding `random_state` to random forecasters like `AutoETS` I went ahead and limited myself to testing with `ThetaForecaster` and the `NaiveForecaster`.

sktime/forecasting/compose/tests/test_multiplex.py
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/env python3 -u
+# -*- coding: utf-8 -*-
+# copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+"""Tests for MultiplexForecaster and associated dunders."""
+
+__author__ = ["aiwalter", "miraep8"]
```
fkiraly, line 6: why is aiwalter an author? Did you take code from elsewhere?

sktime/forecasting/compose/tests/test_multiplex.py
```diff
@@ -0,0 +1,91 @@
...
+
+from sktime.datasets import load_shampoo_sales
+from sktime.forecasting.all import (
```
fkiraly, line 9: better not use `forecasting.all`, it has not been used in a while, and who knows what will happen with it.

In case you want to help fix the `all` import, have a look at this issue: https://github.com/alan-turing-institute/sktime/issues/1928 (until then I would recommend to not use it)

sktime/forecasting/compose/tests/test_multiplex.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+    from numpy.testing import assert_array_equal
+
+    y = load_shampoo_sales()
+    # Note - we select two forecasters which are deterministic.
+    forecaster_tuples = [
+        ("naive", NaiveForecaster()),
+        ("theta", ThetaForecaster()),
+    ]
+    forecaster_names = [name for name, _ in forecaster_tuples]
```
fkiraly, line 51: I think we need to be careful, there is a subtle point here:
objects are being passed by reference in python. So, both the `MultiplexForecaster` hyper-parameters and the `forecasters` list are pointing to the *identical* forecaster objects!
They should be cloned inside the `MultiplexForecaster`, but you could think up a situation where two bugs cancel each other out - i.e., if clone is not done properly.
I would recommend a deepcopy to have the two sides of the comparison, or separate construction from the class.

fkiraly (changes requested):
Nice, thanks!

One change request, related to ensuring that the two sides of the comparison or different objects and not references to the same object.

(kindly let me know if you think my reasoning is wrong here)

sktime/forecasting/compose/tests/test_multiplex.py
miraep8: Oof, my bad!  Very important point! Will be sure to fix that immediately, thanks for spotting!

sktime/forecasting/compose/tests/test_multiplex.py
miraep8: Sure, will fix this to be more specific with the imports!  Thanks for the heads up :)

miraep8: I actually think I may have copied that from the MutliplexForecaster example haha, so will be sure to go and fix it there as well. :)

sktime/forecasting/compose/tests/test_multiplex.py
miraep8: Sorry - chalk this up to me being new to collaborative coding/open source projects!  I wanted to ere on the side of giving as much credit as possible (as I mentioned in the **Any other comments** section above!) But no - the code including the unfortunate reference error is mine! (though I certainly did look at the example `MultiplexForecaster` code as well as other test scripts for inspiration!).

sktime/forecasting/compose/tests/test_multiplex.py
miraep8: Oh actually - I do remember this! I think I was being a bit lazy as I noticed that MultiplexForecaster does technically make a deep copy via sklearn.base.clone (see below)  But still, this code is ugly regardless and I will fix it. 

  ```
  def _set_forecaster(self):
        self._check_selected_forecaster()
        if self.selected_forecaster is not None:
            for name, forecaster in self.forecasters:
                if self.selected_forecaster == name:
                    self.forecaster_ = clone(forecaster)
```

fkiraly: > as I noticed that MultiplexForecaster does technically make a deep copy via sklearn.base.clone (see below)

Yes, I know - my point being, if that gets broken somehow (and it is not difficult to break this, with all the underscores or not-underscores floating around), then the test will no longer pick up if the tested property fails...

sktime/forecasting/compose/tests/test_multiplex.py
miraep8: Actually - if it all right with you I might actually fix the MultiplexForecaster input example in my upcoming `|` dunder PR.  As it stands it is all mixed up with those changes I was working on but didn't necessarily want to be a part of this PR.

sktime/forecasting/compose/tests/test_multiplex.py
miraep8: Ah, got it!  Sorry, and yes I totally agree.  Better to have code that is easier to break.  Thanks! Hopefully this resolves it for you!

sktime/forecasting/compose/tests/test_multiplex.py
fkiraly: so, unless @aiwalter objects, I´d just suggest to remove it.

This is not like academia where you have to put the department head, their cousin and their dog on the paper 😄 

Also, whoever is in the author´s list is the person to blame for bugs in the file 😁

sktime/forecasting/compose/tests/test_multiplex.py
fkiraly: sure, makes sense - generally it´s a good idea to keep PR localized (by file or by content)

sktime/forecasting/compose/tests/test_multiplex.py
miraep8: It has been removed!   Blame me for the bugs!! 🐛 🐛 🐛

fkiraly (approved):
Thanks!

Looks good and tests seem to pass.

PR #2521, opened 2022-04-21
[ENH] simplify `load_from_tsfile`, support more mtypes

fkiraly:
This PR simplifies `load_from_tsfile`:

* removes duplication with `load_from_tsfile_to_dataframe`
* adds support for more panel mtypes (all of them)
* replaces reliance on panel conversion module internals with calls to public `convert`

Relies on https://github.com/alan-turing-institute/sktime/pull/2517 fix.


TonyBagnall (changes requested):
it would be good if load_from_ts assumed the file extension .ts if none was given, so that these two worked equivalently

sktime/datasets/_data_io.py
```diff
@@ -347,102 +347,39 @@ def load_from_tsfile(
     ValueError if return_data_type = numpy3d but the data are unequal length series
     ValueError if return_data_type = numpy2d but the data are multivariate and/
     or unequal length series
-
     """
+    if return_data_type is None:
```
TonyBagnall, line 351: return_data_type now defaults to "nested_univ", so should this still be an option? User unlikely to pass None as an argument

fkiraly: we could deprecate the `None` arg, that would become effective in 0.13.0

TonyBagnall (approved):

PR #2527, opened 2022-04-21
[ENH] faster test parameter settings for `StatsForecastAutoARIMA`

fkiraly:
Faster test parameter settings for `StatsForecastAutoARIMA` in order to speed up the tests.
(lower orders for parameter search range)


aiwalter (approved):

fkiraly:
ok, the `truncate` parameter seems to break it? 😢 

@FedericoGarza, would you be able to have a quick look and come up with a test parameter setting that has good coverage but is fast to run? One fit with the current settings and the test data used everywhere takes 20 sec in the test suite, ranking among the slowest tests.

FedericoGarza:
Yes, sure :)

fkiraly:
any help, @FedericoGarza?

PR #2532, opened 2022-04-22
[MNT] added pytest timeout of 5 minutes

fkiraly:
This PR adds a dependency on `pytest-timeout` in the `dev` dependency set, and a timeout-fail of 5min on all individual tests.

This will address tests hanging for 6 hours such as in https://github.com/alan-turing-institute/sktime/issues/2448 (they will fail earlier, at the 5min bound)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -87,7 +87,7 @@ jobs:
           cp .coveragerc testdir/
           cp setup.cfg testdir/
           cd testdir/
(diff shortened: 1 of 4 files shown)


aiwalter:
Can we not have this flag in setup.cfg instead repeating it 4 times? Also other flags might go there

fkiraly:
> Can we not have this flag in setup.cfg instead repeating it 4 times? Also other flags might go there

I would agree with having all the flags in one place, but the duplication/multiplication is already present, e.g., with the `--cov` flag.

Hence, removing the duplication/multiplication is a scope separate from this PR.

aiwalter (approved):

PR #2533, opened 2022-04-23
[ENH] Logger update - `__init__` removal, private `log` attribute

fkiraly:
This PR makes the following changes to `_MockEstimatorMixin`:

* changes the name of the log attribute from `_log` to `__log`. That way, the attribute becomes hidden and cannot be affected directly.
* write logic is slightly changed, with `log` getting a setter that appends to the log (instead of overwritin it); `__init__` is removed entirely, first access creates the list and not the `__init__`.

Reason to make these changes:

* `__init__` clashes with linear inheritance structure of `sktime` estimators, in particular the pattern of passing args to `super` (which it blocks)
* clashes with `BaseObject` extension here: https://github.com/alan-turing-institute/sktime/pull/2531

Also adds the `utils` module to the ignore-list of the `all_estimators` retrieval utility, since (a) the test estimators are not actually ones a user will want to see, and (b) the combined mock/logger estimators change state in `predict` etc and fail some tests. But, for logger estimators a change of state in `predict` etc is actually intended (addition of the log).

### DIFF
--- a/sktime/registry/_lookup.py
+++ b/sktime/registry/_lookup.py
@@ -129,7 +129,7 @@ def all_estimators(
     import sys
     import warnings
 
-    MODULES_TO_IGNORE = ("tests", "setup", "contrib", "benchmarking")
+    MODULES_TO_IGNORE = ("tests", "setup", "contrib", "benchmarking", "utils")
 
     all_estimators = []
     ROOT = str(Path(__file__).parent.parent)  # sktime package root directory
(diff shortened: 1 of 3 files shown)


fkiraly:
FYI @ltsaprounis

ltsaprounis:
Thanks Franz - that looks good. 

My only feedback is that making the `log.setter` appending the new values is a bit confusing. I would prefer to have a method called `update_log` instead so that the normal python assigning operator `=` behaviour is not something that you wouldn't expect. 

This is not a blocker though, so approving it. 
But if you agree you can change it.

ltsaprounis (approved):

fkiraly:
ok, changed it to a method called `add_log_item`, is that what you meant?

ltsaprounis (approved):
yep! all good now

PR #2534, opened 2022-04-23
[ENH] Added FitInTransform transformer

aiwalter:
#### Reference Issues/PRs
#2267 #2382 

##### For all contributions

##### For new estimators


fkiraly (commented):
I'd put it in `transformations.compose` where the pipelines are, what do you think?

sktime/transformations/series/compose.py
```diff
@@ -430,3 +433,145 @@ def _check_is_pdseries(z):
...
+            if X_inner_mtype is list, _update must support all types in it
+            Data to update transformer with
+        y : Series or Panel of mtype y_inner_mtype, default=None
+            Additional data, e.g., labels for tarnsformation
+
+        Returns
+        -------
+        self: reference to self
+        """
+        self.transformer_.update(X=X, y=y)
```
fkiraly, line 546: I think `update` should be empty. The idea is for this to be a "stateless" version that is always applied to the batch.

sktime/transformations/series/compose.py
```diff
@@ -430,3 +433,145 @@ def _check_is_pdseries(z):
...
+        y : Series or Panel of mtype y_inner_mtype, default=None
+            Additional data, e.g., labels for transformation
+
+        Returns
+        -------
+        transformed version of X
+        """
+        self.transformer_ = clone(self.transformer)
+        self.transformer_.fit(X=X, y=y)
+        return self.transformer_.transform(X=X, y=y)
```
fkiraly, line 508: this will fail the "transform does not change state" test.

The way to avoid this would be to not write to `self` at all, just remove the `transformer_` attribute, and create the clone locally in `_transform`.

sktime/transformations/series/compose.py
```diff
@@ -8,6 +8,9 @@
 from sklearn.utils.metaestimators import if_delegate_has_method
 
 from sktime.transformations.base import BaseTransformer
+from sktime.transformations.series.exponent import (
+    ExponentTransformer as _ExponentTransformer,
```
fkiraly, line 12: estimators used only in the test params should be imported inside `get_test_params`

aiwalter:
> I'd put it in transformations.compose where the pipelines are, what do you think?

hm but then e.g. `OptionalPassthough` should be there as well? Do we actually still need the `series` and `panel` modules at all?

sktime/transformations/series/compose.py
aiwalter: but I need to write to `self` in order to enable `inverse_transform` 🤔

fkiraly: would `inverse_transform` not call `fit.inverse_transform`?

fkiraly:
> hm but then e.g. OptionalPassthough should be there as well? Do we actually still need the series and panel modules at all?

I think the division still makes sense, but it hasn't been really adhered to correctly. So I was considering removing it, but we haven't decided on it.

Either way, I think all compositors should go in `transformations.compose`, at least the new ones.
I was going to move the `OptionalPassthrough` there, too.

sktime/transformations/series/compose.py
aiwalter: what is `fit` here?

aiwalter:
also tests are passing? So state change seems to be fine?

sktime/transformations/series/compose.py
fkiraly: of the inner trafo

fkiraly:
>also tests are passing? So state change seems to be fine?

No, you need to test with an inner transformer where not `fit_is_empty`, otherwise the wrapper does nothing...

sktime/transformations/tests/test_fitintransform.py
```diff
@@ -0,0 +1,40 @@
...
+from sktime.transformations.series.exponent import ExponentTransformer
+from sktime.utils._testing.forecasting import make_forecasting_problem
+
+y, X = make_forecasting_problem(make_X=True)
+y_train, y_test, X_train, X_test = temporal_train_test_split(y, X)
+
+
+def test_transform_fitintransform():
+    """Test fit/transform against ExponentTransformer."""
+    fitintransform = FitInTransform(ExponentTransformer())
```
fkiraly, line 22: same here, you want to use the boxcox or similar transform that does not have empty `_fit`.
why not use `create_test_instance`?

fkiraly:
It appears that the tests were turned off since there were some estimators make state changes in `predict`, `transform`.

That is "cheating", I guess we turned it off in a refactor to the more strict interface, and didn't properly turn it on again.

Can you kindly ensure nothing writes to `self` or changes attributes of `self` in `transform`, while I fix the tests?

aiwalter:
I need to write to `self`, otherwise I  cant do `inverse_trasnform`. Also why is it actually not allowed to write to `self`?

fkiraly:
>I need to write to `self`, otherwise I  cant do `inverse_transform`. 

Can you explain that? I thought `inverse_transform` only requires `fit`, but not `transform` to have been run.

> Also why is it actually not allowed to write to `self`?

Because the state change design for all estimators allows state change only through `fit` and `update`.
This is also written down in the base class docstrings, and all estimators must adhere to that.

The entire logic is based around that assumption, so we can't change it on an ad-hoc basis.
It's one of the "core interface assumptions".

If you are asking for the rationale behind that:

* sklearn `BaseEstimator` interface makes the same assumptions
* it makes sense to localize state change to only a small number of methods. The more methods change state, the more possibilities you have to be mindful of in transitions. Worst case, you get combinatorial explosion of state change sequences which is really hard to control, test, and explain to a user.

aiwalter:
> Can you explain that? I thought inverse_transform only requires fit, but not transform to have been run.

well how can I do `self.forecaster.inverse_transform()` if this fitted `forecaster_` is not soted? If I just to `return self._forecaster_.fit_transform(X, y)` in `_transform()`, then I dont store this fit and then I cant do `inverse_transform`

fkiraly:
> well how can I do `self.forecaster.inverse_transform()` if this fitted `forecaster_` is not stored? If I just to `return self._forecaster_.fit_transform(X, y)` in `_transform()`, then I dont store this fit and then I cant do `inverse_transform` 

Why does `fit(X, y).inverse_transform(X, y)` not work?

sktime/transformations/tests/test_fitintransform.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python3 -u
+# -*- coding: utf-8 -*-
+# copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+"""Unit tests of Imputer functionality."""
```
fkiraly, line 4: this should say `FitInTransform`

sktime/transformations/compose.py
```diff
@@ -663,3 +664,137 @@ def _underscore_join(iterable):
...
+    >>> from sktime.transformations.compose import FitInTransform
+    >>> from sktime.transformations.series.impute import Imputer
+
+    >>> y, X = load_longley()
+    >>> y_train, y_test, X_train, X_test = temporal_train_test_split(y, X)
+    >>> fh = ForecastingHorizon(y_test.index, is_relative=False)
+
+    >>> # we want to fit the Imputer only on the predict (=transform) data.
+    >>> # note that NaiveForecaster cant use X data, this is just a show case.
+    >>> pipe = ForecastingPipeline(
```
fkiraly, line 705: I´d recommend a `ClassifierPipeline` becaues there the algorithm actually makes sense...

sktime/transformations/compose.py
```diff
@@ -663,3 +664,137 @@ def _underscore_join(iterable):
         """Create flattened column names from multiindex tuple."""
         iterable_as_str = [str(x) for x in iterable]
         return "__".join(iterable_as_str)
+
+
+class FitInTransform(BaseTransformer):
+    """Transformer compostion to always fit a given transformer on the transform data only.
+
+    Some transformers like Imputer(method="mean") might be more accurate for some use
```
fkiraly, line 672: I would recommend to be more precise here, along the following lines. It´s not about "accuracy", but you may want to apply the transformation to the same sample. This can make sense in classification, but usually doesn´t make sense in forecasting (since it can lead to information leakage).

sktime/transformations/compose.py
```diff
@@ -663,3 +664,137 @@ def _underscore_join(iterable):
         """Create flattened column names from multiindex tuple."""
         iterable_as_str = [str(x) for x in iterable]
         return "__".join(iterable_as_str)
+
+
+class FitInTransform(BaseTransformer):
+    """Transformer compostion to always fit a given transformer on the transform data only.
```
fkiraly, line 670: typo

fkiraly (approved):
Ok, good for me.

I´ve left some non-blocking, docstring related comments.

sktime/transformations/compose.py
aiwalter: sure we can add this as well, can you make a proposal? I am not so familiar with the classification module

fkiraly: `pipe = imputer * SummaryClassifier()`

aiwalter: I think we shouldnt use dunder methods in examples, if s.o. sees it first time it just looks confusing

fkiraly: so just pick one of the examples from `ClassificationPipeline` and replace the transformer

fkiraly:
@aiwalter, do you want to fix the docstrings before we merge?

sktime/transformations/compose.py
aiwalter: i dont find a class `ClassificationPipeline`?

sktime/transformations/compose.py
aiwalter: pls provide description you want to have

aiwalter:
its ready to be merged

fkiraly:
tests seem stuck, let´s restart

sktime/transformations/compose.py
```diff
@@ -663,3 +664,135 @@ def _underscore_join(iterable):
         """Create flattened column names from multiindex tuple."""
         iterable_as_str = [str(x) for x in iterable]
         return "__".join(iterable_as_str)
+
+
+class FitInTransform(BaseTransformer):
+    """Transformer composition to always fit a given transformer on the transform data only.
+
+    Some transformers like Imputer(method="mean") might be more accurate for some use
```
fkiraly, line 672:
```suggestion
    In panel settings, e.g., time series classification, it can be preferable (or, necessary) to fit and transform on the test set, e.g., interpolate within the same series that interpolation parameters are being fitted on.
    `FitInTransform` can be used to wrap any transformer to ensure that `fit` and `transform` happen always on the same series, by delaying the `fit` to the `transform` batch.
    Warning: use of `FitInTransform` will typically not be useful, or can constitute a mistake (data leakage) when naively used in a forecasting setting.
```

sktime/transformations/compose.py
fkiraly: done

sktime/transformations/compose.py
```diff
@@ -663,3 +664,142 @@ def _underscore_join(iterable):
...
+
+    In panel settings, e.g., time series classification, it can be preferable
+    (or, necessary) to fit and transform on the test set, e.g., interpolate within the
+    same series that interpolation parameters are being fitted on. `FitInTransform` can
+    be used to wrap any transformer to ensure that `fit` and `transform` happen always
+    on the same series, by delaying the `fit` to the `transform` batch.
+
+    Warning: The use of `FitInTransform` will typically not be useful, or can constitute
+    a mistake (data leakage) when naively used in a forecasting setting.
+    cases if they are fitted only on the transform/predict data. The FitInTransform
```
fkiraly, line 680: this looks like a formatting accident??

fkiraly (dismissed):
Happy to merge.

Looks like there is a formatting accident in the docstring, but happy to merge (and possibly fix) once tests pass again.

PR #2535, opened 2022-04-23
[MNT] Added pytest flags to setup.cfg

aiwalter:
#### What does this implement/fix?
Consolidation of `pytest` flags

##### For all contributions

##### For new estimators


Makefile
```diff
@@ -28,7 +28,7 @@ test: ## Run unit tests
 	mkdir -p ${TEST_DIR}
 	cp .coveragerc ${TEST_DIR}
 	cp setup.cfg ${TEST_DIR}
-	cd ${TEST_DIR}; python -m pytest -v -n auto --showlocals --durations=20 $(PYTESTOPTIONS) --pyargs $(PACKAGE)
+	cd ${TEST_DIR}; python -m pytest
```
aiwalter, line 31:
```suggestion
	cd ${TEST_DIR}; python -m pytest $ (PACKAGE)
```

aiwalter:
..

aiwalter:
@lmmentel @freddyaboulton could you two as our CI/CD experts pls have a look? 🙂

setup.cfg
```diff
@@ -12,6 +12,15 @@ addopts =
     --ignore examples
     --ignore docs
     --doctest-modules
+    --durations 10
+    --timeout 300
```
aiwalter, line 16:
```suggestion
    --timeout 600
```

aiwalter:
for some reason when I edit the `build_wheels.sh` file to do the changes there as well, I get a `pre-commit` failure of the `shellcheck`. Should I just ignore that?

lmmentel:
Great work, how about moving the configuration into `pyproject.toml`? `setup.py` is being flagged as legacy packaging backend so we might as well move the configuration.

setup.cfg
```diff
@@ -12,10 +12,21 @@ addopts =
     --ignore examples
     --ignore docs
     --doctest-modules
+    --durations 10
+    --timeout 600
+    --cov sktime
+    --cov-report xml
+    --cov-report html
+    --showlocals
+    --pyargs
```
lmmentel, line 21: I don't think this is needed since the invocation does not pass any arguments

setup.cfg
```diff
@@ -12,10 +12,21 @@ addopts =
...
+    --verbose
+    -n 2
 filterwarnings =
     ignore::UserWarning
     ignore:numpy.dtype size changed
     ignore:numpy.ufunc size changed
+testpaths =
```
lmmentel, line 28: this switches test discovery from package based to path based, have you verified that the same number of tests are discovered with and without this change?

setup.cfg
aiwalter: ok, I didnt know what `--pyargs` is used for, but let me then remove it

setup.cfg
aiwalter: yes I just verified that, its same

aiwalter:
I think deprecating `setup.cfg` can be done in a separate PR.

fkiraly (approved):
Looks good to me - thanks for consolidating this!

@lmmentel, are you fine with this?

lmmentel (approved):

setup.cfg
```diff
@@ -12,10 +12,21 @@ addopts =
     --ignore examples
     --ignore docs
     --doctest-modules
+    --durations 10
```
fkiraly, line 15: the comment is no longer accurate, "ignore certain folders and pytest warnings". Perhaps one comment per block of things?

PR #2538, opened 2022-04-23
BUG] bugfix for default `_predict_var` implementation

fkiraly:
The default `_predict_var` implementation had an unreported bug where it compared column names in the multivariate case.

This is now fixed.


fkiraly:
merging as approved as part of https://github.com/alan-turing-institute/sktime/pull/2539

PR #2539, opened 2022-04-23
[ENH] MockForecaster without logging, MockUnivariateForecaster clean-up

fkiraly:
This PR introduces a `MockForecaster` without logging, for testing the test utilities.

It also cleans up the `MockUnivariateForecaster`, including a name change to `MockUnivariateForecasterLogger`, to distinguish mocks that log and mocks that do not log.
`MockUnivariateForecaster` methods are also cleaned up, it had some unnecessary and duplicative assignments.

Finally, the PR makes the following changes to speed up tests:

* replace `NaiveForecaster` in testing of `check_estimator` with `MockForecaster`, this reduces the `check_estimator` run from 2-5min to 20sec.

The PR relies on:
* https://github.com/alan-turing-institute/sktime/pull/2538 , a bug which the testing of `MockForecaster` uncovered.

FYI @ltsaprounis

### DIFF
--- a/sktime/utils/estimators/__init__.py
+++ b/sktime/utils/estimators/__init__.py
@@ -3,7 +3,14 @@
 
 __author__ = ["ltsaprounis"]
 
-__all__ = ["MockUnivariateForecaster", "make_mock_estimator"]
+__all__ = [
+    "MockUnivariateForecaster",
+    "MockUnivariateForecasterLogger",
+    "make_mock_estimator",
+]
 
 from sktime.utils.estimators._base import make_mock_estimator
-from sktime.utils.estimators._forecasters import MockUnivariateForecaster
+from sktime.utils.estimators._forecasters import (
+    MockUnivariateForecaster,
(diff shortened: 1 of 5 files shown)


ltsaprounis:
Looks good! Maybe we can simplify things even more by removing `MockUnivariateForecasterLogger` completely. You can get exactly the same functionality with the following: 
```python
MockUnivariateForecasterLogger = make_mock_estimator(MockForecaster)
```

Fewer things to maintain this way!

fkiraly:
Tried and it broke a lot of things - perhaps best to keep things separate and later we refactor (as discussed on slack)

ltsaprounis (approved):
Looks good to me. 

As discussed in slack, there is some redundancy in `make_mock_estimator` and hardcoded logging in `MockUnivariateForecasterLogger`. In future refactoring, we will add a `MockUnivariateForecaster` without logging and replace `MockUnivariateForecasterLogger` with `make_mock_estimator(MockUnivariateForecaster)`.

Issue #2459, opened 2022-04-13
[ENH] Forecaster composition dunder methods


fkiraly:
Let's implement some dunder methods for forecaster and transformer composites (and, possibly, discuss design):

- [ ] `**` for applying a transformer to `X` into a forecaster, via `ForecastingPipeline`
- [ ]  `-` for `OptionalPassthrough` applied to a trasnformer
- [ ] `|` for `MultiplexForecaster`
- [ ] implement `MultiplexTransformer`
- [ ] `|` for `MultiplexTransformer`

miraep8:
I was interested by this suggestion,

> * [ ]  `**` for applying a transformer to `X` into a forecaster, via `ForecastingPipeline`

 though I was wondering if you could clarify what you were envisioning here as I wasn't sure exactly what you had in mind.  Was it something like this:

`new_pipe = trans**forecaster`

so that it is essentially as if we did this:

` new_pipe = ForecastingPipeline(steps=[trans, forecaster])`

or perhaps more something like this:

`new_pipe = trans**pipe`

where `pipe` would some pre-existing `ForecastingPipeline` (with n steps s.t. n >= 1) and we wished to prepend `trans` to its `steps` attribute?  (and which to me at least seems like it would be the easiest to implement)

Or perhaps something that supports both?  So one could do:

`new_pipe = trans_1**trans_2**trans_3**...**trans_n**forecaster`

In truth, I would be curious as to your thoughts on use cases as it seems like all the cases I listed could be achieved by creating a new `ForecastingPipeline` with the desired transformations/forecasters added as steps?  

Realize I might be off base entirely here, thanks for your patience if that is the case!

fkiraly:
> Or perhaps something that supports both? So one could do:

Yes, exactly!

This behaviour is actually already implemented for single `*` and the `TransformedTargetForecaster`, if you look at the dunder methods both in that class and in `BaseForecaster`. There are multiple behaviours, depending whether you are dealing with pipelines or not. The aim is to keep the pipeline structure flat, not nested. In total, there are four cases, depending on whether `self` or `other` is already a pipeline, or not:

If `*` is put between a transformer `a` and a forecaster `b`, 
* if neither of which are a pipeline, then a `TransformedTargetForecaster` is created with `steps=[a, b]`.
* if `a` is not a pipeline but `b` is, then `a` is prepended to the steps of `b` in a copy
[...]

PR #2496, opened 2022-04-17
[ENH] metrics rework part I - output format tests

fkiraly:
This PR is part I of a rework of metrics.

It introduces output format tests to enable testing in the refactoring process:
* the current `numpy` based functions in the `performance_metrics._functions` module
* the classes in the `performance_metrics._classes` module

This is necessary since refactoring should be test anchored, and the current tests for metrics did not test the output format.

Contains fix of a bug discovered by this:
* `geometric_mean_squared_error` and `geometric_mean_absolute_error` were inconsistent with the general intervace, missing `kwargs`. This was added.

PR #2540, opened 2022-04-23
[ENH] Added '|' dunder method to MultiplexForecaster

miraep8:
#### Reference Issues/PRs

Implements part of #2459. **Note this PR is dependent on PR #2458**

Have implemented a `|` dunder method for both `MultiplexForecaster`/`BaseForecaster`. Have also introduced new tests for this code.

Other things in this PR:
- added a small `_get_forecaster_names()` helper method to `_HeterogenousEnsembleForecaster` seemed useful to me!
- updated the example code in for using `MultiplexForecaster` so that it no longer imports from all.

No

In addition to the dunder behaving as expected - would be great to get feedback on how I decided to do the unique name handling.

Essentially a use case I was imagining was that a user might want to do this:
``
multiplex_forecaster = AutoETS() | NaiveForecaster() | ThetaForecaster()
``
or potentially even this:
``
multiplex_forecaster = NaiveForecaster(method="last") | NaiveForecaster(method="mean") | NaiveForecaster(method="drift")
``

In these cases we would want to generate unique names for them.   Would appreciate feedback on whether the approach I took to do this seems reasonable. (mostly I tried to leverage the `super()` methods which already existed and were relevant)

##### For all contributions


fkiraly (changes requested):
Nice!

First of all, I want to note that the "heterogeneous ensemble" estimators are a bit in flux currently, so feedback is appreciated.

Questions and/or change requests:

* I was intending to remove the `_HeterogeneousEnsembleForecaster` gradually, using only the `_HeterogenousMetaEstimator`. Do you think this is a good idea or not? It would avoid proliferation of the "intermediate" methods, like the `_get_forecaster_names` you added. Why not use `_get_estimator_names` directly?
* why did you decide to make an `_ensure_unique_names` method? This seems to duplicate `_get_estimator_tuples` which you can get from `_HeterogenousMetaEstimator` (in your case, with `clone_ests=True` arg). Am I missing something? Or did you miss the `_get_estimator_tuples` function?

sktime/forecasting/base/_base.py
```diff
@@ -170,6 +170,27 @@ def __rmul__(self, other):
...
+        other: `sktime` forecaster or sktime MultiplexForecaster
+
+        Returns
+        -------
+        MultiplexForecaster object
+        """
+        from sktime.forecasting.compose import MultiplexForecaster
+
+        # generate a name for our forecaster based on class name:
+        self_name = type(self).__name__.lower()
```
fkiraly, line 189: I think the `MultiplexForecaster` shouhld take care of generating the name - I would just do `MultiplexForecaster(self)`

sktime/forecasting/base/_base.py
```diff
@@ -170,6 +170,27 @@ def __rmul__(self, other):
...
+        # create wrap self in MultiplexForecaster, pass most of '|' logic to that class
+        multiplex_self = MultiplexForecaster([(self_name, self)])
+        return multiplex_self | other
```
fkiraly, line 192: we need to make a case distinction - you do the above if `other` is an instance of `BaseForecaster`, otherwise you need to return `NotImplemented` (see the other dunders)

sktime/forecasting/compose/_multiplexer.py
```diff
@@ -102,6 +103,78 @@ def _check_selected_forecaster(self):
...
+            other_forecasters = self._ensure_unique_names(other.forecasters)
+            new_multiplex_forecaster = MultiplexForecaster(
+                self.forecasters + other_forecasters
+            )
+            return new_multiplex_forecaster
+        # If is anyother type of forecaster, simply add it to forecasters:
+        elif isinstance(other, BaseForecaster):
+            other_name = type(other).__name__.lower()
+            other_tuple = self._ensure_unique_names([(other_name, other)])
+            self.forecasters.append(other_tuple[0])
```
fkiraly, line 170: here, I would construct a new `MultiplexForecaster`, instead of modifying the current one. Otherwise we get all kinds of weird side effects from partially fitted chains etc.

fkiraly: in all similar places as well

fkiraly:
Also, another note.

Overall, this is really nice work!

But, I think you seem to have probably not seen my advice where I said, that I have already been refactoring the `MultiplexForecaster` here: https://github.com/alan-turing-institute/sktime/pull/2458
and I have been using the "unique name" methods and done the integration with the new heterogeneous ensemble forecaster

I was also under the impression you would look at the `MultiplexTransformer`.

The two PR on the `MultiplexForecaster` are now, unfortunately, clashing.

Would you mind taking my PR https://github.com/alan-turing-institute/sktime/pull/2458 as a starting point, and try to reconcile it with the dunder methods you have added?

Sorry for the hassle, I should have been more clearer.

And, please let this not discourage you!
Your PR is one of the very highest quality pieces of work on `sktime` I've seen, I am serious.

sktime/forecasting/base/_base.py
miraep8: So - two things I would appreciate some clarity on!  
1) I actually intententionally had the code return a `ValueError` if other isn't a `BaseForecaster` or a `MultiplexForecaster` instead of `NotImplemented`. (My impression is that [they plan to phase out NotImplemented](https://docs.python.org/3/library/constants.html#NotImplemented), so I decided to just use an error instead).  But given that would you prefer I stick with the `NotImplemented` all the same?   
2) is your issue that there are not checks (as there are - they are just wrapped up in the `MultiplexForecaster` `__or__`) or that you want them in 2 places?  I can understand the logic for the second as it makes it clearer what is going on at all stages, but does lead to code bloat a bit too..... Hence the clarification question :)

miraep8: Whoops!  Just kidding - I think I misread the python docs page!  Just went back and reread it!  I guess they are just depreciating the use of it in a boolean context not all together!

miraep8: In which case I am happy to stick with it since that is what the other dunders use!  Though personally I might find a `ValueError` more useful haha.

sktime/forecasting/compose/_multiplexer.py
miraep8: Sounds good will do!

sktime/forecasting/base/_base.py
miraep8: ok!

miraep8:
Hey!

Sorry! My bad, I had of course seen your PR #2458, it just hadn't initially seemed to me that we had too many clashes (at least in logic!) as believe the dunder method should work with your PR as well! (I tried my best not to touch any of the other functions in MultiplexForecaster so that hopefully we would avoid any direct conflicts!) I guess I also figured since you made your PR first it would probably get merged first and then I would handle all the conflicts on my end as needed!  I think I was perhaps unclear in my comments in #2459  - I had actually intended to first do the dunder for `MultiplexForecaster` to get familiar with how this works before implementing `MultiplexTransformer` to try to be consistent with how you guys have built these objects already!

As for the base classes - I am more than happy to switch over to using all the base classes in `_HeterogenousMetaEstimator` instead of `_HeterogenousMetaForecaster`.  The `_get_forecaster_names()` method was just a convenience to not have to type `multiplexer._get_estimator_names(multiplexer.forecasters)`, but I can definitely see the logic with getting away with `_HeterogenousMetaForecaster` entirely - at the moment it doesn't seem to add too much! I will do a commit with these changes later this evening!  I had actually somehow missed the `get_estimator_tuples` method, so in that refactor I will revisit it! 

Long story short - sorry about the clashes, I thought we were on the same page! Lets just chalk it up to an unfortunate miscommunication! I would certainly be happy to handle any merge conflicts that arise as a result! :)  Why don't we just leave this PR until after #2458 is merged and I will take care of it then.  

Thanks for the helpful feedback, I will make another commit handling some of your suggestions soon!

fkiraly:
>Long story short - sorry about the clashes, I thought we were on the same page! Lets just chalk it up to an unfortunate miscommunication! I would certainly be happy to handle any merge conflicts that arise as a result! :)

My suggestion would be: create another branch, off your fork branch `multiplex_additions`. Into that branch, merge #2458. This will lead to (possibly horrific) conflicts, but you can develop off what is likely the end state of #2458.

Of course, #2458 could change based on review (incl comments that you leave there!), but I do not expect that to be severe.

fkiraly:
feel free to make another draft PR, btw - that will let allow you to run the remote CI. Of course, you can also use `check_estimator` locally, that should also work.

fkiraly:
generally, there are two advanced techniques relevant here if you are working on a major feature.
"feature stream" - a feature branch into which smaller PR are merged, then the entire feature branch is merged into dev/main
"chunk & chain" - multiple PR that depend on each other, and you start merging them in sequence (first to last)

If we had planned in advance to work on the same thing together, I would have gone for the first, since dunder & name/hierarchical seem independent. Now I'm suggesting the second - another example is #2496, #2500, #2502.

fkiraly:
(win 3.8 failure is due to `pytest-timout` - I'll raise the limit in #2541 to 10min)

miraep8:
Done! :) Fortunately as expected our only merge conflict ended up being the `[author]` line in multiplex.py, and otherwise all our changes were compatible! Thanks for the tips regarding best practices for the major feature - will be very useful going forward! I will go ahead and change the comments in this commit to reflect the fact that this PR is "Dependent" on your PR!

sktime/forecasting/base/_base.py
fkiraly: I think the convention with dunders is returning `NotImplemented` if they are not implemented, and it is important to do so.

Why? There is a dunder parse order, let's say we have an expression `a * b * c`. If `a` does not implement `__mul__` for `b`, then python looks next whether `__rmul__` of `b` implements for `a`, which can be the case, and so on.

Had you raised an exception in `a.__mul__`, this would have broken in the first step of parsing. But, expected behaviour and implementation is that python needs to check whether one of the two objects implements it. The information is passed via `NotImplemented`, without raising an exception.

fkiraly:
> I am curious if we actually needed to merge?

I'm surprised and a little confused. Didn't you do all that stuff in the `__init__` and with name unification? Surely that must be clashing on some level. Or so I thought.

But, maybe I overestimated the trouble, and you did a good job in encapsulating your changes even within the class. If that's the case (which it seems), kudos.

sktime/forecasting/base/_base.py
miraep8: No worries - it is implemented like that in the new code! :)

miraep8:
>  Didn't you do all that stuff in the __init__ and with name unification? Surely that must be clashing on some level. Or so I thought.

Hmm, I played around with the `__init__` in one of my first commits, but by the time I actually made the PR (several commits later) I got rid of all that because it was messy :) I can see how that could have been confusing though!

sktime/forecasting/base/_base.py
miraep8: But this is helpful to know/makes sense

fkiraly:
Hm, seems to work!

From my perspective, we could merge - but, due to the dependence on my own PR #2458, we need to wait for someone to review that.
You can´t approve since you are not a core dev, but you could help other reviewers by looking at it and pasting your remarks or change requests. Due to your contributions here, you are probably currently the other person who understands the `MultiplexForecaster` best.

fkiraly (approved):
Very nice contribution!

In the end, the merge was less problematic than I thought.

Will merge as soon as #2458 is merged.

PR #2541, opened 2022-04-23
[ENH] raise `pytest-timeout` time limit to 10min

fkiraly:
This raises the per-test time limit in `pytest-timout` from 5min to 10min.

Reason: some tests, depending on time of day and total number of CI jobs run, may exceed 5min.


fkiraly:
emergency merge, timeout limit was too harsh and let windows tests fail.

PR #2544, opened 2022-04-24
[DOC] expanding content in testing section of "adding estimator" developer docs

aiwalter:
Added pyhton examples to doc of how to run particular tests/estimators


docs/source/developer_guide/add_estimators.rst
```diff
@@ -133,10 +133,30 @@ or third party modules that aim for interface compliance with ``sktime``,
...
+
+    >>> from sktime.utils.estimator_checks import check_estimator
+    >>> from sktime.forecasting.naive import NaiveForecaster
+    >>> check_estimator(NaiveForecaster)
+
 *   importing test classes, e.g., ``test_all_estimators.TestAllEstimators`` or
     ``test_all_forecasters.TestAllForecasters``. The imports will be discovered directly
     by ``pytest``. The test suite also be extended by inheriting from the test classes.
 
+.. code-block:: python
```
fkiraly, line 147: this is not how you would do it if you want to run it via pytest.

pytest runs them if the class is *present* already, it does not need to be run.

fkiraly (changes requested):
These are useful code blocks!

Although, the additions may be in the wrong place.

The section you added them in is using the `sktime` tests as part of a third party test suite. But, the examples are about running the tests locally, so should be perhaps in that section?

In a test suite, either `check_estimator` has to be called within a test. Or, the classes need to be imported, but no methods need to becalled.

docs/source/developer_guide/add_estimators.rst
aiwalter: can you give the code snippet for `pytest`?

fkiraly: sth along the lines of

```python

from sktime.tests.test_all_estimators import TestAllEstimators

from mypackage.registry import list_of_all_estimators_getter()

class TestEstimatorsInMyPackage(TestAllEstimators):

    # this overrides the retrieval of estimators to the third party package
    def _all_estimators(self):
        return list_of_all_estimators_getter()

    def test_my_new_custom_test(self, estimator_class, scenario):
        # do sth useful here, fixtures available are as in BaseFixtureGenerator
```

to override fixture generation or add more fixtures, one needs to add `_generate_` functions; decorating a test with a `pytest.mark.parametrize` also works.

fkiraly: this is the complicate route, for extending the *test framework*

fkiraly: the easier route is sth like

```python

@pytest.mark.parametrize("estimator", ESTIMATOR_LIST)
def test_my_estimators(estimator):
    check_estimator(estimator, return_exception=False)
```

aiwalter: I just wanted to document somewhere how to run test for a dedicated estimator, so the last one seems good for that

fkiraly: but, this is for a test in an external package.

If you are *developing*, you just call `check_estimator`

fkiraly:
@aiwalter, I cleaned this up a bit and added some details.

Kindly have a look, I hope you concur that this is an improvement and can go in?

PR #2546, opened 2022-04-24
[ENH] early fit for NaiveVariance

fkiraly:
This PR refactors and improves the `NaiveVariance` prediction interval wrapper:

* the computing of the sliding residual matrix is factored out into a function
* more efficient behaviour is added if `fh` has been passed early, in `fit`. In that case, `NaiveVariance` also computes the residual matrix early, avoiding a recomputation in the fitted state
* adds an `initial_window` parameter that allows to flexibly determine minimal window size to fit the forecaster to, in order to obtain the matrix of forecast residuals. By default, it is set to 1 which may be sub-optimal but is current behaviour.

This refactor could further be continued by:

* replacing the residual matrix computation by an external utility, used across multiple estimators making use of it, e.g., the conformal prediction interval wrapper https://github.com/alan-turing-institute/sktime/pull/2542
* allowing general resampling schemes to be passed to compute the residual matrix on


fkiraly:
FYI @IlyasMoutawwakil

PR #2551, opened 2022-04-26
[ENH] Refactor sliding and expanding window splitters to allow timdelta forecasting horizon

khrapovs:
#### Reference Issues/PRs

Next step in #1737

In general, I have refactored splitters (mainly `SlidingWindowSplitter` and `ExpandingWindowSplitter`) so that they accept `pd.Timedelta` as forecasting horizons.

More detailed list of changes:
- Conditions in `_check_window_lengths` are simplified taken into account that we now support only two options: either everything (windows, horizons, etc.) is integer, or everything is timdelta-like.
- `get_cutoffs` now returns only integer arrays. This is a unification step to make it clear that we return `k` instead of either `k` or `t(k)`. All implementations, docstrings and test function `_check_cutoffs_against_train_windows` are updated accordingly.
- New method `_split_windows_generic` in `BaseWindowSplitter`. It encapsulates code repetition from two implementations `_split_windows` in `SlidingWindowSplitter` and `ExpandingWindowSplitter`. The difference is regulated by boolean argument `expanding`.
- Most importantly usage of `_coerce_duration_to_int` with hardcoded `freq="D"` is gone. Now we do not care if timedelta is daily, weekly, or any other frequency.
- All test functions for `SlidingWindowSplitter` and `ExpandingWindowSplitter` have additional test cases `TEST_FHS_TIMEDELTA`.

No

##### For all contributions
- [X] The PR title starts with either [ENH], [DOC] or [BUG] indicating whether the PR topic is related to enhancement, documentation or bug


khrapovs:
@fkiraly @aiwalter Sorry to bother you, but would you be so kind to take a look at this PR? Thanks!

fkiraly:
sorry to let you wait, forgot this over GSoC and 0.11.3 and all. Sure!

fkiraly:
looks ok, let me just quickly test whether it would break in combination with #2599

khrapovs:
Thanks for suggestions! As you requested, I have updated dosctrings for `_split_windows_generic` and for `get_cutoffs`. I made it even more explicit that `get_cutoffs` always return `.iloc[]` integer indices.

sktime/forecasting/model_selection/_split.py
```diff
@@ -472,8 +471,8 @@ def get_cutoffs(self, y: Optional[ACCEPTED_Y_TYPES] = None) -> np.ndarray:
 
         Returns
         -------
-        cutoffs : np.array
-            The array of cutoff points.
+        cutoffs : np.ndarray
+            The array of integer cutoff points.
```
fkiraly, line 475: I would suggest to write:
1D np.ndarray of `int`
`iloc` location indices, in reference to `y`, of cutoff indices
(is that correct?)

sktime/forecasting/model_selection/_split.py
```diff
@@ -707,41 +700,81 @@ def _split_for_initial_window(self, y: pd.Index) -> SPLIT_ARRAY_TYPE:
...
+            Index of time series to split
+        fh : ForecastingHorizon
+            Single step ahead or array of steps ahead to forecast.
+        expanding : bool
+            Expanding (True) or sliding window (False) splitter
+
+        Yields
+        ------
+        training_window : np.ndarray
+            Training window indices
```
fkiraly, line 749: same here. Are these `iloc` or `loc`references? And, reference, to what?

fkiraly (changes requested):
Excellent!

Could you just be a little bit more specific?
Made suggestions, but I´m not sure whether they are correct.

sktime/forecasting/model_selection/_split.py
khrapovs: yes, totally agree. changed in all 3 palces

khrapovs:
> Could you just be a little bit more specific? Made suggestions, but I´m not sure whether they are correct.

All suggestions are fine with me. Tests pass. Anything else I can do here?

fkiraly:
you said you changed it in 3 places, but only 1 place is changed. Probably you just forgot to push.

khrapovs:
these two lines:
```
cutoffs : 1D np.ndarray of int
            iloc location indices, in reference to y, of cutoff indices
```
can be found in line 474, 631, 853, and 1071. You are right, I forgot to push one. Thanks.

fkiraly:
I think it´s still missing in line 749, 751? Did you push everything?

khrapovs:
> I think it´s still missing in line 749, 751? Did you push everything?

Lines 748-751:
```
train : 1D np.ndarray of int
    Training window iloc indices, in reference to y
test : 1D np.ndarray of int
    Test window iloc indices, in reference to y
```

fkiraly (approved):
Thanks! all good now.

PR #2557, opened 2022-04-26
[BUG] fix for `update_predict` state handling bug, replace detached cutoff by deepcopy

### LINKED ISSUE #2556: [BUG] `BaseForecaster.update_predict` broken upon second call, `detached_cutoff` does not reset component cutoffs
`update_predict` is broken for all composite forecasters in the following way:

If it is run once for a composite with a component, the `cutoff` of the composite is reset, but not that of the component. This leads to unexpected behaviour and non-compliant predictions when `update_predict` is called a second time.

This is due to a fundamental bug in the design of the `detached_cutoff` that is used internally.

An example is the test failure in https://github.com/alan-turing-institute/sktime/pull/2248, in combination with #2555.

Another minimal example of this behaviour is this, in #2248:

```python
from sktime.forecasting.compose import BaggingForecaster
from sktime.forecasting.model_selection import SlidingWindowSplitter

X = _make_series()
X1 = X[:24]
X2 = X[24:36]
f = BaggingForecaster.create_test_instance()
f.fit(X1, fh=[1, 2])

cv = SlidingWindowSplitter(start_with_window=False)
f.update_predict(X2, cv=cv)
f.update_predict(X2, cv=cv)
```

The expected behaviour is that returns of both `update_predict` are identical, and that the `cutoff` of the internal estimator does not change.

fkiraly:
This PR fixes #2556, the `update_predict` state handling bug, by replacing the `detached_cutoff` by a call of `deepcopy` and working with the deep copy.

This is a temporary fix which ensures correctness - long-term we should probably deprecate the `update_predict` behaviour that resets the `cutoff`.

One secondary problem was that some tests for the online forecaster were *assuming* the behaviour that was problematic in #2556, i.e,. that the inner state *does* get updated by `update_predict`.
This was accounted for by adding a `reset_forecaster` argument to `update_predict`, which can be used to switch between the two behaviours currently assumed in the code base, of `update_predict`.

Docstrings have also been added to clarify what `update_predict` actually does.

### DIFF
--- a/sktime/forecasting/base/_base.py
+++ b/sktime/forecasting/base/_base.py
@@ -748,6 +748,7 @@ class BaseForecaster(BaseEstimator):
         cv=None,
         X=None,
         update_params=True,
+        reset_forecaster=True,
     ):
         """Make predictions and update model iteratively over the test set.
 
@@ -760,11 +761,13 @@ class BaseForecaster(BaseEstimator):
             self.cutoff, self._is_fitted
             If update_params=True, model attributes ending in "_".
 
-        Writes to self:
+        Writes to self, if reset_forecaster=False:
             Update self._y and self._X with `y` and `X`, by appending rows.
             Updates self.cutoff and self._cutoff to last index seen in `y`.
             If update_params=True,
                 updates fitted model attributes ending in "_".
+
+        Does not update state if reset_forecaster=True.
 
         Parameters
         ----------
@@ -795,6 +798,13 @@ class BaseForecaster(BaseEstimator):
                 X.index must contain y.index and fh.index both
             there are no restrictions on number of columns (unlike for y)
         update_params : bool, optional (default=True)
+            whether model parameters should be updated in each update step
+        reset_forecaster : bool, optionel (default=True)
+            if True, will not change the state of the forecaster,
+                i.e., update/predict sequence is run with a copy,
+                and cutoff, model parameters, data memory of self do not change
+            if False, will update self when the update/predict sequence is run
+                as if update/predict were called directly
 
         Returns
         -------
@@ -815,6 +825,7 @@ class BaseForecaster(BaseEstimator):
             cv=cv,
             X=X_inner,
             update_params=update_params,
+            reset_forecaster=reset_forecaster,
         )
 
     def update_predict_single(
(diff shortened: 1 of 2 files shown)


fkiraly:
@GuzalBulatova, I think this will, finally, also fix your bug now!

fkiraly:
this should do it!
https://github.com/alan-turing-institute/sktime/pull/2557

sktime/forecasting/base/_base.py
```diff
@@ -795,6 +798,13 @@ def update_predict(
                 X.index must contain y.index and fh.index both
             there are no restrictions on number of columns (unlike for y)
         update_params : bool, optional (default=True)
+            whether model parameters should be updated in each update step
+        reset_forecaster : bool, optionel (default=True)
```
ltsaprounis, line 802: typo: optionel -> optional

ltsaprounis, line 802: typo: optionel -> optional

sktime/forecasting/base/_base.py
```diff
@@ -1984,15 +1977,47 @@ def _predict_proba(self, fh, X, marginal=True):
...
+        cv : temporal cross-validation generator, optional (default=None)
+        X : time series in sktime compatible format, optional (default=None)
+                Exogeneous time series for updating and forecasting
+            Should be of same scitype (Series, Panel, or Hierarchical) as y
+            if self.get_tag("X-y-must-have-same-index"),
+                X.index must contain y.index and fh.index both
+            there are no restrictions on number of columns (unlike for y)
+        update_params : bool, optional (default=True)
+            whether model parameters should be updated in each update step
+        reset_forecaster : bool, optionel (default=True)
```
ltsaprounis, line 2015: typo

ltsaprounis, line 2015: typo

ltsaprounis (approved):
Thanks Franz, that was really confusing me with the composite tests.
Flagged a couple of small typos (optional).

sktime/forecasting/base/_base.py
fkiraly: fixed

sktime/forecasting/base/_base.py
fkiraly: fixed

ltsaprounis (approved):

Issue #2555, opened 2022-04-26
[BUG] test framework retains state between different test runs


fkiraly:
It would appear that the test framework retains object state between different runs of tests.

E.g., if the test `test_update_predict_predicted_index` is run twice, the second run will be with the state the object has at the end of the first run.
Example of an issue caused by this is here: https://github.com/alan-turing-institute/sktime/pull/2248

Expected behaviour is that a new estimator instance is passed, not a reference to the last one.

PR #2558, opened 2022-04-26
[BUG] fixing side effects between test runs of the same test in the test suite

fkiraly:
This PR fixes within-test side effects discovered by https://github.com/alan-turing-institute/sktime/issues/2555.

High-level, the changes ensure that clones and copies are made by test *run*, as opposed to by *test* but not by *test run* (a test could have multiple test runs).

Following changes are made to address and solve #2555:
* new tests in `TestAllEstimators` that test if estimator instance state is preserved across tests or test instances, for a given estimator. Without any additional changes, the "within test" variant fails, i.e., the `estimator_instance` fixture retains state changes across runs of the same test. This needed to be addressed in `TestAllEstimator` and `check_estimator`, which had similar but slightly different logic to pass fixtures to the tests.
* the issue is addressed in `TestAllEstimator` by making the `estimator_instance` fixture indirect, and creating a new instance of `estimator_instance` (via `clone`) via the `pytest.fixture` and "indirect fixture" formalism. A class variable is added to `TestAllEstimator`, usable by descendants, to indicate which fixtures are indirect.
* the issue is addressed in `check_estimator` by making a `deepcopy` before arguments are passed on to tests. This also seems to speed up the `check_estimator` runtime and tests considerably.

Further changes:
* docstring changes to uniformize occurrences of "parametrize" and "parameterize"
* a `deepcopy_fixtuers` argument in `create_conditional_fixtures_and_names` which allows finer control on whether returned fixtures are reference identical or not. The default is set to `False`, i.e., fixtures are returned reference identical, since now the testing machinery makes the clones/copies, per test run.


sktime/tests/test_all_estimators.py
```diff
@@ -662,6 +685,20 @@ def test_has_common_interface(self, estimator_class):
         if hasattr(estimator, "predict_proba"):
             assert hasattr(estimator, "predict")
 
+    def test_no_cross_test_side_effects_part1(self, estimator_instance):
+        """Tests that there are no side effects across tests."""
```
GuzalBulatova, line 689: Docstrings should start with bare infinitive, so 'Test ...', in the ones below as well. The wording in the PR also seems clearer I think, `Test if estimator instance state is preserved across tests or test instances.` Other than that looks all good to me

fkiraly: ok, I updated the docstring.

I did not go with your suggestion, since I think that can be read to imply that "preserving the state" (preserving being a positive verb) is the desired condition.

fkiraly:
@GuzalBulatova said "all good" but forgot to approve. Interpreting that as approval.

sktime/tests/test_all_estimators.py
GuzalBulatova: Sure, makes sense

Issue #2559, opened 2022-04-27
[ENH] resolve duplication of package/version specs between `pyproject.toml` and `test-windows` workflow


fkiraly:
I think we would like to control all dependencies and version bounds from `pyproject.toml` or another central location, but some packages are directly installed in the windows CI job in `test-windows`, with their own version bounds.

Possibly we will need different bounds for windows/conda, but in any case we should ensure they are controlled somewhere centrally, in my opinion.

May be related as a facilitating factor to https://github.com/alan-turing-institute/sktime/issues/2552

fkiraly:
This is now resolved - not sure by which PR, but the CI now installs windows in a matrix along with all other OS.

Issue #2552, opened 2022-04-26
[BUG] `FileNotFoundError` - `gfortran` in all CI/CD


fkiraly:
PRs error out on windows CI with this, since recently:

`FileNotFoundError: Could not find module 'C:\Miniconda\envs\test\lib\site-packages\scipy\.libs\libbanded5x.TNNMG3IXHGJK7NIBT5J6YPEO5XWTOQAJ.gfortran-win_amd64.dll' (or one of its dependencies). Try using the full path with constructor syntax.`

Unclear what is causing it, likely some package upgraded recently (within the last 12 hours)?

PR for diagnostic runs of CI here: #2550

This needs to be solved before we can make the 0.11.3 release :-(

lmmentel:
I can only see breaking CI runs on windows with conda, seems to be a conda related issue.

lmmentel:
I found this discussion that suggests changing the conda channel for installing scipy https://github.com/pytroll/satpy/issues/1835

fkiraly:
some observations:

* `scipy` coming from `conda-force` seems to have upgraded to 1.8.0 a few days ago - even though the release happened months ago
* this PR https://github.com/alan-turing-institute/sktime/pull/2560 introduces a bound and has `scipy 1.7.3`, but the error still occurs
* switching the `scipy` and `matplotlob` mirror from `conda-force` to `anaconda` here https://github.com/alan-turing-institute/sktime/pull/2561 also ensures `scipy 1.7.3`, and does not seem to suffer from the problem

I do not understand why #2560 does not solve it, but #2561 does.

We should also not close this until we have understood the reason and whether the problem is waiting to happen again once `anaconda` moves up to `1.8.0`.

lmmentel:
How about making the conda deps more transparent and adding an `environment.yaml` with explicit deps, channels and versions?

fkiraly:
good idea - I opened an issue about the general problem this suggestion tries to address here, already:
https://github.com/alan-turing-institute/sktime/issues/2559

fkiraly:
Has been resolved by https://github.com/alan-turing-institute/sktime/pull/2561

PR #2560, opened 2022-04-27
[BUG] bound scipy to <1.8.0 on windows to resolve `gfortran` `FileNotFoundError` in all CI/CD

fkiraly:
Attempts the fix for #2552 suggested by @lmmentel, of bounding `scipy<1.8.0` in windows.

This was previously attempted by @chrisholder #2554 but with a mistake: adding the bound in the "wrong" place, which is easy to do because of the problem highlighted in #2559, where some windows bounds need to go in a hard to find place.


fkiraly:
well, that did not work either :-(

PR #2248, opened 2022-03-18
[ENH] Bagging Forecaster

ltsaprounis:
#### Reference Issues/PRs

Fixes #2247 

##### For all contributions
- [ ] I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/main/.all-contributorsrc).
- [ ] Optionally, I've updated sktime's [CODEOWNERS](https://github.com/alan-turing-institute/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
- [ ] I've added unit tests and made sure they pass locally.

##### For new estimators
- [ ] I've added the estimator to the online documentation.
- [ ] I've updated the existing example notebooks or provided a new one to showcase how my estimator works.

PR #2562, opened 2022-04-27
[ENH] add `reset` at the start of all `fit`

fkiraly:
This PR adds `reset` at the start of all `fit` methods.

This prevents issues with state handling of composites, such as discussed in https://github.com/alan-turing-institute/sktime/pull/2531, especially:

* incorrectly set dynamic tags when calling `set_params` and then calling `fit`
* incorrectly set state variables such as `fh`, `cutoff` (forecasting) or `classes_`, `n_classes_` (classification), in components of composites, when calling `set_params` and then calling `fit`
* incorrectly set `cutoff` and `fh` in components of composites when calling `update_predict` twice (forecasting)

Also adds tests that would fail on `main` (but should work) and are fixed by the above.

Relies on fixes of bugs unmasked by the above:

* https://github.com/alan-turing-institute/sktime/pull/2563


fkiraly:
@miraep8, do you still have the failing code that we could add as a test here saved somewhere?
I seem to have lost it.

@ltsaprounis, we should also also add the `BaggingForecaster` `update_predict` example to this PR once #2248 is merged.

miraep8:
I *think* this is what you are referring to: 
```
from sktime.forecasting.var import VAR
from sktime.forecasting.compose import MultiplexForecaster

f = MultiplexForecaster([ThetaForecaster(), VAR()])
X = _make_series()
f.fit(X)
f.set_params(selected_forecaster="VAR")
X_multivariate = _make_series(n_columns=2)
f.fit(X_multivariate)
```

Or potentially some other version of the same?

fkiraly:
Yes, this is it. Thanks

fkiraly:
the two columns 

```python
X = _make_series()
f.fit(X)
```
are superfluous, I think, to test what we want to test

miraep8:
I agree

PR #2563, opened 2022-04-27
[BUG] fix random state overwrite in MiniRocketMultivariate

fkiraly:
Correctly resetting estimators in https://github.com/alan-turing-institute/sktime/pull/2562 unmasked a hidden bug with `MiniRocketMultivariate`, which was overwriting its `random_state` variable in `__init__`. This has been fixed.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -258,7 +258,7 @@
     },
     {
       "login": "angus924",
-      "name": "angus924",
+      "name": "Angus Dempster",
       "avatar_url": "https://avatars0.githubusercontent.com/u/55837131?v=4",
       "profile": "https://github.com/angus924",
       "contributions": [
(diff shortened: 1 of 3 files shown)


ltsaprounis:
Not sure if this should be fixed in the context of this PR or it's a separate issue, BUT:

Setting the numpy seed as done in line [140](https://github.com/alan-turing-institute/sktime/blob/2ab70b78caf30f3a5fa1ea516ba4179dd44ba632/sktime/transformations/panel/rocket/_minirocket_multivariate.py#L140) is considered bad practice: 
https://numpy.org/doc/stable/reference/random/generated/numpy.random.seed.html?highlight=seed#numpy.random.seed

We should change the randomnesss control to something like this: 

```python
from sklearn.utils import check_random_state

...

rng = check_random_state(self.random_state)

# replace all calls from np.random to rng
rng.choice(...)
rng.randint(...)
```

fkiraly:
I´d say out of scope, but kindly report as an issue.

In-scope for this PR is "fixing non-compliant interface", I don´t know much about multirocket.

ltsaprounis (approved):
Approved as a temp fix of the overwriting behaviour. In the future, randomness handling should be changed as described in #2568.

PR #2567, opened 2022-04-27
[BUG] Fixes the index name dependencies in WindowSummarizer

### LINKED ISSUE #2504: [BUG] issues with WindowSummarizer in hierarchical case
I think there are two bugs (or problems) with `WindowSummarizer`:

* it assumes that levels are named "instances" etc. This assumption is no longer made in the `pd-multiindex` mtype
* despite claims made in various places, it is not actually ready for hierarchical data - `X_inner_mtype` should include `pd_multiindex_hier`, but in that case the transformer will break.

FYI @danbartl

ltsaprounis:
#### Reference Issues/PRs

Fixes #2504 

Removed any dependency on level names. Only requires the agreed convention for `pd-multiindex` and `hierarchical` that the time dimension is the last level.

Validate the quality of the hierarchical case unit tests

##### For all contributions

### DIFF
--- a/sktime/transformations/series/summarize.py
+++ b/sktime/transformations/series/summarize.py
@@ -343,7 +343,8 @@ class WindowSummarizer(BaseTransformer):
             bfill = False
         for cols in target_cols:
             if isinstance(X.index, pd.MultiIndex):
-                X_grouped = X.groupby("instances")[cols]
+                hier_levels = list(range(X.index.nlevels - 1))
+                X_grouped = X.groupby(level=hier_levels)[cols]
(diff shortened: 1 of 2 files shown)


fkiraly (approved):
Thanks for fixing this!

FYI @danbartl

ltsaprounis:
🎵 *There's something happening here. What it is ain't exactly clear* 🎵
Looks like a parallel processing error:
https://github.com/alan-turing-institute/sktime/runs/6207028528?check_suite_focus=true#step:7:40198

ltsaprounis:
I manually re-run the failed jobs to see if I'm lucky this time.

PR #2575, opened 2022-04-29
[ENH] refactored dunder concatenation logic

fkiraly:
This PR refactors repetitive boilerplate code in dunder concatenation logic across `sktime` and moves it into one place, into the `_HeterogeneousMetaEstimator`. The new method `_dunder_concat` exposes variable parts of the boilerplate, which are the base class, pipeline-like composite class, attribute name for the concatenation, and whether the concatenation is left or right (for left vs right dunder).

Dunder concatenators that are simplified by the refactor:
* `__mul__` and `__rmul__` of `TransformerPipeline`
* `__add__` of `FeatureUnion`
* `__or__` of `MultiplexForecaster`

Adds some missing right versions of dunders, using `_dunder_concat`:
* `__radd__` of `FeatureUnion`
* `__ror__` of `MultiplexForecaster`

Makes additional changes:
* docstrings - fixes of typos and uniformization across dunders.
* `test_multiplex_or_dunder` seemed to have a typo, `_get_estimator_names` should be invoked on `forecasters_`, not `forecasters`


fkiraly:
FYI @miraep8, @aiwalter

fkiraly:
@miraep8, I hope you don't mind me completely overwriting one of the two `__or__`-s you added.

As in all refactors, I'd see the pre-refactor state work as necessary conditionality for the refactor. So even when the original code completely (or partially, as in this case) disappears, it is still a crucial "part of" its new, hopefully slightly nicer re-incarnation.

miraep8:
That's fine by me!  Will try and take a look at the changes later and see if I have any comments etc.

fkiraly:
the test for the multiplexer, `test_multiplex_or_dunder` is failing.
I think it is because it accidentally looks at the constructor parameter, `forecasters`, instead of the "made unique" names in `forecasters_`?

sktime/forecasting/compose/_multiplexer.py
```diff
@@ -126,45 +126,50 @@ def _check_selected_forecaster(self):
...
+            base_class=BaseForecaster,
+            composite_class=MultiplexForecaster,
+            attr_name="forecasters",
+            concat_order="left",
+        )
 
-        Raises
-        ------
-        ValueError if other is not of type MultiplexForecaster or BaseForecaster.
+    def __ror__(self, other):
```
miraep8, line 151: So, I could be wrong, but do we actually need this  `__ror__` case?  Why I ask - in theory `__ror__` covers 2 legitimate cases:

- 1 `BaseForecaster` `|` `MultiplexForecaster`
- 2 `MultiplexForecaster` `|` `MultiplexForecaster`

But I think both these cases are already covered!  

1 - should be handled by the `__or__` method defined in `BaseForecaster`. In this case we would convert the base forecaster to a multiplex forecaster, then pass it to the multiplex forecaster `|` dunder. (so it essentially becomes case 2)
2 - can be handled by the fact that the object on the left is already is a `MultiplexForecaster` and thus is covered by `__or__` already.

I guess in theory the `__ror__` could be helpful for raising `NotImplemented` if we tried to do something like this:

"this shouldn't work" | `MultiplexForecaster`

But I would imagine this would already be raised without us needed to explicitly define it.  I also worries it will be confusing/raise bugs to have this unnecessary `__ror__`/I don't completely understand which dunder would take precedent in the case that multiple could handle the case at hand...  (ie, could be fun to see in the case of `MultiplexForecaster` `|` `MultiplexForecaster` if `__or__` or `__ror__` gets called. (my guess is `__or__` but might play around later with the debugger to see!)) 

Please let me know if you feel differently though/that `__ror__` is needed.

miraep8: As a follow up - I guess I would have similar questions about `__radd__` and `__rmul__`, though I am less familiar with whether these bases are fully covered already (though one might imagine that if there were not a similar type casting could happen at the `BaseForecaster` or `BaseTransformer` stage.

miraep8:
> the test for the multiplexer, test_multiplex_or_dunder is failing.
I think it is because it accidentally looks at the constructor parameter, forecasters, instead of the "made unique" names in forecasters_?

Correct, good catch/fix!

sktime/forecasting/compose/_multiplexer.py
fkiraly: > So, I could be wrong, but do we actually need this  `__ror__` case?

Two reasons:

* In the dunders I wrote, I also allowed the named tuple case, e.g.,
`("tf1", trafo) * ("tf2", trafo2) * forecaster`
which works as long as the "special" element (e.g., forecaster, classifier) is not a tuple.
One could argue, one does not want this, though

* this is the "real" reason imo that makes the right function necessary, imagine the forecaster pipeline `trafo1 * trafo2 * forecaster * postproc` - how would this work if you do not want to attach to the transformers forecaster specific functionality? It can only work via the `__rmul__` of `forecaster`.
The alternative would require one to add more and more case distinctions to `transformer.__mul__` which violates the open/closed principle.

fkiraly: so for the "almost commutative" compositions which operate only on classes of the same type, I guess one could get away with the left dunder only. But, if you have type inhomogeneous ones, I think you need the right dunder.

miraep8: Hmm, I had initially also wanted to support tuples, though in the end I couldn't see a good way to have something like `("ets", AutoETS()) | ("naive", NaiveForecaster)` end up as a `MultiplexForecaster`, and to me the requirement that you need at least one forecaster/`MultiplexForecaster` seemed like it might be a weird requirement to the user. (and might cause more confusion than it saved, so I decided to limit myself to only supporting `BaseForecaster` and `MultiplexForecaster` as inputs).  So personally I might be in the camp of not wanting this, at least for the case of the `MultiplexForecaster`...

As to the second point, I am still not sure that this applies to `MultiplexForecaster` though/the need for an `__ror__` though? Also - it could just be me (and probably is just me) but as soon as there are multiple things one could want from  `trafo1 * trafo2 * forecaster * postproc` I worry that having the dunders handle this automatically gets a bit opaque/confusing?  For example, I read it as - make a `TransformedTargetForecaster` and then apply some post processing, but as soon as there are multiple ways to do this and the dunder is handling it automatically (ie either attaching or not attaching the transformers forecaster specific functionality) it gets a bit confusing/I wonder if we should support this.  (again - probably ignorance on my part that is leading to this confusing, so in that case apologies!)

fkiraly: > but as soon as there are multiple things one could want from  `trafo1 * trafo2 * forecaster * postproc` 

Quick question - what are the multiple things one could want, i.e., why do you think this is ambiguous?

miraep8: Sorry - looking back I realized I just misread this comment:
> how would this work if you do not want to attach to the transformers forecaster specific functionality?
And thought you were saying something else.  So now I am not thinking that you want multiple things. 

However... Upon re-reading I think now I potentially have a solution for you - if you have the `__mul__` in `BaseTransformer` simply create a new `TransformedTargetForecaster`, and then pass it off to the `__mul__` in `TransformedTargetForecaster` to actually create the new object (ie akin to what I do in `BaseForecaster` for the `|` dunder), then you could have this work without having to attach to the transformer forecaster specific functionality? Thoughts? :)

fkiraly: No, that would not solve it, because then the `TransformedTargetForecaster` creation logic would be attached to `BaseTransformer`? So, forecaster logic being attached to all transformers.

miraep8: Yes, I see your point/if you want to keep references to forecasters out of `BaseTransfromer` this solution makes sense. 

Though I still don't think we need `__ror__`. :)

fkiraly: I think we need it in `BaseForecaster`, but we might not need it in `BaseTransformer`, you might be right. Although I like to play it safe in cases like this, e.g., end a case distinction with `else raise ValueError("unreachable state")`. You could call that (over?) defensive programming https://en.wikipedia.org/wiki/Defensive_programming

Issue #2416, opened 2022-04-08
[ENH] transformer reconcilers - further work


fkiraly:
I'll merge #2287 so we have it available before the holidays, but there's still some things that should be cleaned up in a separate PR.
I'm listing them here.

* complete checker for data frames for "has totals fields". The current checker only checks presence of `__total` in the top level, not whether it exists for all hierarchy nodes.
* a checker for whether nodes add up correctly to totals. Not necessary to enforce, but useful for checks in certain places.
* moving these checks to the `datatypes` module
* updating the transformer output checks to take into account reconcilers properly that add the `__total` indices. Currently, a small part of the checks which checks number of instances is escaped.
* extension of `Aggregator` to the case where not all series have the same indices

fkiraly:
FYI @ciaran-g

PR #2577, opened 2022-04-29
[ENH] transformer reconcilers - add tests and improve

ciaran-g:
#### Reference Issues/PRs

Addresses a few bullet points in #2416

- Removed a restriction on the `Reconciler` which required the bottom level nodes to have unique names by changing the `_get_s_matrix()` function
- Added check for any difference between indexes used in `.fit()` and `.transform()`
- Test the reconciled predictions actually sum appropriately
- Test the index/columns on the summation (S) and G matrices match
- Test the methods work equally for named and unnamed indexes

No

The added tests :-)


sktime/transformations/hierarchical/reconcile.py
```diff
@@ -61,7 +66,7 @@ class Reconciler(BaseTransformer):
 
     METHOD_LIST = ["bu", "ols", "wls_str"]
 
-    def __init__(self, method="bu"):
+    def __init__(self, method="ols"):
```
fkiraly, line 69: we cannot change the default without deprecation now. The reconcilers are being used since 0.11.2, so this change would require a deprecation message, which is displayed whenever `"bu"` is used. Are we certain we want this change in default?

sktime/transformations/hierarchical/reconcile.py
```diff
@@ -94,13 +99,14 @@ def _fit(self, X, y=None):
         if X.index.nlevels < 2:
             return self
 
-        # check index and bottom level of hierarchy are named correctly
-        #   if not, add totals to X
-        if not _check_index_good(X):
+        # check index for no "__total", if not add totals to X
+        if _check_index_no_total(X):
```
fkiraly, line 103: I don't think this warning is sensible for `_fit`, since nothing is returned. The user would be confused about where things are added.

fkiraly (changes requested):
Overall, looks good!

Blocking change requests are minor, but blocking as it would break downwards compatibility:

* default parameter change, if desired, must be preceded by a deprecation cycle. Is it sensible to change the default? I think `"bu"`, as current, is a more sensible default because it is very quick? What is your reasoning to change this?

Non-blocking changes:

* the warning in `_fit` seems confusing, I would recommend not to raise it.
* explanation of `s_matrix` return in `_get_s_matrix` got deleted, is that an accident? Every input and output should have an explanation, which `s_matrix` would lose. Further, kindly describe entries, indices, and how it relates to input. Optimally, link to a place in the Hyndman paper to avoid lengthy definitions.
* same for `g_matrix` in `_get_g_matrix_bu`.

sktime/transformations/hierarchical/reconcile.py
ciaran-g: Good point, I think the `"bu"` is the more intuitive as well - changed back for now. Although `"ols"` may be the preferred method for users in the long-term since it *typically* is more skillful - forecasts at all levels are used in the reconciliation process.

ciaran-g:
I posted my reasoning for the default change in the comment above @fkiraly, although I didn't think about deprecation. Happy to change back as in the latest commit.

Added more info to the doc strings and removed the warning as well.

fkiraly:
you say you changed it back, but the PR still shows `"ols"` as a default - did you forget to push?

fkiraly (approved):
Thanks, looks good to me now!

fkiraly:
(hm, apparently the pre-PR state had docstring default inconsistent with init default - good that this is fixed)

PR #2589, opened 2022-04-30
[ENH] Fix incorrect `update_predict` arg default and docstring on `cv` arg

aiwalter:
Fix update_predict default `cv` param. `None` was raising an exception, so it should not default to it.


fkiraly:
would it perhaps be better to have a default of "time stamps are fed to the updating individually"?

aiwalter:
> would it perhaps be better to have a default of "time stamps are fed to the updating individually"?

Where?

fkiraly:
in `update_predict`

aiwalter:
dont get what you mean 🤔

fkiraly:
as a default for `cv` - if not passed, then all the time stamps should be individually added in sequence?

fkiraly (approved):
This is correct, according to `check_cv`.

I.e., not an interface change, that requires deprecation, but correction of the docstring.

PR #2599, opened 2022-05-01
[ENH] BaseSplitter extension: hierarchical data, direct splitting of series

fkiraly:
This PR extends all splitters, via `BaseSplitter`, with the ability to split hierarchical data.
The base class implements default functionality which splits individual series (at bottom hierarchy) using the internal `_split` and collates the iloc references to the individual series into iloc references to train/test splits for the entire hierarchical series.

Besides, the following methods are added for convenience:

* `split_loc`, which returns `loc` references to the train/test sets
* `split_series`, which returns the actual train/test splits for a series passed

The following, additional changes are made:
* the docstrings are also updated to clarify the difference between the different methods, and the expectations on the inputs
* input checks and conversions in the splitters - internally, conversion happens to one of the `pandas` based containers. The conversion allows passing of any sktime compatible container to the splitter.


fkiraly:
FYI @khrapovs, do you see any interaction with your work? I am staying in the `BaseSplitter` so it should be in a different place, but it can of course affect the logic...

sktime/forecasting/model_selection/_split.py
```diff
@@ -412,41 +415,211 @@ def __init__(
...
-        y_index = get_index_for_series(y)
-        for train, test in self._split(y_index):
-            yield train[train >= 0], test[test >= 0]
+        y_index = self._coerce_to_index(y)
+
+        if not isinstance(y_index, pd.MultiIndex):
+            for train, test in self._split(y_index):
+                yield train[train >= 0], test[test >= 0]
+
+        if isinstance(y_index, pd.MultiIndex):
```
khrapovs, line 440: may I ask, why not `elif`?

khrapovs: also, why not pack everything after `if` into private methods, say, `_split_flat_index` and `_split_multi_index`? just to improve readability

fkiraly: > may I ask, why not elif?

should be a simple else, no? thought, for readability. I added a comment now.

fkiraly: > also, why not pack everything after if into private methods, say, `_split_flat_index` and `_split_multi_index`? just to improve readability

Yes, that is the "obvious" best-practice thing to do here.
I am, however, not sure how to structure it, or whether this can be refactored easily with `VectorizedDF`.

My plan was to refactor this to sth more cleaner once it works together with metrics and tuning, but let me know if you have a good idea for function signatures and utilities here.

khrapovs: ok, nevermind about `if ... else`. this issue is too small to block anything here.

regarding my second comment. I probably did not state it clear enough. My suggestion is very simple. Everything that goes after `if isinstance(y_index, pd.MultiIndex):` move into a separate method, so that you have something like:
```
if isinstance(y_index, pd.MultiIndex):
    yield self._split_multi_index(y_index)
```
Nothing more complicated than that.

fkiraly: > ok, nevermind about if ... else. this issue is too small to block anything here.

Already changed it and added a comment

khrapovs:
> FYI @khrapovs, do you see any interaction with your work? I am staying in the `BaseSplitter` so it should be in a different place, but it can of course affect the logic...

At least on the idea level it looks fine from my perspective. As long as you take existing spliiters' output as given and only generalize to more than one index level, then no effect on my generalization attempts.

sktime/forecasting/model_selection/_split.py
```diff
@@ -412,41 +415,216 @@ def __init__(
...
+        """Split `y` into training and test windows.
+
+        Parameters
+        ----------
+        y : pd.Index or time series in sktime compatible time series format (any)
+            Time series to split, or index of time series to split
+
+        Yields
+        ------
+        train : pd.Index
```
khrapovs, line 519: I think it should be
```
1D np.ndarray of dtype int
```
same as in `_split`

fkiraly: not in `_split_loc`, this is a new method which returns an index

sktime/forecasting/model_selection/_split.py
fkiraly: > Everything that goes after `if isinstance(y_index, pd.MultiIndex):` move into a separate method

oh, that makes sense. And then we can refactor the "multiindex" function without affecting `_split`. Good idea, done.

fkiraly:
> As long as you take existing spliiters' output as given and only generalize to more than one index level, then no effect on my generalization attempts.

Unless, of course, you wrote lasagna code, which may be worth checking ... 😁

sktime/forecasting/model_selection/_split.py
fkiraly: iloc references are always integer
loc references are typically index

khrapovs: `split_loc` calls `self.split` which in turn returns `1D np.ndarray of dtype int`. docstrings seem to be inconsistent.

sktime/forecasting/model_selection/_split.py
```diff
@@ -412,41 +414,234 @@ def __init__(
...
-        test_window : np.ndarray
-            Test window indices
+        train : 1D np.ndarray of dtype int
+            Training window indices, iloc references to training indices in y
+        test : 1D np.ndarray of dtype int
+            Test window indices, iloc references to test indices in y
         """
         raise NotImplementedError("abstract method")
 
+    def _split_vectorized(self, y: pd.Index) -> SPLIT_GENERATOR_TYPE:
```
khrapovs, line 462: Is `pd.MultiIndex` so different from `pd.Index` that this function hast to be separate from `_split`? Or, in other words, is `_split_vectorized` not general enough to cover a simple `pd.Index` too? Or, yet in other words, what would happen if you completely replace `_split` with `_split_vectorized` without differentiating between `pd.MultiIndex` and `pd.Index`?

sktime/forecasting/model_selection/_split.py
fkiraly: > docstrings seem to be inconsistent.

I don't think so? Call `split_loc` an you should see it returns an index.
That this is so is also tested in `test_split_loc`.

khrapovs: Sure, makes sense. But, please, find `assert isinstance(window, np.ndarray)` in the tests which clearly states that the return type of `.split` is `np.ndarray`. This resolves to `False`: `isinstance(pd.Index([], dtype=int), np.ndarray)`. Hence, there is inconsistency in both docstrings and in type suggestions (see `SPLIT_GENERATOR_TYPE`).

fkiraly: ok, I´m getting confused here.
Can you kindly be precise, and tell me which function precisely, in your opinion, has an incorrect docstring, and which part of the docstring is incorrect, and what it should be instead?

sktime/forecasting/model_selection/_split.py
fkiraly: >Or, yet in other words, what would happen 

It would break, because type and methods-wise, `pd.MultiIndex` is different from ordinary `pd.Index` subtypes?
Further, I think it is safer use `_split` directly when it can be used.

khrapovs: ok, understood. thanks for clarification!

sktime/forecasting/model_selection/_split.py
```diff
@@ -412,41 +414,234 @@ def __init__(
...
+        for i in train_multi.keys():
+            train_multi[i] = np.sort(train_multi[i])
+            test_multi[i] = np.sort(test_multi[i])
+
+        for i in train_multi.keys():
+            train = train_multi[i]
+            test = test_multi[i]
+            yield train, test
+
+    def split_loc(self, y: ACCEPTED_Y_TYPES) -> SPLIT_GENERATOR_TYPE:
```
khrapovs, line 526: In response to https://github.com/alan-turing-institute/sktime/pull/2599/files#r863041191
```suggestion
    def split_loc(self, y: ACCEPTED_Y_TYPES) -> Generator[Tuple[pd.Index, pd.Index], None, None]:
```

sktime/forecasting/model_selection/_split.py
khrapovs: Sorry, I got confused myself. Docstrings are ok. Type suggestion is wrong. See another comment to `split_loc` method.

sktime/forecasting/model_selection/_split.py
```diff
@@ -412,41 +414,234 @@ def __init__(
...
+            Training window indices, loc references to training indices in y
+        test : pd.Index
+            Test window indices, loc references to test indices in y
+        """
+        y_index = self._coerce_to_index(y)
+
+        for train, test in self.split(y_index):
+            yield y_index[train], y_index[test]
+
+    def split_series(self, y: ACCEPTED_Y_TYPES) -> SPLIT_GENERATOR_TYPE:
```
khrapovs, line 546:
```suggestion
    def split_series(self, y: ACCEPTED_Y_TYPES) -> Generator[Tuple[ACCEPTED_Y_TYPES, ACCEPTED_Y_TYPES], None, None]:
```

sktime/forecasting/model_selection/_split.py
```diff
@@ -412,41 +414,234 @@ def __init__(
...
+        y, y_orig_mtype = self._check_y(y)
+
+        for train, test in self.split(y.index):
+            y_train = y.iloc[train]
+            y_test = y.iloc[test]
+            y_train = convert_to(y_train, y_orig_mtype)
+            y_test = convert_to(y_test, y_orig_mtype)
+            yield y_train, y_test
+
+    def _coerce_to_index(self, y):
```
khrapovs, line 570: type suggestions for inputs and outputs? sorry, I really find them useful

sktime/forecasting/model_selection/_split.py
```diff
@@ -412,41 +414,234 @@ def __init__(
...
+        y_index : y, if y was pd.Index; otherwise _check_y(y).index
+        """
+        if not isinstance(y, pd.Index):
+            y, _ = self._check_y(y, allow_index=True)
+            y_index = y.index
+        else:
+            y_index = y
+        return y_index
+
+    def _check_y(self, y, allow_index=False):
```
khrapovs, line 591: type suggestions?

khrapovs (changes requested):

sktime/forecasting/model_selection/_split.py
fkiraly: added type annotations

sktime/forecasting/model_selection/_split.py
fkiraly: here it would be quite complex, since we are talking about mtypes. Not sure how to do that.

sktime/forecasting/model_selection/_split.py
khrapovs: cool! thanks!

sktime/forecasting/model_selection/_split.py
khrapovs: yeah, I can see that. let's leave it like that then. but what you could do is to convert this method into static. my PyCharm is very insistent on that change ;)

fkiraly: for a general private method, I'd do that, but I would prefer not to do this for `_check_y`, out of overall consistency with other `_check_something` input check methods across base classes, some of which do access `self`.

fkiraly: similar story to the more "strict" uniformization pattern where all `_transform` methods have a `y` argument, even if they do not internally make use of `y`

khrapovs: hm, interesting argument. not sure that I entirely "buy" it, but I can definitely live with it. not a place where I would argue too much ;)

fkiraly:
> I still did not carefully look into the core of this PR, namely _split_vectorized method. I would be happy to come back to it on Tuesday...

Explanation: it is basically an "apply-by", over individual series.
If it were `loc` indexed, it is in fact just that.

Unfortunately, we want `iloc` indices, and an "apply-by" would start counting every sub-series from 0 again, so we need to do something to get` iloc indices that start counting at the zero of the hierarchical data frame, not of the individual series. That complicates things a lot, since I do not think there are off-shelf methods in `pandas` for this.

khrapovs:
@fkiraly I looked a bit closer inside of the `sktime.forecasting.model_selection._split.BaseSplitter._split_vectorized`. Again, I completely understand the result, for each series in the hierarchy you want to get train/test windows with numeration in the context of one big stacked time series. But I am worried about the implementation of this method. Obviously, you have the same concerns. There are several enclosed loops, several temporary objects. Since I have no right to either block or approve this PR, you can completely ignore my words and ask one of the core developers to merge this "as is". But let me maybe throw an idea, that I have no time to implement myself.

For each time series in hierarchy you can easily get train/test windows that start counting from zero in isolation for each series. What you want to do next is shift those windows by the iloc indices of each series beginning. Say you have two windows in isolation: [0, 1, 3] (out of [0, 1, 2, 3]), and [0, 2] (out of [0, 1, 2]). The index of the whole series is [0, 1, 2, 3, 4, 5, 6]. The "anchors" are then 0 and 4. So your final windows are [0, 1, 3] + 0 = [0, 1, 3], and [0, 2] + 4 = [4, 6]. I know, it is just a toy example, but isn't it easy to get those iloc "anchors"?

sktime/forecasting/model_selection/tests/test_split.py
```diff
@@ -482,3 +483,69 @@ def _check_train_test_split_y(fh, split):
...
+        assert len(train) == 10 * n_instances
+        assert isinstance(test, np.ndarray)
+        assert test.ndim == 1
+        assert test.dtype == np.int64
+        assert len(test) == 1 * n_instances
+
+    for train, test in cv.split_loc(y):
+        assert isinstance(train, pd.MultiIndex)
+        assert len(train) == 10 * n_instances
+        y.loc[train]
```
khrapovs, line 537: is this here just for debugging?

fkiraly: no - this will raise an exception if `train` is not an index subset of `y.index`

fkiraly:
> I know, it is just a toy example, but isn't it easy to get those iloc "anchors"?

Perhaps - I also considered that. That introduces the problem of the addition which you now have to carry out.
I thought indexing is less brittle than addition, especially since unusual types are involved (indices).

Another problem is, things that seem conceptually useful appear to be difficult in `pandas`.

Although it´s perfectly possible that what I doing is unnecessarily complicated, and there is a much better way.
How much code would your approach create?

khrapovs:
> Perhaps - I also considered that. That introduces the problem of the addition which you now have to carry out. I thought indexing is less brittle than addition, especially since unusual types are involved (indices).

train/test windows are just `np.ndarray`s, nothing from pandas. Adding integer anchors to arrays is a perfectly natural operation. But again, I did not spend any time in comparison to you on this problem.

> Although it´s perfectly possible that what I doing is unnecessarily complicated, and there is a much better way. How much code would your approach create?

Sorry, I am just airing my thoughts here. It is also perfectly possible that your implementation is the the only minimal possible and robust. But I can not prove otherwise, until I try myself.

So, after reminding you that I have no right to merge this, you should start the conversation with one of the core developers ;)

sktime/forecasting/model_selection/tests/test_split.py
khrapovs: ok, I see. Is it possible to make it a bit more explicit statement, something like `assert x.issubset(y)`. sorry, I am away from python console to suggest the exact statement...

fkiraly:
> train/test windows are just np.ndarrays, nothing from pandas. Adding integer anchors to arrays is a perfectly natural operation. But again, I did not spend any time in comparison to you on this problem.

Yes, you are right! I should have done that on the numpy arrays, not on the index.

Although for the moment I'd focus on other work, unless it becomes a blocker.
The logic is encapsulated, so it should be easy to refactor it to something more compact.

sktime/forecasting/model_selection/tests/test_split.py
fkiraly: done. Took me 10min to find this.

PR #2601, opened 2022-05-02
[ENH] metrics rework part IV - hierarchical metrics

fkiraly:
This PR continues rework of the current metrics interface.

This PR uses the tests in https://github.com/alan-turing-institute/sktime/pull/2496 and builds upon #2500 and #2502

The following changes are made relative to #2502:

* extending `BaseForecastingErrorMetricFunc` to accept hierarchical input
* enable averaging options via the `multilevel` parameter


sktime/performance_metrics/forecasting/_classes.py
```diff
@@ -68,391 +75,463 @@ class _BaseForecastingErrorMetric(BaseMetric):
...
+            Hierarchical scitype: pd.DataFrame with 3 or more level row MultiIndex
+        y_pred :time series in sktime compatible data container format
+            Forecasted values to evaluate
+            must be of same format as y_true, same indices and columns if indexed
 
         Returns
         -------
-        loss : float
-            Calculated loss metric.
+        loss : float or np.ndarray
```
eenticott-shell, line 163: Just wanted to raise that for prob metrics I have float or pd.series (or pd.DataFrame if by eval-by-index). It would be desirable that the outputs of metrics and prob metrics are consistent, so we should decide which format works best.

eenticott-shell (commented):
Looks good, there are some inconsistencies with the prob metrics which we should discuss.

Regarding what is done for `multilevel` and `multioutput`, is this logic meant to be implemented in `_evaluate`? For the prob metrics I specify what the output from `_evaluate` should look like so that the logic for averaging over `multioutput` (or `multiscore` in my case) can be done within the outer `evaluate` so someone implementing a new concrete class doesn't need to think about it. Is it possible to do this here or does the `multilevel`/`multioutput` logic vary by metric?

sktime/performance_metrics/forecasting/_classes.py
fkiraly: for completeness, can you kindly list the different output types, by condition? As per your mental model?
I might have gotten confused here.

Regarding the case of `evaluate` and `multiindex`, we need to keep this `np.ndarray`, at least for a deprecation cycle, because of downwards compatibility.
Same for the proba metrics - it needs to stay whatever it was, until it underwent a deprecation cycle.

Mid-term, what should it be? I feel we can´t change the numpy output easily, due to interface compatibility with sklearn.

fkiraly:
mac failures unrelated

Issue #2603, opened 2022-05-04
[ENH] General Delegator Design Discussion


miraep8:
### Overview:

This issue is intended to be a spot for a design discussion for how to implement a new more general delegation functionality per a suggestion from @fkiraly .

Many of the composite classes related to multiple estimator testing/comparison - e.g. `MultiplexForecaster` and the related `BaseGridSearch` classes essentially behave *like* a wrapper where most of their forecasting functionality is delegated to a wrapped forecaster. In implementing `MultiplexTransformer` it became clear a more general delegation functionality (that also extended to transformers) would be useful. #2570 is an attempt to solve this problem, however at @fkiraly's suggestion it seemed like it would be worth having a more detailed design discussion. (hence this issue)

### Summary of the Two Approaches:

(*there are of course more than 2 ways to solve this problem, but this summarizes the approach I have implemented and the mixin approach already in place*)

##### Current approach via the mixin (as it was implemented):
- Have a `DelegatedForecaster` class each class requiring delegation inherits from (as a "mixin")
- This class overwrites all the underscore methods (eg `_fit()`, `_predict()` etc) so that they are called on the delegated forecaster instead.
- This means the original functions are called on all wrapped estimators - which can have various issues (namely we now need to dynamically set the tags of the composite classes so that when we call `fit()` on them we accept inputs. 
- The solution to this is to add some logic to `__init__`, and have a `reset()` function that is called at the start of each call to `fit()`. 
- The advantage of this approach is only the estimator actually in need of having a delegation ability is exposed to the delegation logic by inheriting from the mixin. 

##### Proposed new approach via the decorator (as it was implemented):
- Add a decorator to all the main functions in `BaseEstimator` (eg `fit()`, `predict()` etc) which will delegate that function to the wrapped estimator if one is provided.
- Add an attribute to store the name of the attribute storing the estimator these functions should be delegated to. Defaults to `None` so estimators that do not delegate should not be affected. 
- This approach has a disadvantage in that all `BaseEstimator`'s now have delegation logic attached to them even though many of them might not need it. 
- Because `fit()` is no longer being called on the "main estimator" we now no longer need to dynamically set tags, (at least not to get `fit()` to work).  This makes sense to me at least.
[...]

PR #2612, opened 2022-05-07
[ENH] Created _DelegatedTransformer

miraep8:
#### Reference Issues/PRs

Related to convo in #2603, and replaces #2570

Creates a DelegatedTransformer.  Method is identical to creation of DelegatedForecaster. 

No


sktime/transformations/_delegate.py
```diff
@@ -0,0 +1,253 @@
+# -*- coding: utf-8 -*-
+"""Delegator mixin that delegates all methods to wrapped forecaster.
+
+Useful for building estimators where all but one or a few methods are delegated.
+For that purpose, inherit from this estimator and then override only the methods
+    that are not delegated.
+"""
+# copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+
+__author__ = ["fkiraly", "miraep8"]
```
fkiraly, line 10: does not need me here, thanks for the thought

sktime/transformations/_delegate.py
```diff
@@ -0,0 +1,253 @@
+# -*- coding: utf-8 -*-
+"""Delegator mixin that delegates all methods to wrapped forecaster.
```
fkiraly, line 2: transformer

fkiraly (changes requested):
Thanks!

I'd be happy with this as an intermediate step.

I don't think this is the "final form" of delegation, e.g., due to duplication and the discussion in #2603. I think it should be sth closer to #2570 ultimately.

I've added some comments related to typos, non-blocking.

Few small blocking issues:

* `transform` should not return self but a data container (the output of the inner transform), just like `inverse_transform`
* the template does not have an inner `_fit_transform`, this can be removed (perhaps it should have, but that's another topic)

sktime/transformations/_delegate.py
miraep8: sure thing!  I figured since it was so heavily based on your _DelegatedForecaster I would try to err on the side of caution, but you have now been removed.

sktime/transformations/_delegate.py
miraep8: Good catch!

miraep8:
Thanks for the comments @fkiraly, I think I should have resolved the blocking/non blocking comments!

- I do not think we need a `_fit_transform`, since all the delegation should be properly handled by delegating `_fit` and `_transform` My mistake in for including it in the first place/good catch!
- `_transform` return has been fixed as well. 

I figure this PR makes it easier to proceed with implementing `MultiplexTransformer`.  I had initially wanted something more similar to #2570 as well, but I realized in actually implementing it with the design constraints (ie that it would work as a mixin for both forecasters and transformers) I realized there was just a lot of complexity/cases to handle.  I would be happy to provide feedback on a future approach that attempts to resurrect that approach though!

fkiraly:
>but I realized in actually implementing it with the design constraints (ie that it would work as a mixin for both forecasters and transformers) I realized there was just a lot of complexity/cases to handle.

I agree. I think if there weren´t the need to stick with the design patterns, then, in isolation, your second solution (the one you called "hybrid" and that localizes the delegation concern in one class) would be "better". But there´s always an architectural context, e.g., the world´s best motor might not fit well into an existing car design and might even cause problems to the car if you try to put it in nonetheless.

Maybe we can get closer to #2570 with another refactor at some point. I still think the design ideas are superior, but need some re-thinking around the highest-level design patterns in place for sktime currently.

sktime/transformations/_delegate.py
```diff
@@ -0,0 +1,191 @@
...
+        return getattr(self, self._delegate_name)
+
+    def _fit(self, X, y=None):
+        """Fit transformer to X, optionally to y.
+
+        State change:
+            Changes state to "fitted".
+
+        Writes to self:
+            Sets is_fitted flag to True.
```
fkiraly, line 45: non-blocking issue - this is not correct. The setting of the `is_fitted` flag is done by `fit`, not `_fit`.

sktime/transformations/_delegate.py
```diff
@@ -0,0 +1,191 @@
...
+
+    def _transform(self, X, y=None):
+        """Transform X and return a transformed version.
+
+        State required:
+            Requires state to be "fitted".
+
+        Accesses in self:
+            Fitted model attributes ending in "_".
+            self._is_fitted
```
fkiraly, line 76: also not true, only `transform` accesses `is_fitted`.

sktime/transformations/_delegate.py
```diff
@@ -0,0 +1,191 @@
...
+                Example: all series in the panel are detrended individually
+            if `X` is `Series` or `Panel` and `transform-output` is `Primitives`
+                then the return is `pd.DataFrame` with as many rows as instances in `X`
+                Example: i-th row of the return has mean and variance of the i-th series
+            if `X` is `Series` and `transform-output` is `Panel`
+                then the return is a `Panel` object of type `pd-multiindex`
+                Example: i-th instance of the output is the i-th window running over `X`
+        """
+        estimator = self._get_delegate()
+        return estimator.transform(X, y=None)
```
fkiraly, line 121: should be y=y

fkiraly (changes requested):
Blocking change request: I think it should be `y=y` in the `transform` return.

Otherwise, some non-blocking docstring comments.

sktime/transformations/_delegate.py
miraep8: Correct, also removed/modified.

sktime/transformations/_delegate.py
miraep8: Definitely!

miraep8: Change has been made

fkiraly (approved):
In terms of "best practice", I hesitate to merge without direct or at least indirect (usage/application) tests.

But, it looks good and I trust you will know what do do next, when looking at a new `MultiplexTransformer` and, possibly, a refactor of `OptionalPassthrough` (delegates either to the transformer or an "identity transformer").

miraep8:
I am with you with the tests... definitely best practice to have them.  I had been planning to add them along with `MultiplexTransformer` but if you prefer can throw together a quick test suite for this along now.(perhaps just thumbs up this comment if that is your preference and I can throw it together).

fkiraly:
no, I think it´s fine, since we will test indirectly via very simple `MultiplexTransformer` which, if it weren´t in addition useful to the user, would be more or less equivalent to the "obvious" `DelegatorTesterMock`.

PR #2620, opened 2022-05-09
[ENH][DOC] Forecasting notebook update

fkiraly:
This PR updates the forecasting tutorial notebook with various content introduced between 0.9 and 0.12:

* changes to the data container intro
* small section on hierarchical data, 1.6
* estimator lookup and tag functionality added to section 2
* multiplication dunder presented in pipeline section
* section 4 on creating and testing estimator updated

Relies on #2619 to ensure that dunder pipeline inspection works properly.


sktime/forecasting/compose/_pipeline.py
```diff
@@ -154,7 +154,15 @@ def _get_inverse_transform(self, transformers, y, X=None, mode=None):
...
         """Map the steps to a dictionary."""
-        return dict(self.steps)
+        return dict(self._steps)
+
+    @property
+    def _steps(self):
+        return self._get_estimator_tuples(self.steps, clone_ests=False)
+
+    @_steps.setter
+    def _steps(self, value):
```
GuzalBulatova, line 164: Just out of curiosity question, I got lost a bit with inner steps. Where is this setter used?

fkiraly: this is part of the fix in #2619, not the notebook PR.

It's required to write `_steps` write to `steps`, to ensure that `steps` and `_steps` are always in sync.

It's required for the `_HeterogeneousMetaEstimator._set_params` to work, since that targets `_steps` and not `steps`, and `_get_params` reads from `_steps` which is a standard formatted version of `steps`.

GuzalBulatova: Thanks for clarifying, I think I understand now

PR #2622, opened 2022-05-10
DRAFT: diagnostic combination of #2551 and #2599

fkiraly:
This PR is a diagnostic combination of #2551 and #2599, both making core modifications to the splitters, to see whether they would together break remote CI-CD.


fkiraly:
well, what do you say! This works, and not even a single merge conflict! FYI @khrapovs

khrapovs:
> well, what do you say! This works, and not even a single merge conflict! FYI @khrapovs

The fact that we work on the same file does not automatically mean that we have to collide ;) But I agree, this time we were lucky :)

fkiraly:
#2551 is merged, so no further need for diagnostics now.

PR #2629, opened 2022-05-12
[ENH] Test clone with all parameters set to non-default values

fkiraly:
Bug https://github.com/alan-turing-institute/sktime/issues/2026 was not detected by our test suite, despite the fact that we did have a test that, for all estimators, tests whether `clone` works.

The problem with the current test is that `clone` will not fail if a problematic parameter (e.g., overwritten in `__init__` is set at its default.

To cover all potentially problematic parameters, a new test is added where before invoking `clone` all parameters are overwritten at least once to a non-default value.


fkiraly:
superseded by testing multiple param settings

PR #2633, opened 2022-05-12
[MNT] Fix compatibility failures after sklearn 1.1.0 update, bump sklearn version bound to 1.2.0

### LINKED ISSUE #2631: [BUG] sktime breaks with new scikit-learn 1.1.X due to private import
Someone has been importing private methods again 😠 

This time it's `sklearn` 1.1.0 with `_check_param_grid from sklearn.model_selection._search` being imported in `forecasting/model_selection/_tune`

... breaking much of `sktime`

fkiraly:
This fixes various incompatibilities of `sktime` with the recent sklearn 1.1.0 release:
* fixes #2631 by moving the private `_check_param_grid` (which broke at sklearn 1.1.0) from sklearn 1.0.2 to sktime
* fixes breakage caused by `DecisionTreeClassifier` of `sklearn` no longer accepting float for `max_depth`

This PR also moves the newly introduced upper sklearn bound to 1.2.0.


sktime/forecasting/model_selection/_tune.py
```diff
@@ -483,9 +485,33 @@ def __init__(
         )
         self.param_grid = param_grid
 
+    def _check_param_grid(self, param_grid):
+        """_check_param_grid from sklearn 1.0.2, before it was removed."""
+        if hasattr(param_grid, "items"):
```
Lovkush-A, line 490: seems like an indirect way to check that p is a dictionary.

Lovkush-A (approved):
changes are straightforward. one comment only - does not require any changes.

sktime/forecasting/model_selection/_tune.py
fkiraly: yes, I know. This is 1:1 copied from sklearn. A proper refactor would probably just ditch the function entirely and replace it by better check functionality. Perhaps that´s what happened in 1.1.0, but I don´t think the time investment is worth it, separate from a proper clean-up.

PR #2639, opened 2022-05-13
[ENH] TWE switch to use euclidean distance

chrisholder:

This pr switches the distance metric TWE uses to euclidean. This is because the minkowski distance (original proposed for use with TWE) was causing issues with large datasets. Therefore Euclidean is now being used.

##### For all contributions

##### For new estimators


fkiraly (changes requested):
Why not, but you need to deprecate due to a change in default.

TonyBagnall:
I would not use **2 to square I think it may be slow on numba

TonyBagnall (approved):

TonyBagnall:
> Why not, but you need to deprecate due to a change in default.

also dont think TWE is in 11.0 anyway

TonyBagnall:
> oh, is it experimental? How do I see that?

You classified distances module as experimental this not needing deprecation. If you insist I'll add back parameter but it's holding up experiments I want to set off over the weekend

TonyBagnall:
I'm going to put this in when it passes as I need to set them off, unless @fkiraly seriously objects

fkiraly:
> You classified distances module as experimental thus not needing deprecation. If you insist I'll add back parameter but it's holding up experiments I want to set off over the weekend

Yes, you are right, @TonyBagnall! Just checked.
Consider my comment non-blocking then.

TonyBagnall:
> > You classified distances module as experimental thus not needing deprecation. If you insist I'll add back parameter but it's holding up experiments I want to set off over the weekend
> 
> Yes, you are right, @TonyBagnall! Just checked. Consider my comment non-blocking then.

Thanks, could you unblock this or then please?

PR #2641, opened 2022-05-13
[DOC] Classification notebook clean-up, added new pipelines

fkiraly:
This PR cleans up the classification notebook a bit and adds new features added since 0.8.0.

Clean-up:

* moving imports to sections and cells to avoid huge initial import and allow easier copy-pasting of self-contained code blocks
* some formulations clarified

New features now covered:

* loading of data in arbitrary, and most common panel formats `pd-multiindex` and `nested_univ`, explained
* `numpyflat` aka `numpy2d` type is kept and presented
* classifiers accepting arbitrary input formats, including the above
* `make_pipeline` and dunder operators for constructing TSC classifier pipelines


TonyBagnall:
I dont like this, I want to encourage 2D numpy usage, as it is the simplest use case and the most accessible entry into TSC

fkiraly:
> I dont like this, I want to encourage 2D numpy usage, as it is the simplest use case and the most accessible entry into TSC

We have clashing specifications! How do we sort that out?

fkiraly:
I've added back the `numpy2d` type, everyone should be happy now?

fkiraly:
In review:

* there was not an explicit votedown of this PR, although @TonyBagnall was unhappy due to initial removal of `numpy2d` data type
* the unhappiness should be addressed with adding back the `numpy2d` data type, hopefully
* this PR is blocking an important bugfix, as explained in #2638, which also caused use of `numpy2d` to be broken
* more than the 7 day lazy consensu period has elapsed

Therefore, I'm merging this so #2638 can go in for 0.12.0.

PR #2643, opened 2022-05-15
[ENH] `make_pipeline` utility to create linear pipelines of any type

fkiraly:
This PR adds a `make_pipeline` utility to create linear pipelines of any type, similar to the utility of the same name in `sklearn`.

Under the hood, it uses the `*` dunder for case dispatch.

### DIFF
--- a/sktime/pipeline/_make_pipeline.py
+++ b/sktime/pipeline/_make_pipeline.py
@@ -1,0 +1,57 @@
+# -*- coding: utf-8 -*-
+"""Pipeline making utility."""
+# copyright: sktime developers, BSD-3-Clause License (see LICENSE file).
+
+__author__ = ["fkiraly"]
+
+
+def make_pipeline(*steps):
+    """Approximations of uniform order statistic medians.
+
+    Parameters
+    ----------
+    steps : tuple of sktime estimators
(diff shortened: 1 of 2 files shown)


sktime/pipeline/_make_pipeline.py
```diff
@@ -0,0 +1,56 @@
+# -*- coding: utf-8 -*-
+"""Pipeline making utility."""
+# copyright: sktime developers, BSD-3-Clause License (see LICENSE file).
+
+__author__ = ["fkiraly"]
+
+
+def make_pipeline(*steps):
+    """Approximations of uniform order statistic medians.
```
ltsaprounis, line 9: This description doesn't look correct to me

ltsaprounis (changes requested):
Looks good, one small issue in the docstring.

sktime/pipeline/_make_pipeline.py
fkiraly: yes, fixed.
(I just copied a random function to get started)

fkiraly:
docstring issue is fixed, thanks for spotting this.

ltsaprounis (approved):

PR #2646, opened 2022-05-15
[BUG] Fixed lag definition in WindowSummarizer

GrzegorzRut:
#### Reference Issues/PRs
Mentioned  in [#2424](https://github.com/alan-turing-institute/sktime/pull/2424#issuecomment-1125423143)

This fixes the the behavior of `WindowSummarizer` with provided lags. If lag is set to zero, there should be no shift.

Didn't touch truncate_start as it seems fine for now. The documentation seems fine as it is as well.
Maybe @danbartl would like to take a look?


fkiraly (approved):
Yes, I think this fixes it.

Happy with this as a bugfix, since logic was discrepant with docstring which seems to be correct, as you say.

What would be great if you (or someone else) could add something to `transformations/series/tests/test_summarize` to check whether the output has the correct indices. Currently there are no tests of this?

I´d consider adding tests separate from this PR, since there were no tests to start with.
(aware this was merged quickly due to deadlines, but we shouldn't leave it in this state)

I'll also wait for @danbartl for a few days before I merge this, just to confirm that there's no problem I'm unaware of.

GrzegorzRut:
I'm closing this PR as the changes are redundant. @danbartl, many thanks for double checking this!

`WindowSummarizer` with `lag_config` implicitly adds 1 to the lag  but `lag_config` is going to be deprecated soon.
```python
import pandas as pd

index = pd.date_range("2000-01-01", "2000-01-05", freq="D")
length = len(index)
y = pd.DataFrame(range(length), index=index)
kwargs = {
    "lag_config": {"lag": ["lag", [[0,0]]]},
}
transformer = WindowSummarizer(**kwargs, n_jobs=1)
transformer.fit_transform(y)
                      |   0_lag_1 |
|:--------------------|----------:|
| 2000-01-01 00:00:00 |       nan |
| 2000-01-02 00:00:00 |         0 |
| 2000-01-03 00:00:00 |         1 |
| 2000-01-04 00:00:00 |         2 |
| 2000-01-05 00:00:00 |         3 |
```
`WindowSummarizer` with `lag_feature` does not do any implicit additions.
```python
kwargs = {
    "lag_feature": {
        "lag": [0],
    },
}
transformer = WindowSummarizer(**kwargs, n_jobs=1)

transformer.fit_transform(y)
|                     |   0_lag_0 |
|:--------------------|----------:|
| 2000-01-01 00:00:00 |         0 |
| 2000-01-02 00:00:00 |         1 |
| 2000-01-03 00:00:00 |         2 |
| 2000-01-04 00:00:00 |         3 |
| 2000-01-05 00:00:00 |         4 |
```
As @danbartl described it in a discussion,
> the fix in my view just refers to the old specification we keep for legacy reasons
> but after discussion with Franz we changed the old notation, and noted that in the documentation
> warnings.warn(                "Specifying lag features via lag_config is deprecated since 0.12.0,"                + " and will be removed in 0.13.0. Please use the lag_feature notation"                + " (see the documentation for the new notation)."            )
> back then the logic for lag spefication was different
> I suggest not changing the old logic we just keep for deprecation reasons

PR #2647, opened 2022-05-15
[ENH] update TSR base class, kNN time series regression

fkiraly:
This PR upgrades the time series regression base class to the same patterns as time series classification.
Currently, a copy-paste-plus-adaptation. A future PR should probably refactor joint functionality to a joint mixin or intermediate base class.

As a prototype for the new base class design, this PR also:
* implements a modular KNN time series regressor, based on https://github.com/alan-turing-institute/sktime/pull/1998
* refactors the `TimeSeriesRegressor` to the new base class

Other legacy TSR classes were left as-is, except addition of empty `_fit` and `_predict` to satisfy abc which otherwise would raise an error.

The tests in `test_all_regressors` were refactored to the new class based format, redundant test removed, and new tests integrated with `check_estimator`.


TonyBagnall:
did this not get reviewed?

TonyBagnall:
I would rather you split the KNN regressor from the base class update. This knn does not use the distance factory, which means it needs updating for every new distance function, and it uses precomputed distance matrix, which I think relies on distances being metrics. DTW is not a metric, so this is not guaranteed to find the nearest neighbour, I think

fkiraly:
> did this not get reviewed?

I think it did not.

> This knn does not use the distance factory, which means it needs updating for every new distance function, and it uses precomputed distance matrix, which I think relies on distances being metrics. DTW is not a metric, so this is not guaranteed to find the nearest neighbour, I think

I think this is incorrect - it is modular, so you can pass any distance to its constructor. There is no need for updating, the design is extensible.

> and it uses precomputed distance matrix, which I think relies on distances being metrics. 

I don't understand why you think that, can you kindly explain? I think your statement is mathematically incorrect.
The distance matrix can have any values in it, including any combination of values obtained from pairwise functions that are not metrics.

> DTW is not a metric, so this is not guaranteed to find the nearest neighbour

That would depend on the algorithm being used. E.g., sth that leverages triangle equality vs exhaustive minimization (`brute`).
The algorithm can be selected via direct patchthrough to the sklearn interface, and in any case that would be based on the distance matrix in the first instance, as a mathematical object at least.

fkiraly:
> I would rather you split the KNN regressor from the base class update. 

More generally, is the statement you are implicitly making here that you would prefer a hypothetical, undesigned and unimplemented version of the knn regressor above the actual, implemented, tested, documented, and merged one, and that's why you would prefer to not have the one that exists, at all, and instead prefer to have none?

I don't think that's a good attitude.
I welcome your attempts at implementating knn and potential contributions, but I am opposed to blocking anyone else working in the area just because it's not exactly how you prefer it.

PR #2664, opened 2022-05-19
[ENH] Top-down reconciler based on forecast proportions

ciaran-g:
#### Reference Issues/PRs

Added the top down method based on forecast proportions to the Reconcilers. See #2157 

- The new g matrix function for top down - `_get_g_matrix_td_fcst()`
- An update function for the g matrix for this method, which changes for each forecast - `_update_td_fcst()`
- A function for returning a parent/child look up table for all methods - `_parent_child_df()`
- Re-format of the `_transform()` function to incorporate the new method, also uses column/index matching with `df.dot()` now
- Improvements to documentation

No


fkiraly (approved):
Great addition!

Well-structured, and follows existing pattern in estimator.

PR #2674, opened 2022-05-22
[ENH] Compositor for forecasting of exogeneous data before using in exogeneous forecaster

fkiraly:
This PR adds a compositor, `ForecastX`, that makes forecasts on exogeneous data before using that forecast to forecast endogeneous data.

The typical use case is exogeneous data which is available only up until current cutoff, but forecasts are made of exogeneous data to apply another forecaster to endogeneous data which requires exogeneous data at the forecasting horizon time stampe.

### DIFF
--- a/sktime/forecasting/compose/__init__.py
+++ b/sktime/forecasting/compose/__init__.py
@@ -11,6 +11,7 @@ __all__ = [
     "AutoEnsembleForecaster",
     "TransformedTargetForecaster",
     "ForecastingPipeline",
+    "ForecastX",
     "DirectTabularRegressionForecaster",
     "DirectTimeSeriesRegressionForecaster",
     "MultioutputTabularRegressionForecaster",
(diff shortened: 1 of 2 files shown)


aiwalter:
nice, I recently had the same idea to have sth like this, just forgot to create an issue. I was thinking about name of `ExogForecaster` but `ForecastX` or `ForecasterX` is also fine I think. There is just one edge case that I was thinking about, the case where you have some columns of `X_pred` given and only want to create `X` forecasts for some other missing columns. But could be done in a separate PR

LarsLZondag:
@fkiraly , how would this solution work if only a subset of the features in `X` should exhibit this behavior while the rest of the features should be used in a traditional manner. Would it be better to add another parameter to this transformer called `columns`, or something similar, in which the user can specify which features should be forecast into the future and which features are known in the future?

fkiraly:
Yes, @LarsLZondag, I guess adding a `columns` argument makes it easier.

I've thought how one could have the effect with `ColumnForecaster` or `ForecastingPipeline` etc but could think of no easy alternative way.

fkiraly:
>  There is just one edge case that I was thinking about, the case where you have some columns of `X_pred` given and only want to create `X` forecasts for some other missing columns. But could be done in a separate PR

@aiwalter, do you want to check whether I've addressed your edge case properly?

PR #2676, opened 2022-05-22
[DOC] Fixes in class description ExpandingWindowSplitter

keepersas:
Minor fixes in class description 'ExpandingWindowSplitter'. No code changes.

No

##### For all contributions
- [ x] The PR title starts with either [ENH], [DOC] or [BUG] indicating whether the PR topic is related to enhancement, documentation or bug

### DIFF
--- a/sktime/forecasting/model_selection/_split.py
+++ b/sktime/forecasting/model_selection/_split.py
@@ -943,7 +943,7 @@ class ExpandingWindowSplitter(BaseWindowSplitter):
 
     Split time series repeatedly into an growing training set and a fixed-size test set.
 
-    For example for `window_length = 5`, `step_length = 1` and `fh = 3`
+    For example for `initial_window = 5`, `step_length = 1` and `fh = 3`
     here is a representation of the folds::
 
     |-----------------------|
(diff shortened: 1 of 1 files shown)


fkiraly (approved):
Nice, thanks!

I was looking at it as well today and got confused.

Btw, do you know what happens if `start_with_window` is `False`?
I didn't understand what that parameter does. Might be worth explaining that better, in case you know what it does.

keepersas:
> Nice, thanks!
> 
> I was looking at it as well today and got confused.
> 
> Btw, do you know what happens if `start_with_window` is `False`? I didn't understand what that parameter does. Might be worth explaining that better, in case you know what it does.

It will make like 'initial window = 0'. However, 'initial window' must be > 1.
start_with_window = False

Reference example wit 'start_with_window = True':
start_with_window = True

I'll try to update the documentation to cover this option.

fkiraly:
> It will make like 'initial window = 0'. However, 'initial window' must be > 1.

That sounds pretty weird, it sounds like an unnecessary parameter.
Shouldn't this be controlled simply by `initial_window` then??

We should open an issue and possibly fix this in a separate PR.

fkiraly:
I opened an issue here:
https://github.com/alan-turing-institute/sktime/issues/2678
thanks for explaining what that strange parameter did!

fkiraly:
@keepersas, feel free to add yourself to allcontributor.src with a `doc` contribution!

PR #2683, opened 2022-05-23
[ENH] Adding more tests for probabilistic metrics

eenticott-shell:
Implements testing for all cases of input to the probabilistic tests. Also fixes some issues that were discovered whilst testing. Namely fringe cases of having 0.5 quantiles given to an interval metric, this creates an interval with 0 coverage which causes some failures in the tests so I now remove this column inside the tests with a warning.


fkiraly:
this is conflicting now, since it was not updated from the metrics, kindly resolve

PR #2686, opened 2022-05-23
[ENH] make_mock_estimator passing constructor args for the mocked class

### LINKED ISSUE #2685: [ENH] `make_mock_estimator` should allow passing constructor args for the mocked class
**Describe the bug**

The current implementation of `sktime.utils.estimators.make_mock_estimator` does not allow to pass any arguments to the returned Mock Estimator Class `__init__`. As a result, we can only use it to reproduce the behaviour of the estimator class with it's default arguments.

**To Reproduce**

```python
from sktime.utils.estimators import make_mock_estimator
y = load_airline()
mock_estimator_class = make_mock_estimator(NaiveForecaster)
mock_estimator_instance = mock_estimator_class(strategy=last, sp=1})
```

**Expected behavior**

There should be a way to pass the __init__ arguments of the input estimator class to the mock estimator.

ltsaprounis:
#### Reference Issues/PRs

fixes #2685 

This deals with the problem of not being able to pass the relevant arguments in the __init__ of the returned mock estimator class (described in #2685) by allowing the user to pass a dictionary that contains the kwargs of the input estimator class.

### DIFF
--- a/sktime/utils/estimators/_base.py
+++ b/sktime/utils/estimators/_base.py
@@ -56,6 +56,9 @@ def make_mock_estimator(
 ) -> BaseEstimator:
     r"""Transform any estimator class into a mock estimator class.
 
+    The returned class will accept the original arguments passed in estimator_class
+    __init__ as a dictionary of kwargs.
+
     Parameters
     ----------
     estimator_class : BaseEstimator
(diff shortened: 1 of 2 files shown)


fkiraly (approved):
Looks reasonable.

Minor issue, would this not break things like `get_params`, `set_params`, `repr`?

fkiraly:
Problem:

```python
mock_estimator_instance.get_params()
```
gives output
```
{'estimator_kwargs': {'strategy': 'last', 'sp': 1}}
```

but I'd expect
```
{'strategy': 'last', 'sp': 1}
```
!

ltsaprounis:
Yes it would, but I can't think of another way. 
Any suggestions are more than welcome!

fkiraly:
A hacky way would be overriding `get_params`, `set_params`?

fkiraly:
is there no way to have it behave exactly like the normal class?

ltsaprounis:
> is there no way to have it behave exactly like the normal class?

I bet there are many ways, but I simply cannot think of any.

I gave it a try with some extremely dodgy code: 
```python
    class _MockEstimator(estimator_class, _MockEstimatorMixin):
        args = inspect.getfullargspec(estimator_class.__init__).args
```

but in the end we would need to do something impossible like: 
```python
def __init__(self, *args, **kwargs)
```

fkiraly:
I know of some ways but they are close to hacking the python language. Another approach would involve https://smarie.github.io/python-makefun/

Maybe let's stay away from that for a moment, seems like a separate concern that may require a separate, principled solution first.

PR #2689, opened 2022-05-24
[ENH] deprecation warning for start_with_window in ExpandingWindowSplitter

atxtechbro:
Helps to make the code less confusing for the users and lets them set values in a more intuitive and direct manner. This PR lays the groundwork for the planned deprecation of `start_with_window` in `ExpandingWindowSplitter`. The tasks this PR seeks to realize are:

1. provide users the ability to set initial_window to 0 _and have it actually be set to 0_
2. displays a new `FutureWarning` message prompted when the user sets the default for `start_with_window` to be `False` instead of the default value which is `True`. This way we can warn our users about the coming change and allow them time to make the necessary changes.

### DIFF
--- a/sktime/forecasting/model_selection/_split.py
+++ b/sktime/forecasting/model_selection/_split.py
@@ -1178,11 +1178,16 @@ class ExpandingWindowSplitter(BaseWindowSplitter):
         super(ExpandingWindowSplitter, self).__init__(
             fh=fh,
             window_length=initial_window,
-            initial_window=None,
+            initial_window=initial_window,
             step_length=step_length,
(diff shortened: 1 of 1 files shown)


fkiraly (changes requested):
Thanks!

Though, I think the logic here is not correct?
Have you tested this for the `initial_window=0` case?

This is also not deprecation ready.
There need to be comments on what to do at deprecation.
For instance, you could not simply remove the `start_with_window` arg as it is passed inside the function. I would recommend you build some code at the start of the function, which can be removed together with the argument, and ensures the rest of the code works with the new argument set.

PR #2677, opened 2022-05-22
[MNT] Remove hcrystalball dependency

aiwalter:
#### Reference Issues/PRs
#2675 

#### What does this implement/fix? Explain your changes.
Remove `hcrystalball `. To be merged before release v0.13.0 and after v0.12.0

PR #2694, opened 2022-05-25
[ENH] Address deprecation of `freq` and `freqstr` in `pd.Timestamp`

### LINKED ISSUE #1750: [ENH] address FutureWarning: Timestamp.freq is deprecated and will be removed in a future version

### LINKED ISSUE #2617: [MNT] replace deprecated timestamp.freqstr throughout sktime
A very large amount of deprecation messages around `timestamp.freqstr` are raised throughout the CI/CD, especially around the `ForecastingHorizon`:

```
FutureWarning: Timestamp.freqstr is deprecated and will be removed in a future version.
```

We should address this, as the `ForecastingHorizon` is a central object, so it would be catastrophic if it breaks.

At the moment, deprecation seems to be in the future since there is no specified version number yet, but we should prepare and take care of it earlier than later.

khrapovs:
#### Reference Issues/PRs

Closes #1750

Prohibit the use of `pd.Timestamp` type for `cutoff` in `ForecastingHorizon`.

None

This solves the problem of extracting frequency from an object that will not have this property in some future (see this comment https://github.com/pandas-dev/pandas/issues/15146#issuecomment-273266806 and the whole deprecation PR for details). Although the problem is solved, the solution itself is far from being ideal. The major negative consequence is the mismatch in types between a time series index, which is normally `pd.DatetimeIndex`, and the absolute forecasting horizon of `pd.PeriodIndex` obtained given `pd.Period` cutoff. This happens when the cutoff is taken to be one of the `y.index` (`pd.DatetimeIndex`) elements, which is `pd.Timestamp`. This is not allowed, so we have to convert the cutoff to `pd.Period`. The result of `fh.to_absolute(cutoff).to_pandas()` is then a `pd.PeriodIndex`.

Ideally, this conversion should happen under the hood, without a user noticing it. But it seems impossible without major refactoring. In particular, `ForecastingHorizon` is not aware of the time series itself and its frequency. So it has to obtain the frequency somehow. In the current implementation it is only `cutoff`. An alternative to consider would be to pass frequency explicitly as an argument to `ForecastingHorizon`.

#### Any other comments?

This is very controversial and touches the very core of the library, so I hope for a very careful consideration of this PR.

##### For all contributions
- [X] The PR title starts with either [ENH], [DOC] or [BUG] indicating whether the PR topic is related to enhancement, documentation or bug


khrapovs:
This is in DRAFT mode. Here I am just having fun. This is an experiment. What will happen if we prohibit pd.Timestamp completely 👹

khrapovs:
> @khrapovs, we can´t remove hcrystalball without deprecation. That will happen in 0.13.0 via #2677.

Sure, sure. I assume that this PR will take A LOT of time to merge, if at all. So, I decided to remove some obstacles in advance.

khrapovs:
For the record: this commit https://github.com/alan-turing-institute/sktime/pull/2694/commits/e696757a430db239df4c14be48b80e54970b9164 has no mention of `Timestamp.freq` amonth future warnings (https://github.com/alan-turing-institute/sktime/runs/6705874624?check_suite_focus=true).

fkiraly:
Looks like your personal vendetta against those deprecation messages has come to a spectacular end... Thanks!

sktime/datatypes/_utilities.py
```diff
@@ -109,50 +110,58 @@ def get_cutoff(obj, cutoff=0, return_index=False):
     if len(obj) == 0:
         return cutoff
 
+    obj_copy = copy(obj)
```
fkiraly, line 113: why do we copy, and how does thie relate to `freqstr` etc?

fkiraly: I thought this method had no side effects

fkiraly: so might be worth testing for that instead of adding a `copy` if you doubt it

sktime/datatypes/_utilities.py
```diff
@@ -109,50 +110,58 @@ def get_cutoff(obj, cutoff=0, return_index=False):
...
+            cutoff_ind = obj_copy.shape[0] + cutoff
         if return_index:
             return pd.RangeIndex(cutoff_ind - 1, cutoff_ind)
         else:
             return cutoff_ind
 
-    if isinstance(obj, pd.Series):
-        return obj.index[[-1]] if return_index else obj.index[-1]
+    if hasattr(obj_copy, "index") and isinstance(obj_copy.index, pd.DatetimeIndex):
+        obj_copy.index = obj_copy.index.to_period()
```
khrapovs, line 126: @fkiraly This is why I need the copy. That's the new side effect.

sktime/datatypes/_utilities.py
khrapovs: > why do we copy, and how does thie relate to `freqstr` etc?

This method returns the index of cutoffs. In this PR all of them have to be either `int` or `pd.Period`, not `pd.Timestamp`.

fkiraly (changes requested):
@khrapovs, this is great work, really nice! You seem to have identified the main sources of the `freqstr` problem complex!

My main concern is the double-D, design & deprecation.

What I mean:
* you are performing what looks like extensive surgery to public interface elements of `ForecastingHorizon`. This cannot be done without proper deprecation.
* Having said that, is there maybe not a better way to keep the high-level functionality of `ForecastingHorizon` intact? You are partially "exploding" `ForecastingHorizon`, by moving the "time like" index treatment from `ForecastingHorizon` methods to loose methods, e.g., `to_absolute` being replaced by `convert_fh_to_datetime_index` im places. Can that not be kept back in, but treated by dispatch? Or by a less invasive re-design?
* if we do a major re-design, I think we need a design document, including a clean deprecation strategy. I am aware that there is no original design document for `ForecastingHorizon`, which is coming from legacy times, but given its importance it might be good to do some bookkeeping going forward.

khrapovs:
@fkiraly Thanks a lot for your suggestions!

> * you are performing what looks like extensive surgery to public interface elements of `ForecastingHorizon`. This cannot be done without proper deprecation.

Sure. But let's first agree on what we are ready to deprecate. In the light of the comments below, I am not sure that this PR will survive in the current state.

> * Having said that, is there maybe not a better way to keep the high-level functionality of `ForecastingHorizon` intact? You are partially "exploding" `ForecastingHorizon`, by moving the "time like" index treatment from `ForecastingHorizon` methods to loose methods, e.g., `to_absolute` being replaced by `convert_fh_to_datetime_index` im places. Can that not be kept back in, but treated by dispatch? Or by a less invasive re-design?

I am stuck exactly at the same place. I really want to keep all interfaces "as is", but it is very hard to find a minimal non-invasive change. Especially given the fact that tests run for almost an hour :D

So the main problem is again the following. On master `ForecastingHorizon` needs to know the frequency. Currently, this bit of information is obtained from `cutoff` (either `int`, or `pd.Timestamp`, or `pd.Period`), which in turn is derived from one of the items in `y.index`. Prohibiting `pd.Timstamp` means we have to do the conversion somewhere. `pd.DatetimeIndex` has frequency, so `y.index.freq` is still perfectly valid. Or, `cutoff.to_period(freq=y.index[-1])` is also ok. In both cases we need the time series itself. But obviously, we do not want to pass the whole `y` to `ForecastingHorizon`. So we pass its frequency together with `cutoff`. In my PR `pd.Timestamp` is prohibited, and so the outcome of `.to_absolute` can not be `pd.DatetimeIndex` which breaks type connection with `y`.

Another "cleaner" option is to prohibit `pd.DatetimeIndex` for `y.index`, but I did not even propose it since it sounds too crazy.

Dispatch? I am not familiar with this concept. Started reading n it. Looks applicable. Let me think about it...

> * if we do a major re-design, I think we need a design document, including a clean deprecation strategy. I am aware that there is no original design document for `ForecastingHorizon`, which is coming from legacy times, but given its importance it might be good to do some bookkeeping going forward.

I have never written "design documents". Do you have an example I can follow?

Thanks!

khrapovs:
@fkiraly What would you say to yet another option: new optional argument `freq` in all `ForecastingHorizon` methods that internally require frequency? The consequence is mostly this:
```
fh_abs = fh.to_absolute(y.index[-1], freq=y.index.freq).to_pandas()
```
Please, let me know if it looks less invasive, and then I can try implementing it as an alternative PR.

khrapovs:
@fkiraly Here are some notes that I managed to write before I leave for my vacation: https://github.com/sktime/enhancement-proposals/pull/21

fkiraly:
> I have never written "design documents". Do you have an example I can follow?

example, as discussed in our call:
https://github.com/sktime/enhancement-proposals

fkiraly:
suggested additions:
https://github.com/sktime/enhancement-proposals/pull/22

My preference would be option 2a' with some kind of `_check_freq`

khrapovs:
> My preference would be option 2a' with some kind of `_check_freq`

Cool. I like it too. Will implement as soon as I am back from vacation (June 20).

fkiraly:
> Cool. I like it too. Will implement as soon as I am back from vacation (June 20).

Ok, so I guess 0.12.0 will still torture our users with those `pandas` deprecation messages...
but don't worry, enjoy your vacation!

khrapovs:
This PR will most likely be replaced by https://github.com/alan-turing-institute/sktime/pull/2834

khrapovs:
Was solved by #2932

PR #2699, opened 2022-05-25
[ENH] transformer adaptor for `pandas` native methods

fkiraly:
This PR adds a transformer adaptor for `pandas` native methods.

This exposes applications of pandas methods as an `sktime` transformer.

### DIFF
--- a/sktime/transformations/series/adapt.py
+++ b/sktime/transformations/series/adapt.py
@@ -12,7 +12,7 @@ from sktime.transformations.base import BaseTransformer
 
 
 class TabularToSeriesAdaptor(BaseTransformer):
-    """Adapt scikit-learn-like  transformations to time series setting.
+    """Adapt scikit-learn transformation interface to time series setting.
 
     This is useful for applying scikit-learn :term:`tabular` transformations
     to :term:`series <Time series>`, but only works with transformations that
@@ -23,7 +23,7 @@ class TabularToSeriesAdaptor(BaseTransformer):
         ``fit(X)`` fits a clone of ``transformer`` to X (considered as a table)
         ``transform(X)`` applies transformer.transform to X and returns the result
         ``inverse_transform(X)`` applies tansformer.inverse_transform to X
-    If fit_in_transform = Truee and X is a series (pd.DataFrame, pd.Series, np.ndarray):
+    If fit_in_transform = True and X is a series (pd.DataFrame, pd.Series, np.ndarray):
         ``fit`` is empty
         ``transform(X)`` applies transformer.fit(X).transform.(X) to X,
             considered as a table, and returns the result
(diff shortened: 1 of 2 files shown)


sktime/transformations/series/adapt.py
```diff
@@ -201,3 +201,188 @@ def get_test_params(cls, parameter_set="default"):
...
+            Parameters to create testing instances of the class
+            Each dict are parameters to construct an "interesting" test instance, i.e.,
+            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+            `create_test_instance` uses the first (or only) dictionary in `params`
+        """
+        params1 = {"method": "diff"}
+        params2 = {"method": "diff", "kwargs": {"periods": 2}, "apply_to": "all_subset"}
+
+        def add1(x):
+            return x + 1
```
ltsaprounis, line 380: Looks like this is not used anywhere

ltsaprounis (changes requested):
This is really useful! 
I currently don't see how the the transformation is applied by instance in the hierarchical case. I would expect to see a groupby or something for that case. 

A few tests would be useful too, in case we want to refactor later etc.!

fkiraly:
> I currently don't see how the the transformation is applied by instance in the hierarchical case. I would expect to see a groupby or something for that case.

The framework does that automatically, since the internally supported scitype is `Series` only.

I just added it in the docstring, so users do not get confused, in case they pass a hierarchical dataframe and expect the output to be applied to the entire `DataFrame`.

sktime/transformations/series/adapt.py
fkiraly: indeed. Wanted to use that in the `transform` case, but the pickling test fails, since functions apparenctly cannot be pickled.

fkiraly:
> A few tests would be useful too, in case we want to refactor later etc.!

What would you suggest, apart from the interface tests?

Validity tests, checking against the `pandas` operation? Coming up.

ltsaprounis:
> The framework does that automatically, since the internally supported scitype is `Series` only.

Ha! yes, I missed that. Fair enough.

> Validity tests, checking against the `pandas` operation? Coming up.

Exactly! + some on hierarchical as well!

sktime/transformations/series/adapt.py
fkiraly: removed

fkiraly:
> Exactly! + some on hierarchical as well!

That is already tested via the generic vectorization tests in `test_base`, I don't think we need to test this per estimator?

fkiraly:
@ltsaprounis, ready for re-review.

Added tests, and guess what - I found a bug, for one of the three parameter settings, the output was inconsistent with the intention.

PR #2706, opened 2022-05-27
[ENH] Conformal intervals additions 

bethrice44:
**Reference Issues/PRs**
Building on https://github.com/alan-turing-institute/sktime/pull/2542.

**What does this implement/fix? Explain your changes.**

1. Move `_compute_sliding_residuals` to it's own function like in `NaiveVariance`
2. Add early computation of residuals matrix like in `NaiveVariance`
3. Add option to pass a `sample_frac` argument at initialisation to speed up computing the residuals matrix
4. Found an issue where `forecaster.predict_residuals` doesn't always return a series, so assigning it to the residuals matrix can fail here:
```
residuals_matrix.loc[id] = forecaster.predict_residuals(y_test, X_test)
```

Fixed with:
```
residuals_matrix.loc[id] = forecaster.predict_residuals(y_test, X_test).iloc[:,0]
```

5. Add `get_slice` to `datatypes._utilities` and use to abstract slicing logic so it always has exclusive end, for use in `conformal.py`


sktime/datatypes/_utilities.py
```diff
@@ -290,3 +290,69 @@ def get_window(obj, window_length=None, lag=0):
     raise ValueError(
         "bug in get_latest_window, unreachable condition, ifs should be exhaustive"
     )
+
+
+def get_slice(obj, start=None, end=None):
+    """Slice obj with start (inclusive) and end (exclusive) indices.
+
+    Returns time series or time series panel with time indices
+        strictly greater than cutoff - lag - window_length, and
```
fkiraly, line 299: not correct, needs to be fixed

sktime/datatypes/_utilities.py
```diff
@@ -290,3 +290,69 @@ def get_window(obj, window_length=None, lag=0):
...
+        equal or less than cutoff - lag.
+    Cutoff if of obj, as determined by get_cutoff.
+
+    Parameters
+    ----------
+    obj : sktime compatible time series data container or None
+        if not None, must be of one of the following mtypes:
+            pd.Series, pd.DataFrame, np.ndarray, of Series scitype
```
fkiraly, line 307: this is also not correct, no?

Currently, slicing works only if `obj` is one of the `Series` types.

It will break if the `pd.DataFrame` that reaches line 342ff has a row `MultiIndex`.

You don't need to generalize this, but the docstring should be correct about the assumptions.

fkiraly (changes requested):
Docstring changes are great, thanks!

You also added an example to the docstring, also nice!

Unfortunately, that example is now failing (we test all docstring examples to actually run), could you kindly investigate?

fkiraly:
This code is minimally failing (identical with your doc example):

```python
from sktime.forecasting.conformal import ConformalIntervals
y = load_airline()
forecaster = NaiveForecaster(strategy="drift")
conformal_forecaster = ConformalIntervals(forecaster)
conformal_forecaster.fit(y, fh=[1,2,3])
pred_int = conformal_forecaster.predict_interval()
```

it is also not failing on `main`, currently.

sktime/datatypes/_utilities.py
```diff
@@ -290,3 +290,75 @@ def get_window(obj, window_length=None, lag=0):
...
+def get_slice(obj, start=None, end=None):
+    """Slice obj with start (inclusive) and end (exclusive) indices.
+
+    Returns time series or time series panel with time indices
+        strictly greater and equal to start index and less than
+        end index.
+
+    Parameters
+    ----------
+    obj : sktime compatible time series Series type or None
```
fkiraly, line 304: can you make sure this is correct? Currently it will be incorrect if it is numpy, because no slicing is done

fkiraly (changes requested):
Looks great!

Small exception which we should nonetheless adress before merging: `get_slice` docstring claims correctness for input types which are currently not supported. E.g., if you pass a numpy array, then no slicing takes place. The type will be ok (so the test for output type passes), but the content will be wrong (you do not test that).

fkiraly (approved):
Thanks for clarifying on slack, @bethrice44.
I made my comment in error, if `obj` is `np.ndarray`, and `start` or `end` are `int`,  the function works as intended.

Having some input checks on `start` or `end` and some commenting around line 346 (this is the "twice `int`" case) would be useful, because this is where I got confused, but the lack thereof is not a blocking problem. So I approve.

sktime/datatypes/_utilities.py
fkiraly: erroneous comment by me, see below.

PR #2708, opened 2022-05-28
[DOC] Added list of interns to website

aiwalter:
Added list of Interns to website.


fkiraly (changes requested):
Nice!

I think though we should also include mentees?

Further, you are missing Taiwo from Outreachy.

aiwalter:
Taiwo didnt answer if she is fine with it, mentors we would have to ask them first. So feel free to do so and make a PR

fkiraly:
we could just make PRs and ask for confirmation?

The information is not secret, they have already put it there as issues on the `mentoring` repo.

aiwalter:
Lets merge this already

fkiraly:
I think from multiple perspectives, it would be better to include all the mentees, not just paid interns.

Perspectives from which I see benefits: inclusivity towards the individuals; branding and visibility of `sktime` mentoring; completeness of information.

aiwalter:
But this is the sponsoring section

fkiraly:
> But this is the sponsoring section

Ah, I see. Then, we should include Taiwo at least.
I would also add a link to the issue tracker of the `mentoring` repo, which has details about the internships.

aiwalter:
@thayeylolu pls confirm

aiwalter:
@fkiraly lets merge this then

fkiraly:
merge it?

PR #2719, opened 2022-05-29
[MNT] testing sktime without softdeps

fkiraly:
This PR adds a CI element, `test-nosoftdeps`, which tests `sktime` working correctly without any soft dependencies installed.

Currently, this selectively retrieval functionality in `registry`, which is, possibly, the only major piece of functionality that users might expect to work if soft dependencies are not installed, but which has in the past been subject to failures due to erroneous soft dependency handling in `sktime` internal module imports.

To avoid test proliferation, we only test under the condition: no soft deps installed, python 3.9, ubuntu OS.


fkiraly:
@lmmentel, @freddyaboulton, I must be doing something trivial wrong - I want to run the tests in `registry` only, without having `pytest` try to run all other files which have imports. Any ideas what it is?

fkiraly:
>@lmmentel, @freddyaboulton, I must be doing something trivial wrong - I want to run the tests in `registry` only, without having `pytest` try to run all other files which have imports. Any ideas what it is?

Ah, it seems to work now. Secret was, apparently, module separator being `.` and not `/`.

PR #2727, opened 2022-05-31
[MNT] Add [MNT] tag to PR template

khrapovs:

Add [MNT] tag to PR template.


aiwalter:
maybe we need a definition of "maintenance"? Like eveything that is not affecting code?

aiwalter (approved):

fkiraly:
> Like eveything that is not affecting code?

Well, "infrastructure as code" is the best practice ideal, so subject to that best practice ideal, all maintenance matters would affect some code?

aiwalter:
I meant the core package code

Issue #2730, opened 2022-06-01
[BUG] ARIMA model sarimax_kwargs doesn't exist after fit 


ilkersigirci:
**Describe the bug**

ARIMA model has `**sarimax_kwargs` as init parameters but after fitting the model, they don't exist in the model anymore. `sarimax_kwargs` just returns an empty dictionary.

**To Reproduce**

```python
import numpy as np
from sktime.datasets import load_airline
from sktime.forecasting.arima import ARIMA

y = load_airline()

model = ARIMA(
    order=(5, 1, 4),
    seasonal_order=(3, 0, 3, 7),
    suppress_warnings=True,
    enforce_stationarity=False,
)
print(model.__dict__)

model.fit(y=y)
print(model.__dict__)
```
**Expected behavior**

After fitting, `enforce_stationarity=False`,which is an [sarimax_kwargs parameter](https://alkaline-ml.com/pmdarima/modules/generated/pmdarima.arima.ARIMA.html), should be included in sarimax_kwargs dictionary

**Additional context**

**Versions**
<details>

System:
python: 3.8.13
machine: Linux-5.13.0-44-generic-x86_64-with-glibc2.10

Python dependencies:
pip: 22.1.1
setuptools: 62.3.2
sklearn: 1.0
sktime: 0.11.4
statsmodels: 0.13.2
numpy: 1.21.5
scipy: 1.8.1
pandas: 1.3.5
matplotlib: 3.5.2
joblib: 1.1.0
numba: 0.55.2
pmdarima: 1.8.5
tsfresh: None

</details>

fkiraly:
I've verified the bug.

This is due to non-compliance of `ARIMA` with the scikit-learn interface (the parameter access part), where all parameters in the constructor need to be written out, or otherwise will not be detected by parameter access methods and funniness may happen.

The solution is to write out all the sarimax kwargs in the constructor, instead of catching them in a kwargs.

We should also test for this, since it can lead to problems.

fkiraly:
This should help: https://github.com/alan-turing-institute/sktime/pull/2731

What would be appreciated is a second set of eyes on the defaults for these parameters.
Moving them to the construtor did require to set defaults, and the docstrings of `pmdarima` are not 100% clear there.

ilkersigirci:
Thank you for the quick fix, everything looks good to me now

fkiraly:
well, thanks for checking! Will fix.

fkiraly:
[...]

PR #2732, opened 2022-06-01
[ENH] interface compliance test to ensure sklearn compliance of constructor

fkiraly:
This PR makes the following changes:
* reactivates the `check_constructor` test, which was not executed since the test refactor since we forgot to rename it to start with `test`
*  adds a test to `TestAllEstimators` which checks for interface non-compliance along the lines of the bug #2730, i.e., presence of a `**kwarg` signature element in `Class.__init__`.

Why does having `kwargs` signature elements mean interface non-compliance? `sktime` aims to be compliant with `scikit-learn`, and in particular the nested parameter interface of `scikit-learn`. That assumes that all parameters in the class constructors are explicitly named keyword arguments, and parameters exposed by the `scikit-learn` parameter interface are precisely the named keyword arguments. Passing parameters via unnamed keyword arguments is not caught by the parameter inspection and setter interface of `scikit-learn` or `sktime`. So far, this has been possible without setting off a test, and could lead to unexpected and buggy behaviour, as #2730 shows.

Before the fix #2731 for #2730 is merged, this correctly catches out `ARIMA` as non-compliant.

Relies on the fixes in #2773, of #2733, to test that these are indeed fixes.


fkiraly:
Well, looks like it caught much more...
https://github.com/alan-turing-institute/sktime/issues/2733

ltsaprounis (approved):

PR #2735, opened 2022-06-01
[ENH] automated init args write to self by `BaseObject`

### LINKED ISSUE #2733: [BUG] fixing constructor non-compliance in various estimators
The following estimators fail the extended `test_constructor` from https://github.com/alan-turing-institute/sktime/pull/2732 and need to be fixed, along the lines of the previously undetected bug #2730. Each fix can be modelled upon #2731.

- [x] `ARIMA` https://github.com/alan-turing-institute/sktime/pull/2731
- [ ] `AutoARIMA` 
- [x] `AutoETS` https://github.com/alan-turing-institute/sktime/pull/2736
- [ ] `KNeighborsTimeSeriesClassifier`
- [x] `KNeighborsTimeSeriesRegressor` https://github.com/alan-turing-institute/sktime/pull/2737
- [x] `PCATransformer` https://github.com/alan-turing-institute/sktime/pull/2734
- [ ] `ShapeDTW`
- [ ] `TimeSeriesKMeans`
- [ ] `UnobservedComponents`

fkiraly:
Trying to fix #2733, it is truly atrocious to write long blocks of `self.foo = foo` and then `self.bar = bar` and then `self.supercalifragilistic = supercalifragilistic`.

Should't all estimators just write all `__init__` args to self automatically?

This is now done by `BaseObject`, and writes of parameters to `self` are no longer necessary.

As a proof-of-concept, `AutoETS` is refactored by removing the self writes.

### DIFF
--- a/sktime/base/_base.py
+++ b/sktime/base/_base.py
@@ -70,7 +70,12 @@ class BaseObject(_BaseEstimator):
     """
 
     def __init__(self):
+        # initializing dynamic tag dictionary
         self._tags_dynamic = dict()
+        # this writes all __init__ args to self
+        for key in inspect.getfullargspec(self.__init__).args:
+            setattr(self, key, eval(key))
+
         super(BaseObject, self).__init__()
 
     def reset(self):
--- a/sktime/forecasting/ets.py
+++ b/sktime/forecasting/ets.py
@@ -204,37 +204,7 @@ class AutoETS(_StatsModelsAdapter):
         ignore_inf_ic=True,
         n_jobs=None,
         random_state=None,
-        **kwargs
     ):
-        # Model params
-        self.error = error
-        self.trend = trend
-        self.damped_trend = damped_trend
-        self.seasonal = seasonal
-        self.sp = sp
-        self.initialization_method = initialization_method
-        self.initial_level = initial_level
-        self.initial_trend = initial_trend
-        self.initial_seasonal = initial_seasonal
-        self.bounds = bounds
-        self.dates = dates
-        self.freq = freq
-        self.missing = missing
-
-        # Fit params
-        self.start_params = start_params
-        self.maxiter = maxiter
-        self.full_output = full_output
-        self.disp = disp
-        self.callback = callback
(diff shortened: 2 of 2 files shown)


ilkersigirci:
I believe [dataclass](https://docs.python.org/3/library/dataclasses.html) decorator exactly fulfill this purpose. It automatically  assigns self to its init arguments

```python
from dataclasses import dataclass

@dataclass
class InventoryItem:
    """Class for keeping track of an item in inventory."""
    name: str
    unit_price: float
    quantity_on_hand: int = 0
```

fkiraly:
well, @ilkersigirci, I did consider that, or decorating the `__init__` itself.
Is there any way that this is done automatically, without decorating?

Because we also need to look at the "extender" contract, i.e., people implementing their own estimators.
Having to add a decorator in the right place is not obvious. Inheriting from a base class is simpler and easier to explain.

fkiraly:
maybe decorating `__init__` and call it `write_args_to_self` would be more obvious?

ilkersigirci:
> well, @ilkersigirci, I did consider that, or decorating the `__init__` itself.
> Is there any way that this is done automatically, without decorating?
> 
> Because we also need to look at the "extender" contract, i.e., people implementing their own estimators.
> Having to add a decorator in the right place is not obvious. Inheriting from a base class is simpler and easier to explain.

Not that I know of.

ilkersigirci:
> maybe decorating `__init__` and call it `write_args_to_self` would be more obvious?

Nice thought. If it doesn't break sklearn/sktime api reference, I think it would significantly simplify `__init__` boilerplate codes.

fkiraly:
>If it doesn't break sklearn/sktime api reference, I think it would significantly simplify __init__ boilerplate codes.

Well, that's precisely what's worrying me...
`sklearn` does a lot of direct signature inspection - which is really bad style imo, but we can't do away with it since it would break compatibility. A decorator might break those inspection interfaces.

fkiraly:
approach does not work, closing

Issue #2670, opened 2022-05-20
[ENH] MultiplexTransformer Discussion


miraep8:
During the community meetup today @aiwalter and @indinewton suggested it would be useful to create this issue to discuss the implementation of a  `MultiplexTransformer`.  After getting a chance to check out `OptionalPassthrough` I think we might not need build this thing at all (I agree it could be a bad idea to have multiple areas of the code base with such overlapping/similar functionality).  But - it did leave me with some questions/ideas about some potential use cases that might be worth expanding the current model to support: 

#### Some Background:

- The original idea  to make a `MultiplexTransformer` came from a suggestion in #2459 that it would be nice to create a `MultiplexTransformer` to complement the `MultiplexForecaster`.
- As of #2612 we have implemented a `_DelegatedTransformer` to facilitate implementation of `MultiplexTransformer` in a similar way to how `MultiplexForecaster` has been written. 
- As was mentioned in the community meeting however - it seems like the `OptionalPassthrough` transformer can already do most of what we would want to do with the `MultiplexTransformer` though performing a grid search over its `passthrough` variable.  (A solution which I was not aware of, thank you for bringing it up!)

#### Some remaining questions/ideas I had:

-  **Question 1 - What if we want to use exactly 1 Transformer from a list of `n` transformers?**.  It seems to me with the current system if we did something like this (borrowed from the docstring example for `OptionalPassthrough`):

```
pipe = TransformedTargetForecaster(steps=[
    ("deseasonalizer", OptionalPassthrough(Deseasonalizer())),
    ("scaler", OptionalPassthrough(TabularToSeriesAdaptor(StandardScaler()))),
    ("forecaster", NaiveForecaster())])
# putting it all together in a grid search
cv = SlidingWindowSplitter(
    initial_window=60,
    window_length=24,
    start_with_window=True,
    step_length=48)
param_grid = {
    "deseasonalizer__passthrough" : [True, False],
    "scaler__transformer__transformer__with_mean": [True, False],
    "scaler__passthrough" : [True, False],
    "forecaster__strategy": ["drift", "mean", "last"]}
gscv = ForecastingGridSearchCV(
    forecaster=pipe,
    param_grid=param_grid,
    cv=cv,
    n_jobs=-1)
gscv_fitted = gscv.fit(load_airline())
```
[...]

PR #2738, opened 2022-06-01
[ENH] Add MultiplexTransformer

miraep8:
#### Reference Issues/PRs

Related to #2670 and #2459.  Note this implementation is meant to be very similar to how `MultiplexForecaster` currently works - see #2458 making use of the `_DelegatedTransformer` in #2612

With `MultiplexTransformer` one now in theory has a new way to multiplex over multiple transformers.  (Note one could optionally include various transformers with the pre-existing `OptionalPassthrough` but `MultiplexTransformer` should be useful when people want exactly one transformer.

No

As it stands - it seems the easiest use case is to bundle `MultiplexTransformer` into either `TransformedTargetForecaster` or `ForecastingPipeline` and then evaluate which transformer is best using `ForecastingGridCV`.  However - one might imagine that you could wish to use `MultiplexTransformer` outside of the forecasting context.  Which begs the question - is it worth building some sort of `GridSearch` functionality into transformers as well? If so - what might be the best way to "evaluate" which transformer is best? Just wondering if this would be worth opening an issue for. 

##### For all contributions

##### For new estimators
I guess this is technically a new estimator.... I did not do these things.  But, I can. Lmk! 


fkiraly:
> Which begs the question - is it worth building some sort of `GridSearch` functionality into transformers as well? If so - what might be the best way to "evaluate" which transformer is best? Just wondering if this would be worth opening an issue for. 

That's an interesting question.
I think the natural use case of `MultiplexTransformer` etc is together with `GridSearch`.

The non-obvious observation is that the multiplexer can be *separated* from grid search, and perhaps should, because - as you say! - transformers appear in pipelines of different estimator types.

For me, the key argument whether to separate or merge would be "parsimony of specification" and avoiding "proliferation of classes" (these two design principles do not always give the same answer).

Do you have any arguments or examples in favour of certain merges?

miraep8:
And clearly there is an ongoing effort to expand the pipelining functionality to other applications as well!  I didn't mean to suggest there wasn't - just suggesting that a more general `GridSearch` functionality should maybe be part of that conversation too! :)

fkiraly:
> I was just wondering if a more general `GridSearch` functionality should maybe be part of that conversation too! :) 

Absolutely!
Though, I would have done this by know if I knew what the best architecture would be.

My favoured option right now is to do it via a generic `benchmark` utility, which is called internally and dispatches by estimator type internally.

Although, that would heavily depend on progress in the benchmarking workstream, and at least some amount of design discussion imo.

miraep8:
Yea I knew the tests were failing!  I sort of buried the reference to that  in the bottom of my essay on `GridSearch`. But..... they aren't failing now!

And regarding opening an issue - I agree it would be a better place for this discussion. I can do that, though wondering if it might be worth waiting until the benchmarking workstream is more established as part of a 'benchmarking wishlist'?

sktime/transformations/multiplexer.py
```diff
@@ -0,0 +1,253 @@
...
+# copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+"""Implements forecaster for selecting among different model classes."""
+
+from sklearn.base import clone
+
+from sktime.base import _HeterogenousMetaEstimator
+from sktime.transformations._delegate import _DelegatedTransformer
+from sktime.transformations.base import BaseTransformer
+
+__author__ = ["miraep8"]
```
GuzalBulatova, line 12: Also, don't forget to update `CODEOWNERS` and `docs/source/api_reference/transformations.rst` :)

GuzalBulatova (commented):
Great work, Mirae, very well written code! I only have minor notes.

sktime/transformations/tests/test_multiplexer.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def test_multiplex_transformer_alone():
+    """Test behavior of MultiplexTransformer.
+
+    Because MultiplexTransformer is in many ways a wrapper for an underlying
+    transformer - we can confirm that if the selected_transformer is set that the
+    MultiplexTransformer delegates all its transformation responsibilities as expected.
+    """
+    from numpy.testing import assert_array_equal
```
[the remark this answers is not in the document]
miraep8: Good catch!

sktime/transformations/tests/test_multiplexer.py
```diff
@@ -0,0 +1,112 @@
...
+        test_transformer = clone(transformers[ind])
+        y_transform_indiv = test_transformer.fit_transform(X=y)
+        multiplex_transformer.selected_transformer = name
+        # Note- MultiplexForecaster will make a copy of the forecaster before fitting.
+        y_transform_multi = multiplex_transformer.fit_transform(X=y)
+        assert_array_equal(y_transform_indiv, y_transform_multi)
+
+
+def _find_best_transformer(forecaster, transformers, cv, y):
+    """Will evaluate all the forecasters on y and return the name of best."""
```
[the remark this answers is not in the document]
miraep8: Yes, that is definitely the right way to do it! :) Thank you!

sktime/transformations/multiplexer.py
miraep8: Thank you for the reminder on the CODEOWNERS bit.  And I actually haven't changed the docs directly like this before, so the pointer was really helpful.  I will leave this as unresolved until I can check out how the read the docs build looks after the changes. :)

fkiraly:
>@fkiraly, I totally agree - I do think that it makes sense to use `MultiplexTransformers` with some sort of a `GridSearch`, but does it have to be the forecasting one?  I guess to me the fact that if you want to do a grid search you are forced into a choice - either use the sklearn `GridSearch`, or you can use the `GridSearch` functionality under forecasting -> model selection seems odd. 

I think the "natural" way based on the current design would be to wrap parts of a pipeline possibly in multiplexers, and then use the type of grid search that is specific for the estimator type. E.g., `MultiplexTransformer` in a `TransformedTargetForecaster`, and then the forecaster grid search.

> Idk, I guess it was striking me as somewhat strange to have these 2 different ways of approaching `GridSearch` when one might expect a more unified approach?

Yes, I think ultimately, having similar syntax for grid search *and* multiplexer would be the best end state. Having the same dunder symbol but dispatching by type comes close.

sktime/transformations/multiplexer.py
```diff
@@ -0,0 +1,253 @@
...
+__all__ = ["MultiplexTransformer"]
+
+
+class MultiplexTransformer(_DelegatedTransformer, _HeterogenousMetaEstimator):
+    """Facilitate an AutoML based selection of the best transformer.
+
+    When used in combination with either TransformedTargetForecaster or
+    ForecastingPipeline in combination with ForecastingGridSearchCV MultiplexTransformer
+    provides a framework for transformer selection.  It can be used with exogenous
+    transformers (ie data to be applied to exogenous X variables) or transformers on
```
fkiraly, line 22: perhaps clarify better - there is no such thing as an "exogeneous transformer", being exogeneous is not a property of the transformer but a contextual property, i.e., of how it is used within the different pipeline constructs.

sktime/transformations/multiplexer.py
```diff
@@ -0,0 +1,253 @@
...
+    the y data, simply select the appropriate pipeline.
+
+    MultiplexTransformer has 2 parameters:
```
fkiraly, line 25: seems redundant with the `Parameters` section?
Though, might be useful.

What would be important is to describe *behaviour*. I.e., what will the transformer do in `fit`, `transform`, `inverse_transform`?

sktime/transformations/multiplexer.py
```diff
@@ -0,0 +1,253 @@
...
+
+    def _set_transformer(self):
+        self._check_selected_transformer()
+        # clone the selected transformer to self.transformer_
+        if self.selected_transformer is not None:
+            for name, transformer in self._get_estimator_tuples(self.transformers):
+                if self.selected_transformer == name:
+                    self.transformer_ = clone(transformer)
+        else:
+            # if None, simply clone the first forecaster to self.forecaster_
```
fkiraly, line 155: should be "transformer"

fkiraly (dismissed):
Agreed with @GuzalBulatova, excellent contribution!

Non-blocking comments:
* I noted some typos and improvement suggestions for docstrings and comments.
* some redundancy with `MultiplexForecaster` obviously, and we are aware there is a lot of open questions about how to resolve them, here: https://github.com/alan-turing-institute/sktime/issues/2603
we should come back to that later and think about refactor and improvement. Before approaching the "delegator" problem again, I suggest to think about a generic multiplexer, which seems easier to reach (but still difficult). Separate concern though, not for this PR.

sktime/transformations/multiplexer.py
```diff
@@ -0,0 +1,253 @@
+#!/usr/bin/env python3 -u
+# -*- coding: utf-8 -*-
+# copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+"""Implements forecaster for selecting among different model classes."""
```
fkiraly, line 4: transformer

fkiraly: also not "model classes" but "estimators", no?

sktime/transformations/multiplexer.py
```diff
@@ -0,0 +1,253 @@
...
+            Otherwise, if not passed in _fit, guaranteed to be passed in _predict
+        X : optional (default=None)
+            guaranteed to be of a type in self.get_tag("X_inner_mtype")
+            Exogeneous time series to fit to.
+
+        Returns
+        -------
+        self : reference to self
+        """
+        self._set_transformer()
```
fkiraly, line 187: `__init__` is actually run now at the start every time `fit` is run, so this could be *completely empty* imo

fkiraly: i.e., I think it should be ok to remove `_fit`

sktime/transformations/multiplexer.py
miraep8: Yes, good spots here, and elsewhere!

sktime/transformations/multiplexer.py
miraep8: Yes, I agree it would be better to clarify this.

sktime/transformations/multiplexer.py
miraep8: Ah yea, I did forget about that change.  Sounds good.

sktime/transformations/multiplexer.py
miraep8: Agreed, I got rid of it and put more clarity on what tasks are delegated and how.

sktime/transformations/multiplexer.py
```diff
@@ -0,0 +1,253 @@
...
+        # test with 2 simple detrend transformations with selected_transformer
+        params1 = {
+            "transformers": [
+                ("imputer_mean", Imputer(method="mean")),
+                ("imputer_near", Imputer(method="nearest")),
+            ],
+            "selected_transformer": "imputer_near",
+        }
+        # test no selected_transformer
+        params2 = {
```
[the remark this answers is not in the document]
miraep8: This is a good suggestion.  It is behaving strangely though.  Will push code with fix when I have it sorted out.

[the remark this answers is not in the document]
fkiraly: well, that's one of the signs of a good test - that it catches bugs 😄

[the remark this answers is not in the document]
fkiraly: I'll wait for the fix then

fkiraly:
@miraep8, do you also want to open an issue on generic grid search and/or generic multiplexer?

miraep8:
@fkiraly, sure I will open one either this evening or tomorrow.  Still working on the bug, hope to have that wrapped up early next week as well.

fkiraly:
my conjecture with the bug: this is a confusion between `transformers_` and `_transformers`.

`_transformers` is the "always name/estimator pair" version of `transformers`, not fitted.
`transformers_` is the fitted/fittable copy.

The `get_params` etc should refer to `_transformers`, not `transformers_`, I think?

miraep8:
That is a reasonable conjecture, unfortunately I don't *think* it fixes the problem. This might be due to some confusion on my part though. 

In my implementation of `MultiplexTransformer` I was only using `transformers_` not `_transformers`. I am open to adding `_transformers` as well, but I guess I am a bit confused why this would be needed for `MultiplexTransformer` and not `MultiplexForecaster` (ie, I went back and looked at that code and I *think* you only use `forecasters_` there, never `_forecasters`....?)

Personally, I think that the issue is coming up because we are using the `TransformerFitTransformPanelUnivariateWithClassY` on `MultiplexTransfomer` when it is wrapping `Imputer` which should only be applied to series data, not panel data.  I would have thought that copying the tags from `Imputer` as we do in init would filter out this scenario from being applied, but clearly that is not the case.  Been looking more about how the scenarios are selected and/or if there is another tag that should be applied to MultiplexTransformer to fix this!

fkiraly:
> In my implementation of `MultiplexTransformer` I was only using `transformers_` not `_transformers`. I am open to adding `_transformers` as well, but I guess I am a bit confused why this would be needed for `MultiplexTransformer` and not `MultiplexForecaster` (ie, I went back and looked at that code and I *think* you only use `forecasters_` there, never `_forecasters`....?)

Yes, you are right - only one of these is needed, since there is only one transformer that needs to be fitted.

Btw, if I were to implement it today, I would make `_transformers` a property, like `_steps` in `_Pipeline`. That would avoid the write to self which I think is unnecessary.

sktime/transformations/multiplexer.py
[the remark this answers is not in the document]
miraep8: Well Franz, I guess in the end we go with `_transformers` not `transformers_` after all.  Lmk if you see anything else that needs fixing!

fkiraly:
For the record: the bug was due to a clash of the reserved variable `transformers_`  from `BaseTransformer` with the (obvious named) `transformer_` that @miraep8 introduced.

That is a bit unfortunate, @miraep8 had no other way to know than reading the entirety of `BaseTransformer` which is nothing we can ask of an implementer.

I have tried to address the issue of "reserved variables are not known to the implementer" in PR https://github.com/alan-turing-institute/sktime/pull/2769

fkiraly:
Explanation why it happened only for the fixture "with y": the reason was not the `y`, but that `X` was a panel (not just a single series). Because of that, the transformer vectorization would kick in, and that in turn would use the reserved `transformers_` attribute and write to it.

I guess that highlights a good choice of scenarios, since they cover a bit, but also it highlights the difficulty to identify the condition in the scenario that causes the breakage, if there are multiple unique aspects to a scenario.

docs/source/api_reference/transformations.rst
```diff
@@ -407,3 +407,15 @@ MovingBlockBootstrapTransformer
     :template: class.rst
 
     MovingBlockBootstrapTransformer
+Transformer Selection
+---------------------
+MultiplexTransformer
+~~~~~~~~~~~~~~~~
```
fkiraly, line 413: number of tildes should equal number of characters above

sktime/transformations/multiplexer.py
```diff
@@ -0,0 +1,201 @@
...
+           in transformers.
+        selected_transformer represents the name of the transformer MultiplexTransformer
+           should behave as (ie delegate all relevant transformation functionality to)
+
+    Attributes
+    ----------
+    transformer_ : sktime transformer
+        clone of the transformer named by selected_transformer to which all the
+        transformation functionality is delegated to.
+    _transformers : copy of the list of transformers passed.  If transformers was
```
fkiraly, line 53: I think these are not copies but direct references

miraep8: Yes, I can see how the language would be confusing.

fkiraly:
> I thought we had a design policy about not changing user provided attributes, ie `transformers`. 

That's correct ... ish.
We follow the `scikit-learn` design, where nothing must change *init arguments* (more precisely, their copies written to self to attributes of the same name), *except* the `set_params` interface, which is the dedicated interface point for changing those in an existing estimator instance.

fkiraly (approved):
Looks good now

fkiraly:
@miraep8, tests were failing on main, not sure why - I think we should reopen and check

miraep8:
Ok!  Sounds reasonable to me.  Can you point me to where these tests are failing and I will take a look?

miraep8:
Ah, nevermind I see what you mean.

fkiraly:
I suspect it is an interaction with the changes to the metrics, where the default setting for MAPE has been changed from symmetric to asymmetric
the changes to metrics are coming from the deprecation/removal actions prior to 0.12.0 release, and I merged them at the same time

PR #2744, opened 2022-06-03
[ENH] Greedy Gaussian Segmentation

lmmentel:
This PR implements the Greedy Gasussian Segmentation Estimator, based on

    Hallac, D., Nystrup, P. & Boyd, S.
    Greedy Gaussian segmentation of multivariate time series.
    Adv Data Anal Classif 13, 727–751 (2019).
    https://doi.org/10.1007/s11634-018-0335-0

- source code adapted based on: https://github.com/cvxgrp/GGS
- paper available at: https://stanford.edu/~boyd/papers/pdf/ggs.pdf

Pseudo-code from the original publication for reference

image
 
FYI @Lovkush-A :point_up: 

## Notable design decisions

- the initialization happens through [attrs](https://www.attrs.org/en/stable/) with three separate groups of arguments defined explicitly:
    - `__init__` args that are used to instantiate the estimator: `k_max`, `lamb`, `max_shuffles`, `verbose`, `random_state`
    - private instance state attributes starting with `_`: `_intermediate_change_points` and `_intermediate_ll` 
    - attributes set by `fit`: `change_points_` 
- the compatibility layer with sklearn is coded in `GGSEstimator` by providing the common interface methods
- Thanks to `attrs` default `__repr__` shows not only the kwargs but also internal instance state based on the attributes that are not passed to `__init__`.

## Minimal example

```python
import numpy as np
from sktime.annotation.ggs import GreedyGaussianSegmentation
from sklearn.preprocessing import StandardScaler

x = np.concatenate(tuple(np.ones(5) * i**2 for i in range(4)))
X = x[:, np.newaxis]

ggs = GreedyGaussianSegmentation()
print(ggs)
cp = ggs.predict(X)
print("Predicted segmentation vector", cp)
print(ggs)
print("Predicted change points:", ggs._adaptee.change_points_)

pipe = Pipeline([('scale', StandardScaler()), ('segment', GreedyGaussianSegmentation())])
predicted = pipe.fit_predict(X)
print("Predicted from pipeline: ", predicted)
```

which gives:

```python
GGS(k_max=10, lamb=1.0, max_shuffles=250, verbose=False, random_state=None, change_points_=[], _intermediate_change_points=[], _intermediate_ll=[])
Predicted segmentation vector [0 0 0 0 0 1 1 1 1 1 2 2 2 2 2 3 3 3 3 3]
GGS(k_max=10, lamb=1.0, max_shuffles=250, verbose=False, random_state=None, change_points_=[0, 5, 10, 15, 21], _intermediate_change_points=[[0, 21], [0, 15, 21], [0, 10, 15, 21], [0, 5, 10, 15, 21]], _intermediate_ll=[-50.11068443455932, -2.8697641499393534, 39.44974322647064, 52.18875824868201])
Predicted change points: [0, 5, 10, 15, 21]
Predicted from pipeline:  [0 0 0 0 0 0 0 0 0 0 1 1 1 1 1 2 2 2 2 2]
```


lmmentel:
FYI @miraep8

sktime/annotation/ggs.py
```diff
@@ -0,0 +1,296 @@
...
+            data: time series data
+            change_points: change points indexes
+            newInd:
+
+        Returns
+        -------
+            change_points:
+        """
+        bp = change_points[:]
+        random.seed(0)
```
miraep8, line 182: Another interesting way to approach this can be to randomly generate the seed, but put it in the logger, so that if weird stuff comes up it is reproducible. (and can sometimes help find more edge case stuff than just using a fixed seed while in debug mode).

lmmentel: I pulled `random_state` into intialization

lmmentel:
This should be ready for review. Some notable design decisions:

- the initialization happens through [attrs](https://www.attrs.org/en/stable/) with three separate groups of arguments defined explicitly:
    - `__init__` args that are used to instantiate the estimator: `k_max`, `lamb`, `max_shuffles`, `verbose`, `random_state`
    - private instance state attributes starting with `_`: `_intermediate_change_points` and `_intermediate_ll` 
    - attributes set by `fit`: `change_points_` 
- the compatibility layer with sklearn is coded in `GGSEstimator` by providing the common interface methods
- Thanks to `attrs` default `__repr__` shows not only the kwargs but also internal instance state based on the attributes that are not passed to `__init__`.

Here is a minimal example:

```python
import numpy as np
from sktime.annotation.ggs import GGSEstimator
from sklearn.preprocessing import StandardScaler

x = np.concatenate(tuple(np.ones(5) * i**2 for i in range(4)))
X = x[:, np.newaxis]

ggs = GGSEstimator()
print(ggs)
cp = ggs.predict(X)
print("Predicted segmentation vector", cp)
print(ggs)
print("Predicted change points:", ggs._adaptee.change_points_)

pipe = Pipeline([('scale', StandardScaler()), ('segment', GGSEstimator())])
predicted = pipe.fit_predict(X)
print("Predicted from pipeline: ", predicted)
```

which gives:

```python
GGS(k_max=10, lamb=1.0, max_shuffles=250, verbose=False, random_state=None, change_points_=[], _intermediate_change_points=[], _intermediate_ll=[])
Predicted segmentation vector [0 0 0 0 0 1 1 1 1 1 2 2 2 2 2 3 3 3 3 3]
GGS(k_max=10, lamb=1.0, max_shuffles=250, verbose=False, random_state=None, change_points_=[0, 5, 10, 15, 21], _intermediate_change_points=[[0, 21], [0, 15, 21], [0, 10, 15, 21], [0, 5, 10, 15, 21]], _intermediate_ll=[-50.11068443455932, -2.8697641499393534, 39.44974322647064, 52.18875824868201])
Predicted change points: [0, 5, 10, 15, 21]
Predicted from pipeline:  [0 0 0 0 0 0 0 0 0 0 1 1 1 1 1 2 2 2 2 2]
```

Let me know what you think @miraep8 @KatieBuc @fkiraly

sktime/annotation/ggs.py
```diff
@@ -0,0 +1,320 @@
...
+        """Fit and predict."""
+        return self.fit(X, y).predict(X, y)
+
+    def get_params(self) -> Dict:
+        """Return initialization parameters."""
+        return asdict(self._adaptee, filter=lambda attr, value: attr.init is True)
+
+    def set_params(self, *args, **kwargs):
+        """Initialize the estimator from parameters."""
+        self._adaptee = self._adaptee_class(*args, **kwargs)
```
miraep8, line 315: I slightly worried about doing `set_params` this way for 2 reasons:
- by creating a new estimator rather than changing the previous one I wonder if you might create weird bugs for code that needs the same object instance in several places. (This might be an edge case, but an example of this came up recently in pywatts where a pipeline that uses one estimator as a transformer in one place, and then wishes to use the inverse version of that same object instance later). 
- if you do want to do it this way I think you need to copy over all the other params that have been set, whereas I think with this implementation you only instantiate with the 'new' parameters.

sktime/annotation/ggs.py
```diff
@@ -0,0 +1,320 @@
...
+
+import numpy as np
+import numpy.typing as npt
+from attrs import asdict, define, field
+from sklearn.utils.validation import check_random_state
+
+logger = logging.getLogger(__name__)
+
+
+@define
```
miraep8, line 28: I think using attrs is an *excellent* idea!!  I might try it out as well! Nice.

sktime/annotation/ggs.py
fkiraly: agreed, likely to cause problems due to teh above.

sktime/annotation/ggs.py
lmmentel: Thanks, for a moment I was thinking about alternatives but this worked quite well. The added benefit is that you need to declare **all** the attributes up front which help with state management of the class and forces a cleaner design. Looking forward to seeing your take on this :+1:

lmmentel:
FYI @KatieBuc

sktime/annotation/ggs.py
miraep8: I mean overall I think it is nice to be consistent with the way `sklearn` does it, where the object is modified rather than a new object is created. 

I agree though, concrete examples are always a good idea.  (And perhaps I should also say that both points were probably more targeted at a general `BaseObject`-like thing with `set_params`, rather than GGS in particular).  Along those lines, here are some general examples that might work for both points:

1) I think the first example is in line with what I mentioned before - ie we use our estimator as a transformer, call `set_params` on it and then want to use the inverse transform of the same object we did the original transformation with (this is important if there were fitted params that play a role in both `transform` and `inverse_transform`).  Could test this by having such an transformer and making sure you get the same data back if you `transform`, `set_params` and then `inverse_transform`.  
    - If this is something that is actually likely to happen? Perhaps not? Perhaps it is worth having a discussion as to what the expectations are here. But again, don't think defaulting to what `sklearn` does is such a bad idea. 
2) - If I first use set_params to set attribute a, then use it to set attribute `b` I would expect the final estimator that is returned to have both the `a` and the `b` estimator updated.  Another think to potentially check is that any attributes we set in init are correctly maintained in the object returned if they were not the attribute being changed via `set_params`.

Personally I wouldn't say that you need to implement either test here (unless you strongly want to!) as I think they won't really be as relevant to what is going on in GGS.  And I think your fix is now consistent.  But think it could be useful to do for the more general `BaseObject`/`BaseEstimator`. 

Think we can improve on the `set_params` etc in the more general `BaseObject` PR! :)  But figure its not worth going too much into the details on GGS.

miraep8: > I'm am not necessarily a fan partially modifying the internal state of an estimator - since it leaves it essentially in a corrupted state where __init__ parameters do not correspond to the attributes set at fit. I though that cutting a fresh estimator may prevent problems rather than create but probably a concrete case would help to reduce the uncertainty here.

I can see your point here though as well. I *think* the interpretation via `sklearn` is that `set_params` is meant to be the approved interface for changing the `__init__` parameters, so that this is ok! But it is an assumption.  Again I think it boils down to - does the user expect a clean state when calling `set_params`?

lmmentel: Thanks for providing a specific example but I am not sure how relevant it is since this estimator is not a transformer and therefore it has no `inverse_transform` method. Might be worth adding this to the base class discussion.

lmmentel:
This should now be ready to be merged @fkiraly 

FYI @Lovkush-A have a look if you are interested, as a follow-up on the base class conversation.

lmmentel:
@Lovkush-A @GuzalBulatova @fkiraly @aiwalter could you approve this PR?

Lovkush-A:
I am happy to approve with assumptoin that this is likely going to be changed after we decide on a uniform interface. Is that what you expecting too, @lmmentel?

General comments:
- documentation will need adjusting for the next version. E.g. just having single word 'fit' or 'predict' for a docstring is not helpful for a user!  also, does not use numpy format that is used elsewhere in sktime
- the `attr` package makes things look tidy! cant say i have understood exactly what it does (e.g. consequences of defining the intermediate variables in fit, rather than in init), but looks promising!
- i have not tried at all to understand at all the nitty gritty of GGS. i presume it works. would need some more tests to really check it does what we intend
- i have not thought about this much, but i do not see benefit of having two classes, with one wrapping the other. should be possible to add the methods in the Estimator class directly into the GGS class?

lmmentel:
Thanks @Lovkush-A 


> I am happy to approve with assumptoin that this is likely going to be changed after we decide on a uniform interface. Is that what you expecting too, @lmmentel?

Yup, I do not consider the annotation module "finalized" although the exposed interface if GGS should be compatible with the ecosystem as shown in the minimal example :point_up: 

> General comments:
> 
>     * documentation will need adjusting for the next version. E.g. just having single word 'fit' or 'predict' for a docstring is not helpful for a user!  also, does not use numpy format that is used elsewhere in sktime
> 
I totally agree, right now the documentation lags behind and the current module will not be exposed in the docs https://www.sktime.org/en/stable/api_reference/annotation.html

>     * the `attr` package makes things look tidy! cant say i have understood exactly what it does (e.g. consequences of defining the intermediate variables in fit, rather than in init), but looks promising!
> 
I'm happy to prep a short presentation for one of the devs meetings if more people are interested. 

>     * i have not tried at all to understand at all the nitty gritty of GGS. i presume it works. would need some more tests to really check it does what we intend
> 
Understood, there is a goal we set at some early stage to have a notebook at some point illustrating some of the segmentation methods, would that help in this case or are you thinking about a deeper dive into the theory behind GGS?

>     * i have not thought about this much, but i do not see benefit of having two classes, with one wrapping the other. should be possible to add the methods in the Estimator class directly into the GGS class?

Main reason is separation of concerns between the algorithm implementation (that can be reused in different contexts) and the adapter ensuring compatibility with sktime/sklearn.

Lovkush-A:
> I'm happy to prep a short presentation for one of the devs meetings if more people are interested.

this is a good idea!

> Understood, there is a goal we set at some early stage to have a notebook at some point illustrating some of the segmentation methods, would that help in this case or are you thinking about a deeper dive into the theory behind GGS?

Ideally, I only approve a PR if I am confident the code does what we intend it to do. In this case, I neither know what psuedo-code is for GGS (what the intention for the code is) nor what the code actually does (i have not read any of the code corresponding to the algorithm itself).  this is separate from the theory. notebook might help with this - depends what is in the notebook!

> Main reason is separation of concerns between the algorithm implementation (that can be reused in different contexts) and the adapter ensuring compatibility with sktime/sklearn.

i see. not obvious if such 'algorithm classes' should live in sktime or outside. cf hmmlearn.

Lovkush-A (approved):
happy with pr, given discussion above about expected further changes after designs finalised

fkiraly (changes requested):
Looks good!

I'd kindly request better docstrings before we merge this.
In this state, it can be understood only by the author.

We need docstrings at least for the class (which algorithm is this?) and for the primary interface methods - inputs, outputs, state change. Could you kindly add those?

fkiraly:
>In this case, I neither know what psuedo-code is for GGS (what the intention for the code is) nor what the code actually does

@Lovkush-A, neither do I, and for me that is actually a reason to request this via docstring as a blocking change request.

sktime/annotation/ggs.py
```diff
@@ -0,0 +1,412 @@
...
+class GGS:
+    """
+    Greedy Gaussian Segmentation.
+
+    Parameters
+    ----------
+        k_max: maximum number of change points to find
+        lamb: regularization parameter
```
fkiraly, line 36: why not call this `lambda`?

sktime/annotation/ggs.py
```diff
@@ -0,0 +1,412 @@
...
+logger = logging.getLogger(__name__)
+
+
+@define
+class GGS:
+    """
+    Greedy Gaussian Segmentation.
+
+    Parameters
+    ----------
```
fkiraly, line 34: please provide types and default values for all parameters

sktime/annotation/ggs.py
```diff
@@ -0,0 +1,412 @@
...
+
+    def initialize_intermediates(self):
+        """Fit."""
+        self._intermediate_change_points = []
+        self._intermediate_ll = []
+
+    def log_likelihood(self, cov: float, nrows: int, ncols: int) -> float:
+        """
+        Compute the GGS log likelihood.
+
```
fkiraly, line 60: please state *which* log-lokelihood is being computed here.

fkiraly (changes requested):
Looks ok (as experimental code)

blocking comments:
* we should follow PEP8 conventions on naming
* please ensure docstrings are complete: variable, type, default, description (numpydoc style)
* for functions corresponding to formulae or routines in the paper, please link those precisely
* `sklearn` compatible estimators should not have `kwargs` in `__init__`, but set all arguments verbatim to `self` attributes

Non-blocking comments:
* `add_new_change_point` seems inefficient, since you are recomputing means, covariances, chol decompositions, determinants etc. For all of these, efficient update rules exists, which are not much more complicated than the equations you are using.

sktime/annotation/ggs.py
lmmentel: lambda is a reserved keyword in python

sktime/annotation/ggs.py
lmmentel: not sure what you mean, have you checked the paper?

Lovkush-A (approved):
happy for this to be merged, with assumption that future improvements will be made and after design decisions are made.

For example, using PEP8 conventions, using numpy docstring format, adding more details to documentation (e.g. 'time series data' is ambiguous. is it multivariate? panel? is numpy array 1d or 2d? what do rows and columns represents?)

Am ok with risk that lukasz is currently only one who understands code. we have links to the paper provided so if lukasz does suddenly disappear, it is possible for it to be understood.

fkiraly (changes requested):
Looks good!

Can you kindly ensure documentation is complete:

* ensure the docstrings are in `numpydoc` format, see other docstrings in `sktime. E.g., first chracter should be 4 white spaces from line start.
* Please add a description of the algorithm and/or a reference to the paper in the class docstring.

lmmentel:
@fkiraly I updated the docstrings, hope this will be enough

fkiraly (changes requested):
Thanks, @lmmentel!

This addresses the first point, great!

For the second, I think you still should write in the various docstrings, at least in the class docstring, what objects/algorithms are represented. The parameters, variables, etc, appear without reference and context. Could you either describe the algorithm and objects, or link to the respective objects in the paper?

Non-blocking comment:
could you add the reference as a `numpydoc` reference?

sktime/annotation/ggs.py
```diff
@@ -0,0 +1,456 @@
...
+        """
+        return asdict(self._adaptee, filter=lambda attr, value: attr.init is True)
+
+    def set_params(self, **parameters):
+        """Set the parameters of this object.
+
+        Parameters
+        ----------
+        parameters : dict
+            Initialization parameters for th estimator.
```
KatieBuc, line 444:
```suggestion
            Initialization parameters for the estimator.
```

fkiraly (approved):
Looks good!

Happy to approve under the proviso that we collect doc debt in a separate issue and work on that separately. This is in an experimental module, so we don´t need to have everything at 100% right away...

Issue #2751, opened 2022-06-06
[BUG] after change of MAPE `symmetric` default to False, grid search algorithms fail `test_score`


fkiraly:
After changing the default of the `symmetric` parameter to `False`, the grid search estimators `ForecastingGridSearchCV` and `ForecastingRandomizedSearchCV` fail the `test_score` test.

All other estimators pass the test, indicating that there is something not quite right with the grid search algorithms.
This seems unlikely to be anything related to the unified interface, since all algorithms other than grid search pass the test.

I suspect it is some kind of side effect that is specific to the grid search algorithms.

Run with error log is here:
https://github.com/alan-turing-institute/sktime/runs/6747419876
although the error log is uninformative regarding the root cause, since it simply says that the output of `score` does not agree with the manually computed result.

FYI @aiwalter, I think you worked on the grid search algorithms most recently (author is @mloning).

fkiraly:
ok, I think I know what is happening.

The grid search algorithms have their own `score` function, which override the default of "calculate MAPE". I don't quite know why this is the case and the default is not used. My proposed fix would be to simply delete the custom `score` function since the default seems sensible anywhere, but I feel we first need to understand the rationale why it is there.

@aiwalter, could you kindly elucidate? That part of the code seems to be coming from @mloning who also might shed some light on it, but Ithink you've worked on it later, @aiwalter, and are still active.

PR #2752, opened 2022-06-06
[BUG] remove non-standard `score` function in `BaseGridSearch`

fkiraly:
This PR removes the non-standard `score` function in `BaseGridSearch` which overrides the default `score` from `BaseForecaster`, and seems to cause interface non-compliance with respect to the specification of `score`, as described in https://github.com/alan-turing-institute/sktime/issues/2751.

Before we merge, we should understand why this was there in the first place, as discussed in #2751.


aiwalter (dismissed):

aiwalter:
@fkiraly no I think this was "legacy" code

aiwalter:
wait I think now it can only score with default `mean_absolute_percentage_error`?

fkiraly:
I think that's what `score` is always supposed to do? For `sklearn` compatibility, it has to default to one particular metric always afaik, in the case of forecasters the MAPE

aiwalter:
ah right, there is 2 score functionalities

aiwalter:
I dont have a strong opinion on this PR then

fkiraly:
seems to introduce interface non-conformance, since `score` is always assumed to be MAPE elsewhere, so would be in favour because of making the interface conformant.

aiwalter (dismissed):
ok then I approve it :)

PR #2648, opened 2022-05-16
[DOC] Text changes example notebooks

lbventura:
#### What does this implement/fix? Explain your changes.

First contribution to get a feel of the pull request process. Applied minor text changes - upper-case in section headers and at the start of bullet points - to 01_forecasting.ipynb and 01a_forecasting_sklearn.ipynb and added an image displaying the difference between supervised regression and forecasting. No changes in text nor code content.

#### Does your contribution introduce a new dependency? If yes, which one?
No

##### For all contributions
- [ ] I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/main/.all-contributorsrc).
- [ ] Optionally, I've updated sktime's [CODEOWNERS](https://github.com/alan-turing-institute/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
- [ ] I've added unit tests and made sure they pass locally.
- [x] The PR title starts with either [ENH], [DOC] or [BUG] indicating whether the PR topic is related to enhancement, documentation or bug

##### For new estimators
- [ ] I've added the estimator to the online documentation.
- [ ] I've updated the existing example notebooks or provided a new one to showcase how my estimator works.

PR #2754, opened 2022-06-06
New Explanatory Fig for 01a_forecasting_sklearn.ipynb

lbventura:
#### Reference Issues/PRs

Split from #2648. Adds an explanatory figure (which requires further work, see discussion in  #2648) to 01a_forecasting_sklearn.ipynb

Adds an explanatory figure (which requires further work, see discussion in  #2648) to 01a_forecasting_sklearn.ipynb.
##### For all contributions

##### For new estimators


fkiraly (changes requested):
Thanks for this nice idea!

I thought for longer about the illustration, and I think it is missing a crucial motif: the separation between what is seen in `fit` and what is seen in `predict`.

In particular, there is potential for confusion by not displaying the training data (seen in `fit`) for regression, but showing it for forecasting.

Could you try to rework it so we have on display:

* the training data, fed to `fit`
* any data fed to `predict`
* some indication that a forecasting horizon` also needs to be passed in forecasting (this is different!)
* some way to link commonalities in regression and forecasting, e.g., colour coding of data seen in fit/predict

For supervised learning, my attempt is the "robot" picture in the no.1 tutorial - though that is perhaps not too good for comparison, and the more schematic approach that you have taken might be better.

PR #2756, opened 2022-06-06
[ENH] option to exclude tests/fixtures in `check_estimator`

fkiraly:
This PR adds arguments to `check_estimator` and the underlying `run_tests` utility that allows to exclude tests and fixture/test combinations by name.

In-principle, this would already be possible by passing a positive list and removing the tests to exclude, but currently there is no easy way to get the positive list. Hence, the argument is an easier (non-redundant) way to do this.


ltsaprounis (dismissed):
Looks good. I'd prefer if `default = none` in the docstrings is changed to `default = None`, but it's not a blocker.

Issue #2702, opened 2022-05-26
[BUG] `compute_sliding_residuals` in `NaiveVariance` has overlap between train and test


bethrice44:
**Describe the bug**

In [this line](https://github.com/alan-turing-institute/sktime/blob/df683766f4af7fd7fbe339fbf306d7a560827574/sktime/forecasting/naive.py#L635), a slice is taken to get train and test sets. The slicing behaviour appears to be inclusive for start and end for series so there is one element that is in both.

**To Reproduce**

```
from datetime import datetime
import pandas as pd

index = pd.date_range(datetime(2022, 5, 1, 9), datetime(2022, 5, 2, 8), freq="H")

y = pd.Series(index=index, data=[1]*len(index))

y_index = y.index

for id in y_index:
  y_train = y[:id]
  print(y_train)
  y_test = y[id:]
  print(y_test)
```

**Expected behavior**
Train and test sets should be exclusive. This can be achieved with:

```
y_train = y[:id][:-1]
```

**Versions**
<details>

<!--
Please run the following code snippet and paste the output here:
 
System:
    python: 3.8.12 (default, Oct 22 2021, 18:39:35)  [Clang 13.0.0 (clang-1300.0.29.3)]
executable: /Users/..../python
   machine: macOS-12.1-x86_64-i386-64bit

Python dependencies:
          pip: 22.0.3
   setuptools: 60.6.0
      sklearn: 1.0.2
       sktime: 0.11.3
  statsmodels: 0.13.2
        numpy: 1.21.5
        scipy: 1.7.3
       pandas: 1.4.1
   matplotlib: 3.5.1
       joblib: 1.1.0
        numba: 0.55.1
     pmdarima: None
      tsfresh: 0.19.0

fkiraly:
What a mess!
Thanks for pointing this out, @bethrice44!

It would appear that the slicing behaviour is inconsistent when you compare datetime indexing to `iloc` indexing!
Compare this to `y[:5]` vs `y[5:]`, in which the split is non-overlapping.

I think there is a deeper problem here, with `pandas`, where this should be raised if it hasn't already been.

On the `sktime` side, we probably need to deal with this by case distinction in the slicing logic.

fkiraly:
I confirm the issue on python 3.8 and windows.

bethrice44:
> What a mess! Thanks for pointing this out, @bethrice44!
> 
[...]

PR #2760, opened 2022-06-07
[BUG] Fixing overlap in `NaiveVariance` train/test set due to inclusive indexing for timestamp limits

bethrice44:
Issue #2702

Pandas treats slicing with integer vs timestamp index / period index differently (the first has an exclusive end and the second inclusive). `get_slice` ensures both are exclusive.


lmmentel (approved):
LGTM, thanks :raised_hands:

Issue #2757, opened 2022-06-06
[ENH] Add support to get_slice for multi-index and hierarchical data


bethrice44:
**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is.

In order to deal with [this issue ](https://github.com/alan-turing-institute/sktime/issues/2702), a function called `get_slice` has been created [here](https://github.com/alan-turing-institute/sktime/pull/2706) so slicing objects is always inclusive at the start and exclusive at the end. 

**Describe the solution you'd like**
As in `get_window` ([link](https://github.com/alan-turing-institute/sktime/blob/a761368f32b12913f2d7c22d87776e6112cc785d/sktime/datatypes/_utilities.py#L222)) we need to add some additional conditional logic to deal with multi-index and hierarchical data structures. 

We should also expand the tests in [datatypes/tests/test_utils.py](https://github.com/alan-turing-institute/sktime/blob/main/sktime/datatypes/tests/test_utils.py) to cover these cases.

fkiraly:
solved by @bethrice44 in #2761, forgot to close

PR #2761, opened 2022-06-07
[ENH] Add support to get_slice for multi-index and hierarchical data

bethrice44:
#### Reference Issues/PRs
Issue #2757 

Adds support for multi-index and hierarchical data types to `get_slice` (based on `get_window`)

No.

Are the assumptions for index types valid? I believe `get_window` is implicitly making the same assumptions.

If they are valid, can we amend the examples so series/ dfs have a timestamp index by default for testing?


sktime/datatypes/_utilities.py
```diff
@@ -331,33 +333,35 @@ def get_slice(obj, start=None, end=None):
...
 
-    if (
-        isinstance(start, pd.Timestamp)
-        or isinstance(end, pd.Timestamp)
-        or isinstance(start, pd.Period)
-        or isinstance(end, pd.Period)
-    ):
+    # numpy3D (Panel) or np.npdarray (Series)
+    # Assumes the index is integer so will be exclusive by default
+    if isinstance(obj, np.ndarray):
```
fkiraly, line 338: hm, I think this is not quite right.

`numpy3D` (for a panel of series) has index (instance, variable, time), so you need to subset the *third* index in that case.
`np.ndarray` has (time, variable) indexing.

I know, this is strangely inconsistent, but it is individually consistent with community preference and the `tslearn` package for time series classification.

fkiraly (changes requested):
I think this is good, but does not cover `numpy3D` correctly.

You need to cover the types in `GET_LATEST_WINDOW_SUPPORTED_MTYPES`, the others are dealt with by the conversion.

Out of these, I think `numpy3D` is being dealt with incorrectly, see above.

sktime/datatypes/_utilities.py
bethrice44: Looks good

fkiraly (approved):
All fine now & thanks for the review of my fix.

fkiraly:
What did you mean by this, btw?

> Are the assumptions for index types valid? I believe get_window is implicitly making the same assumptions.
> If they are valid, can we amend the examples so series/ dfs have a timestamp index by default for testing?

Which assumptions?

Issue #2499, opened 2022-04-18
[ENH] Add VARMAX statespace model in sktime


Ris-Bali:
**Is your feature request related to a problem? Please describe.**
I was just going through the code base and couldn't find VARMAX model in the forecasting submodule.

**Describe the solution you'd like**
Add or interface VARMAX model into sktime.

**Additional context**
This can be interfaced through statsmodels . 
https://www.statsmodels.org/dev/generated/statsmodels.tsa.statespace.varmax.VARMAX.html
Forecasting Template (for reference)
https://github.com/alan-turing-institute/sktime/blob/main/extension_templates/forecasting.py

juanitorduz:
I thinks the implementation could be similar to the [`UnobservedComponents`](https://github.com/alan-turing-institute/sktime/blob/main/sktime/forecasting/structural.py) model  implementation right? I could take this one maybe in a couple of weeks (I'm a bit busy at the moment) or support someone ini the process as it seems a good first issue :)

fkiraly:
yes, it probably will be, although the devil is always in the details... thanks!

juanitorduz:
> yes, it probably will be, although the devil is always in the details... thanks!

Totally agree! I will come back to this issue in a couple of weeks if no one has taken over :)

fkiraly:
was done by @KatieBuc in https://github.com/sktime/sktime/pull/2763 but forgot to close

PR #2763, opened 2022-06-08
[ENH] statsmodels VARMAX adapter

KatieBuc:
#### Reference Issues/PRs
[Add VARMAX statespace model in sktime #2499](https://github.com/alan-turing-institute/sktime/issues/2499)

Interfacing VARMAX from statsmodels, with corresponding unit tests

No

#### Any other comments?
Questions:
- Is there a list of estimators I need to add VARMAX to?
- In var.py, there's code in predict() that I don't understand. I initially copied across this (with out-sample predictions using forecast and in-sample predictions using the diff of residuals from the fitted object). I'm not sure why? Instead, for this, I used the predict() method from statsmodels (returns In-sample predictions / Out-of-sample forecasts). Checking that I'm not missing anything here!
https://www.statsmodels.org/devel/generated/statsmodels.tsa.statespace.varmax.VARMAXResults.predict.html
- get_prediction() in statsmodels also gives confidence intervals, do we want to include this?
- In the example (provided in the doc string), it fails the test for stationarity and hence we get: UserWarning: Non-stationary starting autoregressive parameters found. Using zeros as starting parameters.
should we normalise this dataset in the example(s)? (same data in the VAR doc string example)

##### For new estimators

@fkiraly @lmmentel @Lovkush-A


KatieBuc:
Some set-up issues to hopefully resolve with mentors next session:
- pre-commit extension not working with my VScode (need to check linting)
Extension installed.
Settings > Pre-commit-helper: Run On Save. Select, which hooks to run after saving a file.
Set to all hooks
terminal > "The terminal process failed to launch: Path to shell executable "C:\Users\n10907700\repos\sktime\pre-commit" does not exist."

- pytest discovery error in search bar of VScode (settings changed, restated. For now, tested via notebook)
image

@fkiraly @lmmentel @Lovkush-A

sktime/forecasting/varmax.py
```diff
@@ -0,0 +1,312 @@
...
+        params : dict or list of dict, default = {}
+            Parameters to create testing instances of the class
+            Each dict are parameters to construct an "interesting" test instance, i.e.,
+            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+            `create_test_instance` uses the first (or only) dictionary in `params`
+        """
+
+        # todo: set the testing parameters for the estimators
+        # Testing parameters can be dictionary or list of dictionaries.
+        # Testing parameter choice should cover internal cases well.
```
Lovkush-A, line 291: this needs to be done. simplest option is just to have an empty dictionary (so you this class will only be tested on the defaults). but would be good to include two or three options, depending on if there are any edge cases in VARMAX

sktime/forecasting/tests/test_varmax.py
```diff
@@ -0,0 +1,57 @@
...
+from statsmodels.tsa.api import VARMAX as _VARMAX
+from sktime.forecasting.base import ForecastingHorizon
+from sktime.forecasting.varmax import VARMAX
+
+index = pd.date_range(start="2020-01", end="2021-12", freq="M")
+df = pd.DataFrame(
+    np.random.randint(0, 100, size=(23, 3)),
+    columns=list("ABC"),
+    index=pd.PeriodIndex(index),
+)
```
Lovkush-A, line 19: if we are creating a random dataset in testing, would be good to ensure this is reproducible. E.g. setting random seed, or creating rng object in numpy. (I unfortunatley cannot remember best way to do this)

Lovkush-A (changes requested):
overall looks good! i like the unittests - seems like the best way to ensure the wrapper does what we claim it does.

Lovkush-A:
> Some set-up issues to hopefully resolve with mentors next session:

unfortunately i cannot help (immediately anyway) with this. i have not used precommit with vscode, and i do not remember what i did to get pytest to work in vscode... :/

can you get pre-commit to work in terminal? i.e. Can you manage to get pre-commit to do stuff when you run `git commit ...` in terminal?

sktime/forecasting/varmax.py
```diff
@@ -0,0 +1,312 @@
...
+            exog=X,
+            order = self.order,
+            trend = self.trend,
+            error_cov_type = self.error_cov_type,
+            measurement_error = self.measurement_error,
+            enforce_stationarity = self.enforce_stationarity,
+            enforce_invertibility = self.enforce_invertibility,
+            trend_offset = self.trend_offset,
+        )
+        self._fitted_forecaster = self._forecaster.fit(
```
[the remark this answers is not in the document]
fkiraly: this is sensible, because statsmodels has an interface which is `fit_results = model.fit(data)` and `prediction = model.predict(new_data)`, where `fit_results` is not an estimator type object or a reference to a state updated `model`, but a "fitting results" type objects.

This mirrors R philosophy with stateless container objects, rather than python/sklearn where fitting is state update.

Convention-wise, any fitted parameters should *end* in an underscore, which is not the case here, but let´s not change the adapter for now...

fkiraly:
you can fix linting issues manually as well, either by manually running the formatting packages, or by typing out the suggested changes (which may be the most convenient option if there are only few)

fkiraly:
@KatieBuc, your branch has a lot of linting (= code formatting) issues.

The best way to deal with these is to learn how to run the code formatters locally, and/or how to install precommit. See the developer docs here: https://www.sktime.org/en/stable/developer_guide/continuous_integration.html

Another way would be to go on "details" next to the failed checks on the GitHub web interface and see the logs.

lmmentel:
I'm not sure if VSCode plugin installs `pre-commit` itself but it might be worth checking if you have `pre-commit` installed and if your plugin configuration is pointing the right location for `pre-commit` i.e. your virtual env.

fkiraly:
> I'm not sure if VSCode plugin installs `pre-commit` itself 

I think it doesn´t, one has to do it manually

sktime/forecasting/varmax.py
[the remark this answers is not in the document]
KatieBuc: Yes, and I think you meant statsmodel has the interface `fit_results = model.fit(data)` and `prediction = fit_results.predict()`?

https://www.statsmodels.org/stable/generated/statsmodels.tsa.statespace.varmax.VARMAXResults.predict.html

Ah okay, noted regarding the convention.

sktime/forecasting/varmax.py
```diff
@@ -0,0 +1,312 @@
...
+            optim_complex_step = self.optim_complex_step,
+            optim_hessian = self.optim_hessian,
+            flags = self.flags,
+            low_memory = self.low_memory,
+        )
+        return self
+
+    def _predict(self, fh, X=None):
+        """
+        Wrap Statmodel's VARMAX forecast method.
```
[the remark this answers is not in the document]
KatieBuc: You're right, it's basically the same except in `_StatsModelsAdapter`

`y_pred = self._fitted_forecaster.predict(start=start, end=end, exog=X)`

and here

`self._fitted_forecaster.predict(start = start,
            end = end,
            dynamic = self.dynamic,
            information_set = self.information_set,
            signal_only = self.signal_only,
            exog = X,
        )`

we pass in 'dynamic', 'information_set' and 'signal_only'

sktime/forecasting/varmax.py
KatieBuc: quick question @Lovkush-A, do the unit tests iterate through all of these parameter sets?

sktime/forecasting/tests/test_varmax.py
KatieBuc: yeah, true, okay I set a seed

KatieBuc:
Okay, thanks everyone, pre-commit can run locally

sktime/forecasting/varmax.py
Lovkush-A: @KatieBuc The test suites (e.g. test_all_forecasters and test_all_estimators) will iterate through all of these parameter sets.

The unit tests you have written won't (unless you write code to loop through these parameters)

sktime/forecasting/varmax.py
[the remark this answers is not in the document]
Lovkush-A: Makes sense. Thanks for explaining! Might be worth adding a short comment above `def _predict` to explain why we couldn't just use the `_predict` from `_Statsmodelsadapter`

sktime/forecasting/varmax.py
```diff
@@ -0,0 +1,299 @@
...
+    Wrapper for statsmodels VARMAX model
+
+    Vector Autoregressive Moving Average with eXogenous regressors model (VARMAX)
+
+    Parameters
+    ----------
+    y : array_like
+        The observed time-series process :math:`y`, , shaped n_obs x k_endog.
+    X : array_like, optional
+        Array of exogenous regressors, shaped n_obs x k.
```
Lovkush-A, line 23: I dont think y and X need to be here. The stuff that goes here are things that user can provide during initialisation. Maybe this should be moved to _fit_forecaster?

```suggestion
```

sktime/forecasting/varmax.py
```diff
@@ -0,0 +1,299 @@
...
+        matrices or for Kalman filtering options. See `Representation`, and
+        `KalmanFilter` for more details.
+    Attributes
+    ----------
+    order : iterable
+        The (p,q) order of the model for the number of AR and MA parameters to
+        use.
+    trend : str{'n','c','t','ct'} or iterable
+        Parameter controlling the deterministic trend polynomial :math:`A(t)`.
+        Can be specified as a string where 'c' indicates a constant (i.e. a
```
Lovkush-A, line 64: This attributes section is a copy of (a subset of) the parameters section. I think it is safe to just delete this section, as it does not provide any additional information for a user.

Lovkush-A (changes requested):
I like the changes since my first review! Few more suggestions, all around docstrings/documentation.

sktime/forecasting/varmax.py
Lovkush-A: See this method, which creates list of instances: https://github.com/alan-turing-institute/sktime/blob/7be01f62e580db77da1420823291d5d03675b45d/sktime/base/_base.py#L351

KatieBuc: Hmm okay. I'll put this as a question for today's catch up. 
`params = [
            {"order": (0, 0)},
            {"order": (1, 0)},
            {"order": (0, 1)},
            {"order": (1, 1)},
        ]
'
for `order = (p,q)` but when `q=1`, then `X` must be populated (by default `X=None`)

KatieBuc: Wait, ignore me that was wrong^^

Lovkush-A:
@KatieBuc Looking good! Making progress on the automated tests - got passed the code quality. 💪 

It looks like the parameter setting `{"order": (0, 0)}` is invalid (at least one of p and q should be non-zero, is what error message from statsmodel says) and that is causing a bunch of the automated tests to fail.

sktime/forecasting/varmax.py
```diff
@@ -0,0 +1,378 @@
...
+        start, end = fh.to_absolute_int(self._y.index[0], self.cutoff)[[0, -1]]
+
+        return self._fitted_forecaster.predict(
+            start=start,
+            end=end,
+            dynamic=self.dynamic,
+            information_set=self.information_set,
+            signal_only=self.signal_only,
+            exog=X,
+        )
```
Lovkush-A, line 350: I think we need to select a subset of the predictions provided from `self._fitted_forecaster`. See final line in the _statsmodels adapter.

```
# statsmodels forecasts all periods from start to end of forecasting
        # horizon, but only return given time points in forecasting horizon
        return y_pred.loc[fh.to_absolute(self.cutoff).to_pandas()]
```

But maybe wait to see if automated tests raise any errors?

Lovkush-A:
> * Is there a list of estimators I need to add VARMAX to?

I am not sure. @fkiraly?

> * In var.py, there's code in predict() that I don't understand. I initially copied across this (with out-sample predictions using forecast and in-sample predictions using the diff of residuals from the fitted object). I'm not sure why? Instead, for this, I used the predict() method from statsmodels (returns In-sample predictions / Out-of-sample forecasts). Checking that I'm not missing anything here!

- For out-sample, seems like using predict() is the only option.
- For in-sample, seems like there are two options: one is to use residuals and the other to use predict().  I do not know VARMAX or statsmodels well enough to say what the difference is between these too (or even if there is a difference).
- Question just to check: do you know what 'in sample' and 'out of sample' means in this context?
- Question: do you know how the residuals and the predictions are calculated, and whether they produce different answers?


> * get_prediction() in statsmodels also gives confidence intervals, do we want to include this?

Yes, I think so. If you look in [forecasting extension template](https://github.com/alan-turing-institute/sktime/blob/main/extension_templates/forecasting.py), there are method(s) one should implement for confidence intervals. Sounds like `_predict_interval` is the method to implement.

> hence we get: UserWarning: Non-stationary starting autoregressive parameters found. Using zeros as starting parameters.
> should we normalise this dataset in the example(s)?

I do not know VARMAX, so cannot say whether normalising dataset would avoid the error. Based on the error message, an alternative solution could be to explicitly provide `start_params`.

fkiraly:
>>Is there a list of estimators I need to add VARMAX to?

> I am not sure. @fkiraly?

In general not, the registry retrieves estimators automatically from the code base.
Maybe in `build_tools.azure.check_soft_dependencies`, but we're not using Azure anymore, so that list does not need being added to, right? @lmmentel ?

sktime/forecasting/varmax.py
KatieBuc: okay, interesting... very interesting.

these tests are currently failing 
image

changing to 'return y_pred.loc[fh.to_absolute(self.cutoff).to_pandas()]' made a whole bunch of test_predict_... tests fail, but not the original tests. 

image


Okay still digging.

fkiraly:
(very superficial look at CI failure traceback: looks like a `pandas` index related syntax error in `_predict`)

KatieBuc:
ignore this last commit :)

fkiraly:
ok, ignoring this last commit and definitely not looking into it, or observing, for instance, that it says linting but does not look like linting but addition of actual conversion/indexing logic

KatieBuc:
> ok, ignoring this last commit and definitely not looking into it, or observing, for instance, that it says linting but does not look like linting but addition of actual conversion/indexing logic

I'm so glad you didn't notice that @fkiraly 

Okay - debugging - that was a wild ride. I got deep into the codebase but made it out alive. As it turns out, statsmodels have an indexing quirk.

This gives a quick example of it:
`
import pandas as pd

y_range = _make_series(
    n_columns=2, index_type='range', n_timepoints=5
)

print(y_range)

y_int = _make_series(
    n_columns=2, index_type='int', n_timepoints=5
)

print(y_int)

from statsmodels.tsa.api import VARMAX as _VARMAX

stats = _VARMAX(y_range)
stats_fit = stats.fit()
y_pred_stats = stats_fit.predict(start=0, end=10)
print(y_pred_stats)

stats = _VARMAX(y_int)
stats_fit = stats.fit()
y_pred_stats = stats_fit.predict(start=0, end=4)
print(y_pred_stats)

# ValueWarning: No supported index is available. Prediction results will be given with an integer index beginning at `start`.
stats = _VARMAX(y_int)
stats_fit = stats.fit()
y_pred_stats = stats_fit.predict(start=0, end=20)
print(y_pred_stats)

# ValueWarning: No supported index is available. Prediction results will be given with an integer index beginning at `start`.
stats = _VARMAX(y_int)
stats_fit = stats.fit()
y_pred_stats = stats_fit.predict(start=9, end=20)
print(y_pred_stats)
`
Basically, when `index_type='int'` and for the out-of-sample forecasting, statsmodels returns a new index reset at zero! Hence the new if statement.

I'm wondering if there's a better way (involving 'capturing' the ValueWarning somehow). This is passing all unit tests though.

sktime/forecasting/varmax.py
```diff
@@ -0,0 +1,392 @@
...
+            optim_score=self.optim_score,
+            optim_complex_step=self.optim_complex_step,
+            optim_hessian=self.optim_hessian,
+            flags=self.flags,
+            low_memory=self.low_memory,
+        )
+        return self
+
+    # defining _predict method to pass in dynamic, information_set and signal_only
+    # which are not passed into the _predict method in _StatsModelsAdapter
```
Lovkush-A, line 324: suggestion: adjust this as there are now *two* reasons why we need this. 
```suggestion
    # defining `_predict`, instead of inheriting from `_StatsModelsAdapter`,
    # for two reasons:
    # 1. to pass in `dynamic`, `information_set` and `signal_only`
    # 2. to deal with statsmodel integer indexing issue
```

sktime/forecasting/varmax.py
```diff
@@ -0,0 +1,392 @@
...
+            signal_only=self.signal_only,
+            exog=X,
+        )
+
+        # statsmodel returns zero-based index when index is of type int with the
+        # following warning
+        # ValueWarning: No supported index is available. Prediction results will be
+        # given with an integer index beginning at `start`...
+        # but only when out-of-sample forecasting, i.e. when forecasting horizon is
+        # greater than zero
```
Lovkush-A, line 359: I like this thorough explanation! This is very useful for any future people who look at this.

Lovkush-A (commented):
excellent detective work with the bug! it is stuff like this that makes sktime uniform interface so valuable.

KatieBuc:
@Lovkush-A :)

One thing it's still complaining about is 
image

But I've defini9tely added a blank line 
[weird docstring linting](https://github.com/alan-turing-institute/sktime/pull/2763/commits/c77fbe479d017446324ac1d22fad3ada638bf4d7)
[c77fbe4](https://github.com/alan-turing-institute/sktime/pull/2763/commits/c77fbe479d017446324ac1d22fad3ada638bf4d7)

image

I'm not sure what this error wants from me!

fkiraly:
as discussed in standup:
the message is not about a blank new line, but a blank character.

e.g., the first line should be `>>> from etc` instead of `>>>from etc`, and similar for the other lines.

fkiraly:
The failure is in the first line, not the second one as far as I can tell.
`FAILED forecasting/varmax.py::sktime.forecasting.varmax.VARMAX `

means that the example, from the docstring, when run, is failing.

The easiest way to check what is happening is to copy all the lines starting with `>>> ` into a vs code python interpreter window and run them (it will run it despite the `>>> `). If they break, or the printout does not exactly equal example printouts in the docstring example, the doctest will fail.

sktime/forecasting/varmax.py
```diff
@@ -0,0 +1,397 @@
...
+    >>> import numpy as np
+    >>> import warnings
+    >>> from sktime.forecasting.varmax import VARMAX
+    >>> from sktime.datasets import load_longley
+    >>> warnings.filterwarnings("ignore")
+    >>> _, y = load_longley()
+    >>> train, test = y.iloc[:-3,], y.iloc[-3:,]
+    >>> forecaster = VARMAX()
+    >>> forecaster.fit(train[['GNPDEFL', 'POP']], X=train[['ARMED']])
+    VARMAX()
```
lmmentel, line 203: This currently fails since it `fit` returns` a fitted forecaster with all the hyperparameters assigned and your line has an "empty" `VARMAX()`. Consider assigning the fitted estimator to a variable and testing against a print statement/output that is unlikely to change or depend on data. 

As a side note I would say that the scope of this doctest is already covered by explicit tests so consider keeping the example and skipping the doctest altogether.

sktime/forecasting/varmax.py
```diff
@@ -0,0 +1,397 @@
...
+    .. [1] Lütkepohl, Helmut. 2007.
+       New Introduction to Multiple Time Series Analysis.
+       Berlin: Springer.
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> import warnings
+    >>> from sktime.forecasting.varmax import VARMAX
+    >>> from sktime.datasets import load_longley
```
aiwalter, line 197: I recommend using the `load_macroeconomic` data, it just returns a multivariate `y`. You could add a second example with exog data if you want, then we could use the `load_longley`

aiwalter (commented):
looks good, could you kindly also add it to docs? You can check `VAR` e.g. of how to do that. Otherwise I can support :)

sktime/forecasting/varmax.py
```diff
@@ -0,0 +1,397 @@
...
+    >>> warnings.filterwarnings("ignore")
```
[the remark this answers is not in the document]
KatieBuc: it was a `UserWarning: Non-stationary starting autoregressive parameters found. Using zeros as starting parameters.
  warn('Non-stationary starting autoregressive parameters'`

could add `suppress_warnings`, with default `False`?

sktime/forecasting/varmax.py
KatieBuc: yeah, this dataset makes more sense - I restricted the columns so that we didn't get a (bad fit) convergence warning

KatieBuc:
> looks good, could you kindly also add it to docs? You can check `VAR` e.g. of how to do that. Otherwise I can support :)

ohhhhh @fkiraly just showed me, and that answers my very first question! Okay, pushed

fkiraly:
>In var.py, there's code in predict() that I don't understand. I initially copied across this (with out-sample predictions using forecast and in-sample predictions using the diff of residuals from the fitted object). I'm not sure why? Instead, for this, I used the predict() method from statsmodels (returns In-sample predictions / Out-of-sample forecasts). Checking that I'm not missing anything here!

It would look like `statsmodels` `varmax` can indeed deal with both in-sample and out-of-sample predictions. So, not doing anything fancy/custom but juts calling the `statsmodels` functionality that already exists seems the right thing to do.

> `get_prediction()` in `statsmodels` also gives confidence intervals, do we want to include this?

Are these confidence intervals (on the predictive mean), or predictive intervals? Which are not the same concept. We would wanti t only in the latter case, since that is what `predict_interval` is meant to return.

> In the example (provided in the doc string), it fails the test for stationarity and hence we get: UserWarning: Non-stationary starting autoregressive parameters found. Using zeros as starting parameters.
should we normalise this dataset in the example(s)? (same data in the VAR doc string example)

I suppose? Probably some non-stationarity test internal to `varmax` raises the alarm here.
Normalisation might not help, differencing typically does.

fkiraly (approved):
Looks good! Great!

We should just make a note, perhaps in the VARMAX issue, that predictive intervals are not implemented, and that it is not clear whether they can be obtained from `statsmodels` (as opposed to conf intervals).

PR #2773, opened 2022-06-09
[BUG] fixing multiple instances of constructor non-compliance with sklearn specifications

fkiraly:
This PR fixes the remaining estimators in #2733.

Most were failing due to `kwargs` instead of explicit arguments in the construtor.
Some were not writing the default value to self.

This has all been fixed, and #2732 should run through.

Additional fixes:
* docstring of `KNeighborsClassifier` corrected
* passing of the `sarimax_kwargs` in `ARIMA` was corrected


fkiraly:
Reviewed by @ltsaprounis as part of #2732 and approved.

PR #2779, opened 2022-06-10
[MNT] `check_estimators` to run without soft dependencies

fkiraly:
This PR ensures that `check_estimator` can always run without soft dependencies, and adds a test for it.

It addresses a recent issue fixed by https://github.com/alan-turing-institute/sktime/pull/2755 where `check_estimator` could only be run with one specific soft dependency installed, and the problem that `check_estimator` runs for probabilistic forecasters only when `tensorflow-probability` is installed (for `predict_proba`).

This PR adds:
* a test for checking that `check_estimators` runs if no soft dependencies are installed, to the "no soft dependencies" CI checks.
* a `"none"` severity level for the dependency checker functions, which results in a return of True/False based on whether packages are installed or not
* a conditional escape from testing `predict_proba` in forecasters if `tensorflow-probability` is not installed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,9 @@ test_softdeps: ## Run unit tests
 	mkdir -p ${TEST_DIR}
 	cp .coveragerc ${TEST_DIR}
 	cp setup.cfg ${TEST_DIR}
-	cd ${TEST_DIR}; python -m pytest -v -n auto --showlocals --durations=20 -k 'test_all_estimators' $(PYTESTOPTIONS) --pyargs sktime.registry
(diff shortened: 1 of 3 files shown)


ltsaprounis (approved):
Looks good to me! Maybe it's best to throw a warning if the algorithm has` predict_proba` but the dependencies are not installed though, not a blocker.

fkiraly:
>Maybe it's best to throw a warning if the algorithm has `predict_proba` but the dependencies are not installed though, not a blocker.

Yes, I thought about that too, I'll test that in a separate PR.

PR #2780, opened 2022-06-10
[ENH] Add MultiplexTransformer - reverted & restored to fix remaining CI/CD issue

fkiraly:
Restored #2738 to allow fixing remaining CI/CD failure.

The failure is fixed by aligning both sides of the comparison in the test `test_multiplex_transformer_in_grid`, by using a transformer without fit that is element-wise (`ExponentTransformer`). Previously, the sides would not necessarily agree, since in one side fit/transform is done on the entire batch, in the other sequentially/stream-wise.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,7 @@ sktime/datasets/local_data/
 # Installer logs
 pip-log.txt
 pip-delete-this-directory.txt
+pip-wheel-metadata/
 
 # Unit test / coverage reports
 htmlcov/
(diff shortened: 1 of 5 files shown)


fkiraly:
this is a fix for an incomplete merge of accepted PR, so merging

fkiraly:
Restored #2738 to allow fixing remaining CI/CD failure caused by merge incompatibility.

The failure is fixed by aligning both sides of the comparison in the test `test_multiplex_transformer_in_grid`, by using a transformer without fit that is element-wise (`ExponentTransformer`). Previously, the sides would not necessarily agree, since in one side fit/transform is done on the entire batch, in the other sequentially/stream-wise.

Issue #2526, opened 2022-04-21
[ENH] lagging transformer


fkiraly:
Can it be that we don´t have a simple lagging transform?
It seems to be the case (although the `WindowSummarizer` can be "abused" for this).

We should have a `Lag` transformer that can take an integer or timedelta in construction, and all it does is to lag the series by the given lag value, using `pandas.shift`.

Should be pretty straightforward with the transformer extension template - I would recommend implementing it only for `X_inner_mtype = "pd.DataFrame"`, and relying on automatic vectorization for panel and hierarchical data.

One word of caution is that `pandas.shift` is inconsistent between integer and timedelta use, so it should be made consistent and one or the other behaviour should be steered by something like a `keep_index` parameter (I would give it the default `False`).

cgoxo:
Can I work on this issue? I want to know more about this how to proceed.

fkiraly:
great, thanks, of course!

Here is our detail guide on implementing new estimators, in this case it would be a transformer.
https://www.sktime.org/en/stable/developer_guide/add_estimators.html

fkiraly:
@cgoxo, are you still working on this?

cgoxo:
Sorry Sir, I got busy related to my college activities. Now I am free and can work on this problem.

fkiraly:
FYI @cgoxo, I've started here now: https://github.com/alan-turing-institute/sktime/pull/2783
(in case you were still working on it)

cgoxo:
Really sorry sir. I got busy on my end sem exams.

fkiraly:
no worries, no harm done!

PR #2783, opened 2022-06-11
[ENH] Lagging transformer

fkiraly:
This PR introduces a transformer for lagging for use in pipelines and and potential component in more complex algorithms that rely on lagging, e.g., a reduction algorithm refactor. Implements #2526.

A particular challenge is to cover both range/integer and time-like indices, since `pandas.shift` handles those inconsistently.
In the transformer, this is handled by a manual step-out for integer indices.


fkiraly:
I originally thought it could be useful for pydata London, FYI @ltsaprounis. Although it has departed from being "very simple" due to the fiddlyness between int and time-like indices in `pandas`.

fkiraly:
FYI @danbartl, @ltsaprounis - this is an isolated lagging trafo, which we could use inside more complex compositors such as the window summarizer, or reduction.

Issue #2739, opened 2022-06-02
[BUG] Detrender in TransformedTargetForecaster does not work


shagn:
**Describe the bug**
When using a Detrender in a TransformedTargetForecaster and trying to create predictions, the inverse transformation seems to fail and therefore no predictions can be made. 
Detrender and Forecaster work fine when not concatenated in a TransformedTargetForecaster. 

**To Reproduce**
```python
from sktime.forecasting.trend import PolynomialTrendForecaster
from sktime.transformations.series.detrend import Detrender
from sktime.forecasting.compose import TransformedTargetForecaster
from sktime.forecasting.naive import NaiveForecaster
from sktime.datasets import load_airline

y = load_airline()

trans_fc = TransformedTargetForecaster(
    [
        ("detrender", Detrender(forecaster=PolynomialTrendForecaster(degree=1))),
        ("forecaster", NaiveForecaster(strategy="last")),
    ]
)
trans_fc.fit(y)
trans_fc.predict(1)
```
![image](https://user-images.githubusercontent.com/16029092/171620553-68a421e8-a9b5-4dac-a99c-ff7a406875de.png)

**Expected behavior**
The inverse transformation of a Detrender should work in a TransformedTargetForecaster. 

**Versions**
<details>
System:
    python: 3.8.13 (default, Mar 28 2022, 11:38:47)  [GCC 7.5.0]
executable: /anaconda/envs/hog-forecast/bin/python
   machine: Linux-5.4.0-1056-azure-x86_64-with-glibc2.17

Python dependencies:
          pip: 21.2.4
   setuptools: 61.2.0
      sklearn: 1.0.2
       sktime: 0.11.4
  statsmodels: 0.13.2
        numpy: 1.22.3
        scipy: 1.7.3
       pandas: 1.4.2
   matplotlib: 3.5.1
       joblib: 1.1.0
        numba: 0.53.1
     pmdarima: None
      tsfresh: None

</details>

fkiraly:
confirmed with current main, python 3.8 windows.

Problem looks like trouble with column renaming in the `Detrender`, combined with `sklearn`'s new input check on equal column names.

fkiraly:
also fails with other forecasters instead of `PolynomialTrendForecaster`, so not a problem of that

fkiraly:
I think I understand the root cause.

The problem materializes under the following conditions:
[...]

PR #2788, opened 2022-06-11
[BUG] attempted fix 1 to #2739 - keeping `pd.DataFrame` column name as `pd.Series` name after conversion

fkiraly:

The fix is ensuring that, when converting a univariate `pd.DataFrame` to `pd.Series`, that the output `pd.Series` retains the  `pd.DataFrame` column name as `pd.Series` name.

Unfortunately, this leads to breakage through `NaiveForecaster` which handles columns and names inconsistently.
That is addressed by a hard-fix of storing names/columns in `_fit` and overwriting whatever comes out of `_predict`.
At some point, we may want to untangle the nested, organically grown mess that is `NaiveForecaster` and clean it up, but until that day it at least has the right return format now.


fkiraly:
superseded by https://github.com/sktime/sktime/pull/4161

PR #2789, opened 2022-06-11
[BUG] attempted fix 2 to #2739 - removing `pd.Series` name as soon as seen

fkiraly:

The fix is ensuring that, whenever seeing a `pd.Series`, its `name` is removed.


fkiraly:
attempt, not preferred, closing

PR #2790, opened 2022-06-11
[ENH] forecasting pipeline test which triggers conversions and failure condition in #2739

fkiraly:
This adds a test that tests the specific pipeline from bug https://github.com/alan-turing-institute/sktime/issues/2739 .

The pipeline tests an integration condition which failed in #2739 due to subtle issues in handling `pd.Series.name`:

* a `pd.Series` `y` is converted to `pd.DataFrame`, then back to `pd.Series`, then back to `pd.DataFrame`, by `convert` or `convert_to`
* the initial series `y` has a `name` attribute set
* there can be "pairing" through `converter_store`, but it is between the first two, not the last two conversions
* there can be transformations in-between the conversions, as long as they do not change column names or type

Then, it happens that:
* after the first conversion, we have a `pd.DataFrame` with column name `y.name`
* after the last conversion, we have a `pd.DataFrame` with column name `0`

I.e., the name gets lost, and the column names are inconsistent between the two occurrences.

If the conversions occur in a context where naming consistency is assumed (like in the pipeline above), an exception is raised.

The context is created specifically by the pipeline in the test.


fkiraly:
Tests fail since the bug on `main` is correctly found to exist.

ltsaprounis (approved):
Approved via approving #2793

fkiraly:
will merge as part of #2793 once tests pass

PR #2791, opened 2022-06-11
[BUG] attempted fix 3 to #2739 - storing and restoring `pd.Series` name

fkiraly:

The fix is ensuring that, when converting `pd.Series` to `pd.DataFrame` and back, that the `name` is stored in the `converter_store` and restored at back-conversion.


fkiraly:
superseded by https://github.com/sktime/sktime/pull/4161

PR #2793, opened 2022-06-11
[BUG] add patch to ensure column/name preservation in `NaiveForecaster`

fkiraly:
This PR adds a patch to `NaiveForecaster` which currently seem to handle `pd.DataFrame` column names and `pd.Series` names inconsistently, potentially losing them in the process of applying some of the strategies.

This is addressed by a hard-fix of storing names/columns in `_fit` and overwriting whatever comes out of `_predict`.
At some point, we may want to untangle the nested, organically grown mess that is `NaiveForecaster` and clean it up, but until that day at least outputs have the right return format now.


ltsaprounis (approved):
LGTM + relevant test added 🦾

PR #2794, opened 2022-06-12
[DOC] team update: remove Tony from CC

fkiraly:
This PR updates the "team" page and the package maintainers:

* removes @TonyBagnall from community council, following his resignation last week
* removes @TonyBagnall from package maintainers and code owners, following his resignation
* no change of @TonyBagnall membership to core devs


fkiraly:
(Sajay was added in a different PR)

PR #2798, opened 2022-06-12
[DOC] cleaning up forecasting API reference

fkiraly:
This cleans up the forecasting API reference and adds some missing forecasters.


ltsaprounis (approved):
Nice! headers are more intuitive now.

PR #2800, opened 2022-06-12
[DOC] add missing Sajay references

fkiraly:
@sajaysurya was missing as a former core dev.

Also replaces instances where his name was in the author field and not the GitHub ID.

Also lints the files where he appears.


sajaysurya:
Thanks for adding me!

sajaysurya (approved):

PR #2810, opened 2022-06-13
[ENH] Adding | dunder for MultiplexTransformer

miraep8:
Related to design discussion on transformer dunder methods in #2459.  Also related to #2670 #2738

The `|` dunder is implemented the same way it is for `MultiplexForecaster` (ie makes use of `dunder_concat`). There are also tests for this that pass locally.

### DIFF
--- a/sktime/transformations/base.py
+++ b/sktime/transformations/base.py
@@ -40,7 +40,7 @@ State:
     fitted state inspection - check_is_fitted()
 """
 
-__author__ = ["mloning, fkiraly"]
+__author__ = ["mloning, fkiraly", "miraep8"]
 __all__ = [
     "BaseTransformer",
     "_SeriesToPrimitivesTransformer",
@@ -195,6 +195,29 @@ class BaseTransformer(BaseEstimator):
         if isinstance(other, BaseTransformer) or is_sklearn_transformer(other):
             self_as_pipeline = TransformerPipeline(steps=[self])
             return other * self_as_pipeline
+        else:
+            return NotImplemented
+
+    def __or__(self, other):
+        """Magic | method, return MultiplexTranformer.
+
+        Implemented for `other` being either a MultiplexTransformer or a transformer.
+
+        Parameters
+        ----------
+        other: `sktime` transformer or sktime MultiplexTransformer
+
+        Returns
+        -------
+        MultiplexTransformer object
+        """
+        from sktime.transformations.compose import MultiplexTransformer
+
+        if isinstance(other, MultiplexTransformer) or isinstance(
+            other, BaseTransformer
+        ):
+            multiplex_self = MultiplexTransformer([self])
+            return multiplex_self | other
         else:
             return NotImplemented
 
--- a/sktime/transformations/base.py
+++ b/sktime/transformations/base.py
@@ -40,7 +40,7 @@ State:
     fitted state inspection - check_is_fitted()
 """
 
-__author__ = ["mloning, fkiraly", "miraep8"]
+__author__ = ["mloning", "fkiraly", "miraep8"]
 __all__ = [
     "BaseTransformer",
     "_SeriesToPrimitivesTransformer",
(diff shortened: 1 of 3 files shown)


miraep8:
Also, one thought I have had with this is, where would be the best way to put an example of the `|` dunder of both `MultiplexTransformer` and `MultiplexForecaster`?  Is there is a good way to put more than one example in the docstring?

sktime/transformations/base.py
```diff
@@ -40,7 +40,7 @@ class name: BaseTransformer
     fitted state inspection - check_is_fitted()
 """
 
-__author__ = ["mloning, fkiraly"]
+__author__ = ["mloning, fkiraly", "miraep8"]
```
fkiraly, line 43: looks like missing double quotes here, could you kindly add them too?

sktime/transformations/base.py
```diff
@@ -198,6 +198,29 @@ def __rmul__(self, other):
...
+        ----------
+        other: `sktime` transformer or sktime MultiplexTransformer
+
+        Returns
+        -------
+        MultiplexTransformer object
+        """
+        from sktime.transformations.compose import MultiplexTransformer
+
+        if isinstance(other, MultiplexTransformer) or isinstance(
```
fkiraly, line 216: this expression is the same, logically, as `isinstance(other, BaseTransformer)`, because all `MultiplexTransformer` instances are also `BaseTransformer` instances. `isinstance` checks for subclass instance, not "just" direct instance.

fkiraly (changes requested):
Great!

I left some non-blocking comments.

Blocking comments, for discussion or implementation (need not implement if you think it´s a bad idea or not relevant).
* I think it would be "nice" if you could multiplex with an `sklearn` transformer, and it would be silently converted into an `sktime` transformer. That behaviour is is currently implemented in `TransformerPipeline.__mul__`, line 216.
* if we do this, might be worth considering factoring out an `adapt_sklearn` function or similar thing.
* In the past, there has been odd behaviour with pipelines, since some pipelines for forecasters were also transformers. See line 212 in `TransformerPipeline`. That problem is perhaps no longer present after all the pipeline refactors, but worth checking.

sktime/transformations/base.py
miraep8: that makes sense

fkiraly (changes requested):
Nice, thanks!

Small blocking details with `check_dunder_compose`:
* you´ve also input a classifier, which I think doesn´t make sense for anything except the `TransformerPipeline` ... I think? Am I thinking right or not?
* since you use the same function in right and left multiplication, now on occasion the order is wrong

Also, how about a function `coerce_to_sktime` or similar? As an alternative.

miraep8:
I think you are correct.  Too fast and furious with my correction. Will take a more in depth look this afternoon and fix it. Thanks for the in depth check! And sure - can break it out as `coerce_to_sktime_transformer` instead.  Might move it to `utils` as well rather than living in this file.

fkiraly:
> Might move it to utils as well rather than living in this file.

I'd do that only if it is required by more than one module, and even then I'd be hesitant since I think it's good to keep the composition logic localized.

fkiraly:
(if you think it's ready to re-review, click the "ready for review" symbol next to my name in the reviewers list)

miraep8:
Will do, not there yet, have yet to make the changes, but will do so when it is ready!

fkiraly (approved):
Looks good to me!

As a non-blocking side note, my "boilerplate" senses are tingling, seeing `coerce_to_sktime` so closely to `_dunder_concat` multiple times. Were these functions without their context, I'd definitely move the former into the latter.

However, `_dunder_concat`, and the `_HeterogeneousMetaEstimator` it lives in, are very abstract and do not know about concrete wrapper/adapter classes in `sktime`, as `_coerce_to_sktime` does. The move would therefore create "lasagna", i.e., mix architectural layers ("base object functionality" with "sktime adapter functionality").

In case you have a good idea, we can look at it in a separate PR.

miraep8:
Just making a small (and hopefully inconsequential change) I might make before this PR is merged is to hint that the conversion function is private (ie `coerce_to_sktime` -> `_coerce_to_sktime`) (meant to do that originally, but your comment made me realize that I did not). For future could consider breaking adding some of these things on as decorators, seems like `coerce_to_sktime` could be a good candidate for that, though I agree perhaps in a future PR.

Issue #2767, opened 2022-06-08
[ENH] Transformed Target Forecaster (prediction quantiles & intervals) implementation request. 


HishamSalem:
The Transformed Target forecaster is not implemented in sktime, is it possible if it gets implemented? Is there an alternative to manually implement prediction intervals ensemble models?

Have the basic forecasting coverage with the option to utilize bootstrap for confidence intervals  
https://otexts.com/fpp2/prediction-intervals.html

If there is a way to implement this for ensemble XGBoost models that would be excellent :)

**Additional context**

fkiraly:
@HishamSalem, the `TransformedTargetForecaster` does implement prediction intervals currently! For that, the forecaster in it needs to support prediction intervals.

If you want to *add* prediction intervals to a forecaster that doesn´t natively have them, you can use one of the wrappers that add them to any forecaster, like `NaiveVariance` or `ConvormalIntervals`.

HishamSalem:
@fkiraly  Thank you for the advice, I tried this and it still gives a not implemented error. 

Code:

from sktime.datasets import load_airline
from sktime.forecasting.naive import NaiveVariance
from sktime.forecasting.compose import TransformedTargetForecaster,make_reduction #Make reduction
from xgboost import XGBRegressor

y = load_airline()
forecaster = TransformedTargetForecaster(
    [        
      (
          "forecast",
          make_reduction(XGBRegressor()   
              ,
              scitype="tabular-regressor",
              window_length=16,
              strategy="recursive",),
      ),
  ])
#forecast
variance_forecaster = NaiveVariance(forecaster)
variance_forecaster.fit(y)
var_pred = variance_forecaster.predict_interval(fh=[1,2,3],coverage=0.9)



NotImplementedError: Generating in-sample predictions is not yet implemented for RecursiveTabularRegressionForecaster.

HishamSalem:
I do not think the ConformalIntervals is in my modules yet. Do I need to update sktime? 

I did try to take the module directly into my pc, can you share the installation wheel so that I can import ConformalIntervals?

fkiraly:
bootstrap bagging is in the works here, btw:
https://github.com/alan-turing-institute/sktime/pull/2248

HishamSalem:
I updated the package as instructed and I do get an indexing error specifically for sliding window models. The error mainly has to do with the make reduction package. 

from sktime.datasets import load_airline
from sktime.forecasting.conformal import ConformalIntervals
[...]

PR #2815, opened 2022-06-14
[BUG] Fixing type conversion for proba interval wrappers - option 2, convert for computing residuals logic

bethrice44:
#### Reference Issues/PRs
Alternative fix for #2814, implements "alternative" option discussed in #2814 = option 2 in #2767

See linked PR for details

No


fkiraly (approved):
This seems to work now, so I'll merge it.
I'll also think about tests how we catch these composite situations better.

PR #2818, opened 2022-06-15
[DOC] cleaning up transformer API reference

fkiraly:
This PR cleans up the transformer API reference:

* a lot of transformers were simply missing. Included a number of transformers contributed by non-core-devs, and hierarchical transformers (aggregation, reconciliation). They were added.
* the high-level taxonomy of transformers was re-ordered, e.g., to be clear about what the return is (primitives, series, panel, etc). This is explained in a table at the start.
* new mid-level taxonomy of transformers. Parts correspond to the TSC estimator taxonomy, parts based on methodological details.

Further, it seems that the transformers have been placed in the `series` and `panel` folders more or less at random, historically. The taxonomy does not reflect well the folder structure, so we may like to move around things in the folders at some point in the future.

### DIFF
--- a/docs/source/api_reference/transformations.rst
+++ b/docs/source/api_reference/transformations.rst
@@ -122,19 +122,6 @@ These transformers extract simple summary features.
     RandomIntervalFeatureExtractor
     FittedParamExtractor
 
-Dictionary-based features
-~~~~~~~~~~~~~~~~~~~~~~~~~
-
-.. currentmodule:: sktime.transformations.panel.dictionary_based
-
-.. autosummary::
-    :toctree: auto_generated/
-    :template: class.rst
-
-    PAA
-    SFA
-    SAX
-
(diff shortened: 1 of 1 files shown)


docs/source/api_reference/transformations.rst
```diff
@@ -367,26 +442,50 @@ FeatureSelection
...
+.. autosummary::
+    :toctree: auto_generated/
+    :template: class.rst
+
+    PCATransformer
+
+Series-to-Panel transformers
+----------------------------
+
+Thee transformers create a panel from a single series.
```
ltsaprounis, line 488: Typo! These

ltsaprounis (dismissed):
Nice!

docs/source/api_reference/transformations.rst
fkiraly: fixed

Issue #2821, opened 2022-06-16
[ENH] Using OpenML as a benchmarking framework - Implementation in sktime


TNTran92:
**Is your feature request related to a problem? Please describe.**
Refer to #2777 

OpenML is a benchmarking framework that allows ML experiments to be reproduced easily and transparently. Each dataset and task has a unique ID that is share-able, allowing other users to repeat the same task on the same dataset. At the moment, it has limited support for time series dataset. It will be beneficial to investigate its potential to host sktime's dataset and our benchmark tasks.

**Describe the solution you'd like**
A demonstration of implementation of sktime to OpenML. Or demonstration why it is not possible.

**Describe alternatives you've considered**
Kotsu #2804 and Gluon-ts

**Additional context**

TNTran92:
OpenML not compatible with sktime due to having a requirement to have target column in input DataFrame. Closing this issue.

Issue #2777, opened 2022-06-10
[ENH] Benchmarking Discussion


TNTran92:
During the benchmarking meeting today, a number of topics was discussed regarding benchmarking. This issue summarizes/expands these discussions. This documents will be updated after meetings:
Three packages were among the discussions on 06/10/2022:
* Benchmarking-wise, the `Evaluator()` class of `gluonts` requires about 3 adapters to make it work with `sktime`. I do recall exactly but one of the adapter would be for sktime's `metrics`. Need to consider additional options. There are 2 candidates: Kotsu and OpenML
* Kotsu is a high level and light-weight package that performs benchmarking that has no dependency on specific packages. At a high level, it allows users to register a dataset with a predicting task and then apply estimators to that registration. Will need more investigation about application to sktime.
* OpenML performs a similar task to kotsu. An additional plus is that such registration has a share-able ID that can be shared on the cloud, allowing other users to repeat the same task. At the moment, it has limited support for time series dataset. Will also need to investigate its application in `sktime`.

**To do tasks (High priority):**
- [x] Investigate the implementation of `kotsu` into `sktime` - refer to Issue #2804
- [x] Investigate implementation of `OpenML` into sktime - refer to Issue #2821
- [x] Investigate implementaition of  `Mlflow` into sktime - refer to Issue #2916 
- [x] Compare available benchmarking approaches
- [ ] Converter for Multiindex DF into ListDataset -- refer to Issue #2860 
- [ ] Converter for ListDataset into Multiindex DF

Update: 6/20/2022
- [x] kotsu's experiment: Test all estimators for point forecasting
- [x]  kotsu's experiment: Test all estimators for interval forecasting
- [ ] kotsu's experiment: Investigate Issue #2742
- [ ] Prep M4 dataset for sktime
- [ ] Mock a benchmarking study with vanilla kotsu
- [ ] kotsu's experiment: Draft interface design for kotsu
- [ ] kotsu's experiment: Try with annotations

**Good to have (Lower priority):**
N/A

**Additional thoughts (Not action item)**
If implemented,  the `gluonts`'s adapter would look similar to statsmodel's adapter.

ltsaprounis:
@DBCerigo - FYI for kotsu

fkiraly:
[...]

PR #2822, opened 2022-06-16
OpenML application to sktime

TNTran92:
#### Reference Issues/PRs

Solution to Issue #2821. See also Issue #2777

Demonstrate how OpenML's data format is incompatible with sktime.

TLDR:
- In an OpenML dataset, each variable is in a column, each time point is a row, just like in a DataFrame.
- Correction from daily stand-up: Not required to have `target` column but since OpenML does not support Pd.Series inside DataFrame => Need to turn time series into rolling-window format.
- Because dataset has to be rolling window, adapters need to be made for .ts format, estimator, CV splitter and `metrics`
- Still not sure how to make a 1:1 conversion of MultiIndex DataFrame to OpenML Dataset 

Yes, OpenML

Performance, design and viability of implementing OpenML in sktime

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/benchmarking/OpenML_notebook.ipynb
+++ b/sktime/benchmarking/OpenML_notebook.ipynb
@@ -1,0 +1,590 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# OpenML Dataset Creation and Properties"
+   ]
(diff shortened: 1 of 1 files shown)


fkiraly:
past experimental PR - moving to `kotsu` based implementation

Issue #2813, opened 2022-06-14
[ENH] Intrinsic Dimensionality + Hidalgo implementation


KatieBuc:
Tags: multivariate, batch, unsupervised

Implement segmentation based on the local intrinsic dimension, the [Hidalgo](https://arxiv.org/abs/1902.10459) algorithm, as an estimator in the annotation module.

Tasks:
- [x] draft pull request with scaffold of implementation (not inheriting from BaseSeriesAnnotator)
- [x] figure out Gibbs sampling implementation (see below)
- [ ] design annotator base class collaboratively
- [ ] refactor code to adhere to the agreed framework

Hidalgo relies on Gibbs sampling (MCMC) with a custom likelihood function in the model. Specifically, it is assumed that the neighborhood of a point is more likely to contain points from the same manifold than from different manifolds. This requirement is enforced with an additional term in the likelihood that penalizes local inhomogeneity [points belonging to different manifolds].

Options for implementing the Gibbs sampler include:
1. including the C code from [here](https://github.com/micheleallegra/Hidalgo/blob/master/python/gibbs.c)
2. re-writing the model and Gibbs sampler in python
3. interfacing pymc, seeing if our model can be fed into an 'out-of-the-box' Gibbs method

PR #2828, opened 2022-06-17
[ENH] Data segmentation based on Hidalgo (Heterogeneous Intrinsic Dimension Algorithm; Allegra et al, 2019)

KatieBuc:
#### Reference Issues/PRs
Drafting a template for the Hidalgo algorithm, an estimator in the annotation module (issue #2813). Ideas will be discussed as a community about an annotation base class object, what it could look like.

Introduces a new concept/algorithm for segmentation given a (multivariate) timeseries.

intrinsdim.py -> output is scalar value representing intrinsic dimensionality
hidalgo.py -> extends the above model, assuming the data exists on a mixture of manifolds (each with corresponding dimension), output is a vector of size `N x 1` corresponding to each the multivariate input of size `N x c`. For example, if we assume there are `K=3` manifolds then an output vector for `N=10` would look like `(d_1, d_3, d_2, d_2, d_2, d_3, d_1, d_1, d_1, d_1)`, where `d_k` is the dimensionality for manifold `K`.

TDB.
Hidalgo relies on Gibbs sampling (MCMC) with a custom likelihood function in the model. Specifically, it is assumed that the neighborhood of a point is more likely to contain points from the same manifold than from different manifolds. This requirement is enforced with an additional term in the likelihood that penalizes local inhomogeneity [points belonging to different manifolds]. 

Options for implementing the Gibbs sampler include:
- including the C code from https://github.com/micheleallegra/Hidalgo/blob/master/python/gibbs.c
- re-writing the model and Gibbs sampler in python
- interfacing pymc, seeing if our model can be fed into an 'out-of-the-box' Gibbs method

### Notes
References: https://arxiv.org/abs/1902.10459


fkiraly:
Quick question for the PR title: tried to clarify this since "Hidalgo" does not tell much about the contents.

I checked the paper and there is no Hidalgo amongst the authors, and the name does not appear in the abstract.
Is this the correct linked paper? I have a slow internet connection so can't currently look into the paper.

fkiraly:
Re options:

I think we do *not* want to pull any C code into `sktime`. We had a phase (until 0.8.0 or so) where `sktime` had C code, and it turned the installation into an OS-specific nightmare that made it difficult for a lot of users. Any C code should remain localized to soft dependencies.

Interfacing and using `pymc` is probably the cleanest and lowest implementation or maintenance effort of the options? Another option might be leveraging `tensorflow-probability` for MCMC (Gibbs is not directly available from `tfp` afaik)

I'd be hesitant starting to implement Gibbs samplers given that well-tested and robust implementations are available.

KatieBuc:
> Quick question for the PR title: tried to clarify this since "Hidalgo" does not tell much about the contents.
> 
> I checked the paper and there is no Hidalgo amongst the authors, and the name does not appear in the abstract. Is this the correct linked paper? I have a slow internet connection so can't currently look into the paper.

Yeah - it's a bit hidden but Hidalgo stands for Heterogeneous Intrinsic Dimension Algorithm (a mixed model where data exists on the union of K manifolds with varying dimensions, an extension of previous models assuming one manifold/dimeson for the data set).

KatieBuc:
> Re options:
> 
> I think we do _not_ want to pull any C code into `sktime`. We had a phase (until 0.8.0 or so) where `sktime` had C code, and it turned the installation into an OS-specific nightmare that made it difficult for a lot of users. Any C code should remain localized to soft dependencies.
> 
> Interfacing and using `pymc` is probably the cleanest and lowest implementation or maintenance effort of the options? Another option might be leveraging `tensorflow-probability` for MCMC (Gibbs is not directly available from `tfp` afaik)
> 
> I'd be hesitant about starting to implement Gibbs samplers, given that well-tested and robust implementations are available.

Yes I agree, I can by trying to implement the required model using/interfacing pymc.

Question for our next mentor meeting (on the hackmd): what's the best environment to run C code in? to be able to break it down into the components and see the inputs/outputs we get.

fkiraly:
>what's the best environment to run C code in? to be able to break it down into the components and see the inputs/outputs we get.

Are you planning to write C code? If you are using `pymc`, then you might not need to?

PR #2829, opened 2022-06-17
[ENH] VECM model as forecaster 

AurumnPegasus:
#### Reference Issues/PRs

Completes Issue #929 

Added a wrapper around statsmodel's VECM class to make it easy for users to use VECM forecaster

No

I have written the `test_vecm.py` referring the `test_var.py` without really implementing the difference between the two (for example, how to use `exog_coint`). Any update wrt how to make test more robust would be appreciated.

##### For all contributions

##### For new estimators


fkiraly (commented):
`VECMResults.predict` is able to produce "confidence intervals" - we should check whether these are confidence intervals on the mean prediction, or predictive intervals that, for instance, `_predict_interval` could return.

AurumnPegasus:
Ah the test which is failing is because I was implementing `_predict_interval` but forgot to comment it out for this PR. I will just push the new changes and it should work.

fkiraly:
Great!

Are you not implementing `predict_interval`, though? Or do you mean, you'll do in a separate PR incrementally, and we move this in first?

AurumnPegasus:
I had a couple of doubts related to `predict_interval`.
1. Is there a difference between `confidence interval` and `prediction interval` ? From what I see in the documentation of `statsmodel/vecm` there is functionality to only get confidence interval 

> alpha : float, 0 < `alpha` < 1 or None
            If None, compute point forecast only.
            If float, compute confidence intervals too. In this case the
            argument stands for the confidence level.
>

^ This is according to statsmodels/vecm. Is this the result we want in `predict_interval` or is it something else? I just googled around [here](https://towardsdatascience.com/confidence-intervals-vs-prediction-intervals-7b296ae58745) and there seems to be a conceptual difference between the two. 

2. If both are same or `statsmodels/vecm` has implemented `predict interval` correctly (maybe documentation is wrong), then my question is this: According to `statsmodels/vecm` the predict interval is returning `lower` and `upper` forecast only for `outsample` prediction (I took `_pmdarima` as a reference and it has functionality to take `fh` as both `in sample` and `out sample`). So what should be done here? Is `predict interval` only valid for out of sample `fh` in `vecm` ? Or does `statsmodel` have an incomplete version of `vecm` or have I understood something wrong?

fkiraly:
For completeness, I am pasting here what I said on the slack contributors channel:

"Coverage of a predictive interval and confidence interval are not interchangeable concepts! A confidence interval is connected to a statement of confidence, which is a frequentist statement w.r.t. the full data sampling distribution (subject to which the confidence is stated). A predictive interval (and its coverage) relate to frequentist statements w.r.t. the conditional sampling distribution of the quantity predicted, conditional on the quantity observed.

So these are in general not the same! And confidence intervals refer to an estimate, whereas predictive interval refer to a specific prediction. In some cases they may coincide, if the estimate is that of a predictive interval, but they may also be different in the context of a prediction, e.g., if the confidence interval is that of a predictive mean.

Either way, I'm 70% certain that the statsmodels documentation is wrong about the `alpha` being confidence interval levels (for the point forecast), and they are, in fact, the interval violation probability of a symmetric predictive interval, rather than a confidence interval on the point forecast. I.e., what @Lovkush-A has said is correct, `sktime` coverage being the same as `statsmodels` `1-alpha` for VECM. But that being due to implementation, and the `statsmodels` documentation being erroneous.

fkiraly:
What you would still need to do is to check, is the implementation in `statsmodels` producing confidence or predictive intervals? The docstring may be wrong, so you have to look at the code.

fkiraly:
>So what should be done here? Is predict interval only valid for out of sample fh in vecm ? Or does statsmodel have an incomplete version of vecm or have I understood something wrong?

I don't know, this is sth you can learn only from the code. Intention you can learn from asking a `statsmodels` dev.

For practical purposes, you may not want to spend too much time on this, and instead document what you do and the decision you made.

AurumnPegasus:
Okay I will just close this PR for now since this needs me to go a lot in depth into `statsmodel`'s code and intentions. Maybe `predict_interval` could be a separate PR.  For now, what I know of `statsmodel` and `VECM`:
1. They do not have a direct implementation of `predict_intervals`, the closest any function comes to it is `forecast_intervals`, which is called from `predict` in `VECMResults`. This `forecast_intervals` is only called when an `alpha` (or `coverage` according to `sktime`) is specified. 
2. `forecast_intervals` claims to return `confidence_intervals` according to the docstring. I did a few experimentation and I think it actually returns `predict_intervals` and is probably a mistake on `statsmodel`'s end (although one could confirm this once). 
3. Only issue here is that `statsmodel`'s `predict` takes `fh` to be out of sample predictions only (for insample predictions I took inspiration from `VAR`'s implementation in `sktime`). This is an issue when it comes to `predict_intervals` since I do not know how to compute the same for in sample predictions (unlike `pmdarima`, which was my reference in this case)

AurumnPegasus:
@fkiraly  re-opened the PR, ideally it should be merge-able for current functionalities

fkiraly:
>I did a few experimentation and I think it actually returns `predict_intervals` and is probably a mistake on `statsmodel`'s end

Can you kindly describe what you did and how you conclude it is probably that?

fkiraly:
I would suggest, can we just patch it to the `predict_interval`, and document in the class docstring what we have done, so users can then judge whether they want to use it that way or not?

fkiraly (approved):
Nice, tests pass now!

We work on `predict_interval` in a separate PR, as agreed.

PR #2830, opened 2022-06-17
[ENH] A new `ReconcilerForecaster` estimator for reconciling forecasts using base model residuals 

ciaran-g:
#### Reference Issues/PRs

Added a new estimator `ReconcilerForecaster` which contains three reconciliation methods. See https://github.com/alan-turing-institute/sktime/issues/2157

A new `ReconcilerForecaster` class which is a composite of a base forecaster and a reconciliation method. The idea/methods are similar to the `Reconciler`, except here the reconciliation matrix is estimated using variations on the base-forecast residual covariance matrix. The three methods either use the sample covariance matrix `method="mint_cov"`, a shrunk covariance matrix `method="mint_shrink"`, or only the on-diagonal elements `method="wls_var"`.

No

#### Any other comments?

I was wondering if it would be easier or more intuitive for users if all of the reconciliation methods, including the transformers, were accessible though this estimator?

##### For all contributions

##### For new estimators


sktime/forecasting/reconcile.py
```diff
@@ -0,0 +1,355 @@
...
+# !/usr/bin/env python3 -u
+# copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+"""Implements reconciled forecasters for hierarchical data."""
+
+__all__ = ["ReconcilerForecaster"]
+__author__ = [
+    "ciaran-g",
+]
+
+# todo: include the reconciler transformers?
```
fkiraly, line 11: that's not a silly idea at all!

You could import the transformer, and if a string for the "transformer-like" reconciliation is given, you could just do `Reconciler(strategy) * forecaster` inside. (i.e., I would not copy-paste code but wrap the transformer, optionally depending on strategy)

sktime/forecasting/reconcile.py
```diff
@@ -0,0 +1,355 @@
...
+    _parent_child_df,
+)
+
+
+class ReconcilerForecaster(BaseForecaster):
+    """Hierarchical reconcilation forecaster.
+
+    Reconciliation is applied to make the forecasts in a hierarchy of
+    time-series sum together appropriately.
+
```
fkiraly, line 33: maybe a few more lines how it relates to the `forecaster` arg, i.e., that forecasts are made by `forecaster` and then etc

sktime/forecasting/reconcile.py
```diff
@@ -0,0 +1,355 @@
...
+        X : pd.DataFrame, default=None
+            Exogenous variables for the base forecaster
+
+        Returns
+        -------
+        self : reference to self
+        """
+        self._check_method()
+
+        # # check the length of index if not hierarchical just return forecaster
```
fkiraly, line 109: well, not 100% precise, you fit the forecaster and return reference to self (which is the correct thing to do)

sktime/forecasting/reconcile.py
```diff
@@ -0,0 +1,355 @@
...
+
+from warnings import warn
+
+import numpy as np
+import pandas as pd
+from numpy.linalg import inv
+
+from sktime.forecasting.base import BaseForecaster, ForecastingHorizon
+from sktime.transformations.hierarchical.aggregate import _check_index_no_total
+from sktime.transformations.hierarchical.reconcile import (
```
fkiraly, line 22: hm... should this move somewhere? The "parent child", or more generally, graph stuff, could fit in `datatypes` as utilities for the hierarchical data. Although, for the moment, let´s leave it where it is to avoid refactor complexity sneaking in. Worth thinking about though.

sktime/forecasting/reconcile.py
```diff
@@ -0,0 +1,355 @@
...
+        self.forecaster_ = self.forecaster.clone()
+        self.forecaster_.fit(y=y, X=X, fh=fh)
+
+        # now summation matrix
+        self.s_matrix = _get_s_matrix(y)
+
+        # parent child df
+        self.parent_child = _parent_child_df(self.s_matrix)
+
+        # bug in self.forecaster_.predict_residuals() for heir data
```
fkiraly, line 133: oops - well, can we fix it there?

sktime/forecasting/reconcile.py
```diff
@@ -0,0 +1,355 @@
...
+        fh_resid = ForecastingHorizon(
+            y.index.get_level_values(-1).unique(), is_relative=False
+        )
+        self.residuals = y - self.forecaster_.predict(fh=fh_resid, X=X)
+
+        if self.method == "mint_cov":
```
fkiraly, line 139: instead of having strings, would it make sense to have `shrink` and `diag_only` directly as hyperparameters?
Although the current solution seems better to me if we will have methods here that do not use `_get_g_matrix_mint` or a simple extension, in the future.

sktime/forecasting/reconcile.py
```diff
@@ -0,0 +1,355 @@
...
+        self.s_matrix = _get_s_matrix(y)
+
+        # parent child df
+        self.parent_child = _parent_child_df(self.s_matrix)
+
+        # bug in self.forecaster_.predict_residuals() for heir data
+        fh_resid = ForecastingHorizon(
+            y.index.get_level_values(-1).unique(), is_relative=False
+        )
+        self.residuals = y - self.forecaster_.predict(fh=fh_resid, X=X)
```
fkiraly, line 137: this should be `residuals_` according to the condition that this is a fitted parameter of the model

sktime/forecasting/reconcile.py
```diff
@@ -0,0 +1,355 @@
...
+        X : pd.DataFrame, default=None
+            Exogenous variables based to the base forecaster
+        update_params : bool, optional (default=True)
+            whether model parameters should be updated
+
+        Returns
+        -------
+        self : reference to self
+        """
+        if y.index.nlevels < 2:
```
fkiraly, line 205: the update seems incorrect or problematic.

Let's say we update only with a single time point, i.e., a length 1 `y`, then we have only a single residuals and compute matrices on only a single residuals!

Should we not use, say, `self._y` instead? In which case this would be equivalent to `fit` on all seen data, but the matrix updates could be done more efficiently, e.g., by the WSM formula. (although that might be overkill for an initial PR?)

sktime/forecasting/reconcile.py
```diff
@@ -0,0 +1,355 @@
...
+
+        # copy for further mods
+        resid = self.residuals.copy()
+        # mean scale
+        grp_range = np.arange(self.residuals.index.nlevels - 1).tolist()
+        resid = resid.groupby(level=grp_range).apply(lambda x: x - x.mean())
+        # cov matrix
+        resid = resid.unstack().transpose()
+        nobs = len(resid)
+        cov_mat = resid.transpose().dot(resid) / (nobs - 1)
```
fkiraly, line 268: can't we use numpy.cov here?

fkiraly (changes requested):
Looks great, thanks!

Blocking issue: I think `update` is not correct?

Other things are not blocking.

Further, can you kindly point me to the precise location fo the shrinkage formulae in the references? The things you are implementing starting line 272.

sktime/forecasting/reconcile.py
ciaran-g: Ok cool 👍

sktime/forecasting/reconcile.py
ciaran-g: added a few more lines

sktime/forecasting/reconcile.py
ciaran-g: Yeah I think we should maybe leave it as is then? I agree it will be easier to add methods that use the residuals or training data but not the covariance matrix, e.g. top down historical proportions.

sktime/forecasting/reconcile.py
ciaran-g: yep, I simplified this section a wee bit! 👍

ciaran-g:
Just updated @fkiraly with changes to the update function, added the `Reconciler()` transformers, and simplified the shrinkage calc a wee bit.

The shrinkage calculation is detailed in Appendix A of this [paper](https://doi.org/10.2202/1544-6115.1175). I was originally doing it the calculation the long way round as detailed in the paper when I started off, but realised you can do it faster with some re-arranging thanks to the folks at the [fable](https://github.com/tidyverts/fabletools) R package

ciaran-g:
Yep, just updated to also store the matrices in `ReconcilerForecaster()` if the reconciliation method is one of the transformers, so it matches the other methods

sktime/forecasting/reconcile.py
fkiraly: ok, makes sense

fkiraly (approved):
`update` looks good now, thanks.

Non-blocking comment:

* what `self.forecaster` is is not consistent across the methods. One time, it is the fitted forecaster, one time it is the entire pipeline. I would ensure it is the pipeline in both cases, i.e., either bind it to the fittable pipeline element, or execute the pipeline manually.
* The `self.forecaster_` attribute should be mentioned in the docstring.

fkiraly:
moved the remaining points here: https://github.com/alan-turing-institute/sktime/issues/2875 and merged, so it goes in for 0.12.1

aiwalter:
@ciaran-g do you mind to add this new forecaster also to docs?

PR #2831, opened 2022-06-17
[ENH] subsetting transforms

fkiraly:
This PR contains two transformers for subsetting/slicing:

* `IndexSubset`, which subsets a time series `X` to indices of `y`
* `ColumnSelect`, which subsets a multivariate time series `X` to a pre-specified set of column names or integer indices of columns

Tests specific to the transformers are included.

### DIFF
--- a/docs/source/api_reference/transformations.rst
+++ b/docs/source/api_reference/transformations.rst
@@ -442,6 +442,20 @@ These transformers select features in `X` based on `y`.
 
     FeatureSelection
 
+Subsetting time points and variables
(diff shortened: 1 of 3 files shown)


ilkersigirci:
Isn't `ColumnSelect` class has the same functionality as `FeatureSelection` ? What is the extra use case for this new class?

```

 X = load_longley()[1]

transformer = FeatureSelection(method="columns", columns=["GNPDEFL", "POP", "FOO"])

transformed_X  = transformer.fit_transform(X)
```

PR #2834, opened 2022-06-20
[ENH] Pass frequency explicitly in `ForecastingHorizon`

### LINKED ISSUE #1750: [ENH] address FutureWarning: Timestamp.freq is deprecated and will be removed in a future version

khrapovs:
#### Reference Issues/PRs

Partially addresses #1750. More changes are necessary, but I would prefer to break the change into smaller logically complete PRs.

Following this conversation: https://github.com/alan-turing-institute/sktime/pull/2694#issuecomment-1146769665, and [this enhancement proposal](https://github.com/sktime/enhancement-proposals/blob/main/steps/17_forecasting_horizon_freq/step.md#2a-earlies-possible-pass-in-constructor), I added a new optional argument `freq` to the constructor of `ForecastingHorizon`.

Frequency is inferred in `ForecastingHorizon` methods indirectly, via the forecaster's `cutoff` passed to methods. In all uses, this is equal to the `cutoff` of the forecaster, which is the last time index of `y` seen in `fit`. This currently carries the `freq` for all cases of time-like indexing, but will lose it after deprecation, since single time indices will no longer carry `freq`.

In this PR frequency is passed directly upon construction or via the setter method. Cutoff is no longer used as a source of frequency information.

The biggest consequence of this design update is necessity to pass frequency to the `ForecastingHorizon` at some point for the forecasters to preserve their behavior. The minimal code change is to assign frequency via setter method to already existing `fh` in a few of `sktime.forecasting.base._base.BaseForecaster` methods. Frequency is obtained from `y` passed to `fit`, `predict`, and other public methods of this base class.

No

##### For all contributions


sktime/forecasting/base/_base.py
```diff
@@ -260,6 +266,7 @@ def fit(self, y, X=None, fh=None):
         self._is_vectorized = vectorization_needed
         # we call the ordinary _fit if no looping/vectorization needed
         if not vectorization_needed:
+            fh = self._update_fh_freq(fh=fh, index=y_inner.index)
```
fkiraly, line 269: could we instead move it to `_check_fh` which gets an additional argument, `y_inner`?

khrapovs: I had the same idea and I tried myself. But `sktime.forecasting.base._base.BaseForecaster.predict_residuals` stops me (for reference, to reproduce run `pytest --no-cov sktime/forecasting/tests/test_all_forecasters.py::TestAllForecasters::test_predict_residuals[DirRecTabularRegressionForecaster-fh=1-int-int-True]`). This method something strange with `fh` and I can not unravel it.

fkiraly: can you push a branch where that breaks? It does not break on this branch or on `main`, does it?

khrapovs: I have pushed one more commit. I have managed to do what you asked. Hopefully there are no side effects... I run only tests that are most likely to fail with this change and they pass.

The usual problem, the longer you refactor, the higher the probability that the code gives up :D

sktime/forecasting/base/_base.py
```diff
@@ -260,6 +257,7 @@ def fit(self, y, X=None, fh=None):
         self._is_vectorized = vectorization_needed
         # we call the ordinary _fit if no looping/vectorization needed
         if not vectorization_needed:
+            fh = self._check_fh(fh=fh, index=y_inner.index)
```
fkiraly, line 260: this will be problematic, since we will still need to do this when we vectorize!

Suggest to do this directly after `_check_X_y` (line 249 or 250)  and not in the vectorization case distinction.

sktime/forecasting/base/_fh.py
```diff
@@ -212,9 +257,11 @@ def _new(
         values : pd.Index, pd.TimedeltaIndex, np.array, list, pd.Timedelta, or int
             Values of forecasting horizon.
         is_relative : bool, default=same as self.is_relative
-        - If None, determined automatically: same as self.is_relative
-        - If True, values are relative to end of training series.
-        - If False, values are absolute.
+            - If None, determined automatically: same as self.is_relative
+            - If True, values are relative to end of training series.
+            - If False, values are absolute.
+        freq : str, optional (default=None)
```
fkiraly, line 263: did you consider making this polymorphic, as discussed in the STEP?

I.e., allowing multiple typers of input for `freq`, including `pandas` data containers or indices? If those are passed, then `infer_freq` will unwrap. So you can do `_check_fh(fh, y)` instead of `_check_fh(fh, index)`

fkiraly: (also the docstring should reflect that)

sktime/forecasting/base/_base.py
khrapovs: I do not entirely understand the effect of `_vectorize` method, but my guess was that it takes the method in the first argument, say `fit`, rearranges the data, and runs this `fit` with it, effectively returning back to  `vectorization_needed = False`. So, `_check_fh` is still called.

fkiraly: perhaps it still works, but it confounds the layers and makes the structure (and with it, future refactors) more difficult. Better leave it where it was...

fkiraly: yes, a reference to the original container is in the `X` attribute, but now that you point it out it is not too good to do it that way.

Argh, I see, this gets us into coupling territory... there are two ways to die, couple the `ForecastingHorizon` with `VectorizedDF`, or do what you did.

Is there a solution that allows us to decouple `ForecastingHorizon` and `VectorizedDF`?

fkiraly: how do you handle `numpy` anyway?

khrapovs: > how do you handle `numpy` anyway? Staying away from `VectorizedDF` for a moment...

`_check_X_y` converts the data into "Series, Panel, or Hierarchical object, or VectorizedDF". Is it even possible to get `y_inner` or `X_inner` as numpy arrays?

khrapovs: As you asked, I moved `_check_fh` back to its original position outside of the `if...else` statement. In some cases it is moved one or two lines lower just to get access to `y_inner` which appears only after `_check_X_y`.

I have also changed the arguments of `_check_fh` so that `y_inner` or `self._y` are passed instead of their indices. the problem of index extraction is transferred to `infer_freq` which now uses `singledispatch` to address objects of various types.

fkiraly:
@khrapovs, there is already a utility that produces a `pandas.Index` for most `sktime` data containers, `get_time_index`, could that work?
https://github.com/alan-turing-institute/sktime/pull/2852

khrapovs:
> @khrapovs, there is already a utility that produces a `pandas.Index` for most `sktime` data containers, `get_time_index`, could that work? #2852

Thanks for pointing this out! Now this function is used in `infer_freq`.

fkiraly (changes requested):
I think this is getting better!
My main concern is the high coupling being introduced in the `BaseForecaster` class, and that `fit` etc calls can now fail for exotic mtypes, we need to ensure that conversions happen to the assumed mtypes of the frequency inference.

My suggestion:
* in `_check_fh`, use `self._y` instead of `y_inner`. Why: `self._y` is always in one of the "utility friendly" formats. For this, `_check_fh` needs to happen after the first `_update_y_X` , so you need to move the position in `fit` (but only there).
* The call to `self_y` in `_check_fh` can be silent, it already looks up private parameters. This should be added in the docstring.
* personally, I would then push the frequency inference down to `check_fh` (without underscore, in the `utilities` module)

sktime/forecasting/base/_fh.py
fkiraly: I think the docstrings are now not accurate anymore, could you kindly update to what is actual?

khrapovs: Which docstrings do you have in mind? I briefly went through those that are affected by this PR and they look ok to me.

khrapovs:
> * in `_check_fh`, use `self._y` instead of `y_inner`. Why: `self._y` is always in one of the "utility friendly" formats. For this, `_check_fh` needs to happen after the first `_update_y_X` , so you need to move the position in `fit` (but only there).
> * The call to `self_y` in `_check_fh` can be silent, it already looks up private parameters. This should be added in the docstring.
> * personally, I would then push the frequency inference down to `check_fh` (without underscore, in the `utilities` module)

Awesome! Again, it was not immediately obvious, that `y` is always stored in `self._y` in acceptable format. I have implemented all three suggestion above. As a result, the change in `_base.py` reduced even more. Thanks!

sktime/forecasting/base/_base.py
```diff
@@ -1408,6 +1411,7 @@ def _check_fh(self, fh):
         Reads and writes to self._fh
         Writes fh to self._fh if does not exist
         Checks equality of fh with self._fh if exists, raises error if not equal
+        Assigns the frequency inferred from self._y
```
fkiraly, line 1414: to where/what?

fkiraly (changes requested):
Blocking comments, to be addressed or discussed (e.g., not necessary):
* why is there a change to the forecasting notebook? Seems like only execution dates etc are changed. If this is the case, please reset and remove the change from this PR.
* why do we need the `freq` arg in `_to_absolute` and `_to_relative`? Can´t we get it from the `ForecastingHorizon`?
* can we push `infer_freq` into `ForecastingHorizon`, so it can take `freq` a string, or an index? For instance, `test_to_absolute_freq`, line 302, could then be `fh = ForecastingHorizon([1, 2, 3], freq=train)`. Or, line 1471 in `BaseForecaster._check_fh` could be `fh = check_fh(fh=fh, self._y)`

sktime/utils/validation/forecasting.py
```diff
@@ -293,6 +295,9 @@ def check_fh(fh, enforce_relative=False):
     if enforce_relative and not fh.is_relative:
         raise ValueError("`fh` must be relative, but found absolute `fh`")
 
+    if fh.freq is None:
```
fkiraly, line 298: why do we need this? Seems superfluous

sktime/utils/validation/forecasting.py
```diff
@@ -271,6 +271,8 @@ def check_fh(fh, enforce_relative=False):
         Forecasting horizon specifying the time points to predict.
     enforce_relative : bool, optional (default=False)
         If True, checks if fh is relative.
+    freq : str, optional (default=None)
+        Frequency string
```
fkiraly, line 275: please add details to the docstring how the parameters influence the output, and what errors are raised

sktime/forecasting/base/_base.py
khrapovs: added "to the returned forecasting horizon object"

sktime/utils/validation/forecasting.py
khrapovs: A user or some internal function may pass `ForecastingHorizon` object to `check_fh` which was not initialized with `freq` argument (default = `None`). If so, we update this attribute here.

sktime/utils/validation/forecasting.py
khrapovs: details are added

khrapovs:
> * why is there a change to the forecasting notebook? Seems like only execution dates etc are changed. If this is the case, please reset and remove the change from this PR.

I had to add `freq="M"` to the construction of `ForecastingHorizon`. Otherwise, conversion `to_relative` fails in the following cells.

> * why do we need the `freq` arg in `_to_absolute` and `_to_relative`? Can´t we get it from the `ForecastingHorizon`?

We don't. Sorry, my lack of attention. It was a remnant from an earlier version where `freq` was not an attribute.

> * can we push `infer_freq` into `ForecastingHorizon`, so it can take `freq` a string, or an index? For instance, `test_to_absolute_freq`, line 302, could then be `fh = ForecastingHorizon([1, 2, 3], freq=train)`. Or, line 1471 in `BaseForecaster._check_fh` could be `fh = check_fh(fh=fh, self._y)`

I believe it was our mutual agreement (https://github.com/sktime/enhancement-proposals/blob/main/steps/17_forecasting_horizon_freq/step.md#2a-earlies-possible-pass-in-constructor) that passing frequency string as a minimal information to `ForecastingHorizon` is our preferred option. Same with `check_fh`, we pass not a monster `y`, but rather a simple string. You now think differently?

fkiraly:
>> why is there a change to the forecasting notebook? Seems like only execution dates etc are changed. If this is the case, please reset and remove the change from this PR.

> I had to add freq="M" to the construction of ForecastingHorizon. Otherwise, conversion to_relative fails in the following cells.

Can you kindly ensure that the changes are localized to those cells only, and cells are not executed?

fkiraly:
> I believe it was our mutual agreement (https://github.com/sktime/enhancement-proposals/blob/main/steps/17_forecasting_horizon_freq/step.md#2a-earlies-possible-pass-in-constructor) that passing frequency string as a minimal information to `ForecastingHorizon` is our preferred option. Same with `check_fh`, we pass not a monster `y`, but rather a simple string. You now think differently?

My preference would be for the argument to be polymorphic, that it can be a string, or the entire `y` and it extracts the string (for easy declaration syntax).

That´s this section: https://github.com/sktime/enhancement-proposals/blob/main/steps/17_forecasting_horizon_freq/step.md#accepting-more-freq-inferrable-formats
(but I´m not too strongly opined about this option)
Well, we did not agree on a particular choice

sktime/forecasting/base/_base.py
fkiraly: but then this is in the wrong section, this is the section about what it does to `self`

khrapovs:
> Can you kindly ensure that the changes are localized to those cells only, and cells are not executed?

It was not easy, but I managed to reduce the change in the notebook to just 3 lines.

sktime/forecasting/base/_base.py
khrapovs: sorry, I am confused. your question was "to where/what?". I have added the end of the sentence to answer this question. What am I missing?

khrapovs:
> My preference would be for the argument to be polymorphic, that it can be a string, or the entire `y` and it extracts the string (for easy declaration syntax).
> 
> That´s this section: https://github.com/sktime/enhancement-proposals/blob/main/steps/17_forecasting_horizon_freq/step.md#accepting-more-freq-inferrable-formats (but I´m not too strongly opined about this option) Well, we did not agree on a particular choice

Let me disagree and voice a couple of arguments for a simple `freq` argument.

1. It is simple. Forecasting horizon gets the object ready to be used internally. It does not have to bother with any extraction. If we start giving it all kinds of objects, it will get yet another responsibility of extracting frequency from all of them.
2. Right now there are not many places where we actually need to pass the frequency. And not many objects to extract frequency from. So, I would prefer to postpone polymorphism implementation to the point when it is really needed.

Again, I trust your judgement more than mine, so would be happy to implement your request if you insist.

khrapovs:
> How about option 2?

Perfect! Implemented and pushed.

fkiraly (changes requested):
Sorry for repeated change request.

Question: why is it not possible to leave the example in the forecasting tutorial as it is? Why does the `freq` arg need to be passed to it *twice*? I do not think this should be necessary, and `freq` should be automatically extracted from the index object if it has a `freq`.

This would be a significant interface change, we should avoid it!

I might have missed it, but I always assumed it is inferred from the first argument of the constructor if it can be, but looking through the code it does not seem to be the case.

sktime/forecasting/base/tests/test_fh.py
```diff
@@ -335,7 +343,7 @@ def test_relative_to_relative_with_integer_horizon(freqstr):
     """Test converting between relative and absolute with integer horizons."""
     # Converts from relative to absolute and back to relative
     train = pd.Series(1, index=pd.date_range("2021-10-06", freq=freqstr, periods=3))
-    fh = ForecastingHorizon([1, 2, 3])
+    fh = ForecastingHorizon([1, 2, 3], freq=freqstr)
```
fkiraly, line 346: does this not work without `freq`?
Should it not pick up the frequency now from `train.index`?

sktime/utils/validation/forecasting.py
fkiraly: makes sense

fkiraly (changes requested):
Excellent!

Seems that the interface does not change much anymore, so we are almost there!

One final question: there are instances in the tests where a `ForecastingHorizon` is constructed with an iterable of integer, and `freq` is passed. Should that not automatically be inferred now?

I think we should leave the existing tests as they are and ensure they pass, since they test common usage, specifically `test_to_absolute_freq` and `test_absolute_to_absolute_with_xxx_horizon` and `test_relative_to_relative_with_integer_horizon` should remain unchanged and pass.

sktime/forecasting/base/tests/test_fh.py
khrapovs: Unfortunately, no. `train.index` is `pd.DatetimeIndex`, `train.index[-1]` is `pd.Timestamp` which is precisely what we don't want to be the source of frequency information.

fkiraly:
>Yes, it can be inferred from `cutoff`, but only if it is `pd.Period` and not `pd.Timestamp` as it is in all test cases in `test_fh.py`.

I think there are two problems with it:

* I think this is factually incorrect, it can be inferred, and easily (?)
* even if it were correct, we cannot change common usage patterns of `ForecastingHorizon`, at least not without major deprecation effort

Why I think your statement is factually incorrect:

```python
import pandas as pd

idx = pd.date_range(start='2020/12/01', end='2020/12/30', periods=30)
pd.infer_freq(idx)
# gives 'D'
type(idx[0])
# gives Timestamp
```

khrapovs:
> * I think this is factually incorrect, it can be inferred, and easily (?)

Sure, `pd.DatetimeIndex`, `pd.PeriodIndex`, and `pd.Period` objects all have `freq` and `freqstr` attributes. But `pd.Timestamp` will not have this in some future. This is exactly the fact that I am trying to address here. One way or another we have to rule out frequency extraction from `pd.Timstamp`. Some sacrifices have to be made and this is already my second proposal, as you may remember.

> * even if it were correct, we cannot change common usage patterns of `ForecastingHorizon`, at least not without major deprecation effort

Please, let me know what kind of deprecation would you like to see. Should I replace `FutureWarning` that we currently get all over the place from `pandas` with our own?

> Why I think your statement is factually incorrect:
> 
> ```python
> import pandas as pd
> 
> idx = pd.date_range(start='2020/12/01', end='2020/12/30', periods=30)
> pd.infer_freq(idx)
> # gives 'D'
> type(idx[0])
> # gives Timestamp
> ```

This code snippet is correct, but...
1. `pd.Timestamp` will not have `freq` attribute in the future.
2. `idx` is not passed at any stage to `ForecastingHorizon`. We decided to pass frequency inferred from `idx` via `freq` argument in construction. Following your request I even preserved the possibility to infer frequency from `cutoff`, but only if it happens to be `pd.Period` object, not `pd.Timstamp` as in your code snippet.

fkiraly:
> This code snippet is correct, but...
> 1. `pd.Timestamp` will not have `freq` attribute in the future.

Yes, you are right. I was wrong in saying that this was factually incorrect, since you were clearly stating this in the post-deprecation state of `pandas` where this will no longer work, so my counterargument is incorrect.

It's vexing, it does not seem to be easy to solve without major interface changes.

So I went back to the drawing board here, trying to minimize public API changes:

https://github.com/alan-turing-institute/sktime/issues/1750
https://github.com/sktime/enhancement-proposals/pull/24

Let me know what you think.

khrapovs:
> So I went back to the drawing board here, trying to minimize public API changes:
> 
> #1750 [sktime/enhancement-proposals#24](https://github.com/sktime/enhancement-proposals/pull/24)
> 
> Let me know what you think.

In this PR I have solved the problem for a user. I do not see an alternative solution in your enhancement proposal. It keeps frequency with cutoff internally, but does not solve the problem of a user passing `pd.Timestamp` as a `cutoff` to `ForecastingHorizon`.

fkiraly:
>  It keeps frequency with cutoff internally, but does not solve the problem of a user passing `pd.Timestamp` as a `cutoff` to `ForecastingHorizon`. 

It does, because it uses your solution for `ForecastingHorizon`, i.e., with the additional `freq` argument. My addition concerns the `BaseForecaster` internal `cutoff`.

fkiraly:
for completeness: after discussion on slack, I branched off this PR and modified it based on "freq as property of `ForecastingHorizon`" idea: https://github.com/alan-turing-institute/sktime/pull/2932

sktime/forecasting/base/tests/test_fh.py
fkiraly: yes, you are right - that would not work.
Still, not passing `freq` *should* work for the user. This test, in #2932, has the old syntax for `fh` construction.

fkiraly:
@khrapovs, should we close this, or are you going to use parts of this?

khrapovs:
Closing as this branch was subsumed by https://github.com/alan-turing-institute/sktime/pull/2932

Issue #2573, opened 2022-04-28
[ENH] design - init, parameter check, dynamic tags - conflicting interface requirements


fkiraly:
We need to have a design decision on conflicting elements of our current interface specification.

These are:

1. `__init__` should contain no logic for input checking, and optimally no other logic. Such logic should be placed at the start of `_fit`
2. dynamic tags need to be set in the `__init__`, since tags are read at the start of the `fit` call which precedes the `_fit` call
3. we have the `fit_is_empty` tag, which can mean that `_fit` is never called for a given estimator, as well as estimators/object types that have no `fit` (such as distances)

Here, 1. conflucts with 2. and with 3., but 2. and 3. do not conflict with each other (I think).

I feel we need to think about potential ways to resolve this situation.

Related: https://github.com/alan-turing-institute/sktime/issues/1798

fkiraly:
FYI @lmmentel who highlighted this.
FYI @miraep8 who recently thought about it in #2540 and #2458.
FYI @aiwalter, @TonyBagnall

fkiraly:
My personal opinion is, 1. should go.
The requirement of 1. is coming from `sklearn`, the rationale being that calling `fit` multiple times, and possibly `set_params` in-between, should be possible and result in the same state as after the last `fit` call.

`sklearn` has no choice in this imo, since there is no "boilerplate" layer and all the checks etc need to be implemented by a user every ime they write an estimator.
We, however, do have a choice - we can call parameter setting and checking logic from the layer between `fit` and `_fit`.

Like this: https://github.com/alan-turing-institute/sktime/pull/2562

We still would be `sklearn` compatible, since the external logic would be indistinguishable.

There is one - positive! - exception:
with this, the user receives feedback at *specification* of an invalid estimator, not just at `fit` which is later.
This would be in line with the principle "give feedback on invalid input as early as possible", and `sklearn` is *not*  in line with that!

fkiraly:
a more extreme variant of this would be dedicated parameter check methods, as here:
https://github.com/alan-turing-institute/sktime/issues/1798
though that might be overkill.

miraep8:
[...]

PR #2835, opened 2022-06-20
[ENH] `set_params` to call `reset`, to comply with `sklearn` parameter interface assumptions

fkiraly:
This PR modifies `set_params` to also execute a full `reset`, which calls any `__init__` code such as dynamic tag setting.

Also adds parameter inspection related class methods to `BaseObject`:
* `get_param_names` - retrieves list of parameter names
* `get_param_defaults` - retrieves name/default dict of parameter defaults

Why `set_params` should execute a `reset`: @aiwalter pointed me to this specification document in `sklearn` which makes a key interface assumption explicit, namely that the result of `set_params` must be the same as `__init__(**self.get_params())`.

Since `sktime` may have some `__init__` content besides `self` writes (in deviation from `sklearn` due to varia discussed in #2573), we can ensure consistency here by calling `reset` from `set_params`.
This relates to discussion in, and also partially addresses https://github.com/alan-turing-institute/sktime/issues/2573 .

As a secondary effect of this, the `test_set_params` test was failing in instances where parameter related logic was present in `__init__`, e.g., cloning estimators to `component_name_`, due to testing logic in `sklearn`'s `check_set_params` which was called from `test_set_params`. The reason for that is that `sklearn`'s test would overwrite parameters with `np.inf` etc to mimic a change, which is fine in `sklearn` since it strictly has no logic in `__init__`. However, `sktime` estimators may have logic, such as dynamic tag setting, which therefore would break the `set_params` now identical with `__init__`.

Therefore, the `test_set_params` test was changed from overwriting with `np.inf` etc, to overwriting with values from other test cases as available through `get_test_params` and `get_param_defaults` (needed for full set), which ensures validity.

### DIFF
--- a/sktime/base/_base.py
+++ b/sktime/base/_base.py
@@ -54,6 +54,7 @@ __all__ = ["BaseEstimator", "BaseObject"]
 
 import inspect
 import warnings
+from collections import defaultdict
 from copy import deepcopy
 
 from sklearn import clone
@@ -113,6 +114,48 @@ class BaseObject(_BaseEstimator):
         Equal in value to `type(self)(**self.get_params(deep=False))`.
         """
         return clone(self)
+
+    def set_params(self, **params):
+        """Set the parameters of this object.
+
+        The method works on simple estimators as well as on nested objects.
+        The latter have parameters of the form ``<component>__<parameter>`` so that it's
+        possible to update each component of a nested object.
+
+        Parameters
+        ----------
+        **params : dict
+            BaseObject parameters
+
+        Returns
+        -------
+        self : reference to self (after parameters have been set)
+        """
+        if not params:
+            # Simple optimization to gain speed (inspect is slow)
+            return self
+        valid_params = self.get_params(deep=True)
+
+        nested_params = defaultdict(dict)  # grouped by prefix
+        for key, value in params.items():
+            key, delim, sub_key = key.partition("__")
+            if key not in valid_params:
+                raise ValueError(
+                    "Invalid parameter %s for object %s. "
+                    "Check the list of available parameters "
+                    "with `object.get_params().keys()`." % (key, self)
+                )
+
+            if delim:
+                nested_params[key][sub_key] = value
+            else:
+                setattr(self, key, value)
+                valid_params[key] = value
+
+        self.reset()
+
+        for key, sub_params in nested_params.items():
+            valid_params[key].set_params(**sub_params)
 
     @classmethod
     def get_class_tags(cls):
--- a/sktime/base/_base.py
+++ b/sktime/base/_base.py
@@ -14,9 +14,11 @@ Interface specifications below.
 
     class name: BaseObject
 
-Hyper-parameter inspection and setter methods
-    inspect hyper-parameters      - get_params()
-    setting hyper-parameters      - set_params(**params)
+Parameter inspection and setter methods
+    inspect parameter values      - get_params()
+    setting parameter values      - set_params(**params)
+    list of parameter names       - get_param_names()
+    dict of parameter defaults    - get_param_defaults()
 
 Tag inspection and setter methods
     inspect tags (all)            - get_tags()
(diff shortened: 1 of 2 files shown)


sktime/base/_base.py
```diff
@@ -114,6 +115,51 @@ def clone(self):
...
+                    "with `object.get_params().keys()`." % (key, self)
+                )
+
+            if delim:
+                nested_params[key][sub_key] = value
+            else:
+                setattr(self, key, value)
+                valid_params[key] = value
+
+        self.reset()
```
fkiraly, line 155: for reviewers: line 155 is added. The rest is identical with `sklearn`'s `set_params`

sktime/tests/test_all_estimators.py
```diff
@@ -776,8 +775,47 @@ def test_set_params(self, estimator_instance):
...
+        is_equal, equals_msg = deep_equals(
+            estimator.get_params(), params, return_msg=True
+        )
+        msg = (
+            f"get_params result of {type(estimator).__name__} (x) does not match "
+            f"what was passed to set_params (y). Reason for discrepancy: {equals_msg}"
+        )
+        assert is_equal, msg
+
+    def test_set_params_sklearn(self, estimator_class):
```
fkiraly, line 791: this simply mimicks `sklearn` `check_set_params`, the only change is replacing `np.inf` dummy values with alternative test parameter sets from `get_test_params`. `sklearn` can't do this since it does not store test parameter sets, and assumes `np.inf` is always fine.

fkiraly:
the bug was not with the multiplexers, but with the test (parameters not overwritten by defaults in the loop)

sktime/base/_base.py
```diff
@@ -111,9 +114,126 @@ def clone(self):
...
+            # Simple optimization to gain speed (inspect is slow)
+            return self
+        valid_params = self.get_params(deep=True)
+
+        nested_params = defaultdict(dict)  # grouped by prefix
+        for key, value in params.items():
+            key, delim, sub_key = key.partition("__")
+            if key not in valid_params:
+                raise ValueError(
+                    "Invalid parameter %s for object %s. "
```
miraep8, line 218: Any reason not to use f-strings over % formatting?  I think it was released in python 3.6, so should be valid for all python versions we support.  Not a blocker by any means, I just find it reads better.  Can use triple double quotes to split over multiple lines.

sktime/base/_base.py
```diff
@@ -111,9 +114,126 @@ def clone(self):
...
+        # fetch the constructor or the original constructor before
+        # deprecation wrapping if any
+        init = getattr(cls.__init__, "deprecated_original", cls.__init__)
+        if init is object.__init__:
+            # No explicit constructor to introspect
+            return []
+
+        # introspect the constructor arguments to find the model parameters
+        # to represent
+        init_signature = inspect.signature(init)
```
miraep8, line 145: Thoughts on using inspect.getargspec(init) instead.  From my understanding it returns a named tuple of the args, but also the `varargs` and `keywords` (ie `args` and `kwargs`) which seems like it would make what comes next much easier.... ie:
if len(args) > 0 -> p = args[1:]
  if `varargs` -> raise error

sktime/base/_base.py
```diff
@@ -111,9 +114,126 @@ def clone(self):
...
+            keys are all parameters of cls that have a default defined in __init__
+            values are the defaults, as defined in __init__
+        """
+        parameters = cls._get_init_signature()
+        default_dict = {
+            x.name: x.default for x in parameters if x.default != inspect._empty
+        }
+        return default_dict
+
+    def set_params(self, **params):
```
miraep8, line 192: Personally - I think one great thing to consider building in to the set_params function (since we are building our own) is a check for type!! I did a very sketchy version of this in my draft PR of the HMM estimator, but I think an even better way to do it would be to make use of the inspect signature library (ie can get annotations for parameters if type hints are provided).  Putting in a check like this can ensure that variables always remain the type intended.  Anyway - just an idea.  Seems like we are moving in the direction of using more type hints though - so could be a good idea?

sktime/base/_base.py
miraep8: Hmm, just kidding, I guess `getfullargspec` is the most up to date version of that.  However, it does seem that signature is generally the preferred method, so I contradict myself, what you initially had is probably fine.

miraep8:
Sorry I missed the original request to review a few days ago.  Glad the bug was resolved.  Threw in a few ideas/thoughts that came up while looking over the code.

sktime/base/_base.py
fkiraly: > Any reason not to use f-strings over % formatting?

I tend to use f-strings over %, the specific reason here is that these are code blocks copy-pasted from scikit-learn, which are probably older and use %.

sktime/base/_base.py
fkiraly: > Thoughts on using inspect.getargspec(init) instead?

This is refactored scikit-learn code, and they use `signature`.
I try to be consistent in the content. Anyway, I think it is not an issue, more of a discussion.

sktime/base/_base.py
fkiraly: If at all, it should not happen in `set_params` but in `__init__`.
Whether it should happen in init or not is part of this design discussion https://github.com/alan-turing-institute/sktime/issues/2573, feel free to put your opinions there!

fkiraly:
> Glad the bug was resolved

Well, it wasn´t one, I made a stupid mistake.

I was testing whether you could `set_params` with a second feature set, by getting param sets from `get_test_params`. In cases where the first parameter set had parameters that the second did not overwrite, this could lead in inconsistent parameter combinations. To obtain consistent parameter combinations, one needs to revert those parameters (from the first set that the second does not overwrite) to the defaults, which the test now does.

lmmentel (approved):

Issue #2841, opened 2022-06-21
[BUG] FeatureSelection on pd-multiindex:  Error in check_is_mtype, more than one mtype identified: ['pd-multiindex', 'pd_DataFrame_Table']


CelieDs:
**Describe the bug**

Hello everyone ! First, thank you for the great tool!
I encountered an error when trying to select the features of my dataset containing time-series panel data.
I converted the dataset to pd-multiIndex, following the 'examples/AA_datatypes_and_datasets.ipynb' tutorial and when I try to apply the fit_transform function, the following error appears:

```python
TypeError                                 Traceback (most recent call last)
Input In [9], in <cell line: 2>()
      1 transformer = FeatureSelection(method="feature-importances", n_columns = 20)
----> 2 Xt_GradientBoostingRegressor = transformer.fit_transform(X, y)

File ~\Miniconda3\lib\site-packages\sktime\transformations\base.py:434, in BaseTransformer.fit_transform(self, X, y)
    380 """Fit to data, then transform it.
    381 
    382 Fits the transformer to X and y and returns a transformed version of X.
   (...)
    430         Example: i-th instance of the output is the i-th window running over `X`
    431 """
    432 # Non-optimized default implementation; override when a better
    433 # method is possible for a given algorithm.
--> 434 return self.fit(X, y).transform(X, y)

File ~\Miniconda3\lib\site-packages\sktime\transformations\base.py:293, in BaseTransformer.fit(self, X, y)
    290     raise ValueError(f"{self.__class__.__name__} requires `y` in `fit`.")
    292 # check and convert X/y
--> 293 X_inner, y_inner = self._check_X_y(X=X, y=y)
    295 # checks and conversions complete, pass to inner fit
    296 #####################################################
    297 vectorization_needed = isinstance(X_inner, VectorizedDF)

File ~\Miniconda3\lib\site-packages\sktime\transformations\base.py:690, in BaseTransformer._check_X_y(self, X, y, return_metadata)
    687 elif X_scitype == "Hierarchical":
    688     y_possible_scitypes = ["Table", "Panel", "Hierarchical"]
--> 690 y_valid, _, y_metadata = check_is_scitype(
    691     y, scitype=y_possible_scitypes, return_metadata=True, var_name="y"
    692 )
    693 if not y_valid:
    694     raise TypeError("y " + msg_invalid_input)

[...]

Issue #2711, opened 2022-05-28
[BUG] Hierarchical Forecasting does not work with reconciler 


ngupta23:
**Describe the bug**
Hierarchical Forecasting does not work with reconciler (possibly due to empty X dataframe)

**To Reproduce**
Modified from pydata Berlin notebooks. Complete code & data [here](https://github.com/ngupta23/sktime-tutorial-pydata-berlin-2022/blob/109e25cafe798dc2d24d96427a5fd62866bab901/notebooks/minimal_issue_hierarchical.ipynb).

```python
steps=[
        ("daily_season", DateTimeFeatures(ts_freq="D")),  
        ("daily_season2", DateTimeFeatures(manual_selection=["week_of_month", "day_of_quarter"])),
]  

forecaster_hier = NaiveForecaster()

# Version 1A: Local Vectorized (Hierarchical) Forecasts (without reconciliation)
pipe_hier_norecon = ForecastingPipeline(steps= steps + [("forecaster", forecaster_hier)])

# Version 1B: Local Vectorized (Hierarchical) Forecasts (with reconciliation)
pipe_hier_recon = Aggregator() * pipe_hier_norecon * Reconciler(method="ols")

# Version 1A: Local Vectorized (Hierarchical) Forecasts (without reconciliation)
# This works
_ = pipe_hier_norecon.fit(y_train_hier, X_train_hier, fh=FH)

# Version 1B: Local Vectorized (Hierarchical) Forecasts (with reconciliation)
# Does not work with X_train (is it because of empty X dataframe?)
# If 1A works with X_train, shouldn't 1B also work?
_ = pipe_hier_recon.fit(y_train_hier, X_train_hier, fh=FH)

# This works (without passing X)
# _ = pipe_hier_recon.fit(y_train_hier, fh=FH)
```

```python-traceback
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Input In [10], in <cell line: 4>()
      1 # Version 1B: Local Vectorized (Hierarchical) Forecasts (with reconciliation)
      2 # Does not work with X_train (is it because of empty X dataframe)
      3 # If 1A works with X_train, shouldn't 1B also work?
----> 4 _ = pipe_hier_recon.fit(y_train_hier, X_train_hier, fh=FH)

File ~\.conda\envs\sktime-pydata\lib\site-packages\sktime\forecasting\base\_base.py:265, in BaseForecaster.fit(self, y, X, fh)
[...]

PR #2844, opened 2022-06-21
[BUG] Vectorization in transformers overwrote y with X

fkiraly:
Thie fixes the following unreported bug related to https://github.com/alan-turing-institute/sktime/issues/2841 and https://github.com/alan-turing-institute/sktime/issues/2711:

when a transformer entered vectorization mode and `y` was passed (and also vectorized), then the transformer would pass the vectorized `X` to the inner methods where a vectorized `y` should have passed, i.e., passing `X` as `X` (correctly), and `y` as `X` as well (incorrectly). This has been fixed.

(was difficult to spot since the condition is a rare combination of (a) a transformer uses `y` at all, and (b) vectorization happens)

Relation to the bugs:
* after the type inference bug in #2841 is fixed, the test example posted by me breaks due to this unreported bug
* in #2711, it seem like `X` in the forecaster is not properly aggregated. There is not a clear causal link, and the vectorization in forecasters seem correct. Still there is something going on that seems similar, so I reference it.

### DIFF
--- a/sktime/transformations/base.py
+++ b/sktime/transformations/base.py
@@ -743,7 +743,7 @@ class BaseTransformer(BaseEstimator):
         #   then apply vectorization, loop method execution over series/panels
         elif case == "case 3: requires vectorization":
             iterate_X = _most_complex_scitype(X_inner_scitype)
-            X_inner = VectorizedDF(X=X, iterate_as=iterate_X, is_scitype=y_scitype)
+            X_inner = VectorizedDF(X=X, iterate_as=iterate_X, is_scitype=X_scitype)
             # we also assume that y must be vectorized in this case
             if y_inner_mtype != ["None"] and y is not None:
                 # raise ValueError(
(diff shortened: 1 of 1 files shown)


patrickzib (approved):
Looks good :)

chrisholder (approved):
Some of the best code I've ever had the pleasure of reviewing - well done!

fkiraly:
so... no one asked why it is correct that the last arg is `X_scitype` and not `y_scitype`??

fkiraly:
These were also switched! How could you!

fkiraly:
>Young padawan has a long way to go

"if you end your review now - if you choose the quick and easy path as Vader did - you will become an agent of evil"

patrickzib:
> > Young padawan has a long way to go
> 
> "if you end your review now - if you choose the quick and easy path as Vader did - you will become an agent of evil"

Do. Or do not. There is no try!

fkiraly:
jokes aside, really sorry for this. I should have noticed earlier...
If you wouldn't mind having another look?

Issue #2826, opened 2022-06-16
[ENH] addition of window based change point detection algorithm 


lielleravid:
**Is your feature request related to a problem? Please describe.**
Segmentation/annotation not currently implemented in sktime. 
Using change point algorithms one can segment time series data by change points found. Window-based change point is one of many change point algorithms.

**Describe the solution you'd like**
Add support for window based change point detection by interfacing or adding as an adapter into Sktime. 
There is an option to interface from the rupture open source library. 
https://centre-borelli.github.io/ruptures-docs/code-reference/detection/window-reference/#ruptures.detection.window.Window

**Additional context**
for more information about the algorithm and it's usages:
https://centre-borelli.github.io/ruptures-docs/user-guide/detection/window/

duydl:
I'm currently developing an adapter for rupture. As rupture algorithms are batch processing and unsupervised, the adapter structure would closely resemble the currently implemented algorithms in sktime, inheriting from BaseTransformer. There would not be much friction.

duydl:
As we may be talking about different ways of categorizing CPD Algos here, I would like to clarify that ruptures algorithms are **optimization-based** CPD, which finds change points by optimizing some cost functions. Window sliding is just one of its procedures to search for the solutions, with the other like segmentation (binary, bottom-up), dynamic programming (Pelt). 

I think an adapter to ruptures would have an interface allowing users to select from all cost functions and search procedures, not just limiting to window-based. 

Ruptures does not support combining multiple cost functions, so I think this is one of the areas where we could enhance from the original.

fkiraly:
@duydl, I didn't read your reply as rude, just noting a few things:

* @lielleravid worked on this as part of an academic project in 2022, this was one of multiple "optional" items which did not get implemented in the end. I.e., she is unlikely to work on it anymore (please correct us if wrong)
* this issue is about window-based change point detection *in general*, could be a de-novo implementation, or interfacing from somewhere, as @lielleravid also explains. Interfacing from `ruptures` - which has a wider coverage of algorithms, as you say, @duydl - would be just one way to satisfy the requirement.

So, possible approaches to this issue would be:

* implement window-based change point detection from scratch
* take @lielleravid's incomplete PR https://github.com/sktime/sktime/pull/2851 and get it over the finish line
* interface `ruptures` or another package - @madhuri723 has opened an issue here: #6156

PR #2851, opened 2022-06-23
[ENH] window based change point detection adapter #2826

lielleravid:
Issue #2826

I have implemented a Window based change point detection algorithm in the form of an adapter for the ruptures Window class 

The contribution creates a new soft dependency as such, pyproject.toml and check_soft_dependencies.py have been updated as well.

As there is no template chosen yet for annotation/change point/segmentation I went ahead and chose what I thought was best and was to inherit from BaseSeriesAnnotator class. 
I can refactor any changes if needed or if any other template would be better I will make the changes. Suggestions are more than welcome!

##### For all contributions
- [x ] I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/main/.all-contributorsrc).

##### For new estimators

### DIFF
--- a/build_tools/azure/check_soft_dependencies.py
+++ b/build_tools/azure/check_soft_dependencies.py
@@ -35,6 +35,7 @@ SOFT_DEPENDENCIES = {
     "sktime.alignment.dtw_python": ["dtw"],
     "sktime.clustering": ["tslearn"],
     "sktime.transformations.series.kalman_filter": ["pykalman", "filterpy"],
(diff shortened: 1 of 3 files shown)


lmmentel:
Great start! It seems the test is failing when ruptures internals are trying to call `.reshape` probably expecting a `np.ndarray` but getting `pd.Series`. Might be worth looking into what ` annotation/tests/test_all_annotators.py::test_output_type[WindowBasedChangePoint]` is actually doing.

PR #2852, opened 2022-06-23
[BUG] fixing `get_time_index` for 1D and 2D numpy formats

fkiraly:
The utility function was incorrect for `get_time_index` for 1D and 2D numpy formats, where the 0-th index is time. This has been fixed.

Also adds functionality for `VectorizedDF` (by looking for the `X` attribute)

### DIFF
--- a/sktime/datatypes/_utilities.py
+++ b/sktime/datatypes/_utilities.py
@@ -9,8 +9,12 @@ import pandas as pd
 def _get_index(x):
     if hasattr(x, "index"):
         return x.index
+    elif isinstance(x, np.ndarray):
+        if x.ndim < 3:
+            return pd.RangeIndex(x.shape[0])
+        else:
+            return pd.RangeIndex(x.shape[-1])
     else:
-        # select last dimension for time index
         return pd.RangeIndex(x.shape[-1])
 
 
@@ -23,7 +27,7 @@ def get_time_index(X):
     in one of the following sktime mtype specifications for Series, Panel, Hierarchical:
     pd.DataFrame, pd.Series, np.ndarray, pd-multiindex, nested_univ, pd_multiindex_hier
     assumes all time series have equal length and equal index set
-    will *not* work for list-of-df, pd-wide, pd-long
+    will *not* work for list-of-df, pd-wide, pd-long, numpyflat
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


fkiraly:
@khrapovs, there was a bug in `get_index`, and this also adds functionality for `VectorizedDF`. Would appreciate if you can have a look, it should directly benefit your PR..

sktime/datatypes/_utilities.py
```diff
@@ -45,10 +49,17 @@ def get_time_index(X):
             return X.index
     # numpy3D and np.ndarray
     elif isinstance(X, np.ndarray):
-        return _get_index(X)
+        # np.ndarray
+        if X.ndim < 3:
```
khrapovs, line 53: That's a repetition of `_get_index` function. I think you can leave
```
elif isinstance(X, np.ndarray):
    return _get_index(X)
```
as it was before.

khrapovs (changes requested):
I am strongly convinced that any change in code should be accompanied by unit tests that fail without the proposed change. Although I do not see any problems in this PR, I am not convinced that this PR fixes anything. Test were passing before and they still pass. I would kindly ask you to add a few unit tests for the `get_time_index` function.

fkiraly:
>I would kindly ask you to add a few unit tests for the get_time_index function.

Yes, good idea. I think originally @danbartl wrote the function, and did not add unit tests, being short on time in prep for pydata Berlin.

>Although I do not see any problems in this PR, I am not convinced that this PR fixes anything.

The problem is that what is being fixed is currently outside test scope.
The utility is used only in forecasting, and there afaik we do not test `numpy` based fixtures for runtime reasons. Perhaps an oversight that we should also correct, at least test a single estimator or similar.

khrapovs:
> The problem is that what is being fixed is currently outside test scope. The utility is used only in forecasting, and there afaik we do not test `numpy` based fixtures for runtime reasons. Perhaps an oversight that we should also correct, at least test a single estimator or similar.

How about testing this utility function directly, regardless of any context it would be used in the future? E.g. from my own PR: https://github.com/alan-turing-institute/sktime/pull/2834/files#diff-efc08ccad6fac520ff7827e6511a6fc39b0685d914605c9f38b215a2c6fbd372R65

fkiraly:
> How about testing this utility function directly

That doesn´t cover all the cases. I was thinking using the fixtures from `datatypes` and then running conversions to cover all the mtypes.

sktime/datatypes/_utilities.py
fkiraly: well, I thought, it was an unnecessary duplication of `get_time_index`, so I removed it in order to have the logic in one place.
I left `_get_index` in existence (and fixed it) since it was being called from elsewhere.

khrapovs (changes requested):
Tests are coming, right?

fkiraly:
>Tests are coming, right?

Yes, yes, but I can't be everywhere ... 😁

fkiraly:
more tests!

khrapovs (approved):

PR #2854, opened 2022-06-23
[ENH] `CutoffSplitter` with missing observations

khrapovs:
#### Reference Issues/PRs

Next step in #1737

Handle missing observations in `CutoffSplitter`. In particular, if an observation or a few are missing from the time series `y`, then `CutoffSplitter` still returns correct train/test window `.iloc` indices referencing "incomplete" original `y`. To achieve this,
- "incomplete" index of `y` is first regularized to make it equally spaced
- then standard splitting is applied to the regularized index
- and finally, indices existing in the original `y` are returned as values in train/test windows.

No

Do the new train/test windows match the intuition/expectation of a user in case of missing observations? For more long-term vision, please, refer to https://github.com/sktime/enhancement-proposals/pull/23.

#### Any other comments?

I will generalize other splitters one by one in separate PRs to make each one easily digestible.

##### For all contributions


fkiraly (changes requested):
question before I review:

what you mean by "missing", is not actually missing data but irregular indices in which some indices are missing. Correct or wrong?

Either way, can you quickly describe in math or code an example behaviour that previously would not work or work differently, and how it works after the PR?

fkiraly:
>Before I start writing anything new, please, take a look at this open PR

Ah, I missed that, I somehow overlooked that header. Thanks for pointing me to the STEP, that is sufficient to explain the idea.

fkiraly (changes requested):
Ok, I understand what this is doing.

First of all: good observation, great idea!

But, I think we cannot merge this.

Why: this carries out some bloody surgery on the `BaseSplitter` interface, and it is a substantial interface change:

* `freq` argument added to the splitters
* the generic `split` now converts the index to a regular one
* the `iloc` references are no longer to the original `y` but to `y_regular`

There are two comments, one about procedure and one about content.

Procedure: if you make changes as substantive as these, this needs to be described in the STEP. I.e., what exactly are you going to change the base class interface to, and why it is needed. E.g., what you done with `freq` and change to the `split` interface.

Content 1: I do not think we want `split` indices in return to refer to a `y_regular`, but still to `y`. Or, we cannot. Why: in the past, splitters would work for any time series index. Now, they will break with time series index where a `freq` cannot be inferred or attached.

Content 2: Internally, I would use `get_slice` instead of trying to smash it into regular format. The former always works, the latter only works if your indices are multiples of some regular period, see "content 1". This would also not require inferring, or passing, a `freq`.

Issue #2799, opened 2022-06-12
[ENH] [Annotation Workstream] Implement HMM annotation estimator


miraep8:
I am opening this up as an issue for the new annotation/segmentation workstream. 

My plan is to implement a Hidden Markov Model (HMM) estimator for annotation of time series data. You can read more about the algorithm [here](https://en.wikipedia.org/wiki/Hidden_Markov_model), but for a brief overview:

- You want to model a series of variables over time $X$, where $x_t$ represents the variable state at time $t$. 
- Unfortunately, you cannot measure or observe $X$ directly, it is a *hidden* variable, but fortunately you can observe a series of related variables $Y$, where $y_t$ represents your observation of $Y$ at time $t$. 
- You assume both that $y_t$ depends on $x_t$ (and only on $x_t$), and that $x_{t+1}$ depends on $x_t$ but only on $x_t$ (this assumption that each state is only influenced by the state immediately before it is known as a *Markov* assumption).
- One can use the [Baum-Welch algorithm](https://en.wikipedia.org/wiki/Baum%E2%80%93Welch_algorithm) (a special case of EM) to help find the unknown parameters that define the probabilistic relationship that relates $x_t$ and both $y_t$ and $x_{t+1}$.
- If you have defined the distributions that relate $x_t$ to $y_t$ (known as the emission distribution) and $x_t$ to $x_{t+1}$ (known as the transmission distribution) then you can calculate the most likely sequence of $X$ given the observed sequence $Y$ using a variety of algorithms (for example the [Viterbi algorithm](https://en.wikipedia.org/wiki/Viterbi_algorithm#Example))

This is an attractive algorithm for a variety of reasons - its been around for a long time (helped with the lunar landing if I am not mistaken), its a tool that people from a variety of different fields are familiar with, and has good statistical/algorithmic tools to work with it!   

To start with we will work with something that assigns non-overlapping single dimensional labels to univariate time series data (we will expand later, but getting this simple case to work would be nice). This will be a batch, not online approach, and will be an unsupervised algorithm. 

There are two paths forward:

[...]

PR #2855, opened 2022-06-24
[ENH] HMM annotation estimator

miraep8:
I wanted to share a draft PR of what I have been working on with HMM estimator. 

<details>
<summary>More Details:</summary>
<br>

- rather than assuming a particular emission distribution function, which I believe is what `hmmlearn` does (ie you can choose to use a Guassian HMM vs a Multinomial Discrete HMM, and you need to either build, or hope an HMM with your emission distribution has already been built), we simply accept a list of functions which behave as pdf and use those to calculate the emission probability.  The viterbi algorithm gets to treat those as a black box. 
    - the hope here is that we can add some functionality which automatically generates some nice functions for people if needed (ie give me an n-dimensional gaussian!), but it also gives the flexibility to do more interesting things!  - Have an empirical distribution you want to act as a "pdf"?  Have some other distribution you are getting from some other system that you would like to use? As long as it can return a valid probability then it *should* work with our HMM estimator, no need to write your own!
- I have spent some time optimizing the algorithms with numpy (in particular, the `_calculate_trans_mats` function which is the slowest part), the plan is to push this a bit more and then conduct speed tests between this draft estimator and `hmmlearn`, so we can have a more quantitative comparison of speed. 
- I implemented a very simple version of a base object, named `SimpleBaseEstimator` and used that here.  It is very similar to what Lukasz did in his annotation flavored PR actually, just added in the `get_params()` and `set_params()` functions, which are technically required for it to be a sklearn like estimator. 
    - I am not proposing we replace our Base Object with what I wrote at all, just wanted to see how many of the bells and whistles in Base Estimator I could get away without! Again - just for testing purposes.

### Example of how to use it: 

```
from sktime.annotation.hmm import HMM
from scipy.stats import norm
from numpy import asarray

centers = [3.5,-5]
sd = [.25 for i in centers]
emi_funcs = [(norm.pdf, {'loc': mean, 'scale': sd[ind]}) for ind, mean in enumerate(centers)]

test = HMM(emi_funcs, asarray([[0.25,0.75], [0.666, 0.333]]))
obs = asarray([3.7,3.2,3.4,3.6,-5.1,-5.2,-4.9])
test.fit(obs)
print(test.predict())
```

<\details>

@fkiraly, @lmmentel, @GuzalBulatova


fkiraly:
>the hope here is that we can add some functionality which automatically generates some nice functions for people if needed

there currently is a dormant project which meant to introduce distribution interfaces to the ML context:
https://github.com/alan-turing-institute/skpro
maybe we can revisit that sometime. E.g., have `BaseDistribution` inherit from `BaseObject` 😄

fkiraly:
>I am not proposing we replace our Base Object with what I wrote at all, just wanted to see how many of the bells and whistles in Base Estimator I could get away without! Again - just for testing purposes.

hmmmmm..... all this denial just makes it more suspicious :pouting_cat:

miraep8:
Also - I **think** unrelated to this PR, there seem to have been some strange failures on the last CI/CD runthrough.  Can  see details by clicking on the small red x next to the failed run, but copied the error message below in case it is a more general issue, note the error only cropped up in `Install and test / test-mac (3.7) (pull_request)` and the other os/python versions seemed to work ok....  Probably just a one off, but flagging this just in case. 

```
=================================== FAILURES ===================================
____ [doctest] sktime.datasets._single_problem_loaders.load_UCR_UEA_dataset ____
[gw1] darwin -- Python 3.7.13 /Users/runner/hostedtoolcache/Python/3.7.13/x64/bin/python
088         series associated with each case. If return_X_y is False, column
089         n_dimensions+1 contains the class labels/target variable.
090     y: numpy array, optional
091         The class labels for each case in X, returned separately if return_X_y is
092         True, or appended to X if False
093 
094     Examples
095     --------
096     >>> from sktime.datasets import load_UCR_UEA_dataset
097     >>> X, y = load_UCR_UEA_dataset(name="Yoga")
UNEXPECTED EXCEPTION: ContentTooShortError('retrieval incomplete: got only 65422 out of 20908976 bytes')
  File "/Users/runner/hostedtoolcache/Python/3.7.13/x64/lib/python3.7/doctest.py", line 1337, in __run
    compileflags, 1), test.globs)
  File "<doctest sktime.datasets._single_problem_loaders.load_UCR_UEA_dataset[1]>", line 1, in <module>
  File "/Users/runner/hostedtoolcache/Python/3.7.13/x64/lib/python3.7/site-packages/sktime/datasets/_single_problem_loaders.py", line 99, in load_UCR_UEA_dataset
    return _load_dataset(name, split, return_X_y, extract_path)
  File "/Users/runner/hostedtoolcache/Python/3.7.13/x64/lib/python3.7/site-packages/sktime/datasets/_data_io.py", line 147, in _load_dataset
    extract_path=extract_path,
  File "/Users/runner/hostedtoolcache/Python/3.7.13/x64/lib/python3.7/site-packages/sktime/datasets/_data_io.py", line 73, in _download_and_extract
    urlretrieve(url, zip_file_name)
  File "/Users/runner/hostedtoolcache/Python/3.7.13/x64/lib/python3.7/urllib/request.py", line 288, in urlretrieve
    % (read, size), result)
urllib.error.ContentTooShortError: <urlopen error retrieval incomplete: got only 65422 out of 20908976 bytes>
/Users/runner/hostedtoolcache/Python/3.7.13/x64/lib/python3.7/site-packages/sktime/datasets/_single_problem_loaders.py:97: UnexpectedException
```

fkiraly:
@miraep8, clearly, your content is too short. Please fix the issue by making your content longer.

miraep8:
By the way @lmmentel, not saying that the test I am using is great (it is in fact not), but have a snippet of code in the testing file for generating a gaussian with shifting mean.  Plan to make this better (ie - just generate a file and load that as well instead of generating the data in real time) and also add some other tests.

miraep8:
Related to my issue #2799

miraep8:
Testing update, have been comparing with the Gaussian estimator from `hmmlearn`.  It seems like their package is much faster than mine, about 20-50 times after it seems, but that (at least in the synthetic data I have been running it on so far) ours seems to get closer to the truth! 

Here is code I have run to test it on!

```
#imports
from sktime.annotation.hmm import HMM
from scipy.stats import norm
import numpy as np
from hmmlearn.hmm import GaussianHMM
import time

#generating fake data: 
centers = [3.5,-5, 8, 10, .5]
sd = [.25 for i in centers]
seg_lengths = [1000, 500, 250 , 700, 150]
gauss_data = np.zeros([sum(seg_lengths)])
labels = np.zeros([sum(seg_lengths)])
i = 0
for ind, s_len in enumerate(seg_lengths):
    gauss_data[i:i+s_len] = np.random.default_rng().normal(loc=centers[ind], scale=sd[ind], size=[s_len])
    labels[i:i+s_len] = np.asarray([ind for _ in range(s_len)])
    i+= s_len

#running comparison:   
def convert_hmmlearn_output(old_array):
    output_dict = dict()
    count = 0
    new_array = []
    for i in old_array:
        if not i in output_dict.keys():
            output_dict[i] = count
            count += 1
        new_array.append(output_dict[i])
    return new_array

## Time for sktime HMM estimator:
start_time = time.time()
emi_funcs = [(norm.pdf, {'loc': mean, 'scale': 1}) for mean in centers]
transition_matrix = np.asarray([[((1/(s_len/14))/(len(centers)-1)) for state in centers] for s_len in seg_lengths])
for ind, s_len in enumerate(seg_lengths):
    transition_matrix[ind, ind] = 1 - (1/(s_len/14))
sktime_hmm = HMM(emi_funcs, transition_matrix)
sktime_hmm.fit(gauss_data)
sktime_predicted_labels = sktime_hmm.predict()
sktime_time = time.time() - start_time
print(f'sktime HMM estimator took {sktime_time:.4f} seconds to run')

## Time for hmmlearn HMM estimator:
start_time = time.time()
transition_matrix = np.asarray([[((1/(s_len/14))/(len(centers)-1)) for state in centers] for s_len in seg_lengths])
hmmlearn_hmm = GaussianHMM(n_components=len(centers), means_prior = np.asarray(centers).reshape(-1, 1), covars_prior = sd, covariance_type='spherical',
                           transmat_prior = transition_matrix)
hmmlearn_hmm.fit(gauss_data.reshape(-1, 1))
hmmlearn_predicted_labels = hmmlearn_hmm.predict(gauss_data.reshape(-1, 1))
hmmlearn_time = time.time() - start_time
print(f'hmmlearn HMM estimator took {hmmlearn_time:.4f} seconds to run')

if sktime_time > hmmlearn_time:
    print(f'Current sktime implementaiton is {(sktime_time/hmmlearn_time):.1f} time slower than hmmlearn')
else:
    print(f'Current sktime implementaiton is {(hmmlearn_time/sktime_time):.1f} time slower than hmmlearn')
hmmlearn_predicted_labels =convert_hmmlearn_output(hmmlearn_predicted_labels)
print(f'The percent agreement between the two algorithms {100*sum(hmmlearn_predicted_labels==sktime_predicted_labels)/len(gauss_data): .1f}%')
print(f'sktime agreement with ground truth: {100*sum(sktime_predicted_labels == labels)/len(labels): .1f}%')
print(f'hmmlearn agreement with ground truth: {100*sum(hmmlearn_predicted_labels == labels)/len(labels): .1f}%')

```

Which prints out: 

```
sktime HMM estimator took 0.8916 seconds to run
hmmlearn HMM estimator took 0.0371 seconds to run
Current sktime implementaiton is 24.0 time slower than hmmlearn
The percent agreement between the two algorithms  84.7%
sktime agreement with ground truth:  100.0%
hmmlearn agreement with ground truth:  84.7%
```

This is just one pass through, so will be collecting more samples, but thought this was an interesting observation.

miraep8:
@lmmentel @KatieBuc @fkiraly - I have refactored my HMM estimator to make use of the current BaseAnnotation class.  The current plan is to get it merged in its currently compatible format, and move some of the BaseEstimator stuff I was originally exploring here into a different PR. I also plan to add an option - 'use_hmmlearn' as an option to this estimator in a future PR. Ready for review! :)

fkiraly:
>I also plan to add an option - use_hmmlearn as an option to this estimator in a future PR

I would do that in a separate class - we have had similar discussions earlier on the multiple ARIMA models (I cannot find the issue anymore), on whether they should be in a single class or in multiple. The consensus was, they should be multiple, as long as it is coming from different packages with different interfaces - to isolate dependencies, and to follow the conceptual model of "one estimator = one parametric algorithm" more clearly.

sktime/annotation/hmm.py
```diff
@@ -0,0 +1,275 @@
...
+
+__author__ = ["miraep8"]
+__all__ = ["HMM"]
+
+
+class HMM(BaseSeriesAnnotator):
+    """Implements a simple HMM fitted with viterbi algorithm.
+
+    The HMM annotation estimator facilitates the appilcation of
+    the vertibi algorithm to fit a sequence of 'hidden state' class
```
fkiraly, line 24: Viterbi

sktime/annotation/hmm.py
```diff
@@ -0,0 +1,275 @@
...
+    annotations (represented by an array of integers the same size
+    as the observation) to a sequence of observations.
+
+    This is done by finding the most likely path given the emission
```
fkiraly, line 28: It would be crucial to be more precise to describe what the output is, given input.
I.e., given the X, what is being produced in `predict`? What is the fit doing?

The current docstring is just a high-level description of HMM, not of the specific procedure that fit/predict below carries out.

sktime/annotation/hmm.py
```diff
@@ -0,0 +1,275 @@
...
+    probabilities - (ie the probability that a particular observation
+    would be generated by a given hidden state), the transition prob
+    (ie the probability of transitioning from one state to another or
+    staying in the same state) and the initial probabilities - ie the
+    belief of the probability distribution of hidden states at the
+    start of the observation sequence).
+
+    Parameters
+    ----------
+    emission_funcs : a list of functions, which should be properly
```
fkiraly, line 38: must be more precise. What signature do the functions have?
Should also probably be callable.

sktime/annotation/hmm.py
```diff
@@ -0,0 +1,275 @@
...
+        normalized PDFs over the same space as the observed data.
+    transition_prob_mat: a nxn array of probabilities, where 'n'
```
fkiraly, line 40: please provide python type

sktime/annotation/hmm.py
```diff
@@ -0,0 +1,275 @@
...
+        represents the number of hidden states in the HMM.  Each
+        row should sumn to 1 in order to be properly normalized
+        (ie the j'th column in the i'th row represents the
+        probability of transitioning from state i to state j.)
+    initial_probs: optional, a 1d array of probabilities across
```
fkiraly, line 45: please specify python types, and specify which dimensions must be equal, e.g., should the length match something else, above?

sktime/annotation/hmm.py
```diff
@@ -0,0 +1,275 @@
...
+    --------
+    >>> from sktime.annotation.hmm import HMM
+    >>> from scipy.stats import norm
+    >>> from numpy import asarray
+    >>> # define the emission probs for our HMM model:
+    >>> centers = [3.5,-5]
+    >>> sd = [.25 for i in centers]
+    >>> emi_funcs = [(norm.pdf, {'loc': mean,
+    ...  'scale': sd[ind]}) for ind, mean in enumerate(centers)]
+    >>> test = HMM(emi_funcs, asarray([[0.25,0.75], [0.666, 0.333]]))
```
fkiraly, line 59: re naming, avoid calling things `test` or `temp` or similar. Names, even if short, should be as descriptive as possible. Better, perhaps: `hmm`, `hmm_est`

sktime/annotation/hmm.py
```diff
@@ -0,0 +1,275 @@
...
+        emission_funcs: list,
+        transition_prob_mat: np.ndarray,
+        initial_probs: np.ndarray = None,
+    ):
+        self.initial_probs = initial_probs
+        self.num_states = len(emission_funcs)
+        self.emission_funcs = emission_funcs
+        self.transition_prob_mat = transition_prob_mat
+        self.states = [i for i in range(self.num_states)]
+        self._validate_init()
```
fkiraly, line 80: should happen after `super.__init__` call

sktime/annotation/hmm.py
```diff
@@ -0,0 +1,275 @@
...
+        super(HMM, self).__init__(fmt="dense", labels="int_label")
+
+    def _validate_init(self):
+        """Verify that the parameters passed to init are well behaved."""
+        tran_mat_len = len(self.transition_prob_mat[:, 0])
```
fkiraly, line 85: clearer: `self.transition_prob_mat.shape[0]`

sktime/annotation/hmm.py
```diff
@@ -0,0 +1,275 @@
...
+        # transition_prob_mat should be square:
+        if not len(self.transition_prob_mat.shape) == 2 or not tran_mat_len == len(
```
fkiraly, line 87: use `x.ndim` instead of `len(x.shape)`

sktime/annotation/hmm.py
```diff
@@ -0,0 +1,275 @@
...
+                "Transtion Probability must be 2D square, but got an"
+                f"object of size {self.transition_prob_mat.shape}"
+            )
+        # number of states should be consistent!
+        init_prob_len = self.num_states
+        if self.initial_probs is not None:
+            init_prob_len = len(self.initial_probs)
+        if not tran_mat_len == self.num_states == init_prob_len:
+            raise ValueError(
+                "Number of hidden states is inconsistent!  Was passed "
```
fkiraly, line 100: say more explicitly:
* what *needs* to be the case
* which variables are being referred to here, so the user can compare to the function docstring by matching variable names rather than trying to match text

sktime/annotation/hmm.py
```diff
@@ -0,0 +1,275 @@
...
+    @classmethod
+    def _calculate_trans_mats(
+        cls,
+        initial_probs: np.ndarray,
+        emi_probs: np.ndarray,
+        transition_prob_mat: np.ndarray,
+        num_obs: int,
+        num_states: int,
+    ) -> Tuple[np.array, np.array]:
+        """Calculate the transition mats used in the viterbi algorithm.
```
fkiraly, line 124: capitalize Viterbi (name of a person)

sktime/annotation/hmm.py
```diff
@@ -0,0 +1,275 @@
...
+    def _hmm_viterbi_label(self) -> np.array:
+        """Assign hidden state ids to all observations based on most likely path.
+
+        Parameters
+        ----------
+        self - an HMM instance which already has already
+
+        Returns
+        -------
+            -hmm_fit: a 2xn array with the first column being position and the second
```
fkiraly, line 206: typo?

sktime/annotation/hmm.py
```diff
@@ -0,0 +1,275 @@
...
+            self.num_obs,
+            self.num_states,
+        )
+        self.trans_prob = trans_prob
+        self.trans_id = trans_id
+        return self
+
+    def _predict(self, X):
+        """Use the transition matrices to calculate the most likely state."""
+        return self._hmm_viterbi_label()
```
fkiraly, line 240: this can´t be right - `_hmm_viterbi_label` not using `X`? Can this only be used in-sample?

fkiraly (changes requested):
I think this is just the Viterbi algorithm that infers hidden labels based on known initial and transition probas, not Baum-Welch, correct?
In that case, I would not implement a `_fit`, or check in `_ predict` whether the `_fit` labels have already been seen. 
The `_predict` should always produce the inferred hidden states for its argument `X`,  but currently it produces those for `X`.

The more "interesting" version of thie would use Baum-Welch or another form of EM in `fit`, and `Viterbi` in `predict`.

sktime/annotation/hmm.py
```diff
@@ -0,0 +1,275 @@
+# -*- coding: utf-8 -*-
+"""
+HMM Annotation Estimator.
+
+Implements a basic Hidden Markov Model (HMM) as an annotation estimator.
+To read more about the algorithm, check out the wikipedia page:
+(https://en.wikipedia.org/wiki/Hidden_Markov_model
```
lmmentel, line 7: how about?
```suggestion
To read more about the algorithm, check out the `HMM wikipedia page <https://en.wikipedia.org/wiki/Hidden_Markov_model>`_.
```

sktime/annotation/hmm.py
lmmentel, line 24: alternative wording 

```suggestion
    The HMM annotation estimator uses
    the Veterbi algorithm to fit a sequence of 'hidden state' class
```

sktime/annotation/hmm.py
```diff
@@ -0,0 +1,275 @@
...
+        self,
+        emission_funcs: list,
+        transition_prob_mat: np.ndarray,
+        initial_probs: np.ndarray = None,
+    ):
+        self.initial_probs = initial_probs
+        self.num_states = len(emission_funcs)
+        self.emission_funcs = emission_funcs
+        self.transition_prob_mat = transition_prob_mat
+        self.states = [i for i in range(self.num_states)]
```
lmmentel, line 79: all attributes that are not part of the `__init__` signature should be assigned in `fit` to avoid side effects when using `set_params`

sktime/annotation/hmm.py
```diff
@@ -0,0 +1,275 @@
...
+        if self.initial_probs is not None:
+            init_prob_len = len(self.initial_probs)
+        if not tran_mat_len == self.num_states == init_prob_len:
+            raise ValueError(
+                "Number of hidden states is inconsistent!  Was passed "
+                f"{self.num_states} number of emission funcs, {tran_mat_len}"
+                f" as the size of the transition matrix, and  "
+                f" {init_prob_len} number of initial state probs"
+            )
+        for i in range(len(self.transition_prob_mat)):
```
lmmentel, line 105: how about using `np.isclose` and compare against a vector of ones to avoid the loop?

sktime/annotation/tests/test_all_annotators.py
```diff
@@ -23,4 +24,4 @@ def test_output_type(Estimator):
     estimator.fit(*args)
     args = _make_args(estimator, "predict")
     y_pred = estimator.predict(*args)
-    assert isinstance(y_pred, pd.Series)
+    assert isinstance(y_pred, pd.Series) or isinstance(y_pred, np.ndarray)
```
lmmentel, line 27: this could work as well

```suggestion
    assert isinstance(y_pred, (pd.Series, np.ndarray))
```

sktime/annotation/hmm.py
miraep8: Yes, that is cleaner

sktime/annotation/hmm.py
miraep8: Yes, also implemented Viterbi in another situation, and the comment was more relevant there. (biological annotation)

sktime/annotation/hmm.py
miraep8: Yes, good suggestion!

sktime/annotation/hmm.py
miraep8: I added some more details to that effect (and also moved all the logic to predict as you had suggested). Are you happier with it now, or does it still need more?

sktime/annotation/hmm.py
miraep8: Well, as you point out below the estimator (as it is) is only applying Viterbi to a single observation (and originally all the logic that extracted information from `X` was in `_fit`.  I have since moved everything to `_predict`.

miraep8:
Thanks both @lmmentel and @fkiraly for the useful feedback!  I have implemented your suggestions.

Regarding Baum-Welch/other EM - I agree that would definitely be more interesting.  I think my original plan was to add that on in a future PR, if that is ok with you? 

However I have now moved all logic out of `_fit` as was suggested and everything has been consolidated in `_predict`

sktime/annotation/hmm.py
```diff
@@ -0,0 +1,314 @@
...
+    proceeds it in the most likely path.  This logic is mostly carried
+    out by helper function _calculate_trans_mats.
+    Next, these matrices are used to calculate the most likely
+    path (by backtracing from the final mostly likely state and the
+    id's that proceeded it.)  This logic is done via a helper func
+    hmm_viterbi_label.
+
+    Parameters
+    ----------
+    emission_funcs : a list of length n.  Should be either a list of
```
GuzalBulatova, line 53: My suggestions for improvement regarding the docstrings: an example of conventional docs can be found in [boss.py file](https://github.com/alan-turing-institute/sktime/blob/main/sktime/classification/dictionary_based/_boss.py). For example, first line is the type, and broader description of a parameter starts from the next line, which makes it further easier to read. Sth like
```
parameter : float, optional, default=1.5
   Broader description of the parameter.
```

sktime/annotation/hmm.py
```diff
@@ -0,0 +1,314 @@
...
+        hmm_fit: (np.ndarray) an mx1 array where m is the length of the X (obs).
+            each entry in the array is an int representing a hidden id state
+            that has been assigned to that observation.
+        """
+        hmm_fit = np.zeros(self.num_obs)
+        # Now we trace backwards and find the most likely path:
+        max_inds = np.zeros(self.num_obs, dtype=np.int32)
+        max_inds[-1] = np.argmax(self.trans_prob[:, -1])
+        hmm_fit[-1] = self.states[max_inds[-1]]
+        for index in reversed(list(range(1, self.num_obs))):
```
GuzalBulatova, line 251: This might read less intuitive but we could use just `range` with all its params `(start, stop, step)` here, like 
`for index in range(self.num_obs, 0, -1)`, 
what do you think?

GuzalBulatova (commented):
Without exaggeration, this is a great contribution. Amazing work, well done, Mirae! 
I personally am especially grateful for type hints in methods, we should (or maybe already have started to) make it a common practice in sktime :)
I only have some minor style suggestions

sktime/annotation/hmm.py
miraep8: I think that reads a lot better actually!  Thanks for the suggestion!

sktime/annotation/hmm.py
miraep8: Thank you!  It was useful to see a more concrete example of what a good example of docstring looks like :)

GuzalBulatova: This looks great, thanks for the changes! 
I think the only docstr left is `_calculate_trans_mats`, and it'll be good to go.

miraep8: Ah yes, good catch!  Will work on that now.  Thanks again for the helpful feedback! :)

fkiraly (approved):
I think this is ok, great!

Re docstring, I think this is still a bit confusing - lot of text, but doesn't tell the user precisely. Not a blocker, but let's discuss at some later point.

lmmentel:
@fkiraly I cannot merge since the button is inactive for me since CI is waiting for the new checks. I'm happy for this to be merged.

PR #2859, opened 2022-06-24
& ris-bali [ENH] statsmodels DynamicFactor interface

lbventura:
#### Reference Issues/PRs

Addresses #2349. There are multiple failed tests due to the behavior of the `_predict` method, but further exploration in `testfile.ipynb` suggests that the method's behavior is actually correct. Could be that tests are wrong?

Deeper look into #2349 failed tests using check_estimator.

#### Any other comments?

(At least) Two outstanding points:
1. Implement the `predict_proba` method;
2. Create an example notebook - one cool example would be to forecast economic variables;

##### For all contributions

##### For new estimators


fkiraly:
I think I understand the problem now
the test is fine, it intentionally checks for indices that do not start at zero
the problem is with the (inner, statsmodels) dynamic factor model, it discards indices and produces forecasts starting at len(y) rather than at y.index[-1]
the solution would be to manually correct for that and simply overwrite the indices of what comes out of the model

lbventura:
Weirdly pre-commit is failing in "Install and test / code-quality (pull_request)" due to flake8, but worked in my local machine? See the latter below
Failing_Test

fkiraly:
You probably don't have all the linting checks installed locally then?
You can always go on "details" here to see what's wrong:
image

(I know what's wrong, and it is easy to fix - but kindly have a look yourself 😄 )

sktime/forecasting/dynamic_factor.py
```diff
@@ -0,0 +1,429 @@
+# -*- coding: utf-8 -*-
+# copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+
+"""Implements DynamicFactor Model."""
```
fkiraly, line 4: should say sth like "interface to statemodels etc", like in the class docstring

sktime/forecasting/dynamic_factor.py
```diff
@@ -0,0 +1,429 @@
...
+    statsmodels.tsa.statespace.dynamic_factor.DynamicFactor
+    statsmodels.tsa.statespace.dynamic_factor.DynamicFactorResults
+    References
+    ----------
+    [1] Lütkepohl, Helmut. 2007. New Introduction to Multiple Time Series Analysis.
+    Berlin: Springer.
+    """
+
+    _tags = {
+        "scitype:y": "multivariate",
```
fkiraly, line 121: this means that the data *must* have 2 or more variables. Does it not also work for 1 variable? Then the tag should be `"both"`.

sktime/forecasting/dynamic_factor.py
```diff
@@ -0,0 +1,429 @@
...
+        "ignores-exogeneous-X": False,
+        "handles-missing-data": True,
+        "y_inner_mtype": "pd.DataFrame",
+        "X_inner_mtype": "pd.DataFrame",
+        "requires-fh-in-fit": False,
+        "X-y-must-have-same-index": True,
+        "enforce_index_type": None,
+        "capability:pred_int": False,
```
fkiraly, line 129: hm, I thought, based on the formulation, that the model should be able to make proba predictions?

Though, I cannot find a method for that, so maybe let's leave this to another PR but make a note in the issue?

sktime/forecasting/dynamic_factor.py
```diff
@@ -0,0 +1,429 @@
...
+        error_var=False,
+        enforce_stationarity=True,
+        start_params=None,
+        transformed=True,
+        includes_fixed=False,
+        cov_type=None,
+        cov_kwds=None,
+        method="lbfgs",
+        maxiter=50,
+        full_output=False,  # Changed from 1
```
fkiraly, line 147: what does this mean?

sktime/forecasting/dynamic_factor.py
```diff
@@ -0,0 +1,429 @@
...
+import inspect
+
+import numpy as np
+from statsmodels.tsa.statespace.dynamic_factor import DynamicFactor as _DynamicFactor
+
+from sktime.forecasting.base.adapters import _StatsModelsAdapter
+from sktime.utils.validation._dependencies import _check_soft_dependencies
+
+_all_ = ["DynamicFactor"]
+__author__ = ["Ris-Bali"]
```
fkiraly, line 15: please add yourself as an author, too.

fkiraly (changes requested):
Looks good!

Blocking comments:
* we need to remove the notebook `testfile.ipynb`.
* the estimator should be added to the API reference in an appropriate section.

I also left a number of small, non-blocking comments, above.

sktime/forecasting/dynamic_factor.py
```diff
@@ -0,0 +1,429 @@
...
+
+_check_soft_dependencies("dynamicfactor", severity="warning")
```
fkiraly, line 17: there is no such package, at least it is not imported. Also, `statsmodels` is a core dependency, so you can remove this line.

sktime/forecasting/dynamic_factor.py
fkiraly: and in allcontributors, if you haven't done so (if you want to take the credit 😄). If @Ris-Bali is not there, we should also add them for the partial work.

sktime/forecasting/dynamic_factor.py
lbventura: I disabled the output when I was conducting testing because of "verboseness" but will reset to the defaults.

lbventura:
> the estimator should be added to the API reference in an appropriate section.



> Looks good!
> 
> Blocking comments:
> 
>     * we need to remove the notebook `testfile.ipynb`.
> 
>     * the estimator should be added to the API reference in an appropriate section.
> 
> 
> I also left a number of small, non-blocking comments, above.

On the second point, what exactly should be done? One example would be helpful :)

fkiraly:
> On the second point, what exactly should be done? One example would be helpful :)

goto `docs.source.api_reference.forecasting` and add an entry, following the pattern.
If not sure, jump quickly on discord.

fkiraly:
... odd, looks correct.

fkiraly:
ah, looks like a typo in the "references" section of the new estimator. See "details" next to the doc build, maybe you can find out what exactly the typo is.

lbventura:
It is complaining that the output is too long? Could it be due to the fact that I changed `full_output` to `True`?

> ... odd, looks correct.

It is complaining that the output is too long? Could it be due to the fact that I changed `full_output` to `True`?

fkiraly:
hm, estimator tests are failing now - any idea why? Perhaps you would like to change it back to the last state where they were not?

lbventura:
> hm, estimator tests are failing now - any idea why? Perhaps you would like to change it back to the last state where they were not?

The only reasons I can think of are:

1.  Having changed the sciptype from multivariate to both. I reverted back to multivariate;
2. Having changed the outputs of the model (the `full_output` point discussed above).

Let us see if the new commit addresses the problem.

fkiraly:
>The only reasons I can think of are:

>1.  Having changed the sciptype from multivariate to both. I reverted back to multivariate;

But, should it not be `both`? This estimator can deal with univariate data, can it not?

> 2. Having changed the outputs of the model (the `full_output` point discussed above).

Why did you do this?

The docs are still not building, btw, kindly check for typos.

lbventura:
> > The only reasons I can think of are:
> 
> > 1. Having changed the sciptype from multivariate to both. I reverted back to multivariate;
> 
> But, should it not be `both`? This estimator can deal with univariate data, can it not?

This was just a quick edit to see whether the problem comes from here. I think it should also work, but I will check.
> 
> > 2. Having changed the outputs of the model (the `full_output` point discussed above).
> 
> Why did you do this?

The docs were being built fine in a previous commit, reverted to see if this was the change that broke it, but aparently not.

> The docs are still not building, btw, kindly check for typos.

Will do

fkiraly:
> I think it should also work for univariate, but I will check. EDIT: it also does not work in statsmodels. `y` is just a 1D series generated with `_make_series`.

Ah, ok - then the value `"multivariate"` that you originally had was correct.

lbventura:
Tried to replicate `arima.py` forecaster docstring, hopefully this solves the readthedocs problem. Unfortunately the build logs are not very clear.

fkiraly:
> hopefully this solves the readthedocs problem. Unfortunately the build logs are not very clear.

Yes, looks like it! All green!

fkiraly (approved):
Nice, thanks!

lbventura:
> Nice, thanks!

Great, on to the next topic!

PR #2861, opened 2022-06-24
[ENH] `all_estimators` authors variable

fkiraly:
The registry lookup utility file was missing an `__author__` variable, this has been added based on commit history.

FYI @mloning, @KatieBuc, @miraep8, @xloem.


fkiraly:
for reviewers: @xloem contributed before the file was moved to its current location, from `sktime.utils.__init__`
(yes, the utility once lived in an `__init__` module)

PR #2862, opened 2022-06-24
[ENH] test for more than one parameter sets per estimator

fkiraly:
In order to test parameter settings for estimators properly, each important parameter should have been set to a non-default.
This PR introduces a test which checks that there are at least two test parameter sets (if the estimator has at least one parameter), and correctness of these sets.

Requires:
* https://github.com/sktime/sktime/pull/4279 for the general test for `get_test_params` (no specific number assumed)
* https://github.com/alan-turing-institute/sktime/pull/2835 as it uses the parameter name interface
* https://github.com/alan-turing-institute/sktime/pull/3428 to pass the `no-softdeps` step.

Current estimators are not tested due to differential testing.

If diff testing is turned off, helps tracking down the estimators that do not yet have two parameter sets, see #3429.


fkiraly:
FYI @julia-kraus, the test failures point precisely to the estimators with only one test parameter set.

PR #2866, opened 2022-06-25
[ENH] fix `get_window` utility when `window_length` was `None`

fkiraly:
The `get_window` utility was broken when `window_length` was `None`, it would return the object `obj` unchanged.

This has now been fixed.

In addition, the `lag` default has been set to zero, which is interpreted as a zero integer or `timedelta`, depending on `obj`.

Test cases for `None` input behaviour have been added.

### DIFF
--- a/sktime/datatypes/_utilities.py
+++ b/sktime/datatypes/_utilities.py
@@ -269,9 +269,11 @@ def get_window(obj, window_length=None, lag=None):
         # and always subset on first dimension
         if obj.ndim > 1:
             obj = obj.swapaxes(1, -1)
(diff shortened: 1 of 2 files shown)


khrapovs (changes requested):
Would you be so kind to update `sktime.datatypes.tests.test_utils.test_get_window_expected_result` to reflect the changes you made in this PR? Thanks!

fkiraly:
Good catch! No bugfix without tests that nail the correct behaviour down, optimally. Done.

khrapovs (approved):

PR #2870, opened 2022-06-26
[ENH] make `get_cutoff` compatible with all time series formats, fix bug for VectorizedDF input

### LINKED ISSUE #2872: [BUG] forecaster cutoff does not update when data requires vectorization
Forecasters do not update the `cutoff` when the data is vectorized in the hierarchical or panel case.

Example:

```python
from sktime.datatypes import get_examples

y = get_examples("pd_multiindex_hier")[0]
f = ARIMA()
f.fit(y, fh=1)

f.cutoff  # prints nothing
type(f.cutoff)  # it's None
```

Expected behaviour: `f.cutoff` should be `2` and type should be `int`.

The reason for this is the `get_cutoff` utillity which returns `None` if an unexpected type is passed, in this case `VectorizedDF` that `y_inner` is in case of vectorization.

The bug is addressed in https://github.com/alan-turing-institute/sktime/pull/2870.

fkiraly:
This PR:
* makes `get_cutoff` compatible with all time series input formats by putting a conversion step at the start of the function.
* fixes #2872, by extending `get_cutoff` so it can deal with `VectorizedDF`, and adds tests to check correct behaviour.

Other changes:
* updated docstrings, and corrected docstrings for functions in the module
* add `reverse_order` arg to `get_cutoff` to obtain the *first* time stamp instead of the *last* time stamp
* add `check_input` arg (default False) for input checks in `get_cutoff`, and optional input checks

### DIFF
--- a/sktime/datatypes/_utilities.py
+++ b/sktime/datatypes/_utilities.py
@@ -293,9 +293,11 @@ def get_window(obj, window_length=None, lag=None):
     window_length : int or timedelta, optional, default=-inf
         must be int if obj is int indexed, timedelta if datetime indexed
         length of the window to slice to. Default = window of infinite size
-    lag : int or timedelta, optional, default = 0 (correct type zero is inferred)
-        must be int if obj is int indexed, timedelta if datetime indexed
+    lag : int, timedelta, or None optional, default = None (zero of correct type)
         lag of the latest time in the window, with respect to cutoff of obj
+        if None, is internally replaced by a zero of type compatible with obj index
+        must be int if obj is int indexed or not pandas based
+        must be timedelta if obj is pandas based and datetime indexed
 
     Returns
     -------
--- a/sktime/datatypes/_utilities.py
+++ b/sktime/datatypes/_utilities.py
@@ -124,6 +124,10 @@ def get_cutoff(
     """
     from sktime.datatypes import check_is_scitype, convert_to
 
+    # deal with VectorizedDF
+    if hasattr(obj, "X"):
+        obj = obj.X
+
     if check_input:
         valid = check_is_scitype(obj, scitype=["Series", "Panel", "Hierarchical"])
         if not valid:
--- a/sktime/datatypes/_utilities.py
+++ b/sktime/datatypes/_utilities.py
@@ -83,7 +83,7 @@ GET_CUTOFF_SUPPORTED_MTYPES = [
     "np.ndarray",
     "pd-multiindex",
     "numpy3D",
-    "nested_univ"
+    "nested_univ",
     "df-list",
     "pd_multiindex_hier",
     "np.ndarray",
--- a/sktime/datatypes/_utilities.py
+++ b/sktime/datatypes/_utilities.py
@@ -91,7 +91,9 @@ GET_CUTOFF_SUPPORTED_MTYPES = [
 ]
 
 
-def get_cutoff(obj, cutoff=0, return_index=False, reverse_order=False):
+def get_cutoff(
+    obj, cutoff=0, return_index=False, reverse_order=False, check_input=False
+):
     """Get cutoff = latest time point of time series or time series panel.
 
     Assumptions on obj are not checked, these should be validated separately.
@@ -111,7 +113,9 @@ def get_cutoff(obj, cutoff=0, return_index=False, reverse_order=False):
         note: return_index=True may set freq attribute of time types to None
             return_index=False will typically preserve freq attribute
     reverse_order : bool, optional, default=False
-        if False, returns largest time index. If True, returns smallest time index.
+        if False, returns largest time index. If True, returns smallest time index
+    check_input : bool, optional, default=False
+        whether to check input for validity
 
     Returns
     -------
@@ -120,9 +124,10 @@ def get_cutoff(obj, cutoff=0, return_index=False, reverse_order=False):
     """
     from sktime.datatypes import check_is_scitype, convert_to
 
-    valid = check_is_scitype(obj, scitype=["Series", "Panel", "Hierarchical"])
-    if not valid:
-        raise ValueError("obj must be of Series, Panel, or Hierarchical scitype")
+    if check_input:
+        valid = check_is_scitype(obj, scitype=["Series", "Panel", "Hierarchical"])
+        if not valid:
+            raise ValueError("obj must be of Series, Panel, or Hierarchical scitype")
 
     obj = convert_to(obj, GET_CUTOFF_SUPPORTED_MTYPES)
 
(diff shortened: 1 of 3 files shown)


sktime/datatypes/_utilities.py
```diff
@@ -86,23 +108,43 @@ def get_cutoff(obj, cutoff=0, return_index=False):
...
+    check_input : bool, optional, default=False
+        whether to check input for validity, i.e., is it one of the scitypes
+    convert_input : bool, optional, default=True
+        whether to convert the input (True), or skip conversion (False)
+        if skipped, function assumes that inputs are one of GET_CUTOFF_SUPPORTED_MTYPES
 
     Returns
     -------
     cutoff_index : pandas compatible index element (if return_index=False)
         pd.Index of length 1 (if return_index=True)
```
khrapovs, line 132: Please, add `Raises` section

sktime/datatypes/_utilities.py
```diff
@@ -77,7 +78,28 @@ def get_index_for_series(obj, cutoff=0):
     return pd.RangeIndex(cutoff, cutoff + obj.shape[0])
 
 
-def get_cutoff(obj, cutoff=0, return_index=False):
+GET_CUTOFF_SUPPORTED_MTYPES = [
```
khrapovs, line 81:
```suggestion
CUTOFF_SUPPORTED_MTYPES = [
```
This is not a function, just a constant list. Why `GET_`?

sktime/datatypes/_utilities.py
```diff
@@ -210,7 +262,7 @@ def update_data(X, X_new=None):
         return X_new.combine_first(X)
 
 
-GET_LATEST_WINDOW_SUPPORTED_MTYPES = [
+GET_WINDOW_SUPPORTED_MTYPES = [
```
khrapovs, line 265:
```suggestion
WINDOW_SUPPORTED_MTYPES = [
```

sktime/datatypes/_utilities.py
```diff
@@ -219,7 +271,7 @@ def update_data(X, X_new=None):
 ]
 
 
-def get_window(obj, window_length=None, lag=0):
+def get_window(obj, window_length=None, lag=None):
```
khrapovs, line 274: please, update argument description in the docstring. specifically, default value

khrapovs (changes requested):
Again, I am missing simple unit tests that touch all new arguments in the `get_cutoff` function. Please, update `sktime.datatypes.tests.test_utils.test_get_cutoff`.

sktime/datatypes/_utilities.py
fkiraly: this refers to the function `get_cutoff`, i.e., the list of mtypes that is supported by the function `get_cutoff`

sktime/datatypes/_utilities.py
fkiraly: same - refers to the function `get_window`

sktime/datatypes/_utilities.py
khrapovs: if I read it like that, ok, accepted

sktime/datatypes/_utilities.py
khrapovs: ok

sktime/datatypes/_utilities.py
fkiraly: not in the scope of this PR, but sure

fkiraly: actually, I think the docstring was fine. The arg default is `None`, but this is interpreted as "0 of the right type". I added more explanation to the docstring to avoid user confusion. Is that ok?

fkiraly:
> Again, I am missing simple unit tests that touch all new arguments in the `get_cutoff` function. Please, update `sktime.datatypes.tests.test_utils.test_get_cutoff`.

Added test for `reverse_order` arg, is that what you meant?

sktime/datatypes/tests/test_utils.py
```diff
@@ -46,7 +48,9 @@ def test_get_cutoff(scitype, mtype, return_index):
         if fixture is None:
             continue
 
-        cutoff = get_cutoff(fixture, return_index=return_index)
+        cutoff = get_cutoff(
```
khrapovs, line 51: `check_input` and `convert_input` are not tested

sktime/datatypes/_utilities.py
```diff
@@ -230,14 +287,13 @@ def get_window(obj, window_length=None, lag=None):
...
-            pd.multiindex, numpy3D, nested_univ, df-list, of Panel scitype
-            pd_multiindex_hier, of Hierarchical scitype
+        if not None, must be of Series, Panel, or Hierarchical scitype
+        all mtypes are supported via conversion to internally supported types
+        to avoid conversions, pass data in one of GET_WINDOW_SUPPORTED_MTYPES
     window_length : int or timedelta, optional, default=-inf
         must be int if obj is int indexed, timedelta if datetime indexed
         length of the window to slice to. Default = window of infinite size
-    lag : int or timedelta, optional, default = 0
+    lag : int or timedelta, optional, default = 0 (correct type zero is inferred)
```
khrapovs, line 296: default for this argument is `None`, isn't it?

khrapovs (changes requested):

sktime/datatypes/tests/test_utils.py
fkiraly: how would you test them?

`convert_input` is just there to avoid unnecessary conversions in case the user already knows it is in one of the internal types, so the only way to test what it does would be to look into what lines are being executed?

`check_input` raises errors for incompatible input, so I could write a negative test.

sktime/datatypes/_utilities.py
khrapovs: I am a user. perhaps I am a dumb one, but 0 is not the same as `None`. As a user I would prefer to see the same default value in the docstring as I see in the function's signature. So, if possible, please, `default = None` as in many other functions in sktime.

fkiraly: ok, pushed a suggestion

khrapovs: my internal dumb user is happy now:D thanks!

sktime/datatypes/tests/test_utils.py
fkiraly: > Do all of these cases pass without errors? Do we get the same cutoff regardless?

The case where it makes a difference is if the object has a type not internally supported. If conversions are turned off, the function crashes (as the docstring tells you)

fkiraly:
added tests for the two remaining parameters to ensure expected behaviour

sktime/datatypes/tests/test_utils.py
fkiraly: done, please look

khrapovs: great! I am sure we will all be thankful for these tests later in the future :)

khrapovs (approved):

PR #2871, opened 2022-06-26
[ENH] `temporal_train_test_split` compatibility with all data formats and time difference input

fkiraly:
This PR rewrites `temporal_train_test_split` using `get_cutoff` and `get_window` from `sktime.datatypes._utilities`.

Feature/capability changes:
* `temporal_train_test_split` can now be applied to any sktime compatible data format. For panel and hierarchical data containers, the split is applied across a joint timeline, of the instances.
* `temporal_train_test_split` now accepts time difference as `train_size` and `test_size`, and splits accordingly

Relies on the upgraded `get_cutoff` from #2870.

FYI @khrapovs

### DIFF
--- a/sktime/datatypes/_utilities.py
+++ b/sktime/datatypes/_utilities.py
@@ -83,7 +83,7 @@ GET_CUTOFF_SUPPORTED_MTYPES = [
     "np.ndarray",
     "pd-multiindex",
     "numpy3D",
-    "nested_univ"
+    "nested_univ",
     "df-list",
     "pd_multiindex_hier",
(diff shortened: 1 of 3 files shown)


fkiraly:
converting to draft since the logic requires division of time offsets, which `pandas` does not support. The same logic will have to be replaced by sth equivalent.

khrapovs:
Besides "division of time offsets" this PR seems to be a combination of at least two others that I have already reviewed. Please, let me come back to this one once the other two are merged. Thanks.

PR #2881, opened 2022-06-27
[ENH] Pass frequency explicitly to `sktime.utils.datetime._shift` function

### LINKED ISSUE #1750: [ENH] address FutureWarning: Timestamp.freq is deprecated and will be removed in a future version

khrapovs:
#### Reference Issues/PRs

Partially addresses #1750 

On master function `_shift` extracts the frequency information from `pd.Timestamp` object which in future will cot contain it. Hence, here I explicitly pass frequency from the outside. This frequency is used to calculate offset for `pd.Timestamp` input, Frequency is inferred from `self._y` and stored in `self._freq` attribute of `sktime.forecasting.base._base.BaseForecaster` after `fit` is invoked.

No

##### For all contributions

### DIFF
--- a/sktime/forecasting/base/_base.py
+++ b/sktime/forecasting/base/_base.py
@@ -54,7 +54,7 @@ from sktime.datatypes import (
     update_data,
 )
 from sktime.forecasting.base import ForecastingHorizon
-from sktime.utils.datetime import _shift
+from sktime.utils.datetime import _shift, infer_freq
 from sktime.utils.validation._dependencies import _check_dl_dependencies
 from sktime.utils.validation.forecasting import check_alpha, check_cv, check_fh, check_X
 from sktime.utils.validation.series import check_equal_time_index
@@ -103,6 +103,7 @@ class BaseForecaster(BaseEstimator):
         # forecasting horizon
         self._fh = None
         self._cutoff = None  # reference point for relative fh
+        self._freq = None
 
         self._converter_store_y = dict()  # storage dictionary for in/output conversion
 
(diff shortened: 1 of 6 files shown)


khrapovs:
This PR has an overlap with https://github.com/alan-turing-institute/sktime/pull/2834 which should be merged first.

fkiraly (changes requested):
Thanks, nice work dealing with `_shift`!

Change requests:
* `self._freq` is never read in the `BaseForecaster`, so it should not be written in `BaseForecaster`
* from a user perspective, `_shift` should not require a `freq` argument. Why are we passing it? Can we not infer it internally?

fkiraly:
>from a user perspective, `_shift` should not require a `freq` argument. Why are we passing it? Can we not infer it internally?

Ok, I understand that we cannot do this for `pd.Timestamp` anymore.

May I request instead that `_shift` is extended so it can also deal with suitable `pandas.Index` elements? I.e., defaults to `pandas.Index.shift` if an index is passed?

This is building functionality towards indices that might be useful to have later.

fkiraly:
Tried the alternative approach here, to deal with `_shift`:
https://github.com/alan-turing-institute/sktime/pull/2909

`freq` is not explicitly passed, but passed as part of the `index` object `_cutoff`.
Might be a less invasive kind of surgery...

khrapovs:
> `freq` is not explicitly passed, but passed as part of the `index` object `_cutoff`. Might be a less invasive kind of surgery...

`_shift` is an internal function only used in a few places. Please, let's not waste time on preserving its signature by overcomplicated treatment of `cutoff`.

fkiraly:
>Please, let's not waste time on preserving its signature

It's not `_shift` I'm worried about but `BaseForecaster`.

khrapovs:
Was solved by #2932

PR #2885, opened 2022-06-28
[ENH] Pass frequency explicitly to `_make_fh` function

khrapovs:
#### Reference Issues/PRs

closes https://github.com/alan-turing-institute/sktime/issues/1750. There are no more future warnings on this branch.

Pass frequency explicitly to `_make_fh` function so that it is no longer extracted from `cutoff` argument. This argument may be `pd.Timestamp` and its `freq` attribute will be deprecated in the future.

##### For all contributions


khrapovs:
This PR has an overlap with https://github.com/alan-turing-institute/sktime/pull/2834 which should be merged first.

fkiraly:
Question: in the tests, should `fh` not be created without a `freq`?

Because the `fh` that the user makes will typically not have it.

Having the option to pass `freq` makes sense though - for this, would a `fh.freq = freq` at the end be sufficient?

khrapovs:
> Question: in the tests, should `fh` not be created without a `freq`?
> Because the `fh` that the user makes will typically not have it.

Whenever `fh` is converted to absolute/relative, there must be a `freq` attribute present. At the current state of `main` it can be either at construction or from a proper `cutoff`. Passing string at construction seems to me personally the most clear unambiguous and explicit way to do that. 

> Having the option to pass `freq` makes sense though - for this, would a `fh.freq = freq` at the end be sufficient?

This is 100% equivalent to `ForecastingHorizon(..., freq=freq)`.

fkiraly:
>> Having the option to pass `freq` makes sense though - for this, would a `fh.freq = freq` at the end be sufficient?

>This is 100% equivalent to `ForecastingHorizon(..., freq=freq)`.

Yes, indeed it is!

sktime/forecasting/base/tests/test_fh.py
```diff
@@ -80,7 +80,13 @@ def test_fh(index_type, fh_type, is_relative, steps):
     cutoff = y_train.index[-1]
 
     # generate fh
-    fh = _make_fh(cutoff, steps, fh_type, is_relative)
+    fh = _make_fh(
+        cutoff=cutoff,
+        steps=steps,
+        fh_type=fh_type,
+        is_relative=is_relative,
+        freq=infer_freq(y_train),
```
fkiraly, line 88: with the change, line 91 is redundant!

sktime/forecasting/model_selection/tests/test_split.py
```diff
@@ -462,7 +462,13 @@ def test_split_by_fh(index_type, fh_type, is_relative, values):
         # )
     y = _make_series(20, index_type=index_type)
     cutoff = y.index[10]
-    fh = _make_fh(cutoff, values, fh_type, is_relative)
+    fh = _make_fh(
+        cutoff=cutoff,
+        steps=values,
+        fh_type=fh_type,
+        is_relative=is_relative,
+        freq=infer_freq(y),
```
fkiraly, line 470: I think we want this test to be "stronger", i.e., to work without `freq` pased to `_make_fh`

fkiraly (changes requested):
The general theme behind my change request is:

If `freq` is not passed to `_make_fh`, the initial `fh` knows less, and the test is hence "stronger", because it also tests that `fh` "learns" the `freq` from the objects it interacts with.

I think we should always carry out the "stronger" version of the test, since that is typically what the user will do, in a case where it will not trigger the deprecated functionality internally.

These are, I think the following tests:
* `test_predict_time_index`
* `test_predict_time_index_with_X`
* `test_predict_time_index_in_sample_full`

(the other ones do not matter since `fh` is not modelling "user constructed" behaviour)

khrapovs:
Resolved by https://github.com/alan-turing-institute/sktime/pull/2965

PR #2896, opened 2022-06-29
[MNT] cleaning up CI workflow

lmmentel:
This PR simplifies the CI by consolidating the test workflow since there are no longer any platform dependent differences between mac and linux, i.e. merging mac/linux via a matrix.

##### For all contributions

##### For new estimators

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,7 +35,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
-          python-version: 3.7
+          python-version: 3.9
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
@@ -50,6 +50,7 @@ jobs:
     strategy:
       matrix:
         python-version: [3.7, 3.8, 3.9, '3.10']
+
     steps:
       - uses: actions/checkout@v2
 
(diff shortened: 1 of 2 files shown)


fkiraly (changes requested):
Looks great!

Quick question before we merge: have you checked whether we can also subsume windows under this, now that `pip install` also works on windows?

@RNKuhns has managed to do it, here!
https://github.com/sktime/BaseObject/blob/main/.github/workflows/test.yml

lmmentel:
> Quick question before we merge: have you checked whether we can also subsume windows under this, now that `pip install` also works on windows?
> 
> @RNKuhns has managed to do it, here! https://github.com/sktime/BaseObject/blob/main/.github/workflows/test.yml

That would  increase the scope, but we can improve it further in a following PR.

fkiraly (approved):
>That would increase the scope, but we can improve it further in a following PR.

Makes sense.

fkiraly:
stupid questions:

* can you merge this, since the test names have changed?
* I supposed once we merge, we need to change the list of required tests in the admin interface for the same reason?

lmmentel:
> 
>     * can you merge this, since the test names have changed?
> 
Nope, I cannot merge the button is not active,

>     * I supposed once we merge, we need to change the list of required tests in the admin interface for the same reason?

Probably, but I cannot be sure since I don't have access.

fkiraly:
>Probably, but I cannot be sure since I don't have access.

You'd need admin rights which only CC members have, I'll take care of that.

PR #2903, opened 2022-06-30
[ENH] `FittedForecaster` for using serialized/fitted forecasters

fkiraly:
Draft PR containing a design study for use of serialized forecasters, e.g., pickled or fitted neural networks.

The `FittedForecaster` wraps an already fitted forecaster pre-trained on some other data, and exposes it under the `sktime` interface.
`fit` needs to be called again, but that does not change the original pre-trainedmodel. Instead, depending on parameter settings, it leaves the model unchanged, or updates it.

### DIFF
--- a/sktime/pipeline/pipeline.py
+++ b/sktime/pipeline/pipeline.py
@@ -165,7 +165,7 @@ class Pipeline(BaseEstimator):
             "such as TransformedTargetForecaster, ForecastingPipeline, "
             "ClassificationPipeline, etc., see for instance "
             "notebooks 01_forecasting.ipynb and "
-            "02_classification.ipynb at"
+            "02_classification.ipynb at "
             "https://github.com/sktime/sktime/blob/main/examples/.",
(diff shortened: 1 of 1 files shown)


jobs-git:
Any updates on this? is this same as refit for new data?

fkiraly:
> Any updates on this? is this same as refit for new data?

No, it is same as not refit as all - in "fit" it just loads an already fitted model.

jobs-git:
> > Any updates on this? is this same as refit for new data?
> 
> No, it is same as not refit as all - in "fit" it just loads an already fitted model.
> 
> Would you like to use it?

I am thinking of like if there would be a really huge data that could not be placed in vram, so we have to batch it, or in case there where unintended interruption so we can resume on checkpoint.

fkiraly:
This is 1736 commits behind - I will update it and see whether it can be merged.

We should probably rename to `FrozenForecaster`, after `sklearn` introduced something similar last month.

PR #2908, opened 2022-07-02
[ENH] preserve `index.freq` in `get_cutoff`

fkiraly:
This PR ensures that the `freq` attribute is preserved if `get_cutoff` returns an index.

Also simplifies the `return_index` case distinctions by refactoring it into a function `sub_index`, and fixes an unreported bug for `pd-multiindex` where the returned index would always be the last, even if `reverse_order=True`.

### DIFF
--- a/sktime/datatypes/_utilities.py
+++ b/sktime/datatypes/_utilities.py
@@ -152,6 +152,18 @@ def get_cutoff(
 
     if cutoff is None:
         cutoff = 0
+    elif isinstance(cutoff, pd.Index):
+        if not len(cutoff) == 1:
+            raise ValueError(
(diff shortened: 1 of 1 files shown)


khrapovs (changes requested):
Please, add unit tests. Especially since you are claiming that this PR fixes some unreported bugs. These tests should fail on master. Thanks.

fkiraly:
> Please, add unit tests. Especially since you are claiming that this PR fixes some unreported bugs. These tests should fail on master. Thanks.

Here you go, some freshly cooked tests coming up 😄

khrapovs (approved):

PR #2909, opened 2022-07-02
[ENH] turn private cutoff of forecasters into an index that carries `freq`

fkiraly:
This PR is an attemted partial fix to the `freq` issue, by ensuring the private `_cutoff` attribute of forecasters is always a `pandas.Index` that carries `freq` (when `freq` is present or inferrable).

Relies on PR https://github.com/alan-turing-institute/sktime/pull/2908 which ensures that `get_cutoff` extracts and preserves `freq`.

The public `cutoff` is unchanged, a `pandas` compatible index element rather than an index, which means that no deprecation is necessary.

### DIFF
--- a/sktime/utils/datetime.py
+++ b/sktime/utils/datetime.py
@@ -110,8 +110,6 @@ def _shift(x, by=1, return_index=False):
         warn("_shift no longer supports x: pd.Timestamp fom 0.14.0", DeprecationWarning)
         if not hasattr(x, "freq") or x.freq is None:
             raise ValueError("No `freq` information available")
-        by *= x.freq
-        return x + by
         # raise TypeError("_shift no longer supports x: pd.Timestamp")
 
     # we ensure idx is pd.Index, x is first (and usually only) element
--- a/sktime/utils/datetime.py
+++ b/sktime/utils/datetime.py
@@ -114,7 +114,6 @@ def _shift(x, by=1, return_index=False):
         return x + by
         # raise TypeError("_shift no longer supports x: pd.Timestamp")
 
-
     # we ensure idx is pd.Index, x is first (and usually only) element
     if isinstance(x, pd.Index):
         idx = x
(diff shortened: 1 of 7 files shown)


fkiraly:
Credit to @khrapovs for identifying the places where `_shift` is called on `self.cutoff`, in https://github.com/alan-turing-institute/sktime/pull/2881.

sktime/utils/datetime.py
```diff
@@ -88,26 +89,54 @@ def _get_freq(x):
...
+            return idx + by
+        else:
+            return idx.shift(by)
+
+    assert isinstance(x, (pd.Period, pd.Timestamp, int, np.integer)), type(x)
+    assert isinstance(by, (int, np.integer)) or is_integer_index(by), type(by)
+
+    # if we want index element, we can still use add, except if timestamp
+    # deprecate this in 0.13.0 and remove in 0.14.0
+    if isinstance(x, pd.Timestamp):
```
khrapovs, line 136: This block looks like a duplicate of lines 109-114... Is it intended?

khrapovs (changes requested):

sktime/utils/datetime.py
fkiraly: yes, that was a duplication - removed it

fkiraly: yes, indeed. I looked into it again, and in fact it is *not* a duplication, it just looks like one.

Why: the initial `x` can be index, or index element.
The final `x` is always an index element.

Logic-wise, the cases the two pieces of code treat are different:
* the initial `x` was a `pd.Timestamp`. In this case, we will not have a `freq` after pandas deprecation, and a warning (later exception) should be raised. This is treated in the first paragraph, and will be deprecated once we move away from `freq` in `pd.Timestamp`.
* the initial `x` was a time-like index, and its element is `pd.Timestamp`. In this case, we have a `freq` right now, and after deprecation. This is treated in the second paragraph, and should not be deprecated.

So, the logic was actually not redundant, but the exception raised the second time was (it would never be reached), and the deprecation action was incorrect (second paragraph should not be deprecated).

Hope that makes sense?

khrapovs (approved):

PR #2912, opened 2022-07-02
[MNT] delete duplicate classifier tests

fkiraly:
This PR cleans up some stray tests in the classifier package:

* deletes duplcate tests from sub-folders of the classifier module.
The tests duplicate the tests on unit test and basic motions in `TestAllClassifiers`.
* moves repeated tests on `_get_train_proba` into `TestAllClassifiers`


MatthewMiddlehurst:
I may have missed this addition, but unless the general tests have changed to include running `_get_train_probs` then these are not duplicates.

fkiraly:
I see - would you be so kind then to add that to `TestAllClassifiers`?
Easy to write, just do `if hasattr(estimator, "_get_train_probs")`, no?

MatthewMiddlehurst (approved):

fkiraly (commented):
also happy with this

PR #2920, opened 2022-07-03
[ENH] soft dependency testing

fkiraly:
This PR moves inactive legacy code on soft dependency testing for the no longer in use Azure CI/CD to the `test_softdeps` test action on GitHub actions.
Original location is in the Azure build tools, `build_tools.azure.check_soft_dependencies`.
When this is moved, the legacy Azure content can probably be removed without a loss (see Pr https://github.com/alan-turing-institute/sktime/pull/2917 that does this)

The test code is adapted and based on the package dependency tags in PR https://github.com/alan-turing-institute/sktime/pull/2915.

Tests being carried out:

* `_check_pyhon_version` error is raised when trying to construct an estimator requiring a python version incompatible with spces (according to tags)
* `_check_soft_dependencies` error is raised when trying to import a module requiring a soft dependency that is not installed
* `_check_soft_dependencies` error is raised when trying to construct an estimator requiring a soft dependency that is not installed
* no `ModuleImportError`-s are encountered when trying to construct estimators without soft dependencies and compatible python version (according to tags)
* no `ModuleImportError`-s are encountered when trying to construct estimators where python version is compatible and soft dependencies (according to tag) are present in the environment


fkiraly:
for reviewers: it looks like the failures are all legit failures of current estimators not raising the right errors.

PR #2925, opened 2022-07-04
[ENH] `predict_interval` capability for `VECM`

### LINKED ISSUE #929: [ENH] VAR and VECM models

AurumnPegasus:
#### Reference Issues/PRs

Fixes #929 

Implements predict interval for VECM

##### For all contributions


AurumnPegasus:
Hey @fkiraly ,
In continuation to our discussing in [my previosu PR](https://github.com/alan-turing-institute/sktime/pull/2829) related to VECM, I have completed implementation of predict intervals.
Just a minor hiccup:
While running the test, the code passes tests for `predict_interval` but fails for `predict_quantiles`. On further investigation it is because of 
```
raise ValueError("Lengths must match to compare")
ValueError: Lengths must match to compare
```

I compared it with ARIMA output and found that for
```
>>> y_pred_int = forecaster.predict_interval(coverage=0.9)
>>> y_pred_int
           Coverage            
                0.9            
              lower       upper
1961-01  371.535093  481.554608
1961-02  344.853205  497.712761
1961-03  324.223995  508.191104
>>> yp = forecaster.predict_quantiles(alpha=[0.05, 0.95])
>>> yp
          Quantiles            
               0.05        0.95
1961-01  371.535093  481.554608
1961-02  344.853205  497.712761
1961-03  324.223995  508.191104
```

As you can see, the indices are different for different functions. I am sure that `predict_intervals` implementation in my code is giving the correct output format, since it passes all the tests (and looks the same as well). 

I felt this to be inconsistent since the [extension template](https://github.com/alan-turing-institute/sktime/blob/main/extension_templates/forecasting.py) says that 

```
    # implement one of _predict_interval or _predict_quantiles (above), or delete both
    #
    # if implementing _predict_quantiles, delete _predict_interval
    # if not implementing either, delete both methods
```

An easy solution to this would be to implement `predict_quantiles` as well, and just combine the two columns of it to give us `predict_interval`'s output. Just wanted to confirm if it is not a silly mistake on my end before proceeding.

fkiraly:
Hm, it sounds like there might be a bug with the default implementation of `predict_quantiles`.

Can you kindly resolve merge conflicts so the tests run and we see exactly what is happening and we can debug?

AurumnPegasus:
Hey @fkiraly,
can you kindly direct me to some multivariate forecasters in sktime which have implemented predict intervals? Currently my lead to solve this problem is that `predict_quantiles` in the base class is not handling multivariate properly but I wanted to check with other estimators to see if that is the case or if I am making some mistake

fkiraly:
> can you kindly direct me to some multivariate forecasters in sktime which have implemented predict intervals?

I don't think there are any implementing this natively, this all works via vectorization currently.

> Currently my lead to solve this problem is that `predict_quantiles` in the base class is not handling multivariate properly but I wanted to check with other estimators to see if that is the case or if I am making some mistake

I think your suspicion is very plausible.

As you point out, none of the current estimators seem to implement it natively, so the sequence of calls would be (1) univariate default implementation of `predict_quantiles`, (2) vectorization. So, it could well be that the default implementation of `predict_quantiles` based on `predict_interval` is incorrect.

What would help if you could write up some code that isolates a call and produces input/output in one case where you think it is wrong. Actual output, and expected output, and where they do not agree.

fkiraly:
Related PR from me a while ago, I can't remember anymore why I stopped working on this:

https://github.com/alan-turing-institute/sktime/pull/2321

fkiraly:
Are you still working on this, @AurumnPegasus?

I think you wanted to debug?

AurumnPegasus:
Hey,
Sorry for the late reply, I will check it out. Thanks a lot

Issue #2890, opened 2022-06-28
[MNT] Cut down test time to under 10 min


lmmentel:
**Describe the bug**

The current testing suite is horrendously long and creates a poor developer experience with the test suite taking ~1h to complete.

fkiraly:
here is an earlier attempt
https://github.com/alan-turing-institute/sktime/pull/2024

but I do not understand why this did not lead to the expected gains. Sth might not work with the caching scope?

khrapovs:
Btw, who is paying for all this computational time on GitHub?.. It's not just about developer experience, is it?

fkiraly:
so, I just notice that in the #2024 attempt, the `lru_cache`-d function is inside a class, which won't work properly.

That might be it?

I.e., the solution would be:

* take #2024 and adapt it into the current test framework, resolving perhaps a million merge conflicts
* take the cached function `_cached_estimator_fitting`, and move it *outside* any of the classes

... done?

miraep8:
I this this is really great. Some ideas to throw out there:

- what would people's thoughts be on just running CI/CD tests on files changed until a particular PR is ready to merge, and then on the merge action we trigger the full CI/CD run? Might help with increasing iteration speed (seems like most bugs come up on files directly related to the PR) but ensures that a full QC check is done before the branch is merged. 
- also with regards to caching and how to store/access the file - it is actually possible to do this via [Github Actions directly](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows).  Not sure if that helps with this particular discussion, but seems like it might.

fkiraly:
>what would people's thoughts be on just running CI/CD tests on files changed until a particular PR is ready to merge, and then on the merge action we trigger the full CI/CD run?

That's a brilliant idea, and basically what `pytest-testmon` does, this avenue was explored in issue https://github.com/alan-turing-institute/sktime/issues/2530.

I think we got stuck at the question "but how would you know which tests are those that touch the changed files/lines"? There is also a problem with `testmon` that it does not detect files that the `all_estimator` crawler does, say, if an entire estimator is added.

I have to apologize to @tarpas who put work into this, we got stuck and I was thinking about how to ensure that really all classes get tested, for weeks, until the point I eventually forgot by accident that the discussion was still ongoing, sorry @tarpas!
(should have tracked that on one of the boards)

fkiraly:
[...]

PR #2928, opened 2022-07-04
[MNT] removal of pytest pyargs in CI/CD

fkiraly:
This PR removes use of `pyargs` from `pytest`.

According to @tarpas, --pyargs are no longer needed since we no longer have cython, and --pyargs stand in the way of the `testmon` approach to #2890

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ test: ## Run unit tests
 	mkdir -p ${TEST_DIR}
 	cp .coveragerc ${TEST_DIR}
 	cp setup.cfg ${TEST_DIR}
-	cd ${TEST_DIR}; python -m pytest
+	python -m pytest
 
 test_softdeps: ## Run unit tests
 	-rm -rf ${TEST_DIR}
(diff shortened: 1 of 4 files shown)


fkiraly:
Well, that didn't take long to break

fkiraly:
according to @tarpas, --pyargs are no longer needed since we no longer have cython, --pyargs stand in the way of the `testmon` approach to #2890

Makefile
```diff
@@ -28,7 +28,7 @@ test: ## Run unit tests
 	mkdir -p ${TEST_DIR}
 	cp .coveragerc ${TEST_DIR}
 	cp setup.cfg ${TEST_DIR}
-	cd ${TEST_DIR}; python -m pytest
```
aiwalter, line 31: what about the other `pyargs` in this file? Also there are more flags, we could remove them as well. I might have forgotten them last time when I consolidated them or someone has added them recently?

fkiraly: I don't know - tbh, I was just hacking around semi-blindly until it worked.

Can you explain to me what specifically you mean with "what about"?

If you mean the `nosoftdeps` test, we can leave `pyargs` in there since it runs only a small number of specific tests which we probably always want, they take only a few minutes.

aiwalter (approved):

PR #2931, opened 2022-07-05
[ENH] test that `set_params` recognizes unique suffixes as aliases for full parameter string

fkiraly:
Update: The logic in this PR became obsolete after merging it upstreak in `scikit-base 0.6.1`, see
https://github.com/sktime/skbase/pull/229

The PR still adds the tests in `sktime`, with a `scikit-base` version condition, but does not add the logic anymore.

---

This PR introduces experimental functionality to `set_params` to enable it to recognize unique suffixes as aliases for full parameter strings.

That is, in the example in https://github.com/alan-turing-institute/sktime/discussions/2929, it enables `set_params` - and, in consequence, the `param_grid` of grid search - to recognize `max_depth` instead of `forecaster__forecast__estimator__max_depth`. 

This is primarily meant as convenience sugar for the user of grid search or, more generally, `set_params`.

Contains tests for both nested parameter setting behaviour, as well as parameter key aliasing in nested composites.


fkiraly:
hm, maybe it is best if this change is made only in `scikit-base`, that way there is no need for deprecation handling in `sktime`.

fkiraly:
Updated this PR as the change is now in `scikit-base 0.6.1`.

PR #2932, opened 2022-07-05
& khrapovs [ENH] addressing `freq` deprecation in `ForecastingHorizon`

fkiraly:
This addresses the `freq` deprecation in `ForecastingHorizon`.

Continuation/modification of #2834 which aims to keep public interfaces of `ForecastingHorizon` intact. Changes to the public interface have been reverted.

The idea is to streamline the flow of the `freq` information from the data in `BaseForecaster` to the methods of `ForecastingHorizon`, as follows:

* `BaseForecaster._cutoff` is ensured to be `pd.Index`, whenever data has been seen, and with `freq` whenever that information is available. This is ensured through the `cutoff` setters.
* `fh`, as soon as passed to `BaseForecaster`, acquires the `freq` information from `BaseForecaster`, into the `_freq` attribute, as a `pandas` offset. This is ensured through `_check_fh` which calls `check_fh`. Since `fh` is always passed together or later than the data, this ensures that the internal `fh` of any forecaster always has the `freq` information.
* finally, it is ensured that `ForecastingHorizon` methods, which require frequency strings, not offsets, get this from `_freq`, via the `freq` attribute reader which extracts the `freqstr`. This replaces `_get_freq` and similar.
* for robustness, `ForecastingHorizon` can acquire `freq` also from its arguments, and will raise an error if the `freq` from `BaseForecaster` is discrepant.


fkiraly:
@khrapovs, I think this is almost ready and already good enough to get the "idea".

The tests that are failing are, I think, because I put a hard exception into the coercion routine so it dredges up cases where timestamps are passed and freq is unavailable.

sktime/forecasting/base/tests/test_fh.py
```diff
@@ -323,10 +334,11 @@ def test_absolute_to_absolute_with_integer_horizon(freqstr):
     """Test converting between absolute and relative with integer horizon."""
     # Converts from absolute to relative and back to absolute
     train = pd.Series(1, index=pd.date_range("2021-10-06", freq=freqstr, periods=3))
+    cutoff = get_cutoff(train, return_index=True)
```
khrapovs, line 337: That's what I was afraid of. You are forcing a user to use `get_cutoff` function. You could also force a user to use `pd.period_range` above. Or force him/her to pass `freq`? Why is this choice better?

khrapovs (changes requested):

sktime/forecasting/base/tests/test_fh.py
fkiraly: Two arguments:
1. this is not actually a user will have to do, typically.

If they *specify* the forecasting horizon, they will either do one of:
* direct construction, absolute timestamps -> has freq
* direct construction, relatie timestamps -> has freq
* direct construction, integers -> does not have freq, but will be interpreted by forecaster

and not call one of the conversion methods.

2. passing `freq` instead is also possible, as per your interface.

fkiraly: we could also test for both, i.e., passing `cutoff` and `freq`?

khrapovs: I hope you do not mean `pd.Timestamp` when you say "timestamps" ;)

fkiraly: > I hope you do not mean `pd.Timestamp` when you say "timestamps" ;)

I fear that would be a user expectation...

fkiraly: > Sure, the more tests the merrier :D

I added a test where `freq` is passed to `ForecastingHorizon`

sktime/forecasting/base/_fh.py
```diff
@@ -138,6 +139,56 @@ def _check_values(values: Union[VALID_FORECASTING_HORIZON_TYPES]) -> pd.Index:
...
+        return obj
+    elif hasattr(obj, "_cutoff"):
+        return _check_freq(obj._cutoff)
+    elif isinstance(obj, (pd.Period, pd.Index)):
+        return _extract_freq_from_cutoff(obj)
+    elif isinstance(obj, str) or obj is None:
+        return to_offset(obj)
+    else:
+        return None
+    # else:
```
khrapovs, line 167: please, don't forget to remove commented lines

sktime/forecasting/base/_fh.py
```diff
@@ -138,6 +139,56 @@ def _check_values(values: Union[VALID_FORECASTING_HORIZON_TYPES]) -> pd.Index:
     return values.sort_values()
 
 
+def _check_freq(obj):
+    """Coerce obj to a pandas frequency offset for the ForecastingHorizon.
+
+    Parameters
+    ----------
+    obj : pd.Index, pd.Period, pd offset, or None
```
khrapovs, line 147:
```suggestion
    obj : pd.Index, pd.Period, pandas offset, or None
```

sktime/forecasting/base/_fh.py
```diff
@@ -138,6 +139,56 @@ def _check_values(values: Union[VALID_FORECASTING_HORIZON_TYPES]) -> pd.Index:
...
+    #         f" but found freq of type {type(obj)}"
+    #     )
+
+
+def _extract_freq_from_cutoff(x) -> Optional[str]:
+    """Extract frequency string from cutoff.
+
+    Parameters
+    ----------
+    x : pd.Period
```
khrapovs, line 180:
```suggestion
    x : pd.Period, pd.PeriodIndex, pd.DatetimeIndex
```

sktime/forecasting/base/_fh.py
```diff
@@ -153,12 +204,16 @@ class ForecastingHorizon:
         - if None, the flag is determined automatically:
             relative, if values are of supported relative index type
             absolute, if not relative and values of supported absolute index type
+    freq : str, pd.Index, pandas offset, or BaseForecaster, optional (default=None)
```
khrapovs, line 207: what do you mean `BaseForecaster`? is it a mistake?

sktime/forecasting/base/_fh.py
```diff
@@ -262,36 +386,48 @@ def to_numpy(self, **kwargs) -> np.ndarray:
         """
         return self.to_pandas().to_numpy(**kwargs)
 
+    def _coerce_cutoff_to_index_element(self, cutoff):
+        """Coerces cutoff to index element, and updates self.freq with cutoff."""
+        self.freq = cutoff
+        if isinstance(cutoff, pd.Index):
+            assert len(cutoff) > 0
+            cutoff = cutoff[-1]
```
khrapovs, line 394: I guess we will see the new function here from https://github.com/alan-turing-institute/sktime/pull/2939, right?

sktime/forecasting/base/tests/test_fh.py
khrapovs: > I fear that would be a user expectation...

Exactly. And in this case we should push a user to other means of passing frequency information.

khrapovs (changes requested):

sktime/forecasting/base/_fh.py
fkiraly: meant that it can be a forecaster, in which case the `freq` will be obtained from the `cutoff` stored. Clarified the docstring

sktime/forecasting/base/_fh.py
fkiraly: hm, we could use it, but then again this is only 3 lines?

khrapovs (approved):
Looks good. Unfortunately, this is not the whole solution to the main issue. The last bit I think is this PR: https://github.com/alan-turing-institute/sktime/pull/2885. I will clean it up and open for review once this one is merged to main. Thanks a lot!

PR #2940, opened 2022-07-06
[DOC] `ReconcilerForecaster` and hierarchical transformers update

ciaran-g:
#### Reference Issues/PRs

Some missing doc info from #2830 and the hierarchical transformers

- Added docstring examples for the `ReconcilerForecaster`, `Aggregator`, and `Reconciler`.
- Added the forecaster to the `docs/source/api_reference/forecasting.rst` file

No


fkiraly (approved):
Thanks!

Moved the location in the API reference (this is not a pipeline element), but happy with it apart from that

PR #2954, opened 2022-07-07
[ENH] "dont refit or update" option in `evaluate`

fkiraly:
This adds a third strategy, `no-update_params`, to the `evaluate` utility, which does not change the forecaster model after the first `fit`. See discussion https://github.com/alan-turing-institute/sktime/discussions/2938.


sktime/forecasting/model_evaluation/_functions.py
```diff
@@ -42,9 +42,11 @@ def evaluate(
...
     X : pd.DataFrame, default=None
         Exogenous variables
-    strategy : {"refit", "update"}
-        Must be "refit" or "update". The strategy defines whether the `forecaster` is
-        only fitted on the first train window data and then updated, or always refitted.
+    strategy : {"refit", "update", "no-update"}, optional, default="refit"
+        defines the ingestion mode when the forecaster sees new data when window expands
+        "refit" = forecaster is refitted to each training window
+        "update" = forecaster is updated with training window data, in sequence provided
+        "no-update" = fit to first training window, and re-used without fit or update
```
aiwalter, line 49: I dont get what `no-update` is going or not doing? Cant see where this is used in the code?

fkiraly: `no-update` just skips any internal `update`

aiwalter (approved):

PR #2965, opened 2022-07-10
[ENH] Avoid accessing `.freq` from `pd.Timestamp` by converting `cutoff` to `pd.Index`

### LINKED ISSUE #1750: [ENH] address FutureWarning: Timestamp.freq is deprecated and will be removed in a future version

khrapovs:
#### Reference Issues/PRs

closes #1750 

Pass `cutoff` as `pd.Index` object to `_make_fh` which avoids accessing `.freq` attribute of `pd.Timestamp`. This is an alternative solution to #2885.

##### For all contributions


fkiraly (approved):
Little bit of a hack since we need to subset `y` first, but I prefer this above #2885 for the tests.
Happy for this to be merged, since it avoids the `freq` access.

Should we still allow `freq` to be passed to `_make_fh` for convenience, and merge this into #2885?

khrapovs:
Thanks for approving this so fast! I can see the light in the end of this tunnel :D

> Should we still allow `freq` to be passed to `_make_fh` for convenience, and merge this into #2885?

Sorry, but now that I have removed all necessity to pass freq explicitly to `_make_fh`, I do not see why we should change it. If there was at least a single use case in the code for that, I would be happy to do it.

fkiraly:
>If there was at least a single use case in the code for that, I would be happy to do it.

Well, I thought there might be, and you have looked into it, so I'm happy either way what you think is best.

PR #2972, opened 2022-07-10
[ENH] add check for forecast to have correct columns

fkiraly:
This adds a test that forecast `predict` output has the same column names as the training data, if both have column names.

Surprisingly, this was not yet tested.

Relies on PRs with fixes of bugs that would make this fail:
* `Prophet.predict` https://github.com/alan-turing-institute/sktime/pull/2973


aiwalter:
Should this also je testted for transformers in `transform` and `inverse_transform`?

fkiraly:
> Should this also je testted for transformers in `transform` and `inverse_transform`?

Not all transformers are 1:1 in terms of columns, or column names.

The counterpart for transformers are these:
* uniqueness (weaker but catches the most common problem) https://github.com/alan-turing-institute/sktime/pull/2969
* automated naming https://github.com/alan-turing-institute/sktime/pull/2958

aiwalter:
But we could test that after `transform` _and_ `inverse_transform` we  have the same columns?

fkiraly:
> But we could test that after `transform` _and_ `inverse_transform` we  have the same columns?

Yes, good idea.
Naively, I thought this was already tested, but indeed, `test_transform_inverse_transform_equivalent` does not seem to check for that.
I fear there might be pandora's box opening if we add the check to the test ... perhaps we may like to do it in a separate PR...

fkiraly:
should we merge this and #2973, @aiwalter?
We can approximate step by step the desired end state of stronger output checks across estimator types.

PR #2975, opened 2022-07-11
[ENH] Implement get_fitted_params for tuning forecasters

### LINKED ISSUE #1496: [ENH] implement get_fitted_params for tuning forecasters
Tuning forecasters like `ForecastingGridSearchCV` should implement a `get_fitted_params` method.
Note that here, by applying a tuning method, the hyperparameters that are tuned become "normal" (fitted) parameters.

The `get_fitted_params` of tuning should return two things in its dictionary:

* the hyper-parameters that are tuned, with their tuned values
* either: the "best estimator" in its fitted state; or, better, parameters of the "best estimator" with a nesting structure like for `sklearn` `get_params`  (this might be slightly fiddly.

For implementation, it's fine to start with one of the two.

See also discussion in #1486, FYI @boukepostma and @asksteph.

FYI @aiwalter.

ZiyaoWei:
#### Reference Issues/PRs

Fixes #1496.

This implements `get_fitted_params` for `ForecastingGridSearchCV` and `ForecastingRandomizedSearchCV`.

No.

##### For all contributions


ZiyaoWei:
A few comments:

- This only implements the "bare minimum" in that it just returns the hyper params and the best model - I did that since even if the best model doesn't support `get_fitted_params` I feel this should still return something useful.
- I updates the tests to use `tuner` instead of `gscv` since there are now two tuners.
- Not sure if we should just update the default implementation in `BaseGridSearch`?
- Since we are changing what this method returns (`BaseGridSearch` has an implementation where the `get_fitted_params` of the best model is returned), should we add a release note so that if users were getting the params directly out of the return value, they now should get it from the best model or the best hyperparameters instead? Or there could be a better implementation of this, any suggestions are welcome.

Thanks!

sktime/forecasting/model_selection/tests/test_tune.py
```diff
@@ -37,6 +38,12 @@
 TEST_METRICS = [MeanAbsolutePercentageError(symmetric=True), MeanSquaredError()]
 
 
+def _check_no_fitted_params_before_fitting(tuner):
```
fkiraly, line 41: This is already tested by `test_raises_not_fitted_error`, duplicate, kindly remove

sktime/forecasting/model_selection/_tune.py
```diff
@@ -526,6 +541,21 @@ def __init__(
         self.n_iter = n_iter
         self.random_state = random_state
 
+    def get_fitted_params(self):
```
fkiraly, line 544: Nice! Both estimators have this though, we should move this up to the common parent class, `BaseGridSearch`.

fkiraly:
> Since we are changing what this method returns (`BaseGridSearch` has an implementation where the `get_fitted_params` of the best model is returned), should we add a release note so that if users were getting the params directly out of the return value, they now should get it from the best model or the best hyperparameters instead? Or there could be a better implementation of this, any suggestions are welcome.

Hm, we could do both, by taking the union of keys and contents? With a clearly defined and documented precendence role in case there are any clashes.

ZiyaoWei:
Now it's failing for `ForecasterFitPredictMultivariateNoX`, for some reason IIUC the outer `ForecastingGridSearchCV` doesn't have it's best_forecaster_ populated.

fkiraly:
>Now it's failing for `ForecasterFitPredictMultivariateNoX`, for some reason IIUC the outer `ForecastingGridSearchCV` doesn't have it's best_forecaster_ populated.

Ah...
That is coming from the vectorization, not your fault.

The problem is with a new feature, where you can pass multivariate data to all forecasters, and it will just apply the forecaster to all columns. The individual forecasters (one per column) are stored in the `forecasters_` parameter, in that case, which in turn have individual parameters.

I think we need to think about how to make this more principled.

For the time being, you should be fine with adding the following line:
```python
if self._is_vectorized:
    return {"forecasters_": self.forecasters_}
```
which will return a data frame containing all the forecasters.

The way to make this more principles should have something to do with nesting similar to `get_params`, but that is another story, and I would suggest going with the workaround for now (and maybe you would like to help later with nested fitted param access 😄 )

ZiyaoWei:
@fkiraly - Ah, not familiar with vectorization in this context at all but that makes sense, thanks!

PR #2884, opened 2022-06-27
[WIP] [ENH] Reimplement benchmarking with `kotsu`

DBCerigo:
#### To do:
- [ ] Generally, shift use case from "benchmarking all sktime estimators" to "enable a user to benchmark something they care about"
- [ ] Spike on implementing new interface that wraps kotsu and only relies on sktime objects, for easy user life
- [ ] Implement storing folds and predictions
- [ ] Add dependencies 
- [ ] Move nb to https://github.com/alan-turing-institute/sktime/tree/main/examples
- [ ] Expand nb to include examples, to guide a user on its usage (?)

#### Usage instructions:
- Dependencies to install specific for this PR: `kotsu`, `jupyter`, `autoreload` `typing_extensions`
- Start a jupyter server and run the `sktime/benchmarking/run.ipynb`

#### Reference Issues/PRs
See discussion: https://github.com/alan-turing-institute/sktime/issues/2777
See discussion: https://github.com/alan-turing-institute/sktime/issues/2804
See spike: https://github.com/alan-turing-institute/sktime/pull/2805

#### What does this implement/fix? Explain your changes.
Implement benchmarking using `kotsu`. 
- Instantiation of a `validation_registry` and `estimator_registry` in `benchmarking`'s `init`
- A `validations.py` module which defines a paramaterised validation run, which takes a `dataset_loader: Callable`, a `cv_splitter: BaseSplitter`, and a list of `scorers: List[BaseMetric]`. The module registers an example specification.
- An `estimators.py` module which imports and registers estimators + their configs (i.e. param specifications)
- A `run.ipynb` which runs the registry of estimators through the validations, storing the benchmark results

Likely we will eventually want to have separate registries corresponding to different types of predictions (classification, regression, etc.), for now this just implements for regression (point prediction).  

#### Does your contribution introduce a new dependency? If yes, which one?
- `kotsu` - soft dependency
- TBC: `jupyter` - soft dependency - depending on discussions outcome

#### What should a reviewer concentrate their feedback on?

The understandability, flexibility, extendability, and maintainability, of this benchmarking system implementation.

#### To discuss at Tuesday 28th June meeting:
("we" refers to @DBCerigo @alex-hh @ali-tny) 
- [x] What [type of dependency](https://www.sktime.org/en/latest/developer_guide/dependencies.html#types-of-dependencies) will `kotsu` be? We think `soft`.
    - Answer: TBD, but maybe a core dep given benchmarking is intended to be a core functionality. 
[...]

PR #2805, opened 2022-06-13
[ENH] Kotsu application to sktime

TNTran92:
#### Reference Issues/PRs

Solution to Issue #2804. See also Issue #2777

#### What does this implement/fix? Explain your changes.

Demonstration in notebook showing the kotsu could be implemented with sktime. 
This this notebook the `evaluate` method was used as CV splitter method.

#### Does your contribution introduce a new dependency? If yes, which one?

Yes, kotsu

#### What should a reviewer concentrate their feedback on?

Performance, design and viability of implementing kotsu in sktime

##### For all contributions
- [x] I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/main/.all-contributorsrc).
- [ ] Optionally, I've updated sktime's [CODEOWNERS](https://github.com/alan-turing-institute/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
- [x] I've added unit tests and made sure they pass locally.
- [x] The PR title starts with either [ENH], [MNT], [DOC], or [BUG] indicating whether the PR topic is related to enhancement, maintenance, documentation, or bug.

##### For new estimators
- [ ] I've added the estimator to the online documentation.
- [ ] I've updated the existing example notebooks or provided a new one to showcase how my estimator works.

Issue #2804, opened 2022-06-13
[ENH] Using Kotsu as a benchmarking framework - Implementation in sktime


TNTran92:
**Is your feature request related to a problem? Please describe.**
Reference Issue #2777
Kotsu is a high level and light-weight package that performs benchmarking that has no dependency on specific packages. At a high level, it allows users to register a dataset with a validation task (CV splitter method) and then apply estimators to that registration. It will be beneficial to investigate its application to sktime's benchmarking.

**Describe the solution you'd like**
Input: estimator, dataset, cross validation splitter
Output: a Dataframe showing result per estimator per dataset
*: In this notebook, the result is returned as a .csv saved to the same folder as the notebook

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

fkiraly:
suggestion: let´s be more specific about what "apply" means. E.g., would one expect a wrapper function like `benchmark(estimators, metrics, data, config)` that returns results or dataframe?

TNTran92:
> suggestion: let´s be more specific about what "apply" means. E.g., would one expect a wrapper function like `benchmark(estimators, metrics, data, config)` that returns results or dataframe?

I've modified the title to be more specific. Will keep adding comments throughout the day

TNTran92:
Note to self
Tested all estimators for point forecasting. 3 estimators don't work. Those are 
- [ ]  AutoARIMA
- [ ] Prophet
- [ ] VAR

DBCerigo:
> Note to self Tested all estimators for point forecasting. 3 estimators don't work. Those are
> 
>     * [ ]   AutoARIMA
> 
>     * [ ]  Prophet
> 
>     * [ ]  VAR

I can help look into what's causing the issues for those estimates if help wanted :+1:

fkiraly:
I've looked into `kotsu` and the draft PR https://github.com/alan-turing-institute/sktime/pull/2977 .

I think overall it is great, in simplifying functionality!

One design point that bothers me, though, is the `kotsu` design which separates factories or generators from their kwargs, which in the case of `sktime` would mean separating classes from their hyper-parameter settings (objects cannot be passed, since the factories are called with the kwargs, and objects in general do not have a `__call__` defined).

That is design inconsistent with `sktime` (and `sklearn`) where objects are used as "blueprints", especially composites.
Classes, in isolation, cannot specify runtime composites.

Do I understand this correctly?

[...]

PR #2977, opened 2022-07-11
[ENH] Reimplement benchmarking with kotsu v2

alex-hh:
#### Reference Issues/PRs
See previous discussions #2884 and #2805 with previous examples of integrating kotsu into sktime by @TNTran92 and @DBCerigo 
The current PR is intended to provide an example of how benchmarking could be implemented by wrapping kotsu in such a way that users only need to deal with sktime objects. 

This PR provides a candidate benchmarking interface for sktime estimators/datasets by wrapping kotsu.
Currently only supports forecasters, but is intended to be flexible.
It is intended to allow users to create their own benchmarks by adding sets of estimators and tasks.

kotsu

* To avoid sktime users interacting directly with kotsu, we interface kotsu via Benchmark objects designed to handle sktime objects
* Currently the only output allowed is a df / csv file, however it might be desirable to provide simple ways to save models / predictions etc., by passing appropriate arguments to kotsu's benchmarking
* Only a forecasting benchmark and a base benchmark are currently implemented. The idea is that other types of modelling problem could extend the base benchmark in similar ways however it wasn't clear to me how best to do this and  there may be other suggestions!
* No tests are implemented currently

##### For all contributions

##### For new estimators


examples/04_benchmarking_v2.ipynb
```diff
@@ -0,0 +1,288 @@
...
+       "1                        0.051265  "
+      ]
+     },
+     "execution_count": 5,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "results_df = benchmark.run(\"./forecasting_results.csv\")\n",
```
TNTran92, line 262: In my run, `forecasting_results.csv` shows up in the root directory of the repo. In this by design? If yes, I think it's better in the working directory. 

Though I can easily `pd.read_csv` this file, why are we not saving it to memory?

TNTran92, line 262: In my run, `forecasting_results.csv` shows up in the root directory of the repo. In this by design? If yes, I think it's better in the working directory. 

Though I can easily `pd.read_csv` this file, why are we not saving it to memory?

DBCerigo:
([Slides](https://docs.google.com/presentation/d/1ac0Cz5-_v-J9WUAaIZGobghBGtW8Fuwm4lLupPtn4os/edit) from dev-days-2022 talk on 12-7-22 on "sktime - easy benchmarking", which is relevant to this PR)

ltsaprounis:
@DBCerigo - FYI sktime template pattern motivation / explanation https://github.com/sktime/sktime-workshop-pydata-london-2022/blob/main/notebooks/2_basic_extensions.ipynb

sktime/benchmarking/benchmarks.py
```diff
@@ -0,0 +1,50 @@
+# -*- coding: utf-8 -*-
+"""Benchmarking interface for use with sktime objects.
+
+Wraps kotsu benchmarking interface.
+"""
+from typing import Callable, Optional, Type, Union
+
+import kotsu
```
fkiraly, line 8: I would suggest to move the import *inside* the constructor and inside the `run`.

To complete soft dependency encapsulation, kindly add a `_check_soft_dependencies` at the start of the `__init__`.

fkiraly (changes requested):
Looks great as a minimal example!

I have suggestions above, the blocking one is the one on isolating the soft dependency.
The other one would be nicer from a pattern/design perspective imo.

examples/04_benchmarking_v2.ipynb
DBCerigo: > In my run, `forecasting_results.csv` shows up in the root directory of the repo. In this by design? If yes, I think it's better in the working directory.

I don't think there was much thought gone into where it should be saving yet - the example notebook is intended to be an example, and the hope would be users would configure where they wanted to save their results.  
 
> Though I can easily `pd.read_csv` this file, is there a reason to write to a output file?

When writing `kotsu` we generally wanted results to be persisted; you run some experiments one day, and we want those results to still be there when we come back to the project/work say a week later. Leaving results in memory isn't a reliable way of achieving that.

TNTran92:
Looks like at runtime, the `run` [method](https://github.com/datavaluepeople/kotsu/blob/ee98e1550eb6283b353da5e25822ca979d4a2979/kotsu/run.py#L94) turns models into objects by running `model = model_spec.make()`
This method, returns an [entity](https://github.com/datavaluepeople/kotsu/blob/ee98e1550eb6283b353da5e25822ca979d4a2979/kotsu/registration.py#L96) that has the form of `factory(**_kwargs)` 
Do you think we can pass object directly to `run` instead of having to go through `spec`?

sktime/benchmarking/benchmarks.py
DBCerigo: Done in [4e18ac4](https://github.com/alan-turing-institute/sktime/pull/2977/commits/4e18ac46611ae0cea5aba52c5afbc753c105dc6b)

sktime/benchmarking/benchmarks.py
```diff
@@ -0,0 +1,50 @@
...
+        estimator_kwargs: Optional[dict] = None,
+        estimator_id: Optional[str] = None,
+    ):
+        """Register an estimator to the benchmark."""
+        estimator_id = estimator_id or f"{estimator_entrypoint.__name__}-v1"
+        self.estimators.register(
+            id=estimator_id, entry_point=estimator_entrypoint, kwargs=estimator_kwargs
+        )
+
+    def _add_task(
```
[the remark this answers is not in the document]
DBCerigo: Having considered this, I can't figure out how this pattern can be usefully applied here, at least not in the same way it is usefully applied to estimators. 

For forecasters:
```
class BaseForecaster:
    ...
    def fit(self, y, X=None, fh=None):
        .... # do some data validation etc.
        self._fit(y=y_inner, X=X_inner, fh=fh)
        ...

...

class _ProphetAdapter(BaseForecaster):
    ...
    def _fit(self, y, X=None, fh=None):
        ...
```
is useful because `fit` and all subclass `_fit`s have the same signature. 

For benchmarkers, the signatures are not the same, as the task parameters are specific to the subclass (in the same way the init parameters are specific to the forecaster subclasses). The only way I could see this working is for `BaseBenchmark`'s `add_task` to take something like `*args, **kwargs`, but then with no assurances on what the input to the function is there can't be any reliable pre-processing/checks that is dependent on the inputs before `_add_task` is called as happens for `fit`. 

Could you advise if I'm missing something here?

DBCerigo:
Now ready for re-review @fkiraly. 

Some replies made directly on review comments above, and another in #2804.

## PR checklist

sktime/benchmarking/benchmarks.py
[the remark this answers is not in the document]
fkiraly: hm, I suppose you are right.

Originally, I thought, all the tasks would have comparable signature, but I guess in the current design they do not.

fkiraly (changes requested):
I suppose this is good, as a design study!

Blocking change request:
* please add docstrings to public methods and user facing classes

Questions:
* how would carrying out tests integrate with this? You need access to individual predictions for that.
* how would storing partial results on the HD or via a data interface class work?
* I do not fully understand what type or signature you require for `dataset_loader`, could you elaborate?

DBCerigo:
> Blocking change request:
> 
>     * please add docstrings to public methods and user facing classes

Just to confirm, you'd like the doc strings extended to contain `Parameters` sections etc. right? 
 
> Questions:
> 
>     * how would carrying out tests integrate with this? You need access to individual predictions for that.

What do you mean by "tests"? 
If you mean statistical tests and the like - anything that is for a single estimator can we achieved using the system in the PR. Anything that is cross/multi-estimator, that will require the storing of the predictions, which we'll do in the subsequent PR.
If you mean unit tests - such tests are present in this PR. 

>     * how would storing partial results on the HD or via a data interface class work?

Can you clarify what you mean by "partial results"? 

>     * I do not fully understand what type or signature you require for `dataset_loader`, could you elaborate?

This implementation isn't intending to add a new object or interface for a `dataset_loader`, it was intending to follow what was defined in `sktime.datasets`.

fkiraly:
> Just to confirm, you'd like the doc strings extended to contain `Parameters` sections etc. right? 

Yes, `Parameters` and, if appropriate, `Returns`. These end up in the public documentation.

>>how would carrying out tests integrate with this? You need access to individual predictions for that.

> What do you mean by "tests"?

Apologies, I meant "statistical hypothesis tests", on comparing performance of estimators. Not a blocking comment, anyway.

> Can you clarify what you mean by "partial results"?

For instance, predictions, of m out of n estimators involved in the experiment - similar to the old framework.

> This implementation isn't intending to add a new object or interface for a `dataset_loader`, it was intending to follow what was defined in `sktime.datasets`.

Ah, I see - docstrings would be helpful that detail the assumptions, then.

DBCerigo:
@fkiraly this ready for (possibly final) re-review now. Doc strings all updated in [3e28d29](https://github.com/alan-turing-institute/sktime/pull/2977/commits/3e28d29382dd369d3ede233315a4201c6247104e). Nice one.

TNTran92:
Great work. Thanks a lot @DBCerigo 

The classification factory is also up (#3278); I'm still working on its unit tests

fkiraly (changes requested):
Really nice, thanks!
I would be happy to approve, let me know if we should merge and then work on this, or work in this PR.
I have some suggestions for improvement, which might be good to address by 0.14.0 (mid Sep) when we advertise this.

* really minor thing, but let'd not introduce deprecated calls. `frame.append` will be replaced by `concat`.
* I have thought about our earlier discussion, and I think there is a "cleaner" way to structure the class and use the template pattern. Below, not necessary for approval, but please let me know what you think. I think it's cleaner and easier to extend.

Here's how I'd structure the class:
* `add_task` looks like it would be the same for all tasks? I.e., passing its `kwargs` on to `_add_task`. I'd put that in the base class, with generic `kwargs` code.
* The first arg passed to `self._add_task`, the factory, is the point of variation. I would use the template pattern here, i.e., give the `BaseBenchmark` a function called `benchmark`, or similar. This is abstract and is overridden in children, e.g., by what `forecasting_validation` currently is.
* The `benchmark` method is cranked through the factory and then directly referenced by `add-task`. The trick is to always call it `benchmark` and `benchmark_factory` or similar, so you can reference it in `add_task` in `BaseBenchmark`.

An implementer would then just implement the `benchmark` function, in the standard case. The rest works automatically (via `kwargs` and `inspect`).

Do you think this makes sense?

fkiraly:
🎉 welcome back, @DBCerigo 🎉

DBCerigo:
@fkiraly good to be back :smile_cat: This is now ready for re-review. The change since last review is https://github.com/sktime/sktime/pull/2977/commits/c25de09cd40e42ff084c235092e4db63c2e1a50a, as discussed on slack.

fkiraly:
Great! Good to have you back.

There's one point to address, it is dependency isolation.

If `kotsu` becomes a soft dependency, dependent functionality should be isolated. To see examples how this is to be done, search for:

* `pytest.mark.skipif`
* `_check_soft_dependencies`

in the code base. Also see

DBCerigo:
@fkiraly thanks for the pointers for the test skips - all implemented now. 

Let's wait until we have 2+ subclasses of the base class before doing any more design changes, as having multiple examples will make it plainer to see and verify the usefulness of any further design changes. So I would suggest merging this PR as is. Then implementing and merging 1+ more benchmarks (I think @TNTran92 has a PR for classifiers already? though likely needs some retweaking now), then reviewing the implementation as a whole.

fkiraly (approved):
Massive PR, extremely useful functionality!

Comments are either addressed, or make sense to address separately, e.g., minor rewors to the interface.

Issue #2978, opened 2022-07-11
[ENH] Relax numpy dependency to allow numpy 1.22


jlopezpena:
`numpy` 1.23 is already released, and pretty much all the `numpy` related tooling (specifically `numba`) already support `numpy` 1.22, so it should be possible to relax the `sktime` dependency to `numpy < 1.23`.

Getting this change on would be useful because `numpy` 1.22 introduced a forward-incompatible ABI change, and `pip` sometimes does not do the right thing when it comes to building/linking C extensions. If a dependecy-of-a-dependency relies on `numpy` but hasn't specified it as a _build time_ dependency the building step links against the latest available `numpy`, which results in a run-time error when trying to load a 1.22-linked extension from 1.21 sue to different header sizes.

These errors typically look like this:
```
ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject
```

I tested locally and it "just worked". I can make a PR if deemed appropriate (it is literally just a 1-byte change!) but wanted to open this issue in order to have some discussion on whether it is appropriate to do so.

fkiraly:
hm, the version bump looked good!

Should we remove the bound altogether, in the spirit of the "avoid upper bounds" discussion? What do you think, @lmmentel?

`pandas` is consistently breaking things on version increase, so I´d leave that one, but we never had problems with numpy.

aiwalter:
Closed by #2979

PR #2979, opened 2022-07-11
[MNT] Update numpy to <=1.22

jlopezpena:
Enable using numpy 1.22. This is the change suggested in #2978


lmmentel:
Not sure where the upper bound came from but if the tests are passing we can relax the version.

fkiraly:
>but if the tests are passing we can relax the version.

or remove the upper bound? I vaguely remember it might have been sth with `numba`.

PR #2980, opened 2022-07-11
[BUG] Early classification test fixes

MatthewMiddlehurst:
Fixes the duplication of classifier testing caused by an import in the early classification test file.

Adds early classification scenarios for basic tests, so they are actually being run now.

Excludes TEASER from three tests.
- `test_methods_do_not_change_state` does not really work the way we have things set up currently, as early classifiers intentionally retain information from pervious predict calls.
- `test_fit_idempotent` and `test_persistence_via_pickle` break due to predict/predict_proba returning two items and that breaking assert_array_equal. This should be easier to fix, but would require editing the tests to check for multiple outputs/a tuple.


fkiraly (approved):
Looks ok now, thanks!

I think `_make_fit_args` is no longer used anywhere? Anyway, that's not a problem of this PR.

PR #2986, opened 2022-07-12
PERMANENT DIAGNOSTIC: tests failing without soft dependencies - do not merge, do not close

fkiraly:
DO NOT MERGE, DO NOT CLOSE

PERMANENT DIAGNOSTIC PR to investigate whether soft dependencies have been correctly isolated in tests.

Based on PRs:
* PR https://github.com/alan-turing-institute/sktime/pull/2985 which ensures the main test suite only runs with estimators that do not require soft dependencies
* PR https://github.com/alan-turing-institute/sktime/pull/2990 which isolates some of the remaining soft dependencies

Any failures should indicate one of:
* custom (non-suite) tests (or test modules) that import a soft dependency, including, possibly, doctests
* bugs with the soft dependency isolation in the main test suite in #2985

PR #2990, opened 2022-07-12
[ENH] isolate soft dependencies

fkiraly:
This PR isolates soft dependencies at runtime and test time as follows:

* moving imports into functions and classes
* tests requiring soft dependencies are only run when required soft dependencies are installed
* in some tests, classes that require soft dependencies are replaced by equivalents (for test purpose) without soft dependencies
* doctests with soft dependencies are all skipped (all seem special cases of things tested elsewhere)

This is a continuation of PR https://github.com/alan-turing-institute/sktime/pull/2985 for custom/manual tests, with the aim to isolate all instances of soft dependency imports in the above way.


fkiraly:
Ready for review now - all soft dependencies are isolated, as #2986 shows.

Issue #971, opened 2021-06-17
[ENH] Nested `get_fitted_params` and `summary` interface


fkiraly:
This is a feature suggestion arising from discussion in #877, #931, #942.

I suggest two new/refactored interface points:
* `summary` which is attached to `BaseObject` (as in post-merge #877)
* `get_fitted_params` which is attached to `sktime`'s (not `scikit-learn`'s) `BaseEstimator`

Update: it is probably better than the more convoluted suggestion below if the returns of both are `dict` and behave just like `get_params`. Any logic for retrieval should be provided in there, or a private version `_get_fitted_params`, rather than in "populate" functions.
 
---
Old API suggestion

Both should behave precisely like `scikit-learn`'s `get_params`, except that parameters are looked up in dedicated dictionaries `reports` and `fitted_params` rather than in self-attributes that correspond to `__init__` named arguments.

`fit` or `__init__` could populate these dictionaries, this could happen by `_populate_fitted_params` and `_populate_summary` private interfaces called at the end of `fit` (the non-private with the "plumbing" logic), which is specific to individual estimators/objects, or directly in `_fit` (not sure which is the better convention?).

For functionality, I think we can copy-paste `scikit-learn`'s `get_params` code here, and adapt it by making the change in where parameters are being retrieved. This would ensure nested access, which for example `get_fitted_params` doesn't have at current.

What do you think?
@mloning, @aiwalter, @RNKuhns, @ltsaprounis

fkiraly:
Minor point: if you agree, what should be the name for the method retrieving information like AIC, R-squared, etc? `get_reports`, `inspect`, `inspect_reports`, `inspect_model`, `report_model`, `get_report_params`, `ishmael`, ...?

mloning:
If reports are populated in `fit`, it should be attached to the `BaseEstimator` rather than `BaseObject` as not all objects will have a `fit` method. Or do you have other kinds of information in mind that other objects may have?

As to the naming:
* `inspect` seems too general given that we have `get_fitted_params`
* the terms `report` and `model` aren't used anywhere else, but `get_report` seems understandable
* `get_summary` may be an alternative, that's used in `statsmodels` already

fkiraly:
>get_summary may be an alternative, that's used in statsmodels already

Alright, agree with the suggestion that's not on my list.

>If reports are populated in fit, it should be attached to the BaseEstimator rather than BaseObject as not all objects will have a fit method. Or do you have other kinds of information in mind that other objects may have?

[...]

PR #2991, opened 2022-07-13
[ENH] Changes to CNN Classifier

### LINKED ISSUE #2935: [ENH] Changing and Adding features to CNN Classifier
As discussed in the daily-stand up on 05-07-12, I wanted to propose some enhancements / changes to CNN Classifier (and hence to Deep Learning based estimators)

- [ ] Moving the `convert_y_to_keras` function in `BaseDeepClassifier` into a utils function
- [x] Enable returning the `label_encoder` and `one hot encoder` to the user from `convert_y_to_keras` (it can be useful for the users in case they want to use it to encode other data / or to decode data later on)
- [x] Add option to have multiple linear layers in output layer / different activation function. Also add the `use_bias` parameter to output layer. (simply as a customisation option to users, multiple linear layers are useful to create deeper networks depending on dataset / choice of activation function)
- [x] option to choose optimiser (usually its Adam but just for customisation's sake we can give this as option as well). Also should allow specifying `lr` either within the optimiser or as parameter to the class itself.
- [x] Accept parameter as input in `CNNClassifier` to specify `random seed`
- [ ] Train/fit the model on GPU depending on users choice (maybe take an input parameter for this as well)
- [x] Option to view / return the loss of fit (either as a graph or simply as a list / array ) / A better idea is to add a summary function which can return model summary as well as `fit` history. 

**Additional context**
These are some additional features / changes I wanted to add to the `CNNClasifier`. Any inputs / additional comments are appreciated

AurumnPegasus:
#### Reference Issues/PRs

Fixes #2935 

- [ ] Moving the `convert_y_to_keras` function in `BaseDeepClassifier` into a utils function
- [x] Enable returning the `label_encoder` and `one hot encoder` to the user from `convert_y_to_keras` (it can be useful for the users in case they want to use it to encode other data / or to decode data later on)
- [x] Add option to have multiple linear layers in output layer / different activation function. Also add the `use_bias` parameter to output layer. (simply as a customisation option to users, multiple linear layers are useful to create deeper networks depending on dataset / choice of activation function)
- [x] option to choose optimiser (usually its Adam but just for customisation's sake we can give this as option as well). Also should allow specifying `lr` either within the optimiser or as parameter to the class itself.
- [x] Accept parameter as input in `CNNClassifier` to specify `random seed`
- [ ] Train/fit the model on GPU depending on users choice (maybe take an input parameter for this as well)
- [x] Option to view / return the loss of fit (either as a graph or simply as a list / array )

##### For all contributions


AurumnPegasus:
For the first task
>  Moving the `convert_y_to_keras` function in `BaseDeepClassifier` into a utils function

Any idea exactly _where_ should I place this utils function?

AurumnPegasus:
For GPU functionality:
1. A user would need `tensorflow-gpu` and `cuda-toolkit` installed. We will have to add them as a soft dependency? I am unsure if these are required since pytorch (which I am comfortable with) does not really require them
2. We can also have the functionality of multi-gpu: https://keras.io/guides/distributed_training/ . In my estimate majority of the people using sktime would be on kaggle / collab in which case there is a single gpu used. And time series data should not be so heavy that multi-gpu are required, but there might be few using it on larger data / maybe college gpu cluster where they might have easy access to gpus and would prefer running the models with multiple gpus? It might be worth having in those cases

AurumnPegasus:
Update: I am also ignoring tests `test_fit_does_not_overwrite_hyper_params` for `CNNClassifier` since that is giving issues wrt pickling tf models (more specifically, it is unable to pickle a lambda funciton which is used for optimizer).
[Stack overflow](https://stackoverflow.com/a/49858357) mentions that pickle by default does not support lambda functions, and doing any work around for the same would require going too deep into tf implementation instead.

fkiraly:
> Update: I am also ignoring tests `test_fit_does_not_overwrite_hyper_params` for `CNNClassifier` 

What is the problem with that test?

AurumnPegasus:
The test uses pickle to save the model, and I think once we update the save function as per #3022 we can stop ignoring it.

```
    raise PicklingError(
_pickle.PicklingError: ("Can't pickle <function make_gradient_clipnorm_fn.<locals>.<lambda> at 0x7f780541a3a0>: it's not found as keras.optimizers.optimizer_v2.utils.make_gradient_clipnorm_fn.<locals>.<lambda>", 'PicklingError while hashing <keras.optimizers.optimizer_v2.adam.Adam object at 0x7f78041cefd0>: PicklingError("Can\'t pickle <function make_gradient_clipnorm_fn.<locals>.<lambda> at 0x7f780541a3a0>: it\'s not found as keras.optimizers.optimizer_v2.utils.make_gradient_clipnorm_fn.<locals>.<lambda>")')

```

AurumnPegasus:
Hey @fkiraly  @TonyBagnall ,
Any idea why the single test is failing for mac? I can not find any leads

fkiraly:
re single test, for anyone who reads this, it was discussed on slack, and is related to the sporadic failure #2368

sktime/classification/deep_learning/base.py
```diff
@@ -119,4 +131,4 @@ def convert_y_to_keras(self, y, label_encoder=None, onehot_encoder=None):
             y = y.reshape(len(y), 1)
             y = onehot_encoder.fit_transform(y)
 
-        return y
+        return y, label_encoder, onehot_encoder
```
fkiraly, line 134: If you make this a public function, please provide a docstring.

If not meant to be public, please provide a docstring and make it private - three returns are non-obvious why/what.

sktime/classification/deep_learning/base.py
```diff
@@ -66,6 +66,18 @@ def build_model(self, input_shape, n_classes, **kwargs):
         """
         ...
 
+    def get_metrics(self):
```
fkiraly, line 69: could you call this `summary` to be in line with #971 and `statsmodels` type models?

sktime/classification/deep_learning/cnn.py
```diff
@@ -59,6 +80,10 @@ def __init__(
         verbose=False,
         loss="mean_squared_error",
         metrics=None,
+        random_seed=0,
```
fkiraly, line 83: default should be `None`

fkiraly (changes requested):
Thanks! Looks almost good to go to me!

I left some comments related to interface conformance, `summary` and `random_seed`. The rest is not blocking.

Also, question: can you explain why the test `test_fit_does_not_overwrite_hyper_params` did *not* fail prior to this PR? Which change makes it fail? Perhaps that could be avoided somehow?

sktime/classification/deep_learning/base.py
AurumnPegasus: Added in the documentation.

sktime/classification/deep_learning/base.py
AurumnPegasus: Done

AurumnPegasus:
> Also, question: can you explain why the test `test_fit_does_not_overwrite_hyper_params` did _not_ fail prior to this PR? Which change makes it fail? Perhaps that could be avoided somehow?

From what I know, it is because of specifying `lr` in the optimizer by default. The full error description is this

```
'test_fit_does_not_overwrite_hyper_params[CNNClassifier-ClassifierFitPredictMultivariate]': PicklingError("Can't pickle <function make_gradient_clipnorm_fn.<locals>.<lambda> at 0x7f0a6918aaf0>: it's not found as keras.optimizers.optimizer_v2.utils.make_gradient_clipnorm_fn.<locals>.<lambda>", 'PicklingError while hashing <keras.optimizers.optimizer_v2.adam.Adam object at 0x7f0a6917e940>: PicklingError("Can\'t pickle <function make_gradient_clipnorm_fn.<locals>.<lambda> at 0x7f0a6918aaf0>: it\'s not found as keras.optimizers.optimizer_v2.utils.make_gradient_clipnorm_fn.<locals>.<lambda>")')
```

On a bit of investigation, the error is when I use Adam (or any optimizer) and store it as a class variable. For example, in the current code I am doing `self.optimizer = Adam()` in line 140, which leads to the `test_fit_does_not_overwrite_hyper_params` failure. On the other hand, if I were to not write that line, and instead directly assign Adam within the `model.compile` at line 149, it will show all tests passed.

Directly specifying the optimizer is not feasible, since we want to give the option to specify the optimizer to the user. Ofcourse, if instead of optimizer we were to directly take input of `learning_rate` we can do it, but it feels like a work around, and it would take away the choice of optimizer from the user (although Adam is the most popular one)

sktime/classification/deep_learning/cnn.py
AurumnPegasus: Done

sktime/classification/deep_learning/base.py
AurumnPegasus: Let me know if its okay, or if I need to change the wordings.

fkiraly:
Ah, I see.

An `sklearn` interface requirement is to never overwrite `self.paramname` if `paramname` is an `__init__` arg.

The test is failing correctly and pointing out that you are not interface compliant!

Instead, write the optimizer to `self.optimizer_` in `__init__`, and overwrite that if it is `None`, so `self.optimizer` remains unchanged.

fkiraly:
PS: generally, please do *not* except tests if you do not understand what the tests are testing.

AurumnPegasus:
@fkiraly 
Updated the code, it works now. Thanks for the suggestion!

AurumnPegasus:
Hey, before merging this I am planning to make a couple more design changes. I will formalise it in a bit, just informing so we dont merge this

AurumnPegasus:
@fkiraly made some changes:
I converted `label_encoder` and `onehot_encoder` to class variables instead of function variables as they were. I dont know why they were function variables in the first place (maybe I did not realise it at first) but `convert_y_to_keras` (the function which takes in the `label_encoder` and `onehot_encoder` as input in `BaseDeepClassifier`) is called from `CNNClassifier` which itself has no parameter for `label_encoder` and `onehot_encoder` (hence, no way for the user to specify it!)

Now, I am asking for them as input in `__init__` of `CNNClassifier` so users can specify required encoders. One thing I want you to note/comment on is line 174-175 where I am doing:

```python
self.label_encoder = self.label_encoder_
self.onehot_encoder = self.onehot_encoder_
```

In `cnn.py`. The reason for this is so that `BaseDeepClassifier` can directly change `self.label_encoder` from the `convert_y_to_keras` method, and hence I do not have to send it parameters, or receive the two encoders along with `y` (which as you stated, is non obvious). If you think this is unnecessary, I can change it back so that we send and receive parameters, and update the documentation to as it was in the previous commit.

sktime/classification/deep_learning/cnn.py
```diff
@@ -59,6 +80,12 @@ def __init__(
         verbose=False,
         loss="mean_squared_error",
         metrics=None,
+        random_seed=None,
+        activation="sigmoid",
+        use_bias=True,
+        optimizer_=None,
```
fkiraly, line 86: the init variables should have no underscore at the end.

What I meant is that these remain as you intended, without underscore.

You then copy a coerced/substituted version to the underscore version.

fkiraly (changes requested):
Thanks!

Two comments:

* can you explain what purpose exposing the label and onehot encoder in the `__init__` serves? What would be a second example that makes sense here, i.e., another value that the user can pass?
* `__init__` arguments should never be mutated. The line `self.label_encoder = self.label_encoder_` creates a (pointer) reference and not a copy, so calling `fit_transform` on it will mutate the reference.

sktime/classification/deep_learning/cnn.py
AurumnPegasus: I have made some changes, hope it satisfies. I am still unsure about what can be done for it, should I do a `deepcopy`?

AurumnPegasus:
> * can you explain what purpose exposing the label and onehot encoder in the `__init__` serves? What would be a second example that makes sense here, i.e., another value that the user can pass?

My essential idea was to allow for custom functions instead of the default one by `sklearn`. For example, if I want to use a specific Label Encoder (and I send the encoded data to CNNClassifier), the CNNClassifier will run the sklearn's default LabelEncoder on top of it which might cause it to lose some information. Maybe a better solution would have been to have a variable OR a check telling if it has already been encoded?

fkiraly:
> Maybe a better solution would have been to have a variable OR a check telling if it has already been encoded? 

Can you give two or three examples of intended usage here? Not sure whether I fully understand.

AurumnPegasus:
Okay so i thought a bit about it and I was wrong, I do not think it is required:
My original though revolved around this particular scenario:
Lets say my inputs are ['>90', '80-90'....'0-10'] (these are the classes I want to classify in. For some reason I want to do different types of grouping, where '50-100' becomes when group and '0-50' becomes another. In this case I would either have to change the data, or simply create a new Label Encoder which would do encodings are required. This is a customisation I thought would be useful
But, if the user has to create a Label Encoding function themself, they can easily just run the test data on it and pass it on to the CNN Classifier. Another doubt I had about information loss would not hold either, since running `sklearn` Label Encoder on encoded data would not give  different results in such a case.
I will remove them and set the default as sklearn's `LabelEncoder` and `OneHotEncoder`

fkiraly:
ok, also fine for me.

I don't have an opinion either way, only applying the general principle that function or init args should be justified by at least two, optimally three different, meaningful examples of values/references that users may want to pass to them.

fkiraly:
Let me know when you think this is read for another review.

AurumnPegasus:
@fkiraly  it is, have made the requested changes.

fkiraly (changes requested):
Great!

* `random_seed` should be `random_state` for consistency with `sklearn` and the rest of `sktime`
* we have already released `BaseDeepClassifier` with args `batch_size` and `random_state`, so should not remove them without deprecation

Things fixed:

* handling of `self.optimizer_` is good now!
* `convert_y_to_keras` looks good now

fkiraly (approved):
Looks good now, thanks!

PR #2996, opened 2022-07-13
[ENH] Data generator for annotation

lmmentel:
#### Reference Issues/PRs

first draft of the data generator for annotation

##### For all contributions

##### For new estimators


lmmentel:
@miraep8 and @KatieBuc I would appreciate your input on this. Drop a comment if you see this as ready to be merged or missing some functionality.

sktime/annotation/datagen.py
```diff
@@ -0,0 +1,138 @@
...
+"""Synthetic data generating functions."""
+
+from typing import Union
+
+import numpy as np
+import numpy.typing as npt
+from sklearn.utils.validation import check_random_state
+
+# what if we could pass data_gen functions - rather than assuming means?
+#   - could make it easier to swap out different mean assumptions
```
miraep8, line 11: Wondering - is it worth leaving this comment/commented code at the bottom here?  Could be cleaner to remove it and just add ideas elsewhere?  But can also see the value of leaving some ideas in the code as well.  If the latter though maybe  we could make it more organized?

sktime/annotation/datagen.py
```diff
@@ -0,0 +1,138 @@
...
+    """Data generator base class in order to allow composition."""
+
+    def __init__(self, means, lengths, noise, random_state=None):
+        self.means = means
+        self.lengths = lengths
+        self.noise = noise
+        self.random_state = random_state
+
+    def sample(self):
+        """Generate data sample."""
```
miraep8, line 112: Should we also add another function `label` to this class which generates the labels and returns them?  From what I can tell this isn't explicitly included in the class but could be nice.

miraep8 (commented):
Hey!  Looking it over - I think its great as a first pass/simple data gen class, just had a few small comments.  Think we can add some of the more complex layers later as needed.  Thanks for adding the random state param, think that will help make things repeatable.  I personally would be in favor of merging it, and the more complicated stuff can be added at another stage if needed. :) 

Some things to potentially add down the road:
- make a multi-variate option. 
- make it easier to pass in an arbitrary probability distribution?
- make the labeling more generic.
- add some more tests?  To verify that the data generation is behaving as expected.

But, my 2 cents is definitely lets go ahead and merge it!

sktime/annotation/datagen.py
lmmentel: I don't have a good idea yet on how that class could be used so I'll leave it be, unless you want to tackle it?

sktime/annotation/datagen.py
lmmentel: ... and cleaned up the comments :wink:

sktime/annotation/datagen.py
```diff
@@ -0,0 +1,134 @@
...
+    Each segment has length specified in ``lengths`` and data sampled from a normal
+    distribution with a mean from ``means`` and standard deviation from ``noise``.
+
+    Parameters
+    ----------
+    means : array_like
+        Means of the segments to be generated
+    lengths : array_like
+        Lengths of the segments to be generated
+    noise : float ir array_like
```
KatieBuc, line 29: noise : float ir array_like

KatieBuc: specify int as option too?

sktime/annotation/datagen.py
```diff
@@ -0,0 +1,134 @@
...
+    """
+    Generate labels for a series composed of segments.
+
+    Parameters
+    ----------
+    means : array_like
+        Means of the segments to be generated
+    lengths : array_like
+        Lengths of the segments to be generated
+    noise : float ir array_like
```
KatieBuc, line 96: **or

KatieBuc:
Looks good!! I think it's ready but some minor questions/comments

I was also wondering about the ``labels``, if it could be an argument to ``sample``, calling ``label_mean_shift``

Adding to Mira's wish list:
- known standard deviation changes (as opposed to just noise)

lmmentel:
@ltsaprounis @fkiraly @aiwalter could you review and approve?

fkiraly (approved):
All good with me, sensibly documented and structured.

It's experimental, so I assume it will change one or few times still.

GuzalBulatova (approved):
This looks very good to me, thank you Łukasz!

KatieBuc:
Sure!! @lmmentel @miraep8 

> Could you give an example of what you mean? Right now we can do either same noise/std for every segment by passing `noise=2.0` for example or passing noises for each of the segments by `noise=[1.0, 2.0, 5.0]` <- assuming 3 segments. Is there an additional case you would like covered?

Like, for example, this: https://github.com/alan-turing-institute/sktime/compare/main...KatieBuc:sktime:data_generator_multivar

Oh I realise what I said didn't make much sense - what I meant to say was "known covariance changes (as opposed to just independent noise/ variance)" - in the multivariate setting. I've made a flexible parametrisation (i.e. you can give it covariance or just variance and makes an assumption of independence. Not sure if this is good, but it's a first pass.

Note the inconsistency between the multivariate function naming convention of noise (as std dev) vs. vars (as variance). Maybe there should also be a check to see if input covs are symmetric and positive *semi*definite - but I'll keep working and start a separate pull request.

PR #3002, opened 2022-07-14
[ENH] Add multivariate CNN example (from sktime-dl)

tobiasweede:
Adds the corresponding jupyter notebook with minor changes to work within sktime.

Don't know if you have plans to port everything from sktime-dl one-by-one or in a single big effort.
In the latter case, please just discard my PR.

### DIFF
--- a/examples/02a_classification_multivariate_cnn.ipynb
+++ b/examples/02a_classification_multivariate_cnn.ipynb
@@ -1,0 +1,137 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Multi-variate time series classification using a simple CNN\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


fkiraly:
Thanks a lot, @tobiasweede!

Linting (code formatting) checks are failing though, please look at "details" next to the `code-quality` CI. Happy to show how to enable local code formatting on your computer in one of the Fri sessions

tobiasweede:
The linting is fixed.
I also applied the suggested naming convention in  #3003.

fkiraly (approved):
Thanks!

Not sure whether this should be the first notebook after the main classification notebook, or whether there should be a bigger notebook on neural networks first. Think about reading things in order, is it in logical sequence for the reader?

Either way, we can bring things into a nicer order later - thanks for working on this!

PR #3003, opened 2022-07-14
[ENH] add inception time with example (from sktime-dl)

tobiasweede:
Towards #3351

Migrates `InceptionTime` classifier, network and the corresponding jupyter notebook from `sktime-dl`, with minor changes to work within `sktime`.

Don't know if you have plans to port everything from sktime-dl one-by-one or in a single big effort.
In the latter case, please just discard my PR.


fkiraly:
> Don't know if you have plans to port everything from sktime-dl one-by-one or in a single big effort.

I don't know about that for sure, I think we wanted to do it step by step? @TonyBagnall can tell you more.

In either case, any help is appreciated!

fkiraly (changes requested):
Excellent, thansk for the port!

May I suggest one thing, to avoid unstructured clutter in the `examples` dir:
could you rename the notebook to sth that makes clear it is about classification, sub-topic neural networks? E.g., `02a_classification_deeplearning_inceptiontime`?

Or something similarly structured so that a lexicographic sort makes clear where it's from.
You may also like to think about order and naming in the context of the other notebook(s) you are adding - for instance, doesn't need to be 02 *a* as long as there is a consistent, logical order to things.

fkiraly:
@tobiasweede, FYI, the grid search has the `best_params_` attribute etc only after `fit` has been called.

sktime/utils/data.py
```diff
@@ -0,0 +1,107 @@
...
+
+__author__ = "James Large"
+
+import pandas as pd
+
+from sktime.datatypes._panel._convert import from_nested_to_3d_numpy
+from sktime.utils.validation.panel import check_X, check_X_y
+
+
+def check_and_clean_data(X, y=None, input_checks=True):
```
fkiraly, line 12: much of this seems redundant with functionality now in `datatypes`.
Have a look at `convert_to`.

Also, if this is specific to neural networks, should it live in `networks`?

fkiraly:
FYI, I do not consider the "refactor" comments as blocking.
What I would consider blocking is my suggestion to localize the specific checks as private, in the same file or module as the estimator, so we do not duplicate existing utilities modules. We can refactor to `convert_to` etc later. Or, you could refactor/replace now, as said the "refactor" part is not blocking.

tobiasweede:
@fkiraly Thank you for you comments. I will work on the case and try to address the issues.

I also plan show up at a collab code session.
Probably not this Friday, but during the next weeks...

fkiraly:
> I also plan show up at a collab code session.

Nice, looking forward to that!
Would be appreciated if you send a message on the slack contributors channel a few days ahead, so the people interested in neural networks know.

achieveordie:
Hey, @tobiasweede Are you still working on this? We could really make use of `InceptionTime` in `sktime`. It'll be incredibly helpful, thanks!

sktime/classification/deep_learning/inceptiontime.py
```diff
@@ -0,0 +1,200 @@
...
+            X,
+            y_onehot,
+            batch_size=self.batch_size,
+            epochs=self.n_epochs,
+            verbose=self.verbose,
+            callbacks=self._callbacks,
+            validation_data=validation_data,
+        )
+
+        self._is_fitted = True
```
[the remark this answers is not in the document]
fkiraly: I removed this

sktime/utils/data.py
fkiraly: I replaced this

fkiraly:
I've taken the freedom to wrap this up.
FYI @gaudium

fkiraly (approved):
all fixed up now

phylopylo:
I have tested the InceptionTimeClassifier and confirmed that it runs without error.

PR #3019, opened 2022-07-15
add testdir to .gitignore

Lovkush-A:
when one runs `make test` it produces a directory `testdir/`. this should be ignored by git.


fkiraly (approved):
Makes sense.

lmmentel (approved):

PR #3021, opened 2022-07-15
[BUG] fix bug with data loading from timeseriesclassification when extract_path is not None

TonyBagnall:
problem caused when loading data from the website given a specific path, caused by not wanting to write data to datasets/data when no path is given. There is no test function for this, since it relies on tsc.com, but the following use cases all work.

```python
from sktime.datasets import load_UCR_UEA_dataset
# NO EXTRACT PATH
# Use case 1: baked in data set. Should load without downloading
X, y = load_UCR_UEA_dataset(name="ArrowHead")
print(" Case 1 Arrowhead y[0] =", y[0])
# Use case 2: not baked in, not already downloaded: should download to local_data
X, y = load_UCR_UEA_dataset(name="BeetleFly")
print(" Case 2 BeetleFly y[0] =", y[0])
# Use case 3: not baked in, but already downloaded: should load from local_data
X, y = load_UCR_UEA_dataset(name="BeetleFly")
print(" Case 3 BeetleFly y[0] =", y[0])

# Absolute extract path given: C:/temp    or C:/Temp/ or C:\\Temp or C:\\Temp\\
# Use case 4: dataset name currently in path/
X, y = load_UCR_UEA_dataset(name="ArrowHead", extract_path="C:\\Temp")
print(" Case 4 a. Arrowhead y[0] =", y[0])
X, y = load_UCR_UEA_dataset(name="ArrowHead", extract_path="C:\\Temp\\")
print(" Case 4 b. Arrowhead y[0] =", y[0])
X, y = load_UCR_UEA_dataset(name="ArrowHead", extract_path="C:/Temp")
print(" Case 4 c. Arrowhead y[0] =", y[0])
X, y = load_UCR_UEA_dataset(name="ArrowHead", extract_path="C:/Temp/")
print(" Case 4 d. Arrowhead y[0] =", y[0])

# Use case 5: dataset name NOT path/ but path exists
X, y = load_UCR_UEA_dataset(name="BeetleFly", extract_path="C:\\Temp")
print(" Case 5 BeetleFly y[0] =", y[0])

# Use case 6: dataset name NOT path/ and path does not exist
X, y = load_UCR_UEA_dataset(name="BeetleFly", extract_path="C:\\Temp\\Temp")
print(" Case 6 BeetleFly y[0] =", y[0])

# Use case 7: dataset name and path exists
X, y = load_UCR_UEA_dataset(name="BeetleFly", extract_path="C:\\Temp\\Temp")
print(" Case 7 BeetleFly y[0] =", y[0])

# Use case 8: relative path name given e.g. Temp or ./Temp
X, y = load_UCR_UEA_dataset(name="BeetleFly", extract_path="Temp")
print(" Case 8 a BeetleFly y[0] =", y[0])
X, y = load_UCR_UEA_dataset(name="Chinatown", extract_path="./Temp")
print(" Case 8 b Chinatown y[0] =", y[0])
"""


MatthewMiddlehurst (approved):

fkiraly (changes requested):
should this not be accompanied by a test that fails pre-PR and runs post-PR?

fkiraly:
> we removed all data downloading related tests a while back

is there a way to test the post-download logic, because that is what this seems to be about?

TonyBagnall:
> > we removed all data downloading related tests a while back
> 
> is there a way to test the post-download logic, because that is what this seems to be about?

I'm not really sure, you might need to check locally what is available in terms of data set, create a bespoke path then delete it subsequently. This is a two line fix for a small bug

TonyBagnall:
so we could include  zip file in the test folder, then simulate downloading from tsc.com by extracting this into a temp directory, then delete the temp directory. I would worry something would go wrong with the deletion, there will be permission issues I suspect there will be permission issues in the virtual machines. Seems excessive to me, but I guess I will do it if you feel strongly about it

TonyBagnall:
what do you want to do here @fkiraly ? I want this two line PR to go in as it is to fix a minor bug in the code base. Do you want to continue to block that?

fkiraly (approved):
looks reasonable

PR #3026, opened 2022-07-15
[ENH] GraphPipeline Design Study

miraep8:
@benHeid @fkiraly @SMEISEN

Draft Design PR based on what I discussed during the pywatts discussion today! :)

### DIFF
--- a/sktime/base/_simple_base.py
+++ b/sktime/base/_simple_base.py
@@ -1,0 +1,72 @@
+#!/usr/bin/env python3 -u
+# -*- coding: utf-8 -*-
+# copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+"""A minimal BaseObject and BaseEstimator."""
+
+__author__ = ["miraep8"]
+__all__ = ["SimpleBaseObject", "SimpleBaseEstimator"]
+
+import abc
+
+
+class SimpleBaseObject:
+    """Simple class that facilitates getting and setting params for all estimators"""
+
+    input_checks = []
+
+    def run_input_checks(self, kwargs):
+        for check in self.input_checks:
+            check.check(kwargs)
+
+    def get_params(self, name_str="", deep=True):
+        """Get all parameters for this Estimator."""
+        param_return = {}
+        for parameter in self.parameters.keys():
+            param_key = "".join([name_str, parameter])
+            param_return[param_key] = getattr(self, parameter)
+            if deep and isinstance(param_return[param_key], SimpleBaseEstimator):
+                param_return.update(
+                    param_return[param_key].get_params(
+                        name_str="".join([parameter, "__"])
+                    )
+                )
+        return param_return
+
+    def make_list(self, obj):
+        """Make obj a list if it isn't already."""
+        if isinstance(obj, list):
+            return obj
+        return [obj]
+
+    def set_params(self, **kwargs):
+        """Update all parameters for valid parameters and update types."""
+        for parameter, value in kwargs.items():
+            if parameter in self.parameters.keys() and type(value) in self.make_list(
+                self.parameters[parameter]
+            ):
+                setattr(self, parameter, value)
+            else:
+                raise ValueError(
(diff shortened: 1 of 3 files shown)


sktime/pipeline/_graph_pipe.py
```diff
@@ -0,0 +1,210 @@
...
+
+    def __init__(
+        self,
+        steps: list = [],
+    ):
+        params = GraphPipeline.assemble(steps)
+        self.set_params(**params)
+
+    def add(
+        self, name: str, estimator: BaseEstimator, input_edges, input_indices, tags
```
benHeid, line 172: I am unsure if we need the input_indices besides input_edges. We also should discuss if input_edges is a list or a dict. In the pyWATTS implementation of your API it is currently an API to enable a matching of the predecessors to the parameters of the fit.transform (pyWATTS allows more arguments than x and y).

sktime/pipeline/_graph_pipe.py
```diff
@@ -0,0 +1,210 @@
...
+                    output = node.estimator.inverse_transform(**transform_args)
+                else:
+                    output = node.estimator.fit_transform(**transform_args)
+                if "y" in node.tags:
+                    input_y = output
+                else:
+                    input_X = output
+            if isinstance(node.get_estimator, BaseForecaster):
+                node.estimator.fit(y=input_y, X=input_X)
+                input_y, input_X = node.estimator.y, node.estimator.X
```
benHeid, line 208: This can be probably extracted into the steps or graphnodes (we have to decide on the naming graphnode vs step). Then this could be solved via inheritance. On specific graphnode per base class.

sktime/pipeline/_graph_pipe.py
```diff
@@ -0,0 +1,210 @@
...
+from sktime.base._simple_base import SimpleBaseEstimator, SimpleBaseObject
+from sktime.base._tests_and_checks import LengthChecker
+from sktime.forecasting.base import BaseForecaster
+from sktime.transformations.base import BaseTransformer
+
+__author__ = ["miraep8"]
+__all__ = ["GraphOutput", "GraphNode", "GraphPipeline"]
+
+
+class GraphOutput:
```
benHeid, line 19: Is the idea of this class to manage and collect the results of all graphnodes and steps? 

I like this idea, this probably makes the pipeline class easier to understand.

sktime/base/_tests_and_checks.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env python3 -u
+# -*- coding: utf-8 -*-
+# copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+"""collection of different verificiation and checking classes that can be used in composition."""
```
benHeid, line 4: Thank you for adding _tests_and_checks.py Which assumption we have on the input data and thus which checks we need.

sktime/base/_simple_base.py
```diff
@@ -0,0 +1,72 @@
...
+                            f"{type(self).__name__} doesn't have a parameter ",
+                            f"{parameter} which accepts input of type",
+                            f"{type(value).__name__}. ",
+                            f"Valid options and types are {self.parameters}",
+                        ]
+                    )
+                )
+
+
+class SimpleBaseEstimator(SimpleBaseObject, metaclass=abc.ABCMeta):
```
benHeid, line 61: Is the pipeline an estimator or a transformer. I am not sure what fits better. We should discuss this. However, currently I tend to an estimator since pipelines can contain estimators and thus the pipeline needs to be fitted.

benHeid (commented):
Than you @miraep8 for this API proposal. I tried to adapt it to [pywatts-pipeline](https://github.com/KIT-IAI/pywatts-pipeline/pull/1) I hope that you and @fkiraly have access rights.

miraep8:
Closing since a lot of this has been migrated to the pywatts interface already/this was just a design PR. :)

PR #3030, opened 2022-07-16
[MNT] xfail known sporadic test failure #2368

fkiraly:
This adds a test xfail of the known sporadic test failure reported as  bug #2368.

The cause of the bug is not known, and the skip is added to avoid sporadic test crashes unrelated to the addition.


fkiraly:
quick merge to prevent wasted CI time - due to larger number of CI items (after 3.10), we have more fails.

PR #3039, opened 2022-07-17
[ENH] test estimator `fit` without soft dependencies

fkiraly:
This PR extends the `test-nosoftdep` CI element with additional tests:

* test that test parameters returned by `get_test_params` do not require soft dependencies.
* test that estimators without soft dependencies (according to tag) can be fitted without a `ModuleNotFoundError`. Currently, the `test-nosoftdep` CI only tests that estimators construct, which may be too weak, since soft dependencies may show up only in `fit`.

### DIFF
--- a/sktime/tests/test_softdeps.py
+++ b/sktime/tests/test_softdeps.py
@@ -17,6 +17,7 @@ from importlib import import_module
 import pytest
 
 from sktime.registry import all_estimators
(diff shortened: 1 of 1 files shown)


fkiraly:
Tests are failing on two transformers that have not been refactored yet, in the `_deseasonalize` module.
This is being done here: https://github.com/alan-turing-institute/sktime/pull/3040

Issue #3037, opened 2022-07-17
[BUG] address imminent removal of deprecated `sklearn` `_pprint`


celestinoxp:
**Describe the bug**

next release 1.2.0 of scikit-learn will move pprint function and sktime will produce an error.

check this commit: https://github.com/scikit-learn/scikit-learn/commit/c63e38add5d0baa345cc363460dc0d0e18c87903

**To Reproduce**

install scikit-learn 1.2.0dev

```python
pip install --pre --extra-index https://pypi.anaconda.org/scipy-wheels-nightly/simple scikit-learn
```

**Expected behavior**

**Additional context**

**Versions**
<details>

</details>

fkiraly:
Thanks for the pointer! Should be a quick fix, since we can replace by `BaseObject` repr in all cases where the warning is raised.

fkiraly:
>I think only needs to delete line 23 "from sklearn.base import _pprint" in file "_split.py" from sktime\forecasting\model_selection\

Yes, that's one of the two instances and the easier one:
https://github.com/alan-turing-institute/sktime/pull/3048

The other instance occurs in `benchmarking.tasks` and seems less straightforward to address, since `BaseTask` does not inherit `BaseObject`.

fkiraly:
@celestinoxp, looks like we got it now. Thanks for reporting!

fkiraly:
PS: the second PR is not merged yet, so you would have to install from the branch to test it.

PR #3049, opened 2022-07-18
[MNT] remove custom `__repr__` from `BaseTask`, inherit from `BaseObject`

fkiraly:
This PR addresses imminent removal/change in `sklearn` `_pprint` and the private import in `sktime` `BaseTask`, and partially addresses #3037.

`BaseObject` has its own `__repr__` and could be used here, but `BaseTask` did not inherit from `BaseObject`.

To address the change in `__repr__`, `BaseTask` was changed to inherit from `BaseObject`, with some compatibility changes towards the `sklearn` interface. The complete removal of the custom `__repr__` in favour of `BaseObject.__repr__` should address this instance of `_pprint` import.

### DIFF
--- a/sktime/benchmarking/tasks.py
+++ b/sktime/benchmarking/tasks.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 """Unified high-level interface for various time series related learning tasks."""
 
 from inspect import signature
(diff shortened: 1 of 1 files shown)


fkiraly:
ok, well, that did not work...
any good ideas, @celestinoxp?

fkiraly:
ok, looks like this has worked.

PR #3055, opened 2022-07-19
[ENH] flexible `update` behaviour of forecasting tuners

### LINKED ISSUE #3051: [ENH] flexible `update` behaviour of forecasting tuners
Suggestion inferred from #3046.

There should be an `update_best_params: bool` parameter in the forecasting tuners, and a custom `_update` function.

The parameter `update_best_params` should determine whether the grid search is repeated in `update`, or whether only parameters of the wrapped estimator are updated.

The location to add this would be `BaseGridSearch._update`.

Update: what could even be nicer is an `update_param_grid` which allows to specify a tuning grid separate from the one used in fit. Only the parameters specified in there are re-tuned, on update.

fkiraly:
Fixes #3051, also see #3046.

Adds an `update_behaviour` parameter to grid tuners which allows the user to select which of tuning parameters and inner parameters are updated in `update` of the grid search wrapper. Also:

* cleans up docstrings which were missing parameters
* adds the new parameter to the tests

### DIFF
--- a/sktime/forecasting/model_selection/_tune.py
+++ b/sktime/forecasting/model_selection/_tune.py
@@ -220,6 +220,29 @@ class BaseGridSearch(_DelegatedForecaster):
 
         return self
 
+    def _update(self, y, X=None, update_params=True):
+        """Update time series to incremental training data.
+
+        Parameters
+        ----------
+        y : guaranteed to be of a type in self.get_tag("y_inner_mtype")
+            Time series with which to update the forecaster.
+            if self.get_tag("scitype:y")=="univariate":
+                guaranteed to have a single column/variable
+            if self.get_tag("scitype:y")=="multivariate":
+                guaranteed to have 2 or more columns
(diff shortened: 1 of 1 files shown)


fkiraly:
Tests are passing, issue is with codecov #3057

fkiraly:
Hmm, come to think of it: would it be even better, if the user could specify an `update_param_grid` different from the `param_grid` used in `fit`?

fkiraly:
well, the above is just the special case where `update_param_grid` is completely empty...

h-t-w:
> well, the above is just the special case where `update_param_grid` is completely empty...

So your saying, implementation wise it almost makes no difference, one would just need to make the `update_param_grid` public available as parameter in the `update` method? In this case: why not.

fkiraly:
>implementation wise it almost makes no difference

It would, implementing the more general version takes more time - I have, in fact, implemented only the simpler one in #3057.

PR #3058, opened 2022-07-19
[MNT] deactivate codecov (temporary)

fkiraly:
Temporary measure to fix https://github.com/alan-turing-institute/sktime/issues/3057 until we understand the problem.

Codecov is not essential for the CI and would be nice to have, so we can merge temporarily to ensure that the CI is passing again.


fkiraly:
codecov works again, this is no longer needed

PR #3059, opened 2022-07-19
[MNT] Update codecov github action from v2 to v3

miraep8:
Just trying some stuff out.  Will ping if it works.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -75,6 +75,13 @@ jobs:
       - name: Show dependencies
         run: python -m pip list
 
+      - name: Run tests
+        run: |
+          mkdir -p testdir/
+          cp .coveragerc testdir/
+          cp setup.cfg testdir/
+          python -m pytest
+
       - name: Publish code coverage
         uses: codecov/codecov-action@v3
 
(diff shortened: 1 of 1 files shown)


miraep8:
Seems like codecov is passing on simple test with v3 instead of v2.  I am adding back all the other tests I deleted just to see if the version change addressed the codecov problem, so lets see if that works

fkiraly:
it works again, but maybe we still like to change the version

miraep8:
They probably re-requested certifications! :) But yea, from what I can tell a lot of the "breaking changes" from v2 to v3 aren't really relevant to us, and seems like the packages are more up to date in v3 vs v2?

fkiraly:
> But yea, from what I can tell a lot of the "breaking changes" from v2 to v3 aren't really relevant to us, and seems like the packages are more up to date in v3 vs v2?

The latter would be an argument to upgrade. The argument to not upgrade is "don´t change a working setup", but we also know via CI that v3 is working too. Thoughts, @lmmentel, @freddyaboulton?

lmmentel (approved):

lmmentel:
Let's merge it.

Issue #2860, opened 2022-06-24
[ENH] data conversion adapters to gluonts


fkiraly:
We should write adapters to the glonts time series.

This should involve:
* conversion data frame (sktime pd.DataFrame and multiindex format) to gluonts
* conversion glounts to data frame (sktime pd.DataFrame and multiindex format)

fkiraly:
FYI @TNTran92, I opened an issue

TNTran92:
> FYI @TNTran92, I opened an issue

Thanks

TNTran92:
Added a draft PR that implement adapater from pd-multiindex to ListDataset

TNTran92:
Closed PR #2877 and opened #2893. PR is ready for review

shlok191:
@fkiraly, is this issue already taken by someone? If not, I'd be happy to begin work on this like we discussed in the mentorship meeting!

fkiraly:
@shlok191, great!

I believe the converters already exist, you'd need to add new datatypes next, and possibly test them. The adapters are in `datatypes.adapters`, `gluonts_to_pd_multiindex`, and `pd_multiindex_to_list_dataset`.

PR #3061, opened 2022-07-20
[ENH] Unit test for back converter from Gluonts to Multindex DF

TNTran92:
#### Reference Issues/PRs

Final portion of the solution to Issue #2860

Implement unit test for the back convert from Gluonts's result to multiindex DF
Use `to_timestamp()` to go from period index to timestamp index`

Dependency: gluon-ts

##### For all contributions

##### For new estimators

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,7 +35,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
-          python-version: 3.7
+          python-version: 3.9
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
@@ -50,6 +50,7 @@ jobs:
     strategy:
       matrix:
         python-version: [3.7, 3.8, 3.9, '3.10']
+
     steps:
       - uses: actions/checkout@v2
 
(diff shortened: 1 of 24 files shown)


TNTran92:
@fkiraly, I have checked and don't think it's possible to remove mxnet from gluonts. The reason is that MXNet to Gluon is like Keras to Tensorflow. 

Another reason is that, if we want to take advantage of the models of Gluonts,  we'll need Mxnet as soft dependency eventually since they are written in MXNet.

fkiraly:
> Another reason is that, if we want to take advantage of the models of Gluonts, we'll need Mxnet as soft dependency eventually since they are written in MXNet.

I suppose this is true for using `gluonts`, but `MXNet` is a modelling library, and the data container should not require it.

Could you spell out the exact dependency requirements, and explain whether/why `MXNet` is required for the data container?

TNTran92:
> > Another reason is that, if we want to take advantage of the models of Gluonts, we'll need Mxnet as soft dependency eventually since they are written in MXNet.
> 
> I suppose this is true for using `gluonts`, but `MXNet` is a modelling library, and the data container should not require it.
> 
> Could you spell out the exact dependency requirements, and explain whether/why `MXNet` is required for the data container?

There is no requirement as in `MXNet` has to be used for the data container. Even in `Gluonts`, `MXNet` is a soft dependency. I chose MXNet because most models in `Gluonts` are `MXNet`-based, so I thought it'd be convenient to use `MXNet` to produce an example of a typical estimator's output. There are other options, for example: `PyTorch`-based models and `Prophet` that can be used in the same manner. `Prophet` is already part of sktime, so maybe that is okay too.

Eventually, what I am trying to do is to get a representative sample from any Gluon estimator. Do you know of a better way this can be done?

fkiraly:
>Eventually, what I am trying to do is to get a representative sample from any Gluon estimator. Do you know of a better way this can be done?

I do not know `gluonts` well enough, but surely (hopefully?) there is a way to construct the container without the need to import any soft dependency of `gluonts`, or any modelling package at all.

What I am concerned about is the coupling via imports. Generally we want to minimize unnecessary coupling, and coupling a data container to some modelling package that doesn't even use said data container seems very wrong to me.

fkiraly:
FYI @TNTran92, @benHeid has found out how to handle soft dependencies in the `datatypes` module, see here: https://github.com/alan-turing-institute/sktime/pull/3255

TNTran92:
Thanks, I will learn (steal) his method

fkiraly:
>Thanks, I will learn (steal) his method

It's open source (permissive license) and this is a publicly visible record of credit, so it's not stealing :-)

PR #3067, opened 2022-07-21
[ENH] filter transformer from `sktime-neuro`

fkiraly:
Experimental: moves the filter transformers from `sktime-neuro` into sktime.

Changes:
* adds `mne` as a soft dependency
* now inherits from `BaseTransformer` and implements `_transform`
* the `kwargs` in `__init__` are changed to an argument, to ensure interface compliance (no init kwargs)
* the `sktime` `BaseTransformer` no longer distinguishes between series or panel input, so both `FilterForSeries` and `FilterForPanel` are simply moved to a single transformer

FYI @SveaMeyer13


fkiraly:
CI failures unrelated to the PR. See https://github.com/alan-turing-institute/sktime/issues/3103

PR #3078, opened 2022-07-23
DRAFT - KNN refactor, but passing distance as function, not precomputed - DO NOT MERGE

fkiraly:
An experimental version of https://github.com/alan-turing-institute/sktime/pull/1998, where the distance is passed to sklearn as a function (lazy evaluation), as opposed to the PR https://github.com/alan-turing-institute/sktime/pull/1998 which precomputes the distance matrix.

This is for completess of the record only, DO NOT MERGE.

The runtime behaviour of passing the distance as a function is much worse than precomputing, as it appears.


fkiraly:
With the better solution in https://github.com/alan-turing-institute/sktime/pull/1998, I think we can happily put this attempt of mine on the trash heap.

PR #3085, opened 2022-07-24
[BUG] fixes to inverse transform pipeline logic

### LINKED ISSUE #3084: [BUG] Imputer and log transformer do not work together (inverse transform not applied)
**Describe the bug**

- When the `TransformerPipeline` consists of an Imputer followed by a log transformer, predictions do not apply inverse log transform (seems to be a bug)
- When the `TransformerPipeline` consists of only a log transformer, predictions seem to be correct (inverse transform is applied correctly).

**To Reproduce**

```python


from sktime.transformations.series.boxcox import LogTransformer
from sktime.transformations.compose import TransformerPipeline
from sktime.forecasting.compose import ForecastingPipeline, TransformedTargetForecaster

preprocess = TransformerPipeline(
    [("imputer", Imputer()), ("transform", LogTransformer())]
)

pipeline = ForecastingPipeline(
    [
        (
            "forecaster",
            TransformedTargetForecaster(
                [
                    ("preprocess", preprocess),
                    ("model", NaiveForecaster()),
                ]
            ),
        )
    ]
)

y = load_airline()

# Fit Pipeline
pipeline.fit(y)

# Peedict
preds = pipeline.predict([1, 2, 3])

print(preds)
```

image

**Expected behavior**
Predictions should be around the same magnitude as the data but it looks like the inverse transform is not applied for the `LogTransformer` when `Imputer` is also present in the `TransformerPipeline`

**Additional context**
**May be important. The above code works as expected if the `Imputer` is removed from the `TransformerPipeline`**, i.e. change this line

```python
preprocess = TransformerPipeline(
    [("transform", LogTransformer())]
)
```

**Versions**
<details>

System:
    python: 3.8.13 (default, Mar 28 2022, 06:59:08) [MSC v.1916 64 bit (AMD64)]
executable: C:\Users\Nikhil\.conda\envs\pycaret_dev_sktime_13p0\python.exe
   machine: Windows-10-10.0.19044-SP0

Python dependencies:
          pip: 22.1.2
   setuptools: 61.2.0
      sklearn: 1.1.1
       sktime: 0.13.0
  statsmodels: 0.13.2
        numpy: 1.22.4
        scipy: 1.8.1
       pandas: 1.4.3
   matplotlib: 3.5.2
       joblib: 1.1.0
        numba: 0.55.2
     pmdarima: 1.8.5
      tsfresh: None

</details>

fkiraly:
This PR fixes #3084.

This fixes some inconsistent logic in chaining inverse transforms in a pipeline that has some transformers skipping the inverse.

* `TransformerPipeline` incorrectly skipped all inverses in the pipeline, if at least one element skipped its inverse. This has been fixed to the correct rule, all inverses are skipped if and only if all elements skip inverses.
* The tag `skip-inverse-transform` was only used in interfacing composites like the `ForecastingPipeline`. This is spreading/decoupling. Instead, the logic to skip was moved to `BaseTransformer`: under the presence of the tag, the `inverse_transform` behaves like the identity transform now.
* All the above connected to a third possible inconsistency in transformers, which had the `capability:inverse_transform` tag set to `False` while `skip-inverse-transform` was set to True, this would lead to contradictory behaviour, between point two (should pass through) and the capability tag behaviour (raised an error). Logic in the `TransformerPipeline` and tests has been reconciled.
* `Aggregator` had the tags set inconsistently and was fixed.

Adds test case which is a simplified version of #3084.


sktime/transformations/compose.py
```diff
@@ -171,7 +171,7 @@ def __init__(self, steps):
         self._anytagis_then_set("scitype:instancewise", False, True, ests)
         self._anytagis_then_set("fit_is_empty", False, True, ests)
         self._anytagis_then_set("transform-returns-same-time-index", False, True, ests)
-        self._anytagis_then_set("skip-inverse-transform", True, False, ests)
+        self._anytagis_then_set("skip-inverse-transform", False, True, ests)
```
fkiraly, line 174: for reviewers: old = "skip inverse iff at least one component skips the inverse" -> new: "skip inverse iff all components skip the inverse"

sktime/transformations/compose.py
```diff
@@ -171,8 +171,13 @@ def __init__(self, steps):
         self._anytagis_then_set("scitype:instancewise", False, True, ests)
         self._anytagis_then_set("fit_is_empty", False, True, ests)
         self._anytagis_then_set("transform-returns-same-time-index", False, True, ests)
-        self._anytagis_then_set("skip-inverse-transform", True, False, ests)
-        self._anytagis_then_set("capability:inverse_transform", False, True, ests)
+        self._anytagis_then_set("skip-inverse-transform", False, True, ests)
+
+        # self can inverse transform if for all est, we either skip or can inv-trasform
```
fkiraly, line 176: for reviewers: we consider the pipeline to be able to inverse transform if each components can either inverse transform, or are skipped

Issue #3065, opened 2022-07-21
[ENH] Implement update method for AutoETS


h-t-w:
In relation to #3063 and #3064.

`AutoETS` - as a forecaster with autotuning capabilities such as `AutoARIMA` and `TBATS` - should ideally have a dedicated `update` method, which allows updating the model coefficients, without retuning the model hyperparameters when `AutoETS(auto=True)`. This would enable the user to separate hyperparameter tuning from updating/fitting the best forecaster selected by `AutoETS`. Default as of now should be full refit.

fkiraly:
This is a native implementation, so I suppose someone needs to dive into the code...

Let's ask the original author:
@HYang1996, are you still around?

HYang1996:
hey @fkiraly, unfortunately I'm no longer actively contributing to `sktime` at the moment. This implementation was adapted from the [fable version of R](https://github.com/tidyverts/fable/blob/master/R/ets.R), which doesn't seem to contain such a method, I guess the `update` method will have to be written from scratch or be adapted from other sources.

fkiraly:
>However, this implementation was wrapped around the [`ETSModel`](https://www.statsmodels.org/devel/generated/statsmodels.tsa.exponential_smoothing.ets.ETSModel.html) from `statsmodels` which does offer an `update` option, so I guess that would work?

Yes, that would! Where does that sit, in the class?

HYang1996:
According to the [example notebook](https://github.com/alan-turing-institute/sktime/blob/5767dbaac759ed6ac43a8977f4c46beb93d0f18c/examples/01_forecasting.ipynb), the `update` method of `AutoETS` class seems to be updating the forecaster with the new data. To access the wrapped `ETSModel`, one could do the following:
```python
# December 1957

# this is the data known in December 1957
y_1957Dec = y[:-36]

# step 1: specifying the forecasting strategy
forecaster = AutoETS(auto=True, sp=12, n_jobs=-1)

# step 2: specifying the forecasting horizon: one year ahead, all months
fh = np.arange(1, 13)

# step 3: this is the first time we use the model, so we fit it
forecaster.fit(y_1957Dec)

# step 4: obtaining the first batch of forecasts for Jan 1958 - Dec 1958
y_pred_1957Dec = forecaster.predict(fh)

# Get ETSModel object, which is `self._forecaster` in `AutoETS`
ETSModel_object = forecaster._forecaster
```
The `update` method of `ETSModel` should then be accessible.

[...]

Issue #3064, opened 2022-07-21
[ENH] Implement update method for AutoARIMA


h-t-w:
As discussed in #3046, the `AutoARIMA` forecaster should ideally have a dedicated `update` method, which enables updates of coefficients of the best model, without fully refitting the forecaster/re-tuning the hyperparameters (e.g. p,d,q). A full refit - as of now - should remain the default.

The `AutoARIMA` class in the `pmdarima` package does have such an `update` method, which would in turn need to be referenced in the `sktime` wrapper. As far as I can tell this would require adding a `_update` method to the `_PmdArimaAdapter` class. As discussed with @fkiraly `update_pdq: boolean` should control re-fit/update.

fkiraly:
`pmdarima` has an `update` interface, so maybe it's really easy? https://github.com/alan-turing-institute/sktime/pull/3068

fkiraly:
Yes, #3068 seems to have worked.

PR #3086, opened 2022-07-24
[ENH] dedicated `_update` for `BATS` and `TBATS`

### LINKED ISSUE #3063: [ENH] Implement update method for TBATS
Want to perform walk-forward predictions using TBATS model. without refitting parameters every time. This is crucial for validation, especially since TBATS model is slow to train. 

Implement update method for TBATS... same as interface used for other time series forecasters in sktime.

jelc53:
…r BATS and TBATS wrappers

#### Reference Issues/PRs
Fixes #3063. See also https://github.com/intive-DataScience/tbats/issues/35 for discussion. 

Implements update method for tbats, which allows users to update state fo the model with new observations (fast) without refitting model parameters (slow). Particualrly, useful for walk-forward prediction and validation.

No new dependencies were introduced.

There are a number of potential places I could implement this change. I thought the _tbats.py file with the adapter class was the most natural but let me know if this was not the intended design / pattern. 

#### Any other comments?
This is my first contribution to a public code base. Would appreciate any feedback and, if happy, would be interested in contributing more (eg. related issues #3064 and #3065).

### DIFF
--- a/sktime/forecasting/base/adapters/_tbats.py
+++ b/sktime/forecasting/base/adapters/_tbats.py
@@ -124,12 +124,12 @@ class _TbatsAdapter(BaseForecaster):
         """
         if update_params:
             # update model state and refit parameters
-            self._fit(y=self._y)  # _fit re-runs model instantiation 
-            
+            self._fit(y=self._y)  # _fit re-runs model instantiation
+
         else:
             # update model state without refitting parameters
             self._forecaster.fit(y=self._y)
-            
+
         return self
 
     def _predict(self, fh, X=None):
--- a/sktime/forecasting/base/adapters/_tbats.py
+++ b/sktime/forecasting/base/adapters/_tbats.py
@@ -102,10 +102,39 @@ class _TbatsAdapter(BaseForecaster):
         self._yname = y.name
 
         return self
+    
+    def _update(self, y, X=None, update_params=True):
+        """Update time series to incremental training data.
 
-    def _predict(self, fh, X):
+        Derived from example script provided by core devs in TBATS repository
+        https://github.com/intive-DataScience/tbats/blob/master/examples/re_fit_model.py
+
+        Parameters
+        ----------
+        y : pd.Series
+            Target time series to which to fit the forecaster.
+        X : pd.DataFrame, optional (default=None)
+            Exogenous variables (ignored)
+        update_params : bool, optional (default=True)
+            whether model parameters should be updated
+
+        Returns
+        -------
+        self : reference to self
+        """
+        if update_params:
+            # update model state and refit parameters
+            self._fit(y=self._y)  # _fit re-runs model instantiation script 
+            
+        else:
+            # update model state without refitting parameters
+            self._forecaster.fit(y=self._y)
+            
+        return self
+
+    def _predict(self, fh, X=None):
         """Forecast time series at future horizon.
-
+        
         Parameters
         ----------
         fh : int, list, np.array or ForecastingHorizon
--- a/sktime/forecasting/base/adapters/_tbats.py
+++ b/sktime/forecasting/base/adapters/_tbats.py
@@ -129,14 +129,14 @@ class _TbatsAdapter(BaseForecaster):
 
         else:
             # update model state without refitting parameters
-            # out-of-box fit tbats method will not refit parameters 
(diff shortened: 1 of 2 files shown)


sktime/forecasting/base/adapters/_tbats.py
```diff
@@ -102,10 +102,39 @@ def _fit(self, y, X=None, fh=None):
...
+        X : pd.DataFrame, optional (default=None)
+            Exogenous variables (ignored)
+        update_params : bool, optional (default=True)
+            whether model parameters should be updated
 
+        Returns
+        -------
+        self : reference to self
+        """
+        if update_params:
```
fkiraly, line 125: I think instead this should be as follows:

`if update_params`, then run `self._forecaster.fit`.

Else, *only return self*.

Why: `update_params=False` means data is ingested, but the internal models are *unchanged*.

fkiraly (changes requested):
Great, thanks!

Two change requests:
* code formatting (linting) fails, kindly check "details" with the first CI element. It is also a good idea to set up linting checks locally.
* `update_params=False` mode should not change any self attributes. This mode is used when only data is ingested, but nothing whatsoever is updated.

sktime/forecasting/base/adapters/_tbats.py
jelc53: I think this is what happens by default (update model state only, not params) with the TBATS fit method if the model is not re-instantiated. 

A bit weird I know, so maybe I am misinterpretting. See example script here: https://github.com/intive-DataScience/tbats/blob/master/examples/re_fit_model.py. I was able to reproduce something similar for my own project.

jelc53: When ``update_params=False`` my understanding is that we want to change two things only: (a) self._y, (b) cutoff index within TBATS model class. We handle (a) I think already in the ``BaseForecaster`` class ``update`` method by calling ``_update_y_X`` but I think need to call TBAT's fit method for this to fix the cuttoff such that we can access other TBATS methods like forecast, etc. after the data update.

fkiraly: I see, thanks for explaining!

Then this is ok.

fkiraly:
linting still failing, please click here to see what´s wrong:

image

jelc53:
Oh gotcha, looks like still a trailing white space issue. Will fix when back at my laptop.

fkiraly:
Linting still failing 😁 

Consider it a fun scavenger hunt 😄

jelc53:
Damn, aplogoies for the back and forth. I've tried again (shifting comments to not be end-of-line, found whitespace on an unused line between methods), but not sure how to find and test the linting error systematically without workflow run permissions. Let me know if there is a way I can easily recreate linting requirements locally if it fails again.

fkiraly:
>Let me know if there is a way I can easily recreate linting requirements locally if it fails again.

Sure! The easiest way is installing precommit:
https://www.sktime.org/en/stable/developer_guide/continuous_integration.html

If you want to avoid automated changes to your files, you can alternatively install `black` and `flake8` and activate them in your developer IDE (e.g., in vs code you need to enable them in preferences) to get problems listed.

fkiraly:
I expanded the docs on setting up local code quality checks, feel free to look into the expanded version of the docs:
https://github.com/alan-turing-institute/sktime/pull/3089

jelc53:
Setup ``black`` and ``flake8`` on vscode locally, plus installed ``pre-commit`` so should have the linting part resolved
image

fkiraly:
>Setup ``black`` and ``flake8`` on vscode locally, plus installed ``pre-commit`` so should have the linting part resolved

Indeed! Now we can look at the "serious" tests 😁 

Would also appreciated feedback what you think about the improved documentation around linting in #3089 that I wrote on this occasion. Is that more clear and helpful? Can it be improved?

fkiraly (approved):
Nice! Tests pass! Thanks!

jelc53:
Great to hear! Appreciate you helping me through it. Re feedback on #3089, the pre-commit setup made life much easier for me so I think that’ll be helpful for others too. Only question for me is whether you also make clear that another option is to configure the ide to flag black and flake8 conventions for those who don’t want to use pre-commit. Not sure. Might be one of those things where more info is more confusing than helpful.

PR #3093, opened 2022-07-25
[DOC] Added NumFOCUS to sponsors website

aiwalter:
Added Logo and mentioning of NumFOCUS to website


fkiraly (changes requested):
Why is the GSoC image added, if it is not referenced by the new addition?

aiwalter:
I changed it to be a .png picture, so no white background. We should do that for the other logos as well actually, if we can get .png

fkiraly (approved):
Makes sense - I just got suspicious of a potential error since it wasn't mentioned in the PR description.

PR #3102, opened 2022-07-26
[ENH] refactor `RocketClassifier` to pipeline delegate

fkiraly:
Many classifiers in the classification module are just "special names" for simple [transformer][classifier] pipelines.

This is the first PR in an intended refactor of the classification module to expose those "special classifier pipeliens" as shorthand delegates for said pipelines, in the form of `ClassifierPipeline`.

This PR:
* adds `_DelegatedClassifier`, similar to `_DelegatedForecaster` and `_DelegatedTransformer`
* refactors `RocketClassifier` as a specific `_DelegatedClassifier`

Minor changes:
docstrings in the extension template for classifier seemed out of date, these were fixed.

Relies on https://github.com/alan-turing-institute/sktime/pull/3104 which fixes unreported bugs in `predict_proba` defaulting of classifiers that this has uncovered.

Related PR: https://github.com/alan-turing-institute/sktime/pull/767
(was going through old PRs and seeing what could be done easily)

Related issue: https://github.com/alan-turing-institute/sktime/issues/2603
(on delegator design - I notice the delegator class for regression/classification could be identical, if it weren't for docstrings!)

FYI @MatthewMiddlehurst, @victordremov (re `RocketClassifier`) and @miraep8 (re delegator)

### DIFF
--- a/sktime/classification/base.py
+++ b/sktime/classification/base.py
@@ -170,6 +170,7 @@ class BaseClassifier(BaseEstimator, ABC):
 
         self.classes_ = np.unique(y)
         self.n_classes_ = self.classes_.shape[0]
+        self._X_metadata = X_metadata
         self._class_dictionary = {}
         for index, class_val in enumerate(self.classes_):
             self._class_dictionary[class_val] = index
--- a/sktime/classification/base.py
+++ b/sktime/classification/base.py
@@ -524,9 +524,10 @@ class BaseClassifier(BaseEstimator, ABC):
             2nd dimension indices correspond to possible labels (integers)
             (i, j)-th entry is predictive probability that i-th instance is of class j
         """
-        dists = np.zeros((X.shape[0], self.n_classes_))
         preds = self._predict(X)
-        for i in range(0, X.shape[0]):
+        n_pred = len(preds)
+        dists = np.zeros((n_pred, self.n_classes_))
+        for i in range(n_pred):
             dists[i, self._class_dictionary[preds[i]]] = 1
 
         return dists
--- a/sktime/classification/base.py
+++ b/sktime/classification/base.py
@@ -67,11 +67,13 @@ class BaseClassifier(BaseEstimator, ABC):
     }
 
     def __init__(self):
-        self.classes_ = []
-        self.n_classes_ = 0
-        self.fit_time_ = 0
+        # reserved attributes written to in fit
+        self.classes_ = []  # classes seen in y, unique labels
+        self.n_classes_ = 0  # number of unique classes in y
+        self.fit_time_ = 0  # time elapsed in last fit call
         self._class_dictionary = {}
         self._threads_to_use = 1
+        self._X_metadata = []  # metadata/properties of X seen in fit
 
         # required for compatability with some sklearn interfaces
         # i.e. CalibratedClassifierCV
--- a/sktime/classification/kernel_based/_rocket_classifier.py
+++ b/sktime/classification/kernel_based/_rocket_classifier.py
@@ -142,41 +142,27 @@ class RocketClassifier(_DelegatedClassifier):
                 f"{self.VALID_MULTIVAR_VALUES}, but found {use_multivariate}"
             )
 
+        common_params = {
+            "num_kernels": self.num_kernels,
+            "random_state": self.random_state,
+            "max_dilations_per_kernel": self.max_dilations_per_kernel,
+            "n_jobs": self._threads_to_use,
+        }
+
         if rocket_transform == "rocket":
-            univar_rocket = Rocket(
-                num_kernels=self.num_kernels,
-                random_state=self.random_state,
-                n_jobs=self._threads_to_use,
-            )
+            del common_params["max_dilations_per_kernel"]
+            univar_rocket = Rocket(**common_params)
             multivar_rocket = univar_rocket
+
         elif rocket_transform == "minirocket":
-            multivar_rocket = MiniRocketMultivariate(
-                num_kernels=self.num_kernels,
-                max_dilations_per_kernel=self.max_dilations_per_kernel,
-                random_state=self.random_state,
-                n_jobs=self._threads_to_use,
-            )
-            univar_rocket = MiniRocket(
-                num_kernels=self.num_kernels,
-                max_dilations_per_kernel=self.max_dilations_per_kernel,
-                random_state=self.random_state,
-                n_jobs=self._threads_to_use,
-            )
+            multivar_rocket = MiniRocketMultivariate(**common_params)
+            univar_rocket = MiniRocket(**common_params)
+
         elif self.rocket_transform == "multirocket":
-            multivar_rocket = MultiRocketMultivariate(
-                num_kernels=self.num_kernels,
-                max_dilations_per_kernel=self.max_dilations_per_kernel,
-                n_features_per_kernel=self.n_features_per_kernel,
-                random_state=self.random_state,
-                n_jobs=self._threads_to_use,
-            )
-            univar_rocket = MultiRocket(
-                num_kernels=self.num_kernels,
-                max_dilations_per_kernel=self.max_dilations_per_kernel,
-                n_features_per_kernel=self.n_features_per_kernel,
-                random_state=self.random_state,
-                n_jobs=self._threads_to_use,
-            )
+            common_params["n_features_per_kernel"] = self.n_features_per_kernel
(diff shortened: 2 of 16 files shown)


fkiraly:
PS: this is not an exact refactor - instead of selecting the transformer depending on number of `X` columns (multivariate/univariate), this is now selected as a constructor argument `use_multivariate`.

If an exact refactor is desired, this could be achieved by overriding `_get_delegate` by logic conditional on the number of dimensions seem by `fit`, and delegating to one of two delegates (univariate/multivariate).

But, I think that is not desirable, since that does not leave the user with a choice in the univariate case?

fkiraly:
@miraep8, I think this PR establishes an interesting connection: delegators may be similar across tasks - but base classes and other compositors may also be. I am wondering, is there a good way to attach the "right base class" dynamically to a composite?

E.g., delegate all methods, an then inherit from the right base class? Seems impossible to me with native python, but there might be a way around.
For instance, a multiplexer base class, that delegates to all base classes, according to a tag?

miraep8:
Yes! I totally agree that this is an interesting connection/motivation, and I would say is connected with some of the discussion we have had with the pywatts people about whether or not there is any value in "considering everything a graph"  (which personally I think there is)!

Several thoughts with this - I think that trying to dynamically set the base class is perhaps messy...  I think it is possible that it *might* be able to be done, but I personally am not sure it is the best line of evidence.

One way that you could imagine doing this is having a very simple base class that simply tests the types of objects it wraps /has been passed and applies functionality appropriately. - ie do I have a forecaster in my sequence of estimators? ok then I can behave as a forecaster, if not raise an error.  Same for annotators. 

Personally though I think this is further evidence that there is merit in a very simple BaseEstimator/Object where additional functionality can be added on via composition!  But of course I am biased in favor of that solution/would be interested in seeing if anyone else had a better solution in mind! :)

miraep8:
And yes, on this note to some degree I think that some of the issues that came up in the discussion about a more general Multiplexer/a more general Delegator are related to the design issues that were brought up here as well/the current templating design.

MatthewMiddlehurst:
I am not really a fan of including an option for multivariate series, we should be able to infer that from the input and process accordingly, rather than have the user tell us which it is (this also applies to the transform itself, which for some reason is two separate ones).

I understand this may be preferable from our point of view, but it seems like a downgrade for actual usage.

sktime/classification/kernel_based/_rocket_classifier.py
```diff
@@ -34,11 +34,13 @@ class RocketClassifier(BaseClassifier):
...
-        Valid inputs = ["rocket","minirocket","multirocket"]
-    max_dilations_per_kernel : int, default=32
+        Valid inputs = ["rocket", "minirocket", "multirocket"]
+    max_dilations_per_kernel : int, optional, default=32
         MiniRocket and MultiRocket only. The maximum number of dilations per kernel.
-    n_features_per_kernel : int, default=4
+    n_features_per_kernel : int, optional, default=4
         MultiRocket only. The number of features per kernel.
+    use_multivariate : bool, optional, default=True
+        whether to use multivariate rocket transforms (True) or univariate ones (False)
```
MatthewMiddlehurst, line 43: This one is minor, but what makes the bottom two parameters optional and the first not?

fkiraly: nothing, just an oversight. All with defaults should have "optional" in docstr

fkiraly: fixed

fkiraly:
> I am not really a fan of including an option for multivariate series, we should be able to infer that from the input and process accordingly, rather than have the user tell us which it is (this also applies to the transform itself, which for some reason is two separate ones).

> I understand this may be preferable from our point of view, but it seems like a downgrade for actual usage.

Is the "multivariate" transformer not behaving the same as the univariate version, when presented with univariate data?

If yes, I see no need to have the univariate version at all.

If no, why is it preferable to not give the user a choice which one they pick, in the univariate case?

MatthewMiddlehurst:
Honestly, I have no idea. I have worked under the assumption the multivariate version will break if given univariate series, because otherwise why would there be two?

- If the output is the same, then I agree, why split it up.
- If the output is not the same, I can very easily see someone overlooking the option and either crashing or obtaining wrong/strange results (second is worse IMO). 

I think at this point, it is standard for multivariate capable sktime classifiers and transformers to be able to infer the number of dimensions. Unless I am unaware of some other case like this, I don't think we should break that.

This is more of a rocket transformer issue than a problem with the delegation stuff.

fkiraly:
>Honestly, I have no idea. I have worked under the assumption the multivariate version will break if given univariate series, because otherwise why would there be two?

The tests pass with a default of `use_multivariate=True`, so not only does it not break but seems to produce the same results as before (if you put any in the results registry).

> or obtaining wrong/strange results (second is worse IMO).

Is that not excluded because the tests pass? I thought you check for consistency of predictive performance and individual predictions even.

> This is more of a rocket transformer issue than a problem with the delegation stuff.

I can code it easily either way, but it seemed strange/unnecessary to me.

MatthewMiddlehurst:
If it works, it works I guess. RocketClassifier doesnt produce any probabilities, but all the predictions are the same and it didnt crash if the test passed.

May be worth adding a note on the parameter? I didnt like the split transformer that much previously, but like it even less now 🙂. I cant imagine it is that much more efficient having a dedicated univariate version.

fkiraly:
> May be worth adding a note on the parameter? 

done that now

> I cant imagine it is that much more efficient having a dedicated univariate version.

I thought you knew what is going on there.

MatthewMiddlehurst:
No, the transformers were implemented by the author, I believe. Up until now, I thought it would just crash if you gave it the wrong input 🤷.

fkiraly:
but we were already testing *all* transformers with univariate input!

MatthewMiddlehurst:
If not crash, then at least produce a different output. Why have two versions if it's just the same, Never really thought about it too much, I guess.

I'm pretty sure rocket just selects random dimensions per kernel for multivariate, I can't see there being a massive cost to having it select the only dimension over performing no selection.

fkiraly:
so, what do we do for the purpose of this PR?

MatthewMiddlehurst:
Up to you, I still think it is worth inferring the correct transformer on the off chance there is a significant efficiency difference or the test did not catch a difference between the two.

Not going to block it though.

fkiraly:
ok, I updated it now. `use_multivariate` now can be `"auto"`, which is default and replicates current behaviour. This is, therefore a 1:1 refactor with features added.

fkiraly:
would appreciate a review after all the comments...

sktime/classification/base.py
```diff
@@ -170,6 +170,7 @@ def fit(self, X, y):
 
         self.classes_ = np.unique(y)
         self.n_classes_ = self.classes_.shape[0]
+        self._X_metadata = X_metadata
```
MatthewMiddlehurst, line 173: Probably want to add this to _init_ if we are writing to it in every fit.

MatthewMiddlehurst (dismissed):

sktime/classification/base.py
fkiraly: doesn't work in `__init__`, because back then we won't have seen any data yet

fkiraly:
merging with approval. mac is currently erroring out unrelated to this PR.

sktime/classification/base.py
MatthewMiddlehurst: we dont know the classes prior to fit, but we have self.classes_ = [] in __init__

MatthewMiddlehurst: unless self._X_metadata is used anywhere prior to fit which i am missing

fkiraly: what precisely are you suggesting? `X_metadata` is available earliest in `fit`, so we cannot write it in `__init__`

MatthewMiddlehurst: `self._X_metadata = []`? 

This is what happens for every other variable we write to in fit im pretty sure

fkiraly: can add this, but it wouldn't change anything, sure, if that's convention

MatthewMiddlehurst: I think its more just to indicate that it exists and will be written to prior to calling fit, very minor but it is convention currently.

fkiraly: done

PR #3104, opened 2022-07-26
[BUG] fixes to `BaseClassifier._predict_proba` default and `SklearnClassifierPipeline` in case `predict_proba` is not implemented

fkiraly:
This fixes some unreported bugs with `BaseClassifier` default methods and `SklearnClassifierPipeline` for `predict_proba`.

* `BaseClassifier._predict_proba` worked only for certain mtypes, since `n_instances` was obtained as `len(X)` which is correct only for some mtypes. This has been replaced by `len(y)` which is always the same as `n_instances`.
* `SklearnClassifierPipeline._predict_proba` would break if the `sklearn` classifier interfaced had no `predict_proba`. This was fixed by defaulting to the `BaseClassifier._predict_proba`.
* `SklearnClassifierPipeline` had the `capability:multivariate` tag set exactly the wrong way, a `not` was missing. This did not cause a breakage but a lot of incorrect warnings.

### DIFF
--- a/sktime/classification/base.py
+++ b/sktime/classification/base.py
@@ -524,9 +524,10 @@ class BaseClassifier(BaseEstimator, ABC):
             2nd dimension indices correspond to possible labels (integers)
             (i, j)-th entry is predictive probability that i-th instance is of class j
         """
-        dists = np.zeros((X.shape[0], self.n_classes_))
         preds = self._predict(X)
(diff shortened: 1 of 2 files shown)


fkiraly:
approved as part of https://github.com/alan-turing-institute/sktime/pull/3102

PR #3105, opened 2022-07-26
[ENH] Dynamic factor predict_interval and predict_quantiles

lbventura:
#### Reference Issues/PRs

Implements predict_interval, which was missing in the original PR [#2859](https://github.com/alan-turing-institute/sktime/pull/2859). 

#### What does this implement/fix? Explain your changes.

Implements the predict_interval method, using the get_forecast and conf_int methods predefined in the statsmodels package. This is tested explicitly in the ipynb.

#### What should a reviewer concentrate their feedback on?

As it can be seen in the "Practical Example" section of the ipynb, `predict_interval` behaves as expected and generates predictions for the two time-series. However, the derived method `predict_quantiles` only generates one. EDIT (29-07-22): This has now been corrected as per #3106.

##### For all contributions
- [x] I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/main/.all-contributorsrc).
- [ ] Optionally, I've updated sktime's [CODEOWNERS](https://github.com/alan-turing-institute/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
- [ ] I've added unit tests and made sure they pass locally.
- [x] The PR title starts with either [ENH], [MNT], [DOC], or [BUG] indicating whether the PR topic is related to enhancement, maintenance, documentation, or bug.

##### For new estimators
- [ ] I've added the estimator to the online documentation.
- [ ] I've updated the existing example notebooks or provided a new one to showcase how my estimator works.

PR #3106, opened 2022-07-26
[BUG] fix forecaster default `predict_quantiles` for multivariate data

fkiraly:
This fixes a bug reported by @lbventura in https://github.com/alan-turing-institute/sktime/pull/3105: the default implementation of `_predict_quantiles`, when `_predict_interval` is implemented, would return quantiles only for the first variable - which is fine in univariate forecasting, but produces too few columns in multivariate forecasting.

This has now been fixed, the correct columns are now returned in both cases, under the condition that `_predict_quantiles` is implemented but `_predict_interval` is not.

Odd that this was not tested so far, I still need to investigate why.

To the bets of my knowledge, we should be testing the proba predict methods with multivariate scenarios?

### DIFF
--- a/sktime/forecasting/base/_base.py
+++ b/sktime/forecasting/base/_base.py
@@ -1911,8 +1911,12 @@ class BaseForecaster(BaseEstimator):
             # now we need to subset to lower/upper depending
             #   on whether alpha was < 0.5 or >= 0.5
             #   this formula gives the integer column indices giving lower/upper
-            col_selector = (np.array(alpha) >= 0.5) + 2 * np.arange(len(alpha))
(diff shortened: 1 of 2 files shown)


fkiraly:
Reason why tests did not catch this:

vectorization takes priority over the default fill in, so forecasters that are not genuinely multivariate will call the `_predict_interval` from `_predict_quantiles` via vectorization, which is fine.

The problem with the logic therefore occurs only for genuinely multivariate, probabilistic forecasters, of which there were, so far, none in the code base.

Somewhat surprising since `VAR` is in the code base, but that does not have probabilistic prediction implemented.

fkiraly:
Included as part of #3105 and #2925

Issue #3103, opened 2022-07-26
[BUG] memory related test failures post CI simplification PR


fkiraly:
Since we merged the "CI simplification" PR https://github.com/alan-turing-institute/sktime/pull/2896, I have observed a number of mem-out failures of tests, in the OS/version combinations that the CI simplification has addressed.

I cannot claim a causal relation, but the timing makes me suspicious.

Examples of mem-outs:
https://github.com/alan-turing-institute/sktime/runs/7524247553
https://github.com/alan-turing-institute/sktime/runs/7522495684

Any good ideas, @lmmentel?
Maybe @freddyaboulton, @RNKuhns have an idea?

fkiraly:
FYI, #3095 is purely doc related (code of conduct).

While the proposed changes might be contentious to some, it is hard to imagine any plausible technical reason for why the changes would lead to failure in the classification module...

fkiraly:
One more, also on mac: https://github.com/alan-turing-institute/sktime/runs/7523631638

All classification related, maybe that's another possible lead? @MatthewMiddlehurst ?

lmmentel:
Mac failures might be related to macos-10.15 being deprecated https://github.com/actions/virtual-environments/issues/5583 I'll make a PR to bump the OS version of the CI host.

MatthewMiddlehurst:
This seems to be affecting #3090 as well for a range of classifiers and transformers. Most of the failures have been untouched for quite a while as far as I am aware.

fkiraly:
I have an explanation for 14 workers, @lmmentel: they are not concurrent, but sequential. One after another, the workers crash, so `pytest-xdist` spins up new ones in an attempt to try again.

fkiraly:
Revert of recent CI changes seems to fix the issues, see #3118.

The recent changes that are reverted consist of three things:

* testing notebooks on py 3.9, not 3.7, by @fkiraly 
* codecov-action changed from version 2 to version 3, by @miraep8 
* merging separate linux and mac CI elements into a single matrix, by @lmmentel

My suspicion is towards the merge of linux/max since it appears as the most substantial change, but we cannot rule anything out at the moment imo.

fkiraly:
(@lmmentel, @miraep8, not trying to play the blame game here - after all I've approved the PR, and all looked very sensible to me too!)

miraep8:
[...]

PR #3107, opened 2022-07-26
[MNT] Bump macos github actions host to macos-11

lmmentel:
Bumping the CI host/runner for macos from 10.15 to 11 since github is going to deprecate macos-10.15 soon. 

#### Reference Issues/PRs

see: https://github.com/actions/virtual-environments/issues/5583

ongoing migration work might be related to some issues with macos referenced in #3103


fkiraly (approved):
The 10.15 macOS tests no longer seem to run, so I suppose this was a good call.

It does not resolve the HIVE-COTE memory failures though?

PR #3108, opened 2022-07-26
DRAFT [ENH] experimental polymorphic base class

fkiraly:
Experimentation with polymorphic base class to be used in polymorphic delegator, see discussion in https://github.com/alan-turing-institute/sktime/issues/2603

Two approaches are attempted:

* polymorphism by copying methods
* polymorphism by delegation

Does not seem to work yet. Properties and other decorated attributes seem to pose a problem to these naive approaches.

### DIFF
--- a/sktime/base/_poly.py
+++ b/sktime/base/_poly.py
@@ -1,0 +1,300 @@
+#!/usr/bin/env python3 -u
+# -*- coding: utf-8 -*-
+# copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+"""Implements polymorphic base class."""
+
+__author__ = ["fkiraly", "miraep8"]
+__all__ = ["BasePolymorph"]
+
+
+from sktime.base import BaseObject
+from sktime.registry._lookup import _check_estimator_types
+
+
+class BasePolymorph(BaseObject):
+    """Handles parameter management for estimators composed of named estimators.
+
+    Partly adapted from sklearn utils.metaestimator.py.
+    """
+
+    def __init__(self, estimator_type="base"):
+
+        self.estimator_type = estimator_type
+        baseclass = _check_estimator_types(estimator_type)[0]
+
+        super(BasePolymorph, self).__init__()
+
+        attr_to_copy = set(dir(baseclass)).difference(dir(self))
+
+        for attr in attr_to_copy:
+            setattr(self, attr, getattr(baseclass, attr))
+
+        # baseclass.__init__(self)
+
+
+def _delegate(func):
+
+    def inner(self, **kwargs):
+        return func(self._delegate, **kwargs)
+
+    return inner
+
+
+class BasePolymorphByDelegation(BaseObject):
+    """Handles parameter management for estimators composed of named estimators.
+
+    Partly adapted from sklearn utils.metaestimator.py.
+    """
+
+    def __init__(self, estimator_type="base"):
+
+        self.estimator_type = estimator_type
+        baseclass = _check_estimator_types(estimator_type)[0]
+        self._delegate = baseclass
+
+        super(BasePolymorphByDelegation, self).__init__()
+
+        attr_to_copy = set(dir(baseclass)).difference(dir(self))
+
+        print(attr_to_copy)
+
+        for attr in attr_to_copy:
+            setattr(self, attr, _delegate(getattr(baseclass, attr)))
+
+        # baseclass.__init__(self)
+
+
+
+
+import numpy as np
+from sklearn.linear_model import RidgeClassifierCV
+from sklearn.pipeline import make_pipeline
+from sklearn.preprocessing import StandardScaler
+
+from sktime.transformations.panel.rocket import (
+    MiniRocket,
+    MiniRocketMultivariate,
+    MultiRocket,
+    MultiRocketMultivariate,
+    Rocket,
+)
+
+
+class RocketClassifier(BasePolymorphByDelegation):
+    """Classifier wrapped for the Rocket transformer using RidgeClassifierCV.
+
+    This classifier simply transforms the input data using the Rocket [1]_
+    transformer and builds a RidgeClassifierCV estimator using the transformed data.
+
+    Parameters
+    ----------
+    num_kernels : int, default=10,000
+        The number of kernels the for Rocket transform.
+    rocket_transform : str, default="rocket"
+        The type of Rocket transformer to use.
+        Valid inputs = ["rocket","minirocket","multirocket"]
+    max_dilations_per_kernel : int, default=32
+        MiniRocket and MultiRocket only. The maximum number of dilations per kernel.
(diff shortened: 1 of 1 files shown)


sktime/base/_poly.py
```diff
@@ -0,0 +1,300 @@
+#!/usr/bin/env python3 -u
+# -*- coding: utf-8 -*-
+# copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+"""Implements polymorphic base class."""
+
+__author__ = ["fkiraly", "miraep8"]
```
miraep8, line 6: No need to make me an author :)  Thanks for the thought though, I can add myself later if I contribute code.

sktime/base/_poly.py
```diff
@@ -0,0 +1,300 @@
...
+class BasePolymorph(BaseObject):
+    """Handles parameter management for estimators composed of named estimators.
+
+    Partly adapted from sklearn utils.metaestimator.py.
+    """
+
+    def __init__(self, estimator_type="base"):
+
+        self.estimator_type = estimator_type
+        baseclass = _check_estimator_types(estimator_type)[0]
```
miraep8, line 23: I still think copying attributes from the first class in the polymorphism may not be the way to go?  ie - I think it would be nice to have something like `BasePolymorph` work for multiple types of composed estimators, and by enforcing this estimator to look like the first estimator provided may cause some issues/make that goal harder to realize.

fkiraly: there is just one class? if you mean the `[0]`, this is since `_check_estimator_type` always coerces to list, but in this case it always is a one-element list.

sktime/base/_poly.py
```diff
@@ -0,0 +1,300 @@
...
+
+        attr_to_copy = set(dir(baseclass)).difference(dir(self))
+
+        for attr in attr_to_copy:
+            setattr(self, attr, getattr(baseclass, attr))
+
+        # baseclass.__init__(self)
+
+
+def _delegate(func):
```
miraep8, line 35: I am thinking for this to work you would also need to pass/take self?  Something that might be worth trying actually could simply be getting the function from the delegated class using `get_attr` (functions are also attributes after all!) This could be cleaner/easier to do via decorators?  (ie functions that should delegated should be decorated!).  But this is just one way to achieve this.

fkiraly: I was thinking repeatedly that `_delegate` may need to take `self`, but *does it*? `func` has `self` as an argument, so all is fine, I thought. Then, it suffices that `inner` has `self` as arg.

I don't entirely see how it would look like with `get_attr`?

Further, the issue that remains unsolved is properties.

sktime/base/_poly.py
miraep8: Ah you are correct, I had read this a bit too quickly, and thought this code was trying to do something different than it is.

sktime/base/_poly.py
miraep8: Sure, if it works it works.

fkiraly: I don't know whether it works because I'm stuck on the properties

sktime/base/_poly.py
```diff
@@ -0,0 +1,300 @@
...
+            Reference to self.
+
+        Notes
+        -----
+        Changes state by creating a fitted model that updates attributes
+        ending in "_" and sets is_fitted flag to True.
+        """
+        _, n_dims, _ = X.shape
+
+        if self.rocket_transform == "rocket":
```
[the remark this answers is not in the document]
fkiraly: oh, I picked an estimator at random and then saw the same thing as you do, I could not leave it like that...

Here is the refactor: https://github.com/alan-turing-institute/sktime/pull/3102

The purpose of the "rocket" here is just an arbitrary stand-in for an estimator that the polymorphic base class could be transplanted to.

This is just for playing around, and will be deleted again.

sktime/base/_poly.py
fkiraly: the solution I arrived at now does not use delegation

benHeid:
> @miraep8, @benHeid, this is working now but it makes me feel a bit uneasy due to its hackiness.
> 
> The problem is that the inheritance is dynamic based on an instance's object attribute, which seems to cause problems.

I am not sure if I understand the problem correctly. Thus, I ask if the problem is the `self` in line 27 of the module `_poly`? If this is the case, you might override `__new__` of `BasePolymorph`. The effect would be that you manipulate the class/object before it is instantiated. I am not sure if this would solve your issue. Additionally, this solution would have the disadvantage that you have to use the keyword argument `estimator_type` in the `__init__` too even if it does nothing.

```python
class BasePolymorph(BaseObject):
    """Handles parameter management for estimators composed of named estimators.

    Partly adapted from sklearn utils.metaestimator.py.
    """

    def __new__(cls, *args, estimator_type="base", **kwargs):
        from sktime.registry._lookup import _check_estimator_types
        obj = super().__new__(cls)
        baseclass = _check_estimator_types(estimator_type)[0]

        attr_to_copy = set(dir(baseclass)).difference(dir(obj))

        for attr in attr_to_copy:
            setattr(obj, attr, getattr(baseclass, attr))
        return obj

    def __init__(self, estimator_type="base"):
        super().__init__()
        self.estimator_type = estimator_type
```

fkiraly:
@benHeid, I think that's the solution!

>Additionally, this solution would have the disadvantage that you have to use the keyword argument `estimator_type` in the `__init__` too even if it does nothing.

It does something there, it is a parameter that is seen by the `sklearn` parameter interface, simply due to its presence in the signature. So that would be fine!

>Thus, I ask if the problem is the `self` in line 27 of the module `_poly`? If this is the case, you might override `__new__` of `BasePolymorph`.

Not sure whether we are referring to the same code or line, but the problem is that I cannot call `super(baseclass, self).__init__()`, it would fail since it does not inherit from the base class `baseclass`.

But, moving it to `__new__` is the right idea, thanks! We can even ensure that the right class is constructed!

fkiraly:
@benHeid, I modified your idea a bit - the new class does not need `estimator_type`, but has a method that can be overridden.

For instance, a `GraphPipeline` that turns out to be a forecaster can have this function being the logic of inferring the type from the final estimator.

fkiraly:
FYI @benheid; @RNKuhns for `skbase`/pipelines; @miraep8 in case you are still following this.

I think I've figured out how to create a polymorphic delegator, and a polymorphic pipeline using this - see the `_Delegator` and `Pipeline` classes in this PR.

The only thing that I still haven't figured out is how to unherit from the polymorphic `_Delegator`.
The `Pipeline`, when instantiated, is a `_Delegator` and not a `Pipeline`.

I now could copy-paste code from the `_Delegator`, but that would defeat the purpose of inheriting from the delegator...

Either way, worst case, the polymorphic pipeline would be a copy of the internals of `_Delegator`.

PR #3094, opened 2022-07-25
[BUG] Fix HIVE-COTE2 sporadic test failure

MatthewMiddlehurst:
Fixes #3083 by ensuring classifier OOB train estimate functions subsample more than 1 class value.

PR #3110, opened 2022-07-27
[BUG] Enforce minimum no. cases and unique class values for OOB train estimates

MatthewMiddlehurst:
It is technically possible to infinitely loop currently if you were to input data with 1 case or 1 unique class value to certain OOB train estimate functions.

Adds a min number of instances and unique class values to these functions to prevent this.


fkiraly (changes requested):
The change will simply make those estimators crash instead of hang, no?

Is there not a better solution, e.g., discard resamples with the bad property?

TonyBagnall (approved):
we should take a fresh look at the _get_train_probs structure again, we should probably avoid using the check_X_y method these days. However, this PR does what it says on the tin, so I'm fine with it.

MatthewMiddlehurst:
It is impossible to obtain a good resample if you only have 1 class or 1 case, unfortunately. It will either loop forever looking for a good resample, or crash if that loop is removed (the original fix).

I could change the min number of cases to 2 from 3, but you won't get anything good from 2 either.

MatthewMiddlehurst:
This unrelated to the current macOS issue as far as I am aware, just noticed the possibility after it was talked about on slack.

fkiraly:
>It is impossible to obtain a good resample if you only have 1 class or 1 case, unfortunately. It will either loop forever looking for a good resample, or crash if that loop is removed (the original fix).

Why do you need to raise an exception or loop infinitely? If all you see is a single class, than just always predict that class.

MatthewMiddlehurst:
I would say that is pretty useless, better to just exit gracefully.

fkiraly:
>I would say that is pretty useless, better to just exit gracefully.

Disagreed, raising an exception is not "graceful" but more similar to a train wreck.

My main worry is this: when the estimator is used inside a complex pipeline, let's say where we have nested sub-sampling, and 1 of 1000 sub-samples has the property that makes it crash, then the entire pipeline will crash.
Instead, we should simply let it through, in 1 of 1000 instances produces perhaps a useless output but not an exception.

From this argument, the heaviest we should do is raise a warning, but not an exception.

MatthewMiddlehurst:
If your pipeline ends up inputting a single class into the classifier, there is something wrong with the pipeline. Being able to handle a single class value is an exception, not the norm.

For sklearn classifiers, an acceptable alternative to being able to handle single class values is to raise an interpretable error.

fkiraly:
> If your pipeline ends up inputting a single class into the classifier, there is something wrong with the pipeline. Being able to handle a single class value is an exception, not the norm.

Disagreed. Please consider my example where we subsample 1000 times. There is a small but positive probability that the pipeline will break, leading to the kind of sporadic errors that we already keep seeing in the classification module.

MatthewMiddlehurst:
I don't think we are going to agree on this. All the current classifiers should reject inputs with 1 class, it is not an edge case, it is an invalid input. We end up catching these issues because exceptions are raised and tests break.

We don't seem to have a test like the sklearn one which verifies that estimators can properly handle a single class value or see that a proper exception is raised, but we probably should have one.

fkiraly:
I think, all classifiers that see a single class in `fit` should always predict that class in `predict`.

I am not going to let a PR through that introduces an exception for this in general.

As I have explained: this would increase the number of sporadic failures coming from the classifiers, due to the various algorithms that do stochastic sub-sampling, which can produce, on occasion, row subsets with a single class.

Most of these algorithms will be perfectly fine if you always predict the one class seen in `fit` if one class is encountered, but will otherwise crash in an exception sporadically under the same condition.

MatthewMiddlehurst:
Both methods are valid in the end, so im not going to fight over it, but I think outputting a single class is a useless function for the majority of cases which dont involve poorly created pipelines.

You may want to alter the base class or create a test if this is going to be an expectation going forward. Probably worth raising a warning as well.

fkiraly:
> Both methods are valid in the end, so im not going to fight over it, but I think outputting a single class is a useless function for the majority of cases which dont involve poorly created pipelines.

Please consider my argument, I think what you are saying is not correct.

The problem can occur in perfectly valid composites.

example: bagging a classifier.
Bagging sub-samples rows, so there is a non-zero chance that one of the instances inside ends up with a single class, and the chance approaches certainty as the number of internal instances approaches infinity.

Bagging is the one of the, if not the, most common ensemble technique(s).

fkiraly:
Re warning, I would not mind, see what I said above:

> From this argument, the heaviest we should do is raise a warning, but not an exception.

fkiraly:
Exercise: training set with 50 series, 30/70 class balance, we subsample rows with 50% independent probability, and fit 200 instances in an ensemble, on the subsamples. Per test run, we fit five times and test on 12 operating systems. If random number generators are independent for each fit and run, what is the probability for seeing at least one crash in the PR? I get something around 10%.

MatthewMiddlehurst:
Thanks for the lesson Franz. Since we are going a different direction regarding dealing with single class values I am going to close this

fkiraly:
instead of closing, would you be able to make the change, instead?
That would be useful.

fkiraly:
```python
if len(self._class_dictionary) == 1:
    return np.repeat(list(self._classdictionary.values()), len_X)
```
?

fkiraly:
I also think the bug is real? Would be appreciated if you could report it, I don't think there is an issue for this.

MatthewMiddlehurst:
Probably better to just revert #3094 and implement your proposal for the two vector classifiers.

PR #3114, opened 2022-07-28
Data generator for annotation (normal multivariate mean shift)

KatieBuc:
#### Reference Issues/PRs
Continuing the discussion from #2996

#### What does this implement/fix? 
This is the normal multivariate data generation extension, with:
- a flexible parametrisation (i.e. you can give it covariance or just variance and makes an assumption of independence)
- changed naming convention of `noise` to `std_dev` in univariate case, for consistency and specificity 

Maybe there should also be a check to see if optional input `covs` are symmetric and positive semidefinite.

@lmmentel @miraep8 what do you reckon?

##### For all contributions


miraep8:
Hey Katie!  I think it will be great to extend this to support multi-variate data generation! :) I think for me I would appreciate it for testing purposes as I think about making HMM support multi-variate data and I imagine it would be more useful to you with Hidalgo as well 🎉  :)

Just skimming it it occurred to me that there are probably a few other things to add in order to make it fully support the switch to multivariate:

- I like your idea of adding a test to see that the matrix is positive semi-definite and symmetric! :)  (Correct me if I am wrong but I think if we just look at the eigenvalues and they are positive/real that should be sufficient? Something along the lines of:)
```
eig_vals = np.linalg.eigvals(std_dev)
assert np.isreal(eig_vals).all()
....
```
- I think it could also be nice to add some checks that it is a square matrix first before running it through this, and make sure that the shape matches what we expect from the means!)
- Could also be nice to add some tests to make sure that it works as expected with the multivariate assumption! :) 
- I think we also need to think about how the means change as a function of time as well?  ie- I guess we should really be providing a 2D array of means now of shape [num_dimensions x num_segments]?  I think modifying the code to support sampling with the change in means would also be important? 
- I think it raises an interesting question as to whether we want to change the labeling part of this as well! :) But perhaps for now it is sufficient to just keep it checking for any changes in mean/std_dev. 

But yea - great suggestion and excited to see this merged! :)

sktime/annotation/datagen.py
```diff
@@ -8,25 +8,139 @@
...
+
+    L, N = np.array(means).shape
+
+    rng = check_random_state(random_state)
+    assert len(lengths) == L
+
+    # if no covariance is specified, build it from variance
+    # assuming independent random variables
+    if covs is None:
+        assert vars is not None
```
miraep8, line 102: I am curious since we default to vars = 1.0 do we need this?  I guess it guards against someone passing `vars= None`

sktime/annotation/datagen.py
```diff
@@ -8,25 +8,139 @@
...
+        return cov
+
+    L, N = np.array(means).shape
+
+    rng = check_random_state(random_state)
+    assert len(lengths) == L
+
+    # if no covariance is specified, build it from variance
+    # assuming independent random variables
+    if covs is None:
```
miraep8, line 101: Overall I like the code you added for building up covs if it is none! :)

sktime/annotation/datagen.py
```diff
@@ -8,25 +8,139 @@
 from sklearn.utils.validation import check_random_state
 
 
+def multivariate_mean_shift(
+    means: npt.ArrayLike,
+    lengths: npt.ArrayLike,
+    vars: Union[npt.ArrayLike, float] = 1.0,
```
lmmentel, line 14: FYI `vars` is a reserved keyword in python https://docs.python.org/3/library/functions.html#vars, would be worth changing it to avoid clashes

sktime/annotation/datagen.py
```diff
@@ -8,25 +8,139 @@
...
+    Returns
+    -------
+    data : np.array
+        multivariate time series as np.array of shape (X, N)
+        where X = sum(lengths)
+
+    Examples
+    --------
+    >>> from sktime.annotation.datagen import multivariate_shift
+    >>> multivariate_shift([[1, 1], [2, 2], [3, 3]], lengths=[2, 3, 1], random_state=2)
```
lmmentel, line 48: the name of the function is `mutivariate_mean_shift`

lmmentel:
Looks good @KatieBuc. Overall I think we should consider renaming both functions since they in principle can generate mean shifted series, variance shifted series and simultaneous mean and variance shifted series depending on the input parameters.

I think something like `piecewise_random` or `piecewise_normal` would be more appropriate, WDYT?

I see you bundled two changes that are not very related: adding multivariate data generation and changing the parameter names for existing function. In general I would suggest splitting unrelated issues into separate PRs since that makes it easier to manage and merge.

sktime/annotation/datagen.py
KatieBuc: yeah you're right, it might be excessive to check, but it's the one thing we need to build up covariance - in terms of debugging I think it's worth leaving it!

sktime/annotation/datagen.py
```diff
@@ -8,25 +8,149 @@
...
+    ----------
+    means : array_like
+        Means of the segments to be generated of shape (L, N)
+    lengths : array_like
+        Lengths of the segments to be generated of shape (L,)
+    variances : float or array_like (default=1.0)
+        Variance of the segments to be generated
+    covariances : array_like (default=None)
+        Covariances of segments to be generated of shape (L, N, N)
+        If None, this will be constructed from variances
```
lmmentel, line 36: Consider adding a sentence on how the covariances will be generated

sktime/annotation/datagen.py
```diff
@@ -8,25 +8,149 @@
...
+    random_state : int or np.random.RandomState
+        Either a random seed or RandomState instance
```
lmmentel, line 38:
```suggestion
        Either a random seed or ``RandomState`` instance
```

sktime/annotation/datagen.py
```diff
@@ -8,25 +8,149 @@
...
+
+    Returns
+    -------
+    data : np.array
```
lmmentel, line 42:
```suggestion
    data : array_like
```

sktime/annotation/datagen.py
```diff
@@ -8,25 +8,149 @@
...
+        multivariate time series as np.array of shape (X, N)
+        where X = sum(lengths)
```
lmmentel, line 44:
```suggestion
        Multivariate time series as ``np.array`` of shape (sum(lengths), n_series)
```

sktime/annotation/datagen.py
```diff
@@ -8,25 +8,149 @@
...
+
+    Examples
+    --------
+    >>> from sktime.annotation.datagen import piecewise_normal_multivariate
+    >>> piecewise_normal_multivariate([[1, 1], [2, 2], [3, 3]],\
```
lmmentel, line 49:
```suggestion
    >>> piecewise_normal_multivariate(means=[[1, 1], [2, 2], [3, 3]],\
```

sktime/annotation/datagen.py
```diff
@@ -8,25 +8,149 @@
...
+    array([[ 0.58324215,  0.94373317],
+           [-1.1361961 ,  2.64027081],
+           [ 0.20656441,  1.15825263],
+           [ 2.50288142,  0.75471191],
+           [ 0.94204778,  1.09099239],
+           [ 3.55145404,  5.29220801]])
+
+    >>> from sktime.annotation.datagen import piecewise_normal_multivariate
+    >>> piecewise_normal_multivariate([[1, 1], [2, 2], [3, 3]], lengths=[2, 3, 1],\
+        variances = [[1.0, 1.0], [1.0, 1.0], [1.0, 1.0]], random_state=2)
```
lmmentel, line 60:
```suggestion
    >>> piecewise_normal_multivariate(means=[[1, 1], [2, 2], [3, 3]], lengths=[2, 3, 1],\
        variances=[[1.0, 1.0], [1.0, 1.0], [1.0, 1.0]], random_state=2)
```

sktime/annotation/datagen.py
```diff
@@ -8,25 +8,149 @@
...
+           [-1.1361961 ,  2.64027081],
+           [ 0.20656441,  1.15825263],
+           [ 2.50288142,  0.75471191],
+           [ 0.94204778,  1.09099239],
+           [ 3.55145404,  5.29220801]])
+
+    >>> from sktime.annotation.datagen import piecewise_normal_multivariate
+    >>> piecewise_normal_multivariate([[1, 1], [2, 2], [3, 3]], lengths=[2, 3, 1],\
+        covariances = [[[1.0, 0], [0, 1.0]], [[1.0, 0], [0, 1.0]], [[1.0, 0],\
+        [0, 1.0]]], random_state=2)
```
lmmentel, line 71:
```suggestion
    >>> piecewise_normal_multivariate(means=[[1, 1], [2, 2], [3, 3]], lengths=[2, 3, 1],\
        covariances=[[[1.0, 0], [0, 1.0]], [[1.0, 0], [0, 1.0]], [[1.0, 0],\
        [0, 1.0]]], random_state=2)
```

sktime/annotation/datagen.py
```diff
@@ -8,25 +8,149 @@
...
+    array([[ 0.58324215,  0.94373317],
+           [-1.1361961 ,  2.64027081],
+           [ 0.20656441,  1.15825263],
+           [ 2.50288142,  0.75471191],
+           [ 0.94204778,  1.09099239],
+           [ 3.55145404,  5.29220801]])
+
+    >>> from sktime.annotation.datagen import piecewise_normal_multivariate
+    >>> piecewise_normal_multivariate([[1, 3], [4, 5]], lengths=[3, 3],\
+        covariances = [[[0.5, 0.3], [0.3, 1.0]], [[1.0, 0.3], [0.3, 0.7]]],\
```
lmmentel, line 81:
```suggestion
    >>> piecewise_normal_multivariate(means=[[1, 3], [4, 5]], lengths=[3, 3],\
        covariances=[[[0.5, 0.3], [0.3, 1.0]], [[1.0, 0.3], [0.3, 0.7]]],\
```

sktime/annotation/datagen.py
```diff
@@ -8,25 +8,149 @@
...
+
+    assert np.array(covariances).shape[0] == L
+    assert np.array(covariances).shape[1] == N
+
+    segments_data = [
+        rng.multivariate_normal(mean=mean, cov=cov, size=length)
+        for mean, cov, length in zip(means, covariances, lengths)
+    ]
+
+    return np.array(np.concatenate(tuple(segments_data)))
```
lmmentel, line 132: Would this work?

```suggestion
    return np.concatenate(
        rng.multivariate_normal(mean=mean, cov=cov, size=length)
        for mean, cov, length in zip(means, covariances, lengths)
    )
```

KatieBuc:
Thanks so much for your comments @miraep8 @lmmentel - looks ready to merge now :)

lmmentel (approved):
LGTM!

PR #3118, opened 2022-07-28
[MNT] revert recent CI changes

fkiraly:
Update: CI does not fail on this, compared to current PR where almost always sth fails. This indicates that we have probably identified a change condition, even if not the root cause.

This indicates that revert will fix the issues, until the point where we understand the root cause better.

---
This is a diagnostic PR to investigate https://github.com/alan-turing-institute/sktime/issues/3103

Purpose: see whether the tests complete when these recent changes to CI are reverted:
https://github.com/alan-turing-institute/sktime/pull/2896
https://github.com/alan-turing-institute/sktime/pull/3107

If the theory is correct thatsaid changes are the culprit, this should pass. Otherwise, it would fail.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -80,7 +80,6 @@ jobs:
           mkdir -p testdir/
           cp .coveragerc testdir/
           cp setup.cfg testdir/
-          cd testdir/
           python -m pytest
 
       - name: Publish code coverage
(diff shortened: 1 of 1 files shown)


fkiraly:
restarting for good measure, see if it breaks on a second run

fkiraly:
1st and 2nd time: win and unix runs
1st time: mac breaks with same issue

fkiraly:
Has served its purpose - we know that the CI changes are not at fault since reverting them leaves us with the same issues.

Issue #3120, opened 2022-07-28
[ENH] ensure all `BaseObjects` are tested, split object from estimator tests


fkiraly:
Currently, objects (`BaseObject` descendants) that are not `BaseEstimator`s are not tested for `sklearn` compatibility, but they should be. The testing framework has all interface compliance tests needed for that, so could be adapted in-principle.

Example bug that was not detected due to this: https://github.com/alan-turing-institute/sktime/issues/3119

Affected objects:
* splitters, `BaseSplitter` descendants
* performance metrics, `BaseMetric` descendants
* rarely used objects present in the codebase like `BaseTask`

The key problem that needs to be addressed is that currently the test framework assumes inheritance from `BaseEstimator`, and therefore presence of `fit` and associated interface assumptions.

Changes that would need to be made:
* split `TestAllEstimators` into a part for objects and a part for estimators. Also relates to work on https://github.com/sktime/BaseObject where this splitting is currently ongoing. There should be a `TestAllObjects`, and, separately, `TestAllEstimators`.
* add base classes to the base class registry in `registry._base_classes`, including `BaseObject` as a base class that is not there yet
* optionally, add additional tests that are specific to the more specific base classes like `BaseSplitter`, `BaseMetric`. Existing tests should be moved into the test class
* ensure the newly tested classes are interface compliant, by fixing any bugs similar to #3119 that the upgraded testing framework discovers

fkiraly:
This is done, with #3122.

Issue #3119, opened 2022-07-28
[BUG] clone is not working correctly on ExpandingWindowSplitter


SzymonStolarski:
**Describe the bug**

- Sklearn's `clone()` function is not working correctly on `ExpandingWindowSplitter`
- The `window_length` parameter is not copied
- `initial_window` is copied into `window_length` which is fine according to docstrings, however is counterintuitive and incompatible with `clone()`
- This generates problems when cloning models that have cross-validation embeded
- Example: `MultiplexForecaster` combined with `ForecastingGridSearchCV` cannot be evaluated using sktime's `evaluate()`

**To Reproduce**

```python
from sktime.forecasting.model_selection import ExpandingWindowSplitter
from sklearn.base import clone


cv = ExpandingWindowSplitter(
    start_with_window=True,
    initial_window=10,
    fh=[1])

print('pre-clone: ', cv.__dict__)
print('post-clone: ', clone(cv).__dict__)
```
```
pre-clone:  {'step_length': 1, 'start_with_window': True, 'initial_window': None, 'window_length': 10, 'fh': [1]}
post-clone:  {'step_length': 1, 'start_with_window': True, 'initial_window': None, 'window_length': None, 'fh': [1]}
```

**Expected behavior**

Pre-clone and post-clone CVs should be identical.

**Additional context**

**Versions**
<details>

System:
    python: 3.8.10 (default, Mar 15 2022, 12:22:08)  [GCC 9.4.0]
executable: /local_disk0/pythonVirtualEnvDirs/virtualEnv-3d157b3d-b81c-4d78-b9fe-a70430ada65a/bin/python
   machine: Linux-5.4.0-1086-azure-x86_64-with-glibc2.29

Python dependencies:
          pip: 21.0.1
   setuptools: 52.0.0
      sklearn: 0.24.1
       sktime: 0.13.0
  statsmodels: 0.13.1
        numpy: 1.22.4
        scipy: 1.6.2
       pandas: 1.2.4
   matplotlib: 3.4.2
       joblib: 1.0.1
        numba: 0.55.2
     pmdarima: 1.8.4
      tsfresh: None
</details>

fkiraly:
Yes, I can confirm that.

I wondered why the tests do not catch this, since all objects meant to be sklearn compatible are run through the test suite which tests precisely this.

[...]

PR #3121, opened 2022-07-28
[BUG] `ExpandingWindowSplitter` constructor `sklearn` conformace fix

fkiraly:
This PR fixes the non-conformance with `sklearn` specs in the `ExpandingWindowSplitter` constructor reported in #3119.

The problem is that `ExpandingWindowSplitter` calls its parent constructor, of `BaseWindowSplitter`, with certain parameter names switched. This has the effect of overwriting `self.initial_window` with a value that is in general not equal to that passed to the constructor, which breaks the `sklearn` constructor interface expectation.

The solution is as follows:

* `BaseWindowSplitter` internally now accesses a property `self._initial_window` wherever it previously accessed `self.initial_window`
* that property reads from `self.initial_window` if exists, otherwise returne `None`
* `ExpandingWindowSplitter` overrides the property, and does not overwrite `self.inital_window`, which fixes the interface conformance while retaining the patterns used

Tests are not being added here, they are in a separate PR #3122 which includes this PR, and they are verified to catch this issue.

Why: this is an instance of #3120 which should be addressed systemically, doing otherwise would proliferate sporadic test cases that are already covered in the framework.
The corresponding PR #3122 adds tests for all `BaseObject` descendants, and which also catches the `ExpandingWindowSplitter` issue (verified).

I have also manually verified that the example provided in #3119 now gives the expected results.

### DIFF
--- a/sktime/forecasting/model_selection/_split.py
+++ b/sktime/forecasting/model_selection/_split.py
@@ -904,7 +904,9 @@ class BaseWindowSplitter(BaseSplitter):
         """
         start = self._get_start(y=y, fh=fh)
         split_points = self.get_cutoffs(pd.Series(index=y, dtype=float)) + 1
-        split_points = split_points if self._initial_window is None else split_points[1:]
+        split_points = (
+            split_points if self._initial_window is None else split_points[1:]
+        )
         for split_point in split_points:
             train_start = self._get_train_start(
                 start=start if expanding else split_point,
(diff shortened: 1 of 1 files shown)


khrapovs (approved):
Looks good to me. Some strange magic in `ExpandingWindowSplitter` looks weird at the first glance, but thinking about it for a moment makes sense.

PR #3122, opened 2022-07-28
[ENH] test all `BaseObject` descendants for sklearn compatibility

fkiraly:
Addresses https://github.com/alan-turing-institute/sktime/issues/3120 by ensuring that all `BaseObject` descendants are tested for `sklearn` compatibility.

Approach is adding `BaseObject` to the base class registry, and splitting off a class `TestAllObjects` from `TestAllEstimators`.

Secondary changes needed for this:
* `check_estimator` to be adapted to the split
* base class strings added in registry for `"estimator"` and `"object"`
* `scitype` function modified to return most specific scitype (e.g., exclude `"object"` from returns if inherits from subclass)

Issues found and fixed:
* `CNNNetwork` did not have dependency `tensorflow` isolated
* some splitters needed default test params via `get_test_params`
* performance metrics and their tags needed to be properly registered in the registry
* `PinballLoss` constructor was not `sklearn` compatible, similar to the bug in #3119

Includes the fix for `ExpandingWindowSplitter` from https://github.com/alan-turing-institute/sktime/pull/3121


fkiraly:
> If correct, this should find the bug in #3119 before the fix is merged.

Confirmation: it does!

```
FAILED sktime/tests/test_all_estimators.py::TestAllObjects::test_constructor[ExpandingWindowSplitter]
```

Issue #2880, opened 2022-06-27
[ENH] Design Discussion around pipelines and transformers which change element size. 


miraep8:
Related to issue #2807, and potentially also #2806 and #2879. 

I think we might be able to solve some of these problems by not treating the pipeline object as a forecaster until the very last step.

More broadly, I think this might be an argument in favor of making the pipelines themselves more modular.  

Just opening this issues to kick off the discussion and connect related issues.

fkiraly:
Could you perhaps be a bit more explicit what you mean by "not treating as forecaster until the very last step"?

This "relaxing of typing" PR could be related - is it?
https://github.com/alan-turing-institute/sktime/pull/2842

miraep8:
Yes I would agree the "relaxing of typing" PR is relevant, thank you for linking it. 

And yes - would be happy to be more explicit!  I am still in the process of tracing all the code relevant to this, so this is probably a less than perfect understanding, but from what I could tell: 
- in the example which came up in issue 2807 `ForecastingGridSearch` we end up pre-trimming all `X` in a `fh` dependent manner via `_split` before passing it to `TransformedTargetForecaster`
- but in reality, if we are working with something like `TransformedTargetForecaster` I think it doesn't really makes sense to trim the `X` until right before the forecasting step? (Usually it isn't an issue, but it will come up with transformers that change the size of `X` like we saw) 
- so the thought was, is there a way we can delay some of these pre-processing steps that are forecaster-specific until the step in `TransformedTargetForecaster` that is right before the actual forecasting part? 

I am not actually sure that the second bullet point above actually holds.  Is this a valid point, or are their actual cases where we would need to trim the `X` in this manner for transformers as well? I think my understanding on this might be less than perfect!

fkiraly:
> so the thought was, is there a way we can delay some of these pre-processing steps that are forecaster-specific until the step in TransformedTargetForecaster that is right before the actual forecasting part?

Isn't that what #2842 was trying to achieve, by setting pipeline tags to the "relaxed" one where no trimming takes place, instead of copying that one from the forecaster?

fkiraly:
> but in reality, if we are working with something like `TransformedTargetForecaster` I think it doesn't really makes sense to trim the `X` until right before the forecasting step

I think your understanding is correct, but also I thought `TransformedTargetForecaster` does not trim `X`?

fkiraly:
[...]

Issue #2807, opened 2022-06-13
[BUG] long fh does not work for specific forecasting pipeline


fkiraly:
posted by @anthonygiorgio97 in #2452
(added some missing imports)
---

Another strange behaviour is that if `fh=[1,2]` the code below works, but if `fh=[1,2,3]` or `[1,2,3,4],...` or longer, I have error:

```python
from sktime.forecasting.compose import make_reduction, ForecastingPipeline, TransformedTargetForecaster
from sktime.forecasting.model_selection import ExpandingWindowSplitter, temporal_train_test_split, ForecastingGridSearchCV
from sktime.forecasting.fbprophet import Prophet
from sktime.transformations.series.adapt import TabularToSeriesAdaptor
from sktime.transformations.series.boxcox import LogTransformer
from sktime.transformations.series.compose import OptionalPassthrough
from sktime.transformations.series.difference import Differencer

from sklearn.preprocessing import StandardScaler
from sklearn.svm import SVR
from sktime.datasets import load_longley

y, X = load_longley()

#split train/test both y and X
fh = [1,2,3]
train_model, test_model = temporal_train_test_split(y, fh = fh)
X_train = X[X.index.isin(train_model.index)]
X_test = X[X.index.isin(test_model.index)]

#pipeline
pipe = TransformedTargetForecaster(steps = [
            ("log", OptionalPassthrough(LogTransformer())),
            ("differencer", Differencer(drop_na = True)),
            ("scaler", TabularToSeriesAdaptor(StandardScaler())),
            ("myforecasterpipe", ForecastingPipeline(steps = [
                ("logX", OptionalPassthrough(LogTransformer())),
                ("differencerX", Differencer(drop_na = True)),
                ("scalerX", TabularToSeriesAdaptor(StandardScaler())),
                ("myforecaster", make_reduction(SVR()))
        ]))  
])

#cv setup
N_cv_fold = 1
step_cv = 1
cv = ExpandingWindowSplitter(initial_window = len(train_model) - (N_cv_fold - 1) * step_cv - len(fh),
                            start_with_window = True,
                            step_length = step_cv,
                            fh = fh)

param_grid = [{"log__passthrough": [False],
               "myforecasterpipe__logX__passthrough": [False],
               "myforecasterpipe__myforecaster__window_length": [2,3],
               "myforecasterpipe__myforecaster__estimator__C": [10,100]
              },
              {"log__passthrough": [True],
               "myforecasterpipe__logX__passthrough": [True],
[...]

PR #3125, opened 2022-07-28
[BUG] Delay trimming in ForecastingGridSearchCV until after transforming

miraep8:
@anthonygiorgio97 noticed + helpfully pointed out a strange bug that came up when using `Differencer` within `TransformedTargetForecaster` in `ForecastingGridSearchCV`.  (see #2807 and #2880)

From what I could tell this bug was caused by the fact that `Differencer` reduces the size of `X`.  Usually this isn't a problem, but because `ForecastingGridSearchCV` was pre-trimming all `X` input before passing it into the underlying forecaster, `X` ended up being too small by the time it got to the final forecaster and was causing problems. 

My proposed solution is thus simply - avoid pre-trimming `X` in the `_split` function and instead trim it to the right size right before the actual forecasting step. (I do this specifically in `_predict_last_window`, but there is a decent chance it might need to be changed elsewhere as well).


fkiraly:
ah, it was coming from `evaluate`! I was wondering where we were trimming.

miraep8:
Oh dear.  Something seems to have gone terribly wrong with what should have been a fairly vanilla merge from main branch.  Apologies for all the code review requests.  I am planning on closing this PR and starting a new one....

PR #767, opened 2021-03-31
[ENH] adapting ROCKETClassifier to ROCKETRegressor

RavenRudi:
#### Reference Issues/PRs
Contributes to #212

#### What does this implement/fix? Explain your changes.
Refactor ROCKETClassifier to ROCKETRegressor

#### Does your contribution introduce a new dependency? If yes, which one?
No.

##### For all contributions
- [ ] I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/main/.all-contributorsrc).
- [ ] Optionally, I've updated sktime's [CODEOWNERS](https://github.com/alan-turing-institute/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
- [ ] I've added unit tests and made sure they pass locally.

##### For new estimators
- [ ] I've added the estimator to the online documentation.
- [ ] I've updated the existing example notebooks or provided a new one to showcase how my estimator works.

PR #3126, opened 2022-07-28
[ENH] Rocket regressor and regressor pipelines, regressor delegators

fkiraly:
This PR reprises the old PR https://github.com/alan-turing-institute/sktime/pull/767 and attempts to solve it with abstract composition functionality instead of a rocket specific refactor.

This PR introduces compositors for regressors similar to those of classifiers:

* two regressor pipelines, an sktime one and an sklearn one, with dunders
* a delegator for regressors

With the above, the intended translation of classifier to regressor in #767 is as simple as taking the rocket classifier, and changing every string "classifier" in the code to "regressor".

There is a lot of copy-paste, so a common parent such as in #767 could be attempted based on this.

### DIFF
--- a/sktime/regression/_delegate.py
+++ b/sktime/regression/_delegate.py
@@ -1,0 +1,93 @@
+# -*- coding: utf-8 -*-
+"""Delegator mixin that delegates all methods to wrapped regressors.
+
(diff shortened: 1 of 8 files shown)


fkiraly:
FYI @RavenRudi, @mloning

PR #3129, opened 2022-07-29
[Bug] Fix override/defaulting of "prediction intervals" adders 

### LINKED ISSUE #3009: [ENH] check override/defaulting of "prediction intervals" adders
From the presentation of @bethrice44:

if we take a "prediction intervals adder" wrapper, e.g., one adding `predict_intervals`, to add prediction intervals to an estimator that already has a *different* probabilistic prediction method implemented, this may lead to unexpected behaviour.

For example, assume the inner method has `predict_quantiles` implemented, and `predict_intervals` uses the default method which calls `predict_quantiles`. 

The wrapper overwrites `predict_intervals`, and it might not overwrite `predict_quantiles`. Then, the predictoins from `predict_intervals` and `predict_quantiles` will be inconsistent.

Instead, the wrapper perhaps should (?) overwrite `predict_quantiles` with the base class default?

bethrice44:
#### Reference Issues/PRs
Fixes #3009 

This adds the base class `_predict_quantiles` method into `conformal.py` to ensure that the `_predict_quantiles` method for the forecaster that is wrapped with `conformal.py` is consistent with `_predict_interval`.

No

I have just copied and pasted the base class `_predict_quantiles` method and removed the `implements_interval` check . Should this be amended at all? 

##### For all contributions

##### For new estimators


fkiraly (changes requested):
Makes sense!

Question, should this perhaps be an explicit call to `BaseForecaster._predict_quantiles` instead?

Because if we fix a bug in that, it would not propagate to here.

This is very non-hypothetical, see this bug in the default `_predict_quantiles`!
https://github.com/alan-turing-institute/sktime/pull/3106

bethrice44:
> Makes sense!
> 
> Question, should this perhaps be an explicit call to `BaseForecaster._predict_quantiles` instead?
> 
> Because if we fix a bug in that, it would not propagate to here.
> 
> This is very non-hypothetical, see this bug in the default `_predict_quantiles`! #3106

Good shout - I'll call directly to `BaseForecaster`

fkiraly (approved):
Thanks!

fkiraly:
CI failures unrelated, see https://github.com/alan-turing-institute/sktime/issues/3103

fkiraly:
Failures unrelated, caused by #3103

PR #3130, opened 2022-07-29
[ENH] ClearSky transformer for solar time series

ciaran-g:
#### Reference Issues/PRs

No reference issue

This PR is a new transformer which is intended to be used with solar (related) time-series. The method approximates the clear sky value for a given day-of-year and time-of-day based on the input time-series. Values are then transformed into a percentage domain using this estimate of the clear sky as the denominator.

The clear sky is approximated via a weighted quantile, where the weights are defined using von-mises kernels with bandwidths chosen by the user.

This transformation can be inaccurate at low values, in the solar example during early morning and late evening. Therefore, clear sky values below a threshold can be fixed to zero in the transformed domain. Denominator values of zero are set to zero in the transformed domain by default.

This transformer is based on the work detailed in this [paper](https://doi.org/10.1016/j.solener.2009.05.016)

No 

#### Any other comments?

- I added a `_load_solar()` function which queries [this](https://www.solar.sheffield.ac.uk/pvlive/) dataset for the example
- I wasn't sure whether to check for clock changes in the input, is this already checked somewhere else?

##### For all contributions
- [] Optionally, I've updated sktime's [CODEOWNERS](https://github.com/alan-turing-institute/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
- [] I've added unit tests and made sure they pass locally.

##### For new estimators


sktime/transformations/series/clear_sky.py
```diff
@@ -0,0 +1,286 @@
...
+        indx_seasonal = pd.MultiIndex.from_arrays([yday, tod], names=["yday", "tod"])
+
+        # look up values and overwrite index
+        csp = self.clearskypower[indx_seasonal].copy()
+        csp.index = X.index
+        X_trafo = X * csp
+
+        return X_trafo
+
+    def get_fitted_params(self):
```
fkiraly, line 215: changed this recently to `_get_fitted_params`

sktime/utils/_testing/series.py
```diff
@@ -62,3 +62,33 @@ def _make_index(n_timepoints, index_type=None):
 
     else:
         raise ValueError(f"index_class: {index_type} is not supported")
+
+
+def _load_solar(
```
fkiraly, line 67: Nice!

I would actually give this a docstring, and move it to `datasets` as a "full" data set, not just one for testing.

* question: is this reliable, as in, always producing the same data?
* I would suggest to add a test for this
* reference should be added in `docs/source/api_reference/datasets`

fkiraly (changes requested):
Nice!

One change request: `get_fitted_params` should be `_get_fitted_params` (recent interface change - not your fault since it has not been published in release notes or extension templates, but saves us some time)

No other blockers from my side, but I would suggest to move the data loader because other people might find it interesting.

sktime/utils/_testing/series.py
ciaran-g: Good shout, update now including a test

ciaran-g:
just updated @fkiraly

sktime/transformations/series/clear_sky.py
```diff
@@ -0,0 +1,286 @@
+# -*- coding: utf-8 -*-
```
aiwalter, line 1: can we rename this file, e.g. to `solar.py` or can we even have a name that is independent on the data domain?

fkiraly:
Huh, now it seems that the transform/inverse transform tests are failing. They did not fail before! Any good guesses why?

ciaran-g:
Yeah it's strange, the tests are all passing on my local machine with

```python

from sktime.transformations.series.clear_sky import ClearSky

check_estimator(ClearSky)

```

sktime/transformations/series/clear_sky.py
ciaran-g: Could change it if more solar specific/related transformations come up?

ciaran-g:
> Huh, now it seems that the transform/inverse transform tests are failing. They did not fail before! Any good guesses why?

Think this is due to the `min_thresh` default, which is to prevent dividing by zero in the transformation for overnight values lol. Still, not sure why the tests were passing locally and not remotely. Changed the default for testing to see if it passes

fkiraly (approved):
Happy now since the tests pass.

Re name, we could move it later to an "astronomy" or "calendar" related selection, perhaps together with datetime etc.

fkiraly:
>I wasn't sure whether to check for clock changes in the input, is this already checked somewhere else?

@ciaran-g, what did you mean by this question?

sktime/transformations/series/clear_sky.py
fkiraly: we can move this later (after the next release), we need to clean up transformers folder structure anyway...

ciaran-g:
Hi @fkiraly, was on hols there 🌞 

> @ciaran-g, what did you mean by this question?

Just that if the index of `X` is `pd.DatetimeIndex` and is localized to a specific timezone, rather than UTC or something, I was thinking we should raise a warning, or deal with the local clock changes before calculating the day-of-year and hour-of-day and fitting the transformer. Wasn't sure of the best way to deal with this.

```python
from sktime.datasets import load_solar

# by default the time-series in in UTC
X = load_solar(
    start="2022-03-27",
    end="2022-03-28",
)
# but if it was in local time we could deal clock changes before fitting the transformer or raise a warning?
X_tz = X.copy()
X_tz.index = X_tz.index.tz_convert("Europe/London")

(X.index.hour, X_tz.index.hour)

```

PR #3090, opened 2022-07-25
[ENH] Make ProbabilityThresholdEarlyClassifier a BaseEarlyClassifier

MatthewMiddlehurst:
Draft PR for moving ProbabilityThresholdEarlyClassifier to the early classification base class and changes to early classification tests.

PR #3133, opened 2022-07-29
[ENH] Deprecate ProbabilityThresholdEarlyClassifier

MatthewMiddlehurst:
Deprecates ProbabilityThresholdEarlyClassifier for #3090.

Removes the capability:early_prediction tag, which was scheduled to be removed in v0.13.0 (#2895).


fkiraly (changes requested):
Thanks!

In this case, we are changing the location and type of classifier.

So, I would do this together with the alternative, so the message can directly point to what users should do/use instead, e.g., with the message

`"ProbabilityThresholdEarlyClassifier has been refactored to a BaseEarlyClassifier typed estimator, and will be deprecated in v0.15.0, in favour of the early classifier of the same name, in modulename. Please use ProbabilityThresholdEarlyClassifier in modulename instead."`

MatthewMiddlehurst:
I think that would be better, but unfortunately it is not changing location, just base class.

Which would mean we would need a different name/location for the new version. This would have to be permanent, or immediately also be deprecated to return it back, I think?

MatthewMiddlehurst:
Also it seems like a notebook is still loading data from timeseriesclassification.com? Seems like what that notebook test failure is.

fkiraly:
> Also I think a notebook is still loading data from timeseriesclassification.com? Seems like what that notebook test failure is.

What do you mean, "still"?

fkiraly:
> I think that would be better, but unfortunately it is not changing location, just base class.

Hm, that's trickier.
Is there any option of a transition state? Or is the end state fully downwards compatible?

MatthewMiddlehurst:
At some point we made an effort to remove methods which loaded data from timeseriesclassification.com from notebooks and tests, as you cannot reliably control whether it will be able to retrieve them.

MatthewMiddlehurst:
Both versions use different methods with different inputs/outputs to return decisions on whether a prediction is safe or not, I don't think it would be possible to make the new version backwards compatible without essentially having both versions in a single class and making it extremely messy.

Best option was to deprecate the classifier two releases ago, but I did not know when I would be able to get around to this 🙂.

fkiraly:
is the difference only the `predict` return? If so, we could introduce a parameter to the init which controls what is returned. That can be added without deprecation since additional params are downwards compatible. The deprecation message then advises to flip the parameter.

MatthewMiddlehurst:
No, you would pretty much need two versions of predict_proba controlled by this parameter in a single class, and then half the methods most likely wouldn't work depending on which option you chose.

I would honestly rather just wait or name the new version something different temporarily i.e. ThresholdEarlyClassifier.

fkiraly:
ok, let's go with a simple deprecation message, which should be clear that the interface of the class will change to early classifier.

MatthewMiddlehurst:
Is the current message ok? I will edit the other PR to make sure it is not accidentally merged if so.

fkiraly (approved):
Yes, happy with this.

PR #3137, opened 2022-07-30
[ENH] partition design for test matrix to reduce test time to a third

fkiraly:
This PR addresses long test runtimes by a random partition design of allocating estimators to OS/python version combinations.
Addresses https://github.com/alan-turing-institute/sktime/issues/2890.

The idea is to test a third of estimators per OS/version combination only, but to ensure that all estimators are tested on all OS, and all versions (in the marginal view).

The algorithm is simple: we take a fixed random partition of estimators, and select the i + j mod 3 th part where i is integer index for OS, and j is integer index for version, assuming integer indexing is consecutive.

The condition is ensured as long as we test at least 3 python versions and at least 3 operating systems.


fkiraly:
reduces test time to 30-40min

aiwalter:
can I still test locally as before?

fkiraly:
>can I still test locally as before?

Yes, that was the question I had yesterday, I didn't know how to do this.

The solution is to control this via an extra `pytest` parameter that turns the partition design on. When not given (e.g., local standard call), you have the old behaviour - including local tests, or `check_estimator` which should behave just as before.

PR #3140, opened 2022-07-30
[ENH] classifier single class handling

fkiraly:
Handles "single class seen in `fit`" scenario that can lead to infinite loops according to unreported bug discussed in PR #3110.

Alternative solution which simply returns "that class" predicted every time, instead of erroring out. Includes tests.

### DIFF
--- a/sktime/classification/base.py
+++ b/sktime/classification/base.py
@@ -152,11 +152,13 @@ class BaseClassifier(BaseEstimator, ABC):
         start = int(round(time.time() * 1000))
         # convenience conversions to allow user flexibility:
         # if X is 2D array, convert to 3D, if y is Series, convert to numpy
-        X, y = _internal_convert(X, y)
-        X_metadata = _check_classifier_input(X, y)
+        X, y = self._internal_convert(X, y)
+        X_metadata = self._check_classifier_input(X, y)
         missing = X_metadata["has_nans"]
         multivariate = not X_metadata["is_univariate"]
         unequal = not X_metadata["is_equal_length"]
+        self._X_metadata = X_metadata
+
         # Check this classifier can handle characteristics
         self._check_capabilities(missing, multivariate, unequal)
         # Convert data as dictated by the classifier tags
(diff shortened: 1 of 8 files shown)


MatthewMiddlehurst:
While this will ensure all classifiers return a single class from predict, it won't stop them potentially crashing in fit. Easy solution for now is to just add another class_dictionary check in fit and stop building if it is length 1.

fkiraly:
>While this will ensure all classifiers return a single class from predict, it won't stop them potentially crashing in fit. Easy solution for now is to just add another class_dictionary check in fit and stop building if it is length 1.

I did indeed consider that, but then I thought, this would make the "fitted" interface polymorphic. I.e., you would not have fitted parameters etc.
Would it crash the problematic estimators if we do not skip `fit`?

MatthewMiddlehurst:
Not going to block for any of the above, just a concern.

fkiraly:
> If we don't do it in the base class, we would probably require a test for all classifiers making sure it does interact as intended.

Yes, you are right.
We either need to handle it in the base class and test the base class, or ensure the individual implementations guarantee it, by test.

I'll add a test that checks `fit` only.

fkiraly:
> Prehaps the fit check would be the place to raise a warning that only a single class has been seen and certain parameters have not been fit?

Yes, you are also right here, this is a good place to raise a warning (rather than the exception).

fkiraly:
added test & warning now

fkiraly:
Hm, I actually went with your earlier suggestion of just skipping `fit`. I think there were too many estimators that broke with just one class.

MatthewMiddlehurst (dismissed):
Probably don't need the test_all_classifiers test any more if it is handled in fit, but if there is a reason I'm missing to keep it, sure.

fkiraly:
> Probably don't need the test_all_classifiers test any more if it is handled in fit, but if there is a reason I'm missing to keep it, sure.

It seems very quick (no fitting actually happens), so I feel it makes sense to keep it for regression testing in case we change the logic, but no strong reason.

Issue #652, opened 2021-02-02
Behaviour modifying compositors for forecasting


fkiraly:
**Is your feature request related to a problem? Please describe.**
Indirectly arising from discussion on Feb 2, it might be useful to implement wrapper compositors that modify the behaviour of fitting/updating:
* "sliding window update modifier" - wraps a `Forecaster` and replaces `update` of the wrapped estimator with a sliding window re-fit. Hyper-parameters are sliding window parameters.
* "re-fit after T time modifier" - wraps a `Forecaster` and replaces `update` with the following meta-algorithm: if time since last `fit` is <T, then call `update` of wrapped forecaster. Otherwise, re-fit to a pre-specified window (fixed length or all past). Hyper-parameters are T and the window specs.
* "re-fit once forecast error is too large" - wraps a `Forecaster` and re-fits if forecast error, evaluated in a certain way, exceeds a threshold. Internally calls the `evaluator`. Hyper-parameters are: evaluation parameters, including loss metric; threshold that needs to be passed for re-fit to be triggered. Could be dynamic, so might be interesting to think about how to make this more abstract to allow arbitrary smoothers/anomaly detectors on the loss metric.

**Describe the solution you'd like**
A clear and concise description of what you want to happen, ideally taking into consideration the existing toolbox design, classes and methods. 

**Describe alternatives you've considered**
The alternative would be to have "merged/contracted" estimators or composites in cases where these compositors might be applied instead, e.g., as hyper-parameter options of grid search tuners.

fkiraly:
@aiwalter, @mloning, would be interested in your opinion.

mloning:
Sounds interesting, they could be combined into a "conditional refitter" with different options for what the condition is, i.e. in case 1 it's always true, in case 2 it's some duration threshold, in case 3 some error threshold. 

The condition could perhaps even be expressable as an estimator itself (think anomaly detector) - similar to the syntax and composition tools you find in [adtk](https://adtk.readthedocs.io/en/stable/index.html) where we have operators for estimators (e.g. an aggregator). So you could express more complex rules like "refit if an anomaly was detected given some anomaly detection estimator".

fkiraly:
@mloning, yes, your idea sounds like an improvement over mine. Perhaps one "conditional refitter (simple)" and one "pipeline conditional refitter (adtk style)"?

fkiraly:
"refit after time" is implemented as `UpdateRefitsEvery`

[...]

PR #3144, opened 2022-07-30
[ENH] Add `_ConditionalRefitForecaster` and refactor `UpdateRefitsEvery` to subclass it

ZiyaoWei:
#### Reference Issues/PRs

Initial steps of #652.

This adds a `_ConditionalRefitForecaster` as mentioned [here](https://github.com/alan-turing-institute/sktime/issues/652#issuecomment-772585593), to be used in implementations of different refitting forecasters.

No.

The API design - is this okay and can it be exposed to the public?

##### For all contributions

##### For new estimators


fkiraly:
Nice! May I suggest:

* make the `ConditionalRefitForecaster` a separate estimator
* don't start with an underscore, or tests framework will not detect it
* leave `UpdateRefitsEvery` unchanged per functionality, as we have already released it. Fine if you refactor it (subclass or common superclass) but please ensure that the functionality does not change, at least for the next deprecation cycle.

ZiyaoWei:
Thanks for the quick response! To make sure I understand:

> * make the `ConditionalRefitForecaster` a separate estimator
Do you mean putting it in a separate file?

> * don't start concrete estimators with an underscore, or tests framework will not detect it. Joint superclasses can be private, of course.
So we are making `ConditionalRefitForecaster` public, and users would subclass it? (see [this comment](https://github.com/alan-turing-institute/sktime/issues/652#issuecomment-1200247799) as well)

> * leave `UpdateRefitsEvery` unchanged per functionality, as we have already released it. Fine if you refactor it (subclass or common superclass) but please ensure that the functionality does not change, at least for the next deprecation cycle.
I'd love to add some tests for it to make sure the functionality doesn't change. Maybe I can file a ticket for adding tests and refactoring?

fkiraly:
> Do you mean putting it in a separate file?

No, just a separate class.

> So we are making `ConditionalRefitForecaster` public, and users would subclass it? (see [this comment](https://github.com/alan-turing-institute/sktime/issues/652#issuecomment-1200247799) as well)

answering there

> I'd love to add some tests for it to make sure the functionality doesn't change. Maybe I can file a ticket for adding tests and refactoring?

good idea. Please note that generic tests already cover the forecaster, so the only things that need to be tested are whether the functionality is correct.

ZiyaoWei:
Thanks @fkiraly! I've updated the PR to remove the refactoring and added a public class `ConditionalRefitForecaster`, which just takes in a bunch of functions and pass them into `_ConditionalRefitForecaster`. Hope this makes some sense and further discussions are most welcome!

@fkiraly - you mention `although there may be a case for or against having a joint parent class` in the comments in #652 - can you elaborate a bit? There are currently both the parent class and the public class in the PR, but I'll be happy to remove one of the classes if it proves to be not of much use.

fkiraly:
> @fkiraly - you mention although there may be a case for or against having a joint parent class in the comments in https://github.com/alan-turing-institute/sktime/issues/652 - can you elaborate a bit? There are currently both the parent class and the public class in the PR, but I'll be happy to remove one of the classes if it proves to be not of much use.

I thought the parent could be a parent of `UpdateRefitsEvery` as well, or be disconnected of it. Not sure what is better.

ZiyaoWei:
I'll change this to ready for review - I'll add tests for this but want to make sure it looks reasonable at least.

fkiraly:
Hm, overall looks nice!

I think though the logic of the condition and data retrieval can be simplified, could be wrapped in less functions.

Is it not possible to achieve the same effect by a single function that checks the condition, for the simplest version?
And perhaps second function for the data memory or subsetting.

Could you explain what the `_set_refit_condition` and `_update_refit_condition` are meant to do?

PR #3145, opened 2022-07-31
[ENH] Refactor `CutoffSplitter` using `get_window` function

khrapovs:
#### Reference Issues/PRs

Next step in #1737 

This PR is a simple refactoring. Use `sktime.datatypes._utilities.get_window` in `sktime.forecasting.model_selection._split.CutoffSplitter._split`.

No

##### For all contributions

### DIFF
--- a/sktime/forecasting/model_selection/_split.py
+++ b/sktime/forecasting/model_selection/_split.py
@@ -673,6 +673,9 @@ class CutoffSplitter(BaseSplitter):
     which using the notation provided in :class:`BaseSplitter`,
     can be written as :math:`(k_1,\ldots,k_n)` for integer based indexing,
     or :math:`(t(k_1),\ldots,t(k_n))` for datetime based indexing.
+
+    For a cutoff :math:`k_i` and a `window_length` :math:`w`
+    the training window is :math:`(k_i-w,k_i-w+1,k_i-w+2,\ldots,k_i)`.
     Training window's last point is equal to the cutoff.
 
     Test window is defined by forecasting horizons
--- a/sktime/forecasting/model_selection/_split.py
+++ b/sktime/forecasting/model_selection/_split.py
@@ -730,10 +730,14 @@ class CutoffSplitter(BaseSplitter):
         _check_cutoffs_fh_y(cutoffs=cutoffs, fh=fh, y=y)
 
         for cutoff in cutoffs:
-            train_end = y[cutoff] if is_int(cutoff) else cutoff
-            training_window = get_window(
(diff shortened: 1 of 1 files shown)


fkiraly:
hm, something is broken here? tests fail

khrapovs:
While trying to make this PR work I noticed these lines in `sktime.forecasting.base._sktime._BaseWindowForecaster._predict_in_sample`:
```
cutoffs = fh.to_relative(self.cutoff) + len(y_train) - 2
cv = CutoffSplitter(cutoffs, fh=1, window_length=self.window_length_)
```
This results in `cutoffs = [-1, 0, 1, 2, ...]`. If integer cutoffs are treated as `iloc` references, then `-1` is the reference to the very last element in `y_train` which drives `CutoffSplitter` crazy. I have treated this case in the splitter separately, but it seems to me we have a bigger problem here.

fkiraly:
>but it seems to me we have a bigger problem here.

Could this be the source of the problems with `update_predict` tests that creates empty folds?

khrapovs:
> Could this be the source of the problems with `update_predict` tests that creates empty folds?

I am not aware of these problems, but it sounds very much related. `cutoff = -1` results in the empty `train_window = np.array([])`. For whatever historical reasons this is ok for the splitters, it is a tested outcome. But I still struggle to understand this.

khrapovs:
Just for a fun of it I tried to remove negative cutoffs like that:
```
cutoffs = fh.to_relative(self.cutoff) + len(y_train) - 2
cv = CutoffSplitter(cutoffs[cutoffs >= 0], fh=1, window_length=self.window_length_)
```
but this resulted in failure of some estimators. I decided to localise the changes inside of splitters.

fkiraly:
I am getting very confused.

My typical fallback in this case is the docstring, which imo is not detailed enough.
Could you kindly add your understanding there what the splitter does:

* let's start with a full index set $t_1, ..., t_N$
* and an input of cutoffs $k_i$ and a forecasting horizon indices $h_i$
* what is the set of training indices output? what is the set of test indices output precisely?
* for instance, line 682, it should be the test set, not the training set, right?

fkiraly:
> Just for a fun of it I tried to remove negative cutoffs like that

Not to judge, but I can imagine 100 things that seem more fun, such as cutting my own nose off, or eating a can of nails

khrapovs:
I have updated the docstring with train_window math definition. 

> * for instance, line 682, it should be the test set, not the training set, right?

Yes, changed.

PR #3149, opened 2022-07-31
[ENH] Adding `VAR._predict_intervals`

lbventura:
#### Reference Issues/PRs

Adds `predict_intervals` to the VAR module.

Adds `predict_intervals` to the VAR module.

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/forecasting/var.py
+++ b/sktime/forecasting/var.py
@@ -18,6 +18,7 @@ class VAR(_StatsModelsAdapter):
     """
     A VAR model is a generalisation of the univariate autoregressive.
 
+    Direct interface for `statsmodels.tsa.statespace.dynamic_factor`
     A model for forecasting a vector of time series[1].
 
     Parameters
--- a/sktime/forecasting/var.py
+++ b/sktime/forecasting/var.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 """Vector Auto Regressor."""
 __all__ = ["VAR"]
-__author__ = ["thayeylolu", "aiwalter"]
+__author__ = ["thayeylolu", "aiwalter", "lbventura"]
 
 import numpy as np
 import pandas as pd
@@ -75,6 +75,7 @@ class VAR(_StatsModelsAdapter):
         "requires-fh-in-fit": False,
         "univariate-only": False,
         "ignores-exogeneous-X": False,
+        "capability:pred_int": True,
     }
 
     def __init__(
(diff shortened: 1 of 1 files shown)


sktime/forecasting/var.py
```diff
@@ -184,6 +185,112 @@ def _predict(self, fh, X=None):
...
+        pre_output_df = concat_df[list(concat_df_columns)]
+
+        pre_output_df_2 = pd.DataFrame(
+            pre_output_df.values,
+            columns=pd.MultiIndex.from_tuples(
+                [col.split(" ") for col in pre_output_df]
+            ),
+        )
+
+        final_columns = [
```
fkiraly, line 277: ising `zip` might be more elegant than a loop?

sktime/forecasting/var.py
```diff
@@ -184,6 +185,112 @@ def _predict(self, fh, X=None):
...
+        steps = fh_int[-1]
+        n_lags = model.k_ar
+
+        y_cols_no_space = [str(col).replace(" ", "") for col in self._y.columns]
+
+        df_list = []
+
+        for cov in coverage:
+
+            alpha = -0.5 * cov + 0.5
```
fkiraly, line 237: the docs say "significance level", so I think we should have `alpha = 1 - cov`, instead

fkiraly (changes requested):
Nice, great work!

Small but blocking detail: I think your formula for `alpha` is wrong.

According to the `statsmodels` docs, `alpha` must be the significance level (often denoted `p`), and not the quantile point of the lower bound, which is the formula you used.

sktime/forecasting/var.py
lbventura: Corrected in the upcoming PR

sktime/forecasting/var.py
lbventura: I am not sure how you suggest using `zip`, wouldn't `zip` just compute the internal product of the three lists (?) when we actually want the Cartesian product. I did something neater (but more complicated and needs a package import) in the upcoming PR.

lbventura:
> Nice, great work!
> 
> Small but blocking detail: I think your formula for `alpha` is wrong.
> 
> According to the `statsmodels` docs, `alpha` must be the significance level (often denoted `p`), and not the quantile point of the lower bound, which is the formula you used.

Thanks for highlighting this, I should have checked this the first place. I went back today and confirmed this was in fact the case.

sktime/forecasting/var.py
```diff
@@ -184,6 +188,126 @@ def _predict(self, fh, X=None):
...
+        pre_output_df = concat_df[concat_df_columns]
+
+        pre_output_df_2 = pd.DataFrame(
+            pre_output_df.values,
+            columns=pd.MultiIndex.from_tuples(
+                [col.split(" ") for col in pre_output_df]
+            ),
+        )
+
+        # final_columns = [
```
fkiraly, line 284: please remove commented out code

sktime/forecasting/var.py
```diff
@@ -75,6 +78,7 @@ class VAR(_StatsModelsAdapter):
         "requires-fh-in-fit": False,
         "univariate-only": False,
         "ignores-exogeneous-X": False,
```
fkiraly, line 80: saym is this correct? It does not look like `X` is used inside.

I.e., this is VAR, not VAR-X or VARMAX. Right?

fkiraly: while we're here, would also be great if the class docstring could say which statsmodels object it interfaces.

fkiraly (approved):
Looks good now!

Let me know if/when you want it merged, or whether you want to look after the minor details before that.

sktime/forecasting/var.py
fkiraly: ah, you are right, sorry. You want `itertools.product`, not `zip`.

fkiraly: ... which you actually used, all fine then!

lbventura:
> Looks good now!
> 
> Let me know if/when you want it merged, or whether you want to look after the minor details before that.

I'll handle these tonight and let you know  👏

fkiraly (approved):
Great, thanks!

PR #3154, opened 2022-08-01
[GOV] clarifications on algorithm maintainers, CC decision making, arbitration by CoCC

fkiraly:
This PR is a proposal following my action to clarify certain points in the governance document. For discussion at the next CC meetings.

Clarifications/changes.

1. the role of an algorithm maintainer is better clarified
    * the responsibility they have with respect to classes/estimators is clarified
    * it is clarified that the responsibility and ownership is to/of python classes, not abstract concepts
    * in particular, algorithm maintainers have no rights/responsibilities regarding other instances of the same abstract concept
    * algorithm maintainers are better delineated from core developers

2. clarification of core developer role
    * common expectations are outlined, to separate this from algorithm maintainer role better
    * outlined expectations do not change the formal appointment process

New/changed rules:'
* inactivity rule for algorithm maintainer, similar to core developer
* it is clarified that CoC committee can (and is expected to) judicate on how various rules are being interpreted. This is to avoid deadlocks and back/forth as on occasion in the past where core devs would disagree on interpretations of rules. Note: this widens the role of CoC committee to judication on matters not restricted to CoC violations.
* clarified end of tenure from CC by removal. Increased to 80% majority from current default of 2/3 (means: no removals possible until we have at least 5 CC members and 4 or more of them agree).

Typos:
* CoC section accidentally referred to CC, fixed
* apostrophes fixed


aiwalter (changes requested):
had a quick look to this PR, I dont see a need to introduce a new formal role. Everyone who contributes significantly to an algo can put himself to the `CODEOWNER` file. Doing this requires a PR which has to be reviewed by a core-dev, so this process is in my opinion good enough.

fkiraly:
>  I dont see a need to introduce a new formal role. 

I think you may have skimmed the content diagonally here?

This PR does not introduces a new role - algorithm maintainer is already on the list that we have.
It just fleshes it out, i.e., expectations etc.

I'm open to stating different expectations here, of course. Just repeating, this is not *introducing* a new role, which might have been a misunderstanding.

There are also other changes, outlined at the top - the more substantial change, governance-wise, is the CoCC being empowered as arbitration body on certain procedural matters.

aiwalter:
ther current definition is 

> sktime's algorithm maintainers as specified in GOVERNANCE.md

Which is fine with me. I was not aware that people who are listed in this file are considered as "algorithm maintainers" but thats totally fine with me. I still dont see a need to further formalize this role. But happy to hear also opinion of @GuzalBulatova 

Other parts of this PR look good, so if we can split that we can merge the other changes already?

fkiraly:
@aiwalter, as per our discussion at the last CC meeting (Oct 18), I have split this PR in two parts:

* #3675, which contains the parts that you thought were uncontentious and fine to adopt.
* #3676, which contains the parts on algorithm maintainers that you thought were contentious and needed more discussion.

fkiraly:
closed in favour of #3675 and #3676

Issue #3099, opened 2022-07-26
[ENH] Further Improve HMM Estimator


miraep8:
Currently HMM Estimator has been implemented (#2855) ! :)  However there are some changes/improvements I plan to add. Will add and check things off as I get to them:

- [ ] add Baum-Welch logic to `_fit` so that the parameters which are used in the Viterbi algorithm can be dynamically extracted from the data. 
- [ ] Expand support to multi-dimensional data (should be a simple extension!)
- [ ] Move beyond the assumption of equal time spacing
- [ ] Add support for partially observed data
- [ ] For a "fitted" HMM add the option to "sample" from the model.
- [ ] Write a wrapper for hmmlearn
- [ ] write an example use case/notebook of how to use it
- [ ] Look at implementing other statistical methods eg CRF

FYI @lmmentel

Issue #484, opened 2020-11-13
[DOC] Fill in missing docstrings


mloning:
#### Describe the issue linked to the documentation

Our documentation is a bit patchy in some places, filling in the missing information would be a great help! Thanks to @chrisholder for creating the list! 

| File                                                          | Comments                                                                                                                                                             | Github link                                                                                                                                                                                                                                       |
| ------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| sktime/benchmarking/data.py                                   | Multiple methods classes <br>undocumented                                                                                                                            | [https://github.com/alan-turing-institute/sktime/blob/master/sktime/benchmarking/data.py](https://github.com/alan-turing-institute/sktime/blob/master/sktime/benchmarking/data.py)                                                                |
| sktime/benchmarking/evaluation.py                             | Parameters for constructor and <br>some methods missing                                                                                                              | [https://github.com/alan-turing-institute/sktime/blob/master/sktime/benchmarking/evaluation.py](https://github.com/alan-turing-institute/sktime/blob/master/sktime/benchmarking/evaluation.py)                                                    |
| sktime/benchmarking/metrics.py                                | Multiple methods classes <br>undocumented                                                                                                                            | [https://github.com/alan-turing-institute/sktime/blob/master/sktime/benchmarking/metrics.py](https://github.com/alan-turing-institute/sktime/blob/master/sktime/benchmarking/metrics.py)                                                          |
| sktime/benchmarking/orchestration.py                          | Multiple methods classes <br>undocumented                                                                                                                            | [https://github.com/alan-turing-institute/sktime/blob/master/sktime/benchmarking/orchestration.py](https://github.com/alan-turing-institute/sktime/blob/master/sktime/benchmarking/orchestration.py)                                              |
| sktime/benchmarking/results.py                                | No docstrings                                                                                                                                                        | [https://github.com/alan-turing-institute/sktime/blob/master/sktime/benchmarking/results.py](https://github.com/alan-turing-institute/sktime/blob/master/sktime/benchmarking/results.py)                                                          |
[...]

PR #3156, opened 2022-08-01
[ENH] Wrapper for hmmlearn

miraep8:
Related to my issue #3099 on how I plan to further improve our support of HMM-like algorithms in `sktime` and also the issue I opened in `hmmlearn` ([#484](https://github.com/hmmlearn/hmmlearn/issues/484))

Will make a non-draft PR when it is ready for review.  At a high level the plan is o have a simple base class that interfaces with the `BaseAnnotator` class and provides interface compatibility.  Then each HMM-like class which inherits from this base class simply imports the relevant HMM estimator from `hmmlearn`. 

Still to do:
- make sure parameter setting is consistent/works!
- add some more tests that things work as expected.


sktime/tests/test_all_estimators.py
```diff
@@ -878,7 +877,8 @@ def test_constructor(self, estimator_class):
         assert isinstance(estimator, estimator_class)
 
         # Ensure that each parameter is set in init
-        init_params = _get_args(type(estimator).__init__)
+        # init_params = _get_args(type(estimator).__init__)
+        init_params = vars(estimator)
```
miraep8, line 881: Whoops, this was a small workaround I had put into place to deal with the fact that I had constructed a different __init__ (via define).  Will fix this before merging

sktime/annotation/hmm_learn.py
```diff
@@ -0,0 +1,154 @@
...
+
+    def sample(self, n_samples=1, random_state=None, currstate=None):
+        """Interface class which allows users to sample from their HMM."""
+        return self._hmm_estimator.sample(n_samples, random_state, currstate)
+
+
+_check_soft_dependencies("hmmlearn.hmm", severity="warning")
+
+
+@define
```
fkiraly, line 57: hm, you *had* to sneak this in, didn't you?
I suppose it's ok in an experimental part of sktime, let's see whether this causes any problems...

Can you check whether the top level of the test suite is executed on `GaussianHMM`? This should be the case if it ultimately inherits from `BaseObject`.

sktime/tests/test_all_estimators.py
fkiraly: ok, I just scrolled down and do see that the little `define` thing causes errors...

we can't change the core interface contract because of that.
`vars`, for instance, does not return all init args, so this change breaks the test!

fkiraly (changes requested):
All looks good, except the implicit change to the interface contract in "object".

Is there a way without changing the heart of sktime interface specs?

Worst case, lots of typing of `self.=`...
for a slightly less exhausting route, have a look at the `ARIMA` class.

sktime/tests/test_all_estimators.py
miraep8: I totally agree, it was a spandrel I left in by mistake.  Currently working to see if I can still make the `attrs` work despite this though! :)

sktime/annotation/hmm_learn.py
miraep8: Yes, I do believe it ultimately ends up running the test suite on `GaussianHMM` given the fact that I had to debug a few of them :)

sktime/tests/test_all_estimators.py
fkiraly: well, I've also tried and failed...
https://github.com/alan-turing-institute/sktime/pull/2735

sktime/tests/test_all_estimators.py
```diff
@@ -879,7 +879,7 @@ def test_constructor(self, estimator_class):
 
         # Ensure that each parameter is set in init
         init_params = _get_args(type(estimator).__init__)
-        invalid_attr = set(init_params) - set(vars(estimator)) - {"self"}
+        invalid_attr = set(init_params) - set(dir(estimator)) - {"self"}
```
miraep8, line 882: with `attrs`, `vars` seems not to work properly.  I think this has something to do with it not playing well with the `__dict__` function that `vars` seems to rely on.  They do get added as valid attributes in the `__init__` however, which is captured by using `dir` instead, and I think is what we are really interested in???

They are slightly different functions though, as is summarized a bit [here](https://stackoverflow.com/questions/980249/difference-between-dir-and-vars-keys-in-python).  Let me know if people think this is a reasonable change or not.  (I believe it should still execute this test as expected for other estimators, ie if something is passed in init, but is not added as an attribute then this test**shoudl**still fail as expected).

fkiraly: could we separate out the `attrs` business from this PR? In a separate PR?

I think that's a different issue - modification of the base object interface.

The separate PR should be accompanied by tests and dummy examples for the `attr` variant.

miraep8: Yea ok, I will leave this PR as the `attrs` PR but will open another HMM wrapper PR without it.

miraep8:
Removed the attrs bit now and still seems to work ok

fkiraly (changes requested):
Looks good, but your doctest is failing.

If you want expressions that continue over multiple lines, you need to use `...` from the 2nd line on, instead of `>>>`

sktime/annotation/hmm_learn.py
```diff
@@ -0,0 +1,232 @@
...
+        self.n_iter = n_iter
+        self.tol = tol
+        self.verbose = verbose
+        self.params = params
+        self.init_params = init_params
+        self.implementation = implementation
+        super(GaussianHMM, self).__init__()
+
+    def _fit(self, X, Y=None):
+        # import inside _fit to avoid hard dependency.
```
KatieBuc, line 188: I'm wondering what type of expected type X is (i.e. pd.DaraFrame/Series or np.Array) - a docstring?

sktime/annotation/hmm_learn.py
```diff
@@ -0,0 +1,232 @@
...
+        if series:
+            X = (X.to_numpy()).reshape((-1, 1))
+        self._hmm_estimator = self._hmm_estimator.fit(X)
+        return self
+
+    def _predict(self, X):
+        series = isinstance(X, pd.Series)
+        if series:
+            index = X.index
+            X = (X.to_numpy()).reshape((-1, 1))
```
KatieBuc, line 42: I like how you've taken care of reshaping the data in the Base Class, is this also true for when inputs are numpy arrays?

sktime/annotation/hmm_learn.py
```diff
@@ -0,0 +1,232 @@
...
+        * (n_components, n_features, n_features)  if "full",
+        * (n_features, n_features)                if "tied".
+
+    Examples
+    --------
+    >>> from sktime.annotation.hmm_learn import GaussianHMM
+    >>> from sktime.annotation.datagen import piecewise_normal
+    >>> data = piecewise_normal(
+    ...    means=[2, 4, 1], lengths=[10, 35, 40], random_state=7
+    ...    ).reshape((-1, 1))
```
KatieBuc, line 141: ah but then do we need a reshape here?

KatieBuc:
Looks great @miraep8!

sktime/annotation/hmm_learn.py
miraep8: Yep, there should be some docstrings here.  Adding them now!

fkiraly (changes requested):
Code looks good to me!

May I ask for some changes re soft dependency handling, the general rule is "don't import soft dependencies directly, anywhere". So:

* in the test, use a `pytest.skipif` and import `hmmlearn` inside the test. Compare how `pmdarima` is handled, search for `pytest.skipif`.
* set the `python_dependencies` tag in the new classes (base class is enough)
* skip the example doctest (search for `# doctest` to see examples). If this reduces coverage, move the examples into another test decorated with `pytest.skipif`.

miraep8:
Since GMMHMM has already been merged and that PR was a superset of this one I am going to close this PR.

fkiraly:
I see, I should have merged them in a sequence.

Sorry for this.

Next time, may I kindly ask to write sth explicit like "contains PR # 3156" in the other PR instead of sth flowery like "building on success in #3156" (which can simply mean you are following the same pattern).

PR #3171, opened 2022-08-04
[ENH] Adding CNTC network for classification from sktime-dl

### LINKED ISSUE #3153: [ENH] Adding CNTC DL model for classification
In line with discussions and #2894 , adding CNTC model for classification from sktime-dl

AurumnPegasus:
#### Reference Issues/PRs

Fixes #3153.

One thing I was trying was to make the code a bit more readable (compared to what it was on sktime-dl). So let me know any changes I can make towards that goal.

##### For all contributions

### DIFF
--- a/sktime/networks/cntc.py
+++ b/sktime/networks/cntc.py
@@ -1,0 +1,234 @@
+# -*- coding: utf-8 -*-
+"""a combination of CCNN and CLSTM as feature extractors."""
+
+__author__ = ["James-Large, Withington, TonyBagnall", "AurumnPegasus"]
+
+from sktime.networks.base import BaseDeepNetwork
+from sktime.utils.validation._dependencies import _check_dl_dependencies
+
+_check_dl_dependencies(severity="warning")
+
+
+class CNNNetwork(BaseDeepNetwork):
+    """Combining contextual neural networks for time series classification.
+
+    Adapted from the implementation used in [1]
+
+    Parameters
+    ----------
+    kernel_size     : int, default = 7
+        specifying the length of the 1D convolution window
+    avg_pool_size   : int, default = 3
+        size of the average pooling windows
+    n_conv_layers   : int, default = 2
+        the number of convolutional plus average pooling layers
+    filter_sizes    : array of int, shape = (nb_conv_layers)
+    activation      : string, default = sigmoid
+        keras activation function
+    random_state    : int, default = 0
+        seed to any needed random actions
+
+    Notes
+    -----
+    Adapted from the implementation from Fullah et. al
+    https://github.com/AmaduFullah/CNTC_MODEL/blob/master/cntc.ipynb
+
+    References
+    ----------
+    .. [1] Network originally defined in:
+        @article{FULLAHKAMARA202057,
+        title = {Combining contextual neural networks for time series classification},
+        journal = {Neurocomputing},
+        volume = {384},
+        pages = {57-66},
+        year = {2020},
+        issn = {0925-2312},
+        doi = {https://doi.org/10.1016/j.neucom.2019.10.113},
+        url = {https://www.sciencedirect.com/science/article/pii/S0925231219316364},
+        author = {Amadu {Fullah Kamara} and Enhong Chen and Qi Liu and Zhen Pan},
+        keywords = {Time series classification, Contextual convolutional neural
+            networks, Contextual long short-term memory, Attention, Multilayer
+            perceptron},
+       }
+    """
+
+    def __init__(self, rnn_layer, filter_sizes, kernal_sizes, lstm_size):
+        self.rnn_layer = rnn_layer
+        self.filter_sizes = filter_sizes
+        self.kernal_sizes = kernal_sizes
+        self.lstm_size = lstm_size
+        _check_dl_dependencies(severity="error")
+
+    def build_network(self, input_shape, **kwargs):
+        """Construct a network and return its input and output layers.
+
+        Arguments
+        ---------
+        input_shape: tuple,
+            The shape of the data fed into the input layer
+
+        Returns
+        -------
+        input_layer: a keras layer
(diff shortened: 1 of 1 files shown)


sktime/networks/cntc.py
```diff
@@ -0,0 +1,244 @@
+# -*- coding: utf-8 -*-
+"""a combination of CCNN and CLSTM as feature extractors."""
+
+__author__ = ["James-Large, Withington, TonyBagnall", "AurumnPegasus"]
```
fkiraly, line 4: typos

sktime/networks/cntc.py
```diff
@@ -0,0 +1,244 @@
...
+
+from sktime.networks.base import BaseDeepNetwork
+from sktime.utils.validation._dependencies import _check_dl_dependencies
+
+_check_dl_dependencies(severity="warning")
+
+
+class CNNNetwork(BaseDeepNetwork):
```
fkiraly, line 12: if you have base class for neural network classifiers, why are you not inheriting from it?
`BaseDeepClassifier`, not `BaseDeepNetwork`

fkiraly: there is already a class named `CNNNetwork`

sktime/networks/cntc.py
```diff
@@ -0,0 +1,244 @@
...
+                Time series data is fed into a CLSTM and CCNN networks simultaneously
+                and is perceived differently. In the CLSTM block, the input data is
+                viewed as a multivariate time series with a single time stamp. In
+                contrast, the CCNN block receives univariate data with numerous time
+                stamps
+            "
+
+        Arguments
+        ---------
+        X: tuple,
```
fkiraly, line 196: plase be more specific on the format, surely the input cannot be any tuple, and the output cannot be any tuple

sktime/networks/cntc.py
```diff
@@ -0,0 +1,244 @@
...
+        self.lstm_size = lstm_size
+        self.dense_size = dense_size
+        _check_dl_dependencies(severity="error")
+
+    def build_network(self, input_shape, **kwargs):
+        """Construct a network and return its input and output layers.
+
+        Arguments
+        ---------
+        input_shape: tuple,
```
fkiraly, line 77: be more specific, please

sktime/networks/cntc.py
fkiraly: ok, after discussion understood why this is inheriting from `BaseDeepNetwork` - the classifier would be somewhere else

sktime/networks/cntc.py
AurumnPegasus: About data type specificity, I am still not very clear about them. I will do a run based on test dataset to get better idea. I have kept it empty for now for the same reason.

sktime/networks/cntc.py
AurumnPegasus: Same as above

sktime/networks/cntc.py
AurumnPegasus: Can you please point them out? is it a grammatical mistake or typo?

AurumnPegasus:
I have a basic design question which I wanted to confirm if I am doing correctly / what can be done:
1. For `cntc`, the input `X` needs to be modified into `[X2, X, X]`.
2. `X2` is determined by a function `prepare_input` in `CNTCClassifer`
3. It works correctly for fit, since I am able to modify the input before sending it to the model.
4. For predictions, if I were to write `_predict_proba` (get `X2` here) in `CNTCClassifier` (and then call `super()._predict_proba()), the issue is that it has a line `X = X.transpose(0, 2, 1)` , which needs to be done before calculation of `X2`

Is there a way to take care of this? My current approach would be to completely overwrite both `_predict` and `_predict_proba` in `CNTCClassifier` and not call the functions in the super class at all, but was wondering if there is a better solution to this?

AurumnPegasus:
This particular estimator requires adding a soft dependancy (`keras_self_attention`)

fkiraly (changes requested):
Looks good!

One change request, we need to ensure that all tests run without soft dependencies, and that includes the doctests. Kindly skip all doctests using doctest skip.
If that reduces coverage, please move the tests to pytest test functions, and decorate them with `pytest.skipif` that skips if the required dependency is not present.

fkiraly:
@miraep8 wrote it up here:
https://github.com/alan-turing-institute/sktime/pull/3416
(also in the latest doc now)

AurumnPegasus:
@fkiraly 
Done, please check and let me know if this is correct?

sktime/networks/cntc.py
```diff
@@ -0,0 +1,187 @@
...
+__author__ = ["James-Large", "Withington", "TonyBagnall", "AurumnPegasus"]
+
+from sktime.networks.base import BaseDeepNetwork
+from sktime.utils.validation._dependencies import (
+    _check_dl_dependencies,
+    _check_soft_dependencies,
+)
+
+_check_dl_dependencies(severity="warning")
+_check_soft_dependencies(severity="warning")
```
fkiraly, line 13: this should have the arg `keras_self_attention`

sktime/networks/cntc.py
```diff
@@ -0,0 +1,187 @@
...
+        random_state=0,
+        rnn_layer=64,
+        filter_sizes=(16, 8),
+        kernel_sizes=(1, 1),
+        lstm_size=8,
+        dense_size=64,
+    ):
+        super(CNTCNetwork, self).__init__()
+        _check_dl_dependencies(severity="error")
+        _check_soft_dependencies(severity="error")
```
fkiraly, line 72: same here

fkiraly (changes requested):
Thanks!

Almost done - please use the `_check_soft_dependencies` function correctly (see above).

PR #3178, opened 2022-08-04
[MNT] skip known failure case for `VARMAX`

fkiraly:
One specific test for `VARMAX` on windows terminates with an exception:
https://github.com/alan-turing-institute/sktime/issues/3176

Since this happens frequently, removing the test until we understand what is going on.


fkiraly:
will emergency merge

PR #3197, opened 2022-08-06
DIAGNOSTIC - DO NOT MERGE - partition test design with intentionally added failures

fkiraly:
This is https://github.com/alan-turing-institute/sktime/pull/3137 with two intentionally added failures (in `Catch22Classifier` and `NaiveForecaster`), in order to test that the failures are indeed picked up by the partition design.


fkiraly:
works!

PR #3200, opened 2022-08-06
[DOC] added more detail to step 4 of high-level steps to implementing an es…

kcc-lion:
…timator

#### Reference Issues/PRs

No Reference

I added some more detail to the high-level steps outline on how to implement an estimator. More specifically, I added a reference to the section on how to use the `check_estimator` facility and the tutorial on checking interface conformance. These references make sense as both resources should be introduced close to the top of the page.

No


fkiraly (changes requested):
Thanks!

I would ask for these changes:

* keep the bullet points short and crisp, no distractions. E.g., write sth like "run the ``sktime`` test suite and/or the ``check_estimator`` utility on the implemented estimator"
* add the reference to the tutorial at the bottom
* adding links to the sections is a good idea, but avoid writing too much text for the "go here etc". Better: just add hyperlinks to words, or a very short "see here"

fkiraly (approved):
Nice and crisp. Thanks!

PR #3204, opened 2022-08-06
[BUG] allowing single class case in sklearn classifiers (trees/forests)

### LINKED ISSUE #3202: [BUG] `DrCIF` failure due to single class in sub-sample
`DrCIF` fails stochastically under windows, on `test_contracted_drcif`, the line `assert len(drcif.estimators_) > 1`, because `drcif.estimators_` ends up having length 1.

Do you have an idea what is going on here, @MatthewMiddlehurst?

We are fixing the random seed, but the resulting samples might change with versions of the underlying packages or the operating system.

Example here: https://github.com/alan-turing-institute/sktime/runs/7706482325?check_suite_focus=true

fkiraly:
This allows single class in training set in the sklearn classifiers `RotationForest` and `ContinuousIntervalTree`.

Fixes #3202.

### DIFF
--- a/sktime/classification/sklearn/_continuous_interval_tree.py
+++ b/sktime/classification/sklearn/_continuous_interval_tree.py
@@ -126,8 +126,10 @@ class ContinuousIntervalTree(BaseEstimator):
         for index, classVal in enumerate(self.classes_):
             self._class_dictionary[classVal] = index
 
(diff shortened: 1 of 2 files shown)


fkiraly:
Merging to address an existing failure on `main` before release.

Issue #3206, opened 2022-08-07
[BUG] contracted classifier tests fail stochastically on assertion "number estimators > 1"


fkiraly:
The contracted classifier tests fail stochastically on a common assertion "number estimators > 1". The failure is on `main`.

Estimators in questions are:
* `DrCIF`
* `ShapeletTransformClassifier`
* `TemporalDictionaryEnsemble`

I have commented out the assertion, since, I suppose, stochastically the runtime of a single estimator component can be so high that the contract expires after the first component, so there isn't a good reason to expect that "number estimator > 1"? It should just be > 0 (which is tested indirectly already by asking for a prediction in the test suite).

I am commenting the assertion out temporarily to keep failures from `main`, and leave this issue open to resolve this more systematically.

TonyBagnall:
I dont think its a bug, its just how we configure contracting. The assertion should be "n_estimators>0".

If you set a small contract, the classifiers are set up to build at least one estimator, even if it breaks the contract. We deem it better to break the contract that not build a classifier at all and break the experiment. The use case for contracting is for large problems that may take days or weeks to run fully. Contracting and checkpointing allow you to control the build time and restart the classifier later if required, but are just guides, it is not simple or desirable to enforce a strict upper limit, and setting very short contracts makes little sense in practice.

fkiraly:
thanks for confirming, would appreciate a fix.

PR #3207, opened 2022-08-07
[BUG] skip check for no. estimators in contracted classifiers

fkiraly:
This PR removes a stochastic failure from `main` coming from checks in tests of contracted classifiers, see #3206.

#3206 should be used for finding a better resolution of the failure, this PR simply removes the problem from `main`.


fkiraly:
Merging to remove existing stochastic failure on `main`

PR #3211, opened 2022-08-07
[DOC] cleaning up the user docs

fkiraly:
Small changes to improve the documentation by pointing towards the example notebooks, removing outdated notebooks from the sphinx docs.

Goes towards https://github.com/alan-turing-institute/sktime/issues/2127

### DIFF
--- a/docs/source/tutorials.rst
+++ b/docs/source/tutorials.rst
@@ -3,8 +3,9 @@
 Tutorials
 =========
 
-We have a number of tutorial notebooks. If you like to run them interactively, you can launch them on Binder without
-having to install anything.
+We have a number of tutorial notebooks. If you like to run them interactively, you can
+`launch them on binder <https://mybinder.org/v2/gh/alan-turing-institute/sktime/main?filepath=examples>`_
+without having to install anything.
 
 We assume basic familiarity with `scikit-learn`_. If you haven’t worked with scikit-learn before, check out their
 `getting-started guide`_.
@@ -17,4 +18,4 @@ The notebook files can be found `here <https://github.com/alan-turing-institute/
 .. nbgallery::
     :glob:
 
-    examples/*
+    https://github.com/alan-turing-institute/sktime/blob/main/examples


aiwalter (commented):
where is `docs/source/examples/feature_extraction_with_tsfresh.ipynb` now living, is it remoed completely?

aiwalter:
I think we should not link tutorials from other repos as they will not be in synch with `main` in future.

fkiraly:
> where is `docs/source/examples/feature_extraction_with_tsfresh.ipynb` now living, is it remoed completely?

it still lives in `examples` (in root), together with 12 or so other notebooks that are very algorithm specific but it is no longer displayed in the docs.

I thought it might be better to only show the "big" user guides, and leave the algorithm specific vignettes out of it.

If you disagree, I don´t mind, but could then someone with unix or mac run the `symlink_examples.sh` and commit the changes?

fkiraly:
> I think we should not link tutorials from other repos as they will not be in sync with `main` in future. 

I would invoke the principle "accept PR if they are an improvement".

Links to existing, external repos that are currently in sync, are preferable above the empty pages that were there since 2 years.

"being in sync" is also partially covered by some of the external notebooks being in the repository, and the binder set-up in the external repositories.

aiwalter:
> I would invoke the principle "accept PR if they are an improvement - don't reject PR because they are not perfect".

I might be an improvement for now as the notebooks might still pass on `main`, but they will get outdated. Therefore, I am against linking static notebooks from other repos. We should only link notebooks that are in the same repo, I thought also we wanted to migrate the PyData notebooks to main repo so they will always be tested.

fkiraly:
> Therefore, I am against linking static notebooks from other repos.

I'm not sure what the alternative would be.

How would linking to the youtube videos work then? Those videos were made with the notebooks that are in those repositories. There is no way to update those notebooks, *and also* the video, without re-recording it at some event in the future.

How about the following compromise: allow linking to those repositories as long as the notebooks in there are covered by `sktime` tests, here, and raise a "red alert" if someone changes the notebooks in `sktime` at the danger of getting them out of sync.

The notebooks all showcase some stable functionality in sktime, so not much should change.
For instance, the "probabilistic forecasting" notebook is already 1:1 in `sktime`.

More concretely, I would suggest:

* merge as is, so we have some decent user documentation in this week's release
* we'll build the safeguards to ensure those notebooks are evergreen, as descried above
* allow a short period (few weeks or so?) where the safeguards are not in place, because links are better than empty user docs

aiwalter:
do you mind to remove the links to external repo? For the the videos and notebooks from talks are a snapshot of `sktime`, so its not good to guide users there. Linking the videos on README is still ok as user is aware that a video can be outdated.

fkiraly:
I would disagree with "bad idea to guide users to videos". For many people, they are primary entry points and helpful as a means of explanation!

How about I add a warning that these are snapshots, and let the user decide? There is a link to the user guides as well, on the same page, for those who want the notebook/binder experience instead.

PR #3218, opened 2022-08-09
[ENH] Evaluate_classification v2

### LINKED ISSUE #3172: [ENH] - Evaluation method to for classification
At the moment, sktime doesn not have an `evaluate` method for claculating metrics for classfication task. 

A clear and concise description of what you want to happen, ideally taking into consideration the existing toolbox design, classes and methods.
The `evaluate` method will take as argument a classifier, X, y, metrics and cv-index and produce a computed metrics for each cv fold.

Doing things manually, but it's too tedious and not able to use with other functionality.

**Additional context**

TNTran92:
#### Reference Issues/PRs

Fix #3172 

Implement an `evaluate` for classification. This methods allows quick calculation of classification task metrics. Furthermore, it allows us to write kotsu-type benchmarking tool for classification task.

None

- Design and implementation

#### Any other comments?

- Unit test added
- Example docstring added

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/classification/model_evaluation/_function.py
+++ b/sktime/classification/model_evaluation/_function.py
@@ -1,0 +1,86 @@
+# -*- coding: utf-8 -*-
+from typing import Optional
+
+import pandas as pd
+from sklearn.metrics import get_scorer, get_scorer_names
+from sklearn.model_selection import BaseShuffleSplit, cross_validate
+
+from sktime.classification.base import BaseClassifier
+
+
+def evaluate_classification(
+    classifier: BaseClassifier,
+    X,
+    y,
+    scoring=None,
+    cv: Optional[BaseShuffleSplit] = None,
+    return_data=True,
+):
+    """Evaluate classifier using sklearn cross-validation.
+
+    Parameters
+    ----------
+    classifier : Any sktime classifier
+    X : Nested Univariate Pandas DataFrame
+        All training data
+    y : np.array
+        Np.array of labels.
+    scoring : name of classification metrics, default="accuracy_score."
+        For a complete list of acceptable metrics, see
+        https://scikit-learn.org/stable/modules/model_evaluation.html
+    cv : Cross-validation splitter
+        Result of cv-splitters in sklearn.model_selection
+    return_data : bool, default=False
+        Returns three additional columns in the DataFrame, by default False.
+        The cells of the columns contain each a pd.Series for y_train,
+        y_pred, y_test.
+
+    Returns
+    -------
+    pd.DataFrame
+        DataFrame that contains several columns with information regarding each
+        fold of the classifier.
+
+    >>> from sktime.classification.model_evaluation import evaluate_classification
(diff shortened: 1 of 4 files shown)


TNTran92:
@fkiraly @aiwalter This is ready for review

pyproject.toml
```diff
@@ -49,6 +49,7 @@ dependencies = [
     "scikit-learn>=0.24.0,<1.2.0",
     "statsmodels>=0.12.1",
     "scipy<1.9.0",
+    "backoff",
```
fkiraly, line 52: please remove backoff from the core dependencies

sktime/classification/model_evaluation/_function.py
```diff
@@ -0,0 +1,84 @@
...
+        _SCORERS = _scorer.SCORERS
+        list_of_scorer = get_scorer_names()
+    else:
+        _SCORERS = _scorer._SCORERS
+        list_of_scorer = get_scorer_names()
+
+    if scoring is None:
+        scoring = "accuracy"
+    if scoring not in list_of_scorer:
+        print("Metrics is not valid.")  # noqa
```
fkiraly, line 69: should this not be an exception instead?

sktime/classification/model_evaluation/_function.py
```diff
@@ -0,0 +1,84 @@
...
+        print("See sklearn.metrics.get_scorer_names()")  # noqa
+        print("for acceptable classsification metrics.")  # noqa
+    scoring_method = get_scorer(scoring)
+    scores = cross_validate(
```
fkiraly, line 73: I would use the current `fit_predict` instead, inside `BaseClassifier`. That handles input checks, conversions, etc

fkiraly (changes requested):
Thanks, looks good!

Left a few comments.

Blocking ones:
* `backoff` should not be a core dependency. I assume this is coming from here: https://github.com/alan-turing-institute/sktime/pull/3242
so it should be fixed in the other PR, and this PR should not depend on it.
* for the cross-validation, I would use `sktime` instad of `sklearn`. `fit_predict` of the classifiers should have what you need. That also allows to make the input format general (any mtype)

pyproject.toml
TNTran92: backoff removed

sktime/classification/model_evaluation/_function.py
TNTran92: One issue I had with that option is that `fit_predict` doesn't handle metrics. Supposing we do that, there will have to be a methods like `accuracy_score`, `f1_score`, etc to get requested scoring methods. sklearn handles this issue by having method `get_scorer` that returns `make_scorer` I don't think we have any for classification

fkiraly: yes, what I meant is that you get `y_pred` from `fit_predict(cv, etc)`, and then stick it into `scoring_metic` with `y`

PR #3225, opened 2022-08-09
[BUG] fixing bugs in metrics base classes and custom performance metric

### LINKED ISSUE #3221: [BUG] make_forecasting_scorer attribute error
**Describe the bug**

When trying to create a custom forecasting scorer with `make_forecasting_scorer`, it fails with `AttributeError: '_DynamicForecastingErrorMetric' object has no attribute 'lower_is_better'`.

**To Reproduce**

```python
from sktime.performance_metrics.forecasting import make_forecasting_scorer
import numpy as np

y = load_airline()

def custom_mape(y_true , y_pred) -> float:

    eps = np.finfo(np.float64).eps

    result = np.mean(np.abs(y_true - y_pred) / np.maximum(np.abs(y_true), eps))

    return float(result)


fc_scorer = make_forecasting_scorer(
    func=custom_mape,
    name="custom_mape",
    greater_is_better=False,
)

fc_scorer(y,y)
```

**Expected behavior**

Creating custom_score should be done without any error.

**Additional context**

This was working as expected in earlier version IIRC. I think this is happens after `metric refactor` implementations.

**Versions**
<details>

System:
    python: 3.7.13 (default, Apr 24 2022, 01:04:09)  [GCC 7.5.0]
executable: /usr/bin/python3
   machine: Linux-5.4.188+-x86_64-with-Ubuntu-18.04-bionic

Python dependencies:
          pip: 21.1.3
   setuptools: 57.4.0
      sklearn: 1.0.2
       sktime: 0.13.0
  statsmodels: 0.13.2
        numpy: 1.21.6
        scipy: 1.7.3
       pandas: 1.3.5
   matplotlib: 3.2.2
       joblib: 1.1.0
        numba: 0.56.0
     pmdarima: 1.8.5
      tsfresh: None
</details>

fkiraly:
Fixes #3221.

There were two problems with the dynamic metric after refactoring:

* `lower_is_better` tag was only set if false
* `func` attribute was not properly written or invoked

Further bugs in the metrics base classes were discovered in the process of fixing:

* the `BaseForecastingMetric` did not call `super.__init__` in its `__init__`, breaking dynamic tag functionality
* the `BaseForecastingMetric` would overwrite `self.name` even if it was set by a child class already, leading to potential `sklearn` incompatibility (parameter get/set)

The bug was also not caught since the dynamic metric started with an underscore, so was not collected in tests.
This has been remedied by adding to tests:

* a version of the example in #3221
* the custom metric from #3221 as test parameters for the dynamic metric
* a full run of `check_estimator` on the resulting metric

### DIFF
--- a/sktime/performance_metrics/forecasting/_classes.py
+++ b/sktime/performance_metrics/forecasting/_classes.py
@@ -533,8 +533,7 @@ class _DynamicForecastingErrorMetric(BaseForecastingErrorMetricFunc):
         self.name = name
         super().__init__()
 
-        if not lower_is_better:
-            self.set_tags(**{"lower_is_better": False})
+        self.set_tags(**{"lower_is_better": lower_is_better})
 
 
 class _ScaledMetricTags:
--- a/sktime/performance_metrics/forecasting/_classes.py
+++ b/sktime/performance_metrics/forecasting/_classes.py
@@ -8,7 +8,7 @@ Classes named as ``*Error`` or ``*Loss`` return a value to minimize:
 the lower the better.
 """
 from copy import deepcopy
-from inspect import signature
+from inspect import isfunction, getfullargspec, signature
 from warnings import warn
 
 import numpy as np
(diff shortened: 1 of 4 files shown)


fkiraly:
Merging early to ensure inclusion in the release. Hope this is ok, @aiwalter.

PR #3231, opened 2022-08-10
[BUG] fix failure of some conversions in `_load_provided_dataset()`

### LINKED ISSUE #3169: [BUG] IndexError when selecting `return_X_y=False` and `return_type` is not referencing to a pandas dataframe.
**Describe the bug**

As per the discussion in #2697, it was [found](https://github.com/alan-turing-institute/sktime/issues/2697#issuecomment-1196290311) that `_load_provided_dataset()` must throw an error if `return_type` is not a reference to `pd.DataFrame` and `return_X_y=False`. This happens because of code appends the response variable into predictors under the column name 'class_val'.

**To Reproduce**

```python
import sktime
from sktime.datasets._data_io import _load_provided_dataset

def test_load_provided_dataset():
    name = "UnitTest"                                    # referencing to dataset available for testing in the directory 'datasets/data/'
    X, y = _load_provided_dataset(name, return_type='np2D', return_X_y=False)
    print(X.shape, y.shape)

if __name__ == "__main__":
    print("Staring program")
    test_load_provided_dataset()
    print("Done")
```

**Output of above code**
```
PS D:\Projects\sktime\sktime\sktime\datasets\tests> D:/Anaconda/envs/sktime-dev/python temp_test_data_io.py
Staring program
  File "temp_test_data_io.py", line 11, in <module>
    test_load_provided_dataset()
  File "temp_test_data_io.py", line 6, in test_load_provided_dataset
    X, y = _load_provided_dataset(name, return_type='np2D', return_X_y=False)
  File "d:\projects\sktime\sktime\sktime\datasets\_data_io.py", line 224, in _load_provided_dataset
    X["class_val"] = pd.Series(y)
IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices
```

**Expected behavior**

This clearly is an issue because we are not checking the instance type of X, which could be `np.array` or `pd.DataFrame`.

**Solutions(?)**
* _Solution 1: Check instance and append in the appropriate axis_
  Consider the following cases 
  - Univariate, 2D, equal length data (`X=np.random.randn(20, 24), y=np.random.randn(20,)`) then `output.shape=(20,25)`
  - Univariate 3D, equal length data (`X=np.random.randn(20,1,24), y=np.random.randn(20,)`) then `output.shape=(20,1,25)`
  - Multivariate, equal length data (`X=np.random.randn(20,6,24), y=np.random.randn(20,)`) then `output.shape=(20,6,25)`; requires tiling labels which might not be desirable.
 
  I think it should also work for unequal length data but I need confirmation. The only drawback of this solution apart from the repetition of y in the multivariate case is that it might be less intuitive.

* _Solution 2: Force convert the data into pandas dataframe and give a warning about the conversion._

**Final Comments**
- There are more solutions to this but I think these two are the most straightforward. I also have the code ready for the first solution and will send a pull request if that is chosen, otherwise I'll take a day's time to send a pull request if we move forward with any other solution.
- I can also add unit tests regarding the same, I don't know if having tests for private methods are a thing or not, but I think it won't hurt having more tests.

**Versions**
<details>

```
D:\Anaconda\envs\sktime-dev\lib\site-packages\_distutils_hack\__init__.py:33: UserWarning: Setuptools is replacing distutils.
  warnings.warn("Setuptools is replacing distutils.")

System:
    python: 3.8.13 | packaged by conda-forge | (default, Mar 25 2022, 05:59:00) [MSC v.1929 64 bit (AMD64)]
executable: D:\Anaconda\envs\sktime-dev\python.exe
   machine: Windows-10-10.0.22000-SP0

Python dependencies:
          pip: 22.1.2
   setuptools: 63.1.0
      sklearn: 1.1.1
       sktime: 0.13.0
  statsmodels: 0.13.2
        numpy: 1.22.4
        scipy: 1.8.1
       pandas: 1.4.3
   matplotlib: 3.5.2
       joblib: 1.1.0
        numba: 0.55.2
     pmdarima: None
      tsfresh: None
```
</details>

achieveordie:
#### Reference Issues/PRs

Fixes #3169

- Modify `_load_provided_dataset()` under `datasets/_data_io.py` to inhibit failures from data-type conversions.
- Add `return_type` parameter to `load_UCR_UEA_dataset()` under `datasets/_single_problem_loaders.py` giving even more flexibility.

##### For all contributions


fkiraly:
@achieveordie, looks like there is one more problem:

the legacy data loader used some mtype strings that are not on the list of mtype, but were aliased, e.g., `"numpy3d"` would be interpreted as `"numpy3D"`.

I fear we need to keep doing this to ensure downwards compatibility, the best way is probably via an alias dictionary.

Aliases I know of:

* `"np3d"`, `"np3D"`, `"numpy3d"`, for `"numpy3D"`
* `"np2d"`, `"np2D"`, `"numpy2d"`, `"numpy2D"`,  for `"numpyflat"`

I wonder whether this is worth deprecating, or whether we should just "silently" accept these - ensuring downwards compatibility, but not mentioning in the docstring.

I would also change this in the notebooks that break, since we replaced by `mtype` strings. I.e., change the `"numpy3d"` in the notebooks to `"numpy3D"`,  and so on.

achieveordie:
Could you point me to the legacy data loaders that aren't making use of `_load_provided_dataset()`? We're checking for the mentioned aliases in `load_from_tsfile()` which is the only function that is invoked from `_load_provided_dataset()`, so as long as these loaders make use of the private loader functions, aliases can still be passed and corrected for, unless I'm incorrectly understanding the problem at hand.

fkiraly:
According to the failure in `run-notebook-examples` and its traceback, one of the loaders (not legacy but "green" and used in many examples) is `load_arrow_head` which is directly calling `_load_provided_dataset` as far as I can tell.

achieveordie:
Got it. Wouldn't it be better to use a private function to correct this behaviour instead of defining a (global?) alias dict; especially since the number of aliases are very small.

If you'd like lookup-based implementation, I'll revert to that.

I'll make this ready for review after the altered version of #3250 is accepted, I think everything else works fine.

fkiraly:
> Got it. Wouldn't it be better to use a private function to correct this behaviour instead of defining a (global?) alias dict; especially since the number of aliases is very few?

Yes, I would fix it in the private function - I was not suggesting a global alias dict.
I was simply suggesting to use a local alias dict.

achieveordie:
Ah, I misunderstood you, sorry :)

Hope everything is in its right place now; thank you so much for helping me out throughout the pull request.

fkiraly:
are the remaining failures going to be resolved once we merge the converter fix?

achieveordie:
Yes, I checked the logs, they are the same errors that will be fixed from that PR.

fkiraly:
failures are unrelated to this PR but are likely caused by today´s release of `pmdarima` 2.0.0

fkiraly (approved):
Indeed it passes now.

Thanks a lot for the contribution, *and* for your help with tracking down all those bugs!

PR #3241, opened 2022-08-10
Release v0.13.1 - changelog items missed by changelog generator

fkiraly:
Changelog updated with items that the changelog generator apparently missed.

There is a lot that it missed! FYI @lmmentel


fkiraly:
7 contributors and 40 PR were missed.

PR #3242, opened 2022-08-11
[ENH] Retry url request after HTTPError

khrapovs:
#### Reference Issues/PRs

Follow up on Slack conversation: "Looks like test_load_solar is causing CICD to fail. Does anyone experience something similar?" (2022-08-11 in Contributors channel). More precisely, here is the `HTTPError` causing the failure: https://github.com/alan-turing-institute/sktime/runs/7774941244?check_suite_focus=true#step:10:1676

Use [backoff](https://pypi.org/project/backoff/) library to retry data load from url in `sktime.datasets._single_problem_loaders.load_solar`.

[backoff](https://pypi.org/project/backoff/)

Does this failure happens so often that it is worth introducing a new dependency?

##### For all contributions


fkiraly (changes requested):
Good idea!

But, we can't introduce `backoff` as a core dependency, since it's used only in the tests.
It should be moved to the `dev` dependency set.

fkiraly:
assuming you are on holiday or otherwise off, @khrapovs, so I´m just moving this to the other dependency set myself (to avoid the load errors)

khrapovs:
I wanted to come back to it on Monday, but you are welcome to take it over if you think it becomes a pressing issue.

fkiraly:
> I wanted to come back to it on Monday, but you are welcome to take it over if you think it becomes a pressing issue.

ok, sorry - feel free to revert my changes.

I was trying to isolate `backoff` as a soft dependency, as moving it to the soft dependency set was breaking things here and there.

fkiraly:
merging since the error is popping up frequently on PR

PR #3244, opened 2022-08-11
[BUG] Fix grid searches NotFittedError if refit is False

### LINKED ISSUE #3234: [BUG] NotFittedError raised in BaseGridSearch when refit is False
**Describe the bug**
related discussion: #3220

grid search is trying to predict on `self.best_forecaster_ = self.forecaster.clone().set_params(**self.best_params_)` which was not fitted and raises therefore a `NotFittedError`. The problem is that` evaluate() `does not return the last fitted forecaster but it can optionally return the `y_train `for each forecaster, so we can take that data to fit the forecaster if not refit :)

**To Reproduce**

```python
from sktime.datasets import load_shampoo_sales
from sktime.forecasting.model_selection import (
    ExpandingWindowSplitter,
    ForecastingGridSearchCV,
    ExpandingWindowSplitter)
y = load_shampoo_sales()
fh = [1,2,3]
cv = ExpandingWindowSplitter(
    start_with_window=True,
    fh=fh)
forecaster = NaiveForecaster()
param_grid = {"strategy" : ["last", "mean", "drift"]}
    forecaster=forecaster,
    param_grid=param_grid,
    cv=cv,
    refit=False)
gscv.fit(y)
gscv.predict(fh=[1,2,3])
```

**Expected behavior**
If `refit is False`, then the last `y_train` from cross-validation should be used as fit data 

**Additional context**

**Versions**
`main`

aiwalter:
#### Reference Issues/PRs
closes #3234

### DIFF
--- a/sktime/forecasting/model_selection/_tune.py
+++ b/sktime/forecasting/model_selection/_tune.py
@@ -343,7 +343,8 @@ class ForecastingGridSearchCV(BaseGridSearch):
         -1 means using all processors.
     refit: bool, optional (default=True)
         True = refit the forecaster with the best parameters on the entire data in fit
-        False = best forecaster remains fitted on the last fold in cv
+        False = best forecaster remains fitted on the last fold in cv. Setting to False
+        makes only sense when using a SlidingWindowSplitter.
     verbose: int, optional (default=0)
     return_n_best_forecasters: int, default=1
         In case the n best forecaster should be returned, this value can be set
@@ -532,6 +533,7 @@ class ForecastingGridSearchCV(BaseGridSearch):
             "cv": SingleWindowSplitter(fh=1),
             "param_grid": {"window_length": [2, 5]},
             "scoring": MeanAbsolutePercentageError(symmetric=True),
+            "refit": [True, False],
         }
         params2 = {
             "forecaster": ExponentialSmoothing(),
@@ -594,7 +596,8 @@ class ForecastingRandomizedSearchCV(BaseGridSearch):
         -1 means using all processors.
     refit: bool, optional (default=True)
         True = refit the forecaster with the best parameters on the entire data in fit
-        False = best forecaster remains fitted on the last fold in cv
+        False = best forecaster remains fitted on the last fold in cv. Setting to False
+        makes only sense when using a SlidingWindowSplitter.
     verbose: int, optional (default=0)
     return_n_best_forecasters: int, default=1
         In case the n best forecaster should be returned, this value can be set
--- a/sktime/forecasting/model_selection/_tune.py
+++ b/sktime/forecasting/model_selection/_tune.py
@@ -158,7 +158,7 @@ class BaseGridSearch(_DelegatedForecaster):
                 axis=1,
             )
 
-            # store y_train and X_train for the case refit param is True
+            # store y_train and X_train for the case refit param is False
             row = out[out["cutoff"] == out["cutoff"].max()].iloc[0]
             self._y_evaluate = row["y_train"]
             self._X_evaluate = row["X_train"]
--- a/sktime/forecasting/model_selection/_tune.py
+++ b/sktime/forecasting/model_selection/_tune.py
@@ -141,10 +141,30 @@ class BaseGridSearch(_DelegatedForecaster):
                 strategy=self.strategy,
                 scoring=scoring,
                 error_score=self.error_score,
+                # get cv data in order to use it for fitting if self.refit is True
+                return_data=True,
             )
 
             # Filter columns.
-            out = out.filter(items=[scoring_name, "fit_time", "pred_time"], axis=1)
+            out = out.filter(
+                items=[
+                    scoring_name,
+                    "fit_time",
+                    "pred_time",
+                    "y_train",
+                    "X_train",
+                    "cutoff",
+                ],
+                axis=1,
+            )
+
+            # store y_train and X_train for the case refit param is True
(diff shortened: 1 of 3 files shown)


aiwalter:
@acartro I have a PR, now there is just one question. When `refit is False`, the last window will be used to fit the best forecaster instead of all data. This comes with a problem bcz the `cutoff` (last training time point) will not be same as with all data bcz in my example above with an `fh=[1,2,3]` in cv, the cutoff of cv data will 3 steps in the past. What would be your expected behavior as a user in this case? Should we then at least update the best forecaster with this 3 data points in order to move the cutoff to be same with all data?


any thoughts @fkiraly ?

aiwalter:
probably the "clearest" solution would be to move the last window 3 points in the future in above example with `fh=[12,3]`. With `update` the window size would increase by 3 data points, so probably better to just move it to be in sync with full training data.

fkiraly:
> any thoughts @fkiraly ?

My suggestion would be: run `update`, but with `update_params=False`. That way, the model parameters do not get updated, but the "present state" is set to the latest time stamp.

aiwalter:
Did it now with `update`, pls review

sktime/forecasting/model_selection/_tune.py
```diff
@@ -141,10 +141,30 @@ def _fit_and_score(params):
...
+                    "fit_time",
+                    "pred_time",
+                    "y_train",
+                    "X_train",
+                    "cutoff",
+                ],
+                axis=1,
+            )
+
+            # store y_train and X_train for the case refit param is True
```
aiwalter, line 161:
```suggestion
            # store y_train and X_train for the case refit param is False
```

fkiraly:
what happened to the cancelled CI runs? Restarting

aiwalter:
Did it now with `update`, pls review

sktime/forecasting/model_selection/_tune.py
```diff
@@ -311,8 +356,10 @@ class ForecastingGridSearchCV(BaseGridSearch):
         None means 1 unless in a joblib.parallel_backend context.
         -1 means using all processors.
     refit: bool, optional (default=True)
-        True = refit the forecaster with the best parameters on the entire data in fit
-        False = best forecaster remains fitted on the last fold in cv
+        True = refit the forecaster with the best parameters on the entire data in fit.
+        False = best forecaster remains fitted on the last fold in cv but will b
+        updated to move cutoff to present by means of update(update_params=False).
```
miraep8, line 361: I think you should make it a little clearer what you mean regarding  "move cutoff to present".  I think it was clear what you intended from discussion on the PR, but think its a little more confusing here without that context.

sktime/forecasting/model_selection/tests/test_tune.py
```diff
@@ -131,3 +132,29 @@ def test_rscv(forecaster, param_grid, cv, scoring, error_score, n_iter, random_s
...
+    y_pred_refit = gscv_refit.predict(fh=[1, 2, 3])
+
+    # no refit
+    gscv_no_refit = ForecastingGridSearchCV(
+        forecaster=forecaster, param_grid=param_grid, cv=cv, refit=False
+    )
+    gscv_no_refit.fit(y)
+    y_pred_no_refit = gscv_no_refit.predict(fh=[1, 2, 3])
+
+    # predicts have to be different because of different training data
```
miraep8, line 158: Nice test 👍

sktime/forecasting/model_selection/tests/test_tune.py
```diff
@@ -131,3 +132,29 @@ def test_rscv(forecaster, param_grid, cv, scoring, error_score, n_iter, random_s
         ParameterSampler(param_grid, n_iter, random_state=random_state)
     )
     _check_cv(forecaster, rscv, cv, param_distributions, y, X, scoring)
+
+
+def test_refit():
+    """Test refit param."""
```
miraep8, line 138: Could be nice to add another mini test within this to show that `x_train` and `y_train` are the sizes you expect them to be. (and also that cutoff behaves as expected).

miraep8 (commented):
Overall it seems like a sensible change/doesn't have a high likelihood of breaking things. Small non-blocking suggestions:
- I think it would be nice to add more clarity as to what is going on with `cutoff`. Either in docstring or inline with that code. 
- can add one or two more consistency checks that cutoff and x_train/y_train get updated as expected in the testing file.

fkiraly (changes requested):
Yes, I agree with @miraep8 about the `cutoff`.

I would even make a stronger statement: the `cutoff` of the wrapped estimator, at the end, should be the `cutoff` of all data seen in `fit` by the tuner, not the last time stamp seen in the last test fold.

Therefore, I would do sth like run an `update` with `update_params=False` or sth to set the `cutoff` to the latest time stamp seen.

Am I making sense, or is this a bad idea for some reason?

fkiraly:
@aiwalter, what's the plan? Do you disagree with the `cutoff` handling?
I have no strong opinion, and doing sth is better than an exception.

But, I think it can cause trouble down the line if the `cutoff` is inconsistent between the two settings.

aiwalter:
Will come back to this PR next week 😊

fkiraly:
should we merge and release with 0.13.2, or let it wait?
Pro: replaces exception with sth reasonable
Con: if we merge now, change may require deprecation later

fkiraly:
still working on this, @aiwalter, or should we merge?

fkiraly:
@aiwalter , ping

aiwalter:
closing due to #4153

PR #2934, opened 2022-07-06
[ENH] Elbow class selection/pairwise dimension selection for multivariate time series classification

TonyBagnall:
#### Reference Issues/PRs
This PR subsumes this one
https://github.com/alan-turing-institute/sktime/pull/1821
it is a joint effort between @a-pasos-ruiz and @haskarb. Draft PR, will add more details shortly. 

implementation of three dimension/channel selection algorithm sdescribed here
https://link.springer.com/chapter/10.1007/978-3-030-91445-5_3
algorithms named ECP/ECS and kmeans in the paper. These have been relocated to a file dimension_selection in transformations/panel, and will have classes ElbowChannelSelection, ElbowChannelPairwise and ClusterBasedChannelSelection.

##### For all contributions
- [x] I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/main/.all-contributorsrc).
- [ ] Optionally, I've updated sktime's [CODEOWNERS](https://github.com/alan-turing-institute/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
- [ ] I've added unit tests and made sure they pass locally.
- [x ] The PR title starts with either [ENH], [MNT], [DOC], or [BUG] indicating whether the PR topic is related to enhancement, maintenance, documentation, or bug.

##### For new estimators
- [ ] I've added the estimator to the online documentation.
- [ ] I've updated the existing example notebooks or provided a new one to showcase how my estimator works.

PR #3248, opened 2022-08-12
[ENH] Channel selection for multivariate time series classification.

haskarb:
#### Reference Issues/PRs
#2934
Partially related to mentioned PR. The methods are implemented from [this](https://link.springer.com/chapter/10.1007/978-3-030-91445-5_3) paper. 

##### For all contributions

##### For new estimators

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1747,7 +1747,17 @@
       "contributions": [
         "bug"
       ]
-    }
+    },
+    {
+      "login": "haskarb",
+      "name": "Bhaskar Dhariyal",
+      "avatar_url": "https://avatars.githubusercontent.com/u/20501023?v=4",
+      "profile": "https://haskarb.github.io/",
+      "contributions": [
+        "code",
+        "test"
+      ]
+    },
   ],
   "projectName": "sktime",
   "projectOwner": "alan-turing-institute",
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -36,6 +36,7 @@ sktime/transformations/series/kalman_filter.py @NoaBenAmi
 sktime/transformations/panel/augmenter.py @MrPr3ntice @iljamaurer
 sktime/transformations/multiplex.py @miraep8
 sktime/transformations/tests/test_multiplexer.py @miraep8
+sktime/transformations/panel/channel_selection.py @haskarb @a-pasos-ruiz @TonyBagnall
 
 sktime/forecasting/base/ @fkiraly @mloning @aiwalter
 sktime/forecasting/base/adapters/_statsforecast.py @FedericoGarza
--- a/examples/channel_selection.ipynb
+++ b/examples/channel_selection.ipynb
@@ -1,0 +1,513 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 8,
+   "id": "89a363cd-8944-475a-88ea-4401785218c5",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import warnings\n",
+    "\n",
+    "warnings.filterwarnings(\"ignore\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "b45cbeda-65bc-4b10-9f9c-9afd43a54d20",
+   "metadata": {},
+   "source": [
+    "# Channel Selection in Multivariate Time Series Classification \n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "08decb5b-8dfb-4666-a66b-3a2349960956",
+   "metadata": {},
+   "source": [
+    "## Overview"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "da743484-17d3-4cec-8eaa-8d8293ce6f35",
+   "metadata": {},
+   "source": [
+    "Sometimes every channel is not required to perform classification; only a few are useful.  The [1] proposed a fast channel selection technique for Multivariate Time Classification. "
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "dcbe2174-a691-4093-ab80-d796edb5121d",
+   "metadata": {},
+   "source": [
+    "[1] : Fast Channel Selection for Scalable Multivariate Time Series Classification [Link](https://www.researchgate.net/publication/354445008_Fast_Channel_Selection_for_Scalable_Multivariate_Time_Series_Classification)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 9,
+   "id": "d1779970-eefb-4577-9c4e-e0a19ceadcc1",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from sklearn.linear_model import RidgeClassifierCV\n",
+    "from sklearn.pipeline import make_pipeline\n",
+    "\n",
+    "from sktime.datasets import load_UCR_UEA_dataset\n",
+    "from sktime.transformations.panel import channel_selection\n",
+    "from sktime.transformations.panel.rocket import Rocket"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "0437ca7a-5b5a-4e28-b565-0b2df4eac60d",
+   "metadata": {},
+   "source": [
+    "# 1 Initialise the Pipeline"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 10,
+   "id": "830137a3-10c3-49b9-9a98-7062dc7ab1d8",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "cs = channel_selection.ElbowChannelSelection()  # ECS\n",
+    "# cs = channel_selection.ElbowClassPairwise() # ECP"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 11,
+   "id": "89443793-7cf0-4a4c-a4b1-d928a46a1bb2",
+   "metadata": {},
+   "outputs": [],
+   "source": [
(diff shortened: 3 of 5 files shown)


fkiraly (changes requested):
Nice, thanks!

One thing I would like to suggest:
you are duplicating distance calculations that are available in `sktime` already as abstract objects, and it would also be nice to allow generic distances to be used.

The euclidean distance that you are currently using, I think, can be obtained as follows

```python
from sktime.dists_kernels import AggrDist, ScipyDist

my_dist = AggrDist(ScipyDist())

# to compute distance, do this, instead of _distance_matrix.distance
distance_frame = my_dist(centroid_frame)
```

Instead of hard-coding this, I would suggest to allow the user to pass an `__init__` argument `distance`, with a default of `None`, and if `None`, you just use `AggrDist(ScipyDist())`.

fkiraly:
Hm, it might actually be `FlatDist` (from this PR https://github.com/alan-turing-institute/sktime/pull/3249) instead of `AggrDist`, i.e., `FlatDist(ScipyDist())`

fkiraly:
Also, your doctests are failing. The code in there must run with `sktime` installed, and it does not, e.g., there is a type where you confused `sktime` and `sklearn`.

To test whether the docstring is running, easy way is to copy paste it into a visual studio code notebook and run it. You can copy paste with `>>>` and `...` and it will still run.

haskarb:
Thanks for leaving comments.

So in the paper, we have used euclidean to save computation time, so we have fixed euclidean distance. We also tried to use 'dtw' in another [paper](https://arxiv.org/pdf/2206.09274.pdf); it is under review. I will update/contribute the code once the review process is over.

Also, I can't find FlatDist [here](https://github.com/alan-turing-institute/sktime/blob/main/sktime/dists_kernels/compose_tab_to_panel.py)

fkiraly:
> In the paper, we have used euclidean to save computation time, so we have fixed euclidean distance. 

Well, just leave that as the default then, and we keep the distance modular.
If you don't want to do that, I can do that in a separate PR later, if you don't mind?

fkiraly:
> Also, I can't find FlatDist [here](https://github.com/alan-turing-institute/sktime/blob/main/sktime/dists_kernels/compose_tab_to_panel.py)

Still in a PR, so to use it , you would have to merge it in.

Question: is it euclidean distance on flattened series, or is it average of euclidean distances between time pts?

haskarb:
> Or, if you don't want to do that, I can do that in a separate PR later, if you don't mind?

Yes, I'm okay with that.

> Question: is it euclidean distance on flattened series, or is it average of euclidean distances between time pts?

It is euclidean distance between two vectors (or two univariate time-series)

fkiraly:
> Also, I see some of the tests failing, I'm curious what's going wrong?

Do you know how to see the failing tests, and how to get the traceback? Go on "details" next to the test.

haskarb:
Ah, the channel selection was not defined. Thanks, think I found the problem.

fkiraly:
still failing - try to copy-paste the example block in a visual studio code and run it, to see what fails

fkiraly:
(if unclear what I'm saying, ping me on slack and I can show you when I'm online - or ask in the contributors channel)

haskarb:
Sure, thanks. I will try to work on this tomorrow.

fkiraly:
Fixed it for you because it's non-obvious.

If you have an expression going over multiple lines in a docstring, the second line needs to start with triple dots and a space (see what I fixed).

fkiraly (changes requested):
Looks good now.

A few remaining issues:
* please ensure to add `Parameters` section to the docstring, that describe all `__init__` parameters. See other classes for eamples.
* `self.paramname` should never be overwritten by a value different from that passed to `__init__`.  You hav a problem with `random_state` in `ElbowClassPairwise`. Instead, write the coerced variable to another attribute, e.g., `_random_state`
* you should not write to `_is_fitted`, the public part of the interface already deals with that.
* `channels_selected` should end in an underscore, since it is a parameter that is fitted - it should be `channels_selected_`.

Recommendations of improvement (not blocking):
* users will find it inconvient if `channels_selected_` is a list of integers, even if the data format is `pandas` based - in that case, they would expect column names rather than integers.
* you could enable the above by using the `pd-multiindex` or `nested_univ` format, internally. We can look at this in another PR.

fkiraly:
Re docstring: I fixed only the first one, you should be able to do the 2nd by following the pattern.

haskarb:
Thanks for the recommendations.

> * please ensure to add `Parameters` section to the docstring, that describe all `__init__` parameters. See other classes for samples.
Fixed.
> * `self.paramname` should never be overwritten by a value different from that passed to `__init__`.  You hav a problem with `random_state` in `ElbowClassPairwise`. Instead, write the coerced variable to another attribute, e.g., `_random_state`
I removed random_state variable, not required in the code.
> * you should not write to `_is_fitted`, the public part of the interface already deals with that.
Fixed.
> * `channels_selected` should end in an underscore, since it is a parameter that is fitted - it should be `channels_selected_`.
Fixed.
> * `distance_frame_` descriptions seems to be inaccurate, that's not a distance matrix? Entries are grand sum of distance matrix for that variable.
Every cell value in the dataframe, represents distance between class-centroid pair (column name) and channel(row).
> 
> Recommendations of improvement (not blocking):
> 
> * users will find it inconvient if `channels_selected_` is a list of integers, even if the data format is `pandas` based - in that case, they would expect column names rather than integers.
> * you could enable the above by using the `pd-multiindex` or `nested_univ` format, internally. We can look at this in another PR.

fkiraly:
another thing: if you have a line like `my_clf.fit(sthsth)`, the return is `self`, and doctest checks whether what comes after this is the expected printout, e.g., `ClassName(param1=etc)`. Since this is not reliably printed across OS and versions, you can capture the params with ellipsis, e.g., `ClassName(...)`.

fkiraly (changes requested):
Thanks for fixing!
Tests pass now.

Issues that are still unresolved:

> please ensure to add Parameters section to the docstring, that describe all __init__ parameters. See other classes for samples.

It seems you removed all parameters, incl `random_state`. Was that intended?

> you should not write to _is_fitted, the public part of the interface already deals with that.

You still write to `_is_fitted`.

> `distance_frame_` descriptions seems to be inaccurate, that's not a distance matrix? Entries are grand sum of distance matrix for that variable.

Please add a proper description in the docstring what this is.

haskarb:
> It seems you removed all parameters, incl `random_state`. Was that intended?
Yes, no randomisation is involved in our methods. This code was built over Tony's code, they might have used for their work.
 
> > you should not write to _is_fitted, the public part of the interface already deals with that.
> 
> You still write to `_is_fitted`.
I will fix this.
> > `distance_frame_` descriptions seems to be inaccurate, that's not a distance matrix? Entries are grand sum of distance matrix for that variable.
> 
> Please add a proper description in the docstring what this is.

Will update shortly.

fkiraly (changes requested):
Code looks good to me now!

I think the class docstring needs improvement, to clarify two things:

* channel selection is the same as variable selection, i.e., this will column subset a data frame, or row subset a numpy 3D array. Channel selection is not a too common term.
* explanation of the `distance_frame_´ is still a bit nebulous. Optimally, please refer to the paper (e.g., mathematical objects therein), or add some formalism in the docstring, say using the `:math:` environment. Explain what the rows/columns of the matrix are, and what the (i, j)-th entry is.

haskarb:
> * channel selection is the same as variable selection, i.e., this will column subset a data frame, or row subset a numpy 3D array. Channel selection is not a too common term.

Yes, it is a column subset, not the rows. Every column is a channel. 

> * explanation of the `distance_frame_´ is still a bit nebulous. Optimally, please refer to the paper (e.g., mathematical objects therein), or add some formalism in the docstring, say using the `:math:` environment. Explain what the rows/columns of the matrix are, and what the (i, j)-th entry is.

I modified it a bit, however, if people are still confused they can refer to the linked paper, where we have explained it with the help of a diagram.

fkiraly:
>Yes, it is a column subset, not the rows. Every column is a channel.

I know that! But the typical user might not (the result of which would be ignoring your estimator!).
Therefore, I´m asking for an explanation in the class docstring.

> I modified it a bit, however, if people are still confused they can refer to the linked paper, where we have explained it with the help of a diagram.

Sorry, I think it is not enough yet, but we are almost there.
Could you:

* for all the attributes, state the type clearly - e.g., `list of int, integers being column index of selected variable/channel`, not just `list`, etc
* for the `distance_matrix_` explanation, please ensure that the user can easily find the explanation of what "distance between centroids for channel" means, this is not a clear formal definition. One way is to define in the docstring, the other way is to link directly to an object in your paper (e.g., "the matrix denoted by sthsth in algorithm sthsth" or similar)

haskarb:
> I know that! But the typical user might not (the result of which would be ignoring your estimator!). Therefore, I´m asking for an explanation in the class docstring.
> 
> If you explain that this is another term for variable selection, and channel = variable, I am convinced you´ll get more users!

Thanks, I made a remark in class docstring,

> > I modified it a bit, however, if people are still confused they can refer to the linked paper, where we have explained it with the help of a diagram.
> 
> Sorry, I think it is not enough yet, but we are almost there. Could you:
> 
> * for all the attributes, state the type clearly - e.g., `list of int, integers being column index of selected variable/channel`, not just `list`, etc
> * for the `distance_matrix_` explanation, please ensure that the user can easily find the explanation of what "distance between centroids for channel" means, this is not a clear formal definition. One way is to define in the docstring, the other way is to link directly to an object in your paper (e.g., "the matrix denoted by sthsth in algorithm sthsth" or similar)

I gave the reference to the table from paper, the table illustrates it in a much nicer way.

fkiraly (approved):
Ok, fine.

haskarb:
Thanks @fkiraly, for the help!

PR #3250, opened 2022-08-12
[ENH] treat non nested cols in conversion `nested_univ` to `pd-multiindex`

fkiraly:
This PR extends the converter `nested_univ` to `pd-multiindex` to allow it to deal with non-nested columns.

See discussion in https://github.com/alan-turing-institute/sktime/issues/3169

Bit clunky (double loop), but I currently have no better idea, and there is no compute overhead if all the columns are nested.

### DIFF
--- a/sktime/datatypes/_panel/_convert.py
+++ b/sktime/datatypes/_panel/_convert.py
@@ -770,14 +770,17 @@ def from_nested_to_multi_index(X, instance_index=None, time_index=None):
     -------
     X_mi : pd.DataFrame
         The multi-indexed pandas DataFrame
-
     """
     # this contains the right values, but does not have the right index
     #   need convert_dtypes or dtypes will always be object
     # explode by column to ensure we deal with unequal length series properly
     X_mi = pd.DataFrame()
 
-    for c in X.columns:
+    X_cols = X.columns
+    nested_cols = [c for c in X_cols if isinstance(X[[c]].iloc[0, 0], pd.Series)]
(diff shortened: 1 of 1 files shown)


achieveordie:
This works perfectly to ensure that we don't have `NaN` for unnested columns. I have a suggestion though, the resulting dtype will still be an object in the case of `class_val`; Causing checks to fail with `obj should not have a column of 'object' dtype` from the checker function `check_pddataframe_series()`.

Explicitly changing dtype to `int` corrects this behaviour. But since we don't know the type beforehand let pandas handle it with `.convert_dtypes()`.

```python
# X_mi[[c]] = X_mi[[c]].astype(X[[c]].dtypes[0])
X_mi[[c]] = X_mi[[c]].convert_dtypes()
```
passes all the tests from `test_data_io.py` locally.

fkiraly:
Hm, why does `astype` not fix it? Isn't `class_val` a numeric?

achieveordie:
My reference was `UnitTest` data, where `class_val` was being loaded as an object type. We could use `astype()` by explicitly passing int but the reason I suggested using `convert_dtypes()` was because the un-nested column might not always be of integer type. 

If you believe something is off, I can check again.

fkiraly:
> If you believe something is off, I can check again; I could also edit my previous comment with the code to show the problem if required.

Not needed, I just wanted to understand the difference condition.
Have changed it to `convert_dtypes`.

PR #3255, opened 2022-08-13
[ENH] support for `xarray` `DataArray` & mtypes

### LINKED ISSUE #2655: [ENH] sktime-pywatts integration - `xarray` support
Part of [pywatts integration design questions](https://github.com/alan-turing-institute/sktime/issues/2653).

`pywatts` uses `xarray` as central data container, while `sktime` does not have `xarray` as one of its supported input data types.

Adding it to the `datatypes` module would be very easy, but a few design questions need to be answered:

* should `xarray` become a core dependeny, or only a soft dependency?
* if only a soft dependency, how do we handle data types that are soft dependencies? We probably would have to insert some machinery in converters etc?
* should we introduce `xarray` for `Series` (1D, 2D), `Panel` (3D)  and `Hierarchical` (4D plus), or only some of these?

benHeid:
#### Reference Issues/PRs

resolves #2655 

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/datatypes/_series/_check.py
+++ b/sktime/datatypes/_series/_check.py
@@ -263,7 +263,7 @@ if _check_soft_dependencies("xarray", severity="none"):
 
         metadata["is_empty"] = len(index) < 1 or len(obj.values) < 1
         # The second dimension is the set of columns
-        metadata["is_univariate"] = len(obj[obj.dims[1]]) < 2
+        metadata["is_univariate"] = len(obj.dims) == 1 or len(obj[obj.dims[1]]) < 2
 
         # check that columns are unique
         msg = f"{var_name} must have " f"unique column indices, but found {obj.dims}"
--- a/sktime/datatypes/_series/_check.py
+++ b/sktime/datatypes/_series/_check.py
@@ -254,7 +254,7 @@ if _check_soft_dependencies("xarray", severity="none"):
             return ret(False, msg, None, return_metadata)
 
         # we now know obj is a xr.DataArray
-        if len(obj.dims) > 2: # Without multi indexing only two dimensions are possible
+        if len(obj.dims) > 2:  # Without multi indexing only two dimensions are possible
             msg = f"{var_name} must have two or less dimension, found {type(obj.dims)}"
             return ret(False, msg, None, return_metadata)
 
@@ -285,8 +285,8 @@ if _check_soft_dependencies("xarray", severity="none"):
         # Check time index is ordered in time
         if not index.is_monotonic:
             msg = (
-                f"The (time) index of {var_name} must be sorted monotonically increasing, "
-                f"but found: {index}"
+                f"The (time) index of {var_name} must be sorted "
+                f"monotonically increasing, but found: {index}"
             )
             return ret(False, msg, None, return_metadata)
 
--- a/sktime/datatypes/_series/_check.py
+++ b/sktime/datatypes/_series/_check.py
@@ -242,16 +242,12 @@ if _check_soft_dependencies("xarray", severity="none"):
     import xarray as xr
 
     def check_xrdataarray_series(obj, return_metadata=False, var_name="obj"):
-        metadata = dict()
-        # TODO multiindexing is not allowed!
-        # TODO check number of time index
-        # TODO check order of indexes (First time, second column names?)
+        metadata = {}
 
         def ret(valid, msg, metadata, return_metadata):
             if return_metadata:
                 return valid, msg, metadata
-            else:
-                return valid
+            return valid
 
         if not isinstance(obj, xr.DataArray):
             msg = f"{var_name} must be a xarray.DataArray, found {type(obj)}"
@@ -261,14 +257,17 @@ if _check_soft_dependencies("xarray", severity="none"):
         if len(obj.dims) > 2: # Without multi indexing only two dimensions are possible
             msg = f"{var_name} must have two or less dimension, found {type(obj.dims)}"
             return ret(False, msg, None, return_metadata)
-        index = obj.indexes[obj.dims[0]] # TODO check the relation between coords and dims
+
+        # The first dimension is the index of the time series in sktimelen
+        index = obj.indexes[obj.dims[0]]
 
         metadata["is_empty"] = len(index) < 1 or len(obj.values) < 1
-        metadata["is_univariate"] = len(obj[obj.dims[1]]) < 2 # TODO requires that the second dimension describes the columns.
+        # The second dimension is the set of columns
+        metadata["is_univariate"] = len(obj[obj.dims[1]]) < 2
 
         # check that columns are unique
         msg = f"{var_name} must have " f"unique column indices, but found {obj.dims}"
-        assert len(obj.dims) == len(set(obj.dims)), msg # TODO Check column names
+        assert len(obj.dims) == len(set(obj.dims)), msg
 
         # check whether the time index is of valid type
         if not is_in_valid_index_types(index):
@@ -284,7 +283,7 @@ if _check_soft_dependencies("xarray", severity="none"):
             return ret(False, msg, None, return_metadata)
 
         # Check time index is ordered in time
-        if not index.is_monotonic: # TODO Check if index is monotonic
+        if not index.is_monotonic:
             msg = (
                 f"The (time) index of {var_name} must be sorted monotonically increasing, "
                 f"but found: {index}"
@@ -305,9 +304,3 @@ if _check_soft_dependencies("xarray", severity="none"):
         return ret(True, None, metadata, return_metadata)
 
     check_dict[("xr.DataArray", "Series")] = check_xrdataarray_series
-
-
-    #def check_xrdataset_series(obj, return_metadata=False, var_name="obj"):
-    #    pass
-
-    #check_dict[("xr.DataSet", "Series")] = check_xrdataset_series
--- a/sktime/datatypes/_series/_check.py
+++ b/sktime/datatypes/_series/_check.py
@@ -40,6 +40,7 @@ __all__ = ["check_dict"]
 import numpy as np
 import pandas as pd
 
+from sktime.utils.validation._dependencies import _check_soft_dependencies
 from sktime.utils.validation.series import is_in_valid_index_types
 
 VALID_INDEX_TYPES = (pd.RangeIndex, pd.PeriodIndex, pd.DatetimeIndex)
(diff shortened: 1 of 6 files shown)


benHeid:
#TODOs 
* [ ] make xarray to a softdependency (conditional import)
* [ ] resolve TODOs in the code
* [ ] Discuss if 
   * DataSets should also be supported
   * Further scitypes should support xarray

fkiraly:
for conditional execution (based on presence of package in the env), you can use:

```python
if _check_soft_dependencies("xarray", severity="none"):
    import xarray
    etc
```

fkiraly:
let me know if you think you are stuck - I can look at this is more detail, we can do pair programming, or I can give it a try.

benHeid:
> let me know if you think you are stuck - I can look at this is more detail, we can do pair programming, or I can give it a try.

Thank you, I let you know if I am stuck

benHeid:
I have two questions:
* I added the `extend_conversion` function from tables to series. (In panel, there exists something similar). Should I move this method to another file in the datatypes file, or should an extra PR do this? 
* Should I add xarray as a soft dependency somewhere that the automatic tests can install xarray?

fkiraly:
> I added the `extend_conversion` function from tables to series. (In panel, there exists something similar). Should I move this method to another file in the datatypes file, or should an extra PR do this? 

Depends what you need it for. If needed in this PR, we can leave it in its location and refactor.
If you want it as a more general utility and not needed for the `xarray` topic, it is better putting in a separate PR.

> Should I add xarray as a soft dependency somewhere that the automatic tests can install xarray?

Yes, `pyproject.toml`, `all_extras` dependency set.

benHeid:
> > I added the `extend_conversion` function from tables to series. (In panel, there exists something similar). Should I move this method to another file in the datatypes file, or should an extra PR do this?
> 
> Depends what you need it for. If needed in this PR, we can leave it in its location and refactor. If you want it as a more general utility and not needed for the `xarray` topic, it is better putting in a separate PR.
> 
Ok, it is not needed for the xarray topic.

fkiraly:
> A solution for this will probably also solve also the problem with the tests in python 3.7 (it seems that in python 3.7 the index is a BaseIndex after the conversion, while for 3.8 and 3.9 it is a PeriodIndex)

Ouch... how odd. I suppose this is one of those non-obvious problems that occur...

fkiraly:
👏 nice! Looks like it passes now!

sktime/utils/_testing/deep_equals.py
```diff
@@ -108,6 +119,9 @@ def ret(is_equal, msg):
             return ret(False, f".dtype, x.dtype = {x.dtype} != y.dtype = {y.dtype}")
         return ret(np.array_equal(x, y, equal_nan=True), ".values")
     # recursion through lists, tuples and dicts
+    elif check_xarray(x, y)[0]:
```
benHeid, line 122: Do you have an idea how to solve this in a more elegant way? 

I need to wrap this check behind a method for checking if xarray is installed.

sktime/datatypes/_series/_check.py
```diff
@@ -235,3 +236,71 @@ def _index_equally_spaced(index):
...
+            msg = f"{var_name} must be a xarray.DataArray, found {type(obj)}"
+            return ret(False, msg, None, return_metadata)
+
+        # we now know obj is a xr.DataArray
+        if len(obj.dims) > 2:  # Without multi indexing only two dimensions are possible
+            msg = f"{var_name} must have two or less dimension, found {type(obj.dims)}"
+            return ret(False, msg, None, return_metadata)
+
+        # The first dimension is the index of the time series in sktimelen
+        index = obj.indexes[obj.dims[0]]
```
benHeid, line 262: Where should I document the required format of xr.DataArray? Currently, the xr.DataArray must have two dimensions. The first dimension is the index, and the second dimension describes the columns.

sktime/datatypes/_series/_convert.py
```diff
@@ -172,3 +174,67 @@ def convert_np_to_UvS_as_Series(obj: np.ndarray, store=None) -> pd.Series:
...
+def _concat(fun1, fun2):
+    def concat_fun(obj, store=None):
+        obj1 = fun1(obj, store=store)
+        obj2 = fun2(obj1, store=store)
+        return obj2
+
+    return concat_fun
+
+
+def _extend_conversions(mtype, anchor_mtype, convert_dict):
```
benHeid, line 189: If datatypes is refactored, then this method should be placed somewhere else. Currently similar methods are implemented in panels and tables.

sktime/datatypes/_series/_convert.py
```diff
@@ -172,3 +174,67 @@ def convert_np_to_UvS_as_Series(obj: np.ndarray, store=None) -> pd.Series:
...
+        ("xr.DataArray", "pd.DataFrame", "Series")
+    ] = convert_xrdataarray_to_Mvs_as_Series
+
+    def convert_Mvs_to_xrdatarray_as_Series(
+        obj: pd.DataFrame, store=None
+    ) -> xr.DataArray:
+        if not isinstance(obj, pd.DataFrame):
+            raise TypeError("input must be a xr.DataArray")
+
+        result = xr.DataArray(obj.values, coords=[obj.index, obj.columns])
```
benHeid, line 229: Note, I don't use the pandas.to_xarray here since this has some strange effects on the dimension which are reversed in that case. The workaround of using transpose would change the naming of the index and the row dim. Thus, creating an xr.DataArray explicitly here is more transparent.

benHeid:
> 👏 nice! Looks like it passes now!

Yeah 🥳  
I think you can now take a more detailed look at this PR. I added already some comments, and I would appreciate your feedback there.

Should we add the support for other mtypes and the dataset in this PR or should I create additional PRs for the other?

fkiraly:
> Should we add the support for other mtypes and the dataset in this PR or should I create additional PRs for the other?

I would go with the "minimal modular change" principle and separate PR, also finalizing this one first.
Which other mtypes do you have in mind?

sktime/utils/_testing/deep_equals.py
fkiraly: * the last condition defaults to `__eq__`, which is called on the object. Does that not suffice, or does that check identity in reference rather than just in value?
* otherwise no good idea, but I would really like the `deep_equals` module decoupled from any soft dependencies (not just conditionally decoupled but absolutely)

I assume this is because tests are using this, and it breaks without the modification?

"best practice" would be using a dispatch pattern, but that would be a major refactor.
Could do "cheap" dictionary dispatch, on type, like in datatypes, or "full blown" dispatch.

sktime/datatypes/_series/_convert.py
fkiraly: odd, but as long as it works?

sktime/datatypes/_series/_convert.py
fkiraly: yes, indeed - too much copy-paste!
Though I would go with incurring technical debt due to priorities...

Re refactor directions:

@chrisholder started sth, but abandoned it.
https://github.com/alan-turing-institute/sktime/pull/1460
Haven´t had the time to salvage it.

This would be another option:
https://github.com/dylan-profiler/visions/issues/194

sktime/datatypes/_series/_check.py
fkiraly: currently I've collected specifications in `examples/AA_datatypes`, although this is a bit ad-hoc, and long-term we may at some point like to split the "exhaustive list" (specification reference) from a "tutorial notebook". For now, I'd add it there, or perhaps a second notebook which has the less commonly used mtypes (too many mtypes would dilute the tutorial style experience)?

fkiraly (changes requested):
Really nice, thanks!

if this were not a draft, I would have only one blocking change, that's the coupling introduced in `deep_equals`, can we avoid it?

sktime/utils/_testing/deep_equals.py
benHeid: * `__eq__` does element wise comparison as `np.arrays`. Consequently, the elif fails since the result is ambiguous. An alternative would be wrapping the `__eq__` with `np.all` or `np.any`. However, this only checks if the values are the same and do not perform checks on the index or the columns, which I think is expected when a deep_equals is called.

Yes, the tests break without this modification.

benHeid: I tested it, we can remove this check_xarray and replacing the last elif statement with `np.any(x != y)`. It would enable to check the equality of the values of xr.DataArrays and if `x!=y` results into a bool, it still works since `np.any(True)` is `True` and `np.any(False)` is `False`.

benHeid:
# Update
* I changed the check in `deep_equals`.
* I manually create a dataframe instead of using to_pandas. This ensures that xr.DataArrays are always transformed to DataFrames and not to pd.Series if the xr.DataArray contains only one dimension. Note, `to_dataframe` does not work well, since it does some multi-indexing  stuff in case of two dimension which needs to be unstacked.

sktime/utils/_testing/deep_equals.py
```diff
@@ -118,7 +118,7 @@ def ret(is_equal, msg):
         )
     elif type(x).__name__ == "ForecastingHorizon":
         return ret(*_fh_equals(x, y, return_msg=True))
-    elif x != y:
+    elif np.any(x != y):
```
fkiraly, line 121: this is going to break things!

I would do instead:
1. compute `x != y`
2. check if boolean. if yes, return that
3. else, return `np.any` . (perhaps check first if that makes sense, and if not, you could raise an exception)

benHeid: I changed it as you proposed.

Checking if `np.any` makes sense difficult since `np.any` allows anything as input which is array_like. This does not mean that it will work with all array_like objects. E.g., `np.any(np.array("string"))` will fail with `TypeError: cannot perform reduce with flexible type`. Honestly, I do not know how to test it. 
However, as long as `__ne__` returns bool or a list of bools it should work.

fkiraly: ok, this works, I suppose.

fkiraly (approved):
Works!

Do you want me to merge this, or wait for panel?

benHeid:
> Works!
> 
> Do you want me to merge this, or wait for panel?

I would merge it and add the panel with a second PR if I need this. Currently, I can test and go on with the sktime-pyWATTS integration without implementing Panels. Perhaps Issue #2655 should stay open in that case.

benHeid:
@fkiraly Do we have to adapt utils.validation.series? In that file, something like VALID_DATA_TYPES is defined. Do I have to add xr.DataArray?

fkiraly:
> @fkiraly Do we have to adapt utils.validation.series? In that file, something like VALID_DATA_TYPES is defined. Do I have to add xr.DataArray? 

I think that's used only in legacy code, so I would recommend to not touch it (who knows what will break).

The "critical touch point" is `check_series`, which is still linked to a few estimators that have not been fully refactored, such as `ColumnwiseTransformer` - but mostly decoupled from the vast majority of the code base.

PR #3256, opened 2022-08-13
[ENH] channel selection (Dhariyal et al 2021) with arbitrary distance

fkiraly:
Illustrative refactor of one of the two algorithms in https://github.com/alan-turing-institute/sktime/pull/3248 to:

* allow arbitrary distances as input
* retain variable names in the indices, e.g., if variables are named in a `pandas` object. These are returned as `channels_selected_idx_`

Additional changes:
* reduced the dataset in the examples to avoid long test run times

For @haskarb, as illustration.
You can then use this as a blueprint to extend the second transformer, if you want.


fkiraly:
@haskarb, could you kindly have a look, whether this is still the same algorithm?
(I think it should be, just more general)

fkiraly:
I'm not sure why you would be expecting that to work, the `distance` must be a distance object.

To get the cosine distance, you would have to do this:

```python
from sktime.dists_kernels import FlatDist, ScipyDist
from sktime.transformations.panel.channel_selection import ElbowClassSum

distance = FlatDist(ScipyDist("cosine"))
cs = ElbowClassSum(distance)
```

Though I get that the expectation for a string is to use the respective `scipy` distance - so I could make it that, if a string is passed, it will automatically construct `FlatDist(ScipyDist(the_string))` and use that.

fkiraly:
pushed a mini commit, str should work now as parameter, and it defaults as stated above (tabular distance applied to flattened time series)

haskarb:
Looks fine to me.

fkiraly:
alright!

fkiraly:
@haskarb, feel free to extend the other transformer in the same way, you have a blueprint now that you can use.

PR #3257, opened 2022-08-13
[BUG] Fix tag in DirectReductionForecaster

KishManani:
`DirectReductionForecaster` is meant to allow for exogenous variables. However, because it inherits from `BaseForecaster` the default tag for exogenous variables was: ` "ignores-exogeneous-X": True`. This meant that any instance of `DirectReductionForecaster` was ignoring any exogenous variables passed to it during training and prediction. This PR fixes the tag so that `DirectReductionForecaster` uses `X`.


fkiraly:
oops - well, good that we put the warning in there.

Thanks, @KishManani!

fkiraly (approved):

PR #3260, opened 2022-08-13
[ENH] set default `ignores-exogeneous-X` to `False`

fkiraly:
This PR sets the default for the forecaster tag `ignores-exogeneous-X` to `False`.

This is the safer default, or otherwise exogeneous `X` will be inored under the default setting, and this might be accidental and unexpected for an implementer.

See bug here:
https://github.com/alan-turing-institute/sktime/pull/3257

### DIFF
--- a/sktime/forecasting/base/_base.py
+++ b/sktime/forecasting/base/_base.py
@@ -88,7 +88,7 @@ class BaseForecaster(BaseEstimator):
     # default tag values - these typically make the "safest" assumption
     _tags = {
         "scitype:y": "univariate",  # which y are fine? univariate/multivariate/both
-        "ignores-exogeneous-X": True,  # does estimator ignore the exogeneous X?
+        "ignores-exogeneous-X": False,  # does estimator ignore the exogeneous X?
         "capability:pred_int": False,  # can the estimator produce prediction intervals?
         "handles-missing-data": False,  # can estimator handle missing data?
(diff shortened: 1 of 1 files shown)


aiwalter:
I think this was not a very good choice as when I run now
```python
from sktime.registry import all_estimators
out = all_estimators("forecaster", filter_tags={"ignores-exogeneous-X": False})
```
I get a long list of compositions that actually handle `X` mostly optionally depending on the inner forecaster. Classy Franz PR that was not reviewed. Either you dont change default tag or you dynamically adjust the tags then also in the compositions.

Also e.g. `Croston` has now a wrong tag, probably other forecasters as well.

fkiraly:
> Classy Franz PR that was not reviewed

Martin, not constructive.

You complain about not reviewing.

Maybe think: who could be one person who could have reviewed this?

Hint: starts with "M", ends with "artin".

aiwalter:
can you pls fix the tag in `Croston`, its a bug that was caused by this PR

fkiraly:
@aiwalter.

* please remain constructive and assume best intentions.
* please submit a proper bug report in the issue tracker, not in a closed PR where no one looks.
* please stop with putdowns ("Franz PR") and blame. `sktime` is a volunteer driven open source project.
* in particular, please do not put pressure on other contributors to fix or write something you want
* as you are a core developer, you can consider fixing bugs that you find, or giving constructive PR reviews

PR #3261, opened 2022-08-13
[ENH] in `Lag`, make column naming consistent between single-lag and multi-lag case

KishManani:
Problem:
`Lag` outputs column names prefixed with "lag_x__" when using multiple lags. If a single lag is specified then the output column name is the same as the original column name (if the input is a `pd.DataFrame` rather than a `pd.Series`) without "lag_x__" prefixed to the original column name. It would be desirable for the output column names to be consistent, irrespective if a single lag or multiple lags is specified.

Fix implemented:
`Lag` now outputs the same column name irrespective of whether a single lag or multiple lags is specified.

### DIFF
--- a/sktime/transformations/series/lag.py
+++ b/sktime/transformations/series/lag.py
@@ -260,14 +260,9 @@ class Lag(BaseTransformer):
             Xt_list.append(Xt)
 
         lag_names = self._yield_shift_param_names()
-        if len(Xt_list) == 1:
-            Xt = Xt_list[0]
-            lag_name = next(lag_names, None)
-            Xt = Xt.add_prefix(lag_name + "__")
-        else:
-            Xt = pd.concat(Xt_list, axis=1, keys=lag_names, names=["lag", "variable"])
-            if self.flatten_transform_index:
-                Xt.columns = flatten_multiindex(Xt.columns)
+        Xt = pd.concat(Xt_list, axis=1, keys=lag_names, names=["lag", "variable"])
+        if self.flatten_transform_index:
+            Xt.columns = flatten_multiindex(Xt.columns)
 
         # some pandas versions do not sort index automatically after concat
         # so removing will break specific pandas versions
--- a/sktime/transformations/series/lag.py
+++ b/sktime/transformations/series/lag.py
@@ -259,10 +259,12 @@ class Lag(BaseTransformer):
 
             Xt_list.append(Xt)
 
+        lag_names = self._yield_shift_param_names()
         if len(Xt_list) == 1:
             Xt = Xt_list[0]
+            lag_name = next(lag_names, None)
+            Xt = Xt.add_prefix(lag_name + "__")
         else:
-            lag_names = self._yield_shift_param_names()
             Xt = pd.concat(Xt_list, axis=1, keys=lag_names, names=["lag", "variable"])
             if self.flatten_transform_index:
                 Xt.columns = flatten_multiindex(Xt.columns)
--- a/sktime/transformations/series/tests/test_featureizer.py
+++ b/sktime/transformations/series/tests/test_featureizer.py
@@ -28,4 +28,4 @@ def test_featurized_values():
     exp_transformer = ExponentTransformer()
     expected_len = lags + len(y_test)
     y_hat = exp_transformer.fit_transform(y[-expected_len:])
-    assert_array_equal(X_hat["TOTEMP"].values, y_hat.values)
+    assert_array_equal(X_hat[f"lag_{lags}__TOTEMP"].values, y_hat.values)
--- a/sktime/transformations/series/tests/test_lag.py
+++ b/sktime/transformations/series/tests/test_lag.py
@@ -4,6 +4,8 @@
 
 __author__ = ["fkiraly"]
 
+import itertools
+
 import pandas as pd
 import pytest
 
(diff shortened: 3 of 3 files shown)


fkiraly (changes requested):
Yes, I did think about this a lot, but went with the current solution, because I thought that would be the user expectation. This is a trade-off between user frustration and consistency.

In hindsight, I suppose I overestimated user frustration? Consistency might be more important if others use this as a component somewhere, which is a common use case because `Lag` will rarely be used in isolation.

Could you give a bit more insight into why you would like to change it this way? I.e., comment about the "user frustration" point, if a single variable is changed to lag-underscore-etc?

Re code:
If you want to remove the case distinction, I think it would be possible more easily, by just removing the "if" branch of the "if/else", and running only the "else" part?

KishManani:
The context which drove me to want to make a change was related to creating a pipeline of transformers. I have a pipeline where I want to apply the `Lag` transformer to different columns with different lags, in some cases I want to apply a single lag to a given column like so:

```
feat_eng_pipe = ColumnTransformer(
                  transformers=[
                    ('', 'passthrough', ['x1']),
                    ('lag_x2', Lag(lags=[3]), ['x2']),
                    ('lag_x3', Lag(lags=[1,2,3]), ['x3']),
                    ],
                  remainder='drop',
                  preserve_dataframe=True
                  )
```

The resulting dataframe from applying this pipeline will give different column names to lag variables created from `x2` compared to `x3`. This inconsistency is troublesome because I will now have a dataframe where the lag variables of `x3` are prefixed with `lag_{}__` but the lag variable of `x2` will just be called `x2` which could be mistaken for just passing through `x2`. Note, in this example because no two transformers act on the same column we do not get the transformer names being prefixed to the output columns.

> Re code:
If you want to remove the case distinction, I think it would be possible more easily, by just removing the "if" branch of the "if/else", and running only the "else" part?

Yes you're right! I've made the change and also added a unit test. It would be good to get your feedback on the unit test also if you think it looks a bit verbose.

fkiraly:
>This inconsistency is troublesome

I see, I'm convinced now - the inconsistency is a frustrated user expectation bordering on being a bug, in this case.

> I've made the change and also added a unit test. It would be good to get your feedback on the unit test also if you think it looks a bit verbose.

The new test looks good to me.

Planning to approve/merge once/if the CI suite passes.

fkiraly:
PS: since 0.13.1, you should be able to do sth like this:

```python
feat_eng_pipe = Lag()['x1'] + Lag(lags=[3])['x2'] + Lag(lags=[1,2,3])['x3']
```
where the square brackets apply the transformer only to those variables, and the addition means feature union.

Optional passthrough is not yet a dunder, but was thinking of the tilde.

KishManani:
> PS: since 0.13.1, you should be able to do sth like this:
> 
> ```python
> feat_eng_pipe = Lag()['x1'] + Lag(lags=[3])['x2'] + Lag(lags=[1,2,3])['x3']
> ```
> 
> where the square brackets apply the transformer only to those variables, and the addition means feature union.
> 
> Optional passthrough is not yet a dunder, but was thinking of the tilde.

Thanks! Very cool! Will be using this in the future. Tilde sounds very reasonable. An alternative would be to implement an identity transformer if you wanted a more explicit, but more verbose, syntax.

fkiraly:
> Thanks! Very cool! Will be using this in the future. Tilde sounds very reasonable. An alternative would be to implement an identity transformer if you wanted a more explicit, but more verbose, syntax.

Say, what do you think of the variable naming that happens in this case? I think you get a double prefix with `Lag`, once for the transformer, once for the lag. That looks a bit odd.

fkiraly (approved):
looks good, thanks!

KishManani:
> > Thanks! Very cool! Will be using this in the future. Tilde sounds very reasonable. An alternative would be to implement an identity transformer if you wanted a more explicit, but more verbose, syntax.
> 
> Say, what do you think of the variable naming that happens in this case? I think you get a double prefix with `Lag`, once for the transformer, once for the lag. That looks a bit odd.

Hi @fkiraly. 

It's a tricky one. You get the double prefix in these cases:

1) Two or more transformers applied to a single column

```
data = pd.DataFrame(data={'x1':[1,2,3], 'x2':[1,2,3], 'x3':[1,2,3]})

feat_eng_pipe = ColumnTransformer(
                  transformers=[
                    ('', 'passthrough', ['x1']),
                    ('lag_foo', Lag(lags=[1]), ['x2']),
                    ('lag_bar', Lag(lags=[2]), ['x3', 'x2']),
                    ],
                  remainder='drop',
                  preserve_dataframe=True
                  )

feat_eng_pipe.fit_transform(data)
```
Resulting column names: `['__x1', 'lag_foo__lag_1__x2', 'lag_bar__lag_2__x3',
       'lag_bar__lag_2__x2']`

2) Using the alternative method of constructing a FeatureUnion mentioned above irrespective of whether a column is used by more than one transformer

```
data = pd.DataFrame(data={'x1':[1,2,3], 'x2':[1,2,3], 'x3':[1,2,3]})

feat_eng_pipe = Lag()['x1'] + Lag(lags=[1])['x2'] + Lag(lags=[2])[['x3']]

feat_eng_pipe.fit_transform(data)
```
Resulting column names: `['TransformerPipeline_1__lag_0__x1', 'TransformerPipeline_2__lag_1__x2',
       'TransformerPipeline_3__lag_2__x3']`

I'm not sure how you can make `Lag` stand alone and also work with pipelines in a way that the column names are simple in both.

Issue #3267, opened 2022-08-14
[BUG] Updating a `TransformedTargetForecaster` with `Deseasonalizer` breaks


fkiraly:
From discord help-desk, @foobar

```python
from sktime.forecasting.theta import ThetaForecaster
from sktime.forecasting.compose import TransformedTargetForecaster
from sktime.transformations.series.detrend import Deseasonalizer
from sktime.datasets import load_airline

all_df = load_airline().to_frame()


model = TransformedTargetForecaster(
    [
        ("deseasonalize", Deseasonalizer(model="additive", sp=12)),
        ("forecast", ThetaForecaster()),
    ]
)
train_df = all_df["1949":"1950"]
model.fit(train_df)
model.update(y=all_df["1951"])
```

Gives an error `TypeError: _fit() got an unexpected keyword argument update_params`

Workaround: Pass `Deseasonalizer` into a `OptionalPassthrough` object 
i.e.  `sktime.transformations.series.compose.OptionalPassthrough(Deseasonalizer(...))`

fkiraly:
confirmed, the line with the `_fit` call should not have `update_params` in it, that is not a `fit` parameter.

PR #3270, opened 2022-08-14
[ENH] forecasting pipeline test case with `Detrender`

fkiraly:
Switches the transformer in one of the three forecasting pipeline test cases out for the `Detrender`, to cover transformer `update` from a component, and the bug https://github.com/alan-turing-institute/sktime/issues/3267.

No coverage is removed, since the replaced `ExponentTransformer` was used in another test case already, as the transformer.

Depends on:
* https://github.com/alan-turing-institute/sktime/pull/3287 since it triggers the edge case with `infer_freq` on short time series
* https://github.com/alan-turing-institute/sktime/pull/3290 as it also triggers bugs around the `name` attribute


fkiraly:
Should detect #3267 before the fix is merged.

fkiraly:
Detects #3267, and some other stuff - trying to vectorize over single indices?
Looks like a mistake in not recognizing the instance index, and instead treating time index as instance index.

PR #3272, opened 2022-08-14
[ENH] refactor of `OptionalPassthrough` as a delegator

fkiraly:
This PR refactors `OptionalPassthrough` as a delegator, and adds an identity transformer `Id` which is used as part of this.

Features added through this approach:
* `self.transformer_` is consistent, always a transformer
* `update` is now properly delegated (previously, it was not), more generally, any delegation logic or bugfixes are delegated (pun intended) to `_DelegatedTransformer`


fkiraly:
Say, @aiwalter, are there any specific tests of `OptionalPassthrough`? I.e., that it works as the transformer or identity?

aiwalter:
> Say, @aiwalter, are there any specific tests of `OptionalPassthrough`? I.e., that it works as the transformer or identity?

Oh yep, I just created a PR: #3276

PR #3276, opened 2022-08-15
[ENH] Added test for OptionalPassthrough

aiwalter:
#### Reference Issues/PRs
#3272

### DIFF
--- a/sktime/transformations/series/tests/test_optional_passthrough.py
+++ b/sktime/transformations/series/tests/test_optional_passthrough.py
@@ -12,7 +12,7 @@ from sktime.utils._testing.series import _make_series
 
 @pytest.mark.parametrize("passthrough", [True, False])
 def test_passthrough(passthrough):
-    """Test that pandas.method and PandasTransformAdaptor.fit.transform is the same."""
+    """Test that passthrough works as expected."""
     y = _make_series(n_columns=1)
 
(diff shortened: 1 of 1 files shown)


fkiraly:
* I think we should also cover `fit`, with an estimator where `fit` is not empty. I would recommend `BoxCoxTransformer` as a second transformer
* the `try`/`assert` construction seems brittle - I would instead compare to a direct application of the estimator (e.g., `ExponentTransformer`) without wrapping in `OptionalPassthrough`

fkiraly:
I meant, in addition, rather than, instead. I.e., mark.parametrize over estimator which can be exponential or boxcox?

fkiraly:
(worst case, just copy-paste your old test in, and I´m also happy with that, mark.parametrize would be more elegant)

aiwalter:
I anyway call `fit_transform` ,so where is the difference if fit is empty?

fkiraly:
>I anyway call `fit_transform` ,so where is the difference if fit is empty?

If the wrapped `fit` is empty, the tag will be copied to `OptionalPassthrough`, so we do not test the delegation in `_fit`. Conversely, if the wrapped `fit` is not empty, we delegate, and do not test the skipping logic (both branches could hide bugs).

fkiraly (approved):
I think this is good enough as a basis for refactor, because it tests the more important of the branches (`fit` delegation), so I merge - would still appreciate covering the different cases here (separate PR)

PR #3292, opened 2022-08-19
[ENH] Migrated LSTMFCN Classifer from sktime-dl

### LINKED ISSUE #3193: [ENH] Adding LSTMFCN DL model for classification
On the topic of migrating sktime-dl models to sktime, adding FCN model for classification from sktime-dl

AurumnPegasus:
#### Reference Issues/PRs

Fixes #3193 

##### For all contributions

##### For new estimators


AurumnPegasus:
Hey @fkiraly ,
Wanted to confirm if the creation of `modules` directory here is in line with what you were recommending when we discussed this.

AurumnPegasus:
Sorry I forgot to comment on this,
The issue I am facing with this classifier is that authors in sktime-dl have created there on keras layer (LSTMAttention) for this in a separate file: [layer_utils in sktime_dl](https://github.com/sktime/sktime-dl/blob/b565b7499f58f43da7314f1bf26eccce94e88134/sktime_dl/utils/layer_utils.py#L424)

To migrate the same structure into sktime I was thinking to create a modules directory (and then add appropriate layers there) (I have done this in current PR)

Need clarification if I can approach with this (also there seem to be dependancy issues wrt it, so some help on that) or if there is a better and cleaner option

fkiraly:
can this not go in the `network` dir? Let's not proliferate subdirs of src root.

AurumnPegasus:
Hey,
Facing a soft dependancy issue in this, we need to import `keras.layers` in the file in this in the file (not within the class/function as is recommended) in the file `attentionlstm.py`
Is there a possible workaround for it?

fkiraly:
stick the import inside the function

AurumnPegasus:
@fkiraly it cant, since those imports are essentially base classes from which our defined classes `AttentionLSTM` and `AttentionLSTMCell` are inheriting

fkiraly (changes requested):
Thanks for looking into the soft dependency issue!

Few comments:
* this PR has merge conflicts with `main` would be great if you could resolve
* I thought the idea of the `BaseNetwork` class was to wrap networks, e.g., the layer that you are now wrapping in a `_getsthsth` function. Can we not wrap it as a `BaseNetwork` instead, for architectural consistency?

AurumnPegasus:
@fkiraly 
From what I understand, the idea behind `Network` classes was such that any keras network would be wrapped around it, yes, but in a way which would allow us to directly use the particular network for the estimators (like `CNNNetwork` can be easily used within `CNNClassifier` just by calling an object of it)
In this case, the `AttentionLSTM` network is not meant to be used in a way where we directly use it within the estimator, just that its a network which is used within another network

A possible solution could be that we stick the functions `_getAttentionLSTMCell` and `_getAttentionLSTM` (basically all the functions in the file `networks/attentionlstm.py`) within the `LSTMFCNNetwork`, which would make the network class a bit crowded, but would still preserve the architectural consistency.

sktime/tests/_config.py
```diff
@@ -60,6 +60,10 @@
         "test_fit_does_not_overwrite_hyper_params",
         "test_save_estimators_to_file",
     ],
+    "LSTMFCNClassifier": [
```
GuzalBulatova, line 63: Any minimal tests specific for the class would be great to have!

sktime/networks/lstmfcn.py
```diff
@@ -0,0 +1,128 @@
...
+    random_state    : int, default = 0
+        seed to any needed random actions
+
+    Notes
+    -----
+    Implementation of LSTMFCNClassifier from Karim et al (2019).
+
+    References
+    ----------
+    .. [1] @article{Karim_2019,
```
GuzalBulatova, line 37: Please make the format of article reference the same as you made on line 52 in lstmfcn.py

GuzalBulatova (changes requested):
Great job, @AurumnPegasus ! Looks good already.
A blocking moment would be tests for these networks - something to make sure the main functionality is behaving as expected.
Also a reminder - please update from main.

sktime/tests/_config.py
AurumnPegasus: I agree, although I think this should be a new PR on its own, since we can add common basic tests to all DL Classes.

sktime/networks/lstmfcn.py
AurumnPegasus: Seems like I had made a mistake in referencing as well!
Wrote the references in the format needed, let me know if its cool now.

PR #3302, opened 2022-08-20
[ENH] `pmdarima 2.0.0` compatibility fix - use absolute index in return

### LINKED ISSUE #3300: [BUG] pmdarima 2.0.0 incompatibility - return indices
Soft dependency `pmdarima` 2.0.0 was released a few hours ago, and it seems to break our `ARIMA` interface.

Suspicion: return indices are now absolute and no longer relative to the cutoff.

fkiraly:
Fixes #3300 (hopefully) by changing cutoff relative references in return to absolute index.


fkiraly:
Seems to work as a fix now!

Idea is to be more conservative in interfacing - simply ignoring the index returned by `pmdarima`, and overwriting it with the expected index. By that, it might be more error prone (because we are no longer implicitly checking for correct index return) - so would appreciate some eyes on whether this is still fine.

FYI @ngupta23, would also appreciate your opinion.

PR #3306, opened 2022-08-20
[ENH] `BaseTransformer` data memory - always enabled

fkiraly:
This adds a data memory to transformers, similar to forecasters.

It stores all data seen in an attribute `self._X`, as a reference if possible (due to coercions, storing a reference may not be possible).

This allows to refactor some transformers which already to that internally, and remove a layer of boilerplate:

* `Imputer`
* `Lag`
* `IndexSubset`
* `PandasTransformAdaptor`
* `WindowSummarizer`

While convenient, this may cause some memory overhead, especially in pipelines with many transformers, so reviewers should also put that to consideration.

An alternative may be turning this off in transformers that do not need it, say, via a tag, or a base class init argument. The option with a tag is carrie out in https://github.com/alan-turing-institute/sktime/pull/3307, perhaps that is preferable to minimize effect on memory in transformers that do not need the memory.

### DIFF
--- a/sktime/datatypes/_utilities.py
+++ b/sktime/datatypes/_utilities.py
@@ -321,6 +321,8 @@ UPDATE_DATA_INTERNAL_MTYPES = [
 def update_data(X, X_new=None):
     """Update time series container with another one.
 
+    Coerces X, X_new to one of the assumed mtypes, if not already of that type.
+
     Parameters
     ----------
     X : None, or sktime data container, in one of the following mtype formats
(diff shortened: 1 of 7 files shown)


fkiraly:
converting to draft - unrefactored transformers are clashing with the data memory introduction. #3307 should not have the same issue, as we enable the memory only punctually.

fkiraly:
not a good idea due to memory usage. Let´s do #3307 instead

PR #3327, opened 2022-08-23
[ENH] global/local setting for `DirectReductionForecaster`

fkiraly:
This introduces a parameter `pooling` to `DirectReductionForecaster`, which can be set to `"global"` for global pooling, `"local"` for per-instance pooling, and `"panel"` for panel-wise pooling. Pooling refers to the `X` and `y` passed to the supervised sklearn regressors, which could be constructed from single instances, or the entire panel/hierarchy, in case there are more than one instances.

Giving the user a choice is the way to go here imo, as all settings can make sense.

The logic of `DirectReductionForecaster` also had to be extended to also accommodate for potential multi-indexed `pandas.DataFrame` appearing in the `fit`/`predict` parts. Unfortunately this is a bit fiddly, since it seems `pandas` interfaces are quite discrepant between the 1-level case and the 2- and multi-level case.

Testing was done locally using the full panel/hierarchical suite introduced in https://github.com/alan-turing-institute/sktime/pull/3321, but that is *not* included in this PR, to minimize size of PR and interdependencies. #3321 should be merged first, therefore.

---

Update: #3321 is merged now, so test passing pass with full hierarchical test suite in #3321.

### DIFF
--- a/sktime/forecasting/compose/_reduce.py
+++ b/sktime/forecasting/compose/_reduce.py
@@ -1259,6 +1259,25 @@ def _coerce_col_str(X):
     return X
 
 
+def slice_at_ix(df, ix):
+    """Slice pd.DataFrame at one index value, valid for simple Index and MultiIndex.
+
+    Parameters
+    ----------
+    df : pd.DataFrame
+    ix : pandas compatible index value
+
+    Returns
+    -------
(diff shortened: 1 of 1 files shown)


fkiraly:
FYI @danbartl, @KishManani, you might be interested in this. Still prototypical, so leaving the warning in.

PR #3331, opened 2022-08-24
[DOC] Some tips on getting virtual environments to work

miraep8:
I recently ran into some import issues with sktime using jupyter notebooks that essentially stemmed from not having my virtual environment properly linked as a kernel!

Moving forward I plan to just use jupyter notebooks in VScode (and recommend others do too), but figured I would add some tips to the docs (as well as some general resources for using virtual environments) to perhaps save some time for future contributors. :)

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -217,8 +217,8 @@ adding your virtual environment as a new kernel for your notebook.
 Other Startup Resources
 -----------------------
 
-Setting up a virtual environment
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Virtual environments
+~~~~~~~~~~~~~~~~~~~~
 
 Two good options for virtual environment managers are:
  - `conda <https://uoa-eresearch.github.io/eresearch-cookbook/recipe/2014/11/20/conda/>` (many of the sktime community uses this
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -169,10 +169,29 @@ This section outlines the ``sktime`` build requirements. These are required for:
 * installing ``sktime`` from source, e.g., development versions
 * the advanced developer set-up
 
+Setting up a virtual environment
+""""""""""""""""""""""""""""""""
+
+We recommend that you do all developement for sktime within a virtual environment. Using a virtual environment helps keep all the dependencies for the package seperated
+from other python installations on your device, which makes it easier to track new dependencies and avoid clashes from conflicting package versions! Two good options for
+setting your own virtual environment are `conda <https://uoa-eresearch.github.io/eresearch-cookbook/recipe/2014/11/20/conda/>` (many of the sktime community uses this
+virtual environment, so this is recommended) and `venv <https://realpython.com/python-virtual-environments-a-primer/>` (also quite good!).
+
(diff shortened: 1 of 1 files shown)


docs/source/installation.rst
```diff
@@ -207,6 +207,22 @@ Some users have experienced issues when installing NumPy, particularly version 1
     Another option under Windows is to follow the instructions for `Unix-like OS`_, using the Windows Subsystem for Linux (WSL).
     For installing WSL, follow the instructions `here <https://docs.microsoft.com/en-us/windows/wsl/install-win10#step-2---check-requirements-for-running-wsl-2>`_.
 
+Some quick tips on virtual environments
+---------------------------------
```
fkiraly, line 211: header should be at the same level as "Setting up a development environment", and underline should be same length as header text

docs/source/installation.rst
```diff
@@ -207,6 +207,22 @@ Some users have experienced issues when installing NumPy, particularly version 1
     Another option under Windows is to follow the instructions for `Unix-like OS`_, using the Windows Subsystem for Linux (WSL).
     For installing WSL, follow the instructions `here <https://docs.microsoft.com/en-us/windows/wsl/install-win10#step-2---check-requirements-for-running-wsl-2>`_.
 
+Some quick tips on virtual environments
```
fkiraly, line 210: not a good header, very vague. Headers should be concrete and short. Can you find sth clearer?

docs/source/installation.rst
```diff
@@ -207,6 +207,22 @@ Some users have experienced issues when installing NumPy, particularly version 1
...
+---------------------------------
+
+As mentioned above -  we recommend that you do all developement for sktime within a virtual environment.  Two good options for doing so are
```
fkiraly, line 213: avoid duplications, text should be crisp. Header should indicate what this section is about.

fkiraly (changes requested):
Thanks for the addition, I think this is useful advice!

Change requests relate to making this nice and short:

* avoid duplications in the text to the subsection above. Focus on new content and "defining content" for the section. I suppose that is specifics about vs code and jupyter notebooks?
* ensure it is clear to the reader what the section is about. Think about the user journey, eyes go from section header to subsection header and then to content. A reader should understand what the subsection will be about before reaching it!

docs/source/installation.rst
miraep8: Replaced with `Setting up a virtual environment`

docs/source/installation.rst
miraep8: Makes sense, I moved this section before the `Setting up a development environment` and removed references to that.

fkiraly: I think there's better ways to address the comments - the new section is very verbose. Good instruction manuals outline clear short steps.

Think of it as writing code. You are executing code "on the reader", i.e., telling them what to do. It is not very clear what to do from the new section. There is an "if using vscode, elif using sth else", e.g., which is not too clear.

The second section also seems to be a condition for the first section, because you need to have your python environment before using it in notebooks! So it seems, at least in parts, the wrong way round...

fkiraly (approved):
Structure and placement are good now, text is brief and crisp, great!

Issue #3316, opened 2022-08-22
[BUG] Errors when using the recursive strategy in make_reduction when forecasting panel data


wregter:
**Describe the bug**
 I experience errors when using panel data and the recursive strategy in make_reduction. I would like to use scikit-learn regressors to perform global forecasting on panel data but the predict method is giving me errors that I do not know how to solve.

Side note: in the reproducible code I set the multiindex to integer values because otherwise this also produces an error.

**To Reproduce**

```python
<
import pandas as pd
import numpy as np
from sktime.utils._testing.hierarchical import _bottom_hier_datagen
from sktime.forecasting.compose import make_reduction
from sktime.forecasting.model_selection import temporal_train_test_split
from sktime.forecasting.base import ForecastingHorizon
from sklearn.linear_model import LinearRegression
from sktime.datatypes import check_raise

y = _bottom_hier_datagen(no_levels=1)
check_raise(y, mtype="pd-multiindex", scitype="Panel")

date_length = y.index.get_level_values(1).shape[0]
level_length = y.index.get_level_values(0).shape[0]
y = y.set_index(y.index.set_levels(np.arange(date_length), level=1))
y = y.set_index(y.index.set_levels(np.arange(level_length), level=0))

y_train, y_test = temporal_train_test_split(y, test_size=5)

forecaster = make_reduction(LinearRegression(), strategy='recursive')
horizon = pd.unique(y_test.index.get_level_values(1))
fh = ForecastingHorizon(horizon, is_relative=False)
forecaster.fit(y_train, fh=fh)
y_pred = forecaster.predict()
>
```

Error message preview:

TypeError                                 Traceback (most recent call last)
c:\Users\WREGTE16\Documents\GitHub\nl_sales_forecast\testing\a.ipynb Cell 1 in <cell line: 24>()
     [22](vscode-notebook-cell:/c%3A/Users/WREGTE16/Documents/GitHub/nl_sales_forecast/testing/a.ipynb#W0sZmlsZQ%3D%3D?line=21) fh = ForecastingHorizon(horizon, is_relative=False)
     [23](vscode-notebook-cell:/c%3A/Users/WREGTE16/Documents/GitHub/nl_sales_forecast/testing/a.ipynb#W0sZmlsZQ%3D%3D?line=22) forecaster.fit(y_train, fh=fh)
[...]

PR #3333, opened 2022-08-24
[ENH] global/local recursive reduction prototype

fkiraly:
Global/local recursive reduction prototype along the lines of https://github.com/alan-turing-institute/sktime/pull/3327.

Dependent on https://github.com/alan-turing-institute/sktime/pull/3327 as it uses a joint reduction mixin, refactored from https://github.com/alan-turing-institute/sktime/pull/3327.

Addresses https://github.com/alan-turing-institute/sktime/issues/3224 and #3316


vattyamp:
I think the new 'RecursiveReductionForecaster' needs to be referenced in` __all__` in init.py and _reduction.py for importing purposes for importing purposes

fkiraly:
> I think the new 'RecursiveReductionForecaster' needs to be referenced in __all__ in init.py and _reduction.py for importing purposes for importing purposes

Yes, that's right, and eventially it will also have to be referenced in the API reference. 

But currently it's experimental, and we do not have the final form yet, so I don't want to make it "too public", until the interfaces are consolidated.
This way, no one will import it by accident...

PR #3336, opened 2022-08-24
[ENH] design for generic save/load aka serialization/deserialization of estimators

fkiraly:
This adds a prototype design for serializing and deserializing estimators.
This follows the current (Aug 19) draft STEP 27 https://github.com/sktime/enhancement-proposals/pull/27.

This is done via `save` and two `load` methods which can be overridden by class, and a generic `load` (class agnostic), in the `BaseObject` module.

Also changes the pickle/unpickle test to a generic serialize/deserialize test which allows custom, class specific implementations of serialize and deserialize.

FYI @AurumnPegasus - design alternative to https://github.com/alan-turing-institute/sktime/pull/3128 which avoids the workflow where we have to construct an object first when we load.

### DIFF
--- a/sktime/base/__init__.py
+++ b/sktime/base/__init__.py
@@ -9,7 +9,9 @@ __all__ = [
     "BaseObject",
     "BaseEstimator",
     "_HeterogenousMetaEstimator",
+    "load",
 ]
 
 from sktime.base._base import BaseObject, BaseEstimator
 from sktime.base._meta import _HeterogenousMetaEstimator
+from sktime.base._serialize import load
--- a/sktime/base/__init__.py
+++ b/sktime/base/__init__.py
@@ -12,6 +12,6 @@ __all__ = [
     "load",
 ]
 
-from sktime.base._base import BaseObject, BaseEstimator
+from sktime.base._base import BaseEstimator, BaseObject
 from sktime.base._meta import _HeterogenousMetaEstimator
 from sktime.base._serialize import load
--- a/sktime/base/_base.py
+++ b/sktime/base/_base.py
@@ -616,7 +616,7 @@ class BaseObject(_BaseEstimator):
         Returns
         -------
         if `path` is None - in-memory serialized self
-        if `path` is file location - BaseObjectFile with reference to location
+        if `path` is file location - ZipFile with reference to location
         """
         import pickle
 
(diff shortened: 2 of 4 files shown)


AurumnPegasus:
I see,
I do not know if serialisation/deserialisation works for keras models (will check it out once). 
I was thinking something along the lines of (better explained in the step document I was working on, will push that soon):
1. Creating a `save` function which does 2 things:
    a. Creates pickle of class dictionary (EXCLUDING `optimizer`)
    b. Saves the network using `keras.save`
In this case, we will be able to solve both the problems my design was having, firstly being able to save estimator parameters, and secondly, there would be no pickle error since we are excluding the optimizer from the state dict.
2. A generic load function would then first load from the `pickle` file, then load in the network using `keras.load`

fkiraly:
with this design, you don't have to implement separate functions or logic, just overload the three new `BaseObject` methods for neural networks.

AurumnPegasus:
Hey @fkiraly ,
Can you please explain a doubt I had wrt this implementation?

```python
if isinstance(serial, tuple):
        cls = serial[0]
        stored = serial[1]
        return cls.load_from_serial(stored)
    elif isinstance(serial, (str, ZipFile)):

        if isinstance(serial, str):
            zipfile = ZipFile(serial)

        with zipfile.open("metadata", mode="r") as metadata:
            cls = metadata.read()
        with zipfile.open("object", mode="r") as object:
            return cls.load_from_path(object.read())
```
Here, I am assuming `cls` would refer to the particular estimator. 
What is this line doing? `cls.load_from_path(object.read())` (is the `object.read` a pickle file?)

fkiraly:
>What is this line doing? `cls.load_from_path(object.read())` (is the `object.read` a pickle file?)

The idea is that `object.read()` can be anything, but the contract is that `cls.load_from_path` must convert it into an instance of `cls`.

In the default implementation, `object.read` is a pickle file, but you could override `load_from_path` (a class method) if the object us unpickleable.

achieveordie:
Could you confirm if this works for DL Classifiers, like `CNNClassifier()`? I altered `CNNClassifier` directly to serialize contents(`save(), load_from_serial(), __getstate__()` etc) and I got error for the line `meta_file.write(type(self))`:
```python
TypeError: object of type 'ABCMeta' has no len()
```

fkiraly:
> Could you confirm if this works for DL Classifiers

@achieveordie, in the tests, this is currently skipped for all DL classifiers, since they are not compatible with `pickle` (skip is in the `_config` file).

Re your proposed change, is there a draft PR? I could look at it.

Your error message might be related to `BaseClassifier` inheriting from `ABC`. Perhaps it works if you remove the inheritane from `ABC`?

fkiraly:
> Could you confirm if this works for DL Classifiers

@achieveordie, I suppose I have misunderstood your question.

Did you want to know whether the default in this PR works for DL classifiers?

The answer is "no", and I know that it doesn't work.

However, having the abstract methods in place, it should be possible to override this in the deep learning base class, with something that works - as you have probably tried to (this was the idea, I just don't know what precisely would work).

Issue #3337, opened 2022-08-25
[ENH] Add STRAY anomaly detection


KatieBuc:
Add STRAY (Search TRace AnomalY) anomaly detection [1]. 

"The HDoutliers [2] algorithm is a powerful unsupervised algorithm for detecting anomalies in high-dimensional data, with a strong theoretical foundation. However, it suffers from some limitations that significantly hinder its performance level, under certain circumstances. In this article, we propose an algorithm that addresses these limitations. We define an anomaly as an observation where its k-nearest neighbor distance with the maximum gap is significantly different from what we would expect if the distribution of k-nearest neighbors with the maximum gap is in the maximum domain of attraction of the Gumbel distribution. An approach based on extreme value theory is used for the anomalous threshold calculation. Using various synthetic and real datasets, we demonstrate the wide applicability and usefulness of our algorithm, which we call the stray algorithm. "

This task involves re-writing [this R code](https://github.com/pridiltal/stray/tree/master/R), as an estimator in the annotation module.

References: 
1. Talagala, Priyanga Dilini, Rob J. Hyndman, and Kate Smith-Miles. "Anomaly detection in high-dimensional data." Journal of Computational and Graphical Statistics 30.2 (2021): 360-374.
2. Wilkinson, Leland. "Visualizing big data outliers through distributed aggregation." IEEE transactions on visualization and computer graphics 24.1 (2017): 256-266.

lmmentel:
closed by #3338

PR #3338, opened 2022-08-25
[ENH] STRAY anomaly detection

KatieBuc:
#### Reference Issues/PRs
Issue #3337

Implements the novel anomaly detection algorithm STRAY (Search TRace AnomalY), converted from R code.

An initial implementation, assuming this will be refactored once we have more of an idea about annotators. This is an interesting example to reference because a lot of logic in the function `_find_HDoutliers` deals with missing data (i.e. removes all missing data, then re-indexes back to the original data). Using pandas would make this indexing conversion easier, I think?
In any case, it highlights an open question: how should we handle this in general.

Similar to Hidalgo, all the logic is done in `fit`, so that `predict` just returns a specific reference to `self`.

Happy to hear any feedback!

@lmmentel @Lovkush-A @fkiraly @miraep8 @aiwalter


KatieBuc:
@lmmentel @Lovkush-A @fkiraly @miraep8 @aiwalter this is ready for review!

As suggested, for experiment, this anomaly detection algorithm inherits from `BaseTransformer`. The only thing I find clunky is that we can't write to self in `_transform`, hence the warning. A suggestion would be to have instead an `annotate` function that can write to self (instead of `predict` or `transform`). It would be nice to be able to bundle the logic in `annotate` and have public variables updated to self. Does that make any sense? What does everyone else think?

sktime/annotation/stray.py
```diff
@@ -0,0 +1,283 @@
...
+        ----------
+        X : np.ArrayLike
+            Data for anomaly detection (time series).
+
+        Returns
+        -------
+        dict of anomalies and their corresponding scores
+        """
+        r = np.shape(X)[0]
+        idx_dropna = np.array([i for i in range(r) if not np.isnan(X[i]).any()])  # tag
```
lmmentel, line 201: how about [np.isnan](https://numpy.org/doc/stable/reference/generated/numpy.isnan.html) instead?

sktime/annotation/stray.py
```diff
@@ -0,0 +1,283 @@
...
+        # remember X for transform
+        self._X = X
+
+        info_dict = self._find_HDoutliers(X)
+        self.score_ = info_dict["outlier_scores"]
+        self.y_ = info_dict["outlier_bool"]
+
+        return self
+
+    def _transform(self, X, y=None):
```
lmmentel, line 250:
```suggestion
    def _transform(self, X: npt.ArrayLike, y: npt.ArrayLike = None) -> npt.ArrayLike:
```

sktime/annotation/stray.py
```diff
@@ -0,0 +1,283 @@
...
+        outlier_scores = np.array(
+            [list_scores.pop(0) if i in idx_dropna else np.nan for i in range(r)]
+        )
+
+        return {
+            "outlier_scores": outlier_scores,
+            "outlier_bool": outlier_bool,
+        }
+
+    def _fit(self, X, y=None):
```
lmmentel, line 226:
```suggestion
    def _fit(self, X: npt.ArrayLike, y: npt.ArrayLike = None):
```

sktime/annotation/stray.py
```diff
@@ -0,0 +1,283 @@
...
+        bound = np.Inf
+
+        for i in range(start, n):
+            if gaps[i] > log_alpha * ghat[i]:
+                bound = outlier_score[order][i - 1]
+                break
+
+        return np.where(outlier_score > bound)[0]
+
+    def _use_KNN(self, X, n):
```
lmmentel, line 155: Consider changing the name to something a bit more descriptive, e.g. `find_nearest_neighbors`.
```suggestion
    def _use_KNN(self, X: npt.ArrayLike, n: int) -> Dict:
```

lmmentel (changes requested):

sktime/annotation/stray.py
```diff
@@ -0,0 +1,283 @@
...
+    Attributes
+    ----------
+    score_ : pd.Series
+        Outlier score of each data point in X.
+    y_ : pd.Series
+        Outlier boolean flag for each data point in X.
+
+    References
+    ----------
+    [1] Talagala, Priyanga Dilini, Rob J. Hyndman, and Kate Smith-Miles.
```
aiwalter, line 72:
```suggestion
    .. [1] Talagala, Priyanga Dilini, Rob J. Hyndman, and Kate Smith-Miles.
```

sktime/annotation/stray.py
```diff
@@ -0,0 +1,283 @@
...
+    "Anomaly detection in high-dimensional data." Journal of Computational
+    and Graphical Statistics 30.2 (2021): 360-374.
+    [2] Wilkinson, Leland. "Visualizing big data outliers through
```
aiwalter, line 75:
```suggestion
    .. [2] Wilkinson, Leland. "Visualizing big data outliers through
```

sktime/annotation/stray.py
```diff
@@ -0,0 +1,283 @@
...
+    distributed aggregation." IEEE transactions on visualization and
+    computer graphics 24.1 (2017): 256-266.
+
+    Examples
+    --------
+    >>> from sktime.annotation.stray import STRAY
+    >>> from numpy import asarray
+    >>> X = asarray([3.7,3.2,3.4,3.6,-5.1,-5.2,-4.9])
```
aiwalter, line 83: please use an existing data from `sktime.datasets`

sktime/annotation/stray.py
aiwalter: should we really use `npt.ArrayLike`? In docstring we always write `pandas` so far :thinking:

sktime/annotation/stray.py
```diff
@@ -0,0 +1,283 @@
...
+        -------
+        self :
+            Reference to self.
+        """
+        # remember X for transform
+        self._X = X
+
+        info_dict = self._find_HDoutliers(X)
+        self.score_ = info_dict["outlier_scores"]
+        self.y_ = info_dict["outlier_bool"]
```
aiwalter, line 246: can you please look into `HampelFilter`, there we replace outliers with NaN or `bool`:
```
    return_bool : bool, optional
        If True, outliers are filled with True and non-outliers with False.
        Else, outliers are filled with np.nan.
```
Lets have same interface for both? Having NaN is convenient so user can directly apply an `Imputer` after to correct outliers. Whats do you think?

aiwalter: I think for now its best to follow `HampelFilter` interface until we have agreed on annotation output, e.g. storing stuff in index. Then we can change both to have same interface?

sktime/annotation/tests/test_stray.py
```diff
@@ -0,0 +1,274 @@
...
+__author__ = ["KatieBuc"]
+
+import numpy as np
+
+from sktime.annotation.stray import STRAY, standardize
+
+
+def test_default_1D():
+    """Test with default parameters and 1D input array."""
+    X = np.array(
```
aiwalter, line 13: can you use `make_forecasting_problem` to create test data?

aiwalter (commented):
nice :) left some comments.

sktime/annotation/stray.py
```diff
@@ -0,0 +1,283 @@
...
+    if diff == 0:
+        return np.zeros(len(x))
+    return (x - min(x)) / diff
+
+
+def standardize(x):
+    return (x - np.median(x)) / iqr(x)
+
+
+class STRAY(BaseTransformer):
```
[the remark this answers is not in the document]
KatieBuc: only because of the convenience of tags in `BaseTransformer`, specifically:
```
"X_inner_mtype": "np.ndarray"
```

currently not implemented in `BaseSeriesAnnotator`

sktime/annotation/stray.py
```diff
@@ -0,0 +1,283 @@
...
+        y : pd.Series, optional
+            Not used for this unsupervsed method.
+
+        Returns
+        -------
+        self :
+            Reference to self.
+        """
+        # remember X for transform
+        self._X = X
```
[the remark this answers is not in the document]
KatieBuc: It's more an artifact of not being able to write to self (for both `_transform` and `_predict`) - forcing all the logic to be done in `_fit`. So that `_transform` (or  `_predict`) just returns a specific output already stored in self that conforms to standard output. 

But then in `_transform`, since we pass in `X`, and this may of course differ from the `X` that was given to `_fit`, and we need to check for this (and refit if different) - hence the need to store the `X` for comparison. I think this is a bit clunky and would be happy to hear any alternatives!

sktime/annotation/tests/test_stray.py
KatieBuc: ahhh I created these arrays in R, and checked directly against the R output of the original implementation - I could include (as a comment) the R code so that developers know the context of this input data? Or you still think it's best to sever the connection to the R implementation and use an sktime data generator?

sktime/annotation/stray.py
```diff
@@ -0,0 +1,283 @@
...
+        Threshold for determining the cutoff for outliers. Observations are
+        considered outliers if they fall in the (1- alpha) tail of
+        the distribution of the nearest-neighbor distances between exemplars.
+    k : int, optional (default=10)
+        Number of neighbours considered.
+    knn_algorithm : str {"auto", "ball_tree", "kd_tree", "brute"}, optional
+        (default="brute")
+        Algorithm used to compute the nearest neighbors in
+        sklearn.neighbors.NearestNeighbors
+    normalize : callable {unitize, standardize} (default=unitize)
```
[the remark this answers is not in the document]
KatieBuc: It's necessary in the sense that some kind of normalisation is expected to happen (so the assumption in the logic hold true), and is defaulted to min-max normalization. If it were a preprocessing step, how could we incorporate that? Maybe have some sort of check to see if `is_normalized`?

sktime/annotation/stray.py
KatieBuc: ahh, yeah - changed it to `idx_dropna = np.argwhere((~np.isnan(X)).all(axis=1)).flatten()` instead

sktime/annotation/stray.py
KatieBuc: Can do - but I have a question: what if the input data has np.nan? How do we distinguish this from an anomaly when `return_bool = False`?

sktime/annotation/stray.py
KatieBuc: damn, changing back because `AxisError: axis 1 is out of bounds for array of dimension 1` in 1D case

sktime/annotation/stray.py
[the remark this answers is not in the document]
lmmentel: If the data is assumed to be normalized then that's something that should go into the docs. By analogy PCA works way better when input data is normalized but it is not required nor checked.

sktime/annotation/stray.py
KatieBuc: I have changed to boolean flags for output, for now. 

Exploring the other option, introducing np.nan for the anomalous values, would mean input
```
    X = np.array(
        [
            [-1.20706575, -0.57473996],
            [0.27742924, -0.54663186],
            [1.08444118, np.nan],
            [-2.3456977, -0.89003783],
            [0.42912469, -0.4771927],
            [0.50605589, -0.99838644],
        ]
    )
```
gives expected boolean output 
`y = [False, False, False, True, True, True]`
the algorithm specifies any na values are assumed normal data.

Having an output that is the original data with anomalies -> na gives
``` 
X = np.array(
        [
            [-1.20706575, -0.57473996],
            [0.27742924, -0.54663186],
            [1.08444118, np.nan],
            [np.nan, np.nan],
            [np.nan, np.nan],
            [np.nan, np.nan],
        ]
    )
```
which changes dimensionality, and is a bit confusing because the second most important information are the corresponding anomaly scores
scores_ = `[0.5233413, 0.5233413, np.nan, 0.7248368, 0.6035040, 0.8704687]`

Do we still think the np.nan anomaly overwrite makes sense for this use case? @aiwalter @lmmentel

KatieBuc:
Hey @lmmentel @Lovkush-A @fkiraly - something to flag about this normalization transformation pipeline.

1. In the original implementation of the algorithm, the NA's were dropped before normalization. Since one NA value deemed the whole row invalid. However, applying a column-wise transform for data with NA results in different normalization values i.e.
image
where the top are the computed values within the algorithm, the bottom are when the transformation is applied _first_, then na removed.

A simple solution would be to add the tag "handles-missing-data: False" (does this mean if na's are passed in they will be automatically filtered out?) and what does this mean for the output? 

For example, 
with the input data
```
    X = np.array(
        [
            [-1.20706575, -0.57473996],
            [0.27742924, -0.54663186],
            [1.08444118, np.nan],
            [-2.3456977, -0.89003783],
            [0.42912469, -0.4771927],
            [0.50605589, -0.99838644],
        ]
    )
```

The anomaly boolean flags and scores are:
`y_expected = np.array([0, 0, 0, 1, 1, 1])`
`    y_scores_expected = np.array(
        [0.5233413, 0.5233413, np.nan, 0.7248368, 0.6035040, 0.8704687]
    )
`
according to the original index (with na)

2. Another thing to note is that
```
    >>> scaler = MinMaxScaler()
    >>> X = scaler.fit_transform(X)
```
throws an error when the input is 1D, hence `X = np.array(X).reshape(-1, 1)` is needed.

3. Also, in another example (DOBIN), the scaler changes type from `pd.DataFrame` to `np.Array`
```
    >>> _, X = load_uschange()
    >>> scaler = MinMaxScaler()
    >>> X = scaler.fit_transform(X)
    >>> model = DOBIN()
    >>> X_outlier = model.fit_transform(pd.DataFrame(X))
```
I think 2,3 make a case for interfacing, not sure?

and 1 is something for us to consider when thinking about anomaly outputs with missing data.

sktime/annotation/stray.py
lmmentel: I don't see any value in overriding the input data. How would that be used, and under what circumstances? For me in the context of time series we are flagging anomalous **time point** or **periods of time** therefore a boolean 1D array should be the standard output. If the algorithm can **in addition** flag individual points I would say that a boolean mask with the same dimensionality as the input would be a good choice since you can then filter the input data whichever way you want and get either only anomalous data or only normal data. If the algorithm provides even more information like anomaly scores then I would apply same logic.

lmmentel:
@KatieBuc thanks for elaborated comments. Here's some of my thoughts.

## Ad 1.

Does not seem like an algorithmic problem, rather a convention. You can decide what data your implementation expects and how to deal with it. To be honest I'm not seeing the issue here since it seems that for tests you are taking input and output from the R implementation (which was more of a pipeline and not a modular estimator) and trying to reproduce it here. Why not simply test the estimator on some other data? The rest of the issues are probably best left as docs comments/example on what are the options for preparing data, i.e.
- filter out `NaN`s -> normalize -> STRAY
- normalize -> filter out `NaN`s -> STRAY
- normalize -> STRAY

I'm not sure what your recommendation would be but the options above could be examples on how STRAY can be used not necessarily tests since their scope is larger that the estimator.

## Ad 2.

Yup, I think all sklearn X's are assumed to be 2D i.e. tabular data.

## Ad 3.

It seems your example already fixed the issue of type conversion right? This seems like another good example for the docs on "how to use STRAY with the existing sktime/sklearn preprocessors".

sktime/annotation/stray.py
```diff
@@ -0,0 +1,268 @@
...
+        -------
+        self :
+            Reference to self.
+        """
+        # remember X for transform
+        self._X = X
+
+        info_dict = self._find_outliers(X)
+        self.score_ = info_dict["outlier_scores"]
+        self.y_ = info_dict["outlier_bool"]
```
aiwalter, line 234: do we really want to stor `y_` here as public attribute?

sktime/annotation/stray.py
```diff
@@ -0,0 +1,268 @@
...
+    >>> from sktime.annotation.stray import STRAY
+    >>> from sktime.datasets import load_airline
+    >>> from sklearn.preprocessing import MinMaxScaler
+    >>> import numpy as np
+    >>> X = load_airline().head(10)
+    >>> X = np.array(X).reshape(-1, 1)
+    >>> scaler = MinMaxScaler()
+    >>> X = scaler.fit_transform(X)
+    >>> model = STRAY(k=3)
+    >>> y = model.fit_transform(X)
```
aiwalter, line 76: can we put `MinMaxScaler` and `STRAY` into a `TransformerPipeline` here? Then you dont have type issues and you dont need to reshape? `MinMaxScaler` would have to be used via `TabularToSeriesAdopter`. Its best in docstring example not to do any `numpy `operations in my opinion.

aiwalter (dismissed):
left minor comments

sktime/annotation/stray.py
lmmentel: creating a pipeline seems a bit  of an overkill for a basic example, how about
```python
X = load_airline().to_frame().to_numpy()
```
That creates a column vector.

sktime/annotation/stray.py
```diff
@@ -0,0 +1,268 @@
...
+    ):
+        self.alpha = alpha
+        self.k = k
+        self.knn_algorithm = knn_algorithm
+        self.p = p
+        self.size_threshold = size_threshold
+        self.outlier_tail = outlier_tail
+        super(STRAY, self).__init__()
+
+    def _find_threshold(self, outlier_score: npt.ArrayLike, n: int):
```
lmmentel, line 106:
```suggestion
    def _find_threshold(self, outlier_score: npt.ArrayLike, n: int) -> npt.ArrayLike:
```

lmmentel (commented):
Only minor suggestions from me.

lmmentel (approved):

PR #3339, opened 2022-08-25
 [ENH] Optimize `from_3d_numpy_to_nested` converter function

paulbauriegel:
Building a nested column with all instances at once trough pd.DataFrame instead of interating over every instance improves performance by factor 3.

I was doing some inference on a big dataset 500k observations using a numpy array. The inference on big datasets is very slow. In this case one of the issues is that the numpy array was converted to a nested dataframe with `from_3d_numpy_to_nested`. And the function is very slow for bigger Dataframe sizes.

Have a look at this example:
grafik

No

The changes are rather small. You can test how the performance improves for bigger Dataframes, see my example provided above.

##### For all contributions

### DIFF
--- a/sktime/datatypes/_panel/_convert.py
+++ b/sktime/datatypes/_panel/_convert.py
@@ -910,7 +910,12 @@ def from_3d_numpy_to_nested(X, column_names=None, cells_as_numpy=False):
 
     column_list = []
     for j, column in enumerate(column_names):
-        nested_column = pd.DataFrame(X[:, j, :]).apply(lambda x: [container(x)], axis=1).str[0].rename(column)
+        nested_column = (
+            pd.DataFrame(X[:, j, :])
+            .apply(lambda x: [container(x)], axis=1)
+            .str[0]
+            .rename(column)
+        )
         column_list.append(nested_column)
     df = pd.concat(column_list, axis=1)
     return df
--- a/sktime/datatypes/_panel/_convert.py
+++ b/sktime/datatypes/_panel/_convert.py
@@ -892,6 +892,7 @@ def from_3d_numpy_to_nested(X, column_names=None, cells_as_numpy=False):
     df : pd.DataFrame
     """
     n_instances, n_columns, n_timepoints = X.shape
+    array_type = X.dtype
 
     container = np.array if cells_as_numpy else pd.Series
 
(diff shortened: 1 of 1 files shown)


fkiraly (approved):
Thank you for giving this old code some attention!
Very nice & looks good to me.

Btw, not a blocker, but would you quickly be able to describe where these conversions are happening, e.g., in what estimator or use case?
Perhaps there is a way to avoid them altogether, 1min for conversion is still a bit long.

paulbauriegel:
@fkiraly Yes of course. I don't  know all the algortihms that use the nested function, but it should be almost all that have the `_tags = {  "X_inner_mtype": "nested_univ" }`,  one example from the tutorial would be the `ColumnEnsembleClassifier` :

```python
from sktime.classification.compose import ColumnEnsembleClassifier
from sktime.classification.dictionary_based import TemporalDictionaryEnsemble
from sktime.classification.interval_based import DrCIF

clf = ColumnEnsembleClassifier(
    estimators=[
        ("DrCIF0", DrCIF(n_estimators=10), [0]),
        ("TDE3", TemporalDictionaryEnsemble(max_ensemble_size=5), [3]),
    ]
)

clf.fit(X, y)
clf.predict(X)
```

Apparently for the `ColumnEnsembleClassifier` we have the following call stack:
1. `ColumnEnsembleClassifier` inherits from `BaseClassifier`
2. `BaseClassifier` calls the function ` _check_convert_X_for_predict` within `predict`
3. ` _check_convert_X_for_predict` calls the function  `_convert_X`
    - which uses `convert_to`
    - which uses in `sktime/datatypes/_convert.py` in the `convert` function
    - `convert_dict` is exposed as `convert_dict_Panel` in `sktime/datatypes/_panel/__init__.py`
    - Line 930: `convert_dict[("numpy3D", "nested_univ", "Panel")] = from_3d_numpy_to_nested_adp`
    -  `from_3d_numpy_to_nested_adp` just calls  `from_3d_numpy_to_nested`

There are also other base classes that use the same `convert_to`, the callstack to get there is just different e.g.
- `BaseTransformer` calls the function ` _check_X_y` and `_convert_output` within `transform`

I hope that answers you question partially. In general the library seems convert almost every dataset into some other format if the algorithm data input format `X_inner_mtype` does not match the data format.

PR #3343, opened 2022-08-25
[ENH] option to keep column names in `Lag`

fkiraly:
This PR adds a parameter `keep_column_names` (default=`False`) to the `Lag` transformer, which enables `Lag` to return the same column names in `transform` as in the input. This is for avoiding to produce column names `lag_x__varname1`, `lag_x__varname2`, etc, if only one lag `x` is input.

The default is `False` and not `True` according to the reasoning in the recent https://github.com/alan-turing-institute/sktime/pull/3261 which made the defaults consistent.

The use case for not having too long variable names is in cases where we lag twice, or the lagging happens internally in another estimator, such as the reduction prototypes: https://github.com/alan-turing-institute/sktime/pull/3333

FYI, what do you think, @KishManani?


KishManani:
Looks good to me!

PR #3346, opened 2022-08-25
& ilkersigirci [BUG] ensure `Differencer` always inverts properly

fkiraly:
Fixes https://github.com/alan-turing-institute/sktime/issues/3345.
There was a strange shortcut from `_inverse_transform` of `Differencer`, which just returned data seen before if the `transform` indices had all been seen (not if some had been seen). This is highly unexpected behaviour, as @ilkersigirci pointed out.

The fix was not easy due to domino effects in the code, effectifely I rewrote the `Differencer` entirely.

The new behaviour is as in the expectation @ilkersigirci has outlined: when `inverse_transform` is called, previously seen data is taken into account when inverting, in a window before the data to be inverted that has the length of cumulative lag.
For multiple lag values, this requires logic to compute iterated differences in that window.

As a result, the `Differencer` now always properly inverts in `inverse_transform`.

This PR also adds:
* the test case of @ilkersigirci, minimally modified, so as not to include the value at the first index which is required to be outside the `fh` for proper inversion.
* a parameter `memory` for the `Differencer` which allows to control how much of the data seen in `fit`/`update` should be remembered for potential inversion.


fkiraly:
hm, this is now succeeding with @ilkersigirci´s code, but old/regression test code seems to fail. Need to investigate why. I suspect it is because of how the first (earliest) time stamp is handled.

fkiraly:
@ilkersigirci, @rnkuhns, @aiwalter, I think the differencer is finally fixed - this was not easy.

I would appreciate eyes on this especially since I basically rewrote it.

PR #3360, opened 2022-08-29
[ENH] Dictionary Classifiers Speedup (1-2 Orders of Magnitudes)

patrickzib:
This implements some major changes to dictionary classifiers:
- Major rewrite of SFA called SFA_fast using numba, vectorization and dropping dictionaries as output
- Adapt WEASEL, BOSS and CBOSS, MUSE to the new SFA_fast class
- Deprecate `typed-dict` for BOSS and CBOSS, and `alphabet_size=4` for WEASEL.

Overall, the speedup is within 1-2 orders of magnitude (see https://github.com/alan-turing-institute/sktime/pull/3360#issuecomment-1235272891 and https://github.com/alan-turing-institute/sktime/pull/3360#issuecomment-1243315668):

- Old implementation: no runtime is below minutes
- New code: all runtimes are within seconds! (Improve by 10-100x speedup)

Features missing in the new SFA-Fast-Class and will not be implemented for now:
- Spatial Pyramids (still have to figure out this)
- Support for large words (>64 bit), as this breaks numba compability
- Is "use_fallback_dft" needed?


fkiraly (changes requested):
two comments I think we need to resolve:
* does it make sense to have the "expected outputs" tests if we change the expected outputs at every change of the classifier? This PR violates the principle of changing at most one of regression tests vs the thing being tested.
* can we please not add more children of `_PanelToPanelTransformer` which is to be deprecated. Instead, let´s figure out first how we make these transformers compliant with the rest of `sktime`.

patrickzib:
> * does it make sense to have the "expected outputs" tests if we change the expected outputs at every change of the classifier? This PR violates the principle of changing at most one of regression tests vs the thing being tested.

You can't make an egg without breaking an omelet :) 

> * can we please not add more children of `_PanelToPanelTransformer` which is to be deprecated. Instead, let´s figure out first how we make these transformers compliant with the rest of `sktime`.

I will address it, once all tests pass.

> * the classifiers you are changing are reference implementations used in historical experiments, I think? So, changing these would no longer allow to reproduce results of past experiments. Should the new classifiers therefore not be new versions of these classifiers, instead of replacing the old ones?

No, the changed classifiers are re-implementations of Java Code, which I wrote in: https://github.com/patrickzib/SFA. 

In my publications, I have never used Python code. Thus, the reference implementation is Java-based. The code in sktime was always orders of magnitude slower when compared to the Java code.

I will push some plots on time and accuracy, once the benchmarks finish. But it takes time to run the old sktime code on the UCR benchmark.

MatthewMiddlehurst:
Numba Dicts are a pain with pickling. Line 720+ in _tde.py was my solution (turn it into a regular dictionary when saving, and back into a numba Dict when loading). The tests don't explode at least 🙂.

patrickzib:
Finally, all tests pass & I have included the BaseTransformer to SFA_FAST. 

@MatthewMiddlehurst @fkiraly : would be grateful for your opinion.

I will update the discussion with results of a full UCR run as soon as it finishes. But it takes quite long for some classifiers on some datasets.

fkiraly:
> does it make sense to have the "expected outputs" tests if we change the expected outputs at every change of the classifier? This PR violates the principle of changing at most one of regression tests vs the thing being tested.

If this is the expected workflow, we should just remove all "expected output" tests, as they serve no purpose. @MatthewMiddlehurst, any opinions?

MatthewMiddlehurst:
Knowing whether the outputs have changed or not is useful information. As Patrick says, he is running experiments to see the impact of this change. If you can justify the change in outputs, I don't see an issue with the expected ones being updated.

fkiraly (dismissed):
Ok, thanks, some blocking points are addressed.

Some blockers that I had overlooked:
* public parameters of estimators must not change without deprecation. Please ensure to deprecate `ContractableBOSS.typed_dict`
* description of `WEASEL.support_probabilities` is unclear and does not seem to agree with the docstring. The docstring also refers to `chi2-threshold` etc which does not appear as parameter. There are more problems of similar type.
* classes should be `UpperCamelCase` named, not true for `SFA_NEW`
* `deep_equals` should optimally not carry any additional dependencies, can we somehow avoid that?

patrickzib:
> Ok, thanks, some blocking points are addressed.
> 
> Some blockers that I had overlooked:
> 
> * public parameters of estimators must not change without deprecation. Please ensure to deprecate `ContractableBOSS.typed_dict`

I see. I may keep it and set it to deprecated. 

> * description of `WEASEL.support_probabilities` is unclear and does not seem to agree with the docstring. The docstring also refers to `chi2-threshold` etc which does not appear as parameter. There are more problems of similar type.

Sure, will look into it.

> * classes should be `UpperCamelCase` named, not true for `SFA_NEW`

Sure. Yet, SFA is an abbreviation, but I could call it SFANew, if this is better?

> * `deep_equals` should optimally not carry any additional dependencies, can we somehow avoid that?

You are referring to the import of csr_matrix in deep_equals? Unfortunately the current code breaks deep_equals with csr_matrix. Every other solution would be a hack. As sklearn also commonly supports it, it is not that uncommon. What would you suggest?

patrickzib:
> Knowing whether the outputs have changed or not is useful information. As Patrick says, he is running experiments to see the impact of this change. If you can justify the change in outputs, I don't see an issue with the expected ones being updated.

Unfortunately, I ran TDE with my benchmark, which has the highest runtimes among all. I think, we should address this in a separate PR.

fkiraly:
> Sure, will look into it. Yet, you might be referring to p_threshold, which is a parameter?

No, `chi-threshold`, which is not a parameter, but appears in the docstring. Search for it.

fkiraly:
> Sure. Yet, SFA is an abbreviation, but I could call it SFAFast, if this is better?

Better - matter of taste, I suppose. `UpperCamelCase`- yes.

fkiraly:
>You are referring to the import of csr_matrix in deep_equals? Unfortunately the current code breaks deep_equals with csr_matrix. Every other solution would be a hack. As sklearn also commonly supports it, it is not that uncommon. What would you suggest?

Well, this will upset the refactor of `BaseObject`, as it would man taking a dependency of `baseobject` on `scipy`.

Why do we need to compare csr_matrices, can you kindly elaborate? What breaks?

patrickzib:
> No, `chi2-threshold`, which is not a parameter, but appears in the docstring. Search for it.

Sure, found it and fixed it.

> Better - matter of taste, I suppose. `UpperCamelCase`- yes.

Fixed.




> Why do we need to compare csr_matrices, can you kindly elaborate? What breaks?

The problem being:

```
np.any(x != y):
[399](https://github.com/alan-turing-institute/sktime/runs/8088520358?check_suite_focus=true#step:10:400)
        ret        = <function deep_equals.<locals>.ret at 0x0000011BCF867550>
[400](https://github.com/alan-turing-institute/sktime/runs/8088520358?check_suite_focus=true#step:10:401)
        return_msg = True
[401](https://github.com/alan-turing-institute/sktime/runs/8088520358?check_suite_focus=true#step:10:402)
        x          = <10x1 sparse matrix of type '<class 'numpy.uintc'>'
[402](https://github.com/alan-turing-institute/sktime/runs/8088520358?check_suite_focus=true#step:10:403)
	with 4 stored elements in Compressed Sparse Row format>
[403](https://github.com/alan-turing-institute/sktime/runs/8088520358?check_suite_focus=true#step:10:404)
        y          = <10x1 sparse matrix of type '<class 'numpy.uintc'>'
[404](https://github.com/alan-turing-institute/sktime/runs/8088520358?check_suite_focus=true#step:10:405)
	with 4 stored elements in Compressed Sparse Row format>
```

csr_matrix does not support this comparison `np.any(x != y)` and the code fails. Instead, it has to be compared by, e.g.: `np.allclose(x.A, y.A)`

As a potential solution: We may do: `type(x).__name__ == "csr_matrix"` - this addresses the import issue. Would that be ok?

Why we need csr_matrix: SFA generates (word,count)-pairs stored in dicts beforehand. This was inefficient and had to be converted using sklearn.CountVectorizer to CSR-Matrix before further use in sklearn. Now, SFAFast simply returns a CSV-Matrix directly without conversion-operations.

patrickzib:
So here are benchmark results:

Accuracy:
- WEASEL performs more or less equal to before
- BOSS performs slightly less accurate than before

image

Runtimes:
- Runtimes improved by 1-2 orders of magnitude for fit and predict times.
- WEASEL's predict runtime is in a similar range as Rocket now, while fit is slightly slower.

image

image

image

image

I was not able to implement the asymmetric distance with BOSS using CSR-matrix, yet. I think this is where the main difference comes from. This could be done in a future PR. Still, the speedup is so significant, that I would like to use the new BOSS Implementation, as the old one is hardly usable.

patrickzib:
Total fit and predict times in hours to run all 116 ds:

image

image

E.g. fit-time drops 

- from >100 hours to ~6 for BOSS 
- from 6 hours to 0.8 hours for WEASEL.

fkiraly (changes requested):
Thanks, looks all good!

Change request (or discussion if no change is needed):
* can you kindly explain how you handle downwards compatibility?
* it still appears that you are breaking interface contracts, e.g., removal of `typed_dict` param

Any changes in the public interface require deprecation.
It seems some defaults and prior behaviour is changing, and in general this needs deprecation.

patrickzib:
> Thanks, looks all good!

Thank you. 

> Change request (or discussion if no change is needed):
> 
> * can you kindly explain how you handle downwards compatibility?

Good call. For now, the former class SFA is still there for downward compatibility. 

In the long run, I would like to move all dict classifiers to the new class SFAFast, introducing the former functionality to SFAFast. But I am not sure, if it is technically possible to move everything to numba, e.g. python has support for large datatypes (>64bits), which numba does not.

- I am working to port MUSE, too, to gain the speedup, but this still takes some time and thought.
- TDE could be ported, too. Maybe @MatthewMiddlehurst + can step in at some point.

> * it still appears that you are breaking interface contracts, e.g., removal of `typed_dict` param
> Any changes in the public interface require deprecation. It seems some defaults and prior behaviour is changing, and in general this needs deprecation.

Sorry, missed it in BOSS and CBOSS. I changed it to deprecated and included a warning now.

fkiraly:
Could you be more specific on the defaults? For all possible calls to affected classes (including defaults, i.e., parameters not passed):
* is the post-PR behaviour identical (executed algorithm)?
* are the defaults identical?

If not, there should be deprecation, unless we can strongly argue that the change substantially improves match with user expectation.

patrickzib:
> Could you be more specific on the defaults? For all possible calls to affected classes (including defaults, i.e., parameters not passed):
> 
> * is the post-PR behaviour identical (executed algorithm)?
> * are the defaults identical?
> 
> If not, there should be deprecation, unless we can strongly argue that the change substantially improves match with user expectation.

This question is tricky ;)

There is a single major change in default internal parameters: BOSS and WEASEL changed the alphabet size from 4 to 2. This reduces the memory consumption significantly. However, alphabet-size was not exposed as a class-parameter but as a local variable (with no option to set it from outside). Thus, I do not see how this should be deprecated?

fkiraly:
> Thus, I do not see how this should be deprecated?

Leave the default at 4, and add a warning that is raised whenever the value is 4.
Deprecate and change the default to 2 at 0.15.0.
Add a todo that records the change to be made in 0.15.0.

patrickzib:
> > Thus, I do not see how this should be deprecated?
> 
> Leave the default at 4, and add a warning that is raised whenever the value is 4. Deprecate and change the default to 2 at 0.15.0. Add a todo that records the change to be made in 0.15.0.

Ok. I will keep alphabet_size=2 and I have deprecated alphabet_size=4.

I would like to move in 0.14, though. We may add a warning in 0.14 that it changed to 2. 
As noted above, it was not a parameter to the class but an internal parameter set without access by the user.

patrickzib:
To further add, MUSE is ported now, too:

image

image

image


A little bit more accurate on `alphabet_size=4`, yet more than one order of magnitude faster.

patrickzib:
> > I would like to move in 0.14, though, as 4 has no advantage but a major disadvantage in memory footprint. We may add a warning in 0.14 that it changed to `alphabet_size=2`.
> 
> Sorry, we have a deprecation policy that deprecation messages must be present and functionality must be preserved for a full MINOR version cycle. https://www.sktime.org/en/stable/developer_guide/deprecation.html#deprecation-policy
> 
> This is to give users a chance to react to the change that would affect their code. If we change in 0.14.0, that might not be enough time.

I do not understand this. As the user could not choose alphabet_size, there is no reaction needed by the user. I have set a class internal parameter to a different value. It would be different if I dropped a function or a parameter, but this is not the case

fkiraly:
> How does this affect the user?

`alphabet_size` of the default changes from 4 to 2 silently, so the algorithm changes.
This should be accompanied by deprecation with a warning present for a full MINOR release cycle.

patrickzib:
> > How does this affect the user?
> 
> `alphabet_size` of the default changes from 4 to 2 silently, so the algorithm changes. This should be accompanied by deprecation with a warning present for a full MINOR release cycle.

Every bugfix changes an algorithm. Do not see the point there. Anyhow, you should update the referenced documentation, if changing a default parameter has to be done via deprecation. This is not what is stated in your cited link.

fkiraly:
> Every bugfix changes an algorithm. Do not see the point there.

Bugfix is not the same as change in functionality. There was no bug in the algorithm, you are just changing default parameters.

> Anyhow, you should update the referenced documentation, if changing a default parameter has to be done via deprecation. This is not what is stated in your cited link.

I added clarification: https://github.com/alan-turing-institute/sktime/pull/3411

The relevant definition is that of a non-downwards compatible change of public functionality or interface.

You need to imagine user code, and ask the question, does it break or does the functionality change.

patrickzib:
> > Anyhow, you should update the referenced documentation, if changing a default parameter has to be done via deprecation. This is not what is stated in your cited link.
> 
> I added clarification: #3411
> 
> The relevant definition is that of a non-downwards compatible change of public functionality or interface.
> 
> You need to imagine user code, and ask the question, does it break or does the functionality change.

Done as requested.

fkiraly (dismissed):
Thanks, changes to `BOSSEnsemble` and `ContractableBOSS` look great!

* Would appreciate the same to `IndividualBOSS` which is public, right?
* The deprecation of `alphabet_size==4` in `Weasel` seems overzealous, I would raise the warning only if it is the default, i.e., `if self.alphabet_size=4`. Code that sets `alphabet_size` to anything else will not break or change.
* docstring of `WEASEL` is still not fully clear, i.e., that chi2 p-threshold is the same as `p_threshold`

patrickzib:
> Thanks, changes to `BOSSEnsemble` and `ContractableBOSS` look great!
> 
> * Would appreciate the same to `IndividualBOSS` which is public, right?
> * The deprecation of `alphabet_size==4` in `Weasel` seems overzealous, I would raise the warning only if it is the default, i.e., `if self.alphabet_size=4`. Code that sets `alphabet_size` to anything else will not break or change.
> * docstring of `WEASEL` is still not fully clear, i.e., that chi2 p-threshold is the same as `p_threshold`

Thanks. All addressed.

fkiraly (approved):
Looks good to me now, thanks, and sorry for the repeated change requests.

PR #3362, opened 2022-08-29
[ENH] add GMMHMM from hmmlearn

miraep8:
Building on success I have had with wrapping GuassianHMM in   #3156 this is a simple PR which wraps GMMHMM from `hmmlearn`. 

I intend this to be a reference as to how some of the other estimators from `hmmlearn` can also be wrapped and incorporated into `sktime`.


miraep8:
As a note this PR depends on #3156 being merged as well.

miraep8:
Note, tests are failing, but seems to be due to a potentially sporadic failure of ClearSky?  Not sure if this is something folks have seen before or not?

fkiraly (changes requested):
Looks great!

Same comments as to #3156 apply - kindly isolate soft dependencies (see there)

fkiraly (approved):
Dependencies isolated properly now, thanks!

You forgot to re-request a review, methinks.

PR #3366, opened 2022-08-30
[BUG] fixing HMM last read bug

miraep8:
There was an issue with the way the HMM estimator was setting the last annotated segment in the list.  Thanks to @lmmentel for spotting it/letting me know.  I think this small change should fix it.  Will also add the example provided as a test shortly.


lmmentel:
:+1:  might as well fix a typo in this line https://github.com/alan-turing-institute/sktime/blob/3c56f535dc1d060c0053ab6c7f1a1a7199133539/sktime/annotation/hmm.py#L73

miraep8:
Actually just discovered there seems to be another bug, which the test helpfully revealed!  Will fix that and the typo!

miraep8:
In summary - there were actually 2 issues - one was that the last entry was actually being overwritten, and there was also an issue with all the labels being shifted by 1.  I think I have addressed both issues and have also added a sanity check (ie - does HMM work to label a very simple example as expected?)

sktime/annotation/hmm.py
```diff
@@ -374,26 +374,25 @@ def _predict(self, X):
...
-        self.states = [i for i in range(self.num_states)]
+        self.states = list(range(self.num_states))
         self.num_obs = len(X)
         emi_probs = HMM._make_emission_probs(self.emission_funcs, X)
         init_probs = self.initial_probs
-        # if no initial_probs were supplied assign all states equal prob:
         if self.initial_probs is None:
-            init_probs = 1.0 / (self.num_states) * np.ones(self.num_states)
+            init_probs = 1.0 / self.num_states * np.ones(self.num_states)
         trans_prob, trans_id = HMM._calculate_trans_mats(
```
lmmentel, line 383: FYI since you are using `@staticmethod` you can do
```suggestion
        trans_prob, trans_id = self._calculate_trans_mats(
```

lmmentel (dismissed):

lmmentel:
Nice one @miraep8 :raised_hands: I left a comment in case you want to consider changing `HMM` references to `self` but I'm happy to merge as is.

sktime/annotation/hmm.py
miraep8: Nice :+1 I am not really sure what the difference is between these two things, but seems reasonable to me.  Will look more into it, but in the meantime went ahead and make the change! :)

PR #1694, opened 2021-12-04
Add Fourier feature transformer

mloning:
#### Reference Issues/PRs
Resolves #1693 

#### What does this implement/fix? Explain your changes.
* add `FourierFeatures` transformer

#### Does your contribution introduce a new dependency? If yes, which one?
No

#### Any other comments?
Based on pmdarima implementation: https://alkaline-ml.com/pmdarima/modules/generated/pmdarima.preprocessing.FourierFeaturizer.html

##### For all contributions
- [x] I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/main/.all-contributorsrc).
- [x] Optionally, I've updated sktime's [CODEOWNERS](https://github.com/alan-turing-institute/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
- [ ] I've added unit tests and made sure they pass locally.

##### For new estimators
- [ ] I've added the estimator to the online documentation.
- [ ] I've updated the existing example notebooks or provided a new one to showcase how my estimator works.

PR #3374, opened 2022-09-02
[ENH] Fourier features transformer

### LINKED ISSUE #2650: [ENH] Fourier series components
### Discussed in https://github.com/alan-turing-institute/sktime/discussions/2649

<div type='discussions-op-text'>

<sup>Originally posted by **GrzegorzRut** May 16, 2022</sup>
Hi!
I've recently stumbled upon [this](https://robjhyndman.com/hyndsight/longseasonality/) and [this](https://medium.com/intive-developers/forecasting-time-series-with-multiple-seasonalities-using-tbats-in-python-398a00ac0e8a).

As far as I can tell, `sktime` so far does not have functions like [fourier](https://www.rdocumentation.org/packages/forecast/versions/8.16/topics/fourier). We do have the `CosineTransformer` but this can probably be extended to something like the `fourier` function mentioned above. Seems to be really helpful with for long or multiple seasonalities.

Would it be helpful to have something like that? </div>

ltsaprounis:
#### Reference Issues/PRs

Fixes #2650. See also #1694

This transformer is applied on the time index to generate Fourier series features as described here -> https://robjhyndman.com/hyndsight/longseasonality/

When fit is called the minimum date of X is stored. This date serves as the reference point for the time-steps passed, t.
So in the transform step, t is going to be the difference between the minimum date of X passed on fit (based on the frequency of the period index). 

This behaviour is checked in the test `sktime.transformers.series.tests.test_fourier.test_fit_transform_behaviour`

No

- [ ] input checks are happening on the right place
- [ ] fit and transform behaviour is the expected one. 

##### For all contributions

##### For new estimators


docs/source/api_reference/transformations.rst
```diff
@@ -393,6 +393,17 @@ Datetime feature generation
 
     DateTimeFeatures
 
+Trigonometric Features
```
fkiraly, line 396: I would say spectral features or frequency features or similar. Trigonometric is easy to confuse with the category of cosine transform etc

sktime/transformations/series/fourier.py
```diff
@@ -0,0 +1,202 @@
...
+from sktime.transformations.base import BaseTransformer
+
+
+class FourierFeatures(BaseTransformer):
+    """Fourier Features for time series seasonality.
+
+    Fourier Series terms can be used as explanatory variables for the cases of multiple
+    seasonal periods and or complex / long seasonal periods [1][2]. For every seasonal
+    period, sp and fourier term k pair there are 2 fourier terms sin_sp_k and cos_sp_k:
+    - sin_sp_k = sin(2*pi*k*t/sp)
```
fkiraly, line 23: Could you write out what precisely the series is this is transformed to?
Otherwise this might be very confusing.

fkiraly (changes requested):
Looks good!

One blocking change:
I think the docstring needs improvement, e.g., what columns and what entries will appear as a return from `transform` (optimally, with a bit of math of what the i-th entry is).

What I had expected is that the transformed series are actually the Fourier spectrum, which does not seem to be the case?

ltsaprounis:
> What I had expected is that the transformed series are actually the Fourier spectrum, which does not seem to be the case?

Nope - Fourier features from Fourier series for forecasting are the analogous to polynomial regression and Taylor series. i.e. a bunch of non-linear terms that if linearly combined can approximate a non-linear (and in the fourier case periodic) function. They give a chance to linear models (e.g. regression with ARMA) errors to capture non-linear seasonality patterns.

docs/source/api_reference/transformations.rst
ltsaprounis: I think we should stick to "Trigonometric terms" as it's the term used consistently in the literature. 

Here are 2 of the earlier papers I could find:

1. Harvey, A., Koopman, S. J. & Riani, M. (1997), ‘The modeling and seasonal adjustment of 558 weekly observations’, Journal of Business and Economic Statistics 15, 354–368.
2. Harvey, A. & Koopman, S. J. (1993), ‘Forecasting hourly electricity demand using time556 varying splines.’, Journal of the American Statistical Association 88, 1228–1236.

and here is the TBATS paper: 
1. [De Livera, A.M., Hyndman, R.J., & Snyder, R. D. (2011), Forecasting time series with complex seasonal patterns using exponential smoothing, Journal of the American Statistical Association, 106(496), 1513-1527.](https://robjhyndman.com/papers/ComplexSeasonality.pdf#page=37&zoom=100,76,825)

fkiraly: how about "seasonality and calendar features", and put it together with the date/time transformer?

ltsaprounis: OK - that makes sense

fkiraly (approved):
Thanks, brilliant! Clear explanation.

PR #3376, opened 2022-09-03
[ENH] skip `ClearSky` doctest to avoid `load_solar` crash

fkiraly:
This PR skips the `ClearSky` doctest to avoid `load_solar` crashing upon intermittent connection to the data repository.

The problem has already been addressed by using the `backoff` library in `load_solar` itself and testing, and test coverage is not reduced as `ClearSky` is still subject to the full transformer test suite, and `load_solar` is tested separately.

### DIFF
--- a/sktime/transformations/series/clear_sky.py
+++ b/sktime/transformations/series/clear_sky.py
@@ -59,12 +59,12 @@ class ClearSky(BaseTransformer):
 
     Examples
     --------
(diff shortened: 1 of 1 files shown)


patrickzib:
I would very much appreciate this, as my workflow just crashed for the very same reason!

patrickzib (approved):

PR #3379, opened 2022-09-03
[DOC] fix notebook/example symlinks

khrapovs:
Fixes https://github.com/alan-turing-institute/sktime/issues/3246 by ensuring the symlinks are part of git


fkiraly:
Looks like all is fixed, thanks!

PR #3380, opened 2022-09-04
[ENH] Add Format Specification for `.ts` files.

### LINKED ISSUE #2043: [ENH] file format specification for ts and tsf
We should have file format specifications for the `ts` and `tsf` files used in `sktime`.

The speifications can be linked (if exist elsewhere), or part of `sktime`.

Update:
from #3303, there seem to be inconsistencies in the code in referencing `@classlabels` vs `@class_labels`, and `@equallength` vs `@equalLength`.

achieveordie:
#### Reference Issues/PRs

Fixes #2043

Adding a reStructuredText file that will enable users to create their own `.ts` files, given a dataset. 
This document intends to:
- Introduce the idea of string identifiers and why are they required.
- Give an overview of different string identifiers present.
- Present a concise set of instructions for adding these identifiers in the dataset.  

- [x] Add introduction
- [x] Create an overview
- [x] Provide Instructions

Possible changes:
- [x] Restructure the overview table
- [ ] Add missing string identifiers 

##### For all contributions


docs/source/api_reference/data_format.rst
```diff
@@ -0,0 +1,248 @@
...
+    :hidden: false
+
+    Introduction
+    Purpose
+    Overview
+    Instructions
+
+Introduction
+------------
+Suppose you have comma seperated values for your dataset as:
```
TonyBagnall, line 17: separated

docs/source/api_reference/data_format.rst
```diff
@@ -0,0 +1,248 @@
...
+
+How would ``sktime`` know how to interpret the dataset (whether or not it contains multiple dimensions, whether or not
+all instances are of same length etc.)? Additionally, how would another user know about these specifications regarding the dataset?
+The answer is by having additional information at the start of file, simply by
+adding lines in a specific format at the start of the file.
+
+This document formalizes string identifiers used in `.ts` file format. Such files can be opened via any basic
+editor like notepad for visual inspection. String identifiers refer to strings beginning with ``@`` in the file.
+
+To rather understand how data is stored and loaded from these formats, visit
```
TonyBagnall, line 45: remove "rather"

docs/source/api_reference/data_format.rst
```diff
@@ -0,0 +1,248 @@
...
+    :widths: 10 40 30 20
+    :header-rows: 1
+
+    * - Identifier
+      - Description
+      - Additional Comments
+      - Example
+    * - ``@problemname``
+      - The name of the dataset.
+      - Value cannot be space seperated
```
TonyBagnall, line 95: separated

docs/source/api_reference/data_format.rst
```diff
@@ -0,0 +1,248 @@
...
+    * - ``@timestamps``
+      - Whether timestamps are present.
+      - ``true`` / ``false`` only
+      - ``false``
+    * - ``@missing``
+      - Whether there are missing values.
+      - ``true`` / ``false`` only
+      - ``false``
+    * - ``@univariate``
+      - Whether there is only one variable.
```
TonyBagnall, line 106: perhaps clarify that each time time series has one dimension

docs/source/api_reference/data_format.rst
```diff
@@ -0,0 +1,248 @@
...
+Instructions
+------------
+This section provides full set of instructions to create your format specification for your dataset that is compatible with ``sktime``.
+Remember that this begins with the assumption that you have the dataset readily available in
+expected `format <https://github.com/alan-turing-institute/sktime/blob/main/examples/loading_data.ipynb>`_.
+
+Few points to keep in mind while creating the dataset:
+
+  1. The general order of identifiers **does not** matter with the exception that ``@data`` should be the last string identifier.
+  2. For lines containing an identifier, it **must** begin with it.
```
TonyBagnall, line 143: lines containing an identifier **must** begin with that identifier.

docs/source/api_reference/data_format.rst
```diff
@@ -0,0 +1,248 @@
...
+  3. The **only** place to give a space is between identifier and its corresponding value.
```
TonyBagnall, line 144: The **only** place a space is allowed is between an identifier and its corresponding value.

TonyBagnall (changes requested):
thanks for this, I've spotted a couple of spelling mistakes. I'm not sure how this is translated to docs/source, but I cannot see some of the information in data_format in api_reference (e.g. the data line 23

achieveordie:
> thanks for this, I've spotted a couple of spelling mistakes. I'm not sure how this is translated to docs/source, but I cannot see some of the information in data_format in api_reference (e.g. the data line 23

Thanks for pointing those out, I've changed them and have also reformated a bit. I'm not sure what information you cannot see but please check if that is still an issue.

docs/source/api_reference/data_format.rst
```diff
@@ -0,0 +1,273 @@
...
+Introduction
+------------
+This document formalizes string identifiers used in ``.ts`` file format.
+Encoded in ``utf-8``, ``.ts`` files stores time-series dataset and its corresponding metadata and can be
+opened via any basic editor like notepad for visual inspection.
+String identifiers refer to strings beginning with ``@`` in the file.
+
+``.ts`` files contains information blocks in the following order:
+
+1. All the comments to describe the dataset. (begins with ``#`` in the file)
```
fkiraly, line 34: I would say "a description block. All lines in this block begin witih ``#``".

In all the blocks, I would also link to a more detailed description of the block, to be given below. E.g., "a description block. All lines in this block begin witih ``#``. For detail description, see section (reference to section"

docs/source/api_reference/data_format.rst
```diff
@@ -0,0 +1,273 @@
...
+2. All the String Identifiers to give information about the metadata (begins with ``@`` in the file).
```
fkiraly, line 35: "a metadata block. (now describe the block high-level) (and reference the detail description of the block"

docs/source/api_reference/data_format.rst
```diff
@@ -0,0 +1,273 @@
...
+3. Actual dataset.
```
fkiraly, line 36: note enough detail. "actual dataset" is not a concrete term. Only use concrete words and formulations. The meaning of these words is unclear.

docs/source/api_reference/data_format.rst
```diff
@@ -0,0 +1,273 @@
...
+
+See `Basic Motion.ts`_ for a concrete example.
```
fkiraly, line 38: I would link the file, but perhaps giving extracts directly here is better.

I.e., show the file with some annotation on what the blocks are. Then, ensure to link to the sections that give detail descriptions.

docs/source/api_reference/data_format.rst
```diff
@@ -0,0 +1,273 @@
...
+
+These identifiers serve the purpose of containing metadata for the dataset.
```
fkiraly, line 40: "these" refers to content above, but there are no identifiers above.

You also need to explain the meaning of "identifier".

Also, it is unclear why you are talking about them now. Should this not be in a section about the metadata block? Which should be preceded by a subsection about the description block?

fkiraly (changes requested):
earlier PR: https://github.com/alan-turing-institute/sktime/pull/3361

This is good and useful, thanks!

Suggestions:
* think about the order, the file should be structured top-down in detail, and sequentially in terms of file structure
* so, I would suggest to explain the three blocks in order, in a later detail part with subsection per block. And precede it by an overview that links to the detail explanation of the block.
* ensure you provide enough high-level information to make the context clear of everything you explain. For instance, you explain the order and types of metadata identifiers, without explaining clearly where they arise, or where they can/should be located (in the metadata block, and that block consists of identifier/value pairs in different lines, I presume).
* ensure you use only words with meaning that is clear at the point you use them. Examples of vague formulations or words that are undefined when they appear: "actual dataset" (see above), or "identifier" (when the word first appears)

docs/source/api_reference/data_format.rst
achieveordie: Are you suggesting adding the contents from that file into the doc? I think that will needlessly increase the size, especially since we already have an illustration. I could just add a cross-ref to the illustration.

docs/source/api_reference/data_format.rst
achieveordie: Do you have another term? "A dataset block" sounds equally confusing to me and I can't think of any other way of putting it.

achieveordie: Like other blocks, this will also require a small description. I can't think of anything since I find it pretty self-explanatory. Also, what do we write in the detailed subsection for 'Dataset block'? This document is only meant to describe the first two blocks; adding information on how the dataset is supposed to be present seems like an unnecessary addition to me.

Not having a subsection for the last block is also weirdly asymmetric.

docs/source/api_reference/data_format.rst
achieveordie: How about:
```
All lines in this block begin with # and provides citations and comments about the dataset. See [:ref:`subsection on description block <comment description>`](https://github.com/alan-turing-institute/sktime/pull/3380#id1) for more details.
```

What all things do we add in the `comment description` subsection? I've mentioned that it's optional and is used for comments and citations. I don't know if we need to write more.

docs/source/api_reference/data_format.rst
achieveordie: How about:
```
2. A metadata block.
      All lines in this block begin with ``@`` and each line has one string identifier and corresponding value.
      See :ref:`subsection on metadata block <metadata description>` for more details.
```

docs/source/api_reference/data_format.rst
achieveordie: You're right, I've moved this under the "Metadata Block" subsection under "Description".

docs/source/api_reference/data_format.rst
fkiraly: the instruction is vague, e.g., "all the comments", this refers to "the" comments, as if the reader should know what comments to put there. Vague descriptions are not good in a tech documment.

Alt suggestion:
```
1. description block
the description contains any number of (contiguous) lines starting with `#`, each `#` followed by an arbitrary (utf-8) sequence of symbols. The `ts` specification does not prescribe any content for the description block, but it is common to include a description of the data set contained, e.g., a full data dictionary.
```

docs/source/api_reference/data_format.rst
fkiraly: better, but it does not fully spell out how lines look like.

docs/source/api_reference/data_format.rst
fkiraly: > since I find it pretty self-explanatory

It's not self-explanatory. There is no clear reference frame, sorry.

Try to make it clearer what the block contains.

docs/source/api_reference/data_format.rst
fkiraly: I would just show extracts, not the entire file. Or move the reference to later, so the reaser is not lead to another file too early.

achieveordie:
@fkiraly I've written a more verbose introduction section that would help in the "top-down" approach mentioned. I hope that I've gotten away with all vague words as well, but if I missed any or if I need to make more specific changes, do let me know.

TonyBagnall (dismissed):

fkiraly (changes requested):
Excellent! Really great, precision has much improved!

I have two requests for the identifier table which is crucial:
* there should be a column "values" that describes the valid values, e.g., integers >0
* there should be a legend that explains the differen columns

And one request regarding the place in the documentation:
if you go through the documentation, the page is linked as, and has header "file format specifications". A person looking at the table of contents does not know what file format. I would put an intermediate page where all the file formats are listed (currently there is just one), and name the page `ts file format specification`

achieveordie:
@fkiraly I was thinking about updating the structure as well, thanks. I've added legends to the table as well.

TonyBagnall (approved):
looks good. Generally, I would put a space between text and brackets , e.g. simplest case(when -> simplest case (when, and the refs do not render correctly in "display the rich diff" view on github, e.g. See :ref:`subsection on metadata block <metadata description>`, but that may just be the view.  Also the code blocks do not show on this view, but again thats probably just a github thing.

achieveordie:
It seems like there's no perfect editor to render `.rst`. GitHub and [online editors](https://livesphinx.herokuapp.com/) don't render code blocks. PyCharm seems to be rendering them fine though. References also work correctly on the aforementioned editor (doesn't render nicely in PyCharm).

Added some spaces too, it'll be awful if I missed any now.

TonyBagnall (approved):

fkiraly (approved):
Really nice! Thanks!

Issue #1497, opened 2021-10-07
[ENH] `get_fitted_params` umbrella issue


fkiraly:
Collecting some issues related to implementing meaningful `get_fitted_params` methods.

General framework:

- [x] nested interface like `get_params` #971

Individual estimators:

- [x] (Auto)ARIMA #962

For composite forecasters:

- [x] tuning forecasters #1496 
- [x] reduction forecasters #1486 
- [x] forecasting pipelines #1468 

FYI @boukepostma, @asksteph, @aiwalter 

Sounds like an interesting project?

fkiraly:
all of the items on the list are implemented in merged PR, this is now done.

PR #3381, opened 2022-09-04
[ENH] enable default `get_fitted_params` for forecasters and delegated estimators

fkiraly:
This PR enables a default implementation of `get_fitted_params` for forecasters and delegated estimators:

* for forecasting, `BaseForecaster` accidentally overrode the recently introduced default implementation of `get_fitted_params`. The abstract method has been removed from `BaseForecaster` now, which should enable the default in all forecasters. (all other estimator types were unaffected)
* for delegators, `_get_fitted_params` now delegates to the delegate's `get_fitted_params`, in line with all other delegation defaults.
* the default `get_fitted_params` is robustified so that it queries fitted parameters (recursion) only from fitted components
* a conditional override is added to `BaseForecaster` and `BaseTransformer` for the vectorized case (to avoid unfitted estimators present in underscore attributes)

Also see https://github.com/alan-turing-institute/sktime/issues/1497.

This should solve #1497 for all compositors that are delegates.

Relies on:
* #3391 for solving vectorization/`get_fitted_params` clashes
* #3393 for solving vectorization/`get_fitted_params` clashes


fkiraly:
Note to self: works, but the `BaseForecaster` override masked a bug where vectorization and the default interact.
In case of vectorization, components may remain uninitialized/unfitted, and the vectorized components do not get returned via `get_fitted_params`.

PR #3382, opened 2022-09-04
[ENH] direct reducer prototype rework based on feedback

fkiraly:
Partial rework of the direct reducer prototype based on discussion in https://github.com/alan-turing-institute/sktime/issues/3224

* factors out the creation of the feature matrix into a transformer `ReducerTransform` that lives with `Lag`. Internally, this uses `FeatureUnion`, `Lag`, `Impute`, and `YtoX` to avoid multiplication of functionality.
* ensure that case with no lagged `y` variables and no exogeneous `X` or both are properly treated
* allow for passing arbitrary transformers to apply to `y`
* adds a `subset_index` option to `YtoX` that I did not end up needing


fkiraly:
this is ready for review now.

FYI @danbartl, @topher-lo, this includes a `ReducerTransform` that creates the `X` for applying the sklearn estimator.

The problem with the longer runtime remains, still not a good grip where it is coming from. At least, this PR does not make it worse compared to the current prototype.

PR #3383, opened 2022-09-04
[BUG] Fix default conformal intervals initial_window parameter

### LINKED ISSUE #2889: [BUG] ConformalIntervals wrapped ARIMA/AutoARIMA breaks with default parameters
### Discussed in https://github.com/alan-turing-institute/sktime/discussions/2869

<div type='discussions-op-text'>

<sup>Originally posted by **linaamic** June 26, 2022</sup>
Hi everyone, I am currently trying out the ConformalIntervals feature from sktime.forecasting. However, I can't get it to work for both ARIMA and AutoARIMA. Is there an inherent difference between ARIMA forecasters and others? Do I have to call or wrap them differently or is ConformalPrediction not possible with ARIMA?
With other forecasters I have no issues.

This is the error I get:
File "...\site-packages\statsmodels\tsa\statespace\sarimax.py", line 843, in _conditional_sum_squares
    Y = endog[r:]
IndexError: too many indices for array: array is 0-dimensional, but 1 were indexed

Code:
```python
from sktime.forecasting.arima import AutoARIMA, ARIMA
from sktime.forecasting.conformal import ConformalIntervals

y = load_airline()
trial_forecaster = AutoARIMA()
conformal_forecaster = ConformalIntervals(trial_forecaster, method="conformal")
conformal_forecaster.fit(y, fh=[1,2,3])
```

Any help is much appreciated! </div>

bethrice44:
#### Reference Issues/PRs
Fixes #2889 

Sets default `initial_window` to either 10 or 10% of the data (whichever is higher) and allows either an integer or a float to be passed

Is the choice of default the right one? I'm not sure we can have a foolproof default but this should solve most cases.

##### For all contributions

##### For new estimators


fkiraly (approved):
Looks good, thanks!

I think the default (albeit a little arbitrary) makes sense, as it scales with the total amount of data.

From a cleanness of code perspective, I would split the parsing of the `initial_window` parameter from the sliding residuals function.
One of the options here is to leave the input to be an `int`, and translate to an integer before - i.e., a function that takes `initial_window` and the data, and returns what is now `n_initial_window`.

But, it's not a blocker.

fkiraly (changes requested):
Actually, it breaks due to a comparison between float and a date - not sure why.

Could you kindly test this using `check_estimator` and debug?

fkiraly:
I think the problem might be that the fraction computation is not correct in the case of datetimes.

fkiraly (approved):
Looks good now!

I would recommend writing a proper docstring for `parse_initial_window`, but I do not consider this a blocker because it is private and the calling method indirectly documents it.

PR #3386, opened 2022-09-06
[ENH] Add TapNet DL Model for Classification

### LINKED ISSUE #3372: [ENH] Adding TapNet DL Model for Classification
Migrating TapNet Classifier model from sktime-dl to sktime.

achieveordie:
#### Reference Issues/PRs

Fixes #3372

Migrate TapNet Network and Classifier from sktime-dl into sktime. Changes are kept to a minimum. Remove a few commented blocks of code. Ensure the addition of soft-dependency is handled appropriately.

*Yes*, introduces `keras-self-attention` as a soft dependency.

- [x] Complete `tapnet` network
- [x] Complete `tapnet` classifier
- [x] Tested `check_estimator()` for `tapnet` network.
- [ ] Tested `check_estimator()` for `tapnet` classifier. (currently failing)

#### Any other comments?

Currently, working on rectifying the problem and will ask to review the code when it is working locally.

##### For all contributions

##### For new estimators


achieveordie:
Apart from the ongoing discussion on saving/loading of DL models (issue #3022). The reason current PR fails to even save from `self.model_.save()` is due to `keras.layers.Lambda()` which is incompatible with [(de)serialization](https://keras.io/api/layers/core_layers/lambda/#lambda-class) attempts.

I could subclass a layer instead of using a `Lambda` to avoid this issue(this is the suggested way as per the previous link).

@fkiraly If this seems like a reasonable change, let me know so I could start working on it.

fkiraly:
Currently, the two failing tests are  being skipped for all DL models until a solution for the save/load issue is found.

You can add the network to the skip list in `tests._config`.

fkiraly (changes requested):
Quick question: how common is it to install `keras-self-attention` if you already have `tensorflow` or `keras`? Should this be a core dl dependency or just a soft dependency?

achieveordie:
> Quick question: how common is it to install `keras-self-attention` if you already have `tensorflow` or `keras`? Should this be a core dl dependency or just a soft dependency?

I'm aware there are other networks that make use of `keras-self-attention` in `sktime`, but I believe we do not need to have it as a core dl dependency yet. We could add it as so if we have more models that make use of it in the future. Personally, I've never used this myself and have seen it only a few times before this.

fkiraly:
ok, let's make it a soft dependency then

fkiraly (changes requested):
Thanks!

Almost ready to go.

I think we need to:
* add a check for the soft dependency at the top of the module (warning level)
* add another check in the init (error level)

Have a look at `MatrixProfileTransformer` as an example.

achieveordie:
I see I was missing a warning-level soft dependency check for the network. I've added that, I think it should be good now.

On a side note [`MatrixProfileTransformer`](https://github.com/sktime/sktime/blob/main/sktime/transformations/series/matrix_profile.py) has no explicit error-level check in the constructor. Should it be this way?

fkiraly:
> On a side note [MatrixProfileTransformer](https://github.com/sktime/sktime/blob/main/sktime/transformations/series/matrix_profile.py) has no explicit error-level check in the constructor. Should it be this way?

Yes, it's done by the `python_dependencies` tag and `super.__init__`.

fkiraly (approved):
All good now, thanks!

Issue #3394, opened 2022-09-08
[BUG] sporadic bug in `test_all_annotators.py::test_output_type[HMM]`


fkiraly:
There seems to be a sporadic bug with `test_all_annotators.py::test_output_type[HMM]`:

```
         if np.any(np.isinf(trans_prob[:, -1])):
>           raise ValueError("Change parameters, the distribution doesn't work")
E           ValueError: Change parameters, the distribution doesn't work
```

see here: https://github.com/alan-turing-institute/sktime/runs/8257610342?check_suite_focus=true

FYI @miraep8, @lmmentel

miraep8:
Thank you for raising issue!

PR #3396, opened 2022-09-08
[ENH] Fix for hmm sporadic test failure

miraep8:
Thank you to @fkiraly for bringing this error to my attention in #3394 

I believe the issues was arising from potential randomness in the generation of fake data for the testing of output types for the estimator.  Depending on the data set that was generated, I believe the probability of seeing that data was accidentally running into a -Inf.  This is behavior I should fix in a more elegant manner (ie by detecting that we are approaching a -Inf and either recalibrating or using a modulo solution), but for now I believe I have fixed the error by changing two things:

- Increasing the sd of the emission functions for my test parameters.  This should make a range of values far more likely than what we saw before, helping the tests to avoid the region where it encountered errors
- changing the value error to a warning instead of an error. 

Happy to merge @fkiraly's PR (#3395) which simply skips this test for the meantime and we can think if we need more to solve this problem or not, but I think these two changes should solve it!

### DIFF
--- a/sktime/annotation/hmm.py
+++ b/sktime/annotation/hmm.py
@@ -6,6 +6,7 @@ Implements a basic Hidden Markov Model (HMM) as an annotation estimator.
 To read more about the algorithm, check out the `HMM wikipedia page
 <https://en.wikipedia.org/wiki/Hidden_Markov_model>`_.
 """
+import warnings
 from typing import Tuple
 
 import numpy as np
@@ -259,7 +260,7 @@ class HMM(BaseSeriesAnnotator):
             trans_prob[:, i] = np.max(paths, axis=0)
 
         if np.any(np.isinf(trans_prob[:, -1])):
-            raise ValueError("Change parameters, the distribution doesn't work")
+            warnings.warn("Change parameters, the distribution doesn't work")
 
         return trans_prob, trans_id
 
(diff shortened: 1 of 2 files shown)


fkiraly:
could we work with a random seed to ensure the error case does not appear at all? Would this not make it just less likely, but leave it sporadic?

fkiraly (changes requested):
Thanks!
looks like it addresses the problem.

Change requests:
* could we/you confirm that the fix does not just reduce the probability of the bug, but altogether eliminates it? I.e., probability of occurrence should be zero.
* in this PR, can you kindly revert my temporary fix https://github.com/alan-turing-institute/sktime/pull/3395 so we run the test again on HMM?

miraep8:
Addressing these two points:

- Just made a commit that *should* restart the tests for HMM
- I have run some tests that convince me that this error shouldn't cause us any issues going forward.  Details in the dropdown below:

<details>
<summary> Tests on sporadic bug fix </summary>
<br>

Two changes to test here:
-1 how often do we see this bug with the new parameters (note I switched the sd of the emission distribution from .25 which it was at before to 100).
    - to test this I simply re-created the test that was failing and counted the number of failures with the old parameters vs the new parameters over 100,000 simulations.  We end up seeing about 3% error rate with the parameters we had before (which explains why it might not have come up in my CI/CD but came up in others shortly after!), and no errors with the new parameters:
 
image

- 2 Second thing to test is what effect changing the ValueError to a warning has on handling errors when they do arise.  
     - When we run the above test with that change (making no other changes) we see that there are now 0 failures in both the new and the old parameter cases. 

image

Together these two observations convince me that the new parameter setting have made this failure mode practically impossible, and in the unlikely event this failure mode did arise it would now no longer cause the test to fail.
</details>

fkiraly (approved):
happy with that, thanks!

PR #3399, opened 2022-09-09
[ENH] Add Information Gain Temporal Segmentation Estimator

lmmentel:

Information Gain-based Temporal Segmentation.

Information Gain Temporal Segmentation (IGTS) is a method for segmenting
multivariate time series based off reducing the entropy in each segment [1]_.

The amount of entropy lost by the segmentations made is called the Information
Gain (IG). The aim is to find the segmentations that have the maximum information
gain for any number of segmentations.

References
----------
.. [1] Sadri, Amin, Yongli Ren, and Flora D. Salim.
    "Information gain-based metric for recognizing transitions in human activities.",
    Pervasive and Mobile Computing, 38, 92-109, (2017).
    https://www.sciencedirect.com/science/article/abs/pii/S1574119217300081

NO.

##### For all contributions

##### For new estimators


lmmentel:
FYI @miraep8, @KatieBuc here is my take on implementing IGTS. Let me know if you have any thoughts or suggestions.

lmmentel:
@fkiraly some test automation breaks with a logging statement that has an f-string (I think). Would be worth for you to ave a look at https://github.com/alan-turing-institute/sktime/actions/runs/3074385309/jobs/4967169980

fkiraly (changes requested):
Could you kindly explain why the base tests break?
Would be great if you could look into it, what are the error messages coming from the test suite?

lmmentel:
All the test are passing, I would be grateful if you could review the PR if you can so that we can move forward with it. 

I have since removed the offending logging statement. I have not investigated why it was failing tests but the error was coming from an `eval` statement.

fkiraly:
> All the test are passing, I would be grateful if you could review the PR if you can so that we can move forward with it.

@lmmentel, I don't think that's a true statement.

You've added the following skips/excepts for tests in the test config file. Either these tests are failing, or the skips are unnecessary.

I an kindly asking you to explain these - these are tests of sklearn and test framework compatibility that every sktime estimator should pass.

```python
    "InformationGainSegmentation": [
        "test_constructor",
        "test_inheritance",
        "test_create_test_instance",
    ],
```

lmmentel:
These test are a bit too constraining given that they test implementation rather than functionality. This estimator is compatible with sklearn interface.

fkiraly:
> These test are a bit too constraining given that they test implementation rather than functionality. This estimator is compatible with sklearn interface.

Apologies, that's not a very clear statement to me, can you please explain?

The tests above test specific interface assumptions required for `sklearn` compatibility, failing them will typically make your estimators not `sklearn` compatible.

We need to be `sklearn` compatible even if we otherwise go with a custom interface.

fkiraly:
I've added some explanations to the respective tests here:
https://github.com/sktime/sktime/pull/3555

to explain why these tests are important and how to comply with them.
Hope that is useful?

lmmentel:
Not sure what you're asking for here. Are you saying that those tests have to included and pass or you want some more explanations from me?

lmmentel:
FYI: similar tests are skipped for GGS https://github.com/sktime/sktime/blob/13239f38ec9ac8157f70aa2c70487c7978e38ac9/sktime/tests/_config.py#L110

fkiraly:
> Not sure what you're asking for here. Are you saying that those tests have to included and pass or you want some more explanations from me?

Yes, at least one of the two, please.
GGS passes `test_constructor` which is required for much of `sklearn` compatibility, so it's less "bad" than this.

Preferably, all tests pass, although I'm happy with skipping `test_inheritance` until the base class question is solved.

sktime/annotation/igts.py
```diff
@@ -0,0 +1,486 @@
...
+Gain (IG). The aim is to find the segmentations that have the maximum information
+gain for any number of segmentations.
+
+References
+----------
+.. [1] Sadri, Amin, Yongli Ren, and Flora D. Salim.
+    "Information gain-based metric for recognizing transitions in human activities.",
+    Pervasive and Mobile Computing, 38, 92-109, (2017).
+    https://www.sciencedirect.com/science/article/abs/pii/S1574119217300081
+
```
aiwalter, line 18: this should go into class docstring if not there already

sktime/annotation/igts.py
```diff
@@ -0,0 +1,486 @@
...
+"""
+
+# import logging
```
aiwalter, line 21:
```suggestion
```

sktime/annotation/igts.py
```diff
@@ -0,0 +1,486 @@
...
+import pandas as pd
+from attrs import asdict, define, field
+
+from sktime.base import BaseEstimator
+
+__all__ = ["InformationGainSegmentation"]
+__author__ = ["lmmentel"]
+
+
+# logger = logging.getLogger(__name__)
```
aiwalter, line 36:
```suggestion
```

sktime/annotation/igts.py
```diff
@@ -0,0 +1,486 @@
...
+            )
+
+            # logger.info(
+            #     f"BEST {ig_max=:.5f} for {best_candidate=} "
+            #     + f"current {current_change_points=}"
+            # )
+        return current_change_points
+
+
+class SegmentationMixin:
```
aiwalter, line 261: should this not go into another file? Do you think its useful to be used by other algos?

sktime/annotation/igts.py
```diff
@@ -0,0 +1,486 @@
...
+       IGTS does not work very well for univariate series but it can still be
+       used if the original univariate series are augmented by an extra feature
+       dimensions. A technique proposed in the paper [1]_ us to subtract the
+       series from it's largest element and append to the series.
+
+    Parameters
+    ----------
+    k_max: int, default=10
+        Maximum number of change points to find. The number of segments is thus k+1.
+
```
aiwalter, line 319:
```suggestion
```

sktime/annotation/igts.py
```diff
@@ -0,0 +1,486 @@
...
+    References
+    ----------
+    .. [1] Sadri, Amin, Yongli Ren, and Flora D. Salim.
+       "Information gain-based metric for recognizing transitions in human activities.",
+       Pervasive and Mobile Computing, 38, 92-109, (2017).
+       https://www.sciencedirect.com/science/article/abs/pii/S1574119217300081
+
+    Example
+    -------
```
aiwalter, line 349:
```suggestion
    Examples
    --------
```

sktime/annotation/igts.py
```diff
@@ -0,0 +1,486 @@
...
+    >>> from sktime.annotation.datagen import piecewise_normal_multivariate
+    >>> from sklearn.preprocessing import MinMaxScaler
+    >>> X = piecewise_normal_multivariate(lengths=[10, 10, 10, 10],
+    ... means=[[0.0, 1.0], [11.0, 10.0], [5.0, 3.0], [2.0, 2.0]],
+    ... variances=0.5)
```
aiwalter, line 354:
```suggestion
    >>> X = piecewise_normal_multivariate(
    ... lengths=[10, 10, 10, 10],
    ... means=[[0.0, 1.0], [11.0, 10.0], [5.0, 3.0], [2.0, 2.0]],
    ... variances=0.5,
    ... )
```

sktime/annotation/igts.py
```diff
@@ -0,0 +1,486 @@
...
+    >>> X_scaled = MinMaxScaler(feature_range=(0, 1)).fit_transform(X)
+    >>> from sktime.annotation.igts import InformationGainSegmentation
```
aiwalter, line 356: would put all imports at top

sktime/annotation/igts.py
```diff
@@ -0,0 +1,486 @@
...
+            Placeholder for compatibility with sklearn-api, not used, default=None.
+
+        Returns
+        -------
+        y_pred : array_like
+            1D array with predicted segmentation of the same size as the first
+            dimension of X. The numerical values represent distinct segments
+            labels for each of the data points.
+        """
+        return self.fit(X=X, y=y).predict(X=X, y=y)
```
aiwalter, line 434: this should go and we can inherit from `from sklearn.base import BaseEstimator` which is implementing that?

sktime/annotation/tests/test_igts.py
```diff
@@ -0,0 +1,55 @@
+# -*- coding: utf-8 -*-
+"""Tests for igts module."""
+
+import numpy as np
+import pytest
+
+from sktime.annotation.igts import IGTS, InformationGainSegmentation, entropy
```
aiwalter, line 7: I am wondering why you can import `IGTS` and `entropy` here bcz you have in that file `__all__ = ["InformationGainSegmentation"]
`?

sktime/annotation/igts.py
```diff
@@ -0,0 +1,486 @@
...
+            self.intermediate_results_.append(
+                ChangePointResult(
+                    k=k, score=ig_max, change_points=current_change_points.copy()
+                )
+            )
+
+            # logger.info(
+            #     f"BEST {ig_max=:.5f} for {best_candidate=} "
+            #     + f"current {current_change_points=}"
+            # )
```
aiwalter, line 257:
```suggestion
```

sktime/annotation/igts.py
```diff
@@ -0,0 +1,486 @@
...
+
+        for k in range(self.k_max):
+            ig_max = 0
+            # find a point which maximizes score
+            for candidate in self.get_candidates(n_samples, current_change_points):
+                try_change_points = {candidate}
+                try_change_points.update(current_change_points)
+                try_change_points = sorted(try_change_points)
+                ig = self.information_gain_score(X, try_change_points)
+                # logger.info(f"{ig=:.5f} for {candidate=} with {try_change_points=}")
```
aiwalter, line 241:
```suggestion
```

aiwalter (commented):
left some comments.

sktime/annotation/igts.py
lmmentel: Yes, that would be ultimate goal, however I would like to keep this PR small, and work on the changes in following PRs.

sktime/annotation/igts.py
lmmentel: Right now `BaseEstimator` does not have this method https://github.com/sktime/sktime/blob/main/sktime/base/_base.py and it's required to pass the tests.

sktime/annotation/tests/test_igts.py
lmmentel: `__all__` is only relevant when you do `from foo import *` but here I'm importing directly by names so it has no effect

sktime/annotation/igts.py
lmmentel: It is also present in the class docstrings

sktime/annotation/igts.py
lmmentel: I think it's more readable with a line break

sktime/annotation/igts.py
lmmentel: Thanks, but I hope this is non blocking

lmmentel:
`test_constructor` tests if you are exactly doing the `self.ags=arg` in ` __init__` which is testing exactly **one** way of achieving that result in that sense it is testing the implementation syntax and not function and therefore should probably be dropped or allowed to skip. I'm setting the attributes with attrs which in my opinion is superior and cleaner.

fkiraly:
> `test_constructor` tests if you are exactly doing the `self.arg=arg` in ` __init__` 

I don't think that's accurate:

* the test is not looking at the code but the result.
* the tests is testing more, e.g., signature expectations. See the improved docstring in #3555 for an explanation.

> I'm setting the attributes with attrs which in my opinion is superior and cleaner.

Yes, but I believe that breaks the signature expectations that `sklearn`/`sktime` have which are also tested, which is why the test fails.

It may be superior or cleaner but it seems to break `sklearn` compatibility.

lmmentel:
> > `test_constructor` tests if you are exactly doing the `self.arg=arg` in ` __init__`
> 
> I don't think that's accurate:
> 
>     * the test is not looking at the code but the result.
> 
>     * the tests is testing more, e.g., signature expectations. See the improved docstring in [[ENH] improve base class test docstrings and clone test #3555](https://github.com/sktime/sktime/pull/3555) for an explanation.
> 
> 
> > I'm setting the attributes with attrs which in my opinion is superior and cleaner.
> 
> Yes, but I believe that breaks the signature expectations that `sklearn`/`sktime` have which are also tested, which is why the test fails.
> 
> It may be superior or cleaner but it seems to break `sklearn` compatibility.

Nope, it only breaks the test because how the test it structured.

sktime/annotation/igts.py
```diff
@@ -0,0 +1,481 @@
...
+import pandas as pd
+from attrs import asdict, define, field
+
+from sktime.base import BaseEstimator
+
+__all__ = ["InformationGainSegmentation"]
+__author__ = ["lmmentel"]
+
+
+
```
lmmentel, line 35:
```suggestion
```

fkiraly:
> Nope, it only breaks the test because how the test it structured.

That's contradicting me without explanation/reasoning. I gave you my explanation and reasoning, for a productive discussion we should exchange arguments.

Here's the code, verbatim, that checks init attributes:

```python
        # Ensure that each parameter is set in init
        init_params = _get_args(type(estimator).__init__)
        invalid_attr = set(init_params) - set(vars(estimator)) - {"self"}
```

this looks like it is checking the results to me, against signature conventions.

These are also three lines in a test that is ca 70 lines long, so the implicit claim above that this is all the test does is also incorrect (you can look up that the 67 other lines are contentful).

lmmentel:
You haven't mentioned how exactly my implementation breaks sktime/sklearn compatibility so I'm not sure why would you accuse me of making this discussion unproductive.

fkiraly:
>  so I'm not sure why would you accuse me of making this discussion unproductive.

apologies if I wasn't clear here, I'm not accusing you of anything.

> You haven't mentioned how exactly my implementation breaks sktime/sklearn compatibility

I don't think that's correct, self-citing myself above:

> Yes, but I believe that breaks the signature expectations that `sklearn`/`sktime` have which are also tested, which is why the test fails.

> The tests above test specific interface assumptions required for `sklearn` compatibility, failing them will typically make your estimators not `sklearn` compatible.

> I've added some explanations to the respective tests here:
https://github.com/sktime/sktime/pull/3555
to explain why these tests are important and how to comply with them.
Hope that is useful?

lmmentel:
Writing the words "specific assumptions" does not reveal much about what you have in mind. Also this seems like a nitpick. Without a specific example of what assumption is violated this discussion is pointless.

fkiraly:
well, that's precisely why I made PR #3555 and mentioned it in the discussion above multiple times, trying to explain that it contains explanations.

But for your convenience I'm happy to copy-paste the relevant parts:

```
        """Check that the constructor has sklearn compatible signature and behaviour.

        Based on sklearn check_estimator testing of __init__ logic.
        Uses create_test_instance to create an instance.
        Assumes test_create_test_instance has passed and certified create_test_instance.

        Tests that:
        * constructor has now varargs
        * tests that constructor constructs an instance of the class
        * tests that all parameters are set in init to an attribute of the same name
        * tests that parameter values are always copied to the attribute and not changed
        * tests that default parameters are one of the following:
            None, str, int, float, bool, tuple, function, joblib memory, numpy primitive
            (other type parameters should be None, default handling should be by writing
            the default to attribute of a different name, e.g., my_param_ not my_param)
        """
```

fkiraly:
> Without a specific example of what assumption is violated this discussion is pointless.

What kind of specific example do you want? The new class fails the test. How does this need an example, besids the new class?

miraep8:
Hey all I think I actually ran into a similar issue with the init check when using `define`/`attrs`.

With `attrs` (and I think with what you are doing here @lmmentel), vars seems not to work properly. I think this has something to do with it not playing well with the `__dict__` function that `vars` seems to rely on. However the attributes are correctly added as expected during  `__init__`, (which is captured by using `dir` instead), and I think is what we are really interested in???

`vars` and `dir` are  of course slightly different functions though, as is summarized a bit [here](https://stackoverflow.com/questions/980249/difference-between-dir-and-vars-keys-in-python). 

Basically I was able to get the tests to pass when I made the following change to the init tests:
```
init_params = _get_args(type(estimator).__init__)
invalid_attr = set(init_params) - set(vars(estimator)) - {"self"}
```

change to:

```
init_params = _get_args(type(estimator).__init__)
invalid_attr = set(init_params) - set(dir(estimator)) - {"self"}
```

I ended up not implementing this change in part because @fkiraly was not a fan of the change to the sktime interface tests, but I think since it has come up twice now I think it is worth re-evaluating if making this small change (ie `vars` -> `dir`) is consistent with what the test is looking for.  I personally think it is, and I also think it would fix this problem?  Thoughts?

miraep8 (commented):
Just adding a few random comments based on quick impressions.  I may return with others later, overall impressions are that it looks good though and my suggestions are not meant to be blocking.

sktime/annotation/igts.py
```diff
@@ -0,0 +1,486 @@
...
+from sktime.base import BaseEstimator
+
+__all__ = ["InformationGainSegmentation"]
+__author__ = ["lmmentel"]
+
+
+# logger = logging.getLogger(__name__)
+
+
+@dataclass
```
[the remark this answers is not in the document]
lmmentel: You are right attrs and dataclasses have similarities on the definition side but for me they serve different purposes. attrs serves to reduce boilerplate for classes and has methods for controlling and tuning that, whereas dataclasses are a data container for moving structured objects around - before dataclasses I would use sth like namedtuple or a dict but a dataclass combines the strengths of both. Hope that helps.

sktime/annotation/igts.py
```diff
@@ -0,0 +1,486 @@
...
+
+def entropy(X: npt.ArrayLike) -> float:
+    """Shannons's entropy for time series.
+
+    As defined by equations (3) and (4) from [1]_.
+
+    Parameters
+    ----------
+    X: array_like
+        Time series data as a 2D numpy array with sequence index along rows
```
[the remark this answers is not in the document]
lmmentel: Not sure there was anything explicit in the paper about that, I would have to go in and check but the method won't produce anything sensible for univariate series.

sktime/annotation/igts.py
```diff
@@ -0,0 +1,486 @@
...
+
+def generate_segments(X: npt.ArrayLike, change_points: List[int]) -> npt.ArrayLike:
+    """Generate separate segments from time series based on change points."""
+    for start, end in zip(change_points[:-1], change_points[1:]):
+        yield X[start:end, :]
+
+
+def generate_segments_pandas(X: npt.ArrayLike, change_points: List) -> npt.ArrayLike:
+    """Generate separate segments from time series based on change points."""
+    for interval in pd.IntervalIndex.from_breaks(sorted(change_points), closed="both"):
```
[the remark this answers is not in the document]
lmmentel: I thought that `yield` make more sense especially for larger time series. Thanks for the comment, I updated the docstrings.

lmmentel:
> Hey all I think I actually ran into a similar issue with the init check when using `define`/`attrs`.
> 
> With `attrs` (and I think with what you are doing here @lmmentel), vars seems not to work properly. I think this has something to do with it not playing well with the `__dict__` function that `vars` seems to rely on. However the attributes are correctly added as expected during `__init__`, (which is captured by using `dir` instead), and I think is what we are really interested in???
> 
> `vars` and `dir` are of course slightly different functions though, as is summarized a bit [here](https://stackoverflow.com/questions/980249/difference-between-dir-and-vars-keys-in-python).
> 
> Basically I was able to get the tests to pass when I made the following change to the init tests:
> 
> ```
> init_params = _get_args(type(estimator).__init__)
> invalid_attr = set(init_params) - set(vars(estimator)) - {"self"}
> ```
> 
> change to:
> 
> ```
> init_params = _get_args(type(estimator).__init__)
> invalid_attr = set(init_params) - set(dir(estimator)) - {"self"}
> ```
> 
> I ended up not implementing this change in part because @fkiraly was not a fan of the change to the sktime interface tests, but I think since it has come up twice now I think it is worth re-evaluating if making this small change (ie `vars` -> `dir`) is consistent with what the test is looking for. I personally think it is, and I also think it would fix this problem? Thoughts?

Thanks for sharing. My point was a bit broader than fixing the test. I would say that this test serves little purpose other than slowing things down. I have searched and I couldn't find sklearn checking similar aspects so it seems sktime is more dogmatic in this area which by necessity reduces design flexibility but not in way that's desirable.

aiwalter (approved):

PR #3403, opened 2022-09-10
[ENH] Segmentation metrics

lmmentel:
##### For all contributions

##### For new estimators


fkiraly (dismissed):
Nice!

Should this not inherit from `BaseMetric` and live in `performance_metrics`?

lmmentel:
> Nice!
> 
> Should this not inherit from `BaseMetric` and live in `performance_metrics`?

I'm happy to move it into `performance_metrics` but I see little reason for a base class here.

fkiraly:
> I see little reason for a base class here.

Many metrics are parametric, so inheriting from `BaseMetric` makes sense.

lmmentel:
Sorry I don't see your point, do you care to elaborate. If you refer to metrics in this PR specifically that would be helpful.

lmmentel:
@Lovkush-A @aiwalter any chance you could share your thoughts here? Is there a need for a base class where a simple function would do the job?

sktime/performance_metrics/annotation/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+
+"""Metrics to assess performance of segmentation task."""
+
+from sktime.performance_metrics.annotation.metrics import (  # noqa
+    annotation_error,
+    hausdorff_error,
+    prediction_ratio,
+)
+
+__author__ = ["lmmentel"]
```
Lovkush-A, line 11: do we want an `__all__` variable in here?

(there must be an automatic way of creating the __all__ variable based on the imports, right??)

sktime/performance_metrics/annotation/metrics.py
```diff
@@ -0,0 +1,85 @@
...
+from sklearn.utils import check_array
+
+__author__ = ["lmmentel"]
+
+
+def annotation_error(
+    true_change_points: npt.ArrayLike, pred_change_points: npt.ArrayLike
+) -> float:
+    """
+    Annotation error measuring discrepancy in the number of change points.
```
Lovkush-A, line 21:
```suggestion
    Annotation error counting the difference in the number of change points.
```
- 'discrepancy' to 'difference' is simpler language and more specific
- 'counting' is more specific than 'measuring'.

sktime/performance_metrics/annotation/metrics.py
```diff
@@ -0,0 +1,85 @@
...
+
+import numpy as np
+import numpy.typing as npt
+from scipy.spatial.distance import directed_hausdorff
+from sklearn.utils import check_array
+
+__author__ = ["lmmentel"]
+
+
+def annotation_error(
```
Lovkush-A, line 17: is this the standard name for this? the phrase 'annotation error' sounds like any potential way of measuring error between annotations.

sktime/performance_metrics/annotation/metrics.py
```diff
@@ -0,0 +1,85 @@
...
+    true_change_points: npt.ArrayLike, pred_change_points: npt.ArrayLike
+) -> float:
+    """
+    Annotation error measuring discrepancy in the number of change points.
+
+    Parameters
+    ----------
+    true_change_points: array_like
+        Indexes of true change points
```
Lovkush-A, line 26: I think we should say whether this is iloc or loc indexes or both.

sktime/performance_metrics/annotation/metrics.py
```diff
@@ -0,0 +1,85 @@
...
+    true_change_points: np.array, pred_change_points: np.array, symmetric: bool = True
+) -> float:
+    """
+    Hausdorff metric measures how far two subsets of a metric space are from each other.
+
+    Parameters
+    ----------
+        true_change_points: true indices of change points
+        pred_change_points: predicted indices of change points
+        symmetric: if `True` symmetric Hasudorff distance will be used
```
Lovkush-A, line 49:
```suggestion
        symmetric: if `True` symmetric Hausdorff distance will be used
```

sktime/performance_metrics/annotation/metrics.py
```diff
@@ -0,0 +1,85 @@
...
+    pred_change_points = check_array(pred_change_points, ensure_2d=False)
+    return abs(true_change_points.size - pred_change_points.size)
+
+
+def hausdorff_error(
+    true_change_points: np.array, pred_change_points: np.array, symmetric: bool = True
+) -> float:
+    """
+    Hausdorff metric measures how far two subsets of a metric space are from each other.
+
```
Lovkush-A, line 44: worth explaining in the docstring which metric space the true and predicted change points live in

Lovkush-A (dismissed):
various minor suggestions.

also, type hints are inconsistent for the true/predicted change points.

Lovkush-A:
> Is there a need for a base class where a simple function would do the job?

I am in favour of base class, to ensure we have uniform interface. right now, the hausdorff distance has an extra parameter compared to the others. so if we were to do some kind of benchmarking, we cannot simply do: `for metric in metrics: metric(true, pred))`

sktime/performance_metrics/annotation/metrics.py
Lovkush-A: also worth saying in docstring that you are wrapping scipy function. (do we want user to be able to choose all parameters for scipy function. e.g. it has a seed?)

sktime/performance_metrics/annotation/__init__.py
lmmentel: `__all__` would be redundant since it would have all three functions which is equivalent to `*`

sktime/performance_metrics/annotation/metrics.py
lmmentel: This name was used in one of the papers, can't remember which one. It is used in a specific context so I wouldn't say it's a big deal. Happy to accept a suggestion on what to change it to.

lmmentel:
> > Is there a need for a base class where a simple function would do the job?
> 
> I am in favour of base class, to ensure we have uniform interface. right now, the hausdorff distance has an extra parameter compared to the others. so if we were to do some kind of benchmarking, we cannot simply do: `for metric in metrics: metric(true, pred))`

I'm not sure I see you point here, you can still call the function with additional params and a class implementation would have the issue.

Lovkush-A:
> I'm not sure I see you point here, you can still call the function with additional params and a class implementation would have the issue.

The additional params would become init params (at least, that is one option). Compare how sklearn has uniform interface despite random forests and linear regressors having different params.

sktime/performance_metrics/annotation/metrics.py
Lovkush-A: some options:
- 'count_error'
- 'counting_error'
- 'number_error'
- 'number_changepoints_error'
- 'difference_in_count'...

Lovkush-A (approved):
happy with changes since last review. accepting with assumption that whether or not we should use basemetric is going to be an ongoing discussion point.

PR #3411, opened 2022-09-12
[DOC] clarifications on deprecation notes

fkiraly:
This adds clarifications and a worked example to the deprecation policy.

The policy itself does not change, only clarifying details are added.

### DIFF
--- a/docs/source/developer_guide/deprecation.rst
+++ b/docs/source/developer_guide/deprecation.rst
@@ -18,9 +18,26 @@ Deprecation policy
 sktime `releases <https://github.com/alan-turing-institute/sktime/releases>`_ follow `semantic versioning <https://semver.org>`_.
 A release number denotes <major>.<minor>.<patch> versions.
 
-Our current deprecation policy is that we remove functionality after one minor release cycle.
-For example, if some functionality has been deprecated in v0.9.0 or any patch of that minor version, we will raise a warning throughout the next minor version v0.10.* and all of its patches.
-The functionality will then be removed in v0.11.0.
+Our current deprecation policy is as follows:
+
+* all interface breaking (not downwards compatible) changes to public interfaces must be accompanied by deprecation.
+  Examples: changes to defaults of existing parameters, removal of parameters.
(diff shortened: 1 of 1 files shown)


TonyBagnall (changes requested):
Do we still have the distinction between module development status? It was something like experimental, progressing and mature? (cant remember the exact terms). If we are keeping that distinction, it would be good to clarify the deprecation policy related to different statuses, because as I understand it experimental modules do not currently require deprecation

fkiraly:
@TonyBagnall, to the best of my knowledge, there has never been a distinction in the documented ("official") deprecation policy based on "maturity status".

We were generally watching out for public interfaces not getting broken.

So, introducing such a distinction would be a change to the deprecation policy - which is outside the scope of this pull request. This PR just clarifies the existing policy.

TonyBagnall:
@fkiraly is there still an "unofficial" distinction in policy for the deprecation policy for stable/maturing/experimental modules as defined here? https://github.com/alan-turing-institute/sktime/blob/main/README.md in the past we have not required a full cycle to deprecate code in experimental modules. Is this policy changing?

fkiraly:
> in the past we have not required a full cycle to deprecate code in experimental modules

I can remember this only for *private* modules.

Generally, we should not break any interfaces that someone could be using subject to common conventions ("public means stable") if we can avoid it.

TonyBagnall:
it definitely was the case that experimental did not require full deprecation, I remember distinctly with the clustering module. Thank you for clarifying that this is no longer the case though.

fkiraly:
> it definitely was the case that experimental did not require full deprecation, I remember distinctly with the clustering module. Thank you for clarifying that this is no longer the case though.

"was the case", I suppose you mean that PR got merged without proper deprecation in that module.
That sounds more like a mistake (or lenience) than an actual policy.

Either way, would appreciat if you comment whether the change request still needs action, as the discussion does not seem to have to do with the scope of this PR = making the existing rules clearer.

TonyBagnall (approved):

PR #3413, opened 2022-09-12
[BUG] Possible solution for pickle persistency fail for DL models

achieveordie:

Initially, `test_persistence_via_pickle()` of `test_all_estimators.py` was skipped for all DL models since there were problems correctly serializing the objects. So, added possibly corrected `__getstate__()` function for both `BaseDeepRegressor()` and `BaseDeepClassifier()`.

The current implementation seems to be working locally for `CNNClassifier`

No

##### For all contributions

### DIFF
--- a/sktime/classification/deep_learning/base.py
+++ b/sktime/classification/deep_learning/base.py
@@ -125,12 +125,46 @@ class BaseDeepClassifier(BaseClassifier, ABC):
         return y
 
     def __getstate__(self):
-        """Magic method called during serialization."""
+        """Get Dict config that will be used when a serialization method is called.
+
+        Returns
+        -------
+        copy : dict, the config to be serialized
+        """
         copy = self.__dict__.copy()
-        del copy["optimizer"]
-        del copy["model_"]
-        if copy.get("history") is not None:
-            del copy["history"]
-        if copy.get("optimizer_") is not None:
-            del copy["optimizer_"]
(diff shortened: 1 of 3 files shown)


achieveordie:
The failures are due to the known issue of not having the `optimizer` stored. Will fix it in the upcoming days.

fkiraly:
yes, this is part of the interface contract.

Note, we do not need to rely on pickle, we could rely on a save/load interface instead that replaces pickle with sth else.

achieveordie:
This PR can only be developed further after #3425 gets accepted since that deals with `.save()` / `.load()` class methods that will be used instead of pickle's `.dump()` / `.load()` for `test_persistence_via_pickle()`.

fkiraly:
hm, what I would do is to leave the save/load tests skipped, and instead write a test for pickling?
So we test the functionality added by this PR with a test also added in this PR.
Not as part of the generic test suite, but for one example of neural networks.

What do you think?

achieveordie:
The functionality in this PR is an outdated version of #3425. We still need to improve/correct/add test functionality which is the aim of this PR.
So after that PR is merged, we can continue updating the tests here. I could also copy-paste the relevant functions from that PR to this one but I think the former way is better.

achieveordie:
Closing in favour of #3425. This feature will be implemented there.

PR #3416, opened 2022-09-12
[DOC] Instructions for how to skip tests for new soft dependencies.

miraep8:
@fkiraly had some useful suggestions for how to fully isolate new soft dependencies.  In particular there were a few comments about how to skip (or selectively skip) tests for estimators with a new soft dependency.  Just copying those over to the developer docs.

### DIFF
--- a/docs/source/developer_guide/dependencies.rst
+++ b/docs/source/developer_guide/dependencies.rst
@@ -78,8 +78,8 @@ To add an estimator with a soft dependency, ensure the following:
    line in the doctest to skip. Check out the arima estimator as as an example. If concerned that skipping the test will reduce test coverage,
    consider exposing the doctest example as a pytest test function instead, see below how to handle soft dependencies in pytest functions.".
 *  Decorate all pytest tests that import soft dependencies with a ``@pytest.mark.skipif(...)`` conditional on a check to ``_check_soft_dependencies``
-   for your new soft depenency.  Be sure that all soft dependencies which are imported for testing are imported within the test funciton itself
-   rather than for the whole module)!  This decorator will then skip your test unless the system has the required packages installed.  Doing this is
+   for your new soft depenency.  Be sure that all soft dependencies which are imported for testing are imported within the test funciton itself,
+   rather than for the whole module!  This decorator will then skip your test unless the system has the required packages installed.  Doing this is
    helpful for any users running ``check_estimator`` on all estimators, or a full local `pytest` run without the required soft dependency.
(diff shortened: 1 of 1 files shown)


docs/source/developer_guide/dependencies.rst
```diff
@@ -74,6 +74,13 @@ To add an estimator with a soft dependency, ensure the following:
    utility should be used in ``__init__``. Both the warning and constructor call should use the ``package_import_alias`` argument for this.
 *  If the soft dependencies require specific python versions, the ``python_version``
    tag should also be populated, with a PEP 440 compliant version specification ``str`` such as ``"<3.10"`` or ``">3.6,~=3.8"``.
+*  Skip the corresponding doctests.  To do this add a ``# doctest: +SKIP`` to the end of each line in the doctest to skip.
```
fkiraly, line 77: Good point! I would formulate it from the position of not having examples yet rather than having examples and having to skip them.

E.g., sth like "if including docstring examples that use soft dependencies, ensure to skip doctest etc etc. To avoid reducing test coverage, consider exposing the example as a `pytest` test function instead, see below how to handle soft dependencies in `pytest` functions."

docs/source/developer_guide/dependencies.rst
```diff
@@ -74,6 +74,13 @@ To add an estimator with a soft dependency, ensure the following:
...
+   If concerned about reducing testing coverage, can add the doctest back in the testing file. See arima as an example.
+*  Decorate all tests for the estimator with the soft dependency with a ``@pytest.mark.skipif(...)`` conditional on a check to ``_check_soft_dependencies``
```
fkiraly, line 79: Maybe also more general, "all tests that import the soft dependency". Also mention that it should not be imported in the module, it should be imported inside the test.

docs/source/developer_guide/dependencies.rst
```diff
@@ -74,6 +74,13 @@ To add an estimator with a soft dependency, ensure the following:
...
+   for your new soft depenency.
+   This will skip your test unless the system has the required packages installed and is helpful for any users running ``check_estimator``
+   on all estimators.
```
fkiraly, line 82: ... or a full local `pytest` run without the soft dependency.

fkiraly (changes requested):
Great idea! I should have documented this, but I missed it.

Left some small remarks, not blocking but I think they will improve the doc.

fkiraly (approved):
Thanks, perfect!

One typo, search for `)!`

PR #3422, opened 2022-09-13
[ENH] Rename PluginParamsForecaster

aiwalter:


aiwalter:
I think actually that forecaster should live in forecasting module, but for now as it is experimental it can stay where it is probably

fkiraly:
ah, come on :-)

I think it is one of many pipeline adapters for the parameter plugin functionality, so it should stay where it is.

I also like the name.

fkiraly:
suggestion: let's not rename this or change location until it works for multiple estimator types

aiwalter:
My name is fine for multiple estimator types

PR #3425, opened 2022-09-14
[ENH] Save/Load Functionality for both Classical and DL Estimators

### LINKED ISSUE #3022: [ENH] save/load functionality for models

achieveordie:
#### Reference Issues/PRs

Fixes #3022

The Save/Load function for classical estimators is from #3336 and has been expanded for DL models along the same design. Keras is known to be problematic for certain kinds of (de)serialization, so this uses `model.save()` from `keras` along with `pickle.dump()`. The estimators are saved as `.zip` files.

Components of this directory include:
1. Metadata
2. Object
3. Keras History
4. Keras Model subdirectory

The first two components are the same as previously mentioned PR, stored differently to [avoid](https://github.com/alan-turing-institute/sktime/pull/3336#issuecomment-1242013162) this problem.

All the attributes are saved and restored (optimizers/history etc.) appropriately.

No

One thing in `__getstate__()` and `__setstate__()` might be a source of contention, I've explained why it was necessary and I believe it is the simplest solution.

##### For all contributions


achieveordie:
Following is a small code-snippet to showcase how "states are preserved" / "contract is maintained":

```python
import sktime.classification.deep_learning as dl
from sktime.base import load, load_dl
import numpy as np
from pprint import pprint
from pathlib import Path


if __name__ == "__main__":
    save_folder_location = "temp_saved"
    train_X = np.random.randn(15, 24, 16)
    train_y = np.random.randint(0, 2, size=(15, ))
    test_X = np.random.randn(5, 24, 16)

    cnn = dl.cnn.CNNClassifier(n_epochs=1)
    cnn.fit(train_X, train_y)

    cnn.save(save_folder_location)

    new_cnn = load_dl(save_folder_location)

    pprint(cnn.__dict__)
    pprint(cnn.model_.optimizer)
    pprint(cnn.optimizer)
    pprint(cnn.optimizer_)
    pprint(cnn.history)

    print("--" * 50)

    pprint(new_cnn.__dict__)
    pprint(new_cnn.model_.optimizer)
    pprint(new_cnn.optimizer)
    pprint(new_cnn.optimizer_)
    pprint(new_cnn.history)

    print("Does both give same result to same data?", np.allclose(
        cnn.predict(X=test_X),
        new_cnn.predict(X=test_X)
    ))

```

The output:
```
{'_X_metadata': {'has_nans': False,
                 'is_empty': False,
                 'is_equal_length': True,
                 'is_equally_spaced': True,
                 'is_one_series': False,
                 'is_univariate': False,
                 'mtype': 'numpy3D',
                 'n_instances': 15,
                 'scitype': 'Panel'},
 '_callbacks': [],
 '_class_dictionary': {0: 0, 1: 1},
 '_estimator_type': 'classifier',
 '_is_fitted': True,
 '_network': CNNNetwork(),
 '_tags_dynamic': {},
 '_threads_to_use': 1,
 'activation': 'sigmoid',
 'avg_pool_size': 3,
 'batch_size': 16,
 'callbacks': None,
 'classes_': array([0, 1]),
 'fit_time_': 3424,
 'history': <keras.callbacks.History object at 0x000001F77ECD9BB0>,
 'input_shape': (16, 24),
 'kernel_size': 7,
 'label_encoder': LabelEncoder(),
 'loss': 'mean_squared_error',
 'metrics': None,
 'model_': <keras.engine.functional.Functional object at 0x000001F778B355E0>,
 'n_classes_': 2,
 'n_conv_layers': 2,
 'n_epochs': 1,
 'onehot_encoder': OneHotEncoder(sparse=False),
 'optimizer': None,
 'optimizer_': <keras.optimizers.optimizer_v2.adam.Adam object at 0x000001F778B95760>,
 'random_state': None,
 'use_bias': True,
 'verbose': False}
<keras.optimizers.optimizer_v2.adam.Adam object at 0x000001F778B95760>
None
<keras.optimizers.optimizer_v2.adam.Adam object at 0x000001F778B95760>
<keras.callbacks.History object at 0x000001F77ECD9BB0>
----------------------------------------------------------------------------------------------------
{'_X_metadata': {'has_nans': False,
                 'is_empty': False,
                 'is_equal_length': True,
                 'is_equally_spaced': True,
                 'is_one_series': False,
                 'is_univariate': False,
                 'mtype': 'numpy3D',
                 'n_instances': 15,
                 'scitype': 'Panel'},
 '_callbacks': [],
 '_class_dictionary': {0: 0, 1: 1},
 '_estimator_type': 'classifier',
 '_is_fitted': True,
 '_network': CNNNetwork(),
 '_tags_dynamic': {},
 '_threads_to_use': 1,
 'activation': 'sigmoid',
 'avg_pool_size': 3,
 'batch_size': 16,
 'callbacks': None,
 'classes_': array([0, 1]),
 'fit_time_': 3424,
 'history': <keras.callbacks.History object at 0x000001F7022DC430>,
 'input_shape': (16, 24),
 'kernel_size': 7,
 'label_encoder': LabelEncoder(),
 'loss': 'mean_squared_error',
 'metrics': None,
 'model_': <keras.engine.functional.Functional object at 0x000001F70113F850>,
 'n_classes_': 2,
 'n_conv_layers': 2,
 'n_epochs': 1,
 'onehot_encoder': OneHotEncoder(sparse=False),
 'optimizer': None,
 'optimizer_': <keras.optimizers.optimizer_v2.adam.Adam object at 0x000001F70300C0A0>,
 'random_state': None,
 'use_bias': True,
 'verbose': False}
<keras.optimizers.optimizer_v2.adam.Adam object at 0x000001F70300C0A0>
None
<keras.optimizers.optimizer_v2.adam.Adam object at 0x000001F70300C0A0>
<keras.callbacks.History object at 0x000001F7022DC430>
1/1 [==============================] - 0s 253ms/step
1/1 [==============================] - 0s 174ms/step
Does both give same result to same data? True
```

@AurumnPegasus  I would also like for you to code-review this PR and give some suggestion for possible tests.

AurumnPegasus:
Hey @achieveordie ,
Thanks for picking this up! I was not able to find a solution to adding keras history and loading it from Zip file (from Franz's PR #3336 ) and am glad you did.
In case they would be helpful, I had a design doc in work for the same as a [STEP Document](https://github.com/sktime/enhancement-proposals/pull/27), though as mentioned above, there were some solutions I was not able to find.

I went through your PR and I think the code is good. I agree with the `__getstate__` and `__setstate__` point, and think that it is neccessary (generally required because you cannot pickle `lambda` functions, and `keras.optimizers` use them in there implementation, and hence there is no way around it.) I like the idea of having a seperate `load_dl` as well (since I think working with Zip files was proving troublesome for DL)

About the tests, there were a few I had in mind:
1. Idempotent test: Compare the predictions of the fitted model before and after saving the model. This will ensure that the model loaded is the one which was saved. 
This is an existing test which had been ignored for all DL Estimators because we were not getting exact results before and after saving (this was done before optimizer was added as a parameter, and hence we could work this out with `pickle`). The reason it was ignored was although the answers were very similar, they were not able to satisfy the threshold for similarity (which was around 1e-6) due to randomness of DL Estimators.
2. Saving with other parameters: Although I do not think any of the current parameters will give this issue, I still think we should test by sending some value of all parameters (and not taking `None` in any of them) so as to check they dont give the `lambda` issue `optimizer` gives us. It might be more useful for a future use case where someone adds a parameter which uses the `lambda` function.

achieveordie:
Thanks for the review!

My only problem with a separate `load_dl()` was that we'd have to modify tests to call `load_dl()` instead of `load()` for relevant cases, not to mention it might cause a bit of confusion on the user's end.
The easiest solution that I can think of is if we modify the code to only have `load()` as an entry point which would either call `_load_classical()` or `_load_dl()` depending if `path` is a file or a directory, or simply add another parameter where the user passes in that information. 
I think @fkiraly would be against the latter solution since it would require the user to have some pre-information about the model.

As for tests for the current implementation, I agree with your points. However, the idempotent test seems to be testing for fitting model twice as opposed to save/load. I might be wrong since I just glanced at it. For the second point, I was wondering if there is already an implementation that I could look at in the existing tests?

fkiraly (changes requested):
I think the idea looks good, but we need to get to a state where there is only a single `load` function. Why: think of the workflow, you are breaking the strategy pattern. This would introduce a case distinction of selecting the right "load" file, this does not scale well and gets confusing. Knowledge of what estimator is in the file/directory should not be assumed.

All specific functionality for loading objects should be localized in an override of one of the `load_` methods of the class. The idea is that `load` is universal, and any specific functionality sits in the class. This is achieved by a file format that wraps the class and a custom store.

AurumnPegasus:
> The easiest solution that I can think of is if we modify the code to only have load() as an entry point which would either call _load_classical() or _load_dl() depending if path is a file or a directory

This looks quite easy to do, and should not require user to know anything about either the estimator or the saved file/directory. Maybe the python command `os.path.isdir(path)` would be of help? I am sure it works as required for unix-like systems, and unsure about windows since I have not used it on that. 
This would allow an easy (although "hacky") solution to the problem

fkiraly:
> The easiest solution that I can think of is if we modify the code to only have load() as an entry point which would either call _load_classical() or _load_dl() depending if path is a file or a directory

This is the same idea that I was pursuing, I was just attatching these as overrides to the *class* they are relevant for.

So, yes, but `_load_classical` already is in the base class in my design, as methods `load_from`, etc, and `_load_dl` should simply override the `load_from_` methods in the neural network base class.

`load` should be universal and not require extension or modification - if at current state it does, then it means we need to move more in the class specific loaders.

achieveordie:
If we add new estimators that require a different logic than already provided in base classes(having custom objects for example), then the author would have to expand the class method `load_from_path() ` / `load_from_serial()`. The same is true for your design. The private functions that I've added are simply modularized, done to reduce boilerplate code. These external functions call class methods similar to your original design.

fkiraly:
Hm, can you explain why you added more code in `load` then?

My concern is that subject to your design, the same might be necessary if, say, we add a `pytorch` adapter.

Though I admit I have perhaps not fully understood how you see the extension contract, so I'm happy to be convinced that this is the last time something has to be added to `load`.

fkiraly:
I've updated the STEP to clarify what I had in mind:
https://github.com/sktime/enhancement-proposals/pull/27

achieveordie:
Perhaps we don't understand each other. The additional code is due to the fact that non-DL and DL estimators are being stored in different ways. For the sake of discussion, if we were to save non-DL estimators in a directory (not zip them), then we can get rid of `_load_classical()` and `_load_dl()` functions since they'd have the same workflow. 

Going from zip to directory-based for non-DL estimators seems easy, whereas going from directory to file-based storage (for DL estimators) needs more exploration. It might be easy/difficult/impossible. We need to discuss this further.

fkiraly:
>Perhaps we don't understand each other. 

Yes, I suppose so.

>The additional code is due to the fact that non-DL and DL estimators are being stored in different ways. 

I'm not saying "there should not be additional/different code for DL estimators".

I'm saying "there will likely be different/additional code, and it should live in the base class of the estimator, not in the `load` function, because we want to go with the strategy object orientation pattern".

fkiraly (approved):
hm, nice!

I think this solves the extensibility issue neatly.

Didn't know `shutil` until now.

Will merge this post 0.13.3 so we have some time for testing, and since it would make it 0.14.0 (major interface addition).

fkiraly:
question: is there a way to avoid the creation of the temp folder?
This may upset virus scanners etc.

achieveordie:
> question: is there a way to avoid the creation of the temp folder? This may upset virus scanners etc.

Wouldn't this be a problem with `keras`'s folder creation too? Or it's problematic because the code creates AND deletes it?

fkiraly:
> Or it's problematic because the code creates AND deletes it?

I was just thinking to minimize file write/delete actions, as heuristic detection may find more steps more odd, but that's just a gut feeling.

fkiraly (changes requested):
Thanks for this!

Sorry for this change request, this came up when I tried to merge.

You said that you based this on https://github.com/sktime/sktime/pull/3336, but it looks like you did not branch off but just copy-pasted code?

Further, you removed the tests from #3336, so I added them back.
Kindly check whether all still works as intended.

Overall, great contribution in solving the one-file and deep learning issues of serialization!

fkiraly:
hm, seems like save/load is failing when you try to run it for some deep learning estimators - could you kindly have a look?

achieveordie:
It was based on the implementation of #3336, not branched from it - sorry I didn't explicitly mention that.

I had removed the tests here since they are being worked on in #3413. You'd see similar error traces in job failures in both cases. Looking through it, I sense that there's a problem with how `scenarios` run. Since there are plenty of files and you have authored them, I think we can discuss why it's failing tomorrow in the meeting. I strongly suspect something is not working the way it's supposed to in the `_testing/` directory. The error indicates `.fit()` was probably not called before the proceeding method.

PS: there's also a doctest failure.

fkiraly:
> I strongly suspect something is not working the way it's supposed to in the `_testing/` directory.

Ok, how does that explain then that everything on `main` (i.e., the PR #3336) is passing?

achieveordie:
I could be wrong but the reason why #3336 passes and this doesn't might be because all (non-DL) estimators comply with expected test specifications. In contrast, DL estimators probably don't because the tests were written with non-DL estimators in mind.

You'd say the functionalities should be altered to satisfy tests and not the other way around - and you'd be right. I'm saying that there could be certain assumptions in tests which are not true DL estimators. If that's the case we would have to update the testing strategy, assuming implementation is correct for DL estimators (which, after manual testing, I think it is).

In any case, I'll try to come up with specific questions regarding how `scenarios` work by tomorrow and that could help us identify the problem.

achieveordie:
I see we've been discussing two different tests here. Let's break the conversation down:

There are two 'problematic' test functions that we were initially skipping - `test_persistence_via_pickle()` & `test_fit_idempotent()`. You were talking about the former while I was talking about the latter.

Easy to debug first: `test_persistence_via_pickle()` is failing because the default in-memory serialization from #3336 doesn't take into account foreign objects that need to be serialized, in our case that is the `keras` model, hence the error. I did look around to try and find a way to serialize `keras` model into memory and found [this](https://github.com/keras-team/keras/issues/9343#issuecomment-440903847). Kindly look into it and LMK.

`test_fit_idempotent()` needs discussion, will update the comment post that.

fkiraly:
Yes, I think you are right on two points that I missed:

* there should be a test for save/load from file, indeed - we do not cover that properly! I agree we need to.
* I missed that `test_fit_idempotent` was failing for another reason. We should probably rewrite this in terms of save/load instead of using `deepcopy`.

fkiraly:
for `test_persistance_via_pickle`, I think we can leave the test as is (but we should probably rename the test, this may require deprecation since users may be referring to it via `check_estimator`), and we should implement the save/load interfaces against it.

achieveordie:
So here's what I propose - we do not change `test_persistence_via_pickle`, just update the doc to explicitly mention that we are checking for in-memory serialization. We add a `test_file_persistence_via_pickle` where estimators are saved to files. This way we avoid deprecation.

> I did look around to try and find a way to serialize keras model into memory and found [this](https://github.com/keras-team/keras/issues/9343#issuecomment-440903847). Kindly look into it and LMK.

As the first step, I'll start updating the implementation as per the link, which should resolve `test_persistence_via_pickle()` (and the new `test_file_persistence_via_pickle`).

After that, we can update the runner to solve `test_fit_idempotent()`. I'll also close #3413 since that PR was aimed at fixing this.

LMK if this sounds good so I can start.

fkiraly:
yes, sounds good!

Minor comment, I would avoid using `pickle` in the test name. The old test requires deprecation since it might be in use (let´s do that separately), but we can avoid use of the term `pickle` in the new test.

achieveordie:
@fkiraly When we're saving to memory, I don't think it's possible to serialize two objects into a single variable, so we'd have to modify the in-memory serialization as:
```python
if path is None:
    type(self), pickle.dumps(self), None
```
for non-DL estimators. The last variable above will store the serialized `keras` model.

However, if you think this is suboptimal or breaks any user contract, I'd like to hear a different approach to do this.

achieveordie:
I'm making an early commit to explain how `keras` serialization is supposed to work for in-memory.

As explained in my comment above, we'd now return a tuple of size three, with the first two elements same as before. The third element will be `None` for non-DL estimators whereas it'll be a tuple of size two for DL estimators.

The first element will be in-memory serialized `keras` model produced via `h5py` whereas the second element will be in-memory history produced via `pickle.dumps()`.

The following cases now work:
1. Manually loading the in-memory-saved `keras` model:

```python
from sktime.classification.deep_learning import CNNClassifier
import h5py
from keras.models import load_model
import numpy as np

X = np.random.randn(15, 28, 16)
y = np.random.randint(0, high=2, size=(15, ))

cnn = CNNClassifier(n_epochs=1)

cnn.fit(X, y)

# Manual loading check
a, b, c = cnn.save()

# Only care about seeing if keras-model
# loads correctly, so skip a, b, c[1]
with open("temp_file", 'wb') as raw_file:
    raw_file.write(c[0])

with h5py.File("temp_file", 'r') as h5file:
    loaded_model = load_model(h5file)

X = X.transpose([0, 2, 1])
print(
    np.allclose(
        cnn.model_.predict(X),
        loaded_model.predict(X)
    )
)
# True
```

2. Using the dedicated `load` from `sktime`:

```python
from sktime.classification.deep_learning import CNNClassifier
import numpy as np
from sktime.base import load


X = np.random.randn(15, 28, 16)
y = np.random.randint(0, high=2, size=(15, ))

cnn = CNNClassifier(n_epochs=1)

cnn.fit(X, y)

serial = cnn.save()
loaded_cnn = load(serial)

print(
    np.allclose(
        cnn.predict(X),
        loaded_cnn.predict(X))
)
# True
```

fkiraly:
>As explained in my comment above, we'd now return a tuple of size three, with the first two elements same as before. The third element will be None for non-DL estimators whereas it'll be a tuple of size two for DL estimators.

This is violating the strategy pattern - you are changing the base interface to accomodate for DL estimators. Again, think about the extension case. If we add a third estimator type (e.g., super-DL), would you extend the tuple size to four?

A simple solution to keep interface homogeneity is to return a tuple of size two, and the second element can be a tuple of size two. Also has three elements, but the general format is the same ("what is in the second element" being type specific).

fkiraly:
perhaps a very stupid question: is there a way to avoid the h5py dependency? I thought HDF5 was outdated for tensorflow, is there a way to use `SavedModel`?

fkiraly:
ps: if you want the tests to run, you need to add the soft dependency (also, please isolate it)

achieveordie:
> A simple solution to keep interface homogeneity is to return a tuple of size two

Sure thing, I'll modify the code to do that.

> is there a way to use `SavedModel` ?

I'm not sure. `SavedModel` is a `tensorflow` exclusive whereas `h5py` has a more extensive scope and dedicated support. Plus `SavedModel` makes a directory so we'd have to deal with serializing multiple objects in memory and then make sure they communicate the same way they do when stored on disk.

> (also, please isolate it)

How do I do that? I didn't update the setup since I just wanted to show how it'll work. Will add the softdeps and other things today.

achieveordie:
Things added:
- in-memory serialization of deep estimators.
- test to save an estimator as a file.
- in-memory serialization of deep regression estimators.

Tests should still fail either because of faulty individual estimators (see #3538 and #3539) or because of current `test_fit_idempotent()` implementation(see [comment](https://github.com/sktime/sktime/pull/3425#issuecomment-1267957975)).

I suggest excluding these again (was removed during this PR) and work on them separately.

fkiraly:
> because of current test_fit_idempotent() implementation

Should we change that to use deepcopy, or save/load? (not sure which one would be better)

> Tests should still fail either because of faulty individual estimators

I'd separate fixing these from the general framework issue, i.e., except them in the `_config` and link to the issue number in a comment

achieveordie:
> Should we change that to use deepcopy, or save/load? (not sure which one would be better)

AFAIK currently, deepcopy is being used under the hood for `test_fit_idempotent`. We'd have to diagnose why exactly it's failing. I'll add an issue regarding it in the upcoming days.

> except them in the _config and link to the issue number in a comment

Could you please add the exceptions in the configuration file? I think all the current failures are to be excluded because these are problems with the test/estimator rather than save/load interface. But I might be wrong and so it'll be great if you could confirm it.

fkiraly:
> Could you please add the exceptions in the configuration file? I think all the current failures are to be excluded because these are problems with the test/estimator rather than save/load interface. 

I added excepts for the failures that are covered by currently linked issues, to the best of my knowledge.

>But I might be wrong and so it'll be great if you could confirm it.

I do think the `FCNClassifier` failure is not covered by any known issue, and in fact the tests pass on `main`.

achieveordie:
`test_fit_idempotent()` was failing for `FCNClassifier`. Since we know that this test is problematic, how come it passed in the main branch?

achieveordie:
When this PR was created, `FCNClassifier` was skipping `test_fit_idempotent()` like other DL estimators, as seen in the files changed. When was it added back?

fkiraly:
> When this PR was created, FCNClassifier was skipping test_fit_idempotent() like other DL estimators, as seen in the files changed. When was it added back?

Don't know - probably a side effect of copy-pasting instead of branching. If it was skipped on main, feel free to add it back.

fkiraly:
quickly looked into it - it's not the merging as I claimed, but we removed the skipped tests for deep learning estimators in order to test the new save/load for deep learning.

So, this would be a genuine failure of the new interface, but it seems to be estimator specitic?

achieveordie:
error_ss

This is the output from the latest run. All failures in `CNNRegressor` is due to the known issue #3538. There seems to be a genuine problem in `MLPClassifier` for `test_fit_idempotent`. The outputs are close but they don't satisfy the criteria of `decimal=6`. This is probably sporadic, we might have to look more into it.

fkiraly:
can you please explain what is going on with `CNNRegressor`, to the best of your understanding?

Re `MLPClassifier`, does that mean the `random_state` parameter does not work properly?

achieveordie:
My comment [here]() explains the problem with `CNNRegressor`. 

AFAIK, All other failures have `test_fit_idempotent()` as a common denominator. There seem to be two main reasons why this is failing.

- Type-1: [Assert failures](https://github.com/sktime/sktime/actions/runs/3210369162/jobs/5247990847#step:10:538) are sporadic failures that occur when the idempotent property of a method doesn't satisfy the `decimal` accuracy we want.
- Type-2: [Attribute failures](https://github.com/sktime/sktime/actions/runs/3210369162/jobs/5247990716#step:7:360) are due to (faulty?) implementation of the test function. Under the hood, `deepcopy` is being employed to store parameters which in turn uses `__getstate__()`, similar to pickle. This is problematic because the function during which this is called is `fit`, and at that point of execution, `self.model_` isn't defined, hence we get an error. It's weird that this isn't happening always. One would probably need to understand how `deepcopy` works, but this is what I get from those errors.

fkiraly:
> My comment here explains the problem with `CNNRegressor`.

This links to this very PR, not a specific comment. Do you mean this issue? https://github.com/sktime/sktime/issues/3538

fkiraly:
> Type-1: Assert failures are sporadic failures that occur when the idempotent property of a method doesn't satisfy the decimal accuracy we want.

In theory, this should not be happen - if a `random_state` is set, the same `fit` should lead to the same output with high decimal accuracy. That indicates to me that `random_state` is not working as intended.

> Type-2: Attribute failures are due to (faulty?) implementation of the test function. Under the hood, deepcopy is being employed to store parameters which in turn uses __getstate__(), similar to pickle. This is problematic because the function during which this is called is fit, and at that point of execution, self.model_ isn't defined, hence we get an error. It's weird that this isn't happening always. One would probably need to understand how deepcopy works, but this is what I get from those errors.

This sounds odd, I thought `deepcopy` worked with any `python` object, including those that are not pickleable. But I may be wrong.

In this case, would replacing `deepcopy` with `save` or `load` work?

Either way, we need to reduce the problem to minimal code, e.g., applying `deepcopy` to a fitted estimator, or our understanding will remain obscure and based on conjecture.

You could test yours by applying `deepcopy` to a suitable fitted estimator and seeing whether it breaks in the same way?

fkiraly:
From https://github.com/sktime/sktime/issues/3538, it seems to become apparent that you had assumed in your implementation of save/load that all deep learning estimators have an `optimizer` attribute.

I don't think that is a good assumption to make? It is certainly not a documented interface assumption. Or do we expect all deep learning estimators to have an `optimizer` attribute? If yes, why?

achieveordie:
I'll make the changes where the code doesn't expect `optimizer` to be present for a DL estimator. It seemed reasonable to me that all DL estimators have an `optimizer` since we employ `compile` from `keras` in our current interface. Even if this interface changes radically in the future (say the user has the ability to define their own custom training step), optimizer would still be pivotal, unless the interface stops using `tensorflow` altogether. 

Besides, I don't think there is any contention in the fact that `CNNRegressor` should have an `optimizer` parameter in its [constructor](https://github.com/sktime/sktime/blob/13239f38ec9ac8157f70aa2c70487c7978e38ac9/sktime/regression/deep_learning/cnn.py#L50). If not for the current implementation, this minor bug would probably have gone unnoticed for longer.

fkiraly:
> Besides, I don't think there is any contention in the fact that `CNNRegressor` should have an `optimizer` parameter in its [constructor](https://github.com/sktime/sktime/blob/13239f38ec9ac8157f70aa2c70487c7978e38ac9/sktime/regression/deep_learning/cnn.py#L50). If not for the current implementation, this minor bug would probably have gone unnoticed for longer.

Ok - just to spell this out, I'm less concerned about individual estimators than making this an interface expectation on all deep learning estimators.

If we want to make `optimizer` an interface expectation on all deep learning estimators, it should be a documented interface point in the base classes, perhaps even a property.

achieveordie:
After taking care of the issue with `optimizer`, there are two kinds of failures:
- `TapNet` failures are due to the known issue #3539 
- Failures from the problematic `test_fit_idempotent()`

If you think that either of them is in the scope of this PR, LMK, otherwise one can skip these tests and work on it in another PR.

achieveordie:
@fkiraly It seems that the only errors remaining are the `AssertionError` from `test_fit_idempotent()`. Could you look into them and tell me how you would like them to be handled? If it is beyond the scope we'd be better off creating an issue, if not then we'd have to look more into it.

Do leave some comments about the added tests for base deep classifiers, these tests were meant to check if  `__getstate__()` and `__setstate__()` work properly in edge cases or not. I might add more tests in future.

fkiraly:
Thanks! looks great!

Did you check (by deactivating `--matrixdesign` locally) whether TapNet and FCN are the only estimators affected?

sktime/classification/deep_learning/base.py
```diff
@@ -123,3 +124,242 @@ def convert_y_to_keras(self, y):
...
+        if `path` is None - in-memory serialized self
+        if `path` is file location - ZipFile with reference to the file
+        """
+        import pickle
+        import shutil
+        from pathlib import Path
+        from zipfile import ZipFile
+
+        if path is None:
+            import h5py
```
fkiraly, line 244: we need to make sure we have proper isolation of soft dependencies.
This would simply break at use and test.

For use, I would put a check and an error message, i.e., "in-memory serialization requires `h5py`" or similar.

For the testing, we need to think about what to skip when.
The serialization test (save/load to memory) will fail without `h5py`. Since it is part of the standard test suite, this creates a problem for which I can't see a nice solution.

fkiraly: (the general requirement being, the full test suite of sktime should run without failures with any set of soft dependencies installed)

fkiraly:
> @fkiraly It seems that the only errors remaining are the `AssertionError` from `test_fit_idempotent()`.

These look like genuine bugs to me, i.e., `random_state` not working properly for the two estimators TapNet and FCN.
I would open two bug issues, skip the tests in `_config`, and reference the bug issues in the skip comment.

fkiraly (changes requested):
Summarizing remaining items afaik:

* we need to deal somehow with the issue that test suite should not break for a user who has `tensorflow` but not `h5py` installed, and it still should have maximal possible coverage
* handling `fit_test_idempotent` `AssertionError` (see above)
* answer to "do we really need a special save/load now" below

fkiraly:
> Do leave some comments about the added tests for base deep classifiers, these tests were meant to check if  `__getstate__()` and `__setstate__()` work properly in edge cases or not. I might add more tests in future.

Tests look good to me!

Stupid question: now that we got set/getstate to work, and therefore `pickle`, do we really still need a custom save/load for deep learning estimators? I might be missing something obvious.

achieveordie:
> Did you check (by deactivating --matrixdesign locally) whether TapNet and FCN are the only estimators affected?

If you remember in one of the conversations I mentioned that my system fails to run the entire suite. It just hangs till all the worker crash. So I couldn't try this out. I think that all the deep estimators are affected (see `CNNClassifier` failing on this in the latest ubuntu runner).

> deal somehow with the issue that test suite should not break

I don't know the feasibility of doing this here's a possible solution - We create a fixture which runs only when certain soft-deps are included in the built. Otherwise, the tests in this fixture are just skipped. For now, we might have to skip all those tests regardless.

 > do we really still need a custom save/load

Yes. We are deleting the most essential part of the object - the `keras` model - before pickling. So the custom save/load ties up the work that `__getstate__()` / `__setstate__()` is doing (by using pickle) along with saving the model/history/optimizer associated with that model.

achieveordie:
So if I were to re-iterate, our revised remaining items will be:
- Skipping tests that require soft dependency till there's a solution for that.
- Skip all the deep estimators for `test_fit_idempotent()`.
- Add `ImportError` in case soft dep `h5py` is missing during import

If we agree, then I can move forward and make these changes in the next commit.

sktime/classification/deep_learning/base.py
achieveordie: `tensorflow` is also a part of the soft dependencies, right? Or we do not consider it since it's also included in `dl` package?

fkiraly:
> Skipping tests that require soft dependency till there's a solution for that.

Not sure what you mean by this, can you explain?

> Skip all the deep estimators for `test_fit_idempotent()`.

I thought with `getstate` solved issus for most. Are we having `AssertionError` for all deep learning models?

That would mean that `random_state` is not working at all for these, no?

> Add `ImportError` in case soft dep `h5py` is missing during import

Yes

achieveordie:
> Not sure what you mean by this, can you explain?

As you said that user's test suite may not have soft dependencies installed. So we can skip those tests. Specifically, any test that uses `h5py` (so in-memory serialization tests etc). I don't have a concrete idea if that's the best thing to do or not, but I'll leave that decision-making to you.

> I thought with `getstate` solved issus for most. Are we having AssertionError for all deep learning models?

The updated `getstate` solves the [Attribute Failure](https://github.com/sktime/sktime/pull/3425#issuecomment-1272471010). `AssertionFailure` seems more of an issue with the test function / scenario rather than the estimators, but that's just my initial hypothesis. In any case, it is an entirely different issue and is not concerned with save/load code.

fkiraly (approved):
Thanks, great contribution!

achieveordie:
@ilkersigirci thanks for the much-needed input. I think we were expecting to find some bugs since this has just been implemented. What I would suggest is that we should tackle this on a new issue. So maybe open a new issue (possibly with reproducible code) and let's take on it from there.

Thanks!

ilkersigirci:
Ofc, I will open a bug issue. BTW, in the `sktime.BaseObject` instead of using pickle, if we use `cloudpickle` or `dill`, all of this issues will be solved. This is because, we won't need to modify `__get__state__` and `__set__state__`. `cloudpickle` and `dill` can successfully serialize optimizer and lambda functions. Hence, I am in favor for this approach which changes `pickle` as discussed in #3672

PR #3430, opened 2022-09-15
[ENH] Add E-Agglo estimator for hierarchical agglomerative cluster estimation

### LINKED ISSUE #3397: [ENH] Hierarchical agglomerative estimation: E-Agglo
### Details 
E-Agglo is a nonparametric method for performing hierarchical agglomerative estimation of multiple change points based on [1][2]. Neighboring segments are then sequentially merged to maximize a goodness-of-fit statistic. The goodness-of-fit statistic is the "between-within distance" [3] among adjacent segments i.e.

image

where || . || is the Euclidean norm. Which is the scaled sample measure of divergence below

image

i.e.
``e(A, B) = mn/(m+n) * [minimum energy clustering]`` as evidenced in [4].

Note that minimum energy clustering can be computed with `scipy.stats.energy_distance(...)` [5].

#### References
[1] Matteson, David S., and Nicholas A. James. "A nonparametric approach for multiple change point analysis of multivariate data." Journal of the American Statistical Association 109.505 (2014): 334-345. URL: https://www.tandfonline.com/doi/full/10.1080/01621459.2013.849605
[2] James, Nicholas A., and David S. Matteson. "ecp: An R package for nonparametric multiple change point analysis of multivariate data." arXiv preprint arXiv:1309.3295 (2013). URL: https://arxiv.org/pdf/1309.3295.pdf
[3] Szekely, Gabor J., and Maria L. Rizzo. "Hierarchical clustering via joint between-within distances: Extending Ward's minimum variance method." Journal of classification 22.2 (2005): 151-184. URL: https://home.ipipan.waw.pl/j.mielniczuk/Szekely-Rizzo_2005.pdf
[4] https://en.wikipedia.org/wiki/Hierarchical_clustering#cite_ref-7
[5] https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.energy_distance.html

## Option 1: Interface sklearn.cluster.AgglomerativeClustering

[sklearn.cluster.AgglomerativeClustering](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.AgglomerativeClustering.html#sklearn.cluster.AgglomerativeClustering) has parameters `linkage{‘ward’, ‘complete’, ‘average’, ‘single’}, default=’ward’`.

We need another linkage type not offered in this list. Therefore, this task involves writing another version of [ward_tree](https://github.com/scikit-learn/scikit-learn/blob/36958fb24/sklearn/cluster/_agglomerative.py#L169), with distnace between clusters `e(A,B)` from above instead of Ward's linkage defined as

image
(the visualisation of linkage parameters comes from [here](https://towardsdatascience.com/machine-learning-algorithms-part-12-hierarchical-agglomerative-clustering-example-in-python-1e18e0075019))

and interfacing `AgglomerativeClustering(ClusterMixin, BaseEstimator)` with this extended functionality.

### Dependencies

- [ ] `scipy.cluster.hierarchy.ward(X)` s called, where
image

which is (almost) the same as the desired incremental metric from [2]:
image

- [ ] `ward_dist` relies on c++ ` _hierarchical.compute_ward_dist` - can this be fully replaced with the scipy function? [5]
- [ ] `ward_dist` relies on c++  `_hierarchical._get_parents` - can this be called?

## Option 2: rewrite R code as stand alone

The R code is: https://github.com/cran/ecp/blob/master/R/e_agglomerative.R

- [x] getWithin, getBetween are C functions that need to be re-written
https://github.com/cran/ecp/blob/master/src/energyChangePoint.cpp

but maybe they can be replaced by `scipy.stats.energy_distance(...)` [5] - wouldn't take too long to verify.

KatieBuc:
closes #3397 

E-Agglo is a nonparametric method for performing hierarchical agglomerative estimation of multiple change points based on [1][2]. Neighboring segments are then sequentially merged to maximize a goodness-of-fit statistic. The goodness-of-fit statistic is the "between-within distance" [3] among adjacent segments.

The code is being written as an annotator, re-written from this R code: https://github.com/cran/ecp/blob/master/R/e_agglomerative.R

[1] Matteson, David S., and Nicholas A. James. "A nonparametric approach for multiple change point analysis of multivariate data." Journal of the American Statistical Association 109.505 (2014): 334-345. URL: https://www.tandfonline.com/doi/full/10.1080/01621459.2013.849605
[2] James, Nicholas A., and David S. Matteson. "ecp: An R package for nonparametric multiple change point analysis of multivariate data." arXiv preprint arXiv:1309.3295 (2013). URL: https://arxiv.org/pdf/1309.3295.pdf
[3] Szekely, Gabor J., and Maria L. Rizzo. "Hierarchical clustering via joint between-within distances: Extending Ward's minimum variance method." Journal of classification 22.2 (2005): 151-184. URL: https://home.ipipan.waw.pl/j.mielniczuk/Szekely-Rizzo_2005.pdf

@lmmentel @fkiraly @Lovkush-A


sktime/annotation/e_agglo.py
```diff
@@ -280,6 +285,40 @@ def get_cluster(estimates):
...
+        if not X.index.equals(self._X.index):
+            X_full = X.combine_first(self._X)
+            new_eagglo = EAGGLO(
+                member=self.member,
+                alpha=self.alpha,
+                penalty=self.penalty,
+            ).fit(X_full)
+            warnings.warn(
+                "Warning: Input data X differs from that given to fit(). "
+                "Refitting with new input data, not storing updated public class "
```
Lovkush-A, line 315: Should say you refit with *both* the data in fit and new input data. Current wording suggests fitting with only new input data.

sktime/annotation/e_agglo.py
```diff
@@ -280,6 +285,40 @@ def get_cluster(estimates):
...
+                "attributes. For this, explicitly use fit(X) or fit_transform(X)."
+            )
+            return new_eagglo.cluster_
+
+        return self.cluster_
+
 
 def get_within(X, alpha):
```
Lovkush-A, line 323: the code in get_within and get_between is basically impossible to read. I am guessing this is what automatically is done by `black`, but there should be cleaner way of writing this.

Edit: i had a closer look and I think it should be possible to do this using numpy broadcasting / numpy functionality.

sktime/annotation/tests/test_e_agglo.py
```diff
@@ -9,7 +9,7 @@
 from sktime.annotation.e_agglo import EAGGLO
 
 
-def test_default_params_univariate():
+def test_fit_default_params_univariate():
     """Test data."""
     X = pd.DataFrame([-7.207066, -5.722571, 5.889715, 5.488990])
```
Lovkush-A, line 14: add a comment explaining where these numbers (and where teh expected numbers) come from.

sktime/annotation/e_agglo.py
Lovkush-A, line 323: i recommend changing name of function, to include distance/metric in the name.

Lovkush-A (changes requested):
Looks good! Few minor in line comments below.

Main recommendation is to add docstrings as you write the code, instead of only writing them at the end.

Another recommendation is to add unit-tests for some of the helper functions. E.g. for get_within.  Try testing with some 'extreme' values of inputs where the output answer is obvious. Maybe if X is constant, or if alpha is 1 or 0, or ...

sktime/annotation/e_agglo.py
```diff
@@ -0,0 +1,427 @@
...
+                X.loc[
+                    member_ == i,
+                ],
+                self.alpha,
+            )
+            for i in range(N_)
+        ]
+
+        # dataframe of between between-within distances
+        D_ = pd.DataFrame(index=range(2 * N_), columns=range(2 * N_))  # .fillna(np.Inf)
```
lmmentel, line 84: consider renaming `D_` to sth like `distances`

sktime/annotation/e_agglo.py
```diff
@@ -0,0 +1,427 @@
...
+
+        # check if sorted.
+
+        sizes_ = np.repeat(0, 2 * N_)
+        sizes_[:N_] = [
+            sum(member_ == i) for i in range(N_)
+        ]  # calculate initial cluster sizes_
+
+        # array of within distances
+        within = [
```
lmmentel, line 73: I'm not sure if a list comprehension is a good choice here, I would stick to numpy arrays an vectorized operations as much as possible.

sktime/annotation/e_agglo.py
```diff
@@ -0,0 +1,427 @@
...
+        super(EAGGLO, self).__init__()
+
+    def _process_data(self, X):
+        """Docstring."""
+        member_ = (
+            self.member if self.member is not None else np.array(range(X.shape[0]))
+        )
+
+        u = np.sort(np.unique(member_))  # unique array of cluster labels
+        N_ = len(u)  # number of clusters
```
lmmentel, line 58: how about renaming `N_` to `n_cluster` the you don't need the comment to explain it

sktime/annotation/e_agglo.py
lmmentel: Also, is there an advantage of using a DataFrame here instead of a regular np.array?

KatieBuc:
@lmmentel @fkiraly @Lovkush-A ready to merge? :)

Lovkush-A (changes requested):
whole chunk of suggestions. almost all about variable names or documentation.

some other suggestions:
- include unit tests for individual methods, e.g. for get_distance.
- include unit tests that directly check the code works, rather than only relying on unit-tests that assume the R implementation is correct. e.g. think of extreme inputs where we can determine what the output should be from our understanding of the algorithm. possibilities for extreme examples:
  - `member` is constant (so starting point is that everything is in one cluster already)
  - X is constant`[0,0,0,....,0]`
  - X is piecewise constant, e.g. `[0,0,0,...0,1,1,1,...1, 0,0,0,....,0]`, or other similar examples

sktime/annotation/e_agglo.py
```diff
@@ -0,0 +1,435 @@
...
+        # set initial GOF value
+        gof = np.array(
+            [
+                sum(
+                    [
+                        distances[i, left[i]] + distances[i, right[i]]
+                        for i in range(n_cluster)
+                    ]
+                )
+            ]
```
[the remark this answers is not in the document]
lmmentel: I am not sure I understand fully but I noticed that `distances` is `2 * n_clusters` x `2 * n_clusters` and both `left` and `right` are `2 * n_cluster - 1` size but this sum runs only until `n_clusters`, am I correct assuming that only a quarter of the distance matrix should be summed here?

sktime/annotation/e_agglo.py
```diff
@@ -0,0 +1,435 @@
...
+        )
+
+        # change point progression
+        progression = np.empty((n_cluster, n_cluster + 1))
+        progression[0, :] = [
+            sum(sizes[:i]) if i > 0 else 0 for i in range(n_cluster + 1)
+        ]  # N + 1 for cyclic mergers
+
+        # array to specify the starting point of a cluster
+        lm = np.repeat(0, 2 * n_cluster - 1)
```
lmmentel, line 170: Again this could be 
```suggestion
        lm = np.zeros(2 * n_cluster - 1)
```

sktime/annotation/e_agglo.py
```diff
@@ -0,0 +1,435 @@
...
+                    ]
+                )
+            ]
+        )
+
+        # change point progression
+        progression = np.empty((n_cluster, n_cluster + 1))
+        progression[0, :] = [
+            sum(sizes[:i]) if i > 0 else 0 for i in range(n_cluster + 1)
+        ]  # N + 1 for cyclic mergers
```
lmmentel, line 167: Isn't this a cumulative sum? Could use `np.cumsum` instead.

sktime/annotation/e_agglo.py
```diff
@@ -0,0 +1,435 @@
...
+        self.n_cluster = n_cluster
+        self.sizes = sizes
+        self.distances = distances
+        self.left = left
+        self.right = right
+        self.open = open
+        self.merged = merged
+        self.gof = gof
+        self.progression = progression
+        self.lm = lm
```
lmmentel, line 184: It's probably less of a gotcha if those are assigned directly where they are computed.

sktime/annotation/e_agglo.py
```diff
@@ -0,0 +1,435 @@
...
+            Data for anomaly detection (time series).
+        y : pd.Series, optional
+            Not used for this unsupervsed method.
+
+        Returns
+        -------
+        self :
+            Reference to self.
+        """
+        self._X = X
```
lmmentel, line 317: Is this necessary?

sktime/annotation/e_agglo.py
```diff
@@ -0,0 +1,435 @@
...
+    alpha-th moment). Estimation is performed in a manner that simultaneously
+    identifies both the number and locations of change points.
+
+    Parameters
+    ----------
+    member : array_like (default=None)
+        1D `array_like` representing the initial cluster membership for input
+        data, X.
+    alpha : float (default=1.0)
+        fixed constant alpha in (0, 2) used in the divergence measure, as the
```
lmmentel, line 35: based on `assert self.alpha > 0 and self.alpha <= 2` from `_fit` this should be
```suggestion
        fixed constant alpha in (0, 2] used in the divergence measure, as the
```

sktime/annotation/e_agglo.py
```diff
@@ -0,0 +1,435 @@
...
+
+        # find which clusters optimize the GOF and then update the distances
+        for K in range(self.n_cluster - 1, 2 * self.n_cluster - 2):
+            i, j = self._find_closest(K)
+            self._update_distances(i, j, K)
+
+        def filter_na(i):
+            return list(
+                filter(
+                    lambda v: v == v,
```
lmmentel, line 331: what might this be checking?

sktime/annotation/e_agglo.py
```diff
@@ -0,0 +1,435 @@
...
+                alpha=self.alpha,
+                penalty=self.penalty,
+            ).fit(X_full)
+            warnings.warn(
+                "Warning: Input data X differs from that given to fit(). "
+                "Refitting with both the data in fit and new input data, not storing "
+                "updated public class attributes. For this, explicitly use fit(X) or "
+                "fit_transform(X)."
+            )
+            return new_eagglo.cluster
```
lmmentel, line 398: This is quite unexpected for me and suggests that there might be an ill fit for using a transformer here. Feel free to ignore this comment since I don't have a suggestion to improve this atm :slightly_smiling_face:

sktime/annotation/tests/test_e_agglo.py
```diff
@@ -0,0 +1,128 @@
...
+
+    cluster_actual = fitted_model.cluster
+    fit_actual = fitted_model.gof
+
+    assert np.allclose(cluster_actual, cluster_expected)
+    assert np.allclose(fit_actual, fit_expected)
+
+
+def test_fit_other_params_univariate():
+    """Test univariate data with alternative starting clusters."""
```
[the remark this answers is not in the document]
lmmentel: why?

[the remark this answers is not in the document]
Lovkush-A: if you include it for some, may as well include it for all, no? may as well be consistent.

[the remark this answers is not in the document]
lmmentel: Consistency is nice, but that would be lowest on my priority list given how little value that actually brings.

sktime/annotation/e_agglo.py
```diff
@@ -0,0 +1,435 @@
...
+        alpha=1.0,
+        penalty=None,
+    ):
+        self.member = member
+        self.alpha = alpha
+        self.penalty = penalty
+        super(EAGGLO, self).__init__()
+
+    def _process_data(self, X):
+        """Initialize parameters and store to self."""
```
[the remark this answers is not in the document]
lmmentel: - :100:  on the type hits
- generally my preference is limiting writes to self, but if that's unavoidable next best thing is docs. Would you have that at the method level? Do you have any examples of good practice around this in sktime?

fkiraly:
@KatieBuc, are you still working on this? What is the status here?

sktime/annotation/eagglo.py
```diff
@@ -0,0 +1,430 @@
...
+
+    def _find_closest(self, K: int) -> Tuple[int, int]:
+        """Determine which clusters will be merged_, for K clusters.
+
+        Greedily optimize the goodness-of-fit statistic by merging the pair of adjacent
+        clusters that results in the largest increase of the statistic's value.
+
+        Parameters
+        ----------
+        K: int
```
miraep8, line 328: Sorry if I am being pedantic here, but I do not think `K` is as you say it is. Since where this is called `K` is in the range (self.n_cluster - 1, 2 * self.n_cluster - 2).  It is related to the number of original clusters yes... My suggestion would be to actually have it be range(self.n_cluster) and do additions needed to make it easier for indexing here. Just found this part slightly confusing as is.

sktime/annotation/eagglo.py
```diff
@@ -0,0 +1,430 @@
...
+        # which clusters were merged_, info only
+        self.merged_[K - self.n_cluster + 1, 0] = (
+            -i if i <= self.n_cluster else i - self.n_cluster
+        )
+        self.merged_[K - self.n_cluster + 1, 1] = (
+            -j if j <= self.n_cluster else j - self.n_cluster
+        )
+
+        # update left and right neighbors
+        ll = self.left[i]
```
miraep8, line 361: Personally I think it might be cleaner to have a function which takes care of getting the left/right neighbors rather than maintaining two separate arrays as such. Not necessary change though - just have seen approaches like this lead to unexpected errors.  Easier when there is one set of ground truth and other values are extracted in relation to that.

sktime/annotation/eagglo.py
```diff
@@ -0,0 +1,430 @@
...
+        return self.cluster_
+
+    def _initialize_params(self, X: pd.DataFrame) -> None:
+        """Initialize parameters and store to self."""
+        self._member = np.array(
+            self.member if self.member is not None else range(X.shape[0])
+        )
+
+        unique_labels = np.sort(np.unique(self._member))
+        self.n_cluster = len(unique_labels)
```
miraep8, line 212: I find it a bit strange that `self.n_cluster` is never updated? At least from what I could find from Cntl+F of "self.n_cluster =". I would have naively expected this to be updated each time we merge 2 clusters... 🤔

sktime/annotation/tests/test_eagglo.py
```diff
@@ -0,0 +1,128 @@
...
+
+    These numbers are generated from the original implemention
+    in R, with the following code:
+
+    set.seed(1234)
+    X <- c(rnorm(15, mean = -6), 0, rnorm(16, mean = 6))
+    X <- as.matrix(X[c(1,2,17,18)])
+    ret = e.agglo(X)
+    """
+    X = pd.DataFrame([-7.207066, -5.722571, 5.889715, 5.488990])
```
miraep8, line 23: Tests seem reasonable.  Feels a bit roundabout to expect agreement via the R implementation, but I think that the expected cluster assignments seem reasonable for the synthetic data so in the end they are good tests.

miraep8 (commented):
Nice.  I do have a few suggestions.  However I would consider them largely non-blocking. Overall feedback - it might be worth simplifying the logic here somewhat? ie lots of extra attributes and complications in how the algorithm is written.  I will admit to being a bit confused at certain points reading the code even though I think the underlying algorithm is pretty straightforward.  This could totally just be a me problem though. 😂 

In general my feedback regarding that would be that we could merge the PR as is and then look into improving in a future PR. Think that line of improvement could stall this PR for a long time - while I do think it is in a nearly-mergeable state as is. 

Nice tests overall though the roundabout comparison via the R package is a bit strange.  However I was satisfied that it was probably testing algorithm performance in terms of clustering in a reasonable manner. 

Only "blocking" change would be to include a `get_test_params` as I think this is useful and to also explain that penalty can also take strings (currently missing).

sktime/annotation/eagglo.py
```diff
@@ -0,0 +1,430 @@
...
+            Not used for this unsupervsed method.
+
+        Returns
+        -------
+        self :
+            Reference to self.
+        """
+        self._X = X
+
+        if self.alpha <= 0 or self.alpha > 2:
```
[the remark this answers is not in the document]
lmmentel: Unfortunately your suggestion goes directly against sklearn api design principles to not do **any** parameter validation in `__init__` see https://scikit-learn.org/dev/developers/develop.html#parameters-and-init I'll leave it as is to be consistent with sklearn.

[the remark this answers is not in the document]
miraep8: Whoops.  I actually had a sinking suspicion that might be the case when I suggested it.  Sounds good, leave as is. 👍

sktime/annotation/eagglo.py
```diff
@@ -0,0 +1,430 @@
...
+import pandas as pd
+from scipy.spatial.distance import cdist
+
+from sktime.transformations.base import BaseTransformer
+
+__author__ = ["KatieBuc"]
+__all__ = ["EAGGLO"]
+
+
+class EAGGLO(BaseTransformer):
```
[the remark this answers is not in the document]
lmmentel: Sure, I'm not attached to either name :wink:

sktime/annotation/eagglo.py
```diff
@@ -0,0 +1,430 @@
...
+    ----------
+    member : array_like (default=None)
+        Assigns points to to the initial cluster membership, therefore the first
+        dimension should be the same as for data. If `None` it will be initialized
+        to dummy vector where each point is assigned to separate cluster.
+    alpha : float (default=1.0)
+        Fixed constant alpha in (0, 2] used in the divergence measure, as the
+        alpha-th absolute moment, see equation (4) in [1]_.
+    penalty : functional (default=None)
+        Function that defines a penalization of the sequence of goodness-of-fit
```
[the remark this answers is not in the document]
lmmentel: Yup, good catch

sktime/annotation/eagglo.py
```diff
@@ -0,0 +1,430 @@
...
+        ----------
+        K: int
+            Number of clusters
+
+        Returns
+        -------
+        result : Tuple[int, int]
+            Tuple of left cluster and right cluster index values
+        """
+        best_fit = -1e10
```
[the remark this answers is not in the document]
lmmentel: If it starts as `None` the the comparison below won't won't works since comparing a `float` to `None` will result in a `TypeError`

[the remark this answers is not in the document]
miraep8: But that is where or statements come in handy.  Only first 1/2 gets evaluated if the first half is true. So can do if best_fit == None or gof > best_fit or something like that.

sktime/annotation/eagglo.py
```diff
@@ -0,0 +1,430 @@
...
+        self.n_cluster = len(unique_labels)
+
+        for i in range(
+            self.n_cluster
+        ):  # relabel clusters to be consecutive numbers (when user specified)
+            self._member[np.where(self._member == unique_labels[i])[0]] = i
+
+        # check if sorted
+        if not all(sorted(self._member) == self._member):
+            raise ValueError("'_member' should be sorted")
```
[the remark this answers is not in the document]
lmmentel: Not sure what you mean here, `member` should be sorted but since it's coming from the users it might be anything :wink:

sktime/annotation/eagglo.py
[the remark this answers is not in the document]
lmmentel: Not sure if that improves redability

sktime/annotation/eagglo.py
lmmentel: Can you elaborate on what unexpected errors might happen and under what circumstances?

sktime/annotation/eagglo.py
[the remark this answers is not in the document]
lmmentel: Added a line explaining the string params and describing options

sktime/annotation/eagglo.py
lmmentel: check `_update_distances` function for example

sktime/annotation/eagglo.py
lmmentel: I'm not sure, I'm guessing this follows a convention from the paper/original R implementation.

lmmentel: Maybe @KatieBuc could comment?

lmmentel:
Hey @KatieBuc could you have a glance and check if you're OK with the changes?

sktime/annotation/eagglo.py
```diff
@@ -0,0 +1,430 @@
...
+                    lambda v: v == v,
+                    self.progression[
+                        i,
+                    ],
+                )
+            )
+
+        # penalize the gof_ statistic
+        if self.penalty is not None:
+            penalty_func = get_penalty_func(self.penalty)
```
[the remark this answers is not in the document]
lmmentel: done

sktime/annotation/eagglo.py
miraep8: Ah - I think it is fine as is!  This was not a change request.  I was more thinking that you can derive a lot of these arrays (eg left/right, open etc) by looking at relationships to the current cluster array. There isn't anything wrong with keeping them as separate array though, so no need to change it. All I meant is that once you have a bunch of redundancy it can be easy to have errors sprout up if you are not on top of maintaining that redundancy everywhere.  Please consider this resolved though, was more of a future suggestion.

miraep8: ie was more thinking - in the future writing this differently might reduce bugs.

miraep8 (approved):
I think it looks good now (and further improvements can be made in a future PR).  Would be nice to wait and make sure @KatieBuc is on board with all changes - so perhaps give it a few days before merging?

sktime/annotation/eagglo.py
miraep8: I agree 100% 👍

lmmentel:
> I think it looks good now (and further improvements can be made in a future PR). Would be nice to wait and make sure @KatieBuc is on board with all changes - so perhaps give it a few days before merging?

Sounds good, I'm happy to give @KatieBuc some time to have a look. Thanks for looking into this PR @miraep8, your suggestions were quite valuable.

Issue #2254, opened 2022-03-19
[ENH] Cleaning up Catch22


fkiraly:
While refactoring, I cam across `Catch22`.

This is a bit of a mess (old legacy code structure) and should be refactored as follows, imo:
* `_transform` should return a `pd.DataFrame`, with columns being the names in `feature_names`
* `Catch22` should accept a parameter `features` such as used in various functions. This is a `str` or list of `str`, of codes in `feature_names`. If passed, the transformer will calculate only those features.
* `X_inner_mtype` should also accept `pd.DataFrame`, i.e., a single series (in which case no loop is run over instances)
* the docstring should describe all parameters properly.

Good first issue, I think - most of these changes should be straightforward, albeit a bit fiddly.

fkiraly:
@benfulcher has pointed me to this installable version of catch22:
https://github.com/chlubba/catch22/
I suggested a discussion on the topic of the "final form" of catch22, and possibly on running some benchmarking experiments, say, against tsfresh and the like.

benfulcher:
There is a group of people working with catch22, including some improvements in pull requests about to be incorporated. We may also in the next days move the repo to an organizational repo. It's in active development.
Note that catch22 is a subset of the larger hctsa (>7k features) set (which is independently being ported to python [here](https://github.com/fairscape/hctsa-py)).

MatthewMiddlehurst:
Happy to hear there is a Python version of hctsa coming 🙂.

The chlubba/catch22 package is what we originally used in sktime as a soft dependency, but after some issues with differing results by OS and one other issue I can't remember exactly breaking tests, we decided to just re-implement it using numba.

If this has been resolves/people are willing to overlook it, i don't see why we can't reintroduce it as an alternative.

benfulcher:
We now have a dedicated python repo: https://github.com/DynamicsAndNeuralSystems/pycatch22. May be best to develop any modifications you require onto this repo for consistency (or to add the OS dependency tests)?

fkiraly:
> or to add the OS dependency tests

Any plans for CI/CD in the repo?

benfulcher:
Being a static feature set, we don't anticipate changes, and are not proficient in implementing CI/CD, so this isn't on the cards.
Not sure why we need the transformer there…? Users can use pycatch22 to simply extract the features, and use a combination workflow with `sktime` to do any extras.

fkiraly:
[...]

PR #3431, opened 2022-09-15
Catch22 Update

MatthewMiddlehurst:
Updates Catch22 transformer and adds a Catch22Wrapper #2254


sktime/transformations/panel/catch22.py
```diff
@@ -45,27 +63,28 @@ class Catch22(BaseTransformer):
     """
 
     _tags = {
-        "scitype:transform-input": "Series",
-        # what is the scitype of X: Series, or Panel
+        "scitype:transform-input": "Panel",
```
fkiraly, line 66: I don't think that's right - one feature, one series

fkiraly (changes requested):
I know it's still a draft, but some comments:

* afaik Catch22 transforms a single series into features and does not look at other series in a panel. So it should be series-to-primitives. I would also avoid loop over instances internally, unless the computation becomes faster that way. Still, I would avoid using `numpy3D` internally, since that unnecessarily restricts the transformer to equal length series.
* if you introduce a `features` parameter that can be `str`, the set of valid `str` should be listed.

sktime/transformations/panel/catch22.py
MatthewMiddlehurst: Does that include multivariate series? i.e. dimension lengths (10, 4, 20) would output (10, 88) by default. Can change it back.

MatthewMiddlehurst:
I prefer the panel for multithreading, I imagine the overhead from swapping between formats using the transformer base class is pretty miniscule. Good point with unequal length, I didn't consider that but it should be a relatively easy change to DataFrame.

Will add the features to the doc, the names are very long which is a bit of a pain.

fkiraly:
to prevent the failures, ensure:
* the soft dependency is added to sktime
* the soft dependency is registered/isolated in the estimator

see instructions here: https://www.sktime.org/en/stable/developer_guide/dependencies.html

fkiraly (changes requested):
Please ensure all the tests pass

fkiraly (changes requested):
Thanks!

Is it correct that we are  changing the signature and logic in the `Catch22` transformer?

May I request some changes then:
* I think it should be possible to allow `int` and `list` of `int` for `features` - did we not allow this in the past?
* mean and variance (as in catch-24) should also be on the list
* if we change the parameter signature of the old transformer, there needs to be deprecation
* internally, I would use either numpy or `pd-multiindex`. `pandas` has deprecated functionality around handling of `nested_univ` so it is a risk to implement in that mtype.
* if there is no efficient way to loop over instances, implement the transformer for a single series instead and rely on broadcasting functionality from the base class

MatthewMiddlehurst:
> if we change the parameter signature of the old transformer, there needs to be deprecation

do you mean n_jobs?

fkiraly:
>> if we change the parameter signature of the old transformer, there needs to be deprecation

> do you mean n_jobs?

any - I don't know what was changed since the old parameters were not well documented

MatthewMiddlehurst:
the only parameter change is n_jobs -1 to 1. features is new (see __init__).

>I think it should be possible to allow int and list of int for features - did we not allow this in the past?

>mean and variance (as in catch-24) should also be on the list

>internally, I would use either numpy or pd-multiindex. pandas has deprecated functionality around handling of nested_univ so it is a risk to implement in that mtype.

Can look at these. Would rather have catch24 as a parameter rather than in the default features. Dont think I have seen/used pd-multiindex before

fkiraly:
> Can look at these. Would rather have catch24 as a parameter rather than in the default features. Dont think I have seen/used pd-multiindex before

Use `datatypes.get_examples` to see an example.

> the only parameter change is n_jobs -1 to 1. features is new (see __init__). I can deprecate -1 to use the old functionality for now?

I think that should be deprecated, since it is a change of default.

MatthewMiddlehurst:
I have deprecated the n_jobs value and included the new parameter options. Changing the internal data type can wait for another PR I think, it is listed as a bonus in the issue anyway.

TonyBagnall (approved):

Issue #2942, opened 2022-07-06
[ENH] clean up the `_contrib` folder, aim to remove


fkiraly:
There apparently is a `_contrib` folder in `sktime` with ... stuff ... in it.
Tbh it is unclear to me when it was decided that we should have it, what the scope of it is, or how it got populated with a collection of seemingly random content.

I think we need to deal with it, stop putting more random stuff in it, and eventually get rid of it.

* decide what should stay in `sktime` and what should go.
* if stays, move content in appropriate folders, e.g., classifiers in `classification`, docs into documentation, and so on. Ensure content is clean, well-documented, and interface compliant.
* content could also move to different repositories.

FYI @TonyBagnall, what is all this?

fkiraly:
some discussion by @TonyBagnall in the duplicate issue https://github.com/alan-turing-institute/sktime/issues/2947

fkiraly:
added a list of things we should do imo.

FYI @MatthewMiddlehurst, @chrisholder

TonyBagnall:
so I was just going to delete it once we had extracted the sklearn classifiers then you can put anything anyone wants back in later. All of this is experimental or redundant code and what you suggest involves a significant amount of work for very little benefit. But if anyone wants to do as you describe thats great.

PR #3434, opened 2022-09-16
[ENH] remove contrib

TonyBagnall:
#### Reference Issues/PRs
#2942 

contrib was a folder we included at the beginning of sktime as a working area for code not suitable or ready for sktime proper. We made it private a year or so ago, but I think there is no need for it any more. My group certainly doesnt need it. The simplest solution, and the one I am proposing, is just deleting the directory.  I understand @fkiraly wants it recycled, see #2942, which proposes working up certain elements to include in sktime proper. No one has volunteered for these tasks and I dont want to do it. Much of the code in contrib requires significant work to get it in, and much is just inferior or old versions of existing code.  If we go with #2942 I think we will just be left with contrib indefinitely


aiwalter (approved):

fkiraly (changes requested):
"no one has volunteered" is inaccurate and also wrongly framed.

No one needs to "volunteer", you just make PR.

Further, I was slowly working on it, and so were others, e.g., see the PR on the shapelet transforms. There is also ongoing work for which the folder might be useful.

Generally, I think it's not a good idea to go for an "approval model" like you seem to be imply, i.e., that people have to tell @TonyBagnall specifically, or it's considered as "not being worked on".

I'm voting this formally down because people are actually working on moving the content (albeit slowly), it just seems that @TonyBagnall is not aware of it.

aiwalter:
we can still work on getting stuff from this folder into the repo, you can always find the folder in the tag history of the repo. So I dont see this as an argument to not remove it actually.

TonyBagnall:
I would have thought people working on it might comment on your PR or might be assigned. I don't think the original contributors wanted to do it, hence why I assumed no one was interested. But hey, good luck to all who want to contribute,  I am not looking for an approval model or anyone telling me anything. I simply think it would better to remove contrib then add things in you want later, code is still accessible.

fkiraly:
> I simply think it would better to remove contrib then add things in you want later, code is still accessible.

> we can still work on getting stuff from this folder into the repo, you can always find the folder in the tag history of the repo. So I dont see this as an argument to not remove it actually.

Imo a very strong argument for not removing it before it has been refactored is to not leave a gap in functionality for anyone using it - even though it is a private module, it has functionality that our users seems to have been using on occasion - e.g., the user recently who was looking for the shapelet transform (on discord I think).

fkiraly:
FYI, two observations:
* there is newer and useful code in `_contrib` that Tony et al never moved to the main modules
* there is useful code by third parties who were told by Tony et al to put their code there and are no longer active, so we shouldn't just delete it but ensure it is moved

See for instance the discussions here:
https://github.com/sktime/sktime/pull/3518
https://github.com/sktime/sktime/pull/3519

I propose to close this PR.

PR #3435, opened 2022-09-17
[ENH] predictive variance and quantiles for naive forecaster

### LINKED ISSUE #1905: [ENH] predictive intervals and quantiles for naive forecaster
- [ ] Describe naive predictive intervals (what is the method?)
- [ ] implement for `NaiveForecaster`

Related to the variance forecaster of @IlyasMoutawwakil

topher-lo:
#### Reference Issues/PRs
Closes #1905

- Implements `predict_var` and `predict_quantile` for `NaiveForecaster`
- We apply the following algorithm from FPP3 Chapter 5. (under the assumption of uncorrelated residuals):
  1. Compute "past" residuals (`y - y_pred`)
  2. Compute residuals MSE and residuals SE from "past" residuals, with degrees of freedom equal to:
     - number of nans in residuals for "last" and "mean" strategies
     - number of nans in residuals + 1 (for the drift coefficient SE) for "drift" strategy
  3. Compute forecast SE using residuals SE according to formulas from FPP3 Table 5.2 (see screenshot and references below).
     - If `cov == False`, we return a vector of forecast SE squared, otherwise if `cov == True`, we return a diagonal matrix with the vector of forecast SEs squared as the diagonal.
  5. (For quantiles) Assume normality and use z-scores to compute quantiles from forecast SE.
- `_predict_var` and `_predict_quantile` both return `pd.DataFrame` with same shape and schemas as `NaiveVariance`'s corresponding methods 

Screenshot 2022-09-17 001240

Two unit tests:
1. [x] `test_naive_predict_var_backwards`: For `fh=1` and large `T`, the forecast standard error is approximately equal to the residual standard errors. We use this property and the fact that the equation for computing forecast SE from residuals SE is fully invertible. 
2. [x] `test_naive_predict_quantiles_against_R_naive` Compare predicted quantiles with results from R naive forecast library

#### References
- [FPP3 Table 5.2](https://otexts.com/fpp3/prediction-intervals.html#benchmark-methods)
- [R naive forecast code](https://github.com/robjhyndman/forecast/blob/master/R/naive.R)

No.

1. Check that the logic for `test_naive_predict_var_backwards` passes the smell test. It is not a rigorous test proof-wise, but it applies the "working backwards" method for checking computation.
2. Compare the formulas specified in `partial_se_formulas` in `_predict_var` with the equations in FPP3 Table 5.2.

#### Any other comments?
- Not sure if these methods extend to the multivariate case
- In line 415, I used `np.squeeze(self.predict(fh=1))` to convert the result from `self.predict` into a scalar. Does this always work for all possible return signatures? I am not familar with all of sktime's acceptable datatypes. Does this need to be tested or is this behavior guaranteed according to some design contract (when `fh=1`)?
- R naive also has a "bootstrap" option for computing forecast variance and quantiles, but this functionality is already implemented in sktime's `BaggingForecaster`.

##### For all contributions


fkiraly:
Thanks!

Will dig into this later.

Quick question, is the idea that this is the same as `NaiveVariance(NaiveForecaster())`, but more efficiently computed due to explicit formulae in the case of the `NaiveForecaster`?

topher-lo:
> Thanks!
> 
> Will dig into this later.
> 
> Quick question, is the idea that this is the same as `NaiveVariance(NaiveForecaster())`, but more efficiently computed due to explicit formulae in the case of the `NaiveForecaster`?

The general idea is the same, but `NaiveVariance` will give biased estimates. The formulas are adjusted for degrees of freedom and (to my knowledge) the formulas are derived directly from the likelihoods for each naive strategy.

This implementation should be unbiased estimates of forecast variance. In fact, we aren't far behind the R implementation, they also added in a [bias adjustment 16 days ago](https://github.com/robjhyndman/forecast/commit/d9fe7102b883ecc6f5bfc4b7c6f26400ad1feeda).

fkiraly (changes requested):
`_predict_var` looks good!

Change request: `_predict_quantiles` should call `_predict` and `_predict_var`, i.e., the private functions, not the public ones.

This could also become a default to implement `_predict_quantiles` if `_predict_var` is implemented, but perhaps let's do that in a separate PR.

topher-lo:
> `_predict_var` looks good!
> 
> Change request: `_predict_quantiles` should call `_predict` and `_predict_var`, i.e., the private functions, not the public ones.

@fkiraly I agree with your assessment about using the private methods in `_predict_quantiles`. However, this property is not consistent across the code base (e.g. in `NaiveVariance`: https://github.com/alan-turing-institute/sktime/pull/3435/files#diff-bfe91a3126b90a3d6721ec8f362fc927ce1bc439ce90a103b63965e765d3fdbaL483).

We simply followed precedent (i.e. `_predict_quantiles` in `NaiveVariance`) for `NaiveForecaster._predict_quantiles`.

> 
> This could also become a default to implement `_predict_quantiles` if `_predict_var` is implemented, but perhaps let's do that in a separate PR.

Agreed. I have a feeling (untested) that `_predict_quantiles` in `NaiveVariance` (and perhaps over forecasters that have `_predict_var` implemented) are not necessary because the `BaseForecaster`'s `predict_quantiles` public method supports intervals / quantiles under the normality assumption as long as `_predict_var` is implemented.

My suggestion:
1. Merge this PR as is (i.e. with `predict_var` and `predict` public methods used inside `_predict_quantiles`): to keep this consistent with how `NaiveVariance` `_predict_quantiles` has been implemented.
2. Create a GitHub issue checklist to audit that ALL forecasters aren't using public "predict" methods inside private "predict" methods.
3. Create a GitHub issue checklist to remove `_predict_quantiles` from forecasters that satisfy the following:
  1. Have `_predict_var` implemented
  2. Assume uncorrelated residuals and normality

Thoughts?

fkiraly:
Your suggestion makes sense - because the inconsistency is already in there.

Generally, calling public methods of self from private ones is not a good idea, since it can upset state related to to-and-back conversion of input types. It seems less problematic in `predict` though, I cannot construct a failure case here, because calls to `predict` type methods do not change state, so this causes only unnecessary conversions and checks to run.

>My suggestion:
> 1. Merge this PR as is (i.e. with `predict_var` and `predict` public methods used inside `_predict_quantiles`): to keep this consistent with how `NaiveVariance` `_predict_quantiles` has been implemented.

Ok

>2. Create a GitHub issue checklist to audit that ALL forecasters aren't using public "predict" methods inside private "predict" methods.

I feel that might take too long. Is there a way to test for it, instead? We could use one of the logger classes, e.g., `MockUnivariateForecasterLogger`.
Either way, the first step should probably be an issue that highlights the need for testing (or manual checking, but I prefer testing).

> 3. Create a GitHub issue checklist to remove `_predict_quantiles` from forecasters that satisfy the following:
    1. Have `_predict_var` implemented
    2. Assume uncorrelated residuals and normality

Though, first we'd need to implement a default in the base class for the case where `_predict_var` is implemented but none of `_predict_quantiles` or `_predict_ interval`?
Default infill currently only goes the other way round if I read the base class correctly.

fkiraly (changes requested):
All great.

I consider the issue with the private/public call to internal methods resolved for this PR, since we will open an issue to resolve the inconsistency for all cases.

However, I noticed another thing where I would like to request a change:

the docstring needs to explain how the method produces proba forecasts, optimally with a reference. It should not be difficult since we can link to the Hyndman book?
The three different cases need to be explained clearly.

topher-lo:
@fkiraly Added an explaination of the predict variance / quantiles algorithm. 

> The three different cases need to be explained clearly.
Instead of writing out the equations (which can be found the FPP3 textbook), we added context to **why** (namely "small sample bias") these three cases have different formulas for turning residuals variance into forecast variance.

fkiraly (approved):
Explanation and reference are great, thanks! All green now.

fkiraly:
To answer your questions:

> Not sure if these methods extend to the multivariate case

It automatically does, per vectorization.
(not the most efficient way, but it does)

> In line 415, I used `np.squeeze(self.predict(fh=1))` to convert the result from `self.predict` into a scalar. Does this always work for all possible return signatures? I am not familar with all of sktime's acceptable datatypes. Does this need to be tested or is this behavior guaranteed according to some design contract (when `fh=1`)?

The current API contract ensures `predict` returns the same mtype that was passed in `fit`. Currently that means a univariate `pandas.DataFrame` (tested), a `pd.Series` (tested) or `np.ndarray` (not tested, I think, but probably does not break).

It is also not extension stable, if more input types are added, e.g., one where `squeeze` will for certain break.

Anyway, once you change to `_predict`, as we discussed, it will be `pd.Series` and the potential breakage will be fixed.

> R naive also has a "bootstrap" option for computing forecast variance and quantiles, but this functionality is already implemented in sktime's `BaggingForecaster`.

Yes, and it does not sound very naive - it is, in fact, a wrapper rather than an isolated forecaster.

PR #3441, opened 2022-09-17
[ENH] Predicted variance for `STLForecaster`

### LINKED ISSUE #3437: [ENH] Predict variance / quantiles for `STLForecaster`
- `STLForecaster` does not have `predict_var` or `predict_quantile` capabilities

Since the current implementation already assumes a model for the residuals, we could take the prediction intervals (if `forecaster_resid` has that capability) computed by `forecaster_resid` then add `y_pred` to it. The prediction intervals will **definiitely be too narrow** (as they do not incorporate model uncertainty from `forecaster_trend` and `forecaster_seasonal`), but my suggestion is consistent with the extremely strong assumptions already imposed with having three separate forecasters for seasonal, trend and residual components.

Alternatively, we could naively add prediction intervals from `forecaster_resid`, `forecaster_trend` and `forecaster_seasonal` together. However, I am not sure if additive separability applies for both point and interval predictions here. If it does, then this feature request could be trivial to implement. First add checks that all three underyling forecasters have "capacity:pred_var" or "capacity:pred_int", then just sum predicted intervals up.

1. Users could use wrap `STLForecaster` around `BaggingForecaster` or `NaiveVariance`. Both methods however rely on computing residuals across many windows to estimatethe forecast probablity distribution. This is likely too slow the most common use-case for `STLForecaster`: i.e. as a quick and dirty benchmark with naive forecasters underneath.
2. The strategy from [R forecast library's mstl forecaster](https://github.com/robjhyndman/forecast/blob/master/R/mstl.R#L312). According to the docs from the R library[^1], the developers first take the prediction intervals given by the forecaster on the deseasonalized component (i.e. trend + residuals), then they reseasonalize the prediction intervals ***using the last year of the seasonal component***. Decided against this as it is currently non-trivial to extend `STLForecaster` to use the same strategy as `R mstl`. `STLForecaster` in sktime fits three models for seasonal, trend, and residual components.

NOTE: Alternative 2. suggests that `inverse_transform` could be implemented in `STLTransformer` under similar assumptions: i.e. just repeat the most recent computed seasonality (up to some period e.g. 1 year).

[^1] I confirmed this in the R implemetation. Here is the relevant code block: https://github.com/robjhyndman/forecast/blob/839062c26afe6d1ca2eb485a65b6f4e265548a88/R/mstl.R#L312

**Additional context**

topher-lo:
#### Reference Issues/PRs
Closes #3437 

- Implements `_predict_var` for `STLForecaster`
- Adds conditional tagging logic under `__init__`
- Fixes missing docstrings, failed linting, and flaky tests of existing `Detrender` and `PolynomialTrendForecaster` unit tests.

No.

- Conditional logic under `def __init__` to set "capability:pred_int" and "capability:pred_var" as True.
- Docstring for `_predict_var`
 

#### Any other comments?

Had to make changes to a few existing flaky tests that failed linting.

##### For all contributions

### DIFF
--- a/sktime/forecasting/tests/test_trend.py
+++ b/sktime/forecasting/tests/test_trend.py
@@ -1,9 +1,9 @@
 #!/usr/bin/env python3 -u
 # -*- coding: utf-8 -*-
 # copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+"""Trend and STL forecaster tests."""
 
-__author__ = ["Markus Löning"]
-__all__ = ["get_expected_polynomial_coefs"]
+__author__ = ["Markus Löning", "topher-lo"]
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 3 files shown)


topher-lo:
@fkiraly 👋 It seems that the existing code in `test_trend.py` file (unrelated to the current enhancement) fails the code quality test (perhaps it was committed before pre-commit hooks weres setup). What is your Git workflow to deal with these errors? Many thanks!

fkiraly:
are you still working on this, @topher-lo ?

Seems like linting is failing.

topher-lo:
@fkiraly Done but I had to rewrite to a few existing tests (Detrender) in the same file. And remove a cross dependency between transformers detrender tests and this trend forecastet tests.

topher-lo:
Closing this as the implementation itself was straightforward...it's just existing code that seems to be failing linting. Will sort out linting first. Then revisit.

PR #3442, opened 2022-09-17
[BUG] fix `get_cutoff` for numpy format

fkiraly:
Fixes `get_cutoff` for `numpy` formats.

Was apparently not tested.

Todo: tests.


fkiraly:
didn't have time to write tests - better a fix without test than no fix :-(

PR #3443, opened 2022-09-18
[ENH] Added permutation param to pipelines

### LINKED ISSUE #3439: [ENH] Tuning order of transformers in pipelines as a hyper-parameter
We have currently three sequential pipeline classes `TransformedTragetForecaster`, `ForecastingPipeline` and `TransformerPipeline`. Having several transformers chained raises the question which is the best order (permutation) of the given transformers.
For example in the pipeline below, what is the best order of transformers?
```python
    >>> pipe = TransformedTargetForecaster(
        steps=[
    ...     ("outlier", HampelFilter()),
    ...     ("impute", Imputer()),
    ...     ("detrend", Detrender()),
    ...     ("deseason", Deseasonalizer()),
    ...     ("minmax", TabularToSeriesAdaptor(MinMaxScaler())),
    ...     ("difference", Differencer()),
    ...     ("forecast", NaiveForecaster(strategy="drift")),
    ... ])
```

To tune the permutations of transformers, we would need another parameter that allow to give several sortings of the steps.
A solution could look as follows:
```python
    >>> pipe = TransformedTargetForecaster(
        steps=[
    ...     ("outlier", HampelFilter()),
    ...     ("impute", Imputer()),
    ...     ("detrend", Detrender()),
    ...     ("deseason", Deseasonalizer()),
    ...     ("minmax", TabularToSeriesAdaptor(MinMaxScaler())),
    ...     ("difference", Differencer()),
    ...     ("forecast", NaiveForecaster(strategy="drift")),
    ... ],
    ... permutation=["difference", "outlier", "deseason", "detrend", "impute", "minmax", "forecast"],
    ... )
```
This could be implemented as a `Mixin` class so all sequential pipeline classes could inherit and implement it. To be discussed if `"forecast"` step should be part of `permutation` list or if it should just stay at position where it is given. The param `permutation` could also accept `permutation="all"` to try all `n!` combinations with `n` as number of given transformers.

I thought first of having a separate transformer composition for this, but now I think its actually best to integrate this into the existing pipeline classes.

**Additional context**
Grid search, tuning

@miraep8  would be interested about your feedback as you wrote the `MultiplexTransformer` which solves the 1:n selection of transformers.
Also pinging @benHeid and @fkiraly

aiwalter:
#### Reference Issues/PRs
closes #3439

### DIFF
--- a/sktime/forecasting/compose/_pipeline.py
+++ b/sktime/forecasting/compose/_pipeline.py
@@ -91,8 +91,8 @@ class _Pipeline(
             if not sorted([x[0] for x in estimator_tuples]) == sorted(permutation):
                 raise ValueError(
                     f"""Permutations must contain the same step names as the pipeline,
-                    found tuple names {[x[0] for x in estimator_tuples]} but got
-                    permutation {permutation}."""
+                    found tuple names {sorted([x[0] for x in estimator_tuples])} but got
+                    permutation {sorted(permutation)}."""
                 )
 
             # sort steps
--- a/sktime/forecasting/compose/_pipeline.py
+++ b/sktime/forecasting/compose/_pipeline.py
@@ -5,6 +5,7 @@
 __author__ = ["mloning", "aiwalter"]
 __all__ = ["TransformedTargetForecaster", "ForecastingPipeline", "ForecastX"]
 
+
 import pandas as pd
 
 from sktime.base import _HeterogenousMetaEstimator
@@ -27,7 +28,7 @@ class _Pipeline(
         """Get the index of the first forecaster in the list."""
         return self._get_pipeline_scitypes(estimators).index("forecaster")
 
-    def _check_steps(self, estimators, allow_postproc=False):
+    def _check_steps(self, estimators, allow_postproc=False, permutations=None):
         """Check Steps.
 
         Parameters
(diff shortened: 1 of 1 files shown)


fkiraly:
@aiwalter, I think the idea of a permutation transformer is great!

I do not agree design-wise to attach it to the heterogeneous meta-estimator, or to the pipelines themselves - I think it "has no business there" in the sense of separation of concerns, and would be opposed to overloading functionality in the pipeline classes.

I would suggest a wrapper design, se #3439.

fkiraly:
I'd really go about it the following way:

* inherit from `_DelegatedForecaster`
* init args are `estimator`, `permute_param_name` and `permutation`
* the entire logic sits in `__init__`, it passes on the parameter permuted, and copies over tags
* there is no other method, i.e., only `__init__`

aiwalter:
can I put it into `TransformerPipeline`? I dont want to create another pipeline just for this 🤔

fkiraly:
I'd rather not, since I think (a) that's too much responsibility - keep estimators modular, and (b) we would have to duplicate this for both pipelines. Why not the delegator idea which can be applied to any heteoregenous estimator?

aiwalter:
superseded by #3689

PR #3450, opened 2022-09-20
[MNT] Relax pandas dependency

jlopezpena:
Pandas 1.5 is out, this is a WIP just testing wether upgrading it breaks anything

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,7 @@ dependencies = [
     "deprecated>=1.2.13",
     "numba>=0.53",
     "numpy>=1.21.0,<1.23",
-    "pandas>=1.1.0,<1.5.0",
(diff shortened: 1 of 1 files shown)


fkiraly:
Looks like a confirmation that it was a good idea to put some defensive bounds on `pandas`...

image

Seems like things are breaking around the period/frequency index interface.

Was this not caught by deprecation messages? Does not seem so, how odd.

fkiraly:
FYI @khrapovs, this is `freq` related, some remaining instances, but not in the parts that you fixed as far as I see - kudos.

fkiraly:
I cannot reproduce the failures of the `Lag` tests on python 3.10, `pandas` 1.5

fkiraly:
Thanks @jlopezpena for pointing out the new pandas release!

#3457 should make sktime compatible.

PR #3458, opened 2022-09-22
[ENH] test for correct return type of `load_basic_motions`

fkiraly:
Under `pandas 1.5.0`, it seems that `load_basic_motions` returns a non-mtype-compliant object for `nested_univ` return, see https://github.com/alan-turing-institute/sktime/actions/runs/3101253841/jobs/5022506912.

This is not caught by a dedicated test but only by accident, because we did not test this.

Oddly, the test for the generic loader `_load_provided_datast` seems to pass.

To diagnose this further and to improve test coverage, I have added a test for `load_basic_motions`.

Contains bugfix #3463 as this change uncovers a bug in the case where `return_X_y=True` and scalar columns are present.

FYI @achieveordie, you worked on these a while ago - any good ideas? This seems to be the last issue with `pandas` 1.5.0.

### DIFF
--- a/sktime/classification/deep_learning/fcn.py
+++ b/sktime/classification/deep_learning/fcn.py
@@ -1,0 +1,173 @@
+# -*- coding: utf-8 -*-
+"""Fully Connected Neural Network (CNN) for classification."""
+
+__author__ = ["James-Large", "AurumnPegasus"]
+__all__ = ["FCNClassifier"]
+
+from sklearn.utils import check_random_state
+
+from sktime.classification.deep_learning.base import BaseDeepClassifier
+from sktime.networks.fcn import FCNNetwork
+from sktime.utils.validation._dependencies import _check_dl_dependencies
+
+_check_dl_dependencies(severity="warning")
+
+
+class FCNClassifier(BaseDeepClassifier):
+    """Fully Connected Neural Network (FCN), as described in [1].
+
+    Parameters
+    ----------
+    should inherited fields be listed here?
+    n_epochs       : int, default = 2000
+        the number of epochs to train the model
+    batch_size      : int, default = 16
+        the number of samples per gradient update.
+    random_state    : int or None, default=None
+        Seed for random number generation.
+    verbose         : boolean, default = False
+        whether to output extra information
+    loss            : string, default="mean_squared_error"
+        fit parameter for the keras model
+    optimizer       : keras.optimizer, default=keras.optimizers.Adam(),
+    metrics         : list of strings, default=["accuracy"],
+    activation      : string or a tf callable, default="sigmoid"
+        Activation function used in the output linear layer.
+        List of available activation functions:
+        https://keras.io/api/layers/activations/
+    use_bias        : boolean, default = True
+        whether the layer uses a bias vector.
+    optimizer       : keras.optimizers object, default = Adam(lr=0.01)
+        specify the optimizer and the learning rate to be used.
+
+    Notes
+    -----
+    .. [1] Zhao et. al, Convolutional neural networks for
+    time series classification, Journal of
+    Systems Engineering and Electronics, 28(1):2017.
+
+    Adapted from the implementation from Fawaz et. al
+    https://github.com/hfawaz/dl-4-tsc/blob/master/classifiers/cnn.py
+
+    Examples
+    --------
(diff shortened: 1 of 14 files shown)


TonyBagnall:
I ran into this yesterday with a new install of version sktime 0.13.2
image
for me it is happening when resampling, the experiment on the default resample works fine. load_basic_motions also merges the data, so it has to be something with combining dataframes. It does not happen with sktime version 0.12.0 (just verified this, and I have run through newarly all classifiers/datasets with 0.12). I will look closer now

TonyBagnall:
I cant reproduce precisely atm as an experiment is running on the machine where it occurs and its fine with pandas 1.4.2. I'll look closer but I suspect it is related to pd.concat, which is used in both _load_provided_dataset (when TRAIN or TEST are not specified)
        # extract data from corresponding indices
        train_instances = all_data.iloc[train_indices, :]
        test_instances = all_data.iloc[test_indices, :]
        train_labels = all_labels[train_indices]
        test_labels = all_labels[test_indices]
        # concat onto current data from previous loop iterations
        X_train = pd.concat([X_train, train_instances])
        X_test = pd.concat([X_test, test_instances])

 and in resample
        # take the first lot of instances for train, remainder for test
        train_indices = indices[0:num_instances]
        test_indices = indices[num_instances:]
        del indices  # just to make sure it's not used!
        # extract data from corresponding indices
        train_instances = all_data.iloc[train_indices, :]
        test_instances = all_data.iloc[test_indices, :]
        train_labels = all_labels[train_indices]
        test_labels = all_labels[test_indices]
        # concat onto current data from previous loop iterations
        X_train = pd.concat([X_train, train_instances])
        X_test = pd.concat([X_test, test_instances])

TonyBagnall:
I can reproduce this error with pandas 1.4.4 and sktime 13.1, works with version 13.0.  I'm not sure what changed in the datatypes module, but I believe all other code was the same between versions

```python
from sktime.datasets import load_basic_motions
from sktime.datasets import load_from_tsfile_to_dataframe
from sktime.utils.sampling import stratified_resample
from sktime.classification.kernel_based import RocketClassifier
X, y = load_basic_motions()
trX, trY = load_from_tsfile_to_dataframe(
    full_file_path_and_name="C:\Code\sktime\sktime\datasets\data\BasicMotions"
                            "\BasicMotions_TRAIN.ts")
testX, testY = load_from_tsfile_to_dataframe(
    full_file_path_and_name="C:\Code\sktime\sktime\datasets\data\BasicMotions"
                            "\BasicMotions_TEST.ts")

newTrainX, newTrainY, newTestX,newTestY = stratified_resample(
                trX, trY, testX, testY, 2)
from sktime.datatypes import check_is_mtype, MTYPE_LIST_PANEL
from sktime.datatypes._panel._registry import MTYPE_REGISTER_PANEL
print(MTYPE_LIST_PANEL)
print(MTYPE_REGISTER_PANEL)
cls = RocketClassifier()
cls.fit(trX, trY)
p1 = cls.predict(testX)
cls.fit(newTrainX, newTrainY)
p2 = cls.predict(newTestX)
cls.fit(X,y)
print(p1.shape)
print(p2.shape)
print(p3.shape)
```
working version, sktime 13.0
image


datatypes error thrown versions 13.1
image

TonyBagnall:
Internally, it seems a bit odd, since check_is_scitype is failing a standard dataframe. This is the one check_is_scitype passes

image

and this one fails! 

image

seems there is problem in check_is_scitype introduced between 13.0 and 13.1

fkiraly:
@TonyBagnall, can you elaborate what makes you think it is the checker and not the loader where something has changed? Both have dependencies that could have changed, and direct changes.

TonyBagnall:
> 

my thinking on why its the datatypes: the data is loaded correctly into a dataframe, it all looks fine before it is passed to the classifier (and internally (see debugger); the bug is caused by resampling an already loaded dataframe, and resampling has not to my knowledge changed; it seems unlikely the loaders would have changed much between 13.0 and 13.1. I'm afraid that I'm not sure what changed from 13.0 to 13.1, that would narrow it down

khrapovs (approved):

TonyBagnall:
bug recreated without a data loader https://github.com/alan-turing-institute/sktime/issues/3471

PR #3459, opened 2022-09-22
DIAGNOSTIC, DO NOT MERGE: `pandas` 1.5.0 fixes without `test_resizing`

fkiraly:
This is a diagnostic PR, please do not merge.

Current #3457 without:
* `test_resizing` that fails due to `load_basic_motions` output (?)
* new test of `load_basic_motions` from #3458

This PR allows to check whether there are any other failures from the `pandas` 1.5.0 - there could theoretically be, since we dispatch estimators across OS/py-version combinations.


fkiraly:
works without `test_resizing`, has served its purpose to diagnose

PR #3460, opened 2022-09-22
[BUG] Fix `ConformalIntervals` update does not update residuals_matrix

### LINKED ISSUE #3419: [BUG] `ConformalIntervals` update does not update residuals_matrix
**Describe the bug**

`ConformalIntervals` update only updates the forecaster, not the residuals_matrix for predicting intervals or quantiles.

**To Reproduce**

`_update` is defined as:

```
    def _update(self, y, X=None, update_params=True):
        self.forecaster_.update(y, X, update_params=update_params)
        return self
```

which is only updating the forecaster and no the residuals_matrix. Since the forecaster is already fitted when `update` is called, the residuals_matrix is set to the fitted one with:

```
  if self.fh_early_:
      residuals_matrix = self.residuals_matrix_
```

**Expected behavior**

The residuals_matrix should be updated with the new data.

bethrice44:
#### Reference Issues/PRs
Fixes #3419 

As described in the bug, the `update` routine does not update the `residuals_matrix` which means the conformal intervals will be incorrect on the updated data. This PR computes the sliding residuals on the remaining part of the `residuals_matrix` in the case of an update.

No

This PR adds new rows and new columns populated for the new rows, but does not extend the existing rows as it stands. Is this an acceptable approximation? The way it is used this will only matter where old rows are not extending the length of the forecasting horizon. Happy to explain this aspect in more detail if this doesn't make sense. 

##### For all contributions

##### For new estimators


sktime/forecasting/conformal.py
```diff
@@ -351,11 +370,23 @@ def _compute_sliding_residuals(self, y, X, forecaster, initial_window, sample_fr
...
+            if len(remaining_y_index) != len(full_y_index):
+                residuals_matrix.loc[
+                    self.residuals_matrix_.index, self.residuals_matrix_.columns
+                ] = self.residuals_matrix_
+            y_index = remaining_y_index
+        else:
+            y_index = full_y_index
 
-        if sample_frac:
+        if sample_frac and len(y_index.to_series().sample(frac=sample_frac)) > 2:
```
fkiraly, line 389: you are sampling twice, this can lead to the sample in the "if" satisfying the condition and the other sample not. So, should we not sample and then check the lenght of *that* sample?

fkiraly (changes requested):
Looks good overall!

Change requests or questions:
* does the change of `initial_window` require deprecation?
* there is probably an issue with the if-statement in line 389

sktime/forecasting/conformal.py
bethrice44: Oh yes - fixed now

bethrice44:
Fixed the first point - good spot.

On the second point, I don't think so

fkiraly:
> On the second point, I don't think so

That's the one where you said

> Oh yes - fixed now

so there no longer is a problem

fkiraly (approved):
Thanks!

PR #3462, opened 2022-09-22
[ENH] simulate feature for base class and statsmodels

### LINKED ISSUE #3461: [ENH] Forecast simulations
Proposing to add a new method in the base forecaster class called `simulate`. 
This method will return a multiindex m-type containing multiple forecast simulations, similar to the Bootstrapping transformers return type.

This is useful for multiple applications.
1. Getting multistep prediction intervals in cases where only the 1-step forecast variance is available. e.g. https://otexts.com/fpp3/ets-forecasting.html
2. Intervals of temporal aggregations of forecasts. https://otexts.com/fpp3/aggregates.html

A lot of packages that we have wrappers for already have this functionality: 
1. [statsmodels](https://www.statsmodels.org/dev/generated/statsmodels.tsa.exponential_smoothing.ets.ETSResults.simulate.html)
2. [prophet](https://github.com/facebook/prophet/blob/419ee68ac6b218b7a3f06a9ddaa6bc08041be1d6/python/prophet/forecaster.py#L1445)

For the above, we can interface this method directly.

For other models we can use our native `predict_proba` and sample from the one-step ahead distribution to create multiple sample paths.

example from a statsmodels protoype: 

```python
import numpy as np
y = load_airline()
forecaster = AutoETS()
forecaster.fit(y)

y_hat = forecaster.simulate(np.arange(1,12)) 

display(y_hat)
```

image

ltsaprounis:
#### Reference Issues/PRs

fixes #3461 

Prototype for simulate method in forecaster base class with a statsmodels implementation.

to do before ready for review:
- [x] capability tag `"capability:simulate"`
- [ ] make existing implementations API compatible:
    - [x] UnobservedComponents
    - [ ] Dynamic Factor
- [x] tests in test_all_forecasters.py
- [x] clean up the docs
- [ ] maybe add default logic using predict_proba? (I think I'll skip this for this PR if that's OK.)

##### For all contributions


ltsaprounis:
Results from simply adding the simulate method in the PREDICT_METHODS list in `BaseForecaster._vectorize`

```python
import numpy as np
import pandas as pd

y = load_airline().to_frame()
y_hier = (
    pd.concat([y.assign(ts_id="ts_1"), y.assign(ts_id="ts_2")])
    .set_index("ts_id", drop=True, append=True)
    .reorder_levels(["ts_id", "Period"])
)

forecaster = AutoETS()
forecaster.fit(y_hier)

y_hat = forecaster.simulate(np.arange(1,12)) 

display(y_hat)
```
image

still WIP, but looking promising!

ltsaprounis:
Look at all these tests failing in the CI!

image

ltsaprounis:
HA! some estimators already have a simulate method!
1. [UnobservedComponents](https://github.com/sktime/sktime/blob/29209ecabce57e3586ee0a7121b7c19101f2f56b/sktime/forecasting/structural.py)
2. [DynamicFactor](https://github.com/sktime/sktime/blob/29209ecabce57e3586ee0a7121b7c19101f2f56b/sktime/forecasting/dynamic_factor.py)

ltsaprounis:
This error is unrelated and really strange...
[FAILED sktime/transformations/series/tests/test_clear_sky.py::test_clearsky_trafo_vals - urllib.error.HTTPError: HTTP Error 503: SERVICE UNAVAILABLE](https://github.com/sktime/sktime/actions/runs/3344084474/jobs/5538139324#step:7:1689)

ltsaprounis:
There's an issue with the statsmodels prediction index for integer index not starting from 0. 
I've raised an issue and PR there to fix it upstream (https://github.com/statsmodels/statsmodels/pull/8488), but might add a temp fix here with a warning to make this feature available here a bit sooner and not high lower bound on statsmodels only because of this.

ltsaprounis:
image

fkiraly (changes requested):
[removed earlier version since not a real review]

miraep8:
@fkiraly  - dismissing previous review as I don't think that it was pertinent to this PR. @ltsaprounis I would be happy to review this PR when it is ready for review.  See it is still marked as a draft PR, but feel free to ping me when ready to be checked out!

fkiraly (changes requested):
Sure.

* I would like a discussion of one point (here or in a STEP, as this is a base class interface): Why should this be part of the base class? It seems like most examples in this proof-of-concept have extra parameteres that are specific for the simulator and are not used by any other method. That indicates we should compare this to an alternative design where this is not a method of a forecaster, but we have a simulator/sample base class, that a forecaster can return when fitted? This would also allow simulators that are not coming from forecasters, e.g., Brownian motion with fixed parameters.
* we should also test the vectorized case(s), I would do that separate from the test suite only once or a few time for selected forecasters, like the tests in `test_base` for vectorized forecasts.

ltsaprounis:
I will no longer work on this project. See #4152

fkiraly:
seems useful as a basis for a simulator interface, I'll reopen as a draft

Issue #3469, opened 2022-09-22
[BUG] bug in `NaiveForecaster` variance formulae?


fkiraly:
From https://github.com/alan-turing-institute/sktime/pull/3435/files which I've merged and only now noticed potential issues in the formula for `predict_var`.

The main issue (which I missed on first read) was that I think we need rolling forecasts, no?

* should the `mean` strategy not have rolling mean predictions used, instead of the single prediction at the `cutoff`? So, `y.rolling(window_length).mean` instead of y_pred? With an appropriate shift per `fh`?
* `last` seems correct only for `fh` indices that are smaller or equal to `sp`.
* the `slope` strategy should use `window_length` instead of `T`, and also do this on rolling basis?

FYI @topher-lo

fkiraly:
So, for instance, for `"last"`:

```python
if self.strategy == "last":
    for h in fh:  # this may need conversion to integer index first
        sp = self.sp
        numsp = h // sp
        y_res[h] = y - y.shift(numsp * sp)
```

topher-lo:
@fkiraly An oversight on our part as my team haven't used `window_length` in our existing codebase and I don't think it's an argument in the R implementation either. We will sort this issue out **before the end of the day.** Seems relatively straightforward imo.

1. `mean`: Use pandas rolling mean
2. `last`: This is interesting. I never thought of it this way that `fh > sp` should cycle through the most recent `sp` observations. We will add this in as well! 
3. `drift`: We can make it rolling 👍

One last point, I think `last` and `drift` rolling operations can be expressed purely in matrix notation. We will implement this in a fully vectorized way without the need for for loops.

Apologies for the oversight!

jasmineliaw:
For `last` method with fh > sp, we compared the result from our implementation with R and both produced the same result.

Refer to screenshots below for results from R and our implementation respectively, tested using sp=24 and fh=30.

- From R:
![image](https://user-images.githubusercontent.com/93530957/191888350-b30dfeb4-7369-4bb9-9f26-37f5a0727230.png)

- From sktime:
![image](https://user-images.githubusercontent.com/93530957/191889415-6892fbc9-fea6-42d7-b46c-628d291aad23.png)

To reproduce, please refer to below:

- From R:
```
library(fpp3)
library(tsibbledata)

# export y to csv from sktime example below
data <- read.csv("y.csv")
[...]

PR #3470, opened 2022-09-22
[MNT] turn off `pred_var` in `NaiveForecaster`

fkiraly:
Temporarily turns off probabilistic predictions in `NaiveForecaster` until #3469 is resolved.


fkiraly:
breaks since we check consistency. So we should leave or revert - for now, leave, as per discussion in #3469.

PR #3475, opened 2022-09-24
[BUG] Fix side-effect of predict_residuals

### LINKED ISSUE #3473: [BUG] Calling predict_residuals() is writing in-sample index to self.fh which should not happen
**Describe the bug**
Calling `forecaster:predict_residuals()` is setting the `forecaster.fh` to the in-sample index. This should not happen.

**To Reproduce**

```python
from sktime.performance_metrics.forecasting import mean_absolute_percentage_error
from pandas.testing import assert_series_equal

y = load_airline()
y_train, y_test = temporal_train_test_split(y)
forecaster = NaiveForecaster(sp=12)
forecaster.fit(y_train, fh=fh)

y_pred_1 = forecaster.predict()
print(forecaster.fh)
y_resid = forecaster.predict_residuals()
print(forecaster.fh)
y_pred_2 = forecaster.predict()
assert_series_equal(y_pred_1, y_pred_2)
```

**Expected behavior**
Above example should work.

**Additional context**

**Versions**
<details>

</details>

aiwalter:
#### Reference Issues/PRs
closes #3473

### DIFF
--- a/sktime/forecasting/base/_base.py
+++ b/sktime/forecasting/base/_base.py
@@ -1064,29 +1064,33 @@ class BaseForecaster(BaseEstimator):
                 Series, Panel, Hierarchical scitype, same format (see above)
         """
         self.check_is_fitted()
+
+        # clone self to avoid any side-effects to self
+        self_copy = deepcopy(self)
+
         # if no y is passed, the so far observed y is used
         if y is None:
-            y = self._y
+            y = self_copy._y
 
         # we want residuals, so fh must be the index of y
         # if data frame: take directly from y
-        # to avoid issues with _set_fh, we convert to relative if self.fh is
+        # to avoid issues with _set_fh, we convert to relative if self_copy.fh is
         if isinstance(y, (pd.DataFrame, pd.Series)):
-            fh = ForecastingHorizon(y.index, is_relative=False, freq=self._cutoff)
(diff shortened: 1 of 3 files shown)


fkiraly (changes requested):
Happy with the idea in general.

Some change requests:
* I think this will probably fail if an `fh` has already been passed to `fit` in a case where it does not need to be
* is there a way to do this without a copy of self? That can be large. Can we get away with just avoiding steps that write `fh` to self, e.g., `_check_fh`?
* we should also add a test for the expected logic. If we think `fh` being written is a bug, we should add a test in `test_base` that pins down the behaviour we want.
* the docstring should clearly mention what is written to self or not, like in `fit` or `predict`.

aiwalter:
> we should also add a test for the expected logic. If we think fh being written is a bug, we should add a test in test_base that pins down the behaviour we want.

I added a test?

aiwalter:
> I think this will probably fail if an fh has already been passed to fit in a case where it does not need to be

This case is covered by the test.

> is there a way to do this without a copy of self? That can be large. Can we get away with just avoiding steps that write fh to self, e.g., _check_fh?

I copy now only `fh`, that will be faster

> the docstring should clearly mention what is written to self or not, like in fit or predict.

Done

lmmentel (approved):
LGTM!

PR #3490, opened 2022-09-26
Revert "[DOC] Update README.md"

aiwalter:
Reverts alan-turing-institute/sktime#3480


aiwalter:
there was no link changed only package name, which was a wrong change

fkiraly:
but the link works now. `fbprophet` is just the link alias, and that is how it´s referenced, too?

fkiraly:
@aiwalter, I'm confused - wouldn't the revert break the link again?

fkiraly:
I tested the link and it works now. It would break with the revert.
No comment was posted since a month.

I therefore assume this revert was made in error.

PR #3495, opened 2022-09-28
[BUG] Fix residuals formula in NaiveForecaster predict_var for non-null window_length

### LINKED ISSUE #3469: [BUG] bug in `NaiveForecaster` variance formulae?

topher-lo:
#### Reference Issues/PRs
Fixes #3469 

The current `_predict_var` implementation for `NaiveForecaster` does not deal with "mean" strategies with `sp > 1` or `window_length != None`. Moreover, the "drift" strategy does not account for the fact that the "starting" point for the line is set to `y[-window_length]` if `window_length` is not null.

This PR fixes both issues by computing residuals for rolling means (seasonal and non-seasonal) and setting the starting point of the drift line to `-(window_length or len(y))`

No.

- Logic for computing residuals for rolling means and rolling seasonal means
- Prior art (R fable): https://github.com/sktime/sktime/issues/3469#issuecomment-1256862149

#### Any other comments?
Please note the debate on computing residuals across rolling windows vs computing residuals against a single y hat array (#3469).

##### For all contributions

### DIFF
--- a/sktime/forecasting/naive.py
+++ b/sktime/forecasting/naive.py
@@ -436,8 +436,10 @@ class NaiveForecaster(_BaseWindowForecaster):
                     # T / self.sp times to match the length of trained y
                     # NOTE: +1 extra tile to defend against off-by-one errors
                     reps = math.ceil(T / sp) + 1
-                    seasonal_means = self.predict(fh=list(range(1, sp + 1))).to_numpy()
-                    y_pred = np.tile(seasonal_means, reps)[0:T]
+                    seasonal_means = self.predict(fh=list(range(1, sp + 1)))
+                    if isinstance(seasonal_means, pd.DataFrame):
+                        seasonal_means = seasonal_means.squeeze()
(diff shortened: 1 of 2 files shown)


fkiraly (approved):
Thanks, looks ok to deal with NA.

We need to look in more detail at the root issue though (wrong formula here as well as in R?), but that does not impact what this PR fixes.

PR #3501, opened 2022-09-28
Dlforecast

### LINKED ISSUE #3500: [ENH] Adding DL Forecasters to sktime
Adding DL Forecasters to sktime around the lines of how Classifiers and Regressor were migrated from sktime-dl. In this case, since there is no pre-existing code base for forecasters, we will have to create DL Forecasters roughly following the structure of other DL Estimators. 

The solution would involve 2 major parts:
1.`BaseDeepForecaster`: like `BaseDeepClassifer` and `BaseDeepRegressor`, the `BaseDeepForecaster` would derive from `BaseForecaster` and have functions specific to DL Forecasters (which will be its child classes)
2. `FCNForecaster`: it would be the simplest forecaster to create since the network for it already exists (`FCNNetwork`) and it is the least complex forecaster

**Additional context**
Some good resources related to it which can be referred to:
1. [Tensorflow guide to Time Series Forecasting](https://www.tensorflow.org/tutorials/structured_data/time_series)
2. [Generating Sequences wtih RNN by Alex Graves](https://arxiv.org/pdf/1308.0850.pdf)

AurumnPegasus:
#### Reference Issues/PRs

Fixes #3500 

Currently, I have implemented a basic `FCNForecaster` using keras in the notebook `RNN_Forecaster.ipynb`. Based on the structure used in the trial code, the plan is to divide this into a cleaner class structure having `BaseDeepForecaster` and `FCNForecaster`. It might also require us to add a class called `WindowGenerator` to get forecasting values in a sliding window fashion. 

No

Since currently the PR is in the thinking stage, I would appreciate any help/points related to structure/design of the forecaster. Also, I would like to know opinions/feedback on the idea of adding WindowGenerator (or if its needed at all)

##### For all contributions


AurumnPegasus:
Current Status:
Have created and checked `FCNForecaster`, it works currently for univariate data, but will work for multivariate as well imo (need to just test this locally once).
I also need to add `predict` method to the forecaster, which will be iterative in nature. For consistency across forecasters, I think I will need to use the `ForecastHorizon` for it.
I have not started creation/design of `BaseForecaster` yet, since I wanted to complete the forecaster before deciding which functions will be redundant.

AurumnPegasus:
Status Update:
So there are a couple of issues that I am facing
technically, the fit and predict work as required, but the problem is when I try to give it very basic data for forecasting, it gives out horrible results. Now it might be possible that its due to the choice of DL forecaster itself (since the one I am implementing right now is not benchmarked against anything) but I highly doubt that.

I tried with a simple input where `y=[10, 20, 30, 40, 50, 60, 70, 80, 90]` and `steps=3`, in case someone wants to try this out.
@TonyBagnall maybe you could have a look at the code and see if I am not making any logical silly error? I have been on it for some time now and am not able to comprehend where the issue is

Another information which might be useful, the loss seems to decrease initially but then seems to stagnate:

```
model.history.history['loss'][-10:]: [4387.83056640625,
                                          4387.85107421875,
                                          4387.830078125,
                                          4387.86328125,
                                          4387.84130859375,
                                          4387.82666015625,
                                          4387.8291015625,
                                          4387.83056640625,
                                          4387.8310546875,
                                          4387.82958984375]

```

These are the last 10 loss updates for a 2000 epoch run, which are quite bad. 

Currently, I will try doing the same implementation on `MLPNetwork`, which are benchmarked, so that I can understand if it is due to the network itself or my implementation (although I do think my implementation is lacking somewhere)

AurumnPegasus:
Hey, so there is a design issue wrt forecasters and DL estimators I wanted to discuss:

In general, forecasters take input in `_fit` as `(y, fh=None, X=None)`, where `y` is the main data we need to use for predictions, and predict, and `X` is the exogeneous data which may or may not be provided.

But in DL Estimators, `y` represents the target, and `X` represents the input to the model.

There is an inherent difference in what each variable means in these cases. In my current implementation of this, I have tried to adhere to both norms as much as possible (in `mlp.py` I have taken input as expected of a forecaster, but then converted into a datatype which is more appropriate for a DL model as `source` and `target`), but once I try to adhere to the overall structure for `BaseDeep` classes, it breaks since it is still checking for `X` and `y` in the context of DL models. 

There is an implementation which directly works for now in the notebook, but for implementation which is consistent with the style, either we need to take a step and change how DL Forecasters work separately (not preferred) or we could create a function which does the conversion from forecasting type data to DL type data, and accept the DL type data as the input.

PR #3504, opened 2022-09-28
& canbooo [BUG] fix `ColumnEnsembleForecaster` for str index; extend `ColumnEnsembleForecaster` to allow application of multivariate forecasters

### LINKED ISSUE #3502: [BUG] ColumnEnsembleForecaster fails with `str` indices
**Describe the bug**
`ColumnEnsembleForecaster` raises an error, when defined over `str` indices (of a dataframe). Specifically [this](https://github.com/sktime/sktime/blob/4194aacaa105ef59807ea56c1bb3497b37eb1613/sktime/forecasting/compose/_column_ensemble.py#L397-L400) error is raised, since the logic there does not account for string indice.

**To Reproduce**

```python
>>> df = pd.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6]})
>>> fc = ColumnEnsembleForecaster([(f"trans_{col}", NaiveForecaster(), col) for col in "ab"])
>>> fc.fit(df)

ValueError: One estimator per column required. Found 2
```

**Expected behavior**
Being able to use `str` indices as describe in the [docstring](https://github.com/sktime/sktime/blob/4194aacaa105ef59807ea56c1bb3497b37eb1613/sktime/forecasting/compose/_column_ensemble.py#L23-L24)
>     forecasters : sktime forecaster, or list of tuples (str, estimator, int or str)
        if tuples, with name = str, estimator is forecaster, index as str or int

**Additional context**
I suspect that the same error will be produced with other `ColumnEnsemble...` objects but I did not check.

**Versions**
<details>
System:
    python: 3.9.5 (default, Nov 23 2021, 15:27:38)  [GCC 9.3.0]
executable: .../bin/python
   machine: Linux-5.4.0-1084-aws-x86_64-with-glibc2.31

Python dependencies:
          pip: 21.2.4
   setuptools: 58.0.4
      sklearn: 0.24.2
       sktime: 0.13.4
  statsmodels: 0.12.2
        numpy: 1.20.3
        scipy: 1.7.1
       pandas: 1.3.4
   matplotlib: 3.4.3
       joblib: 1.0.1
        numba: 0.56.2
     pmdarima: None
      tsfresh: None

</details>

fkiraly:
Fixes #3502, `ColumnEnsembleForecaster` failed for str indices.

This was a little involved, as it seems that string indices were not properly parsed at all.

Also adds a test based on @canbooo's example

Also extends `ColumnEnsembleForecaster` so that multiple columns can be passed, in which case a forecaster is applied as multivariate.

### DIFF
--- a/sktime/forecasting/compose/_column_ensemble.py
+++ b/sktime/forecasting/compose/_column_ensemble.py
@@ -62,7 +62,8 @@ class ColumnEnsembleForecaster(_HeterogenousEnsembleForecaster):
     ...     [("foo", NaiveForecaster(), "a"), ("bar", NaiveForecaster(), "b")]
     ... )
     >>> fc.fit(df, fh=[1, 42])
-    >>> fc.predict()
+    ColumnEnsembleForecaster(...)
+    >>> y_pred = fc.predict()
 
     Applying one forecaster to multiple columns, multivariate:
     >>>     df = pd.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6], "c": [7, 8, 9]})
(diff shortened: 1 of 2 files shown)


GuzalBulatova:
Looked briefly, all seems good, but I'd like to come back to it. Consider approved if I don’t come back by the end of this weekend / Martin approves it.

fkiraly:
end of week-end, so considered as approval

PR #3511, opened 2022-10-01
[EHN] Parallelized `evaluate` with {joblib, dask}

topher-lo:
- Parallelizes `evaluate` with two possible backends: `joblib` and `dask`
- Adds flag `compute` which is set to True returns a lazy dask DataFrame (if backend is set to "dask")

Soft dependencies: `dask>=2022.8.0` and `distributed>=2022.8.0`

- To parallelise the loop, I had to move the logic executed within each loop into a separate function: `_evaluate_window`
- To construct a Dask DataFrame from a list of lazy Pandas DataFrames (i.e. each result from `_evaluate_window`), we have to specify the dtypes of the Dask DataFrame beforehand (i.e. the `meta` argument in `dd.from_delayed` (line 283)). However, the mixed type used for `cutoff` and `y_pred` (i.e. `np.nan` to represent null values and `pd.Period` / `pd.Series` otherwise) is not recognised by Dask (see screenshot below). Therefore, I had to change the null value for `cutoff` / `y_pred` to the Pandas nullable `pd.NA` and pass in `cutoff_dtype` into each `_evaluate_window` call (to coerce the `cutoff` column dtype).
image

- Updated schema of `result` DataFrame:
```python
{
  score_name: "float",
  "fit_time": "float",
  "pred_time": "float",
  "len_train_window": "int",
  "cutoff": str(y.index.dtype),  (with `pd.NA` as the null representation if `dtype != PeriodDtype` else `pd.Period(pd.NaT)`)
  "y_train": "object",   (with `pd.NA` as the null representation)
  "y_test": "object",   (with `pd.NA` as the null representation)
  "y_pred": "object",   (with `pd.NA` as the null representation)
}
```
- Dask unfriendiness with lazy dataframes with mixed dtype columns aside, imo it is more reasonable to set `y_train`, `y_test`, and `y_pred` as an object dtype (with `pd.NA` set if `return_data = False`) as opposed to the existing default of `np.nan`.

#### Any other comments?
Screenshot of dask dashboard:
Screenshot 2022-09-27 005652

- I replaced `_subset_keep_freq` with defensive freq reassignment for both single and multi-index.
- Removed isinstance checks with `y.index.nlevels` (faster).
- Moved the fh selection logic into it's own function `_select_fh_from_y`
- I replaced a few conditionals with `try except`, which is more Pythonic (i.e. ask forgiveness not permission).

##### For all contributions
- [x] I've added unit tests and made sure they pass locally. (Yes tested all new combinations of backend)


topher-lo:
@fkiraly What''s the protocol for adding soft dependencies? Should `dask` be added to a separate dependency group e.g. `distributed`?

sktime/forecasting/model_evaluation/_functions.py
```diff
@@ -62,10 +165,18 @@ def evaluate(
         Value to assign to the score if an exception occurs in estimator fitting. If set
         to "raise", the exception is raised. If a numeric value is given,
         FitFailedWarning is raised.
+    backend : {"dask", "loky", "multiprocessing", "threading"}, by default None.
+        Runs parallel evaluate if specified.
+    compute : bool, default=True
+        Only applied if backend is to set "dask". If set to True, returns
```
topher-lo, line 171:
```suggestion
        Only applied if backend is to set "dask". If set to True, returns `dd.DataFrame`
```

fkiraly:
protocol for soft dependencies here:

Generally, all tests must pass without soft dependencies installed, or be decorated with skipif or internal check that avoids the import if not present.

fkiraly:
https://github.com/sktime/sktime/pull/3542 is now merged (unlocking dependencies in grid search and reduction) - let me know if I should help integrate.

topher-lo:
@fkiraly All tests are passing except 1 "doctest" (https://github.com/sktime/sktime/actions/runs/3258070873/jobs/5349886697). Haven't seen this before and the code quality check / pre-commits didn't pick this one up. Can you help out with this?

After that, I think this PR is good to go!

fkiraly:
excellent - looks good to go?

topher-lo:
@fkiraly Good to go. We should run benchmarks with the M3 / M4 datasets and this parallelized evaluate. I can help turn this into a blog post as well.

chillerobscuro:
When starting fresh with a new copy of this branch, I'm getting a circular import error. I can run the following code only if I comment out line 19 from sktime/forecasting/model_selection/_tune.py, but that presumably limits functionality from _tune.py. (Commented out line is `from sktime.forecasting.model_evaluation import evaluate`)
```

y = load_airline()
forecaster = NaiveForecaster(strategy="mean", sp=12)
cv = ExpandingWindowSplitter(initial_window=12, step_length=3, fh=[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12])
results = evaluate(forecaster=forecaster, y=y, cv=cv)
ImportError                               Traceback (most recent call last)
<ipython-input-1-b8e7e546881f> in <module>
      1 from sktime.datasets import load_airline
      2 
----> 3 from sktime.forecasting.model_evaluation import evaluate
      4 
      5 from sktime.forecasting.model_selection import ExpandingWindowSplitter

~/Code/opensource/sktime/sktime/forecasting/model_evaluation/__init__.py in <module>
      7 __all__ = ["evaluate"]
      8 
----> 9 from sktime.forecasting.model_evaluation._functions import evaluate

~/Code/opensource/sktime/sktime/forecasting/model_evaluation/_functions.py in <module>
     18 from sktime.exceptions import FitFailedWarning
     19 from sktime.forecasting.base import BaseForecaster, ForecastingHorizon
---> 20 from sktime.forecasting.model_selection._split import BaseSplitter
     21 from sktime.utils.validation.forecasting import check_cv, check_scoring
     22 

~/Code/opensource/sktime/sktime/forecasting/model_selection/__init__.py in <module>
     20 from sktime.forecasting.model_selection._split import SlidingWindowSplitter
     21 from sktime.forecasting.model_selection._split import temporal_train_test_split
---> 22 from sktime.forecasting.model_selection._tune import ForecastingGridSearchCV
     23 from sktime.forecasting.model_selection._tune import ForecastingRandomizedSearchCV

~/Code/opensource/sktime/sktime/forecasting/model_selection/_tune.py in <module>
     17 from sktime.exceptions import NotFittedError
     18 from sktime.forecasting.base._delegate import _DelegatedForecaster
---> 19 from sktime.forecasting.model_evaluation import evaluate
     20 from sktime.utils.validation.forecasting import check_scoring
     21 

ImportError: cannot import name 'evaluate' from partially initialized module 'sktime.forecasting.model_evaluation' (most likely due to a circular import) (/Users/loganduffy/Code/opensource/sktime/sktime/forecasting/model_evaluation/__init__.py)
```

fkiraly:
I can confirm @chillerobscuro's bug - it happens on this branch, and not on `main` (python 3.8, windows).

I suggest we add it as a one-off test and see what happens?

fkiraly (changes requested):
Looks great!

We need to find out though why this bug is appearing.

fkiraly:
I think it is all the imports that this PR added in `model_evaluation._functions` for typing, especially

```python
from sktime.forecasting.model_selection._split import BaseSplitter
```

I think you asked a while ago why we do not use typing everywhere, @topher-lo? I forgot one reason - risk of circular imports.

Let's minimize imports that introduce unnecessary coupling (worst case, circular imports).

Can we get to having the same imports as before?

topher-lo:
Ahh it just be due to the typing. Thanks for the heads up here @fkiraly. I'm currently travelling as well. Will revert the changes first thing next week.

fkiraly (changes requested):
I wanted to merge for the release but noticed this was failing dependency isolation tests, then I noticed some other things. I think we need to delay this to 0.14.1, apologies.

* dependency `dask` was not isolated. Full test suite should run if `dask` is not present. Made some proposed changes in latest commit, kindly check whether you are happy with that.
* `"loky"` and `"multithreading"` options don't seem to do anything, this does not seem right. Probably should be passed to `joblib`?
* `dask` is not used in `"update"` mode, is that correct?
* docstring was incomplete, this should properly describe what happens for what argument. I tried to fill in but cannot, since I am not sure what should happen in the `"update"` case or `"multithreading"` case, i.e., if the current behaviour that can be reverse engineered is a bug or intentional.

topher-lo:
> * dependency `dask` was not isolated. Full test suite should run if `dask` is not present. Made some proposed changes in latest commit, kindly check whether you are happy with that.
👍 
> * `"loky"` and `"multithreading"` options don't seem to do anything, this does not seem right. Probably should be passed to `joblib`?
https://github.com/sktime/sktime/pull/3511/commits/4097494cff757e73ce3e434bb3db8699650e94c6
> * `dask` is not used in `"update"` mode, is that correct?
Correct. In fact, no parallelization joblib or dask is used in update mode (unit tests for update would fail otherwise).
> * docstring was incomplete, this should properly describe what happens for what argument. I tried to fill in but cannot, since I am not sure what should happen in the `"update"` case or `"multithreading"` case, i.e., if the current behaviour that can be reverse engineered is a bug or intentional.
https://github.com/sktime/sktime/pull/3511/commits/85083112d0e569eae6fddd7a085815426f4ba25e

fkiraly:
> Correct. In fact, no parallelization joblib or dask is used in update mode (unit tests for update would fail otherwise).

Hmmm, can you explain why that is the case?

fkiraly (dismissed):
Looks like all resolved, thanks!

topher-lo:
> > Correct. In fact, no parallelization joblib or dask is used in update mode (unit tests for update would fail otherwise).
> 
> Hmmm, can you explain why that is the case?

Because update must work sequentially (from one window to the next on the same estimator), but parallel execution will fit different windows on unfitted estimators simultaneously.

fkiraly:
oh, of course.

PR #3513, opened 2022-10-01
[ENH] `dask` conversion adapters for multi-indexed `pandas.DataFrame`

fkiraly:
Adds adapters for conversion between `pd.DataFrame` and `dask`.

For use in estimators or future converters to enable parallelism or compatibility with `dask` `DataFrame`.

The adapters are necessary to deal with `pandas.MultiIndex` which `dask` does not support, but is required for hierarchical and panel data.


fkiraly:
@topher-lo, would appreciate your opinion - is this a good mapping of multiindex onto `dask` structure?

topher-lo:
@fkiraly What workflows do you have in mind that would necessitate pandas-to-dask and vice versa conversion?

It is generally an anti-pattern to perform pandas (in memory) to dask (distributed) conversion. For distributed DataFrame computing, the best practice is to *never* have the full dataset loaded in RAM at any time. The workflow is to first lazy read pre-partitioned parquet files from blob storage (e.g. S3) into a Dask DataFrame (which has pointers to the filepaths in blob storage). Then the lazy Dask DataFrame is passed into some batch processing algorithm (e.g. distributed stochastic gradient descent) that transforms the data into some lower dimensional representation / approximation (e.g. coefficients in a linear regression). Only this lower dimensional representation should be stored in its entirety in RAM.

fkiraly:
>@fkiraly What workflows do you have in mind that would necessitate pandas-to-dask and vice versa conversion?

Use as an `sktime` data container.

Currently, all new data containers need to be registered in `datatypes` and be present with at least one converter.

So, if we want `sktime` to accept `dask` dataframes, we need to:

* register them in the mtype registry
* implement checkers
* implement converters

The last one seems trickiest, so that is what I have started with.

The open question was, how do we represent multi-indexed/hierarchical time series as a dask object.

> it is just many Pandas Dataframes stacked on top of each other

I do think I have understood that already.

topher-lo:
> > @fkiraly What workflows do you have in mind that would necessitate pandas-to-dask and vice versa conversion?
> 
> Use as an `sktime` data container.
> 
> Currently, all new data containers need to be registered in `datatypes` and be present with at least one converter.
> 
> So, if we want `sktime` to accept `dask` dataframes, we need to:
> 
> * register them in the mtype registry
> * implement checkers
> * implement converters
> 
> The last one seems trickiest, so that is what I have started with.
> 
Ah makes sense now. Thanks for the explaination. I'll take another look at this PR from this angle.

> The open question was, how do we represent multi-indexed/hierarchical time series as a dask object.
> 
> > it is just many Pandas Dataframes stacked on top of each other
> 
> I do think I have understood that already.
Yes indeed! The larger point here though (as discussed in our Slack DMs) is these Pandas DataFrames are stacked and then partitioned by divisions along the index. But this point, on second look, is not directly related to this PR, which as your mentioned seems to be a generic converter (regardless of the multi-index dtypes).

fkiraly:
@topher-lo 
This won't work in a distributed setting. When you run set_index on Dask DataFrame, dask partitions your data by the index (where each partition is a Pandas DataFrame). So in your case, if you have "time" instead of "instance / entity" as the index, then each partition is a cross-sectional table (i.e. the values across entities for a single timestamp). Dask parallelizes Pandas operations by running functions on partitions separately across available workers.
Therefore, for panel data, it is often more useful to let the index be the "instance / entity column", so that each partition is a time-series table (i.e. values across time for a single entity). I say it is more useful because the most common preprocessing for forecasting (e.g. lag operator, rolling transforms) work along the time dimension (per entity).

@fkiraly 
I'm talking about what the index is, not how we partition.
Maybe I do not understand what you mean by "dask partitions your data by the index"

@topher-lo 
https://docs.dask.org/en/stable/dataframe-design.html#partitions

@topher-lo 
https://stackoverflow.com/questions/69570717/dask-dataframe-can-set-index-put-a-single-index-into-multiple-partitions
Stack OverflowStack Overflow
[Dask dataframe: Can `set_index` put a single index into multiple partitions?](https://stackoverflow.com/questions/69570717/dask-dataframe-can-set-index-put-a-single-index-into-multiple-partitions)
Empirically it seems that whenever you set_index on a Dask dataframe, Dask will always put rows with equal indexes into a single partition, even if it results in wildly imbalanced partitions. Here ...
^*Empirically it seems that whenever you set_index on a Dask dataframe, Dask will always put rows with equal indexes into a single partition*

@topher-lo 
Note: this behavior seems to be an accidental feature, but either way, Dask partitions your DataFrame by "divisions" along your set index (https://docs.dask.org/en/stable/generated/dask.dataframe.DataFrame.repartition.html). For panel data, you want to divide your dataset by instances not time if you want to perform time-based rolling operations per entity.
So for Dask DataFrames, the index is directly related to how you intend to partition.

fkiraly:
summary: index should be sth that puts an instance in the same partition, and that is impacted by index choice - partitions are vertical by sorted index

fkiraly:
@fkiraly 
I see. So you want the index to be the lowest level instance index?
is it not a problem that this would lead to duplicate indices?

@topher-lo 
Indeed there would be duplicate indices, which would be a problem if we are optimizing for row retrieval, but it seems unproblematic if we are using these duplicate indices as partition keys for parallel transforms. (edited) 
Is there another concern that you have in mind?

fkiraly:
@topher-lo, how about the following solution: we move *all* index vectors from multiindex into the dataframe, including time.

We replace this by a string index, e.g., str coercion of the multiindex. Would that work? Or does dask have problems with string indices?

topher-lo:
Dask DataFrame works really well with string indices 👍

topher-lo:
Dask DataFrame has support for all vectorized `str` operations (https://docs.dask.org/en/stable/dataframe-api.html#string-accessor) that you would find in Pandas

topher-lo:
The surrogate key design is also great for downstream Dask work. E.g. you create surrogate keys in the following way:
```python
# Creates index e.g. ["2016-01-01$usa", "2016-01-02$usa", ..., "2016-01-30$fbr"]
df["__index__"] = df["time"].astype(str) + "$" + df["country"].astype(str)
```
You can then partition (after sorting the DataFrame along the index) with divisions ["2016-01-01$afg", "2016-01-01$ala", ..., "2016-01-01$zwe"]. The partitions will contain all timestamps for a single country as the index is sorted in alphanumeric order.

fkiraly:
> It looks like you are going for a "surrogate key" solution,

Yes, that's precisely what it is. Unique surrogate key, specifically a product key.

fkiraly:
@topher-lo, done, the conversion back/forth now uses the surrogate key (string)

PR #3515, opened 2022-10-01
[ENH] isolate `statsmodels` imports, part 2

fkiraly:
Isolates remaining `statsmodels` imports, continues #3445 (see description there).


sktime/_contrib/datasets/_single_problem_loaders.py
```diff
@@ -936,6 +936,9 @@ def load_macroeconomic():
     .. [3] Data Source: Bureau of Labor Statistics, U.S. Department of Labor;
           http://www.bls.gov/data/; accessed December 15, 2009.
     """
+    _check_soft_dependencies("statsmodels")
```
GuzalBulatova, line 939: Shouldn't we introduce `_check_soft_dependencies("statsmodels")` before every statsmodels import then? In #3445 most (maybe all) of the updated files don't have that.

GuzalBulatova (commented):
I don't like how these imports look, but I understand the why. Is this a statsmodels-specific thing?

sktime/_contrib/datasets/_single_problem_loaders.py
fkiraly: the trick is in the `python_dependencies` tag, which has the value `"statsmodels"`, in those cases.

If that is set, the check is called by the base class (which reads the tag). Single locus of responsibility, for lookup as well as dependency checking.

fkiraly: > Is this a statsmodels-specific thing?

No, generally we want ot isolate soft dependencies to inside classes or functions. Isolating statsmodels is a "good thing" imo, see #3445, even if it remains a core dependency.

GuzalBulatova: Ok, I understand, thanks for clarification, Franz

PR #3516, opened 2022-10-01
[MNT] 0.16.0 change/deprecation action - `statsmodels` as soft dependency

fkiraly:
Update: it was decided that we turn `statsmodels` into a soft dependency. This is now a change/deprecation action for 0.16.0.

---

Turning `statsmodels` into a soft dependency.

We should decide whether we should turn `statsmodels` into a soft dependency.

This PR shows that we could, if we wanted to - the two contained PR successfully isolate `statsmodels` in estimators, tests, and minor utility functions. Dependencies are located mostly inside specific estimators.

Depends on:
#3515
https://github.com/sktime/sktime/pull/3632
https://github.com/sktime/sktime/pull/3821
https://github.com/sktime/sktime/pull/4035


fkiraly:
(failures are due to `statsmodels` estimators being used in tests and as estimator test parameters)

ltsaprounis:
> (failures are due to `statsmodels` estimators being used in tests and as estimator test parameters)

In a lot of cases we should replace these with the MockForecaster.
This should speed up tests too!

fkiraly:
Hm, don't know, @ltsaprounis - the `BaggingForecaster` seems to fail at `MockForecaster`?

fkiraly:
from internal anon vote - removal of `statsmodels` as core dependency seems uncontroversial:

image

caveat: when we merge this, the next version will be 0.15.0

fkiraly:
hm, do we need deprecation?

I suppose we do, since people could have, at the same time:

* `pip install sktime` (no deps)
* `import SthWithStatsModels`

in their deployment scripts, so we ought to give them a warning.

PR #3518, opened 2022-10-01
[ENH] refactor/integrate `_contrib` - `datasets`

fkiraly:
Carries out part of https://github.com/sktime/sktime/issues/2942 - `datasets`.

* some `_config.datasets` contents were duplicates, these were removed
* some contents were *newer* versions with the `return_mtype` argument which @TonyBagnall et al never moved to the main module. These had a lot of duplicative code as they never underwent the refactor by @achieveordie. This has been addressed by refactoring the existing versions to have the additional argument
* there was also tests in the file `test_single_problem_loaders` which did not exist in the main repository - in fact, the `single_problem_loaders` were not tested in the main repo. The tests were moved to the main repo. They were also refactored, a lot of manual loops were replaced by `pytest.mark.parametrize`. These tests now cover the new argument `return_mtype` in the main module.
* docstrings were also corrected in the data loaders in the main module.


sktime/datasets/_data_io.py
```diff
@@ -135,7 +136,38 @@ def _list_available_datasets(extract_path):
 
 
 def _load_dataset(name, split, return_X_y, return_type=None, extract_path=None):
```
TonyBagnall, line 138: maybe this could become public? It is used in testing elsewhere I think

TonyBagnall (approved):

sktime/datasets/_data_io.py
fkiraly: well, that would not be a refactor/integration but a separate change.

Makes sense to me, but perhaps better to do this in a separate PR.

Issue #1788, opened 2021-12-21
[BUG] "update_params=False" for statsmodels adapter should call apply


raphaccf:
The current logic of  `update(update_params=False)` in the `statsmodels` interfaces is not correct, the model simply produces predictions it would have made before the update, instead of applying the fitted model to new data.

Instead, the `apply` interface from `statsmodels` should be used appropriately, see the discussion below. Previously, this was available only for a few models, but seems to be a general interface point now.

---
**Describe the bug**

After estimated a model, I would like to make forecast after my test set, but without updating parameters (changing the model parameters). This is specially important for comparison, moreover to forecast with new data without changing the model.
I have tried to set `update_params=False` in "update()"  and in "update_predict_single()", but it seems that is not working. Dates (timestamp) are updated, but forecast is the same for the previous period. I do not know If I am doing something wrong, or it is a bug. I have an example with auto.arima (code bellow and pdf attached).

Kind regards

**To Reproduce**

```
import pandas as pd
import numpy as np
from sktime.datasets import load_airline
from sktime.forecasting.arima import AutoARIMA
from sktime import show_versions 
show_versions()
```
```
## data split
y=load_airline() ## total observations
y1 = y.head(y.shape[0]-6)  #train observations
y2 = y.tail(6) # test observations
```
```
forecaster = AutoARIMA(sp=12, suppress_warnings=True,random_state=123)
z=forecaster.fit(y1)
AutoARIMA(...)
y_pred = forecaster.predict(fh=[1,2,3,4,5,6])
### model configuration and parameters for train
z.summary()
```
```
y_pred_semfit = forecaster.predict(fh=[1,2,3,4,5,6]) ## o cutoff 1960-06
y_pred_semfit
```
```
## using a function from stkime for not update_params, but to forecast using the test portion of the data set (to append to train y1)
y_pred = forecaster.update_predict_single(y_new=y2, update_params=False)
y_pred
### dates changed, but forecasted values are the same as above
```

```
### just changing the input of the function
y_pred = forecaster.update_predict_single(y=y, update_params=False)
y_pred
```

```
### another function from stkime, without update/retrain model parameters using update_params=False
updated_model =  forecaster.update(y=y, update_params=False)
updated_model.predict()
```
```
### however, if we set "update_params=True", time for command excution increase and forecasted values change
updated_model =  forecaster.update(y=y2, update_params=True)
updated_model.predict()
```

```
## but model parameters/configuration are different
updated_model.summary()
```

```
[...]

PR #3527, opened 2022-10-03
[BUG] Fixes 1788 - Add statsmodels append() to _update() in _StatsModelsAdapter

chillerobscuro:
#### Reference Issues/PRs

Fixes [1788](https://github.com/sktime/sktime/issues/1788)

When `update_params=False`, calling `update` with any `_StatsModelsAdapter` model does not work as expected. This PR changes behavior only when `update_params=False` and `self.is_composite=False`, and now uses the `statsmodels` `append` functionality to add new observed data without refitting the model.

No

The following outstanding questions
#### Any other comments?

1. For now if the user doesn't pass data with the correct index to `update()`, they will get the error message from `statsmodels` that `ValueError: Given endog does not have an index that extends the index of the model.` I could add functionality to check if the updated data intersects with the original data and only take the new subset, but could use advice on which data types we would need to handle (pandas only? also numpy? etc).
2. In the base `update()` method, if `self._is_vectorized=True`, then `_vectorize("update", ...)` is called instead of `_update()`. Do we need to add logic to handle this case?
3. Do we need to update the logic if both `self.is_composite=True` and `update_params=False`? This PR does not affect current behavior in the composite case.

##### For all contributions

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1760,17 +1760,6 @@
         "test"
       ]
     },
-    {
-      "login": "chillerobscuro",
-      "name": "Logan Duffy",
-      "avatar_url": "https://avatars.githubusercontent.com/u/5232872?v=4",
-      "profile": "https://github.com/chillerobscuro",
-      "contributions": [
-        "code",
-        "ideas",
-        "bug"
-      ]
-    },
   ],
   "projectName": "sktime",
   "projectOwner": "alan-turing-institute",
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1760,6 +1760,17 @@
         "test"
       ]
     },
+    {
+      "login": "chillerobscuro",
+      "name": "Logan Duffy",
+      "avatar_url": "https://avatars.githubusercontent.com/u/5232872?v=4",
+      "profile": "https://github.com/chillerobscuro",
+      "contributions": [
+        "code",
+        "ideas",
+        "bug"
+      ]
+    },
   ],
   "projectName": "sktime",
   "projectOwner": "alan-turing-institute",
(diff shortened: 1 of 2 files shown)


chillerobscuro:
The test failed as only the "state space" family of `statsmodels` have the `append` method. For the classes in our wrapper this means:
    Models with `append` support: [SARIMAX, DynamicFactor, VARMAX, UnobservedComponents]
    Models without: [ExponentialSmoothing, AutoETS, ETS, VAR, VECM]

`statsmodels` does have state space implementations of the Exponential smoothing models, with slightly limited performance and functionality (no multiplicative models). But our wrapper calls the dedicated ES model instead.

Seems worthwhile to implement the fix in this PR as it will improve the `update()` method of several models. We can also add warning to the rest of the models that there is no true update without a refit. If this sounds right, how should I check which model is calling `_update`? I can create a list like that above and compare with `type(self._fitted_forecaster)`, but there is likely a cleaner way.

fkiraly:
> If this sounds right, how should I check which model is calling _update?

The idea is to do this in `update`, not `_update`, see https://github.com/sktime/sktime/issues/1788, what I wrote:

>I would suggest to do the following in a separate PR (which you merge into this):
move the warning to update in BaseForecaster, and raise it only when the _update method is overridden. This can be checked by self._has_implementation_of("_update").

I would do this in a separate PR though.

chillerobscuro:
This fix now passes the tests and improves results of `update()` and `update_predict()`, for example:
```
from sktime.forecasting.sarimax import SARIMAX

y = load_airline()
ytrain, ytest = y[:110], y[110:125]

model = SARIMAX()
model.fit(ytrain)
test_data_to_plot = model.update_predict(ytest, update_params=False)
```
image ---> image

`update()` was fixed with the first commit but to fix the behavior for `update_predict()`, this code now takes the set difference of the data passed in and the data the model has already seen using `pandas.Index.difference()`. This code only affects models within the `_StatsModelsAdapter` class, when `update_params=False`.

fkiraly (approved):
Looks good, nicely done!

Issue #899, opened 2021-05-21
Unit testing for contracted classifiers and train estimates


MatthewMiddlehurst:
Pre-empting this somewhat, but over the summer we plan to expand and standardise contracting (#159) and generation of train estimates (#727) for classifiers. This would be through a standard parameter name i.e. _time_limit_in_minutes_ or method __get_train_probs_, though this hasn't been decided.

It would be good to have unit tests for these similar to the _n_jobs_ parameter for parallelisation when it gets a bit further along.

PR #3528, opened 2022-10-04
Classifier tests for capability:contractable and capability:train_estimate

MatthewMiddlehurst:
Adds generalised tests for classifiers which implement the `capability:train_estimate` and `capability:contractable` tags (#899, #3206).

The contracting test requires the classifier to implement a `time_limit_in_minutes` parameter for the classifier, and will use a "contracting" parameter set for the test if available. If the returned parameter set does not set `time_limit_in_minutes`  to 5 minutes or more, an error will be thrown. Low contract times lead to the errors found in #3206. None of the current tests should reach 5 minutes (or even 10 seconds in most cases) as they are configured to end early after a certain point (i.e. 2 trees built). 

The train estimate test ensures a `_get_train_probs` method is implemented, looks for a "train_estimate" parameter set, and finally tests that the output probabilities conform to what is expected (same as predict_proba). The parameter set is less important for this one, the default can be used.

Removes old tests and adds parameter sets for affected classifiers.


fkiraly (dismissed):
* I think we haven't fully consolidated the `train_estimate` capability. This seems to be related to a private interface point that does not have a public counterpart, which I think is bad design.
I would move or connect this with the functionality from `fit_predict`, and at least have a contract specification, e.g., which method needs to be present and do what.
* if you add parameter sets, they need to be described in the docstrings and the extension template
* why do we need extra parameter sets? Can we not use the default parameter sets, and set the additional parameter in the test, instead of putting part of the test logic into the parameter set?

MatthewMiddlehurst:
> I think we haven't fully consolidated the train_estimate capability. This seems to be related to a private interface point that does not have a public counterpart, which I think is bad design.
> I would move or connect this with the functionality from fit_predict, and at least have a contract specification, e.g., which method needs to be present and do what.

This is just to test what is currently there, as it is what all the classifiers which use the tag do at the moment. I don't mind discussing changes in an issue to change the expectations for the tag.

> if you add parameter sets, they need to be described in the docstrings and the extension template

Ok

> why do we need extra parameter sets? Can we not use the default parameter sets, and set the additional parameter in the test, instead of putting part of the test logic into the parameter set?

Parameters such as `contract_max_n_parameter_samples` and `save_transformed_data` are not universal in the interfaces, and are not required for the default parameter set.

MatthewMiddlehurst:
The parameter sets are now documented in the template and methods. Classifiers which dont implement _get_train_probs but set the tag to true just skip the test now.

fkiraly (changes requested):
* I think as it is currently done proliferates the interface contract on `get_test_params` too much. Instead I think we need to set the additional parameter, e.g., `time_limit_in_minutes`, in the test. This should also be consistent across estimators.
* the added interface assumptions should be documented in the classifier extension template
* may I also kindly ask you to write a short STEP on what precisely the interface assumption are, arising from the two tags?

MatthewMiddlehurst:
> I think as it is currently done proliferates the interface contract on get_test_params too much. Instead I think we need to set the additional parameter, e.g., time_limit_in_minutes, in the test. This should also be consistent across estimators.

Unfortunately, we have to give the contract tests a large contract time to avoid issues like #3206. We don't want these tests lasting 5 minutes each, however, so we reduce the time taken by setting other limits (i.e. `contract_max_n_estimators=2`). There is no static interface point for this (there is no guarantee the estimator has an `n_estimators` to limit or that it is being used for the contracting). While we could set `time_limit_in_minutes` in the test, more is required to prevent these tests taking much longer than they should. I personally don't see the issue with adding more test sets as long as they are documented.

> the added interface assumptions should be documented in the classifier extension template

> may I also kindly ask you to write a short STEP on what precisely the interface assumption are, arising from the two tags?

While I agree this should be done, I don't really want to change these interfaces in this PR. I can edit the contracting test so it does not fail when the `time_limit_in_minutes` argument or `get_test_params` set are missing, the goal is to test what is currently there (and replace the tests causing #3206).

PR #3531, opened 2022-10-04
[BUG] add get_test_params method to TapNet

### LINKED ISSUE #3525: [BUG] TapNet test failures and long test times
TapNet is currently failing the macOS 3.7 test (`Install and test / test-unix (3.7, macOS-11) (pull_request)`) in PR #3523. The PR makes no changes relevant to TapNet or DL estimators, and the test times are extremely high.

https://github.com/sktime/sktime/actions/runs/3174214839/jobs/5170841022

image

TonyBagnall:
#### Reference Issues/PRs

Fixes #3525 , the tests for TapNet were either timing out or taking a very long time. 

adds a get_test_params to the TapNet classifier. 

no

the parameter settings in the generator test. Ive tried these
{"n_epochs": 100, "batch_size": 2, "dilation": 2}
but would appreciate feedback/review from @achieveordie on this.

### DIFF
--- a/sktime/classification/deep_learning/tapnet.py
+++ b/sktime/classification/deep_learning/tapnet.py
@@ -230,3 +230,35 @@ class TapNetClassifier(BaseDeepClassifier):
         )
 
         return self
+
+
+@classmethod
+def get_test_params(cls, parameter_set="default"):
+    """Return testing parameter settings for the estimator.
+
+    Parameters
+    ----------
+    parameter_set : str, default="default"
+        Name of the set of test parameters to return, for use in tests. If no
+        special parameters are defined for a value, will return `"default"` set.
+        For classifiers, a "default" set of parameters should be provided for
+        general testing, and a "results_comparison" set for comparing against
+        previously recorded results if the general set does not produce suitable
+        probabilities to compare against.
+
+    Returns
+    -------
+    params : dict or list of dict, default={}
+        Parameters to create testing instances of the class.
+        Each dict are parameters to construct an "interesting" test instance, i.e.,
+        `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+        `create_test_instance` uses the first (or only) dictionary in `params`.
+    """
+    return {"n_epochs": 100, "batch_size": 2, "dilation": 2}
+
+
+# dropout = 0.5, \
+# filter_sizes = (256, 256, 128), \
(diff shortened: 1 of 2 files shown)


achieveordie (commented):
Hey sorry, wasn't able to look at this issue and thanks for taking this up. I'll give a more extensive review in a while after I look more into it but do you reckon we should do the same for `TapNetRegressor` as well since that estimator too was causing problems in that issue?

TonyBagnall:
hi  @achieveordie, I suggest we sort out classification test times first, seems to be taking four hours? I will try run them locally and take a closer look

sktime/classification/deep_learning/tapnet.py
```diff
@@ -230,3 +230,35 @@ def _fit(self, X, y):
...
+
+    Returns
+    -------
+    params : dict or list of dict, default={}
+        Parameters to create testing instances of the class.
+        Each dict are parameters to construct an "interesting" test instance, i.e.,
+        `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+        `create_test_instance` uses the first (or only) dictionary in `params`.
+    """
+    return {"n_epochs": 100, "batch_size": 2, "dilation": 2}
```
achieveordie, line 257: Wouldn't it make more sense to increase the `batch_size`? We could also try to decrease `n_epochs` to 50 and also decrease the default `filter_sizes` to maybe `(128, 128, 64)` & dense sizes (via `layers` to `(200, 100)`).

This network is bigger than others that are currently present here and these values come from paper's recommendation. As far as I understand, in these tests, we're not testing for convergence but for proper working. These values might not be optimal for convergence but should be good for testing.

achieveordie:
Even with the current values, it takes about 4 hours across different runners. Maybe that's because there's no `get_test_params()`  for `TapNetRegressor`? I don't see the 10 slowest runs so I can't say for sure, it's just speculation.

On a side note, should we also try to remove `self._is_fitted` in the constructor of both classes? I see that there's a property in the superclass referring to it already so I think while we're refactoring this, we could also discuss removing that line.

TonyBagnall:
hi @achieveordie yes you are correct about batch size! Feel free to make the changes you see fit to this branch, or I can do it if you prefer. 
_is_fitted is set in the base class fit method so not required as you say, so feel free to remove that line

TonyBagnall:
replaced by #3541

Issue #3429, opened 2022-09-14
[ENH] ensure that all estimators have two test parameter sets


fkiraly:
We should ensure that all estimators (that have parameters) possess at least two test parameter sets.

The two (or more) parameter sets should:

* be fast to run together - `fit` is the bottleneck (so we should not overdo it with too many parameter sets)
* cover substantially different settings for all the important parameters, i.e., substantially different typical cases and/or important edge cases

Recipe:

1. search for estimators which have parameters but only a single test parameter set. These are estimators with no `get_test_params` implemented, or `get_test_params` returning only a single dictionary instead of a list of two (or more) dictionaries.
2. come up with a parameter set satisfying the above constraints and add it to the return (should be list of two or more dictionaries)

An example PR that adds second parameter sets for some estimators can be found here: https://github.com/alan-turing-institute/sktime/pull/3428

Finding some estimators that have only one parameter set can also be done speedily by using this PR https://github.com/alan-turing-institute/sktime/pull/2862 which adds a test for two parameter sets - either run the test suite from the branch locally, or look into the failing CI.

Abelarm:
Hi @fkiraly 

I’d like to to tackle this if it’s ok for you.

fkiraly:
sure, @Abelarm - pick an estimator!

janpipek:
I am picking SARIMAX.

julia-kraus:
which estimators are still left?

fkiraly:
@julia-kraus, the failures in this diagnostic PR https://github.com/sktime/sktime/pull/2862 correspond to the ones that do have only one - it might not be 100% up to date, I'll restart it so it is:

fkiraly:
I would like to add the test params for the `HampelFilter`, that sounds interesting.

iaryangoyal:
I have worked on Rocket regression #6134

namita0210:
Hi , can I take up this issue for the estimator:  "TimeSeriesForestClassier"
@fkiraly

fkiraly:
@namita0210, absolutely! All yours!

MMTrooper:
Hi, I will try to tackle the MatrixProfileClassifier. @fkiraly

fkiraly:
great, thanks, @MMTrooper!

shankariraja:
Hi @fkiraly,

I'm currently working on adding new test parameter sets for the estimators identified in the issue. I'll be focusing on ``TimeSeriesKMeansTslearn``.
[...]

PR #3544, opened 2022-10-08
[BUG] fast test parameters for `TapNet` estimators and docstring/interface cleanup

### LINKED ISSUE #3525: [BUG] TapNet test failures and long test times

achieveordie:
#### Reference Issues/PRs

Fixes #3525

- Add `get_test_params()` for `TapNetRegressor` and `TapNetClassifier` with two parameter set.
- Remove the redundant `is_fitted` variable from TapNet estimators.
- Correct minor typos.

No

@TonyBagnall LMK if we should make any more changes here.

### DIFF
--- a/sktime/classification/deep_learning/tapnet.py
+++ b/sktime/classification/deep_learning/tapnet.py
@@ -134,8 +134,6 @@ class TapNetClassifier(BaseDeepClassifier):
         self.callbacks = callbacks
         self.verbose = verbose
 
-        self._is_fitted = False
-
         self.dropout = dropout
         self.use_lstm = use_lstm
         self.use_cnn = use_cnn
--- a/sktime/classification/deep_learning/tapnet.py
+++ b/sktime/classification/deep_learning/tapnet.py
@@ -253,10 +253,20 @@ class TapNetClassifier(BaseDeepClassifier):
             `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
             `create_test_instance` uses the first (or only) dictionary in `params`.
         """
-        return {
+        param1 = {
             "n_epochs": 50,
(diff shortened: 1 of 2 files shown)


fkiraly:
@achieveordie, every estimator should have at least 2 sets of test parameters, covering a good range of typical values.

achieveordie:
So should I change it to have a different set of parameters for a different value of `parameter_set`? Something like the one that I have for `default` and larger parameter values for a new one called ``slower`? Or should I just change the default values since these values are already there for `TapNetClassifier`? I'm not sure I understand.

fkiraly:
look at the example in `NaiveForecaster`.

The goal is to get a good decision coverage and parameter value coverage:
https://en.wikipedia.org/wiki/Code_coverage#Modified_condition/decision_coverage
https://en.wikipedia.org/wiki/Code_coverage#Parameter_value_coverage

TonyBagnall:
I think he wants you to define two parameter sets and return them as a list like this https://github.com/sktime/sktime/pull/3428/files

TonyBagnall:
https://github.com/sktime/sktime/issues/3429
could you do this for the tap net classifier too? (please make them small parameters!)

fkiraly:
yes, but not just any parameter sets, but with a good test coverage

achieveordie:
Alright, I understood, will do this in a while. Thanks!

PR #3542, opened 2022-10-07
[ENH] extend `evaluate` to hierarchical and panel data

fkiraly:
Carries out item from https://github.com/sktime/sktime/issues/2243.

Extends `evaluate` to accept hierarchical `X` and `y`.

By extension, should enable `GridSearchCV` to handle hierarchical data and resolve bug https://github.com/sktime/sktime/issues/3464.

Less hacky version of https://github.com/sktime/sktime/pull/3509 which uses the centralized checker functions and hierarchical functionality of splitters instead of custom duplicates.

FYI @aiwalter, @danbartl, @ngupta23

PR #3548, opened 2022-10-08
[ENH] extend forecasting grid/random search to hierarchical and panel data

fkiraly:
Carries out item from https://github.com/sktime/sktime/issues/2243.

Extends `ForecastingGridSearchCV` and `ForecastingRandomSearchCV` to work with hierarchical and panel type data.

Adds a test to check the hierarchical/panel logic.

Builds on and includes https://github.com/sktime/sktime/pull/3542.


aiwalter (approved):
can we get this in?

PR #3549, opened 2022-10-08
[BUG] Fix pipeline tag for NaN values

### LINKED ISSUE #3547: [BUG] Forecasting Pipeline containing an imputer transformer fails when the underlying forecaster can not handle missing data
**Describe the bug**

A forecasting pipeline that **contains an imputer** and a forecaster that **does not accept missing values** fails to work in the latest release of sktime (0.13.4). This used to work fine in 0.13.2 release.

**To Reproduce**

```python
# Works in 0.13.2, but fails in 0.13.4

import numpy as np
from sktime.forecasting.trend import TrendForecaster
from sktime.transformations.compose import TransformerPipeline
from sktime.forecasting.compose import ForecastingPipeline, TransformedTargetForecaster

y, X = load_longley()
y[10] = np.nan  # Add missing values


transformer_y = TransformerPipeline(steps = [("imputer", Imputer())])

forecaster = TransformedTargetForecaster(
    steps = [
        ("transformer_y", transformer_y),
        ("model", TrendForecaster())
    ]
)

pipe = ForecastingPipeline(steps = [("forecaster", forecaster)])

pipe.fit(y)
```

```python-traceback
[<ipython-input-18-6f4acc86cd81>](https://localhost:8080/#) in <module>
----> 1 pipe.fit(y)

2 frames
[/usr/local/lib/python3.7/dist-packages/sktime/forecasting/base/_base.py](https://localhost:8080/#) in fit(self, y, X, fh)
    286 
    287         # check and convert X/y
--> 288         X_inner, y_inner = self._check_X_y(X=X, y=y)
    289 
    290         # set internal X/y to the new X/y

[/usr/local/lib/python3.7/dist-packages/sktime/forecasting/base/_base.py](https://localhost:8080/#) in _check_X_y(self, X, y)
   1298                 )
   1299 
-> 1300             _check_missing(y_metadata, "y")
   1301 
   1302         else:

[/usr/local/lib/python3.7/dist-packages/sktime/forecasting/base/_base.py](https://localhost:8080/#) in _check_missing(metadata, obj_name)
   1244                     )
   1245                 if metadata["has_nans"]:
-> 1246                     raise ValueError(msg)
   1247 
   1248         # retrieve supported mtypes

ValueError: ForecastingPipeline cannot handle missing data (nans), but y passed contained missing data.
```

**Expected behavior**
Would have expected this to work because there is an imputer in the pipeline.

**Additional context**
This used to work fine up until the 0.13.2 release. This started breaking after the recent 0.13.4 release (maybe even 0.13.3, but I have not explicitly tested that). **The initial impact was observed in pycaret when the unit tests started failing on GitHub.**

Working Notebook in 0.13.2: https://gist.github.com/ngupta23/fe385efcad52b1d6c4c11f955d13da8e
Notebook with error in 0.13.4: https://gist.github.com/ngupta23/c35b23ae7b786faff2f020a36efcbb38

**Versions**
<details>

</details>

aiwalter:
closes #3547 

@ngupta23 please have a look

I removed the tag inheritance bcz if the forecaster cant handle missing values, it will raise an exception then anyways but it will work if an `Imputer` was used in the pipe.

### DIFF
--- a/sktime/forecasting/compose/_pipeline.py
+++ b/sktime/forecasting/compose/_pipeline.py
@@ -299,7 +299,7 @@ class ForecastingPipeline(_Pipeline):
         "X_inner_mtype": SUPPORTED_MTYPES,
         "ignores-exogeneous-X": False,
         "requires-fh-in-fit": False,
-        "handles-missing-data": False,
+        "handles-missing-data": True,
         "capability:pred_int": True,
         "X-y-must-have-same-index": False,
     }
@@ -311,7 +311,6 @@ class ForecastingPipeline(_Pipeline):
         tags_to_clone = [
             "ignores-exogeneous-X",  # does estimator ignore the exogeneous X?
             "capability:pred_int",  # can the estimator produce prediction intervals?
-            "handles-missing-data",  # can estimator handle missing data?
             "requires-fh-in-fit",  # is forecasting horizon already required in fit?
             "enforce_index_type",  # index type that needs to be enforced in X/y
         ]
@@ -713,7 +712,7 @@ class TransformedTargetForecaster(_Pipeline):
         "X_inner_mtype": SUPPORTED_MTYPES,
         "ignores-exogeneous-X": False,
         "requires-fh-in-fit": False,
-        "handles-missing-data": False,
+        "handles-missing-data": True,
         "capability:pred_int": True,
         "X-y-must-have-same-index": False,
     }
@@ -727,7 +726,6 @@ class TransformedTargetForecaster(_Pipeline):
         tags_to_clone = [
             "ignores-exogeneous-X",  # does estimator ignore the exogeneous X?
             "capability:pred_int",  # can the estimator produce prediction intervals?
-            "handles-missing-data",  # can estimator handle missing data?
             "requires-fh-in-fit",  # is forecasting horizon already required in fit?
             "enforce_index_type",  # index type that needs to be enforced in X/y
         ]
--- a/sktime/forecasting/compose/tests/test_pipeline.py
+++ b/sktime/forecasting/compose/tests/test_pipeline.py
@@ -38,6 +38,7 @@ from sktime.transformations.series.impute import Imputer
 from sktime.transformations.series.outlier_detection import HampelFilter
 from sktime.utils._testing.estimator_checks import _assert_array_almost_equal
 from sktime.utils._testing.series import _make_series
+from sktime.utils.estimators import MockForecaster
 
 
 def test_pipeline():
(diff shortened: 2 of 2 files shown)


ngupta23:
Thanks @aiwalter . I am not sure what the best solution would be in this case. The tag for handling missing data is set to True, but this assumes that there is an imputer in the pipeline (which the user must add). Also, even if there is an imputer, some subsequent transformation step may introduce NA values. The onus is on the user to make sure that the pipeline will impute missing values appropriately and and I am ok with that responsibility being put on the user.

fkiraly (changes requested):
Oops - I suppose that was a mistake since we indeed can have an imputer.

Do I understand correctly that this sets the "handles missing data" tag to `True` no matter what? It is a fix but the tag will be incorrect for some pipelines.

I don't mind as a temporary fix, but a more prinipled fix should be not too difficult:
The cleaner way would be using `_tagchain_is_linked_set` as in lines 208ff in `TransformerPipeline`.

fkiraly:
> Thanks @aiwalter . I am not sure what the best solution would be in this case. The tag for handling missing data is set to True, but this assumes that there is an imputer in the pipeline (which the user must add). Also, even if there is an imputer, some subsequent transformation step may introduce NA values. The onus is on the user to make sure that the pipeline will impute missing values appropriately and and I am ok with that responsibility being put on the user.

@ngupta23, we have a tag that says whether the transformer removes all NA. Indeed transformers may produce NA (and there is no tag for that), but at least we can check that the transformers preceding the first transformer removing NA can handle NA (see above).

fkiraly:
also, kindly add the (simplest possible version of) example from #3547 to `test_pipeline`

aiwalter:
@fkiraly I think adding test should not be part of this PR. Its a bigger topic to add generic tests for compositions. I dont want to add a test for just one particular case.

fkiraly:
> I dont want to add a test for just one particular case.

My recommendation: just copy-paste @ngupta23's code so we know we fixed it, and we don't regress.

We can think about more general tests later.

aiwalter:
I have tested it locally and it works now

fkiraly:
> I have tested it locally and it works now

That's not the same. We need to make sure that future changes don't break it again, that we can do only if we add it as a test case.

E.g., let's say we introduce the "advanced" handling of the tag, how would we know we didn't break @ngupta23's use case?

General best practice: if you fix a bug, include a test that "certifies" the fix.

fkiraly:
thanks, @aiwalter  :-)
@ngupta23's examples have saved us a number of times from breaking things in updates...

fkiraly (approved):
Thanks!

Issue #3512, opened 2022-10-01
[ENH] refactor datatypes mtype related functionality into classes


fkiraly:
There are multiple pieces of metadata and functions such as checkers and converteres floating around in the `datatypes` module.

Currently, these are loose objects, and with recent addition of soft dependency handling I think it has passed the threshold where it would make sense to bundle these in a class.

I think the right design is:

* `BaseObject` descendant
* one class for: mtype checks, metadata
* one class for conversions (from/to) or pair mtype, metadata
* perhaps one class for examples - or should they be with the mtype class?

PR #3551, opened 2022-10-08
[ENH] enforces row index of `pandas.DataFrame` based mtypes to be sorted and unique

fkiraly:
This adds two lines to the checkers for `pandas.DataFrame` based mtypes which enforce row indices to be lexsorted and unique.

### DIFF
--- a/sktime/datatypes/_hierarchical/_check.py
+++ b/sktime/datatypes/_hierarchical/_check.py
@@ -85,7 +85,15 @@ def check_pdmultiindex_hierarchical(obj, return_metadata=False, var_name="obj"):
         msg = f"{var_name} must have a MultiIndex, found {type(obj.index)}"
         return _ret(False, msg, None, return_metadata)
 
-    # check that columns are unique
+    # check that row index is unique
+    msg = f"{var_name} must have unique row indices, but found {obj.index}"
+    assert obj.index.is_unique, msg
+
+    # check that rows are sorted
+    msg = f"{var_name} must have lexsorted row index, but found {obj.index}"
+    assert obj.index.is_monotonic_increasing, msg
+
+    # check that column index is unique
     msg = f"{var_name} must have " f"unique column indices, but found {obj.columns}"
     assert obj.columns.is_unique, msg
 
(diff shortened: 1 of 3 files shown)


fkiraly:
fails since reconcilers seem to produce non-lexsorted indices

patrickzib (changes requested):
Is there no better way than having copies of the same code snippet all over different classes? Eg call some helper function?

fkiraly:
I suppose one could use sth like `pandera`.

The trade-off for sth written here seems to be coupling (via the utility if custom) vs readability of code and independence of the individual checkers.

Ultimately, I would like to refactor the checkers/converters into individual classes https://github.com/sktime/sktime/issues/3512, that is the stage where it would be best to resolve multiplications here.

sktime/datatypes/_hierarchical/_check.py
```diff
@@ -85,9 +85,20 @@ def check_pdmultiindex_hierarchical(obj, return_metadata=False, var_name="obj"):
...
         return _ret(False, msg, None, return_metadata)
 
-    # check that columns are unique
+    # check that row index is unique
+    msg = f"{var_name} must have unique row indices, but found {obj.index}"
+    if not obj.index.is_unique:
+        return _ret(False, msg, None, return_metadata)
+
+    # check that rows are sorted
+    msg = f"{var_name} must have lexsorted row index, but found {obj.index}"
```
GuzalBulatova, line 94: Agree with @patrickzib, but if this will be refactored in #3512 then we can leave the copied code.

fkiraly: yes, I'd leave it to then

GuzalBulatova:
@fkiraly pinging you in case this PR got lost :)

fkiraly:
@GuzalBulatova, it got sort-of lost - I got stuck when I saw some failures from the reconciler returning unsorted columns, according to the check. But, according to the debugger, and manual inspection, the returned indices *did* seem sorted, so I don't quite understand what is going on here.

PR #3553, opened 2022-10-08
[ENH] replace inplace sort by non-inplace sort in `Reconciler`

fkiraly:
Replaces inplace sort by non-inplace sort in `Reconciler` to avoid any inplace related issues (e.g., side effects)


patrickzib (approved):
Looks fine to me

PR #3554, opened 2022-10-08
[ENH] `dask` mtypes - part 1, `Series`

fkiraly:
Adds mtypes for `dask` containers.

First part, adds `Series` (non-hierarchical) mtype to validate the idea and allow estimator prototypes built on top.


fkiraly:
FYI @topher-lo

fkiraly:
really strange - this works locally, on python 3.8 and 3.10 (window), but on remote CI python 3.10, `get_examples` seems to return empty lists of examples.

@topher-lo, any good idea what is going on here?

topher-lo:
I'm almost entirely certain it's just a missing dependency problem. We should add `dask>=2022.6.0` explicitly in optional dependencies.

Screenshot 2022-10-11 183048
Screenshot 2022-10-11 183120

fkiraly:
ah, thanks for the review!!

I made two mistakes that "cancelled each other out" on certain versions:

* forgetting `dask` as a softdependency in the pyproject.toml
* accidentally switching the input check of `dask` with `xarray`

So part of the data type framework would think it could support `dask` types. To determine number of examples, the first element in a numpy array is used, which sorts non-dask string on 3.7-3.9, but dask first in 3.10, which causes the bug.

PR #3557, opened 2022-10-09
[ENH] `numpy` integer support for `ColumnEnsembleForecaster`

fkiraly:
This PR adds support for `numpy` based integer types in `ColumnEnsembleForecaster`, previously this would have broken.

Also changes one of the current tests to cover this case.


GuzalBulatova:
Sorry I missed the requested review. Thank you Franz.

PR #3561, opened 2022-10-10
[ENH] Add more relevant parameters to `CNNRegressor` for user flexibility

### LINKED ISSUE #3538: [BUG] `CNNRegressor` missing essential initializing parameters
**Describe the bug**

`CNNRegressor` misses parameters like `optimizer` in its constructor. There might be more things which don't coincide with `CNNClassifier` which should be added but since I'm not sure what all things might be missing, I'm opening this issue to bring it to notice.

achieveordie:
#### Reference Issues/PRs

Fixes #3538

- Add the following parameters into constructor:
  - `optimizer`
  - `activation`
  - `use_bias`

No.

@AurumnPegasus do let me know if you see more changes that we can add in this PR.

##### For all contributions

### DIFF
--- a/sktime/regression/deep_learning/cnn.py
+++ b/sktime/regression/deep_learning/cnn.py
@@ -1,8 +1,10 @@
 # -*- coding: utf-8 -*-
 """Time Convolutional Neural Network (CNN) for regression."""
 
-__author__ = ["AurumnPegasus"]
+__author__ = ["AurumnPegasus", "achieveordie"]
 __all__ = ["CNNRegressor"]
+
+from sklearn.utils import check_random_state
 
 from sktime.networks.cnn import CNNNetwork
 from sktime.regression.deep_learning.base import BaseDeepRegressor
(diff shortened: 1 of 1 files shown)


TonyBagnall (approved):

fkiraly (approved):
Thanks, makes sense.

Good spot on `random_state` which had the wrong name too.

TonyBagnall:
@fkiraly doesn't the change of variable name require deprecation?

achieveordie:
@TonyBagnall, other estimators were using `random_state` for the outer interface (constructor parameter) as well as the inner interface (`self.random_state`). `CNNRegressor` was using `random_state` as a parameter constructor but was using `self.random_seed` internally. I just changed the name to `self.random_state` so there's uniformity in the naming of variables used internally.

Since no changes have been made to the user interface, no deprecation is involved. LMK if I misunderstood your point otherwise.

TonyBagnall:
I agree, I have nothing against the change, but other PRs are being blocked for changing parameter default values or other minor changes, I think we should be consistent. 

image

achieveordie:
Well, that does make sense

fkiraly:
> The parameter name changes from `random_seed` to `random_state`, minor and only really relevant if the estimator has been in a release.

I noticed this but would consider that a bug, since per interface contract this should be `random_state`, and it was the reason for test failure afaik. Bugfixes do not require deprecation.

fkiraly:
I would apply the same rule if, say, you would be fixing a bug where `X` in `_fit` is overwritten, or another argument is used by accident. Yes, this changes interface and behaviour, but it fixes a bug.

fkiraly:
@TonyBagnall, @MatthewMiddlehurst, for reference, here is another issue from July where I have requested deprecation from @AurumnPegasus but not applied the deprecation rule to `random_state`: https://github.com/sktime/sktime/pull/2991#pullrequestreview-1062150757

fkiraly:
> Is the n_jobs -1 default from https://github.com/sktime/sktime/pull/3431 considered a bug because the generally accepted functionality of n_jobs=-1 is to use all available threads?

No, that's not a tested interface convention.

PR #3562, opened 2022-10-10
[BUG] Fix typo in `set_tags` call in `AggrDist.__init__`

aiwalter:
Fix typo in method `set_tags()`

### DIFF
--- a/sktime/dists_kernels/compose_tab_to_panel.py
+++ b/sktime/dists_kernels/compose_tab_to_panel.py
@@ -125,7 +125,7 @@ class AggrDist(BasePairwiseTransformerPanel):
         # importing inside to avoid circular dependencies
         from sktime.dists_kernels import ScipyDist
 
-        return {"transformer": ScipyDist()}
+        return {"transformer": ScipyDist(), "aggfunc_is_symm": [True, False]}
 
 
 class FlatDist(BasePairwiseTransformerPanel):
--- a/sktime/dists_kernels/compose_tab_to_panel.py
+++ b/sktime/dists_kernels/compose_tab_to_panel.py
@@ -65,7 +65,7 @@ class AggrDist(BasePairwiseTransformerPanel):
         super(AggrDist, self).__init__()
 
         if self.aggfunc_is_symm:
-            self.set_tag("symmetric", True)
+            self.set_tags(**{"symmetric": True})
 
     def _transform(self, X, X2=None):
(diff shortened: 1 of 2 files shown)


sktime/dists_kernels/compose_tab_to_panel.py
```diff
@@ -125,7 +125,7 @@ def get_test_params(cls, parameter_set="default"):
         # importing inside to avoid circular dependencies
         from sktime.dists_kernels import ScipyDist
 
-        return {"transformer": ScipyDist()}
+        return {"transformer": ScipyDist(), "aggfunc_is_symm": [True, False]}
```
fkiraly, line 128: this is not how this works, you need to return a list of parameter dicts, not a dict of lists

fkiraly (changes requested):
Great, except for the `get_test_params` signature which has a small mistake

sktime/dists_kernels/compose_tab_to_panel.py
aiwalter: Hm why are tests then passing?

fkiraly: don't know. Probably because there is no input parameter validation and the checks are "if aggfunc_if_symm = something else something else"

aiwalter:
@fkiraly  test params should be fine now? But tests are failing, do you mind to look? Its "your" code

fkiraly:
Hm, failures seem to be unrelated and an instance of https://github.com/sktime/sktime/issues/3525

Where did you branch off? Before or after the fix?

fkiraly:
Just checked - *after* the fix, so the supposed fix did not fix it.

achieveordie:
I know this has been merged and done, But just FYI, looking at the error trace in the latest windows [runner](https://github.com/sktime/sktime/actions/runs/3259756522/jobs/5353813766), it seems like `FCNClassifier` crashes due to OOM. The other two failures are due to the workers crashing, which makes sense since OOM should crash the entire process. `FCNClassifier` crashes due to lack of a `get_test_params()` (this function reduces the size of the estimator, enabling faster tests and avoiding failures like this). The fix was only for `TapNet` estimator, which isn't the estimator that caused this failure, it was just running on a different worker.

_Tl;dr_ The fix works, the problem is due to the same problem in a different DL estimator. We'll try to fix all the current estimators in #3620 or a new PR since that's probably just for diagnosis.

PR #3566, opened 2022-10-10
[BUG] fix `check_estimator` exclude arguments not working for non-base scitype tests

### LINKED ISSUE #3565: [BUG] check_estimator tests_to_exclude doesn't exclude the specified tests
**Describe the bug**

Specifying a list of tests in the `tests_to_exclude` variable in `check_estimator` doesn't exclude the specified tests.

**To Reproduce**

```python

log = check_estimator(
    NaiveForecaster, return_exceptions=True, tests_to_exclude=["test_predict_residuals"]
)

should_be_empty = [key for key in log.keys() if "test_predict_residuals" in key]

assert len(should_be_empty) == 0
```

**Expected behavior**

Any tests in the tests_to_excluded should be actually excluded!
(FYI for whoever takes this on. The problem is somewhere [here](https://github.com/sktime/sktime/blob/fe6f43c2e06e586c2dd16102aef5eef4bd4b6ce6/sktime/tests/test_all_estimators.py#L378))

**Versions**
<details>

sktime==0.13.4  
python==3.8.5
</details>

fkiraly:
Fixes #3565, i.e.,. `check_estimator` exclude arguments would not work for non-base scitype specific tests

This was due to the exclude arguments not being passed on properly to `run_tests`.

The existing test did not cover this failure because it looked at base scitype tests only. It has been modified to also include some non-base scitype tests.

### DIFF
--- a/sktime/utils/estimator_checks.py
+++ b/sktime/utils/estimator_checks.py
@@ -133,6 +133,8 @@ def check_estimator(
             return_exceptions=return_exceptions,
             tests_to_run=tests_to_run,
             fixtures_to_run=fixtures_to_run,
+            tests_to_exclude=tests_to_exclude,
+            fixtures_to_exclude=fixtures_to_exclude,
         )
         results.update(results_scitype)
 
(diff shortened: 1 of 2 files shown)


ltsaprounis (approved):
That was fast! tested on my example in #3565 and it now works as intended.

PR #3568, opened 2022-10-10
[ENH] NA handling for forecasting pipelines

fkiraly:
Continuation of https://github.com/sktime/sktime/pull/3549, adds dynamic handling of NA tag to forecasting pipelines based on components.


aiwalter:
not sure if this is the best solution, if the actual forecater raises the exception its more clear to user where the error is coming from so another forecaster could be chosen or an `Imputer` be used?

PR #3569, opened 2022-10-11
[MNT] Remove `ABCMeta` inheritance from `_HeterogeneousMetaEstimator`

fkiraly:
This PR removes `ABCMeta` inheritance from `_HeterogeneousMetaEstimator` which fails currently due to `flake8` `B024`, i.e., abstract base class without abstract decorators.

In fact, `_HeterogeneousMetaEstimator` is a mixin and not an abstract base class, so this is indeed a style/pattern failure which was not picked up earlier (possibly since `flake8` was not enabled back then).

### DIFF
--- a/sktime/base/_meta.py
+++ b/sktime/base/_meta.py
@@ -6,13 +6,12 @@
 __author__ = ["mloning, fkiraly"]
 __all__ = ["_HeterogenousMetaEstimator"]
 
-from abc import ABCMeta
 from inspect import isclass
 
 from sktime.base import BaseEstimator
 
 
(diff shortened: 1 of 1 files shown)


TonyBagnall (approved):
I think the two methods should return a more informative message, such as not implemented, since strictly speaking they are not abstract methods (i.e. base class does not have to implement them)

def get_params(self, deep=True):
        """Return estimator parameters."""
    def set_params(self, **params):
        """Set estimator parameters."""

fkiraly:
these methods are implemented by the top base class, so will always be implemented.

Issue #3370, opened 2022-08-31
[ENH] Migrate from `sphinx-panels` to `sphinx-design`


achieveordie:
**Is your feature request related to a problem? Please describe.**
Currently, we're using `sphinx-panel` as an [extension](https://github.com/alan-turing-institute/sktime/blob/3c56f535dc1d060c0053ab6c7f1a1a7199133539/docs/source/conf.py#L51) to build documentation. However, in the current state, `sphinx-panel` is not being actively [maintained](https://github.com/executablebooks/sphinx-design/issues/51) and they suggest using [`sphinx-design`](https://github.com/executablebooks/sphinx-design) instead.

**Describe the solution you'd like**
There is a migration [guide](https://sphinx-design.readthedocs.io/en/latest/get_started.html#migrating-from-sphinx-panels) which can be followed. Changes involve replacing `panel`, `link-button` etc. directives with their updated versions. It might not be as simple as refactoring the names since there seem to be some differences in their formatting as well.

miraep8:
Is this issue now close-able?

fkiraly:
Yes, the PR https://github.com/sktime/sktime/pull/3575 was merged, so this is resolved.

PR #3575, opened 2022-10-12
[DOC] Replace `sphinx-panels` with `sphinx-design`

MatthewMiddlehurst:
Replaces the `sphinx-panels` dependency with `sphinx-design` #3370.

The cards/panels are slightly smaller than before, and the button colour has changed since the default grey bugged out on darkmode for me with the new package.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -4,7 +4,7 @@
 Welcome to sktime
 =================
 
-  A unified framework for machine learning with time series. :)
+  A unified framework for machine learning with time series.
 
 
 Mission
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -4,7 +4,7 @@
 Welcome to sktime
 =================
 
-  A unified framework for machine learning with time series
+  A unified framework for machine learning with time series. :)
 
 
 Mission
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -47,103 +47,131 @@ Contents
    developers
    about
 
-.. panels::
-    :card: + intro-card text-center
+.. grid:: 1 2 2 2
+    :gutter: 3
 
-    ---
+    .. grid-item-card::
+        :text-align: center
 
-    Get Started
-    ^^^^^^^^^^^
+        Get Started
 
-    Get started using ``sktime`` quickly.
+        ^^^
 
-    +++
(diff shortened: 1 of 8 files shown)


TonyBagnall (approved):

fkiraly (changes requested):
Why is this needed? What does it improve?

MatthewMiddlehurst:
See #3370. You gave it a thumbs up, so I assumed you agreed with the change.

RNKuhns:
@MatthewMiddlehurst I'll give this a review in the next day.

@fkiraly the `sphinx-panels` package that has been used in the `sktime` documentation is no longer maintained. `sphinx-design` provides similar functionality and is suggested as an alternative by [sphinx-panels](https://sphinx-panels.readthedocs.io/en/latest/index.html). Plus `sphinx-design` looks nice (IMHO).

fkiraly:
ah yes, I remember, thanks. Was confused by the PR text which said something about panels being smaller, which struck me odd as a reason for a change, but you have to follow the link to #3370. Please consider this unblocked (but not approved)

fkiraly:
any chance to get this working using `sphinx-design`?
https://github.com/sktime/sktime/pull/3211

Using `sphinx-panels`, it was impossible to get two buttons on a panel and have both work.

achieveordie:
LGTM

MatthewMiddlehurst:
@RNKuhns do you still intend to review this? Not rushing you to do so if you are.

docs/source/index.rst
```diff
@@ -47,103 +47,131 @@ Contents
...
-    User Guide
-    ^^^^^
+            Get Started
 
-    Find user documentation.
+    .. grid-item-card::
+        :text-align: center
 
-    +++
+        User Guide
```
RNKuhns, line 74: This is really pointing to the broader category of "User Documentation". Could consider changing these labels to "User Documentation" from "User Guide". But this PR does successfully convert to sphinx-design and retain the original label.

RNKuhns (approved):
@MatthewMiddlehurst this looks good to me and the new buttons point to the same places as the old buttons. I think we should be able to merge this.

I did make one comment on the label for the "User Guide" on the index. It is really pointing to the "User Documentation". But you maintained the prior label, so maybe we shouldn't change that as part of this PR.

PR #3579, opened 2022-10-12
[BUG] `ClearSky` fix missing value problem after transform

ciaran-g:

Missing values which should be retained as missing in the transformed domain were being overwritten to zero.

* Simple fix to only set transformed values to zero when they are below the clear sky power threshold **and** if the original value is not null

### DIFF
--- a/sktime/datasets/_single_problem_loaders.py
+++ b/sktime/datasets/_single_problem_loaders.py
@@ -13,6 +13,7 @@ __author__ = [
     "aiwalter",
     "jasonlines",
     "achieveordie",
+    "ciaran-g",
 ]
 
 __all__ = [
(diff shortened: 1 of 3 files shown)


fkiraly (changes requested):
makes sense to me.

Would it be easy to add a test to certify for the fix?
(or modify an existing one, e.g., checking the output in a special case and expected values vs produced)

ciaran-g:
> Would it be easy to add a test to certify for the fix?

Added this franz 👍

Also added the author/function name in `sktime/datasets/_single_problem_loaders.py` which I just realised I forgot to add at the time

fkiraly (approved):
Great, thanks!

PR #3587, opened 2022-10-16
[MNT] loosen `scipy` bound to <2.0.0

### LINKED ISSUE #3586: [BUG] sktime incompatible with the latest versions of scipy
**Describe the bug**

The latest versions of scipy (>= 1.9.0) are incompatible with the latest sktime version.

**To Reproduce**

1. In a clean virtual environment, create this requirements.txt file:
```
scipy==1.9.2
sktime==0.13.4
```
2. Install the requirements with pip:
```
% pip install -r requirements.txt
```

**Bug**:
```
Collecting scipy==1.9.2
  Using cached scipy-1.9.2-cp310-cp310-macosx_10_9_x86_64.whl (34.2 MB)
Collecting sktime==0.13.4
  Using cached sktime-0.13.4-py3-none-any.whl (7.0 MB)
Collecting numpy<1.26.0,>=1.18.5
  Using cached numpy-1.23.4-cp310-cp310-macosx_10_9_x86_64.whl (18.1 MB)
Collecting statsmodels>=0.12.1
  Using cached statsmodels-0.13.2-cp310-cp310-macosx_10_9_x86_64.whl (9.7 MB)
Collecting deprecated>=1.2.13
  Using cached Deprecated-1.2.13-py2.py3-none-any.whl (9.6 kB)
INFO: pip is looking at multiple versions of <Python from Requires-Python> to determine which version is compatible with other requirements. This could take a while.
INFO: pip is looking at multiple versions of scipy to determine which version is compatible with other requirements. This could take a while.
ERROR: Cannot install -r requirements.txt (line 2) and scipy==1.9.2 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested scipy==1.9.2
    sktime 0.13.4 depends on scipy<1.9.0

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict
```

**Expected behavior**

I would expect the latest release versions of scipy and sktime to be mutually compatible, given they are likely to frequently be used together.

**Additional context**

I could take a stab at the fix, but I would need someone to explain the why behind the upper bound in [pyproject.toml](https://github.com/sktime/sktime/blob/main/pyproject.toml), so I know what to replace it with:

```
    "scipy<1.9.0",
```

I'm assuming there's some implicit SemVer policy behind the upper bound, though it could be something else, like sktime depends on implicit behavior in scipy in a way that breaks with the latest version. If it's the latter case, this is probably beyond my scope, since this would be my first contribution.

**Versions**

N/A: You can't actually run `show_versions` when the package install fails.

fkiraly:
This PR loosens the upper version bound for `scipy` to `<2.0.0`, thus removing any restriction on `1.X.X` versions.

We had the MINOR bound introduced after repeated breakage, this was mostly on `sktime` side using private methods (which was a mistake). Afaik we are no longer doing that, and `scipy` is very good at deprecation, so I think we can safely move the bound up to the MAJOR cycle.

Fixes #3586.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,7 @@ dependencies = [
     "pandas>=1.1.0,<1.6.0",
     "scikit-learn>=0.24.0,<1.2.0",
     "statsmodels>=0.12.1",
-    "scipy<1.9.0",
+    "scipy<2.0.0",
 ]
 
 [project.optional-dependencies]


fkiraly:
FYI @adamcataldo

aiwalter (approved):

PR #3590, opened 2022-10-16
[ENH] tests for `get_fitted_params` interface contract by estimator

fkiraly:
The `get_fitted_params` interface contract was not covered in per-estimator tests so far.

This PR adds testing for the basic interface contract on `get_fitted_params` on a by-estimator basis..

Composition properties and base class functionality is already covered in tests of `BaseEstimator`, in `test_base`.

On a more general note, testing of `BaseEstimator` abstract methods is added to `test_methods_do_not_change_state`, which will be renamed to `test_non_state_changing_method_contract` in 0.15.0 after a deprecation introduced.
The deprecation is "safe" in that the deprecation message recommends adding both the new and old name to arguments of `check_estimator`, in a case where the old argument has been used.

Includes https://github.com/sktime/sktime/pull/3595 to address methods with empty argument set in scenarios.

Includes fixes for bugs found through this:
* https://github.com/sktime/sktime/pull/3598
* https://github.com/sktime/sktime/pull/3599

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1752,7 +1752,8 @@
       "name": "An Hoang",
       "profile": "https://github.com/AnH0ang",
       "contributions": [
-        "bug"
+        "bug",
+        "code"
       ]
     },
     {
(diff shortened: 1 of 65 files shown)


fkiraly:
Failures seem all genuine

Issue #3363, opened 2022-08-29
[ENH] Import more estimators from hmmlearn


miraep8:
[hmmlearn](https://github.com/hmmlearn/hmmlearn) is a nifty python package with a lot of different sci-kit learn like estimators for using HMM algorithms. 

We now have a simple way to wrap and thus import new HMM estimators into `sktime`, and I thought that it could be a great task for anyone new to `sktime` or open source in general to get started with if you are looking for an easy issue to contribute. 

The estimators which still need to be wrapped are:
  - [ ] CategoricalHMM
  - [ ] MultinomialHMM
  - [ ] PoissonHMM

Here is a rough guide for what needs to be done in order to wrap an estimator from `hmmlearn`:
   - in the `sktime.annotations.hmm_learn` file add your new estimator to the `__all__` attribute list near the top of the file.
   - make a new class for your relevant `hmmlearn` estimator.
   - copy over the docstring from `hmmlearn` into your new class.  [Here](https://github.com/hmmlearn/hmmlearn/blob/main/lib/hmmlearn/hmm.py) is where the code lives on Github and you can check it out there.  (may need to look into both the `__init__` function and the main class docstring in order to get both the `Parameters` section and the `Attributes` slash general description. 
   - update the `__init__` and `_fit` functions to properly collect and store all the attributes needed for your wrapped estimator.  In `_fit` pass these parameters to a wrapped version of your estimator. 
   - update the docstring example
   - update the get_test_params method with some relevant params for your new estimator.
   - add a quick test in test_hmm_learn that gaurantees your new wrapped estimator agrees with the estimator it is wrapping in `hmmlearn`.  It will be important to make sure both implementations have the same random state. 
   
   To see an example of what this should look like please check out my recent PR (#3362) for GMMHMM which has all these steps in it.  If you would like top take on an estimator just comment in this issue so other people know you are working on it! :)

kejsitake:
Hi! Since no one else has taken over, I am planning to work on importing these estimators this weekend, after a paper deadline tomorrow!

miraep8:
Fantastic, thank you! :D  Let me know if you have any questions on this and I would be happy to review your PR when its ready, just ping me

JonathanBechtel:
So I think myself and a few co-workers are also going to work on this issue.  Seeing that issue #3597 was never resolved, and there does not seem to be any additional HMM modules imported into the library since this issue, we're planning on seeing if we can get some of these taken care of.  

[...]

PR #3597, opened 2022-10-17
[ENH] Imports wrapped CategoricalHMM estimator

kejsitake:
#### Reference Issues/PRs

Part of https://github.com/sktime/sktime/issues/3363

This PR adds a wrapper around CategoricalHMM

Yes, this PR uses hmmlearn as a soft dependency.

#### Any other comments?

I noticed a type in the setup.cfg file that was not allowing me to run the parametrized pytest. Fixing the typo, I was able to succesfully run them.


miraep8:
Hey! Good spot on the typo with pytest!  Surprising that no one has noticed that before now........ 🤔 

I just took a quick look over and think it looks pretty good to go, once you figure out a better set of test parameters to feed it! I would recommend playing around with some example data in a jupyter notebook or something until you find a set of data/parameters that play well together.   I suspect that starting with discrete data rather than the piecewise normal might get you most of the way there though! 

One small request with this PR - would you mind adding this new estimator to the documentation for annotators?  To see how can check out #3603 .  That way people can find and use your estimator once we add it! 😄 

Thanks again a lot for taking this on and please ping me if you want help debugging the test parameters issue!

kejsitake:
Hi! I have been wondering what the best way to handle the categorical case in testing is. In forecasters, I believe we use tags, but there is no tags for annotators.

More specifically the tests are failing because in sktime/annotations/tests/tests_all_annotators CategoricalHmm is initiated with non categorical values. If we had tags that marked this estimator as categorical then I could add an if statement, but I am not sure what is the current best way to handle this?

miraep8:
Just to copy over/summarize my notes from the slack channel:

Personally I support your proposed solution  i.e. adding a categorical tag to the annotation class and an if statement in `test_all_annotators`.

I think that this issue and your proposed solution would actually be worth brining up in the next meeting we have an base annotators etc! (as we discuss more the designs of the base classes).

Thanks again for implementing this, sorry that you ended up with an estimator that was a bit trickier to implement but glad this is getting us thinking a bit more about the design questions for base annotators that have been a bit on the back burner!  Good luck! :)

PR #3606, opened 2022-10-18
[BUG] fix behaviour of `FourierFeatures` with `pd.DatetimeIndex`

### LINKED ISSUE #3605: [BUG] `FourierFeatures` with `pd.DatetimeIndex`
**Describe the bug**
When using data with a `pd.DatetimeIndex` in the `FourierFeatures` transformer, it gets converted into nanoseconds. This means you have to specify all of your sp in terms of nanoseconds which is not clear.

**To Reproduce**

```python
from sktime.transformations.series.fourier import FourierFeatures

df = load_airline()
df.index = df.index.to_timestamp(freq="M")
from sktime.transformations.series.fourier import FourierFeatures
F_transform = FourierFeatures(sp_list = [12], fourier_terms_list=[2]) # should be yearly
sp1 = F_transform.fit_transform(df)

# Convert seasonality to nanoseconds to get correct result
F_transform = FourierFeatures(sp_list = [12*30*24*3600000000000], fourier_terms_list=[2])
sp2 = F_transform.fit_transform(df)

plot_series(sp1["sin_12_1"], sp2["sin_31104000000000000_1"])
```

**Expected behavior**
Would expect it to work from the frequency of the index in the `DatetimeIndex` case. 

**Additional context**
Have added functionality for the expected behaviour in #3606.

eenticott-shell:
#### Reference Issues/PRs
Fixes #3605 

Adds an extra case into `FourierFeatures` for dealing with a `pd.DatetimeIndex`. These are stored in nanoseconds and so the seasonalities have to also be specified in nanoseconds. Wanted to avoid this and work from frequency instead for the DatetimeIndex case.


sktime/transformations/series/fourier.py
```diff
@@ -152,6 +165,15 @@ def _fit(self, X, y=None):
         # store the integer form of the minimum date in the prediod index
         # this is used to make sure that time t is calculated with reference to
         # the data passed on fit
+        if isinstance(X.index, pd.DatetimeIndex):
+            if X.index.freq is None:
+                if self.freq is None:
+                    ValueError("X has no known frequency and none is supplied")
```
ltsaprounis, line 171: You can try to enforce a frequency here:
`X.index.freq = pd.infer_freq(df.index)`

We're successfully using this to overcome some mysterious pandas behaviour here:
https://github.com/sktime/sktime/blob/18b502e969d9c8d8dafce5cea78eb0c43fe23d91/sktime/datatypes/_vectorize.py#L387

you'd still need to throw the exception if `pd.infer_freq` returns None.

sktime/transformations/series/fourier.py
```diff
@@ -152,6 +165,15 @@ def _fit(self, X, y=None):
         # store the integer form of the minimum date in the prediod index
```
ltsaprounis, line 165: move the comment to be just above the line:
`self.min_t_ = np.min(X.index.astype(int))`

sktime/transformations/series/fourier.py
```diff
@@ -115,7 +123,7 @@ def __init__(self, sp_list: List[Union[int, float]], fourier_terms_list: List[in
 
         super(FourierFeatures, self).__init__()
 
-    def _fit(self, X, y=None):
+    def _fit(self, X, y=None, freq=None):
```
ltsaprounis, line 126: Remove the variable `freq` here, it's not being used and it's changing the signature of _fit for no reason.
You are correctly using the `self.freq` instead, in your code!

ltsaprounis (changes requested):
Thanks for the bugfix! left a few comments, but nothing major.
I'm keen to see @fkiraly's thoughts on my comment [here](https://github.com/sktime/sktime/issues/3605#issuecomment-1285995000), before we merge. I either misinterpreted the functionality of the tag "enforce_index_type", or there's another bug in the tag behaviour.

ltsaprounis (commented):
Based on this fix and on the discussion on #3605, I suggest we add `pd.DateTimeIndex` in the `enforce_index_type` list

eenticott-shell:
So the real bug was in the enforce index! I've added `pd.DatetimeIndex` as an allowed type, and addressed the other changes you suggested. Also changed the logic when choosing the frequency to avoid so many nested `if` statements.

sktime/transformations/series/fourier.py
```diff
@@ -149,9 +166,21 @@ def _fit(self, X, y=None):
                         "exists from other seasonal period, fourier term pairs."
                     )
 
-        # store the integer form of the minimum date in the prediod index
         # this is used to make sure that time t is calculated with reference to
         # the data passed on fit
+        if isinstance(X.index, pd.DatetimeIndex):
+            # Chooses first non None value
+            freq = X.index.freq or self.freq or pd.infer_freq(X.index)
```
ltsaprounis, line 173: I think we need to store this `freq` as an attribute, maybe `self.freq_` and use this instead in the predict step.

sktime/transformations/series/fourier.py
```diff
@@ -174,6 +203,10 @@ def _transform(self, X, y=None):
         transformed version of X
         """
         X_transformed = deepcopy(X)
+
+        if isinstance(X.index, pd.DatetimeIndex):
+            X.index = X.index.to_period(self.freq)
```
ltsaprounis, line 208: use the new attribute `self.freq_` described above instead and do the operation on X_transformed that at this point is a deepcopy of X to avoid any side effects on X.

ltsaprounis (changes requested):
Looking good! 
I left a few suggestions in the comments.

ltsaprounis:
Tests pass now! 
One last thing, can you just add an extra test in the [test_fourier.py](https://github.com/sktime/sktime/blob/main/sktime/transformations/series/tests/test_fourier.py) with something similar to your example in the issue? 

Just parametrise the test and add a Y_datetime. 
If you specify a monthly freq, the outputs etc should be the same as in the existing test.

ltsaprounis:
@eenticott-shell - In the case where the index is datetime, will the transformed series index maintain the datetime index, or will it change to period? Cause I think the transform method might be changing the index permanently here. Can you quickly check and maybe add it in the test that the index is not changed. Something like:
```python
from pandas.testing import assert_index_equal
assert_index_equal(left, right, exact=True)
```

ltsaprounis:
> Will approve and merge once the CI is done.

@fkiraly - given it's a bug, can we merge this before the release? your call...

eenticott-shell:
Thanks for your help! I think one of the tests might fail, I am having trouble running pytest locally but I am currently trying to fix!

eenticott-shell:
Think that should have fixed it

sktime/transformations/series/fourier.py
```diff
@@ -184,6 +220,8 @@ def _transform(self, X, y=None):
             X_transformed[f"sin_{sp}_{k}"] = np.sin(int_index * 2 * k * np.pi / sp)
             X_transformed[f"cos_{sp}_{k}"] = np.cos(int_index * 2 * k * np.pi / sp)
 
+        # Ensure transformed X has same index
+        X_transformed.index = X.index
```
ltsaprounis, line 224: do a deepcopy(X.index) here for good measure

ltsaprounis (commented):
the failing test is not related to your transformer:
[FAILED sktime/transformations/panel/rocket/tests/test_Rocket.py::test_rocket_on_gunpoint - assert 0.9933333333333333 == 1.0](https://github.com/sktime/sktime/actions/runs/3379230667/jobs/5610684533#step:7:814)

Left some comments for avoiding a potential side effect  in the transform index and removing the redundant index copy in fit.

sktime/transformations/series/fourier.py
fkiraly: we're testing for side effects, using `test_methods_have_no_side_effects`, so I think that is not needed?

fkiraly:
> @fkiraly - given it's a bug, can we merge this before the release? your call...

You need to approve and resolve conversations, @ltsaprounis if that's what you advocate.

sktime/transformations/series/fourier.py
```diff
@@ -149,9 +166,22 @@ def _fit(self, X, y=None):
                         "exists from other seasonal period, fourier term pairs."
                     )
 
-        # store the integer form of the minimum date in the prediod index
+        X = deepcopy(X)
```
GuzalBulatova, line 169: I'm curious if we should change the variable name here? Or what's the point of deep copy in this case?

GuzalBulatova (commented):
Looks good overall, thank you @eenticott-shell ! Have only one question

GuzalBulatova:
@ltsaprounis could you resolve the pending changes if you're satisfied with them?

sktime/transformations/series/fourier.py
ltsaprounis: no side-effects on the real X!

ltsaprounis (approved):
Looking good! thanks for the fix and your patience with the change requests @eenticott-shell! Will merge once the tests are done.

PR #3612, opened 2022-10-19
fix typo that stops pytests running locally

TonyBagnall:
fixes a typo in setup.cfg that stops pytest running correctly locally

addopts -> adopts

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,7 +7,7 @@ multi_line_output = 3
 
 [tool:pytest]
 # ignore certain folders and pytest warnings
-addopts =
+adopts =
     --ignore build_tools
     --ignore examples
     --ignore docs


lmmentel:
Hmmm according to the docs https://docs.pytest.org/en/7.1.x/reference/customize.html it should be `addopts` what was the exact issue that you saw?

TonyBagnall:
ah sorry, thought it was a typo. Perhaps if I set it to adopts it just gets ignored and this is just a pycharm thing. If I run in pytests with addopts in pycharm, it will not run tests unless I comment out these lines
    --durations 10
    --timeout 600
    --cov sktime
    --cov-report xml
    --cov-report html
    --showlocals
    --matrixdesign True
I'll close this PR and investigate further, problem may just be local, because having said this, it now seems to be running.

PR #3618, opened 2022-10-19
[BUG] bug in the interaction between Numba and np.zeros identified in #2397

### LINKED ISSUE #2397: [BUG] ShapeletTransformClassifier numba error when dtype is not float64
**Describe the bug**
Seems that when using `ShapeletTransformClassifier` there is some Numba accelerated functions that break if the data in the input data frame are of type `int32`.

**To Reproduce**
MRE as below:

```python
import warnings
warnings.simplefilter('ignore', category=FutureWarning)

import numpy as np
import pandas as pd

from sklearn.model_selection import train_test_split

from sktime.classification.shapelet_based import ShapeletTransformClassifier
from sktime.contrib.vector_classifiers._rotation_forest import RotationForest

# make fake data
data = pd.DataFrame(np.random.random((5000, 250))).astype(np.float32)

# reshape to input into Shapelet Classifier
data4train = data.apply(lambda row: pd.Series({
  'time-series': pd.Series(row.values)
}), axis=1)

# make targets
targets = pd.Series(2500 * [1] + 2500 * [0])

# train test split
X_train, X_test, y_train, y_test = train_test_split(
  data4train, targets, test_size=0.7, random_state=42
)

# train
clf = ShapeletTransformClassifier(
    estimator=RotationForest(n_estimators=3),
    n_shapelet_samples=500,
    max_shapelets=20,
    batch_size=100,
)

clf.fit(X_train, y_train)
```

**Expected behavior**
will not throw an error, and also enforce conversion to float32 or float64 within the classifier?
**Additional context**
removing conversion to `float32` (hence `dtype == float64`) will make the code running without issues.

**Versions**
numba 0.55.1
sklearn 0.24.1
sktime 0.11.0
pandas 1.4.2
python 3.8.10

**Stacktrace output**
```bash
TypingError: Failed in nopython mode pipeline (step: nopython frontend)
Cannot unify array(float64, 1d, C) and array(float32, 1d, C) for 'X_n.2', defined at /path_to_mypython/python/lib/python3.8/site-packages/sktime/utils/numba/general.py (39)

File "../python/lib/python3.8/site-packages/sktime/utils/numba/general.py", line 39:
def z_normalise_series(X):
    <source elided>

    return X_n
```

TonyBagnall:
#### Reference Issues/PRs

 Fixes #2397 

#2397 identified that STC could fail when passed float32 data instead of float64. On investigation, it turns out that the actual problem was with the call to np.zeros in the numba method z_normalise_series. Numba does some strange type checking, and it does not like the return signature of np.zeros when called with float32. We could fix this by setting the dtype for np.zeros, but this would not work with int32 or int64. 

So the simplest solution is to replace the function with this

```
@njit(fastmath=True, cache=True)
def z_normalise_series(X):
    """Numba z-normalisation function for a single time series."""
    std = np.std(X)
    if std > 0:
        X_n = (X - np.mean(X)) / std
    else:
        X_n = X - np.mean(X)
    return X_n
```

adapting the original code in #2397 

```
def debug_numba_stc_2397(type):
    """See https://github.com/sktime/sktime/issues/2397."""
    import warnings
    warnings.simplefilter("ignore", category=FutureWarning)
    from sklearn.model_selection import train_test_split
    from sktime.classification.shapelet_based import ShapeletTransformClassifier
    from sktime.classification.sklearn import RotationForest
    # make fake data
    if type == "int32" or type == "int64": # Ensure not all zeros
        data = pd.DataFrame(100 * np.random.random((500, 25))).astype(type)
    else:
        data = pd.DataFrame(np.random.random((500, 25))).astype(type)

    # reshape to input into Shapelet Classifier
    data4train = data.apply(
        lambda row: pd.Series({"time-series": pd.Series(row.values)}), axis=1
    )

    # make targets
    targets = pd.Series(250 * [1] + 250 * [0])

    # train test split
    X_train, X_test, y_train, y_test = train_test_split(
        data4train, targets, test_size=0.7, random_state=42
    )

    # train
    clf = ShapeletTransformClassifier(n_shapelet_samples=100)

    clf.fit(X_train, y_train)
```

this now works

```
    debug_numba_stc_2397("int32")
    debug_numba_stc_2397("float64")
    debug_numba_stc_2397("float32")
    debug_numba_stc_2397("int64")
```

I have added a unit test for this function. 

no

#### Any other comments?

there is a wider issue as to whether the datatypes converter should automatically convert input data into a standard type (float32 probably), but that is beyond the scope of this tiny bug fix!

##### For all contributions
- [ x] I've added unit tests and made sure they pass locally.

### DIFF
--- a/sktime/utils/numba/tests/test_general.py
+++ b/sktime/utils/numba/tests/test_general.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+"""Tests for numba functiond."""
+
+__author__ = ["TonyBagnall"]
+
+import numpy as np
+import pytest
+from numpy.testing import assert_array_equal
+from utils.numba.general import z_normalise_series
+
+DATATYPES = ["int32", "int64", "float32", "float64"]
+
+
+@pytest.mark.parametrize("type", DATATYPES)
+def test_z_normalise_series(type):
+    """Test the function z_normalise_series."""
+    a = np.array([2, 2, 2], dtype=type)
+    a_expected = np.array([0, 0, 0], dtype=type)
+    a_result = z_normalise_series(a)
+    assert_array_equal(a_result, a_expected)
--- a/sktime/utils/numba/tests/test_general.py
+++ b/sktime/utils/numba/tests/test_general.py
@@ -6,7 +6,8 @@ __author__ = ["TonyBagnall"]
 import numpy as np
 import pytest
 from numpy.testing import assert_array_equal
-from utils.numba.general import z_normalise_series
+
+from sktime.utils.numba.general import z_normalise_series
 
 DATATYPES = ["int32", "int64", "float32", "float64"]
 
--- a/sktime/utils/numba/tests/test_general.py
+++ b/sktime/utils/numba/tests/test_general.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-"""Tests for numba functiond."""
+"""Tests for numba functions."""
 
 __author__ = ["TonyBagnall"]
 
--- a/sktime/utils/estimators/tests/__init__.py
+++ b/sktime/utils/estimators/tests/__init__.py
@@ -1,4 +1,4 @@
 # -*- coding: utf-8 -*-
-"""Tests for Mock Estimnators."""
+"""Tests for Mock Estimators."""
 
 __author__ = ["ltsaprounis"]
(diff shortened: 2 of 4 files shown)


sktime/utils/numba/tests/test_general.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+"""Tests for numba functiond."""
```
lmmentel, line 2:
```suggestion
"""Tests for numba functions."""
```

TonyBagnall: I fixed one typo and created a new one :)

patrickzib (approved):

Issue #3604, opened 2022-10-18
[BUG] CI fails on win with no code changes


lmmentel:
**Describe the bug**

CI fail on windows with only changes being to sphinx documentation rst files and no code changes. Noticed in  https://github.com/sktime/sktime/pull/3603

![image](https://user-images.githubusercontent.com/8989838/196434625-6fbb1d66-41a6-4d85-8204-d2df8e068923.png)

**To Reproduce**

```python
<Paste your code here>
```

**Expected behavior**

**Additional context**

**Versions**
<details>

</details>

TonyBagnall:
this is odd, failing on mine too, and I ran all tests locally on windows and it all passed
![image](https://user-images.githubusercontent.com/9594042/196446691-29cbee22-b763-46dc-9efa-b0cec43dba59.png)

lmmentel:
Thanks for sharing. The error seems to be coming from the `coverage` github action so you won't see it locally unless you are running the whole CI workflow.

miraep8:
I also noticed this, and anecdotally it seems to only be showing up for Windows 3.7?

TonyBagnall:
not sure if this is related, but running pytest on test_all_estimators locally generates a large number of files called .coverage during the run which persist on my machine. There is also a directory htmlcov. Is it meant to do this? Perhaps this might cause I/O issues ?
![image](https://user-images.githubusercontent.com/9594042/196676469-a4d627a5-b1e2-4e4b-9bc4-8e149a78f7e6.png)

TonyBagnall:
and further to this, I am writing a small test and when I run pytest I get all this, and it creates this htmlcov files locally 
![image](https://user-images.githubusercontent.com/9594042/196746694-e5adcfae-e81a-457b-af20-910f6c5c2adf.png)

Issue #3351, opened 2022-08-26
[ENH] Migrating Estimators from sktime-dl


AurumnPegasus:
**Is your feature request related to a problem? Please describe.**
Related to the task of migrating estimators from [sktime-dl](https://github.com/sktime/sktime-dl) to sktime. The main goal of doing so is to have all networks/models in a single package, following similar structure and quality. 

## Steps for Migration

1. Choose one of the Classifiers or Regressors, and create an issue for it.
2. You will need to create 2 new files for any DL estimator you choose:
    a. In `sktime/sktime/classification/deep_learning`: [here](https://github.com/alan-turing-institute/sktime/tree/main/sktime/classification/deep_learning), create a new file named after the network.
   b. In `sktime/sktime/networks`: [here](https://github.com/alan-turing-institute/sktime/tree/main/sktime/networks) create a new file named after the network.

## DL Classifiers:
- [x] [CNN Classifier](https://github.com/sktime/sktime-dl/blob/master/sktime_dl/classification/_cnn.py) #2991 #2447 
- [ ] [CNTC Classifier](https://github.com/sktime/sktime-dl/blob/master/sktime_dl/classification/_cntc.py) #3171 
- [ ] [Encoder Classifier](https://github.com/sktime/sktime-dl/blob/master/sktime_dl/classification/_encoder.py) #3293 
- [ ] [FCN Classifier](https://github.com/sktime/sktime-dl/blob/master/sktime_dl/classification/_fcn.py) #3233 
- [ ] [InceptionTime Classifier](https://github.com/sktime/sktime-dl/blob/master/sktime_dl/classification/_inceptiontime.py) #3003 
- [ ] [LSTMFCN Classifier](https://github.com/sktime/sktime-dl/blob/master/sktime_dl/classification/_lstmfcn.py) #3292 
- [ ] [MACNN Classifier](https://github.com/sktime/sktime-dl/blob/master/sktime_dl/classification/_macnn.py)  #3294 
- [ ] [MCDCNN Classifier](https://github.com/sktime/sktime-dl/blob/master/sktime_dl/classification/_mcdcnn.py) #3295 
- [ ] [MCNN Classifier](https://github.com/sktime/sktime-dl/blob/master/sktime_dl/classification/_mcnn.py)
- [x] [MLP Classifier](https://github.com/sktime/sktime-dl/blob/master/sktime_dl/classification/_mlp.py) #3232 
[...]

PR #3620, opened 2022-10-20
[BUG] Test CI fails

TonyBagnall:

This is to examine why CI is failing without any actual changes. There may be two unrelated problems:

Something to do with coverage See #3604
Intermittent tapnet failures #3525

### DIFF
--- a/sktime/_contrib/classification_experiments.py
+++ b/sktime/_contrib/classification_experiments.py
@@ -4,7 +4,6 @@
 This file is configured for runs of the main method with command line arguments, or for
 single debugging runs. Results are written in a standard format.
 
-Change text to test CI.
 """
 
 __author__ = ["TonyBagnall"]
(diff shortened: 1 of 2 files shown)


TonyBagnall:
seems FCN doesnt have get_test_params

achieveordie:
> seems FCN doesnt have get_test_params

We should add this as a requirement for all the migrating/new DL estimators especially as they are large in size per the default values. This means adding another step in #3351 since that's the only written documentation we have of the steps to be followed for migration.

TonyBagnall:
> > seems FCN doesnt have get_test_params
> 
> We should add this as a requirement for all the migrating/new DL estimators especially as they are large in size per the default values. This means adding another step in #3351 since that's the only written documentation we have of the steps to be followed for migration.

@achieveordie would you like to add an issue to update current DL classifiers with get_test_params? thanks, Tony

PR #3622, opened 2022-10-20
[ENH] SupervisedIntervals transformer and cleaned numba functions

MatthewMiddlehurst:

Adds a SupervisedIntervals transformer which uses the same interval extraction process as the current STSF classifier with extended parameterisation. Adaptation of the authors @stevcabello implementation (https://github.com/stevcabello/r-STSF/) with permission.

Cleans up the Numba utilities, implements faster alternatives for some.

No

New transformer and numba utility files.

##### For all contributions


sktime/transformations/panel/supervised_intervals.py
```diff
@@ -0,0 +1,471 @@
...
+        """
+        self.reset()
+        if y is None:
+            raise ValueError("SupervisedIntervals requires `y` in `fit`.")
+        X, y = self._check_X_y(X=X, y=y)
+
+        y = self._fit_setup(X, y)
+        X_norm = z_normalise_series_3d(X)
+
+        fit = Parallel(n_jobs=self._threads_to_use)(
```
patrickzib, line 201: Why not `n_jobs`?

sktime/transformations/panel/supervised_intervals.py
```diff
@@ -0,0 +1,471 @@
...
+            for i in range(self.n_intervals)
+        )
+
+        (
+            intervals,
+            transformed_intervals,
+        ) = zip(*fit)
+
+        self._intervals = []
+        for i in intervals:
```
patrickzib, line 218: `self._intervals = list(intervals)` ?

sktime/transformations/panel/supervised_intervals.py
```diff
@@ -0,0 +1,471 @@
...
+            Xt = np.hstack((Xt, transformed_intervals[i]))
+
+        self._is_fitted = True
+        return Xt
+
+    def _fit(self, X, y=None):
+        y = self._fit_setup(X, y)
+        X_norm = z_normalise_series_3d(X)
+
+        fit = Parallel(n_jobs=self._threads_to_use)(
```
patrickzib, line 234: `n_jobs`?

sktime/transformations/panel/supervised_intervals.py
```diff
@@ -0,0 +1,471 @@
...
+            for i in range(self.n_intervals)
+        )
+
+        (
+            intervals,
+            _,
+        ) = zip(*fit)
+
+        self._intervals = []
+        for i in intervals:
```
patrickzib, line 251: see above question in fit_transform

sktime/transformations/panel/supervised_intervals.py
```diff
@@ -0,0 +1,471 @@
...
+            self._intervals.extend(i)
+
+        self._transform_features = [True] * len(self._intervals)
+
+        return self
+
+    def _transform(self, X, y=None):
+        transform = Parallel(n_jobs=self._threads_to_use)(
```
patrickzib, line 259: n_jobs?

sktime/transformations/panel/supervised_intervals.py
```diff
@@ -0,0 +1,471 @@
...
+                candidate_agg_feats.extend(candidate_agg_feat_R)
+                if keep_transform:
+                    Xt = np.hstack((Xt, Xt_R))
+
+        if keep_transform:
+            Xt = Xt[:, 1:]
+
+        return candidate_agg_feats, Xt
+
+    def _transform_intervals(self, X, idx, skip):
```
patrickzib, line 367: can this be turned into njit? seems to be a bottleneck?

patrickzib (changes requested):
Some small questions/issues

sktime/utils/numba/general.py
```diff
@@ -4,15 +4,30 @@
 import numpy as np
 from numba import njit
 
+import sktime.utils.numba.stats as stats
+
 
 @njit(fastmath=True, cache=True)
 def unique_count(X):
```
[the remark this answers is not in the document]
MatthewMiddlehurst: the function is supported, but returning counts for each unique item is not. http://numba.pydata.org/numba-doc/latest/reference/numpysupported.html#other-functions

sktime/transformations/panel/tests/test_intervals.py
```diff
@@ -0,0 +1,1790 @@
...
+"""RandomIntervals test code."""
+import numpy as np
+from numpy import testing
+
+from sktime.datasets import load_basic_motions
+from sktime.transformations.panel.random_intervals import RandomIntervals
+from sktime.transformations.panel.supervised_intervals import SupervisedIntervals
+
+
+def test_random_intervals_on_basic_motions():
```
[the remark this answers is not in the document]
MatthewMiddlehurst: That is only for classification, there is no equivalent file/system for transformations.

sktime/transformations/panel/supervised_intervals.py
MatthewMiddlehurst: It is the var name we use in classification. I don't want to overwrite the input `self.n_jobs` when calling `self._threads_to_use = check_n_jobs(self.n_jobs)`

patrickzib: Maybe it is better to have a method get_n_jobs() then? It can return self._threads_to_use. But it is highly confusing that you set the number of jobs with one variable, and the one used has a different name.

MatthewMiddlehurst: changed the name to `self._n_jobs`

sktime/transformations/panel/supervised_intervals.py
MatthewMiddlehurst: this would create a list of lists rather than a single list im pretty sure

sktime/transformations/panel/supervised_intervals.py
MatthewMiddlehurst: we dont know if the `feature` function stored in `self._intervals` will be numba compatible. Best leave it to the user to pass in efficient functions, the default ones are already all numba.

MatthewMiddlehurst:
I have made the suggested changes or replied @patrickzib

patrickzib (approved):

fkiraly (approved):
Looks ok.

Please ensure you add the new transformer to the API reference in a separate PR, or here.

PR #3630, opened 2022-10-22
[ENH] tag manager mixin

fkiraly:
This PR factors out the tag management logic into a `_FlagManager` mixin, and generalizes the functionality so it can be used for multiple fields of tags.

Advantages:
* increases cohesion and reduces coupling, separates concerns in the `BaseObject` class: state/parameter management and tag management
* prepares the ground for managing configuration fields, e.g., for implementing a `set_config` similar to the new `sklearn` feature

Test coverage is already there for the base case, but should be added for a different tag field. Will do that with the PR that adds config handling.

(I called the flag manager since not all flags are tags, and interactions with the sklearn `_get_tags`)

A STEP is available here: https://github.com/sktime/enhancement-proposals/pull/29


RNKuhns (changes requested):
@fkiraly I’ve started looking at this (I’ll be looking more at implementation and testing yet, but have initial feedback). My initial thoughts are that I’m not 100% sure of the need for a separate config interface (of which this is a precursor).

I want to make sure a config interface is differentiated from the tag interface. A risk is adding complexity for users/developers needing to know when something is a tag and when something is a config item.

For example, why not have users set a dynamic tag that overrides the global config? Particularly, since we already use tags in relation to the current input/output mtype conversion. Would a config interface related to output types be adequately differentiated to warrant the new interface? I’m not sure and would want a bit more discussion on the topic.

Another consideration is whether exceptions to the global config are better handled through a context manager.

So I’m not so much as requesting changes as requesting further discussion.

fkiraly:
> I want to make sure a config interface is differentiated from the tag interface. A risk is adding complexity for users/developers needing to know when something is a tag and when something is a config item.

Yes, I think this is a valid point. I thought about this.
From a procedural perspective, I would need to write a STEP and explain the "why".

> For example, why not have users set a dynamic tag that overrides the global config that controls the output data container of an estimator? Particularly, since we already use tags in relation to the current input/output mtype conversion. Would a config interface related to output types be adequately differentiated to warrant the new interface? I’m not sure and would want a bit more discussion on the topic.

My answer for this would be: there needs to be a different interface because of difference in workflow.

* tags are intended to be set by *developers* and not changed by *users*. They relate to *general properties* of a given estimator.
* configs are intended to be set by *users*. They model *current behaviour configuration* of a given estimator.

E.g., if my estimator were a washing machine, tags would define the type of water and power connection, and not change once the washing machine is installed. Configs would correspond to the laundry settings, and could change with each use of the washing machine.

fkiraly:
STEP is here: https://github.com/sktime/enhancement-proposals/pull/29

fkiraly:
merged in future upstream `skbase`, I assume approval is transitive.

PR #3631, opened 2022-10-22
[MNT] isolate `numba` - NB: does not make `numba` a soft dependency

fkiraly:
This PR isolates `numba` throughout `sktime`.

Note: does *not* turn `numba` into a soft dependency.

This is required for all approaches to making `sktime` python 3.11 compatible as long as `numba` is not python 3.11 compatible, because that requires being able to run the framework components of `sktime` without `numba` imports.

See also https://github.com/sktime/sktime/issues/3963.

The general recipe to isolate `numba`:

* move `numba` imports into functions and classes, where possible
* where loose functions are decorated with `jit` and `njit`, move them to a private module `_orig_modulename_numba`, and move imports of that into functions or classes
* add the package dependency tag to `numba` dependent estimators to isolate them
* use `pytest.skipif` to conditionally skip `numba` dependent non-suite tests if `numba` is not installed

Makes the following changes:
* changes the `test_module_softdeps` test to ignore private modules (starting with `_`), i.e., to not require a user warning. This is ok since users are not expected to import private modules.

Depends on https://github.com/sktime/sktime/pull/3857


patrickzib:
I am not sure what you intention is? We havent even come to an agreement, yet you progressing with this?

fkiraly:
> We havent come to an agreement, yet. Let us have a vote first, please.

Votes happen on PR, someone would have to vote this down to initiate the vote

fkiraly:
from internal slack:

image

not a 2/3 majority either way.

We should have voted formally on this PR with names to opinions but didn't have any votes here in the end.

Either way, I suppose this is now going to CC

patrickzib:
> Either way, I suppose this is now going to CC

Ok. Why did we have this vote at all, if the CC finally decides? And if the person that starts the PR can easily push it through the CC?

patrickzib:
And IM:  It would be best, if numba checks could only happened at the import level, e.g. the __init__ files.

fkiraly:
> Ok. Why did we have this vote at all, if the CC finally decides?

Standard process. If the outcome is not clear, it goes to phase 3.
https://www.sktime.org/en/stable/get_involved/governance.html#stage-3-conflict-resolution

fkiraly:
> And imo: It would be best, if numba checks could only happened at the import level, e.g. the init files.

I think that is the most sensible option for "loose functions".

For estimator classes, the numba checks should be looped through the class init, using the `python_dependencies` tag.

patrickzib:
I am formally voting this down for missing deprecation warnings. 

My code runs fine under sktime <= = 0.141 on gogole colab, but is broken with this branch due to a mising `numba` libarary.  You must deprecate all affected files first.

These may then go in with sktime 0.16, if acccepted.

patrickzib:
Sorry, it is on you to prove that this PR does not break anything. Until then=> deprecate

patrickzib:
> can you please give an example of code that no longer runs on google colab?
> 
> All tests are passing, and I think there are no public interfaces affected as far as I can see.
> 
> Generally, isolating a dependency should be fully preserving public interfaces, i.e., not require deprecation.

Btw you just admittedly said that this PR bares no functionallity and thus just adds unnecessary boilerplate code? Everthing runs as before. Thus, we can close it.

fkiraly:
@patrickzib, please stop this.

patrickzib:
1. we are between public holidays and everyone is on vacation!
2. We agreed to pause everything until the end of the numfocus investigation.
3. majority vote of core devs already voted this down. I do not see any change in this
4. I am still voting this down.

fkiraly:
Detractors of numba/python 3.11 have announced they will be forking, so there are no longer downvotes to this.

Issue #3567, opened 2022-10-10
[ENH] turning `numba` into a soft dependency


fkiraly:
I think we should turn `numba` into a soft dependency, since it is not core to the framework, but used only in specific estimators.

---

key arguments for:

* the only key change is giving users the choice `numba` yes/no
* for users with `numba` installed, the `sktime` experience remains precisely the same
  * users without `numba` will be able to use `sktime`, currently they cannot
  * users without `numba` will still be able to find `numba` based estimators using `all_estimators`, similar to other soft dependencies
* it should be easier to upgrade `sktime` to newer python or package versions, of core dependencies, as `numba` is slow to update compatibility
* makes python 3.11 compatibility possible, at current date
* architectural: reduces coupling, increases layer separation; minimize soft dependencies, see context below

key arguments against:

I don't really see any except the work required

---

Update: more context on architectural and usage benefits.

the general desideratum is "minimize dependencies". This is an explicit architectural north star of scikit-learn as well as sktime.

See e.g., this paper on sklearn design,
https://mblondel.org/publications/lbuitinck-ecmlpkdd2013.pdf
search for "dependencies" in the text.

In the past 6 months, we have made the interfaces of estimators much clearer, including that we can manage python versions and package dependencies on the estimator level, rather than at the sktime package level

This has prompted multiple lines of refactoring around "minimize dependencies":

* migrate sktime-dl into sktime because we can isolate tensorflow to estimators
* isolate `statsmodels`
* isolate `numba`
* newly added annotation module soft dependencies are fully isolated

`statsmodels` isolation and soft dep is very similar, see this PR: https://github.com/sktime/sktime/pull/3821

---

FAQ

> does this remove `numba` support from `sktime`?

No, `numba` will still be supported and used in estimators.
It will only change to be optional, for people who do not want (or need) to use estimators that depend on `numba`.

> will this impact performance of `sktime`?

No. For users who have `numba` installed, the experience is precisely the same. The logic of `numba` based estimators stays the same.

No change in estimator accuracy or speed. (evidence: look at the tests in https://github.com/sktime/sktime/pull/3843 which test for both)

The only change is that users get a choice whether they want `numba` or not.

> How can this be implemented within the estimators? Numba requires tags/decorators at each function to be compiled.

See https://github.com/sktime/sktime/pull/3631

strategy is the same as any dependency isolation on estimator level and uses the `sktime` base class layer:

* move `numba` imports into functions and classes, where possible
[...]

PR #3634, opened 2022-10-23
[MNT] Move BaseEarlyClassifier and refactor dists_kernels distances into subpackage

MatthewMiddlehurst:
Discussion in #3567 showed that early classifiers are being imported through the `BaseEarlyClassifier` class, as they share an `__init__.py`. In this case, these imports cause numba functions to be compiled, which fail to cache and cause an error likely due to the user's write permissions. 

This PR moves `BaseEarlyClassifier` to the main classification package `sktime/classification/` with a deprecated wrapper in the old location to prevent estimators from being imported with it.

The `dists_kernels` package base classes also have this issue, with multiple distances being imported in the same `__init__.py`. This PR moves these distances to a `sktime/dists_kernels/distances/` subpackage leaving deprecated wrappers in the old location. The base classes have been moved to `sktime/_dists_kernels_base/` temporarily so it can be imported by the registry without the distances, with a temporary wrapper in the old location. The base class can be moved back when the distances have been moved.

There are a lot of files changed, but it is mostly movement of files and changing imports to the new location.


ngupta23:
Thanks @MatthewMiddlehurst!

Hi Reviewers, if this resolves the write issue in #3567, could we please expedite this PR? We would love to use this latest version of sktime for our development :).

Thanks!

MatthewMiddlehurst:
Moved the base class to `sktime.classification`. This means it does not have to go into any classification subpackages and pick up a bunch of random estimators along the way. 

I am running this code with numba uninstalled, which appears to reproduce the issue well:
```
from sktime.forecasting.compose import ForecastingPipeline
from sktime.transformations.series.adapt import TabularToSeriesAdaptor

if __name__ == "__main__":
    y, X = load_longley()
    y_train, _, X_train, X_test = temporal_train_test_split(y, X)
    fh = ForecastingHorizon(X_test.index, is_relative=False)
    pipe = ForecastingPipeline([
        Imputer(method="mean"),
        TabularToSeriesAdaptor(MinMaxScaler()),
        NaiveForecaster(strategy="drift"),
    ])
    pipe.fit(y_train, X_train)
    y_pred = pipe.predict(fh=fh, X=X_test)
    print(y_pred)
```

Unfortunately, looks like the same issue is present for a distances base class.
```
  File "D:\CMP Machine Learning\sktime-workshop-boss\sktime\_contrib\local_code2.py", line 5, in <module>
    from sktime.forecasting.compose import ForecastingPipeline
  File "D:\CMP Machine Learning\sktime-workshop-boss\sktime\forecasting\compose\__init__.py", line 36, in <module>
    from sktime.forecasting.compose._pipeline import (
  File "D:\CMP Machine Learning\sktime-workshop-boss\sktime\forecasting\compose\_pipeline.py", line 12, in <module>
    from sktime.registry import scitype
  File "D:\CMP Machine Learning\sktime-workshop-boss\sktime\registry\__init__.py", line 4, in <module>
    from sktime.registry._base_classes import (
  File "D:\CMP Machine Learning\sktime-workshop-boss\sktime\registry\_base_classes.py", line 67, in <module>
    from sktime.dists_kernels._base import (
  File "D:\CMP Machine Learning\sktime-workshop-boss\sktime\dists_kernels\__init__.py", line 9, in <module>
    from sktime.dists_kernels.dtw import DtwDist
  File "D:\CMP Machine Learning\sktime-workshop-boss\sktime\dists_kernels\dtw.py", line 10, in <module>
    from sktime.distances import pairwise_distance
  File "D:\CMP Machine Learning\sktime-workshop-boss\sktime\distances\__init__.py", line 33, in <module>
    from sktime.distances._distance import (
  File "D:\CMP Machine Learning\sktime-workshop-boss\sktime\distances\_distance.py", line 8, in <module>
    from numba import njit
ModuleNotFoundError: No module named 'numba'
```

MatthewMiddlehurst:
Our culprits:
image

Code runs fine if those two are removed alongside the early classification change. Unfortunetly, it doesnt look like the base classes here could move anywhere unless it was outside the package, and moving the other distance classes would require deprecation.

MatthewMiddlehurst:
In theory, we could move the two base classes to a new temporary package which is used in the registry, import them from the current base class location like is done for the `BaseEarlyClassifier` changes, and deprecate the distance classes to move them to a subpackage.

That seems like a lot of effort though

sktime/dists_kernels/__init__.py
```diff
@@ -5,6 +5,8 @@
     BasePairwiseTransformer,
     BasePairwiseTransformerPanel,
 )
+
+# todo remove these imports in 0.15.0
 from sktime.dists_kernels.compose_tab_to_panel import AggrDist, FlatDist
 from sktime.dists_kernels.dtw import DtwDist
```
patrickzib, line 11: My two cents: Would it be possible to have a soft_dependency check for numba around these conflicting imports? Or push this soft_dependency check somewhere down the stack-trace

patrickzib: E.g. avoid exporting these modules if numba is not installed?

MatthewMiddlehurst:
I have made the changes from my above comment and updated the main PR text. If merged, this should fix the issue @ngupta23 has in theory, though I don't have the means to test their exact scenario.

This probably doesn't catch all cases like this, but ideally we should be isolating the base classes so other estimators are not imported through the registry/testing/whatever. Creating subpackages for `dists_kernels` is probably not a bad move either way as well.

MatthewMiddlehurst:
The 0.14.0 release is soon, so it would be good to have a review (or rejection) prior to then.

ngupta23:
> The 0.14.0 release is soon, so it would be good to have a review (or rejection) prior to then.

If this is released in `sktime` 0.14.0, we can check on our system to see if the issue with `numba` is resolved. Thanks!

MatthewMiddlehurst:
Unfortunately, with the feature freeze tomorrow I don't think this will be in 0.14.0. If the large amount of files changed or the refactor of distances is anyone's concern, that can be moved from this PR.

extension_templates/early_classification.py
```diff
@@ -39,7 +39,7 @@
 
 import numpy as np
 
-from sktime.classification.early_classification import BaseEarlyClassifier
+from sktime.classification.early_classification_base import BaseEarlyClassifier
```
patrickzib, line 42: Why not base.early_classification?

sktime/classification/early_classification/__init__.py
```diff
@@ -10,4 +10,4 @@
     ProbabilityThresholdEarlyClassifier,
 )
 from sktime.classification.early_classification._teaser import TEASER
-from sktime.classification.early_classification.base import BaseEarlyClassifier
+from sktime.classification.early_classification.old_base import BaseEarlyClassifier
```
patrickzib, line 13: Can it stay base without renaming it to old_base? And just contain deprecation stuff?

sktime/dists_kernels/signature_kernel.py
```diff
@@ -10,7 +10,7 @@
 from scipy.sparse.linalg import svds
 from sklearn.base import BaseEstimator, TransformerMixin
 
-from sktime.dists_kernels._base import BasePairwiseTransformerPanel
+from sktime._dists_kernels_base._base import BasePairwiseTransformerPanel
```
patrickzib, line 13: base.dist_kernels?

patrickzib (approved):
I have two suggestions.
1. bundle the base classes within the same folder but use subdirs
2. If possible, avoid the old_base name?

sktime/classification/early_classification/__init__.py
MatthewMiddlehurst: fair point

patrickzib:
> Having the base classes be in the `sktime.base` package was also something I considered suggesting, but it's not a refactor I'm willing to push for a day before release. Same with the current PR, really.
> 
> Every package seems to have a slightly different way for dealing with the base class and `__init__.py` in the top level of the package.

Hmm, so how should we best proceed? Postpone the PR? Split into only the early_classifiaction part without distances?

MatthewMiddlehurst:
See #3660. This PR is essentially option 2 from that issue, will keep it open for now but may close later.

fkiraly (changes requested):
we're not going to move classes around randomly.

* can you please explain how this solves the `numba` issue
* please write a STEP

fkiraly:
formally voting this down. Too many things moving around. Don't create private modules in the root folder.

Please write this up properly, and present at a developer meeting.

> Unfortunetly, it doesnt look like the base classes here could move anywhere unless it was outside the package

How do you conclude that? That sounds wrong.

fkiraly:
I don't think this resolves anything `numba` related?

fkiraly:
> bundle the base classes within the same folder but use subdirs

@patrickzib, this is a major API change, please follow sktime guidance on major API changes.

We can't just do this as an afterthought to a PR, it would break so many things.

I think it's also a bad idea since it increases coupling, decreases cohesion.

fkiraly:
Here's what I think is happening here, @MatthewMiddlehurst, @ngupta23:

@MatthewMiddlehurst is making major changes to `sktime` package structure in an attempt to placate `numba` and prevent issues with imports/compilation.

We don't even know whether this would work, see @TonyBagnall's first post.

What this shows to me is that there are just too many arcane issues caused by the presence of `numba`.

We should *not* be going around `sktime` and change its architecture arbitrarily with the primary goal to fit `numba`. Besides the substantial violations of core designs, it just looks very brittle, who knows when it next break and what it next affects, if we are spreading the ripple effects around the package like this.

Instead, we need to isolate `numba` inside `sktime`, so `sktime` works without `numba`, and keep primary architectural lines intact.

MatthewMiddlehurst:
This is solely for fixing the issue affecting @ngupta23, and I didn't merge or push for it to be merged because I know there are major changes. The changes let you run the code he provided without `numba` even installed, as no estimators are imported through the registry with the base classes.

Closing.

fkiraly:
@MatthewMiddlehurst, why are you closing this?

I asked some valid questions. E.g., why is this fixing the `numba` issue?

Cleaning up the init/import structure is valuable, but we can't do it in a way that breaks things, e.g., by putting new directories in root.

Instead of you closing this, I would have expected you to answer to my concerns.

MatthewMiddlehurst:
I'm not quite sure how putting a new directory in the root breaks things, but if it does, then unfortunately this method will not work.

We can't import from `dists_kernels` without also importing the distances, therefore I changed the package the base classes are imported from temporarily.

MatthewMiddlehurst:
Also I opened #3660 to discuss how best to do this.

Issue #3633, opened 2022-10-23
[BUG] Missing `get_test_params()` class method for DL estimators 


achieveordie:
**Describe the bug**

The default values of DL estimators are chosen as per the recommendation by the author of that estimator. This often results in classifiers/regressors that are big, making it more difficult to test its interface / binding contract in `sktime`. To reduce the testing time, a class method `get_test_param()` is to be used which contains 'smaller' values that will be fed to the constructor for testing purposes.

Take a look at TapNet's [classifier](https://github.com/sktime/sktime/blob/main/sktime/classification/deep_learning/tapnet.py#L235) and [regressor](https://github.com/sktime/sktime/blob/main/sktime/regression/deep_learning/tapnet.py#L216) for how this can be done.

**To Reproduce**

The errors due to the lack of this method are sporadic in nature but have been caught in various runners like #3525, and #3562, where one of the estimators cause the workers to crash. See #3620 for discussion.

**Additional context**

- Apart from decreasing the testing time by decreasing the model size, one also needs to take care of the fact there is significant code coverage from the choices of parameters.
- At least two sets of params should be provided for this. Kindly check the link provided for tapnet for clarity.

fkiraly:
Thanks!

would it be worthwhile to add the two recommendations (low test tim, two param set) to the extension templates?

achieveordie:
> would it be worthwhile to add the two recommendations (low test tim, two param set) to the extension templates?

That'd be great, I don't see an extension template for regression though. I'll quickly add this information to the classification's extension template.

fkiraly:
> That'd be great, I don't see an extension template for regression though.

Oh, that's right! I've opened an issue here so we don't forget.
https://github.com/sktime/sktime/issues/3638

PR #3635, opened 2022-10-23
[ENH] Add important information about `get_test_params()` in the extension template

achieveordie:

This adds two important recommendation when adding parameter set for `get_test_params()` in classification template. See #3633 for further information.

No.

I could copy-paste it to other relevant templates but I was wondering if there was a better way. Since this recommendation is important for only classification at the moment, so I want to at least add it there.

##### For all contributions

##### For new estimators

### DIFF
--- a/extension_templates/classification.py
+++ b/extension_templates/classification.py
@@ -252,9 +252,8 @@ class MyTimeSeriesClassifier(BaseClassifier):
         -------
         params : dict or list of dict, default = {}
             Parameters to create testing instances of the class
-            Each dict are parameters to construct an "interesting"
-            and viable test instance, i.e., `MyClass(**params)` or
-            `MyClass(**params[i])` creates a valid test instance.
+            Each dict are parameters to construct an "interesting" test instance, i.e.,
+            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
             `create_test_instance` uses the first (or only) dictionary in `params`
         """
 
@@ -275,12 +274,13 @@ class MyTimeSeriesClassifier(BaseClassifier):
         #   i.e., "parameter_set not available" errors should never be raised.
         #
         # A good parameter set should primarily satisfy two criteria,
-        #   1. Chosen set of parameters should reduce the testing time.
-        #       This is vital for the case of Deep Classifiers where default values
-        #       lead of "big" models which not only increases test time but also
+        #   1. Chosen set of parameters should have a low testing time,
+        #      ideally in the magnitude of few seconds for the entire test suite.
+        #       This is vital for the cases where default values result in
+        #       "big" models which not only increases test time but also
         #       run into the risk of test workers crashing.
-        #   2. There should be minimum two such parameter set with different values
-        #      to ensure maximum code coverage.
+        #   2. There should be a minimum two such parameter sets with different
+        #      sets of values to ensure a wide range of code coverage is provided.
         #
         # example 1: specify params as dictionary
         # any number of params can be specified
--- a/extension_templates/classification.py
+++ b/extension_templates/classification.py
@@ -252,8 +252,9 @@ class MyTimeSeriesClassifier(BaseClassifier):
         -------
         params : dict or list of dict, default = {}
             Parameters to create testing instances of the class
-            Each dict are parameters to construct an "interesting" test instance, i.e.,
-            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+            Each dict are parameters to construct an "interesting"
+            and viable test instance, i.e., `MyClass(**params)` or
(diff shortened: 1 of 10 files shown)


extension_templates/classification.py
```diff
@@ -252,8 +252,9 @@ def get_test_params(cls, parameter_set="default"):
         -------
         params : dict or list of dict, default = {}
             Parameters to create testing instances of the class
-            Each dict are parameters to construct an "interesting" test instance, i.e.,
-            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+            Each dict are parameters to construct an "interesting"
+            and viable test instance, i.e., `MyClass(**params)` or
```
fkiraly, line 256: what do you mean by "viable"?

Also, if you change it here, we need to change it in all other instances of this docstring.

extension_templates/classification.py
```diff
@@ -273,6 +274,14 @@ def get_test_params(cls, parameter_set="default"):
         # A parameter dictionary must be returned *for all values* of parameter_set,
         #   i.e., "parameter_set not available" errors should never be raised.
         #
+        # A good parameter set should primarily satisfy two criteria,
+        #   1. Chosen set of parameters should reduce the testing time.
```
fkiraly, line 278: "reduce" is vague, since it refers to a baseline point which is unclear.

Better: "keep testing time low"

"low" is also vague, so try to say sth about the order, e.g., less than a second for the entire test suite?

extension_templates/classification.py
```diff
@@ -273,6 +274,14 @@ def get_test_params(cls, parameter_set="default"):
...
+        #       This is vital for the case of Deep Classifiers where default values
```
fkiraly, line 279: this sentence is too specific to DL classifiers, and also true for all other classifiers, so I would leave it out. It is vital in general

extension_templates/classification.py
```diff
@@ -273,6 +274,14 @@ def get_test_params(cls, parameter_set="default"):
...
+        #       lead of "big" models which not only increases test time but also
+        #       run into the risk of test workers crashing.
+        #   2. There should be minimum two such parameter set with different values
```
fkiraly, line 282: bit odd english, perhaps reformulate?

fkiraly (changes requested):
Great!

Left some small comments, mainly about increasing clarity.
We can add this to all other extension templates too, when these are addressed.

fkiraly:
> I could copy-paste it to other relevant templates but I was wondering if there was a better way. Since this recommendation is important for only classification at the moment, so I want to at least add it there.

I have wondered the same.

The right kind of mechanism would be inheritance, but there isn't such a mechanism for documents or "fill-in templates" (that I know of).

extension_templates/classification.py
achieveordie: By viable I meant the set should also be feasible, certain interesting cases might require more resources than, let's say, the default set. This set should be avoided since they are not viable.

But this distinction is made again in the second half, so I'll revert back to what it originally was.

extension_templates/classification.py
achieveordie: How about "lower testing time than the default set"? I think it'll be better than giving a value because otherwise, it sounds like this should actively be enforced (using something like `timeit`).

A lower-than-default time should suffice.

extension_templates/classification.py
achieveordie: You're right! Changed it to be type/class-agnostic via `This is vital for the cases where default values result in "big" models which..`

extension_templates/classification.py
achieveordie: I'm confused about whether you're referring to the first point or the second point (or maybe both?)

extension_templates/classification.py
fkiraly: "lower" can still be hours, I think we should tell people what they should approximatel be aiming for (but also imply it's approximate and order of magnitude)

extension_templates/classification.py
fkiraly: -> there should be *a* minimum
-> parameter *sets*
-> "maximum code coverage" is vague, unspecific

extension_templates/classification.py
achieveordie: Was waiting to clarify other points, updated.

achieveordie:
@fkiraly LMK if these clarifications resolve the initial issues. If they do, I'll add this content to all other relevant templates.

fkiraly:
great, all good! Let's copy-paste...

GuzalBulatova (approved):
Great job, @achieveordie! 👍  I also don't like the amount of copying but since there's no better solution atm this looks all good to me.

fkiraly (approved):
Thanks!

PR #3636, opened 2022-10-23
[ENH]  `TSFreshFeatureExtractor` cleanup, tests, and docstring

### LINKED ISSUE #3526: [ENH] `tsfresh` transformers - improved signature, custom features, docstring
update by @fkiraly:

`TSFreshFeatureExtractor` is not in a good state, we need:

* a clean and clear docstring, which also describes parameters, and explains pre-defined and user specified parameter sets.
* a parameter that allows the user to specify the features they would like to extract (this may already exist)
* possibly some changes in what the parameters are to make it more user friendly (may need deprecation)
* `get_test_params` with at least two test cases, one with prespecified parameter set and one with a custom specified parameter set.

---

Hi, I'm not sure if my issue is related to something not implemented or if I just don't know how to perfom it.
Basically, I would like to use the extract_features() method of tsfresh just to extract a subset of features defined in a dictionary, which would be different than the 'minimal','efficient' or 'comprehensive' option. 
If possible still using default parameters for this subset of features: 

```
from sktime.transformations.panel.tsfresh import TSFreshFeatureExtractor

features_to_calc = [
    'absolute_sum_of_changes',
    'absolute_maximum',
    'agg_linear_trend',
    'large_standard_deviation',
    'longest_strike_above_mean',
    'longest_strike_below_mean',
    'mean_abs_change',
    'mean_change',
    'quantile',
    'variance_larger_than_standard_deviation',
    'variance',
]

features_dict = dict.fromkeys(features_to_calc, None)
```

If I try the following, it extracts every features of the default or other settings (which is obviously very long except for minimal setting), without considering my dictionary (name of features as keys, None as values)

```
from sktime.transformations.panel.tsfresh import TSFreshFeatureExtractor
tsfresh = TSFreshFeatureExtractor(kind_to_fc_parameters=features_dict)

X_transform = tsfresh.fit_transform(panel_data)
```

Am I missing something or is it just not implemented yet for consistency with other transformers?

Would it be possible to extend tsfresh extract_features() method to panel dataset in sktime, for instance using the following:

```
from sktime.transformations.panel.tsfresh import TSFreshFeatureExtractor
tsfresh = TSFreshFeatureExtractor(kind_to_fc_parameters=features_dict)

X_transform = tsfresh.extract_features(panel_data)
```

I referred to  https://github.com/blue-yonder/tsfresh/blob/main/notebooks/examples/03%20Feature%20Extraction%20Settings.ipynb
but only the not so convenient tsfresh dataformat are supported, so maybe instead I should transform my data to fit in one of the tsfresh format? If this case is the easiest, I could use a few lines to help since it is not clear how could i get there with the page https://www.sktime.org/en/stable/examples/AA_datatypes_and_datasets.html

**Additional context**
Relative to that alternative solution, completing the datasets types notebook referred previously with a canva, to convert from one or a few tsfresh format to one or a few sktime format would be ideal (only for a few ultivariate options which are more convenient from one package to the other...)

Thanks very much in advance for your help, and for your work,

Julien

kcc-lion:
#### Reference Issues/PRs
Attempt to fix #3526

- Fixes parameter passing logic of `TSFreshFeatureExtractor` (previously, as mentioned in #3526, the user-specified feature set was overriden by the `default_fc_parameters` argument)
- Improved docstring
- Added test cases (also for user-selected features via `kind_to_fc_parameters`)

No

##### For all contributions

### DIFF
--- a/sktime/transformations/panel/tsfresh.py
+++ b/sktime/transformations/panel/tsfresh.py
@@ -33,7 +33,7 @@ class _TSFreshFeatureExtractor(BaseTransformer):
 
     def __init__(
         self,
-        default_fc_parameters="efficient",
+        default_fc_parameters=None,
         kind_to_fc_parameters=None,
         chunksize=None,
         n_jobs=1,
@@ -83,6 +83,7 @@ class _TSFreshFeatureExtractor(BaseTransformer):
             ComprehensiveFCParameters,
             EfficientFCParameters,
             MinimalFCParameters,
+            from_columns,
         )
         from tsfresh.utilities.dataframe_functions import impute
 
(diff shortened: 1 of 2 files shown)


MatthewMiddlehurst:
Can take a look at this tomorrow if no one else wants to pick it up.

sktime/transformations/panel/tsfresh.py
```diff
@@ -127,15 +128,101 @@ def _get_extraction_params(self):
...
 
 class TSFreshFeatureExtractor(_TSFreshFeatureExtractor):
     """Transformer for extracting time series features.
 
+    Parameters
+    ----------
+    default_fc_parameters : string, FCParameters object or None
+        Specifies pre-defined feature sets to be extracted
+        If string should be in ["minimal", "efficient", "comprehensive"]
+        See [3] for more details.
```
MatthewMiddlehurst, line 148: The default value should be included, i.e. `default_fc_parameters : string, FCParameters object or None, default=None`. In this case it would be good to include what None actually does as well. 

I wouldn't be afraid to take directly from the `tsfresh` docstrings for these if they are available.

sktime/transformations/panel/tsfresh.py
```diff
@@ -33,7 +33,7 @@ class _TSFreshFeatureExtractor(BaseTransformer):
 
     def __init__(
         self,
-        default_fc_parameters="efficient",
+        default_fc_parameters=None,
```
MatthewMiddlehurst, line 36: If `None` has a different output from `"efficient"` this will require deprecation.

MatthewMiddlehurst (changes requested):
I assume the docstring tests need to be skipped due to the soft dependency? Good idea replacing it with a test if so.

It would be good if there was also a test for `kind_to_fc_parameters` to ensure the correct amount of features are being returned in the transform.

MatthewMiddlehurst (changes requested):
I see that None for the `default_fc_parameters` parameter uses the "comprehensive" set rather than "efficient".

Either the parameter should remain as "efficient", or the parameter value should be deprecated. I can give some guidance if you want to do the latter.

MatthewMiddlehurst:
The test failures seem unrelated to this PR.

sktime/transformations/panel/tsfresh.py
```diff
@@ -127,15 +127,104 @@ def _get_extraction_params(self):
...
+        containing strings specifying selected features to be extracted.
+        The naming convention from tsfresh applies, i.e. the strings
+        should be structured as:
+        {time_series_name}__{feature_name}__{param name 1}_
+        {param value 1}__[..]__{param name k}_{param value k}.
+        See [2] for more details and [4] for viable options.
+        Either default_fc_parameters or kind_to_fc_parameters
+        should be passed. If both are passed, only features specified
+        in kind_to_fc_parameters are extracted. If neither
+    n_jobs : int, default=1
```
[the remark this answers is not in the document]
kcc-lion: Thanks! I fixed this doc-error in the commit below. Some guidance on how to do deprecation would be much appreciated. I think that calculating the "comprehensive" set if None is passed is reasonable, as it is in line with the behavior of tsfresh.

fkiraly:
> Thanks! I fixed this doc-error in the commit below. Some guidance on how to do deprecation would be much appreciated. I think that calculating the "comprehensive" set if None is passed is reasonable, as it is in line with the behavior of tsfresh.

Guide for deprecation is here:

High-level, the estimator needs to behave as current with an combination of inputs, and raise a deprecation warning message in every case behaviour will change.

For current, `None` would have to default to `"efficient"`, but you can raise a deprecation warning that this will change to `"comprehensive"` in, I suppose, 0.16.0 (because the earliest this is in `sktime` is 0.14.1, I think)

fkiraly:
fixed the minor outstanding issues (docstring, deprec message change only) so this can go in the release

fkiraly:
@MatthewMiddlehurst, I think your issues are addressed?

MatthewMiddlehurst (approved):
yes

PR #3652, opened 2022-10-26
[ENH] Option to fit `Clearsky` transformer in parallel 

ciaran-g:
#### Reference Issues/PRs

No specific issue

Option to fit the `ClearSky` transformer in parallel
* This is done with joblib
* Updated the docstrings with new options

No


MatthewMiddlehurst:
Is the output `self.clearskypower` the same with the new code as it was previously? I assume so, but good to check.

There is a bit of inconsistency with the new doc options vs the old, i.e. `min_thresh : float, default=0` vs `n_jobs: int, optional (default=None)`. I don't believe there is a consistent way to do it throughout the package, but it should be consistent in the same file at least.

MatthewMiddlehurst:
Unrelated to this PR specifically, but we should probably add a `backend` parameter to other estimators which use joblib and start using the `prefer` argument. May require some evaluation to determine which backend if preferred for each, though.

fkiraly:
FYI @topher-lo, re (in)consistency with current efforts to introduce `dask` and multiprocessing?

ciaran-g:
> Is the output `self.clearskypower` the same with the new code as it was previously? I assume so, but good to check.

Added a wee check for this

> There is a bit of inconsistency with the new doc options vs the old, i.e. `min_thresh : float, default=0` vs `n_jobs: int, optional (default=None)`. I don't believe there is a consistent way to do it throughout the package, but it should be consistent in the same file at least.

Updated the docstrings 👍

MatthewMiddlehurst (dismissed):
Looks good assuming all the tests pass. Won't merge this week due to the release, and ill give a bit of time for others to comment if they want.

FYI there appears to be an issue with the data loaded in the test currently, see #3663. Won't request any changes in this PR for that though unless you want to make them yourself.

ciaran-g:
> FYI there appears to be an issue with the data loaded in the test currently, see #3663. Won't request any changes in this PR for that though unless you want to make them yourself.

Hmm ok, maybe for the tests it would make sense to just store the data in a csv file or something and load them that way rather than calling the api?

MatthewMiddlehurst:
Yes, `sktime.data.datasets` contains some dataset files used mostly for testing, but not this one possibly? If none of the already present datasets fulfil the requirements for the test, it may be worth adding it.

MatthewMiddlehurst:
i.e. `load_unit_test` will load the file from there.

ciaran-g:
> Yes, `sktime.data.datasets` contains some dataset files used mostly for testing, but not this one possibly? If none of the already present datasets fulfil the requirements for the test, it may be worth adding it.

Added the local dataset and the option for loading it via `load_solar(api_version=None)`. Also made some changes to use the new file in the tests for `load_solar`

ciaran-g:
> 



> Looks good assuming all the tests pass. Won't merge this week due to the release, and ill give a bit of time for others to comment if they want.

Any wiggleroom on this? Would be keen for it to be in the new release to make sharing the update easier

MatthewMiddlehurst (dismissed):
Those changes look good to me. `test_load_solar` may still run in to the same issue, but that's not a problem for this PR.

At this stage I'm pretty sure only PRs required for the release will go in, sorry.

fkiraly (changes requested):
There would be wiggle room, but it seems like the tests are failing, sth around loading the dataset?
We need to understand what is going on here before we merge, so it will probably not go in 0.14.0.

sktime/transformations/series/clear_sky.py
```diff
@@ -40,17 +41,24 @@ class ClearSky(BaseTransformer):
 
     Parameters
     ----------
-    quantile_prob : float, default=0.95
```
fkiraly, line 43: I think there should be a space on both sides. Currently not testing but it's numpydoc style afaik

fkiraly (approved):
Hm, the failure seems to be unrelated to this PR, so this can be merged.

But, can you kindly check the bug separately then, @ciaran-g?

fkiraly (changes requested):
I think I was wrong, the failures seem genuinely related to the csv added in this PR?

topher-lo:
> FYI @topher-lo, re (in)consistency with current efforts to introduce `dask` and multiprocessing?

No inconsistency. The idea that we should have a standard `backend` argument across all parallelized estimators (unless we decide that only "threading" should be used e.g. for code that readily releases the GIL) is great. Also having `prefer="threading"` is generally a good idea for most data science situations (unless we know beforehand that the code is GIL hungry e.g. in `evaluate`).

@MatthewMiddlehurst Perhaps we should start a GitHub issue tracker for all uses of joblib so far? Try to standardize the signatures across `sktime`.

ciaran-g:
> There would be wiggle room, but it seems like the tests are failing, sth around loading the dataset? We need to understand what is going on here before we merge, so it will probably not go in 0.14.0.

No worries, I hard-coded the path to the csv so it would only work on windows (lol) should be fixed now!

fkiraly (approved):
Looks good now, hopefully fixes the loader test failure?

Issue #3649, opened 2022-10-26
[ENH] Include uniform scaling technique for uneven length time series


equ1:
Hi,

I have a multivariate classification problem where I'm trying to classify an uneven length time series, but zero padding would not work well for me since the difference between the smallest and largest series is large (550 to 14000). It is well-documented that uniform scaling works better for these kinds of problems (Keogh 2005 - https://link.springer.com/content/pdf/10.1007/b13634.pdf), so I was surprised to find that there is no implementation of this in sktime.

If I've missed an implementation of this in the documentation, apologies.

Thanks

fkiraly:
I think `TSInterpolator` might do what you are looking for, but it is badly documented...

In either case, a PR to improve the docstring, e.g., adding scientific reference to contexts it appears or names it appears under would be appreciated.

fkiraly:
what might also be great is to add support for `pd-multiindex` internally, since `nested_univ` has some issues with recent `pandas` versions

equ1:
Thanks, I'll try out TSInterpolator. On a related note, I tried to run `KNeighborsTimeSeriesClassifier` with dtw on my dataset (pd-multiindex) and it said that it does not support uneven length series despite the algorithm traditionally being able to do so. Is this a Numba thing? It's really hard to find which classifiers work out of the box for uneven length series (I tried `all_estimators(as_dataframe=True, filter_tags={"capability:unequal_length:removes":True})` and `all_estimators(as_dataframe=True, filter_tags={"capability:multivariate":True}, estimator_types="classifier")` but it didn't give me what I'm looking for).

MatthewMiddlehurst:
Hi, I don't think any classifiers can handle unequal length currently.  `print(all_estimators(filter_tags={"capability:unequal_length": True}, estimator_types="classifier"))` will just give you the DummyClassifier, which is not very helpful.

Handling of unequal length data for those where it is possible is a good goal for the classifiers, and contributions would be welcome.

fkiraly:
>  On a related note, I tried to run KNeighborsTimeSeriesClassifier with dtw on my dataset (pd-multiindex) and it said that it does not support uneven length series despite the algorithm traditionally being able to do so.

Well, you can always substitute a distance that can handle unequal length time series, and then the classifier should be able to handle unequal length time series.

I just tried it but there was a bug where the tags accidentally didn't let you do this but I fixed it:
https://github.com/sktime/sktime/pull/3654

fkiraly:
[...]

Issue #2774, opened 2022-06-09
[BUG] `KNeighborsTimeSeriesClassifier` fails with any algorithm except `brute`


fkiraly:
`KNeighborsTimeSeriesClassifier` has one of two bugs:

* a docstring that incorrectly claims that an `algorithm` can be set. Currently, it is set to `brute` internally (when calling the sklearn algorithm), no matter what the hyperparameter is.
* failing with algorithms other than `brute`, e.g., `auto` (the sklearn default) when the implementation is fixed to match the docstring.

So, either, the docstring should be corrected and refrain from claiming that more than one algorithm can be used internally; or, the `algorithm` needs to be properly passed through, and the failures with algorithms other than `brute` fixed.

AurumnPegasus:
@fkiraly I was actually editing the ```KNeighborsTimeSeriesClassifier``` according to #2733 (to remove the kwargs) and noticed the same! For now I simply added ```algorithm``` as a parameter (which can simply be removed for the sake of this bug). Please update me on the approach taken for this bug so that I could reflect it as well.

fkiraly:
I have changed the default to `brute`, so one can set the algorithm, but it will break when choosing anything different

MatthewMiddlehurst:
This was fixed in the refactor, correct?

fkiraly:
hm, I see that the test parameters do not check for this, so we don't really know

PR #3654, opened 2022-10-26
[ENH/BUG] knn estimators for unequal length distances

fkiraly:
This PR extends knn estimators' functionality to work with unequal length distances, and fixes some tag behaviour that would have prevented it:

* `capability:unequal_length` tag is added in the registry for distances
* `capability:unequal_length = True` is set for the aligner adapter
* the two knn estimators (classifier and regresor) have all capability tags set to true to indicate the potential functionality
* the tags are dynamically set based on how distances are passed, in particular it inherits all capability tags from a pairwise panel transformer

In the sum, this allows to plug in distance for unequal length series.

The two examples from https://github.com/sktime/sktime/issues/3649 are added as tests.


TonyBagnall:
```from sktime.classification.distance_based import KNeighborsTimeSeriesClassifier
from sktime.datasets import load_plaid
from sktime.datasets import load_japanese_vowels

trainX, trainy = load_plaid(split="train")
testX, testy = load_plaid(split="test")
```
fails with this
image

Also both of these fail (see #2774)
```from sktime.datasets import load_unit_test
trainX, trainy = load_unit_test(split="train")
testX, testy = load_unit_test(split="test")
knn = KNeighborsTimeSeriesClassifier(algorithm="kd_tree")
knn.fit(trainX, trainy)
preds = knn.predict(testX)
knn = KNeighborsTimeSeriesClassifier(algorithm="ball_tree")
knn.fit(trainX, trainy)
```
image

TonyBagnall (dismissed):
1. This does not run with unequal length data loaded from our baked in datasets, see comment, some tags problem I think 
2. KNeighborsTimeSeriesClassifier does not work with the scikit learn options for algorithm. This option is hard coded in scikit to only accept certain distances, because it relies on the triangle inequality. The algorithm option should be removed from KNN see #2774
3. This seems a over complex approach that relies on a soft dep distance function that is four times slower than our own distances. It would be far easier to just adapt existing distances to allow for variable length, which should be quite easy @chrisholder for most, although it will take some care to handle edge cases
4. Could you add yourself as code owner for KNeighborsTimeSeriesClassifier?
5. I do not like the tight coupling between distances and the classifier. If we add a new distance we have to alter the classifier. The distance function approach was much better imo.

fkiraly:
@TonyBagnall, the two failures that you bring up have nothing to do with this PR, and the first one is expected behaviour.

Let's go through this one-by-one.

Failure no.1:
The distances coming from the `distances` module cannot handle unequal length series.
The default in knn is one of those distances, and you invoke the algorithm with said default.
Thus, the observed behaviour is expected.

You have to stick a distance that can handle unequal length series into the knn algorithm for it to be able to handle unequal length series.

Failure no.2:
This is a known bug and fails on `main` already. It is not affected, and also not fixed, by this PR.

fkiraly:
> 1. This does not run with unequal length data loaded from our baked in datasets, see comment, some tags problem I think 

Explained above. You need to use a distance that handles unequal length data.

> 2. KNeighborsTimeSeriesClassifier does not work with the scikit learn options for algorithm. This option is hard coded in scikit to only accept certain distances, because it relies on the triangle inequality. The algorithm option should be removed from KNN see #2774

Known bug unrelated to this PR, see above.

> 3. This seems a over complex approach that relies on a soft dep distance function that is four times slower than our own distances. It would be far easier to just adapt existing distances to allow for variable length, which should be quite easy @chrisholder for most, although it will take some care to handle edge cases

Non-sequitur. You can put any distance in there.

There is no change in runtime to the existing options, and if you want to implement a faster distance, go ahead. Out of scope.

> 4. Could you add yourself as code owner for KNeighborsTimeSeriesClassifier?

not the current owner, requires consent of current owners, incl @goastler and @jasonlines .

> 5. I do not like the tight coupling between distances and the classifier. If we add a new distance we have to alter the classifier. The distance function approach was much better imo.

I think you misunderstand. The current design where you can plug in arbitrary distances has lower coupling than a design where distances are hard-coded and are retrieved.

Having said that, the current PR does not alter the design in comparison to `main`, so it is not a point that applies to this PR.

fkiraly:
@TonyBagnall, can you please look at my reply? I don't think the points you raised relate to this bugfix.

TonyBagnall:
sorry, somehow I edited your comment rather than make a new one, probably was on my phone.
I said
"imo if you list a classifier as having the capability to handle unequal length series, it should handle unequal length series with all valid parameter settings, or at least the default ones. Its a simple fix to make dtw work with unequal length, @chrisholder might do it"
but I wont block this

fkiraly:
> "imo if you list a classifier as having the capability to handle unequal length series, it should handle unequal length series with all valid parameter settings, or at least the default ones. Its a simple fix to make dtw work with unequal length, @chrisholder might do it"

Ah, yes, I think this is a valid point.

But, setting the default parameter to the "minimum" rather to the "maximum" is not current convention, because of how it behaves in retrieval, e.g., by `all_estimators`

In retrieval the class tag is used. Because users are looking for estimators that have a certain special capability, i.e., unequal length support, estimators that *can* have the capability (rather than *always* have it) appear as having the tag.

If the class tag would be the "minimum" rather than the "maximum", users would not find an estimator that can be made to handle unequal length by choosing certain parameters.

I agree that the default setting should have the "good capability", but in this case that would require deprecation or extension.

PR #3659, opened 2022-10-27
[BUG] Fix handling of `missing_values` in `Imputer`

### LINKED ISSUE #3655: [BUG] Imputer fails when `missing_values` and `method="drift"` are set
**Describe the bug**

Imputation method fails when `missing_values=[np.inf]` and `method="drift"` is set. 

**To Reproduce**

```python
import numpy as np

# create timeseries with inf
y = load_airline()
y.iloc[3] = np.inf

# try to impute inf
Imputer(missing_values=[np.inf], method="drift").fit_transform(y)
```

**Expected behavior**

Imputer should work without errors

**Additional context**

**Versions**
<details>

```
System:
    python: 3.8.13 (default, Aug 11 2022, 15:40:13)  [Clang 13.0.0 (clang-1300.0.29.30)]
    machine: macOS-12.5.1-x86_64-i386-64bit

Python dependencies:
          pip: 22.3
   setuptools: 59.8.0
      sklearn: 1.1.2
       sktime: 0.13.4
  statsmodels: 0.13.2
        numpy: 1.22.4
        scipy: 1.8.1
       pandas: 1.4.4
   matplotlib: 3.6.1
       joblib: 1.2.0
        numba: 0.56.2
     pmdarima: 1.8.5
      tsfresh: 0.19.0
```

</details>

AnH0ang:
#### Reference Issues/PRs

Fixes #3655. This PR solves a bug in `sktime.transformations.series.impute.Imputer`. The problem is that abnormal values such as `np.inf` that are passed to the `Imputer` are not handled before calculating metrics such as `std` or fitting forecasters. This leads to an error.

The values that are specified in `missing_values` are imputed with `np.nan` at three different locations for different reasons:
* `_fit` -> Prevent errors when calculating `mean` or `median` with `np.inf` as an input.
* `_get_random` -> Prevent errors when sampling from uniform distribution with `np.inf` as an input.
* `_impute_with_forecaster` -> Prevent errors when `np.inf` is passed to underlying forecasters.

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/transformations/series/impute.py
+++ b/sktime/transformations/series/impute.py
@@ -3,7 +3,7 @@
 # copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
 """Transformer to impute missing values in series."""
 
-__author__ = ["aiwalter"]
+__author__ = ["aiwalter", "AnH0ang"]
 __all__ = ["Imputer"]
 
 
@@ -50,7 +50,7 @@ class Imputer(BaseTransformer):
         "drift", "mean", "median", "random". For all other methods, the
         transform data is sufficient to compute the impute values.
 
-    missing_values : int/float/str, default=None
+    missing_values : int/float/str/list, default=None
         The placeholder for the missing values. All occurrences of
         missing_values will be imputed, in addition to np.nan.
         If None, then only np.nan values are imputed.
@@ -135,10 +135,16 @@ class Imputer(BaseTransformer):
         -------
         self: reference to self
         """
+        X = X.copy()
+
+        # replace missing_values with np.nan
+        if self.missing_values:
+            X = X.replace(to_replace=self.missing_values, value=np.nan)
+
         self._check_method()
         # all methods of Imputer that are actually doing a fit are
         # implemented here. Some methods dont need fit, so they are just
-        # impleented in _transform
+        # implemented in _transform
         if self.method in ["drift", "forecaster"]:
             self._y = y.copy() if y is not None else None
             if self.method == "drift":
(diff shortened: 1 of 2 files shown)


fkiraly (commented):
Good idea!

I think we are mixing different things though:

* handling multiple nan values
* handling infinity

I would do that differently:

* handling multiple nan values can be done as here
* but inf should not be interpreted as nan by default
* for the min/max thing, we could use `np.nanmin` etc?

fkiraly (changes requested):
thanks for the great contribution!

Have you had time to look at my comments?

AnHo4ng:
> Good idea!
> 
> I think we are mixing different things though:
> 
> * handling multiple nan values
> * handling infinity
> 
> I would do that differently:
> 
> * handling multiple nan values can be done as here
> * but inf should not be interpreted as nan by default
> * for the min/max thing, we could use `np.nanmin` etc?

Sorry 😅 I most forgotten about the PR. 

@fkiraly Your suggestions make total sense! `np.inf` values shouldn't just be overwritten by default. I just implemented the changes.

AnH0ang:
> Good idea!
> 
> I think we are mixing different things though:
> 
> * handling multiple nan values
> * handling infinity
> 
> I would do that differently:
> 
> * handling multiple nan values can be done as here
> * but inf should not be interpreted as nan by default
> * for the min/max thing, we could use `np.nanmin` etc?

Sorry 😅 I most forgotten about the PR. 

@fkiraly Your suggestions make total sense! `np.inf` values shouldn't just be overwritten by default. I just implemented the changes.

fkiraly (changes requested):
Thanks for this!

Quick question: do we need to copy at the start?
There don't seem to be side effects?

AnH0ang:
Thank you for the review! I thought that the `.replace` statement afterward changed the data frame but that does not seem to be the case. Just removed the copy.

AnH0ang:
Hey @fkiraly 👋 I just wanted to check if there is something that I can change regarding this PR. I would be happy to do it.

fkiraly (changes requested):
Sorry for letting this lie, apologies. I missed it in the notifications.
Nice, thanks!

* can we avoid making the copy in line 138? The CI (or `check_estimator`) tests for side effects. Worst case, use `pandas.DataFrame.copy` which keeps references to the values but not to the frame.
* @aiwalter needs to have a look at this, since he is listed as author

Issue #3657, opened 2022-10-27
[BUG] pmdarima estimators break when `X` contains more indices than the forecasting horizon


d4nielmeyer:
**Describe the bug**
I have a dataset containing 1 endogenous and 5 exogenous variables and would like to do cross-validation (via ExpandingWindowSplitter) to evaluate several AutoARIMA models. In particular, my intention was to use an initial_window = 3, step_length = 1 and fh = [1, 2, 3] (as illustrated below). 

| * * * x x x - - - |
| * * * * x x x - - |
| * * * * * x x x - |
| * * * * * * x x x |

X (exogenous variables) is of shape (16, 5), while y (endogenous variable) is of shape (16, ). After running temporal cross-validation I get train and test indices of length 3 for the first split -> train [0, 1, 2] and test [3, 4, 5], which is what I would expect. However, within the temporal CV when I split the data into X/y-train/test I get y_train of shape (3, ), y_test (3, ), X_train (3, 5) but X_test (6, 5). Since X_test: n_rows=6 doesn't match n_periods=3 it is not accepted to be passed inside the predict-method (inside the evaluate-method). As a consequence I get the following error notification:

`ValueError('X array dims (n_rows) != n_periods')`

**To Reproduce**

To reproduce the error I used a simple, but similar dataset from the sktime library:

```python
from sktime.datasets import load_longley
from sktime.forecasting.model_evaluation import evaluate
from sktime.forecasting.model_selection import ExpandingWindowSplitter
from sktime.forecasting.arima import AutoARIMA
from sktime.performance_metrics.forecasting import MeanAbsoluteError
import numpy as np

y, X = load_longley()

forecaster = AutoARIMA()
cv = ExpandingWindowSplitter(initial_window=3, step_length=1, fh=np.arange(1, 4))
loss = MeanAbsoluteError()

results = evaluate(forecaster=forecaster, y=y, X=X, cv=cv, error_score='raise', scoring=loss)
```

**Expected behavior**
I am pretty aware of the facts that: 
- If you fit with exogenous, you must predict with exogenous
- When you are predicting with exogenous, your dimensions must match.

But to some extent I would expect that (S)ARIMAX-models were capable of handling inputs at inference time in a way like:
***A)***
 input:
  ------------------     y(t-3)
  ------------------     y(t-2)
  ------------------     y(t-1)
 x1(t)    x2(t)     x3(t)   

output:
 y(t) 
 y(t+1) 
 y(t+2) 
 y(t+3)

[...]

PR #3667, opened 2022-10-29
[BUG] fix pmdarima interfaces breaking for `X` containing more indices than forecasting horizon

fkiraly:
Fixes https://github.com/sktime/sktime/issues/3657.

The bug was caused by `pmdarima` models breaking when the `X` passed was strictly larger than the indices in the forecasting horizon.

The example code in #3657 has been added as a test (with minor generalization to cover more estimators).

In the future, we should probably also add test scenarios where `X` is strictly larger than the forecasting horizon.

Depends on:
* https://github.com/sktime/sktime/pull/4474, which fixes a bug that was masked by #3657
* https://github.com/sktime/sktime/pull/4483 for slicing `X`


fkiraly:
strange: this test fails under 3.10 but not under 3.8

```python

check_estimator(TransformedTargetForecaster, fixtures_to_run="test_predict_time_index_with_X[TransformedTargetForecaster-1-y:2cols-fh=1-range-int-True]", return_exceptions=False)
```

fkiraly:
this seems to be related to `statsmodels`, not sure what is happening there.

Reverting the change to the `statsmodels` adapter and moving to separate issue for investigation.

SzymonStolarski:
Hi @fkiraly,

It's been quite some time since this has been created, however I was wondering if it would be possible to go back to this PR and check the issue with the tests in order to fix #3657?

I've been also playing around with the change in the `pmdarima` adapter proposed in this PR and it looks like the `evaluate` funtion works fine. One thing, however, that I noticed is the fact that this change loses the possibility of quickly passing `X` assumptions without `PeriodIndex` in `predict`, e.g.:

```python3
y = load_airline()
X = pd.DataFrame({
    'exo_1': y*5,
})

forecaster = ARIMA()
forecaster.fit(y=y, X=X) # Here we are passing data the standard way with `PeriodIndex`

# Here we would like to check the predictions passing quickly exo assumptions
forecaster.predict(fh=[1,2,3], X=pd.Series([123, 213, 321]) 
```

Will throw an error `KeyError: "None of [PeriodIndex(['1961-01', '1961-02', '1961-03'], dtype='period[M]')] are in the [index]"`. 

I assume predict would need to check if data passed in `X` contains `PeriodIndex` and based on that decide if:
```python3
# PeriodIndex case
fh_abs = fh.to_absolute(self.cutoff).to_pandas()
if X is not None:
    X = X.loc[fh_abs]
```
or
```python3
# No PeriodIndex case
fh_idx = fh.to_indexer(self.cutoff)
if X is not None:
    X = X.loc[fh_idx]
```

fkiraly:
> I assume predict would need to check

That might be a fix?
Apologies, @SzymonStolarski, for not working on this - the "obvious" fix didn't work and failures were python version dependent.

Maybe we should have a look how this behaves after the `pandas 2` compatibility upgrade.

Also FYI @yarnabrina, who has recently been working on the `_StatsModelsAdapter`, perhaps he can spot sth. Return objects are highly idiosyncratic and not subject to a unified interface in `statsmodels` (or `pmdarima` which depends heavily on `statodels`).

PR #3675, opened 2022-10-30
[GOV] CC decision making, arbitration by CoCC

fkiraly:
This PR is a proposal following my action to clarify certain points in the governance document. 

As it was discussed in one of the last CC meetings, the original #3154 was contentious in its changes to the algo maintainer role, this should be the uncontentious part (@aiwalter found the first part on algo maintainers contentious).

Changes as below.

Clarification of core developer role:
    * common expectations are outlined, to separate this from algorithm maintainer role better
    * outlined expectations do not change the formal appointment process

New/changed rules:
* inactivity rule for algorithm maintainer, similar to core developer
* it is clarified that CoC committee can (and is expected to) judicate on how various rules are being interpreted. This is to avoid deadlocks and back/forth as on occasion in the past where core devs would disagree on interpretations of rules. Note: this widens the role of CoC committee to judication on matters not restricted to CoC violations.
* clarified end of tenure from CC by removal. Increased to 80% majority from current default of 2/3 (means: no removals possible until we have at least 5 CC members and 4 or more of them agree).

Typos:
* CoC section accidentally referred to CC, fixed
* apostrophes fixed

### DIFF
--- a/docs/source/get_involved/governance.rst
+++ b/docs/source/get_involved/governance.rst
@@ -162,6 +162,21 @@ Rights and responsibilities
    * - Nomination
      - They can nominate new core developers, CoC committee members and CC members.
 
+Expectations
+^^^^^^^^^^^^
+
+Without any restriction of eligibility, common expectations on newly appointed core developers are a history of:
+
+* sustained contributions to ``sktime``, over a period of at 3 months or longer
+* contributions that go substantially beyond "algorithm maintainer" responsibilities.
+  In particular, contributions that go substantially beyond work on a single algorithm or a small set thereof.
+* openness to make valuable non-code contributions, e.g., helping new contributors, helping organize events, running project boards
+* responsiveness to issues, bug reports, decision making procedures
+* participation in community meetings and events, e.g., dev sprints, monthly developer meetings
+* an active commitment to the ``sktime`` community beyond immediate personal networks, being welcoming to all contributors
+
+To clarify, the above are common expectations and not formal prerequisities for appointment or eligibility.
+
 Eligibility
 ^^^^^^^^^^^
 
@@ -177,12 +192,14 @@ they have been nominated, there will be a vote by the current core
 developers.
 
 Voting on appointments is one of the few activities that takes
-place on the project’s private communication channels. The vote will be
-anonymous.
+place on the project's private communication channels. The vote will be
+anonymous and should be set up by a member of the community council.
 
 While it is expected that most votes will be unanimous, a 2/3 majority of
 the cast votes is enough. The vote needs to be open for five days excluding
 weekends.
+
+For details on the voting procedure, see :ref:`decision-making` below.
 
 End of tenure
 ^^^^^^^^^^^^^
@@ -200,17 +217,23 @@ Becoming inactive in the above sense means not contributing for the period via:
 * commenting on pull requests or issues
 * attending one of the regular meetings
 
-Becoming active (after becoming inactive) in the above sense requires one of:
+of a scope that goes beyond algorithm maintainer expectations - i.e., PR or issues related to owned algorithms do not count.
 
-* an approved pull request authored by the core developer
-* a contribution to the community that is minuted in one of the regular meetings
+Becoming active (after becoming inactive) in the above sense requires:
+
+* a contribution within the last month that goes beyond algorithm maintainer expectations, e.g.:
+  * an approved pull request authored by the core developer
+  * a substantial contribution to the community that is minuted in one of the regular meetings
+* success via the same voting procedure as for new core developers, with the exception that former core developers can also self-nominate
 
 .. _coc-committee-members:
 
 CoC committee members
 ~~~~~~~~~~~~~~~~~~~~~
 
-CoC members are contributors with special rights and responsibilities.
+CoC committee members are contributors with special rights and responsibilities.
+The CoC committee is the main judicative body of ``sktime``.
+
 The current members of the CoC committee are listed in the
 `CoC <https://www.sktime.org/en/stable/about/team.html>`__.
 
@@ -225,6 +248,16 @@ They are the point of contact for reporting potential CoC incidents.
 
 In addition, they are responsible for maintaining and improving the CoC.
 
+Expectations
+^^^^^^^^^^^^
+
+Without restriction of eligibility:
+* it is generally expected that CoC committee members
+  maintain a high degree of independence of day-to-day matters in the ``sktime`` community,
+  as well as high personal and professional standards of fairness.
+* it is highly advisable that CoC committee members are specifically selected for
+  being unaffected by any potential power dynamics within the ``sktime`` community.
+
 Eligibility
 ^^^^^^^^^^^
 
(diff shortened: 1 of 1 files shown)


docs/source/get_involved/governance.rst
```diff
@@ -177,13 +192,15 @@ they have been nominated, there will be a vote by the current core
 developers.
 
 Voting on appointments is one of the few activities that takes
-place on the project’s private communication channels. The vote will be
-anonymous.
+place on the project's private communication channels. The vote will be
+anonymous and should be set up by a member of the community council.
```
aiwalter, line 196:
```suggestion
anonymous and should be set up by a core-developer or member of the community council.
```

docs/source/get_involved/governance.rst
```diff
@@ -200,17 +217,23 @@ Becoming inactive in the above sense means not contributing for the period via:
...
-Becoming active (after becoming inactive) in the above sense requires one of:
+of a scope that goes beyond algorithm maintainer expectations - i.e., PR or issues related to owned algorithms do not count.
+
+Becoming active (after becoming inactive) in the above sense requires:
 
-* an approved pull request authored by the core developer
-* a contribution to the community that is minuted in one of the regular meetings
+* a contribution within the last month that goes beyond algorithm maintainer expectations, e.g.:
+  * an approved pull request authored by the core developer
+  * a substantial contribution to the community that is minuted in one of the regular meetings
```
aiwalter, line 226:
```suggestion
```

aiwalter: we can remove this bcz below you write there needs to be a vote as for newly appointed core-devs

docs/source/get_involved/governance.rst
```diff
@@ -200,17 +217,23 @@ Becoming inactive in the above sense means not contributing for the period via:
...
+* success via the same voting procedure as for new core developers, with the exception that former core developers can also self-nominate
 
 .. _coc-committee-members:
 
 CoC committee members
 ~~~~~~~~~~~~~~~~~~~~~
 
-CoC members are contributors with special rights and responsibilities.
+CoC committee members are contributors with special rights and responsibilities.
```
aiwalter, line 234:
```suggestion
CoC committee members are contributors with special rights and responsibilities. CoC member should not be members of CC due to conflicts of interest.
```

docs/source/get_involved/governance.rst
```diff
@@ -225,6 +248,16 @@ They are the point of contact for reporting potential CoC incidents.
...
+Expectations
+^^^^^^^^^^^^
+
+Without restriction of eligibility:
+* it is generally expected that CoC committee members
+  maintain a high degree of independence of day-to-day matters in the ``sktime`` community,
+  as well as high personal and professional standards of fairness.
+* it is highly advisable that CoC committee members are specifically selected for
+  being unaffected by any potential power dynamics within the ``sktime`` community.
+
```
aiwalter, line 260: this is just some bla bla in my opinion. Its not binding or anything and does not provide a new information. I would remove it therefore.

docs/source/get_involved/governance.rst
```diff
@@ -324,6 +357,8 @@ expected to resign voluntarily.
 Tenure also ends automatically when a CC member's tenure as core developer ends,
 e.g., through resignation or inactivity.
 
+CC members can be removed from the CC by an 80% majority of CC members.
+
```
aiwalter, line 361:
```suggestion
CC members can be removed from the CC by an 80% majority of core-developers.

```

aiwalter (changes requested):
left some comments

docs/source/get_involved/governance.rst
GuzalBulatova: @aiwalter maybe "by an 80% majority of all cast votes"?

aiwalter: I think thats same bcz CC members are also core-devs always?

aiwalter: Not sure about the 80%, probably a 2/3 majority should be better?

docs/source/get_involved/governance.rst
fkiraly: adopting that rule would, at current, result in either a CC with 2 members (below numfocus threshold) or an empty CoC.

I'm happy with this proposed rule, but let's make it a condition for appointment instead, in the "appointment" paragraph, so current members (i.e., me) are not automatically removed from either committee.

I'm also happy to sharpen the rule as soon as it seems we are in a state where we have enough members for CoC committee and CC, and I feel that the CoC is operating properly (i.e., addressing problematic issues by the book when they occur).

docs/source/get_involved/governance.rst
fkiraly: come to think of it, there are just formalities. If CoCC gets reponsibility for timing the vote etc, should not a member of CoCC have the responsibility to send out all the announcements etc?

Core developers and CC members may have conflict of interest, and even if unintended this can lead to disagreements on the framing of the vote.

docs/source/get_involved/governance.rst
fkiraly: since this is unrelated to CoCC functionality and may require some additional discussion, I moved it out into a new PR https://github.com/sktime/sktime/pull/3697, and added some details.

docs/source/get_involved/governance.rst
fkiraly: I think it's more than blabla.
Because it highlights what behaviours are expected of CoCC members, which in turn tells core developers what to look for in a vote.

But I'm also not to oattached to this part as it does not define process.

docs/source/get_involved/governance.rst
fkiraly: I think it should require both. Not just a vote, but actually having done something within the last month.

docs/source/get_involved/governance.rst
fkiraly: accepted (moved to expectations, because it is a "should")

docs/source/get_involved/governance.rst
fkiraly: replaced by the "should not be members of CC" which is more concrete

fkiraly:
made the changes you requested, @aiwalter. Is this good to go now?

docs/source/get_involved/governance.rst
aiwalter: Did I propose the "should", if yes it was my mistake. I think it cant be a should, a should is meaningless bcz it means its an optional requirement. It has to be a must, otherwise there is no sense to have it here.

docs/source/get_involved/governance.rst
```diff
@@ -553,6 +593,20 @@ the CC tries to resolve the deadlock.
...
+judicate in contentious cases of application of the following rules by the CC:
+
+* voting rules, vote counts, start/end times
+* determining activity/inactivity of a core developer
+* determining unresponsiveness of an algorithm maintainer
+* inclusion and exclusion criteria for algorithms
+
+Judication is upon written appeal of core developers and algorithm maintainers to the CoC committee.
+A ruling by the CoC committee does not halt timelines, but can reverse decisions made.
+
```
aiwalter, line 609: I would propose to remove that part, I think thats not what CoCC is meant to do, CoCC is meant to make sure that CoC is applied. But this here is adding additional power, e.g. judging a core developers activities. Can we kindly remove that section?

aiwalter (changes requested):
In my opinion its almost good to go, just left 2 comments from my side

PR #3678, opened 2022-10-30
[MNT] 0.16.0 deprecation action - change `BaseForecaster.cutoff` to `pd.Index`

fkiraly:
Scheduled deprecation action for 0.16.0.

Changes public `cutoff` attribute of `BaseForecaster` to be `pd.Index`.

Changes:
* changed type of `cutoff` in `BaseForecaster` to `pd.Index`
* replaced occurrences of `cutoff` in tests with `cutoff[0]`
* replaced occurrences of `cutoff` with `cutoff[0]` in estimators that make direct reference: reducers; `UpdateEvery`
* addressing instance in `evaluate` where `cutoff` is returned in the result data frame


fkiraly:
something doesn't quite work with `update_predict` - we probably need to delay this until 0.15.0

fkiraly:
postponed since it does not seem to work out of the box.

Probably `update` will have to be fixed/reworked first.

fkiraly:
was further postponed to 0.16.0.

I think I've identified the issue now, comes from `_format_moving_cutoff_predictions`. Should be fixed (but will need a clean refactor later on).

PR #3685, opened 2022-11-01
[ENH] Fix for initial_window=0 and depreciating "start_with_window" in ExpandingWindowSplittter

### LINKED ISSUE #2678: [ENH] removing redundant parameter `start_with_window` in `ExpandingWindowSplitter`
The parameter `start_with_window` in `ExpandingWindowSplitter` is redundant with `initial_window`.

If `start_with_window` is `False` (a non-default), the `ExpandingWindowSplitter` behaves as if `initial_window` were zero.
However, `initial_window` cannot be set to zero.

I feel the above choice contributes to user confusion.

I think `start_with_window` should just be removed, and `initial_window=0` should be possible, mapping on the prior behaviour of `start_with_window=False`.

The change needs to be accompanied by deprecation, as follows:

* `initial_window=0` should become possible
* `start_with_window` is retained, but a value of `False` raises a deprecation message along the lines of "please use `initial_window=0` instead"

For deprecation conventions in `sktime`, see this guide:

chillerobscuro:
#### Reference Issues/PRs
Fixes #2678 

`ExpandingWindowSplitter` now has consistent behavior when `start_with_window=False` or `initial_window=0`. In the case of the former, a warning is raised that `start_with_window` will be depreciated in version 0.15.0. 

Due to these changes, `SlidingWindowSplittter` can now accept `initial_window=0`, and no other behavior was changed.

I imported `warn` but no real deps.


khrapovs (changes requested):
Would you mind to add a few test cases that would fail on main but pass on your branch? Would be much appreciated!

.all-contributorsrc
```diff
@@ -1819,6 +1832,11 @@
       "contributions": [
         "doc"
       ]
-    }
-  ]
+    },
```
khrapovs, line 1835: where did this block come from? does not seem related to the PR

sktime/forecasting/model_selection/_split.py
```diff
@@ -834,6 +835,9 @@ def __init__(
         self.step_length = step_length
         self.start_with_window = start_with_window
         self.initial_window = initial_window
```
khrapovs, line 837: please, move this line to the else statement that is now absent from the if statement below. and check for `initial_window`, not `self.initial_window`

sktime/forecasting/model_selection/_split.py
```diff
@@ -885,8 +889,10 @@ def _split_for_initial_window(self, y: pd.Index) -> SPLIT_ARRAY_TYPE:
             raise ValueError(
                 "`start_with_window` must be True if `initial_window` is given"
             )
-        if self._initial_window <= self.window_length:
-            raise ValueError("`initial_window` must greater than `window_length`")
+        if self._initial_window != 0 and self._initial_window <= self.window_length:
+            raise ValueError(
```
khrapovs, line 893: is this exception tested?

khrapovs (changes requested):

.all-contributorsrc
fkiraly: seems to be a merge accident, the same is also at the top.

Please make sure your fork is always up-to-date.

sktime/forecasting/model_selection/_split.py
```diff
@@ -834,6 +835,9 @@ def __init__(
         self.step_length = step_length
         self.start_with_window = start_with_window
         self.initial_window = initial_window
+        if self.initial_window == 0:
+            self.start_with_window = False
```
fkiraly, line 839: this block violates the `sklearn` interface conventions since it the self parameters are not equal to the init parameters.

The tests do not catch this, since `initial_window = 0` is not a test case, so please add that as a test case in `get_test_params`.

sktime/forecasting/model_selection/_split.py
fkiraly: can `initial_window` then ever be zero?

chillerobscuro:
Thanks for your reviews, the branch I've been working on is now far out of date and the changes are muddled, so I am going to open a new PR with a clean fix.

chillerobscuro:
New PR: https://github.com/sktime/sktime/pull/3690

PR #3689, opened 2022-11-01
[ENH] permutation pipeline prototype, delegator design

fkiraly:
Alternative design to https://github.com/sktime/sktime/pull/3443 which does not add the functionality to the existing pipelines, but makes it extrinsic as a delegator.

More modular design-wise, and more flexible as it can be applied to any heterogeneous ensemble (as long as it is a forecaster).


sktime/base/_meta.py
```diff
@@ -17,14 +17,6 @@ class _HeterogenousMetaEstimator(BaseEstimator):
     Partly adapted from sklearn utils.metaestimator.py.
     """
 
-    def get_params(self, deep=True):
-        """Return estimator parameters."""
-        raise NotImplementedError("abstract method")
-
-    def set_params(self, **params):
```
miraep8, line 24: If this is causing issues with get/set params in children - maybe worth making in another PR to make sure it gets merged?

sktime/forecasting/compose/_pipeline.py
```diff
@@ -1448,3 +1450,192 @@ def get_test_params(cls, parameter_set="default"):
...
+
+        if permutation is not None:
+
+            inner_estimators = getattr(estimator, steps_arg)
+            estimator_tuples = self._get_estimator_tuples(inner_estimators)
+
+            estimator_dict = {x[0]: x[1] for x in estimator_tuples}
+
+            # check that all permutation are str values
+            if not all(isinstance(item, str) for item in permutation):
```
miraep8, line 1575: might raise unexpected errors if permutation is not a list

sktime/forecasting/compose/_pipeline.py
```diff
@@ -1448,3 +1450,192 @@ def get_test_params(cls, parameter_set="default"):
...
+        params2 = {
+            "estimator": TransformedTargetForecaster(
+                [ExponentTransformer(0.5), NaiveForecaster(), ExponentTransformer(3)]
+            ),
+            "permutation": [
+                "NaiveForecaster",
+                "ExponentTransformer_1",
+                "ExponentTransformer_2",
+            ],
+        }
```
miraep8, line 1639: Request to add more tests around tuning of permute, and tuning of other parameters other than the steps order when the pipeline is wrapped in `Permute`.

sktime/forecasting/compose/_pipeline.py
```diff
@@ -1448,3 +1450,192 @@ def get_test_params(cls, parameter_set="default"):
...
+    >>> cv = ExpandingWindowSplitter(
+    ...     start_with_window=True,
+    ...     fh=fh)
+    >>> forecaster = NaiveForecaster()
+    >>> # check which of the two sequences of transformers is better
+    >>> param_grid = {
+    ...     "permutation" : [["boxcox", "exp", "naive"], ["exp", "boxcox", "naive"]]
+    ... }
+    >>> gscv = ForecastingGridSearchCV(
+    ...     forecaster=forecaster,
```
miraep8, line 1523: I am somewhat confused by this example.  Shouldn't the "forecaster" here be `permuted` from above? Not NaiveForecaster()?

sktime/forecasting/compose/_pipeline.py
```diff
@@ -1448,3 +1450,192 @@ def get_test_params(cls, parameter_set="default"):
...
+    >>> fh = ForecastingHorizon([1, 2, 3])
+    >>> pipe = ForecastingPipeline(
+    ...     [
+    ...         ("boxcox", BoxCoxTransformer()),
+    ...         ("exp", ExponentTransformer(3)),
+    ...         ("naive", NaiveForecaster()),
+    ...     ]
+    ... )
+    >>> # this results in the pipeline with sequence "exp", "boxcox", "naive"
+    >>> permuted = Permute(pipe, ["exp", "boxcox", "naive"])
```
miraep8, line 1503:
```suggestion
    >>> permuted = Permute(pipe, ["exp", "boxcox", "naive"])
    >>> permuted = permuted.fit(y, fh-fh)
    >>> y_pred = permuted.predict()
```
In order for it to match up with my understanding of what this estimator is doing I think it should be like this ^

miraep8 (commented):
Overall I think the implementation works/I do appreciate the separation of logic out of the base class.

However I think adding another layer of abstraction is potentially confusing for the user, and can also add some complexity to tuning if using `Permute`. (should be tested)

sktime/base/_meta.py
fkiraly: precisely what I also thought when I worked on this PR, and that's why I created this other PR https://github.com/sktime/sktime/pull/3708

Glad to see we think along the same lines :-)

sktime/forecasting/compose/_pipeline.py
fkiraly: this just follows 1:1 the logic of @aiwalter's PR.
I agree though we may want to have a more user friendly check though.

sktime/forecasting/compose/_pipeline.py
fkiraly: confused, what you wrote seems to be identical with what I wrote?

sktime/forecasting/compose/_pipeline.py
fkiraly: it should be `permuted` indeed. Why does this not raise an error?

sktime/forecasting/compose/_pipeline.py
miraep8: I thought I had highlighted multiple lines for suggested change.  Basically - the line after this should be permuted.fit()  - yes?

fkiraly: oh, of course

sktime/forecasting/compose/_pipeline.py
miraep8: Sure, I agree that it should work ok.  My main concern with requesting this (and actually perhaps it is better solved by clear documentation rather than a test)  is - before we might have been able to tune pipeline's parameter foo by telling gscv to tune 'foo' but now we will need to tell it to tune 'pipeline.foo' as pipeline is now wrapped in another layer.  

I don't think that this is broken per say - but I do anticipate it being confusing, and thus worth pointing out somewhere.  Documentation would be fine.

sktime/base/_meta.py
miraep8: 👍

sktime/forecasting/compose/_pipeline.py
fkiraly: fixed

sktime/forecasting/compose/_pipeline.py
fkiraly: fixed

sktime/forecasting/compose/_pipeline.py
fkiraly: hm, what test specifically would you like to add? A grid search where we tune `pipeline__foo`?

sktime/forecasting/compose/_pipeline.py
```diff
@@ -1448,3 +1450,192 @@ def get_test_params(cls, parameter_set="default"):
         params = {"forecaster_X": fx, "forecaster_y": fy}
 
         return params
+
+
+class Permute(_DelegatedForecaster, _HeterogenousMetaEstimator):
+    """Permutation compositor for permuting pipeline steps.
+
+    The compositor can be used to permute the sequence of any meta-estimator,
+    including ForecastingPipeline and TransformedTargetForecaster.
```
aiwalter, line 1459:
```suggestion
    including ForecastingPipeline, TransformedTargetForecaster and
    TransformerPipeline
```

sktime/forecasting/compose/_pipeline.py
aiwalter: I think there is now a example in doctring section `Examples`, so would be fine for me. This is also executed when running tests.

aiwalter (approved):
Left minor comments, not a blocker but nice to have. Another nice to have

Add a `See Also` seciton in docstring that links to other pipeline classes and especially link from other pipeline classes `See Also` section to this class. And also from and to `OptionalPassthrough`. Can we add `Permute` also to docs please?

sktime/forecasting/compose/_pipeline.py
```diff
@@ -1448,3 +1450,192 @@ def get_test_params(cls, parameter_set="default"):
...
+        "ignores-exogeneous-X": False,
+        "requires-fh-in-fit": False,
+        "handles-missing-data": True,
+        "capability:pred_int": True,
+        "X-y-must-have-same-index": False,
+    }
+
+    _delegate_name = "estimator_"
+
+    def __init__(self, estimator, permutation, steps_arg="steps"):
```
aiwalter, line 1541:
```suggestion
    def __init__(self, estimator, permutation=None, steps_arg="steps"):
```
can we default this to None and raise a warning that `steps` is not permuted when `None`?
Having `None` helps when using it with grid search that its not necessary to instantiate `Permute` with any permutation as the permutation can then just be given in `param_grid`.

sktime/forecasting/compose/_pipeline.py
fkiraly: added the check

aiwalter (dismissed):

aiwalter (approved):
ok looks good, nice addition. I will add a more advanced doc string example in a separate PR.

PR #3690, opened 2022-11-01
[ENH] ExpandingWindowSplittter fix for initial_window=0 and depreciating "start_with_window" 

### LINKED ISSUE #2678: [ENH] removing redundant parameter `start_with_window` in `ExpandingWindowSplitter`

chillerobscuro:
#### Reference Issues/PRs
Fixes #2678 - second attempt after PR 3685 got messy.

`ExpandingWindowSplitter` now has consistent behavior when `start_with_window=False` or `initial_window=0`. In the case of the former, a warning is raised that `start_with_window` will be depreciated in version 0.15.0.

Also includes a test case for when `initial_window=0` with `ExpandingWindowSplitter`. 

Unlike the last PR, the scope does not affect behavior of `SlidingWindowSplitter`. 

I imported `warn` but no real deps.

### DIFF
--- a/sktime/forecasting/model_selection/_split.py
+++ b/sktime/forecasting/model_selection/_split.py
@@ -13,6 +13,7 @@ __all__ = [
 __author__ = ["mloning", "kkoralturk", "khrapovs"]
 
 from typing import Iterator, Optional, Tuple, Union
+from warnings import warn
 
 import numpy as np
 import pandas as pd
@@ -986,7 +987,7 @@ class BaseWindowSplitter(BaseSplitter):
         # length.
         if hasattr(self, "start_with_window") and self.start_with_window:
 
-            if self._initial_window is not None:
+            if self._initial_window not in [None, 0]:
 
                 if is_timedelta_or_date_offset(x=self._initial_window):
                     start = y.get_loc(
@@ -1199,7 +1200,7 @@ class ExpandingWindowSplitter(BaseWindowSplitter):
         fh: FORECASTING_HORIZON_TYPES = DEFAULT_FH,
         initial_window: ACCEPTED_WINDOW_LENGTH_TYPES = DEFAULT_WINDOW_LENGTH,
         step_length: NON_FLOAT_WINDOW_LENGTH_TYPES = DEFAULT_STEP_LENGTH,
-        start_with_window: bool = True,
+        start_with_window: bool = True,  # TODO: remove 0.15.0
     ) -> None:
         # Note that we pass the initial window as the window_length below. This
         # allows us to use the common logic from the parent class, while at the same
@@ -1209,8 +1210,14 @@ class ExpandingWindowSplitter(BaseWindowSplitter):
             window_length=initial_window,
             initial_window=None,
             step_length=step_length,
-            start_with_window=start_with_window,
+            start_with_window=start_with_window,  # TODO: remove 0.15.0
         )
+
+        if not start_with_window:  # TODO: remove 0.15.0
+            warn(
+                '"start_with_window" will be depreciated in 0.15.0, '
+                "use initial_window=0 instead"
+            )
 
         # initial_window needs to be written to self for sklearn compatibility
         self.initial_window = initial_window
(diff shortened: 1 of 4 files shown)


khrapovs (changes requested):
Out of curiosity, will you proceed doing the same trick with `SlidingWindowSplitter`?

chillerobscuro:
> Out of curiosity, will you proceed doing the same trick with `SlidingWindowSplitter`?

I'll take a look at that in its own PR. There seemed to be a bit more complexity to match the behavior of the "expected  windows" checks, but it will likely be much simpler when isolated from this work.

sktime/forecasting/model_selection/_split.py
```diff
@@ -13,6 +13,7 @@
 __author__ = ["mloning", "kkoralturk", "khrapovs"]
```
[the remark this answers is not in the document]
khrapovs: @chillerobscuro Please, do not mark conversations as resolved. It is a "privilege" of a reviewer to decide whether a suggestion was really addressed. Just helps to track the changes. Thanks!

sktime/forecasting/model_selection/tests/test_split.py
```diff
@@ -325,6 +325,28 @@ def test_sliding_window_splitter_start_with_empty_window(
             )
 
 
+@pytest.mark.parametrize("y", TEST_YS)
+@pytest.mark.parametrize("fh", [*TEST_FHS, *TEST_FHS_TIMEDELTA])
+@pytest.mark.parametrize("step_length", TEST_STEP_LENGTHS)
+def test_expanding_window_splitter_start_with_initial_window_zero(y, fh, step_length):
+    """Test ExpandingWindowSplitter."""
+    if _inputs_are_supported([fh, step_length, 0]):
```
khrapovs, line 333: please, create a variable `initial_window = 0` above `if` statement and use it instead of zeros below

khrapovs (changes requested):

khrapovs:
> > Out of curiosity, will you proceed doing the same trick with `SlidingWindowSplitter`?
> 
> I'll take a look at that in its own PR. There seemed to be a bit more complexity to match the behavior of the "expected windows" checks, but it will likely be much simpler when isolated from this work.

Kudos to that! Much easier to review.

sktime/forecasting/model_selection/tests/test_split.py
chillerobscuro: done

sktime/forecasting/model_selection/tests/test_split.py
```diff
@@ -325,6 +325,29 @@ def test_sliding_window_splitter_start_with_empty_window(
...
+        )
+        train_windows, test_windows, _, n_splits = _check_cv(
+            cv, y, allow_empty_window=True
+        )
+
+        assert np.vstack(test_windows).shape == (n_splits, len(check_fh(fh)))
+    else:
+        match = "Unsupported combination of types"
+        with pytest.raises(TypeError, match=match):
+            ExpandingWindowSplitter(fh=fh, initial_window=0, step_length=step_length)
```
khrapovs, line 348:
```suggestion
            ExpandingWindowSplitter(fh=fh, initial_window=initial_window, step_length=step_length)
```

khrapovs (changes requested):

sktime/forecasting/model_selection/tests/test_split.py
chillerobscuro: Good catch. Am I doing the right thing by "re-requesting review" each time I complete the changes?

Also, this one auto resolved after I accepted changes, it wasn't me!

khrapovs: If you commit suggestion "as is", it is indeed auto resolved, as expected

re-requesting review is not mandatory. at least in my case, I get email notifications on the PR that I participate in

khrapovs (approved):

PR #3697, opened 2022-11-03
[GOV] vote for removal from CC

fkiraly:
This PR factors out the rules on removal from CC proposed in #3675, in a version similar to that proposed by @aiwalter and @GuzalBulatova, i.e., requiring votes by core devs.

### DIFF
--- a/docs/source/get_involved/governance.rst
+++ b/docs/source/get_involved/governance.rst
@@ -324,7 +324,10 @@ expected to resign voluntarily.
 Tenure also ends automatically when a CC member's tenure as core developer ends,
 e.g., through resignation or inactivity.
 
-CC members can be removed from the CC by an 80% majority of CC members.
+CC members can be removed from the CC by an 80% majority of all core developers,
+i.e., non-participation abstention counts as against.
+This vote requires a 5 working day (excl week-ends) discussion period, followed
+by vote which is open for 5 working days (excl week-ends)
 
 Communications
 ^^^^^^^^^^^^^^
--- a/docs/source/get_involved/governance.rst
+++ b/docs/source/get_involved/governance.rst
@@ -326,7 +326,8 @@ e.g., through resignation or inactivity.
 
 CC members can be removed from the CC by an 80% majority of all core developers,
 i.e., non-participation abstention counts as against.
-This vote requires a 5 working day (excl week-ends) discussion period, followed
+This vote is anonymous and takes place on the private core developer channel.
+It requires a 5 working day (excl week-ends) discussion period, followed
 by vote which is open for 5 working days (excl week-ends)
 
 Communications
--- a/docs/source/get_involved/governance.rst
+++ b/docs/source/get_involved/governance.rst
@@ -324,6 +324,8 @@ expected to resign voluntarily.
 Tenure also ends automatically when a CC member's tenure as core developer ends,
 e.g., through resignation or inactivity.
 
+CC members can be removed from the CC by an 80% majority of CC members.
+
 Communications
 ^^^^^^^^^^^^^^
 


fkiraly:
@GuzalBulatova suggested to lower this to 2/3 in the other PR

RNKuhns:
@fkiraly what do we mean by non-participation or abstention counts as against. 

Do we mean a vote against the CC member being removed or a vote against the CC member being a CC member? I think we want that to be a bit clearer if that parts remains as part of the rule.

I'm also not sure I agree on the non-participation or abstention counting in either direction. If there are core developers who haven't been active for some time (but are still core devs b/c of activity in past year), it wouldn't seem that they should have an impact on the outcome at the margins (e.g., someone still being or someone no longer being a CC member based on inactivity).

Instead, could the process require a participation hurdle? Something like: At least a certain percentage (75%? 80%?) of core developers must participate in vote for it to count? If that threshold is reached then 2/3rd majority of those voting is required for removal. If 75% threshold is not reached then there is no removal.

We may also want to clarify if another removal vote can be called right away if a removal vote fails or we want a cooling off period (e.g., can't have another removal vote for that person for X weeks). I don't have strong feelings either way, but might be worth considering that the situation could arise.

docs/source/get_involved/governance.rst
```diff
@@ -324,6 +324,12 @@ expected to resign voluntarily.
 Tenure also ends automatically when a CC member's tenure as core developer ends,
 e.g., through resignation or inactivity.
 
+CC members can be removed from the CC by an 80% majority of all core developers,
+i.e., non-participation abstention counts as against.
```
aiwalter, line 328:
```suggestion
```

docs/source/get_involved/governance.rst
```diff
@@ -324,6 +324,12 @@ expected to resign voluntarily.
 Tenure also ends automatically when a CC member's tenure as core developer ends,
 e.g., through resignation or inactivity.
 
+CC members can be removed from the CC by an 80% majority of all core developers,
```
aiwalter, line 327:
```suggestion
CC members can be removed from the CC by a majority of all voting core developers.
```

docs/source/get_involved/governance.rst
```diff
@@ -324,6 +324,12 @@ expected to resign voluntarily.
...
+i.e., non-participation abstention counts as against.
+This vote is anonymous and takes place on the private core developer channel.
+It requires a 5 working day (excl week-ends) discussion period, followed
```
aiwalter, line 330:
```suggestion
```

aiwalter (changes requested):

docs/source/get_involved/governance.rst
fkiraly: I think this is important, people need to know this is coming. We have this for appointments, so we should also do this for revocations.

fkiraly:
> @fkiraly what do we mean by non-participation or abstention counts as against.

I mean, in favour of *remaining* a CC member, and as a vote *against* removing the CC member.

The rationale is to avoid the "holiday attack" mentioned above, where a small fraction of people band together and call the vote when others are absent.

This strategy can be prevented by (a) putting the quorum at a fixed fraction of the total (rather than of participants), or (b) requiring at least a certain fraction of voters participating for the vote to count.

> I'm also not sure I agree on the non-participation or abstention counting in either direction. If there are core developers who haven't been active for some time (but are still core devs b/c of activity in past year), it wouldn't seem that they should have an impact on the outcome at the margins (e.g., someone still being or someone no longer being a CC member based on inactivity).

> Instead, could the process require a participation hurdle? Something like: At least a certain percentage (75%? 80%?) of core developers must participate in vote for it to count? If that threshold is reached then 2/3rd majority of those voting is required for removal. If 75% threshold is not reached then there is no removal.

I'm not too precious about specific details on how to prevent the issue.

For instance, I would be happy with a 3/4 participation threshold and a 2/3 removal threshold (of votes cast). That indirectly also ensures that in case of a successful vote, there are at least `3/4 * 2/3 = 1/2` of all potential votes in favour of removal.

> We may also want to clarify if another removal vote can be called right away if a removal vote fails or we want a cooling off period (e.g., can't have another removal vote for that person for X weeks).

Hm, good point. Cooldown seems sensible. In the UK I think this is 12 months after a vote of no confidence, from party statutes (not parliamentary conventions).

FYI @aiwalter, votes of no confidence also requires a time for debate in the UK.

https://commonslibrary.parliament.uk/research-briefings/sn02873/
https://commonslibrary.parliament.uk/votes-of-no-confidence/

fkiraly:
@RNKuhns, @GuzalBulatova, @aiwalter, I think we should combine this with a change in appointment rules to CC.

I would suggest: if there are strictly less than 3 members of CC, the condition that 2/3 of CC members must approve does not apply.

Why: this would ensure that gradually the core developer gremium can replace a CC that does not have the support of at least 1/3 of the core developers with 2/3 majority representatives.

I would also prefer to introduce the cooldown idea by @RNKuhns, e.g., there can be at most one removal every 6 months, an at most one attempt directed at a specific CC member to avoid breaks in handover and "noise votes".

That would, however, create the problem of having to define a proposal and priority process. E.g., can any one, any time propose removal? This could be done by minorty CC supporters to cause a cooldown.

To prevent this, I think there needs to be a discussion period, e.g., alternative removals can be proposed, and not removing anyone is also an option, relative majority determines the removal proposal going forward?

Sounds a bit complicated, but I can't think of a good rule that prevents the tactic "activating the cooldown to support a CC which has only minority support".

fkiraly:
Other option: only removal is voted on first, but not the CC member who would be removed. That would mean that there is not too much advantage in causing a vote if you are member of a minortiy supporting the CC.

GuzalBulatova (commented):
Agree with adding this option. Counter-proposal on details: 
1 week discussion period (as proposed) and 5 working days (excl weekends and public holidays) voting period. 2/3 majority of all cast votes. Wdyt @aiwalter , @fkiraly ?

aiwalter:
I have the opinion that we do not need a separate section for this. A vote for removal from CC, CoC or any other role that we might have in future will just be a [regular core dev vote ](https://github.com/sktime/sktime/blob/main/docs/source/get_involved/governance.rst#stage-2-voting)

fkiraly:
closing this, as we will start discussing the topic again later.

PR #3701, opened 2022-11-04
[BUG] `ConformalIntervals` doesn't work as expected when `alpha = 0.5`

### LINKED ISSUE #3693: [BUG] `ConformalIntervals` `conformal_bonferroni` method doesn't work as expected when alpha = 0.5
**Describe the bug**
When using `predict_quantiles` with `ConformalIntervals`, you would expect `alpha=0.5` to correspond to the output of `predict` (or very close to it). However, when the Bonferroni correction is applied, it doesn't at all and the error becomes large vs just using `predict`.

I think this is due to the way `_predict_quantiles` is converted to `_predict_interval` and how zero coverage is interpreted in the logic.

**To Reproduce**
Toy example using the logic from `conformal_bonferroni` correction [code](https://github.com/sktime/sktime/blob/ad2619ba76675c9b8a9bc9174e5e442ca4730284/sktime/forecasting/conformal.py#L243):

```python
# Set alpha to 0.5 and a toy forecasting horizon and abs_residuals
alpha = 0.5
fh = [1,2,3,4,5]
abs_resids = [0.1, 0.2, 0.3, 0.25, 0.1, 0.0001]

# alpha is converted to coverage in `_predict_quantiles` base forecaster 
coverage = np.abs(1 - 2 * alpha)
# coverage = 0 

coverage2 = np.repeat(coverage, 2)
# coverage2 = [0, 0]

alphas = 1 - coverage2
# alphas = [1, 1]

quantiles = 1 - alphas / len(fh)
# quantiles = [0.8, 0.8]

pred_int_row = np.quantile(abs_resids, quantiles)
# pred_int_row = [0.3, 0.3]

I think even in the `conformal` logic this is (less) incorrect because you take the 0th quantile which is the smallest residual and I'm not sure if this is what you want to do. 

In the toy example:
```
quantiles = coverage2
pred_int_row = np.quantile(abs_resids, quantiles)
# pred_int_row = [0.0001, 0.0001]
```

```

**Expected behavior**
`_predict_interval` with alpha = 0.5 and therefore coverage = 0 should return `y_pred + 0`

**Versions**
0.13.4
System:
    python: 3.8.13 (default, May  8 2022, 17:48:02)  [Clang 13.1.6 (clang-1316.0.21.2)]
executable: /Users/bethrice/Documents/Code/uk-ds-day-ahead-prices-forecast/.venv/bin/python
   machine: macOS-12.1-arm64-i386-64bit

Python dependencies:
          pip: 22.1.2
   setuptools: 63.1.0
      sklearn: 1.1.2
       sktime: 0.13.4
  statsmodels: 0.13.2
        numpy: 1.22.0
        scipy: 1.8.1
       pandas: 1.4.1
   matplotlib: 3.6.0
       joblib: 1.2.0
        numba: 0.56.2
     pmdarima: None
      tsfresh: None

</details>

bethrice44:
#### Reference Issues/PRs

Fixes #3693 

This checks if the coverage = 0, which corresponds to alpha = 0.5 (`coverage` = a list of one alpha value in the translation from `_predict_quantiles` to `_predict_interval`). If this is the case it sets `pred_int_row` to `[0.0, 0.0]` which means that `alpha =0.5` corresponds to the output of `_predict()`. 

No

Is this mathematically correct (should  median = output of `._predict()`?

##### For all contributions

##### For new estimators


fkiraly (changes requested):
Looks a bit like a hack...

have you thought about whether there is a more principled way to fix it?

Because you would expect the prediction still be wrong if coverage is 0.01, say.

Might there simply be a problem in the formula?

bethrice44:
> Looks a bit like a hack...
> 
> have you thought about whether there is a more principled way to fix it?
> 
> Because you would expect the prediction still be wrong if coverage is 0.01, say.
> 
> Might there simply be a problem in the formula?

Is the translation from alpha in `_predict_quantiles` -> coverage in `_predict_interval` correct? Should alpha = 0.5 mean coverage is 0? 

```
coverage = np.abs(1 - 2 * alpha)
```

bethrice44:
In `conformal_beronni` there is this conversion:

```
alphas = 1 - coverage2
quantiles = 1 - alphas / len(fh)
```

so maybe this is the error. `coverage` -> `alpha` is not the inverse of `alpha` -> `coverage`

In that case, the fix would be 

```
alphas = (1-coverage2)/2

```
assuming symmetric intervals.

Does that make sense?

bethrice44:
Revisiting the paper: I can't see where we use `m` (calibration dataset size) in the Bonferroni correction in the code. Should we be multiplying the quantiles code by `len(y) - initial_window`? And if not, why not?
Screenshot 2022-11-04 at 14 00 20

fkiraly:
> Should we be multiplying the quantiles code by len(y) - initial_window + 1?

In my opinion, no.

Bonferroni is only the second term.

The first term, multiplying by the sample size `len(y) - initial_window + 1`, is taken care of by the `np.quantile` (in lines 239, 242, 246, and 249). This is because ` np.quantile` takes a fraction rather than the integer rank, whereas in the paper it is expressed in terms of integer rank.

fkiraly:
maybe there's something going on with the "one minus" or the factor two?

bethrice44:
> maybe there's something going on with the "one minus" or the factor two?

I think the correction makes sense when (say) coverage = 0.9 (i.e. alpha = 0.05 or 0.95). Then you get `alphas = [0.1, 0.1]` (slightly different definition without a factor of 2 in the paper) and the correction gives quantiles [0.02, 0.02] for `len(fh) = 5`, i.e. a higher bar to reach the actual quantile to counteract the multiple comparisons problem. 

However, it doesn't seem right when `alpha = 0.5` => `coverage = 0` => `alphas = [1, 1]` => `quantiles = [0.8, 0.8]` (for `len(fh) = 5`, but I can't quite see or articulate _why_ it seems wrong and is different. And I also wonder if it is wrong just for the special symmetric case of `alpha = 0.5` or if it's more wrong for values close to `0.5`. 

Does that make sense, and do you also feel it is wrong (at least) for the case `alpha = 0.5`?

fkiraly:
@bethrice44, I'm slightly confused. In the second example

> However, it doesn't seem right when `alpha = 0.5` => `coverage = 0` => `alphas = [1, 1]` => `quantiles = [0.8, 0.8]` (for `len(fh) = 5`), but I can't quite see or articulate _why_ it seems wrong and is different. And I also wonder if it is wrong just for the special symmetric case of `alpha = 0.5` or if it's more wrong for values close to `0.5`. 

Shouldn't it be `quantiles = [0.2, 0.2]`?

bethrice44:
> @bethrice44, I'm slightly confused. In the second example
> 
> > However, it doesn't seem right when `alpha = 0.5` => `coverage = 0` => `alphas = [1, 1]` => `quantiles = [0.8, 0.8]` (for `len(fh) = 5`), but I can't quite see or articulate _why_ it seems wrong and is different. And I also wonder if it is wrong just for the special symmetric case of `alpha = 0.5` or if it's more wrong for values close to `0.5`.
> 
> Shouldn't it be `quantiles = [0.2, 0.2]`?

The formula is currently 

```
quantiles = 1 - alphas / len(fh)
quantiles = 1 - [1, 1]/5
quantiles = 1 - [0.2, 0.2]
quantiles = [0.8, 0.8]

```

fkiraly:
more confused - isn't the first example wrong then?

I suspect there's something going on with "one minus"

fkiraly:
I think we need to answer two questions:

(a) is it 1:1 congruent with the algorithm in the paper (applied to sliding windows as if they were independent)?
(b) is it sensible in the coverage = 0 case?

If the answer is (a) yes, (b) no, I think we should leave the method as is, since it explicitly refers to the paper. We can add an alternative, of course. but that should also be reasonable in coverage = 0.01 etc cases.

bethrice44:
> I think we need to answer two questions:
> 
> (a) is it 1:1 congruent with the algorithm in the paper (applied to sliding windows as if they were independent)? (b) is it sensible in the coverage = 0 case?
> 
> If the answer is (a) yes, (b) no, I think we should leave the method as is, since it explicitly refers to the paper. We can add an alternative, of course. but that should also be reasonable in coverage = 0.01 etc cases.
> 
> The `"empirical"` strategy (not in the paper but the semi-naive empirical method) should already behave sensibly, no?

On the first question, I came across [this implementation of the method](https://github.com/kamilest/conformal-rnn) (you may have seen).

The relevant lines for what the quantiles should be are [here](https://github.com/kamilest/conformal-rnn/blob/845b436cc85480457d277ed00f80e31487eb4cfa/models/cfrnn.py#L285). 

```
        q = min((n_calibration + 1.0) * (1 - self.alpha) / n_calibration, 1)
        corrected_q = min((n_calibration + 1.0) * (1 - self.alpha / self.horizon) / n_calibration, 1)

        self.critical_calibration_scores = get_critical_scores(calibration_scores=self.calibration_scores, q=q)

        # Bonferroni corrected calibration scores.
        # [horizon, output_size]
        self.corrected_critical_calibration_scores = get_critical_scores(
            calibration_scores=self.calibration_scores, q=corrected_q
        )

```

The function `get_critical_scores()` uses `torch.quantile` which expects an input tensor and a float or tensor in the range [0, 1], corresponding to the quantiles, which is taken from the index of the sorted input (as you said above).

> To compute the quantile, we map q in [0, 1] to the range of indices [0, n] to find the location of the quantile in the sorted input. 

So vs this implementation for, we don't have the factor of `(n_calibration + 1.0)/n_calibration` but also take the quantiles with linear interpolation of the data rather than via indices of the data.

fkiraly:
> So vs this implementation, we don't have the factor of (n_calibration + 1.0)/n_calibration.

That's almost one, and I don't think it appears in the original paper?

bethrice44:
Yes I don't think that is there. I think the paper implementation is correct as shown but I don't think it's correct in practise for the case `alpha = 0.5`. I might raise an issue on that other repo to query it and maybe link to this issue for context.

fkiraly:
> I might raise an issue on that other repo to query it and maybe link to this issue for context.

Hm, ok - let me know what they say.

I think that the algorithm simply produces unintuitive forecasts for large values of `alpha` due to the significance correction, but it is the actual algorithm as described (therefore should remain as is, since this is an direct implementation).

We can always add calibration methods or similar, as wrappers?

PR #3684, opened 2022-10-31
[GOV] Lazy consensus on GOV PRs

aiwalter:
Clarification on lazy consensus approval for GOV PRs.

PR #3711, opened 2022-11-05
[GOV] Lazy consensus on GOV PRs - FK coutnerproposal

fkiraly:
My counterproposal for https://github.com/sktime/sktime/pull/3684, a version I would be happy with.

(wip draft)


aiwalter (changes requested):
I disagree with this as it affects current discussions around CoC, so we need to first finish them before we should have changes to CoC goernance rules

fkiraly:
> I disagree with this as it affects current discussions around CoC, so we need to first finish them before we should have changes to CoC goernance rules

How does it affect current discussions around CoC? Don't see that.
If an issue is contentious, you can always communicate a rejection.

I don't want any rules that allow a motion to be deadlocked. It should be accepted or rejected, we shouldn't allow it to be "soft rejected" by stalling and maneuvering.

Issue #3715, opened 2022-11-06
[MNT] PyPi publish action for 0.14 failed


jorenham:
Summary by @fkiraly:
The attempted 0.14.0 release has failed at pypi upload of successfully built wheels.

Help on diagnosing and solving this is appreciated.

Original post:
---
Just to let you know: 0.14 has not been published on PyPI yet:

https://github.com/sktime/sktime/actions/runs/3402037795

fkiraly:
You are right 😱 

The error happens right after it attempts to upload to `pypi`, this is it:

```
 ERROR    HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/        
         '=?utf-8?b?RnJhbnogS2lyw6FseSA8...YWxsIDxhamJAdWVhLmFjLnVrPg==?=' is an
         invalid value for Author-email. Error: Use a valid email address See   
         https://packaging.python.org/specifications/core-metadata for more     
         information.   
```

fkiraly:
any ideas what this is, @jorenham? Haven't seen it before, and Professor Google doesn't seem to help...

jorenham:
@fkiraly I'm guessing something with the email address in the `setup.py` or `pyproject.toml`,  but I can't seem to find anything at first glance

jorenham:
@fkiraly It might be easiest to merge the `author` and `author_email` in the `pyproject.toml` and `setup.py`, formatted like a `RFC-822 From: header`, as specified in: https://packaging.python.org/en/latest/specifications/core-metadata/#author-email

fkiraly:
let's keep this open until we are sure it works...

fkiraly:
> @fkiraly I'm guessing something with the email address in the setup.py or pyproject.toml, but I can't seem to find anything at first glance

I had the same guess, though I did sth slightly different. If that does not work, I will try your suggestion next, @jorenham.

Other guess: `sktime` still seems to run on the pypi account on one of the founders, @mloning, who isn't very active anymore.
Could it be that he forgot about this and changed his password, so it no longer agrees with the secret?

Although the wouldn't fit the message which spews something about UTF8.

fkiraly:
my attempt at fix: https://github.com/sktime/sktime/pull/3718

fkiraly:
actually, @jorenham, if you have 10min, it would be appreciated if you could make a draft PR that implements your suggested change? I don't know what exactly I would change.

[...]

PR #3718, opened 2022-11-07
[MNT] addressing failed v0.14.0 pypi upload

fkiraly:
This PR attempts to address the failed pypi upload for v0.14.0, fixes https://github.com/sktime/sktime/issues/3715.

My suspicion is that there is an interaction between `setup.py` and `pyproject.toml` in the email address fields.

* this PR removes personal email addresses from `pyproject.toml`
* in `setup.py`, replaces names of one author with "project developers", and emails with a hardcoded string `"info@sktime.org"`

Also adds instructions to the release process docs how to re-release in a case where the wheel upload fails.


fkiraly:
Emergency merge, let's see whether this fixes it.

PR #3719, opened 2022-11-07
[MNT] PEP 518 compliance

jorenham:
Addresses #3715

@fkiraly

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,15 +3,15 @@ name = "sktime"
 version = "0.14.0"
 description = "A unified framework for machine learning with time series"
 authors = [
-    {name = "Franz Király"},
-    {name = "Tony Bagnall"},
-    {name = "Markus Löning"},
-    {name = "Martin Walter"},
+    "Franz Király <f.kiraly@ucl.ac.uk>",
+    "Tony Bagnall <ajb@uea.ac.uk>",
+    "Markus Löning",
+    "Martin Walter",
 ]
 maintainers = [
-    {name = "Franz Király"},
-    {name = "Guzal Bulatova"},
-    {name = "Martin Walter"},
+    "Franz Király <f.kiraly@ucl.ac.uk>",
+    "Guzal Bulatova",
+    "Martin Walter",
 ]
 readme = "README.md"
 keywords = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,7 @@ authors = [
     "Franz Király <f.kiraly@ucl.ac.uk>",
     "Tony Bagnall <ajb@uea.ac.uk>",
     "Markus Löning",
-    "Martin Walter <mwalter@sktime.org> ",
+    "Martin Walter <mwalter@sktime.org>",
 ]
 maintainers = [
     "Franz Király <f.kiraly@ucl.ac.uk>",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,12 +6,12 @@ authors = [
     "Franz Király <f.kiraly@ucl.ac.uk>",
     "Tony Bagnall <ajb@uea.ac.uk>",
     "Markus Löning",
-    "Martin Walter",
+    "Martin Walter <mwalter@sktime.org> ",
 ]
 maintainers = [
     "Franz Király <f.kiraly@ucl.ac.uk>",
     "Guzal Bulatova",
-    "Martin Walter",
+    "Martin Walter <mwalter@sktime.org>",
 ]
 readme = "README.md"
 keywords = [
--- a/setup.py
+++ b/setup.py
@@ -29,8 +29,8 @@ def setup_package():
     """Set up package."""
     projectname = pyproject["project"]["name"]
     setup(
-        author_email="info@sktime.org",
-        author=f"{projectname} developers",
+        author=pyproject["project"]["authors"][0],
+        maintainer=pyproject["project"]["maintainers"][0],
         classifiers=pyproject["project"]["classifiers"],
         description=pyproject["project"]["description"],
         download_url=pyproject["project"]["urls"]["download"],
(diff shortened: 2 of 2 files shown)


fkiraly:
oh, you mean it should be a string rather than a dict?

The page wasn't very clear about which types are expected. Would be really great if there were an example file.

fkiraly:
actually, looks like it's wrong... CI is failing with this:

```
      OFFENDING RULE: 'type'
      
      DEFINITION:
          {
              "$id": "#/definitions/author",
              "title": "Author or Maintainer",
              "$comment": "https://www.python.org/dev/peps/pep-0621/#authors-maintainers",
              "type": "object",
              "properties": {
                  "name": {
                      "type": "string",
                      "$$description": [
                          "MUST be a valid email name, i.e. whatever can be put as a name, before an",
                          "email, in :rfc:`822`."
                      ]
                  },
                  "email": {
                      "type": "string",
                      "format": "idn-email",
                      "description": "MUST be a valid email address"
                  }
              }
          }
```

jorenham:
@fkiraly I believe it can be both in pyproject.toml, and from what I can tell from PEP 518, it should be valid for setup.py as well.

fkiraly:
well, it's failing the CI. Here's from PEP 621:

```
Format: Array of inline tables with string keys and values

These fields accept an array of tables with 2 keys: name and email. Both values must be strings. The name value MUST be a valid email name (i.e. whatever can be put as a name, before an email, in [RFC 822](https://datatracker.ietf.org/doc/html/rfc822.html)) and not contain commas. The email value MUST be a valid email address. Both keys are optional.
```

What's an array of inline tables? That's not even a python type. Who writes tech docs like this.

fkiraly:
😱

jorenham:
@fkiraly  Yea, but I don't see why though, since it works fine with python-poetry in my personal projects
https://python-poetry.org/docs/pyproject/#authors

fkiraly:
> @fkiraly Yea, but I don't see why though, since it works fine with python-poetry in my personal projects

Well, I don't see it either, it's highly confusing. Let's be sad together.

jorenham:
@fkiraly It's pretty annoying that PEP 621 and PEP 518 seem to contradict each other on this point, see  https://peps.python.org/pep-0621/#authors-maintainers

fkiraly:
> It's pretty annoying that PEP 621 and PEP 518 seem to contradict each other on this point

Is there some other PEP that defines override laws, maybe it's something like [genetic recombination](https://en.wikipedia.org/wiki/Genetic_recombination) and you have to guess what the actual rule is by mixing the right parts from both.

jorenham:
@fkiraly Yea could very well be. Or perhaps it's dependent on the position of the moon with respect to the earth 🤷🏻

fkiraly:
You think we should ask an [`astropy`](https://www.astropy.org/) developer?

fkiraly:
(need to go to sleep now, will check the run in half a day or so)

jorenham:
@fkiraly The only other possibility I could think of, is that it has to do with the authors that have no email address

fkiraly:
any new ideas on the the issues please on #3715 to have them in one place

aiwalter (commented):
Left comment

pyproject.toml
```diff
@@ -3,15 +3,15 @@ name = "sktime"
...
+    "Markus Löning",
+    "Martin Walter",
 ]
 maintainers = [
-    {name = "Franz Király"},
-    {name = "Guzal Bulatova"},
-    {name = "Martin Walter"},
+    "Franz Király <f.kiraly@ucl.ac.uk>",
+    "Guzal Bulatova",
+    "Martin Walter",
```
aiwalter, line 14:
```suggestion
    "Martin Walter <mwalter@sktime.org>",
```

aiwalter (commented):
..

pyproject.toml
```diff
@@ -3,15 +3,15 @@ name = "sktime"
...
 description = "A unified framework for machine learning with time series"
 authors = [
-    {name = "Franz Király"},
-    {name = "Tony Bagnall"},
-    {name = "Markus Löning"},
-    {name = "Martin Walter"},
+    "Franz Király <f.kiraly@ucl.ac.uk>",
+    "Tony Bagnall <ajb@uea.ac.uk>",
+    "Markus Löning",
+    "Martin Walter",
```
aiwalter, line 9:
```suggestion
    "Martin Walter <mwalter@sktime.org> ",
```

aiwalter (commented):
.

pyproject.toml
```diff
@@ -3,15 +3,15 @@ name = "sktime"
...
 description = "A unified framework for machine learning with time series"
 authors = [
-    {name = "Franz Király"},
-    {name = "Tony Bagnall"},
-    {name = "Markus Löning"},
-    {name = "Martin Walter"},
+    "Franz Király <f.kiraly@ucl.ac.uk>",
+    "Tony Bagnall <ajb@uea.ac.uk>",
+    "Markus Löning",
+    "Martin Walter <mwalter@sktime.org> ",
```
aiwalter, line 9:
```suggestion
    "Martin Walter <mwalter@sktime.org>",
```

aiwalter (commented):
Sry there is a space character

wolph:
Isn't the entire `pyproject[...]` stuff in `setup.py` superfluous? When building the package it should automatically interpolate those values from the `pyproject.toml` already.

Looking at the setup.py, I think it might even be possible to fully port it to the `pyproject.toml` so you can delete the `setup.py` entirely.

fkiraly:
> Looking at the setup.py, I think it might even be possible to fully port it to the `pyproject.toml` so you can delete the `setup.py` entirely.

can we try that? PR a appreciated!

wolph:
> > Looking at the setup.py, I think it might even be possible to fully port it to the `pyproject.toml` so you can delete the `setup.py` entirely.
> 
> can we try that? PR a appreciated!

As you wish :)
https://github.com/sktime/sktime/pull/3723

fkiraly:
I suppose this is contained in https://github.com/sktime/sktime/pull/3723 and can be closed? You are credited on the other PR @jorenham so this will go in the release notes with your name too.

Feel free to add yourself to the all-contributors file, and thanks!

PR #3723, opened 2022-11-07
[MNT] Converted `setup.py` to `pyproject.toml`. Depends on `setuptools>61.0.0`

### LINKED ISSUE #3715: [MNT] PyPi publish action for 0.14 failed

wolph:
#### Reference Issues/PRs
Fixes #3715, see also #3719.

It converts all `setup.py` parameters to `pyproject.toml`.

This should be supported from `setuptools` `61.0.0` and up.

It should be noted that this is officially still in beta, but it's been supported since march 2021 with no breaking changes as far as I can see. And that's after a quite a few release (we're at 65.5.1 now).

Not a new dependency but a new version requirement: `setuptools` `>61.0.0` 

Whether the package metadata is still identical. I've compared the metadata as far as I could using `wheel-inspect`

##### For all contributions

##### For new estimators


wolph:
Thanks to @jorenham for the initial work :)

pyproject.toml
```diff
@@ -3,15 +3,16 @@ name = "sktime"
...
+    {name = "Franz Király", email = "f.kiraly@ucl.ac.uk"},
+    {name = "Tony Bagnall", email = "ajb@uea.ac.uk"},
     {name = "Markus Löning"},
-    {name = "Martin Walter"},
+    {name = "Martin Walter", email = "mwalter@sktime.org"},
 ]
 maintainers = [
-    {name = "Franz Király"},
+    {name = "sktime developers", email = "info@sktime.org"},
+    {name = "Franz Király", email = "f.kiraly@ucl.ac.uk"},
```
fkiraly, line 13: I think we should be using the `sktime.org` emails. This is a bit of a relic.

The pattern is `fkiraly@sktime.org` etc, I think.
(@aiwalter is maintaining the email acct)

fkiraly:
FYI @aiwalter, @GuzalBulatova, @freddyaboulton, @lmmentel - this removes the entire `setup.py`. Would be amazing if it worked out of the box!

Any opinions?

wolph:
I should note that we can drop the `setup.cfg` as well in a similar matter, but there are also settings for other tools such as pytest and isort in that file that would need to be moved so I'm not a 100% certain that will work

RNKuhns:
@wolph and @fkiraly does this still allow installation in editable mode? 

I know for a while a minimal setup.py was still needed for that. But it looks like that is no longer True  based on this [StackOverflow](https://stackoverflow.com/questions/69711606/how-to-install-a-package-using-pip-in-editable-mode-with-pyproject-toml). But I saw mention of later versions of setuptools than we are requiring in this PR.


We should verify that the editable install instructions we give to devs in docs will work with all setuptools versions. Looks like version 64 is first to fully support PEP660.

lmmentel:
`setup.py` has been a legacy format for some time now so I'm happy for us to drop it (see [pypa docs](https://pip.pypa.io/en/stable/reference/build-system/setup-py/)).  I have not researched this extensively but adopting `pyproject.toml` allows usage of alternative build systems to setuptools for we don't necessarily need to stick with it if it does not meet our requirements. 


@wolph certainly most configurations can be moved over to appropriate sections in the `pyproject.toml` but I would go for a separate PR for this for better isolation.

fkiraly (approved):
Seems to work - we can change emails later.
Thanks a lot, @jorenham, @wolph, this simplifies things!

I would like to hear from @lmmentel though before someone merges.

lmmentel (approved):
Looks good, thanks 👍

fkiraly:
please feel free to add yourselves to allcontributors, @jorenham and @wolph, thanks!

PR #3724, opened 2022-11-07
[ENH] better error message on transform output check fail

fkiraly:
This PR improves the error message if the transform output check fails, by pointing to the estimator type and printing the output.

### DIFF
--- a/sktime/transformations/base.py
+++ b/sktime/transformations/base.py
@@ -1021,11 +1021,17 @@ class BaseTransformer(BaseEstimator):
             #   we cannot convert back to pd.Series, do pd.DataFrame instead then
             #   this happens only for Series, not Panel
             if X_input_scitype == "Series":
-                _, _, metadata = check_is_mtype(
+                valid, msg, metadata = check_is_mtype(
                     Xt,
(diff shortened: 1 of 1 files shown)


sktime/transformations/base.py
```diff
@@ -1021,11 +1021,17 @@ def _convert_output(self, X, metadata, inverse=False):
...
-                _, _, metadata = check_is_mtype(
+                valid, msg, metadata = check_is_mtype(
                     Xt,
                     ["pd.DataFrame", "pd.Series", "np.ndarray"],
                     return_metadata=True,
                 )
+                if not valid:
+                    raise TypeError(
+                        f"_transform output of {type(self)} does not comply "
+                        "with sktime mtype specifications. Returned error message:"
```
GuzalBulatova, line 1032: possible improvement: add pointer to the mtype specifications?

GuzalBulatova (commented):
Looks all good, added a comment

sktime/transformations/base.py
fkiraly: added

GuzalBulatova (approved):
Good to go imo

PR #3726, opened 2022-11-08
Annotation notebook

KatieBuc:
Creating an example notebook to explain the implemented algorithms in the annotation module! @fkiraly @lmmentel @Lovkush-A


KatieBuc:
I'm not entirely sure what to do about checking for dependencies - 

I'm running `%matplotlib widget` which required `ipympl` to be installed, do we `check_soft_dependencies` in the notebook?

KatieBuc:
Hmm yeah - it's failing on run-notebook-examples because `ModuleNotFoundError: No module named 'ipympl'`.

Do I just put a pip install at the beginning of the notebook or is there a smarter way?

GuzalBulatova:
Great job, @KatieBuc , an option is to get rid of the widget? Is that necessary?

MatthewMiddlehurst:
If you want to keep the dependency it would have to be added to `all_extras` in `pyproject.toml` I think. 

I don't have a strong opinion about whether it should be added as a dependency or not, but may be worth trying just to see if it clears the test.

miraep8:
Just wanted to say I think that this notebook will be a great contribution for the annotation docs!  So happy someone is putting it together! :)

lmmentel:
Nice! Regarding the dependency, there is a `binder` section in the `pyproject.toml` file with deps for the notebooks and you can add `ipympl` there if you want since it's only needed for notebooks.

fkiraly:
superseded by the current detection notebook

PR #3739, opened 2022-11-10
[DOC] Update documentation for Greedy Gaussian Segmentation

### LINKED ISSUE #3375: [DOC] Improve developer documentation for GreedyGaussianSegmentation

Greedy gaussian segmentation was introduced in #2744, however the documentation is insufficient especially when it comes to helping developers take over the maintenance or further development of the estimator.

The docstrings for all private and public methods should be expanded on to provide an easier onboarding to the algorithm and it's implementation and link to specific parts of the existing references such as published papers.

lmmentel:
#### Reference Issues/PRs

closes #3375 

see title

### DIFF
--- a/sktime/annotation/ggs.py
+++ b/sktime/annotation/ggs.py
@@ -65,7 +65,10 @@ class GGS:
         Maximum number of change points to find. The number of segments is thus k+1.
     lamb: : float, default=1.0
         Regularization parameter lambda (>= 0), which controls the amount of
-        (inverse) covariance regularization, see Eq (1) in [1]_.
+        (inverse) covariance regularization, see Eq (1) in [1]_. Regularization
+        is introduced to reduce issues for high-dimensional problems. Setting 
+        ``lamb`` to zero will ignore regularization, whereas large values of
+        lambda will favour simpler models.
     max_shuffles: int, default=250
         Maximum number of shuffles
     verbose: bool, default=False
@@ -381,7 +384,10 @@ class GreedyGaussianSegmentation(BaseEstimator):
         Maximum number of change points to find. The number of segments is thus k+1.
     lamb: : float, default=1.0
         Regularization parameter lambda (>= 0), which controls the amount of
-        (inverse) covariance regularization, see Eq (1) in [1]_.
+        (inverse) covariance regularization, see Eq (1) in [1]_. Regularization
+        is introduced to reduce issues for high-dimensional problems. Setting 
+        ``lamb`` to zero will ignore regularization, whereas large values of
+        lambda will favour simpler models.
     max_shuffles: int, default=250
         Maximum number of shuffles
     verbose: bool, default=False
--- a/sktime/annotation/ggs.py
+++ b/sktime/annotation/ggs.py
@@ -76,7 +76,7 @@ class GGS:
     lamb: : float, default=1.0
         Regularization parameter lambda (>= 0), which controls the amount of
         (inverse) covariance regularization, see Eq (1) in [1]_. Regularization
-        is introduced to reduce issues for high-dimensional problems. Setting 
+        is introduced to reduce issues for high-dimensional problems. Setting
         ``lamb`` to zero will ignore regularization, whereas large values of
         lambda will favour simpler models.
     max_shuffles: int, default=250
@@ -395,7 +395,7 @@ class GreedyGaussianSegmentation(BaseEstimator):
     lamb: : float, default=1.0
         Regularization parameter lambda (>= 0), which controls the amount of
         (inverse) covariance regularization, see Eq (1) in [1]_. Regularization
-        is introduced to reduce issues for high-dimensional problems. Setting 
+        is introduced to reduce issues for high-dimensional problems. Setting
         ``lamb`` to zero will ignore regularization, whereas large values of
         lambda will favour simpler models.
     max_shuffles: int, default=250
--- a/sktime/annotation/ggs.py
+++ b/sktime/annotation/ggs.py
@@ -2,6 +2,16 @@
 """
 Greedy Gaussian Segmentation (GGS).
 
+The method apprixmates solutions for the problem of breaking a
+multivariate time series into segments, where the data in each segment
+could be modeled as independent samples from a multivariate Gaussian
+distribution. It uses a dynamic programming search algorithm with
+a heuristic that allows finding approximate solution in linear time with
+respect to the data length and always yields locally optimal choice.
+
+Notes
+-----
+
 Based on the work from [1]_.
 
 - source code adapted based on: https://github.com/cvxgrp/GGS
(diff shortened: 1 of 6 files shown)


sktime/annotation/ggs.py
```diff
@@ -34,19 +43,46 @@ class GGS:
...
+    search used in other change point detection problems.
+
     Parameters
     ----------
     k_max: int, default=10
-        Maximum number of change points to find
+        Maximum number of change points to find. The number of segments is thus k+1.
     lamb: : float, default=1.0
-        Regularization parameter lambda
+        Regularization parameter lambda (>= 0), which controls the amount of
```
miraep8, line 67: Any chance to get a bit more detail here?  - I don't think you need the full equation but a quick sentence on what affect increasing/decreasing lamba has on the fit would be nice without needing to dive into the paper.

sktime/annotation/ggs.py
```diff
@@ -72,28 +108,27 @@ class GGS:
...
+    def initialize_intermediates(self) -> None:
+        """Initialize the state fo the estimator."""
         self._intermediate_change_points = []
         self._intermediate_ll = []
 
-    def log_likelihood(self, cov: float, nrows: int, ncols: int) -> float:
+    def log_likelihood(self, data: npt.ArrayLike) -> float:
         """
-        Compute the GGS log likelihood.
+        Compute the GGS log-likelihoo of the segmented Gaussian model.
```
miraep8, line 118:
```suggestion
        Compute the GGS log-likelihood of the segmented Gaussian model.
```

sktime/annotation/ggs.py
```diff
@@ -34,19 +43,46 @@ class GGS:
...
+    strength. It follows an interative procedure
+    where a new breakpoint is added and then adjusting all breakpoints to
+    (approximately) maximize the objective. It is similar to the top-down
+    search used in other change point detection problems.
+
     Parameters
     ----------
     k_max: int, default=10
-        Maximum number of change points to find
+        Maximum number of change points to find. The number of segments is thus k+1.
```
miraep8, line 65: 👍 Always a helpful reminder

sktime/annotation/ggs.py
```diff
@@ -307,34 +355,50 @@ def find_change_points(self, data: npt.ArrayLike) -> List[int]:
...
 class GreedyGaussianSegmentation(BaseEstimator):
     """Greedy Gaussian Segmentation Estimator.
 
+    The method apprixmates solutions for the problem of breaking a
+    multivariate time series into segments, where the data in each segment
+    could be modeled as independent samples from a multivariate Gaussian
+    distribution. It uses a dynamic programming search algorithm with
+    a heuristic that allows finding approximate solution in linear time with
+    respect to the data length and always yields locally optimal choice.
+
```
miraep8, line 368: This comment is not meant to be blocking - but I think what would push these docs to the next level would be a simple breakdown of the strengths (eg - no need for prior knowledge of the values of the means/sds, quick run time) and some of the drawbacks (knowledge of k, reliance on Gaussian assumptions etc) of this estimator in simple language.  Or even better - what types of problems does this algorithm work well with?  

I think having this sort of information readily available would make it so much easier for someone comparing the different annotation algorithms to have an idea of which one they actually need.

sktime/annotation/ggs.py
```diff
@@ -34,19 +43,46 @@ class GGS:
...
+    Greedy Gaussian Segmentation (GGS) fits a segmented gaussian model (SGM)
+    to the data by computing the appriximate solution to the combinatorial
+    problem of finding the approximate covariance-regularized  maximum
```
miraep8, line 55: I find it a bit confusing from a developer perspective that the docstring for the `GGS` class is the same as the one for `GreedyGaussianSegmentation` class below. I know this docstring isn't showing up on the website - so it isn't an issue there, but I think it would be helpful for anyone looking at your code in the future to be clued into the fact that `GGS` is the algorithmic implementation that is turned into an estimator in `GreedyGaussianSegmentation` here (just one or two sentences to give people a sense of the design would suffice).

sktime/annotation/ggs.py
```diff
@@ -146,65 +186,72 @@ def add_new_change_point(self, data: npt.ArrayLike) -> Tuple[int, float]:
...
-            time series data
+            2D `array_like` representing time series with sequence index along
+            the first dimension and value series as columns.
         change_points: list of ints
-            change points indexes
+            Locations of change points as integer indexes. By convention change points
+            include the identity segmentation, i.e. first and last index + 1 values.
         new_index: list of ints
-            new change points
+            New change points
```
miraep8, line 248: 👍  much clearer now!

miraep8 (commented):
Happy to take another look/approve once the errors are fixed. Overall I think it looks great/makes it a lot easier to tell what is actually going on in GGS.  Found a few small typos and areas where slightly more detail could be added, but I don't consider my comments to be blocking.

sktime/annotation/ggs.py
lmmentel: :slightly_smiling_face:

sktime/annotation/ggs.py
lmmentel: thanks

sktime/annotation/ggs.py
lmmentel: Added some extra info. The original paper does not go into much depth about this and as with any hyperparameter it's a question of CV or intuition to choose good values.

sktime/annotation/ggs.py
lmmentel: I :100: % agree, that would be great to have. Right now it's hard to source that information since actual studies comparing these methods and  are not out yet. It would be great to come up with such a comparison but than it would probably be better suited to be share as a user guide in a separate doc/notebook rather than docstrings.

sktime/annotation/ggs.py
lmmentel: Added module level paragraph on the design.

miraep8 (approved):
Just need to fix issue with the trailing whitespace it seems.  Looks good 👍

miraep8 (approved):

GuzalBulatova (approved):
Looks ready to me!

GuzalBulatova:
`sktime/datasets/tests/test_data_io.py::test_load_solar` this test has nothing to do with changes made here, any idea why it's failing on this PR?

lmmentel:
> `sktime/datasets/tests/test_data_io.py::test_load_solar` this test has nothing to do with changes made here, any idea why it's failing on this PR?

It seems it's calling an external service (which fails to respond) to get some data, which I would argue is quite a bad practice for testing since as we're seeing it can easily break and has nothing to do with sktime itself. I've created an issue about it here https://github.com/sktime/sktime/issues/3758

PR #3742, opened 2022-11-10
by matthewmiddlehurst [DOC] Add missing estimators to classification API page

fkiraly:
Restored reverted #3735 by @MatthewMiddlehurst since I have a change request.

My change request:

Agreed with the addition of missing classifiers.

Disagree with the change in section sequence and overall sequence, this makes more obscure classifiers more prominent.

First listed should be more well-known classifiers or those that people would like to find, generalizations of common tabular classifiers, and/or sensible baselines.

Re categories, the first ones should be

* composition
* dummy
* feature-based
* distance-based

Inside categories, common classifiers or generalizations of well-known classifiers from sklearn should be first.

* in distance-based, should be KNN
* in interval-based, it should be the forests

Early classification and tabular classification should be separate rubrics.

### DIFF
--- a/docs/source/api_reference/classification.rst
+++ b/docs/source/api_reference/classification.rst
@@ -5,33 +5,6 @@ Time series classification
 
 The :mod:`sktime.classification` module contains algorithms and composition tools for time series classification.
 
-Base
-----
-
-.. currentmodule:: sktime.classification
-
-.. autosummary::
-    :toctree: auto_generated/
-    :template: class.rst
-
-    BaseClassifier
-
-.. currentmodule:: sktime.classification.deep_learning.base
-
-.. autosummary::
-    :toctree: auto_generated/
-    :template: class.rst
-
-    BaseDeepClassifier
-
-.. currentmodule:: sktime.classification.early_classification.base
-
-.. autosummary::
-    :toctree: auto_generated/
-    :template: class.rst
-
-    BaseEarlyClassifier
-
 Composition
 -----------
 
@@ -194,3 +167,30 @@ sklearn
 
     ContinuousIntervalTree
     RotationForest
+
+Base
+----
+
+.. currentmodule:: sktime.classification
+
+.. autosummary::
+    :toctree: auto_generated/
+    :template: class.rst
+
+    BaseClassifier
+
+.. currentmodule:: sktime.classification.deep_learning.base
+
+.. autosummary::
+    :toctree: auto_generated/
+    :template: class.rst
+
+    BaseDeepClassifier
+
+.. currentmodule:: sktime.classification.early_classification.base
+
+.. autosummary::
+    :toctree: auto_generated/
+    :template: class.rst
+
+    BaseEarlyClassifier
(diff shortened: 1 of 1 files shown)


MatthewMiddlehurst:
The structure is just order of appearance in the file structure, which ends up being alphabetical order with items in the same file being coupled. This is very similar to what sklearn does https://scikit-learn.org/dev/modules/classes.html.

I don't see the API reference as a recommendation page, and don't think we should assume what the user is looking for.

TonyBagnall:
I disagree with the proposed changes by @fkiraly and agree with this structure that reflects the code base. Refactoring the organisation of classifiers is a different issue and should be dealt with as such. This is purely updating the docs to be accurate. I also did not realise we could arbitrarily revert approved and merged PRs like this without any discussion.

TonyBagnall (changes requested):
this does not reflect the current code structure, #3735 was bett and approved by two core devs. Rather than reverting, it would seem better to just create a new PR

TonyBagnall (approved):
sorry, got confused, this is still @MatthewMiddlehurst changes, so I approve it (again)

MatthewMiddlehurst:
I was a little confused on what approving actually meant as well 😅

fkiraly:
> which ends up being alphabetical order

???
If you look at the very file you are pushing, the estimators are not in alphabetical order.
I would be happy with alphabetical order btw as an alternative solution.

But this is neither in order of the (modern English) alphabet, nor in order of imo importance or likely interest of the reader.

fkiraly:
> Rather than reverting, it would seem better to just create a new PR

@TonyBagnall, our process is that enought time needs to be given for consideration, please read the governance/decision making document.

If within the decision making period a core developer disagrees, the PR is reverted.

fkiraly:
on this PR: this is a revert of a revert.

So:

* my change request still stands until I declare it resolved
* this PR counts as the one of @MatthewMiddlehurst, and he cannot self-approve

MatthewMiddlehurst:
Alphabetical order of the file structure. We could do it for estimator name, but that could separate things such as BOSSEnsemble and IndividualBOSS. Up to you.

TonyBagnall:
>I would be happy with alphabetical order btw as an alternative solution.

I am also am happy with alphabetic by package and then by estimator

TonyBagnall (dismissed):

MatthewMiddlehurst:
Both packages and estimators should be in alphabetical order now. I also added deep learning, which was missing.

fkiraly:
I would also not put the base classes as a top, since they are irrelevant for a user (only important for extender-developer). Therefore, if at all, I would put them to the end.

MatthewMiddlehurst:
Someone may want to change forecasting however if thats what we are doing instead of just alphabetical

image

GuzalBulatova:
Re user-friendliness I too believe the best way is just to go with alphabetical order. I'd guess that nobody will go scrolling looking for a description - it's way easier to go with search bar instead. Hence doesn't matter where the Base class is imo.

GuzalBulatova (approved):
Looks all good to me

Issue #3743, opened 2022-11-10
[BUG] invalid keras import in tests (for some versions)


ltsaprounis:
It seems like there's some inconsistencies with keras imports between keras versions.
Looking at the toml file I don't see any bounds on the versions. 

In any case I'm on keras 2.8 and the following import in sktime/classification/tests/test_base.py raises an import error:
`from keras.optimizers import Adamax`

This causes the tests to fail during collection, which is annoying (especially for the VS code testing functionality)
to test that without running all the tests you can do: `pytest --collect-only .`

doing some debugging I found that the right import for my case is: 
`from keras.optimizer_v2.adamax import Adamax`

The 2 ways to resolve this (I suggest we do both):
- Add version bounds to avoid this in the future (also why are we using keras and not tensorflow.keras ?)
- move the import inside the test as it's a soft dependency anyway

fkiraly:
hm, but moving the import inside the test wouldn't solve the issue when executing the test?

Third option: check the version bound where the import changes, and have if/else conditional import

ltsaprounis:
> this should also fail in #2986
> 
> we should run the diagnostic more often.

Is it because soft-dependency imports should be inside the test?

fkiraly:
no, even if the soft dep imports are inside the tests, the diagnostic #2986 would still fail, but it would appear as a test failure (because you are then also testing the import).

The way to isolate the soft deps is have a `pytest.skipif` depending on package availability or estimator support, search for `.skipif` to find examples in the code base.

fkiraly:
couple unisolated imports have crept in since we last did this...

We should fix, opened an issue:
https://github.com/sktime/sktime/issues/3748

PR #3744, opened 2022-11-10
keras import quick-fix

ltsaprounis:
Quick fix for #3743 by moving the import inside the test.

### DIFF
--- a/sktime/classification/tests/test_base.py
+++ b/sktime/classification/tests/test_base.py
@@ -8,7 +8,6 @@ import pickle
 import numpy as np
 import pandas as pd
 import pytest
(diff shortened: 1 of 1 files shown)


ltsaprounis:
VS code testing feature: 

before fix: 
image

after fix:
image

fkiraly (approved):
Yes, this is better.

Not affecting this PR, I think we still should do this in a more principled manner to prevent this from happening in the future, see
https://github.com/sktime/sktime/issues/3748
https://github.com/sktime/sktime/pull/3750

fkiraly:
would also be great if you could do the `pytest.skipif` while you're already at it.

Issue #242, opened 2020-04-06
[ENH] Implement `TimeBinner` transformer, regular binning with aggregation for irregular time series


mloning:
Update (@fkiraly): we should implement a `TimeBinner` transformer, which takes as parameters: `bins`, a `pandas.IntervalIndex`, or an iterable of int or timedelta (of bin start/ends); `aggfunc`, a function with signature (1D iterable) -> float. 

Bins time series into the bins, and aggregates by `aggfunc`.

---
Tabularizer implicitly assumes a bin width of 1, creating a column from each time point. TimeBinner generalises the Tabularizer by allowing users to specify (and tune) the bin width and aggregation function.

mloning:
The Tabularizer is a transformer, but effectively reduces the series-as-features setting to the tabular setting. It takes in a nested data frame (or whatever data format we use for the series-as-feature setting) and returns a tabular array which is compatible with the tabular setting of scikit-learn. It thereby breaks the transformer scitype, which should have the same type for input and output. It would be more consistent to encapsulate the Tabularizer as a meta-estimator similar to the reduced forecaster meta-estimators.

sparkingdark:
Are we going to create a classifier for binned data or want to create bin in graphs @mloning

mloning:
@sparkingdark this is meant to be a transformer for binning time series data.

sparkingdark:
Is this up for working

mloning:
In principle yes, but there are some complications around the data container. You can give it a try focusing on 3d numpy arrays where you would up/down sample the third dimension (time points).

sparkingdark:
Okay basically  I can assume it as a f(x,y,time)

mloning:
For panel data in the 3d numpy format, it's `(n_instances, n_variables, n_timepoints)` - this describes the panel data version of the familiar feature matrix `X` from scikit-learn.

fkiraly:
@kcc-lion, are you going to implement the series-to-series version?

PR #3745, opened 2022-11-10
[ENH] `TimeBinner` transformation for temporal bin aggregation

kcc-lion:
#### Reference Issues/PRs

Implements `TimeBinner` transformer as requested in #242 

Adds the option to specify custom intervals which to use for aggregating time series/panel data. In that sense, it generalizes `Tabularizer`

No

#### PR checklist
- [ ] Check for functionality on panel data other than `basic_motions` dataset
- [ ] Implement alternatives to passing a `pd.IntervalIndex`

##### For all contributions

### DIFF
--- a/docs/source/api_reference/transformations.rst
+++ b/docs/source/api_reference/transformations.rst
@@ -100,6 +100,7 @@ Sklearn and pandas adapters
     :template: class.rst
 
     Tabularizer
+    TimeBinner
 
 .. currentmodule:: sktime.transformations.series.adapt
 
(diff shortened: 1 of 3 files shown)


fkiraly (dismissed):
Nice!

Non-blocking suggestions:

I would also allow arbitray functions `aggfunc` to be passed, to be used as a function if it is not a string.

Lines 139-150 I would deal with by a dictionary, i.e., keys being the strings and values the functions.

fkiraly:
oh, and it would be great if you added this to the transformers API reference in `docs`

PR #3607, opened 2022-10-18
[DOCS] Adding invite to fall dev days to website landing page.

miraep8:
Adding an invite to the sktime website for people to register for the upcoming dev days.

PR #3746, opened 2022-11-10
Revert "[DOCS] Adding invite to fall dev days to website landing page."

aiwalter:
Reverts sktime/sktime#3607

The dev days will be over after today, so we can remove that ad then :)


miraep8 (approved):

aiwalter:
however ad will still be on website until next release

PR #3747, opened 2022-11-10
[DOC] added new slack link

hadifawaz1999:
##### For all contributions

##### For new estimators


miraep8 (approved):
🎉 🎉 🎉

MatthewMiddlehurst:
Both links work for me, are people having issues with the previous one?

miraep8:
@MatthewMiddlehurst yes I had at least one person besides Ali tell me that the old link was broken, and we also confirmed with them that this new link works! I agree though - from someone who is already in the slack channel it is hard to tell that the old link doesn't work - one way to tell (I think) is if you try to use the old link in incognito mode you should get the error that new users were experiencing. 

On a plus side this new link should "never expire" or at least that is what slack told me when I made it, so hopefully we don't run into this problem again.

MatthewMiddlehurst (approved):

hadifawaz1999:
@MatthewMiddlehurst yes I actually tried the link for the first time and did not work for me but it did using the new link that was provided to me. Before adding the new link I noticed that the old one is working for me now (after that I have joined the slack) so I tested it in a FireFox private tab and it did not work, such as @miraep8 explained. After adding the new one I tested it and it works.

fkiraly:
🎉 , welcome @hadifawaz1999 !

Issue #3748, opened 2022-11-10
[BUG] `keras` and other soft dep imports should be properly isolated in tests


fkiraly:
See #3743 and https://github.com/sktime/sktime/pull/2986 - soft dependency imports in non-suite tests should:
(a) happen inside tests, and
(b) be isolated via `pytest.skipif`

Currently, some `keras` imports are not properly isolated.

The idea is that the full test suite should be runnable without any soft dependencies installed.

fkiraly:
you can use `_check_soft_dependencies` to return a boolean already, no try/except needed

Didn't know about `importorskip`, seems very useful!

PR #3750, opened 2022-11-10
[MNT] test soft dependency isolation in non-suite tests

fkiraly:
This PR partially addresses #3748 in a systematic manner by adding a CI element which runs all non-suite tests without soft dependencies.

If these are not properly isolated, then test collection or individual tests will fail.

This should catch the current issues mentioned in #3748.

Relies on https://github.com/sktime/sktime/pull/3760 where the remaining problems are being fixed.


fkiraly:
> This should catch the current issues mentioned in https://github.com/sktime/sktime/issues/3748.

Indeed it does, see fixes in #3760

PR #3751, opened 2022-11-10
[DOC] Regression api

TonyBagnall:

This adds CNNRegressor, TapNetRegressor and RocketRegressor to the regression api. I have also expanded the docstring for TapNetRegressor to be a bit more descriptive and added and init to the regression.deep_learning package.  CNNRegressor and TapNetRegressor should be private, but that would require deprecation, and is beyond the scope of this [DOC] PR.


MatthewMiddlehurst (dismissed):
API changes are done in the style of #3742. Should probably wait for that first, but I approve of the changes.

docs/source/api_reference/regression.rst
```diff
@@ -5,9 +5,14 @@ Time series regression
...
-    :no-members:
-    :no-inherited-members:
+All current sktime Regressors can be listed as follows
+
+```
+from sktime.registry import all_estimators
+cls = all_estimators(estimator_types="regressor")
+names = [i for i, _ in cls]
+print(names)
+```
```
MatthewMiddlehurst, line 15: This does not seem to be rendering as intended, it's all on one line. i.e. `from sktime.registry import all_estimators cls = all_estimators(estimator_types="regressor") names = [i for i, _ in cls] print(names)`.

docs/source/api_reference/regression.rst
```diff
@@ -30,3 +47,31 @@ Interval-based
...
+
+    BaseRegressor
+
+.. currentmodule:: sktime.regression.deep_learning
+
+.. autosummary::
+    :toctree: auto_generated/
+    :template: class.rst
+
+    BaseDeepRegressor
```
MatthewMiddlehurst, line 77: Both of these seem to be missing also. Maybe the whitespace is necessary?

MatthewMiddlehurst (changes requested):
Some small bits.

docs/source/api_reference/regression.rst
TonyBagnall: how is it now? cant really tell in the preview

docs/source/api_reference/regression.rst
TonyBagnall: should be fixed

MatthewMiddlehurst (approved):
I made a push fixing the module import, both base classes now show on the page.

PR #3761, opened 2022-11-11
[ENH] Add get_test_params to deep learning classifiers and regressors

TonyBagnall:
#### Reference Issues/PRs

fixes https://github.com/sktime/sktime/issues/3633
may fix https://github.com/sktime/sktime/issues/3525

This PR adds the class method get_test_params to CNN, FCN, MLP classifiers and CNN regressor and adjusts the TapNet parameters to stop the intermittent memory issue in testing. Each get_test_params returns two sets of parameters. 
It also removes TapNet from the excluded classifier list. I have run these new TapNet parameters many times on CI and locally without problem. See #3620 

No

@achieveordie and @AurumnPegasus could you look at the parameter settings and advise if they should change? thanks

##### For all contributions


sktime/classification/deep_learning/tapnet.py
```diff
@@ -254,19 +254,20 @@ def get_test_params(cls, parameter_set="default"):
...
             "dilation": 2,
-            "layers": (50, 25),
+            "layers": (32, 16),
         }
-
         param2 = {
-            "n_epochs": 100,
-            "use_cnn": False,
-            "layers": (25, 25),
+            "n_epochs": 20,
```
achieveordie, line 266: Is there a reason for including convolutional layers here? Since we're already covering it in the first param set, we could directly set `use_cnn: False` since the current set is a subset (and hence more resource-hungry) than what I'm proposing.

The current set is good as well, I was just thinking in terms of improving coverage.

achieveordie (commented):
Only have one doubt, everything else LGTM.

achieveordie:
Also, let's run tests multiple times to make sure everything works since initial failures were sporadic. You're probably doing that already.

TonyBagnall:
> Also, let's run tests multiple times to make sure everything works since initial failures were sporadic. You're probably doing that already.

I have run these on https://github.com/sktime/sktime/pull/3620. The sporadic fault was an exceeding memory problem, so its reasonable to assume these configs remove it. Its simple enough to exclude the classifier again if it crops up

AurumnPegasus:
Hey, just a question about this:

If I am not wrong the current workflow would be:
```
cnn = CNNClassifier()
param1, param2 = cnn.get_test_params()
cnn1 = CNNClassifier(param1)
cnn2 = CNNClassifier(param2)
```

In this case, we need to create an object of `CNNClassifier` to even get the test parameters, right? In my opinion that would be a bit inefficient right?

If we could maybe define a common function which accepts the name of estimator we want the test parameters for, and returns the test parameter from a dict for parameter it contains, imo that would be better for
a. It could become consistent across all estimators, DL or not.
b. You do not need to create an extra object of the class to just get the parameters

So in that case the workflow would be

```
def get_test_params(self, name):
    return paramdict[name]

params = get_test_params('cnnclassifier')
cnn = CNNClassifier(params)
```

sktime/regression/deep_learning/tapnet.py
```diff
@@ -235,18 +235,17 @@ def get_test_params(cls, parameter_set="default"):
             `create_test_instance` uses the first (or only) dictionary in `params`.
         """
         param1 = {
-            "n_epochs": 25,
-            "batch_size": 32,
+            "n_epochs": 10,
+            "batch_size": 4,
             "padding": "valid",
-            "filter_sizes": (64, 64, 64),
+            "filter_sizes": (64, 64, 32),
```
achieveordie, line 241: Let's reduce the size to `(16, 16, 16)` as in the case of classification.

TonyBagnall:
> Hey, just a question about this:
> 
> If I am not wrong the current workflow would be:
> 
> ```
> cnn = CNNClassifier()
> param1, param2 = cnn.get_test_params()
> cnn1 = CNNClassifier(param1)
> cnn2 = CNNClassifier(param2)
> ```
> 
> In this case, we need to create an object of `CNNClassifier` to even get the test parameters, right? In my opinion that would be a bit inefficient.
> 
> If we could maybe define a common function which accepts the name of estimator we want the test parameters for, and returns the test parameter from a dict for parameter it contains, imo that would be better for a. It could become consistent across all estimators, DL or not. b. You do not need to create an extra object of the class to just get the parameters
> 
> So in that case the workflow would be
> 
> ```
> def get_test_params(self, name):
>     return paramdict[name]
> 
> params = get_test_params('cnnclassifier')
> cnn = CNNClassifier(params)
> ```

hi,

thanks for that. The get_test_params function is used in the testing routines, and is not really meant for general use. The test_all_estimators test interface compliance rather than functionality, and they call fit a very large number of times. Because of this we need tiny models to stop the overall test time ballooning (this is a constant problem). 

We do not want to set the default settings to be tiny, as it risks people using them and drawing false conclusions (see here, https://arxiv.org/abs/1703.06777, reviewers didnt like our muppet story :) ), so we use this function. 

It returns a list of parameter settings because thats how the overall toolkit is structured. See https://github.com/sktime/sktime/pull/3544
addressing any issues related to that is well beyond the scope of this PR

sktime/classification/deep_learning/tapnet.py
TonyBagnall: changed to set use_cnn to False

MatthewMiddlehurst (approved):
Better than the current state for sure. If failures start appearing again we can just add them back to the exclude list.

MatthewMiddlehurst:
There is a TapNet crash on main likely related to this PR. https://github.com/sktime/sktime/actions/runs/3481733151/attempts/1

```
[gw0] win32 -- Python 3.7.13 C:\Miniconda\envs\test\python.exe
worker 'gw0' crashed while running 'sktime/tests/test_all_estimators.py::TestAllEstimators::test_fit_returns_self[TapNetRegressor-0-ClassifierFitPredict]'
```

Not a very helpful error message, if another happens probably add both back to exclude.

PR #3767, opened 2022-11-11
[ENH] Warning rather than error for TDE small series

MatthewMiddlehurst:
Rather than throwing an error when the minimum window length is too small, this PR changes `TemporalDictionaryEnsemble` to warn the user and lower the minimum window length instead.


patrickzib (approved):
Looks fine to me!

Issue #3764, opened 2022-11-11
[BUG] `ColumnEnsembleForecaster` fails with hierarchical exogenous data


RikStarmans:
ColumnEnsembleForecaster fails with Aggregator based forecaster and exogenous data.

**To Reproduce**

```python
from sktime.datatypes import get_examples
from sktime.datatypes._utilities import get_window
from sktime.forecasting.sarimax import SARIMAX
from sktime.transformations.hierarchical.aggregate import Aggregator
from sktime.transformations.hierarchical.reconcile import Reconciler
from sktime.forecasting.compose import ColumnEnsembleForecaster

X = get_examples("pd_multiindex_hier")[0]
y = get_examples("pd_multiindex_hier")[1]

X_train = get_window(X, lag=1)
y_train = get_window(y, lag=1)
X_test = get_window(X, window_length=1)

f = Aggregator() * (Aggregator() ** SARIMAX()) * Reconciler()

f.fit(y=y_train, X=X_train, fh=1)
print('this works')

f_hat = ColumnEnsembleForecaster(f)
f_hat.fit(y=y_train, fh=1)
print("still working")
f_hat.fit(y=y_train, X=X_train, fh=1)
print("never reached, crash")
```
**Proposed Solution**
Add the following tag to ColumnEnsembleForecaster
```python
   X_inner_mtype: [
            "pd.Series",
            "pd.DataFrame",
            "pd-multiindex",
            "pd_multiindex_hier",
        ],
```
**Additional context**
ColumnEnsembleForecaster inherits from _HeterogenousEnsembleForecaster which inherits from
BaseForecaster.
If fit is called with exogenous data, the BaseForecaster changes the type of both X and Y to vector.
The problem is triggered here https://github.com/sktime/sktime/blob/18b502e969d9c8d8dafce5cea78eb0c43fe23d91/sktime/forecasting/base/_base.py#L288
And can be resolved by replacing this line with
```python
#X_inner, y_inner = self._check_X_y(X=X, y=y)
X_inner, y_inner = X, y
```
Check_X_y seems to run into issues at this line
https://github.com/sktime/sktime/blob/18b502e969d9c8d8dafce5cea78eb0c43fe23d91/sktime/forecasting/base/_base.py#L1331
If an X is passed, vectorization is triggered for both X and y.
In specific, in line 
[...]

PR #3768, opened 2022-11-11
& RikStarmans [BUG] fix `ColumnEnsembleForecaster` for hierarchical `X`

fkiraly:
Fixes https://github.com/sktime/sktime/issues/3764, with the fix proposed by @rikstarmans (with a minimal change)

Includes the failure case in #3764 as a test.


GuzalBulatova (approved):
Makes sense, all tests pass

PR #3775, opened 2022-11-13
[BUG] Update File Format Specifications page to show list of hyperlinked formats

### LINKED ISSUE #3772: [BUG] Empty File Format page in API Reference
**Describe the bug**

File Format Page (available [here](https://www.sktime.org/en/stable/api_reference/data_format.html)) should have reference to different file formats present in `sktime` (`.ts` for example). But the current page only provides a one-liner with no references.

This is happening because the `.rst` file is improper and needs to be corrected. I'll make a PR correcting it in a while.
 
**To Reproduce**

Here's what I see:
fail

achieveordie:
#### Reference Issues/PRs

Fixes #3772

The problem was in `toctree` part of the `.rst` file. It has been updated and now `.ts` format is visible (in the local build).

No

Feedback on whether the resulting page (`API Reference/File Format Specifications`) look good.

##### For all contributions

##### For new estimators

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -597,7 +597,6 @@
         "bug",
         "code",
         "data",
-        "design",
         "doc",
         "review",
         "test"
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -596,7 +596,11 @@
       "contributions": [
         "bug",
         "code",
-        "doc"
+        "data",
+        "design",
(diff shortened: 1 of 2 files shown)


fkiraly (approved):
Yes, this fixes it.

It's a bit odd that the page is aligned with the bottom and not the top, but I suppose that is default formatting.

Issue #2905, opened 2022-07-01
[BUG] contributors.md not updating?


fkiraly:
Update: there's a "hard" and an "easy" (good first issue) problem here.

easy = running the cmdline tool locally and making a PR with the generated contributors.md

hard = automating this in the CI (e.g., release script)

(and we close it only once we have automation done)

---

It seems the contributors.md has not updated since 0.9.0 - are we doing sth wrong with the release?

Did we accidentally drop sth from the release script when we moved to GHA that re-generated the contributors.md?

FYI @lmmentel, @freddyaboulton

lmmentel:
Good catch, do you know how were they updated in the past? I've checked the history and those updates are a part of release PR but it's hard to say if this was a manual or automated edit.

fkiraly:
> Good catch, do you know how were they updated in the past? I've checked the history and those updates are a part of release PR but it's hard to say if this was a manual or automated edit.

I suspect, one of these:
https://allcontributors.org/docs/en/tooling

@mloning might know more.

fkiraly:
Docs are still here:
https://allcontributors.org/docs/en/cli/installation

>Think it's still manual. 

Do you mean manual, as there is something you can locally and commit - or, manual as in having to translate the all-contributors line by line??

mloning:
> Do you mean manual, as there is something you can locally and commit - or, manual as in having to translate the all-contributors line by line??

Using their command line client.

fkiraly:
hm, looks straighforward enough - so someone should be able to do that locally and simply make a PR with the updated file, until we figure out how to automate it.

achieveordie:
I think the most direct solution will be to have a GitHub bot perform this operation whenever it notices a change in `.all_contributorsrc`. There's already a [bot](https://allcontributors.org/docs/en/bot/overview) that they provide but even that requires a cue to perform a PR so we might as well create one for added flexibility.

PR #3781, opened 2022-11-14
[DOC] Update CONTRIBUTORS.md

achieveordie:
#### Reference Issues/PRs

See #2905 for reference.

Locally ran `all-contributors-cli` via npm to update `CONTRIBUTORS.md` to include new changes. The last update was two months ago so I thought it'll be great to have it updated.

None

##### For all contributions


fkiraly (approved):
Nice!

In total it seems to have added three lines though, can that be correct?

Also, I was wondering whether this can be automated, say, as part of the release workflow?

achieveordie:
> In total it seems to have added three lines though, can that be correct?

Yes, it's correct since the number of contributors has gone up from 176 before this PR to 179 in this PR with one line per new contributor.

> Also, I was wondering whether this can be automated

That's what I've been considering too. I made a comment in the referenced issue that we could automate this via a GitHub bot, but I suppose GitHub doesn't provide native support to bots (they'd have to be deployed somewhere else and integrated to GH). So we could maybe try to add an additional step in the workflow. But since this involves adding and managing node packages I'm not sure if that's the best course of action.

What are your opinions?

fkiraly:
> What are your opinions?

Well, if I had a good solution in mind, I would have posted it in #2905 or already implemented it - whatever works and isn't too complex to maintain and implement.

achieveordie:
In that case, we could find a viable solution in by discussing the possible solutions in #2905 - I gave one of the possible solutions, maybe someone over there knows a better way to do it.

lmmentel:
Nice work @achieveordie. Another possibility for automating would be to add a job to GitHub actions to run the command you ran locally on each PR merge.

achieveordie:
@lmmentel That's the only idea that can be implemented immediately since other solutions (GitHub App / Bot) would require a third-party integration which I understand `sktime` wouldn't want to pursue due to additional governance and cost (not to mention it needs to be set up by a core developer). So I'm going to proceed with the GH Actions.

Initially, my problem with this solution was how often this job would be repeatedly called since `.allcontributorsrc` rarely changes, but I suppose GH Actions must have an answer to it as well.

(Also, I think core developers could give GH Apps / Bots a thought since a lot of maintenance work could then be automated enabling maintainers to focus their time and energy in a better way)

Issue #3769, opened 2022-11-11
[ENH] Extend minirocket_multivariate to unequal / variable length time series


michaelfeil:
**Is your feature request related to a problem? Please describe.**

Currently, just a few unequal length multivariate models exist. From the minirocket_multivariate the authors implemented a option for variable length time series data. (https://github.com/angus924/minirocket/blob/main/code/minirocket_multivariate_variable.py)

**Describe the solution you'd like**
Implement a MiniRocketMultivariate under as `sktime.transformations.panel.rocket` . While the original author's code is already used in the other rocket implementations.
From sktime only a small interface is needed to convert the X (sktime-compatible Panel data format)
into n 1-D series with a additional vector on length information.
```
def fit(X, y=None):
     """
     X : sktime-compatible Panel data format, with n_samples series
```
Variable-Length Input (Experimental)

**Describe alternatives you've considered**
Using the minirocket_multivariate using trunancation or padding transformers. 

**Additional context**

michaelfeil:
I would like to submit a PR on this.

I would include the following:
- interface to https://github.com/angus924/minirocket/blob/main/code/minirocket_multivariate_variable.py as `MiniRocketMultivariateVariable`, essentially converting a `"X_inner_mtype": "df-list"` to a 

 ```
 X : df-list (list of dataframes, with multivariate ts
 
 Returns
        np.array: 2D array of shape = [n_dimensions, sum(length_series(i) for i in n_instances)], np.float32
        np.array: 1D array of shape = [n_instances] with length of each series, np.int32
```
- test on JapaneseVowels (multivariate and unequal length time-series)
- the authors code, as in https://github.com/sktime/sktime/blob/main/sktime/transformations/panel/rocket/_multirocket_multivariate.py, with no modifications except for seeding.

```
accs = []
    for i in range(100):
        classifier = make_pipeline(
            MiniRocketMultivariateVariable(random_state=i),
            StandardScaler(with_mean=False),
            RidgeClassifierCV(alphas=np.logspace(-3, 3, 10)),
        )
        classifier.fit(X_training, Y_training)
        accs.append(
            classifier.score(X_test, Y_test)
        )
```
`[0.9891, 0.9864, 0.9837, 0.9945, 0.9891]`
I get a mean acc of ca. `0.987` and variance of `8.77e-06`.

JapaneseVowels reported results https://arxiv.org/pdf/1711.11343.pdf
[...]

PR #3786, opened 2022-11-14
[ENH] Add minirocket_multivariate_variable for unequal length time series

michaelfeil:
#### Reference Issues/PRs
Refers to #3769 .

This PR adds an unequal length, multivariate time series https://github.com/angus924/minirocket/blob/main/code/minirocket_multivariate_variable.py.
The MinirocketMultivariateVariable works similar to the MinirocketMultivariate, however can process multivariate timeseries with variable length, therefore requires an own Sktime interface. 

No.

As this is my first contribution, give a broad feedback:
- if the class is implemented correctly, and has sufficient tests
- if the formatting is correct (check doc strings)
- is the doc update sufficient (did not find more for other models)

#### Any other comments?
- also updated the notebook, where the `RidgeClassifierCV` has a deprecation warning of `normalizer=True`
- 
##### For all contributions
- [x] I've added unit tests and made sure they pass locally. (Python 3.10 conda env, wsl2 )
- [x]The PR title starts with either [ENH], [MNT], [DOC], or [BUG] indicating whether the PR topic is related to enhancement, maintenance, documentation, or bug.

##### For new estimators


MatthewMiddlehurst:
Can look at this tomorrow. There is currently a conflict with main, however.

michaelfeil:
> Can look at this tomorrow. There is currently a conflict with main, however.

should be solved, was just the contributors.

sktime/transformations/panel/rocket/_minirocket_multivariate_variable.py
```diff
@@ -0,0 +1,1185 @@
...
+
+    .. [2] Angus Dempster, Daniel F Schmidt, Geoffrey I Webb
+           https://github.com/angus924/minirocket
+    """
+
+    _tags = {
+        "univariate-only": False,
+        "fit_is_empty": False,
+        "scitype:transform-input": "Panel",
+        "scitype:transform-output": "Panel",
```
MatthewMiddlehurst, line 78: These should be consistent with the other transformers.

sktime/transformations/panel/rocket/_minirocket_multivariate_variable.py
```diff
@@ -0,0 +1,1185 @@
...
+    A provisional and naive extension of MINIROCKET to multivariate input
+    with unequal length provided by the authors [2]_ .  For better
+    performance, use the sktime class MiniRocket for univariate input,
+    and MiniRocketMultivariate to equal length multivariate input.
+
+    Parameters
+    ----------
+    num_kernels              : int, number of random convolutional kernels
+                               (default 10_000).
+                               Reduced to the closest multiple of 84.
```
MatthewMiddlehurst, line 35: [This](https://sktime--3786.org.readthedocs.build/en/3786/api_reference/auto_generated/sktime.transformations.panel.rocket.MiniRocketMultivariateVariable.html#sktime.transformations.panel.rocket.MiniRocketMultivariateVariable) is the generated docs for the estimator. There are a couple of issues, but the other transformers also have them. Not a requirement to fix, but making you aware.

examples/minirocket.ipynb
```diff
@@ -42,17 +42,25 @@
...
    "source": [
     "import numpy as np\n",
     "from sklearn.linear_model import RidgeClassifierCV\n",
     "from sklearn.pipeline import make_pipeline\n",
+    "from sklearn.preprocessing import StandardScaler\n",
     "\n",
-    "from sktime.datasets import load_arrow_head  # univariate dataset\n",
+    "from sktime.datasets import (\n",
+    "    load_arrow_head,  # multivariate dataset with unequal length; univariate dataset\n",
+    ")\n",
```
MatthewMiddlehurst, line 56: This comment seems incorrect.

examples/minirocket.ipynb
```diff
@@ -120,12 +132,13 @@
    "source": [
     "We suggest using `RidgeClassifierCV` (scikit-learn) for smaller datasets (fewer than ~10,000 training examples), and using logistic regression trained using stochastic gradient descent for larger datasets.\n",
     "\n",
-    "**Note**: For larger datasets, this means integrating MiniRocket with stochastic gradient descent such that the transform is performed per minibatch, *not* simply substituting `RidgeClassifierCV` for, e.g., `LogisticRegression`."
+    "**Note**: For larger datasets, this means integrating MiniRocket with stochastic gradient descent such that the transform is performed per minibatch, *not* simply substituting `RidgeClassifierCV` for, e.g., `LogisticRegression`.\n",
+    "**Note**: While the input features of MiniRocket are unscaled, we recommend scaling the features when passing them to another classifier such as `RidgeClassifierCV`."
```
MatthewMiddlehurst, line 136: Is this something recommended for all classifiers? I know the default RidgeClassifierCV normalises, but do they recommend it for others as well?

MatthewMiddlehurst:
FYI @angus924 in case you want to comment on this also.

sktime/transformations/panel/rocket/_minirocket_multivariate_variable.py
michaelfeil: indeed, the ` "scitype:transform-output": "Panel"`. Can be modified to to `primitive`, like in the other rockets.

`"scitype:transform-input"` can't be modified to numpy3d, as the input data is unequal length time series data (+multivariate). Therefore, this minirocket needs an extra class from sktime. 
MiniRocket / MiniRocketMultivariate strictly maybe do not, however, there is a significant amount of different numba code as well.

sktime/transformations/panel/rocket/_minirocket_multivariate_variable.py
michaelfeil: Okay, should be possible to fit!

examples/minirocket.ipynb
michaelfeil: True, was not meant to be an editorial recommendation. However, `RidgeClassifierCV(normalize=True)` has now a deprecated `normalize` arg, which is why we may use e.g. StandardScaler() in this tutorial for forward-compatability.

As in https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.RidgeClassifierCV.html
`Deprecated since version 1.0: normalize was deprecated in version 1.0 and will be removed in 1.2. `

michaelfeil:
> FYI @angus924 in case you want to comment on this also.
It would be great to have his approval + comment on code before merging it to this library.

angus924:
Hi, sorry, a bit late to this.

You definitely don't need my approval!

I only have a couple of trivial comments, sorry if I've missed anything.

> Generally for all the ROCKET stuff, I'm not really a fan of the new transformer for each capability approach.

I agree.  The problem is the underlying implementation...  For example, the multivariate implementation is (or, at least, was) meaningfully slower than the univariate implementation.  I have never been able to work out why (although it's been quite a while since I looked at it properly).  Hence, we ended up with two versions: the "canonical" univariate version, and the multivariate version which is ok but flawed in some inscrutable way.  Likewise with variable-length input, it requires handling and iterating over the input data differently (you might even describe it as a "hack"), so you end up with another version again.

If there's a way to get it working with a single backend without compromising efficiency too much, that would be great.  Maybe the different versions are not as different in terms of runtime as I remember.

In any case, I'm happy to go with whatever you think makes sense.

Re the ridge classifier... this is a strange one.  The sklearn implementation changed, deprecating the `normalize` argument, and they suggest using StandardScaler instead.  The odd thing is, StandardScaler does something different to what the `normalize` argument did: `normalize` scales by the L2 norm, `StandardScaler` scales by the standard deviation (although I think these quantities should be within a constant factor of each other?).  It shouldn't make a whole lot of difference in the broad scheme of things, but technically you need to scale by the L2 norm to reproduce the behaviour of the deprecated `normalize` argument.

MatthewMiddlehurst:
I can take a look at some point. If they make a difference that's fine, efficient code for these capabilities is better. If that's the case, we could possibly make these private and create a public facing class which selects an implementation based on the input data?

You can set options for `StandardScaler`, the classification package classifier uses `StandardScaler(with_mean=False)`.

examples/minirocket.ipynb
MatthewMiddlehurst: I think the note is fine now. We should probably use `StandardScaler(with_mean=False)` to keep it consistent with the classifier, however.

sktime/transformations/panel/rocket/_minirocket_multivariate_variable.py
MatthewMiddlehurst: These tags should not change the output at all, the mtype tag for df-list is what sets the input datatype.

It is slightly odd but current convention is to set the tags
```
        "scitype:transform-output": "Primitives",
```
for transformers which take time series as an input and output primitives (even if the input is a panel).

If I'm wrong and this does actually impact the transformer, let me know.

TonyBagnall:
I'm fine with this going in once ready, but I agree medium term we need to find a way to unify them, I am just tweaking the docstrings and was scratching my head as to why we had duplicates! so I think we probably need an issue to unify the rocket transformers without compromising efficiency. If no one else solves it we can take a look, probably next year though

sktime/transformations/panel/rocket/_minirocket_multivariate_variable.py
michaelfeil: Indeed, now it works!

michaelfeil:
Should be all done now!

MatthewMiddlehurst (approved):
I am happy with the changes made. Will give it a bit to see if anyone else has input, then merge if not.

PR #3793, opened 2022-11-15
[DOC] `sktime` intro notebook

fkiraly:
This PR adds a consensus version of the mature introduction notebook from various python conferences as notebook 0 to the examples.


achieveordie (commented):
I think this is good to go but there are some non-blocking comments I'd like to make:
- Renaming the title from `0 Introduction to sktime` to just `Introduction to sktime` would be good since this is what a user would see when they visit the User Guide.
- Likewise, renaming sections from 0.1, 0.2, etc to simply 1, 2.
- I'd also recommend adding a statement about `sktime`'s deployment capabilities as the fourth point in `Summary`.

fkiraly:
> sktime's deployment capabilities

excellent idea!

> Renaming the title from 0 Introduction to sktime

I thought of this as a python in-joke, "starting to count with 0". Also the remaining notebooks are already numbered, too, they just start at 1 right now.

achieveordie:
> Also the remaining notebooks are already numbered, too, they just start at 1 right now

By renaming I meant the title of the notebook (marked by the '#' in the markdown file, the filename remains the same) since this line is what will be shown in the user guide which can get [messy](https://github.com/sktime/sktime/pull/5312#issue-1910241649) from my recent experience. There are indeed some notebooks that have a numbered title but the ones in the user guide have a number-free title. Since this notebook will belong there it might be confusing to some new-comers.

Completely agree with 0-indexing it though.

fkiraly:
I see! How about keeping the number in the file name, but losing it in the first header?

achieveordie:
Yes, that's exactly what I had in mind.

fkiraly:
changed the section headers - more content I consider out of scope, for another PR, if desired.

Issue #3776, opened 2022-11-13
[BUG] Transformers which change data from series to panel cannot be used in pipelines with exogenous data


RikStarmans:
**Describe the bug**
A forecaster which contains a transformer for both X and Y and converts series to panel, cannot be used.

**To Reproduce**
```python
from sktime.datasets import load_airline
from sktime.forecasting.naive import NaiveForecaster
from sktime.transformations.bootstrap import STLBootstrapTransformer
y = load_airline()
f = STLBootstrapTransformer()*(STLBootstrapTransformer()**NaiveForecaster())
f.fit(y=y)
# does not crash
f.fit(y=y,
      X=y)
# Error
# TypeError: X and y must have the same scitype
```
**Explanation and solution**
The pipeline looks as follows
```
TransformedTargetForecaster(steps=[STLBootstrapTransformer(),
                                   ForecastingPipeline(steps=[STLBootstrapTransformer(),
                                                              NaiveForecaster()])])
```
The STLBootstrapTransformer is performed on y. Which changes the type from Series to Panel.
The forecastingPipeline then receives and X which is a Series and y which is a panel and fails.
This can only be resolved if it account for the fact that internally, it also transform Series to Panel.

**Versions**
'0.14.0'

fkiraly:
hmmmm - what exactly would you expect the happen in this case, @RikStarmans?

Would you expect `f` to take the i-th bootstrap instance of `X` and the i-th bootstrap instance ot `y`?

The bootstrap transformer creates multiple instances, but they bootstrap instances of `X` and `y` will be independently created of each other.

fkiraly:
I understand where the discrepancy between status quo and expected is coming from:

* currently, we block all instances where `X` and `y` have a different scitype, for atomic forecasters this typically results in an error
* however, in this case it makes sense for the composite (all three components). The problem is how it is built up - in the middle layer, which is also a forecaster, we have `X` and `y` that do not have the same scitype (number of levels in `pandas`).

We could relax the check and trust the user they are building sth sensible - at the cost that error messages are less informative in some genuine error cases.

fkiraly:
(also, I didn't expect anyone to build this kind of pipeline, but it makes sense!)

RikStarmans:
Great.. the code is indeed just an example..

RikStarmans:
See above added example of why I believe I need it.

fkiraly:
That's a very nice transformer!

[...]

PR #3794, opened 2022-11-15
[BUG] Remove check_X_y for pipelines

RikStarmans:
#### Reference Issues/PRs

Fixes https://github.com/sktime/sktime/issues/3776

A forecaster which contains a transformer for both X and Y and converts series to panel, cannot be used.
The following forecaster does not make sense. Operations like these are needed for temporal aggregators.
```python
from sktime.transformations.bootstrap import STLBootstrapTransformer
y = load_airline()
f = STLBootstrapTransformer()*(STLBootstrapTransformer()**NaiveForecaster())
f.fit(y=y)
# fail in the current code base, due to TypeError: X and y must have the same scitype
# it should not fail as X and y have the same type after the trafos
# this is fixed by this push
f.fit(y=y,
      X=y)
# the following still results in a TypeError
f = NaiveForecaster()
try:
    f.fit(y=y,
          X=STLBootstrapTransformer().fit_transform(y))
except TypeError:
    print('Catched a type error as expected')
```

No

Changes are minimal, I did not do any test or added any tests.

##### For all contributions

##### For new estimators


fkiraly (changes requested):
Great, thanks for the initiative!

May I request:

* I would suggest a slightly more principled version of this along the lines in my newest post in #3776, here: https://github.com/sktime/sktime/issues/3776#issuecomment-1315642986
* also, kindly make sure you don't accidentally include your other pull request (is that intended or accident? If intended, comment on this in the PR description, can make sense for testing or conditionalities)

RikStarmans:
> Great, thanks for the initiative!
> 
> May I request:
> 
> * I would suggest a slightly more principled version of this along the lines in my newest post in #3776, here: [#3776 (comment)](https://github.com/sktime/sktime/issues/3776#issuecomment-1315642986)
Fixed
> * also, kindly make sure you don't accidentally include your other pull request (is that intended or accident? If intended, comment on this in the PR description, can make sense for testing or conditionalities)
Fixed

fkiraly (changes requested):
Yes, makes sense.

* linting fails, should be fixed
* should the check not be in the line `if y_scitype != X_scitype:`?

Issue #3788, opened 2022-11-15
[BUG] HIVECOTEV2 with n_jobs>1 crashes on internal dataset


johan-sightic:
**Describe the bug**
I want to fit HIVECOTEV2 on an internal dataset in "df-list" format with two classes. Running it in a singe thread `HIVECOTEV2(time_limit_in_minutes=1)` works as expected and outputs no warnings or errors. However I want to speed it up with multiprocessing `HIVECOTEV2(time_limit_in_minutes=1, n_jobs=14)` but this creates `NumbaPendingDeprecationWarning`:s, overflow warnings and then crashes with a `PicklingError`.

**To Reproduce**
Unfortunate I can't share the data since it contains sensitive personal information, please tell me if I should check something in the data. Hopefully you can figure something out from the logs.

Load and verify data:
```python
samples = load_data()

check_raise(samples[0], "pd.DataFrame")
check_raise(samples, "df-list")

print(any([df.isnull().values.any() for df in samples]))  # False
print(any([~np.isfinite(df.values).all() for df in samples]))  # False
```

Fit single process creates a working model without any warnings or errors.
```python
model = HIVECOTEV2(time_limit_in_minutes=1)
model.fit(samples, np.array(targets, dtype=int))
```

Fit multiprocessing shows warnings and then crashes.
```python
model = HIVECOTEV2(time_limit_in_minutes=1, n_jobs=14)
model.fit(samples, np.array(targets, dtype=int))
```
```
/home/johan/dev/vAll/lib/python3.10/site-packages/numba/core/ir_utils.py:2147: NumbaPendingDeprecationWarning: 
Encountered the use of a type that is scheduled for deprecation: type 'reflected list' found for argument 'sorted_indicies' of function '_online_shapelet_distance'.

For more information visit https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-reflection-for-list-and-set-types

File "../../../../vAll/lib/python3.10/site-packages/sktime/transformations/panel/shapelet_transform.py", line 1473:
@njit(fastmath=True, cache=True)
def _online_shapelet_distance(series, shapelet, sorted_indicies, position, length):
^

  warnings.warn(NumbaPendingDeprecationWarning(msg, loc=loc))
/home/johan/dev/vAll/lib/python3.10/site-packages/numba/core/ir_utils.py:2147: NumbaPendingDeprecationWarning: 
[...]

PR #3796, opened 2022-11-15
[BUG] Fix ContinuousIntervalTree and RandomShapeletTransform deprecation warnings

MatthewMiddlehurst:
Solves the deprecation part of #3788 by changing the catch22 function used in `ContinuousIntervalTree` and integrating `numba` typed-lists into `RandomShapeletTransform`.

Additionally:
- Adds myself as the CODEOWNER of `catch22wrapper.py` (missed from the PR adding it)
- Adds a `parallel_backend` parameter to RandomShapeletTransform and sets the preferred default to "threads"

### DIFF
--- a/sktime/transformations/panel/shapelet_transform.py
+++ b/sktime/transformations/panel/shapelet_transform.py
@@ -17,7 +17,8 @@ from operator import itemgetter
 import numpy as np
 import pandas as pd
 from joblib import Parallel, delayed
-from numba import NumbaPendingDeprecationWarning, njit
+from numba import njit
+from numba.typed.typedlist import List
 from sklearn import preprocessing
 from sklearn.utils import check_random_state
 from sklearn.utils.multiclass import class_distribution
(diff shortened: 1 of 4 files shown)


sktime/transformations/panel/shapelet_transform.py
```diff
@@ -999,6 +1000,9 @@ class RandomShapeletTransform(BaseTransformer):
     n_jobs : int, default=1
         The number of jobs to run in parallel for both `fit` and `transform`.
         ``-1`` means using all processors.
+    parallel_backend : str, default=None
+        Specify the parallelisation backend implementation in joblib, where
```
TonyBagnall, line 1004: what are valid values for this string?

MatthewMiddlehurst: Ah yeah that is missing, will add it now.

TonyBagnall (approved):

PR #3807, opened 2022-11-17
[MNT] Automate updating CONTRIBUTORS.md

### LINKED ISSUE #2905: [BUG] contributors.md not updating?

achieveordie:
#### Reference Issues/PRs

Fixes #2905

Add a GH workflow that is triggered when changes are made to `.all-contributorsrc` in a PR. This works by setting up the environment and calling `all-contributors-cli generate` followed by making an automatic commit to reflect the changes in `CONTRIBUTORS.md`.

None

- Location of this workflow (Does it make sense to add it in a different file as opposed to already existing workflows)
- Is this the most cost-effective way to implement it? (despite being free we do not want any workflow to needlessly run and waste computation)

##### For all contributions


achieveordie:
So there seems to be an unsupported type called "build" that has been added by @jorenham and @wolph. I do not find any support for the type "build" [here](https://allcontributors.org/docs/en/emoji-key). Could you please confirm this and possibly update/let me know what to update it with?

@lmmentel My main concern is that the workflow is running despite being asked to run only when changes are made to `.all-contributorsrc`. Is something wrong with the code that you can see at the first glance?

lmmentel:
Hey @achieveordie, nice work. I think there is still an unrelated error in the workflow file, have you checked it works locally? On the workflow config I would try changing `path:` to `paths:` in the `on:` section of the yaml.

achieveordie:
> an unrelated error in the workflow file

Oh, what might that be? I could only see why we're getting the current error. I might have to look more to understand if there are further problems with it.

> have you checked it works locally?

Sorry, I haven't. What do you suggest I should use? I [see](https://stackoverflow.com/questions/59241249/how-to-run-github-actions-workflows-locally) suggestions but I'm not sure if you're actually referring to that. It escaped my mind that there might be a way to locally test it.

lmmentel:
The error seems to be
```
Error: Unknown contribution type build for contributor jorenham
```

Might be worth checking if there are no typos for this use.

Yeah [act](https://github.com/nektos/act) is something I've used in the past but it takes a while to get set up. Since the workflow isn't that complex you can run the commands locally or in a docker image to ensure isolation.

wolph:
> So there seems to be an unsupported type called "build" that has been added by @jorenham and @wolph. I do not find any support for the type "build" [here](https://allcontributors.org/docs/en/emoji-key). Could you please confirm this and possibly update/let me know what to update it with?

Sorry about that, I thought I could name the categories myself but apparently I was wrong :)
I have no clue what would be the best option to replace it with though, I've never used allcontributors before.
Some possible options:
 - infra: Hosting, Build-Tools, etc. Links to source file (like travis.yml) in repo, if applicable
 - maintenance: People who help in maintaining the repo, links to commits by the user on this project
 - platform: Porting to support a new platform

I think `infra` covers it the best but there's something to be said for all three

fkiraly:
yes, I would also agree that `infra` desribes it best.

Apologies, it escaped my attention in review, this is one of the rarer contribution types.

achieveordie:
So I'd like to take permission from @jorenham and @wolph to change `build` to `infra` since I already needed to make changes to `.all-contributorsrc` to make sure the workflow gets triggered correctly.

wolph:
You have my permission :)

jorenham:
and mine

lmmentel:
Easiest way to test it might be in your on fork since there you would have full control over the CI. I don't get immediately what the issue is with the current version so if you would like some help, please share more details.

achieveordie:
> Easiest way to test it might be in your on fork since there you would have full control over the CI.

The problem is with how that'll differ from the actual use case. It will be trivial to run this workflow in my fork via `pull_request` but things get tricky when the PR is on the actual repo from a fork. That's causing problems with the step that performs a commit.

fkiraly:
@achieveordie, this is still a draft, but it looks like it works?
Or, do we need to test it first in a specific way? If yes, how?

achieveordie:
To reiterate the progress of this PR:
- After writing the initial workflow, there was a "There's still an error" issue which couldn't be resolved due to lack of information. I opened an [issue](https://github.com/all-contributors/cli/issues/346) in the `all-contributors-cli` repo assuming that the source of the error was from there.
- They haven't responded back and this PR went stale. @fkiraly merged main into this PR (possibly to resolve conflicts) yesterday and suddenly that error went away (the implication being that "There's still an error" error could possibly be sporadic).
- Since then, I've corrected the workflow to checkout the correct repository in a non-detached state. The resulting workflow [seems](https://github.com/sktime/sktime/actions/runs/3785791848/jobs/6436118766#step:6:205) to detect changes to `.all-contributorsrc` and make an appropriate commit.

The problem now is that there's an authentication issue as the commit made by the bot couldn't be pushed back to my fork. [According](https://github.com/stefanzweifel/git-auto-commit-action#action-does-not-push-commit-to-repository-authentication-issue) to the author of the commit-bot, it can be resolved by adding a PAT.

@lmmentel @fkiraly What are your suggestions on how to move forward? If I understand correctly, a token will be linked to a single user (since authentication is required for the fork). In this case, even if we merge this PR, every contributor would possibly have to add their own PAT in order to run this workflow file in future.

PS: I'm no expert on authentication and security, we also should think if there's any security vulnerability that we could be introducing here.

achieveordie:
> access tokens should always be secrets, not in the code

Oh, that's not what I meant and I totally agree. If we were to go through this route, we'd have to add another step in the guide for developers, where we ask them to make a `SECRET_TO_WRITE_ON_FORK` token which would be used by this workflow. 

That being said, even I'm against this since this is too much hassle for a workflow file that may be triggered every once in a while (and most likely only once per user). 

> Is there any way we can do this without needing access token of each contributor?

I'm waiting for our in-house expert Lukasz's opinion if he knows some way out already. If not then I'll take another deep dive to try to resolve it in some other way.

fkiraly:
> If not then I'll take another deep dive to try to resolve it in some other way.

Well, there may be other things with higher impact-work ratio (e.g., DL forecasters), but appreciate the sentiment!

fkiraly:
how about ... it triggers only in the *release* workflow, and uses the release manager´s secret?

lmmentel:
Github actions provides a `GITHUB_TOKEN` for each PR that you could try out (see [docs](https://docs.github.com/en/actions/security-guides/automatic-token-authentication)). It should be sufficient to authenticate the commits.

achieveordie:
I can only hypothesize why `GITHUB_TOKEN` isn't working. What's worth noticing though that this secret is already accessed via `github.token`. Here's the quoted text from the doc:

> The token is also available in the `github.token` context. For more information, see "[Contexts](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context)."

This token is accessed by [default](https://github.com/actions/checkout#usage) by the checkout action in the first step.

From my perspective, the reason this isn't working is that the token only has access to the repository that contains the workflow. We could try [increasing](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) the scope of access that this token has, but I don't know whether it includes writing to a fork.

We could possibly bundle this with the release workflow (as suggested by @fkiraly) or I could ask the original author of the action if there's a possible workaround.

lmmentel:
I looked at the error and for some reason you are trying to commit to your fork from an sktime owned workflow which results in a permission error. Please make sure you have everything configure properly and that you are trying to commit to `sktime/sktime` and not your fork.

lmmentel:
Here are the relevant lines from the logs

```
2022-12-28T10:08:40.4636268Z No tagging message supplied. No tag will be added.
2022-12-28T10:08:40.4636827Z INPUT_PUSH_OPTIONS: 
2022-12-28T10:08:40.6826594Z remote: Permission to achieveordie/sktime.git denied to github-actions[bot].
```

achieveordie:
[This](https://github.com/sktime/sktime/actions/runs/3783602998/jobs/6432276265) ran in the context of sktime's repo rather than the fork. It failed as it couldn't find the current branch as it only exists on my fork. I have a question: if we were to run it in the context of the main repo from a PR that has been opened from a fork (like the current one), where will the commits be pushed to?

lmmentel:
Good point @achieveordie, I haven't caught that. An option here might be to change the trigger and run after the PR is merged into `main` that way the commit would be made directly to main, WDYT?

achieveordie:
IMO, the disparity between the development environment (i.e. making sure it works here in this PR) vs the production environment (i.e. making sure it works after merging this PR) is causing all the troubles. Your solution (correct me if I'm wrong) of using an `on-push` trigger will definitely help in that regard. Although, the cost of doing this is that we might have to revert this PR if it doesn't work as intended (and then repeat the same loop in case it fails in the next iteration).

achieveordie:
Just to make sure I understand you correctly, you're suggesting that:
- We only have an `on-push` trigger to the main branch.
- We make sure that workflow points to the `sktime` repo in the checkout action so the commit made by the bot is onto the main rather than a fork.

LMK if you actually meant something else.

lmmentel:
Yeah, it sounds we're on the same page here. With a `on: push` against main we shouldn't have the issues you are experiencing with multiple branches and forks.

lmmentel:
All looks good to me. Did you get a chance to test it with your own fork? That would give us some extra confidence that it won't complain for some odd reason on first run :wink:

achieveordie:
While I could merge the branch that has this workflow (automate_contributors_update) with my fork's main, wouldn't it be better if we run it on sktime's main since that is our target? We have nothing to lose since this doesn't disrupt in case it fails (since it is not a required checker) and we can be assured that there's no difference between the testing and dev environment.

fkiraly (approved):
sounds reasonable to me - shall we try?

When you think it's ready, feel free to merge, and with the next PR to main we'll see whether it works.

RNKuhns:
@achieveordie I'm interested to see if this works. I agree, why not give it a go. If it works, and you are still up for it, I'd appreciate on [skbase](https://github.com/sktime/skbase) too.

achieveordie:
Whoops, it currently fails with the following:
```
[main 46fc2ca] [AUTOMATED] update CONTRIBUTORS.md
 Author: achieveordie <achieveordie@users.noreply.github.com>
 1 file changed, 53 insertions(+), 33 deletions(-)
INPUT_TAGGING_MESSAGE: 
No tagging message supplied. No tag will be added.
INPUT_PUSH_OPTIONS: 
remote: error: GH006: Protected branch update failed for refs/heads/main.        
remote: error: At least 1 approving review is required by reviewers with write access. 17 of 17 required status checks are expected.        
To https://github.com/sktime/sktime
 ! [remote rejected] main -> main (protected branch hook declined)
error: failed to push some refs to 'https://github.com/sktime/sktime'
```

PR #3810, opened 2022-11-17
[ENH] Add TimeSince transformer

KishManani:
This PR implements a transformer, `TimeSince`, which calculates the time elapsed since a reference time. The time elapsed since the start time of a time series is commonly used as a featured to capture trend ([example](https://otexts.com/fpp3/useful-predictors.html)).

- Should the interface be as flexible as it currently is? (i.e., it can accept a lot of different types [int, datetime, period], can convert the time-like result to numeric)
- Understand why an error is raised in the sktime internals when the result contains a dataframe with a column which has type pd.offset.MonthEnd. 

Example usage to cause the error:
```python
import pandas as pd

X = pd.DataFrame(data={"y":[1, 2, 3, 4, 5]}, index=pd.period_range(start="2000-01-01", freq="M", periods=5))
transformer = TimeSince()
Xt = transformer.fit_transform(X)
```
```python
TypeError                                 Traceback (most recent call last)
/Users/kishan_manani/Documents/sktime/Untitled copy 2.ipynb Cell 4 in <cell line: 3>()
      [1](vscode-notebook-cell:/Users/kishan_manani/Documents/sktime/Untitled%20copy%202.ipynb#X62sZmlsZQ%3D%3D?line=0) X = pd.DataFrame(data={"y":[1, 2, 3, 4, 5]}, index=pd.period_range(start="2000-01-01", freq="M", periods=5))
      [2](vscode-notebook-cell:/Users/kishan_manani/Documents/sktime/Untitled%20copy%202.ipynb#X62sZmlsZQ%3D%3D?line=1) transformer = TimeSince()
----> [3](vscode-notebook-cell:/Users/kishan_manani/Documents/sktime/Untitled%20copy%202.ipynb#X62sZmlsZQ%3D%3D?line=2) Xt = transformer.fit_transform(X)

File ~/Documents/sktime/sktime/transformations/base.py:554, in BaseTransformer.fit_transform(self, X, y)
    497 """Fit to data, then transform it.
    498 
    499 Fits the transformer to X and y and returns a transformed version of X.
   (...)
    550         Example: i-th instance of the output is the i-th window running over `X`
    551 """
    552 # Non-optimized default implementation; override when a better
    553 # method is possible for a given algorithm.
--> 554 return self.fit(X, y).transform(X, y)

File ~/Documents/sktime/sktime/transformations/base.py:490, in BaseTransformer.transform(self, X, y)
    488 # convert to output mtype
    489 if not hasattr(self, "_output_convert") or self._output_convert == "auto":
--> 490     X_out = self._convert_output(Xt, metadata=metadata)
    491 else:
    492     X_out = Xt
...
   (...)
   1037     store_behaviour="freeze",
   1038 )

TypeError: 'NoneType' object is not subscriptable
```

I've traced that this is occurring when the mtype is being enforced.

```
System:
executable: /Users/kishan_manani/.pyenv/versions/3.8.7/envs/sktime-dev/bin/python

Python dependencies:
          pip: 22.1.2
   setuptools: 49.2.1
      sklearn: 1.1.1
       sktime: 0.14.0
  statsmodels: 0.13.2
        numpy: 1.21.6
        scipy: 1.8.1
       pandas: 1.5.1
   matplotlib: 3.5.2
       joblib: 1.1.0
        numba: 0.55.2
     pmdarima: 1.8.5
      tsfresh: 0.19.0
```

#### Comments
- The `positive_only` argument facilitates creating the features needed when creating a piece-wise trend using linear regression (see [here](https://otexts.com/fpp3/nonlinear-regression.html)).  The `start` times then become the time points where the knots/change points are defined. 

##### For all contributions


fkiraly:
The errors are currently coming from a syntax error, search for `start_=`.

The other error, do I interpret it correctly as coming from cases where test scenario data index and the start index are not type compatible? (e.g., time delta vs integer)
In which case we may have to turn off the test cases where they are incompatible, simply?

In addition, I've come across the same failure as well - I was working on it separately, introducing a `set_config` like `sktime` has where you can turn off input/output conversions, e.g., inside pipeliens or for diagnostics.

KishManani:
Hi Franz. Apologies, I think you might have picked up a different error as I refactored a static class method to just a private method that sits outside of the class but forgot to remove the reference to self in `transform()`. I've corrected this. Are you now able to reproduce the same error I get in https://github.com/sktime/sktime/pull/3810#issue-1454157491?

fkiraly:
The CI still gets the `start_=` error...

fkiraly:
have you pushed your newest commits?

KishManani:
Hello! Not yet, I still need to investigate the above error.

KishManani:
I think the syntax error is due to using f-strings in a way which is incompatible with older versions of Python. Specifically, using f-strings in this way:
```python
>>> x=5
>>> f"{x=} , did you know x is 5?"
'x=5, did you x is 5?'
```
Let's see if it get's further along the CI/CD pipeline now... 

I've also changed the default value of `to_numeric` to `True` as I think it would be more often used than `False`.

KishManani:
Hi @fkiraly! I need some help with the error being thrown:

```TypeError: _transform output of <class 'sktime.transformations.series.time_since.TimeSince'> does not comply with sktime mtype specifications. See datatypes.MTYPE_REGISTER for mtype specifications. Returned error message: ["obj should not have column of 'object' dtype", "obj must be a pandas.Series, found <class 'pandas.core.frame.DataFrame'>", "obj must be a numpy.ndarray, found <class 'pandas.core.frame.DataFrame'>"].```

The output is a dataframe with a `pd.Period` index and a column of `pd.offset.MonthEnd` objects which has `dtype=object`. Does sktime require that the dtype of columns satisfy a specific criteria?

**Comment by fkiraly** [comment]
yes, currently that's not allowed since most transformers, forecasters, will break if they find a column with exotic column type.

What do you think we should do, @KishManani?
Deactivate the check?
If we do that, it will probably break pipelines we build with this feature.

Also, I'm slightly confused, since according to the docstring, do we not want to return integers?

**Comment by KishManani** [comment]
Originally, I wanted to give the option to return the result as something time-like (e.g., timedelta). That being said, I'm not entirely sure what use case this would support. I'd be happy to remove this option and only return integer outputs. We can always change things if people raise an issue. It would also simplify the transformer and tests. 

An aside, are you happy with the nomenclature I've used here? (e.g., calling the transformer `TimeSince` and the argument names `start`, `positive_only` etc.). I was a little uncertain about whether this was the simplest and clearest naming convention.

**Comment by fkiraly** [comment]
> It would also simplify the transformer and tests.

Oh, I see, it is the custom tests that are failing.

Here's what I suggest: let's `xfail` the loose test and remove the example from the doctest.

I'm currently working on config ability to turn off output checks, someone might find the timedelta/object output uesful.

PS: would appreciate if there are two parameter sets returned by `get_test_params`.

**Comment by KishManani** [comment]
Hi Franz! One unit test in `test_time_since.py` is failing because of the error thrown in my example in the original comment https://github.com/sktime/sktime/pull/3810#issue-1454157491. The transformer's `_transform` method returns the result as expected and would pass the unit test. However, when using `transform` the `TypeError` is raised which causes the unit test to fail.

> Here's what I suggest: let's xfail the loose test and remove the example from the doctest.

Given the above, is this still what you want to do? I'm less familiar with the sktime internals for testing. Could you point me in the right direction? Thank you! 🙏 

> PS: would appreciate if there are two parameter sets returned by get_test_params.

Sure! I found it a little bit tricky to know what is expected here and how it's used in the sktime internals. Could you point me to anything I can read to learn more about what's going on with `get_test_params`? Thanks again!

**Comment by fkiraly** [comment]
> Sure! I found it a little bit tricky to know what is expected here and how it's used in the sktime internals. Could you point me to anything I can read to learn more about what's going on with get_test_params? Thanks again!

Of course. High-level, `get_test_params` returns parameter sets which are used to create test instances to run through various test scenarios.

They are the objects used inside the tests, e.g., see the general "estimator writing" guide on `check_estimator`
https://www.sktime.org/en/stable/developer_guide/add_estimators.html

**Comment by fkiraly** [comment]
> The transformer's _transform method returns the result as expected and would pass the unit test. However, when using the transform method the TypeError is raised which causes the unit test to fail.

Not sure what you mean? The output type check (and conversion) happens in `transform`, on the output of `_transform`, so that´s where one would expect it to fail?

**Comment by fkiraly** [comment]
> Given the above, is this still what you want to do? I'm less familiar with the sktime internals for testing.

Hm, I still think it's a good idea, but I haven't fully understood what you mean about `_transform` vs `transform`.

**Comment by KishManani** [comment]
> Not sure what you mean? The output type check (and conversion) happens in transform, on the output of _transform, so that´s where one would expect it to fail?

Hi Franz! I think we're on the same page. I'm just highlighting that the unit test is failing because of the additional checks done by the sktime internals rather than the underlying implementation in `_transform` being erroneous.

**Comment by fkiraly** [comment]
> Hi Franz! I think we're on the same page. I'm just highlighting that the unit test is failing because of the additional checks done by the sktime internals rather than the underlying implementation in `_transform` being erroneous.

Well, "erroneous" is a matter of definition depending on the interface specification, e.g., what column types are allowed in the return.

You could turn off the output conversion in the test via `_output_convert`, and leave this hack in until we have arrived at the config interface.

**Comment by fkiraly** [comment]
This test failure seems like a "real" one ...

```
FAILED sktime/transformations/series/tests/test_time_since.py::test_fit_transform_datetime_15mins_idx_numeric_output - AssertionError: DataFrame.iloc[:, 0] (column name="time_since_2000-01-01 00:00:00") are different

DataFrame.iloc[:, 0] (column name="time_since_2000-01-01 00:00:00") values are different (80.0 %)
[index]: [2000-01-01T00:00:00.000000000, 2000-01-01T00:15:00.000000000, 2000-01-01T00:30:00.000000000, 2000-01-01T00:45:00.000000000, 2000-01-01T01:00:00.000000000]
[left]:  [0, 225, 450, 675, 900]
[right]: [0, 15, 30, 45, 60]
```

**Comment by KishManani** [comment]
> This test failure seems like a "real" one ...

This is due to a bug in older versions of pandas. In older version of pandas computing the time difference when the frequency is specified like "15T" (15 minutes) gave incorrect results (i.e., the result that you see as the output in the error message from the test).

**Comment by fkiraly** [comment]
> This is due to a bug in older versions of Pandas. In older version of Pandas computing the time difference when the frequency is specified like "15T" (15 minutes) gave incorrect results (i.e., the result that you see as the output in the error message from the test).

Argh yes, I remember. We have a workaround for this somehow to ensure compatibility on lower versions. Let me try to find it.

**Comment by fkiraly** [comment]
Argh, can't find it.

I vaguely remember that it was @khrapovs who encountered this, can that be?

Otherwise, you could do a version check and conditional workaround for the `freq` argument.

For an example version check, look in `_check_python_version`, line 221

**Comment by KishManani** [comment]
> Argh, can't find it.
> 
> I vaguely remember that it was @khrapovs who encountered this, can that be?
> 
> Otherwise, you could do a version check and conditional workaround for the `freq` difference if the version is 3.7.
> 
> For an example version check, look in `_check_python_version`, line 221

Hey Franz! I found the line mentioned. I've got one doubt, given that the error is being raised for a particular version of Pandas, is checking the version of Python the right thing to do here? You would still get the error on later versions of Python if you're still using certain versions of Pandas.

**Comment by fkiraly** [comment]
> You would still get the error on later versions of Python if you're still using certain versions of Pandas.

Yes, I think the version check should be for `pandas`, not `python`. My bad.

My suggestion would indeed be 1) - that is:

1. check `pandas` version condition
2. if "bad" condition, do a manual workaround to compute the right time difference

I vaguely remember @khrapovs had the same problem somewhere and we did the same, so there might be code somewhere if you have a good keyword that finds checks of pandas version.

**Comment by fkiraly** [comment]
there's actually an easier way if this proves too troublesome: set the `python_version` tag to >3.7, an example of this tag is in `SignatureClassifier`. This prevents users from using the estimator on 3.7 , but at least it provides an informative error message (and will pass the test because `sktime` knows not to test it on 3.7)

**Comment by KishManani** [comment]
Hi @fkiraly! I managed to find a workaround that was compatible with older versions of Pandas. It seems to work on the use cases I've tested... 🤞. The tests are all passing now! 🙌

**Comment by fkiraly** [comment]
great! anyone still using 3.7 can now profusely thank @KishManani

**Review by fkiraly — APPROVED** [review]
Looks good, would be happy to approve!

**Review by GuzalBulatova — APPROVED** [review]
Looks good to go! There’s a weird “This codeownes  file contains errors” highlight for me though, and I don’t understand the reason for that, e.g.

```
Unknown owner on line 11: make sure @jasonlines exists and has write access to the repository
…lassification/distance_based/ @jasonlines @goaster @TonyBagnall 
```
 and so on, many more handles
seems like a GitHub error? 🤔

fkiraly:
> and so on, many more handles
seems like a GitHub error?

It's related to a discussion we had earlier on the "algorithm maintainer" role, which afaik so far remains unaddressed with other discussions of governance.

The issue is, for codeowners to "work" and automatically request a review on anything that changes owned code, the code owner needs to have write access to the repo. Otherwise it will be picked up as an error, like this.

A more granular assignment of ownership is possible with separate repos or containers in a market place, e.g., as huggingface is doing it, but that comes with a substantial admin or tech overhead.

fkiraly:
Currently, for non-core-devs, we use the codeowners file as an "as if", i.e., we use it as a basis for manual action in informing the owner.

PR #3812, opened 2022-11-18
[BUG] exclude TapNet from tests

TonyBagnall:
Tests are failing for strange reasons, see for example
https://github.com/sktime/sktime/actions/runs/3492869189/jobs/5847895424
I cannot reproduce locally. TapNet was excluded but recently put back in. This PR is just to test to see if it is the root of the problem

### DIFF
--- a/sktime/tests/_config.py
+++ b/sktime/tests/_config.py
@@ -29,6 +29,9 @@ EXCLUDE_ESTIMATORS = [
     "RandomIntervalClassifier",
     "MiniRocket",
     "MatrixProfileTransformer",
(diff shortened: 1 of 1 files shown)


MatthewMiddlehurst (approved):

fkiraly (changes requested):
Please add an issue number to the comment before we merge this

MatthewMiddlehurst (approved):

achieveordie:
Couldn't find any errors related to TapNet in the job, will follow this issue though

PR #3820, opened 2022-11-19
[BUG] Update `test_deep_estimator_full` to incorporate new versions of tf/keras

### LINKED ISSUE #3816: [BUG] testing deep learning base class fails due to optimizer type
**Describe the bug**

 test_deep_estimator_full[keras-adamax] is failing due to a type check in keras. See, for example

https://github.com/sktime/sktime/actions/runs/3495913658/jobs/5855642783

"ValueError: `optimizer` of type <class 'keras.optimizers.optimizer_v2.adamax.Adamax'> cannot be serialized, it should either be absent/None/str/tf.keras.optimizers.Optimizer object"

it could be that the wrong type of Optimizer is being used, output states

Optimizer  = <class 'keras.optimizers.optimizer_experimental.optimizer.Optimizer'>
whereas it is expecting
tf.keras.optimizers.Optimizer

image

I don't want to hack around in the testing suite, I'll let someone who understands it better do that,

achieveordie:
#### Reference Issues/PRs

Fixes #3816

Replacing `keras.optimizers` with `tf.keras.optimizers` along with few other API-related changes. Also reformatted `save` to check for model presence first, must have gone unnoticed in the original PR.

No

#### Any other comments?

Tested for the latest tensorflow version (2.11.0)

##### For all contributions

### DIFF
--- a/sktime/classification/deep_learning/base.py
+++ b/sktime/classification/deep_learning/base.py
@@ -245,7 +245,7 @@ class BaseDeepClassifier(BaseClassifier, ABC):
             import h5py
 
             in_memory_model = None
-            if self.model is not None:
+            if self.model_ is not None:
                 with h5py.File(
                     "disk_less", "w", driver="core", backing_store=False
                 ) as h5file:
@@ -363,7 +363,7 @@ class BaseDeepClassifier(BaseClassifier, ABC):
                 zip_file.extract(file, temp_unzip_loc)
 
         keras_location = temp_unzip_loc / "keras"
-        if keras_location.exist():
+        if keras_location.exists():
             cls.model_ = keras.models.load_model(keras_location)
         else:
             cls.model_ = None
--- a/sktime/classification/deep_learning/base.py
+++ b/sktime/classification/deep_learning/base.py
@@ -241,6 +241,9 @@ class BaseDeepClassifier(BaseClassifier, ABC):
         from pathlib import Path
         from zipfile import ZipFile
 
+        if self.model_ is None:
+            raise NotFittedError("Model not built yet, call it via `.fit()`")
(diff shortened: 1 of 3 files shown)


fkiraly (changes requested):
The fix to the test looks good.

Regarding the change to the `save` method - should that not also simply work with an unfitted object? I thought there's no harm in that, there may be an use case.

achieveordie:
We're testing for the presence of the model (`self.model_ is not None` should evaluate to `True`) as per the current implementation. This, in my opinion, is implicit in the documentation we have built during serialization (by mentioning that `keras/` subdirectory is present in the zipped file).

I also have trouble thinking of a use case for saving an estimator that lacks the most important element. Perhaps I've probably used the wrong kind of error (`NotFittedError`) to give you a false impression (I probably did that since there was no `NotBuiltError`?).

In any case, I'm making a commit that actually lets you save the estimator even when `self.model_ = None` for DL Classifiers. Let's discuss if this is actually what we want so we can change the documentation/other parts of the code.

achieveordie:
I don't fully understand why the current failure (in 3.10, macOS-11) is occurring. I also highly doubt this failure has anything to do with the current PR's work since the problem is setting up `sktime`.

One possibly relevant piece of information is that `grpcio` is downloaded directly as a wheel in Windows whereas it is downloaded as a `.tar.gz` file for Mac and it is during the building stage of this package where the error is occurring.

fkiraly (approved):
Seems to work now, thanks!

The failure seems to have been sporadic.

As said, I would remove the raised exception in `save` and ensure it works for unfitted models, but it is not a blocker since that behaviour is already on `main`.

PR #3822, opened 2022-11-20
[ENH] Estimator config interface

fkiraly:
This PR adds a config management capabilities to all `sktime` objects, similar to `sklearn`'s new config interface, via two methods `get_config` and `set_config`.

It is based on the flag manager mixin introduced in https://github.com/sktime/sktime/pull/3630.

A STEP is available here: https://github.com/sktime/enhancement-proposals/pull/29

Includes:
* test for `get_config` and `set_config` interface
* proof-fo-concept functionality for transformers to control input/output checks and conversions
* refactor of existing private `_output_convert` config to the new config interface


fkiraly:
FYI @RNKuhns, this is what I had in mind - will also write a STEP

fkiraly:
FYI @RNKuhns, this is how it would look like at the end.

FYI @KishManani, I forgot we already had some internal functionality to turn off output conversions, this is useful for diagnostics. You can set the `_output_convert` attribute to `"off"` and the output conversion will not happen.

This is currently undocumented and not a public interface point, but if you need it for testing or diagnosis.

fkiraly:
FYI @RNKuhns, I wrote a STEP here:
https://github.com/sktime/enhancement-proposals/pull/29

fkiraly:
as proof of concept: global configs can be added like this:
https://github.com/sktime/sktime/pull/4268

Issue #3824, opened 2022-11-21
[ENH] Allow `plot_series()` to plot prediction intervals


chillerobscuro:
**Is your feature request related to a problem? Please describe.**
Making interval forecasts is a key feature of `sktime`, seems like there should be a simple way to plot the intervals. 

**Describe the solution you'd like**
Pass the results of `forecaster.predict_interval()` or `forecaster.predict_quantiles()`  directly to `utils.plotting.plot_series()`. This would entail adding an argument like `intervals_to_plot`. 

**Describe alternatives you've considered**
Plot by hand, like in the [forecasting example](https://www.sktime.org/en/stable/examples/01_forecasting.html). The code is as follows:
```
from sktime.utils import plotting
y_pred = forecaster.predict()
y_pred_ints = forecaster.predict_interval(coverage=.95)

fig, ax = plotting.plot_series(y, y_pred, labels=["y", "y_pred"])
ax.fill_between(
    ax.get_lines()[-1].get_xdata(),
    y_pred_ints["Coverage"][coverage]["lower"],
    y_pred_ints["Coverage"][coverage]["upper"],
    alpha=0.2,
    color=ax.get_lines()[-1].get_c(),
    label=f"{coverage}% prediction intervals",
)
ax.legend();
```
My working solution involves adding the above code to a new function `plot_interval()` in `sktime/utils/plotting.py`, which is called on `ax` if `plot_series()` receives a pandas DataFrame including either `"Coverage"` or `"Quantile"` nested keys.

**Questions**
 - How to check if the interval is included and find its value in the predicted_interval df? It could be nested under `y_pred_ints["Coverage"]` but if the df was created by `forecaster.predict_quantiles()` it may be under `y_pred_ints["Quantiles"]`. Are there any other schemas that need to be checked for/accepted? We could allow an additional descriptive argument (ie `coverage=.95`) to `plot_series()` but I think it's cleaner to only add one new argument to that function, and raise an error if we can't infer the value from what is provided.
 - How to handle the case where multiple series and/or more than one pair of upper/lower intervals are passed?

aiwalter:
related issue: https://github.com/sktime/sktime/issues/541

fkiraly:
yes, duplicate of #541.

> receives a pandas DataFrame including either "Coverage" or "Quantile" nested keys.

I think we need to be careful, in case you did not intend this: the top level is called "Coverage" or "Quantile" only in the univariate case, otherwise it is the name of the variable(s).

[...]

PR #802, opened 2021-04-12
Plotting the prediction interval

Ifeanyi30:
#### Reference Issues/PRs

fixes the issue #541

#### What does this implement/fix? Explain your changes.

Adds the pred_int option to the plot_series function
#### Does your contribution introduce a new dependency? If yes, which one?
No

#### What should a reviewer concentrate their feedback on?
Focus on the plot_series function in the plotting.py file

##### For all contributions
- [ ] I've added myself to the [list of contributors](https://github.com/alan-turing-institute/sktime/blob/main/.all-contributorsrc).
- [ ] Optionally, I've updated sktime's [CODEOWNERS](https://github.com/alan-turing-institute/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
- [ ] I've added unit tests and made sure they pass locally.

##### For new estimators
- [ ] I've added the estimator to the online documentation.
- [ ] I've updated the existing example notebooks or provided a new one to showcase how my estimator works.

PR #3825, opened 2022-11-22
[ENH] Add support for plotting intervals in `plot_series()`

chillerobscuro:
#### Reference Issues/PRs

This would close [Issue-3824](https://github.com/sktime/sktime/issues/3824) as well as [Issue-541](https://github.com/sktime/sktime/issues/541). There have been multiple other PRs for the original issue but as far as I can see they're all closed ([545](https://github.com/sktime/sktime/pull/545), [802](https://github.com/sktime/sktime/pull/802)).

Allows user to pass `pred_interval` to `plot_series()` and plot the confidence interval provided in that argument.

No

Any other edge cases I need to check for? So far there is pretty rigorous check of the schema of the interval DataFrame, and there's a check to ensure the index matches that of the final series passed to the function.

Once there's some buy in for this approach I'll add at least one test case.

I'm not sure what killed the other attempts at this feature, it's been discussed for at least two years. I think we start simple and then expand to the `predict_quantiles()`, multivariate, plotting kwargs, and other more complex cases after.

### DIFF
--- a/sktime/utils/validation/forecasting.py
+++ b/sktime/utils/validation/forecasting.py
@@ -468,10 +468,12 @@ def check_interval_df(interval_df):
     if not isinstance(interval_df, pd.DataFrame):
         raise TypeError("`interval_df` must be instance type pd.DataFrame")
     if interval_df.empty:
-        raise TypeError("`interval_df` is empty, can not plot intervals")
+        raise ValueError("`interval_df` is empty, can not plot intervals")
     levels = interval_df.columns.levels
     if len(levels[0]) != 1:
         raise ValueError("`interval_df` must only contain one variable with interval")
+    if not (levels[0] == "Coverage")[0]:
+        raise ValueError("`interval_df` must have 'Coverage' column label")
     if len(levels) != 3:
         raise ValueError("`interval_df` must have 3 column levels")
     cov_level = levels[1][0]
--- a/sktime/utils/validation/forecasting.py
+++ b/sktime/utils/validation/forecasting.py
@@ -457,18 +457,22 @@ def check_regressor(regressor=None, random_state=None):
     return regressor
 
 
-def check_interval_df(interval_df):
+def check_interval_df(interval_df, index_to_match):
     """
     Verify that a predicted interval DataFrame is formatted correctly.
 
     Parameters
     ----------
     interval_df : pandas DataFrame outputted from forecaster.predict_interval()
+    index_to_match : Index object that must match interval_df.index
     """
     if not isinstance(interval_df, pd.DataFrame):
-        raise TypeError("`interval_df` must be instance type pd.DataFrame")
+        raise TypeError("`interval_df` must be of type pd.DataFrame")
     if interval_df.empty:
         raise ValueError("`interval_df` is empty, can not plot intervals")
+    df_idx = interval_df.index
+    if len(index_to_match) != len(df_idx) or not (index_to_match == df_idx).all():
+        raise ValueError("Prediction interval index must match the final Series index.")
     levels = interval_df.columns.levels
     if len(levels[0]) != 1:
         raise ValueError("`interval_df` must only contain one variable with interval")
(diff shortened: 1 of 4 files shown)


chillerobscuro:
Almost all of the pieces were there but with this PR I think `sktime` has the most impressive <10 lines of code in open source:
```
from sktime.utils import plotting

y = load_airline()
forecaster = AutoARIMA(sp=12)

forecaster.fit(y, fh=list(range(1, 13)))
y_pred = forecaster.predict()
pred_int = forecaster.predict_interval(coverage=.95)
plotting.plot_series(y, y_pred, labels=["y", 'y_pred'], pred_interval=pred_int)
```
image

fkiraly:
> Almost all of the pieces were there but with this PR I think sktime has the most impressive <10 lines of code in open source:

You're too kind 😊 

Let's see whether we can improve on that with the pipelines tutorial at pydata global 2022 :-)

fkiraly:
> And IMO this (with image) should be the marquee code block for the docs, either on the home or "get started" page.

Hm, you're probably right - maybe with a pipeline creation step or so?

sktime/utils/validation/forecasting.py
```diff
@@ -455,3 +455,35 @@ def check_regressor(regressor=None, random_state=None):
             )
         regressor = clone(regressor)
     return regressor
+
+
+def check_interval_df(interval_df, index_to_match):
```
fkiraly, line 460: some of this is redundant with

```python

check_is_mtype(interval_df, "pred_interval")
```

(I think all except the "index to match" check)

fkiraly (dismissed):
Makes sense as an improvement.

Regarding the checkers, I left a non-blocking comment.

In addition, I just looked at the checker at the top - it seems to call `check_y` which enforces univariate inputs, and may also be replaceable by `check_is_scitype`, but I would not consider the checks part of this PR's scope.

fkiraly:
> I'm not sure what killed the other attempts at this feature, it's been discussed for at least two years. 

Maybe just bad luck?

I can imagine that it was also the lack of a clean design for prediction interval return objects that has been introduced in the beginning of 2022 and was then presented at pydata Berlin 2022 (and/or good documentation around the probabilistic prediction interface which was also added around that time).

chillerobscuro:
> > And IMO this (with image) should be the marquee code block for the docs, either on the home or "get started" page.
> 
> Hm, you're probably right - maybe with a pipeline creation step or so?

I think since any sequence of steps is conceptually a "pipeline" that could be foregone for the "marquee" and expanded upon later, for the sake of conciseness.

sktime/utils/validation/forecasting.py
chillerobscuro: I will check that out and replace these lines, good tip

chillerobscuro: After looking into it, I am leaning towards leaving the checks as is. I would typically want to reuse `check` style code but the existing code raises very informative errors and the function needs to exist regardless, at least to check the column name (used in the plotting function) and the matching index as you mentioned. Is that acceptable?

fkiraly: sure, but that would mean you don´t find `check_is_mtype` informative of the same degree, which should be improved then, separately.

chillerobscuro: I was not fully aware of the usefulness of `return_metadata=True`, now that I see it I've added it in. I kept checks for index matching, exact column name as used in the plot function ("Coverage"), and that there is only one pair of interval series. Should be good to go if this passes tests.

FYI you mentioned a "non blocking" comment but it appears this PR can't be completed until all comments are resolved. It's very possible I misunderstood what you meant though.

fkiraly: in the end, a core developer needs to merge it either way

fkiraly (approved):

PR #3828, opened 2022-11-22
Added MultiIndex error message in _fit_forecaster method

aadya940:


fkiraly (changes requested):
Thanks for this!

Can you kindly describe what improvement or issue this is addressing?

I do think there might not be a situation where `MultiIndex` is being seen in `y`, as the base class, `BaseForecaster`, should take care of that?

Or, do you have a code example of the situation this is helpful in?

PR #3843, opened 2022-11-27
[MNT] 0.18.0 change action - `numba` as soft dependency

fkiraly:
This PR turns `numba` into a soft dependency, as scheduled for 0.18.0.

Relies on https://github.com/sktime/sktime/pull/3631 which isolates `numba`, which was merged pre-0.17.0 to prepare the change.


miraep8:
So, I have a question (a sort of a half-baked question): 

Is there any way that we can move the ability to speed up via `numba` outside of the estimator itself? I agree with the point folks have made that it is a great advantage for certain algorithms - but I am still concerned by the edge cases where certain users are not able to install `sktime` or certain parts of `sktime` because of issues with `numba`. It seems this solution, while no longer requiring that `numba` be a hard dependency still requires that some estimators either use `numba`, or not, there is no user choice.

I guess what I am wondering is - is there any way we can move the choice of whether or not to use `numba` to the user end? So for example they can somehow wrap their call to fit in an njit function? To be honest - I am not entirely sure this is possible/what it would look like - I just thought I would ask/throw the idea out there as a possible third solution.

fkiraly:
> I guess what I am wondering is - is there any way we can move the choice of whether or not to use `numba` to the user end? So for example they can somehow wrap their call to fit in an njit function? To be honest - I am not entirely sure this is possible/what it would look like - I just thought I would ask/throw the idea out there as a possible third solution. 

Hm, I think that's an excellent question.

Is there such a thing as conditional decoration with `numba`? The problem is, that `numba` compiles nested calls, so a conditional decorator might break the compilation of nested dependencies.

Of course you could do that manually with a hand-crafted dispatcher between two private modules, one has numba and the other one has all the `njit`-s removed etc, but that does not seem very elegant.

patrickzib:
> Is there any way that we can move the ability to speed up via numba outside of the estimator itself? 

No, numba was not introduced for fun but as the functionality provided is not usable without it. A speedup of 50-100x means that you can train a model or not at all in any reasonable time. 

A full run of the UCR archive using numba and multi-threading takes 30-100 minutes with SotA classifiers like Rocket, Hydra, MultiRocket, MiniRocket, etc. Removing `numba`, this will equate to 50 hours to 416 hours = 17 days!

patrickzib:
The whole UCR archive contains mostly toy examples and equates to **800MB in total.** This is small in an industry scale. 

I have a dataset on satellite image time series that has **3GB alone**. No one wants to use a super computer to train this dataset. Either you are able to install numba, or you will not be able to process this data.

MatthewMiddlehurst:
There is an environmental variable and config option which allows the disabling of Jit functions (https://numba.pydata.org/numba-doc/dev/reference/envvars.html#envvar-NUMBA_DISABLE_JIT), but I would not recommend using it.

patrickzib:
> There is an environmental variable and config option which allows the disabling of Jit functions (https://numba.pydata.org/numba-doc/dev/reference/envvars.html#envvar-NUMBA_DISABLE_JIT), but I would not recommend using it.

Yet, you must install numba to set the flag?

MatthewMiddlehurst:
The only numba issues I have seen which prevent the use of other functionality are related to the writing/reading of compiled functions, not the actual installation itself (you can also change where these functions are written to, FYI). If there are other issues which deal with the installation then sure this won't work.

patrickzib:
> The only numba issues I have seen which prevent the use of other functionality are related to the writing/reading of compiled functions, not the actual installation itself (you can also change where these functions are written to, FYI). If there are other issues which deal with the installation then sure this won't work.

True, and the question was wether sktime estimators using numba could run without numba being installed. Thus, the environment flag is not the solution.

patrickzib:
> We should try to provide a numba implementation for estimators

If the suggestion is to have two code bases: I believe that will be a nightmare to maintain and fix bugs.

Also please note that `numba` allows/enforces writing c-like code using e.g. for-loops. This however is hopelessly slow in plain Python. Just check the Rocket-classes to see this. Thus, this non-numba code will involve major rewrites of all codes. And there again we are at the nightmare to maintain.

fkiraly:
@RNKuhns, @patrickzib, the architecture I have in mind for the refactor is much simpler.

We use the estimator level dependency isolation through the base class.

This requires minimal changes to code structure, and an estimator can be `numba` dependent or not, while the package layer is not `numba` dependent.

This is how it looks like:
https://github.com/sktime/sktime/pull/3631

It is precisely how the current `main` is for `statsmodels`. We could remove `statsmodels` as a soft dep and nothing would break, except the estimators that explicitly depend on it as is inspectable through the `python_dependencies` tag.

fkiraly:
FYI @RNKuhns, this is via the `python_dependencies` tag, the logic is mediated through `BaseObject`, so dependency isolation it is in principle an `skbase` concern.

fkiraly:
> Also, as @MatthewMiddlehurst mentioned, I am currently aware of a single user with numba-related problems.

@patrickzib, the informative KPI here is not the total number but the fraction of users outside your immediate bubble. What is the denominator. Also 1?

fkiraly:
> Thus, this non-numba code will involve major rewrites of all codes using vectorization, etc. And there again we are at the nightmare to maintain.

I don't think making `numba` a soft dependency would involve any of that.

patrickzib:
> > Also, as @MatthewMiddlehurst mentioned, I am currently aware of a single user with numba-related problems.
> 
> @patrickzib, the informative KPI here is not the total number but the fraction of users outside your immediate bubble. What is the denominator. Also 1?

@fkiraly So, in your bubble, how many users are you aware of? Please name the concrete users. Everything else is just an urban legend

patrickzib:
> > Thus, this non-numba code will involve major rewrites of all codes using vectorization, etc. And there again we are at the nightmare to maintain.
> 
> I don't think making `numba` a soft dependency would involve any of that.

I did not ever claim that. You are citing me out of the given context. This statement is in context of providing both numba and non-numba code as @RNKuhns implied.

patrickzib:
> Happy to do that, but I don't think this is how evidence based argumentation goes.

This is a joke - and your argumentation is a joke. You were claiming from the start that there are so many users with numba-related problems, thus you pursue this personal quest to remove numba. So the burden is on you to prove this! So far, you have provided NO evidence for this but just this single user I am not pushing this PR so there is no burden on me. 

Anyhow: I am using sktime for 2 years in my lectures now with more than 200 students that used it by now. NONE of them had numba related problems but all of them faced speed issues! 200 students is more than enough evidence. You have none for your claims.

Also I want to clarify:

- The majority of core developers was against this PR. You are moving on anyways. Thus, this is a personal quest of @fkiraly and you should not use anonymous users as a justification.
- You claimed this could be done minimally invasive. So far you touched **87 files with >8000 lines of code** changed. You split estimators in two, and add lots of wrapper code.
- The number of files and lines shows that: sktime - numba == forecasting? It is not usable without! It simply does not make sense to remove a library that is used so intensively within the project.


I call for a vote again on this concrete PR. You had a vote on the soft-dependence, which you ignored. Now I am calling for a vote on this concrete realization of yours.

TonyBagnall:
I am against this PR and would formally like to express that, so it cannot get merged under the five day lazy consensus rule. I realise it is draft now, but I may not notice if it becomes ready for review as I am otherwise engaged for the next couple of weeks, hence the post now.

fkiraly:
> This is a joke - and your argumentation is a joke. You were claiming from the start that there are so many users with numba-related problems, thus you pursue this personal quest to remove numba:

@patrickzib, your tone and implications are inappropriate. Please remain civil in tone.

fkiraly:
> you had a vote on the soft-dependence, lost, and which you ignored.

It was not ignored. It went to [stage 3 of confluct resolution according to the decision making process as outlined in the governance document](https://www.sktime.org/en/stable/get_involved/governance.html#decision-making) and was discussed in the [November 8 CC meeting](https://github.com/sktime/community-org/blob/main/community_council/previous_meetings/20221108-meeting.md), see agenda point 6 in the minutes.#

Please also keep your tone civil, and your statements factual.

fkiraly:
@patrickzib, kindly refer to the original issue:
https://github.com/sktime/sktime/issues/3567

I have updated the description and arguments for.

I would appreciate discussion and exchange of arguments on there, and not all over the place on the various PR.

patrickzib:
> @patrickzib, kindly refer to the original issue: #3567
> 
> I have updated the description and arguments for.
> 
> I would appreciate discussion and exchange of arguments on there, and not all over the place on the various PR.

I do not see any point in discussing this with you in this PR. Provably, you do not change your mind on anything. Thus, this is a waste of time for us both. It is clear what the majority of other core developers think on this: still you do no care, nor will you listen to them. I wish you really ever spoke to people being relally involved in the classification community like Eamonn Keogh, Webb or Tony. But again you do not care.

fkiraly:
@patrickzib, tone

patrickzib:
I am formally voting this down for missing deprecation warnings. 

My code runs fine under sktime <= = 0.141 on gogole colab, but is broken with this branch due to a mising `numba` libarary.  You must deprecate all affected files first.

These may then go in with sktime 0.16, if acccepted.

PR #3847, opened 2022-11-28
[ENH] _fit_transform

### LINKED ISSUE #3823: [BUG] `SupervisedIntervals` is not interface compliant
`SupervisedIntervals` is not interface compliant, as it overrides `fit_transform`.

It will start failing after the next core interface change, kindly fix this, @MatthewMiddlehurst.

Apologies for not spotting this in the review.

MatthewMiddlehurst:
#### Reference Issues/PRs

Fixes #3823 

Implements a `_fit_transform` method in `BaseTransformer`. `fit_transform` now calls the private method, and contains a set of checks merged from `fit` and `transform`.

This PR also replaces the current `fit_transform` overrides in `SupervisedIntervals` and `TSFreshRelevantFeatureExtractor` with the private method.

Alongside allowing for easier implementation of efficient `fit_transform` methods, the change reduces the repetition of some checks when the method is called.

#### Any other comments?

Happy to make changes or add additional checks if people think they are necessary. As it is, this is essentially a merge of the current `fit` and `transform` methods with some repetition removed.

### DIFF
--- a/sktime/transformations/base.py
+++ b/sktime/transformations/base.py
@@ -549,9 +549,42 @@ class BaseTransformer(BaseEstimator):
                 then the return is a `Panel` object of type `pd-multiindex`
                 Example: i-th instance of the output is the i-th window running over `X`
         """
-        # Non-optimized default implementation; override when a better
-        # method is possible for a given algorithm.
-        return self.fit(X, y).transform(X, y)
+        # if fit is called, estimator is reset, including fitted state
+        self.reset()
+
+        # if requires_y is set, y is required in fit and update
+        if self.get_tag("requires_y") and y is None:
+            raise ValueError(f"{self.__class__.__name__} requires `y` in `fit`.")
+
+        # check and convert X/y
+        X_inner, y_inner, metadata = self._check_X_y(X=X, y=y, return_metadata=True)
+
+        # memorize X as self._X, if remember_data tag is set to True
+        if self.get_tag("remember_data", False):
+            self._X = update_data(None, X_new=X_inner)
+
+        # checks and conversions complete, pass to inner fit_transform
+        #####################################################
+
+        vectorization_needed = isinstance(X_inner, VectorizedDF)
+        self._is_vectorized = vectorization_needed
+
+        # we call the ordinary _fit_transform if no looping/vectorization needed
+        if not vectorization_needed:
+            Xt = self._fit_transform(X=X_inner, y=y_inner)
+        else:
+            # otherwise we call the vectorized version of fit_transform
+            Xt = self._vectorize("fit_transform", X=X_inner, y=y_inner)
+
+        self._is_fitted = True
+
+        # convert to output mtype
+        if not hasattr(self, "_output_convert") or self._output_convert == "auto":
+            X_out = self._convert_output(Xt, metadata=metadata)
+        else:
+            X_out = Xt
+
+        return X_out
 
     def inverse_transform(self, X, y=None):
         """Inverse transform X and return an inverse transformed version.
@@ -1156,6 +1189,25 @@ class BaseTransformer(BaseEstimator):
 
             return Xt
 
+        if methodname == "fit_transform":
+            X, _, Xs, ys, n, row_idx, col_idx = unwrap(kwargs)
+
+            # create container of transformers
+            self.transformers_ = pd.DataFrame(index=row_idx, columns=col_idx)
+            for ix in range(n):
+                i, j = X.get_iloc_indexer(ix)
+                self.transformers_.iloc[i].iloc[j] = self.clone()
+
+            # fit/update the ix-th transformer with the ix-th series/panel
+            Xts = []
+            for ix in range(n):
+                i, j = X.get_iloc_indexer(ix)
+                method = getattr(self.transformers_.iloc[i].iloc[j], methodname)
+                Xts += [method(X=Xs[ix], y=ys[i], **kwargs)]
+            Xt = X.reconstruct(Xts, overwrite_index=False)
+
+            return Xt
+
     def _fit(self, X, y=None):
         """Fit transformer to X and y.
 
@@ -1209,6 +1261,29 @@ class BaseTransformer(BaseEstimator):
         See extension_templates/transformer.py for implementation details.
         """
         raise NotImplementedError("abstract method")
+
+    def _fit_transform(self, X, y=None):
+        """Fit transformer to X and y.
+
+        private _fit_transform containing the core logic, called from fit_transform
+
+        Parameters
+        ----------
+        X : Series or Panel of mtype X_inner_mtype
+            if X_inner_mtype is list, _fit must support all types in it
+            Data to fit transform to
+        y : Series or Panel of mtype y_inner_mtype, default=None
+            Additional data, e.g., labels for tarnsformation
+
+        Returns
+        -------
+        self: a fitted instance of the estimator
+
+        See extension_templates/transformer.py for implementation details.
+        """
+        # Non-optimized default implementation; override when a better
+        # method is possible for a given algorithm.
+        return self._fit(X, y)._transform(X, y)
 
     def _inverse_transform(self, X, y=None):
         """Inverse transform X and return an inverse transformed version.
--- a/sktime/transformations/panel/tsfresh.py
+++ b/sktime/transformations/panel/tsfresh.py
@@ -560,7 +560,7 @@ class TSFreshRelevantFeatureExtractor(_TSFreshFeatureExtractor):
 
         return selection_params
 
-    def fit_transform(self, X, y=None):
+    def _fit_transform(self, X, y=None):
         """Fit to data, then transform it.
 
         Fits the transformer to X and y and returns a transformed version of X.
@@ -616,9 +616,6 @@ class TSFreshRelevantFeatureExtractor(_TSFreshFeatureExtractor):
                 then the return is a `Panel` object of type `pd-multiindex`
                 Example: i-th instance of the output is the i-th window running over `X`
         """
-        self.reset()
-        if y is None:
-            raise ValueError("SupervisedIntervals requires `y` in `fit`.")
         X, y, metadata = self._check_X_y(X=X, y=y, return_metadata=True)
 
         # lazy imports to avoid hard dependency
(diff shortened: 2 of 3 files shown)


sktime/classification/dictionary_based/_tde.py
```diff
@@ -825,7 +825,8 @@ def _fit(self, X, y):
                     n_jobs=self._threads_to_use,
                 )
             )
-            sfa = self._transformers[0].fit_transform(X, y)
```
patrickzib, line 828: Could you comment on why fit_transform does not work here?

patrickzib: (or add a comment to the source code)

sktime/transformations/base.py
```diff
@@ -549,9 +549,42 @@ def fit_transform(self, X, y=None):
                 then the return is a `Panel` object of type `pd-multiindex`
                 Example: i-th instance of the output is the i-th window running over `X`
         """
-        # Non-optimized default implementation; override when a better
-        # method is possible for a given algorithm.
-        return self.fit(X, y).transform(X, y)
+        # if fit is called, estimator is reset, including fitted state
+        self.reset()
```
patrickzib, line 553: As these checks re-occur in fit and transform, I believe it would be better to refactor everything into one method and call it from fit, transform, fit_transform rather than copying all the content.

patrickzib (changes requested):
Some thoughts about copy-paste vs refactoring into one method.

sktime/transformations/base.py
MatthewMiddlehurst: fit and transform have a separate set of checks and datatype changes. I considered making methods like `_transform_checks` `_fit_checks` but it's not really that compatible. i.e. we don't need `self.check_is_fitted()` or multiple calls of `self._check_X_y`.

patrickzib: That is fine to have separate checks in `transform` and `fit`. But it feels bad to copy these to `fit_transform`. Would parametrization work?

sktime/classification/dictionary_based/_tde.py
MatthewMiddlehurst: SFA is not interface compliant, and adding the checks apparently breaks TDE.

patrickzib: Ok, can you add this as a source code comment? Otherwise someone else might simply change it in a future PR.

sktime/transformations/panel/tsfresh.py
```diff
@@ -616,9 +616,6 @@ def fit_transform(self, X, y=None):
                 then the return is a `Panel` object of type `pd-multiindex`
                 Example: i-th instance of the output is the i-th window running over `X`
         """
-        self.reset()
```
patrickzib, line 619: In `fit()` there is:

```python
        # skip everything if fit_is_empty is True and we do not need to remember data
        if self.get_tag("fit_is_empty") and not self.get_tag("remember_data", False):
            self._is_fitted = True
            return self
```

Why is this not needed in `fit_transform()`?

sktime/transformations/panel/tsfresh.py
```diff
@@ -616,9 +616,6 @@ def fit_transform(self, X, y=None):
...
-        if y is None:
-            raise ValueError("SupervisedIntervals requires `y` in `fit`.")
         X, y, metadata = self._check_X_y(X=X, y=y, return_metadata=True)
```
patrickzib, line 619: In `fit()` this reads:

`X_inner, y_inner = self._check_X_y(X=X, y=y)`

(without the metadata). Maybe unify both `fit`and `fit_transform`?

sktime/transformations/panel/tsfresh.py
```diff
@@ -560,7 +560,7 @@ def _get_selection_params(self):
 
         return selection_params
 
-    def fit_transform(self, X, y=None):
```
patrickzib, line 563: If I am not mistaking, this is the part that is mostly equal to `fit`:

```python

self.reset()

        # skip everything if fit_is_empty is True and we do not need to remember data
        if self.get_tag("fit_is_empty") and not self.get_tag("remember_data", False):
            self._is_fitted = True
            return self

        # if requires_y is set, y is required in fit and update
        if self.get_tag("requires_y") and y is None:
            raise ValueError(f"{self.__class__.__name__} requires `y` in `fit`.")

        # check and convert X/y
        X_inner, y_inner = self._check_X_y(X=X, y=y)

        # memorize X as self._X, if remember_data tag is set to True
        if self.get_tag("remember_data", False):
            self._X = update_data(None, X_new=X_inner)
            
```

with some exceptions (see below)

sktime/classification/dictionary_based/_tde.py
patrickzib: > SFA is not interface compliant, and adding the checks apparently breaks TDE.

I guess the problem might be with the input of a 2d numpy array to SFA, which are then treated as one multivariate time series. I have faced this issue before, too :(

sktime/transformations/base.py
MatthewMiddlehurst: The vast majority of the actual functionality is already contained in separate methods i.e. `self._check_X_y`, `update_data`, `self._convert_output`. I don't really feel so bad about duplicating calls to these. What remains is stuff like         
```
if self.get_tag("requires_y") and y is None:
    raise ValueError(f"{self.__class__.__name__} requires `y` in `fit`.")
```
which is small. The overall method is less than 30 lines including comments.

patrickzib: Yes, but the code already now diverges. And at least to me it is not clear why.

sktime/transformations/panel/tsfresh.py
MatthewMiddlehurst: Regardless of whether `fit` is empty or not we have to return the transformed data, we can't just early exit.

patrickzib: > Regardless of whether fit is empty or not we have to return the transformed data, we can't just early exit.

True, but that is just the return self. 

The check could remain

sktime/transformations/panel/tsfresh.py
MatthewMiddlehurst: While `fit` does not need the metadata, `transform` does for 
```
        # convert to output mtype
        if not hasattr(self, "_output_convert") or self._output_convert == "auto":
            X_out = self._convert_output(Xt, metadata=metadata)
        else:
            X_out = Xt
```

sktime/transformations/panel/tsfresh.py
patrickzib:
```
if self.get_tag("fit_is_empty") and not self.get_tag("remember_data", False):
            self._is_fitted = True
```

MatthewMiddlehurst: we `self._is_fitted = True` at the end anyway, no need to do it twice

sktime/transformations/base.py
patrickzib: My point still being: every future change to fit or transform will have to be added to fit_transform, too. This is always a potential source of error. Thus, I am in favor of moving as much as possible into the same functionality.

MatthewMiddlehurst: Fair. I can move these into separate methods
```
        # convert to output mtype
        if not hasattr(self, "_output_convert") or self._output_convert == "auto":
            X_out = self._convert_output(Xt, metadata=metadata)
        else:
            X_out = Xt
```
```
        # if fit is called, estimator is reset, including fitted state
        self.reset()

        # if requires_y is set, y is required in fit and update
        if self.get_tag("requires_y") and y is None:
            raise ValueError(f"{self.__class__.__name__} requires `y` in `fit`.")
```

or do you want the whole block from fit posted below in a single method with parameters?

patrickzib: Does this work / make sense?

```python
    def _fit_checks(self, X, y, return_metadata=False):
        self.reset()

        # skip everything if fit_is_empty is True and we do not need to remember data
        if self.get_tag("fit_is_empty") and not self.get_tag("remember_data", False):
            self._is_fitted = True
            return True, None, None

        # if requires_y is set, y is required in fit and update
        if self.get_tag("requires_y") and y is None:
            raise ValueError(f"{self.__class__.__name__} requires `y` in `fit`.")

        # check and convert X/y
        if return_metadata:
            X_inner, y_inner, metadata = self._check_X_y(X=X, y=y, return_metadata=True)
        else:
            X_inner, y_inner = self._check_X_y(X=X, y=y)
            metadata = None

        # memorize X as self._X, if remember_data tag is set to True
        if self.get_tag("remember_data", False):
            self._X = update_data(None, X_new=X_inner)

        # skip the rest if fit_is_empty is True
        if self.get_tag("fit_is_empty"):
            self._is_fitted = True
            return True, None, None

        if return_metadata:
            return False, X_inner, y_inner, metadata
        else:
            return False, None, None
            
```


and in `fit` call:
```python
     def fit(self, X, y=None):
        fit_is_empty, X_inner, y_inner = self._fit_checks(X, y)
        if fit_is_empty:
            return self
```

and in `fit_transform` call:
```python
     def fit(self, X, y=None):
        _, X_inner, y_inner, metadata = self._fit_checks(X, y, True)
```

MatthewMiddlehurst: That will skip checks and return nothing for `fit_transform` if `self.get_tag("fit_is_empty")`. A bunch of the fit checks are early abandon, but we can't do that in `fit_transform`

patrickzib: Thanks. True. Another try:

```python
    def _fit_checks(self, X, y, early_abandon=True, return_metadata=False):
        self.reset()

        X_inner = None
        y_inner = None
        metadata = None

        # skip everything if fit_is_empty is True and we do not need to remember data
        if not early_abandon or \
                not self.get_tag("fit_is_empty") or \
                self.get_tag("remember_data", False):
            # if requires_y is set, y is required in fit and update
            if self.get_tag("requires_y") and y is None:
                raise ValueError(f"{self.__class__.__name__} requires `y` in `fit`.")

            # check and convert X/y
            X_inner, y_inner, metadata = self._check_X_y(X=X, y=y, return_metadata=True)

            # memorize X as self._X, if remember_data tag is set to True
            if self.get_tag("remember_data", False):
                self._X = update_data(None, X_new=X_inner)

        self._is_fitted = True

        if return_metadata:
            return X_inner, y_inner, metadata
        else:
            return X_inner, y_inner
```



and in `fit` call:
```python
     def fit(self, X, y=None):
        X_inner, y_inner = self._fit_checks(X, y)
        if self.get_tag("fit_is_empty"):
            return self
```

and in `fit_transform` call:
```python
     def fit_transform(self, X, y=None):
        X_inner, y_inner, metadata = self._fit_checks(X, y, False, True)
```

patrickzib: Thanks! Looks all good to me now

patrickzib (approved):

fkiraly (changes requested):
please do not merge until I have had time to review

TonyBagnall:
> please do not merge until I have had time to review

could you review this please? It has been 5 days

PR #3647, opened 2022-10-25
[MNT] python 3.11 support

fkiraly:
update, Feb 23, 2023: `numba` is now fully isolated, and `sktime` release 0.16.1 has been released with experimental 3.11 support.
Issues that remain to fix before full support in 0.17.0:

* soft dependencies need to be checked with 3.11, and restrictions removed where appropriate https://github.com/sktime/sktime/pull/4269
* conda issues with 3.11 need to be fixed https://github.com/conda-forge/sktime-feedstock/issues/66
  this is waiting for migration, and is not necessarily blocking 0.17.0

---

Python 3.11!

* extends GHA CI tests and release tests to include python 3.11
* updates `pyproject.toml`

todo:

* wait until core deps support 3.11 (or turn them into soft deps, e.g., `statsmodels` and `numba`)
* check if anything fails, e.g., estimators dependent on soft dependencies not compatible with python 3.11
* fix any bugs or add version limiters in estimators
* once all is fixed, update documentation to include 3.11 support

PR #3856, opened 2022-11-29
EXPERIMENTAL: [MNT] python 3.11 compatibility with `numba` as soft dependency

fkiraly:
Experimental PR which tests python 3.11 compatibility after `numba` is turned into a soft dependency.

This is a merge of two PR:

* PR https://github.com/sktime/sktime/pull/3843 which turns `numba` into a soft dependency
* PR https://github.com/sktime/sktime/pull/3647 which updates package metadata to 3.11 support

with the modifications that:

* `numba` is part of the `all_extras` only for python 3.10 and lower.
* `tensorflow` is part of the `all_extras` only for python 3.10 and lower.


patrickzib:
89 files touched for an issue that will be solved by the numba developers in the coming months! Also note that conda does not even support python 3.11 by now. This is the main reason, that here is no numba support. Bugfixes should be minimally invasive. I am calling for a vote on this PR, too.

patrickzib:
Here is the reason for the missing numba support: 

> Our build and release process is based around conda. We can't run CI reliably without an official Python 3.11 package being released via anaconda.org. Hence we are currently blocked by that. We hope to get this merged up and a new version released as soon as the required packages become available. Until then, please be patient, thank you!

https://github.com/numba/llvmlite/pull/869#issuecomment-1313215683

fkiraly:
> 89 files touched for an issue that will be solved by the numba developers in the coming months!

No, `numba` developers will not solve the issues of dependency isolation and that our users all are forced to install `numba`

patrickzib:
You are the only one that does not want to install numba.

fkiraly:
why is this trying to install `numpy 1.19`?

fkiraly:
can be closed, same as #3843 under current `main`

PR #3857, opened 2022-11-30
[BUG] fix unreported `set_params` bug in `ClassifierPipeline` and `RegressorPipeline`

fkiraly:
`set_params` bug in `ClassifierPipeline` and `RegressorPipeline` was broken, it would not correctly update parameters.

The failure to detect this is an instance of the known problem https://github.com/sktime/sktime/issues/3429

The bug has been fixed, and is now covered by appropriate tests (addition of a second parameter set).

The fix is as follow:

* the issue was in `set_params` which accidentally had one too many layer of nesting in the param dict indexing, e.g., `classifier__` etc. This would materialize only for doubly nested estimators.
* this was fixed, with a concomitant extension of the dict subset utility in the `_HetereogenousMetaEstimator`.

Depends on https://github.com/sktime/sktime/pull/3858 as the `DummyClassifier` is used as one of two param sets.

### DIFF
--- a/sktime/base/_meta.py
+++ b/sktime/base/_meta.py
@@ -120,10 +120,41 @@ class _HeterogenousMetaEstimator:
                 "{0!r}".format(invalid_names)
             )
 
-    def _subset_dict_keys(self, dict_to_subset, keys):
-        """Subset dictionary d to keys in keys."""
+    def _subset_dict_keys(self, dict_to_subset, keys, prefix=None):
+        """Subset dictionary d to keys in keys.
+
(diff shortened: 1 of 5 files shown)


aiwalter (dismissed):

miraep8:
Hmmm, any thoughts on the error messages?  Seem to be specific to this PR, but only crop up in certain python versions I am guessing?

fkiraly:
> Hmmm, any thoughts on the error messages? Seem to be specific to this PR, but only crop up in certain python versions I am guessing?

The fact that it´s in different versions is coming from estimators being split across version/OS combinations to minimize runtime of tests

PR #3858, opened 2022-11-30
[BUG] in `DummyClassifier`, fix incorrectly set `capability:multivariate` tag

fkiraly:
In `DummyClassifier`, the `capability:multivariate` tag was incorrectly set to `False`.

This has been changed to `True`.

Reason: `DummyClassifier` ignores `X` (besides its length), so it is unimportant whether it was multivariate or not.


miraep8 (approved):
Seems reasonable to me

PR #3861, opened 2022-12-01
[BUG] Fix #3832

### LINKED ISSUE #3832: [BUG] ComposableTimeSeriesForestRegressor cannot instantiate
I can't use ComposableTimeSeriesForestRegressor because it cannot instantiate...

model = ComposableTimeSeriesForestRegressor()

TypeError: Can't instantiate abstract class ComposableTimeSeriesForestRegressor with abstract method _set_oob_score_and_attributes

MatthewMiddlehurst:
Adds a filler method to `ComposableTimeSeriesForestRegressor`.

### DIFF
--- a/sktime/regression/compose/_ensemble.py
+++ b/sktime/regression/compose/_ensemble.py
@@ -373,6 +373,10 @@ class ComposableTimeSeriesForestRegressor(BaseTimeSeriesForest, BaseRegressor):
 
         self.oob_score_ /= self.n_outputs_
 
+    # TODO - Implement this abstract method properly.
+    def _set_oob_score_and_attributes(self, X, y):
+        raise NotImplementedError("Not implemented.")
+
     def _validate_y_class_weight(self, y):
         # in regression, we don't validate class weights
         # TODO remove from regression


MatthewMiddlehurst:
Who is the codeowner for this? I have no interest in maintaining it past this PR.

TonyBagnall:
@AyushmaanSeth and @mloning listed as authors for ComposableTimeSeriesForestRegressor

TonyBagnall (approved):

fkiraly (changes requested):
can you please explain in which respect the parameter was broken?

In case there was a valid way to address it, it requires deprecation.

MatthewMiddlehurst:
All the parameter does is get passed to the base estimator. Except the base estimator does not have this parameter, so it crashes.

I doubt this requires deprecation, since the estimator is literally unusable.

TonyBagnall:
surely we dont have to deprecate unusable code? The code authors have not replied to the original bug report. This fix should go in now imo

MatthewMiddlehurst:
This seems like an unmaintained duplicate either way.

TonyBagnall:
note that ComposableTimeSeriesForestRegressor is not returned by 
'''
all_estimators(estimator_types="regressor")
'''
which seems strange.

Issue #3865, opened 2022-12-02
 [GOV] Evolution of sktime governance


aiwalter:
**Is your feature request related to a problem? Please describe.**
This issue aims to identify sections in sktime governance that are "undemocratic". General idea of this discussion is to give more power to the core developers to avoid that individuals can take hostage of sktime. 

Undemocratic in this context are for example current gov items:
- Single individuals can block nomination of CC or CoC members, 
see #3939
- CC can supersede core dev votes. This can lead to offending majority of core devs,
see #3937
- CC and CoC nominations are currently for lifetime. This should be changed, so a nomination ends e.g. after 2 years. Then another core dev vote has to happen. This is common practice in any org that claims to have a democratic governance. 
see #3868
- Separation of power between CC and CoC, currently there is no separation on member level.
see #3870

**Describe the solution you'd like**
I will make separate PRs for the different topics. Discussion can happen at PR. General discussion can happen at this issue.

fkiraly:
@aiwalter, I think the implication that sktime governance is undemocratic is wholly inappropriate. We have a representative decision making system and voting.

This is also not sufficiently addressing problems we have seen recently emerge in discussion culture.

I think these parallel issues that wikipedia is seeing, especially things like hounding, single-issue vote piling by people who do not really contribute to the community.

The key parallel concept would be "clearly not being here to build a package":
https://en.wikipedia.org/wiki/Wikipedia:Here_to_build_an_encyclopedia#Clearly_not_being_here_to_build_an_encyclopedia

Related issues:

* https://en.wikipedia.org/wiki/Wikipedia:Civil_POV_pushing
* https://en.wikipedia.org/wiki/Wikipedia:POV_Railroad

aiwalter:
I didnt say the governance in undemocratic in general, but there are parts of it that are clearly undemocratic as mentioned above. The general idea of this issue is to give more power to the core developers, who are the main representatives of the community.

fkiraly:
You are making PR that abolish separation of duties/powers and strongly imply that we are currently undemocratic.

fkiraly:
"clearly undemocratic" is a misrepresentation of how representative democracies and multi-chamber systems work.

Democratic does not mean that a given person always get what they want. In representative democracies like the one you live in, it does not even mean that the majority always gets what it wants.

fkiraly:
This is a populist move, and possibly plan to seize control of sktime.

[...]

PR #3867, opened 2022-12-02
[GOV] Remove ability of CC to overrule core dev decisions

aiwalter:
#### Reference Issues/PRs
#3865 

Remove ability of CC to overrule decisions made by core devs.

### DIFF
--- a/docs/source/get_involved/governance.rst
+++ b/docs/source/get_involved/governance.rst
@@ -447,8 +447,6 @@ tracker <https://github.com/sktime/sktime/issues>`__,
 `pull requests <https://github.com/sktime/sktime/pulls>`__ or an :ref:`steps`. Some
 sensitive discussions and appointment votes occur on private chats.
 
-The CC reserves the right to overrule decisions.
-
 We distinguish between the following types of proposed changes. The
 corresponding decision making process is described in more detail below.
 


fkiraly:
I formally vote this down.

This is a substantial change of how we currently decide, and would strengthen patterns like hounding/piling that are currently already problematic.

Scikit-learn and wikipedia both have this kind of overruling mechanism and I strongly argue we need to keep it.

CC should be able to overrule a dogpile of single-issue-voters based on architectural or strategic viewpoints. This proposal would remove the stage 3 option.

aiwalter:
I am not having a strong opinion here, if the PR #3868 is merged then I am fine with not having this one actually.

GuzalBulatova:
I agree with @fkiraly  that CC should have an option to overrule simple majority decisions, given that it's CC consensus that the majority decision isn't in the best interests of sktime

aiwalter:
I still think this PR would be an improvement, but I am happy to put this PR to draft for now. If we have merged the other gov PRs I think this will become more an optional part of lower priority. So lets focus discussion for now on the other gov PRs 👍🏻

TonyBagnall (approved):

lmmentel (approved):

fkiraly:
formal downvote.

Reasons explained here: https://github.com/sktime/sktime/issues/3882

aiwalter:
a core dev vote was invoked on this PR in the private core dev Slack channel and will end after 7 days.

aiwalter:
results of core-dev vote is >2/3 of casted votes, therefore merging this
image

PR #3872, opened 2022-12-02
[BUG] `ClearSky` doesn't raise error for range indexes and when `X` has no set frequency

ciaran-g:
#### Reference Issues/PRs

No open issue

Added some bug fixes which were highlighted by @robertmcleod2 - cheers!

The bugs were

* The `ClearSky` transformer relies on the datetime index to get the time of day. I thought the `enforce_index_type` tag would deal with removing range indexes, but it doesn't at the moment. Added a check for this
* The frequency of the data should be set and there was no check for this. I added one along with a function to infer it if possible. 
* The frequency should be within a range to be suitable for the transformer, so added a check for this as well.

Added tests for these bugs as well. 

No

### DIFF
--- a/sktime/transformations/series/tests/test_clear_sky.py
+++ b/sktime/transformations/series/tests/test_clear_sky.py
@@ -6,6 +6,7 @@
 __author__ = ["ciaran-g"]
 
 import numpy as np
+import pandas as pd
 import pytest
 
 from sktime.datasets import load_solar
@@ -69,7 +70,7 @@ output_chk = [
     reason="skip test if required soft dependency for hmmlearn not available",
 )
 def test_clearsky_trafo_vals():
-    """Tests clear sky trafo with and without missing values."""
+    """Tests clear sky trafo with and without missing values and period intex."""
     y = load_solar(api_version=None)
     cs_model = ClearSky()
     y_trafo = cs_model.fit_transform(y)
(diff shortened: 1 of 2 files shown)


sktime/transformations/series/tests/test_clear_sky.py
```diff
@@ -69,7 +70,7 @@
     reason="skip test if required soft dependency for hmmlearn not available",
 )
 def test_clearsky_trafo_vals():
-    """Tests clear sky trafo with and without missing values."""
+    """Tests clear sky trafo with and without missing values and period intex."""
```
TonyBagnall, line 73: intex -> index

TonyBagnall:
hi, is it possible to reconfigure the test test_clearsky_trafo_vals so it is faster? It currrently takes over 3 minutes, and we would like tests to be under 30 seconds (and preferably much faster!). 

191.13s call     sktime/transformations/series/tests/test_clear_sky.py::test_clearsky_trafo_vals

ciaran-g:
> hi, is it possible to reconfigure the test test_clearsky_trafo_vals so it is faster? It currrently takes over 3 minutes, and we would like tests to be under 30 seconds (and preferably much faster!).

Tests should be around 20-25s now @TonyBagnall - I also relaxed the checks to allow daily frequency data since that's why it was failing the test_all_transformers automatic tests I think. 

I hadn't tested it out on daily data, so excluded it in the previous commit, but after checking that it does work as expected the check is now to allow greater than or equal to daily frequency

TonyBagnall (approved):

PR #3877, opened 2022-12-03
EXPERIMENTAL: [MNT] `scipy` as a soft dependency

fkiraly:
This is an experimental testing PR for turning `scipy` into a soft dependencym to study test failures and their locations.

Relies on the following PR which isolate `scipy`:

* https://github.com/sktime/sktime/pull/3707


fkiraly:
`scipy` is a dependency of `sklearn`, so this is not actually testing what happens.

We can still isolate the `scipy` locus, but we can't make it a soft dependency if it is a core dependency of `sklearn`.

Issue #2608, opened 2022-05-05
[DOC] add explanation of dunder method construction and postprocessing in pipelines to forecasting tutorial


fkiraly:
An explanation of "compact" dunder pipelining with `TransformedTargetForecaster` should be added to the forecasting tutorial - not replacing, but being added to the subsections where pipelines are explained.

For example, instead of
```python
pipe = TransformedTargetForecaster(
    steps=[
        ("imputer", Imputer()),
        ("forecaster", UnobservedComponents())
    ]
)
```

one can also write

```python
pipe = Imputer() * UnobservedComponents()
```

This would also be a great opportunity to explain postprocessing, i.e., transformers can also be added after the forecaster since 0.11.2.

nshahpazov:
@fkiraly I would like to add this. There's already documentation on the dunder method usage, but nothing about post-processing.

fkiraly:
thanks!

PR #3878, opened 2022-12-03
[DOC] added post-processing notes in tutorial

nshahpazov:
### Reference Issues/PRs
- https://github.com/sktime/sktime/issues/2608
### Description of PR

- [x] Checked and confirmed that dunder pipeline methods are described in the tutorial
- [x] Added myself to the [list of contributors](https://github.com/sktime/sktime/blob/main/.all-contributorsrc).
- [x] Added description in the tutorial on how TransformedTargetForecaster can also end with a post-processing Transformer, in this case, I've used a simple predictions rounding


fkiraly (changes requested):
That´s nice, thanks!

May I ask for one thing, to keep the repository size small: can you make sure only the parts are changed where you add content?

I think the current huge diff comes from you rerendering plots.

Worst case, this might require going into the ipynb file with an editor, but should also work if you take the original file and copy-paste the change.

nshahpazov:
@fkiraly I've addressed the proposed changes. The actual diff is visible now.

examples/01_forecasting.ipynb
```diff
@@ -5376,7 +5376,8 @@
    "source": [
```
aiwalter, line 5376: <p>we actually dont have a scitype "post-processing transformer", so maybe lets frame that a little bit different?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/sktime/sktime/pull/3878/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='203'/>

aiwalter:
also please make sure you run pre-commit bcz linting is failing

examples/01_forecasting.ipynb
```diff
@@ -5425,7 +5426,7 @@
...
+    "The `TransformedTargetForecaster` is constructed with a list of steps, each a pair of name and estimator. The last estimator should be of either, a forecaster scitype or a post-processing transformer, the other estimators should be series-to-series transformers which possess both a `transform` and `inverse_transform` method. The resulting estimator is of forecaster scitype and has all interface defining methods. In `fit`, all transformers apply `fit_transforms` to the data, then the forecaster's `fit`; in `predict`, first the forecaster's `predict` is applied, then the transformers' `inverse_transform` in reverse order."
```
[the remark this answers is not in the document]
fkiraly: bit complex, so I would suggest to split the explanation:

* case 1 - only pre-processing transformers
* case 2 - only post-processing transformers
* case 3 - putting it together, and spelling out the sequence

fkiraly (changes requested):
Thanks! Great idea to add this content, will be useful to our users!

Change requests:

* linting
* I would split the content and introduce it step by step, or it may be confusing if it comes all at once, see suggestion above

examples/01_forecasting.ipynb
nshahpazov: Thanks!
I've the post-processing from this paragraph and moved it further below, following the other comments.

nshahpazov:
@fkiraly I've rephrased the paragraphs into three sections like you suggested. Linting is working now. Let me know if you find it ok.

fkiraly (approved):
All good! Just made some minor changes in formatting.

fkiraly (approved):

lmmentel (approved):

PR #3879, opened 2022-12-03
[ENH] Add `prefer="threads"` to classification `Parallel` usage

MatthewMiddlehurst:
Some classifiers have had issues with the current default backend (#3788), and "threads" seems to be a more sensible default for data science applications (discussion in #3652).


TonyBagnall (approved):

fkiraly (changes requested):
I would expose this as a parameter, then deprecate.

TonyBagnall:
this fixes an obscure bug with a classifier #3880 . Exposing then deprecating takes months, it seems completely unnecessary in this instance.

PR #3883, opened 2022-12-04
[ENG] Evaluate function to accept list of scorers

aiwalter:
Function `from sktime.forecasting.model_evaluation import evaluate` to accept list of scorers, this is then added as new columns to the output data.

### DIFF
--- a/sktime/forecasting/model_evaluation/_functions.py
+++ b/sktime/forecasting/model_evaluation/_functions.py
@@ -8,13 +8,15 @@ __all__ = ["evaluate"]
 
 import time
 import warnings
+from typing import List, Optional, Union
 
 import numpy as np
 import pandas as pd
 
 from sktime.datatypes import check_is_scitype, convert_to
 from sktime.exceptions import FitFailedWarning
-from sktime.forecasting.base import ForecastingHorizon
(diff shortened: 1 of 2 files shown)


RNKuhns:
@aiwalter this will be a nice incremental increase in functionality. I'll review once out of draft.

sktime/forecasting/model_evaluation/_functions.py
```diff
@@ -8,13 +8,16 @@
...
 
 import numpy as np
+import numpy.typing as npt
 import pandas as pd
 
 from sktime.datatypes import check_is_scitype, convert_to
 from sktime.exceptions import FitFailedWarning
-from sktime.forecasting.base import ForecastingHorizon
+from sktime.forecasting.base import BaseForecaster, ForecastingHorizon
+from sktime.forecasting.model_selection._split import BaseSplitter
```
fkiraly, line 20: risk of circular imports, I would recommend to avoid doing this

fkiraly (changes requested):
Makes sense.

The "returns" section in the docstring should document the change properly.

fkiraly (approved):
Thanks for completing the docstring.

PR #3888, opened 2022-12-05
[MNT] Add type hint to PR template

aiwalter:
Add type hint to PR template

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -43,7 +43,7 @@ Please go through the checklist below. Please feel free to remove points if they
 - [ ] Optionally, I've updated sktime's [CODEOWNERS](https://github.com/sktime/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
 - [ ] I've added unit tests and made sure they pass locally.
 - [ ] The PR title starts with either [ENH], [MNT], [DOC], or [BUG] indicating whether the PR topic is related to enhancement, maintenance, documentation, or bug.
-- [ ] I have added parameter type hints via `typing` to functions and methods that were changed.
+- [ ] I have checked that all functions and methods that were changed have correct parameter type hints via the `typing` module.
 
 ##### For new estimators
 - [ ] I've added the estimator to the online documentation.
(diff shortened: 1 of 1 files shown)


aiwalter:
we should ideally also add type hints to all the extension templates, I might do that in next days maybe

fkiraly (changes requested):
Apologies, I don't think we should enforce type hints.

We just recently recovered from issues due to "greedy imports" and circular imports, as inputs where typed with `sktime` base classes, which in turn imported modules that would otherwise not have been coupled.

I think we should not enforce type hints until we can ensure that we do not introduce a substantial module coupling footprint through that.

ilkersigirci:
> Apologies, I don't think we should enforce type hints.
> 
> We just recently recovered from issues due to "greedy imports" and circular imports, as inputs where typed with `sktime` base classes, which in turn imported modules that would otherwise not have been coupled.
> 
> I think we should not enforce type hints until we can ensure that we do not introduce a substantial module coupling footprint through that.

I think type hinting is a good way to catch bugs in the development.  To remedy circular imports , one can use the following for circular imports dependencies:

## TYPE_CHECKING
```python

from __future__ import annotations
from typing import TYPE_CHECKING

if TYPE_CHECKING:
	# imports that cause circular dependencies goes here
```

- This way we can easily solve any circular dependency problem because in runtime `TYPE_CHECKING` is always False.
- More info can be found in [mypy documentation](https://mypy.readthedocs.io/en/stable/runtime_troubles.html)

fkiraly:
I see, thanks for the learning moment, @ilkersigirci!

I get that at runtime this does not run the imports - does it still not cause circular imports when the dev IDE does type checking though?

ilkersigirci:
> I see, thanks for the learning moment, @ilkersigirci!
> 
> I get that at runtime this does not run the imports - does it still not cause circular imports when the dev IDE does type checking though?

You are welcome, sir. No, it doesn't cause any circular import when an IDE inferring types, because they don't actually run the code to get the types. I have been using this method for a long time and hasn't faced any problem so far.

ilkersigirci:
Actually what I mean by not running the code is that they don't need python runtime. For example; 
- Vscode uses [Pylance Language Server](https://marketplace.visualstudio.com/items?itemName=ms-python.vscode-pylance)
- Pycharm uses Pyright

to find code error such as indentation error or unmatched types without running the code. For mypy case, they are using [stub files](https://mypy.readthedocs.io/en/stable/stubs.html) to get the correct type hints for the code. I hope I am able to make more lights to the subject.

fkiraly:
no follow-upengagement for 2 months, and "do not enforce type hint" opinion seems more prevalent (although there was no vote). Can reopen if anyone has a strong opinion.

Issue #3712, opened 2022-11-05
[ENH] Wrapper for TSFeatures Transformer


chillerobscuro:
**Is your feature request related to a problem? Please describe.**
As named in [2982](https://github.com/sktime/sktime/issues/2982), a wrapper for the python implementation of [`tsfeatures`](https://pypi.org/project/tsfeatures/), which was originally written in R.

**Describe the solution you'd like**
Wrap the [existing package]((https://pypi.org/project/tsfeatures/)) similar to the existing wrapper around TSFresh.

**Additional context**
Can we verify that this external package is stable/solid enough to build a wrapper for? It is built by Nixtla who is doing some interesting work in the TS space, and hasn't been in active development for about a year now.

I'm happy to take on this task if deemed valuable, just have some questions about what type of inputs we need to accept, some specifics about scitypes, and the necessary architecture of the wrapper that lives on top of `BaseTransformer`.

fkiraly:
I think this is a good idea - I did not know there was a `python` version for tsfeatures.
Interfacing might be an easy solution to #2982.

Nixtla are fairly active, so might be a good choice.
But, we should of course test and review the package before we interface.

chillerobscuro:
@fkiraly I found that `tsfeatures` is failing on our test series when len<12, and working with len>=12 so I made [this ticket for nixtla](https://github.com/Nixtla/tsfeatures/issues/16). 

I could potentially just raise an exception if shorter inputs are passed, but in that case I'm skipping almost all the test cases. Not going to count on Nixtla making a fix but not sure how to proceed otherwise. The wrapper could still have value if had had a "too short input series" warning but would not properly engage the test series.

fkiraly:
hm, that's indeed odd - our test cases are 11 or shorter, so I suppose it's just below...

we could just make the test cases longer, e.g., 14? Surely that shouldn't affect runtime too much.

Do we know why it is failing? What is the error message?

fkiraly:
Made a PR here that should help you: https://github.com/sktime/sktime/pull/3892 - this makes test series in scenarios length 14 or longer.

I also wonder whether sth like `min_series_length` should be a tag. Might be worth opening an issue for that, since the min length of a series is not an inspectable interface element.

Alternatively, if length is 11 or shorter, you could just produce a dummy data frame and raise a warning.

chillerobscuro:
[...]

PR #3892, opened 2022-12-05
[ENH] "typical length" constant in transformer scenarios

fkiraly:
update: the PR no longer makes the example series longer. It only adds the "typical length" constant which can be used to change length of all examples, but does not change the length.

---

This extends the length of time series in the base interface compliance tests for `sktime` transformers.

It also exposes the "typical length" of the time series as a config parameter in the transformer scenarios file.

Should help with this PR: https://github.com/sktime/sktime/issues/3712


fkiraly:
@chillerobscuro, is this still needed for the `tsfeatures` trafo, or have you found a workaround?

chillerobscuro:
> @chillerobscuro, is this still needed for the `tsfeatures` trafo, or have you found a workaround?

The author of the repo put in a [fix](https://github.com/sktime/sktime/issues/3712#issuecomment-1354037097) in the new version, so this change is not necessary for me, as long as anything using `tsfeatures` is `>=0.4.0`.

fkiraly:
Ok, so I'll close this, and you ensure to have a lower version bound in the soft dependency then?

chillerobscuro:
> Ok, so I'll close this

That works for me

> and you ensure to have a lower version bound in the soft dependency then?

In `pyproject.toml` I have `"tsfeatures>=0.4.0"`, since as we discussed `_check_soft_dependencies` doesn't support versions.

fkiraly:
> since as we discussed `_check_soft_dependencies` doesn't support versions.

The newest version of `_check_soft_dependencies` now does!

fkiraly:
actually, I think it still makes sense to merge this since it exposes the length in all the scenarios as a parameter that is easy to change. I'll simply change it to 10 (what it currently is) instead of 14 (the longer version).

PR #3898, opened 2022-12-07
[ENH] Changed grid search parallelization to use backend param from evaluate function

aiwalter:
removes `joblib` from grid search and adds param `backend` and `**kwargs` to use recently added parallelizations directly in `evaluate`.

fyi @topher-lo


aiwalter:
I think would be good to add docstring for the `kwargs` what they can be for which case or even make explicit

aiwalter:
@topher-lo question regardin `kwargs` in `èvaluate`, Icant see `kwargs` used for `backend="dask"` but in docstring it says

> **kwargsKeyword arguments
> Only relevant if backend is specified. Additional kwargs are passed into dask.distributed.get_client or dask.distributed.Client if backend is set to “dask”, otherwise kwargs are passed into joblib.Parallel.

can you clarify if I am missing sth? Also what would you recommend to be default `backend` for grid search and `evaluate`?

fkiraly (changes requested):
* why are you changing the docstring of `evaluate`? Is that a mistake?
* changing public parameters of estimator classes requires deprecation, see here: https://www.sktime.org/en/stable/developer_guide/deprecation.html

aiwalter:
> why are you changing the docstring of evaluate? Is that a mistake?

Yes.

> changing public parameters of estimator classes requires deprecation, see here: https://www.sktime.org/en/stable/developer_guide/deprecation.html

I reverted that, we can actually keep this

topher-lo:
@aiwalter thanks for pointing this out! You are right about kwargs for Dask, the first iteration had client inside the function but I removed it cause it's not strictly necessary (ie it's better to just create your Dask client outside). 

Loky is a good default. Can't go wrong with that 👍

aiwalter:
closing due to #4153

fkiraly:
@hazrulakmal, I would suggest:

* branch off this branch
* update from `main` and fix any merge conflicts
* work/finish/review
* ensure to give credit to aiwalter in the other PR - as it might not show up in the diff, I usually do this by prefixing the PR name with "& githubname [ENH] etcetc", then it will show up in the changelog generator correctly. Also, simply by linking to this PR.

fkiraly:
(this is a PR from a fork, so I'm not sure whether working on this branch would even be technically possible. Branching off, of course, always works)

fkiraly:
superseded by https://github.com/sktime/sktime/pull/5430 which uses unified parallelization backend for both `evaluate` and grid search.

PR #3906, opened 2022-12-08
[MNT] Update sklearn compatibility to `1.2.x`, version bound to `<1.3`

aiwalter:
(@fkiraly update:) This PR updates the `sklearn` version bound to `<1.3`.

Includes fixes for `sklearn 1.2` compatibility:

* https://github.com/sktime/sktime/pull/3918
* https://github.com/sktime/sktime/pull/3919

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,7 @@ dependencies = [
     "numba>=0.55",
     "numpy>=1.21.0,<1.23",
     "pandas>=1.1.0,<1.6.0",
-    "scikit-learn>=0.24.0,<1.2.0",
+    "scikit-learn>=0.24.0,<1.3.0",
     "statsmodels>=0.12.1",
     "scipy<2.0.0",
 ]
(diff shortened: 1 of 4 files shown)


aiwalter:
there is an issue with `from sklearn.neighbors._base import _check_weights`  which ahhpens twice in the file `_time_series_neighbors.py`, ideally we should not import private functions from dependency packages bcz they tend to break as now. @jasonlines @TonyBagnall do you mind to have a look into this?

fkiraly:
FYI, `KNeighborsTimeSeriesClassifier` is no longer the original version of @jasonlines and @TonyBagnall that "hacked" the `sklearn` one and used its private functions.

Some while ago, I change-refactored it to composition where `sklearn` is internally used. The fix should hence be easy (rather than requiring a substantial amount of rework):

https://github.com/sktime/sktime/pull/3918

fkiraly:
moving my work here: https://github.com/sktime/sktime/pull/3922

there's too many changes, and I don't want to fiddle around too much in @aiwalter's PR.

please feel free to revert everything.

fkiraly:
superseded by fixes in https://github.com/sktime/sktime/pull/3922

PR #3908, opened 2022-12-08
[ENH] Add `colors` argument to `plot_series`

chillerobscuro:
Add an option to `plot_series()` to allow the user to pass a list of colors. I added a small util function to check that the length is correct, and use `from matplotlib.colors import is_color_like` to check that they are all valid colors. If either condition is not met, then it emits a warning and defaults back to the current behavior, which is using `sns.color_palette("colorblind", n_colors=n_series)`.

### DIFF
--- a/sktime/utils/plotting.py
+++ b/sktime/utils/plotting.py
@@ -7,7 +7,7 @@ __all__ = ["plot_series", "plot_correlations", "plot_windows"]
 __author__ = ["mloning", "RNKuhns", "Drishti Bhasin", "chillerobscuro"]
 
 import math
-from warnings import simplefilter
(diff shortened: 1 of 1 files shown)


GuzalBulatova (approved):
Looks good to go! Please don't forget to add yourself to the list of contributors, if you haven't yet. Does the colors param works as expected?

chillerobscuro:
> Looks good to go! Please don't forget to add yourself to the list of contributors, if you haven't yet. Does the colors param works as expected?

Thanks Guzal, I'm already on the list and this feature works as expected, at least for me.

fkiraly (approved):
Great addition, thanks!

Issue #3819, opened 2022-11-19
[BUG] codecov no longer works


fkiraly:
I think code coverage via codecov is no longer reported in CI?

Unsure why it stopped working or when, but I cannot see it anymore in most recent PR.

PR #3912, opened 2022-12-08
[ENH] MLflow custom flavor

### LINKED ISSUE #1215: [ENH] Adding sktime to mlflow
I got the idea to add `sktime` as a package to `mlflow` like the following packages that are already included:
https://www.mlflow.org/docs/latest/python_api/index.html

Fot this we would have to contribute to `mlflow` ofc, but I thought anyway to open the issue here first as we might find here someone to work on it more likely maybe. We could also open an issue at `mlfow`.

**Additional context**
Outreach. We might also think of other packages where we could integrate `sktime` similarly to `mlflow`.

benjaminbluhm:
#### Reference Issues/PRs
Resolves #1215

This PR adds mlflow custom model flavor for sktime. The implementation closely follows the approach of mlflow built-in model flavors, see [`pmdarima`](https://www.mlflow.org/docs/latest/models.html#pmdarima-pmdarima) as an example in mlflow repo (for more details see [mlflow.pmdarima](https://github.com/mlflow/mlflow/blob/master/mlflow/pmdarima.py) and [tests.pmdarima.test_pmdarima_model_export](https://github.com/mlflow/mlflow/blob/master/tests/pmdarima/test_pmdarima_model_export.py)). I also added a mlflow notebook under examples which could be added to the sphinx docs. 

New dependencies introduced: `boto3, mlflow, moto`

I tried to add all dependencies as soft dependencies but ran into a challenge for mlflow. We need to import `pyfunc` from mlflow at the module level to enable loading sktime model as pyfunc flavor. This will not work if we import pyfunc inside the functions. 

Would it be acceptable to have mlflow as core dependency or do you have another suggestion?

- how we want to handle the open dependency question
- check the `_SktimeModelWrapper`class which contains the sktime custom inference logic 
- predict method of `_SktimeModelWrapper` is only allowed to take a single  DataFrame as an input (see [mlflow documentation](https://www.mlflow.org/docs/latest/python_api/mlflow.pyfunc.html) for more details) - thus we can pass exogenous regressor only as 2D or 3D numpy array. Based on my understanding this would cover all forecasters utilizing scitypes `Series` and `Panel` - can you confirm this?
- whether you see the need to add more unit tests
- should we add example notebook to sphinx docs? Is there anything in the code that needs to be modified for this inclusion?  

##### For all contributions

##### For new estimators


examples/mlflow.ipynb
```diff
@@ -0,0 +1,749 @@
+{
```
aiwalter, line 1: <p>Really nice notebook! :) </p><p> </p><p>This table seems not correctly rendered, can you pls check?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/sktime/sktime/pull/3912/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

sktime/utils/tests/test_mlflow_sktime_model_export.py
```diff
@@ -0,0 +1,638 @@
...
+@pytest.fixture(scope="module")
+def mock_s3_bucket():
+    """Create a mock S3 bucket using moto.
+
+    Returns
+    -------
+    string with name of mock S3 bucket
+    """
+    import boto3
+    import moto
```
aiwalter, line 42: if those 2 dependencies are only required for testing, I would say we can move them to the `dev` dependencies instead `all_extras`.

sktime/utils/mlflow_sktime.py
```diff
@@ -0,0 +1,534 @@
...
+    "get_default_pip_requirements",
+    "get_default_conda_env",
+    "save_model",
+    "log_model",
+    "load_model",
+]
+
+import logging
+import os
+import pickle
```
aiwalter, line 29: is it possible to use `cloudpickle` instead `pickle`? Or even parametrize this so user could chose?

aiwalter (commented):
Really nice work! :tada: Left some comments, mostly minor things. The unit tests I did not yet fully check.

some general questions:
- what were the linting/test issues related to `# doctest: +SKIP`?
- what were linting issues related to `# noqa: E501`?

There is a merge conflict, can you check so the CI/CD can run?

benjaminbluhm:
> here is a merge conflict, can you check so the CI/CD can run?



> Really nice work! tada Left some comments, mostly minor things. The unit tests I did not yet fully check.
> 
> some general questions:
> 
>     * what were the linting/test issues related to `# doctest: +SKIP`?
> 
>     * what were linting issues related to `# noqa: E501`?
> 
> 
> There is a merge conflict, can you check so the CI/CD can run?

Thanks a lot for your quick feedback and comments! This is really helpful :smiley:

Will address your points over the coming days

fkiraly (changes requested):
Well, excellent!

I will comment in more detail on:
* (soft) dependency handling 
* predict/inference signature

Re "other inference methods", I would say, let's do that in chunks. Let's work out the interface for the "core" methods first, and the dependency thing, we can add more features later

fkiraly:
Re dependency handling:

* I would add the `mlflow` dependency in an extra dependency set, called `mlflow`.
* the current dependency handling requirement is that the full test suite runs with minimal dependencies, i.e., without the `mlflow` module. Also, the crawling utilities (like `all_estimators`) must not encounter any offending imports. This is what `test-nosoftdeps` is testing.
* Isolating the imports should be straightforward for most instances, as they are already inside functions. The offending imports are those outside functions. Unfortunately, they relate to docstring utils. Can the decorators be omitted?

fkiraly:
Re `_SktimeModelWrapper`, I understand that it allows passing of a `dict`?

So, if there's only a single argument, we can use `pd.DataFrame`; for multiple arguments, I would design the interface that it expects a `dict` like `{"X": my_X, "y": my_y}` that is keyed by arguments of the predict/inference function.

fkiraly:
@benjaminbluhm, regarding isolation of module level imports, the one main other case which could serve as half of a blueprint is that of the `numba` package which is currently a core dependency.

As said above, the two requirements for isolation are isolation from (a) `sktime` onboard crawlers and (b) `pytest`.

I have figured out how to isolate crawlers - a combination of using private modules and using a modified `pkgutil.walk_packages`.

I don't know yet how to isolate `pytest`, e.g., tell `pytest` "don't look in this file".

The first soft dependency isolation principle is part of this recent PR on `numba` https://github.com/sktime/sktime/pull/3631
(please ignore the unrelated discussions in the tracker).

fkiraly:
> Naively, it looks like it could be moved into the functions, but perhaps there are some interface assumptions on how the template sits in the module?

I'm curious, trying here what happens: https://github.com/sktime/sktime/pull/3920

fkiraly:
@benjaminbluhm, re dependency isolation:

what I did in https://github.com/sktime/sktime/pull/3920 seems to work.

I have not tracked down the remaining test failures but I think these are genuine, explained as:

* dependency not isolated in doctests. Use doctest skips (search for `+SKIP` for some examples in the codebase). If this would reduce test coverage, move the doctest into a `pytest` test that isolates the soft dependency as you have already done.
* something with handling paths on windows. Not sure what that is, but looks genuine.

benjaminbluhm:
@fkiraly Thank you for all your feedback, much appreciated! And great you found a way for mlflow soft dependency isolation.

I will go through your comments in more detail and address them / provide feedback.

Should I continue working on #3920 and commit my code changes into [mlflow-naive-isolation](https://github.com/sktime/sktime/tree/mlflow-naive-isolation) branch or you have another suggestion?

benjaminbluhm:
> Re `_SktimeModelWrapper` interface, I understand that the `pyfunc` interface allows passing of a `dict`?
> 
> So, if there's only a single argument, we can use `pd.DataFrame`; for multiple arguments, I would design the interface that it expects a `dict` like `{"X": my_X, "y": my_y}` that is keyed by arguments of the predict/inference function.

I also think that using `dict` rather than `pd.DataFrame` for passing multiple arguments would be more intuitive and this would also simplify passing exogenous regressor (e.g. no conversion from DataFrame to numpy array needed). 

The only concern or question I have with using `dict` is possible side effect on using mlflow deployment tools. These tools accept csv or json input for model scoring via API endpoint, see for example [here](https://www.mlflow.org/docs/latest/models.html#deploy-mlflow-models) and [here](https://www.mlflow.org/docs/latest/models.html#deploy-a-python-function-model-on-microsoft-azure-ml) for more details on accepted input formats and code examples. I was wondering whether having single DataFrame input to pyfunc predict function might be required for model scoring via API endpoint to work correctly. Also in the [Inference API documentation](https://www.mlflow.org/docs/latest/python_api/mlflow.pyfunc.html#pyfunc-inference-api) it states that _All PyFunc models will support pandas.DataFrame as input and PyFunc deep learning models will also support tensor inputs in the form of Dict[str, numpy.ndarray] (named tensors) and numpy.ndarrays (unnamed tensors)._

fkiraly:
> I was wondering whether having single DataFrame input to pyfunc predict function might be required for model scoring via API endpoint to work correctly.

Hm, I don't see a reason why it would break, since the outputs are still single data frames (and that's what goes into scoring?)

> Should I continue working on #3920 and commit my code changes into [mlflow-naive-isolation](https://github.com/sktime/sktime/tree/mlflow-naive-isolation) branch or you have another suggestion?

I would suggest, merge into your fork from #3920 what you find useful and continue working here.

benjaminbluhm:
@fkiraly I merged your contributions from #3920 into this PR as you suggested

On the input type for the predict method of the `_SktimeModelWrapper` I investigated a bit further. If we want to allow the user to serve sktime model to API endpoint for inference (scoring) using [mlflow built-in deployment tools](https://www.mlflow.org/docs/latest/models.html#built-in-deployment-tools) we need to have `pd.DataFrame` as input type. This requirement is stated here: https://github.com/mlflow/mlflow/blob/master/mlflow/pyfunc/scoring_server/__init__.py. I believe this is also the reason why the various mlflow [built-in model flavors](https://www.mlflow.org/docs/latest/models.html#built-in-model-flavors) (pmdarima, statsmodels, prophet etc.) expect to pass a `pd.DataFrame` to the predict method. I think it would be important for sktime to also follow this convention so mlflow deployment capabilities can be fully utilized.

I added an end-to-end example for model deployment via local API endpoint in the mlflow example notebook. In short the local host workflow looks like this:

- Use mlflow cli to serve model from selected mlflow run to local host: `mlflow models serve -m runs:/99b7ea689b0843889e3b181030fc44cf/model --env-manager local --host 127.0.0.2`
- Make prediction request to hosted sktime model:
```
host = "127.0.0.2"
url = f'http://{host}:5000/invocations'

prediction_conf = pd.DataFrame(
    [{"predict_method": "predict", "fh": 1}]
)

json_data = {"dataframe_split": prediction_conf.to_dict(orient='split')}
print(json_data)

# Json input is parsed into pandas.DataFrame and passed to the model
response = requests.post(url, json=json_data)
response.json()
```
{'dataframe_split': {'index': [0], 'columns': ['predict_method', 'fh'], 'data': [['predict', 1]]}}

{'predictions': [{'0': 66513.0}]}

fkiraly:
> API endpoint for inference (scoring) using [mlflow built-in deployment tools](https://www.mlflow.org/docs/latest/models.html#built-in-deployment-tools) we need to have `pd.DataFrame` as input type

But that should be fine, shouldn't it?

* in forecasting, we can pass `fh` always in `fit`. In `predict`, we can pass an `X: pd.DataFrame`, or `None` (if `None` is not acceptable, an empty df then)
* in classification, `predict` has the single argument `X` which can be a `pd.DataFrame`

fkiraly:
Other question: do you know whether `mlflow` supports `update` endpoints? I.e., not just batch train/inference model, but stream/online train/update/infer/update/infer etc ?

fkiraly:
PS: unix seems to fail with this:

```ERROR: Could not install packages due to an OSError: [Errno 28] No space left on device```

**Comment by benjaminbluhm** [comment]
> But that should be fine, shouldn't it? in forecasting, we can pass `fh` always in `fit`, so we can ignore it in `predict`. That is, in `predict`, we can pass a single argument `X: pd.DataFrame`, or `None` (if `None` is not acceptable, an empty df then)

Good we can ignore `fh` in `predict`, agree we can then pass a single argument `X: pd.DataFrame` or empty df. 

In case of probabilistic and variance forecasting we would still need to pass `pd.DataFrame` with column `predict_method` (required) and optionally `X: np.ndarray`, `coverage` etc.

**Comment by benjaminbluhm** [comment]
I will comment out the part in the example notebook with prediction request to local API endpoint as this will fail in the pipeline

**Comment by benjaminbluhm** [comment]
> stream/online train/update/infer/update/infer etc ?

Not exactly sure what use case you are looking for, but automatic endpoint updates are certainly possible for models hosted in the mlflow model registry. For example whenever you push a newly trained model to the mlflow model registry the endpoint can automatically use this latest model version for an incoming prediction request.

**Comment by benjaminbluhm** [comment]
> PS: unix seems to fail with this:
> 
> `ERROR: Could not install packages due to an OSError: [Errno 28] No space left on device`
> 
> does this mean we need to split the dependencies in testing because some are very large?

I added `--no-cache-dir` argument to `pip install` to save VM disk space (3f6de89dd47046924b5cfec56c98b7a40ce67aee), this seems to solve it. Let me know in case you prefer a different approach

**Comment by fkiraly** [comment]
>> stream/online train/update/infer/update/infer etc ?

> Not exactly sure what use case you are looking for,

incremental training, where you ingest more data to an already trained model, instead of replacing the old model with one that´s retrained. This is also frequently called stream inference.

**Comment by fkiraly** [comment]
> In the mlflow repo tests are only executed on ubuntu (see [here](https://github.com/mlflow/mlflow/blob/master/.github/workflows/master.yml)) so I guess we cannot expect these tests to pass without errors on other operating systems.

> Would you be okay to follow the same approach here?

> Was thinking about pytest.skip at module level

I suppose that works - we need to make sure of course that the OS string is correct.
Are you sure `mlflow` VMs are always linux?

**Comment by fkiraly** [comment]
> In case of probabilistic and variance forecasting we would still need to pass `pd.DataFrame` with column `predict_method` (required) and optionally `X: np.ndarray`, `coverage` etc.

Hm, can we perhaps use the same trick as with `fh` and pass all non-`X` argument in `fit`?

Because, in a deployment case, you typically already know what you want later on in `predict_sth_sth`.
Or do I miss sth, e.g., the case where you want to return multiple methods' return? (could use column multi-index then with col concat)

**Comment by benjaminbluhm** [comment]
> I suppose that works - we need to make sure of course that the OS string is correct (or no test will be run ever...).

Yes agree, I was trying to access code coverage to check if mlflow tests ran for `linux`, but it looks like there is an error uploading `coverage.xml` (https://github.com/sktime/sktime/actions/runs/3679210003/jobs/6225738392#step:8:38) - is there another way how I can view the code coverage? 

> Are you sure `mlflow` VMs are always linux?

Yes, all their yml files for testing specify to run jobs on only `ubuntu` (more details can be found under their github action workflows [README.md](https://github.com/mlflow/mlflow/tree/master/.github/workflows)).

**Comment by benjaminbluhm** [comment]
> incremental training, where you ingest more data to an already trained model, instead of replacing the old model with one that´s retrained. This is also frequently called stream inference (as opposed to batch inference).

This sounds interesting and a new topic to me, can you suggest any resources to learn about stream inference?

**Comment by benjaminbluhm** [comment]
> Or do I miss sth, e.g., the case where you want to return multiple methods' return?

Yes, I think we should give the user the possibility to apply different predict methods on a single fitted model. For example, user may want to use the fitted model to generate point forecast with `predict` and variance forecast with `predict_var`.  

> (could use column multi-index then with col concat)

Do you have in mind the following approach for probabilistic and variance forecasting with `pyfunc predict`? I think it could be nice as we do not have to convert the exogenous regressor to numpy before passing to `pyfunc predict`, right?

```python
import pandas as pd

# Exogenous regressor
d = {"col1": [1, 2], "col2": [3, 4]}
X = pd.DataFrame(data=d)

# Other predict parameters
d = [{"predict_method": "predict_var", "cov": False}]
params = pd.DataFrame(data=d)

# Input to pyfunc predict method
prediction_conf = pd.concat([X, params], axis=1, keys=["X", "params"])
prediction_conf
```
|    |   X |    | params |    |
|-----------------|-----------------|-----------------|--------------------------------|---------------------|
|   |   **col1** |   **col2** | **predict_method**   |   **cov** |
|               0 |               1 |               3 | predict_var                    |                   False |
|               1 |               2 |               4 | NaN                            |                 NaN |

If we want to go this route I would first like to check on json/csv conversion of MultiIndex dataframe so we make sure this also works with prediction request against API endpoint...

sktime/utils/tests/test_mlflow_sktime_model_export.py
benjaminbluhm: done

sktime/utils/mlflow_sktime.py
benjaminbluhm: Let me check. 

What would be the advantage of using `cloudpickle` or `parametrization` this?

sktime/utils/mlflow_sktime.py
```diff
@@ -0,0 +1,534 @@
...
+    ...     seasonal_order=(0, 1, 0, 12),
+    ...     suppress_warnings=True)
+    >>> forecaster.fit(y)  # doctest: +SKIP
+    ARIMA(...)
+    >>> mlflow.start_run()  # doctest: +SKIP
+    >>> artifact_path = "model"
+    >>> model_info = mlflow_sktime.log_model(  # doctest: +SKIP
+    ...     sktime_model=forecaster,
+    ...     artifact_path=artifact_path)
+    """  # noqa: E501
```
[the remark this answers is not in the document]
benjaminbluhm: correct, now after code update it is line 306, and also line 339 - any suggestion how to deal with this other than `noqa: E501`?

sktime/utils/mlflow_sktime.py
```diff
@@ -0,0 +1,534 @@
...
+
+import logging
+import os
+import pickle
+
+import pandas as pd
+import yaml
+from mlflow import pyfunc  # noqa: F401
+from mlflow.tracking._model_registry import DEFAULT_AWAIT_MAX_SLEEP_SECONDS
+from mlflow.utils.docstring_utils import LOG_MODEL_PARAM_DOCS, format_docstring
```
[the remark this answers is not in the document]
benjaminbluhm: solved by de93789b5770766e76043e0696ffe9a5bdb63e39

sktime/utils/mlflow_sktime.py
```diff
@@ -0,0 +1,534 @@
...
+    >>> from sktime.utils import mlflow_sktime
+    >>> y = load_airline()
+    >>> forecaster = ARIMA(  # doctest: +SKIP
+    ...     order=(1, 1, 0),
+    ...     seasonal_order=(0, 1, 0, 12),
+    ...     suppress_warnings=True)
+    >>> forecaster.fit(y)  # doctest: +SKIP
+    ARIMA(...)
+    >>> model_path = "model"
+    >>> mlflow_sktime.save_model(sktime_model=forecaster, path=model_path)  # doctest: +SKIP
```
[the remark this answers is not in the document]
benjaminbluhm: resolved 20dfd3335ced932de6a714f9817d06374e405fd0

fkiraly:
> Do you have in mind the following approach for probabilistic and variance forecasting with `pyfunc predict`?

No, I was thinking of `fit_conf = {"predict_method": ["predict", "predict_proba"], etc}`, and then the *output* of the inference API point is the col-concat of the `predict` and `predict_proba` output, with `keys=["predict", "predict_proba"]`. Not sure if that results in silly formatting though.

fkiraly:
> This sounds interesting and a new topic to me, can you suggest any resources to learn about stream inference?

* the `sktime` interface point `BaseForecaster.update` is an example of a stram interence interface point see also the forecasting tutorial
* here it is on wikipedia: https://en.wikipedia.org/wiki/Online_machine_learning - it applies in particular to any model that can be expressed as gradient descent mechanism, including most deep learning models. Online learning does not necessarily happen on a temporal data stream, but if it does, it is also sometimes called stream inference, especially in a neural network contet.
* The sklearn-like toolbox for that, on iid data, is the `river` toolbox: https://github.com/online-ml/river (it's explained nicely there)

I have also been imprecise with my terminology:

* as said above, online learning is the task where you update models with new data. 
* there is also online *inference*, but that commonly refers not to model updates but to low-latency serving of predictions (inference results), see here from Google https://cloud.google.com/ai-platform/prediction/docs/online-vs-batch-prediction or Microsoft https://learn.microsoft.com/en-us/azure/cloud-adoption-framework/innovate/best-practices/ml-deployment-inference cloud services
* to complete the confusion, online learning is sometimes referred to as online inference

benjaminbluhm:
> No, I was thinking of `fit_conf = {"predict_method": ["predict", "predict_proba"], etc}`, and then the output of the inference API point is the col-concat of the `predict` and `predict_proba` output, with `keys=["predict", "predict_proba"]` (done only if there are more than one method passed). Not sure if that results in silly formatting though.

Ah okay got you, sure we can do that. Just one thing to keep in mind is that we cannot return DataFrame with `MultiIndex` column structure from `pyfunc predict ` because `infer_signature` function from mlflow will only work with single index column structure (I included a unit test `test_signature_and_example_for_predict_interval_method` for that case), so after column concat we will need to flatten the column index to something like: 

`columns = ["(predict)", "(predict_interval, Coverage, 0.1, lower)", "(predict_interval, Coverage, 0.1, upper)"]`

What do you think?

benjaminbluhm:
> * online learning is sometimes referred to as online inference

Great, thank you for the nice topic overview and pointing out those resources :smiley: Would be very interested to see some real world online learning use cases in time series forecasting as I think it is still pretty common to retrain models on an entire dataset rather than incrementally, excited to read more into the above resources!

sktime/utils/mlflow_sktime.py
```diff
@@ -0,0 +1,557 @@
...
+        Either an iterable of pip requirement strings
+        (e.g. ["sktime", "-r requirements.txt", "-c constraints.txt"]) or the string
+        path to a pip requirements file on the local filesystem
+        (e.g. "requirements.txt")
+    extra_pip_requirements : Union[Iterable, str], optional (default=None)
+        Either an iterable of pip requirement strings
+        (e.g. ["pandas", "-r requirements.txt", "-c constraints.txt"]) or the string
+        path to a pip requirements file on the local filesystem
+        (e.g. "requirements.txt")
+    use_cloudpickle : boolean, optional (default=True)
```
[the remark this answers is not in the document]
aiwalter: actually I am thinking if we should this make default or not, currently we test only `pickle`, probably we should also test `cloudpickle` (but not required for this PR, only nice to have 😄 ) as it seems to be a popular package. I am not sure if all pickle-able objects are also cloudpickle-able? Also our current `save()` method can only use `pickle`, probably it should in future also support `cloudpickle` optionally.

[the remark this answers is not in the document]
benjaminbluhm: > actually I am thinking if we should this make default or not, currently we test only `pickle`, probably we should also test `cloudpickle` (but not required for this PR, only nice to have 😄 ) as it seems to be a popular package. I am not sure if all pickle-able objects are also cloudpickle-able? Also our current `save()` method can only use `pickle`, probably it should in future also support `cloudpickle` optionally.

Good point, then I suggest we make `pickle` the default until this is more thoroughly tested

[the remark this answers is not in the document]
aiwalter: then we can also isolate the `cloudpickle` dependency more, so it will only be required as installation if `serialization="cloudpickle"`

fkiraly:
re code coverage:

this is a known issue with the sktime CI - https://github.com/sktime/sktime/issues/3819
any ideas are appreciated

fkiraly:
> we will need to flatten the column index to something like:
> What do you think?

Makes sense!

`sktime` uses a different flattening convention though - it's a bit secret since column multiindex support is not official (yet?), but imo let's use that for consistency. You can find it as `sktime.utils.multiindex.flatten_multiindex`.

sktime/utils/mlflow_sktime.py
```diff
@@ -0,0 +1,631 @@
+#!/usr/bin/env python3 -u
+# -*- coding: utf-8 -*-
+"""The ``mlflow_sktime`` module provides an MLflow API for ``sktime`` models.
```
aiwalter, line 3:
```suggestion
"""The ``mlflow_sktime`` module provides an MLflow API for ``sktime`` forecasters.
```

aiwalter:
oh you mean if s.o. want to import the module `from sktime.utils import mlflow_sktime`, yes thats actually right, there might be users that might do that. So then lets just keep current name 👍🏻

benjaminbluhm:
> > we will need to flatten the column index to something like:
> > What do you think?
> 
> Makes sense!
> 
> `sktime` uses a different flattening convention though - it's a bit secret since column multiindex support is not official (yet?), but imo let's use that for consistency. You can find it as `sktime.utils.multiindex.flatten_multiindex`.

Cool, sounds good. Will start working on this tomorrow

benjaminbluhm:
Do you publish CI test results somewhere? 

Would be useful for me to check if `mlflow` tests are actually executed on `ubuntu` after adding `pytest.skip` to `sktime/utils/tests/test_mlflow_sktime_model_export.py`

aiwalter:
> Do you publish CI test results somewhere?

I think no, but actually maybe that can be possible? I meanwhile I can think of a solution to set the verbose flag in `pytest` in `setup.cfg` we would have to set `-v` flag additionally. Then we should at least see all test in the raw logs, let me try to do that quickly. The other failing test seems unrelated to this PR

aiwalter:
Idk whats heppening now, many runs were canceled. Seems pytest verbose is creating too much output?

setup.cfg
```diff
@@ -20,6 +20,7 @@ addopts =
     --showlocals
     --matrixdesign True
     -n auto
+    -v
```
aiwalter, line 23:
```suggestion
```

aiwalter (commented):
Removed verbose

benjaminbluhm:
> > Do you publish CI test results somewhere?
> 
> I think no, but actually maybe that can be possible? In meanwhile I can think of a solution to set the verbose flag in `pytest` in `setup.cfg` we would have to set `-v` flag additionally. Then we should at least see all test in the raw logs, let me try to do that quickly. The other failing test seems unrelated to this PR

Maybe something like this could be promising https://github.com/marketplace/actions/test-reporter

benjaminbluhm:
@fkiraly I updated the logic for `pyfunc predict` based on our earlier alignment. Based on the revised implementation we will only pass exogenous regressor  as `DataFrame` to `pyfunc predict`, all other predict params via attribute of type `dict` assigned to fitted model.

Example of passing predict method using default parameters with `Dict[str, list]`:

```python
{"predict_method": ["predict", "predict_interval", "predict_quantiles", "predict_var"]}
```

Example of passing predict method and parameters with `Dict[str, dict]`:

```python
{"predict_method": {"predict": None, "predict_interval": {"coverage": [0.1]}}}
```


I am posting an example workflow below to make sure we are aligned before I update the relevant tests, notebook examples, documentation etc. (in case we do not have exogenous regressor we need to pass an empty DataFrame)

```python
import mlflow
from sktime.utils import mlflow_sktime

y, X = load_longley()

pyfunc_predict_conf = {"predict_method": {"predict": None, "predict_interval": {"coverage": [0.1]}}}

model_path = "model"

with mlflow.start_run():

    forecaster = NaiveForecaster()
    forecaster.fit(y_train, X_train, fh=[1, 2, 3])
    forecaster.pyfunc_predict_conf = pyfunc_predict_conf

    mlflow_sktime.save_model(
        forecaster,
        model_path,
    )

loaded_pyfunc = mlflow_sktime.pyfunc.load_model(model_path)
predictions = loaded_pyfunc.predict(X_test)
print(predictions)
```

|      |   predict__0 |   predict_interval__Coverage__0.1__lower |   predict_interval__Coverage__0.1__upper |
|:-----|-------------:|-----------------------------------------:|-----------------------------------------:|
| 1959 |        66513 |                                  66337.2 |                                  66688.8 |
| 1960 |        66513 |                                  66264.4 |                                  66761.6 |
| 1961 |        66513 |                                  66208.5 |                                  66817.5 |

fkiraly (changes requested):
@benjaminbluhm, looks good to me!

I would recommend to make sure that in the absence of `predict_conf`, the `predict` method is called (unless it already works that way).

fkiraly:
(will approve if tests pass)

fkiraly:
some CI errors still

fkiraly:
@benjaminbluhm, any chance that this would be ready end of Dec 19? Would be a real highlight for the 0.15.0 release!

benjaminbluhm:
@fkiraly Pushed some updates, PR is ready from my end

Would be great if you could approve CI pipeline to see if all checks pass.

Updates:
- added support for `predict_proba`in pyfunc flavor
- added fallback to sktime `predict` in case no `pyfunc_predict_conf`is provided
- fixed issue with parametrizing serialization (pickle/cloudpickle)
- fixed failing unit tests
- updated example notebook and docstrings

sktime/utils/mlflow_sktime.py
```diff
@@ -0,0 +1,534 @@
...
+from sktime import utils
+from sktime.utils.validation._dependencies import _check_soft_dependencies
+
+_check_soft_dependencies("mlflow", severity="warning")
+
+FLAVOR_NAME = "mlflow_sktime"
+_MODEL_BINARY_KEY = "data"
+_MODEL_BINARY_FILE_NAME = "model.skt"
+_MODEL_TYPE_KEY = "model_type"
+PREDICT_METHODS = ["predict", "predict_interval", "predict_quantiles", "predict_var"]
```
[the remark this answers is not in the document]
benjaminbluhm: Add this point it is not trivial to add these methods to the `pyfunc` wrapper class as one would need to allow passing `y` and `X`, but one can only pass single DataFrame to `pyfunc.predict`. Perhaps it could be done by concatenating `y` and `X` but at the very least it would make the inference logic much more involved. Would suggest to keep this in mind for a potential future release.

In any case one can load the model in native sktime format via `mlflow_sktime.load_model(model_uri=model_path)` and then call `update_predict`, `update_predict_single` etc.

sktime/utils/mlflow_sktime.py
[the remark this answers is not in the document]
benjaminbluhm: Resolving this for now unless you have some more advise

sktime/utils/mlflow_sktime.py
```diff
@@ -0,0 +1,534 @@
...
+    --------
+    >>> from sktime.datasets import load_airline
+    >>> from sktime.forecasting.arima import ARIMA
+    >>> from sktime.utils import mlflow_sktime
+    >>> y = load_airline()
+    >>> forecaster = ARIMA(  # doctest: +SKIP
+    ...     order=(1, 1, 0),
+    ...     seasonal_order=(0, 1, 0, 12),
+    ...     suppress_warnings=True)
+    >>> forecaster.fit(y)  # doctest: +SKIP
```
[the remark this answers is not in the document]
benjaminbluhm: Just based on what I read in the docs:  

_If including docstring examples that use soft dependencies, ensure to skip doctest. To do this add a `# doctest: +SKIP` to the end of each line in the doctest to skip._ (https://www.sktime.org/en/stable/developer_guide/dependencies.html?#adding-a-soft-dependency)

Just let me know in case I have some misunderstanding here!

Resolving this for now

benjaminbluhm:
I would suggest to also raise feature request (once new sktime release is published) in mlflow repo to add sktime as community model flavor in mlflow docs here: https://www.mlflow.org/docs/latest/models.html#community-model-flavors

okay with you?

Then I would do that next week (will be off for couple of days from tomorrow onwards)

fkiraly:
> I would suggest to also raise feature request (once new sktime release is published) in mlflow repo to add sktime as community model flavor in mlflow docs here
> okay with you?

with me, yes, although I do not fully understand what this entails - simply a mention on their webpage? Or is there some additional commitment?
If the latter, it may have to be approved at a developer meeting.

benjaminbluhm:
> > I would suggest to also raise feature request (once new sktime release is published) in mlflow repo to add sktime as community model flavor in mlflow docs here
> > okay with you?
> 
> with me, yes, although I do not fully understand what this entails - simply a mention on their webpage? Or is there some additional commitment? If the latter, it may have to be approved at a developer meeting.

Indeed just a mention on their webpage with reference to sktime mlflow documentation and short code usage example of mlflow sktime custom flavor, no additional commitment

fkiraly:
>Indeed just a mention on their webpage with reference to sktime mlflow documentation and short code usage example of mlflow sktime custom flavor, no additional commitment

Sounds great and uncontentious to me, thanks!

aiwalter (commented):
really great work. A minor thing would be to add this to API docs, but can also be a separate PR. Would be this file where we could add a section for `mlflow`: https://github.com/sktime/sktime/blob/main/docs/source/api_reference/utils.rst

benjaminbluhm:
Seems one CI check is failing (https://github.com/sktime/sktime/actions/runs/3730166184/jobs/6328601281), but it does not seem to be related to this PR, correct me if I'm wrong

benjaminbluhm:
> really great work. A minor thing would be to add this to API docs, but can also be a separate PR. Would be this file where we could add a section for `mlflow`: https://github.com/sktime/sktime/blob/main/docs/source/api_reference/utils.rst

Thanks :blush: 

Updated API docs as suggested 9401c2b171f8b155eb8d2e41641cd440c8a8b610

sktime/utils/mlflow_sktime.py
```diff
@@ -0,0 +1,804 @@
...
+SERIALIZATION_FORMAT_CLOUDPICKLE = "cloudpickle"
+SUPPORTED_SERIALIZATION_FORMATS = [
+    SERIALIZATION_FORMAT_PICKLE,
+    SERIALIZATION_FORMAT_CLOUDPICKLE,
+]
+
+_logger = logging.getLogger(__name__)
+
+
+def get_default_pip_requirements(include_cloudpickle=False):
```
aiwalter, line 88: actually I am wondering if we can include the soft dependencies also here somehow? Is it possible to accept a `list` as input for `get_default_pip_requirements` and for `get_default_conda_env`

benjaminbluhm: not sure exactly what you mean, what would be the content of that `list`? and how does it relate to soft dependencies?

aiwalter: I thought this 2 functions are used to restore an environment on which the forecaster was trained on? So if the forecaster used a soft-dependency e.g. `pmdarima` it could also be recreated with that version of `pmdarima`? So we could optionally handover a list of soft-dependencies to store them as well, not only the `cloudpickle`?

benjaminbluhm: Ah got you, I think for this purpose we could hand over a list of requirements to `pip_requirements` argument in `save_model` and `log_model` function and `_process_pip_requirements` would then combine these additional requirements with the default requirements `sktime` and `pickle` (or `cloudpickle`)

aiwalter: yeah sth like that, but can also be done in a separate PR as it would be backwards compatible?

aiwalter (dismissed):
looks good to me, would be great to have this in next release!

benjaminbluhm:
> I would recommend to make sure that in the absence of `predict_conf`, the `predict` method is called (unless it already works that way).

Makes sense, have already implemented this as part of my earlier PR today

fkiraly (approved):
Looks good!

Resolved some merge issues in all-contributorsrc.

Issue #3913, opened 2022-12-08
[BUG] Using `update` with `SlidingWindowSplitter` for `ConformalIntervals` gives a KeyError


bethrice44:
**Describe the bug**
When using `update` with a `SlidingWindowSplitter` (or equivalent CV which misses initial data in subsequent folds), the `ConformalIntervals` wrapper throws a KeyError. 

This is because of [this line](https://github.com/sktime/sktime/blob/aea3e7a353e13f6990700c3400c232733728893d/sktime/forecasting/conformal.py#L382) which aims to avoid recalculation of parts of the `residuals_matrix` which have already been calculated. When it is used with a sliding window there is index in the initial matrix not in the new one, hence the KeyError.

I will make a PR to fix. 

**To Reproduce**

```python

from sktime.forecasting.conformal import ConformalIntervals
from sktime.forecasting.naive import NaiveForecaster
from sktime.datasets import load_airline
from sktime.forecasting.base import ForecastingHorizon
from sktime.forecasting.model_selection import SlidingWindowSplitter
from sktime.forecasting.model_evaluation import evaluate
from sktime.performance_metrics.forecasting.probabilistic import PinballLoss

y = load_airline()

cv = SlidingWindowSplitter(
            fh=np.arange(1, 13),
            window_length=48,
            step_length=12,
        )

forecaster = NaiveForecaster()

conformal_forecaster = ConformalIntervals(forecaster, initial_window=24)

results = evaluate(
        forecaster=conformal_forecaster,
        cv=cv,
        y=y,
        X=None,
        strategy="update",
        scoring=PinballLoss(alpha=[0.1, 0.5, 0.9]),
        return_data=True,
        error_score="raise",
        backend=None,
    )
```

**Expected behavior**
No KeyError and the matrix updated properly without recalculation. 

**Additional context**

**Versions**

System:
    python: 3.8.13 (default, May  8 2022, 17:48:02)  [Clang 13.1.6 (clang-1316.0.21.2)]
executable: /Users/bethrice/Documents/Code/uk-ds-day-ahead-prices-forecast/.venv/bin/python
   machine: macOS-12.1-arm64-i386-64bit

Python dependencies:
          pip: 22.1.2
   setuptools: 63.1.0
      sklearn: 1.1.2
       sktime: 0.14.0
  statsmodels: 0.13.2
        numpy: 1.22.0
        scipy: 1.8.1
       pandas: 1.4.1
   matplotlib: 3.6.0
[...]

PR #3914, opened 2022-12-08
[BUG] in `ConformalIntervals`, fix update of residuals matrix for sliding window splitter

bethrice44:
#### Reference Issues/PRs
Fixes https://github.com/sktime/sktime/issues/3913

Ensures only indices in the new residuals matrix are populated if already calculated and avoids the KeyError.

No

##### For all contributions

##### For new estimators


fkiraly (changes requested):
looks plausible! - may I request to cover this by a test?

It might be as simple as adding a parameter set with the sliding window splitter in `get_test_params`.

bethrice44:
Thanks @fkiraly  will add a test.

While thinking about the test I realised a limitation of the `_update` method. It is set up to not recalculate `residuals_matrix` rows which have already been calculated. However, because the additional rows will add new columns, you end up with the existing rows having NULL in the last x columns. This has the potential to cause issues if for example the window cv is long and you are building a lot of the matrix via the `_update` method. 

Does the problem make sense? 

I have fixed by defining an `_extend_residuals_matrix` function which uses the already fitted forecasters and extends the prediction to the length of the new `y_test`.

I have added a test to check `ConformalIntervals` and `NaiveVariance` with strategy = `refit` or `update` and `cv` =  sliding or expanding. The test threw up an erroneous assertion error in `PinballLoss` which I have also added a fix for (we use `convert_dtypes()` from pandas which converts to `Float64` instead of `float`).

sktime/performance_metrics/forecasting/probabilistic/_classes.py
```diff
@@ -256,7 +256,7 @@ def _check_consistent_input(self, y_true, y_pred, multioutput):
         if not isinstance(y_pred, pd.DataFrame):
             raise ValueError("y_pred should be a dataframe.")
 
-        if not all(y_pred.dtypes == float):
+        if not all(y_pred.dtypes == float) and not all(y_pred.dtypes == "Float64"):
```
GuzalBulatova, line 259: Should it be "or" here?

sktime/forecasting/tests/test_interval_wrappers.py
```diff
@@ -4,15 +4,24 @@
 
 __author__ = ["fkiraly"]
```
GuzalBulatova, line 5: Please add your GitHub handle here too

GuzalBulatova (commented):
Only have a couple of small comments. Great work, @bethrice44!

sktime/performance_metrics/forecasting/probabilistic/_classes.py
bethrice44: That's what I thought initially but because it is `not` it has to be `and` (I think). I tested both and `or` didn't give expected behaviour. So it only raises the error if all of the dtypes are NOT float AND all of the dtypes are NOT Float64 (OR with NOTs give weird behaviour because it will always be true that they are NOT one or the other).

fkiraly: I think it should be `not any(y_pred.dtypes not in [float, Float64])`, or the condition will erroneously fire if some columns are `Float64` and the remaining ones are `float`.

An easier way, I think, is using `is_numeric_dtype`?

bethrice44: Ok, that makes sense - will use `is_numeric_dtype`

fkiraly (approved):
Thanks, @bethrice44!

Re your questions, your handling of extending the residuals matrix look sensible to me.

At some point, we may like to factor this out more cleanly, as it is common to multiple predictive probabilistic wrappers. Would appreciate a design, we can also talk about that together with other devs.

PR #3915, opened 2022-12-09
& fkiraly [MNT] bump numpy version bound to <1.25 and fix compatibility issues #3907

### LINKED ISSUE #3907: [ENH] Loosen the restrictions on the numpy version requirement
Numpy 1.23 has been out for almost half a year now, and numpy 1.24 is on the horizon, as it is currentlly in the RC2 stage.
But sktime 0.14.1 has the requirement `numpy>=1.21.0,<1.23`.

See also:
* https://numpy.org/doc/stable/release.html
* https://github.com/numpy/numpy/releases/tag/v1.23.0
* https://github.com/numpy/numpy/releases/tag/v1.24.0rc2

aquemy:
Fixes #3907 

* moves upper bound for `numpy` to `1.25` (`1.24` is out)
* fixes some compatibility issues due to some high precision numbers in doctests (by @fkiraly)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,7 @@ requires-python = ">=3.7,<3.11"
 dependencies = [
     "deprecated>=1.2.13",
     "numba>=0.55",
-    "numpy>=1.21.0,<1.23",
+    "numpy>=1.21.0,<1.24",
     "pandas>=1.1.0,<1.6.0",
     "scikit-learn>=0.24.0,<1.2.0",
     "statsmodels>=0.12.1",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,7 @@ requires-python = ">=3.7,<3.11"
 dependencies = [
     "deprecated>=1.2.13",
     "numba>=0.55",
-    "numpy>=1.21.0,<1.24",
+    "numpy>=1.21.0,<1.25",
     "pandas>=1.1.0,<1.6.0",
     "scikit-learn>=0.24.0,<1.2.0",
     "statsmodels>=0.12.1",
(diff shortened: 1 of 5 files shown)


fkiraly:
why are these failing? Seems like failure from numerical errors? It is `numpy` related.

jorenham:
@fkiraly It looks like a classic case of round errors
https://github.com/sktime/sktime/actions/runs/3655740138/jobs/6180108575#step:7:470

fkiraly:
yes, but it did not happen before? What changed?

aquemy:
I am bumping the upper bound to 1.24 to test is this was not introduced later.

fkiraly:
yes, looks like 1.23

It also looks like it is only about the printout (number of significant digits printed), my preferred approach would be moving the tests to pytest and testing approximate equality rather than precise equality of the printout.

Would you like to do that? Otherwise I can do that too, it should not take too long.

aquemy:
I am afraid I will not have time to do it this week so feel free to take over.

jorenham:
numpy 1.24 was just realeased FYI
https://github.com/numpy/numpy/releases/tag/v1.24.0

fkiraly:
@jorenham, yes, I was going to fix this and release 0.15.0 with numpy and sklearn compatibilities.

fkiraly:
FYI @RNKuhns, the compatibility issues came from your old doctests!

Not to blame anyone, just a learning moment for me as well - looks like some numerical routines subtly changed in `numpy`, and you put a lot of digits after the comma in the expected doctest output.

I fixed that by moving the test out to `pytest` and increasing the floating point tolerance for an equality check.

PR #3920, opened 2022-12-10
[ENH] MLflow custom flavor - naive soft dependency isolation

fkiraly:
Takes @benjaminbluhm's PR https://github.com/sktime/sktime/pull/3912 and carries out a very naive soft dependency isolation.

I would like to see what happens if `pyfunc` is moved into a function and the docstring stuff is removed.

Surely nothing bad? ...

Update: indeed `pyfunc` must be at module level. How about conditional import (based on env check) then?

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1851,6 +1851,15 @@
       "contributions": [
         "build"
       ]
+    },
+    {
+      "login": "benjaminbluhm",
+      "name": "Benjamin Bluhm",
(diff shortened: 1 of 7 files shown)


fkiraly:
??? what's this?

`ERROR: Could not install packages due to an OSError: [Errno 28] No space left on device` (in the unix run)

fkiraly:
incorporated in https://github.com/sktime/sktime/pull/3912, therefore no longer needed

PR #3921, opened 2022-12-10
[BUG] `sklearn 1.2.0` compatibility - fix invalid elbow variable selection shrinkage parameter passed to `sklearn` `NearestCentroid`

fkiraly:
`ElbowClassSum` and `ElbowClassPairwise` call `sklearn`'s `NearestCentroid` with an invalid shrinkage parameter `shrink_threshold=0`.

This parameter has always been invalid, but is only being checked for validity since `1.2.0`.

To be faithful to the current implementation, I suggest to set it to a small value - an alternative option would be to expose it as a parameter in the constructor, and set it to a small default.

I'm not sure whether `shrink_threshold=None` corresponds to zero mathematically.

@haskarb, what do you think?

### DIFF
--- a/sktime/transformations/panel/channel_selection.py
+++ b/sktime/transformations/panel/channel_selection.py
@@ -230,7 +230,7 @@ class ElbowClassSum(BaseTransformer):
         t = self.distance_
 
         start = int(round(time.time() * 1000))
-        centroid_obj = _shrunk_centroid(0)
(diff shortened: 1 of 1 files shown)


haskarb:
I think it should be fine. Not much use of shrink parameter. I have to take a relook at the code in few days, needs to push new work here; currently wrapping my internship.

PR #3922, opened 2022-12-10
[MNT] Update sklearn compatibility to `1.2.x`, version bound to `<1.3`

fkiraly:
This PR updates the `sklearn` version bound to `<1.3`.

Includes fixes for `sklearn 1.2` compatibility:

* https://github.com/sktime/sktime/pull/3918
* https://github.com/sktime/sktime/pull/3919
* https://github.com/sktime/sktime/pull/3921
* https://github.com/sktime/sktime/pull/3923
* https://github.com/sktime/sktime/pull/3924
* https://github.com/sktime/sktime/pull/3926

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -157,6 +157,15 @@
       ]
     },
     {
+      "login": "nilesh05apr",
+      "name": "Nilesh Kumar",
+      "avatar_url": "https://avatars.githubusercontent.com/u/65773314?v=4",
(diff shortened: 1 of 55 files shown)


fkiraly:
There seems to be one remaining issue, within time series forest.

This is a classifier that "hacks" sklearn and uses many of its private functions.
We could fix it, or deprecate it.

Another option would be to introduce package version bounds on estimators.

PR #3924, opened 2022-12-11
[MNT] `sklearn 1.2.0` compatibility - remove `normalize=False` args from `RidgeClassifierCV`

fkiraly:
This PR removes superfluous `normalize=False` args from `RidgeClassifierCV`.

The arg has been deprecated in `sklearn 1.2.0`, but was default `normalize=False` prior.

Simply removing the arg makes the occurrences compatible with `sklearn 1.2.0`.

Further, this PR removes use of the `normalize` argument from one of the secondary notebooks.

### DIFF
--- a/examples/rocket.ipynb
+++ b/examples/rocket.ipynb
@@ -149,7 +149,12 @@
    "outputs": [
     {
      "data": {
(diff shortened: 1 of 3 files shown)


patrickzib (dismissed):

patrickzib:
Thank you, looks good. I cannot approve, as I do not have write access to sktime, yet.

PR #3931, opened 2022-12-13
[ENH] Add multiplicative option to `Detrender`

KishManani:
This PR adds a `model` argument to `Detrender`, similar to `Deseasonalizer`, which gives the option to divide by the trend component rather than subtract. 

Example usage:
```python
from sktime.transformations.series.detrend import Detrender

y = load_airline()
detrender = Detrender(forecaster, model="multiplicative")

# Divide by the trend component
y_transformed = detrender.fit_transform(y)
```

If the extracted trend has a zero value at any point, it would result in an `np.inf` value when we divide by the trend. What should the expected behaviour be in this case? Throw an error or return an output with an infinite value? 

##### For all contributions

### DIFF
--- a/sktime/transformations/series/detrend/_detrend.py
+++ b/sktime/transformations/series/detrend/_detrend.py
@@ -4,7 +4,7 @@
 """Implements transformations to detrend a time series."""
 
 __all__ = ["Detrender"]
-__author__ = ["mloning", "SveaMeyer13"]
+__author__ = ["mloning", "SveaMeyer13", "KishManani"]
 
 import pandas as pd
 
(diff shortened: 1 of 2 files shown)


sktime/transformations/series/detrend/_detrend.py
```diff
@@ -38,6 +38,9 @@ class Detrender(BaseTransformer):
         The forecasting model to remove the trend with
             (e.g. PolynomialTrendForecaster).
         If forecaster is None, PolynomialTrendForecaster(degree=1) is used.
+    model : {"additive", "multiplicative"}, default="additive"
```
fkiraly, line 41: could you add more detail on this?

I.e., explain that the forecaster is fitted on the original data etc?

fkiraly (changes requested):
Great, thanks!

Could you kindly explain better in the docstring what "multiplicative" means?

sktime/transformations/series/detrend/_detrend.py
KishManani: Of course! I've added more detail. Let me know what you think!

fkiraly (approved):
Looks good! The detrender is not "owned" as far as I can see, so good to go.

fkiraly:
PS: I think `inf` is fine in the "divide by zero" case, that's the expected output.

PR #3932, opened 2022-12-13
[ENH] TSFeatures adaptor

chillerobscuro:
#### Reference Issues/PRs
Closes [Issue-3712](https://github.com/sktime/sktime/issues/3712)

A wrapper for [TSfeatures](https://pypi.org/project/tsfeatures/) library which computes "features used by Montero-Manso, Talagala, Hyndman and Athanasopoulos in [their implementation of the FFORMA model](https://htmlpreview.github.io/?https://github.com/robjhyndman/M4metalearning/blob/master/docs/M4_methodology.html#features)"

TSFeatures package mentioned above

#### This is a draft PR

Because this transformer does not yet pass tests or `check_estimator()`, because each test case requires a `freq` argument that matches the seasonal period of the data, which is not currently supported. Here are some comments from Franz on this issue: 
```
    - currently, the interface contract assumes all parameters work for all data, with potential restrictions through tags
    - there is a somewhat obscure way to give scenarios and estimators tags and then define criteria on “does estimator match the test case”
    - practically, there is a baseline measure if tests don’t pass and it’s not a “true” failure, or one that is hard to fix:
        add exceptions in the tests/_config.py
        open an issue that describes the problem clearly, so it can be fixed later
        link issue number in the comment in tests/_config, next to the exception
    - general design question: parameter validity depending on data
        do we want to allow this at all? Or consider it an interface violation?
            another example: ColumnTransformer referring to column names in the data set that is expected
            this will work only for data sets that actually have those column names
     - if we allow this, what is the interface
     - how to deal with this in the tests/scenarios framework
```

So I'm leaving this here as a progress checkpoint until there is a clear path forward that we're happy with, ie adding many tests to the "test exceptions" in config. The logic for the transformation is simple and just requires getting the data into a 3 column df in the right format. This works for both single and multi series inputs, as long as you pass the matching `freq`. For example:
```
from sktime.datasets import load_airline, load_longley
from sktime.transformations.panel.tsfeatures import TSFeaturesExtractor

y = load_airline()
z = load_longley()[1]

t = TSFeaturesExtractor()
xt = t.fit_transform(y, freq=12)
# or
xt = t.fit_transform(z, freq=1)
```

##### For new estimators

### DIFF
--- a/sktime/transformations/panel/tsfeatures.py
+++ b/sktime/transformations/panel/tsfeatures.py
@@ -1,0 +1,176 @@
+# -*- coding: utf-8 -*-
+"""tsfeatures interface class."""
+# copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+
+__author__ = ["chillerobscuro"]
+__all__ = ["TSFeaturesExtractor"]
+
+# -*- coding: utf-8 -*-
+# copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+# todo: write an informative docstring for the file or module, remove the above
+# todo: add an appropriate copyright notice for your estimator
+#       estimators contributed to sktime should have the copyright notice at the top
+#       estimators of your own do not need to have permissive or BSD-3 copyright
+
+from tsfeatures import tsfeatures
+
+from sktime.transformations.base import BaseTransformer
+from sktime.utils.validation._dependencies import _check_soft_dependencies
+
+_check_soft_dependencies("tsfeatures", severity="error")
+
+
+class TSFeaturesExtractor(BaseTransformer):
+    """Custom transformer. todo: write docstring.
+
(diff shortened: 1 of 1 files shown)


fkiraly (commented):
failure is due to missing soft dependency handling, see here:

chillerobscuro:
@fkiraly thanks for pointing me to that doc, the dependency should be handled now. Now seeing these errors:
```
FAILED sktime/forecasting/model_evaluation/tests/test_evaluate.py::test_evaluate_error_score[threading-update-False-nan] - Failed: DID NOT WARN. No warnings of type (<class 'sktime.exceptions.FitFailedWarning'>,) were emitted.
```
even though this transformer doesn't have a `_fit()` method implemented.

PR #3937, opened 2022-12-14
[GOV] Remove ability of CC to overrule core dev decisions

aiwalter:
Reference Issues/PRs
https://github.com/sktime/sktime/issues/3865

Remove ability of CC to overrule decisions made by core devs.

### DIFF
--- a/docs/source/get_involved/governance.rst
+++ b/docs/source/get_involved/governance.rst
@@ -447,8 +447,6 @@ tracker <https://github.com/sktime/sktime/issues>`__,
 `pull requests <https://github.com/sktime/sktime/pulls>`__ or an :ref:`steps`. Some
 sensitive discussions and appointment votes occur on private chats.
 
-The CC reserves the right to overrule decisions.
-
 We distinguish between the following types of proposed changes. The
 corresponding decision making process is described in more detail below.
 


aiwalter:
a core dev vote was invoked on this PR and will end 21. December

GuzalBulatova (changes requested):
I think that CC should have rights to overrule core-devs decisions, given the consensus/ majority of CC agree that another decision has to be made and provide reasoning behind the overrule.

fkiraly:
This PR (and its predecessor) has violated governance rules on how to vote multiple times.

* gov rules clearly say, we vote publicly. Martin opened votes in private.
* Martin has used some app to vote anonymously in a private slack channel that had people in it who were not eligible to vote.
* this is a substantial change so would formally require a STEP, which implies 7 days discussion period.

Further, we are violating the "spirit" of how to vote.

* some people are now on holiday and will be unable to vote until they return.
* some people wanted to pause from votes after this and will not see the second vote in time.

I think we should just stop this and base the changes on consensus.

I.e., start over in January when everyone is back from holidays or pause.
Then, work out how elections work with term limits and clear rules.

patrickzib (approved):

PR #3942, opened 2022-12-15
[BUG] Fix `StatsForecastAutoARIMA_.predict` incorrect in-sample start index

tianjiqx:
#### Reference Issues/PRs
When I use the StatsForecastAutoARIMA algorithm to forecast the past, the forecast is always returned from the start time of the training, even though fh specifies to start at a moment in the middle of the training time range.

I think the reason is that the index used is incorrect, the fh passed in is not used correctly, and the past results of the prediction are always returned from the start time of the training. We should use from_cutoff to true. I did this and the final result was as I expected.
image

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/forecasting/base/adapters/_statsforecast.py
+++ b/sktime/forecasting/base/adapters/_statsforecast.py
@@ -131,7 +131,7 @@ class _StatsForecastAdapter(BaseForecaster):
         """
         # initialize return objects
         fh_abs = fh.to_absolute(self.cutoff).to_numpy()
-        fh_idx = fh.to_indexer(self.cutoff, from_cutoff=False)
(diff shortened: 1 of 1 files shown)


fkiraly (approved):
Looks like a straightforward fix, thanks!

@FedericoGarza is the owner so should have a look at this.

FYI @FedericoGarza, I think this is a bugfix and not a substantial change to the content. Would appreciate a comment.
We'll merge in a week or so if you do not reply; if you reply later but think this is a substantial change, we will unmerge when you comment.

FedericoGarza:
hey @tianjiqx, @fkiraly! LGTM :)

PR #3916, opened 2022-12-09
&fkiraly [ENH] `Imputer`: docstring clarity improvement, conditional parameter handling logic

aiwalter:
updated content and description by @fkiraly

Minor improvements to `Imputer`:

* updates to the docstring to clarify conditionnality of parameters
* various corrections to the docstring
* change to parameter check logic so setting of conditional parameters is allowed when they are ignored, e.g., setting `forecaster` if `method` is not `"forecaster"`, and tests for this

---

#### What does this implement/fix? Explain your changes.
This changes the exception for `method="constant"` to always accept a `value` that is not `None`. This is required to make grid search over the `method` param and at same time give a `value` value.

PR #3951, opened 2022-12-18
Release v0.15.0 

fkiraly:
Release PR for 0.15.0.

* version number bump
* changelog


aiwalter (changes requested):
Pls include #3916

fkiraly:
I left comments on #3916 which you have not yet replied to.

ltsaprounis (changes requested):
#3975 was merged with change requests pending. The introduced mlflow dependencies set is specific to AWS because of the mock in the tests. This would mean that a user who would install sktime with `pip install sktime[mlfow]` would involuntarily install a lot of AWS things they may not need. We should find a more straightforward way of dealing with our mlflow test dependencies.

fkiraly:
@ltsaprounis, I think that's a valid point. Would this https://github.com/sktime/sktime/pull/3979 be an acceptable solution?
I summarized the different requirements we need to satisfy there imo.

An important requirement is "tests continue running", the last one, which implies afaik some kind of dependency isolation away from `dev` or `all_extras`.

fkiraly:
Decision by me as release manager:

* #3916 is not ready, it has open PR change requests to address. Not a reason to block the release.
* #3979 decision is resolved and explained in the PR

With this, no blocking requests to this PR.

PR #3952, opened 2022-12-19
[MNT] 0.15.0 deprecation actions

fkiraly:
Deprecation actions for the 0.15.0 release:

* changed `check_is_scitype` parameter `msg_legacy_interface` to `False`
* `ExpandingWindowSplitter` had `start_with_window` argument removed. From now on, `initial_window=0` should be used instead of `start_with_window=False`. Contains bugfix https://github.com/sktime/sktime/pull/3953 as the deprecation logic was incomplete.
* Test `test_methods_do_not_change_state` renamed to `test_non_state_changing_method_contract`
* row transformers have been removed, replaced by base class vectorization/broadcasting
* changed default for `default_fc_parameters` in `tsfresh` transformers to follow the default of `tsfresh`, which currently is `comprehensive`
* removed module `transformations.series.compose`. All contents were moved to `transformations.compose`.

Delayed to 0.16.0:
* base class of `ProbabilityThresholdEarlyClassifier` change to `BaseEarlyClassifier`. Unclear what to do here, deprecation action specified breaks the interface compliance tests. Attempt to fix here https://github.com/sktime/sktime/pull/3962 but fails `fit_idempotent` test (unclear why, will drop for this release).
* change of forecaster `cutoff` attribute to `pd.Index` type


sktime/classification/early_classification/base.py
```diff
@@ -353,7 +352,6 @@ def split_indices_and_filter(X, indices, decisions):
         inv_dec = np.invert(decisions)
         return X[inv_dec], indices[inv_dec], indices[decisions]
 
-    @abstractmethod
```
miraep8, line 356: So - I suppose this may have already come up in the original depreciation discussion, in which case if consensus was reached can disregard my comment, but I personally really like using abstract methods.  Reasoning - for developers if forces you to actually write the `_fit` and `_predict` functions.  I think for anyone unfamiliar with the templating design pattern, the fact that the code literally breaks if you don't write a `_fit` makes it easy to figure out that that is where the logic should live. 

I understand that there are also downsides to requiring `abc` though, but just wanted to put that out there - for the type of design we do it is a pretty nice tool.

fkiraly: for the PR at hand: it breaks the deprecation action, but since I removed the deprecation action (was broken/incomplete), we do not need to remove `abc` in here. I erroneously only reverted the estimator file, not the base class file. I will revert the base class file.

for the general design point: yes, I do appreciate the clarity `abc` brings, but the drawback (which I've already encountered) is that inspection becomes much more difficult due to the added decorators (these are decorators, so you may have to unwrap to inspect!)
this has caused problems in inspection heavy use cases such as `reset` like methods (in `BaseObject` work), or work around the polymorphic base class type, although it is currently not breaking anything on `main`.

There is an open issue on the general design point if you want to chip in: https://github.com/sktime/sktime/issues/1923

miraep8 (commented):
Was the depreciation also meant to extend to `start_with_window` in `SlidingWindowSplitter` as well?  (Or should we perhaps introduce a depreciation for this?).  Just thinking that that splitter also has an initial window/could make sense to expand up to `window_length` then begin sliding?

fkiraly:
>Was the depreciation also meant to extend to `start_with_window` in `SlidingWindowSplitter` as well?

It wasn't, since we have not raised a deprecation warning for it.

I have also seen the same thing while going through the code, this could have been done at the same time (but then again, the `ExpandingWindowSplitter` was not deprecated correctly).

Probably worth opening an issue on this and discussing whether the parameter is also redundant in `SlidingWindowSplitter`. I think it is, by virtue of the same argument (if `False` just set the other parameter to 0).

miraep8 (approved):
Thank goodness we got those `OptionalPassthrough` imports in the correct order.

fkiraly:
> Thank goodness we got those OptionalPassthrough imports in the correct order.

Well, the location moved, and with it the import's position in the lexicographical order.

Issue #3961, opened 2022-12-20
[BUG] EnsembleForecaster fails to fit with exogenous data on hierarchical data


shagn:
**Describe the bug**
After upgrading sktime from version 0.12 to 0.14 the fitting of an EnsembleForecaster on a hierarchical dataset fails when receiving exogenous data. 
Without exogeneous data the fitting works without problems. 
It fails with exogeneous data if the models are using the exogeneous data or not. 

**To Reproduce**
```python
import pandas as pd

from sktime.utils._testing.hierarchical import _make_hierarchical
from sktime.forecasting.compose import EnsembleForecaster
from sktime.forecasting.naive import NaiveForecaster
from sktime.forecasting.trend import PolynomialTrendForecaster

data = _make_hierarchical(
    hierarchy_levels = (2, 4),
    n_columns = 2,
    min_timepoints = 38,
    max_timepoints = 38, 
    index_type = "period",
)

target_col = "c0"
y_train = data.loc[:, [target_col]]
X_train =  data.drop(columns=[target_col])

forecasters = [    
    ("trend", PolynomialTrendForecaster()),
    ("naive", NaiveForecaster(strategy="last", sp=12)),
]
forecaster = EnsembleForecaster(forecasters=forecasters)

forecaster.fit(y=y_train, X=X_train)
```
![image](https://user-images.githubusercontent.com/16029092/208664020-95342334-997a-4993-9bb9-cf639dce546f.png)

**Expected behavior**
The EnsembleForecaster works also with exogenous data. 

**Additional context**
None

**Versions**
System:
    python: 3.8.13 (default, Mar 28 2022, 11:38:47)  [GCC 7.5.0]
executable: /anaconda/envs/.../bin/python
   machine: Linux-5.15.0-1022-azure-x86_64-with-glibc2.17

Python dependencies:
          pip: 22.1.2
   setuptools: 65.5.0
      sklearn: 1.1.1
       sktime: 0.14.0
  statsmodels: 0.13.2
        numpy: 1.21.5
        scipy: 1.7.3
       pandas: 1.4.4
   matplotlib: 3.6.2
       joblib: 1.1.0
        numba: 0.56.4
     pmdarima: None
      tsfresh: None

fkiraly:
Bug confirmed on python 3.8 and windows.

I also found the reason: both vectorization/broadcasting and the heterogeneous ensemble base class use the attribute `forecaster_` which leads to a recursion accudent.

[...]

PR #3969, opened 2022-12-20
[ENH] add contract test for hierarchical forecasting

fkiraly:
This PR adds a new contract test to `TestAllForecasters` which tests forecasting with hierarchical endogeneous and exogeneous data, for that it works and returns a forecast of expected mtype and format.

This test should also cover the failure case in https://github.com/sktime/sktime/issues/3961.

Contains fixes for further bugs detected through this:

* https://github.com/sktime/sktime/pull/4001


fkiraly:
@ciaran-g, the failures of the `ReconcilerForecaster` look genuine, do you have an idea what is going on?

ciaran-g:
Hi @fkiraly, just got an email notification on this PR - is there a possible bug here in the `ReconcilerForecaster`?

fkiraly:
Having thought of it, I think it's not a bug - the `ReconcilerForecaster` adds more indices after the forecast (if `__total` levels are not present), so the forecast contains more hierarchy levels than the training data.

The check in the test was just too strict.

What is your opinion, @ciaran-g?

ciaran-g:
> Having thought of it, I think it's not a bug - the `ReconcilerForecaster` adds more indices after the forecast (if `__total` levels are not present), so the forecast contains more hierarchy levels than the training data.
> 
> The check in the test was just too strict.

I see what you mean, yep the index tests should be relaxed then for this case.

However, looking into the code last night I think there might be a related bug in the `ReconcilerForecaster`. When it isn't used in a pipeline with the `Aggregator` or the `X` and `y` values aren't aggregated before predict or update are called. Seems like the class is missing a pre-processing step in the `.predict` and `.update` functions

```py

if X is not None:
  if _check_index_no_total(X):
      X = self._add_totals(X)
```

About to make a PR

ciaran-g:
#3980

aiwalter:
this PR is violating the PR rules defined here: https://www.sktime.org/en/stable/get_involved/governance.html#decision-making.

fkiraly:
@aiwalter, early merge is possible as per the rules and standing policy.

Do you want to vote it down and revert it?

Issue #3670, opened 2022-10-30
[BUG] `statsmodels` interfaces break for `X` containing more indices than forecasting horizon


fkiraly:
Same as #3657, but for `SARIMAX` and `statsmodels` adapter when `X` is present.

Also see #3667 and #3668.

The obvious fix, in https://github.com/sktime/sktime/pull/3669, seems to fail on python 3.10, and not on 3.9 or lower.

GianlucaGuarroSonos:
I am using python version 3.8.13 and run into this issue (for both statsmodel and pmdarima)

fkiraly:
Any good guesses why #3669 fails on 3.10 but not on 3.9 or lower?

fkiraly:
I think this is now fixed by #3972.

PR #3669, opened 2022-10-29
DRAFT: `statsmodels` interface fix for `X` strictly containing `y`

fkiraly:
parallel fix for `statsmodels` adapter to https://github.com/sktime/sktime/pull/3667

Strangely, this fails inside `TransformedTargetForecaster` for some reason. Unclear what is going on there.

PR #3972, opened 2022-12-21
[BUG] fix `statsmodels` estimators when exogenous `X` is passed with more indices than `fh`

### LINKED ISSUE #3830: [BUG] `model_evaluation.evaluate` fails for `statsmodels` forecasters with exogenous features
**Describe the bug**

model_evaluation.evaluate creates X_test containing both train and test samples.  For statsmodels models with exogenous features X_test length must equal the forecast horizon and therefore this currently throws the error:

`ValueError: Provided exogenous values are not of the appropriate shape.`

This bug occurs for UnobservedComponents and SARIMAX.

The root cause of the bug is in the `_split` function:
https://github.com/sktime/sktime/blob/a5f824a5fff1c3fc1babd3f34b94c5055ec040b0/sktime/forecasting/model_evaluation/_functions.py#L60

This could be modified to 'X.iloc[test].sort_index()' for statsmodels models with exogenous features or alternatively by filtering X here:
https://github.com/sktime/sktime/blob/a5f824a5fff1c3fc1babd3f34b94c5055ec040b0/sktime/forecasting/base/adapters/_statsmodels.py#L104

**To Reproduce**

```python
from sktime.forecasting.structural import UnobservedComponents
from sktime.forecasting.model_selection import SlidingWindowSplitter

y, X = load_longley()

forecaster = UnobservedComponents(level='local linear trend')

cv = SlidingWindowSplitter(fh=[1,2,3],
                           window_length = 4,
                           step_length = 1)

results = evaluate(
        forecaster=forecaster, y=y, X = X,
        cv=cv, strategy='refit', return_data=True,
        error_score = 'raise'
        )
```

**Versions**

System:
python: 3.8.12 (default, Nov 20 2022, 10:50:06)  [Clang 14.0.0 (clang-1400.0.29.202)]
machine: macOS-13.0.1-arm64-arm-64bit

Python dependencies:
          pip: 21.2
   setuptools: 63.1.0
      sklearn: 1.1.3
       sktime: 0.14.0
  statsmodels: 0.13.5
        numpy: 1.22.0
        scipy: 1.9.3
       pandas: 1.4.1
   matplotlib: 3.6.2
       joblib: 1.2.0
        numba: 0.56.4
     pmdarima: None
      tsfresh: None

adoherty21:
#### Reference Issues/PRs

Fixes [#3830](https://github.com/sktime/sktime/issues/3830)

Statsmodels with exogenous features would raise an exception in `predict` if `X` contained more indices than the range of `fh`.
This was due to all indices being passed through to `statsmodels` `predict`.  The fix ensures that only the expected `X` indices from the test set are passed to `statsmodels` `predict`.

Fixes #3830 as this was also the root cause of the exception raised when using `evaluate` with `statsmodels` based estimators.

No

##### For all contributions

##### For new estimators


sktime/forecasting/base/adapters/_statsmodels.py
```diff
@@ -100,6 +100,13 @@ def _predict(self, fh, X=None):
         # beginning of the training series when passing integers
         start, end = fh.to_absolute_int(self._y.index[0], self.cutoff)[[0, -1]]
 
+        # bug fix for evaluate function as test_plus_train indices are passed
+        # statsmodels exog must contain test indices only.
+        # For discussion see https://github.com/sktime/sktime/issues/3830
+        if X is not None:
+            ind_drop = self._X.index
+            X = X.loc[~X.index.isin(ind_drop)]
```
aiwalter, line 108: not sure if that is correct bcz how will that work for in-sample predictions? Then you would drop the in-sample `X` values here also? Can you make sure that in-sample prediction still works with exog data?

aiwalter:
might be better to fix it in `_split` as you mentioned in the issue. Not sure why the in `X.iloc[test_plus_train].sort_index()` the train data is appended to the test data. You would have to check if all tests are passing and ideally check why this was implemented like that. Might be that some forecasters need both as input to make predictions?

sktime/forecasting/base/adapters/_statsmodels.py
adoherty21: Thanks for the quick response @aiwalter.  I wasn't considering the requirement for in-sample predictions.  Good catch.  I'll review the code and my changes with this in mind and respond.

Regarding your comments:

> might be better to fix it in `_split` as you mentioned in the issue. 

I'll look into whether it is possible to fix in `_split`.  

> Not sure why the in `X.iloc[test_plus_train].sort_index()` the train data is appended to the test data. You would have to check if all tests are passing and ideally check why this was implemented like that. Might be that some forecasters need both as input to make predictions?

You are right, some transformers require `train_plus_test` to be passed for prediction.  @miraep8 responded to my issue and mentioned that this is related to [#2880](https://github.com/sktime/sktime/issues/2880).  Do you have any thoughts on this?

fkiraly:
@adoherty21, this is the underlying issue imo: https://github.com/sktime/sktime/issues/3670

fkiraly:
my attempted fix is here: https://github.com/sktime/sktime/pull/3669
I do not understand why the fix does not work, if you want to have a look

adoherty21:
Good evening and apologies for the delay in responding.

@aiwalter I checked and this modification does not impact in-sample predictions as statsmodels does not require `exog` to predict in-sample.  In the case where we pass an empty data frame due to dropping the train indices or `None` the underlying statsmodels implementation still produces the correct result.

@fkiraly I have done some investigating and believe that the reason your fix in #3669 does not work is due to statsmodels requiring the entire range of `X_test` up to the final forecast horizon and not just `fh_abs`.  I have made a modification in the latest commit of this PR which I think fixes this.  Would you be able to have a quick look and let me know your thoughts?  I have tested locally and this passes using python 3.8.12. 

Note, in this PR I am only modifying `_predict`.  The other changes in #3669 to `_update` and `_fit`  could be made after this PR is merged.

fkiraly:
> is due to statsmodels requiring the entire range of `X_test` up to the final forecast horizon and not just `fh_abs`

ahhhhh, that makes sense! Thanks a lot for this!

Your linting is failing, here's the guide on how to ensure your code formatting is right:
https://www.sktime.org/en/stable/developer_guide/coding_standards.html

adoherty21:
No problem at all and thanks for the guide @fkiraly.  All tests are passing now.

fkiraly (approved):
Thanks for the fix!

* somehow the formatting in the all contributors file got messed up. I fixed it.
* Have you looked at the `pmdarima` issue, too? Might be the same fix, contributions would be appreciated if it is indeed the same problem (and the same remedy), see https://github.com/sktime/sktime/issues/3657

adoherty21:
Thanks for sorting the contributor file @fkiraly and merging.  

I will happily review the `pmdarima` issue.

PR #3975, opened 2022-12-21
[MNT] isolate `mlflow` and AWS dependency set

fkiraly:
This PR makes two changes:

1. cleans up the dependency set from https://github.com/sktime/sktime/pull/3912 - `mlflow` and AWS dependencies are moved to a separate dependency set `mlflow`.
The CI for `mlflow` is also separated and isolated from the remaining tests, ensuring that the remaining tests, or a user who wants `all_extras` do not need to install a large python env or packages they do not want (e.g., AWS specific ones).

2. removes duplicate API reference from `utils`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -102,7 +102,7 @@ jobs:
 
       - name: Install sktime and dependencies
         run: |
-          python -m pip install .[dev,mlflow] --no-cache-dir
+          python -m pip install .[all_extras,dev,mlflow] --no-cache-dir
 
       - name: Show dependencies
         run: python -m pip list
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -102,13 +102,13 @@ jobs:
 
       - name: Install sktime and dependencies
         run: |
-          python -m pip install .[dev,mlflow]
+          python -m pip install .[dev,mlflow] --no-cache-dir
 
       - name: Show dependencies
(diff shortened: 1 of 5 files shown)


aiwalter (changes requested):
I think only mlflow is needed to run the code. Others are test related?

fkiraly:
FYI @benjaminbluhm

fkiraly:
>I think only mlflow is needed to run the code. Others are test related?

If you want to use mlflow on AWS, I think you need those.
In the CI it is treated as being a test related dependency set - there is no doc that encourages the user to install `sktime[mlflow]`

aiwalter:
Well we dont need to provide installs for particular platform. Also AWS will have this installed by default then.

fkiraly:
>Well we dont need to provide installs for particular platform. Also AWS will have this installed by default then.

Can you please phrase your comments as a change request?

aiwalter:
I think this PR is not required, so change reqeust would be to close it?😊

fkiraly:
can you please provide an explanation why it is not required?
This is a supplement to #3912.

fkiraly:
If you block this, I will revert the `mlflow` PR, otherwise we overload the `all_extras` environment with large packages.

aiwalter:
There was only added mlflow

fkiraly:
I'm interpreting this as an unblock, then.

PR #3979, opened 2022-12-22
[MNT] separate dependency set for AWS based `mlflow` tests

fkiraly:
This PR addresses comments in https://github.com/sktime/sktime/pull/3951 and https://github.com/sktime/sktime/pull/3975 related to handling of mlflow test dependencies.

I think we have heard some valid arguments:

* no user of `sktime` should be forced to install AWS dependencies, as there are other major cloud providers like MS and Google (with which `mlflow` is also compatible). That means the plain, `dev` (power user) and `all_extras` dependency sets
* putting the AWS dependencies used in the tests in the `mlflow` dep set is misleading, as a user could think this is for `mlflow` compatibility - point made by @ltsaprounis.
* we should isolate the `mlflow` AWS deps in the tests as they cause a large python environment and are not used anywhere else in the test suite, leaving them in can cause test and memory issues as we have seen.

Proposed solution in this PR:
* `mlflow` dep set contains only `mlflow`
* `mlflow_test` dep set contains `mlflow` and the AWS deps, this is used in the tests only and for isolation of the AWS deps

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -102,7 +102,7 @@ jobs:
 
       - name: Install sktime and dependencies
         run: |
-          python -m pip install .[all_extras,dev,mlflow] --no-cache-dir
+          python -m pip install .[all_extras,dev,mlflow_tests] --no-cache-dir
 
       - name: Show dependencies
         run: python -m pip list
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,6 +56,7 @@ dependencies = [
 [project.optional-dependencies]
 all_extras = [
     "cloudpickle",
+    "dask",
     "dtw-python",
     "esig==0.9.7; python_version < '3.10'",
     "filterpy>=1.4.5",
@@ -79,7 +80,6 @@ all_extras = [
     "tsfresh>=0.17.0; python_version < '3.10'",
     "tslearn>=0.5.2",
     "xarray",
-    "dask",
 ]
 
 dev = [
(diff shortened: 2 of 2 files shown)


benjaminbluhm:
> Due to the last requirement - test isolation and memory issues with the CI: if we cannot come to an agreement here, it would mean we have to unmerge @benjaminbluhm's excellent PR and not have it in 0.15.0 (and delay it to 0.16.0) .

I cannot really contribute before next week as I do not have a laptop with me, but if the AWS dependency in the unit tests cause a lot of trouble we can also delete the respective test without any harm. I just took this test from one of the built-in model test modules but I think we could also safely remove it given we also do not have any equivalent tests for Azure, GCP etc.

In my opinion PR should not be undone as a result of AWS dependency challenge.

fkiraly:
> In my opinion PR should not be reverted as a result of AWS dependency challenge.

Well, I do share that opinion.

I just wanted to get it in quickly for the release since I suspected you were off, so I thought I was going to handle the remaining dependency issue myself.

fkiraly:
Update:

* Martin and Leonidas have been very active on the internal slack today, up to a point in the discussion where I asked them to provide their preferred handling of this situation (i.e., names of dependency sets, what goes in there, CI jobs and environments).
* after that very active point in the discussion, the internal discussion on the topic dropped entirely silent, while Martin has continued to be active (but not reactive). That was about 4-5 hours ago. I have also not seen anything on the public GitHub channels.

Given this, I will assume that that it is not so important that we need to make changes right now, and decide, as the current release manager, that it is preferable to keep @benjaminbluhm's original PR in the release above delaying it due to this discussion.

This PR satisfies requirements that were brought up, including what was brouht up, as well as avoiding to bloat the user dependency sets.
As we are not substantially modifying the current key dependency sets, we can always make changes after the next release without too much of a deprecation burden.

Issue #3815, opened 2022-11-18
[BUG] ReconcilerForecaster does not support exogenous data out of the box


RikStarmans:
**Describe the bug**

If the sktime example is adapted to a forecaster with exogenous data it fails.

**To Reproduce**

```python
from sktime.forecasting.model_selection import temporal_train_test_split
from sktime.forecasting.exp_smoothing import ExponentialSmoothing
from sktime.transformations.hierarchical.reconcile import Reconciler
from sktime.transformations.hierarchical.aggregate import Aggregator
from sktime.utils._testing.hierarchical import _bottom_hier_datagen
from sktime.transformations.series.date import TempAgg
import numpy as np
fh = 7
y = _bottom_hier_datagen(
    no_bottom_nodes=3,
    no_levels=1,
    random_seed=123,
)
y_train, y_test = temporal_train_test_split(y, test_size=fh)
# first we show how it does work
f0 = Aggregator()*Aggregator()**ReconcilerForecaster(SARIMAX(), method='td_fcst')
f0.fit(y=y_train, X=y_train)
prds = f0.predict(fh=list(range(1, fh+1)), X=y_test)
# now a failure is shown
f1 = ReconcilerForecaster(SARIMAX(), method='td_fcst')
f1.fit(y=y_train, X=y_train)
# following fails
prds = f1.predict(fh=list(range(1, fh+1)), X=y_test)
```
Error
```python
   1749 if methodname != "update_predict_single":
-> 1750     y_preds += [method(X=Xs[i], **kwargs)]
   1751 else:
   1752     y_preds += [method(y=ys[ix], X=Xs[i], **kwargs)]

IndexError: list index out of range
```

**Expected behavior**

No error.

**Additional context**

**Versions**
<details>

</details>

Issue #2875, opened 2022-06-26
[ENH] make internal interface of `ReconcilerForecaster` consistent and visible in docstring


fkiraly:
left over from #2830:

In `ReconcilerForecaster`:
* what `self.forecaster` is is not consistent across the methods. One time, it is the fitted forecaster, one time it is the entire pipeline. I would ensure it is the pipeline in both cases, i.e., either bind it to the fittable pipeline element, or execute the pipeline manually.
* The `self.forecaster_` attribute should be mentioned in the docstring.

PR #3980, opened 2022-12-22
[BUG] `ReconcilerForecaster` bug in predict and update

ciaran-g:
#### Reference Issues/PRs

No open issue on this as far as I can tell

There is a bug in the `ReconcilerForecaster` predict and update functions when the input data is not used in a pipeline with `Aggregator` or the input data is not aggregated before using with the class.

To reproduce

```py
from sktime.datatypes._utilities import get_window
from sktime.forecasting.sarimax import SARIMAX
from sktime.forecasting.reconcile import ReconcilerForecaster

y = _make_hierarchical(
    hierarchy_levels=(2, 4),
    n_columns=1,
    min_timepoints=24,
    max_timepoints=24,
    index_type="period",
)
y_train = get_window(y, lag=2)
y_test = get_window(y, window_length=2)

X = _make_hierarchical(
    hierarchy_levels=(2, 4),
    n_columns=2,
    min_timepoints=24,
    max_timepoints=24,
    index_type="period",
)
X.columns = ["foo", "bar"]
X_train = get_window(X, lag=2)
X_test = get_window(X, window_length=2)

forecaster = SARIMAX()
estimator_instance = ReconcilerForecaster(forecaster, method="mint_shrink")
fh = [1, 2]
# fit works ok
estimator_instance.fit(y=y_train, X=X_train, fh=fh)
# no aggregation in X so it fails
estimator_instance.predict(X=X_test)
# no aggregation in X or y so it fails again
estimator_instance.update(y=y_test, X=X_test)
```

Changes

- Added necessary steps in the functions for automatically aggregating the data. 
- Also formatted the doc string a bit, but I would appreciate it if you can check that this is the correct way

No


fkiraly:
Thanks a lot!

I think this bug *has* already been reported by @rikstarmans, here?
https://github.com/sktime/sktime/issues/3815
Sounds plausible at least that your fix and the bug report are related.

Also, have you ween this?
https://github.com/sktime/sktime/issues/2875

fkiraly (changes requested):
I am a bit concerned manipulating `self._y` etc, since it couples the inner functions with the boilerplate and may hence cause problems down the line - so I would prefer to avoid this.

Is there a way we can avoid manipulating `self._y` this way? (I think we need to resolve the question, could also be "no change needed")

Otherwise, looks fine to me.

I also added the failure case as a test, in a variant of one and multiple variables in `y`.

ciaran-g:
Hi @fkiraly - sorry for the delay with this, I've been off. Just updated with a final commit to try and fix the docs formatting finally.

> I think this bug _has_ already been reported by @RikStarmans, here?
> #3815
> Sounds plausible at least that your fix and the bug report are related.

Whoops, hadn't seen that - aye that's the bug! Tested that this PR does fix it as well.

> Also, have you seen this?
> #2875

I have done, hopefully open a separate PR on that soon.

> I am a bit concerned manipulating `self._y` etc, since it couples the inner functions with the boilerplate and may hence cause problems down the line - so I would prefer to avoid this.
> 
> Is there a way we can avoid manipulating `self._y` this way? (I think we need to resolve the question, could also be "no change needed")

I was also hesitant to change the `self._y`, but I couldn't really think of a quick way around that since the data is already updated (with no-aggregation) through the base class. Happy to have a go if you have any ideas?

> I also added the failure case as a test, exactly the code you posted, in a variant of one and multiple variables in `y`.

Thanks for putting in the test. Quick (maybe daft) question on that: do the pytests not need to end with some sort of assertion?

aiwalter:
@ciaran-g question: if you add the totals in `fit`, are they then also dropped before returning the predictions? I think they should in that case so everything is happening under the hood? otherwise user can still add them in a pipeline in order to keep totals in predictions? Or at least we should parametrize this?

aiwalter:
E.g. the following example can not be run in any grid search or `evaluate` so would be nice if this is then fixed in `ReconcileForecaster` to apply the `Aggregator` and `Reconciler` just under the hood.

```python
from sktime.forecasting.reconcile import ReconcilerForecaster
from sktime.utils._testing.hierarchical import _bottom_hier_datagen
from sktime.transformations.hierarchical.reconcile import Reconciler

y = _bottom_hier_datagen(
    no_bottom_nodes=3,
    no_levels=1,
    random_seed=123,
)

forecaster = TransformedTargetForecaster(steps=[
    ("agg", Aggregator()),
    ("forecaster", NaiveForecaster()),
    ("rec", Reconciler()),
]
)
cv = ExpandingWindowSplitter(
    initial_window=10,
    step_length=2,
    fh=[1],
)

evaluate(cv=cv, forecaster=forecaster, y=y, error_score="raise")
```

fkiraly:
> Thanks for putting in the test. Quick (maybe daft) question on that: do the pytests not need to end with some sort of assertion?

Not necessarily - any piece of code functions as a test for "it runs".

fkiraly:
> I was also hesitant to change the `self._y`, but I couldn't really think of a quick way around that since the data is already updated (with no-aggregation) through the base class. Happy to have a go if you have any ideas?

Here's how I would do it: use a different variable, e.g., `self._y_total`, and don't use `copy` (so it's a referece and saves memory). You may like to add a test that the variable behaves as expected.

ciaran-g:
> @ciaran-g question: if you add the totals in `fit`, are they then also dropped before returning the predictions? I think they should in that case so everything is happening under the hood? otherwise user can still add them in a pipeline in order to keep totals in predictions? Or at least we should parametrize this?

At the moment the predictions for the aggregate levels are returned even if they are not supplied in fit. I did think of excluding them in that case and I'm still undecided of the best way! Obviously keeping the totals in makes the forecaster consistent with it's predictions no matter the input data and is more typically the use case for the `ReconcilerForecaster`, i.e. returning aggregate and bottom node forecasts. Excluding them when they're not supplied in fit keeps it consistent with the input data.  I'll leave it up to you both to say what is the higher priority 👍 

> Here's how I would do it: use a different variable, e.g., `self._y_total`, and don't use `copy` (so it's a referece and saves memory). You may like to add a test that the variable behaves as expected.

Ok cool, I assumed that would mess with things like getting residuals etc. but will look at it

fkiraly:
> Ok cool, I assumed that would mess with things like getting residuals etc. but will look at it

Well, I thought what you are currently doing might mess with residuals etc.
But you might have a point then.

The general point I'm worried about (and which I think needs discussion and resolution, not necessarily the way I propose) is that the "inner" machinery seems jerry-rigged to the "outer" machinery, so you can no longer independently change them and rely on the contracts.

ciaran-g:
I think there are three options when `y` or `X` have no aggregation, i.e. no `__total` indexes. Default aggregation is carried out under the hood and
 1. the predictions returned include aggregate levels. (my preference) 
 2. the predictions returned only include the indexes in `X` and `y`.
 3. we add an function argument to control this behaviour. e.g. `ReconcilerForecaster(base_forecaster, method="mint_shrink", retain_aggregate_if_no_totals=True)`

The reason I prefer 1 is because it would be my expectation for a reconciler. If you are only interested in the bottom nodes then why not use a simple base forecaster for all the bottom nodes? 

The downside is that index_in != index_out and if you want to use evaluate or grid search as they are at the moment, as @aiwalter showed you would have to use as `Aggregator` as a pre-processing step like

```python

from sktime.forecasting.sarimax import SARIMAX
from sktime.forecasting.reconcile import ReconcilerForecaster
from sktime.utils._testing.hierarchical import _bottom_hier_datagen

base_forecaster = SARIMAX()
forecaster = ReconcilerForecaster(base_forecaster, method="mint_shrink")
agg = Aggregator()

y = _bottom_hier_datagen(
    no_bottom_nodes=3,
    no_levels=1,
    random_seed=123,
)
y_total = agg.fit_transform(y)

cv = ExpandingWindowSplitter(
    initial_window=10,
    step_length=2,
    fh=[1],
)
evaluate(cv=cv, forecaster=forecaster, y=y_total, error_score="raise")
```

ciaran-g:
> The general point I'm worried about (and which I think needs discussion and resolution, not necessarily the way I propose) is that the "inner" machinery seems jerry-rigged to the "outer" machinery, so you can no longer independently change them and rely on the contracts.

I agree, i just don't know about the inner machinery enough to make any sensible suggestions lol. Anyway, let me have a root around and I'll get back to you

ciaran-g:
@fkiraly, just updated with a really simple fix for not changing `self._y` etc. here lol: just update the `self.residuals_` data that is defined in fit

fkiraly (approved):
Oh, of course!
Well caught.

Issue #3986, opened 2022-12-23
[BUG] ForecastX is returning wrong predictions


aiwalter:
**Describe the bug**
`ForecastX` returns weird predictions, when I recreate the composition manually I get different results, see below.

**To Reproduce**

```python
from sktime.datasets import load_longley
from sktime.forecasting.arima import ARIMA
from sktime.forecasting.base import ForecastingHorizon
from sktime.forecasting.compose import ForecastX
from sktime.forecasting.var import VAR
from sktime.forecasting.model_selection import temporal_train_test_split
import pandas as pd

y, X = load_longley()
y_train, y_test, X_train, X_test = temporal_train_test_split(y, X, test_size=3)
fh = ForecastingHorizon([1, 2, 3])
columns=["ARMED", "POP"]

# ForecastX
pipe = ForecastX(  
    forecaster_X=VAR(),
    forecaster_y=ARIMA(),
    columns=columns,
)
pipe = pipe.fit(y_train, X=X_train, fh=fh) 
# dropping ["ARMED", "POP"] as those are the columns where we expect not to have future values
y_pred = pipe.predict(fh=fh, X=X_test.drop(columns=columns))
y_pred.to_frame()
```
<html>
<body>

  | 0
-- | --
-285606.156664
-276594.299226
-267325.354023

</body>
</html>

**Expected behavior**

Expected would be following, recreated manually:
```python
# fit y forecaster
arima = ARIMA().fit(y_train, X=X_train)

# fit and predict X forecaster
var = VAR()
var.fit(X_train[columns])
var_pred = var.predict(fh)

# predict y forecaster with predictions from VAR
X_pred = pd.concat([X_test.drop(columns=columns), var_pred], axis=1)
arima.predict(fh=fh, X=X_pred).to_frame()
```
<html>
<body>

  | 0
-- | --
70085.541848
70073.339809
72665.367272

</body>
</html>

This makes more sense if we look into what plain `ARIMA` is forecasting:
```python
arima = ARIMA()
arima.fit(y=y_train, X=X_train, fh=fh)
arima.predict(X=X_test).to_frame()
```
<html>
<body>

  | 0
-- | --
70372.810378
70873.888217
73678.534405

</body>
</html>

**Additional context**
[...]

PR #3987, opened 2022-12-23
[BUG] fix `ForecastX` predictions

fkiraly:
Fixes https://github.com/sktime/sktime/issues/3986 and includes consistency tests.

The reason for the bug is that columns may end up permuted, and the wrapped forecaster may or may not be able to cope with that. The fix is ensuring that the `X` in predict always has columns in the same order as in `fit`.

Probably, we should ensure that all estimators can deal with permuted columns by coercing in the base class, but that's a separate topic.

### DIFF
--- a/sktime/forecasting/compose/_pipeline.py
+++ b/sktime/forecasting/compose/_pipeline.py
@@ -1264,6 +1264,10 @@ class ForecastX(BaseForecaster):
         if X is not None:
             X_pred = X_pred.combine_first(X)
 
+        # order columns so they are in the same order as in X seen
+        X_cols_ordered = [col for col in self._X.columns if col in X_pred.columns]
+        X_pred = X_pred[X_cols_ordered]
+
         return X_pred
 
     def _predict(self, fh=None, X=None):
--- a/sktime/forecasting/compose/tests/test_pipeline.py
+++ b/sktime/forecasting/compose/tests/test_pipeline.py
@@ -456,13 +456,13 @@ def test_forecastx_logic():
     columns = ["ARMED", "POP"]
 
     # ForecastX
-    pipe = ForecastX(  
+    pipe = ForecastX(
         forecaster_X=VAR(),
         forecaster_y=ARIMA(),
         columns=columns,
     )
-    pipe = pipe.fit(y_train, X=X_train, fh=fh) 
-    # dropping ["ARMED", "POP"] as those are the columns where we expect not to have future values
+    pipe = pipe.fit(y_train, X=X_train, fh=fh)
+    # dropping ["ARMED", "POP"] = columns where we expect not to have future values
     y_pred = pipe.predict(fh=fh, X=X_test.drop(columns=columns))
 
     # comparison case: manual execution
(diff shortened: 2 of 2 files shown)


sktime/forecasting/compose/tests/test_pipeline.py
```diff
@@ -436,3 +436,48 @@ def test_subset_getitem():
...
+    var = VAR()
+    var.fit(X_train[columns])
+    var_pred = var.predict(fh)
+
+    # predict y forecaster with predictions from VAR
+    X_pred = pd.concat([X_test.drop(columns=columns), var_pred], axis=1)
+    y_pred_manual = arima.predict(fh=fh, X=X_pred)
+
+    # compare that test and comparison case results are equal
+    assert np.allclose(y_pred, y_pred_manual)
```
aiwalter, line 483: I think we could even use `assert_series_equal` or `assert_frame_equal`?

aiwalter: ah not here sry, I think you are mixing sth up here in the test?

fkiraly: does that really matter? I don't mind

sktime/forecasting/compose/_pipeline.py
```diff
@@ -1264,6 +1264,10 @@ def _get_forecaster_X_prediction(self, X=None, fh=None, method="predict"):
         if X is not None:
             X_pred = X_pred.combine_first(X)
```
aiwalter, line 1265: I think here is another issue, its assuming that the specified columns are `NaN`, if not it will not overwrite the values with the predictions. So this is then making issues in `evalute` and therefore in the grid searches as well. Happy to make a PR if you dont want to fix it? Its a separate bug of `ForecastX` I think?

fkiraly: sure, although I'm not sure what the bug is or what the condition is under which it appears

fkiraly: I remember checking which way round it is, and I do think it prioritizes observations over predictions. I think that's the right way round? Since observations are more "real" than predictions.

fkiraly: I don't think this is a problem of `ForecastX` because it's expected behaviour. Instead, `evaluate` should censor the variables that are not known in advance when feeding them to the estimator.

aiwalter: We could have a param to handle this in `ForecastX` or change the logic (my preferred solution), currently its not possible to backtest this and therefore its not possible to test this in grid searches as it will always use the real observations if I am not wrong?

fkiraly: > We could have a param to handle this in

why not - the default should be "observation overwrites forecast" imo as that is user expectation.

However, guessing what is your train of thought: `evaluate` should not know about the estimators it is getting, or you get a highly coupled design.

To address your problem, I would instead suggest to add a "censored variables" or "temporal variables" option and use that in preparing your `X` and `y` in `evaluate`.

PR #3989, opened 2022-12-23
[BUG] Fix wrong tag in FeatureSelection and add tests

aiwalter:
A tag was set wrongly in `FeatureSelection` as it can also return `None` in `transform` if `method="none"`.
Added unit tests for `FeatureSelection`

No

### DIFF
--- a/sktime/transformations/series/feature_selection.py
+++ b/sktime/transformations/series/feature_selection.py
@@ -73,7 +73,7 @@ class FeatureSelection(BaseTransformer):
     _tags = {
         "scitype:transform-input": "Series",
         # what is the scitype of X: Series, or Panel
-        "scitype:transform-output": ["Series", "Primitives"],
+        "scitype:transform-output": "Series",
         # what scitype is returned: Primitives, Series, Panel
         "scitype:instancewise": True,  # is this an instance-wise transform?
         "X_inner_mtype": ["pd.DataFrame", "pd.Series"],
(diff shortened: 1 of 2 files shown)


fkiraly (changes requested):
I think the tag is currently set up that it can return only one scitype. This will surely break things.

Can you give an example where what is returned is not a `Series`?
In the empty case, should be an empty container then.

aiwalter:
The custom tests were passing but might be that the generic tests are failing. If we do `FeatureSelection(method="none")` then its basically dropping all features by setting `X=None` which is a `Primitive`?

aiwalter:
Example:
```python
from sktime.transformations.series.feature_selection import FeatureSelection

y, X= load_longley()
y_train, y_test, X_train, X_test = temporal_train_test_split(y, X, test_size=3)
t = FeatureSelection(method='none')
t.fit(X_train)
X_hat = t.transform(X_test)
X_hat is None
```

fkiraly (changes requested):
Can you please describe what `FeatureSelection` will do?

Given what condition on parameters and inputs, what output will be returned?

It should be described in the docstring at least, and I'm not sure whether it makes sense at all to have an output polymorphic transformer.

aiwalter:
closing due to #4153

fkiraly:
I still haven't understood what this was supposed to be doing, addressing, or why, so I suppose we can leave this one closed.

PR #3991, opened 2022-12-24
[ENH] speed up mtype check for `pandas` based mtypes with `pd.PeriodIndex`

### LINKED ISSUE #3990: [ENH] Improve space and time performance during `transform` when using `PeriodIndex` for time related transformers
The `DateTimeFeatures` and `TimeSince` transformer are a lot slower and takes up a lot more memory than expected when working with pandas Periods. I think this is due to some of the type checking happening in the `transform` method rather than the implementation in the `_transform` method. On large datasets (e.g., M5 dataset) my kernel surprisingly dies when using `DateTimeFeatures`. 

Here is an example to show the issue:
 ```python
import pandas as pd
import numpy as np

# Create time series with datetime index
ts_datetime = pd.DataFrame(data={"y":np.random.rand(50000)},
                        index=pd.date_range(start="2000-01-01", freq="D", periods=50000))

# Create time series with period index
ts_period = pd.DataFrame(data={"y":np.random.rand(50000)},
                        index=pd.period_range(start="2000-01-01", freq="D", periods=50000))

# Instantiate transformers
date_time_transformer = DateTimeFeatures(manual_selection=[
    "day_of_month",
    "month_of_year",
    "day_of_week",
])

time_since_transformer = TimeSince()

# Select timeseries with datetime or period index
ts = ts_period

# Fit transformers
date_time_transformer.fit(ts)
time_since_transformer.fit(ts)

# Time the runs
%%timeit
results = time_since_transformer.transform(ts)
# 3.66 s ± 199 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
results = time_since_transformer._transform(ts)
# 184 ms ± 4.48 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
result = date_time_transformer.transform(ts)
# 3.41 s ± 131 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
result = date_time_transformer._transform(ts)
# 24 ms ± 9.31 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

The time difference between running `transform` and `_transform` when the time series has a `PeriodIndex` are large (e.g., for `DateTimeFeatures` there is over a factor of 100 in the execution time).  

This issue does not seem to occur when the index is a `DatetimeIndex`. 

**Version**
System:
executable: /Users/kishan_manani/.pyenv/versions/3.8.7/envs/udemy-ts/bin/python3.8

Python dependencies:
          pip: 22.3.1
   setuptools: 49.2.1
      sklearn: 1.1.3
       sktime: 0.15.0
  statsmodels: 0.13.2
        numpy: 1.21.6
        scipy: 1.8.1
       pandas: 1.4.4
   matplotlib: 3.6.2
       joblib: 1.2.0
        numba: 0.56.4
     pmdarima: 1.8.5
      tsfresh: 0.19.0

### PR DESCRIPTION
**Pull request by fkiraly**
This PR fixes #3990; related PR: #3827. FYI @KishManani, @danbartl.

Combining information from @danbartl and @KishManani, I understood where the issue must have been coming from - from a line of `np.diff` on the checked index, in `_index_equally_spaced`, but only in the case of `pd.PeriodIndex` (the latter condition I only understood from @KishManani's analysis).

I have used this insight to replace the time and memory consuming check by a much quicker one, exploiting mathematical assumptions on the `pd.PeriodIndex` (unique and sorted index).

While this PR seems to address the issue, some open questions remain:

* do we want to test this fix, or in general, checks being fast and not memory intensive? If yes, how? Runtimes can vary even on VM.
* is @danbartl's fix in https://github.com/sktime/sktime/pull/3935 redundant or complementary? It seems to address hierarchical and panel types primarily, so I could imagine it either makes no difference, or helps speed up things if the series check is fast.

### DIFF
--- a/sktime/datatypes/_series/_check.py
+++ b/sktime/datatypes/_series/_check.py
@@ -245,7 +245,7 @@ def _index_equally_spaced(index):
         return False
 
     # if we arrive at this stage, and the index is PeriodIndex,
-    # we know it must be equally spaced:
+    # we know it must be equally spaced and sorted:
     # it cannot have duplicates, and by the pigeonhole principle, the
     # two necessary conditions we checked are also sufficient
     if isinstance(index, pd.PeriodIndex):
--- a/sktime/datatypes/_series/_check.py
+++ b/sktime/datatypes/_series/_check.py
@@ -225,14 +225,35 @@ def _index_equally_spaced(index):
     if not is_in_valid_index_types(index):
         raise TypeError(f"index must be one of {VALID_INDEX_TYPES} or integer index")
 
-    # empty and single element indices are equally spaced
-    if len(index) < 2:
+    # empty, single and two-element indices are equally spaced
+    if len(index) < 3:
         return True
 
     # RangeIndex is always equally spaced
     if isinstance(index, pd.RangeIndex):
         return True
 
+    # we now treat a necessary condition for being equally spaced:
+    # the first two spaces are equal. From now on, we know this.
+    if index[1] - index[0] != index[2] - index[1]:
+        return False
+
+    # another necessary condition for equally spaced:
+    # index span is number of spaces times first space
+    n = len(index)
+    if index[n - 1] - index[0] != (n - 1) * (index[1] - index[0]):
+        return False
+
+    # if we arrive at this stage, and the index is PeriodIndex,
+    # we know it must be equally spaced:
+    # it cannot have duplicates, and by the pigeonhole principle, the
+    # two necessary conditions we checked are also sufficient
+    if isinstance(index, pd.PeriodIndex):
+        return True
+
+    # fallback for all other cases:
+    # in general, we need to compute all differences and check explicitly
+    # CAVEAT: this has a comparabily long runtime and memory usage
     diffs = np.diff(index)
     all_equal = np.all(diffs == diffs[0])
 
--- a/sktime/datatypes/_series/_check.py
+++ b/sktime/datatypes/_series/_check.py
@@ -245,8 +245,9 @@ def _index_equally_spaced(index):
         return False
 
     # if we arrive at this stage, and the index is PeriodIndex,
-    # we know it must be equally spaced and sorted:
-    # it cannot have duplicates, and by the pigeonhole principle, the
+    # we know it must be equally:
+    # it cannot have duplicates and must be sorted (other conditions for mtype),
+    # therefore, by the pigeonhole principle, the
     # two necessary conditions we checked are also sufficient
     if isinstance(index, pd.PeriodIndex):
         return True
--- a/sktime/datatypes/_series/_check.py
+++ b/sktime/datatypes/_series/_check.py
@@ -245,7 +245,7 @@ def _index_equally_spaced(index):
         return False
 
     # if we arrive at this stage, and the index is PeriodIndex,
-    # we know it must be equally:
+    # we know it must be equally spaced:
     # it cannot have duplicates and must be sorted (other conditions for mtype),
     # therefore, by the pigeonhole principle, the
     # two necessary conditions we checked are also sufficient
--- a/sktime/datatypes/_series/_check.py
+++ b/sktime/datatypes/_series/_check.py
@@ -254,7 +254,7 @@ def _index_equally_spaced(index):
 
     # fallback for all other cases:
     # in general, we need to compute all differences and check explicitly
-    # CAVEAT: this has a comparabily long runtime and memory usage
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by KishManani** [comment]
Hi @fkiraly! I'm unsure whether your fix renders @danbartl's fix unnecessary. When applying the `DateTimeFeatures` transformer to a large panel time series I still find that the current implementation causes my kernel to crash. Here is an example: 

```python
import pandas as pd
import numpy as np

# Create large panel dataset
df = _make_hierarchical(hierarchy_levels=(100,), min_timepoints=10000, max_timepoints=10000)

# Prepare transformer
date_time_transformer = DateTimeFeatures(manual_selection=[
    "day_of_month",
    "month_of_year",
    "day_of_week",
])

date_time_transformer.fit(df)

# Standard implementation
%%timeit
results = date_time_transformer.transform(df)
# 6.43 s ± 215 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)


# Naive implementation which iterates over
# each timeseries. 
%%timeit
results = []
for tsid, ts in df.groupby(level=0, observed=True):
    results.append(date_time_transformer.transform(ts.loc[tsid]))
result = pd.concat(results, axis=0)
result.index = df.index
# 1.37 s ± 19.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

I'm unsure whether the fix implemented in https://github.com/sktime/sktime/pull/3935 resolves this though.

fkiraly:
@KishManani, thanks for the added info!

Can you clarify, does it crash on `main`, or on this branch?

KishManani:
I've only tested it on this branch. It crashes on this branch when I'm using a subset of the M5 dataset. I generated the above as a smaller example which can at least run.

fkiraly:
Re kernel crash, that indicates a memory issue rather than a runtime issue?

KishManani:
After merging @danbartl's fix there was no difference in the runtime with the above. @fkiraly you are correct that I'm conflating the memory issue with the runtime issue. I need to run more experiments. But here is the result of a pre-liminary one:

image

`df` was created from `_make_hierarchical(hierarchy_levels=(500,), min_timepoints=10000, max_timepoints=10000)`. So it is a dataframe with 5,000,000 rows made of 500 time series. `df2` is from the M5, it has 4,800,000 rows and has 3048 time series. The number of columns in both dataframes is one. For some reason `df2` which has fewer rows but many more time series takes *a lot* longer to run.

KishManani:
The results are easier to see here where I generate a dataframe with the same number of rows but made with different number of time series.

```Python
import pandas as pd
import numpy as np

# Create large panel dataset
df = _make_hierarchical(hierarchy_levels=(50,), min_timepoints=1000, max_timepoints=1000)
df_ = _make_hierarchical(hierarchy_levels=(500,), min_timepoints=100, max_timepoints=100)

# Prepare transformer
date_time_transformer = DateTimeFeatures(manual_selection=[
    "day_of_month",
    "month_of_year",
    "day_of_week",
])

date_time_transformer.fit(df)

# Small number of long time series
%time
results = date_time_transformer.transform(df)
# CPU times: user 1.5 s, sys: 53 ms, total: 1.56 s
# Wall time: 1.56 s

# Large number of small time series
%time
results = date_time_transformer.transform(df)
# CPU times: user 13.7 s, sys: 62.5 ms, total: 13.7 s
# Wall time: 14.6 s
```

I guess this behaviour is to be expected because the number of times we apply `transform` scales linearly with the number of time series. It's just a shame here as for datetime features we could compute them a lot faster if we could just directly compute them from the index in one go rather than looping over each time series.

KishManani:
@fkiraly and @danbartl - if for hierarchical and panel data we expect the last level in the multi-index to be the time column, is it worth considering computing the `datetime` features directly from this rather than generalising from the behaviour that `DateTimeFeatures` has for the `Series`  scitype?

fkiraly:
>  It's just a shame here as for datetime features we could compute them a lot faster if we could just directly compute them from the index in one go rather than looping over each time series.

Well, currently the framework just applies a stupid loop - there are two opportunities to improve the runtime:

* for a given transformer, the developer can implement internal logic for multi-indexed (hierarchical) data frames. For that, `y_inner_mtype` and `X_inner_mtype` need to include the hierarchical types, these are then passed directly to `_transform` instead of to the (fallback) loop in the base class.
* the broadcasting in `VectorizedDF`, could be made smarter - from what is currently a loop, there could be "actual" vectorization logic, or parallelization.

PR #3996, opened 2022-12-25
[ENH] Change `DateTimeFeatures` trafo to work with multi-index data and add option to drop columns.

KishManani:
#### Reference Issues/PRs
https://github.com/sktime/sktime/pull/3991#issuecomment-1364593497

Changes the `transform` method in `DateTimeFeatures` to handle dataframes with a multi-index (e.g., panel and hierarchical data). The datetime features are now computed directly from the time index of the multi-index. This reduces the runtime significantly when using `DateTimeFeatures` with panel or hierarchical data; I measured a factor of 10 improvement.

A `keep_original_columns` argument is added which gives the option to keep or drop the columns in `X` passed to `.transform`. 

`transform` has also been refactored to avoid making a copy of `X`  to make it more memory efficient. 

See https://github.com/sktime/sktime/pull/3996#issuecomment-1364725220 for more context.

No.

#### Any other comments?
 The `transform` method is significantly slower than using `_transform` when working with hierarchical data - I assume that there is some other machinery converting between types under the hood causing this? Some examples with hierarchical data below:

```Python
import numpy as np
import pandas as pd

# Make hierarchical data
df = _make_hierarchical(hierarchy_levels=(50, 2, 2), min_timepoints=1000, max_timepoints=1000)

trafo = DateTimeFeatures()
trafo.fit(df)

%%time
result = trafo.transform(df)
# CPU times: user 1.55 s, sys: 88.8 ms, total: 1.64 s
# Wall time: 1.69 s

%%timeit
result = trafo._transform(df)
# CPU times: user 64.6 ms, sys: 9.08 ms, total: 73.6 ms
# Wall time: 74.2 ms
```

##### For all contributions

##### For new estimators


fkiraly (changes requested):
Thanks, great idea!

There's no need to have two transformers, and it would be confusing to the user.

If you have `X_inner_mtype` being `["pd.DataFrame", "pd-multiindex", "pd_multiindex_hier"]`, you will receive a `pd.DataFrame` in `_fit` and `_transform` and can then make case distinctions based on that.

I.e., do `groupby` in case of `MultiIndex`.

KishManani:
Thanks @fkiraly! I agree with you, I just wasn't sure the best way to achieve this with sktime. Are there any helper methods to check if a given input has an mtype of "pd-multiindex", "pd_multiindex_hier"?

KishManani:
I can actually just use` if isinstance(X.index, pandas.MultiIndex)`.

KishManani:
@fkiraly, I've refactored the PR and made the code changes in the original transformer in the `transformations.series` subpackage. I see the speed boost which is nice. However, I still see a large time difference between `transform` vs `_transform` when working with hierarchical dataframes. Do you have any idea why this may be? Example code below:

```Python
import numpy as np
import pandas as pd

# Make hierarchical data
df = _make_hierarchical(hierarchy_levels=(50, 2, 2), min_timepoints=1000, max_timepoints=1000)

trafo = DateTimeFeatures()
trafo.fit(df)

%%time
result = trafo.transform(df)
# CPU times: user 2.1 s, sys: 85.2 ms, total: 2.19 s
# Wall time: 2.26 s

%%time
result = trafo._transform(df)
# CPU times: user 62.6 ms, sys: 6.65 ms, total: 69.2 ms
# Wall time: 67.8 ms
```

KishManani:
Whilst I'm working on this. I've got another question for @fkiraly and @danbartl. In the current implementation we make a copy of the transformed dataframe for the purpose of concatting the original features at the end. I would like to make a proposal to add an option to *not return the input features* and set this as the default behaviour. The benefits are:

1. In a pipeline where you just want to extract the date related features you don't have to be concerned about the other features being accidentally passed through. Currently I have issues when making a pipeline with `DateTimeFeatures` because it passes through the original time series column. 
2. We don't need to copy the original dataframe which saves memory. This is because the only purpose of making the original copy is to append it to the derived features. If we create an option to drop the original features then for that option we can remove the need for copying the dataframe (i.e., we only create the copy when `if keep_original_columns == True`).
3. Not keeping the input columns is a preferable default behaviour because we can always passthrough columns in a pipeline by using `Id()`. 

What do you both think?

fkiraly:
> However, I still see a large time difference between `transform` vs `_transform` when working with hierarchical dataframes. Do you have any idea why this may be? Example code below:

Which branch? With or without the recent fix(es)?

fkiraly:
@KishManani, agreed basically with everything you say, in-principle.

> I would like to make a proposal to add an option to *not return the input features* and set this as the default behaviour. 

agreed with this - we will need to deprecate though. I.e., (1) add the option, but leave default as pass through and raise a warning; (2) switch the option to don´t pass.

> 2. We don't need to copy the original dataframe which saves memory. 

I believe you are right.
We have a test for side effects which should catch any.

fkiraly:
> this is because both `FourierFeatures` and `DateTimeFeatures` don't drop the original features

Maybe we should do the same there?

KishManani:
> agreed with this - we will need to deprecate though. I.e., (1) add the option, but leave default as pass through and raise a warning; (2) switch the option to don´t pass.

I've added the `keep_original_columns` argument, amended the tests, and added the warning.

> I believe you are right.
We have a test for side effects which should catch any.

I've now removed `.copy()` and instead reference the index of the original dataframe `X`. Could you point me to mentioned test that checks for side effects?

KishManani:
> Which branch? With or without the recent fix(es)?

Without fixes.

fkiraly:
> Could you point me to mentioned test that checks for side effects?

The test name string is `"test_methods_have_no_side_effects"`, it is automatically run in `check_estimator` on any relevant method of the estimator.

fkiraly:
>> Which branch? With or without the recent fix(es)?

> Without fixes.

Then I'd hope it will be resolved once the fixes are merged to `main`?

KishManani:
> Then I'd hope it will be resolved once the fixes are merged to main?

Hopefully? 🤞 If not, I'll raise a separate issue.

fkiraly:
may I kindly ask you to update the PR summary to match the content now, for when @danbartl reads it? (he wrote the original transformer so needs to approve)

danbartl:
Hi thanks for the great work, I'll hope to read it today or tomorrow

KishManani:
Thanks both! I've updated the PR summary now.

fkiraly (changes requested):
Excellent!

I only have a minor (but important) comment: you need to exactly adhere to the deprecation policy, i.e., leave the todo message with versions 0.17.0 in it (for change), or it will likely be missed by the release manager:

KishManani:
>I only have a minor (but important) comment: you need to exactly adhere to the deprecation policy, i.e., leave the todo message with versions 0.17.0 in it (for change), or it will likely be missed by the release manager:

Hi @fkiraly! Thanks for pointing me to this. I've tried to make the change. Could you let me know if it is correct? Thank you!

sktime/transformations/series/date.py
```diff
@@ -33,6 +34,15 @@
 ]
 
 
+# TODO: remove in v0.17.0
```
fkiraly, line 37: the message to the *user* was already ok before this.

You need to write a message to the *developer* that details precisely what needs to be changed when 0.17.0 is released.
Currently, it is wrong, since:
* `deprecated` will raise the message whenever the `DateTimeFeatures` transformer is used, indicating that it will be deprecated as a whole! What you had before, warning in case of parameter set to `True`, was good!
* in the `todo` comment, you should exactly describe what the release manager needs to do - I assume, flip the parameter to `False` and remove the warning? Anything else?

fkiraly (changes requested):
Ok, done - almost there!

See above for a comment on what needs to change.
Let me know if you want help or a quick chat.

danbartl:
@fkiraly @KishManani great changes, thanks a lot and good from my side!

KishManani:
Hi @danbartl! Do you think the following on line 229 is needed?

```Python
        if self.manual_selection is not None:
            df = df[self.manual_selection]
```

On the if block starting on line 194 there is a check that when `self.manual_selection` is not `None` there already appears to be a filter on line 220 and 225 to ensure the calculation is only computed on the subset of options found in `self.manual_selection`. I have commented on this block and checked that the tests still do pass. What do you think? Can we get rid of this?

sktime/transformations/series/date.py
```diff
@@ -154,14 +178,17 @@ def _transform(self, X, y=None):
...
+            time_index = X.index.get_level_values(-1)
+        else:
+            time_index = X.index
+
+        x_df = pd.DataFrame(index=X.index)
+        if isinstance(time_index, pd.PeriodIndex):
+            x_df["date_sequence"] = time_index.to_timestamp().astype("datetime64[ns]")
+        elif isinstance(time_index, pd.DatetimeIndex):
+            x_df["date_sequence"] = time_index
+        elif not isinstance(time_index, pd.DatetimeIndex):
```
KishManani, line 191: @danbartl - as we check whether `time_index` is of type `pd.DatetimeIndex` on line 189, should we change line 191 to be a simple `else`?

fkiraly (dismissed):
With the most recent change looks good to me now.

As algo owner, @danbartl should approve before we merge.

KishManani:
Hi @fkiraly!

I've discovered the potential for a side effect here after removing `X.copy()`. Here is an example:

```Python
import numpy as np
import pandas as pd

# Load data
y = load_airline()
y.head()
Period
1949-01-01    112.0
1949-02-01    118.0
1949-03-01    132.0
1949-04-01    129.0
1949-05-01    121.0
Freq: MS, Name: Number of airline passengers, dtype: float64

# Create datetime features
trafo = DateTimeFeatures(keep_original_columns=False)
trafo.fit(y)
result = trafo.transform(y)

# Modify the index of the result inplace
result.index.values[0] = "2000-01-01"

# Observe that the index of the original data is now changed
y.head()
Period
2000-01-01    112.0
1949-02-01    118.0
1949-03-01    132.0
1949-04-01    129.0
1949-05-01    121.0
Freq: MS, Name: Number of airline passengers, dtype: float64
```

Interestingly the above hack only works with a `DateTimeIndex`.  The `PeriodIndex` doesn't change value if you try to do the above.

I think there are two options here:

1. Leave it and hope that an unscrupulous user doesn't try to modify the index in-place like this. Pandas makes the index immutable so calls like `df.index[0] = 5` throws an error. But you can still bypass this by modifying the `values` attribute of the index.
2. We sacrifice some performance on memory and create a (deep?) copy of just the index to guarantee no side effects.

The trade-off here is about guaranteeing no side effects vs slightly more efficient memory management. What are your thoughts on this?

The same issue will be present in my recent change to the `TimeSince` trafo here: https://github.com/sktime/sktime/pull/4006

KishManani:
> The alternative would be, every estimator must create deep copies at the end. Which I think is not desirable as it would be memory intensive.

I agree!

> Let me know if you think I'm saying sth wrong. E.g., is this also true in `sklearn`?

I believe you're correct. I checked with the `MinMaxScaler` in `sklearn` and you get the same issue on the index (i.e., side effects when changing the index value). 

> but outputs may be identity linked to inputs.

That makes our lives easier then! I didn't know that. 

In which case, there is no issue in removing `X.copy()` here and I shouldn't lose sleep over side effects! 😄

fkiraly:
There's perhaps a more general design discussion around the upcoming config module here, e.g., should "make a copy at the end" be configurable? Either way, seems a bit off-topic for this PR, but keen to discuss it in issues that are on topic if you have thoughts to add.

Related to this PR, if you haven't seen it: I think @danbartl has managed to make enormous speed-ups to the input/output checking logic using `groupy` etc smartly!!
I'm sure he would appreciate you having a review of it since you were working on very similar things recently.
Here's the PR: https://github.com/sktime/sktime/pull/3935

sktime/transformations/series/date.py
KishManani: I've gone ahead and made the change. Let me know if there are any issues with this!

KishManani:
> Hi @danbartl! Do you think the following on line 229 is needed?
> 
> ```python
>         if self.manual_selection is not None:
>             df = df[self.manual_selection]
> ```
> 
> On the if block starting on line 194 there is a check that when `self.manual_selection` is not `None` there already appears to be a filter on line 220 and 225 to ensure the calculation is only computed on the subset of options found in `self.manual_selection`. I have commented out this block and checked that the tests still do pass. What do you think? Can we get rid of this?

I've gone ahead and removed this as it appears redundant. The tests are still passing so I assume functionality remains unchanged.

danbartl:
@KishManani makes sense, thanks 🙏

fkiraly (approved):
Looks good to go, @danbartl has confirmed that he is fine with the changes, too

PR #3997, opened 2022-12-26
[ENH] time bin aggregation transformer

fkiraly:
Fixes https://github.com/sktime/sktime/issues/242

Implements a transformer `TimeBinAggregate` which transforms a series to a series by aggregating in time bins as specified in a hyper-parameter (description in the docstring).


fkiraly:
fails since returns float index which is not admissible in the `sktime` mtype - how do we address this?

PR #3998, opened 2022-12-26
[DOC] add some missing entries in API reference

fkiraly:
adds some missing entries in API reference:

* kernel SVC classifier
* `WeightedEnsembleClassifier` classifier
* link to deployment topical reference
* `TimeSince` transformer


aiwalter:
this PR is violating the PR rules defined here: https://www.sktime.org/en/stable/get_involved/governance.html#decision-making

fkiraly:
@aiwalter, early merge is possible as per the rules.

Do you want to vote it down and revert it?

PR #4000, opened 2022-12-26
[MNT] python 3.11 compatibility, with `numba` as core dependency on 3.7-3.10

fkiraly:
This PR adds python 3.11 compatibility to `sktime`.

Compatibility strategy:

* most packages in `all_extras` are restricted to only python 3.7-3.10. The list can be relaxed gradually once packages become 3.11 compatible.
* core dependency `numba` remains a core dependency on 3.7-3.10
* on python 3.11, `numba` is not present (since not available). This requires the isolation measures in #3631, which leave user experience on 3.7-3.10 exactly the same.

User experience:
* user experience on 3.7-3.10 is entirely unchanged.
* users on 3.11 can install `sktime` now, previously they were unable to. As `numba` is not 3.11 compatible, it is not required.
* 3.11 incompatible estimators are aware of their dependency requirements and will notify the user at an instantiation attempt. Search for estimators via `all_estimators` yields all estimators, irrespective of whether their dependencies are satisfied.

Deprecation and compatibility:

* no deprecation is necessary, as behaviour and user experience on 3.7-3.10 is entirely unchanged.
* once `numba` becomes available on 3.11, it can become a core dependency on 3.11.


patrickzib:
1. we are between public holidays and everyone is on vacation!
2. We agreed to pause everything until the end of the numfocus investigation.
3. majority vote of core devs already voted this down. I do not see any change in this
4. I am still voting this down.

fkiraly:
> voted against the numba change

that was for `numba` to become a soft dependency.

This PR ensures python 3.11 compatibility and keeps `numba` as a core dependency.

> we are between public holidays and everyone is on vacation!

Timeline-wise, there will be enough time to discuss.

We can still work on python 3.11 when we have time to do so.

fkiraly:
@patrickzib, I made extra sure that your requirement of keeping `numba` a core dependency can be satisfied.

I was hoping for a more constructive reception of this - can we maybe agree on giving python 3.11 compatibility to our users, and you can be happy with `numba` as a core dependency?

patrickzib:
Franz. It is a public holiday. 

This is PR voted down until everyone is back.

fkiraly:
> Franz. It is a public holiday.
> This is PR voted down until everyone is back.

No chance of reaching an agreement? I'm not sure what it is you would prefer.

lmmentel:
Thanks for the effort @fkiraly but I do not think it is fair to expect anyone to go over `+9,335 −8,361` changes over 120 files. I do not immediately see what necessitated such a dramatic change to the code base. I consistently encourage everyone to keep PRs small to not end up in limbo since it's quite an effort to go over such a massive amount of changes.

I don't quite see  how all the changes (not having read  through all the changes) play into the issue you are describing. Would it not be sufficient to update dependency restrictions in `pyproject.toml` for now to not install incompatible dependencies under py311 and wait until the ecosystem have caught up with the latest python version?

I'm voting this down for now :-1:

fkiraly:
>  I do not immediately see what necessitated such a dramatic change to the code base.

@lmmentel, the changes are not dramatic - it is basically the same change in every file. Look at the first couple. Scope is narrow.

What changes are the imports, look at #3631 for a description of the general pattern.

> Would it not be sufficient to update dependency restrictions in `pyproject.toml` for now to not install incompatible dependencies under py311 and wait until the ecosystem have caught up with the latest python version?

That is precisely what I'm doing, but the naive approach is not possible since `numba` itself is incompatible and not isolated.

It is possible (and precisely what this PR does) with `numba` imports being isolated. But for that you have to go through every file with `numba` imports and place them either under a conditional or inside the class, as in the case of other dependencies.

See #3631 for a description of the isolation patterns applied.

patrickzib:
There are two votes against it. 

Enjoy public holidays :)

fkiraly:
above detractors/opponents of python 3.11 have publicly announced they are no longer part of sktime, so nothing is blocking python 3.11 anymore.

PR #4001, opened 2022-12-26
[BUG] fix dropped column index in `BaggingForecaster`

fkiraly:
This PR fixes an unreported bug (discovered by #3969) with `BaggingForecaster` which would drop its column names if `pd.DataFrame` were passed.

The reason was that the internal type was set as `pd.Series`, but it always returned a `pd.DataFrame` (and with inconsistent column name, always the integer zero).
The back-conversion of `BaseForecaster` would get confused and not retrieve the input column names as it would with a returned `pd.Series`.


aiwalter:
this PR is violating the PR rules defined here: https://www.sktime.org/en/stable/get_involved/governance.html#decision-making

fkiraly:
@aiwalter, early merge is possible as per the rules.

Do you want to vote it down and revert it, as well as the contract test for hierarchical forecasting in #3969?

PR #4005, opened 2022-12-27
[ENH] isolating `scipy` imports, part 1

fkiraly:
Copy of #3707 after it was reverted, for further consideration.


patrickzib:
For the reference: I am voting it down. Please, let's get back to it in 2023.

fkiraly:
patrickzib has left the project, so no review blocking this anymore.

PR #4006, opened 2022-12-27
[ENH] Enable `TimeSince` trafo to transform multiindex dataframes directly

KishManani:
This PR:
1.  allows the `TimeSince` transformer to operate directly on the time index of multi-index dataframes. This should give a performance boost when working with panel and hierarchical data.
2. simplifies the logic in `transform` and removes an unnecessary resampling of the dataframe when the index is of type `DateTimeIndex`. 
3. removes unnecessary copying of `X` unless `keep_original_columns=True`, in which case a copy of `X` is made so that we can return the original columns.

The use of `get_period_alias` to make it easier to convert from a datetime index to a period index.

##### For all contributions

### DIFF
--- a/sktime/transformations/series/tests/test_time_since.py
+++ b/sktime/transformations/series/tests/test_time_since.py
@@ -29,6 +29,15 @@ def df_datetime_15mins_idx():
 
 
 @pytest.fixture
+def df_datetime_weekly_wed_idx():
+    """Create timeseries with Datetime index, weekly with Wed start frequency."""
+    return pd.DataFrame(
+        data={"y": [1, 1, 1, 1, 1]},
+        index=pd.date_range(start="2000-01-05", freq="W-WED", periods=5),
+    )
+
+
+@pytest.fixture
 def df_datetime_monthly_idx():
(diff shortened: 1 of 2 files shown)


fkiraly (dismissed):
Makes sense to me - fine once the tests pass,

(you are yourself the author, so no author approval needed)

fkiraly:
Re `get_period_alias`, it seems to make sense what you are doing - although in my experience `pandas` is capricious when it comes to `freq` and the devil is in the detail. To the extent I grasp what you are doing, I can spot no issue.

KishManani:
Update:

I think this issue in Pandas https://github.com/pandas-dev/pandas/issues/36939 could cause problems if they change the behaviour of this method. To mitigate this I've added a test for when the data has weekly granularity starting on a Wednesday. If Pandas change the behaviour of `get_period_alias` to fix the issue attached (it's been open for 2 years now it seems) then the added test should fail.

fkiraly (approved):
Makes sense.
Together with the defensive bound for `pandas`, users should be safe and we should be able to detect it at release time.

PR #4008, opened 2022-12-28
[ENH] Add `keep_original_columns` option to `FourierFeatures` trafo

KishManani:
#### Reference Issues/PRs
See discussion here: https://github.com/sktime/sktime/pull/3996#issuecomment-1364733772

This PR makes the following changes:

1. Adds a `keep_original_columns` option and sets the default to `True` with a deprecation warning that in the future this should be set to `False`. This will make it easier to use `FourierFeatures` in the context of a pipeline (or FeatureUnion) where we do not wish to passthrough the other columns of `X`. 
2. Adds tests and modifies the existing tests now that the class can take a `keep_original_columns` as an argument. 
3. Removes the use of deep copies of `X` to make this more memory efficient. 

- Are there any concerns about removing the deep copies of `X` here? @ltsaprounis - would be great to hear your thoughts!

#### Any other comments?
For additional context around the impact of removing copies of `X` see the discussion here: https://github.com/sktime/sktime/pull/3996#issuecomment-1366607145. 

##### For all contributions
- [X] I've added myself to the [list of contributors](https://github.com/sktime/sktime/blob/main/.all-contributorsrc).

### DIFF
--- a/sktime/transformations/series/fourier.py
+++ b/sktime/transformations/series/fourier.py
@@ -6,7 +6,7 @@ __author__ = ["ltsaprounis"]
 
 import warnings
 from distutils.log import warn
-from typing import List, Union
+from typing import List, Optional, Union
 
 import numpy as np
 import pandas as pd
@@ -113,8 +113,8 @@ class FourierFeatures(BaseTransformer):
         self,
         sp_list: List[Union[int, float]],
         fourier_terms_list: List[int],
-        freq=None,
-        keep_original_columns=False,
+        freq: Optional[str] = None,
+        keep_original_columns: Optional[bool] = True,
     ):
         self.sp_list = sp_list
         self.fourier_terms_list = fourier_terms_list
--- a/sktime/transformations/series/fourier.py
+++ b/sktime/transformations/series/fourier.py
@@ -5,7 +5,6 @@
 __author__ = ["ltsaprounis"]
 
 import warnings
-from copy import deepcopy
 from distutils.log import warn
 from typing import List, Union
 
@@ -15,6 +14,8 @@ import pandas as pd
 from sktime.transformations.base import BaseTransformer
 
 
+# TODO: Change the default value of `keep_original_columns` from True to False
+# and remove the warning in v0.17.0
 class FourierFeatures(BaseTransformer):
     r"""Fourier Features for time series seasonality.
 
@@ -51,6 +52,8 @@ class FourierFeatures(BaseTransformer):
         Specifies the frequency of the index of your data. The string should
         match a pandas offset alias:
         https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases
+    keep_original_columns :  boolean, optional, default=True
+        Keep original columns in X passed to `.transform()`
 
     References
     ----------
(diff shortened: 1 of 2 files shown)


fkiraly (approved):
Great, really nice! Deprecation by the book.

@ltsaprounis is listed as owner so would have to approve before we merge.

KishManani:
Whilst we're on this topic, I have a question regarding what the behaviour of this transformer should be for panel data. The `t` that goes inside the sine and cosine functions is currently defined by the start of the time series. For panel data, the current implementation would look at the start index of each time series individually when getting the time indices to pass to the sine and cosine functions. This means that at time `t`, if two different time series have different start times, then the two time series will have different values for their fourier features. I think this could be problematic when fitting global models (e.g., linear regression via reduction). Consider a linear model, whilst the sum of the fourier features (i.e., a sum along axis=1) still represents a fourier series; the model might struggle to find the optimal coefficients because a given fourier feature (e.g., the column sin_7_1) contains not just a single sine wave (i.e.,`sin(2*pi*t/7)`) but instead a sine wave at multiple different phases (i.e.,`sin(2*pi*(t-t_i)/7)` where `t_i` is the start time of time series `i`). 

Here is an example:
```Python
import numpy as np
import pandas as pd
from sktime.transformations.series.fourier import FourierFeatures

# Panel dataframe with two timeseries
# Note there are entries in the two time series which have the same date.
df = pd.DataFrame(data={'c0': 
{('h0_0', pd.to_datetime('2000-01-01 00:00:00')): 2.1482432004268444,
  ('h0_0', pd.to_datetime('2000-01-02 00:00:00')): 1.5974533778936917,
  ('h0_0', pd.to_datetime('2000-02-01 00:00:00')): 2.5374903855394737,
  ('h0_1', pd.to_datetime('2000-02-01 00:00:00')): 1.2991497904155045,
  ('h0_1', pd.to_datetime('2000-02-02 00:00:00')): 1.920059660128214,
  ('h0_1', pd.to_datetime('2000-02-03 00:00:00')): 1.0}})
display(df)
                        c0		
h0_0	2000-01-01	2.1482432004268444
h0_0	2000-01-02	1.5974533778936917
h0_0	2000-02-01	2.5374903855394737
h0_1	2000-02-01	1.2991497904155045
h0_1	2000-02-02	1.920059660128214
h0_1	2000-02-03	1.0

trafo = FourierFeatures(sp_list=[7], fourier_terms_list=[1], freq="D")
trafo.fit(df)
result = trafo.transform(df)
display(result)
		        c0	sin_7_1	cos_7_1
h0_0	2000-01-01	2.148	0.0	1.0
h0_0	2000-01-02	1.597	0.782	0.623
h0_0	2000-02-01 *	2.537	0.434	-0.901 
h0_1	2000-02-01 *	1.299	0.0	1.0 
h0_1	2000-02-02	1.92	0.782	0.623
h0_1	2000-02-03	1.0	0.975	-0.223
```
The rows with the same time points (marked with an *) have different feature values.

@ltsaprounis - do you know if this is problematic or not? If it is, a simple solution would be to use a single reference time (i.e., the earliest time of all the time series) when computing the integer time index to pass to the sine and cosine functions. However, this would be modelling a shared seasonal component across all the multiple time series. What do you think?

fkiraly:
@KishManani, it could simply be an option: shared time axis (reference point possibly being the earliest time stamp occurring), or individual time axis, which is also the current (and recommended future?) default

KishManani:
> @KishManani, it could simply be an option: shared time axis (reference point possibly being the earliest time stamp occurring), or individual time axis, which is also the current (and recommended future?) default

This is a very pragmatic way forward. I'll wait for @ltsaprounis to comment and leave this for a separate PR.

KishManani:
Hi @fkiraly! Any idea of whether this can be merged?

fkiraly:
it's been a month now without reaction, and this moves an existing transformer towards an interface desideratum (do not keep columns). There is also deprecation which makes this easily reversible should discussion come up.
I'm therefore making the call to merge it for 0.16.0.

PR #4010, opened 2022-12-28
[DOC][BUG] Add warning regarding issues with macOS ARM

### LINKED ISSUE #3994: [BUG] Tensorflow failing on macOS
When following development environment setup guide in sktime documentation, conda installation was not able to run properly on macOS doe to missing tensorflow-macos dependency.

To reproduce the issue run on macOS:
```shell
$ conda create -n sktime-dev python=3.8
$ conda activate sktime-dev
$ pip install -e ."[all_extras,dev]"
$ make test
zsh: illegal hardware instruction
```

A factor that can complicate things is that tensorflow-macos cannot be installed using Python 3.8.

Possible solutions can be adding a warning to sktime documentation or adjusting pyproject.toml to tackle this issue.

The expected behavior would be to run successfully all tests after initial installation.

dainelli98:
#### Reference Issues/PRs

Fixes #3994.

This PR adds a warning for macOS users using ARM macs of the issues that they may encounter when installing ``all_extras``, related to some dependencies not compatible with that kind of setup.

This contribution does not add any new dependency.

Check if the warning is enough, or it would be appropriate to add a link to a separate new article in the documentation explaining in more detail the issue, listing the dependencies that cause the error, in case the user wants to install all the other dependencies in ``all_extras`` that do not generate any errors.

##### For all contributions
- [X] I've added myself to the [list of contributors](https://github.com/sktime/sktime/blob/main/.all-contributorsrc).

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -39,7 +39,8 @@ To install ``sktime`` with maximum dependencies, including soft dependencies, in
 
 .. warning::
     Some of the dependencies included in ``all_extras`` do not work on mac ARM-based processors, such
-    as M1, M1Pro, M1Max or M1Ultra. This may cause an error during installation.
+    as M1, M1Pro, M1Max or M1Ultra. This may cause an error during installation. Mode details can be found
+    in troubleshooting section below.
 
 
 Installing sktime from conda
@@ -217,6 +218,24 @@ Import errors are often caused by an improperly linked virtual environment.  Mak
 your environment is activated and linked to whatever IDE you are using.  If you are using Jupyter
 Notebooks, follow `these instructions <https://janakiev.com/blog/jupyter-virtual-envs/>`_ for
 adding your virtual environment as a new kernel for your notebook.
+
+Dependency error on mac ARM
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+If you are using a mac with an ARM processor, you may encounter an error when installing
+``sktime[all_extras]``.  This is due to the fact that some libraries included in ``all_extras``
+are not prepared to run in ARM-based processors.
+
+If you want to still install all the remaining compatible libraries from ``all_extras`` you need
+to take into account the following:
+
+* Avoid the following packages:
+    * esig
+    * prophet
+    * tsfresh
+    * tslearn
+* Replace tensorflow package with the following packages:
+    * tensorflow-macos
+    * tensorflow-metal (optional)
 
 Other Startup Resources
 -----------------------
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -171,7 +171,7 @@ In the ``anaconda prompt`` terminal:
 2. Create new environment with python 3.8: :code:`conda create -n sktime-dev python=3.8`
 
    .. warning::
-       If you already have an environment called "sktime-dev" from a previous attempt you will first need to remove this
+       If you already have an environment called "sktime-dev" from a previous attempt you will first need to remove this.
 
 3. Activate the environment: :code:`conda activate sktime-dev`
 
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -229,13 +229,13 @@ If you want to still install all the remaining compatible libraries from ``all_e
 to take into account the following:
 
 * Avoid the following packages:
-    * esig
-    * prophet
-    * tsfresh
-    * tslearn
-* Replace tensorflow package with the following packages:
-    * tensorflow-macos
-    * tensorflow-metal (optional)
+    * ``esig``
+    * ``prophet``
+    * ``tsfresh``
+    * ``tslearn``
+* Replace ``tensorflow`` package with the following packages:
+    * ``tensorflow-macos``
+    * ``tensorflow-metal`` (optional)
 
 Other Startup Resources
 -----------------------
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -39,8 +39,8 @@ To install ``sktime`` with maximum dependencies, including soft dependencies, in
 
 .. warning::
     Some of the dependencies included in ``all_extras`` do not work on mac ARM-based processors, such
-    as M1, M1Pro, M1Max or M1Ultra. This may cause an error during installation. Mode details can be found
-    in troubleshooting section below.
+    as M1, M2, M1Pro, M1Max or M1Ultra. This may cause an error during installation. Mode details can
+    be found in the :ref:`troubleshooting section<Dependency error on mac ARM>` below.
 
 
 Installing sktime from conda
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -37,6 +37,10 @@ To install ``sktime`` with maximum dependencies, including soft dependencies, in
 
     pip install sktime[all_extras]
 
+.. warning::
+    Some of the dependencies included in ``all_extras`` do not work on mac ARM-based processors, such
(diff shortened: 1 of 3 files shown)


docs/source/installation.rst
```diff
@@ -37,6 +37,10 @@ To install ``sktime`` with maximum dependencies, including soft dependencies, in
 
     pip install sktime[all_extras]
 
+.. warning::
+    Some of the dependencies included in ``all_extras`` do not work on mac ARM-based processors, such
+    as M1, M1Pro, M1Max or M1Ultra. This may cause an error during installation.
```
fkiraly, line 42: do you know any workarounds that one could mention here, or in the troubleshooting section (if longer)?

fkiraly (approved):
Thanks, great!

If you have a workaround that can be added, that would also be nice, but it´s not blocking.

docs/source/installation.rst
dainelli98: I could add the affected libraries and the fix for TensorFlow in the troubleshooting section if you think it is OK.

fkiraly: yes, that would be great! Short note next to the install with a link to the troubleshooting section, where you can add more details and workarounds.

dainelli98: I added a brief explanation in the workaround section. I hope it is not too much detail.

docs/source/installation.rst
```diff
@@ -214,6 +219,24 @@ your environment is activated and linked to whatever IDE you are using.  If you
...
+Dependency error on mac ARM
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+If you are using a mac with an ARM processor, you may encounter an error when installing
+``sktime[all_extras]``.  This is due to the fact that some libraries included in ``all_extras``
+are not prepared to run in ARM-based processors.
+
+If you want to still install all the remaining compatible libraries from ``all_extras`` you need
+to take into account the following:
+
+* Avoid the following packages:
```
fkiraly, line 231: would make it clearer: *do not install*, since "avoid" is a vague verb.

General advice for writing tech docs, think whether it is clear to the reader: How does it translate into action? look away from the computer screem while these are being installed ? :-)

docs/source/installation.rst
```diff
@@ -214,6 +219,24 @@ your environment is activated and linked to whatever IDE you are using.  If you
 Notebooks, follow `these instructions <https://janakiev.com/blog/jupyter-virtual-envs/>`_ for
 adding your virtual environment as a new kernel for your notebook.
 
+Dependency error on mac ARM
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+If you are using a mac with an ARM processor, you may encounter an error when installing
+``sktime[all_extras]``.  This is due to the fact that some libraries included in ``all_extras``
+are not prepared to run in ARM-based processors.
+
+If you want to still install all the remaining compatible libraries from ``all_extras`` you need
```
fkiraly, line 228: I would say more clearly: the workaround is to not install some of the packages, and install Mac compatible replacements for others.

docs/source/installation.rst
```diff
@@ -214,6 +219,24 @@ your environment is activated and linked to whatever IDE you are using.  If you
 Notebooks, follow `these instructions <https://janakiev.com/blog/jupyter-virtual-envs/>`_ for
 adding your virtual environment as a new kernel for your notebook.
 
+Dependency error on mac ARM
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+If you are using a mac with an ARM processor, you may encounter an error when installing
+``sktime[all_extras]``.  This is due to the fact that some libraries included in ``all_extras``
+are not prepared to run in ARM-based processors.
```
fkiraly, line 226: "not prepared" -> not compatible with

docs/source/installation.rst
```diff
@@ -214,6 +219,24 @@ your environment is activated and linked to whatever IDE you are using.  If you
 Notebooks, follow `these instructions <https://janakiev.com/blog/jupyter-virtual-envs/>`_ for
 adding your virtual environment as a new kernel for your notebook.
 
+Dependency error on mac ARM
```
fkiraly, line 222: maybe better section header: "installing `all_extras` on Mac with ARM processor"?

fkiraly (approved):
Excellent, thanks!
I don't think the troubleshooting paragraph is too long, it is clear and helpful.

No blocking comments from my side.

I left some comments on how you could improve formulations - in tech writing and espceially tech instructions, it's worth being very clear and precise (since the reader will be someone who just got frustrated and possibly confused by a failed install).
Not blocking, up to you if you want to give it another pass to sharpen the formulations (just let me know if not).

fkiraly:
Other question: I do sometimes run into trouble with `numba` on M1. Did you encounter that at all?

dainelli98:
> Other question: I do sometimes run into trouble with `numba` on M1. Did you encounter that at all?

To be honest, it has not been that long that I bought my current laptop. As of now, I have not experienced any issue, but that could be due to the fact that I have yet to use a package that leverages ``numba``. Anyway, this also interests me, so I will check. If you want, I can get back to you with what I get.

dainelli98:
> Excellent, thanks! I don't think the troubleshooting paragraph is too long, it is clear and helpful.
> 
> No blocking comments from my side.
> 
> I left some comments on how you could improve formulations - in tech writing and espceially tech instructions, it's worth being very clear and precise (since the reader will be someone who just got frustrated and possibly confused by a failed install). Not blocking, up to you if you want to give it another pass to sharpen the formulations (just let me know if not).

You are right, thanks for the advice. I will make the change, as I do not see any reason to not be as precise as possible.

dainelli98:
On the ``numba`` issue. As you may know already, Python packages can be distributed as wheels or source distribution. The first comes ready to be used, but the second needs to be built, which can cause issues on an ARM-based processor (The same issue occurs with ``prophet`` do to its dependency to ``pystan``, that comes as source distribution).

In the case of ``numba``, the proposed solution is to install it using conda:
```shell
conda install numba
```

Another solution seems to be to install with pip the version 0.55.2, that is distributed as wheel:
```shell
pip install numba==0.55.2
```

It is wierd that they do not provide a wheel for the last versions, but both workarounds seem to work on my mac.

fkiraly:
> It is weird that they sometimes do not provide a wheel for the last versions, but both workarounds seem to work on my mac.

Is that worth adding, too?

fkiraly:
Thanks for ths tipps btw! At work, we have a mix of ARM and non-ARM macs, and the two behave differently...

fkiraly:
> but that could be due to the fact that I have yet to use a package that leverages `numba`

`sktime` has `numba` as a core dependency, but it is used only in a couple estimators (not in the framework).

dainelli98:
> > It is weird that they sometimes do not provide a wheel for the last versions, but both workarounds seem to work on my mac.
> 
> Is that worth adding, too?

The only reason to add it would be because the issue that the last version of ``numba`` that now is fixed could happen in the future with another package if they do a new release without providing the wheel.

As of now, all the packages not mentioned in the workaround do not present this issue, so at this moment this problem should not arise. If you think this is a bit overkill, it is easy to undo the commits.

fkiraly:
ok, makes sense. Let me know when you think you're done, for review and potential merge.

dainelli98:
> ok, makes sense. Let me know when you think you're done, for review and potential merge.

For me everything is okay.

fkiraly (approved):
Excellent, great contribution!

Hope this will be helpful to all the poor souls with ARM macs out there.

fkiraly:
failures unrelated to this PR, see #4033

PR #4015, opened 2022-12-29
[BUG] Fix `TimeSince` check of inconsistency between `time_index` and `start`

KishManani:
There is a check that the type of the time index and the `start` argument are consistent. The check was still being done on `X.index` which is fine for series data, but not for panel or hierarchical data. The check has been changed to the `time_index` variable.

##### For all contributions

### DIFF
--- a/sktime/transformations/series/time_since.py
+++ b/sktime/transformations/series/time_since.py
@@ -182,7 +182,7 @@ class TimeSince(BaseTransformer):
                     f"Period index. Check that `start` is of type "
                     f"pd.Period or a pd.Period parsable string."
                 )
-            elif X.index.is_numeric() and not isinstance(start_, (int, np.integer)):
+            elif time_index.is_numeric() and not isinstance(start_, (int, np.integer)):
                 raise ValueError(
                     f"start_={start_} incompatible with a numeric index."
                     f"Check that `start` is an integer."
--- a/sktime/transformations/series/time_since.py
+++ b/sktime/transformations/series/time_since.py
@@ -58,13 +58,13 @@ class TimeSince(BaseTransformer):
     >>> from sktime.transformations.series.time_since import TimeSince
     >>> X = load_airline()
 
-    Create a single column with time elapsed since start date of time series.
-    The output is in units of integer number of months, same as the index `freq`.
+        Create a single column with time elapsed since start date of time series.
+        The output is in units of integer number of months, same as the index `freq`.
     >>> transformer = TimeSince()
     >>> Xt = transformer.fit_transform(X)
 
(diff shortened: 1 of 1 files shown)


fkiraly (approved):
Good catch!

aiwalter:
@KishManani as you are on it, can you kindly fix the `Examples` section, its not rendering well:
https://www.sktime.org/en/latest/api_reference/auto_generated/sktime.transformations.series.time_since.TimeSince.html?highlight=TimeSince

The comments basically have to be indented

KishManani:
Thanks for spotting that @aiwalter! I've made the change to the examples. But the parameters also don't look to be rendering how they should. 

image

I checked out the docs on [documentation](https://www.sktime.org/en/stable/developer_guide/documentation.html) and the examples of good documentation. I can't see an example with a bullet pointed list in the parameters section. If you know of any examples or how I should change the docstring so that the parameters section renders correctly please let me know!

Making the docs locally by running `make html` takes a very long time. Do you know if there is a quicker way for me to check if my docstrings are being rendered correctly? Thank you!

aiwalter:
You can check `Imputer`. The CI/CD has a build of the PR docs. You can click on it once passed

fkiraly:
> I checked out the docs on [documentation](https://www.sktime.org/en/stable/developer_guide/documentation.html) and the examples of good documentation. I can't see an example with a bullet pointed list in the parameters section.

Then there should be one - let's open an issue for someone to link one (and perhaps add explanation what can be learnt from the examples on the `documentation` page).

KishManani:
> Might be best if you do this, @KishManani, since you know best what you were looking for and didn't find, so you can best specify what the documentation page should have done/explained?

Sure! Will do.

KishManani:
> You can check Imputer. The CI/CD has a build of the PR docs. You can click on it once passed

Thanks @aiwalter, this is very handy indeed! Do you happen to know of a faster workflow? It can take up to 5 minutes or longer for this to run. The workflow for me is currently like the following: 1) look at documentation and spot errors, 2) take a guess as to whether I need to change indentation or some other fix, 3) render the docs and check if the fix worked. I was wondering if you know of a quicker way to do step 3) 😅 ?

For some reason the formatting is not consistent in my list in the parameters section. For some reason "time-like" is bolded.
image

Any suggestions?

fkiraly:
> Any suggestions?

Get an exorcist

fkiraly (approved):
Good for me.

I'd say, the doc formatting issue is different scope and hence a different PR.
Let's merge this so there isn't too much conflict with PR https://github.com/sktime/sktime/pull/4018

PR #4016, opened 2022-12-29
[MNT] full test run on mid-to-lower dependency version range, 2023 state of core deps

fkiraly:
This adds a full test run on the lower supported dependency boundary of the minimal dependency set, under python 3.11 and ubuntu, to test at the 2023 state of `sktime`'s support range.

Depends on:
* https://github.com/sktime/sktime/pull/4041
* https://github.com/sktime/sktime/pull/4443


fkiraly:
any ideas about the strange error in the test-minimaldeps?

@benHeid, @yarnabrina

benHeid:
This is just a guess:
After some investigations, I have the feeling that this error occurs mainly if there are problems with the installed numpy version and the version that is used for compiling other packages. [Comment in an issue in SKLearn on that topic](https://github.com/scikit-learn/scikit-learn/issues/3537#issuecomment-248700575)

If this is really the problem, then we might need to separate the installation into two commands. First, we need to install numpy. Afterwards, we can install the rest.

fkiraly:
yes, I also got that - somehow it seems two `numpy` versions get installed?

This wasn't the case ca 6 months ago when this PR first got opened though, so I also am not sure on what would have changed here.

benHeid:
I guess that the runner has a numpy version installed via apt-get. During the last six months this numpy version was updated. If the updated version now has a new feature which the minimal dependency does not have. Then this would mean that scipy is built with a new feature that is not available in the older numpy version which is installed after building the wheels.

fkiraly:
Correctly picks up the failure in https://github.com/sktime/sktime/issues/8544

PR #4018, opened 2022-12-29
[ENH] Make `TimeSince` trafo faster by changing period diff calculation

KishManani:
Inspired by how @ltsaprounis calculated the integer time difference when using a `PeriodIndex` in the `FourierFeatures` transformer, I've switched to using this alternative method which is *much* faster. I measured a factor of ~100 speed up in `_transform`.

An example to measure the difference:
```Python

df =_make_hierarchical(hierarchy_levels=(10000,), min_timepoints=200, max_timepoints=200)

trafo = TimeSince(freq="D")
trafo._fit(df)
trafo._transform(df)
```

This PR will have conflicting changes with https://github.com/sktime/sktime/pull/4015. Once https://github.com/sktime/sktime/pull/4015 has merged I will resolve the conflicts in this PR. 

Code quality. I'm having to write a lot of comments to explain the context of what's going on. I'm unsure whether this is a bad smell. 

##### For all contributions

### DIFF
--- a/sktime/transformations/series/time_since.py
+++ b/sktime/transformations/series/time_since.py
@@ -229,26 +229,47 @@ class TimeSince(BaseTransformer):
                     freq_ = _remove_digits_from_str(self.freq_)
                     freq_period = get_period_alias(freq_)
 
-                    # Convert `start` and datetime index to period.
-                    start_period = pd.Period(start_, freq=freq_period)
+                    # Convert `start` and `time_index` to period.
+                    # Casting `start_` to PeriodIndex using pd.period_range
+                    # here so we can later cast it to an int using
+                    # `astype(int)` in _get_period_diff_as_int().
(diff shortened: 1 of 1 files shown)


fkiraly:
Merged https://github.com/sktime/sktime/pull/4015 so avoiding conflicts should now be easier.

fkiraly (approved):
I can confirm the speedup.
In a fresh python 3.10 env on windows, the example is reduced from 7 sec to 300 msec.

Functionality seems unaffected.

KishManani:
It appears there are no conflicts 🤷‍♂️ , that saves some time then!

fkiraly:
Re comments: I think this is a bit of a smell, but I would say, ultimately it's coming from `pandas` (and how it handles indices). The dependent code isn't bad.

PR #3852, opened 2022-11-29
&fkiraly [DOC] updates to distances API reference page

MatthewMiddlehurst:
update by @fkiraly 

this PR was split into three: this PR, and #4027 and #4028 for different content.

This PR now only adds the API reference for `numba` distances, to the existing distances API reference.

Additional updates to the API reference:

* adds panel distances and kernels that were missing from the API reference
* adds explanation on how to construct common (flat or aggregate) distances and kernels

---

- Adds distance function API page for `distances`.
- Moves clustering base classes to the bottom of the page to conform with classification and regression
- Adds `all_estimators` snippet from the regression page to classification and clustering

PR #4027, opened 2022-12-31
&matthewmiddlehurst [DOC] `all_estimators` reference on all estimator pages

fkiraly:
This adds a reference to `all_estimators` on all pages for estimators.

Taken partially from https://github.com/sktime/sktime/pull/3852 (which added this for some estimator types).


fkiraly:
(taken from a PR by @matthewmiddlehurst, only slightly extended. Imo can be considered reviewed & approved by four eyes)

PR #4036, opened 2023-01-01
[BUG] Diagnose and fix sporadic failures in the test suite due to `MemoryError`

### LINKED ISSUE #4033: [BUG] LSTM deep learning estimators failing CI on windows
Since recently, a failure of two deep learning estimators has been appearing on windows CI:
```
FAILED sktime/tests/test_all_estimators.py::TestAllEstimators::test_methods_have_no_side_effects[MLPClassifier-1-ClassifierFitPredictMultivariate-predict]
FAILED sktime/tests/test_all_estimators.py::TestAllEstimators::test_methods_have_no_side_effects[LSTMFCNClassifier-0-ClassifierFitPredictMultivariate-predict]
```

Note that the failure appears to be only on python 3.9, but this is due to the matrix design which spreads estimators across version/OS combinations.

In theory, it could be a 3.9 specific failure, but I think that is less likely than windows specifity (although it may be worth to test that by turning the `matrixdesign` flag off in the CI).

achieveordie:
#### Reference Issues/PRs

Fixes #4033

Current code changes improve `LSTMFCNClassifier`'s `get_test_params()` method to increase code coverage as well as decrease the size of the resulting model that is to be used for testing. If this doesn't stop failures, I'll edit this blob to incorporate different changes made to diagnose this problem.

None

##### For all contributions

##### For new estimators


achieveordie:
@solen0id Is there any reason you believe that `X = X.transpose(0, 2, 1)` ([source](https://github.com/sktime/sktime/blob/main/sktime/classification/deep_learning/lstmfcn.py#L181)) from `LSTMFCNClassifier` should be removed? You probably have gone through the paper/implementation recently so you might know more about it than I do. So I just wanted to know your opinion on it before I remove that comment.

achieveordie:
The check has passed once, I'll retrigger it a few more times to make sure everything is smooth.

achieveordie:
Currently failing with the following:

```
worker 'gw1' crashed while running 'sktime/tests/test_all_estimators.py::TestAllEstimators::test_save_estimators_to_file[LSTMFCNClassifier-0-ClassifierFitPredict-predict_proba]'
...
...
=========================== short test summary info ===========================
FAILED sktime/tests/test_all_estimators.py::TestAllEstimators::test_save_estimators_to_file[LSTMFCNClassifier-0-ClassifierFitPredict-predict_proba]
FAILED sktime/tests/test_all_estimators.py::TestAllEstimators::test_non_state_changing_method_contract[CNNClassifier-1-ClassifierFitPredict-get_fitted_params] - tensorflow.python.framework.errors_impl.ResourceExhaustedError: Graph execution error:
```

Comments:
- Seems like I was correct about `LSTMFCNClassifier` causing the failures. In this run, `CNNClassifier` was caught in the collateral damage, as opposed to `MLPClassifier` in the originally linked issue.
- It might be coincidental that the same combination of the matrix is failing (python 3.9, windows). 
- Even the current size might be too big and hence causing `OOM`. 

Based on my comments, I'll reduce the size even further, assuming my original diagnosis is correct (and `OOM` error is being caused by the estimator size). If even then the issue persists, the next step would be to alter the test matrix and remove the current failing element (win-3.9). If errors stop then we'd try to look into why this specific combination is leading to the error.

achieveordie:
@fkiraly FYI, `pip._vendor.urllib3.exceptions.ReadTimeoutError: HTTPSConnectionPool(host='files.pythonhosted.org', port=443): Read timed out.` is still [occuring](https://github.com/sktime/sktime/actions/runs/3901192314/jobs/6662916556#step:5:106). I think this is what you observed a while ago, it occurs while building `sktime` in mac-based systems.

achieveordie:
Progress on this task:
- The root of this problem was corrected in #4067, and I believe this issue has been resolved. I'll re-run the checks a couple of times to be sure that everything works well.

_Number of times the checks have successfully run: 1/1_

achieveordie:
This should be good to merge. Now that we have taken care of the root cause, the smaller set of test parameters should not lead to `MemoryError`. I've run the checkers 4 times already, I don't think we need to check it for the fifth time separately (It will be done during push-to-main).

fkiraly (approved):
Yes, that was probably it!

PR #4040, opened 2023-01-02
[ENH] `hcrystalball` forecaster adapter

fkiraly:
This PR adds back @MichalChromcak's framework adapter to `hcrystalball`.

Since its removal, internal discussions have clarified inclusion policy of estimators in `sktime`, see here: https://www.sktime.org/en/stable/get_involved/governance.html#algorithm-inclusion-guidelines

The consolidated policy sees `sktime` as a library, to which any estimator can be contributed as long as minimal requirements on documentation are met.

I.e., perceived duplication, performance, relevance, personal taste of core developers, are not criteria we have decided we want to use going forward to decide on whether an algorithm should be included.

This PR moves the estimator into a new module `forecasting.adapters`, a location where potentially further adapters can be placed, say, to `gluonts`, `darts`, `pytorch-forecast` and so on (should someone implement those).

`hcrystalball` is not added as soft dependency in `all_extras`, which is fine, as `sktime` now also has full capability for dependency isolation.


sktime/forecasting/adapters/_hcrystalball.py
```diff
@@ -0,0 +1,186 @@
...
+
+__author__ = ["MichalChromcak"]
+
+import pandas as pd
+from sklearn.base import clone
+
+from sktime.forecasting.base import BaseForecaster
+from sktime.utils.validation._dependencies import _check_soft_dependencies
+
+_check_soft_dependencies("hcrystalball", severity="warning")
```
aiwalter, line 14: This is not in `all_extras`

aiwalter (changes requested):
There was a core dev vote not to have this in past

fkiraly:
> There was a core dev vote not to have this in past

* that was before the change in estimator inclusion policy
* which followed the discussion on HIVE-COTE where thought, @aiwalter, that it seemed to be a code of conduct violation to remove someone else's estimator

Same rules for everyone, I would suggest.

sktime/forecasting/adapters/_hcrystalball.py
fkiraly: if you read the PR description, it says:

> `hcrystalball` is not added as soft dependency in `all_extras`, which is fine, as `sktime` now also has full capability for dependency isolation.

(fine in my opinion)

fkiraly:
If you want to block this or downvote this, please specify, on which grounds.

aiwalter:
Lazy consensus failed on this PR

fkiraly:
@aiwalter, that is not a productive statement - it amounts to you saying "I cause the lazy consensus to fail, and my reason is that the lazy consensus fails".

I.e., you say you reject this "just because".

I kindly ask you to reconsider your position on this PR and contribute productively to a discussion.

fkiraly:
no follow-up, and not a clear change request. Can be merged by standing policy.

Issue #3717, opened 2022-11-06
[DOC] Create "sktime people" section on our website


GuzalBulatova:
#### Describe the issue linked to the documentation

We'd like to celebrate our contributors and make the most active / recent ones more visible. After discussion in https://github.com/sktime/community-org/issues/36 , community decided to go with this option for improvement on our .org website.
#### Suggest a potential alternative/fix

Create "sktime people" section similar to https://fastapi.tiangolo.com/fastapi-people/ with sections like "Top Contributors", "Top Reviewers", etc. Ideally under avatar there'd be the type of contribution indicated, like "code", "documentation", "bug reporting", "mentoring" etc.

Aarthy153:
Hi, @GuzalBulatova @miraep8. I would like to take up this issue and fix it. Can you just provide me a short guidance on the actual requirements?

Aarthy153:
Hi, I just got an idea to do this. Instead of marking down all the names. the idea is to use the services from https://contrib.rocks/preview

PR #4048, opened 2023-01-03
[DOC] Update the `README.md` with the contributors section

Aarthy153:
## Description

Adding a separate section for contributors in the readme file. I just found a way to do this efficiently. Instead of marking down all the names, the idea is to use the services from https://contrib.rocks/preview.

## Issue reference
#3717

## Screenshot of the fix

image

## Additional Comments

Clicking the image will redirect you here https://github.com/sktime/sktime/graphs/contributors
And, hovering on the profile image will give you the name of the contributor.


Aarthy153:
@GuzalBulatova @miraep8 Need your reviews on this PR

fkiraly:
nice! I haven't looked in detail, just wanted to let you know of one thing I've noticed: the number of collaborators in the picture does not seem to be accurate. I count precisely 100, whereas the current total is at 215. Perhaps it is capped by 100 (in which case I suppose it's ok?)

fkiraly:
Hmmm, I think https://github.com/sktime/sktime/graphs/contributors contains only the 100 with most merges to `main`.

If you look at the landing page, there are actually 215:
image

Not sure whether it is easy to change (or desirable), I just wanted to point out it's probably not *all* contributors.

There are also contributors who have made contributions other than code - and those are important too! - these may not be listed via the GitHub API, but are listed in the `allcontributors` file.

Aarthy153:
@fkiraly I am fixing it. Yes, the alternative idea is to use the all-contributors file.

Aarthy153:
I have committed the new changes with the data from all-contributors. kindly verify!

achieveordie:
Hey, @Aarthy153 Thanks for this PR!

However, I'm not sure if this directly references the linked problem. We have a [`CONTRIBUTORS.md`](https://github.com/sktime/sktime/blob/main/CONTRIBUTORS.md) that has a table of all our contributors (in line with all-contributors [specification](https://allcontributors.org/docs/en/specification)).

The reason we don't have a contributors section directly in `README.md` is to keep it brief. Here's a quote from the said specification:

> Consider [using a CONTRIBUTORS file](https://allcontributors.org/docs/en/bot/configuration#moving-the-all-contributors-table-into-a-different-file) in the top level of the repository when the number of project contributors exceeds a level at which it is unfeasible to use the README file to acknowledge contributions. In this case, a prominent link to the CONTRIBUTORS file should be included on the README page under the "Contributors" heading

We make use of the `all-contributors-cli` to update this (as you probably may have done) and currently we're in the process of automating it in #3807.

achieveordie:
Oh, it seems like the code quality check is not passing currently, please take a look at [how](https://www.sktime.org/en/stable/developer_guide.html) to set up a dev environment. If you have any troubles, do let me or anyone else know.

miraep8:
I would say the best/easiest way to make sure that your code always passes the quality check is just to [set up pre-commit](https://pre-commit.com/) on your end! :)  That will automatically make sure that you can't push code that won't pass the test.  As already said - please do reach out if you run into any issues on this!

miraep8:
Regarding the PR itself - I would personally request that we maybe add the change to the documentation website instead of the readme.  (I think that was the original intention in Guzal's issue).  I really like that your solution auto-updates already though! :)  As @achieveordie said we are looking at other ways to have the contributor file auto-update, but I think something like what you suggested (perhaps prefaced with a title "Top 100 Code Contributors") could be a great stand in/starting place! :) It could be added to the contributors section already in the docs? 

Thanks for this PR/finding a nifty solution!

Aarthy153:
Hi, I would to like to know in more detail on what actually fixes this issue? Should I add the link to the Readme.md or should I add the main 100 contributors (main commits)?
If you could kindly explain in detail it would be easy for me to fix it as soon as possible.
@fkiraly @achieveordie @fkiraly. Thank you!

achieveordie:
So the linked issue is regarding adding a "sktime people" section to our website https://sktime.org. This section is different from the [contributors](https://www.sktime.org/en/stable/about/contributors.html) section as the latter consists of all contributors as per all-contributors specification and we want the former to be representative of people that are currently active. 

We need something similar to [FastAPI](https://fastapi.tiangolo.com/fastapi-people/), a section that changes with time. Since that may be too complicated, we could take the route that @miraep8 suggested and add a "Top 100 Code Contributors" section for this PR. 

The contents, structure and code to build the website are present in the [docs](https://github.com/sktime/sktime/tree/main/docs) directory of our repo. We use [Sphinx](https://www.sphinx-doc.org/en/master/) to build our documentation(website).

I can think of a couple of ways to try to do it (mind you I'm not an expert). I suggest you to go through the links, make sure you understand the problem and then try to come up with a solution. @miraep8 and I will help you with inputs if you still feel completely lost.

P.S. We don't have to change `README.md` at all. The file that needs to be changed is [docs/source/about/contributors.md](https://github.com/sktime/sktime/blob/main/docs/source/about/contributors.md) which right now directly makes use of `CONTRIBUTORS.md` (present at the top level).

PR #4049, opened 2023-01-03
[BUG] fix default parameter mutation by `Detrender`

fkiraly:
This fixes an unreported but with `Detrender` detected through the new parameter set in https://github.com/sktime/sktime/pull/4043.

If `forecaster` is not passed, it is mutated by the `_fit` method (from `None` to a default forecaster), which does not conform with the interface contract.

This is fixed by adding a `clone` step in-between.


fkiraly:
superseded by #4053

Issue #4051, opened 2023-01-03
[BUG] incorrect multivariate detrending in `Detrender`


fkiraly:
Currently, the `Detrender` transformer implements multivariate detrending incorrectly, and in contrast to what its docstring says.

The difference between "should" and "is" happens if the forecaster passed is genuinely multivariate - in that case, the `Detrender` coerces it to univariate, and fits an univariate version per variable, instead of just applying the multivariate forecaster and detrending via it.

This has the status of a bug since the docstring does not agree with content, and the content is contrary to user expectation.

PR #4053, opened 2023-01-03
[ENH][BUG] clean up `Detrender` and implement correct multivariate behaviour

fkiraly:
This PR improves the `Detrender`:
* it fixes https://github.com/sktime/sktime/issues/4051 and also simplifies the logic in the `Detrender`. As the `BaseForecasting` interface supports loop over hierarchy instances and variables natively, all the case distinctions internally can go, and with it goes the incorrect multivariate implementation.
* adds support for detrending with hierarchical forecasters
* adds support for detrending with forecasters that require `fh` in `fit` already (this was not possible previously)

As a consequence, the test `test_polynomial_detrending` had to be updated since the internal mtype of the `Detrender` changed, from `pd.Series` based to `pd.DataFrame` based (requiring a minor change in how objects are being compared).

### DIFF
--- a/sktime/transformations/series/detrend/_detrend.py
+++ b/sktime/transformations/series/detrend/_detrend.py
@@ -4,7 +4,7 @@
 """Implements transformations to detrend a time series."""
 
 __all__ = ["Detrender"]
-__author__ = ["mloning", "SveaMeyer13", "KishManani"]
+__author__ = ["mloning", "SveaMeyer13", "KishManani", "fkiraly"]
 
 import pandas as pd
 
@@ -73,10 +73,11 @@ class Detrender(BaseTransformer):
         "scitype:transform-output": "Series",
         # what scitype is returned: Primitives, Series, Panel
         "scitype:instancewise": True,  # is this an instance-wise transform?
-        "X_inner_mtype": ["pd.DataFrame", "pd.Series"],
+        "X_inner_mtype": ["pd.DataFrame", "pd-multiindex", "pd_multiindex_hier"],
         # which mtypes do _fit/_predict support for X?
-        "y_inner_mtype": "pd.DataFrame",  # which mtypes do _fit/_predict support for y?
-        "univariate-only": True,
+        "y_inner_mtype": ["pd.DataFrame", "pd-multiindex", "pd_multiindex_hier"],
+        # which mtypes do _fit/_predict support for y?
+        "univariate-only": False,
         "fit_is_empty": False,
         "capability:inverse_transform": True,
         "transform-returns-same-time-index": True,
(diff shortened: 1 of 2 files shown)


fkiraly:
FYI @mloning, @SveaMeyer13, @KishManani, would appreciate a look as you have worked on the detrender before.

sktime/transformations/series/detrend/_detrend.py
```diff
@@ -197,36 +183,15 @@ def _inverse_transform(self, X, y=None):
...
-            for colname in X.columns:
-                X_pred = self.forecaster_[colname].predict(fh=fh, X=y)
-                if self.model == "additive":
-                    X[colname] = X[colname] + X_pred
-                elif self.model == "multiplicative":
-                    X[colname] = X[colname] * X_pred
-
-            return X
+        fh = self._get_fh_from_X(X=X)
+        forecaster = self._get_fitted_forecaster(X=X, y=y, fh=fh)
```
KishManani, line 187: If `X` is the de-trended data, wouldn't passing this to `_get_fitted_forecaster` return a newly fitted forecaster when `if self.forecaster_.get_tag("requires-fh-in-fit", True)`, resulting in the forecaster being fitted on the de-trended data? Is there an example forecaster you have in mind where this is needed to add the trend back to the data?

fkiraly: ah, good catch! Didn't think carefully about this. I suppose fitting on the detrended data does not really make sense here.

Should it then just raise an error, if `self.forecaster_.get_tag("requires-fh-in-fit", True)`?

fkiraly: here's I think what should happen in the `inverse_transform`: the data should *not* be updated with the new data, simply the data in `fit` should be remembered. Changed to this.

KishManani: The only downside is that now we'll be carrying around a copy of the data inside the transformer. Perhaps this is not much of an issue because it's just reference rather than a deep copy of the data. My concern was If I wanted to just use something like the `PolynomialTrendForecaster` to do de-trending and re-trending, then I don't actually require the data after `fit`, I certainly wouldn't want the burden of carrying around a copy of the data unnecessarily. But we don't have the memory burden because it's not a deep copy, so hopefully should be fine.

fkiraly: well, it will not happen for the `PolynomialTrendForecaster` because its tag is `"requires-fh-in-fit": False`.

The memorization happens *only* in cases where this tag is `True` - a case where the transformer previously would just crash.

In the `False` case, behaviour is precisely as before for univariate forecasters, and the "bugfixed" behaviour for multivariate or hierarchical ones.

Issue #3288, opened 2022-08-18
[ENH] `ARDL` leftover items


fkiraly:
Some todo issues remaining in the `ARDL` interface:

- [ ] `_update` with `update_params=False`
- [ ] probabilistic forecast interface
- [ ] add the estimator the API reference (docs)

The first point seems difficult, the last one easy, so good first issue (not `_update`)

xPushpeshx:
hello sir may i  resolve this issue?

xPushpeshx:
@fkiraly  can you elaborate 2nd issue and precisely tell what is estimator refereed here

PR #4059, opened 2023-01-04
[ENH] 1st issue resolved

xPushpeshx:
- solved issue 1 (#3288 )


fkiraly (changes requested):
The default should remain as it was - the issue was not about changing the default, but implementing the case where the parameter is `False`.

Issue #3853, opened 2022-11-29
[DOC] `scikit-learn` copyright statements


fkiraly:
Where we have reused code snippets from `scikit-learn`, we should ensure they appear at the top near the license statement to ensure we cleanly comply with `scikit-learn`'s BSD3 license.

In most cases the reference is somehwere in docstrings, but I think it should also live in modules.

See [here](https://github.com/sktime/skbase/pull/62#discussion_r1031776986) for an example in the `skbase` refactor.

I suppose this mostly affects concerns around the base class and some tests.
I can also imagine that tests, having undergone extensive refactoring, may contain snippets of `sklearn`'s `check_estimator` logic.

PR #4060, opened 2023-01-04
[ENH] complainces with scikit-learn's License

xPushpeshx:
updated license terms complying with scikit-learn's license (#3853)


fkiraly (changes requested):
Thanks!

I think this is not exactly right. I would say:

* the full license in the main folder should stay as it is
* where code from `sklearn` has been copied, it should be acknowleded in the file

As per my understanding, at least - I'm not 100% sure of that though.

xPushpeshx:
got it thank you

Issue #4064, opened 2023-01-04
[BUG] should the `dev` dependency set include requirements of `pre-commit`?


fkiraly:
The `dev` dependency set is intended for `sktime` developers, but does not include requirements of `pre-commit`, even though setting `pre-commit` is a recommended part of the `sktime` developer workflow.

Is this an oversight? Or is it intentional?

yarnabrina:
> Do I need to have black, isort etc installed for pre-commit install to work?

No. 100% certain. Installing `pre-commit` is enough, and what the command will basically do is create a new hook `pre-commit` in `.git/hooks`. Nothing else is required, not even the configuration file.

> vs code integration

I think this will vary because different people may want different types of integrations. For example, I want only syntax errors to be highlighted by VS Code, nothing else, and I use CLI commands myself to get those.

But anyway, I think VS Code integration does not need packages by default. It mainly needs the `Python` extension to start with. Earlier it contained few integrated tools as well, but now Microsoft is releasing `isort`, `black`, `mypy`, etc. extensions separately. Installing these extensions should probably be enough for basic error highlighting.

For more control, I suppose one can configure interpreter per workspace, configure the formatting options from settings etc., but as I said, it depends on what the user's personal preferences.

---

This is my _**biased**_ suggestion, based on what I usually do for my personal projects, such as [this](https://github.com/yarnabrina/learn-python-packaging).

1. Decide on configurations for formatters (`autoflake`, `black`, `docformatter`, `isort`, `pyupgrade`, etc.) / linters (`flake8`, `mypy`, `pydocstyle`, `pylint`, etc.) / both (`ruff`, etc.) to follow on `sktime`.
2. Add these configurations in `pyproject.toml`.
    * Few tools, mainly those that are managed by Anthony Sottile, do not support `pyprject.toml`. For those, use other configuration files, e.g. `.flake8` or `.pre-commit-config.yaml` etc.
3. Pass no configuration in `.pre-commit-config.yaml` for tools that have configurations specified on other standard files (as part of 2), and those configurations will be respected by default.
4. Create separate optional dependencies for formatters and linters, along with those for documentation, testing, release, etc. stuff.
5. Create `dev` optional dependencies for dev, which covers all of the above optional dependencies.
6. Create `Makefile` for all tools with exact specification of files and folders, so that running `make <tool-name>` is enough. It usually gives better/easier control for inclusion/exclusion over `pre-commit`.
[...]

PR #4065, opened 2023-01-04
[ENH] linting & pre-commit hook dependency set

fkiraly:
This is a potential fix for https://github.com/sktime/sktime/issues/4064 and includes an extra dependency set for linters and pre-commit hook dependencies.

This is for developers who may like to set up a python env for auto-linting in their local IDE, or a custom CI element for pre-commits or linting.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -94,6 +94,18 @@ dev = [
     "wheel",
 ]
 
+dev_precommit = [
+    "black",
+    "check-manifest",
+    "flake8",
+    "isort",
+    "nbqa",
+    "nbqa-black",
+    "nbqa-flake8",
+    "nbqa-isort",
+    "pydocstyle",
+]
+
 mlflow = [
     "mlflow",
 ]


fkiraly:
Looking through some old PR.

What do you think about this, @yarnabrina?

pyproject.toml
```diff
@@ -94,6 +94,18 @@ dev = [
     "wheel",
 ]
 
+dev_precommit = [
```
yarnabrina, line 97: The name is a bit surprising, given it doesn't contain `pre-commit`.

pyproject.toml
```diff
@@ -94,6 +94,18 @@ dev = [
...
+    "black",
+    "check-manifest",
+    "flake8",
+    "isort",
+    "nbqa",
+    "nbqa-black",
```
yarnabrina, line 103: Not sure if it exists before or not, but as of now `nbqa-black` does not exist in `PyPI`.

pyproject.toml
```diff
@@ -94,6 +94,18 @@ dev = [
...
+    "nbqa-flake8",
```
yarnabrina, line 104: Not sure if it exists before or not, but as of now `nbqa-flake8` does not exist in `PyPI`.

pyproject.toml
```diff
@@ -94,6 +94,18 @@ dev = [
...
+    "nbqa-isort",
```
yarnabrina, line 105: Not sure if it exists before or not, but as of now `nbqa-isort` does not exist in `PyPI`.

pyproject.toml
```diff
@@ -94,6 +94,18 @@ dev = [
     "wheel",
 ]
 
+dev_precommit = [
+    "black",
+    "check-manifest",
```
yarnabrina, line 99: Is `MANIFEST.in` file modified commonly? If not, this may not be required for usual contributors.

Personal curiosity, can you tell me what does `MANIFEST.in` do? I've few python packages (private for office work) and never needed to use it.

yarnabrina (changes requested):

PR #4078, opened 2023-01-07
[ENH] Added piecewise_poisson

Pyyim:
Co-authored-by: Kevin <kevinlamdata@gmail.com>

#### Reference Issues/PRs

Reference # 3363

Added to datagen.py in preparation for adding to PoissonHMM

No

##### For all contributions


sktime/annotation/datagen.py
```diff
@@ -241,6 +241,53 @@ def label_piecewise_normal(
...
+        Either a random seed or RandomState instance
+
+    Returns
+    -------
+    data : np.array
+        univariate time series as np.array
+
+    Examples
+    --------
+    >>> from sktime.annotation.datagen import piecewise_normal
```
miraep8, line 271: Example needs to be updated for poisson!

sktime/annotation/datagen.py
```diff
@@ -241,6 +241,53 @@ def label_piecewise_normal(
...
+    """
+    Generate series using Possion distribution.
+
+    Each segment has length specified in ``lengths`` and data sampled from a Poisson
+    distribution with expected frequency from ``frequency``.
+
+    Parameters
+    ----------
+    frequency : array_like
+        Expected frequency of the segments to be generated
```
miraep8, line 258: Small change - but I would recommend you call this "lamba" rather than "frequency" (just to follow the typical syntax seen with Poisson distributions) and explain that it is the expected number (and variance! Aren't Poisson distributions cool?) of events within each time interval/data point.

miraep8 (commented):
I have a few small change requests - but overall looks good! May I also request that if either of you are not in our contributors list already you go ahead and add yourselves? :)

miraep8:
Adding a quick test for this under sktime -> annotation -> tests would also not be such a bad idea!

sktime/annotation/datagen.py
lmmentel: This is one point where python and maths don't play well since `lambda` is a reserved keyword in python. I've been in this situation before and still don't have a good solution for an alternative name to something everyone knows should be called `lambda`.

sktime/annotation/datagen.py
```diff
@@ -241,6 +241,53 @@ def label_piecewise_normal(
...
+    >>> piecewise_normal([1, 2, 3], lengths=[2, 4, 8], random_state=42) # doctest: +SKIP
+    array([1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0])
+
+    >>> from sktime.annotation.datagen import piecewise_normal
+    >>> piecewise_normal([1, 3, 6], lengths=[2, 4, 8], random_state=42) # doctest: +SKIP
+    array([2, 2, 2, 2, 2, 2, 1, 0, 1, 0, 1, 0, 1, 0])
+
+    """
+    rng = check_random_state(random_state)
+    assert len(frequency) == len(lengths)
```
lmmentel, line 281: Would you mind turning `assert` into a `try`/`except` block raising human readable errors? The reason is that in some cases of production code there is a way to skip assert statements entirely (see e.g. this [link](https://realpython.com/python-assert-statement/#understanding-the-__debug__-built-in-constant)). Thanks!

lmmentel (changes requested):

lmmentel:
Thanks @Pyyim :raised_hands: All looks good except for the small comment I made on the `assert` statement that could be turned into a try/except to be more robust.

sktime/annotation/datagen.py
miraep8: Oh whoops, of course!  Perhaps in this case `lambdas` would work (since it is an array).  And then could use `lam` or similar when iterating over it.

Pyyim:
Thank you for the quick response. We will get to everything you have mentioned and resolve them by tomorrow.

Pyyim:
@miraep8 @lmmentel we have submitted the changes to the files that were requested. 

- included tests in a new file test_datagen.py
- updated examples, included the try/except block, and changed name of frequency to lambdas in datagen.py
- added ourselves to the .all-contributorsrc file

miraep8:
Excellent!  Thank you for submitting tests for `data_gen` as well! Unfortunately there seem to be some slight issues with the tests (can check out the logs in the CI/CD to see what exactly is failing.  I think it looks like an easy-to resolve issue though! :) 

Thanks again!

Pyyim:
@miraep8 Just finished fixing the issue you just mentioned. Thank you!

miraep8:
Hey @Pyyim, I think there are still some merge conflicts with this branch! :)  I think once those are resolved we will be in a much better place to merge this.

Pyyim:
@miraep8 Sorry about that. I believe it should be good to go now.

miraep8 (dismissed):

miraep8:
Nice work!  🎉 Unfortunately running into the issue with the small merge conflict from the contributors file again.  I tried to change it on my end but was running into some permissions issues, so perhaps might be better for you to fix!  I am planning to also merge the PR from your colleague @JonathanBechtel today as well (once tests pass) so it could just be worth waiting until that event, then resolving the merge conflicts.  Sorry - an unfortunate issue with lots of new contributors!

miraep8:
The other PR from a new contributor has been merged!  I hope now you should be able to get the merge conflict fixed with no more errors and we can hopefully quickly get this merged.  Thank you for your patience and your work on this! :)

Pyyim:
Hi, I currently don't have my computer right now as today is a holiday in the US, but I should be able to get to it first thing in the morning tomorrow. Thank you!

miraep8:
No worries! The "quickly" was more on myself - ie once you resolve the issue I will try to merge it quickly to avoid any future merge conflicts!  :)  Enjoy the day off!

Pyyim:
Hi @miraep8, 
Would it be possible to remove the all contributor file for this PR and create a separate PR for just for that?

miraep8:
If you do that we could then re-run the test suite on just your code changes and once those pass (sure they will based on the above) we can merge this and add your contribution separately! :)

Pyyim:
@miraep8 Would it actually be okay to create a new pull request? I believe I might have messed up with removing the all-contributors file, and it might be cleaner.

miraep8:
No worries :)  That works!

PR #4080, opened 2023-01-07
[ENH] Update latest version of channel selection

haskarb:
#### Reference Issues/PRs
Extension of #3248

Accepted version: https://arxiv.org/abs/2206.09274
No.

##### For all contributions

##### For new estimators


fkiraly (changes requested):
you are reverting the class to no longer accept arbitrary distances, and you are making changes to the content of it.

You are the owner of the algorithm, but this can't be done without deprecation.

The easiest way might be to just have a separate estimator which is a version 2 or so?

haskarb:
> The easiest way might be to just have a separate estimator which is a version 2 or so?

How can I do this? We can deprecate the older version.

fkiraly:
Well, there is a bit of a complication that we need to deal with:

1. I made some more recent changes to your old version (let's call it version A) so the estimator accepts arbitrary distances. This resulted in a version B, which is your old version but takes any distance.
2. it looks like you took version A, not version B, as a basis for the change in this PR, let's call it version C. As such, version C no longer seems to have the improvements in version B, e.g., the flexibility to choose the distance.

I see a few solutions that would make sense - and you can decide since you are the owner.

Solution 1: we replace version B by version C in `ElbowClassSum`, and make a new estimator (`ElbowClassSumWithDist`?) that is version B (owned by me). That would require deprecation since users who were using version B with arbitrary distances no longer can. It will probably be a bit fiddly, see here for the rules: https://www.sktime.org/en/stable/developer_guide/deprecation.html
As said, the requirement is that users who have running code don't suddenly end up with broken code when your new code is released.

Solution 2: we add a new estimator which is version C and corresponds to https://arxiv.org/abs/2206.09274. We could call that `ElbowClassSumV2`, and `ElbowClassSum` stays as it is. This would not require deprecation

Solution 3: you create a "merge" of version B and version C, i.e., an upgraded `ElbowClassSum`, which is downwards compatible with version B. This would not require deprecation, but might require substantial additional work.

My personal recommendation would be solution 2, because it does not require deprecation.

haskarb:
Hi Franz, thanks for the reply.

> Well, there is a bit of a complication that we need to deal with:

We will deal with this :) 

> 1. I made some more recent changes to your old version (let's call it version A) so the estimator accepts arbitrary distances. This resulted in a version B, which is your old version but takes any distance. You gave your approval here: [[ENH] channel selection (Dhariyal et al 2021) with arbitrary distance #3256](https://github.com/sktime/sktime/pull/3256)

Agreed.

> 2. it looks like you took version A, not version B, as a basis for the change in this PR, let's call it version C. As such, version C no longer seems to have the improvements in version B, e.g., the flexibility to choose the distance.

This version is more generalised form. We can use any distance metrics now, including from version A, and version B (thanks to sktime improvement meanwhile), and different class prototypes. 

I think this should be equivalent to solution 3.

fkiraly:
> I think this should be equivalent to solution 3.

I don't think this PR satisfies the conditions of solution 3. The condition it would have to fulfil, if a user has code that currently runs with `ElbowClassSum`, it will continue running after the change, and produce same or similar results.

I think some code can break because parameters and attributes are different.

For example, before the change, there was only a single parameter, `distance`, and it could be any `sktime` distance object.
After the change, it can only be a string, and there are other parameters which are not optional. So a previously valid constructor/specification line would break after the change.

You also deleted the tests that would ensure it doesn't - via the `get_test_params` method, which, if you would have left it in, would ensure that previous parameter choices are valid.

Also, names and number of attributes change which are returned by `get_fitted_params`.

haskarb:
>I don't think this PR satisfies the conditions of solution 3. The condition it would have to fulfil, if a user has code that currently runs with ElbowClassSum, it will continue running after the change, and produce same or similar results.

Makes sense.

How about I keep version B, along with version C (naming ElbowClassSumV2) and add a deprecation warning to version B? Eventually, I want to keep version C as the only method for channel selection.

fkiraly:
> How about I keep version B, along with version C (naming ElbowClassSumV2) and add a deprecation warning to version B? Eventually, I want to keep version C as the only method for channel selection.

Sure, would work.

I think it would make scientific sense to keep version B since it's a historic version, e.g., to perform comparisons between the two and because it's linked to one of your earlier papers - it's the algorithm that your ECML-PKDD 2021 comes with.

Just thinking along the lines of a library: libraries also don't throw away older books by an author just because the newer ones are nicer.

So instead of deprecating, how about adding in the docstring sth like "this version is no longer recommended by the author because...", or maybe you add a reference to the new paper and new class and say sth like this "the improved version described in [3] is available as `ElbowClassSumV2`"?

haskarb:
This is not yet ready for review.

fkiraly:
sure - we generally don't merge drafts

PR #4084, opened 2023-01-09
[DOC] Top100 contributions section added in the About/CONTRIBUTORS.md

Aarthy153:
**Description :** 

Top100 contributions section added in the About/CONTRIBUTORS.md

**Issue reference**

#3717 

**Screenshot**

image

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -28,7 +28,7 @@ If your contribution does add a new hard dependency, we may suggest to initially
 
 <!-- This section is particularly useful if you have a pull request that is still in development. You can guide the reviews to focus on the parts that are ready for their comments. We suggest using bullets (indicated by * or -) and filled checkboxes [x] here -->
 
-### Did you add any tests rather than the exisiting ones?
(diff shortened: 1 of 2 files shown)


Aarthy153:
Need your review @achieveordie @miraep8 @fkiraly

miraep8:
Hey @Aarthy153, will check this out in more detail once the tests run and we can checkout your change in the docs! :)   But - small note - in general best practice is to isolate each new issue/change to its own branch and PR.  I noticed here that you also have the changes from the pull request template change PR here as well!  I think we can work around it in this case (as both changes are pretty small and reasonable!) - but just letting you know best practice for future PRs! 

Also thank you for not letting the previous PR for this issue die and instead creating a new solution!  Much appreciated! :)

achieveordie:
@Aarthy153 I personally prefer "Top 100 Code Contributors" since GH tracks [code/doc] commit as the metric of sorting (in descending order). Regarding the current failures, it seems that you haven't set up your dev environment yet, hence causing the first check to fail. Did you follow the steps mentioned in the [developer guide](https://www.sktime.org/en/stable/developer_guide.html)?

If you have trouble setting it up (or think it could be improved) do let us know.

PR #4085, opened 2023-01-09
[DOC] update list of core devs

fkiraly:
Updates the list of core devs to be current.


patrickzib:
Please do not do this.

fkiraly:
@patrickzib, this is just the status quo. This is not a decision, it just updates the list to be accurate.

PR #4086, opened 2023-01-09
[GOV] core developers must contribute to `sktime`

fkiraly:
This updates eligibility requirements for core developers and CC members.

The gist is, core developers must continue to contribute to `sktime` - code or non-code.
Who wants to lead must first serve, and who is not willing to serve should not lead.

I think that's a pretty reasonable ask.

Separately from this, I think it is worth considering to extend the vote to developers who *would* be eligible but do not want to become core developers, e.g., very active members of the community.

I'm not very precious about the list of what counts as contributions - I probably did not list things that should be there, it might also be worth to make some points precise. Suggestions appreciated.

### DIFF
--- a/docs/source/get_involved/governance.rst
+++ b/docs/source/get_involved/governance.rst
@@ -264,6 +264,9 @@ Core developers are appointed for a period of twelve months.
 For continued tenure, they have to present a summary of their contributions within
 the last twelve months and meet the same contribution threshold.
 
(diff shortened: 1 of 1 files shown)


patrickzib:
Please stop escalating. Take some time to think about it.

fkiraly:
superseded by governance widening working group

PR #4095, opened 2023-01-11
[BUG] fix - Callbacks cause DL estimators to fail

### LINKED ISSUE #4088: [BUG] Callbacks cause DL estimators to fail
**Describe the bug**
Constructing a CNNClassifier with callbacks causes an error during training because the self.callbacks is never assigned to self._callbacks if self.callbacks is not None
Error:
`AttributeError: 'CNNClassifier' object has no attribute '_callbacks'`
problem area: https://github.com/sktime/sktime/blob/8fe509628405eaeaa6cbc26351e946c4a8a54687/sktime/classification/deep_learning/cnn.py#L187

**To Reproduce**
```python
import numpy as np
from sktime.classification.deep_learning import CNNClassifier

model = CNNClassifier(callbacks=[])
X_train = np.array(np.random.random((100, 1000)), dtype=np.float64)
y_train = np.array(np.random.randint(0, 2, size=(100,)), dtype=np.float64)
model.fit(X_train, y_train)
```

**Expected behavior**
self.classifier should be assigned to self._classifier OR self.classifier is what should be accessed on line 187

**Versions**
sktime==0.15.0
Please run the following code snippet and paste the output here:
 
from sktime import show_versions; show_versions()
```
System:
    python: 3.8.10 (default, Jun 22 2022, 20:18:18)  [GCC 9.4.0]
executable: /usr/bin/python3
   machine: Linux-5.10.16.3-microsoft-standard-WSL2-x86_64-with-glibc2.29

Python dependencies:
               pip: 22.3.1
   setuptools: 65.5.1
         sklearn: 1.2.0
          sktime: 0.15.0
 statsmodels: 0.13.5
         numpy: 1.23.5
            scipy: 1.9.3
         pandas: 1.5.2
    matplotlib: 3.6.2
           joblib: 1.2.0
         numba: 0.56.4
    pmdarima: 2.0.2
          tsfresh: 0.19.0
None
```

aaronrmm:
#### Reference Issues/PRs
Fixes #4088.

I replaced each assignment and reference of `self._callback` with `self.callback`

No.

- Is there any reflection code that treats variables starting with underscore differently? 
- Should these be expanded to more classes?
#### Any other comments?
None.

##### For all contributions

##### For new estimators


fkiraly (changes requested):
Thanks!

There is a problem which we need to deal with here: to comply with the `sklearn` interface, arguments of `__init__` must be written to attributes of the same name, and never changed later on (e.g., by `_fit`).
It seems your fix does violate that important interface contract, which is important in tuning, composing with other estimators to pipelines, etc.

Could we think about a fix that:
* ensures that `self.callbacks` is never overwritten after it is set by `__init__`
* fixes the bug
* comes with a test for the `callbacks` that fails before the fix and runs after the fix?

fkiraly:
(that´s what the failing tests confirm - `fit` overwrites the hyperparam `callback`)

aaronrmm:
I'm having trouble getting these tests to fail locally. running python -m pytest in the root directory gives me:
`== 29793 passed, 578 skipped, 9 xfailed, 1 xpassed, 115517 warnings in 614.99s (0:10:14) ==`


My thought on fixing this was to make self.callback a property for these models:

```
def __init__(self, ..., callbacks=None):
  self._callbacks = callbacks or []
  
@property
def callbacks(self):
  return self._callbacks

def _fit(self, X, y):
  # revert fit back to how it was before this PR
```

@fkiraly What do you think?

fkiraly:
> I'm having trouble getting these tests to fail locally.

That is very weird. Have you tried using `check_estimator` instead? That should be quicker as well, than running all the tests in all the estimators.

> My thought on fixing this was to make self.callback a property for these models:

That probably wouldn't work, since the `scikit-learn` contract (which `sktime` follows) assumes for all `__init__` arguments, lets' say, `argname`, that *at any point* the call `self.argname` gives the value of `argname ` that was passed at `__init__`. Using the property trick would break that contract.

fkiraly:
@aaronrmm, @achieveordie, could you kindly help me to understand the issue?

I understand the constract constraints coming from the `scikit-learn` interface, and I suspect they are clashing with another contract assumption around the `callbacks` attribute (see above).

But right now I cannot with certainty spell out that other contract assumption or condition.

Does it make sense to one of you what I'm saying, and could you help formally write down the second assumption/condition/contract that we seem to be clashing with here?

achieveordie:
The problem stems from the fact that the callback parameter itself is changed by tensorflow. ([Here](https://github.com/keras-team/keras/blob/e6784e4302c7b8cd116b74a784f4b78d60e83c26/keras/callbacks.py#L227) is an example when `None` is changed to `[]`). We should assume that any callback that we're passing is bound to be "changed" which is in contrast to the scikit-learn contract.

A way to escape this apparent paradox:

```python
def _fit(self, X, y):
    from copy import deepcopy
    _callbacks = [] if self.callbacks is None else deepcopy(self.callbacks)
    ...
```

(This has been tested by `check_estimator()` and it passes.)

fkiraly:
>A way to escape this apparent paradox:

Ahhh, I get it now!

The original code tried to comply with the `sklearn` code by writing to `_callbacks`, but that caused a problem since it was copied by reference, not just copied as a clone.

So, the solution by @achieveordie would work in general for this - write to `_callbacks`, but also deep-copy.

sktime/regression/deep_learning/cnn.py
```diff
@@ -162,8 +162,7 @@ def _fit(self, X, y):
         -------
         self : object
         """
-        if self.callbacks is None:
-            self._callbacks = []
+        self.callbacks = self.callbacks or []
```
achieveordie, line 165: The problem is here where accidental overwriting is taking place, we should keep the structure the same as it was for classification estimators.

sktime/regression/deep_learning/tapnet.py
```diff
@@ -207,8 +207,7 @@ def _fit(self, X, y):
         -------
         self: object
         """
-        if self.callbacks is None:
-            self._callbacks = []
+        self.callbacks = self.callbacks or []
```
achieveordie, line 210: Same issue as in `CNNRegressor`

achieveordie (changes requested):
Minor changes, nothing substantial. All the classifiers passing shows that we were indeed correct.

sktime/regression/deep_learning/tapnet.py
aaronrmm: Okay. I made the same changes to the deep regressors.

fkiraly (changes requested):
Thank you! This seems to fix it.

May I ask for tests that check whether it's fixed?

The simplest might be having a test parameter set where the parameter `callbacks` is used, for all the estimators fixed?
That would be in `get_test_params`.

aaronrmm:
I had some trouble figuring out how to get these tests to fail in main. I was trying to exclude other tests to isolate the test I wanted and found I couldn't exclude tests, and also noticed test_fit_idempotent was being run for CNNClassifier. I'm not sure if that's a bug or I was just doing thing incorrectly. Anyway at some point I saw the error message I wanted.

fkiraly:
> I had some trouble figuring out how to get these tests to fail in main. I was trying to exclude other tests to isolate the test I wanted and found I couldn't exclude tests, and also noticed test_fit_idempotent was being run for CNNClassifier. I'm not sure if that's a bug or I was just doing thing incorrectly. Anyway at some point I saw the error message I wanted.

Hm, let me know if you would like me to look into this.
If you think something is broken, can you be specific about what exactly happened (and what did you run) and what you think should have happened instead?

fkiraly:
The soft dependency tests are failing since `keras` is not isolated in the estimator creation.

Examples how to do that is in `get_test_params` of `RegressorPipeline`, or of `BaggingForecaster`. Slightly clunky, but ensures that the test suite can be run without soft dependencies.

aaronrmm:
Sorry, I made the last set of changes before I saw this.

> The soft dependency tests are failing since `keras` is not isolated in the estimator creation.
> 
> Examples how to do that is in `get_test_params` of `RegressorPipeline`, or of `BaggingForecaster`. Slightly clunky, but ensures that the test suite can be run without soft dependencies.

This guy in regression.compose._pipeline.py?
```
        if _check_soft_dependencies("numba", severity="none"):
            from sktime.regression.distance_based import KNeighborsTimeSeriesRegressor

            c = KNeighborsTimeSeriesRegressor()

            # construct without names
            params2 = {"transformers": [t1, t2], "regressor": c}
            return [params1, params2]
        else:
            return params1
```

So I'd do something like

```
    def get_test_params(cls, parameter_set="default"):
        param1 = {...}
        param2 = {...}

        if _check_soft_dependencies("keras", severity="none"):
            from keras.callbacks import LambdaCallback

            param3 = {
                "n_epochs": 2,
                "callbacks": [LambdaCallback()],
            }
            return [param1, param2, param3]
        else:
            return [param1, param2]
```
?

fkiraly:
yes, exactly.
The proposed code works - a tiny bit simpler would be only "if" with "then append".

fkiraly (approved):
Looks good now!

achieveordie (approved):
Thanks for the fix.

Issue #3836, opened 2022-11-25
[GOV] security architecture for release and community assets - separation of duties, MFA 


fkiraly:
`sktime` is starting to attract a sizeable community on its social channels (many 100s) and is starting to have a substantial reverse dependency footprint.

We should therefore consider safeguards against malicious actors in our operations model, both from external and internal.

Two critical points in `sktime`'s security architecture are imo currently:

* the release process, and control of the `pypi` wheels, e.g., for a [software supply chain attack](https://en.wikipedia.org/wiki/Supply_chain_attack). These have become more frequent recently.
* community assets, communication and operations channels including GitHub accounts of developers, control of the repositories, and social channels Slack, Discord, twitter

Motives for attackes can differ, e.g.:

* criminal actors using `sktime` as a stepping stone for ransomware attacks via SSCA
* nation state actors using `sktime` for SSCA, e.g., in the context of cyberattacks
* community assets used as leverage in community internal social conflicts 
* community assets being seized as a supportive measure to enhance length, scope, or scale of an SSCA

In my opinion, the key tools to mitigate any of the above risks are:

* [separation of duties](https://en.wikipedia.org/wiki/Separation_of_duties) - reducing the risk that a single actor can seize control of an entire process of all assets, by increasing the number of separate responsibilities or control points
* [multi-factor authentication](https://en.wikipedia.org/wiki/Multi-factor_authentication) - reducing the risk of impersonation of elevated rights by a malicious actor by increasing the number of independent authentication points

This issue is also about discussing:

* are there other key critical points that need to be covered by security architecture?
* are there other measures than the above that we should consider?
* what are the trade-offs, e.g., with operational efficiency
* what is the best implementation of said measures?

fkiraly:
Mapping community assets and maximal roles, with worst case action radius:

* GitHub organisation admin account. This role "trumps" GitHub repo admins. It can seize control of all GitHub assets by removing other admins. There can be multiple GitHub repo or org admins.
* pypi account. This role can seize control of the package on pypi. There can be only one such account afaik.
* Slack primary owner. This role can seize control of the entire slack space. There can be only one such role holder.
* Discord primary owner. This role can seize control of the entire discord space. There can be only one such role holder.
* personal email accounts associated with the above roles. 
* sktime email address accounts.
* not sure about linkedin, twitter

Regarding trump relations:
[...]

PR #4108, opened 2023-01-13
[ENH] Refactor/simplify `sktime.forecasting.model_selection._split.BaseSplitter._split_vectorized`

### LINKED ISSUE #2695: [ENH] Refactor/simplify `sktime.forecasting.model_selection._split.BaseSplitter._split_vectorized`
`sktime.forecasting.model_selection._split.BaseSplitter._split_vectorized` is very hard to read and understand.

This method was introduced in https://github.com/alan-turing-institute/sktime/pull/2599. During the review I proposed the following refactoring idea: https://github.com/alan-turing-institute/sktime/pull/2599#issuecomment-1126736598 (see comments that followed). For convenience, I copy my main suggestion here:

> I looked a bit closer inside of the sktime.forecasting.model_selection._split.BaseSplitter._split_vectorized. Again, I completely understand the result, for each series in the hierarchy you want to get train/test windows with numeration in the context of one big stacked time series. But I am worried about the implementation of this method. There are several enclosed loops, several temporary objects.

> For each time series in hierarchy you can easily get train/test windows that start counting from zero in isolation for each series. What you want to do next is shift those windows by the iloc indices of each series beginning. Say you have two windows in isolation: [0, 1, 3] (out of [0, 1, 2, 3]), and [0, 2] (out of [0, 1, 2]). The index of the whole series is [0, 1, 2, 3, 4, 5, 6]. The "anchors" are then 0 and 4. So your final windows are [0, 1, 3] + 0 = [0, 1, 3], and [0, 2] + 4 = [4, 6]. I know, it is just a toy example, but isn't it easy to get those iloc "anchors"?

mateuja:
#### Reference Issues/PRs

Resolves #2695

* Refactor the `sktime.forecasting.model_selection._split.BaseSplitter._split_vectorized` method to improve readability.
* Simplify other small fragments of code in the same module.

Are there any concerns regarding the new implementation of the `_split_vectorized` method? @fkiraly would be great to hear your thoughts on this, since you wrote the original code.

No, I just checked that previous tests still pass.

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/forecasting/model_selection/_split.py
+++ b/sktime/forecasting/model_selection/_split.py
@@ -363,10 +363,10 @@ class BaseSplitter(BaseObject):
         """
         y_index = self._coerce_to_index(y)
 
-        if isinstance(y_index, pd.MultiIndex):
-            split = self._split_vectorized
-        else:
+        if not isinstance(y_index, pd.MultiIndex):
             split = self._split
+        else:
+            split = self._split_vectorized
 
         for train, test in split(y_index):
             yield train[train >= 0], test[test >= 0]
--- a/sktime/forecasting/model_selection/_split.py
+++ b/sktime/forecasting/model_selection/_split.py
@@ -394,7 +394,7 @@ class BaseSplitter(BaseObject):
         """
         raise NotImplementedError("abstract method")
 
-    def _split_vectorized(self, y: pd.Index) -> SPLIT_GENERATOR_TYPE:
+    def _split_vectorized(self, y: pd.MultiIndex) -> SPLIT_GENERATOR_TYPE:
         """Get iloc references to train/test splits of `y`, for pd.MultiIndex.
 
         This applies _split per time series instance in the multiindex.
@@ -412,51 +412,22 @@ class BaseSplitter(BaseObject):
         test : 1D np.ndarray of dtype int
             Test window indices, iloc references to test indices in y
         """
-        # challenge is obtaining iloc references for *the original data frame*
-        #   todo: try to shorten this, there must be a quicker way
-        train_test_res = dict()
-        train_iloc = dict()
-        test_iloc = dict()
-        y = pd.DataFrame(index=y)
-        y_index_df = y.reset_index(-1)
-        y_index_df["__index"] = range(len(y_index_df))
-        y_index_inst = y_index_df.index.unique()
-        for idx in y_index_inst:
-            train_iloc[idx] = dict()
-            test_iloc[idx] = dict()
-        for idx in y_index_inst:
-            y_inst = y_index_df.loc[idx]
-            y_inst = y_inst.reset_index(drop=True).set_index(y_inst.columns[0])
-            y_inst_index = y_inst.index
-            train_test_res[idx] = list(self._split(y_inst_index))
-            for i, tt in enumerate(train_test_res[idx]):
-                train_iloc[idx][i] = tt[0]
-                test_iloc[idx][i] = tt[1]
-            for i, train in train_iloc[idx].items():
-                train_iloc[idx][i] = y_inst["__index"].iloc[train].values
-            for i, test in test_iloc[idx].items():
-                test_iloc[idx][i] = y_inst["__index"].iloc[test].values
+        srs = pd.DataFrame(index=y).reset_index(-1).iloc[:, 0]
+        index = srs.index
+        anchors = pd.Series(range(len(srs)), index).groupby(index).first().tolist()
+        splits = (self._split(pd.Index(inst.values)) for _, inst in srs.groupby(index))
 
-        train_multi = dict()
-        test_multi = dict()
-        for idx in y_index_inst:
-            for i in train_iloc[idx].keys():
-                if i not in train_multi.keys():
-                    train_multi[i] = train_iloc[idx][i]
-                    test_multi[i] = test_iloc[idx][i]
-                else:
-                    train_multi[i] = np.concatenate(
-                        (train_iloc[idx][i], train_multi[i])
-                    )
-                    test_multi[i] = np.concatenate((test_iloc[idx][i], test_multi[i]))
-        for i in train_multi.keys():
-            train_multi[i] = np.sort(train_multi[i])
-            test_multi[i] = np.sort(test_multi[i])
+        train = []
+        test = []
+        for split_inst, anchor in zip(splits, anchors):
+            train_inst, test_inst = zip(*split_inst)
+            train.append(tuple(indices + anchor for indices in train_inst))
+            test.append(tuple(indices + anchor for indices in test_inst))
 
-        for i in train_multi.keys():
-            train = train_multi[i]
-            test = test_multi[i]
-            yield train, test
+        train = map(np.concatenate, zip(*train))
+        test = map(np.concatenate, zip(*test))
+
+        yield from zip(train, test)
 
     def split_loc(self, y: ACCEPTED_Y_TYPES) -> Iterator[Tuple[pd.Index, pd.Index]]:
         """Get loc references to train/test splits of `y`.
--- a/sktime/forecasting/model_selection/_split.py
+++ b/sktime/forecasting/model_selection/_split.py
@@ -348,7 +348,7 @@ class BaseSplitter(BaseObject):
         super(BaseSplitter, self).__init__()
 
     def split(self, y: ACCEPTED_Y_TYPES) -> SPLIT_GENERATOR_TYPE:
-        """Get iloc references to train/test slits of `y`.
+        """Get iloc references to train/test splits of `y`.
 
         Parameters
         ----------
--- a/sktime/forecasting/model_selection/_split.py
+++ b/sktime/forecasting/model_selection/_split.py
@@ -96,8 +96,7 @@ def _get_end(y_index: pd.Index, fh: ForecastingHorizon) -> int:
     fh_offset = null if fh.is_all_in_sample() else fh[-1]
     if array_is_int(fh):
         return n_timepoints - fh_offset - 1
-    else:
-        return y_index.get_loc(y_index[-1] - fh_offset)
+    return y_index.get_loc(y_index[-1] - fh_offset)
 
 
 def _check_window_lengths(
@@ -180,10 +179,7 @@ def _inputs_are_supported(args: list) -> bool:
     -------
     True if all inputs are compatible, False otherwise
     """
-    if all_inputs_are_iloc_like(args) or all_inputs_are_time_like(args):
-        return True
-    else:
-        return False
+    return all_inputs_are_iloc_like(args) or all_inputs_are_time_like(args)
 
 
 def _check_inputs_for_compatibility(args: list) -> None:
@@ -367,10 +363,10 @@ class BaseSplitter(BaseObject):
         """
         y_index = self._coerce_to_index(y)
 
-        if not isinstance(y_index, pd.MultiIndex):
-            split = self._split
-        else:
+        if isinstance(y_index, pd.MultiIndex):
             split = self._split_vectorized
+        else:
+            split = self._split
 
         for train, test in split(y_index):
             yield train[train >= 0], test[test >= 0]
@@ -637,8 +633,7 @@ class BaseSplitter(BaseObject):
             return np.argwhere(
                 (y >= y[max(train_start, 0)]) & (y <= y[min(split_point, len(y)) - 1])
             ).flatten()
-        else:
-            return np.array([], dtype=int)
+        return np.array([], dtype=int)
 
 
 class CutoffSplitter(BaseSplitter):
@@ -763,8 +758,7 @@ class CutoffSplitter(BaseSplitter):
         """
         if array_is_int(self.cutoffs):
             return check_cutoffs(self.cutoffs)
-        else:
-            return np.argwhere(y.index.isin(check_cutoffs(self.cutoffs))).flatten()
+        return np.argwhere(y.index.isin(check_cutoffs(self.cutoffs))).flatten()
 
     @classmethod
     def get_test_params(cls, parameter_set="default"):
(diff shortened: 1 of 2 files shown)


sktime/forecasting/model_selection/_split.py
```diff
@@ -367,10 +363,10 @@ def split(self, y: ACCEPTED_Y_TYPES) -> SPLIT_GENERATOR_TYPE:
         """
         y_index = self._coerce_to_index(y)
 
-        if not isinstance(y_index, pd.MultiIndex):
```
fkiraly, line 370: hm, this one I would leave like it is, based on "default case first"

not a strong opinion though

fkiraly (changes requested):
Thanks! This looks amazing! Much simpler!

No strictly a blocking comment, but I think we need to explicitly comment on this one:
How do we know the new code does the same as the old code? The devil is in the details usually. Is there a test that covers this sufficiently? (if not, that's probably my bad, but I suspect @khrapovs would not have let me merge without a test)
We should just identify the test and spell out why there is an external check.

Also not a blocker, but a potential bonus task that occurs to me since you are also changing code in `get_test_params`: normally, this should return at least two test parameter sets for any `sktime` object, but in this module it seems it returns only one. This is the original issue: https://github.com/sktime/sktime/issues/3429, and it might be worth doing that in a separate PR if you are interested and anyway working on this module (I suggest separate PR since it may throw up bugs that went undetected due to low test coverage)

sktime/forecasting/model_selection/_split.py
```diff
@@ -96,8 +96,7 @@ def _get_end(y_index: pd.Index, fh: ForecastingHorizon) -> int:
     fh_offset = null if fh.is_all_in_sample() else fh[-1]
     if array_is_int(fh):
         return n_timepoints - fh_offset - 1
-    else:
-        return y_index.get_loc(y_index[-1] - fh_offset)
+    return y_index.get_loc(y_index[-1] - fh_offset)
```
khrapovs, line 99: out of curiosity, why?

sktime/forecasting/model_selection/_split.py
```diff
@@ -666,8 +633,7 @@ def _get_train_window(
             return np.argwhere(
                 (y >= y[max(train_start, 0)]) & (y <= y[min(split_point, len(y)) - 1])
             ).flatten()
-        else:
-            return np.array([], dtype=int)
+        return np.array([], dtype=int)
```
khrapovs, line 636: same here, why?

sktime/forecasting/model_selection/_split.py
```diff
@@ -792,8 +758,7 @@ def get_cutoffs(self, y: Optional[ACCEPTED_Y_TYPES] = None) -> np.ndarray:
         """
         if array_is_int(self.cutoffs):
             return check_cutoffs(self.cutoffs)
-        else:
-            return np.argwhere(y.index.isin(check_cutoffs(self.cutoffs))).flatten()
+        return np.argwhere(y.index.isin(check_cutoffs(self.cutoffs))).flatten()
```
khrapovs, line 761: why?

sktime/forecasting/model_selection/_split.py
```diff
@@ -840,8 +804,7 @@ def __init__(
     def _initial_window(self):
         if hasattr(self, "initial_window"):
             return self.initial_window
-        else:
-            return None
+        return None
```
khrapovs, line 807: why?

sktime/forecasting/model_selection/_split.py
```diff
@@ -1478,7 +1441,6 @@ def _split_by_fh(
     if X is None:
         return y_train, y_test
 
-    else:
-        X_train = X.loc[train]
-        X_test = X.loc[test]
-        return y_train, y_test, X_train, X_test
+    X_train = X.loc[train]
+    X_test = X.loc[test]
+    return y_train, y_test, X_train, X_test
```
khrapovs, line 1446: why?

khrapovs (changes requested):

mateuja:
> Thanks! This looks amazing! Much simpler!
> 
> No strictly a blocking comment, but I think we need to explicitly comment on this one: How do we know the new code does the same as the old code? The devil is in the details usually. Is there a test that covers this sufficiently? (if not, that's probably my bad, but I suspect @khrapovs would not have let me merge without a test) We should just identify the test and spell out why there is an external check.
> 
> Also not a blocker, but a potential bonus task that occurs to me since you are also changing code in `get_test_params`: normally, this should return at least two test parameter sets for any `sktime` object, but in this module it seems it returns only one. This is the original issue: #3429, and it might be worth doing that in a separate PR if you are interested and anyway working on this module (I suggest separate PR since it may throw up bugs that went undetected due to low test coverage)

Thanks a lot for your feedback, I really appreciate it.

I am not very familiar with the testing framework of sktime yet, but to make sure that the new code works correctly, I think that we should confirm that there are enough tests using the `_split_vectorized` method in different situations (for example, different Window Splitters) and that there is at least one of them checking the output of the method in more detail. I executed all tests enabling [dynamic contexts](https://coverage.readthedocs.io/en/coverage-5.0.3/contexts.html#dynamic-contexts), which allowed me to get the following list of tests that cover the method:
- `sktime.forecasting.model_evaluation.tests.test_evaluate.test_evaluate_hierarchical`, which runs the `evaluate` function with an `SlidingWindowSplitter`.
- `sktime.forecasting.model_selection.tests.test_split.test_split_series_hier`, which computes different splits with an `SlidingWindowSplitter` and asserts multiple conditions on them.
- `sktime.forecasting.model_selection.tests.test_tune.test_gscv_hierarchical`, which uses a `ForecastingGridSearchCV` and is parameterized to be used with different `SingleWindowSplitter` instances and an `SliddingWindowSplitter`. 
- `sktime.forecasting.tests.test_all_forecasters.TestAllForecasters.test_get_fitted_params`
- `sktime.forecasting.tests.test_all_forecasters.TestAllForecasters.test_hierarchical_with_exogeneous`

To be honest, I do not really understand how the last two tests are related to the `_split_vectorized` method. Could you give me some information on how they work? I am probably not familiar enough with the code to decide if the existing tests are sufficient or not. In any case, please let me know if you think that I should extend a test or create a new one to validate a functionality that has not been covered yet.

Regarding the `get_test_params` issue (#3429), I am willing to take on that task, but I won't have the time to work on it during the next two weeks. If the issue is still unresolved by the end of the month, I will try to address it then.

sktime/forecasting/model_selection/_split.py
mateuja: I don't have a strong opinion on this either, so I restored the previous code.

sktime/forecasting/model_selection/_split.py
mateuja: It is just to make the code more concise, since the `else` clause is actually not needed. Some linters like pylint discourage the use of an `else` after a return statement ([see this](https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/no-else-return.html)). In any case, it's true that in the end it is probably just a matter of style. I can change it back to the original code if you prefer it.

sktime/forecasting/model_selection/_split.py
mateuja: same as in the other comment (https://github.com/sktime/sktime/pull/4108#discussion_r1070364781)

sktime/forecasting/model_selection/_split.py
mateuja: same as in the other comment (https://github.com/sktime/sktime/pull/4108#discussion_r1070364781)

sktime/forecasting/model_selection/_split.py
mateuja: same as in the other comment (https://github.com/sktime/sktime/pull/4108#discussion_r1070364781)

sktime/forecasting/model_selection/_split.py
mateuja: same as in the other comment (https://github.com/sktime/sktime/pull/4108#discussion_r1070364781)

fkiraly:
Ok, thanks a lot, very illuminating!

Let me give a bit of explanation on tests, perhaps we can piece together the testing of `_split_vectorized`.

First about the general testing strategy:

* `sktime` objects inherit from `BaseObject` which is an `sklearn`-like base interface with `get_params` etc. From it inherit base classes, such as `BaseSplitter` and `BaseForecaster`.
* For each such generic base class, there is usually a suite of tests. E.g., the testing class `TestAllObjects` is applied to all descendants of `BaseObjects`, the class `TestAllForecasters` is applied to all descendants of `BaseForecaster`.
* in addition, there may be "loose tests" using `pytest`.
* further details on the test framework are here: https://www.sktime.org/en/stable/developer_guide/testing_framework.html

Now, about splitters specifically, i.e., the descendants of `BaseSplitter`:
* It seems there is no tester class for the splitters. So, the only generic tests applied to descendants of `BaseSplitters` are the same that it inherits from `BaseObjects`, i.e., the test class `TestAllObjects`.
* there are further tests for splitters in `model_selection.tests.test_splits`. These do not seem comprehensive though.
* Out of these, most do not seem to cover directly the vectorized/broadcast case that would lead to invocation of `_split_vectorized`. However, some call the functionality as part of a bigger integration test, the ones you listed.

fkiraly:
Regarding the tests you listed, the general thing to know about the splitters is that `_split_vectorized` is called whenever a splitter receives panel or hierarchical data, e.g., the data generated by `_make_hierarchical` or `_make_panel`. Every time that happens, we would expect to see a call to `_split_vectorized`.

In the tests you listed, here is why that happens:

-`sktime.forecasting.model_evaluation.tests.test_evaluate.test_evaluate_hierarchical`, the data passed is hierarchical, and `evaluate` applies a splitter to that
- `sktime.forecasting.model_selection.tests.test_split.test_split_series_hier` - this is actually a direct application, a "true" unit test for splitters applied to hierarchical data. It only tests one splitter though, the `SlidingWindowSplitter`.
- `sktime.forecasting.model_selection.tests.test_tune.test_gscv_hierarchical` - grid search tuners use splitters internally. A tuner is applied to hierarchical data, which means a splitter is applied to hierarchical data.
- `sktime.forecasting.tests.test_all_forecasters.TestAllForecasters.test_hierarchical_with_exogeneous` - this is a generic test for forecasters (`BaseForecaster` descendants) with hierarchical data. The coverage must come from the tuners, as they apply splitters internally, as discussed above.
- `sktime.forecasting.tests.test_all_forecasters.TestAllForecasters.test_get_fitted_params` - I'm not actually sure why this one appears.

fkiraly:
(overall, looking at this: I think the tests are currently covering enough of `_split_vectorized`; but, ultimately, we may want to have a specific test class for `BaseSplitter` descendants. I'll open an issue, this may be a bit of a project)

fkiraly (approved):
Thanks for the contribution and discussion, @mateuja!

With the discussion of the tests, I'm convinced that this is probably logic equivalent.

sktime/forecasting/model_selection/_split.py
khrapovs: Interesting. I did not know of this linter warning. Still curious what is the deep reason behind it.

khrapovs (approved):

sktime/forecasting/model_selection/_split.py
mateuja: I have just briefly read some blog posts about this topic ([link](https://blog.mozilla.org/nnethercote/2009/08/31/no-else-after-return-considered-harmful/), [link](https://dmerej.info/blog/post/else-after-return-yea-or-nay/)) and it appears that there is ongoing debate. Ultimately, it seems that the two main points of discussion are readability (which can be quite subjective) and whether it is better to use a functional approach or if it is acceptable to rely on side-effects.

Although I generally prefer the functional style, in this situation I like the conciseness of the approach without else clauses. That being said, after reading those posts, I have come to realize that I may have been following the guidelines of Pylint without questioning the reasoning behind it. If I encounter a similar code pattern in the future, I will likely not make any changes to it as I now understand that it is likely a valid option too.

mateuja:
Thanks @fkiraly and @khrapovs for your feedback. I learned a lot working on this PR, even if it was just a small change. I definitely plan on continuing to contribute to sktime in the future.

Just one more question: is there a certain way you usually handle resolving conversations? Should I handle it or will you take care of it?

sktime/forecasting/model_selection/_split.py
khrapovs: Nice! Thanks! I've learned something too :)

khrapovs:
> Just one more question: is there a certain way you usually handle resolving conversations? Should I handle it or will you take care of it?

Clicking the magic button "Resolve conversation" is entirely the privilege/responsibility of a reviewer. How else can I track if you addressed suggestions/comments/etc? ;)

khrapovs:
> Clicking the magic button "Resolve conversation" is entirely the privilege/responsibility of a reviewer. How else can I track if you addressed suggestions/comments/etc? ;)

Strange... I can not find this magic button. I was able to approve, but can not resolve my own conversations... @fkiraly perhaps you would know, why?

sktime/forecasting/model_selection/_split.py
fkiraly: if it adds anything - I'm neutral on the issue.

I would consider using `else` as slightly more readable, as it is more verbose - it makes it very explicit that the remaining code is executed only if the condition above is `False`. Whereas otherwise one needs to infer (rather than read) that this is the case, but it's not too hard because understanding `return` is python 101.

fkiraly:
> Strange... I can not find this magic button. I was able to approve, but can not resolve my own conversations... @fkiraly perhaps you would know, why?

@khrapovs, we've had some internal discussions that also touched asset security, see here: https://github.com/sktime/sktime/issues/3836
There are some unresolved issues (as you may glean if you look at the private dev channels), until these are resolved I have minimized contributor credentials that do not restrict current contributor behaviour.

Since you haven't been active in a while, and also have told me you probably wouldn't be, I reset your credentials to "read" - I have now upgraded you to "write" again, since you seem to be back :-)

Issue #4103, opened 2023-01-13
Fix automated Contributors.md update


RNKuhns:
Whoops, it currently fails with the following:
```
[main 46fc2ca] [AUTOMATED] update CONTRIBUTORS.md
 Author: achieveordie <achieveordie@users.noreply.github.com>
 1 file changed, 53 insertions(+), 33 deletions(-)
INPUT_TAGGING_MESSAGE: 
No tagging message supplied. No tag will be added.
INPUT_PUSH_OPTIONS: 
remote: error: GH006: Protected branch update failed for refs/heads/main.        
remote: error: At least 1 approving review is required by reviewers with write access. 17 of 17 required status checks are expected.        
To https://github.com/sktime/sktime
 ! [remote rejected] main -> main (protected branch hook declined)
error: failed to push some refs to 'https://github.com/sktime/sktime'
```

__Originally posted by @achieveordie in https://github.com/sktime/sktime/pull/3807#issuecomment-1382086216__

RNKuhns:
Am I understanding it right, that this fails b/c main is protected branch?

Could we just have it open a new PR?

These should be pretty easy approvals.

fkiraly:
> Am I understanding it right, that this fails b/c main is protected branch?

I don't "know" know, but given the error message it feels like a pretty safe bet.

fkiraly:
> Could we just have it open a new PR?

Could it instead just push to the branch that triggers it, pre-merge?

achieveordie:
We could follow the route [recommended](https://github.com/stefanzweifel/git-auto-commit-action#push-to-protected-branches) by the author of this GH Action.

More specifically (emphasis mine):
> **You need a Personal Access Token and** you either have to allow force pushes or **the Personal Access Token needs to belong to an Administrator**.
> ...
> First, you have to create a new [Personal Access Token (PAT)](https://github.com/settings/tokens/new), store the token as a secret in your repository and pass the new token to the [actions/checkout](https://github.com/actions/checkout#usage) Action step.
> ...
> Note: If you're working in an organisation, and you don't want to create the PAT from your personal account, we recommend using a bot-account for such tokens.

achieveordie:
> Could it instead just push to the branch that triggers it, pre-merge?

This is what I originally intended. [This](https://github.com/stefanzweifel/git-auto-commit-action/issues/211) issue might be helpful in that regard. WDYT? @RNKuhns @fkiraly

fkiraly:
[...]

PR #4123, opened 2023-01-18
[ENH] [DO NOT MERGE] Trigger `update_contributors.yml` workflow to run on `pull_request_target`

achieveordie:
#### Reference Issues/PRs

Reference #4103

This is an alternative proposal to the current `update_contributors.yml` according to which the workflow will run pre-merge in the context of base PR. Originally, this was meant to be a replacement for the current structure, but recent findings [show](https://github.com/sktime/sktime/issues/4103#issuecomment-1386533283) that it is not necessary for proper functioning.

**_Important_:** Merging the current workflow changes will [introduce](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) security vulnerabilities and hence should not be merged till finalized.

No

Firstly, we need to discuss if this alternative is required. If it is then we need to rewrite the workflow file in such a way that we don't accidentally introduce any exploitable nodes.

Assuming we choose to move forward with an alternative, there are two things we can do to:
1. Run this workflow only after approval from a core developer (possibly easy to implement but responsibilities fall on core developers to go through changed files to confirm there are no malicious/fishy changes.)
2. Divide the workflow into two steps, where artifacts are built and saved in a less privileged environment followed by downloading the specific changes and getting the result. See [link](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) for example (difficult to implement, resulting in longer runtime of the job but more secure than the first)

No

##### For all contributions

### DIFF
--- a/.github/workflows/update_contributors.yml
+++ b/.github/workflows/update_contributors.yml
@@ -1,9 +1,8 @@
 name: Update Contributors
 
 on:
-  push:
-    branches:
-      - main
+  pull_request_target:
+    types: [edited, opened, reopened, synchronize]
     paths:
       - '.all-contributorsrc'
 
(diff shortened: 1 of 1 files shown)


achieveordie:
Closing, read https://github.com/sktime/sktime/issues/4103#issuecomment-1426778333 for reference.

PR #4121, opened 2023-01-18
[ENH] Added piecewise_poisson

Pyyim:
Co-authored-by: Kevin <kevinlamdata@gmail.com>

#### Reference Issues/PRs

Reference #3363, #4078

#### What does this implement/fix? Explain your changes.

Added to datagen.py in preparation for adding to PoissonHMM

#### Does your contribution introduce a new dependency? If yes, which one?

No

#### Any other comments?
Reopened from PR request 4078 due to merge conflicts. Will add contributor file in a separate PR request in the future.
@miraep8 

##### For all contributions
- [x] I've added myself to the [list of contributors](https://github.com/sktime/sktime/blob/main/.all-contributorsrc).
- [ ] Optionally, I've updated sktime's [CODEOWNERS](https://github.com/sktime/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
- [x] I've added unit tests and made sure they pass locally.
- [X] The PR title starts with either [ENH], [MNT], [DOC], or [BUG] indicating whether the PR topic is related to enhancement, maintenance, documentation, or bug.

PR #4126, opened 2023-01-18
[ENH] Add PoissonHMM estimator

klam-data:
Co-authored-by: Paul [pauljr97@gmail.com]
Co-authored-by: Margaret [margaret.gorlin@gmail.com]
@Pyyim, @mgorlin

Reference Issues/PRs
Reference #3363, #4121

Built PoissonHMM estimator using piecewise_poisson function from previous PR #4121. 

Does your contribution introduce a new dependency? If yes, which one?
No

What should a reviewer concentrate their feedback on?
@miraep8

Follow-up on wrapping estimators from hmmlearn

Any other comments?
PR checklist

For all contributions
[X] I've added myself to the [list of contributors](https://github.com/sktime/sktime/blob/main/.all-contributorsrc).
[ ]Optionally, I've updated sktime's [CODEOWNERS](https://github.com/sktime/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
[X]I've added unit tests and made sure they pass locally.
[X]The PR title starts with either [ENH], [MNT], [DOC], or [BUG] indicating whether the PR topic is related to enhancement, maintenance, documentation, or bug.


miraep8:
Hey all, nice work with this its generally looking good.  I think that you all are actually running into a similar issue that @kejsitake ran into with her wrapper for `CategoricalHMM` in that the default `make_annotation_problem` function that we have for generating random annotation problems doesn't make sense with your estimator type (ie Poisson).  She suggested a good way to get around this that should work for you guys too - ie implement tags for the annotation estimators, and then use those tags to pass to the `make_annotation_problem` function to ensure that we make some dummy data that makes sense for your estimator.   Please ping me either here, or on slack if you want to chat about this more!

klam-data:
@miraep8, @Pyyim and I created a new tag under BaseSeriesAnnotator and modified the test_output_type under test_all_annotators.py to run certain test based on the tag. We ran the test and it looks good. This should also setup future contributors for other types of distributions.

sktime/annotation/tests/test_all_annotators.py
```diff
@@ -20,8 +21,16 @@ def test_output_type(Estimator):
...
-    assert isinstance(y_pred, (pd.Series, np.ndarray))
+    if estimator.get_tag("distribution_type") == "Poisson":
+        arg = piecewise_poisson(lambdas=[1, 2, 3], lengths=[2, 4, 8], random_state=42)
+        estimator.fit(arg)
+        arg = piecewise_poisson(lambdas=[1, 3, 6], lengths=[2, 4, 8], random_state=42)
+        y_pred = estimator.predict(arg)
+        assert isinstance(y_pred, (pd.Series, np.ndarray))
+
+    else:
+        arg = make_annotation_problem(n_timepoints=50)
```
miraep8, line 32: Excellent!  Thank you for implementing the tag system, I think this is looking much better! :) 

However I do have one request (not on the logic itself but rather where we put it. To me - it makes more sense to add the distribution dependent logic within `make_annotation_problem` rather than in this particular test.  The thought being - in the future we may add more tests for the annotator classes in general (this part of the library is still pretty new) and we would want those tests to be able to make use of the `make_annotation_problem` function. Rather than have to check in each test what distribution is needed, it makes more sense to me to pass this logic off to `make_annotation_problem`. 

It should be a simple matter of changing `make_annotation_problem` to (optionally) take in the estimator as well.  Then if an estimator is passed it can run some checks on the tags and tailor the problem to fit the needs of that particular estimator. 

Again, thank you for your efforts team, this is looking good! :)

miraep8 (changes requested):
Change request based on the location of the distribution dependent logic.  Good implementation of the tags though, and thank you all for getting on this so fast! :)

sktime/annotation/tests/test_all_annotators.py
klam-data: Thank you for the review!

We have made the changes to make_annotation_problem as well as make_series in order to accommodate for the change in annotator class. We have left an estimator_type argument in the make_annotation_problem for future instances.

miraep8: Hey!

Thank you guys for updating this in response to my suggestions - it is really looking much better!

I think we can push this just a bit farther and separate the new logic here a bit better from the existing code base.  To motivate what I am thinking/why I am asking for it a bit more:

- in sktime/object oriented programming in general we want to write code that respects "separation of concerns".  In this instance - I am a bit concerned that by putting all this logic about the distribution type in `make_series` we are actually adding some functionality and logic into that class when it doesn't strictly need to be there!  Again, this isn't breaking the code - but the added complexity can make maintenance and debugging a bit more difficult.
- on the other hand you guys just wrote a nice function for generating exactly the type of data that would work well for `PoissonHMM` (or any other estimator that needed Poisson like data!). So the logic we need already exists in sktime!  

Personally I kind of think of the functionality of `make_annotation_problem` to be acting as a delegator.  So far it has only delegated to `make_series`, but we can change it to instead optionally delegate the role of making an example problem to `piecewise_poisson` instead! 

Let me know what you think on the above!  Again, not intended as a huge change request, but I do think it would make things cleaner here, and also is a great opportunity to use that new function you guys wrote as well! :) 

- Mirae

sktime/utils/_testing/annotation.py
```diff
@@ -10,7 +16,17 @@ def make_annotation_problem(
...
+    estimator_type=None,
 ):
+    if estimator_type == "Poisson":
+        rng = check_random_state(random_state)
+        y = piecewise_poisson(
+            lambdas=rng.randint(1, 4, n_timepoints),
+            lengths=rng.randint(1, 5, n_timepoints),
+            random_state=random_state,
+        )
+        return y
```
miraep8, line 28: Its worth thinking about what might happen in a case where we want a poisson distributed `y` but also `X`.  What if we wanted a poisson distributed `X`?  I think it is ok that we do not cover all of these cases here, and I don't think that writing things this way should break anything yet (since we are typically not thinking about `X` at all for the `PoissonHMM` anyway.  I was just thinking it could be useful to leave a comment to this affect! :)

sktime/utils/_testing/annotation.py
```diff
@@ -1,5 +1,11 @@
 # -*- coding: utf-8 -*-
 
+__author__ = ["mloning", "fkiraly", "klam-data", "pyyim", "mgorlin"]
```
miraep8, line 3: hey all!  It might be worth checking with @fkiraly and others if you are going to add them as authors to random files.  In general in sktime we consider authorship to be somewhat of a responsibility rather than something just bestowed here and there.  (I know this because I did the same thing when I got started haha, and that is what @fkiraly told me!).

sktime/annotation/tests/test_hmm_learn.py
```diff
@@ -54,3 +54,26 @@ def test_GMMHMM_wrapper():
...
     sktime_predict = sktime_model.predict(X=data)
     assert array_equal(hmmlearn_predict, sktime_predict)
+
+
+@pytest.mark.skipif(
+    not _check_soft_dependencies("hmmlearn", severity="none"),
+    reason="skip test if required soft dependency for hmmlearn not available",
+)
+def test_PoissonHMM_wrapper():
+    """Verify that the wrapped GMMHMM estimator agrees with hmmlearn."""
```
miraep8, line 64: typo GMMHMM -> Poisson

miraep8 (commented):
Hey all!  Looking good! :)  Thank you for iterating with me and making these changes.  Will give @fkiraly some time to see how he feels about being added as an author throughout and then will check back about getting this merged.

sktime/annotation/tests/test_all_annotators.py
miraep8: Nice!  Looks much better now!  Thank you guys! :)

sktime/utils/_testing/annotation.py
klam-data: Thanks @miraep8 , should I just remove authorship in its entirety from here? It originally started without one.

miraep8: Hmmm, I guess you can still have the author field, but can leave it empty for now.

sktime/utils/_testing/annotation.py
klam-data: We gave the possibility of a poisson distributed `X` some thought and decided to not implement it in this case as we thought it was sufficient to only have a poisson distributed `y`. However, we do have the following code below for the request if needed. 

```
if estimator_type == "Poisson":
        rng = check_random_state(random_state)
        y = piecewise_poisson(
            lambdas=rng.randint(1, 4, n_timepoints),
            lengths=rng.randint(1, 5, n_timepoints),
            random_state=random_state,
        )
        if not make_X:
            return y
            
        X = piecewise_poisson(
            lambdas=rng.randint(1, 4, n_timepoints),
            lengths=rng.randint(1, 5, n_timepoints),
            random_state=random_state,
        )
        return y, X
```

sktime/utils/_testing/annotation.py
```diff
@@ -10,7 +16,17 @@ def make_annotation_problem(
     make_X=False,
     n_columns=2,
     random_state=None,
+    estimator_type=None,
 ):
+    if estimator_type == "Poisson":
+        rng = check_random_state(random_state)
```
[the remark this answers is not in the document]
miraep8: No you should leave `random_state`! My only point was that since you pass it into `piecewise_poisson` anyway (and run the check there) that there is no need to also run the check here too! :)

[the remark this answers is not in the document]
miraep8: Ah I see, nevermind!  I missed that you were using this to generate random `lambdas` and `lengths`!  Sorry, oversight on my part. Yes this is a fine way to do it.

sktime/utils/_testing/annotation.py
miraep8: I tend to agree with your hunch - ie that it probably isn't necessary and is just contributing the code bloat.  It also isn't clear to me that one would necessarily want to for X to also be poisson in this case that we want y to be poisson, so I wouldn't necessarily say we should default to the code you suggested.  Perhaps a better compromise would be to raise a ValueError if someone passes in both `make_X` and `estimator_type` = "Poisson".  That would give someone an informative error if they do run into this edge case int he future, and also serve as a potential warning to future devs that this edge case isn't covered.

sktime/annotation/tests/test_hmm_learn.py
klam-data: Thanks for the catch. Changed

sktime/utils/_testing/annotation.py
klam-data: Thanks @miraep8, 

We have added the following logic to the condition

```
    if estimator_type == "Poisson":
        if make_X:
            raise ValueError("PoissonHMM creates a distribution for y only")
```

miraep8: Nice!

sktime/utils/_testing/series.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3 -u
 # -*- coding: utf-8 -*-
 
-__author__ = ["mloning"]
+__author__ = ["mloning", "fkiraly"]
```
miraep8, line 4: Another place where I am not sure it makes sense to add @fkiraly!  I don't have an issue with you guys adding him, but again, would need to get his permission before we merge this if that is what you guys want to do! :)

miraep8 (approved):
Happy to merge pending the authorship questions in certain files.  It looks really good team! I imagine this took a little more time than anticipated - but nice work persevering - I especially appreciate you laying the groundwork with the estimator_distribution tags! :)

sktime/utils/_testing/series.py
klam-data: Reverted it. Thank you.

miraep8 (approved):

miraep8:
Just now has a merge conflict!

klam-data:
@miraep8. Should be good with merge now.

miraep8:
Hmm, I still see merge conflict on my end.  Otherwise I agree!

klam-data:
@miraep8 , think we're good on the merge now

miraep8 (approved):
Nice work!

Issue #4089, opened 2023-01-10
[ENH]Implement Bandpass Filters


JonathanBechtel:
**Is your feature request related to a problem? Please describe.**
Bandpass filters are common signal extraction techniques used in time series, in particular for the detection of business cycles and they do not have a current implementation in sktime (as far as I can tell)

**Describe the solution you'd like**
I think the use of the BaseTransformer class could be used to extend to the statsmodels.tsa library and/or the spicy implementation, since I believe wrappers are already in place for each of those libraries.

**Describe alternatives you've considered**
Right now the main alternatives I think are to use either the tsa_filters package from statsmodels, or other languages like in Matlab:  https://www.mathworks.com/help/signal/ref/bandpass.html

fkiraly:
I think there is already a multi-purpose spectral filtering class in `sktime`, I ported this a while ago from @SveaMeyer13's excellent summer project. It's the class `Filter` in `sktime.transformations.series.filter`, and interfaces the `filter.filter_data` in `mne`, which I believe is the correlate of Matlab `bandpass` in python.

Have a look whetehr that suits your requirements, and if not we can still add transformers.

fkiraly:
Update: I just looked at it, and it might be worth thinking about what should happen to a collection of unequally spaced or unequal length series, the class does not support that at current.

JonathanBechtel:
So looking into this a little bit further, what I was referring to was a set of specific transformers that are specific to econometric time series data, whereas the `mne.filter_data` part looks like it was transported over from `sktime-neuro`, which I think has slightly different use cases.  

I should have been more specific in my title, my apologies.  We're particularly interested in the baxter-king filter, which can be found in statsmodels here:  https://www.statsmodels.org/dev/generated/statsmodels.tsa.filters.bk_filter.bkfilter.html

The final step is to apply `fftconvolve` from scipy, but there are some additional transformations that come before it.  

I haven't looked over everything in `mne.filter_data`, but it looks like the `fftconvolve` step could be sent into the `mne.filter_data` function as an `fir` filter, but the preceding parts are not supported.  

I don't know if something like this would be in scope with sktime, or if it's preferable to use the existing `Filter` class or port over the tools from statsmodels, but myself and others be happy to try and work on this.

fkiraly:
ah, I see, @JonathanBechtel.
I think the math is pretty similar, on first glance the `statsmodels` BK filter just looks like a special case mathematically, with slightly different terminology, in a different community.

[...]

PR #4127, opened 2023-01-18
& pyyim [ENH] Add BKFilter Transformer

klam-data:
Co-authored-by @Pyyim

Reference Issues/PRs
Reference #4089

- Built BKFilter Transformer using statsmodels time series bandpass filters. 
- Added a separate test "test_bkfilter" to test transformer against statsmodels

Does your contribution introduce a new dependency? If yes, which one?
Yes - statsmodels 0.13.5

What should a reviewer concentrate their feedback on?
@fkiraly 

Any other comments?
PR checklist

For all contributions
[X] I've added myself to the [list of contributors](https://github.com/sktime/sktime/blob/main/.all-contributorsrc).
[ ]Optionally, I've updated sktime's [CODEOWNERS](https://github.com/sktime/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
[X]I've added unit tests and made sure they pass locally.
[X]The PR title starts with either [ENH], [MNT], [DOC], or [BUG] indicating whether the PR topic is related to enhancement, maintenance, documentation, or bug.


fkiraly (changes requested):
Great addition!
Some minor comments (but blocking, as CI fail).

Interface:
* you set the tag `transform-returns-same-time-index` to `True`, but it appears not to be the case. Kindly check whether the failure is expected (in which case the tag should be `False`), or whether it is a failure in the implementation.
* Kindly set the tag `python_dependencies`, as `statsmodels` will become a soft dependency in 0.15.1.
* the doctest is failing, docstring examples should run in the test environment. The module `sm` is not present. I think you are missing an `import statsmodels.api`.

Docstring:
* there is no description block, there should be one
* description block should contain a line which states it is a direct interface to `BKFilter` from `statsmodels`
* description block should contain a reference to scientific description of the filter. This could be a reference to `statsmodels` documentation, or to the original paper.

sktime/transformations/series/bkfilter.py
```diff
@@ -0,0 +1,170 @@
...
+        # this can be a Panel mtype even if transform-input is Series, vectorized
+        "y_inner_mtype": "None",  # which mtypes do _fit/_predict support for y?
+        "requires_y": False,  # does y need to be passed in fit?
+        "enforce_index_type": [
+            pd.RangeIndex
+        ],  # index type that needs to be enforced in X/y
+        "fit_is_empty": True,  # is fit empty and can be skipped? Yes = True
+        "transform-returns-same-time-index": False,
+        # does transform return have the same time index as input X
+        "capability:unequal_length": False,
```
fkiraly, line 101: I think this should be `True`, since it is applied per series (so it does not matter if they are different length)

klam-data: Thanks @fkiraly! Totally missed that tag when we did validate the input/output.

@Pyyim, @mgorlin

klam-data (commented):
Change tag for unequal length

fkiraly (dismissed):
Excellent, this seems to work now.

We can merge the estimator, although I would have some documentation requests. Up to you whether you would like to address these in this PR, or a separate one.

* a link should be added in the API reference, in `docs/source/api_reference/transformations.rst`, in the section "Filtering and denoising" I think.
* while you are at it, it seems that the `Filter` transformer was missing here, it should also be added.
* The math in the docstring is a bit unclear - it needs to be phrased in terms of the input. Some names/symbols appear without prior reference, e.g., `omega`. It is also not clear how weights with positive and negative index are used to compute a weighted moving average. I would recommend, add the missing information, or simply remove the math.

klam-data:
@fkiraly , 

We have made the changes to the documentation that you have requested. 

Please let us know if there is anything else.

Thank you.

@Pyyim

fkiraly (approved):
Nice, thanks!

Issue #4131, opened 2023-01-21
[BUG] `PolynomialTrendForecaster` stalls when fit on timeseries with datetime index


KishManani:
**Describe the bug**
My computer stalls when running `fit` on a dataframe with a `DatetimeIndex` with the `PolynomialTrendForecaster`.

**To Reproduce**
```python
from sktime.datasets import load_airline
from sktime.forecasting.trend import PolynomialTrendForecaster  

df = load_airline()
df.index = df.index.to_timestamp()

trafo = PolynomialTrendForecaster()
trafo.fit(df)
```

**Expected behavior**
The fitting time shouldn't be orders of magnitude different from using a `PeriodIndex`. 

**Versions**
<details>

System:
    python: 3.8.7 (default, May  6 2021, 21:53:45)  [Clang 12.0.0 (clang-1200.0.32.29)]
executable: /Users/kishan_manani/.pyenv/versions/3.8.7/envs/udemy-ts/bin/python3.8
   machine: macOS-11.6.6-x86_64-i386-64bit

Python dependencies:
          pip: 22.3.1
   setuptools: 49.2.1
      sklearn: 1.2.0
       sktime: 0.15.0
  statsmodels: 0.13.2
        numpy: 1.21.6
        scipy: 1.8.1
       pandas: 1.5.2
   matplotlib: 3.6.2
       joblib: 1.2.0
        numba: 0.56.4
     pmdarima: 1.8.5
      tsfresh: 0.19.0

</details>

fkiraly:
```
Unable to allocate 2.74 TiB for an array with shape (376012800001,) and data type int64
```
???

This is certainly a bug.

fkiraly:
Hope this fixes it: https://github.com/sktime/sktime/pull/4133

PR #4133, opened 2023-01-21
[BUG] fix erroneous `int` coercion of `TrendForecaster` and `PolynomialTrendForecaster` on `DatetimeIndex`

### LINKED ISSUE #3230: [BUG] `TrendForecaster` gives incorrect output
**Describe the bug**
The TrendForecaster produces unusual outputs on levels different from the target series.

**To Reproduce**

```python
from sktime.forecasting.trend import TrendForecaster

y = load_airline()
forecaster = TrendForecaster()

y_preds = []
for last_train in [20, 40, 60, 80]:
    forecaster.fit(y[:last_train])
    y_pred = forecaster.predict(fh=[1,2,3,4,5])
    y_preds.append(y_pred)

plot_series(y, *y_preds)
```

**Expected behavior**
The forecasts should be straight lines extrapolating the linear historical trend.

**Additional context**
Would expect that the default behaviour is similar to the default behaviour of the `PolynomialTrendForecaster`. In `TrendForecaster`, the `X` is calculated as
```python
X = y.index.astype("int").to_numpy().reshape(-1, 1)
```
whereas in the polynomial version, it is calculated as
```python
n_timepoints = _get_duration(self._y.index, coerce_to_int=True) + 1
X = np.arange(n_timepoints).reshape(-1, 1)
```

**Versions**
<details>

System:
    python: 3.8.13 (default, Apr 19 2022, 00:53:22)  [GCC 7.5.0]
executable: /usr/bin/python3.8
   machine: Linux-5.4.72-microsoft-standard-WSL2-x86_64-with-glibc2.27

Python dependencies:
          pip: 22.0.4
   setuptools: 56.0.0
      sklearn: 1.1.2
       sktime: 0.13.0
  statsmodels: 0.13.2
        numpy: 1.22.4
        scipy: 1.8.1
       pandas: 1.4.3
   matplotlib: 3.5.2
       joblib: 1.1.0
        numba: 0.56.0
     pmdarima: None
      tsfresh: None

</details>

fkiraly:
Fixes https://github.com/sktime/sktime/issues/4131 and a parallel issue for `TrendForecaster` which is the cause of #3230 (also fixes #3230).

Erroneous coercion caused huge arrays to be allocated when the input data had a `DatetimeIndex`.

Also, previously, any time stamp based fits and coefficients were related to a reset index (naive replacement with `RangeIndex`) rather than the time stamp.

This has been changed to periods since start of 1970 (with correct time stamp conversion). I've now added a clarifying docstring on what the coefficients mean, and changed it consistently to periods since start of 1970.

Since this was subject to a bug previously (and, likely, confusion), we don't need deprecation for this change.

Contains new and updated tests:
*  `test_trendforecaster_with_datetimeindex`, a test for the failure case in #4131, and parallel case.
* The existing test `test_trend` was failing with the bugfix due to the erroneous conversion to `int`. The test has been modified to cover the time stamp case corretly, by using time stamp converted indices in the Vandermonde matrix rather than the `RangeIndex` based indices.
* a test for the time stamp to integer/float conversion logic now used under the hood

### DIFF
--- a/sktime/forecasting/tests/test_trend.py
+++ b/sktime/forecasting/tests/test_trend.py
@@ -2,7 +2,7 @@
 """Test trend forecasters."""
 # copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
 
-__author__ = ["mloning"]
+__author__ = ["mloning", "fkiraly"]
 __all__ = ["get_expected_polynomial_coefs"]
 
 import numpy as np
@@ -11,9 +11,9 @@ import pytest
 
 from sktime.datasets import load_airline
 from sktime.forecasting.trend import (
-    _get_X_numpy_int_from_pandas,
     PolynomialTrendForecaster,
     TrendForecaster,
+    _get_X_numpy_int_from_pandas,
 )
 from sktime.utils._testing.forecasting import make_forecasting_problem
 
--- a/sktime/forecasting/tests/test_trend.py
+++ b/sktime/forecasting/tests/test_trend.py
@@ -16,7 +16,10 @@ from sktime.utils._testing.forecasting import make_forecasting_problem
 
 def get_expected_polynomial_coefs(y, degree, with_intercept=True):
     """Compute expected coefficients from polynomial regression."""
-    poly_matrix = np.vander(np.arange(len(y)), degree + 1)
+    from sktime.forecasting.trend import _get_X_numpy_int_from_pandas
+
+    t_ix = _get_X_numpy_int_from_pandas(y.index).reshape(-1)
+    poly_matrix = np.vander(t_ix, degree + 1)
     if not with_intercept:
         poly_matrix = poly_matrix[:, :-1]
     return np.linalg.lstsq(poly_matrix, y.to_numpy(), rcond=None)[0]
--- a/sktime/forecasting/tests/test_trend.py
+++ b/sktime/forecasting/tests/test_trend.py
@@ -42,6 +42,7 @@ def test_get_X_numpy():
     assert np.isin(intdiffs, [30, 31, 28, 29]).all()
 
     # testing pd.DatetimeIndex with hourly frequency
+    # diffs should be 1/24, since this is converted to float, days since 1970
     df_hourly = pd.DataFrame(
         data=[10, 5, 4, 2, 10],
         index=pd.date_range(start="2000-01-01", periods=5, freq="H"),
--- a/sktime/forecasting/tests/test_trend.py
+++ b/sktime/forecasting/tests/test_trend.py
@@ -10,14 +10,50 @@ import pandas as pd
 import pytest
 
 from sktime.datasets import load_airline
-from sktime.forecasting.trend import PolynomialTrendForecaster, TrendForecaster
+from sktime.forecasting.trend import (
+    _get_X_numpy_int_from_pandas,
+    PolynomialTrendForecaster,
+    TrendForecaster,
+)
 from sktime.utils._testing.forecasting import make_forecasting_problem
+
+
+def test_get_X_numpy():
+    """Test _get_X_numpy_int_from_pandas converts to int/float as expected."""
+    y = load_airline()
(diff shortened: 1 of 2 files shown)


KishManani:
Hi @fkiraly! I can confirm this fixes the bug on my end!

fkiraly:
@KishManani, this seems to fail tests against exact fitted/forecast values.

The question is now, were the expected incorrect in the tests that exist (e.g., just copied over from an output), or is the fix making them incorrect?

Would be thankful for independent verification.

fkiraly:
The tests are failing since the expected coefficients were in a different unit, i.e., not periods since start of 1970 (with correct time stamp conversion), but after naive conversion to a `RangeIndex`.

I've now added a clarifying docstring on what the coefficients mean, and changed it consistently to periods since start of 1970.

Since this was subject to a bug previously (and, likely, confusion), we don't need deprecation for this change.

KishManani:
https://github.com/sktime/sktime/pull/4133#issuecomment-1399427723

Hi @fkiraly! Do you still need a review on this? I think it would be great to sneak it in in time for the next release.

KishManani:
> @KishManani, this seems to fail tests against previously hard coded, exact fitted/forecast values.
> 
> The question is now, were the expected incorrect in the tests that exist (e.g., just copied over from an output), or is the fix making them incorrect?
> 
> Would be thankful for independent verification.

Hi @fkiraly! Do you still need a review on this? I think it would be great to sneak it in in time for the next release.

fkiraly:
> Hi @fkiraly! Do you still need a review on this?

Yes, that would be much appreciated 😄

sktime/forecasting/tests/test_trend.py
```diff
@@ -2,20 +2,60 @@
...
 from sktime.utils._testing.forecasting import make_forecasting_problem
 
 
+def test_get_X_numpy():
+    """Test _get_X_numpy_int_from_pandas converts to int/float as expected."""
+    y = load_airline()
+    X_idx = _get_X_numpy_int_from_pandas(y.index)
+
+    # testing pd.PeriodIndex
+    # this should be a 2D nd.nparray, with diffs being 1 (month)
```
KishManani, line 27: Nit: there is a typo here `nd.nparray`.

sktime/forecasting/tests/test_trend.py
```diff
@@ -2,20 +2,60 @@
...
+
+    y_idx_datetime = y.index.to_timestamp()
+    X_idx_datetime = _get_X_numpy_int_from_pandas(y_idx_datetime)
+
+    # testing pd.DatetimeIndex
+    # this should be a 2D np.ndarray, with diffs being 28, 29, 30, or 31 (days)
+    # because DatetimeIndex time stamps convert to days
+    assert isinstance(X_idx_datetime, np.ndarray)
+    assert X_idx_datetime.shape == (len(y), 1)
+    intdiffs = (np.diff(X_idx_datetime.reshape(-1))).astype(int)
```
KishManani, line 41: I think it would be worth adding a test for sub-daily data (e.g., hourly) when working with `DateTimeIndex`.

Here is an example of trying to taking the `np.diff` on the `DateTimeIndex` with hourly data:
```Python
>>> df=pd.DataFrame(data=[10,5,4,2,10], index=pd.date_range(start="2000-01-01", periods=5, freq="H"))
>>> np.diff(df.index)
array([3600000000000, 3600000000000, 3600000000000, 3600000000000],
      dtype='timedelta64[ns]')
```
Now we just want to check they're all the same.

KishManani (changes requested):

sktime/forecasting/tests/test_trend.py
fkiraly: fixed

sktime/forecasting/tests/test_trend.py
fkiraly: added

KishManani (dismissed):

KishManani:
LGTM!

Issue #4139, opened 2023-01-23
[ENH] Improve performance of pandas multi-index data type operations with many groups


hoesler:
**Is your feature request related to a problem? Please describe.**
The current implementation is showing poor performance with panel or hierarchical datasets in the pandas multi-index mtype format if the number of instances is substantial.
I was observing this with a retail dataset: daily sales data for thousands of articles across hundreds of stores and a number of exogenous features.
I am trying to create a forecasting pipeline for each article, which includes different series transformations (e.g. ColumnSelect, Lag, Impute, ...) and a regression model. So any performance issue is amplified by the number of models and probably only striking because of that.

The root cause, I found, is the iterative indexing approach of sktime to apply functions to each instance (series) in the data frame, even in cases where a more performant solution exists. Pandas grouped operations are very slow if you iterate over the group indices (`df.loc(idx) for idx in groups`). Groupby/apply with a python function is a first good step to improve things, but the most performant way is utilizing grouped operations which are delegated to compiled code.

As an example, consider imputing with a mean function. Currently, the Imputer is basically doing this:

```python
means = [mean(panel.loc[idx]) for idx in get_indices(panel, 'store')]
[impute(panel.loc[idx], mean) for idx, mean in zip(panel, means)]
```

A first step would be to rewrite it as

```python
means = [mean(series) for series in panel.groupby('store')]
[series.fillna(mean) for series, mean in zip(panel.groupby('store'), means)]
```

An even more performant implementation would not use any iteration at all, but apply an optimized pandas operation on the whole multi-index frame:

```python
means = panel.groupby('store').mean()
panel.groupby('store').fillna(means)
```

To get a feeling for the impact, I timed `Imputer(method="mean").fit_transform(X[['price']])` for a dataset with 1000 stores and each series consisting around 2000 observations, the difference I measured was **60.7s** vs **2.4s**. Remember that in my case this will be multiplied by the number of transformations in the pipeline, stores, models, hyperparameter optimization runs, etc.

**Describe the solution you'd like**
Any series wise operation on multi-index frames should use optimized pandas methods, if possible.

Core operations, like `VectorizedDF.as_list` or `check_pdmultiindex_panel` should be refactored (partially covered by [[#3827](https://github.com/sktime/sktime/issues/3827)](https://github.com/sktime/sktime/issues/3827)). Unnecessary calls should be avoided.

[...]

PR #4140, opened 2023-01-23
[ENH] Improve multiindex performance

hoesler:
#### Reference Issues/PRs

#4139

This PR improves the performance of some multi-index operations.

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/datasets/tests/test_data_io.py
+++ b/sktime/datasets/tests/test_data_io.py
@@ -6,7 +6,6 @@ __author__ = ["SebasKoel", "Emiliathewolf", "TonyBagnall", "jasonlines", "achiev
 __all__ = []
 
 import os
-import shutil
 import tempfile
 
 import numpy as np
@@ -44,24 +43,23 @@ _TO_DISABLE = ["pd-long", "pd-wide", "numpyflat"]
 
 @pytest.mark.parametrize("dataset_name", ["UnitTest", "BasicMotions"])
 @pytest.mark.parametrize("return_type", ["nested_univ", "numpy3d"])
-def test_write_panel_to_tsfile_equal_length(dataset_name, return_type):
+def test_write_panel_to_tsfile_equal_length(dataset_name, return_type, tmpdir):
     """Test function to write a dataset.
 
     Loads equal and unequal length problems into both data frames and numpy arrays,
     writes locally, reloads, then compares all class labels. It then delete the files.
     """
     X, y = _load_provided_dataset(dataset_name, split="TRAIN", return_type=return_type)
-    write_panel_to_tsfile(data=X, path="./Temp", target=y, problem_name=dataset_name)
-    load_path = f"./Temp/{dataset_name}/{dataset_name}.ts"
+    write_panel_to_tsfile(data=X, path=tmpdir, target=y, problem_name=dataset_name)
+    load_path = tmpdir / dataset_name / f"{dataset_name}.ts"
     newX, newy = load_from_tsfile(
         full_file_path_and_name=load_path, return_data_type=return_type
     )
     assert np.array_equal(y, newy)
-    shutil.rmtree("./Temp")
 
 
 @pytest.mark.parametrize("dataset_name", ["PLAID", "JapaneseVowels"])
-def test_write_panel_to_tsfile_unequal_length(dataset_name):
+def test_write_panel_to_tsfile_unequal_length(dataset_name, tmpdir):
     """Test function to write a dataset.
 
     Loads equal and unequal length problems into both data frames and numpy arrays,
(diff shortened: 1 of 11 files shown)


danbartl:
Great initiative, and highly welcome :). 

Your approach reduces times from 200 secs to 90 secs in my code. It is still crazy even in this version how much we waste on checks:

image

In my view, we should aim at having checks take only a fraction of the time used for model training and fitting, not the other way round. 

Btw I implemented a simple cross check for are_columns_nested check to jump back to top level  in https://github.com/sktime/sktime/pull/4130, but your rework is much more useful.

fkiraly:
Your code quality checks are failing - see here for a guide on code formatting for sktime:
https://www.sktime.org/en/stable/developer_guide/coding_standards.html

hoesler:
Black errors should be fixed now. I forgot to install the pre-commit hooks in the beginning. BTW: `pre-commit run -a` throws a lot of errors not related to my changes.

fkiraly:
> BTW: `pre-commit run -a` throws a lot of errors not related to my changes.

Does it? Not again...

Would appreciate a bug issue with a summary of the problems (possibly printouts or tracebacks).

hoesler:
Sure. See #4182

fkiraly:
Regarding our discussion on PR scope:

there's a lot of good improvements in there!
Now, they happen in many places, and some are causing knock-on errors (e.g., the notebook failure).

It would be easier to review - and to move "unproblematic" (as in: bug-free) parts incrementally into the codebase - if you were to split this along the lines of modularity and concerns.
E.g., improvements to estimators could go separate, even by-estimator. Improvements to checks could be separated, and so on.

General best practice is to prefer multiple, modular PR above one maxi-PR that collects all your work.
This is not always possible of course, but it's a good guideline to have in mind.

hoesler:
> This is not always possible of course, but it's a good guideline to have in mind.

Sure, i know. Reason for these bulk changes is, that making sktime usable for my use case was my primary goal. The PR came second. I'll try to split the PR.

fkiraly:
> I'll try to split the PR.

Unless of course you see easily how to fix the failure in the notebook...

hoesler:
> Unless of course you see easily how to fix the failure in the notebook...

Yes, I do. So I'll take my chance ;)

fkiraly:
image
👍

sktime/datasets/tests/test_data_io.py
```diff
@@ -44,24 +43,23 @@
 
 @pytest.mark.parametrize("dataset_name", ["UnitTest", "BasicMotions"])
 @pytest.mark.parametrize("return_type", ["nested_univ", "numpy3d"])
-def test_write_panel_to_tsfile_equal_length(dataset_name, return_type):
+def test_write_panel_to_tsfile_equal_length(dataset_name, return_type, tmpdir):
```
hoesler, line 46: @fkiraly The changes in this file are unrelated to my PR, but I ran into problems with a race condition testing my changes.

fkiraly: this is sth I would separate out, actually. Should be a few clicks only.

Would be appreciated if you could describe the race condition somewhere, e.g., in an issue?

fkiraly: you can still have it in this PR, and at the top you say, "depends on PR such-and-such". I.e., if the other PR is merged more quickly, the diff in this one gets smaller.

fkiraly:
> I'll try to split the PR.

I think this still could make sense given that some changes are more "in progress", while others seem more obvious and ready to go.

sktime/datatypes/_check.py
```diff
@@ -237,18 +237,20 @@ def check_raise(obj, mtype: str, scitype: str = None, var_name: str = "input"):
...
+        return_metadata=False,
         var_name=var_name,
     )
 
     if valid:
         return True
     else:
-        raise TypeError(msg)
+        raise TypeError(
+            f"Expected object mtype to be one of {list(mtype)}, was {type(obj)}"
```
fkiraly, line 252: this message would be confusing, as some mtypes are called like class types, e.g., `pd.DataFrame`.

sktime/datatypes/_check.py
```diff
@@ -302,16 +304,16 @@ def mtype(
...
             mtype=m_plus_scitype[0],
             scitype=m_plus_scitype[1],
-            return_metadata=True,
+            return_metadata=False,
         )
         if valid:
             mtypes_positive += [m_plus_scitype[0]]
         else:
-            mtypes_negative[m_plus_scitype[0]] = msg
+            mtypes_negative[m_plus_scitype[0]] = ""
```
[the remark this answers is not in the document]
fkiraly: I see what you're doing - very smart observation imo.

The "clean" way would be a refactor with deprecation.
One way to do it would be to introduce a `return_msg` argument everywhere, which can be True, False, or None (default).
If None, the current behaviour happens and a deprecation warning is raised. If `True`, returns the `msg` return, independently of `return_metadata`. If `False`, never returns the `msg` arg.
After a deprecation cycle, the default changes to `False`.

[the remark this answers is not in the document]
fkiraly: it is nightmarish though since it would have to be one in *all the checker functions*.

I would definitely advise a separate PR for this with only that as the purpose, since PRs should be focused in file footprint, or in scope. It is not possible to focus this in terms of file footprint...

unless you come up with a better idea.

fkiraly:
I would definitely recommend to separate the PR with the estimators, one per estimator.
Better from a modularity perspective.

sktime/datatypes/_check.py
hoesler: Moved discussion to #4191

sktime/datatypes/_check.py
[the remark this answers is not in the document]
hoesler: Moved discussion to #4191

sktime/datasets/tests/test_data_io.py
hoesler: Moved to #4192

hoesler:
Now, that the PR is split up, I guess we can close it.

PR #4143, opened 2023-01-24
[ENH] Add `SimpleRNN` DL Estimator

ArushikaBansal:
In order to implement diffrent estimators in sktime I made the RNN Network which will now be used to make RNN estimator and RNN classifier.

This is a continuation to the issue #3351


achieveordie:
Great job on this PR.

It seems like the dev setup is incomplete - did you install [pre-commit](https://www.sktime.org/en/stable/developer_guide/coding_standards.html#id8) properly? After you set it up, make small changes (I recommend removing extra newlines that are present at the end of some files) and then commit again, pre-commit will automatically work and make changes to adhere to the coding standards that `sktime` follows. In the case it cannot make changes automatically, it will tell you what is causing the failure.

The other thing that I noticed is importing `keras` and `tensorflow` at the top level of the file. These are soft dependencies and are to be imported within the scope of their usage (so systems that don't have these dependencies and do not use them do not fail from an `ImportError`). You can read more about it in [this](https://www.sktime.org/en/stable/developer_guide/dependencies.html#adding-a-soft-dependency) section of the developer guide. Even though it pertains to adding new soft dependencies, it is also applicable to already existing ones.

Hope this helps :)

achieveordie:
Closing since the updated PR is #4185.

PR #4146, opened 2023-01-25
[ENH] transformer interfacing `numpy.fft` for simple fourier transform

blazingbhavneek:
#### No new dependency
Do you think this is enough? Or am I missing something? Also, suggest for docstrings. @fkiraly

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -2064,6 +2064,25 @@
       ]
     },
     {
+      "login": "pranavvp16",
+      "name": "Pranav Prajapati",
+      "avatar_url": "https://avatars.githubusercontent.com/u/94780581?v=4",
+      "profile": "https://www.linkedin.com/in/pranav-prajapati-a5b413226/",
+      "contributions": [
+        "code",
+        "test"
+      ]
+    },
+    {
+      "login": "romanlutz",
+      "name": "Roman Lutz",
+      "avatar_url": "https://avatars.githubusercontent.com/u/10245648?v=4",
+      "profile": "https://www.linkedin.com/in/romanlutz/",
+      "contributions":[
+        "doc"
+      ]
+    },
+    {
       "login": "DBCerigo",
       "name": "Daniel Burkhardt Cerigo",
       "avatar_url": "https://avatars.githubusercontent.com/u/8318425?v=4",


fkiraly:
Nice, thanks!

Small thing: the PR is failing code formatting, here's the guide on how to fix it:
 I was asking about the other mentorship you mentioned

blazingbhavneek:
I think I solved the issue @fkiraly , (there was a problem with my pre-commit) kindly run the checks again

sktime/transformations/series/simple_fourier.py
```diff
@@ -0,0 +1,82 @@
...
+        X_sort = X.sort_index()  # In-case the data is unsorted
+
+        ser_size = X.size
+        sig = X_sort.to_numpy()
+
+        # numpy.fft methods
+        dft_seq = np.abs(fft(sig))
+        sample_freq = fftfreq(ser_size)
+
+        # Removing negative conjugate of frequncies
```
KishManani, line 75: Would using [`rfft`](https://numpy.org/doc/stable/reference/generated/numpy.fft.rfft.html#numpy.fft.rfft) instead of `fft` make this section of the code redundant and also provide a faster solution?

sktime/transformations/series/simple_fourier.py
```diff
@@ -0,0 +1,82 @@
...
+        Parameters
+        ----------
+        X : Series mtype X_inner_mtype
+
+        Returns
+        -------
+        transformed version of X
+        """
+
+        X_sort = X.sort_index()  # In-case the data is unsorted
```
KishManani, line 66: I believe `sktime` already checks that the `X` is pre-sorted and would throw an error if it isn't. So we can remove this line.

KishManani:
@blazingbhavneek is it possible to also add a unit test? You can read more about tests [here](https://www.sktime.org/en/stable/developer_guide/testing_framework.html). Another good place to look for examples is the [existing tests](https://github.com/sktime/sktime/tree/main/sktime/transformations/series/tests) for other transformers.

fkiraly:
there's still one linting issue (missing newline), perhaps a forgotten commit/push?

sktime/transformations/series/simple_fourier.py
fkiraly: indeed - also, instead of manually converting to `numpy`, you can simply use `X_inner_mtype="np.ndarray"`, this ensures that the `X` you see inside `_fit`, `_transform` is already in `numpy` format.

sktime/transformations/series/simple_fourier.py
blazingbhavneek: Yes! According to [this](https://www.researchgate.net/figure/5-Speed-comparison-of-FFT-method-and-RFFT-method_tbl7_49909866) article its efficiency increases with dataset size, I will make changes right away!

sktime/transformations/series/simple_fourier.py
```diff
@@ -0,0 +1,82 @@
...
+__author__ = ["blazingbhavneek"]
+
+
+import numpy as np
+import pandas as pd
+from numpy.fft import fft, fftfreq
+from sktime.transformations.base import BaseTransformer
+
+
+class simple_fourier(BaseTransformer):
```
achieveordie, line 15: Minor grievance: name it `SimpleFourier` to stick to the class naming convention (although the linter might automatically fix it).

achieveordie (changes requested):

sktime/transformations/series/simple_fourier.py
KishManani: @blazingbhavneek, adding to that, out of curiosity why is this called "simple fourier" rather than just "fourier transform"?

sktime/transformations/series/simple_fourier.py
blazingbhavneek: Done!

sktime/transformations/series/simple_fourier.py
blazingbhavneek: Done! @achieveordie 
@KishManani It's a simple Fourier transform because it just interfaces numpy.fft, it doesn't take anything else as an input (and also, it assumes data is sampled uniformly, I raised this issue, but @fkiraly told me to do the simple interfacing task for now and do the other stuff later)

fkiraly:
there's something wrong with your precommit.
Do you want to join a screenshare to get it fixed? I have some time tomorrow, or in a few hours (discord).

sktime/transformations/series/simple_fourier.py
fkiraly: 🙌

fkiraly:
I think it's still failing with the linting. Happy to do a screenshare on precommits and local linting sometime, e.g., Friday in the community session.

sktime/transformations/series/simple_fourier.py
KishManani: Thanks @blazingbhavneek! Are there other Fourier transform implementations that you think could be added, other than an implementation which can handle non-uniformly sampled data? If this implementation is going to handle the majority of use cases, would it be worth just naming it as `FourierTransform`? I'm just mildly concerned if the namespace will have a lot of Fourier Transform implementations whether it's easy for a user to find what they're looking for (e.g., we already have `FourierFeatures`, this would add `SimpleFourier`, in the future we might have `NonUniformFourier` etc.). @fkiraly do you have any opinions on this? I know `sktime` now accepts multiple implementations of the same algorithm, is there any guidance or thought about the tedious topic of naming these 😅?

fkiraly: hm, my gut feeling says: I would only have one file, e.g., `fourier_transform`, for any variations on the topic of fourier transform, and keep adding classes - or options to existing classes - unless the scope grows too large.

Fourier *features* are related but conceptually different in how they are applied, so one could argue either way sensibly, for separate or same module.

fkiraly: >is there any guidance or thought about the tedious topic of naming these

I'm not too good at naming. Names should be descriptive and make clear to the user, optimally without inspection of the docstring, what a function, variable, class, does. The risk is that names get too long.

E.g., `SimpleFourier` is less clear than `FourierTransform`, because what does "simple" mean, and a user might ask, well, Fourier *what*?

fkiraly: regarding applicability, I think we should try to extend the scope of the fourier transform transformer to irregular data etc in the same class.

fkiraly (changes requested):
@blazingbhavneek, could you kindly make a separate pull request for the seasonal estimator?
Instead of putting it in the same PR as the fourier transform.

achieveordie (changes requested):
@blazingbhavneek Excellent work so far!

Regarding the current check failure, you'd also need to add the `seasonal` package to [`pyproject.toml`](https://github.com/sktime/sktime/blob/main/pyproject.toml). [This](https://stackoverflow.com/questions/62983756/what-is-pyproject-toml-file-for) SE question offers insight if you want to learn more.

If `seasonal` is being added as a soft dependency, you must also import the package within the scope of its usage, rather than a file-level import. I've discussed its rationale in the second paragraph [here](https://github.com/sktime/sktime/pull/4143#issuecomment-1405200728).

blazingbhavneek:
Hey! @fkiraly I am sorry for the delay, I had exams in college, but I noticed I ruined my main branch of the forked repository, so I will be making the PR again for this too. I am sorry for the inconvenience.

PR #4155, opened 2023-01-27
[ENH] consistent internal removal of `pd.Series` name

### LINKED ISSUE #4144: [BUG] `Detrender` returns incorrect result when using `inverse_transform`
**Describe the bug**
The `inverse_transform` is not returned correctly when passing a `Series` type which has a different `name` from the `DataFrame` or `Series` used during `fit`.  Instead two columns which are null are returned. Another condition to trigger this bug appears to be when the user supplies data with a different time index during `fit` and `inverse_transform`. 

**To Reproduce**

```python
from sktime.transformations.series.detrend import Detrender
df = load_airline().rename("y").to_frame()

detrender = Detrender()
# Fit on some time period
detrender.fit(df.loc[:"1949-03", "y"])
# De-trend on a different time period
y_detrended = detrender.transform(df.loc["1949-04":, "y"])
# Inverse transform (i.e., add the trend back)
detrender.inverse_transform(y_detrended)
```

image

This behaviour only incurs when the input is a `Series`. If I were to wrap `"y"` above in a list, `["y"]`, so that a `DataFrame` is instead passed to `fit` and `transform`, then we do not get this bug.

**Expected behavior**
A single column with the trend added back resulting in the same data as the original dataframe.

**Versions**
<details>

System:
executable: /Users/kishan_manani/.pyenv/versions/3.8.7/envs/udemy-ts/bin/python3.8

Python dependencies:
          pip: 22.3.1
   setuptools: 49.2.1
      sklearn: 1.2.0
       sktime: 0.15.1
  statsmodels: 0.13.2
        numpy: 1.21.6
        scipy: 1.8.1
       pandas: 1.5.2
   matplotlib: 3.6.2
       joblib: 1.2.0
        numba: 0.56.4
     pmdarima: 1.8.5
      tsfresh: 0.19.0

</details>

fkiraly:
This is an alternative fix, fixes #4144.

The fixing strategy is to always remove the `name` attribute of `pd.Series` when entering the `fit` conversion, and adding it back, at back-conversion.

Experimental, to see what happens.

### DIFF
--- a/sktime/datatypes/_series/_convert.py
+++ b/sktime/datatypes/_series/_convert.py
@@ -54,6 +54,24 @@ for tp in MTYPE_LIST_SERIES:
     convert_dict[(tp, tp, "Series")] = convert_identity
 
 
+def handle_UvS_as_Series_nameattr(obj: pd.Series, store=None) -> pd.DataFrame:
+
+    if not isinstance(obj, pd.Series):
+        raise TypeError("input must be a pd.Series")
+
+    res = obj.copy(deep=False)
+
+    if isinstance(store, dict) and "name" in store.keys():
+        res.name = store["name"]
+    if isinstance(store, dict) and "name" not in store.keys():
+        store["name"] = obj.name
+
+    return res
+
+
+convert_dict[("pd.Series", "pd.Series", "Series")] = handle_UvS_as_Series_nameattr
+
+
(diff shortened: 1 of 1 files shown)


fkiraly:
closed in favour of https://github.com/sktime/sktime/pull/4150 which seems to be the better option

PR #4160, opened 2023-01-27
[DOC] improved reducer docstring formatting

### LINKED ISSUE #4158: [DOC] make_reduction API reference page doesn't render properly

The following page renders a bit strangely:
https://www.sktime.org/en/stable/api_reference/auto_generated/sktime.forecasting.compose.make_reduction.html?highlight=make_reduction

image

A few newlines could help fix this perhaps. Is that `|--|**|--|` notation meant to render as something special?

fkiraly:
Fixes #4158.

The solution is via indentation, in which case the symbol table is rendered as a code block incl newlines.

### DIFF
--- a/sktime/forecasting/compose/_reduce.py
+++ b/sktime/forecasting/compose/_reduce.py
@@ -1359,53 +1359,54 @@ def make_reduction(
         ``*`` = observations, past or future, neither part of window nor forecast.
 
         Assume we have the following training data (14 observations):
-        |----------------------------|
-        | * * * * * * * * * * * * * *|
-        |----------------------------|
+            |----------------------------|
+            | * * * * * * * * * * * * * *|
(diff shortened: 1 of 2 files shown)


fkiraly:
Almost working, it doesn't put newlines at the end though:

image

any ideas?

romanlutz (approved):
Lgtm.
Is there a place where sktime compatible data formats are explained? I'm just wondering how I'd know that series can't have multiindex, for example.

fkiraly:
well, this is not fixed yet...

> Is there a place where sktime compatible data formats are explained? I'm just wondering how I'd know that series can't have multiindex, for example.

Yes, it's in the notebook called `examples/AA_datatypes_and_datasets`, the better docstrings tend to link there

Issue #4162, opened 2023-01-28
[BUG] CI fails in `code_quality` `isort` hook, environment creation, at preparing metadata via `pyproject.toml`


fkiraly:
Since a few minutes ago, `sktime` CI fails at the first step when trying to install the env, with this traceback below.

This seems to be coming from the `code_quality` job, in there the `isort` hook when the python environment is being created, probably when trying to parse a `pyproject.toml`.

As a temporary workaround, `isort` has been removed here: https://github.com/sktime/sktime/pull/4164
but it should be added back.

Below is the traceback:

```
      Preparing metadata (pyproject.toml): started
      Preparing metadata (pyproject.toml): finished with status 'error'
    
stderr:
      error: subprocess-exited-with-error
      
      × Preparing metadata (pyproject.toml) did not run successfully.
      │ exit code: 1
      ╰─> [14 lines of output]
          Traceback (most recent call last):
            File "/home/runner/.cache/pre-commit/repohskwxkw8/py_env-python3/lib/python3.10/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 351, in <module>
              main()
            File "/home/runner/.cache/pre-commit/repohskwxkw8/py_env-python3/lib/python3.10/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 333, in main
              json_out['return_val'] = hook(**hook_input['kwargs'])
            File "/home/runner/.cache/pre-commit/repohskwxkw8/py_env-python3/lib/python3.10/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 152, in prepare_metadata_for_build_wheel
              return hook(metadata_directory, config_settings)
            File "/tmp/pip-build-env-3kaotqva/overlay/lib/python3.10/site-packages/poetry/core/masonry/api.py", line 40, in prepare_metadata_for_build_wheel
              poetry = Factory().create_poetry(Path(".").resolve(), with_groups=False)
            File "/tmp/pip-build-env-3kaotqva/overlay/lib/python3.10/site-packages/poetry/core/factory.py", line 57, in create_poetry
              raise RuntimeError("The Poetry configuration is invalid:\n" + message)
          RuntimeError: The Poetry configuration is invalid:
            - [extras.pipfile_deprecated_finder.2] 'pip-shims<=0.3.4' does not match '^[a-zA-Z-_.0-9]+$'
          
          [end of output]
      
      note: This error originates from a subprocess, and is likely not a problem with pip.
[...]

PR #4164, opened 2023-01-28
[MNT] emergency fix for precommit CI failure - remove `isort`

fkiraly:
Temporary workaround for #4162 to allow other development and release.

Removes the `isort` pre-commit step to which the problem can be isolated.


ir3456:
We were having the same issue but it appears that their 5.12.0 release has resolved the issue at least in out case. Simply updating the `rev: 5.12.0` in our `.pre-commit-config.yaml` fixed the issue.

Issue #4067, opened 2023-01-05
[BUG] Initializing any DL estimator with non-default values constructs these values only superficially


achieveordie:
**Describe the bug**

A logical bug is present in all DL estimators and is caused by not passing custom parameters to its underlying `Network` since `Network` actually makes use of these values to create the DL model of appropriate size. Take a look at the example provided below.

**To Reproduce**

```python
from pprint import pprint
from classification.deep_learning import CNNClassifier

model = CNNClassifier(
    kernel_size=3,
    n_conv_layers=1,
)

# model.summary() fails due to self.history being None, another bug

pprint(vars(model))
print("=="*25)
pprint(vars(model._network))
```

Output:
``` python
{'_X_metadata': [],
 '_class_dictionary': {},
 '_estimator_type': 'classifier',
 '_is_fitted': False,
 '_network': CNNNetwork(),
 '_tags_dynamic': {},
 '_threads_to_use': 1,
 'activation': 'sigmoid',
 'avg_pool_size': 3,
 'batch_size': 16,
 'callbacks': None,
 'classes_': [],
 'fit_time_': 0,
 'history': None,
 'kernel_size': 3,
 'loss': 'mean_squared_error',
 'metrics': None,
 'model_': None,
 'n_classes_': 0,
 'n_conv_layers': 1,
 'n_epochs': 2000,
 'optimizer': None,
 'random_state': None,
 'use_bias': True,
 'verbose': False}
==================================================
{'_tags_dynamic': {},
 'activation': 'sigmoid',
 'avg_pool_size': 3,
 'filter_sizes': [6, 12],
 'kernel_size': 7,
 'n_conv_layers': 2,
 'random_state': 0}
```

**Expected behavior**

This is happening because of `self._network = CNNNetwork()` (or corresponding Network) in the constructor of the estimator. The values that we receive are to be passed into the Network.

**Additional context**

I'm speculating this is the reason why the checks at #4036 are failing despite having very small hyperparameters. In any case, I'll open a PR to resolve this bug.

**Versions**
<details>

```
D:\Anaconda\envs\sktime-dev\lib\site-packages\_distutils_hack\__init__.py:33: UserWarning: Setuptools is replacing distutils.
  warnings.warn("Setuptools is replacing distutils.")

System:
    python: 3.8.13 | packaged by conda-forge | (default, Mar 25 2022, 05:59:00) [MSC v.1929 64 bit (AMD64)]
[...]

PR #4185, opened 2023-02-01
[ENH] Add SimpleRNN DL Estimator

ArushikaBansal:
migrates simple RNN regressor from sktime-dl, see https://github.com/sktime/sktime/issues/3351

### DIFF
--- a/sktime/classification/deep_learning/rnn.py
+++ b/sktime/classification/deep_learning/rnn.py
@@ -6,9 +6,7 @@
 __author__ = ["Markus Löning"]
 __all__ = ["SimpleRNNRegressor"]
 
-import tensorflow as tf
 from sklearn.utils import check_random_state
-from tensorflow import keras
 
 from sktime.classification.deep_learning.base import BaseDeepClassifier
 from sktime.networks.rnn import RNNNetwork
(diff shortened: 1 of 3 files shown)


ArushikaBansal:
Made necessary changes.
issue #3351

ArushikaBansal (commented):
Removed soft dependencies

fkiraly (changes requested):
Very nice, thanks, @ArushikaBansal!

Before we merge, may I ask you to fix the systematic error that @achieveordie had found here?
https://github.com/sktime/sktime/issues/4067

That is, parameteres that should be passed to the undlerying network - here, `RNNNetwork` - were not actually being passed.

Have a look how this is done in `CNNRegressor.__init__` after @achieveordie fixed it, and let us know if you have any questions.

fkiraly:
Looks almost good now!

The failing test says you are probably missing a `super.__init__` call in the constructor.

fkiraly:
I think, still? In `RNNNetwork`, have a look at the error message in the CI test.

fkiraly:
added the `super` call that I think was missing, hope you don't mind (otherwise we can revert)

fkiraly (changes requested):
Nice work!

I think we have now fixed the `super` issues.

However, before merge:

* `__init__` of `SimpleRNNRegressor` passes an argument `batch_size` to the `RNNNetwork` which it does not have - this will break if passed! I'm also surprised none of the tests break, have you checked with `check_estimators`?
* we should also add test parameters, i.e., `get_test_params`, with at least two test parameter settings for each object added. See the extension templates (in `extension_templates`) for how to fill this in.

fkiraly:
I now added test parameters and fixed things like the naming and init.

achieveordie (approved):
LGTM

fkiraly:
will merge after 0.18.0

PR #4190, opened 2023-02-02
[ENH] probabilistic forecasting rework part 1 - backend agnostic probability distributions

fkiraly:
This experimental PR introduces backend agnostic probability distributions, based on `BaseObject`, towards https://github.com/sktime/sktime/issues/4359

Currently, `BaseForecaster.predict_proba` relies on `tensorflow_probability` as a return type, which is a heavy dependency (>300MB in the python env).

This PR, instead, introduces a `BaseObject`-based interface for probability distributions, which is back-end agnostic, but has `tensorflow-probability` as one of the options for a back-end.

The conceptual model mixes indexed objects as in `pandas` and the array-based distributions in `tensorflow-probability`, i.e., array distributions which have `pandas`-based `index`, `columns`, and can be sub-set using `loc` and `iloc` indexing, similar to `pd.DataFrame`.

Advantages:
* row/column subsetting of the return, compatibility with `pandas` indices
* decouples the distributions from the `tensorflow` back-end, foundation for distribution metrics/losses
* allows to add methods to the distribution outside `tensorflow`, in `sktime`, e.g., heuristical integrated cdf or energy statistics required for CRPS metric (see `energy` method)

STEP: https://github.com/sktime/enhancement-proposals/pull/31
Related issue: https://github.com/sktime/sktime/issues/1746

Includes a full test suite for the interface, based on the `skbase` design.

Roadmap for subsequent PR:

* part 2: probabilistic (distributional) metrics, see https://github.com/sktime/sktime/pull/4276
* part 3: use in forecasters as return object if `tensorflow-probability` is not present; deprecation for 0.18.0 or 0.19.0
* part 4: use in tuned forecasters
* further probability distributions native to `sktime`
* interfacing more distributions from `tfp`

FYI @bethrice44, @adoherty21


fkiraly:
I think this is ready for review now, as tests are included.

The distributions are not yet used by anything else - use in metrics and forecasters will be added in subsequent PR.

PR #4191, opened 2023-02-03
[ENH] don't compute mtype metadata if not required

hoesler:
#### Reference Issues/PRs
Split out of #4140
Contributes to #4139

Some calls to `check_is_mtype` request metadata computation, which can be quite expensive especially for multi-index dataframes, only because they require the error message, which is unnecessarily conditional on that.

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/datatypes/_check.py
+++ b/sktime/datatypes/_check.py
@@ -248,7 +248,9 @@ def check_raise(obj, mtype: str, scitype: str = None, var_name: str = "input"):
     if valid:
         return True
     else:
-        raise TypeError(f"Expected object mtype to be one of {list(mtype)}, was {type(obj)}")
+        raise TypeError(
+            f"Expected object mtype to be one of {list(mtype)}, was {type(obj)}"
+        )
 
 
 def mtype(
--- a/sktime/datatypes/_check.py
+++ b/sktime/datatypes/_check.py
@@ -237,18 +237,18 @@ def check_raise(obj, mtype: str, scitype: str = None, var_name: str = "input"):
     ValueError if mtype input argument is not of expected type
     """
     obj_long_name_for_avoiding_linter_clash = obj
-    valid, msg, _ = check_is_mtype(
+    valid = check_is_mtype(
         obj=obj_long_name_for_avoiding_linter_clash,
         mtype=mtype,
         scitype=scitype,
-        return_metadata=True,
+        return_metadata=False,
         var_name=var_name,
     )
 
     if valid:
         return True
     else:
-        raise TypeError(msg)
+        raise TypeError(f"Expected object mtype to be one of {list(mtype)}, was {type(obj)}")
 
 
 def mtype(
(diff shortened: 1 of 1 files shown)


sktime/datatypes/_check.py
```diff
@@ -302,16 +302,16 @@ def mtype(
...
             mtype=m_plus_scitype[0],
             scitype=m_plus_scitype[1],
-            return_metadata=True,
+            return_metadata=False,
         )
         if valid:
             mtypes_positive += [m_plus_scitype[0]]
         else:
-            mtypes_negative[m_plus_scitype[0]] = msg
+            mtypes_negative[m_plus_scitype[0]] = ""
```
hoesler, line 314: Copied from the previous discussion in #4140

> @[hoesler](https://github.com/hoesler) hoesler [2 days ago](https://github.com/sktime/sktime/pull/4140/files#r1092252104)
> @fkiraly This is a TODO. The idea was to reduce computation time of checks by setting return_metadata=False, as metadata is not required here. The problem is, however, that error messages are turned off as well. I think the best would be to always return error messages, but that would require an api change of check_is_mtype. What's your opinion?

> @[fkiraly](https://github.com/fkiraly) fkiraly [2 days ago](https://github.com/sktime/sktime/pull/4140/files#r1092276277) • 
> I see what you're doing - very smart observation imo.
> 
> The "clean" way would be a refactor with deprecation.
> One way to do it would be to introduce a return_msg argument everywhere, which can be True, False, or None (default).
> If None, the current behaviour happens and a deprecation warning is raised. If True, always returns the msg return, independently of return_metadata. If False, never returns the msg arg.
> After a deprecation cycle, the default changes to False.

> @[fkiraly](https://github.com/fkiraly) fkiraly [2 days ago](https://github.com/sktime/sktime/pull/4140/files#r1092277077) • 
> it is nightmarish though since the argument would have to be introduced in all the checker functions.
> (the deprecation logic should sit in some public functions rather than in the checkers)
> 
> I would definitely advise a separate PR for this with only that as the purpose, since PRs should be focused in file footprint, or in scope. It is not possible to focus this in terms of file footprint...
> 
> unless you come up with a better idea.

sktime/datatypes/_check.py
```diff
@@ -237,18 +237,18 @@ def check_raise(obj, mtype: str, scitype: str = None, var_name: str = "input"):
...
-        return_metadata=True,
+        return_metadata=False,
         var_name=var_name,
     )
 
     if valid:
         return True
     else:
-        raise TypeError(msg)
+        raise TypeError(f"Expected object mtype to be one of {list(mtype)}, was {type(obj)}")
```
hoesler, line 251: Copied from the previous discussion in https://github.com/sktime/sktime/pull/4140

> @[fkiraly](https://github.com/fkiraly) fkiraly [2 days ago](https://github.com/sktime/sktime/pull/4140/files#r1092270788)
> this message would be confusing, as some mtypes are called like class types, e.g., pd.DataFrame.

fkiraly:
May I consider this superseded by https://github.com/sktime/sktime/pull/4389, @hoesler?

That follows the same idea, but doesn't turn off the error messages as a side effect.

fkiraly:
open for one week, @hoesler is non-responsive - possibly easter holidays.
I strongly assume he won't have a problem with assuming a "yes" here, given earlier interaction.

PR #4194, opened 2023-02-03
[ENH] add native multi-index/hierarchical data support to `Imputer`

hoesler:
#### Reference Issues/PRs

This PR allows Imputer to transform multi-index dataframes directly instead using an expensive iterative instance wise
transformation for some imputation strategies.

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/transformations/series/impute.py
+++ b/sktime/transformations/series/impute.py
@@ -198,7 +198,9 @@ class Imputer(BaseTransformer):
         if self.method == "random":
             for col in X.columns:
                 X[col] = X[col].apply(
-                    lambda i, col=col: self._get_random(col) if np.isnan(i) else i  # noqa: B023
+                    lambda i, col=col: self._get_random(col)
+                    if np.isnan(i)
+                    else i  # noqa: B023
                 )
             return X
         elif self.method == "constant":
--- a/sktime/transformations/series/impute.py
+++ b/sktime/transformations/series/impute.py
@@ -8,8 +8,10 @@ __all__ = ["Imputer"]
 
 
 import numpy as np
+import pandas as pd
 from sklearn.utils import check_random_state
 
+from sktime.datatypes import VectorizedDF
 from sktime.forecasting.base import ForecastingHorizon
 from sktime.forecasting.trend import PolynomialTrendForecaster
 from sktime.transformations.base import BaseTransformer
@@ -24,7 +26,7 @@ class Imputer(BaseTransformer):
     Parameters
     ----------
     method : str, default="drift"
-        Method to fill the missing values values.
+        Method to fill the missing values.
 
         * "drift" : drift/trend values by sktime.PolynomialTrendForecaster(degree=1)
             first, X in transform() is filled with ffill then bfill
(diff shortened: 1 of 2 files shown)


hoesler:
Guess I forgot to cherry pick the formatting fixes here.

fkiraly (changes requested):
Very nice!

I would request to change the "backwards detour" to `self._vectorize` inside the private logic. This coupling to the boilerplate layer is potentially problematic, e.g., in a case where we want to change the boilerplate layer (example: rename `_vectorize` to `_bob`), it is no longer independent.

Instead, I would do as follows:

* at the end of `__init__`, make a case disticntion based on the `method` parameter
* if it is not `"mean"` or `"median"` or the others that you make more efficient (methods in line 211), dynamically set the `X_inner_mtype` tag to `"pd.DataFrame"` (use `self.set_tags`, search the codebase to find similar uses)
* it's worth adding comments around that, i.e., why do we do that (for developers)
* then, `_fit` and `_transform` for the untreated methods will never receive an `X` with `pd.MultiIndex`, so you can just remove the detour in the first case block

This avoids coupling `_vectorize` into the private layer.
(generally, architecture-wise, we don't want to back-couple the private layer to the boilerplate layer, dependencies should be unidirectional; can be nested in composition, but that should interact only with the public interface)

hoesler:
Your suggestion to not call `_vectorize` directly make sense, given the current design.

While incorporating the request, I stumbled over this code block again:

```
            # fill first/last elements of series,
            # as some methods (e.g. "linear") can't impute those
            X = X.fillna(method="ffill").fillna(method="backfill")
```

This "cleanup" method feels a bit random and is also unable to catch one edge case: a series with only NAs.
Shouldn't this be handled more explicitly by chaining imputers for example?

What's your opinion on this?

fkiraly:
> This "cleanup" method feels a bit random and is also unable to catch one edge case: a series with only NAs.
> Shouldn't this be handled more explicitly by chaining imputers for example?

> What's your opinion on this?

First, I would say (without talking in detail about content): different scope, so different PR (or issue).
From the general principle, PRs should be scope focused (and/or file focused if possible).
And, that when looking through by-line history, the description of the PR that pops up should match the change (which is better done with focused PR).

About the content:

* I think the presence of the lines covers more edge cases than when they were absent - namely, starting and tailing sequences with NA; versus "only NA" series.
* the `Imputer` also has the property `capability:missing_values:removes`, which (when set to `True`) guarantees a contract of removing all NAs. Of course we could set it to `False` for certain inputatoin methods that cannot deal with start/tail, but that might not be as clean as a property guarantee.
* another option could be to have a parameter that controls what happens to NA at start or end. Another common option would be "cut", i.e., remove any starting or tailing NA together with their indices. A third option could be leaving them unchanged.

I don't have a strong opinion, the important thing would be to ensure whatever we do is documented in the docstring (not prominently but at all); and if we make a change, we would need to deprecate. https://www.sktime.org/en/stable/developer_guide/deprecation.html

hoesler:
> First, I would say (without talking in detail about content): different scope, so different PR (or issue).
> From the general principle, PRs should be scope focused (and/or file focused if possible).

Sorry to bring this up here. I'm just trying to copy the old behavior to the multi-index version, so I must understand it and add tests. Thereby I found this edge case, which is not covered yet.

Ok, so I'll leave this case (and tests for it) out of scope and just add a groupby/ffill+bfill.

fkiraly:
> Sorry to bring this up here

No reason at all to be sorry! I think it is excellent that you spotted it.
We just need to think about deprecation etc.

fkiraly (approved):
Nice, all points addressed.

You did the conditional tag setting the other way round, I suppose that's more robust, in case someone adds an imputation method.

PR #4195, opened 2023-02-03
[ENH] Improve vectorization performance

hoesler:
#### Reference Issues/PRs

This PR primarily improves instance iteration performance in `VectorizedDF`, by replacing pandas loc-iteration with groupby. It also fixes a memory leak I observed when `_get_X_at_index` is called multiple times, by returning a copy of a slice (I still don't fully understand the root cause, but this solved it).

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/datatypes/_vectorize.py
+++ b/sktime/datatypes/_vectorize.py
@@ -312,15 +312,24 @@ class VectorizedDF:
 
     def as_list(self):
         """Shorthand to retrieve self (iterator) as list."""
-        if isinstance(self.X_multiindex.index, pd.MultiIndex) and self.iterate_as == "Series" and not self.iterate_cols:
+        if (
+            isinstance(self.X_multiindex.index, pd.MultiIndex)
+            and self.iterate_as == "Series"
+            and not self.iterate_cols
+        ):
+
             def _convert(series):
-                series = series.droplevel(list(range(self.X_multiindex.index.nlevels - 1)))
+                series = series.droplevel(
+                    list(range(self.X_multiindex.index.nlevels - 1))
+                )
                 series = _enforce_index_freq(series)
                 return series
 
             return [
-                _convert(series) for _, series in
-                self.X_multiindex.groupby(level=list(range(self.X_multiindex.index.nlevels - 1)))
+                _convert(series)
+                for _, series in self.X_multiindex.groupby(
+                    level=list(range(self.X_multiindex.index.nlevels - 1))
+                )
             ]
         else:
             return list(self)
--- a/sktime/datatypes/_vectorize.py
+++ b/sktime/datatypes/_vectorize.py
@@ -310,29 +310,42 @@ class VectorizedDF:
             col_ind = [col_ind]
         return row_ind, col_ind
 
+    def __iter__(self):
+        """Iterate over all instances.
+
+        Returns
+        -------
+        An iterator over all instances
+        """
+
+        def _iter_cols(inst):
+            if self.iterate_cols:
+                for col in inst.columns:
+                    yield _enforce_index_freq(inst[[col]])
+            else:
+                yield _enforce_index_freq(inst)
+
+        iter_levels = 0
+
+        if self.is_scitype == "Panel":
+            if self.iterate_as == "Series":
+                iter_levels = 1
+        if self.is_scitype == "Hierarchical":
+            if self.iterate_as == "Panel":
+                iter_levels = 2
+            elif self.iterate_as == "Series":
+                iter_levels = 1
+
+        if iter_levels > 0:
+            group_levels = list(range(self.X_multiindex.index.nlevels - iter_levels))
+            for _, group in self.X_multiindex.groupby(level=group_levels):
+                yield from _iter_cols(group.droplevel(group_levels))
+        else:
+            yield from _iter_cols(self.X_multiindex)
+
     def as_list(self):
         """Shorthand to retrieve self (iterator) as list."""
-        if (
-            isinstance(self.X_multiindex.index, pd.MultiIndex)
-            and self.iterate_as == "Series"
-            and not self.iterate_cols
-        ):
-
-            def _convert(series):
-                series = series.droplevel(
-                    list(range(self.X_multiindex.index.nlevels - 1))
-                )
-                series = _enforce_index_freq(series)
-                return series
-
-            return [
-                _convert(series)
-                for _, series in self.X_multiindex.groupby(
-                    level=list(range(self.X_multiindex.index.nlevels - 1))
-                )
-            ]
-        else:
-            return list(self)
+        return list(self)
 
     def reconstruct(
         self,
--- a/sktime/datatypes/_vectorize.py
+++ b/sktime/datatypes/_vectorize.py
@@ -312,10 +312,18 @@ class VectorizedDF:
 
     def as_list(self):
         """Shorthand to retrieve self (iterator) as list."""
-        return [
-            _enforce_index_freq(item[1].droplevel(level=self.X_multiindex.index.names[:-1])) for item in
-            self.X_multiindex.groupby(level=self.X_multiindex.index.names[:-1])
-        ]
+        if isinstance(self.X_multiindex.index, pd.MultiIndex) and self.iterate_as == "Series" and not self.iterate_cols:
+            def _convert(series):
+                series = series.droplevel(list(range(self.X_multiindex.index.nlevels - 1)))
+                series = _enforce_index_freq(series)
+                return series
+
+            return [
+                _convert(series) for _, series in
+                self.X_multiindex.groupby(level=list(range(self.X_multiindex.index.nlevels - 1)))
+            ]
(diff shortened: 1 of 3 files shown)


sktime/datatypes/_vectorize.py
```diff
@@ -285,9 +285,9 @@ def _get_X_at_index(self, row_ind=None, col_ind=None, X=None):
         elif row_ind is None:
             res = X[col_ind]
         else:
-            res = X[col_ind].loc[row_ind]
+            res = X.loc[row_ind, col_ind]
         res = _enforce_index_freq(res)
-        return res
+        return res.copy()
```
fkiraly, line 290: why do we add a `copy` here? genuinely asking.

sktime/datatypes/_vectorize.py
```diff
@@ -312,7 +312,27 @@ def _get_item_indexer(self, i: int, X=None):
 
     def as_list(self):
         """Shorthand to retrieve self (iterator) as list."""
-        return list(self)
+        if (
+            isinstance(self.X_multiindex.index, pd.MultiIndex)
+            and self.iterate_as == "Series"
```
fkiraly, line 317: would it make sense to also do this for `Panel`, and then just change the "-1" to "-2" below?

fkiraly (changes requested):
Hm, interesting.

Some questions (not yet change requests):

* given that this is more performant, would it make sense to have `__getitem__` and `as_list` rely on a private method that carries out creation of a `GroupBy` object, across rows and potentially cols? In `__getitem__` this would use the `GroupBy` iteration, and in `as_list` it would use the same aggregation strategy as your modification does?
* it shouldn't be too hard to also address multiple columns? A second `groupby`?
* covering the `Panel` case should also be possible, see my comment above

sktime/datatypes/_vectorize.py
hoesler: I mentioned it in the description. Without this, there is a memory leak which I can't really explain. I saw memory increasing above 50G (for ~100M parquet input data) until the python process got killed by the OS. It's probably pandas copying the frame for some reason. Pandas is really unclear to me, when it comes to returning copies vs refs. This observation is actually what started my journey.

sktime/datatypes/_vectorize.py
hoesler: Good idea.

sktime/datatypes/_vectorize.py
fkiraly: This is a bit dangerous as it is in a central location.
It could be something specific about your system.

Could you provide an example so I can verify the memory leak?
We should also raise a bug report then (I can also write it if you give a code example)

hoesler: Here you go. The generated data is similar to what mine looks like.

```
import os

import numpy as np
import pandas as pd
import psutil
from sktime.datatypes import VectorizedDF

index = pd.MultiIndex.from_product([
    np.arange(1000), pd.period_range(end="2022-12-31", periods=365 * 5, freq="D")
], names=["store", "date"])

data = np.random.default_rng().standard_normal((len(index), 50), dtype=np.float32)

df = pd.DataFrame(data=data, index=index)

process = psutil.Process(os.getpid())
print("{mem:.3f} GiB".format(mem=process.memory_info().rss / 1024 ** 3))  # ~0.4GiB

VectorizedDF(df).as_list()
```

This script runs until killed and returns with exit code 137 (oom).
`python oom.py; print $?`

hoesler:
Taking your questions into account, I though it would make sense to implement `__iter__`. So I did. But then I realized, that you replaced iteration for vectorization in 0.16 with a new  `vectorize_est` method, which is using index slicing directly and cannot take advantage of my improved iteration. I need to look into that first, before we can see the performance effects again.

hoesler:
I reimplemented `vectorize_est` taking advantage of fast iteration.

fkiraly:
> But then I realized, that you replaced iteration for vectorization in 0.16 with a new  `vectorize_est` method, which is using index slicing directly and cannot take advantage of my improved iteration. I need to look into that first, before we can see the performance effects again.

Apologies for crossing over here - but I see you looked into it and ... clearly understood the design, made it faster, further refactored and improved it, fixed some bugs on the side ... AMAZING!!!

fkiraly:
PS: when you're done reworking after a review, click this:

image

sktime/datatypes/_vectorize.py
fkiraly: I see you factored this out into its own function. Good idea.

sktime/datatypes/_vectorize.py
```diff
@@ -310,6 +309,118 @@ def _get_item_indexer(self, i: int, X=None):
...
+            or, if X is Hierarchical and iterate_as is "Panel"
+                then the class will iterate over individual Panels in X
+                    (Panel = flat/non-hierarchical collection of Series)
+        iterate_cols : boolean, optional, default=self.iterate_cols
+            whether to iterate over columns, if true, the class will iterate over
+            columns
+
+        Returns
+        -------
+        An iterator over all (row name, column name, instance) tuples.
```
fkiraly, line 350: this is not 100% correct - it doesn't return an iterator, but *is* a generator, no? Also, is this not going to get us into trouble when we, say, want to iterate through it twice?

fkiraly: Update, note to myself: I think no, because each call a new generator is returned.

sktime/datatypes/_vectorize.py
```diff
@@ -310,6 +309,118 @@ def _get_item_indexer(self, i: int, X=None):
...
+        """Get the levels to group by for iteration using iterate_as.
+
+        Parameters
+        ----------
+        iterate_as: The scitype of iteration elements
+
+        Returns
+        -------
+        A list of multiindex levels to group by for iteration
+        """
```
fkiraly, line 386: this could be simplified, no?
```python
iter_levels_dict = {"Panel": 2, "Series": 1}
iter_levels = iter_levels_dict.get(iterate_as, 0)
```
right?

fkiraly: (the compatibility between `is_scitype` and `iterate_as` is already checked at construction)

sktime/datatypes/_vectorize.py
fkiraly: ouch. Thanks, although it seems mysterious to me.

fkiraly (approved):
See above - amazing improvement.

I'll leave it unmerged in case you want to action any of the comments.

sktime/datatypes/_vectorize.py
hoesler: To the cite python docs:
> Python’s [generator](https://docs.python.org/3/glossary.html#term-generator)s provide a convenient way to implement the iterator protocol. If a container object’s __iter__() method is implemented as a generator, it will automatically return an iterator object (technically, a generator object) supplying the __iter__() and [__next__()](https://docs.python.org/3/reference/expressions.html#generator.__next__) methods. More information about generators can be found in [the documentation for the yield expression](https://docs.python.org/3/reference/expressions.html#yieldexpr).

https://docs.python.org/3/library/stdtypes.html#generator-types

And yes, each time a new generator is returned.

sktime/datatypes/_vectorize.py
hoesler: I don't think so. For example if `is_scitype == "Series" and iterate_as == "Series"`, `iter_levels` must be 0 not 1.

fkiraly: ahhhh, I see. You are right.

How about the following then:

```python
if self.is_scitype != iterate_as:
  iter_levels_dict = {"Panel": 2, "Series": 1}
  iter_levels = iter_levels_dict.get(iterate_as, 0)
else:
  iter_levels = 0
```

Although this seems less explicit and more error prone.

I would say, let's leave it then.

sktime/datatypes/_vectorize.py
fkiraly: well, we should at least update the docstring then for it to be correct. It returns a generator, not an iterator.

fkiraly: I hope I didn't just end up confusing myself, so please correct me if I'm wrong.

hoesler: Technically, you are correct. But a generator [behaves like an iterator](https://wiki.python.org/moin/Generators) and I'm my opinion, this is an implementation detail, a consumer of the method shouldn't care about. Don't you think so?

fkiraly: I do, but docstrings should be correct up to the type. A consumer will not care in most cases, but I think it's good practice to be technically correct.

How about saying sth like `generator, iterates over all (row name, column name, instance) tuples. The i-th element returned by the generator is ...`

fkiraly: ok, I'll merge it, and we can talk the docstring separately

Issue #3827, opened 2022-11-22
[ENH] Change logic of checks


danbartl:
Investigation to see why checking for datatypes can take very long.
A description of all checks can be found here.
https://docs.google.com/spreadsheets/d/1cDx_yg2HhFcPfNR1Trc1Vr1FDuSOTkXwid5GTKIS2aE/edit?usp=sharing

After some investigation, most computing time was consumed when having _index_equally_spaced for periodindex.

Some investigations revealed that groupby can optimize / reduce time needed slightly, but periodindex should be avoided.

Findings are included here:
https://github.com/danbartl/sktime/blob/check_test/perftest.ipynb

Conclusion: 

- Will create a pull request to change method to groupby, increasing performance
- periodindex should be avoided for large datasets
- unfortunately, only periodindex supports freq argument for multiindex, further supporting the idea that freq should be outside argument
- polars library performs 5 times as fast as fastest pandas method

fkiraly:
Amazing and impressive!

Based on this, I wonder whether we can integrate a slightly more polished version of the performance testing into `sktime` itself, as a developer utility.

fkiraly:
Re some of these checks:

some changes can be made *by accident* by transformers, but this can be certified in the test suite (as opposed to in an output check).

Regarding whether transformers can change:
* in-principle, transformers could return an empty data frame, e.g., when subsetting or selecting
* transformers can also change the number of time points, number of samples in a panel, number of variables

Issue #2337, opened 2022-03-28
[ENH] Investigate `visions` for use in `datatypes` module


fkiraly:
@lmmentel pointed out the [`visions`](https://github.com/dylan-profiler/visions) library which provides abstract functionality for semantic typing, similar to our `datatypes` module.

We should investigate whether we could use it to replace logic in the `datatypes` module and whether it has sufficient features for our needs.

E.g., could we move our `_check` functions to `contains_op` etc.

I have opened an issue for discussion on the visions repo here: https://github.com/dylan-profiler/visions/issues/194

fkiraly:
FYI @miraep8, @lmmentel, re typing/base classes

PR #4196, opened 2023-02-03
[ENH] Improve panel mtype check performance

hoesler:
#### Reference Issues/PRs

Overlaps with https://github.com/sktime/sktime/issues/3827, https://github.com/sktime/sktime/pull/3991

This PR improves the performance of `check_pdmultiindex_panel`.

It was developed in parallel to the work of @danbartl, so some improvements might be outdated or just implemented differently. I would be grateful, if he could support here.

##### For all contributions

##### For new estimators

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -2062,6 +2062,15 @@
       "contributions": [
         "maintenance"
       ]
+    },
+    {
+      "login": "hoesler",
+      "name": "Christoph Hösler",
+      "avatar_url": "https://avatars.githubusercontent.com/u/1052770?v=4",
+      "profile": "https://www.linkedin.com/in/hoesler/",
+      "contributions": [
+        "code"
+      ]
     }
   ]
 }
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -2073,14 +2073,14 @@
       ]
     },
     {
-       "login": "pranavvp16"
-       "name": "Pranav Prajapati"
-       "avatar_url": "https://avatars.githubusercontent.com/u/94780581?v=4"
-       "profile": "https://www.linkedin.com/in/pranav-prajapati-a5b413226/"
-       "contributions":[
-          "code",
-          "test"
-       ]
+      "login": "pranavvp16"
+      "name": "Pranav Prajapati"
+      "avatar_url": "https://avatars.githubusercontent.com/u/94780581?v=4"
+      "profile": "https://www.linkedin.com/in/pranav-prajapati-a5b413226/"
+      "contributions":[
+        "code",
+        "test"
+      ]
     }
   ]
 }
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -2073,10 +2073,10 @@
       ]
     },
     {
-      "login": "pranavvp16"
-      "name": "Pranav Prajapati"
-      "avatar_url": "https://avatars.githubusercontent.com/u/94780581?v=4"
-      "profile": "https://www.linkedin.com/in/pranav-prajapati-a5b413226/"
+      "login": "pranavvp16",
+      "name": "Pranav Prajapati",
+      "avatar_url": "https://avatars.githubusercontent.com/u/94780581?v=4",
+      "profile": "https://www.linkedin.com/in/pranav-prajapati-a5b413226/",
       "contributions":[
         "code",
         "test"
--- a/sktime/datatypes/_panel/_check.py
+++ b/sktime/datatypes/_panel/_check.py
@@ -220,7 +220,7 @@ def check_pdmultiindex_panel(obj, return_metadata=False, var_name="obj", panel=T
 
     # Check time index is ordered in time
     index_frame = obj.index.to_frame()
-    if not index_frame.groupby(level=list(range(obj.index.nlevels - 1)))[index_frame.columns[-1]].is_monotonic_increasing.all():
+    if not index_frame.groupby(level=list(range(obj.index.nlevels - 1)))[index_frame.columns[-1]].is_monotonic_increasing.astype(bool).all():
         msg = (
             f"The (time) index of {var_name} must be sorted monotonically increasing, "
             f"but found: {index}"
--- a/sktime/datatypes/_panel/_check.py
+++ b/sktime/datatypes/_panel/_check.py
@@ -171,6 +171,8 @@ def check_pdmultiindex_panel(obj, return_metadata=False, var_name="obj", panel=T
         msg = f"{var_name} must have a MultiIndex, found {type(obj.index)}"
         return _ret(False, msg, None, return_metadata)
 
+    index = obj.index
+
     # check that columns are unique
     col_names = obj.columns
     if not col_names.is_unique:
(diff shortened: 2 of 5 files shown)


danbartl:
looks great to me, I think the only "collision" is with my approach to change get_cutoff, but I removed that since?

danbartl:
Hi I am really confused now, to me performance seems to go down with the latest PR?

hoesler:
> Hi I am really confused now, to me performance seems to go down with the latest PR?

What are you referring to with "with the latest PR"?
I am still getting a much better performance for `check_pdmultiindex_panel` here than on main.

0.09s vs 12.4s using this test case:

```
import timeit

import numpy as np
import pandas as pd

from datatypes._panel._check import check_pdmultiindex_panel

index = pd.MultiIndex.from_product(
    [np.arange(1000), pd.period_range(end="2022-12-31", periods=365 * 5, freq="D")],
    names=["store", "date"],
)

data = np.random.default_rng().standard_normal((len(index), 50), dtype=np.float32)

df = pd.DataFrame(data=data, index=index)

print(timeit.timeit(lambda: check_pdmultiindex_panel(df), number=10) / 10)
```

danbartl:
Hi sorry didnt have the time test thoroughly, I use the following code 
```
# -*- coding: utf-8 -*-
from sklearn.linear_model import LinearRegression
from sklearn.pipeline import make_pipeline

from sktime.forecasting.compose import ForecastingPipeline, make_reduction
from sktime.forecasting.compose._reduce import _RecursiveReducer

regressor = make_pipeline(
    LinearRegression(),
)

kwargs = {
    "lag_feature": {
        "lag": [1],
    }
}


forecaster_global = make_reduction(
    regressor,
    scitype="tabular-regressor",
    transformers=[WindowSummarizer(**kwargs, n_jobs=1, truncate="bfill")],
    window_length=None,
    strategy="recursive",
    pooling="global",
)

y = _make_hierarchical(
    hierarchy_levels=(10000,), min_timepoints=1000, max_timepoints=1000
)
from sktime.datatypes._utilities import get_time_index

y_no_freq = get_time_index(y.reset_index().set_index(["h0","time"]))


def Main():

    # from time import perf_counter
    # t1_start = perf_counter()

    _ = forecaster_global.fit(y)

    y_pred_global = forecaster_global.predict(fh=[1,2])

    # t1_stop = perf_counter()
    # print("Elapsed time during the whole program in seconds:", t1_stop-t1_start)


import cProfile
import pstats

cProfile.run("Main()", "output_newmulti_adjust.dat")

```

Which resulted for me in a slower performance when  investigating the  cProfile file. Hope I can dig  in  a bit more tomorrow.

danbartl:
> Ok, so you are testing more than just `check_pdmultiindex_panel`. I think that can be misleading, or were you comparing CPU times for the function within the profiles?

I broke it down to your example 
```
import timeit

import numpy as np
import pandas as pd

from sktime.datatypes._panel._check import check_pdmultiindex_panel

# index = pd.MultiIndex.from_product(
#     [np.arange(1000), pd.period_range(end="2022-12-31", periods=365 * 5, freq="D").to_timestamp()],
#     names=["store", "date"],
# )
y = _make_hierarchical(
    hierarchy_levels=(10000,), min_timepoints=1000, max_timepoints=1000
)

data = np.random.default_rng().standard_normal((len(index), 50), dtype=np.float32)

df = pd.DataFrame(data=data, index=index)

print(timeit.timeit(lambda: check_pdmultiindex_panel(y), number=10) / 10)
```

I think the key point is that you consider PeriodIndex, while I conser DateTimeIndex. With Datetimeindex and many groups, the PR code presented here takes twice as long on my PC. Maybe we need to implement a logic to check PeriodIndex.

hoesler:
Interesting. I will investigate further with your code.
And yes, I was focusing on PeriodIndex. The main finding here was that operations are fast if you use its numeric representation, like with `is_full`.

hoesler:
Did a quick comparison of three scenarios:
My example with two different index types and yours.

**This PR@0e13388:**
panel_period: 0.090
panel_datetime: 0.098
hier_datetime: 1.1557

**main@8a14adb:**
panel_period: 12.308
panel_datetime: 0.628
hier_datetime: 1.472

My code performed consistently better. For your example the difference is small but it's sill better. For the other two it's significant. Not only for the period index.

Very strange, that we observe so different outcomes.

```
import timeit

import numpy as np
import pandas as pd

from datatypes._panel._check import check_pdmultiindex_panel
from utils._testing.hierarchical import _make_hierarchical


def create_panel(time_range_fun):
    index = pd.MultiIndex.from_product(
        [np.arange(1000), time_range_fun(end="2022-12-31", periods=365 * 5, freq="D")],
        names=["store", "date"],
    )

    data = np.random.default_rng().standard_normal((len(index), 50), dtype=np.float32)

    return pd.DataFrame(data=data, index=index)


for key, val in {
    "panel_period": create_panel(pd.period_range),
    "panel_datetime": create_panel(pd.date_range),
    "hier_datetime": _make_hierarchical(
        hierarchy_levels=(10000,), min_timepoints=1000, max_timepoints=1000
    ),
}.items():

    def func():
        check_pdmultiindex_panel(val, return_metadata=False)

    n = 10
    avg_time = timeit.timeit(func, number=n) / n

    print(f"{key}: {avg_time}")
```

hoesler:
Found some further improvements.
Latest benchmark (min runtime, now **with** metadata calculation for the scenarios above except yours now with sorted index):

panel_period: 0.145111
panel_datetime: 0.172680
hier_datetime: 0.991199

vs

panel_period: 12.283228
panel_datetime: 0.544896
hier_datetime: 1.357230

The slowest part is now `_index_equally_spaced`, but I don't see how we could improve that, except for maybe storing it's result as metadata in the index to prevent repeated calculation. Don't know if that is feasible, but worth investigating.
And of course, prevent unnecessary metadata calculation (#4191)!

fkiraly:
FYI, I'm currently working on using the config mechanisms to turn off checks, metadata etc - I suppose that's orthogonal to this PR.

hoesler:
Nice. Thanks. I didn't start yet with an implementation, but my idea was to split check and metadata into two methods and, as a first step, call the existing functions as fallback somehow. But I guess this is indeed orthogonal.

fkiraly:
> but my idea was to split check and metadata into two methods and, as a first step, call the existing functions as fallback somehow. But I guess this is indeed orthogonal.

Hm, good idea. How would it look like, in an example?

fkiraly:
Here are two relevant issues:

* https://github.com/sktime/sktime/issues/3512 - I think the mtypes should be turned into classes, they are a bit all over the place now.
* https://github.com/sktime/sktime/issues/2337 - we also thought of using `visions` in a refactor

danbartl:
@fkiraly @hoesler 
I still get a reduced performance for the 10,000 groups example. Still in my view this can be merged, since performance is real good for less groups and the increase is not too high and apparently hardware dependent.  @fkiraly  if you have time to run the code on your machine would be nice, else it is good to go from my side.

Thanks a lot!

fkiraly:
Performance statistics

`main@c41efee7f` - note that includes #4195
```
panel_period: 37.70587424999103
panel_datetime: 1.4263513099984266
hier_datetime: 2.135691480000969
```

`this@0e1338875`
```
panel_period: 0.2671109699993394
panel_datetime: 0.28593878000974654
hier_datetime: 2.568562889995519
```

fkiraly (approved):
Thanks a lot! Great improvement!

I would also approve it.

fkiraly:
> Still in my view this can be merged, since performance is real good for less groups and the increase is not too high and apparently hardware dependent. @fkiraly if you have time to run the code on your machine would be nice, else it is good to go from my side.

I'll take this as approval, @danbartl - did you want to press the "approve" button?

PR #4200, opened 2023-02-04
[DOC] remove gap between pandas and DataFrame | Series in classification notebook (#4200)

### LINKED ISSUE #4189: [DOC] class and function references are rendering as plain text - ideas for improvement

Example:
image
from https://www.sktime.org/en/stable/examples/02_classification.html?highlight=nested%20dataframe#Datasets-and-Problem-Types

This looks kind of odd and doesn't do a good job of showing that `pandas` and `DataFrame` belong together.

Instead of `pandas` `DataFrame`, why not at least `pandas.DataFrame`?

Better yet, have you considered using something like [intersphinx](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html)? I'm happy to submit a PR to convert a few of these every time I read through a page, but wanted to hear from maintainers if there's a reason this hasn't been done yet.

romanlutz:
#### Reference Issues/PRs

Fixes #4189 

Removes the unnecessary gap between `pandas` and the class names. This only applies in these cases from what I can tell, since in other cases like `sktime` `make_pipeline` we'd need to add the in-between modules. If that is ever desired I suggest switching to `intersphinx` since it will also validate all those references, so in case there's a typo or anything changes you'll know immediately.

no

##### For all contributions

##### For new estimators

### DIFF
--- a/examples/02_classification.ipynb
+++ b/examples/02_classification.ipynb
@@ -165,7 +165,7 @@
     "collapsed": false
    },
    "source": [
-    "Some data sets have unequal length series. Two data sets with this characteristic are shipped with sktime: PLAID (univariate) and JapaneseVowels (multivariate). We cannot store unequal length series in `numpy` arrays. Instead, we use a nested `pandas` `DataFrame`, where each cell is a `pandas` `Series`. This is the default return type for all single problem loaders."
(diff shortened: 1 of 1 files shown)


fkiraly (approved):
Yes, looks like these are all the instances.

PR #3777, opened 2022-11-13
`skbase` refactor - part 2: `all_estimators` lookup

fkiraly:
Part 2 of the change action to move the base module to `skbase`.

Moves `BaseObject` and core `base` functionality to the `skbase` package.

Part 2, relies on:

* part 1: https://github.com/sktime/sktime/pull/3151
* https://github.com/sktime/sktime/pull/4512

Migrated:
* `all_estimators` - uses `skbase.all_objects` under the hood now.
* from part 1 - `BaseObject`. The `sktime` `BaseObject` is a child of `skbase` `BaseObject` with some added features that are currently `sktime` specific.

todo, will happen in different parts of this:
* `TestAllObjects`
* add tests

PR #4208, opened 2023-02-05
Experimental - DO NOT MERGE: `BaseObject` refactor - part 2: `all_estimators` lookup - git/main dependent

fkiraly:
Same as https://github.com/sktime/sktime/pull/3777, but depends on the current GitHub main to test the unreleased fix https://github.com/sktime/skbase/pull/115


fkiraly:
@RNKuhns, I'm not quite sure what is wrong here, but it would look like there is something wrong with inspecting the set of parameters?

There seems to be a real problem, but it is unrelated to the one we wanted to test, i.e., the meta-estimator here: https://github.com/sktime/sktime/pull/3777

fkiraly:
@RNKuhns, found what the issue was - I did not take into account the deprecation of `ignore_modules` in favour of `modules_to_ignore` - that should be fixed now

fkiraly:
Hm, it doesn't seem to be ignoring the modules we want it to ignore. Some bug that crept in through renaming the argument?

fkiraly:
superseded by https://github.com/sktime/sktime/pull/3777

PR #4214, opened 2023-02-09
 [ENH] transformer interfacing numpy.fft for simple fourier transform

blazingbhavneek:
#### No new dependency
Creating a new PR, continued from [here](https://github.com/sktime/sktime/pull/4146)
(Sorry for the inconvenience @fkiraly )


blazingbhavneek:
@fkiraly @achieveordie What should I do to fix this?

sktime/transformations/series/simple_fourier.py
```diff
@@ -0,0 +1,72 @@
...
+        "fit_is_empty": True,
+        "capability:inverse_transform": False,
+        "capability:unequal_length": True,
+        "handles-missing-data": False,
+    }
+
+    def __init__(self):
+        super(SimpleFourier, self).__init__()
+
+    def _transform(self, X):
```
achieveordie, line 53: `transform()` method of [BaseTransformer](https://github.com/sktime/sktime/blob/main/sktime/transformations/base.py#L504) calls `._transform()` with a y parameter as well. In your implementation, there's no y parameter and hence the error. 

I suggest passing y even if the operation is solely on the predictors.

blazingbhavneek: Done!

blazingbhavneek:
@achieveordie It's still giving the same error, what should I do? (Can i set these checks locally or something?)

fkiraly:
I think in this line `Y = pd.Series(dft_seq[1:], index=sample_freq[1:])`, you are passing values and index of different length somehow?

blazingbhavneek:
@fkiraly Hey! I tried using this in a notebook, and it gave me a "not fitted" error, I have checked that `fit_is_empty` is set to `true`, maybe that's the issue. Should I write a fit then?

fkiraly:
no, it shouldn't be required to write an extra `_fit` if `fit_is_empty=True`, that should be fine in this case.

Could you kindly post the code with the unexpected behaviour here? I can then look into it for you.

blazingbhavneek:
@fkiraly Here
image

fkiraly:
ah, even for non-fitting transformers, you must call `fit_transform` and not just `transform`.

A while ago there was a long design discussion whether `transform` should not just call `fit_transform` in this case, but the majority was against that.

fkiraly:
doctest is failing - users shouldn't call the private methods.

blazingbhavneek:
Hey! @fkiraly, I got the issue!

When I called the private method, the transformation went fine, but when called through `fit_transform,` it did the conversion of some sort, which passed through `rfft` created a `ValueError`. By changing the X_inner_mtype to `pd.Series` the conversion was skipped, and the function works fine now!

fkiraly:
hm, checks are failling - what if you set the `X_inner_mtype` to `np.ndarray`?

blazingbhavneek:
> hm, checks are failling - what if you set the `X_inner_mtype` to `np.ndarray`?

@fkiraly It gives `ValueError`, the same as you mentioned above

> I think in this line `Y = pd.Series(dft_seq[1:], index=sample_freq[1:])`, you are passing values and index of different length somehow?

sktime/transformations/series/simple_fourier.py
```diff
@@ -0,0 +1,72 @@
...
+    --------
+    >>> from sktime.transformations.series.simple_fourier import SimpleFourier
+    >>> from sktime.datasets import load_airline
+    >>> X = load_airline()
+    >>> transformer = SimpleFourier()
+    >>> X_ft = transformer.fit_transform(X)
+    """
+
+    _tags = {
+        "scitype:transform-input": "pd.Series",
```
fkiraly, line 36: This should be `Series`, not `pd.Series`, Same for the output.
`X_inner_mtype` should be `pd.Series`.

fkiraly (changes requested):
I think I understand why it´s failing, see above.
Bit of an confusing naming convention.

sktime/transformations/series/simple_fourier.py
blazingbhavneek: Hey! I tried this, but it is giving me a `TypeError`, AND it is also returning an output

image

fkiraly: that won't work, if you look at the tag specification, in the extension template, it needs to be `Series`, `Panel`, or `Primitives`.
I think both should be `"Series"` (just the single string)

blazingbhavneek: Hey! So I expanded the `TypeError` to see whats wrong; Basically the issue was with index of the returned series, it shouldn't be float64 but RangedIndex, which is fine because the simple-interfacing anyways makes the index a ranged float (0-0.5 with size `X.size`)

image

fkiraly: yes, resetting the index would probably solve it.

blazingbhavneek: Should I make it limited to univariate only? Or add support for multivariate as well?

image

fkiraly: I would recommend to track it down where this is coming from - i.e., from the framework or from inside the function. I.e., debug using `check_estimator` with the `raise_exceptions` flag set, and have a look at the traceback.

fkiraly:
yes, that seems to have done the trick!

blazingbhavneek:
So, this is it? Or is there more to be done? @fkiraly

fkiraly (changes requested):
Yes, very nice!

In-principle, this is it code-wise.
Let's go once over the docs, I would kindly request:

* in the docstring say that this computes the Fourier transform of a series - say precisely what values and index it produces for a (pandas) series, refer to the `numpy` method
* kindly add the estimator to the API reference, in `docs/source/api_reference/transformations.rst` so users can easily find it

blazingbhavneek:
All Done! @fkiraly

fkiraly (changes requested):
Thanks!

Two change requests before merging:

* I just checked the old issue #4146 which this started as, there is still a conversation open about naming, with @achieveordie and @KishManani.
I would recommend we name the estimator `FourierTransform` or similar.
* in the docstring, small correction: the Fourier transform is not a periodicity test (we should onyl say this for statistical tests). It is simply a (feature/data) transformation. Tests can be done on top of it.

blazingbhavneek:
All done! @fkiraly (ig?)

fkiraly (approved):
Yup - thanks a lot for this nice contribution!

PR #4215, opened 2023-02-09
 [ENH] interface the seasonal package as a parameter estimator for seasonality

### LINKED ISSUE #3835: [ENH] interface the `seasonal` package as a parameter estimator for seasonality (`BaseParamFitter` descendant)
We now have an interface design for generic parameter estimators, with `BaseParamFitter`.
(original design issue: https://github.com/sktime/sktime/issues/1056)

Still experimental, we should further explore it by interfacing useful parameter estimators.

@mloning suggested the methods in the `seasonal` package in https://github.com/sktime/sktime/issues/1056, given the existing base class interface and seasonality examples it seems like a good first issue.

blazingbhavneek:
Fixes #3835

Interfaces welch/seasonal package as seasonality param_est

Introduces soft dependancey: [Seasonal](https://github.com/welch/seasonal)

### DIFF
--- a/sktime/param_est/seasonality.py
+++ b/sktime/param_est/seasonality.py
@@ -416,11 +416,12 @@ class SeasonalityPeriodogram(BaseParamFitter):
 
     Parameters
     ----------
-    candidate_sp : None, int or list of int, optional, default = None
-        candidate sp to test, and to restrict tests to
-        if None, min_period and max_period will default to 4 and None
-        if int, min_period = candidate_sp and max_period will be None
-        if list of int (size 2), will set min_period and max_period respectively
+    min_period : int
+        Disregard periods shorter than this number of samples.
+        Defaults to 4
+    max_period : int
+        Disregard periods longer than this number of samples.
+        Defaults to None
 
     Attributes
     ----------
@@ -432,7 +433,7 @@ class SeasonalityPeriodogram(BaseParamFitter):
     Examples
     --------
     >>> from sktime.datasets import load_airline
-    >>> from sktime.param_est.seasonality import Seasonality
+    >>> from sktime.param_est.seasonality import SeasonalityPeriodogram
     >>>
     >>> X = load_airline().diff()[1:]  # doctest: +SKIP
     >>> sp_est = Seasonality()  # doctest: +SKIP
@@ -452,8 +453,9 @@ class SeasonalityPeriodogram(BaseParamFitter):
         "python_dependencies": "seasonal",
     }
 
-    def __init__(self, candidate_sp=None):
-        self.candidate_sp = candidate_sp
+    def __init__(self, min_period=4, max_period=None):
+        self.min_period = min_period
+        self.max_period = max_period
         super(SeasonalityPeriodogram, self).__init__()
 
     def _fit(self, X):
@@ -474,28 +476,6 @@ class SeasonalityPeriodogram(BaseParamFitter):
         self : reference to self
         """
         from seasonal.periodogram import periodogram
-
-        self.max_period = None
-
-        if self.candidate_sp is None:
-            self.min_period = 4
-
-        elif isinstance(self.candidate_sp, int):
-            self.min_period = self.candidate_sp
-
-        elif isinstance(self.candidate_sp, list):
-            if len(self.candidate_sp) == 0:
-                self.min_period = 4
-            elif len(self.candidate_sp) == 1:
-                self.min_period = self.candidate_sp[0]
-            elif len(self.candidate_sp) == 2:
-                self.min_period = self.candidate_sp[0]
-                self.max_period = self.candidate_sp[1]
-            else:
-                raise RuntimeError("candidate_sp list must have size <= 2")
-
-        else:
-            raise TypeError("candidate_sp must be of type None, int or list of int")
 
         seasons, _ = periodogram(
             X, min_period=self.min_period, max_period=self.max_period
@@ -530,7 +510,7 @@ class SeasonalityPeriodogram(BaseParamFitter):
             `create_test_instance` uses the first (or only) dictionary in `params`
         """
         params1 = {}
-        params2 = {"candidate_sp": [5, 12]}
-        params3 = {"candidate_sp": [5, 12, 2]}
+        params2 = {"min_period": 5, "max_period": 12}
+        params3 = {"min_period": 5}
 
         return [params1, params2, params3]
--- a/sktime/param_est/seasonality.py
+++ b/sktime/param_est/seasonality.py
@@ -140,7 +140,7 @@ class SeasonalityACF(BaseParamFitter):
 
         candidate_sp = self.candidate_sp
         if candidate_sp is None:
-            candidate_sp = range(2, nlags)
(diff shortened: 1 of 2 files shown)


blazingbhavneek:
@achieveordie Can you tell me what is the issue here? (Last time you told me to change pyproject.toml)

sktime/param_est/seasonality.py
```diff
@@ -2,10 +2,11 @@
 # copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
 """Parameter estimators for seasonality."""
 
-__author__ = ["fkiraly"]
-__all__ = ["SeasonalityACF"]
+__author__ = ["fkiraly", "blazingbhavneek"]
+__all__ = ["SeasonalityACF", "Seasonality"]
 
 import numpy as np
+from seasonal.periodogram import periodogram
```
achieveordie, line 9: Soft imports should not be in the scope of the file but should rather be limited to the method(s) that use it. Take a look at tensorflow/keras in the DL submodule for an example of how to do that (It involves simply importing it within the method where this package is used.)

The reason behind this is simple - being a soft dependency, one might not have installed it but may still want to use a function/object (and this function/object should not be making use of this dependency). A good software design will not throw an error for an import which isn't supposed to be used in the code execution.

sktime/param_est/seasonality.py
```diff
@@ -407,3 +406,93 @@ def get_test_params(cls, parameter_set="default"):
...
+
+        Returns
+        -------
+        params : dict or list of dict, default = {}
+            Parameters to create testing instances of the class
+            Each dict are parameters to construct an "interesting" test instance, i.e.,
+            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+            `create_test_instance` uses the first (or only) dictionary in `params`
+        """
+        params1 = {}
```
achieveordie, line 495: Why leave it empty? Unless I'm misunderstanding something, the entire purpose of this method is to overwrite default params to facilitate faster testing. I suggest making use of a set of params that maximizes code coverage.

achieveordie (changes requested):

sktime/param_est/seasonality.py
```diff
@@ -141,7 +141,7 @@ def _fit(self, X):
 
         candidate_sp = self.candidate_sp
         if candidate_sp is None:
-            candidate_sp = range(2, nlags + 1)
+            candidate_sp = range(2, nlags)
```
ngupta23, line 144: Why is this being reverted back to nlags? nlags + 1 was added for a specific reason?

@fkiraly FYI...

fkiraly: uh-oh, I think this is just an accident, probably from file copy-paste.
Should be reverted.

sktime/param_est/seasonality.py
blazingbhavneek: Done!

sktime/param_est/seasonality.py
blazingbhavneek: I see I made a few mistakes here

1. There are no parameters required for this class, so we don't need them anyways
2. I have not written any tests for it either (simple interfacing task)

(Copied Franz Kiraly's code for reference, but forgot to delete this part)

fkiraly: it has `min_period` and `max_period`, I would expose these as parameters, possibly through `candidate_sp`

sktime/param_est/seasonality.py
```diff
@@ -407,3 +405,71 @@ def get_test_params(cls, parameter_set="default"):
         params2 = {"candidate_sp": [3, 7, 12]}
 
         return [params1, params2]
+
+
+class Seasonality(BaseParamFitter):
```
fkiraly, line 410: I wouldn´t call it `Seasonality`, maybe be more specific, e.g., `SeasonalityPeriodogram` or similar

blazingbhavneek: Done!

sktime/param_est/seasonality.py
blazingbhavneek: Done! I also wrote test parameters for it, will write tests for it later

sktime/param_est/seasonality.py
fkiraly: thanks

fkiraly (changes requested):
Thanks!

Can you please revert the merge accident that @ngupta23 pointed out?

I.e., please do not make any changes to the existing estimator, `SeasonalityACF`.

fkiraly:
I see the commit, but that weren't all changes. I would suggest, just copy `SeasonalityACF` from `main`.

sktime/param_est/seasonality.py
```diff
@@ -407,3 +407,130 @@ def get_test_params(cls, parameter_set="default"):
...
+
+
+class SeasonalityPeriodogram(BaseParamFitter):
+    """Score periodicities by their spectral power.
+
+    Interfacing `seasonal.periodogram` to determine candidate seasonality parameters.
+
+    Parameters
+    ----------
+    candidate_sp : None, int or list of int, optional, default = None
```
fkiraly, line 419: from the code, it looks like the default is 4?

Further, I think the mapping of `candidate_sp` on `min_period` and `max_period` might be confusing to the user.
Perhaps you just want to patch through `min_period` and `max_period`, that might be clearer?

sktime/param_est/seasonality.py
```diff
@@ -407,3 +407,110 @@ def get_test_params(cls, parameter_set="default"):
...
+        Defaults to 4
+    max_period : int
+        Disregard periods longer than this number of samples.
+        Defaults to None
+
+    Attributes
+    ----------
+    sp_ : int, seasonality period at lowest p-level, if any sub-threshold, else 1
+        if `candidate_sp` is passed, will be in `candidate_sp` or 1
+    sp_significant_ : list of int, seasonality periods with sub-threshold p-levels
```
fkiraly, line 430: is this correct, i.e., are they ordered by p-level? If not, the docstring needs to be changed.

sktime/param_est/seasonality.py
```diff
@@ -407,3 +407,110 @@ def get_test_params(cls, parameter_set="default"):
...
+        Disregard periods shorter than this number of samples.
+        Defaults to 4
+    max_period : int
+        Disregard periods longer than this number of samples.
+        Defaults to None
+
+    Attributes
+    ----------
+    sp_ : int, seasonality period at lowest p-level, if any sub-threshold, else 1
+        if `candidate_sp` is passed, will be in `candidate_sp` or 1
```
fkiraly, line 429: there is no argument `candidate_sp`, docstring needs to be adapted

blazingbhavneek:
@fkiraly @achieveordie Requested changes are addressed as far as I know. Am I missing something?

achieveordie (approved):

fkiraly (approved):
yes, all resolved from my side too. Nice contribution, thanks!

fkiraly:
Did you add yourself to the all contributors file, @blazingbhavneek? (if not, I suggest separate PR)

Issue #4220, opened 2023-02-12
[BUG] some released estimators do not construct, tests did not catch this


fkiraly:
A number of released estimators do not construct:

* `ComposableTimeSeriesForestRegressor`
* `TimeSeriesDBScan`

The reason in both cases are abstract methods expected by `ABC` that are not implemented. In both cases it seems that it should not be required to implement said methods, e.g., the parallel `ComposableTimeSeriesForestClassifier` just adds a dummy method that raises an error to keep `ABC` happy.

Also, this was not detected by the testing framework due to #4219.

PR #4227, opened 2023-02-13
[BUG] fix `TimeSeriesDBSCAN` and remove strict `BaseClusterer` abstracts

fkiraly:
Partial fix towards #4220.

Fixes `TimeSeriesDBSCAN`, and removes the requirement in `BaseClusterer` to implement abstract methods.


fkiraly:
Quick merge due to breaking bug in base class and interaction with dependency isolation

Issue #4013, opened 2022-12-29
[ENH] end-to-end `dask` integration


fkiraly:
This issue collects a roadmap for achieving end-to-end integration with `dask` data containers at the framework (not estimator) level. For discussion and design.

* `dask` based mtypes, especieally for panel and hierarchical data
* base classes, esp forecasters and transformers, recognize `dask` data frames and avoid calls to `compute`. Possibly:
    * option: pass to `VectorizeDF`, which recognizes `dask` and handles vectorization via `dask` `groupby`
    * option: if forecaster/transformer does not support `dask` natively, it is wrapped in a `delayed` call
    * option: if transformer is instance-wise, always uses `dask` `groupby` for vectorization
* `dask` support in key transformers: lag, windowing, aggregation/summarization, temporal features

topher-lo:
Another option (same spirit as option 1): pass to `VectorizeDF` and handle parallelisation via `dask` `map_partitions`. This is what nixtla's `mlforecast` does!

fkiraly:
> Another option (same spirit as option 1): pass to `VectorizeDF` and handle parallelisation via `dask` `map_partitions`. This is what nixtla's `mlforecast` does!

hm, indeed. Just from naive inspection, this seems to be two ways to achieve the same end.

Do you have a good grip on which of `groupby` vs `map_partitions` would be "better", for a yet-to-specify multi-faceted definition of "better"?

fkiraly:
@topher-lo, I would be keen to hear how you would design the dispatch inside `VectorizeDF` (if you have the time to think about it)

topher-lo:
Definitely `map_partitions`. Dask maintainers recommend using `map_partitions` AKA "map dataframe chunks" over groupby for general functions. It's generally a lot more performant as well: e.g. in this blog post (https://targomo.medium.com/how-we-learned-to-love-dask-and-achieved-a-40x-speedup-aa14e72d99c0).

topher-lo:
@fkiraly FYI this is how github.com/nixtla/mlforecast implemented a distributed time series data structure: https://github.com/Nixtla/mlforecast/blob/main/nbs/distributed.core.ipynb

PR #4228, opened 2023-02-13
[ENH] Improve vectorized metric calculation and deprecate VectorizedDF.__getitem__ and VectorizedDF.get_iloc_indexer

hoesler:
#### Reference Issues/PRs
Followup to #4195

This PR implements `BaseForecastingErrorMetric._evaluate_vectorized` using `VectorizedDF.vectorize_est`.
Further, by removing with it the last reference to `VectorizedDF.__getitem__`, it deprecates this method. Random access is not needed, and developers should use `__iter__` for iteration instead (implemented in #4195).
Also, unused method `get_iloc_indexer` is marked as deprecated for the same reason.

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/datatypes/_vectorize.py
+++ b/sktime/datatypes/_vectorize.py
@@ -48,6 +48,8 @@ class VectorizedDF:
         as pandas.DataFrame with Index or MultiIndex (in sktime pandas format)
     len(self) or self.__len__()
         returns number of Series/Panels in X
+    get_iter_indices()
+        Returns pandas.(Multi)Index that are iterated over
     reconstruct(self, df_list, convert_back=False)
         Takes iterable df_list and returns as an object of is_scitype.
         Used to obtain original format after applying operations to self iterated
@@ -185,7 +187,7 @@ class VectorizedDF:
 
     # TODO: remove in v0.18.0
     @deprecated(
-        version="0.16.1",
+        version="0.16.2",
         reason="get_iloc_indexer will be removed in v0.18.0",
         category=FutureWarning,
     )
@@ -230,7 +232,7 @@ class VectorizedDF:
 
     # TODO: remove in v0.18.0
     @deprecated(
-        version="0.16.1",
+        version="0.16.2",
         reason="__getitem__ will be removed in v0.18.0",
         category=FutureWarning,
     )
--- a/sktime/datatypes/_vectorize.py
+++ b/sktime/datatypes/_vectorize.py
@@ -8,6 +8,7 @@ import itertools
 
 import numpy as np
 import pandas as pd
+from deprecated.sphinx import deprecated
 
 from sktime.datatypes._check import check_is_scitype, mtype
 from sktime.datatypes._convert import convert_to
@@ -220,6 +221,16 @@ class VectorizedDF:
                 iterate_as=self.iterate_as, iterate_cols=self.iterate_cols
             )
         )
+
+    # TODO: remove in v0.18.0
+    @deprecated(
+        version="0.16.1",
+        reason="__getitem__ will be removed in v0.18.0",
+        category=FutureWarning,
+    )
+    def __getitem__(self, i: int):
+        """Return the i-th element iterated over in vectorization."""
+        return next(itertools.islice(self, i, None))
 
     def items(self, iterate_as=None, iterate_cols=None):
         """Iterate over (group name, column name, instance) tuples.
--- a/sktime/datatypes/_vectorize.py
+++ b/sktime/datatypes/_vectorize.py
@@ -183,6 +183,12 @@ class VectorizedDF:
         """
         return self.iter_indices
 
+    # TODO: remove in v0.18.0
+    @deprecated(
+        version="0.16.1",
+        reason="get_iloc_indexer will be removed in v0.18.0",
+        category=FutureWarning,
+    )
     def get_iloc_indexer(self, i: int):
         """Get iloc row/column indexer for i-th list element.
 
--- a/sktime/datatypes/_vectorize.py
+++ b/sktime/datatypes/_vectorize.py
@@ -230,12 +230,6 @@ class VectorizedDF:
             )
         )
 
-    # TODO: remove in v0.18.0
-    @deprecated(
-        version="0.16.2",
-        reason="__getitem__ will be removed in v0.18.0",
-        category=FutureWarning,
-    )
     def __getitem__(self, i: int):
         """Return the i-th element iterated over in vectorization."""
         return next(itertools.islice(self, i, None))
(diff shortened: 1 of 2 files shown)


fkiraly (changes requested):
I think it's an interesting question whether we should do this.
On one hand, good idea as a refactor step to move away from `__getitem__`, this would be the last step of your intended refactor, I suppose.
You have already isolated the less efficient interfaces and replaced them with the group based one.

On the other hand, I wonder whether partial or random access is sth we want later, if we add support for distributed computing, see, e.g., https://github.com/sktime/sktime/issues/4013

I would recommend to keep that in mind as the northstar for `VectorizedDF`, i.e., lazy computation which passes through distributed objects and inverts the application (estimator within/from `dask` frame rather than the other way round)

Do you have any thoughts on that?

hoesler:
I think it’s good to remove unused code in order to prevent maintenance overhead until you really know the requirements.

Having said that, you raise an interesting question. Actually, I don’t think there is good use for for random access from the perspective of a costumer of a distributed data object. Besides ordering guarantees in such systems might be absent or costly, you don’t want to fetch the data, as you said, but tell the backend do do something with it and return the result, which again will be distributed data at first. In the end, when small enough, you convert it into a pandas frame, list of model (references) etc. and fetch it (iteratively).
That’s how I would do it with spark (which has no random access method for rows. And from what I [read](https://docs.dask.org/en/stable/dataframe-indexing.html#positional-indexing), this also is true for dask).

fkiraly (changes requested):
Yes, I suppose you are right. `dask` would result in a delayed `groupby` object and a mapping applicate or similar, and we still would use sth similar to `items`.

So we are agreed about the end state, removing this.

Now, I recalled that `VectorizedDF` is public, so we should not make a change to its interfaces without deprecation. May I suggest the following alternative, hence?

* we patch `__getitem__` through to `items`, i.e., a full call and then the `i`-th element or similar
* at this opportunity, I would also suggest to change `items` so that the third tuple element (the subsetted dataframe) is an optional return, e.g., via an additional arg. Not necessary for this PR, but this would cover functionality that is removed and may be useful later.

fkiraly:
FYI, deprecation policy is here:
(affecting all public interfaces)

hoesler:
> Now, I recalled that VectorizedDF is public
Uh, thanks for clarifying. I didn't realize it is getting exported in __init__.py.

> we patch __getitem__ through to items, i.e., a full call and then the i-th element or similar
Although just a very temporary workaround, patching `__getitem__` through items would be very inefficient. I would use [multiindex sclicing](https://pandas.pydata.org/pandas-docs/stable/user_guide/advanced.html#advanced-indexing-with-hierarchical-index).

> at this opportunity, I would also suggest to change items so that the third tuple element (the subsetted dataframe) is an optional return, e.g., via an additional arg. Not necessary for this PR, but this would cover functionality that is removed and may be useful later.
I wouldn't squeeze to much use cases into one function. I'd rather add a `keys()` method.

fkiraly:
> Although just a very temporary workaround, patching `__getitem__` through items would be very inefficient. I would use [multiindex sclicing](https://pandas.pydata.org/pandas-docs/stable/user_guide/advanced.html#advanced-indexing-with-hierarchical-index).

I just thought it would be the quickest way to wrap this up, i.e., just call `items`.

>  I would use [multiindex sclicing](https://pandas.pydata.org/pandas-docs/stable/user_guide/advanced.html#advanced-indexing-with-hierarchical-index).

That's also fine with me.

> I wouldn't squeeze to much use cases into one function. I'd rather add a `keys()` method.

Excellent idea! I infer that you want to move this to a dictionary-like design. Also an excellent (design) idea!

hoesler:
> I just thought it would be the quickest way to wrap this up, i.e., just call items.
You were right. I finally decided a more sophisticated implementation is not worth the effort and delegating to `__iter__` is safer from a consistency perspective.

> Excellent idea! I infer that you want to move this to a dictionary-like design. Also an excellent (design) idea!
Yeah, sort of. But maybe it's indeed best postpone that, until there is really the need for such a method, now knowing that it is public API. What's the reason for that, btw?

fkiraly:
> (What's the reason for that, btw?).

Well, you need one of `__getitem__` or `__iter__` or `items` for the general design. It's just the way I designed it originally - acknowledging that yours is probably the better way to go.

hoesler:
I meant the latter. Thanks for the explanation, although I must admit that I don't understand the reasoning. From my gut feeling I would say fewer public classes of functions mean less to maintain and therefore more freedom to change things.

fkiraly:
> From my gut feeling I would say fewer public classes of functions mean less to maintain and therefore more freedom to change things.

Well, my gut feeling is about stability of interfaces. The base classes are quite important, and they depend on `VectorizedDF`. There are multiple base classes depending on it, so it's a bit like a utility package. Making it stable also makes it more reliable in constructing new base classes which haven't been written yet.

hoesler:
Ah, okay. Yeah, makes sense now.

fkiraly (changes requested):
Nice!

Could you kindly split out the work on the forecasting metrics in another PR? Same as before, keeping PR modular.
Also, good spot that we can use `VectorizedDF` there!

Regarding the deprecation, why are you suggesting to deprecate `__getitem__`? I'd say, let's keep it. `get_iloc_indexer` also doesn't hurt (may be useful if you have two vectorized objects).

hoesler:
> Could you kindly split out the work on the forecasting metrics in another PR? Same as before, keeping PR modular.
> Also, good spot that we can use VectorizedDF there!

Ok.

> Regarding the deprecation, why are you suggesting to deprecate __getitem__?
> I'd say, let's keep it. get_iloc_indexer also doesn't hurt (may be useful if you have two vectorized objects).

Fine for me. I was seeing this as a cleanup, with the reasoning I gave earlier: Removal of unused code to reduce maintenance overhead and a fear that people still might iterate in unintended ways. And I thought we agreed on this. But after rereading your message above, I noticed that I missed your advise against deprecation.

fkiraly:
> And I thought we agreed on this. But after rereading your message above, I noticed that I missed your advise against deprecation.

Sorry, I wasn't really clear. I would recommend we keep `__getitem__` and deprecate `get_iloc_indexer`.

fkiraly:
@hoesler, haven't heard back, so I hope you would be ok with the following solution here:

* I reverted the deprecation of `__getitem__`, but left the change in logic (to `islice`)
* anything else is unchanged from your last state, i.e., we deprecate `get_iloc_indexer`
* this gets merged as-is, except for that minor change

fkiraly (approved):
see above - as-is, minus one deprecation

PR #3869, opened 2022-12-02
[ENH] multi-level column support for columnensemble

RikStarmans:
#### Reference Issues/PRs

Fixes #3804 

#### What does this implement/fix? Explain your changes.

Add supports for multi-level columns to ColumnEnsembleForecaster

#### What should a reviewer concentrate their feedback on?

I added changes on three places; check forecasters which ensures all columns are forecasted, coerce_to_pd_index and the way forecasts are combined.
Changes are minimal but I guess it can still be made better. This is mainly to show what is needed.

##### For all contributions
- [ ] I've added myself to the [list of contributors](https://github.com/sktime/sktime/blob/main/.all-contributorsrc).
- [ ] Optionally, I've updated sktime's [CODEOWNERS](https://github.com/sktime/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
- [ ] I've added unit tests and made sure they pass locally.
- [x] The PR title starts with either [ENH], [MNT], [DOC], or [BUG] indicating whether the PR topic is related to enhancement, maintenance, documentation, or bug.

##### For new estimators
- [ ] I've added the estimator to the online documentation.
- [ ] I've updated the existing example notebooks or provided a new one to showcase how my estimator works.

PR #4233, opened 2023-02-14
DRAFT: & RikStarmans - multi-level column support for ColumnEnsembleForecaster

fkiraly:
This naively translates @RikStarmans' PR https://github.com/sktime/sktime/pull/3869 to the state post refactor https://github.com/sktime/sktime/pull/4231.

Currently, this will break just as #3869 does - but I'm refactoring the PR with #4231 so we don't run into merge conflicts later when we fix @RikStarmans' contribution.

(quick question, @RikStarmans - are you still working on this and/or intend to fix #3869?)

### DIFF
--- a/sktime/forecasting/compose/_column_ensemble.py
+++ b/sktime/forecasting/compose/_column_ensemble.py
@@ -151,6 +151,10 @@ class ColumnEnsembleForecaster(_HeterogenousEnsembleForecaster):
 
     def _coerce_to_pd_index(self, obj):
         """Coerce obj to pandas Index."""
+        # multi-index kept unchanged
+        if isinstance(obj, (list, tuple)):
+            return obj
+
         # replace ints by column names
         obj = self._get_indices(self._y, obj)
 
@@ -241,6 +245,9 @@ class ColumnEnsembleForecaster(_HeterogenousEnsembleForecaster):
         keys = self._get_indices(self._y, keys)
 
         if col_multiindex:
+            y_pred = pd.concat(y_preds, axis=1, keys=keys)
+        elif isinstance(keys[0], (list, tuple)):
+            keys = [tuple(key) for key in keys]
             y_pred = pd.concat(y_preds, axis=1, keys=keys)
         else:
             y_pred = pd.concat(y_preds, axis=1)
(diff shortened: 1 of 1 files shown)


fkiraly:
ah, thanks for letting me know, @hstarmans.

I think GitHub support can recover your account, if you had 2FA enabled?

fkiraly:
just noticed I had not committed the move properly, this PR should now contain the changes translated to the post-refactor state correctly.

hstarmans:
The problem is not with 2FA. The client I was working for cut of my account in December.
Currently, I don't have an employer which is able to make money with your code.
As such, my attention drifted back to laser scanning. I spent the last month building a PCB motor.
https://hackaday.com/2023/02/22/laser-scanner-upgraded-to-use-pcb-motor/

I did consider using your code for trading on the market. If you can predict the future, you can use techniques like https://pypi.org/project/pyportfolioopt/, to calculate the optimal allocation.
The main issue you r still left with is slippage, i.e. transaction costs.
I would be happy if my tool could outperform a simple investment in bonds.
This is however quite complicated to embark upon on my own.

I really enjoyed working together on this but if ideally, I am also able to monetize it.

fkiraly:
I see, sorry to hear, @hstarmans.
Your code is not lost, you can clone this branch and work on it if you want!

Interesting pointer to `pyportfolioopt`, that package links together closely with `sktime`.
Do their models not incorporate transaction costs?

hstarmans:
There are several tools involved;
 - you need a forecast this could be done with SKtime
 - you need a strategy given these forecasts, e.g. you distribute over multiple assets use an efficient portfolio (Markowitz 1952) . For this you can use  stuff like pyportfolio
 - you need to backtest your result, account for slippage
 
A library working on this is https://github.com/edtechre/pybroker 

I would be interested in seeing how a simple forecast using arimax, exponential time smoothing and allocation over the main indices SP500 etc. performs.

Currently, I use your library for fmcg companies. I actually talked to a medium size dutch supermarket yesterday. They had worked with gfk and even signs of that company but they were unfamiliar with sktime (spoken to their forecasting unit ) 😀 and was surprised

fkiraly:
> They had worked with gfk and even signs of that company

sktime is not a gfk project, and I'm not contributing in my gfk role - sktime is an open, community governed project and I'm contributing in my free time.
I'm also not using sktime in my day job, this has to do with other people (not me) making the actual decisions about python code, and they prefer building most things from scratch as part of a huge legacy codebase.

It's a big ironic, given how many users sktime has, including in the market forecasting space, slow and fast-moving, but such is life.

This is now moving a bit off-topic, but do you think it would make sense integrating the trading and portfolio stack more closely, e.g., the three packages? sktime has a vision of integrating open souce communities, and we're moving in that direction by making framework code available as stand-alone: https://github.com/sktime/skbase

PR #4248, opened 2023-02-19
[ENH] conditional transform

fkiraly:
See https://github.com/sktime/sktime/issues/4249

This adds a transformer compositor which executes a transform conditionally, e.g., deseasonalizes if seasonality is detected, or differences if stationarity is detected. The condition can be specified flexibly based on any estimator (with a `fit` interface), and the if/else transformation can also be configured.

Relies on https://github.com/sktime/sktime/pull/4360 for the seasonality example.


fkiraly:
Note to self: this works, but the test cases fail with panel data that cause the `BoxCoxTransformer` to vectorize and not have a `lambda` parameter

Issue #4266, opened 2023-02-23
[BUG] "PerformanceWarning: DataFrame is highly fragmented." in ShapeDTW with GridSearchCV


giordanoscerra:
PerformanceWarning: DataFrame is highly fragmented.  This is usually the result of calling `frame.insert` many times, which has poor performance.  Consider joining all columns at once using pd.concat(axis=1) instead. To get a de-fragmented frame, use `newframe = frame.copy()`
  df[i] = data
  
when i try to use shapeDTW with gridsearchCV and i have a very big dataset of shape (198,7)

```python
from sklearn.model_selection import KFold
from sklearn.model_selection import GridSearchCV
from importlib import import_module as imp
import numpy as np

X_frag = [[0, 1, 2, 3, 5, 6, 8], [0, 1, 2, 3, 5, 6, 9], [0, 1, 2, 3, 5, 6, 7], [0, 1, 2, 3, 5, 6, 7], [0, 1, 2, 3, 5, 6, 7]]
y_frag = [0, 0, 1, 1, 1]
X = []
y = []

for i in range(198):
    X.extend(X_frag)
    y.extend(y_frag)
    print(i)


X_arr = np.array(X)
y_arr = np.array(y)


shapedtw_type = 'sktime.classification.distance_based._shape_dtw.ShapeDTW'

module_name, class_name = shapedtw_type.rsplit(".", 1)
model = getattr(imp(module_name), class_name)()

param_grid = {'shape_descriptor_function': ['raw', 'paa'] }
parameter_tuning_method = GridSearchCV(model, param_grid, cv=KFold(n_splits=5), return_train_score=True, verbose=3)
parameter_tuning_method.fit(X_arr, y_arr)

#print(parameter_tuning_method.best_score_)
print(parameter_tuning_method.cv_results_)
```

fkiraly:
I can confirm the issue on python 3.10, windows.

The reason seems to be use of an old data representation internally, where `pandas` objects are nested. This should be replaced by flat pandas objects or ND numpy.

An attempt at this is here: https://github.com/sktime/sktime/pull/4267

fkiraly:
@giordanoscerra, I think I fixed the issue, but runtime seems to be affected slightly adversely. I would appreciate if you could test #4267 on your "real" use case and stop the time, so we can see whether it is better or worse.

fkiraly:
code adapted to the use case from discord - numpy format

```python
from sklearn.model_selection import KFold
[...]

PR #4267, opened 2023-02-23
[ENH] refactoring segmentation transformers to use `pandas` native data types

fkiraly:
This refactors three segmentation transformers, `RandomIntervalSegmenter`, `SlidingWindowSegmenter` and `IntervalSegmenter` to the `pandas` native mtype, `pd-multiindex`, from the deprecated mtype `nested_univ` which had nested pandas objects.

As a consequence, this also fixes https://github.com/sktime/sktime/issues/4266 for which the use of the nested data type seems to have been the cause.

This seems to cause a substantial improvement in runtime too (ca 25% reduction in the example in #4266), we may like to benchmark.

Relies on bugfix https://github.com/sktime/sktime/pull/4270 for an unreported bug was discovered through this.

### DIFF
--- a/sktime/datatypes/_utilities.py
+++ b/sktime/datatypes/_utilities.py
@@ -43,6 +43,9 @@ def get_time_index(X):
         # nested_univ
         elif isinstance(X, pd.DataFrame) and isinstance(X.iloc[0, 0], pd.DataFrame):
             return _get_index(X.iloc[0, 0])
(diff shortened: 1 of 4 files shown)


fkiraly:
update: has side effects on `RandomIntervalSegmenter` which was inheriting from `IntervalSegmenter` - need to fix that too

Issue #4262, opened 2023-02-23
[DOC]Column Concatenator Provides insufficient examples in documentation


JonathanBechtel:
#### Describe the issue linked to the documentation
The documentation for the column concatenator does not provide any code example about how to use it, and lacks a little bit of detail about how its transformations are made

There's no source code provided about how to use it.  Even though it's mechanism is simple, it still makes sense to include starter code for others to see if they're new to the library.  Being able to quickly observe how the dimensions of your data change going into and out of an operation is useful for people who are working with sktime for the first time.

#### Suggest a potential alternative/fix

I can add a simple example to the docstring demonstrating its use, as well as a bit more detail about the order series are concatenated in.

fkiraly:
> I can add a simple example to the docstring demonstrating its use, as well as a bit more detail about the order series are concatenated in.

Excellent idea - this has come up frequently and is quite valuable to do, since volunteer contributors don't like to work on documentation. Much much appreciated!

I've opened an umbrella issue:
https://github.com/sktime/sktime/issues/4264

fkiraly:
resolved by a generous soul in #4272

Issue #4264, opened 2023-02-23
[ENH] add at least one example to docstrings of estimators


fkiraly:
Each estimator should have an example in its docstring.

This example should cover basic usage, and possiblly advanced cases.

The recipe for adding examples is:

* add an example section in `numpydoc` stype
* use as template for a simple example: `NaiveForecaster`
* use as template for complex/multiple examples: `ForecastingGridSearchCV` (search in code base)
* if the estimator relies on a soft dependency, ensure to use doctest skip to avoid import if the dependency is not present. Use as template for an example with soft dependency: `ARIMA`

KishManani:
Hi @fkiraly! Happy to help adding some/more examples in the doc strings of estimators. Do you have a list already of some of the likely most used estimators that could benefit with an example?

fkiraly:
There are currently 90 objects in `sktime` without example, code to generate them:

```python
from sktime.registry import all_estimators

ests = all_estimators()
[x[1] for x in ests if "Example" not in x[1].__doc__ ]
```
(to subset to types, use the `scitype` arg of `all_estimators`)

JonathanBechtel:
@fkiraly After PR #4272 gets accepted I'll work on one for `TimeSeriesForestRegressor` and `ForecastingRandomizedSearchCV`

Verogli:
I am adding examples to docstrings for DBScan.

fkiraly:
FYI @Verogli, there's an example for `DBScan` in the EuroPython tutorial - this notebook: https://github.com/sktime/sktime-tutorial-europython-2023/blob/main/notebooks/05_panel_tasks.ipynb

although the result isn't "nice" in the sense of the clulster split being very clear.

Verogli:
Thank you. I will look into that.

meraldoantonio:
Hi! It's my first time contributing. I plan to add an example for `CNNRegressor` following the example from `KNeighborsTimeSeriesRegressor`. I imagine I'll need to include doctest skip for this?

fkiraly:
Nice! Yes, the doctest should be skipped due to the soft dependency (alternatively, you can make the execution condtional on the soft dependency being present, that's a newer feature in `doctest`).

YashKhare20:
Hi all! It's my first time contributing. I've added an example for DummyClassifier following the example from DummyRegressor. I've included doctest skip due to the soft dependency, hope that is fine.

fkiraly:
[...]

PR #110, opened 2019-08-01
Consistent input validation

mloning:
#### Reference Issues/PRs
Fixes #98

#### What does this implement/fix? Explain your changes.
* consistent input checks

#### Any other comments?
* different input checks for forecasting?

Issue #104, opened 2019-07-31
pmdarima interface comparison


mloning:
**Data container**
* pmdarima assumes for y (target) array of shape=(n_obs,) and for X (exogeneous) array of shape=(n_obs, n_features), crucially with time series observations in rows. 
* by contrast, we have a nested pandas Series or DataFrame, with rows representing iid samples, i.e. in the classical forecasting setting a single row, and columns representing different kinds of measurements (i.e. features), the format is nested because each cell may contain no longer only primitive types but also arrays or pandas Series of shape=(n_obs,), 
* we chose this data container because it allows to have a unified interface for different time series related learning settings including forecasting, but also panel/longitunidal data settings such as time series classification/regression or supervised/panel forecasting.

**Forecasting horizon**
* we allow the forecasting horizon to be an array with the steps ahead, not only the number of steps ahead (`n_periods`), allowing for gaps e.g. predicting only the third and fifth step ahead with `fh = np.array([3, 5])`, may not make a difference for ARIMA, but will do for reduction strategies

**Fit method**
* some methods already require the forecasting horizon in `fit` (e.g. non-dynamic reduction to time series regression where one time series regressor is fitted for each step of the forcasting horizon)
* pmdarima currently takes the forecasting horizon only in `predict`
* should we allow the forecasting horizon when passed in `fit` to be then changed in `predict`?

**Update method**
* pmdarima adds the new data to the old data (kept in the model) and refits the model using the previously fitted params as start params,
* is the new data always in the future?
* alternatively, one could use the new data to update the already fitted parameters using Kalman smoothing/filtering, ideally only the new data is used so that the data seen in training does not have to be stored in the model? 

**In-sample forecasts**
* pmdarima has a separate method for in-sample forecasts
* currently, we don't have a good way to do in-sample forecasts, our forecasting horizon is currently relative to the end of the series seen in training, allowing negative values would perhaps be one option?

**Forecast confidence intervals**
* pmdarima returns intervals from `predict` method via an optional kwarg changing the output of predict 
* should we return intervals from a separate method instead to have a stable method signature for `predict`?
* related to design discussion on probabilistic forecasting #97 

**Composition**
* Pipelines for target variable vs `TransformedTargetForecaster`

tgsmith61591:
Great discussion points, definitely some we need to consider on pmdarima side to unify interfaces, though I'm not sure how we'd handle paneled data..

[...]

PR #4272, opened 2023-02-25
[DOC] Update Docstring for ColumnConcatenator

JonathanBechtel:
#### Reference Issues/PRs
References PR #4262

This PR adds an updated docstring that provides an example for how to use it, as well as a slightly more detailed explanation about its use case.

No

Making sure the new example is adequate to provide context about how to use the tool

Being that this is just an update to a DOC string, there is no need to add a unit test

#### Any other comments?
I think more of these PR's should be created because it often difficult to see where to import different tools from.  It was too difficult to figure out how to use such a simple piece of the program.

##### For all contributions
- [ X] I've added myself to the [list of contributors](https://github.com/sktime/sktime/blob/main/.all-contributorsrc).
- [X ] The PR title starts with either [ENH], [MNT], [DOC], or [BUG] indicating whether the PR topic is related to enhancement, maintenance, documentation, or bug.

### DIFF
--- a/sktime/transformations/panel/compose.py
+++ b/sktime/transformations/panel/compose.py
@@ -215,8 +215,29 @@ class ColumnConcatenator(BaseTransformer):
     """Concatenate multivariate series to a long univariate series.
 
     Transformer that concatenates multivariate time series/panel data
-    into long univariate time series/panel
-        data by simply concatenating times series in time.
+    into single univariate time series/panel data by concatenating
+    each individual series on top of each other from left to right.
+
+    Uses pandas method stack() to do the concatenating
+
+    Examples
+    --------
+    >>> from sktime.transformations.panel.compose # doctest: +SKIP
+    >>> import numpy as np # doctest: +SKIP
+    >>> data = data = np.array([[1, 2, 3], # doctest: +SKIP
+    ...                         [4, 5, 6], # doctest: +SKIP
+    ...                         [7, 8, 9]]) # doctest: +SKIP
+    >>> concatenator = ColumnConcatenator() # doctest: +SKIP
+    >>> concatenator.fit_transform(data) # doctest: +SKIP
+    >>> array([[1.],  # doctest: +SKIP
+    ...        [4.],  # doctest: +SKIP
+    ...        [7.],  # doctest: +SKIP
+    ...        [2.],  # doctest: +SKIP
+    ...        [5.],  # doctest: +SKIP
+    ...        [8.],  # doctest: +SKIP
+    ...        [3.],  # doctest: +SKIP
+    ...        [6.],  # doctest: +SKIP
+    ...        [9.]]) # doctest: +SKIP
     """
 
     _tags = {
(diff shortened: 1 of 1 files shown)


fkiraly (changes requested):
Thanks!

One minor thing, I don't think we need to (or should) skip the doctest lines. This is just in a case where we have soft dependencies.

In doctest, the output also does not need to be preceded by `>>>` (or it will be considered an input). So, you may want to remove this in front of the array at the end.

Further, would using an example panel dataset, e.g., using `_make_panel` from `sktime.utils._testing.panel` (with `n_columns=2`) also be illustrative?

Regarding "making more of these" - absolutely! Much appreciated!
Here's the umbrella issue: https://github.com/sktime/sktime/issues/4264

JonathanBechtel:
All of those changes are fine, and I can update them, with one exception:  `from sktime.transformations.panel.compose import ColumnConcatenator` takes up 92 lines so it fails the flake8 check.  So I think that exception has to stay.  Definitely think the panel dataset would be a good example as well, so will add it.

fkiraly:
> So I think that exception has to stay. Definitely think the panel dataset would be a good example as well, so will add it.

Sure - the `noqa` exception is fine, that doesn't skip it in the doctest.

An alternative, though slightly more ugly for the user (and I prefer your workaround, but just FYI) is having an import with brackets and newline e.g.,

```python
from longmodulename.longmodulename.longmodulename (
    this_is_what_we_import
)
```

JonathanBechtel:
Made the changes.  In the future, is it okay to just reference the umbrella issue when a new PR is opened and skip making it a separate issue?

fkiraly (changes requested):
Thanks!

This is failing due to conventions of doctests.

In doctest:
* `>>>`  is start of an expression
* `... ` is used when the same expression goes over multiple rows. One `>>> ` and all subsequent `... ` are interpreted as one block of code.
* if you want "expeted output", it should start at the line without `...`.

That is, the `...` in lines 228, 229 are correct - in lines 232-240 you probably don't want `...`

RNKuhns:
@JonathanBechtel you can also add comments in between examples (have to have blank lines around those lines).

To make it easy check out the example section in numpydoc’s [docs](https://numpydoc.readthedocs.io/en/latest/format.html) and examples in [skbase](https://github.com/sktime/skbase/blob/main/skbase/validate/_named_objects.py).

JonathanBechtel:
> @JonathanBechtel you can also add comments in between examples (have to have blank lines around those lines).
> 
> To make it easy check out the example section in numpydoc’s [docs](https://numpydoc.readthedocs.io/en/latest/format.html) and examples in [skbase](https://github.com/sktime/skbase/blob/main/skbase/validate/_named_objects.py).

@RNKuhns To be clear, lines #96, #104, and #110 in [skbase] are examples of what you're describing, yes?

JonathanBechtel:
@fkiraly Any idea why some small changes to a docstring would cause compatibility failures for unix operating systems?

fkiraly (approved):
Great, thanks!

PR #3151, opened 2022-07-31
 `skbase` refactor - part 1: `BaseObject` and package dependencies

fkiraly:
First PR in a series of refactor branch to move `sktime` base functionality to a separate package, `skbase`.

Part 1:
Moves `BaseObject` and core `base` functionality to the `skbase` package.

Relies on:
* the `skbase` release which it introduces as a core dependency
* `sktime` sided tests refactor https://github.com/alan-turing-institute/sktime/pull/3122 (this should be merged first)
* requires https://github.com/alan-turing-institute/sktime/pull/3123 to fully decouple `BaseObject` from `_config` test params

Migrated:
* `BaseObject`. The `sktime` `BaseObject` is a child of `skbase` `BaseObject` with some added features that are currently `sktime` specific.

further parts will refactor:
* `all_estimators`
* `TestAllObjects`
* add tests

PR #4283, opened 2023-03-04
Experimental - DO NOT MERGE: `BaseObject` refactor - part 3: test framework

fkiraly:
Experimental refactor branch.

Not for merge until mature, tests are in place, and based on current `skbase`.

Moves `BaseObject` and core `base` functionality to the `baseobject` package.

Part 3, relies on:

* part 1: https://github.com/sktime/sktime/pull/3151
* part 2: https://github.com/sktime/sktime/pull/3777

Migrated:
* `BaseFixtureGenerator`, `QuickTester`, and `TestAllObjects`, uses `skbase.testing` under the hood.
* from part 2 - `all_estimators` - uses `skbase.all_objects` under the hood now.
* from part 1 - `BaseObject`. The `sktime` `BaseObject` is a child of `skbase` `BaseObject` with some added features that are currently `sktime` specific.

This should be the last step of the refactor/migration.


fkiraly:
@RNKuhns, I think I've refactored the `sktime` test framework correctly, but there's now a highly mysterious bug appearing.

The tests seem to be run as expected, but one of the tests - multiprocessing identity (check results are the same with and without multiprocessing) - fails with an odd error, where out of 10 entries in the result a few (four) are by a constant of 1 off.

I cannot explain this with missing excluded tests etc, especially since it's coming from `TestAllEstimators` rather than `TestAllObjects`.

PR #4308, opened 2023-03-09
[BUG] Fix issue #4299; add example DataFrame with different starting times …

### LINKED ISSUE #4299: [BUG] check_pdmultiindex_panel incorrect result for metadata is_equally_spaced
**Describe the bug**

The *check_pdmultiindex_panel* method from sktime.datatypes will return an incorrect result for the "is_equally_spaced" metadata when the time index for each instance has a non-sequential initial time.

**To Reproduce**

```python
import pandas as pd

example_instance_index = [0, 0, 0, 1, 1, 1, 2, 2, 2]
example_time_index = pd.to_datetime([0, 1, 2, 4, 5, 6, 9, 10, 11], unit='s')
example_multiindex = pd.MultiIndex.from_arrays([example_instance_index, example_time_index])

example_data = pd.DataFrame(index=example_multiindex, data=list(range(0, 9)))
"""example_data
                       0
0 1970-01-01 00:00:00  0
  1970-01-01 00:00:01  1
  1970-01-01 00:00:02  2
1 1970-01-01 00:00:04  3
  1970-01-01 00:00:05  4
  1970-01-01 00:00:06  5
2 1970-01-01 00:00:09  6
  1970-01-01 00:00:10  7
  1970-01-01 00:00:11  8
"""

_, _, metadata = check_is_mtype(example_data, mtype="pd-multiindex", return_metadata=True)

assert metadata['is_equally_spaced'] == True
```

**Expected behavior**

The method should return True for "is_equally_spaced", since for each separate instance in the DataFrame the spacing between time indexes is 1 second.

**Additional context**

In the check_pdmultiindex_panel method (sktime/datatypes/_panel/_check.py) the is_equally_spaced metadata is computed via
```python
metadata["is_equally_spaced"] = all(
            _index_equally_spaced(group.index.levels[-1]) for _, group in series_groups
        )
```
If the get_level_values() method is used instead of the levels attribute, this bug should be fixed.
```python
metadata["is_equally_spaced"] = all(
            _index_equally_spaced(group.index.get_level_values(-1)) for _, group in series_groups
        )
```

**Versions**
<details>

System:
    python: 3.9.15 | packaged by conda-forge | (main, Nov 22 2022, 15:55:03)  [GCC 10.4.0]
executable: /home/bin/python
   machine: Linux-4.14.305-10ms-x86_64-with-glibc2.27

Python dependencies:
          pip: 23.0.1
   setuptools: 58.1.0
      sklearn: 1.2.1
       sktime: 0.16.1
  statsmodels: 0.13.5
        numpy: 1.23.5
        scipy: 1.10.1
       pandas: 1.5.3
   matplotlib: None
       joblib: 1.2.0
        numba: 0.56.4
     pmdarima: None
      tsfresh: 0.20.0
</details>

noahleegithub:
…per instance; change is_equally_spaced check from using levels attribute to get_level_values function

#### Reference Issues/PRs

Fixes #4299 

Previously `check_pdmultiindex_panel` used the `levels` attribute of the index to determine whether the panel was equally spaced. This is a bug because the `levels` attribute references the full DataFrame's index when used in a groupby. The fix changes the `levels` attribute to the `get_level_values` method, which works correctly for groupby groups.

No

Reviewers might focus on the newly added example to test the fixed behavior.

Yes, a new fixture example in `sktime/datatypes/_panel/_examples.py` to specifically target the behavior.

##### For all contributions


fkiraly (changes requested):
Thanks!

Also, great, the examples file where you added it is the correct file.

It seems unfortunately that extending the coverage of fixtures has caused someo ther test failures:

* `get_time_index` assumes equal index, and currently examples are not annotated with that property. To address that, I've introduced that here: https://github.com/sktime/sktime/pull/4312 just for the purpose of skipping in the test. Later, it could be inferred in checks.
* There is an issue with `freq` (the 4 remaining tests), as the cutoff getter should retrieve a `freq` attribute. I am not sure whether this is a problem with the test, with the cutoff getting logic, or sth else. I will have a look at this, if you have good idea, that would be appreciated.

fkiraly:
would you mind if I merge #4312 into this to address the error?
Of course, review of #4312 would be appreciated.

There is probably a wider problem around unclear assumptions when it comes to series that are equally spaced but unequal index sets.

fkiraly:
@noahleegithub, I've understood the problem with the remaining failures that are not covered by #4312.

The issue is that the `freq` attribute (frequency) is not correctly detected in this combination case, i.e., unequal index set but equally spaced with equal frequency across instances.

I've described the issue here in more detail:
https://github.com/sktime/sktime/issues/4315
this may be a `pandas` `MultiIndex` issue, your eyes on it would be appreciated.

For now, I've pushed an exception for the failing tests, kindly check - and feel free to revert if you think this is not the right thing to do. With that, all tests should pass.

PR #4312, opened 2023-03-09
[ENH] `is_equal_index` metadata element in checks and examples

fkiraly:
This PR adds an `is_equal_index` metadata element in checks and examples, motivated by #4299 and the test failure in #4308. This tags examples of panel data where the index is not equal across instances.

This PR does not introduce metadata inference for that property, it only enables excluding examples in tests of functions that make assumption of equal index, e.g., `get_time_index`.

Also corrects the docstring of `test_get_time_index`.


fkiraly:
reviewed by @noahleegithub, and indirectly via #4308

PR #4316, opened 2023-03-12
[ENH] remove test for output values in `test_FeatureUnion_pipeline`

fkiraly:
This removes one test for output values in `test_FeatureUnion_pipeline`, as this is sporadically no longer satisfied.
This is due to the refactor https://github.com/sktime/sktime/pull/4267 changing how random seeds map onto estimator states, and the test in question only testing validity of the pipeline.


fkiraly:
quick merge, since the issue is on main

Issue #4182, opened 2023-01-31
[BUG] running pre-commit hook on all files raises errors on main branch


hoesler:
**Describe the bug**
Running pre commit hook on all files in a fresh checkout of the main branch raises errors.

**To Reproduce**

```shell
pre-commit run -a

Check for added large files..............................................[42mPassed[m
Check for case conflicts.................................................[42mPassed[m
Check for merge conflicts................................................[42mPassed[m
Check for broken symlinks................................................[42mPassed[m
Check Yaml...............................................................[42mPassed[m
Debug Statements (Python)................................................[42mPassed[m
Fix End of Files.........................................................[41mFailed[m
[2m- hook id: end-of-file-fixer[m
[2m- exit code: 1[m
[2m- files were modified by this hook[m

Fixing sktime/datasets/data/GunPoint/GunPoint_TRAIN.arff

Fix python encoding pragma...............................................[42mPassed[m
Fix requirements.txt.................................(no files to check)[46;30mSkipped[m
Trim Trailing Whitespace.................................................[41mFailed[m
[2m- hook id: trailing-whitespace[m
[2m- exit code: 1[m
[2m- files were modified by this hook[m

Fixing .github/ISSUE_TEMPLATE/bug_report.md
Fixing sktime/datasets/data/ACSF1/ACSF1_TEST.ts
Fixing sktime/datasets/data/ACSF1/ACSF1_TRAIN.ts

isort....................................................................[41mFailed[m
[2m- hook id: isort[m
[2m- files were modified by this hook[m

Fixing /Users/choesler/Develop/vendor/alan-turing-institute/sktime/sktime/transformations/panel/tests/test_dwt.py
Fixing /Users/choesler/Develop/vendor/alan-turing-institute/sktime/sktime/transformations/panel/tests/test_hog1d_transformer.py
Fixing /Users/choesler/Develop/vendor/alan-turing-institute/sktime/sktime/forecasting/base/__init__.py
Fixing /Users/choesler/Develop/vendor/alan-turing-institute/sktime/sktime/transformations/panel/tests/test_ColumnConcatenator.py
Fixing /Users/choesler/Develop/vendor/alan-turing-institute/sktime/sktime/datatypes/_series/_mtypes.py
Fixing /Users/choesler/Develop/vendor/alan-turing-institute/sktime/sktime/transformations/panel/tests/test_slope_transformer.py
[...]

PR #4319, opened 2023-03-13
[MNT] Fix linting issues in `forecasting` module

SamiAlavi:
#### Reference Issues/PRs
#4182

- isort auto-fixes
- Fix flake8 issue (B028 No explicit stacklevel keyword argument found)
- Fix pydocstyle issues (D104, D100, D205, D400)

No

- isort auto-fixes
- flake8 issue (B028 No explicit stacklevel keyword argument found)
- pydocstyle issues (D104, D100, D205, D400)

No

#### Any other comments?
No

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/forecasting/base/__init__.py
+++ b/sktime/forecasting/base/__init__.py
@@ -8,5 +8,5 @@ __all__ = [
     "BaseForecaster",
 ]
 
-from sktime.forecasting.base._fh import ForecastingHorizon
 from sktime.forecasting.base._base import BaseForecaster
+from sktime.forecasting.base._fh import ForecastingHorizon
--- a/sktime/forecasting/ardl.py
+++ b/sktime/forecasting/ardl.py
@@ -278,7 +278,8 @@ class ARDL(_StatsModelsAdapter):
                 inner_order = 0
                 warnings.warn(
                     "X is none but the order for the exogenous variables was"
-                    " specified. Order was thus set to 0"
+                    " specified. Order was thus set to 0",
+                    stacklevel=2,
                 )
         else:
             if not isinstance(X, pd.DataFrame):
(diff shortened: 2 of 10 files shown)


sktime/forecasting/base/__init__.py
```diff
@@ -8,5 +8,5 @@
     "BaseForecaster",
 ]
 
-from sktime.forecasting.base._fh import ForecastingHorizon
 from sktime.forecasting.base._base import BaseForecaster
+from sktime.forecasting.base._fh import ForecastingHorizon
```
SamiAlavi, line 12: Is `Install and test / run-notebook-examples (pull_request)` and `Install and test / test-nosoftdeps (pull_request)` failing due to this moving of imports?

fkiraly (changes requested):
Hm, this is really strange - it would appear that change the sequence of imports has caused a circular import?

This should not happen, but probably is a bug that has gone undetected due to the wrong sequence of imports.

I think it's the line you pointed out, I would suggest to try tracking down why it is circular.

sktime/forecasting/base/__init__.py
fkiraly: yes, it seems somehow it is causing a circular import - the problem is not the linting change, the problem is that there is probably a bug which was masked by the wrong sequence in the first place.

fkiraly: attempt to resolve this:
https://github.com/sktime/sktime/pull/4329

seems to happen when we start at `conftest` - alternative hypothesis, any import of `test_all_estimators` will cause this

fkiraly: correction: it was the forecaster base module, it was importing from base, but was itself imported from base

fkiraly (approved):
It's fixed now, after the circular import is resolved.

Thanks for the linting!

PR #4320, opened 2023-03-13
[MNT] Fix linting issues in `performance_metrics` module

SamiAlavi:
#### Reference Issues/PRs
#4182

- isort auto-fixes
- Fix pydocstyle issues (D100, D400, D103)

No

- isort auto-fixes
- pydocstyle issues (D100, D400, D103)

No

#### Any other comments?
No

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/performance_metrics/tests/test_performance_metrics_forecasting.py
+++ b/sktime/performance_metrics/tests/test_performance_metrics_forecasting.py
@@ -4,50 +4,51 @@
 
 __author__ = ["Tomasz Chodakowski", "Ryan Kuhns"]
 
-import pytest
 import numpy as np
 import pandas as pd
+import pytest
 from pandas.api.types import is_numeric_dtype
-from sktime.utils._testing.series import _make_series
+
 from sktime.performance_metrics.forecasting import (
-    MeanAbsoluteScaledError,
-    MedianAbsoluteScaledError,
-    MeanSquaredScaledError,
-    MedianSquaredScaledError,
-    MeanAbsoluteError,
-    MeanSquaredError,
-    MedianAbsoluteError,
-    MedianSquaredError,
-    MeanAbsolutePercentageError,
-    MedianAbsolutePercentageError,
-    MeanSquaredPercentageError,
-    MedianSquaredPercentageError,
-    MeanRelativeAbsoluteError,
-    MedianRelativeAbsoluteError,
     GeometricMeanRelativeAbsoluteError,
     GeometricMeanRelativeSquaredError,
+    MeanAbsoluteError,
+    MeanAbsolutePercentageError,
+    MeanAbsoluteScaledError,
     MeanAsymmetricError,
+    MeanRelativeAbsoluteError,
+    MeanSquaredError,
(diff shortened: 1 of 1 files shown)


sktime/performance_metrics/tests/test_performance_metrics_forecasting.py
```diff
@@ -1,53 +1,55 @@
 #!/usr/bin/env python3 -u
 # -*- coding: utf-8 -*-
 # copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+"""Tests for performance metrics forecasting."""
 
 __author__ = ["Tomasz Chodakowski", "Ryan Kuhns"]
```
fkiraly, line 6: could you replace these by GitHub names? Ryan is @RNKuhns, Tomasz is @tch

fkiraly (changes requested):
Thanks!

Could you kindly replace the full names by GitHub names at the top of the module (otherwise the contributions are not auto-linked)

sktime/performance_metrics/tests/test_performance_metrics_forecasting.py
SamiAlavi: There are some more files in which full names are being used instead of GitHub names.

https://github.com/sktime/sktime/blob/47ab3242590364582c9b503b6c75334ce6c3e085/sktime/_contrib/distance_based/_proximity_forest.py#L14

I can also see some inconsistencies. Most of the files have the `__author__` variable type as `string[]`, but some have type `string`. 

Should a new issue be created for fixing both these issues?

fkiraly: that's a good idea!

I think both `str` and `List[str]` are fine for the `__author__` field, btw.

SamiAlavi: The official style guide [PEP 8](https://peps.python.org/pep-0008/#module-level-dunder-names) provides an example with only `one` author. However, web archive's page [Python Coding Guidelines 12/14/07](http://web.archive.org/web/20111010053227/http://jaynes.colorado.edu/PythonGuidelines.html#module_formatting) lists multiple authors as comma separated values in `__author__` with type `str`.

I think this is left to whatever the organization/team prefers. So, I could transform all `__author__` values to type `List[str]` as I believe it looks cleaner than a comma-separated `str`.

fkiraly: ok, that makes sense.
That would be consistent with the extension templates (in the `extension_templates` folder), which has the line

```python
# __author__ = [authorGitHubID, anotherAuthorGitHubID]
```

fkiraly: I would also say, for a single author, `str` is also fine

fkiraly (approved):
agreed with suggestion, GitHub name replacement is scope creep.

Could you kindly open an issue to track that?

SamiAlavi:
Created issue #4335 to track updating author names

PR #4324, opened 2023-03-14
[DOC] updated `dtw_distance` docstring example to include import

JonathanBechtel:
#### Reference Issues/PRs
Adds to #4264

I updated the `dtw_distance` example to include the actual import.  This way someone can copy and paste the example given on the website and re-use it.

No

Make sure the example in the docstring can be re-used if copied and pasted directly into a python file.

Since it's just an example I didn't add a test.

#### Any other comments?
This fixes a small pet peeve of mine with the current documentation, which is that the actual import for the technique is often not listed in the example, so you can't just copy and paste from the API reference to see how it works.  This inconvenience is compounded by the fact that the actual imports you need to use are often irregular and can't be easily inferred from the way different methods are listed in the API docs.  So you often have to go on a bit of a chase just to figure out how to import something.  

I know a lot of the distance / kernel calculations do NOT include the import, which makes it unnecessarily difficult for someone new to the library to just grab an example and go.  

This might need to be a separate issue from #4264 since it's enhancing existing documentation, but for now I put it under that.  

It'd also be relatively simple to fix this, since most of the distance and kernel calculations are contained within a few files and you could add the imports fairly quickly. 

It might also be worth nothing that a potential 3rd example could be used for time series examples with different lengths.  Right now all of the examples only use equilength time series, but it might be instructive to demonstrate how calculations like dtw can be used on time series with dissimilar shapes, since it might not be obvious to people that this is possible.  For instance, with the 2D example you can transpose one of the time series and still get a result.

##### For all contributions

### DIFF
--- a/sktime/distances/_distance.py
+++ b/sktime/distances/_distance.py
@@ -713,6 +713,7 @@ def dtw_distance(
     Examples
     --------
     >>> import numpy as np
+    >>> from sktime.distances import dtw_distance
     >>> x_1d = np.array([1, 2, 3, 4])  # 1d array
     >>> y_1d = np.array([5, 6, 7, 8])  # 1d array
     >>> dtw_distance(x_1d, y_1d)


JonathanBechtel:
I realized I made a mistake when I deleted `compose.py`.  I was thinking that I'd remove it from the PR, not remove it from the actual codebase.  Not sure if it's easier for the reviewer to reinstate it or for me just to create an entirely new PR without the faux pas.

fkiraly:
The best way would be to revert your latest commit which deletes the file.

Second best way: since PRs are squashed to a single commit to main, you can simply copy the file from main and commit it, but this is quite hacky.

JonathanBechtel:
Ok, fixed.

fkiraly (approved):
makes sense, thanks.

I wonder how this could be tested in general, i.e., missing import in a doctest. It seems the test will pass since the module imports are present, but the user will miss it so the code will fail.

JonathanBechtel:
Agreed.  More of a usability issue than a code issue, but I see these omissions a lot throughout sktime documentation and it does throw sand in the gears for getting started.  

Not sure if it's worthwhile to dig deep into the codebase for a cosmetic change like this......but long term it *does* seem like a good way to go about checking for it.

fkiraly:
FYI, have you seen the issue I opened?
https://github.com/sktime/sktime/issues/4332

PR #4329, opened 2023-03-14
[MNT] resolve circular imports in `forecasting.base`

fkiraly:
PR https://github.com/sktime/sktime/pull/4319 has highlighted a potential issue with a circular import in `forecasting.base`.

`forecasting.base` was importing `BaseForecaster` from `forecasting.base._base`, which was importing `ForecastingHorizon` from `forecasting.base`, resulting in a circular import. Fixed by importing `ForecastingHorizon` from its direct location.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -13,8 +13,6 @@ by default, this is off, including for default local runs of pytest
 
 __author__ = ["fkiraly"]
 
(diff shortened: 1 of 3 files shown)


fkiraly:
from https://github.com/sktime/sktime/pull/4319

Issue #4332, opened 2023-03-15
[ENH] discoverable and searchable datasets with taxonomy


fkiraly:
Following ideas of @JonathanBechtel in #4314, it might be good to make datasets discoverable by tags, e.g., is it classification or forecasting, how many classes, instances, etc.

Here's an overview table that @JonathanBechtel suggested in #4314:
![image](https://user-images.githubusercontent.com/7985502/225168189-9e983470-0947-4f90-abf5-4719c25b0a44.png)

Citing @JonathanBechtel:

> Yes, correct.  I'll probably create a new issue for this broader agenda to touch up the datasets section, once it's specced out a bit more.  There are a number of tweaks that could be made in order to make it more usable.  

>In addition to what you mentioned, here are some other items to think about:
> - Some of the datasets return data in an inconsistent format:  some datasets return `(y, X)`, and others return `(X, y)`
> - Some have a `split` argument, but I'm not sure why others wouldn't.  
> - In general the top level documentation is a little difficult to parse.  I found myself having a hard time getting around to the "right" dataset to use for an example, only to find that it didn't exist.  
> - The UCR_UEA dataset isn't actually a dataset at all, but a loader for any number of datasets from their repository, so it's mistitled in my opinion, and should be changed to `fetch_ucr_uea_dataset` or something similar to denote that it's not a pre-installed dataset but something that actively downloads new data.

> Regarding taxonomy, I think a simple improvement would be to have something on the main datasets page akin to this document:  https://docs.google.com/spreadsheets/d/1SNbdgWUoot1tSnuKlxbhKvalYti7L1Z73hnUPXAaZdg/edit?usp=sharing

> We can quibble about the details of what to include, but that seems like it'd be an improvement over what's currently here:  http://www.sktime.net/en/latest/api_reference/datasets.html, which is a little sparse on the information you'd actually need to use for an example.  I found myself having to manually download a lot of different ones to verify what was actually inside them.

My reaction to this:

If the datasets were objects, and the properties in the columns were tags, then the existing `all_estimators` could be leveraged to produce this table quite easily, by running

```python
all_estimators("dataset", as_dataframe=True, return_tags=tag_list)
```
where `tag_list` is a list of tags corresponding to the table columns.

felipeangelimvieira:
[...]

Issue #4314, opened 2023-03-10
[ENH]Add datasets for functional regression


JonathanBechtel:
**Is your feature request related to a problem? Please describe.**
The collection of datasets available in sktime right now broadly fall into three categories:  univariate forecasting, multivariate classification, and forecasting with exogenous data.  However, there is another type of time series problem which does not have an example dataset:  multivariate regression.  This dataset would have the format `(n_instances, n_timepoints)` in `X`, and a corresponding column of continuous numbers for the target in `y`.  Examples of such data would be a collection of time series from sensor readings with a numeric target.

This issue is problematic because there are a number of estimators in sktime that are built for such use cases and are lacking documentation and examples, presumably because no such example datasets exist.  For example see this GitHub issue: https://github.com/sktime/sktime/issues/4264.

**Describe the solution you'd like**
I'd like to add at least 1-2 datasets that are specifically used for the case of functional regression, where a single sample represents an entire time series and the target is a continuous number.  One example would be the tecator dataset:  https://www.openml.org/search?type=data&sort=runs&id=505&status=active.  These would have similar functionality to existing multi class classification datasets, with a `return_X_y` and `split` arguments.

**Additional context**
You need an adequate number of example datasets to provide adequate documentation for the different estimators in sktime, and without said datasets it will be difficult to patch other issues with useful examples for different estimators.

fkiraly:
Yes, great idea!

Indeed the selection of data sets for tasks other than classification is a bit limited at a moment.
For forecasting, one might for example like to have the M4 or M5 data available. I think @TNTran92 started working on that a while ago.

Summarizing discussion from the Friday collab meeting:

* first step it would be really nice to have regression datasets
* seond step, as you suggested, could be thinking about cleaning up the loader module, adding a clean taxonomy for datasets (e.g., univariate, multivariate, which task) and making them queriable. An example where this is done well is the openML project.

JonathanBechtel:
Yes, correct.  I'll probably create a new issue for this broader agenda to touch up the datasets section, once it's specced out a bit more.  There are a number of tweaks that could be made in order to make it more usable.  

In addition to what you mentioned, here are some other items to think about:
 - Some of the datasets return data in an inconsistent format:  some datasets return `(y, X)`, and others return `(X, y)`
 - Some have a `split` argument, but I'm not sure why others wouldn't.  
[...]

PR #4333, opened 2023-03-15
DRAFT [ENH] experimental: datasets as objects

fkiraly:
This is a prototype draft refining some ideas of @JonathanBechtel in #4314.
See #4332.

This introduces a `BaseDataset` class and classification specific descendants, and showcases using the example of the `PLAID` dataset how datasets could be tagged and made be discoverable using `all_estimators`.

### DIFF
--- a/sktime/datasets/base/__init__.py
+++ b/sktime/datasets/base/__init__.py
@@ -1,0 +1,6 @@
+# -*- coding: utf-8 -*-
+"""Base class for datasets."""
+
+__all__ = ["BaseDataset"]
+
+from sktime.datasets.base._base import BaseDataset
--- a/sktime/datasets/base/_base.py
+++ b/sktime/datasets/base/_base.py
@@ -1,0 +1,38 @@
+# -*- coding: utf-8 -*-
+# copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+"""
+Base class template for data sets.
+
+    class name: BaseDataset
+
+Scitype defining methods:
+    loading dataset              - load()
+    loading object from dataset  - load(*args)
+
+Inspection methods:
+    hyper-parameter inspection   - get_params()
+"""
+
+__author__ = ["fkiraly"]
+
+__all__ = ["BaseDataset"]
+
+from sktime.base import BaseObject
+
+from sktime.utils.validation._dependencies import _check_estimator_deps
+
+
+class BaseDataset(BaseObject):
+    """Base class for datasets."""
+
+    # default tag values - these typically make the "safest" assumption
+    _tags = {}
+
+    def __init__(self):
+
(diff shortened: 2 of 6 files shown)


felipeangelimvieira:
Hi @fkiraly! Should I use this PR as a base for the dataset class? I may start implementing it this month

fkiraly:
Welcome back from holiday, @felipeangelimvieira!

Feel free to use this PR or branch off it, and feel free to change the design.
It might be good to have a quick discussion in the issue about the end design, or even write an enhancement proposal in the STEP repo.

Some things that one might consider in today's state of the `sktime` and `sklearn` repos:

* tags are now also "record" style classes. They can still be added as tuples, which may make sense until we have the design consolidated.
* `sklearn` has `Bunch` which could also be used here.
* some thinking may be required later around the topic of caching, if the dataset comes from a remote location. More generally, we want to separate the retrieval of the data from the presentation of the data in the architecture, mid-term.

However, the basic design principle should still be valid.

felipeangelimvieira:
@fkiraly,  I see that `load` is expected to receive some arguments from the user. What kind of arguments did you have in mind for this? I was expecting that the classes would only require arguments in __init__, with load being parameter-free.

felipeangelimvieira:
> @fkiraly, I see that `load` is expected to receive some arguments from the user. What kind of arguments did you have in mind for this? I was expecting that the classes would only require arguments in **init**, with load being parameter-free.

Actually, I now understand that the user should pass "X", "X_test", "X_train", etc., so that `load(*args)` returns the respective sets. I find this approach simple and useful.

I also have some questions about the tags:

* **Redundancy**: I see that some tags contain information already available in others, such as `is_one_panel` and `n_panels`. I assume this was intentional, but I’d like to understand the reasoning. Is the goal to simplify filtering of datasets?
* 
* **Datasets without train/test split**: some datasets, such as airline, do not come with an original train-test split, so the tags `"n_instances_train"=0, "n_instances_test"=0` are zero. In these cases, (a) should we raise an error if the user tries to retrieve "X_test", "y_test", "X_train" or "y_train", or (b) should we provide a default split?
* **Forecasting tags** : In your opinion, does the following tag set work for forecasting?
```
        "is_univariate": True,
        "is_one_panel": True,
        "is_hierarchical" : False,
        "is_one_series": True,
        "is_equally_spaced": True,
        "is_empty": False,
        "has_nans": False,
        "n_instances": 144,
        "n_instances_train": 0, 
        "n_instances_test": 0,
        "frequency": "M",
        "n_dimensions": 1, # Here, we would define is series is univariate or not
        "n_panels": 1,
        "n_hierarchy_levels": 1,  # this could be the number of index levels (except for time index at level -1)}
```

PR #4341, opened 2023-03-15
Feature/4281 generic ducktyping pipeline

### LINKED ISSUE #4281: [ENH] prototype approaches for polymorphic graph pipeline - duck typing
One of two prototype approaches to investigate for polymorphic graph pipelines: duck typing:

* same approach as `sklearn` `Pipeline` - switches on methods on demand, i.e., if estimator thinks it should have method
* use dispatcher to base class methods, with methods of pipeline having only `args`, `kwargs`

implement as prototype only for a linear pipeline, transformers plus one end element that can be transformer, forecaster, or classifier (`BaseTransformer`, `BaseForecaster`, `BaseClassifier`)

the current pipeline logic is in `make_pipeline`, works via dispatch on dunder

benHeid:
First scratch of a ducktyping based generic pipeline. Currently the solution works with a forecasters and a transformer.  A preliminary list of open questions and possible issues with such a solution are listed in the following:

1. predict/fit/... etc methods can only called with key word arguments. E.g. `x=X, fh=ForecastingHorizon(1)...`. I think we need a solution for this.
2.  Currently all potential methods need to exist in the pipeline (all methods from classifiers, forecasters, ...).
3. Until now, this is only tested with forecasters. **edit** Now tested also with transformer and classifier 
5. Problems may arise for different kind of forecasting pipelines. Currently, the transformer only manipulate the exogenous feature, how can we instead transform the time series itself. (e.g. scaling) in such a generalized setting?

#### Reference Issues/PRs
Fixes #4281


benHeid:
@fkiraly FYI. This is the current state, tomorrow I will test this with a classifier and with only transformers.

fkiraly:
Nice!

Here's my prototyping around with the polymorphic base class idea: https://github.com/sktime/sktime/pull/3108

benHeid:
I added two additional examples one with only transformers and one with a classifier in the pipeline.

Furthermore, I added a draft of a even more general pipeline. This pipeline solve the open question 2 in the description of the PR by introducing a method that can handle all transform and all predicts methods. 
During initialisation of the transformer all required methods are added to the pipeline. Probably this method should moved to the `__new__` dunder. However, a major drawback of this solution is that the intellisense systems of IDEs can not detect the available method. Thus, they show warnings.

benHeid:
Implemented by the graphical pipeline

PR #4342, opened 2023-03-16
[ENH] Add HPFilter Transformer

ken-maeda:
Reference Issues/PRs
Reference https://github.com/sktime/sktime/issues/4089

Built HPFilter Transformer using statsmodels time series filters.
Added a separate test "test_hpfilter" to test transformer against statsmodels

Almost similar to #4127
Difference of bkfilter and hpfilter
- input array support only 1d array
- output style( outputs in statsmodels are cycle and trend )

##### For all contributions


fkiraly (changes requested):
Great, thanks for your contribution!

Some change requests:

* kindly don't copy-paste the code from `statsmodels` but import it. That way, improvements and bugfixes directly propagate to sktime.
* The docstring should give clear credit, i.e., this is a direct interface to `statsmodels.etc.etc`
* `x` is not a parameter to the constructor, kindly fix
* kindly ensure your code passes code quality checks: https://www.sktime.net/en/latest/developer_guide/coding_standards.html#coding-standards

ken-maeda:
@fkiraly I appriciate your feedback. I have a question to clarify.

> kindly don't copy-paste the code from statsmodels but import it. That way, improvements and bugfixes directly propagate to sktime. Generally, we should avoid "forking" code (by copy-paste etc).

It means `HPFilter.transform_ ` is just like

```
  import statsmodels.api as sm
  
  cycle = sm.tsa.filters.hpfilter(X, self.lamb)[0]
  return cycle 
```
I thought I was refering BKfilter code to Hpfilter. BKfilter in sktime is like statsmodels code.

I maybe missed something about this enhancement. (I'm sorry for causing you trouble)

About other problem, I'm clear.

fkiraly:
@ken-maeda, you are right, `BKFilter` accidentally does the same thing, copying.
I've fixed it in this PR: https://github.com/sktime/sktime/pull/4346

Sorry for pointing you to a buggy template.

> It means HPFilter.transform_ is just like

Indeed! It is quite simple - one might ever cover it with a single adapter class and then expose the filters as special cases (descendants).

fkiraly:
small linting failures, you can have a look in the GitHub GUI if you click on "details" next to the failure

sktime/transformations/series/hpfilter.py
```diff
@@ -0,0 +1,123 @@
...
+"""
+Interface to Hodrick-Prescott filter from `statsmodels`.
+
+Please see the original library
+(https://github.com/statsmodels/statsmodels/blob/main/statsmodels/tsa/filters/hp_filter.py)
+Interfaces `hp_filter` from `statsmodels.tsa.filters`.
+"""
+# copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+
+__author__ = ["klam-data", "pyyim", "mgorlin", "ken_maeda"]
```
fkiraly, line 11: I think the author should be just you?

sktime/transformations/series/hpfilter.py
```diff
@@ -0,0 +1,123 @@
...
+            There are currently no reserved values for transformers.
+        Returns
+        -------
+        params : dict or list of dict, default = {}
+            Parameters to create testing instances of the class
+            Each dict are parameters to construct an "interesting" test instance, i.e.,
+            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+            `create_test_instance` uses the first (or only) dictionary in `params`
+        """
+        params = {"lamb": 1600}
```
fkiraly, line 122: there should always be at least two parameter sets.

I would go with an empty dict (the default) and this one.

sktime/transformations/tests/test_hpfilter.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
+"""Tests for HPfilter wrapper annotation estimator."""
+
+__author__ = ["klam-data", "pyyim", "ken-maeda"]
```
fkiraly, line 4: this should be just you

docs/source/api_reference/transformations.rst
```diff
@@ -333,6 +333,14 @@ Filtering and denoising
 
     BKFilter
 
+.. currentmodule:: sktime.transformations.series.hpfilter
```
fkiraly, line 336: actually, could you kindly move the `Filter` to the top?
Things should roughly be ordered by how common they are, I think `Filter` (generic band pass) is the most common filter people might be looking for. The others are more finance/econ specific afaik.

fkiraly (changes requested):
left some small comments above.

The only ones blocking are the linting failure and the author lines.

ken-maeda:
@fkiraly I think it should be fixed now.

sktime/transformations/series/hpfilter.py
```diff
@@ -0,0 +1,131 @@
...
+
+class HPFilter(BaseTransformer):
+    """Filter a times series using the Hodrick-Prescott filter.
+
+    This is a wrapper around the `hpfilter` function from `statsmodels`.
+    (see `statsmodels.tsa.filters.hp_filter.hpfilter`).
+
+    Parameters
+    ----------
+    x : array_like
```
fkiraly, line 31: `x` is not a parameter to the estimator, kindly remove this one

fkiraly (changes requested):
Looks great!

Two things:
* the docstring which has a superfluous `x` argument - probably a copy-paste accident.
* the example in the docstring does not work (this is tested), since you are using `sm`, instead import `statsmodels`. Also, kindly use doctest skip in the docstring example to isolate the soft dependency.

fkiraly (approved):
Great, thanks a lot!

Issue #4349, opened 2023-03-16
[ENH] `pytest` failure caused by `dash` version upgrade 2.8.1 -> 2.9.0


fkiraly:
Since a couple hours, CI fails with `pytest-xdist` workers are crashing.
The culprit appears to be `dash` 2.9.0, up from 2.8.1.

### history of diagnostics

* The suspicion is that this is coming from the recent changes in versions.
* the reason is not `pytest-xdist`, as removal does not fix the failure
* diffs in dependency versions show changes of `coverage` 7.2.1 -> 7.2.2, `dash` 2.8.1 -> 2.9.0

Attempted fixes:

* restrict `pytest-xdist<3.2.1` - does not fix - https://github.com/sktime/sktime/pull/4348
* restrict `pytest-xdist<3.2.0` - does not fix - https://github.com/sktime/sktime/pull/4348
* remove `pytest-xdist` - does not fix - https://github.com/sktime/sktime/pull/4350
* forcibly downgrade `coverage`, `dash` - fixes - https://github.com/sktime/sktime/pull/4351
* forcibly downgrade `coverage`, but not `dash` - does not fix - https://github.com/sktime/sktime/pull/4352
* forcibly downgrade `dash`, but not `coverage` - ?  - https://github.com/sktime/sktime/pull/4353

### example traceback

from https://github.com/sktime/sktime/actions/runs/4440696964/jobs/7796826928?pr=4347

```
scheduling tests via LoadScheduling

sktime/utils/tests/test_mlflow_sktime_model_export.py::test_pyfunc_raises_invalid_dict_key 
2023-03-16 21:05:09.652491: I tensorflow/core/platform/cpu_feature_guard.cc:193] This TensorFlow binary is optimized with oneAPI Deep Neural Network Library (oneDNN) to use the following CPU instructions in performance-critical operations:  AVX2 AVX512F FMA
To enable them in other operations, rebuild TensorFlow with the appropriate compiler flags.
2023-03-16 21:05:09.800236: W tensorflow/compiler/xla/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic library 'libcudart.so.11.0'; dlerror: libcudart.so.11.0: cannot open shared object file: No such file or directory; LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.9.16/x64/lib
[...]

PR #4351, opened 2023-03-16
[MNT] addressing `pytest` failure - downgrade `coverage`, `dash`

fkiraly:
Attempted/diagnostic fix to #4349 - attempt is downgrading packages to last known working versions: `coverage` to 7.2.1, from 7.2.2, and `dash` to 2.8.1 from 2.9.0


fkiraly:
diagnostic PR, no longer needed

PR #4356, opened 2023-03-17
[MNT] `pandas` 2!

fkiraly:
Loosens bound to allow `pandas` 2 versions as dependencies - let's see what breaks...


fkiraly:
well, that was painless!
Looks like all those deprecations we hear about in the warnings will kick in only at a later version...

PR #4360, opened 2023-03-18
[BUG] fix seasonality estimators for `candidate_sp` being `int`

fkiraly:
This fixes an unreported bug when `candidate_sp` in two seasonality estimators would be `int`. According to the docstring, this should be treated as a one-element iterable, but instead led to error.

This is fixed by coercing to one-element iterable, and covered by an additional test case.

The estimators concerned are `SeasonalityACF` and `SeasonalityACFqstat`.


fkiraly:
Factored out of #4248 which was open for a while

PR #4382, opened 2023-03-23
Move to rnkuhns to former core developer

RNKuhns:
I'm requesting that I be moved to the former core developer list.

Best of luck on everything, but I can't be part of things currently.


fkiraly:
sorry to see you go, but I understand.
Hope to see you back sometime.

Issue #4363, opened 2023-03-19
[BUG] `BaggingForecaster` returns incorrect index for probabilistic forecasts


fkiraly:
Same as https://github.com/sktime/sktime/issues/4362, but for `BaggingForecaster`

Issue #4362, opened 2023-03-19
[BUG] `DynamicFactor` returns incorrect index for probabilistic forecasts


fkiraly:
`DynamicFactor` returns incorrect index and length returns for probabilistic forecasts, e.g.,

```python
from sktime.utils._testing.series import _make_series
from sktime.forecasting.dynamic_factor import DynamicFactor

n_columns=2
fh=[2, 5]

y_train = _make_series(n_columns=n_columns)

f = DynamicFactor()
f.fit(y_train, fh=fh)

f.predict_interval()
```

returns a length 4 `DataFrame` with integer index, but it should return a length 2 `DataFrame` with datetime index.

This bug was masked by unreported bug https://github.com/sktime/sktime/pull/4361 in the testing framework.

FYI @Ris-Bali, @lbventura as authors.

PR #4393, opened 2023-03-25
[ENH] cleaned up probabilistic forecasting tests for quantile and interval predictions

fkiraly:
This PR  cleans up probabilistic forecasting tests for quantile and interval predictions, i.e., `test_predict_interval` and `test_predict_quantiles`.

It makes the two tests consistent, and ensures both check the expected index and columns, which they previously did not.

Minor additions in this PR connected to the tests:
* adds a probabilistic test case to `ForecastX`, which would surface the bug https://github.com/sktime/sktime/issues/4386 (and covers its fix)
* `BaggingForecaster` and `DynamicFactor` have a broken `predict_interval`, this is a known issue, see #4362 and #4363. They also need to be excepted from the updated `test_predict_interval` which also would surface the same bug now.

Depends on https://github.com/sktime/sktime/pull/4399, as fixes of bugs detected through this are in https://github.com/sktime/sktime/pull/4399


yarnabrina:
Continuing what I said [here](https://github.com/sktime/sktime/pull/4391#issuecomment-1483916233), does it make sense to change from this:

https://github.com/sktime/sktime/blob/e75cfc4508819059d486cb548f3886f992fa6981/sktime/forecasting/tests/test_all_forecasters.py#L406-L410

to the effect of this?

```python3
expected_columns = ["Quantiles"] if isinstance(y_train, pd.Series) else y_train.columns
expected_quantiles = [alpha] if isinstance(alpha, float) else alpha
expected = pd.MultiIndex.from_product([expected_columns, expected_quantiles])
```

And then add at least a list element in here:

https://github.com/sktime/sktime/blob/e75cfc4508819059d486cb548f3886f992fa6981/sktime/forecasting/tests/_config.py#L92

For example:

```python3
TEST_ALPHAS = [0.05, 0.1, [0.25, 0.75]]
```

Please let me know what you think.

fkiraly:
yes, that's better! Much clearer, and adds a test case with multiple alphas.

I'd be happy if you branch off this PR and make the change. Or, suggest the change (using the GitHub graphical web interface) and I'll accept it.

yarnabrina:
> I'd be happy if you branch off this PR and make the change. Or, suggest the change (using the GitHub graphical web interface) and I'll accept it.

Done. See #4394.

fkiraly:
Great, thanks!

I will merge in this sequence:

* #4391 
* this
* #4394 

First I'll update from main and ensure the tests pass here.
They have already failed with `ForecastX`, so I'd expect they now run after the fix #4391 is merged.

If that is the case, we merge to main and update #4394 from main and check whether the extended tests still run.

PR #4394, opened 2023-03-26
[ENH] cover list input case for `test_predict_interval`/`coverage` and `test_predict_quantiles`/`alpha` in forecaster contract tests

yarnabrina:
Updates `test_predict_interval` and `test_predict_quantiles` fixtures for `coverage` and `alpha`, adds list input case.

Note: These tests **should** fail on current `main`, but should start passing after #4391 is merged. Verified locally with following commands.

> pytest -k "test_predict_interval[ForecastX" sktime/forecasting/tests/test_all_forecasters.py

> pytest -k "test_predict_quantiles[ForecastX" sktime/forecasting/tests/test_all_forecasters.py


fkiraly:
Thanks! Any idea why the tests aren't running?
You changed none of the CI config files, so why would this be happening?

fkiraly:
Oh, I see - you want to merge into my branch.
I changed the target to main, and I will merge the PR in this sequence:

* your bugfix https://github.com/sktime/sktime/pull/4391
* https://github.com/sktime/sktime/pull/4393
* this PR

fkiraly (changes requested):
Something doesn't seem quite right here - the mock forecaster is failing the test.

That might be a problem with the fixed test, or with the mock forecaster (`MockForecaster`). I'll try to see whether I can see sth easy/obvious today.

fkiraly:
here's my guess (I would have to dig to confirm):

previously, only one alpha/coverage was tested, so the code that checked whether there is the expected monotonicity was never run. I.e., forecasters could get away with violating the monotonicity property.

So, there could be a bug with that code (error of sign?), or a bug with specific forecasters (e.g., mock forecaster might just return random values which end up not being monotonous)

yarnabrina:
> mock forecaster might just return random values which end up not being monotonous

I guess this is the issue. See this:

https://github.com/sktime/sktime/blob/306e90007666378e04989cb3255c58949235f077/sktime/utils/estimators/_forecasters.py#L324-L369

If `self.prediction_constant` is negative, this can't be monotonically increasing, given `alpha` (a) is positive fraction.

How do you want to handle this?

fkiraly:
> How do you want to handle this?

Ohhhhhh, I see. Good spot, since one of the test examples has the constant being -4.2.

I would simply replace `a` in the formula by `1-a` if the constant is negative, then the monotonicity will be the right way round.
It's a mock estimator anyway, so anything that will turn it the right way should be fine.

From a practical perspective, would you like to make a PR? I would make this completely separate and branch off main - then merge and see what remains failing.

PS, there are some other failures too: https://github.com/sktime/sktime/pull/4393
these also seem genuine

yarnabrina:
> From a practical perspective, would you like to make a PR? I would make this completely separate and branch off main - then merge and see what remains failing.
> 
> PS, there are some other failures too: #4393 these also seem genuine

Created #4397. Will try to take a look at test failures tomorrow, but I'm afraid won't get much time before weekend.

fkiraly:
> Will try to take a look at test failures tomorrow, but I'm afraid won't get much time before weekend.

No worries, I'll have a look at the #4393 failures then, it would be appreciated then if you start with `MockForecaster` if you have any time (seems straightforward given your diagnosis), I'll try to deal with the others first.

fkiraly:
> would be appreciated then if you start with MockForecaster

... which I see you've already done! #4397
Thank you!

fkiraly:
(merged original branch #4393 into this to preserve the order of the still open PR)

fkiraly:
> Since all tests seem to pass in https://github.com/sktime/sktime/pull/4393 and https://github.com/sktime/sktime/pull/4399 now, will you merge them in main?

Eventually - we have a standing policy where we leave PR open for 5 working days, at least until they get a "serious" review.
(there is also a longer period they are considered "under consideration" and could be unmerged)

If you would like to review the PR, that would be much appreciated!

fkiraly:
> I'll merge updated main in my fork, and then can see what's causing the test failures in predict_interval over next 2/3 days.
> Or, if you would prefer me to solve the tests first before updating main, I can do that locally for testing.

I've already updated this branch with #4393 and #4399, apologies if that wasn't clear! As a consequence of my saying in which sequence we merge, the workflow also implies that we keep PR n updated with PR n-1 in the sequence.

So, you can go ahead with testing and investigation, this is the state that the PR would have if the other two were already merged.

yarnabrina:
The tests are failing because `VECM` cannot handle list of float as coverage values.

```python
import numpy as np
import pandas as pd
from sktime.forecasting.vecm import VECM

index = pd.date_range(start="2005", end="2006-12", freq="M")
df = pd.DataFrame(np.random.randint(0, 100, size=(23, 2)), columns=list("AB"), index=pd.PeriodIndex(index))

model = VECM()
model.fit(df)

model.predict_interval(fh=fh, coverage=[0.9, 0.95])
```

It results in the similar error and traceback as per test failures. Let me know if I should fix it here (inner loop results are not going out, so only the last iteration of outer loop is present), or in a new PR.

fkiraly:
> or in a new PR.

that, as you just did - great!

Makes the sequence of PR now a bit non-obvious.
But, since we are using *this* PR to test #4411 (good call!), we can merge #4411 any time before this one.

fkiraly:
hm, this fixes all the bugs now together with #4399 - nice!

fkiraly (approved):
Forgot to approve, I obviously approve (see comments above)

fkiraly:
hm, seems to error out with `VAR` predictive quantiles not being monotonically increasing.
It seems randomized and sporadic though, as it does not fail in all (three) cases where this test is run (it could be OS/version specific, but my guess is sporadic).

yarnabrina:
Since the last run (against https://github.com/sktime/sktime/pull/4394/commits/dd543eea8e2f1ab1c015cb4206f5e71abb73e761) passed all tests, random seems more likely, but to be confirmed, is there a way to trigger that workflow (may be just that test) manually? If it gets repeated, I can take a look tomorrow.

NB: I've not seen `sktime` configurations in detail, but I can trigger workflows manually in my personal repositories with `workflow_dispatch`.

fkiraly:
You can run a given test/fixture/estimator combination locally by using `check_estimator`, see here:
https://www.sktime.net/en/stable/developer_guide/add_estimators.html#using-the-check-estimator-utility

For the failing test, this would be:

```python
check_estimator(VAR, fixtures_to_run="test_predict_quantiles[VAR-0-fh=1-alpha=[0.25, 0.75]]")
check_estimator(VAR, fixtures_to_run="test_predict_quantiles[VAR-0-fh=[2 5]-alpha=[0.25, 0.75]]")
```
(I usually copy-paste the fixture names)

What we could also do, is open a bug issue ("sporadic failure in ..." and explain the monotonicity problem), add a test exception to `_config`, to ensure the bug is being tracked and does not block the test framework code.

yarnabrina:
I think I may not have expressed myself well. I did run the failed tests locally (Ubuntu WSL on Windows 11), and it passed. But since you said there is at least a possibility of the failure being OS specific, it's working on my local (or anyone's local) is probably insufficient to conclude that it's random.

So I was asking if we can run only that test in Github actions or not. So that would ensure same OS, same versions, etc. identical scenario as per the last run. Now if we run that multiple times and if it fails sometimes AND works sometime, we can infer with confidence that it's random. If it always fails in that setup, but never in any other setup, then it's setup specific.

So, basically I was asking if there's a workflow for following purpose:

Trigger: on `workflow_dispatch` instead of on `push` or `pull_request` to `main`
Inputs: OS, architecture (is required?), python version, estimator name, fixture name
Flow: Run `check_estimator` corresponding to inputs

This way we shall hopefully be able to narrow down the setups, and `skip` (or `xfail`) of a particular test can be made more limited instead of skipping always.

fkiraly:
> I did run the failed tests locally (Ubuntu WSL on Windows 11), and it passed

Have you run it more than once? It might fail only 2 times out of 100 or similar, if it is stochastic and sporadic.
First thing I'd do is run it a 100 times and see how often it fails locally.

> So I was asking if we can run only that test in Github actions or not

Ah, I see!
That's not possible with the vanilla set-up, but you can achieve the effect by rewriting the `workflows/test.yml` and putting the pytest command in there, although that's quite hacky.

A "cleaner" way would be using the test dockers, although that does not allow you to specify all the OS - specifically windows where  so it won't help in this case: https://www.sktime.net/en/latest/developer_guide/continuous_integration.html#alternative-dockerized-testing

> I was asking if there's a workflow for following purpose:

No, that would seem useful for testing though (albeit in some specific cases like this).

fkiraly:
> of a particular test can be made more limited instead of skipping always.

If using `_config`, it would be limtied to the particular estimator/fixture combination, that's more granular than `xfail`.

yarnabrina:
Hi, I see a weird environment setup failure on main after this PR is merged: https://github.com/sktime/sktime/actions/runs/4594058440/jobs/8112704779. Is it something caused by some of my changes?

PR #4402, opened 2023-03-28
[ENH] Christiano Fitzgerald filter transformer

ken-maeda:
Reference Issues/PRs
Reference https://github.com/sktime/sktime/issues/4089

Built CFFilter Transformer using statsmodels time series filters.
Added a separate test "test_hpfilter" to test transformer against statsmodels

Almost similar to https://github.com/sktime/sktime/pull/4127
Difference of bkfilter and cffilter

output style(outputs in statsmodels are cycle and trend )


sktime/transformations/series/cffilter.py
```diff
@@ -0,0 +1,130 @@
...
+
+        Returns
+        -------
+        params : dict or list of dict, default = {}
+            Parameters to create testing instances of the class
+            Each dict are parameters to construct an "interesting" test instance, i.e.,
+            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+            `create_test_instance` uses the first (or only) dictionary in `params`
+        """
+        params1 = {"low": 6, "high": 24, "drift": True}
```
fkiraly, line 128: this is identical to the default, I would make all of these values differ.

sktime/transformations/series/cffilter.py
```diff
@@ -0,0 +1,130 @@
...
+    """Filter a times series using the Christiano Fitzgerald filter.
+
+    This is a wrapper around the `cffilter` function from `statsmodels`.
+    (see `statsmodels.tsa.filters.cf_filter.cffilter`).
+
+    Parameters
+    ----------
+    low : float
+        Minimum period of oscillations. Features below low periodicity
+        are filtered out. Default is 6 for quarterly data,
```
fkiraly, line 31: the default is always 6, so this sentence is confusing. I would say "for quarterly data, the default of 6 gives 1.5 years periodicity".

sktime/transformations/series/cffilter.py
```diff
@@ -0,0 +1,130 @@
...
+        giving a 1.5 year periodicity.
+
+    high : float
+        Maximum period of oscillations. Features above high periodicity
+        are filtered out. Default is 32 for quarterly data,
+        giving an 8 year periodicity.
+
+    drift : bool
+        Whether or not to remove a trend from the data.
```
fkiraly, line 40: bit unclear what "remove" means. I assume it means "subtract"?
Also, `x` is not defined, you need to say this refers to a univariate `X` seen in `transform`.

fkiraly (changes requested):
Thanks, looks good!

Change requests are for docstrings and the test parameter settings.

sktime/transformations/series/cffilter.py
ken-maeda: The sentense "remove trend" somethime is used in time series analysis, but it completely is same meaning to substracting trend from the sequence. So I'll replace it.

sktime/transformations/series/cffilter.py
```diff
@@ -0,0 +1,130 @@
...
+
+class CFFilter(BaseTransformer):
+    """Filter a times series using the Christiano Fitzgerald filter.
+
+    This is a wrapper around the `cffilter` function from `statsmodels`.
+    (see `statsmodels.tsa.filters.cf_filter.cffilter`).
+
+    Parameters
+    ----------
+    low : float
```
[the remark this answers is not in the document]
ken-maeda: OK, I should have added this discription to `hpfilter` and `bkfilter`. Should I create another PR for fix them too? or fix in this PR?

sktime/transformations/series/cffilter.py
fkiraly: I know, but it does not always have the same meaning. You could also de-trend multiplicatively. It's worth being as precise as possible, since people from a different user context may use the class.

sktime/transformations/series/cffilter.py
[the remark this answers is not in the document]
fkiraly: I would suggest another PR - it's a good idea to raise PR "by topic" and keep them separate if possible (e.g., if there are multiple topics). Very similar to the "single responsibility princinple" in coding.

Apologies, I did not spot that there were defaults and they were missing in the docs, but you are right!

sktime/transformations/series/cffilter.py
```diff
@@ -0,0 +1,132 @@
...
+        Minimum period of oscillations. Features below low periodicity
+        are filtered out. For quarterly data, the default of 6 gives
+        1.5 years periodicity.
+
+    high : float, optional, default = 32.0
+        Maximum period of oscillations. Features above high periodicity
+        are filtered out. For quarterly data, the default of 32 gives
+        8 year periodicity.
+
+    drift : bool, optional, default =
```
fkiraly, line 39: default is missing, it´s `True` according to the code

sktime/transformations/series/cffilter.py
```diff
@@ -0,0 +1,132 @@
...
+        Whether or not to substract a trend from the data.
+        The trend is estimated as np.arange(nobs)*(X[-1] -X[0])/(len(X)-1).
```
fkiraly, line 41: is this correct? The `X` in `_transform` ends up being a 2D array, not a 1D array.
Also, is `nobs` the same as `len(X)`, which would be clearer?

sktime/transformations/series/cffilter.py
```diff
@@ -0,0 +1,132 @@
...
+
+        Returns
+        -------
+        params : dict or list of dict, default = {}
+            Parameters to create testing instances of the class
+            Each dict are parameters to construct an "interesting" test instance, i.e.,
+            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+            `create_test_instance` uses the first (or only) dictionary in `params`
+        """
+        params1 = {"low": 8, "high": 26, "drift": True}
```
fkiraly, line 130: `drift` should be a non-default, i.e., `False`

fkiraly (changes requested):
Thanks! Very minor change requests and a question.

PR #4404, opened 2023-03-28
[ENH] Clustering plot function now supports alpha (transparency) value for individual timeseries lines

Aylr:

A user with large numbers of time series using the existing clustering plotting function `plot_cluster_algorithm` can suffer from  [over plotting](https://www.quanthub.com/how-do-you-fix-overplotting/). This is easily solved with the addition of transparency.

This PR introduces an optional argument that allows a user to customize the transparency of the individual time series plots.

**Before:**

```python
plot_cluster_algorithm(k_means, df, k_means.n_clusters)
```
Screen Shot 2023-03-28 at 9 59 19 AM

**After:**

```python
plot_cluster_algorithm(k_means, df, k_means.n_clusters, alpha=0.2)
```
Screen Shot 2023-03-28 at 9 59 47 AM

No.

No.

##### For all contributions

##### For new estimators

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -2146,6 +2146,15 @@
         "bug",
         "code"
       ]
+    },
+    {
+      "login": "Aylr",
+      "name": "Taylor Miller",
+      "avatar_url": "https://avatars.githubusercontent.com/u/928247?v=4",
+      "profile": "https://github.com/Aylr",
+      "contributions": [
+        "code"
+      ]
     }
   ]
 }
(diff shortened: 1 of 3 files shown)


fkiraly (changes requested):
Good idea!

The one chane I would request: `alpha` is not special, there are other arguments that users might like to have passed on.

So, I would ask you to replace the `alpha` args with a `kwargs` arg that is passed through to `axes.plot`.

I would also recommend (not a blocker but a suggested feature for another PR if you agree), add a `fig`, `axes` arg to the `plot_cluster_algorithm` function (default=None, then it is created inside), and return `fig`, `axes`.
That way, the plotting function can be used within subplots, too.

fkiraly:
@aylr, are you still working on this? Let us know if we can help!

PR #4411, opened 2023-03-29
[BUG] Support multiple coverage values for VECM

yarnabrina:
#### Reference Issues/PRs

See also #4394 (relevant comment: https://github.com/sktime/sktime/pull/4394#issuecomment-1489052793).

The previous implementation of `VECM` caused `ValueError` when `coverage` was passed as a list of float to `predict_interval`. This was caused because lower/upper bounds determined for different coverages were not stored, and it was trying to match the last iteration with all columns generated independently. This PR addresses that by combining the results from all the different coverage values, and then creating the final results.

No.

No.

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/forecasting/vecm.py
+++ b/sktime/forecasting/vecm.py
@@ -260,6 +260,7 @@ class VECM(_StatsModelsAdapter):
         )
         int_idx = pd.MultiIndex.from_product([var_names, coverage, ["lower", "upper"]])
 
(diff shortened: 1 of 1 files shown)


fkiraly (approved):
Well-diagnosed, well-spotted!

Also, it's covered by tests in https://github.com/sktime/sktime/pull/4394 and currently being (positively) tested by tests in there, so all great!

Issue #4420, opened 2023-04-02
[BUG] sporadic failure in `test_predict_quantiles` during monotonicity check for `VAR`


yarnabrina:
**Describe the bug**

Since quantile function if non-decreasing, predicted quantiles should be monotone, i.e. prediction of smaller quantile should not be larger than that for higher quantile. For `VAR`, this check is failing occasionally.

**To Reproduce**

This script takes a fixture name for `VAR` as an input and number of replications as another. It repeats the corresponding fixture that number of times and then prints the proportion of times it failed.

```python
# test_random_failures.py
import argparse
import statistics
import warnings

from sktime.forecasting.var import VAR
from sktime.utils.estimator_checks import check_estimator

warnings.filterwarnings("ignore")

parser = argparse.ArgumentParser()

parser.add_argument("fixture", type=str)
parser.add_argument("replication", type=int)

parsed_arguments, _ = parser.parse_known_args()

results = list(
    check_estimator(VAR, fixtures_to_run=parsed_arguments.fixture)[
        parsed_arguments.fixture
    ]
    for _ in range(parsed_arguments.replication)
)
print(statistics.mean(result != "PASSED" for result in results))
```

Running the following leads to these results:

```sh
❯ python test_random_failures.py "test_predict_quantiles[VAR-0-fh=1-alpha=[0.25, 0.75]]" 1000
...
0.066
❯ python test_random_failures.py "test_predict_quantiles[VAR-0-fh=[2 5]-alpha=[0.25, 0.75]]" 1000
...
0.048
```

**Expected behavior**

The tests should pass all the time, leading zero failure proportion.

**Additional context**

See https://github.com/sktime/sktime/pull/4394#issuecomment-1493244223.

**Versions**

The above results are on Ubuntu WSL running on Windows 11. The local environment is using `Python 3.8.16` and have editable `sktime` installation same as https://github.com/sktime/sktime/pull/4394/commits/80ccaf7512abd057ee485fe85713a7177cc2e7ca.

Note: `show_versions()` failed with following, so unable to share that information:

> AssertionError: assert '_distutils' in core.__file__, core.__file__

fkiraly:
Thanks for checking!

* `show_versions` seem to be broken, I am getting the same message. Fix and test for it here: https://github.com/sktime/sktime/pull/4421
* I would recommend to add the failing test as a `_config` exception (unless the fix is clear)

fkiraly:
[...]

PR #4425, opened 2023-04-03
[BUG] skip test_predict_quantiles for VAR for sporadic behaviour

yarnabrina:
#### Reference Issues/PRs

Addresses #4420 

`predict_quantiles` for `VAR` was sometimes giving results where lower quantiles had higher values than upper quantiles. This is incorrect behaviour, but it was happening rarely (close to 5% of times). As a quick fix to unblock #4394, this PR disables testing of `predict_quantiles` for `VAR` estimator by explicit exception of `test_predict_quantiles` for `VAR` in `_config.py`.

No.

No.

#### Any other comments?

This PR does not solve the issue itself, it merely avoids it so that test suite does not fail randomly. Possible things to investigate in future:

* look for pattern in data when monotonicity fails
* https://github.com/sktime/sktime/issues/4420#issuecomment-1493365120

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/tests/_config.py
+++ b/sktime/tests/_config.py
@@ -139,6 +139,8 @@ EXCLUDED_TESTS = {
         "test_predict_quantiles",
         "test_predict_proba",
     ],
+    # random quantile monotonicity issue in VAR, refer to #4420
+    "VAR": ["test_predict_quantiles"],
 }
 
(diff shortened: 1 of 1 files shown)


fkiraly (approved):
Thanks!

Issue #4426, opened 2023-04-03
[BUG] `pandas 2` compatibility issues


fkiraly:
Some problems with `pandas 2` as released on Apr 3:

* `ForecastingHorizon` `to_absolute` no longer seems to work with `pandas 2` as released, due to incomaptibility with the index based logic in `to_absolute`
* `pd.DataFrame` constructor no longer accepts `ForecastingHorizon` as `index` or `columns` argument
* conversion from `nested_univ` to `pd-multiindex` causes `pd.Index` indices which are not compatible with data input contracts
* changes that make the contract of `DataFrame.agg` more restrictive and cause exceptions
* `loc` can no longer be used to write to index locations that do not exist
* the `pandas.util.testing` module no longer exists, contents moved to `pandas.testing`
* `Timestamp.freq` is now removed. Despite dealing with this already, there were still some instances in `sktime` using this.
* slicing timelike indices with a single index of lower granularity is no longer possible, e.g., `df["1950"]` on monthly data now raises error instead of returning all months in 1950

Further problems to be mapped by https://github.com/sktime/sktime/pull/4430

fkiraly:
why was this not detected?

It appears that some of the soft dependenies imply `pandas < 2.0.0`, so relaxing the bound in `sktime` is not enough if all the soft dependencies are present.

fkiraly:
code to trigger/diatnose vectorization failure:

```python
from sktime.transformations.series.exponent import ExponentTransformer
from sktime.utils._testing.panel import _make_panel_X

X = _make_panel_X()
t = ExponentTransformer()
t.fit_transform(X)
```

PR #4427, opened 2023-04-03
[MNT] `pandas` 2 compatibility fixes

fkiraly:
Fixes https://github.com/sktime/sktime/issues/4426.
This PR contains some tentative fixes for compatibility issues arising from the `pandas` 2 release on Apr 3:

* fixes to index handling in `ForecastingHorizon.to_absolute`
* fixes to many instances of using `ForecastingHorizon` as `pd.DataFrame` `index` in constructor or setter (`.index = ...`), this is no longer possible
* writing to a `loc` index that does not exist no longer creates it but raises an exception - this was replaced by ensuring the right rows and columns are present before the write - `AutoETS`, `UnobservedComponents`
* replace deprecated `is_integer` and `is_numeric` with recommended replacement calls to `pandas.api.types`
* fixed old `from_nested_to_multi_index` utility producing `object` valued multiindex (instead of integer typed etc)
* same for `_convert_tsf_to_hierarchical` utility in data loaders
* replace deprecated `pandas.util.testing` imports with `pandas.testing`
* https://github.com/sktime/sktime/pull/4435
* https://github.com/sktime/sktime/pull/4436
* replace in-place nested pandas operations with construction of new frame: `test_interpolate`
* https://github.com/sktime/sktime/pull/4438
* replace single-index temporal slicing at lower granularity with slice-based slicing

Relies on https://github.com/sktime/sktime/pull/4429 for testing.

Does *not* change the dependency bounds.


fkiraly:
https://github.com/sktime/sktime/pull/4430 passes - this looks complete now!

fkiraly:
(newest failure due to wrong `is_numeric` replacement - this is post-2.0 deprecation, so I'm replacing that again to restore running tests - will address that separately)

yarnabrina:
Is there any way to test if everything is returning same results on pandas v2 as it was on v1? I'm not even sure if it's possible to test, as an environment will have only one of the two versions, but this seems like a reasonable expectation from user point of view (other than random stuff, which should be independent of pandas).

fkiraly:
> Is there any way to test if everything is returning same results on pandas v2 as it was on v1? I'm not even sure if it's possible to test automatically in CI or even locally in a single environment, as an environment will have only one of the two versions, but this seems like a reasonable expectation from user point of view (other than random stuff, which should be independent of pandas).

Not sure whether there is an easy way to test that pandas 1 and 2 results are the same.

You can, however, run the same batch of tests in a `pandas 1` and a `pandas 2` environment.

What I was thinking is have a single CI element with `pandas 2` in it, e.g., a matrix across the three OS with 1/3 of the estimators each, under python 3.11.

PR #4430, opened 2023-04-03
DIAGNOSTIC [MNT] `pandas` 2 diagnostic PR

fkiraly:
Diagnostic PR, version of https://github.com/sktime/sktime/pull/4427 with CI conditional requirements relaxed so all tests run even if basic tests fail - to surface general incompatibility areas.

DO NOT MERGE

Makes changes for easy diagnostics:

* removes soft dependencies with known incompatibility to `pandas 2`: `xarray`, `statsforecast`
* removes requirement for test conditionality to run all from start

(THIS SHOULD NOT BE MERGED)


fkiraly:
hm, some soft deps seem to imply `pandas < 2.0.0`?

fkiraly:
Notes:

* the fixes to `to_absolute` for `ForecastingHorizon` `seem to fail on the lower end environment with `python 3.7`, `pandas 1.3.5` but work fine from `python 3.8`, on `pandas 1.5.3`

fkiraly:
looks like all passes. Failure on the notebook is due to `statsforecast` which is not in the env (as it is incompatible with `pandas 2` currently)

fkiraly:
no longer needed as we now test for both `pandas 2` and `pandas 1`.

Issue #4431, opened 2023-04-04
[BUG] umbrella issue: sporadic failures of monotonicity in `test_predict_quantiles`


fkiraly:
Multiple forecasters produce, on occasion, quantile forecasts that are not monotonic:

* `VAR` https://github.com/sktime/sktime/issues/4420
* `Prophet`

This is sporadic and happens in a small percentage of cases.

A commonality may be doing Bayesian sampling internally, e.g., sampling is done independently for the different `alpha` or `coverage` values, so the approximated posteriors may end up crossing (while the real posteriors don't)

fkiraly:
FYI @yarnabrina

yarnabrina:
> should we allow this to happen

A very strict way will be to not allow at all, and let it fail. A slightly less strict (and debatable) way can be to run once again and then fail or return based on monotonicity.

> should we force-sort returned prediction intervals

If I am to express my opinion, I do not want this. Even if we allow reverse quantiles, one possibly still justify the reasons causing it and the associated likelihoods, as you did above. I'll prefer un-sorted quantiles with a warning (generated if quantiles != sorted(quantiles) etc.), over force-sorted quantiles. Or, this can be left to users as well, e.g. `warn_on_monotonicity=True`.

> A commonality may be doing Bayesian/MC sampling internally, e.g., sampling is done independently for the different alpha or coverage values, so the approximated posteriors may end up crossing (while the real posteriors don't)

I know some of the probabilistic forecast packages like `gluonts` predict this way, but is it done for all the models? Or is this just an example?

May be we should attempt (somehow) to generate at least one occurrence of unsorted quantile predictions for underlying original estimators. That would at least ensure that nothing in `sktime` is causing it.

fkiraly:
> If I am to express my opinion, I do not want this.

I would agree with you there. "if you build a framework, don't police the instance" would be the design principle, i.e., giving implementers freedom to move within the framework.
If they want crossing quantiles, so be it (perhaps there is a reason why one would like this).

> May be we should attempt (somehow) to generate at least one occurrence of unsorted quantile predictions for underlying original estimators. 

Agreed. We should check whether it's coming from upstream.

> A likely stupid question: is it correct to use predict_interval for predict_quantiles? The former says this interval will contain the actual value {coverage}% cases, and the latter says {alpha} is the chance for actual value to be less than this value. Do these necessarily correspond to each other, across any types of skewness?

[...]

PR #4432, opened 2023-04-04
[MNT] except `Prophet` from `test_predict_quantiles` due to sporadic failures

fkiraly:
Excepts `Prophet` from `test_predict_quantiles` due to sporadic failures of the current test (quantile prediction monotonocity), see https://github.com/sktime/sktime/issues/4431


fkiraly:
quick merge due to the bug being on main

Issue #4377, opened 2023-03-21
[ENH] pass `statsmodels` `fit_params` to constructor


fkiraly:
From discord, indirect request by ckwan48.

The `SARIMAX` model interfacing `statmodels` does not allow to specify the parameters that `statsmodels` would receive in `fit`. These should be specified in the `sktime` constructor.
Good first issue as a recipe is available, here:
(similar to #2344)

---

Recipe for interfacing models with `fit_params` or any other `fit` args, in sktime:

* pass the parameter(s) to the constructor (the `__init__` method), using explicit parameter names and default, not `**kwargs`
* write it to `self` in the constructor
* in `_fit`, read it from `self` and pass it to the wrapped model's `fit` method
* ensure docstring is complete and defaults documented (optimally same as in wrapped estimator)

This allows to expose all parameter choices via the scikit-learn compatible interface (which relies on the constructor to expose all parameter choices).

For an example of this, you can look at `smoothing_level` in `ExponentialSmoothing`.

AishwaryaSatpute1:
I would Like to work on this issue , will you please assign this issue to me ?

fkiraly:
Sure, it's all yours!

fkiraly:
@AishwaryaSatpute1, are you still working on this?
If so, as @yarnabrina noted here https://github.com/sktime/sktime/pull/4439#issue-1656568085, it might make sense to add `simple_differencing` in the same PR.

yarnabrina:
Noting the list of available `fit` parameters for easy reference in future.

https://www.statsmodels.org/dev/generated/statsmodels.tsa.statespace.sarimax.SARIMAX.fit.html#statsmodels.tsa.statespace.sarimax.SARIMAX.fit-parameters

PR #4439, opened 2023-04-06
[ENH] Adds `_predict_interval` to `SARIMAX` to support `predict_interval` and `predict_quantiles`

### LINKED ISSUE #4301: [ENH] `SARIMAX` - implement `predict_interval` or `predict_quantiles` (for `statsmodels` `SARIMAX` interface)
From discord, Huseyin Aytug:

hi when I want to produce confidence intervals with `SARIMAX`, I get this error:

`NotImplementedError: SARIMAX does not have the capability to return quantile predictions. If you think this estimator should have the capability, please open an issue on sktime.`

However, it should be possible according to the documentation http://www.sktime.net/en/latest/api_reference/auto_generated/sktime.forecasting.sarimax.SARIMAX.html#sktime.forecasting.sarimax.SARIMAX.predict_interval

yarnabrina:
#### Reference Issues/PRs

Fixes #4301.

Before this PR, `SARIMAX` did not support prediction of confidence intervals and quantiles. This PR adds those functionalities, and tests to check that results match with underlying `statsmodels` estimator.

No.

Yes.

#### Any other comments?

`SARIMAX` in `statsmodels` has a boolean argument `simple_differencing`. If it's set to be `True`, the forecasts and predictions are about the differenced series. It can be considered to add handling of that, similar to [how `pmdarima` does it](https://github.com/alkaline-ml/pmdarima/blob/18cbe34d62bf7ec2508509b0590ff0962f00af3e/pmdarima/arima/arima.py#LL168-L171).

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/forecasting/sarimax.py
+++ b/sktime/forecasting/sarimax.py
@@ -6,6 +6,8 @@
 __all__ = ["SARIMAX"]
 __author__ = ["TNTran92"]
 
+import pandas as pd
+
 from sktime.forecasting.base.adapters import _StatsModelsAdapter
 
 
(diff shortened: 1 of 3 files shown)


sktime/forecasting/sarimax.py
```diff
@@ -6,6 +6,8 @@
 __all__ = ["SARIMAX"]
 __author__ = ["TNTran92"]
```
fkiraly, line 7: you should add yourself as an author.
(this is not an "owned" estimator, so this is by contribution history)

fkiraly (approved):
Great, as usual!

Non-blocking comments:

* you should add yourself as an author
* the `_predict_interval` looks very similar to that in other `statsmodels` functions. It might make sense to see whether the duplications (or partial duplications) can be refactored to live in a single place. I would consider that out of scope though for this PR, just making a note.

fkiraly:
Regarding questions:

> It can be considered to add handling of that

Yes, that would make sense.
If you look at that, it would be great if you also could add the missing `fit` args, see here: https://github.com/sktime/sktime/issues/4377

> skipped generalisation attempt as of now as suggested https://github.com/sktime/sktime/issues/4301#issuecomment-1493256817.

I suppose that's the same comment I just made in the review? Fair enough, so what you're saying, there are subtle interface inconsistencies?

fkiraly (changes requested):
Another thing I noticed, apologies for not letting you know first time round - I just fixed something similar as part of the `pandas 2` compatibility patch.

lines 242, 243 look identical to lines in `AutoETS`.
These cause a failure on `pandas 2` which no longer allows to write to index locations that do not exist.
Could you kindly follow the fix in https://github.com/sktime/sktime/pull/4427 (`AutoETS`) instead? (this replaces the offending lines by an alternative call that works on both `pandas 1` and `pandas 2`)

I also notice, the entire `_predict_interval` looks identical, except the name of the method of `prediction_results` which is called.
Could a refactor work by adding an attribute or tag to the adapter parent class which has the name of the attribute?
Would appreciate an issue with any observations you might have on similarities and/or how to refactor this (if you don't get to it, I will open one later when I'm done with the `pandas 2` upgrade)

yarnabrina:
> what you're saying, there are subtle interface inconsistencies?

Yes. For example, `DynamicFactor` has a method `predict_interval` that can be used, while `SARIMAX` does not and one has to use `get_prediction` or similar first and on top that use `conf_int`. I can't think of a way to merge these.

fkiraly:
> I can't think of a way to merge these.

Well, but there are at least three instances where the similarities are very high, see above.
We can deal with these, and override `_predict_interval` in cases of substantial difference, e.g., `DynamicFactor`.

fkiraly (approved):
Thanks, that's exactly what I meant!

PR #4442, opened 2023-04-06
[MNT] except `VECM` from `test_predict_quantiles` due to sporadic failures

fkiraly:
Excepts `VECM` from `test_predict_quantiles` due to sporadic failures of the current test (quantile prediction monotonocity), see https://github.com/sktime/sktime/issues/4431


fkiraly:
the failure is on main, therefore merging early.

PR #4443, opened 2023-04-07
[MNT] fix and sharpen soft dependency isolation logic for `statsmodels` and `pmdarima`

fkiraly:
This fixes and sharpens a number of soft dependency isolations.

* supersedes some incorrect soft isolation statements with the below. E.g., the run condition for some tests involving `ARIMA` was `statsmodels` present, while `ARIMA` requires `pmdarima`. This may not have been spotted since `pmdarima` requires `statsmodels`, and users of `statsmodels` might often have `pmdarima` installed.
* replaces all `pmdarima` isolation statements with `_check_estimator_deps` statements, which is clearer to read as it relates directly to the estimator.
* same for all statements affecting `ARDL`
* sharpens the dependency tag of `ARDL` from `statsmodels` to `statmodels>=0.13.0`


fkiraly:
0.17.1 scheduled for early release due to #4449, adding this too (dep isolation)

Issue #4449, opened 2023-04-09
[BUG] `attrs` import error under minimal dependencies


fkiraly:
CI and module lookup (e.g., `all_estimators`) breaks with an `attrs` import error because:

* it seems one of the core dependencies dropped `attrs` as a dependency, causing minimal dependency set of `sktime` to no longer require `attrs` (update: this is the developer core dependency `pytest`, with release 7.3.0)
* two of the experimental modules (in annotation) imported `attrs`, but `attrs` was not one of `sktime`'s direct requirements
* as a consequence, any attempted import of these modules now fails, in an environment with minimal dependency set

This will not impact users who have `attrs` installed or implied, but may break testing and lookup based logic for users whose requirements are small and do no longer imply `attrs`.

Possible fixes:

* fix 1: isolating `attrs`, turn into explicit soft dependency and manage via `sktime` dependency isolation mechanism
* fix 2: removing offending modules
* fix 3: make `attrs` a core dependency

Fix 3 would require a deprecation cycle and seems invasive.
Fix 2 may break dependencies on the experimental module, and will not solve the problem in general.

Fix 1 seems most involved and should not require deprecation. Why no deprecation: there is no change in public facing behaviour - and `attrs` becomes an explicit soft dependency, from being an implied (non-explicit) soft dependency.

fkiraly:
(I will try working on fix 1 - the two others are "trivial" in the sense that they can be executed very quickly)

PR #4450, opened 2023-04-09
[MNT] isolating `attrs` imports

fkiraly:
This PR isolates `attrs` imports and also fixes https://github.com/sktime/sktime/issues/4449, as "fix 1" proposed there.

This turns `attrs` from an implied (non-explict) soft dependency into an explicit soft dependency.
As the "de-facto" state of `attrs` as a soft dependency does not change, this does not require deprecation (instead, if removes the failure condition in #4449 from the status quo).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,6 +54,7 @@ dependencies = [
 
 [project.optional-dependencies]
 all_extras = [
+    "attrs",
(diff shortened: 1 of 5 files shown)


fkiraly:
fast merge due to issue already on main

PR #4463, opened 2023-04-12
[ENH] refactoring of `ForecastingHorizon` to use `Index` based `cutoff` in private methods

fkiraly:
This is a refactoring of `ForecastingHorizon` to use `Index` based `cutoff` in private methods.

One problem that occurs is that `lru_cache` can no longer be directly used, as `pd.Index` (which is now used instead of index element) is unhashable.
Not sure what the impact is on performance would be of dropping it.

I've solved it now by wrapping the `pd.Index` in a hashable `_HashIndex` class when passing through `lru_cache`. The `_HashIndex` has a `__hash__` dunder that uses the `pandas` hash methods internally.


fkiraly:
Sth funny is going on on python 3.7 and earlier `pandas` versions...

fkiraly:
FYI @yarnabrina as you have been looking at the `ForecastingHorizon` object recently.

PR #4452, opened 2023-04-09
[MNT] Changed line endings of `ElectricDevices.csv` and `GunPoint.csv` from `CRLF` to `LF`

yarnabrina:
#### Reference Issues/PRs

See also #4446 

#### What does this implement/fix? Explain your changes.

Most datasets in `sktime` have `LF` line endings, except `ElectricDevices.csv` and `GunPoint.csv`. This triggered auto-fix if configured in global gitattributes or global gitconfig. This PR attempts to convert those into `LF` line endings.

##### Before

```console
❯ git ls-files --eol | grep -E "*.csv"
i/lf    w/lf    attr/                   sktime/datasets/data/Airline/Airline.cs
i/lf    w/lf    attr/                   sktime/datasets/data/Longley/Longley.cs
i/lf    w/lf    attr/                   sktime/datasets/data/Lynx/Lynx.csv
i/lf    w/lf    attr/                   sktime/datasets/data/PBS_dataset/PBS_dataset.csv
i/lf    w/lf    attr/                   sktime/datasets/data/ShampooSales/ShampooSales.csv
i/lf    w/lf    attr/                   sktime/datasets/data/Uschange/Uschange.csv
i/mixed w/mixed attr/                   sktime/datasets/data/segmentation/ElectricDevices.csv
i/mixed w/mixed attr/                   sktime/datasets/data/segmentation/GunPoint.csv
i/lf    w/lf    attr/                   sktime/datasets/data/solar/solar.csv
```

##### After

```console
❯ git ls-files --eol | grep -E "*.csv"
i/lf    w/lf    attr/text               sktime/datasets/data/Airline/Airline.csv
i/lf    w/lf    attr/text               sktime/datasets/data/Longley/Longley.csv
i/lf    w/lf    attr/text               sktime/datasets/data/Lynx/Lynx.csv
i/lf    w/lf    attr/text               sktime/datasets/data/PBS_dataset/PBS_dataset.csv
i/lf    w/lf    attr/text               sktime/datasets/data/ShampooSales/ShampooSales.csv
i/lf    w/lf    attr/text               sktime/datasets/data/Uschange/Uschange.csv
i/lf    w/lf    attr/text               sktime/datasets/data/segmentation/ElectricDevices.csv
i/lf    w/lf    attr/text               sktime/datasets/data/segmentation/GunPoint.csv
i/lf    w/lf    attr/text               sktime/datasets/data/solar/solar.csv
```

#### Does your contribution introduce a new dependency? If yes, which one?

No

[...]

PR #4465, opened 2023-04-12
[ENH] adds `_predict_interval` in `_StatsModelsAdapter` and inherits in other estimator to reduce code duplication

### LINKED ISSUE #4447: [ENH] refactoring of `_predict_interval` inheritance in descendants of `_StatsModelsAdapter`
_Originally posted by @yarnabrina in https://github.com/sktime/sktime/issues/4301#issuecomment-1500896007_

@fkiraly I am planning to work on the refactoring of `_predict_interval` over the next week. This is my idea so far before starting:

1. add a `_predict_interval` to `_StatsModelsAdapter`
    1. similar to the common pattern in `ets`, `sarimax` and `structural`
    2. one extra abstract/non-implemented method to get quantile information from corresponding `PredictionResults` class, which seem to vary between above three
3. override that in subclasses:
    1. no override in classes for which `capability:pred_int` is `False`
    2. for estimators other than above the above three, override `_predict_interval` with current methods
    3. for above three estimators, implement the new method

Please let me know if it seems okay, or if you have some other suggestions.

I have few other doubts as well:

1. What can I expect at to be commit order to start with? Shall I start with `main`, or with #4439, or with #4427, or with both of these?
2. Is it safe to assume classes inheriting `_StatsModelsAdapter` are all the estimators in the scope of this refactoring? Or, can there be some others as well?
3. Are #4362 and #4363 related to this refactoring? Or they are indexing bugs in `_predict_interval`, but not affecting this?

yarnabrina:
#### Reference Issues/PRs

Closes #4447. Depends on #4439 and #4452.

The implementation of `_predict_interval` in `AutoETS`, `SARIMAX` and `UnobservedComponents` were very similar. This PR helps to define these at a single place at base level which should help maintaining easier.

No.

All but `ThetaForecaster` implements `_predict_interval` and hence that is failing the tests. However, the `_predict_quantiles` of `ThetaForecaster` just adds Gaussian quantiles to point predictions. If it is a proper way to do interval prediction, a specific exception can be made for `ThetaForecaster` in the base method. If not, subject to deprecation, can it be considered to remove interval support for `ThetaForecaster`?

No.

#### Any other comments?

Observations on multivariate estimators as being discussed in #4447.

##### For all contributions

##### For new estimators


sktime/forecasting/base/adapters/_statsmodels.py
```diff
@@ -123,6 +123,88 @@ def _predict(self, fh, X=None):
...
+                Third level is string "lower" or "upper", for lower/upper interval end.
+            Row index is fh, with additional (upper) levels equal to instance levels,
+                from y seen in fit, if y_inner_mtype is Panel or Hierarchical.
+            Entries are forecasts of lower/upper interval end,
+                for var in col index, at nominal coverage in second col index,
+                lower/upper depending on third col index, for the row index.
+                Upper/lower interval end forecasts are equivalent to
+                quantile forecasts at alpha = 0.5 - c/2, 0.5 + c/2 for c in coverage.
+        """
+        if not self.get_tag("capability:pred_int"):
```
fkiraly, line 177: why do you have this in the adapter? This is already in `BaseForecaster.predict_interval`.

sktime/forecasting/ets.py
```diff
@@ -416,63 +416,26 @@ def _predict(self, fh, X=None, **simulate_kwargs):
...
-            The forecasting horizon with the steps ahead to to predict.
-        X : optional (default=None)
-            guaranteed to be of a type in self.get_tag("X_inner_mtype")
-            Exogeneous time series to predict from.
-        coverage : list of float (guaranteed not None and floats in [0,1] interval)
-           nominal coverage(s) of predictive interval(s)
+        prediction_results : PredictionResults
+            results class for predictions
+        alpha : float
+            level of significance
```
fkiraly, line 428: for confidence intervals, should this be "nominal coverage"? Or is it "one minus nominal coverage"?
"level of significance" I think is not accurate, as we are not conducting a frequentist hypothesis test here.

sktime/forecasting/base/adapters/_statsmodels.py
```diff
@@ -123,6 +123,88 @@ def _predict(self, fh, X=None):
...
         return y_pred
 
+    @staticmethod
+    def _extract_conf_int(prediction_results, alpha) -> pd.DataFrame:
+        """Construct confidence interval at specified `alpha` for each timestep.
+
+        Parameters
+        ----------
+        prediction_results : PredictionResults
+            results class for predictions
```
fkiraly, line 133: I would be more specific here, "as returned by `self._fitted_forecaster.get_prediction`" ?

fkiraly (commented):
Overall, design-wise, looks good!

Left some minor comments.

Also, you already noted the fiddly task of finding out what happens downstream with the failures and `Theta`.

sktime/forecasting/ets.py
yarnabrina: This is the [documentation](https://www.statsmodels.org/dev/generated/statsmodels.tsa.base.prediction.PredictionResults.conf_int.html#statsmodels.tsa.base.prediction.PredictionResults.conf_int) for the underlying `statsmodels` class, didn't want to change:

> alpha[float](https://docs.python.org/3/library/functions.html#float), optional
The significance level for the prediction interval. The default alpha = .05 returns a 95% confidence interval.

sktime/forecasting/base/adapters/_statsmodels.py
yarnabrina: I checked in `BaseForecaster._predict_interval` by mistake, and found no check. Removed.

sktime/forecasting/base/adapters/_statsmodels.py
yarnabrina: Done.

yarnabrina:
Current failures:

1. The new base method expects all subclasses to use `_predict_interval` to generate probabilistic forecasts. Since `ThetaForecaster` (inheriting from `ExponentialSmoothing`) implements `_predict_quantiles`, it is causing issues.
    * temporarily remove predict proba support from `ThetaForecaster` -> will break significantly, not an option
    * change predict proba support for `ThetaForecaster` by replacing quantiles with interval ([reference](https://github.com/sktime/sktime/issues/4447#issuecomment-1509586502)) - effort/effects?
    * how to tackle other estimators inheriting from `_StatsModelsAdapter` relying on quantiles instead of interval?
2. There is a test (`test_pred_int_tag`) that ensures absence of estimators with implementation of `_predict_interval` or `_predict_quantiles` or `_predict_proba` and without `capability:pred_int` tag. Base method in `_StatsModelsAdapter` is causing failures in all inheriting estimators without predict proba support to fail.
    * skip this test - effects?
    * modify the test to allow for abstract methods - how?

This is the current status, and planning on how to handle these. Any comments/suggestions are very much welcome.

(FYI: @fkiraly)

fkiraly:
1. this is interesting, didn't think of this.

I can see option 2 as well - that might be the cleaner option long term, as all `statsmodels` models return prediction intervals afaik?

However, this seems like scope creep and unclear effort, so I would go for the following strategy:

* aim for an end state where we have a pair of methods for `_predict_interval` and `_predict_quantiles`, i.e., `_predict_quantiles` and `_predict_quantiles_adapt` (does not exist), or `_predict_interval` and `_predict_interval_adapt` (currently called `_extract_conf_int`)
* use `_has_implementation_of` - a base method present in all `skbase` objects - to check which of the adapter pair (`_predict_interval_adapt`, `_predict_quantiles_adapt`) has been implemented, in `_predict_interval`, and `_predict_quantiles` both. If the other one has been implemented, default to the parent `_predict_interval`. If the same has been implemented, run the adapter logic
* for the start, only cover the `_predict_interval` side, at least in this PR. `_predict_quantiles` might not be worth dealing with.

What do you think?

fkiraly:
2. the test `test_pred_int_tag` already has one exception, the `_DelegatedForecaster`. The simplest way would be to skip the impacted forecasters the same way, but that weakens test coverage.

A slightly more elaborate way would be, changing the test so it checks whether the *functionality* is there rather than an *implementation*. That would require a config option which turns off the error message and lets the method run. But I consider this scope creep for this PR, so out of scope.

yarnabrina:
Agreed on out-of-scope point on both 1 and 2. Then, I'm thinking to skip the tag test for now, and mark as a TODO comment + GH issue to be addressed later (hopefully soon). Please let me know if that's not okay.

---

Is it possible to redirect to parent's `_predict_interval` if we are not able to check if `extract_conf_int` is implemented AND "functional"? There's a very high chance that I misinterpreted your suggestion, so can you please elaborate a little bit more?

fkiraly:
> Is it possible to redirect to parent's `_predict_interval` if we are not able to check if `extract_conf_int` is implemented AND "functional"? There's a very high chance that I misinterpreted your suggestion, so can you please elaborate a little bit more?

`_has_implementation_of` should return `True` if and only if the method has been overridden at least once. That should be an if and only if with "functional", right?

It's important to note the difference between "implemented" (appears at least once in the inheritance resolution order), and "overridden at least once" (appears at least twice). `_has_implementation_of` checks the latter.

yarnabrina:
> That should be an if and only if with "functional", right?

Not sure I'm certain on this. Say the adapt example you gave, basically my implementation of `_predict_interval` in `_StatsModelsAdapter`. I'd have said it's non-functional as it can lead to `NotImplementedError`.

I'll give it a try tomorrow, but I've a feeling that `super()` may not be enough (or at least for me). Will update.

fkiraly:
> Not sure I'm certain on this. Say the adapt example you gave, basically my implementation of `_predict_interval` in `_StatsModelsAdapter`. I'd have said it's non-functional as it can lead to `NotImplementedError`.

Ahh, I see - what I meant, you check the `extract_conf_int` with `_has_implementation_of`, not the `_predict_interval`! If that is overridden at least once, it is functional. If I understand the design correctly?

yarnabrina:
> Ahh, I see - what I meant, you check the `extract_conf_int` with `_has_implementation_of`, not the `_predict_interval`! If that is overridden at least once, it is functional. If I understand the design correctly?

As of now, definitely yes. But I suppose theoretically it's possible that in future one can inherit `_StatsModelsAdapter` and have a template `extract_conf_int` to be overridden by it's child classes afterwards. Admittedly a very stretched scenario, but possible I think.

---

As of now, I believe I managed to handle the first issue using your suggestion of dispatch to other methods. As I guessed last night, I failed to take benefit of `super`, as all inherited estimators (with no restriction on how low/high they are on hierarchy) need to call the method present in `BaseForecaster`. I'll highly appreciate a better way, probably with a smart `super` call, to replace `L181`.

https://github.com/yarnabrina/sktime-fork/blob/caadd66f083d6ec44c432b4323e66a9469aa3ee9/sktime/forecasting/base/adapters/_statsmodels.py#L177-L181

All tests on `test_predict_interval`and `test_predict_quantiles` pass locally, and hopefully CI will report just tag failures. If that's the case, will add to `_config.py` for now referring to all failing `statsmodels` based estimators without the tag.

fkiraly:
> CI will report just tag failures.

I would add it as an explicit exception in the test, i.e., pass the test if `_has_implementation_of ("extract_conf_int")` (and add a comment where it is coming from, i.e., the statsmodels adapter)

I would also recommend, let's open an issue to review the test, and test the tag against actual functionality (if called) rather than `_has_implementation_of`.

sktime/forecasting/ets.py
yarnabrina: I really wonder what proportion of users will know/care about the distinction between the two. Admittedly, visualising the difference between ranges of future given past and ranges of expected values of future given past is difficult for me (even assuming I am doing it correctly.)

But I don't think that is a reason to affect the docstring, especially because it's a private method. Updated with your suggestion.

yarnabrina:
Yes, it's ready to review.

If it's complete and merged, will start on #4485 afterwards.

fkiraly (changes requested):
Really great, thanks!

Seeing this, I would do the following:

* in the "returns" of the docstring of `_extract_conf_int`, be crystal clear what the implementation contract is. I.e. that it should return a `DataFrame` with two columns `"lower"`, `"upper"`, in this order, and row index should be integer, relative to cutoff, and should contain the relative integer horizon of `fh` ( and can be strictly larger)
* I would implement the following default logic for `_extract_conf_int`: if there is a unique method ending in `_int` and having argument `alpha`, use that

fkiraly:
> I would implement the following default logic 

that is, I would suggest `[m for m in dir(prediction_results) if m.endswith("_int")`, check that length equals 1 and has `alpha` as arg (via `inspect.fullargspec`), otherwise raise error

fkiraly:
none blockers for now, but for consideration. Will merge tomorrow in this state if no change.

yarnabrina:
Updated the docstring as suggested, please take a look.

I'm not comfortable adding default method based on patterns, skipped it for now. Even if we add it, it's not going to be used in inherited classes unless someone specifically enabled the `capability:pred_int` tag. And I'm personally of the opinion that it's significantly better to explicitly use a method in inherited classes.

fkiraly:
ok, makes sense.

Issue #4468, opened 2023-04-13
[BUG] sktime 0.17.1 breaks in-sample `predict_quantiles` in `pmdarima` adapter


ngupta23:
**Describe the bug**
The latest release of sktime breaks in-sample predictions for ARIMA models

**To Reproduce**

```python
from sktime.datasets import load_airline
from sktime.forecasting.arima import ARIMA

y = load_airline()
forecaster = ARIMA(order=(1, 1, 0), seasonal_order=(0, 1, 0, 12), suppress_warnings=True)
forecaster.fit(y)
forecaster.predict_quantiles(fh=y.index, X=None, alpha=[0.05, 0.95])
```

```python-traceback
'numpy.ndarray' object has no attribute 'to_pandas'
  File "C:\Users\Nikhil\.conda\envs\pycaret_dev_sktime_17p1\Lib\site-packages\sktime\forecasting\base\adapters\_pmdarima.py", line 169, in _predict_in_sample
    index=fh_abs.to_pandas(), columns=["lower", "upper"]
  File "C:\Users\Nikhil\.conda\envs\pycaret_dev_sktime_17p1\Lib\site-packages\sktime\forecasting\base\adapters\_pmdarima.py", line 293, in _predict_interval
    _, y_pred_int = self._predict_in_sample(fh_ins, **kwargs)
  File "C:\Users\Nikhil\.conda\envs\pycaret_dev_sktime_17p1\Lib\site-packages\sktime\forecasting\base\_base.py", line 2079, in _predict_quantiles
    pred_a = self._predict_interval(fh=fh, X=X, coverage=[coverage])
  File "C:\Users\Nikhil\.conda\envs\pycaret_dev_sktime_17p1\Lib\site-packages\sktime\forecasting\base\_base.py", line 559, in predict_quantiles
    quantiles = self._predict_quantiles(fh=fh, X=X_inner, alpha=alpha)
  File "C:\Users\Nikhil\OneDrive\my_libraries\my_python_libraries\pycaret\pycaret_dev\time_series_debug.py", line 28, in <module>
    forecaster.predict_quantiles(fh=y.index, X=None, alpha=[0.05, 0.95])
  File "C:\Users\Nikhil\.conda\envs\pycaret_dev_sktime_17p1\Lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
[...]

PR #4470, opened 2023-04-13
[ENH] testing all forecasters `predict_quantiles`, `predict_interval` in-sample

fkiraly:
Bug https://github.com/sktime/sktime/issues/4468 surfaced that probabilistic forecast methods were not tested in-sample.

This PR adds coverage for probabilistic forecast methods `predict_quantiles`, `predict_interval` used in-sample.

This should also surface #4468 in 0.17.1, and pass on 0.17.0 or post-#4469.

Depends on:
* https://github.com/sktime/sktime/pull/4472 to ensure we test all estimators at the same time
* https://github.com/sktime/sktime/pull/4476 for cleaning up the `test_predict_time_index` test
* fixes in https://github.com/sktime/sktime/pull/4469
* fixes in https://github.com/sktime/sktime/pull/4505
* https://github.com/sktime/sktime/pull/4506


fkiraly:
@yarnabrina, would appreciate your review since you are working on this (or in close vicinity)

fkiraly:
failures coming from:

* `SquaringResiduals`
* `ForecastX`
* `VECM`

sktime/forecasting/tests/test_all_forecasters.py
```diff
@@ -230,6 +230,12 @@ def test_predict_time_index(
             _assert_correct_pred_time_index(y_pred.index, cutoff, fh=fh_int)
             _assert_correct_columns(y_pred, y_train)
 
+            if estimator_instance.get_tag("capability:pred_int"):
```
yarnabrina, line 233: clarification: if this check is specifically made, what can cause `NotImplementedError`?

sktime/forecasting/tests/test_all_forecasters.py
```diff
@@ -230,6 +230,12 @@ def test_predict_time_index(
...
+                y_pred_int = estimator_instance.predict_interval()
+                _assert_correct_pred_time_index(y_pred_int.index, cutoff, fh=fh_int)
+                y_pred_q = estimator_instance.predict_quantiles()
+                _assert_correct_pred_time_index(y_pred_q.index, cutoff, fh=fh_int)
+
```
yarnabrina, line 238: Consider addition of `predit_proba`?

fkiraly: hm, yes. I thought the same but hesitated doing this here due to the current `tensorflow` dependency of the method.
Maybe splitting the blocks smartly and try/except would be an option.

sktime/forecasting/tests/test_all_forecasters.py
fkiraly: I thought something about in-sample or out-of-sample.

But you are right the try/except is a block with an unclear condition, thinking carefully, I do not know for sure why it is there!

So, let's see what happens when we remove it:
https://github.com/sktime/sktime/pull/4476
If it has a good reason to break in cases on `main`, we should probably replace it with a new tag.

sktime/forecasting/tests/test_all_forecasters.py
fkiraly: the same is actually tested in `test_predict_proba` already.

sktime/forecasting/tests/test_all_forecasters.py
fkiraly: I assume we consider this addressed by #4476

PR #4476, opened 2023-04-15
[ENH] testing in-sample forecasting - replace try/except in `test_predict_time_index` by tag and tag dependent contract

fkiraly:
This PR adds tags for in-sample prediction capability - one for point prediction and one for probabilistic prediction.
This changes the implicit contract (can raise `NotImplementedError`) to a queriable, inspectable contract, via the tags.

Also replaces the try/except in `test_predict_time_index` with a conditional based on the tags, to sharpen the coverage of estimators' in-sample forecasting capability, based on explicit tagging.

The following tags are added to forecasters and set appropriately for all forecasters:

* `capability:insample`
* `capability:pred_int:insample`


fkiraly:
FYI @yarnabrina, it looks like the error is raised by some forecasters that cannot make in-sample forecasts.
Looks like a non-explicit convention that some developers have introduced.

I will translate this to a tag and a clear contract condition.

yarnabrina:
Makes sense. Noting one more case of `NotImplementedError` in `test_predict_proba`, which definitely looks related.

https://github.com/sktime/sktime/blob/4ffd8d4ca8b71db4875eb9bf662b81939ba0287a/sktime/forecasting/tests/test_all_forecasters.py#L505-L510

There's a lot of try-except in this module, sometimes even with `predict` method. Are these really necessary? e.g. L272, L301.

fkiraly:
> Are these really necessary?

Probably not, and probably they are dangerous...

When we've moved to test classes, I've mostly taken the code that was there 1:1.
I'd rather not have any try/except in the tests, and instead control explicitly for the contract condition - I would even go so far and insist that if there are contract variations, they should be tag inspectable.
(of course that may be hard, depending on what the conditions actually turn out to be in the try/except)

fkiraly:
On the specific case in `test_predict_proba`, this cannot relate to in-sample/out-of-sample, since all `fh` tested in there are out of sample.

Issue #4359, opened 2023-03-18
[ENH] Umbrella issue - distribution forecast rework


fkiraly:
This is an umbrella issue for the roadmap to build a `BaseObject`-based interface for probability distributions, which is back-end agnostic, but has `tensorflow-probability` as one of the options for a back-end.

The conceptual model mixes indexed objects as in `pandas` and the array-based distributions in `tensorflow-probability`, i.e., array distributions which have `pandas`-based `index`, `columns`, and can be sub-set using `loc` and `iloc` indexing, similar to `pd.DataFrame`.

Advantages:
* row/column subsetting of the return, compatibility with `pandas` indices
* decouples the distributions from the `tensorflow` back-end, foundation for distribution metrics/losses
* allows to add methods to the distribution outside `tensorflow`, in `sktime`, e.g., heuristical integrated cdf or energy statistics required for CRPS metric (see `energy` method)

STEP: https://github.com/sktime/enhancement-proposals/pull/31
Related issue: https://github.com/sktime/sktime/issues/1746

Roadmap:

* part 1: distribution objects, https://github.com/sktime/sktime/pull/4190
* part 2: probabilistic (distributional) metrics, https://github.com/sktime/sktime/pull/4276
* part 3: use in forecasters as return object if `tensorflow-probability` is not present; deprecation for 0.18.0 or 0.19.0. Only "simple" forecast, no vectorization support
* part 4: use in tuned forecasters
* further probability distributions native to `sktime`
* interfacing more distributions from `tfp`
* vectorization across variables and hierarchy levels

fkiraly:
@frthjf on #855, answering here to keep the topicality

> I see, in that case, why move this out of sktime into an independent skpro package if it's required back in anyway? Wouldn't it be easier to port the skpro features into sktime instead? It seems to me that the sktime package already has all the CI and package infrastructure we would have to recreate for a skpro re-release.

fkiraly:
The move is due to scope/architecture standpoint:

* `sktime` covers time series related tasks, and this is just one place where probability distributions might be useful. I'm just planning 2-5 years ahead :-)
* the infrastructure and testing is indeed easier in `sktime` currently, but that will move as a template into `skbase`. So, `skpro` can just adopt the `skbase` template for CI and testing and it will be much less work than a full setup from zero!
[...]

Issue #2473, opened 2022-04-14
[BUG] Windows fatal exception: access violation sktime/benchmarking/tests/test_evaluator.py::test_plots


k1m190r:
**Describe the bug**
`sktime/benchmarking/tests/test_evaluator.py::test_plots` fails with  `Windows fatal exception: access violation`

The line where error surfaces is:
`sktime\benchmarking\evaluation.py::plot_boxplots#145`
`fig, ax = plt.subplots(1)`

Solution from https://github.com/matplotlib/matplotlib/issues/22006
downgrade freetype from conda to 2.10.4

`conda install freetype=2.10.4`

**To Reproduce**
````
pytest sktime/benchmarking/tests/test_evaluator.py::test_plots

sktime/benchmarking/tests/test_evaluator.py::test_plots Windows fatal exception: access violation
````

**Expected behavior**
Test passes

**Additional context**
freetype=2.11.0 (anaconda)

**Versions**
Windows, python 3.{7 8 9}

fkiraly:
is this new??

fkiraly:
Is this related to this one:
https://github.com/alan-turing-institute/sktime/issues/2368

k1m190r:
Yes new using current main. I'll check on Mac. I'm not sure if related...

fkiraly:
we should leave it open until resolved

k1m190r:
Oh you mean till it is resolved in matplotlib/freetype?

fkiraly:
yes, imo we should track this until (a) we understand what the error is, and (b) it is explained here, and (c) it is fixed/resolved (upstream or in sktime wherever it needs to be)

PR #4481, opened 2023-04-16
[ENH] remove superfluous implementation checks in `_predict_interval` and `_predict_quantiles` of `BaseForecaster`

yarnabrina:
In the existing implementation of `_predict_interval` in `BaseForecaster`, first it checks whether any of `_predict_interval`, `_predict_quantiles` or `_predict_proba` is overridden or not. But the check on `_predict_interval` would never evaluate to `True` when called from an inherited forecaster, as if it is indeed overridden, that overridden method would have been called, and not that of the `BaseForecaster`.

(Edit: stroked out no-more-affected areas by PR, because of update as suggested below)

~~On the other hand, even if `_predict_proba` is implemented, as of current implementation it is not used to generate prediction interval in case `_predict_interval` is not implemented. Plus presence of this check can possibly lead to `NameError` from `return pred_int` line, in case `_predict_quantiles` is not implemented.~~

~~Because of the above reasons, the check with `can_do_proba` inside `_predict_interval` is equivalent to check with `implements_quantiles`, and that can simplify the next indented block by avoiding the check because of `raise` before this.~~

Similar arguments of can be made for `_predict_quantiles` in `BaseForecaster` as well, where using `_predict_interval` is sufficient.

This PR attempts to solve the above arguments by removing these unnecessary checks.

### DIFF
--- a/sktime/forecasting/base/_base.py
+++ b/sktime/forecasting/base/_base.py
@@ -2055,9 +2055,8 @@ class BaseForecaster(BaseEstimator):
                 at quantile probability in second col index, for the row index.
         """
         implements_interval = self._has_implementation_of("_predict_interval")
-        implements_quantiles = self._has_implementation_of("_predict_quantiles")
         implements_proba = self._has_implementation_of("_predict_proba")
-        can_do_proba = implements_interval or implements_quantiles or implements_proba
+        can_do_proba = implements_interval or implements_proba
 
         if not can_do_proba:
             raise RuntimeError(
--- a/sktime/forecasting/base/_base.py
+++ b/sktime/forecasting/base/_base.py
@@ -1986,10 +1986,9 @@ class BaseForecaster(BaseEstimator):
                 Upper/lower interval end forecasts are equivalent to
                 quantile forecasts at alpha = 0.5 - c/2, 0.5 + c/2 for c in coverage.
         """
-        implements_interval = self._has_implementation_of("_predict_interval")
         implements_quantiles = self._has_implementation_of("_predict_quantiles")
         implements_proba = self._has_implementation_of("_predict_proba")
-        can_do_proba = implements_interval or implements_quantiles or implements_proba
+        can_do_proba = implements_quantiles or implements_proba
 
         if not can_do_proba:
             raise RuntimeError(


fkiraly (changes requested):
Ah, apologies, the `implements_proba` is indeed unnecessary currently, but we were planning to add that, and a sensible default, based on the new probabilistic interface (for 0.18.0 and 0.19.0).

See https://github.com/sktime/sktime/issues/4359 and https://github.com/sktime/sktime/pull/4290
(if you're interested to join the workstream, send a message on discord)

So, in a sense, `implements_proba` it is a scaffolding that I'd prefer you leave there.

On the other hand, checking whether it implements itself is indeed superfluous, as it will never evaluate to `True` (then it would be overridden).

fkiraly:
looks good!

Design question, since I'm looking at this PR: would it perhaps makes sense to factor this out into a private utility function `_get_proba_implemented` which returns a `dict` of bool, and this is called in both places (and possible future places), and retrieves all the `_has_implemented` results, for all relevant methods?

Or is this overkill?

fkiraly (approved):
I'd be happy with this - see my comment about a potential alternative design.

yarnabrina:
> looks good!
> 
> Design question, since I'm looking at this PR: would it perhaps makes sense to factor this out into a private utility function `_get_proba_implemented` which returns a `dict` of bool, and this is called in both places (and possible future places), and retrieves all the `_has_implemented` results, for all relevant methods?
> 
> Or is this overkill?

If it's only from DRY principle, it makes sense, as it's already being used at least thrice (twice in base, once in tag test). But I think actually all these different tests can be made more specific, so probably I feel good with as it is currently.

yarnabrina:
@fkiraly any suggestion about the `MemoryError`?

fkiraly:
> any suggestion about the MemoryError?

I don't know what that is, it sometimes happens sporadically, perhaps from `pytest-xdist`.
Generally, this is unrelated to any PR (usually). Not sure what to do here. Perhaps tracking it might be a good idea.

Earlier instances seemed to occur around the plotting functions:
https://github.com/sktime/sktime/issues/2473

PR #4500, opened 2023-04-21
[ENH] seasonal forecast wrapper to use any forecaster as seasonal forecaster

fkiraly:
This PR introduces a seasonal forecast wrapper, which can turn any forecaster into a seasonal forecaster by applying it to a seasonally tabulated multivariate series. This is a form of reduction to a multivariate forecaster.

Still work in progress as it requires a version of the tabulation utility with pivot element at the end, not the start - just pushing it so it does not get lost while I work on other things next week.

Depends on https://github.com/sktime/sktime/pull/4490 for the seasonal tabulation utility.


benHeid:
Mhm I am not sure if the solution from #5891 would make this easier. I think the difference would be that the time series is not transformed into pivot table. With grouping it could be transformed into a multi-index time series. Thus, it would require that there is a forecaster trained for each group. 
Thus, I assume that the grouping would create Panels. However, the pivot_table would create a multivariate time series. So grouping might be better than pivots if panels are easier to handle than multivariate time series in this case. What is your feeling on panels vs multivariate forecasting?

PR #4502, opened 2023-04-22
[BUG] fix broken `DynamicFactor._predict_interval`, part 2

fkiraly:
Some missing commits from https://github.com/sktime/sktime/pull/4479 which were not spotted due to merge order.

This should fix the failures in https://github.com/sktime/sktime/pull/4470.

Depends on https://github.com/sktime/sktime/pull/4476


fkiraly:
ok, it seems there was no merge accident in `DynamicFactor` after all. Probably did a diff with a wrong base branch.

PR #4512, opened 2023-04-27
[ENH] add test for `all_estimators` tag filter

fkiraly:
This adds a test for the tag filter parameter of `all_estimators`, i.e., that only the filtered estimators are returned, and the filtered estimators are returned.


fkiraly:
adds safety to release action `skbase`, merging early

PR #4527, opened 2023-04-29
&achieveordie [BUG] fix bug from clash between `ABC` inheritance and RNN `fit` override

fkiraly:
By accident, we merged https://github.com/sktime/sktime/pull/4185 without re-running the tests - this masked a bug. Which is not too grave as it is post-release, and the release CI has all passed.

Details: on the new `main`, the merge caused a clash with the `ABC` inheritance in `BaseRegressor` as one of the methods was not overridden. Also, `fit` should be `_fit`, which is also fixed.

### DIFF
--- a/sktime/classification/deep_learning/rnn.py
+++ b/sktime/classification/deep_learning/rnn.py
@@ -52,10 +52,7 @@ class SimpleRNNClassifier(BaseDeepClassifier):
         self.use_bias = use_bias
         self.optimizer = optimizer
         self.history = None
-        self._network = RNNNetwork(
-            batch_size=self.batch_size,
-            units=self.units,
-        )
+        self._network = RNNNetwork(random_state=random_state, units=units)
 
     def build_model(self, input_shape, **kwargs):
         """
--- a/sktime/classification/deep_learning/rnn.py
+++ b/sktime/classification/deep_learning/rnn.py
@@ -94,7 +94,7 @@ class SimpleRNNClassifier(BaseDeepClassifier):
         model.compile(loss=self.loss, optimizer=self.optimizer_, metrics=metrics)
         return model
 
-    def fit(self, X, y, input_checks=True):
+    def _fit(self, X, y):
         """
         Fit the regressor on the training set (X, y).
 
@@ -106,8 +106,6 @@ class SimpleRNNClassifier(BaseDeepClassifier):
             n_dimensions is assumed to be 1.
         y : array-like, shape = [n_instances]
             The training data class labels.
-        input_checks : boolean
-            whether to check the X and y parameters
 
         Returns
         -------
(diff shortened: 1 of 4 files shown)


fkiraly:
@achieveordie, any idea what is going wrong here?

achieveordie:
It is most likely due to [line 39 in `RNNNetwork`](https://github.com/sktime/sktime/blob/8338a2a513116a3d1e97403ffcadf7ba50648bf8/sktime/networks/rnn.py#L39):
```python
input_layer = keras.layers.Input((input_shape, 1)
```

It should be either be `keras.layers.Input(input_shape)` or `keras.layers.Input((*input_shape, 1))` depending on the input. Currently the code incorrectly adds 1 as the last dimension regardless if the input_shape is dimension, more concretely:

```python
    def build_network(self, input_shape, **kwargs):
        if len(input_shape) == 2:                                             # eg: (5, 10)
            input_layer = keras.layers.Input(input_shape)         # (_, 5, 10)
        elif len(input_shape) == 1:                                          # eg: (5, )
            input_layer = keras.layers.Input(*input_shape, 1)   # (_, 5, 1)
        else:
            raise Error
```

achieveordie:
I'm not sure why are we doing [this](https://github.com/sktime/sktime/blob/8338a2a513116a3d1e97403ffcadf7ba50648bf8/sktime/regression/deep_learning/rnn.py#L114) either:
```python
self.batch_size = int(max(1, min(X.shape[0] / 10, self.batch_size)))
```

This doesn't seem to be the standard, what requires this line?

achieveordie:
I'm taking the liberty of fixing this issue here, I hope that's okay.

achieveordie:
Hmm, it seems like I can't directly push to this branch since it's upstream.
Here are my modifications for `networks/rnn.py/RNNNetwork`

```python
    def build_network(self, input_shape, **kwargs):
        """Construct a compiled, un-trained, keras model that is ready for training.

        Parameters
        ----------
        input_shape : int or tuple
            The shape of the data fed into the input layer. It should either
            have dimensions of (m, d) or m. In case an int is passed,
            1 is appended for d.

        Returns
        -------
        output : a compiled Keras Model
        """
        from tensorflow import keras

        if isinstance(input_shape, int):
            input_layer = keras.layers.Input((input_shape, 1))
        elif isinstance(input_shape, tuple):
            if len(input_shape) == 2:
                input_layer = keras.layers.Input(input_shape)
            elif len(input_shape) == 1:
                input_layer = keras.layers.Input((*input_shape, 1))
            else:
                raise ValueError(
                    "If `input_shape` is a tuple, it must either be "
                    f"of length 1 or 2. Found length of be {len(input_shape)}"
                )
        else:
            raise TypeError(
                "`input_shape` should either be of type int or tuple. "
                f"But found the type to be: {type(input_shape)}"
            )

        rnn = keras.layers.SimpleRNN(units=self.units, return_sequences=True)(
            input_layer
        )

        dense = keras.layers.Dense(units=1)(rnn)

        return input_layer, dense
```

fkiraly:
> Hmm, it seems like I can't directly push to this branch since it's upstream.

Odd, you have write access to the repo.

fkiraly:
I've applied your fix now. You should have been able to push it though.

Issue #4528, opened 2023-04-29
[ENH] review probabilistic methods default dispatch logic


fkiraly:
_Originally posted by @yarnabrina in https://github.com/sktime/sktime/issues/4513#issuecomment-1528621231_

(moving this to separate issue)

> Do you think we should do this differently? Or, same logic, but implement it differently?

I completely agree with the logic, but I think I'm missing how `predict_interval` would work if both `_predict_interval` and `_predict_quantiles` are undefined (in the sense of not overridden at estimator level) and only `_predict_proba` is overridden.

If actual (non-base) estimator do not define `_predict_interval` or `_predict_quantiles`, the checks with `_has_implementation_of` will be `False`, but `_predict_proba` implementation will make `can_do_proba` as `True`. So, it goes pass [L1995](https://github.com/sktime/sktime/blob/195e517667a8ad69c65eec9737cc9160007f477e/sktime/forecasting/base/_base.py#L1995), and also goes pass[L2003](https://github.com/sktime/sktime/blob/195e517667a8ad69c65eec9737cc9160007f477e/sktime/forecasting/base/_base.py#L2003). My question/doubt is how will [L2030](https://github.com/sktime/sktime/blob/195e517667a8ad69c65eec9737cc9160007f477e/sktime/forecasting/base/_base.py#L2030) work then, because I don't see `pred_int` being defined anywhere else in the function.

Do we not need something like this here in `_predict_interval` as well? Of course the logic needs to be a bit different to conform to the output format.

https://github.com/sktime/sktime/blob/195e517667a8ad69c65eec9737cc9160007f477e/sktime/forecasting/base/_base.py#L2106

If I have missed some logic which will call `_predict_quantiles` of `BaseForecaster` from `_predict_interval` of `BaseForecaster` to make use of `_predict_proba` in actual estimator in such cases, please let me know.

fkiraly:
> If I have missed some logic which will call `_predict_quantiles` of `BaseForecaster` from `_predict_interval` of `BaseForecaster` to make use of `_predict_proba` in actual estimator in such cases, please let me know.

Ok, that's case 2 in this post, right?
https://github.com/sktime/sktime/pull/4513#issuecomment-1527459579
What you're saying, you think in this case the dispatch might not work.

Here's how I understand it will work (in the case without vectorization):

1. `predict_interval` is called
2. this enters `_predict_interval` in line 638
3. no concrete implementation is present, so we enter the `BaseForecaster._predict_interval` as a default
[...]

PR #4529, opened 2023-04-29
[ENH] corrected default logic for `_predict_interval` in case `_predict_quantiles` is not implemented but `_predict_proba` is

fkiraly:
As @yarnabrina pointed out in #4528, `predict_interval` would break in the case where neither `_predict_interval` nor `_predict_quantiles` are implemented but `_predict_proba` is.

In this case, `_predict_interval` would run into an exception, since no dispatch would be taking place.

This PR fixes this so that `_predict_interval` dispatches to `_predict_quantiles`, which in turn will dispatch to `_predict_proba`.

### DIFF
--- a/sktime/forecasting/base/_base.py
+++ b/sktime/forecasting/base/_base.py
@@ -2000,32 +2000,31 @@ class BaseForecaster(BaseEstimator):
 
         # we default to _predict_quantiles if that is implemented or _predict_proba
         # since _predict_quantiles will default to _predict_proba if it is not
-        if implements_quantiles or implements_proba:
-            alphas = []
-            for c in coverage:
-                # compute quantiles corresponding to prediction interval coverage
-                #  this uses symmetric predictive intervals
-                alphas.extend([0.5 - 0.5 * float(c), 0.5 + 0.5 * float(c)])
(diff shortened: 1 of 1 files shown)


sktime/forecasting/base/_base.py
```diff
@@ -1998,7 +1998,9 @@ def _predict_interval(self, fh, X=None, coverage=0.90):
                 "This is likely a bug, please report, and/or set the flag to False."
             )
 
-        if implements_quantiles:
+        # we default to _predict_quantiles if that is implemented or _predict_proba
+        # since _predict_quantiles will default to _predict_proba if it is not
+        if implements_quantiles or implements_proba:
```
yarnabrina, line 2003: I think that the check with `can_do_proba`in L1993 and the following `raise` should ensure that check in L2003 will always pass. So it can be skipped, and the following section can be de-indented.

yarnabrina (changes requested):

sktime/forecasting/base/_base.py
fkiraly: you are right - changed

yarnabrina (approved):

PR #4531, opened 2023-04-30
achievordie [BUG] fix RNN classifier and regressor

fkiraly:
This PR turns on the tests for RNN classifier and regressor, post https://github.com/sktime/sktime/pull/4530

### DIFF
--- a/sktime/classification/deep_learning/rnn.py
+++ b/sktime/classification/deep_learning/rnn.py
@@ -32,6 +32,7 @@ class SimpleRNNClassifier(BaseDeepClassifier):
         batch_size=1,
         units=6,
         callbacks=None,
+        add_default_callback=True,
         random_state=0,
         verbose=False,
         loss="mean_squared_error",
@@ -47,6 +48,7 @@ class SimpleRNNClassifier(BaseDeepClassifier):
         self.verbose = verbose
         self.units = units
         self.callbacks = callbacks
+        self.add_default_callback = add_default_callback
         self.random_state = random_state
         self.loss = loss
         self.metrics = metrics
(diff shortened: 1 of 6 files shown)


fkiraly:
@achieveordie, it looks like this is still failing - is this a broken estimator?
Why did the tests pass on the original PR?

achieveordie:
> Why did the tests pass on the original PR?

It's because the estimators were removed from the being tested.

Primarily, apart from the bug we fixed last time there's also a missing line in `._fit()` which transposes the input because Keras's models expect `(batch_size, timestamps, features)` while other estimators in `sktime` expect `(batch_size, features, timestamps)`. This might not be the only problem here but it certainly is a bug.

There are also some cases where the runners run out of memory. Those seem like the case where a runner unluckily took testing heavy estimators.

fkiraly:
do you want to give it a shot to fix this?
Seems like you understand better what has been going wrong here.

I assumed it was fine to merge after the tests passed and you approved.

achieveordie:
This requires proper porting, I checked with the original code and quite a few things are different here than the original one. Let me fix it by tomorrow.

fkiraly:
that's odd. Where is this code then coming from? I didn't touch it except fixing bugs.

achieveordie:
Retriggering to check for sporadic errors

achieveordie:
@fkiraly This has been completed, only caveat that we need to be aware of that is we cannot save this estimator yet in its default state because of the `ReduceLROnPlateau` callback which has a lambda layer in its implementation. However, if `add_default_callback` is set as `False` (default `True`) then this estimator can be saved as normal.

In the original implementation, this callback was being added regardless and since the user might not want it, I decided to add a flag to disable that.

fkiraly:
makes sense - thanks for fixing!

PR #4537, opened 2023-05-02
[ENH] replace `joblib.hash` with `deep_equals` in `test_fit_does_not_overwrite_hyper_params`

fkiraly:
This PR replaces replace `joblib.hash` with `deep_equals` in `test_fit_does_not_overwrite_hyper_params`.

This is since `joblib.hash` seems to have issues with `pandas` objects, see #4534 or https://stackoverflow.com/questions/71312750/joblib-memory-and-pandas-dataframe-inputs


fkiraly:
does not work for `sklearn` estimators

PR #4539, opened 2023-05-02
[ENH] `statsforecast` `AutoTheta` direct interface estimator

yarnabrina:
#### Reference Issues/PRs

Depends on #4629
Part of #4643 

This PR adds and interface to the `AutoTheta` model provided in `statsforecast`. It's exposed to users as `StatsForecastAutoTheta`.

No.

No.

##### For all contributions

##### For new estimators

### DIFF
--- a/docs/source/api_reference/forecasting.rst
+++ b/docs/source/api_reference/forecasting.rst
@@ -206,6 +206,7 @@ All "ARIMA" models below include SARIMAX capability.
     :template: class.rst
 
     StatsForecastAutoARIMA
+    StatsForecastAutoTheta
 
 .. currentmodule:: sktime.forecasting.sarimax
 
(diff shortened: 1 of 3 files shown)


yarnabrina:
> @yarnabrina, great! I notice the import errors on 3.7, this is due to some types like `Literal` only being present from 3.8. I would say, let's try to avoid these imports.
> 
> Regarding the return type, it's a matter of conversion - `sktime` estimators expect the return to be the same data type as the input to `_fit`.

Removed the `Literal` as suggested.

I don't think I could follow your comment on returns. Can you please elaborate? I thought it is failing because of the existing adapter for `statsforeast` which may be too specific for ARIMA case. If it is related to `sktime` expectations as well, I'd need some time to analyse. Will check this over weekend.

fkiraly:
> Can you please elaborate?

Sorry for expressing myself obscurely.

I meant, the output mtype (e.g., `pd.Series` or `pd.DataFrame`) of the `_predict` return should be the same as the input mtyep for `y`, of the `_fit` arguments. Note the underscores.

yarnabrina:
Note: putting it on a hold for around one or two weeks. attempting adapter rewrite for statsforecast. If it works and is approved, will add theta forecaster from that branch in future.

fkiraly (approved):
Excellent, thanks!

Issue #4544, opened 2023-05-03
[ENH] testing `sktime` without `pytest`


fkiraly:
@yarnabrina made us aware of a quite vexing issue that makes my head hurt:

1. some parts of `sktime` that are not tests require `pytest` to run, e.g., `check_estimator`
2. some parts of `sktime` might import modules that should be isolated but aren't and `import pytest`, e.g., `all_estimators`

Simple question: how do we test that `sktime` runs without `pytest`, which is currently not a core dependency (but was an implied dependency prior to the dependency cleanup in 0.18.0 and `scikit-base`)

The conundrum is, currently any testing assumes `pytest` present (because we are using `pytest` for testing!), so we cannot test under the condition that it isn't - because we are testing!

We should at least be able to test that there are no unisolated `pytest` imports.

The same question poses itself for `scikit-base` as well.

The main alternative is, probably, making `pytest` a core dependency of `sktime`.

yarnabrina:
@fkiraly I may be wrong, but there is a chance that the `pytest` issue I faced in work pipelines are different from the one you have found. Can you please confirm?

This is a (hopefully) reproducible setup to duplicate my error.

## Environment

```Dockerfile
FROM python:3-slim-bullseye
RUN pip install "sktime>=0.18" statsforecast
```

## Codes

```pycon
>>> from sktime.forecasting.statsforecast import StatsForecastAutoARIMA
>>> model = StatsForecastAutoARIMA()
>>> model_clone = model.clone()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.11/site-packages/skbase/base/_base.py", line 164, in clone
    from skbase.testing.utils.deep_equals import deep_equals
  File "/usr/local/lib/python3.11/site-packages/skbase/testing/__init__.py", line 5, in <module>
    from skbase.testing.test_all_objects import (
  File "/usr/local/lib/python3.11/site-packages/skbase/testing/test_all_objects.py", line 14, in <module>
    import pytest
ModuleNotFoundError: No module named 'pytest'
>>>
```

## Comment

So, this line is the issue. I really feel equality check should be in `utils` instead of in `testing/utils`, and that'd probably solve the problem as that `deep_equals` does not sesm to need `pytest` after all.

https://github.com/sktime/skbase/blob/9a5c70c358ef89b06993148407d93c524141e07a/skbase/base/_base.py#L96

fkiraly:
[...]

PR #4551, opened 2023-05-04
[MNT] add silent dependencies to core dependency set

fkiraly:
This PR adds all formerly silent dependencies to the core dependency set.

A silent dependency is:

* an indirect dependency, i.e., a dependency of a core dependency, or an indirect dependency of a core dependency
* not a core dependency as per `pyproject.toml`
* imported directly from `sktime`

This is potentially problematic, as the dependency may decide to drop the dependency, which results in instant failure of `sktime` based CI.

An example of this in the past is the failure that https://github.com/sktime/sktime/pull/4450 fixed, where `pytest` dropped `attrs`, but `attrs`.

With https://github.com/sktime/sktime/issues/4544 in addition, it has become clear that we need to manage all silent dependencies and turn them into core dependencies.

This PR does that, as per the detection mechanism in https://github.com/sktime/sktime/pull/4548.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,6 +47,7 @@ dependencies = [
     "deprecated>=1.2.13",
     "numpy>=1.21.0,<1.25",
     "pandas>=1.1.0,<2.0.0",
+    "packaging",
     "scikit-base<0.5.0",
     "scikit-learn>=0.24.0,<1.3.0",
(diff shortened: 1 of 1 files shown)


fkiraly:
merging to enable quick testing of hotfix for `skbase` soft dependency issue

PR #4553, opened 2023-05-04
[ENH] Add a test to validate contract of DL Estimators

### LINKED ISSUE #4535: [ENH] Introduce tests to validate conformity of DL estimators
DL estimators are usually based on certain assumptions (save/load is more comprehensive, expected input dimension is [transposed](https://github.com/sktime/sktime/pull/4531#issuecomment-1529740511) that of normal estimators, expect private attributes like `network_`, `model_` etc) to properly function. This creates multiple points of failure and missing even one of these things creates a buggy estimator could need several PRs to test and debug.

Hence, I propose adding a test that checks for this conformity to avoid situations like #4531, #4527.

Tests that run only on child classes of `BaseDeepRegressor` and `BaseDeepClassifier` (and future Base-Deep Abstract Estimators) to check if the newly added estimator completely conforms to the expected design.

An alternative solution to this problem would involve a lot of human intervention - writing down steps to adding new DL estimators, reviewers checking off each point mentioned in the fore-mentioned recipe, or something like that. 

**Additional context**

achieveordie:
#### Reference Issues/PRs

Fixes #4535

Add a test to validate all the implicit/explicit points that a DL estimator has to adhere to. These points are necessary either for their proper functioning or for the estimator design to remain consistent throughout the framework.

No

Currently, the initial commit only contains the docstring of the test that is to be added. I propose using [`inspect`](https://docs.python.org/3/library/inspect.html#module-inspect) library to implement most points mentioned here. I'm up for different suggestions about how exactly should we go about implementing it.

This is a test :)

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/tests/test_all_estimators.py
+++ b/sktime/tests/test_all_estimators.py
@@ -1488,6 +1488,36 @@ class TestAllEstimators(BaseFixtureGenerator, QuickTester):
             if vars(estimator._network).get(key) is not None:
                 assert vars(estimator._network)[key] == value
 
+    @pytest.mark.skip(reason="Not yet implemented, only conceptualized.")
+    def test_dl_contract_consistency(self, estimator_class):
+        """
+        Test whether all DL estimators follow the required implicit/explicit contract.
+
+        This test checks for the following points:
+            1. The estimator must have a class method called `get_test_params`
+                which returns a list or dict.
+            2. The estimator contains a `._network` attribute after initialization.
+            3. The estimator must implement a `._fit()` method which returns itself.
+            4. After calling `.fit()` on the estimator, private attributes
+                `self.model_`, `self.optimizer_` and
+                `self.callbacks_` must not be None.
+            5. Following naming convention must be consistent during initialization:
+                a. There must be a `num_epochs` attributes and not `nb_epochs`.
+                b. The `verbose` attribute must be of `bool` type.
+            6. Via code-scanning, the following points are checked:
+                a. The input `X` must be transposed inside `.fit()`.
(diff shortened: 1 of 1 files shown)


achieveordie:
To improve the utility of this test, we would need to be aware that not all these tests might be required for some future estimators. In that case, we should also implement a system where all the excluded (and perhaps new tests only relevant to this specific estimator) is to be passed on via a `classmethod` of that estimator.

achieveordie:
One simple way to get away with classmethod and follow the current design is to modularize the current test function into smaller test functions which only test for one condition per function. This is also considered the better way of testing things (one function to test one thing) but in its current state, it will increase boilerplate code (return None if the estimator is not a subclass of Base-Deep class etc).

The problem could be tackled by adding another fixture as a child of the current one (rather than extending the same fixture) but I'm not familiar with pytest and their best practises so I might be wrong.

fkiraly:
Hm, I see.
There are two trade-offs we should think about:

* modular tests vs runtime from re-computin `fit` - the "optimal" in the latter is fitting only once and having only a single test for everything where `fit`is needed with an extremely branching tree. (I haven't figured out how to cache the fitted estimator, even though I've tried...)
* a central list for test excepts which can be considered good design as the exception can be seen "extrinsic" to the estimator, vs design that localizes estimator specific concerns with the estimator

achieveordie:
If we were able to cache the fitted object to be used for different tests as per necessity, we can have the best of both worlds. For the second point, I'd take the former option for the fact that it remains consistent with the design choices we have made till now. If we were able to cache the fitted object, we can modularize the `test_dl_contract_consistency()` test and have the contract formally written somewhere else.

PR #4559, opened 2023-05-07
[DOC} adds sktime internship link on the docs page

fkiraly:
Adds references to the sktime internship on the docs page.


fkiraly:
as discussed in the May 05 CC meeting

PR #4560, opened 2023-05-07
[BUG] Fixes incorrect window indexing in `HampelFilter`

antonioramos1:
#### Reference Issues/PRs
Not found any

Consists on two fixes:

1. SlidingWindowSplitter split assumes a default forecasting horizon of 1 which leaves one datapoint out for test.  
   - Impact: all the datapoints checked for outlier in the last window have miscalculated stats since the window is shifted by 1 position, leaving the last datapoint out.
    - Solution: changed SlidingWindowSplitter forecasting horizon to 0, this creates an additional window including the missing datapoint, thus ensuring the calculations for the last window are accurate.

2. Current way of finding the start and end where the window is incomplete is not accurate for odd-sized windows
    - Impact: for an odd-sized window, the datapoint n - floor(half_window_length) - 2 is not being considered as an outlier.
    - Solution: changed the logic to identify where the window is incomplete at the start and end of the series making the code more explicit, and added different logic to handle the even-sized and odd-sized windows.
  
See screenshot below which gives an overview of the issue and the proposed fix.
image

No

No

##### For all contributions

### DIFF
--- a/sktime/transformations/series/outlier_detection.py
+++ b/sktime/transformations/series/outlier_detection.py
@@ -7,6 +7,7 @@ __author__ = ["aiwalter"]
 __all__ = ["HampelFilter"]
 
 import warnings
+from math import ceil
 
 import numpy as np
 import pandas as pd
@@ -183,7 +184,7 @@ def _hampel_filter(Z, cv, n_sigma, half_window_length, k):
             if cv.window_length % 2 == 0:
                 start_end_win = half_window_length
             else:
-                start_end_win = half_window_length + 1
+                start_end_win = ceil(cv.window_length / 2)
             idx_range = range(len(Z) - start_end_win, len(Z))
         else:
             idx_range = [cv_window[0] + half_window_length]
--- a/sktime/transformations/series/outlier_detection.py
+++ b/sktime/transformations/series/outlier_detection.py
@@ -126,7 +126,10 @@ class HampelFilter(BaseTransformer):
             )
 
         cv = SlidingWindowSplitter(
-            window_length=self.window_length, step_length=1, start_with_window=True
+            fh=0,
+            window_length=self.window_length,
+            step_length=1,
+            start_with_window=True,
         )
         half_window_length = int(self.window_length / 2)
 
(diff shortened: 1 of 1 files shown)


fkiraly (changes requested):
Thanks! This is astutely observed!
Out of curiosity, how did you notice the bug?

May I request some changes for clarity:

* for the half rounded up, I think it is clearer if you do `from math import ceil`, then `start_end_win = ceil(hald_window_length)`
* is the "else" (middle) case correct in every case? Just asking, since shouldn't the rounding also affect that?

antonioramos1:
> Thanks! This is astutely observed! Out of curiosity, how did you notice the bug?

We were testing this implementation of the Hampel filter for an outlier solution and were lucky enough to have an obvious outlier that was not being picked up, so we started digging into the source code :)


>     * for the half rounded up, I think it is clearer if you do `from math import ceil`, then `start_end_win = ceil(half_window_length)`

I have changed to `start_end_win = ceil(cv.window_length / 2) `, let me know if happy with that

>     * is the "else" (middle) case correct in every case? Just asking, since shouldn't the rounding also affect that?

It is correct for odd-sized windows so I will say yes, but I didn't understand the point about rounding

fkiraly:
Thanks! Let's see whether the tests run through.

> It is correct for odd-sized windows so I will say yes, but I didn't understand the point about rounding

What I meant, there are two edge cases and one generic case. Edge cases are first window, last window. Generic case is some window in the middle.

What I was asking, have you checked whether there are any rounding problems for the middle windows? As there are no case distinctions there regarding rounding.

antonioramos1:
> What I was asking, have you checked whether there are any rounding problems for the middle windows? As there are no case distinctions there regarding rounding.

Got you, yes I have checked, the middle case is always accurate, for odd-sized windows it picks up the value in the middle and for even-sized windows it picks up middle value +1 as shown in the excel screenshot.


Please let me know if anything else is needed from my side to get this merged as it is my first time here :)

fkiraly:
Welcome to sktime, @antonioramos1, and thanks a lot for your contribution!

fkiraly:
@antonioramos1, feel free to add yourself to the contributors file!
with `bug` (for reporting) and `code` (for fixing) badges

Issue #4521, opened 2023-04-28
[ENH] Next generation reservoir computing estimator


frthjf:
I've recently implemented [Next Generation Reservoir Computing](https://www.nature.com/articles/s41467-021-25801-2) in the form of a scikit-learn estimator interface.

My code can be found [here](https://gist.github.com/frthjf/bbb13e0ab269bd43fbcd029999f4bad0); it is adapted from the original code by @agrif from [here](https://github.com/quantinfo/ng-rc-paper-code).

I thought I put this here in case anyone is interested in incorporating this into the sktime [template](https://github.com/sktime/sktime/blob/main/extension_templates/forecasting_simple.py).

fkiraly:
NICE! This looks very close to the `sktime` interface (with `fh` etc), so it should probably be an easy copy-paste job, using the extension template `extension_templates/forecasting_simple`.

Setting a "good first issue" tag for anyone who would like to take this on!

Some details on implementation that may be helpful
* The only tags that may have to be set to non-default values are probably the `y_inner_mtype`, `X_inner_mtype` tags (on first superficial glance - I may be wrong)
* the right location is probably a python file inside the `forecasting` module, e.g., `reservoir`

fkiraly:
one question would be who would own and maintain it - by default it is the implementer, but this can of course be shared with @agrif (if they are willing to fix bugs here)

agrif:
I'm afraid RC isn't really my wheelhouse these days -- though I'm always happy to see what people are doing with them.

afzal442:
If none has taken on this, I would like to take this opportunity for my masters. 😄 HTH!

fkiraly:
welcome back, @afzal442, great to hear from you again!!!

of course, that would be nice, and it's great you can do this as part of your masters project!

FYI, kindly see here our quick info for academic projects: https://github.com/sktime/mentoring#academic-mentoring

PR #4562, opened 2023-05-08
[ENH] add Next generation reservoir computing estimator

afzal442:
This PR adds next gen reservoir computing estimator extending simple forecasting template.

For #4521


afzal442:
hi @fkiraly can you plz check the fmt around this source code so that I can go for adding other metadata?

fkiraly:
This looks good, for a start, nice!

For "implementing estimators", we suggest you use the extension template, here is a tutorial (with video if you like):
https://www.sktime.net/en/latest/developer_guide/add_estimators.html

For code formatting standards, here is a guide how to set it up on your computer to ensure the code is formatted automatically:

afzal442:
Thanks @fkiraly , I have been following that fixing code quality, formatting codes, etc. But my concern is about `tags`. Although I found from the docs, I had some caveats pulling me back from working further.

sktime/forecasting/reservoir.py
```diff
@@ -0,0 +1,119 @@
...
+"""Implements Next generation reservoir computing estimator."""
+
+__all__ = ["NextRC"]
+__author__ = ["frthjf"]
+
+import numpy as np
+import sklearn.linear_model
+from sklearn.utils.validation import check_array
+
+from sktime.base import BaseEstimator
```
afzal442, line 12: Does `BaseEstimator` follow `BaseForcaster` extending forecasting template? Silly question.

fkiraly: you'll need to inherit from `BaseForecaster` and follow the `forecasting_simple` extension template. That also explains the tags.

afzal442: Wholesome. Thanks.

fkiraly:
the failure is due to `X` sometimes being `None`. The estimator should be able to work in cases where `X` is not provided (i.e., no exogeneous, only endogeneous data), as well as in cases where `X` is passed.

afzal442:
sorry! some caveats cause falling me apart here. after adding this following condition, I am getting this error `IndexError: index 1 is out of bounds for axis 0 with size 1` Is the original code compatible to what it is supposed to work as expected? Thanks
```python
if X is not None:
            (n_samples, n_features) = X.shape
            X_features = self._feature_vector(X[: fh[1] + 1, :])
        else:
            n_samples = y.shape[0]
            n_features = 1
            X_features = self._feature_vector(np.zeros((n_samples, 1)))
```

fkiraly:
One would have to look into the code and debug - could be anything. Where is the `IndexError` coming from?

afzal442:
Hi, if you get to test interface conformance through debugging a `check_estimator` like below 
```
from sktime.forecasting.reservoir import NextRC
check_estimator(NextRC, raise_exceptions=True)
```

fkiraly:
> Hi, if you get to test interface conformance through debugging

Yes, but what I meant, what is the last call within `sktime` that causes the bug? Which test is failing?

afzal442:
Ah I see. Here what I got is I put up below
```
  File "/workspaces/sktime/sktime/forecasting/reservoir.py", line 141, in _fit
    X_features = self._feature_vector(X[: fh[1] + 1, :])
  File "/workspaces/sktime/sktime/forecasting/base/_fh.py", line 70, in delegated
    return getattr(obj.to_pandas(), method)(*args, **kwargs)
  File "/workspaces/sktime/.venv/lib/python3.10/site-packages/pandas/core/indexes/base.py", line 5320, in __getitem__
    return getitem(key)
IndexError: index 1 is out of bounds for axis 0 with size 1
```

I have no idea to crack it down. While I had an issue encountered, there was like 'NoneType' object has no attribute 'shape' dealing with`(n_samples, n_features) = X.shape`

fkiraly:
Ok, I see - I recommend to use python step through, i.e., debugging, to see how this looks in the happy case (where the code works) and in the failure case.

I.e., which objects end up different?

Generic stackkoverflow post on debugging:
https://stackoverflow.com/questions/4929251/how-to-step-through-python-code-to-help-debug-issues

afzal442:
That is a great advice. I think I have missed collaboration meetings a lot on discord. Hopefully I will attend next time soon. until then I will keep trying out debugging as much as I can.

afzal442:
Hi @frthjf, when I ran your code to check if it works fine or not, it showed up a few errors against X being None. Here is the following logs to let you know where it comes from. LMK how we can handle such. Thanks
https://github.com/sktime/sktime/actions/runs/4934593430/jobs/8819842757?pr=4562#step:7:2741

Could you recheck on the surface of sktime estimators? What if you pass in X=None?

frthjf:
That's right, the original code does not support the case where `X=None`. In that case, I am unsure what the standard behavior for sktime estimators should be, perhaps skipping fitting altogether?

sktime/forecasting/reservoir.py
```diff
@@ -0,0 +1,187 @@
...
+        self.X_ = X
+        self.y_ = y
+        self.delay_ = 2
+
+        if self.regressor is None:
+            self.regressor_ = sklearn.linear_model.Ridge(
+                alpha=1.0e-3, fit_intercept=False
+            )
+
+        (n_samples, n_features) = X.shape
```
afzal442, line 139: Also `n_samples` has no use. We can ignore it. Right?

sktime/forecasting/reservoir.py
```diff
@@ -0,0 +1,187 @@
...
+    >>> y_pred = forecaster.predict(fh=[1,2,3])
+    """
+
+    _tags = {
+        "y_inner_mtype": "pd.Series",
+        "X_inner_mtype": "pd.DataFrame",
+        "requires-fh-in-fit": False,
+        "handles-missing-data": True,
+        "scitype:y": "univariate",
+        "ignores-exogeneous-X": False,
```
afzal442, line 58: I made it `False` here, but still it ignores X data. LMK if I misunderstood here.

sktime/forecasting/reservoir.py
frthjf: Yep, you could write `(_, n_features) = X.shape` or `n_features = X.shape[1]` or just ignore it.

sktime/forecasting/reservoir.py
fkiraly: if `True`, then we always pass `X=None` to the inner `_fit`, `_predict`. If `False`, then `X` can be `None` or non-empty. In either case, the inner `_fit` and `_predict` needs to be able to cope with `X=None` as per the current contract.

sktime/forecasting/reservoir.py
```diff
@@ -0,0 +1,225 @@
+# -*- coding: utf-8 -*-
+# copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+"""Implements Next generation reservoir computing estimator."""
+
+__all__ = ["NextRC"]
+__author__ = ["frthjf"]
```
frthjf, line 6: I think you should be the author, at least co-author.

sktime/forecasting/reservoir.py
```diff
@@ -0,0 +1,225 @@
...
+            )
+        else:
+            self.regressor_ = clone(self.regressor)
+
+        n_features = y.shape[1]
+
+        # use relative fh as time index to predict
+        fh = self.fh.to_absolute_index(self.cutoff)
+
+        # print(y[: fh[-1] + 1, :])
```
frthjf, line 147: Are these comments still meant to be here, or is this a left over from debugging?

frthjf (commented):
This looks great @afzal442!

The only suggestion I have is that you may like to include a test case like [this one](https://gist.github.com/frthjf/bbb13e0ab269bd43fbcd029999f4bad0#file-test_next_rc-py) to check for correctness. Although, I am not sure how this works with the sktime's CI (maybe @fkiraly knows). In any case, it all looks good to me, thanks for working on this!

sktime/forecasting/reservoir.py
afzal442: Yep! spot on. I have passed over it for debugging. I will remove it as soon as our CI is happy.

sktime/forecasting/reservoir.py
```diff
@@ -0,0 +1,225 @@
...
+
+        X_features = self._feature_vector(y[: fh[-1] + 1, :])
```
afzal442, line 149: @frthjf could you help me know what fh[0] and fh[1] correspond to as per your code? In sktime fh[1] hit an error due to fh being `absolute` here. both fh[0] and fh[-1] gives 20 after changing `fh.to_absolute_index`

frthjf: If I remember correctly `fh[0]` is the start and `fh[1]` is the end of the interval used for prediction (index relative to the data array). So if you call `estimator.fit([1, 2, 3, 4, 5], fh=[2,3])` you are operating on `[3,4]`.

fkiraly: I see. In `sktime`, a relative index of `0` is actually the `cutoff`, i.e., the last element of the array. `1` would correspond to an offset of one, i.e., if `y` seen in `fit` has length 20, with indices 0-19, then `1` would correspond to forecasting 20. `0` would correspond to an in-sample forecast of the index 19.

Issue #4303, opened 2023-03-07
[DOC] write complete docstrings for performance metrics classes


fkiraly:
The docstrings of performance metrics classes in `performance_metrics.forecasting._classes` should be completed.

For each metric, the following should be done:

* a mathematical summary of the metric should be provided, including for the output of `evaluate_by_index`
* it should be ensured that the `multilevel` and `multioutput` arguments are included and documented.

[This PR](https://github.com/sktime/sktime/pull/4302) can be taken as a blueprint for improvements/changes to the docstrings.

If you would like to work on this, please indicate here the name of the metrics class, and whether you will work on math, parameters/arguments, or both.

ramsarank:
Hello @fkiraly 

Just wanted to check if there are any pending metrics classes.
Thank you.

fkiraly:
Yes! Examples are `MedianAbsoluteScaledError`, `MeianAbsoluteScaledErro`.

Would be great if you could help, @ramsarank!

ramsarank:
Sure, I shall work on this. Thank you @fkiraly

nkc-137:
Hello @fkiraly, 
I went through the `_classes` file and it looks like all the metric classes have doc strings and parameters already. But I saw that some metrics like `MeanSquaredError` and `MedianAbsoluteError` don't have mathemical formulae as the metric `MeanAbsoluteError` doc string has. Is this something I could potentially work on? 

Thanks!

fkiraly:
> Is this something I could potentially work on?

Absolutely! Thanks!

Another, related, task is to implement efficient `_evaluate_by_index` methods: https://github.com/sktime/sktime/issues/4304

There's some synergy doing the two for one given metric at the same time, but of course it's not required to do so.

sinemkilicdere:
> > Is this something I could potentially work on?
> 
> Absolutely! Thanks!
> 
> Another, related, task is to implement efficient `_evaluate_by_index` methods: [#4304](https://github.com/sktime/sktime/issues/4304)
> 
> There's some synergy doing the two for one given metric at the same time, but of course it's not required to do so.

I would love to work on _mlinex.py.

PR #4563, opened 2023-05-08
[DOC] Improve Docstring for MAPE Metrics

### LINKED ISSUE #4457: [ENH] naming of symmetricMAPE vs adjustedMAPE - clarifying docstring
_Originally posted by @mtorabirad in https://github.com/sktime/sktime/issues/1986#issuecomment-1470320909_

A minor point but shouldn't we call symmetricMAPE adjustedMAPE. After all, sMAPE is not really symmetric and the original developer of the Metric, Armstrong, never called it symmetricMAPE. https://www.youtube.com/watch?v=_uaJWg5eNpE&ab_channel=MLBoost

hazrulakmal:
#### Reference Issues/PRs
Fixes #4457. See also #4303 

1. Added adjusted vs symmetric clarifications for sMAPE.
2. Improved docstring documentation to include a mathematical summary of the metric and additional parameters that the metric class currently supports.

No

#### Any other comments?

This isn't directly related to the PR, but I ran the documentation build locally to confirm that the changes were displayed correctly. During the process, I noticed that it took a while to build the HTML docs. It would be great if this process could be sped up a bit. Maybe consider `sphinx-autobuild` extension to automatically detect changes in docs with live-reload in the browser during local development.

##### For all contributions

### DIFF
--- a/sktime/performance_metrics/forecasting/_classes.py
+++ b/sktime/performance_metrics/forecasting/_classes.py
@@ -1735,15 +1735,8 @@ class MeanAbsolutePercentageError(BaseForecastingErrorMetricFunc):
     MeanSquaredPercentageError
     MedianSquaredPercentageError
 
-    Notes
-    -----
-    In some literature sources, sMAPE is also known as Adjusted MAPE because
-    ,in some cases, over- and under-forecasts are not really penalised equally. [1]_
-
     References
     ----------
-    .. [1] Goodwin.P and Lawton.R (1999) "On the asymmetry of the symmetric MAPE".
-
     Hyndman, R. J and Koehler, A. B. (2006). "Another look at measures of
     forecast accuracy", International Journal of Forecasting, Volume 22, Issue 4.
 
@@ -1868,15 +1861,8 @@ class MedianAbsolutePercentageError(BaseForecastingErrorMetricFunc):
     MeanSquaredPercentageError
     MedianSquaredPercentageError
 
-    Notes
-    -----
-    In some literature sources, sMdAPE is also known as Adjusted MdAPE because
-    ,in some cases, over- and under-forecasts are not penalised equally. [1]_
-
     References
     ----------
-    .. [1] Goodwin.P and Lawton.R (1999) "On the asymmetry of the symmetric MAPE".
-
     Hyndman, R. J and Koehler, A. B. (2006). "Another look at measures of
     forecast accuracy", International Journal of Forecasting, Volume 22, Issue 4.
 
--- a/sktime/performance_metrics/forecasting/_classes.py
+++ b/sktime/performance_metrics/forecasting/_classes.py
@@ -1684,7 +1684,7 @@ class MeanAbsolutePercentageError(BaseForecastingErrorMetricFunc):
     predicted values :math:`\widehat{y}_1, \dots, \widehat{y}_n`,
     at time indices :math:`t_1, \dots, t_n`,
     `evaluate` or call returns the Mean Absolute Percentage Error,
-    :math:`\frac{1}{n}\sum_{i=1}^n |\frac{y_i - \widehat{y}_i}{y_i}|`.
+    :math:`\frac{1}{n} \sum_{i=1}^n \left|\frac{y_i - \widehat{y}_i}{y_i} \right|`.
     (the time indices are not used)
 
     if `symmetric` is True then calculates
@@ -1708,7 +1708,7 @@ class MeanAbsolutePercentageError(BaseForecastingErrorMetricFunc):
 
     `evaluate_by_index` returns, at a time index :math:`t_i`,
     the abolute percentage error at that time index,
-    :math:`|\frac{y_i - \widehat{y}_i}{y_i}|`,
+    :math:`\left| \frac{y_i - \widehat{y}_i}{y_i} \right|`,
     or :math:`\frac{2|y_i - \widehat{y}_i|}{|y_i| + |\widehat{y}_i|}`,
     the symmetric version, if `symmetric` is True, for all time indices
     :math:`t_1, \dots, t_n` in the input.
@@ -1814,7 +1814,7 @@ class MedianAbsolutePercentageError(BaseForecastingErrorMetricFunc):
     predicted values :math:`\widehat{y}_1, \dots, \widehat{y}_n`,
     at time indices :math:`t_1, \dots, t_n`,
     `evaluate` or call returns the Median Absolute Percentage Error,
-    :math:`median(|\frac{y_i - \widehat{y}_i}{y_i}|)`.
+    :math:`median(\left|\frac{y_i - \widehat{y}_i}{y_i} \right|)`.
     (the time indices are not used)
 
     if `symmetric` is True then calculates
--- a/sktime/performance_metrics/forecasting/_classes.py
+++ b/sktime/performance_metrics/forecasting/_classes.py
@@ -1684,15 +1684,16 @@ class MeanAbsolutePercentageError(BaseForecastingErrorMetricFunc):
     predicted values :math:`\widehat{y}_1, \dots, \widehat{y}_n`,
     at time indices :math:`t_1, \dots, t_n`,
     `evaluate` or call returns the Mean Absolute Percentage Error,
-    :math:`\frac{100\%}{n}\sum_{i=1}^n |\frac{y_i - \widehat{y}_i}{y_i}|`.
+    :math:`\frac{1}{n}\sum_{i=1}^n |\frac{y_i - \widehat{y}_i}{y_i}|`.
     (the time indices are not used)
 
     if `symmetric` is True then calculates
     symmetric mean absolute percentage error (sMAPE), defined as
-    :math:`\frac{200\%}{n} \sum_{i=1}^n \frac{|y_i - \widehat{y}_i|}
+    :math:`\frac{2}{n} \sum_{i=1}^n \frac{|y_i - \widehat{y}_i|}
     {|y_i| + |\widehat{y}_i|}`.
 
-    Both MAPE and sMAPE output is non-negative floating point. The best value is 0.0.
+    Both MAPE and sMAPE output non-negative floating point which is in fractional units
+    rather than percentage. The best value is 0.0.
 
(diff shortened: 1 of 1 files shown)


sktime/performance_metrics/forecasting/_classes.py
```diff
@@ -1677,11 +1677,22 @@ def get_test_params(cls, parameter_set="default"):
...
 
-    If `symmetric` is False then calculates MAPE and if `symmetric` is True
-    then calculates symmetric mean absolute percentage error (sMAPE). Both
-    MAPE and sMAPE output is non-negative floating point. The best value is 0.0.
+    For a univariate, non-hierarchical sample
+    of true values :math:`y_1, \dots, y_n` and
+    predicted values :math:`\widehat{y}_1, \dots, \widehat{y}_n`,
+    at time indices :math:`t_1, \dots, t_n`,
+    `evaluate` or call returns the Mean Absolute Percentage Error,
+    :math:`\frac{100\%}{n}\sum_{i=1}^n |\frac{y_i - \widehat{y}_i}{y_i}|`.
```
fkiraly, line 1687: I think it returns the metric in units of a float and not percentage, so we should replace the `100%` by `1` and the `200%` by `2`, right?

Might be worth adding a `Note: the metric returns the MAPE in fractional units, not percentages, that is, 0.5. rather than 50% etc`

sktime/performance_metrics/forecasting/_classes.py
```diff
@@ -1701,15 +1720,27 @@ class MeanAbsolutePercentageError(BaseForecastingErrorMetricFunc):
...
 
     See Also
     --------
     MedianAbsolutePercentageError
     MeanSquaredPercentageError
     MedianSquaredPercentageError
 
+    Notes
+    -----
+    In some literatures, sMAPE is also known as Adjusted MAPE because
```
fkiraly, line 1737: In some literatures -> In some literature sources

sktime/performance_metrics/forecasting/_classes.py
```diff
@@ -1701,15 +1720,27 @@ class MeanAbsolutePercentageError(BaseForecastingErrorMetricFunc):
...
+    ,in some cases, over- and under-forecasts are not really penalised equally. [1]_
```
fkiraly, line 1738: do you have the exact page or section number?

Does this contain both the sMAPE and adj MAPE terminology?

sktime/performance_metrics/forecasting/_classes.py
```diff
@@ -1701,15 +1720,27 @@ class MeanAbsolutePercentageError(BaseForecastingErrorMetricFunc):
...
+
     References
     ----------
+    .. [1] Goodwin.P and Lawton.R (1999) "On the asymmetry of the symmetric MAPE".
+
     Hyndman, R. J and Koehler, A. B. (2006). "Another look at measures of
```
fkiraly, line 1744: this is not cited in the text, would be worth pointing to this, and add an exact section or page reference.

fkiraly (changes requested):
Thanks, really great!

I left some small comments, these are not blocking since small, but kindly have a look.

sktime/performance_metrics/forecasting/_classes.py
hazrulakmal: My intention is to provide readers with additional resources to understand why symmetric MAPE is not truly "symmetric." This explains the pushback against naming it as such. Perhaps I should include a section titled "Further Reading" instead of listing it under "References"?

sktime/performance_metrics/forecasting/_classes.py
hazrulakmal: Unfortunately, I dont have access to the full paper. but the paper snippets the author wants to call it modified MAPE. Anyway, for context, I got the paper from the youtube video that was shared by the person who raised this issue.

sktime/performance_metrics/forecasting/_classes.py
hazrulakmal: Yup. You are right! can't believe I miss that from the code.

sktime/performance_metrics/forecasting/_classes.py
fkiraly: I think "references" is perfectly fine, if you list it as a reference for the statements in relation to the terminology "symmetric".

Having said this, the symmetric MAPE *is* symmetric, no? It satisfies the definition of "symmetric function" (see [wikipedia](https://en.wikipedia.org/wiki/Symmetric_function)) if you consider the prediction vector and training data vector as individual variables. Similarly, the individual sample losses are symmetric in the prediction and training data point.

sktime/performance_metrics/forecasting/_classes.py
```diff
@@ -1691,6 +1702,14 @@ class MeanAbsolutePercentageError(BaseForecastingErrorMetricFunc):
     values are close to zero. In such cases the function returns a large value
     instead of `inf`.
 
+    `multioutput` and `multilevel` control averaging across variables and
+    hierarchy indices, see below.
+
+    `evaluate_by_index` returns, at a time index :math:`t_i`,
+    the abolute percentage error at that time index,
+    :math:`100\%|\frac{y_i - \widehat{y}_i}{y_i}|`,
+    for all time indices :math:`t_1, \dots, t_n` in the input.
```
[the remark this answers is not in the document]
fkiraly: still, I think it is worth a repetition to avoid confusion... "if `symmetric=True` etc

sktime/performance_metrics/forecasting/_classes.py
hazrulakmal: fair point. I suppose I'll remove the reference. rather provide a flag to readers something along the line '_one subtle problem to note is that sMAPE is not symmetric in the sense that over- and under-forecasts are not treated equally. i.e  over-forecasting with y_t = 150 and yhat_t = 200 gives sMAPE = 0.29, while under-forecasting with y_t = 150 and yhat_t = 100 gives sMAPE = 0.4. However, MAPE = 0.33 for both cases_.'. Also, based on this [Wikipedia](https://en.wikipedia.org/wiki/Symmetric_mean_absolute_percentage_error), it appears adjusted MAPE equation is slightly different than the symmetric MAPE equation we have in the library, namely on the denominator.

fkiraly: thanks - this is making more sense then.

What I would avoid though is too extensive discussion of the metrics, which one is better or problematic etc, such as what follows "one subtle problem" (the "asymmetry" is a general problem for relative metrics).

This is better done in scientific papers (which can be referred to).
You could add the original references though, and/or references of papers that discuss it.

hazrulakmal: In light of the wikipedia page, I've removed a discussion on adjusted MAPE. Kindly check the latest changes :)

fkiraly (changes requested):
Excellent, all good to go!

Non-blocking comment: The absolute bars around fractions render as too short, so I would recommend to use the `\left|` and `\right|` commands instead of just `|` if it is around `\frac`

hazrulakmal:
Done!

fkiraly (approved):
Nice! Thanks a lot!

fkiraly (changes requested):
Apologies for taking the approval back, but it looks like there is now an issue with the examples.

As the `r` has been added , the backslash for newline (that was already there!) is now causing trouble. Could you kindly check and possibly reformat?

hazrulakmal:
I see. based on [sphnix doc](https://sphinx-rtd-trial.readthedocs.io/en/latest/ext/math.html),  I added `r` because it seems that sphinx can't detect the backslash for `\frac`. 

> Keep in mind that when you put math markup in Python docstrings read by [autodoc](https://sphinx-rtd-trial.readthedocs.io/en/latest/ext/autodoc.html#module-sphinx.ext.autodoc), you either have to double all backslashes, or use Python raw strings (r"raw").

Didnt realise that it'd affect the `\` at the bottom. I've reformatted it now. It should work.

fkiraly:
> Also, I'm wondering, which test files that run documentation tests for consistency. would be great if I could run the tests locally and detect this sort of problem before pushing the commits.

I think that's an extremely valid point! Would appreciate if you could open an issue that there should be some guide on how to run the doctests locally. The current guide only covers `pytest`.

Could you open an issue to add this to `docs/source/developer_guide/continuous_integration.rst` which has instructions on how to run tests locally? I would add it as a subsection after `Full test suite runs via pytest`.
(if you do that, feel free to say that you're going to do this, if not I'll do this at some point closer to the release)

Here's from `doctest` on running the doctests locally: https://docs.python.org/3/library/doctest.html

fkiraly:
you still need the `r` for the TeX backslashes! You just need to reformat the code example so it does not have a line break.

fkiraly:
> Also regarding doctest, will open an issue about it soon! but has doctest already been implemented internally, or is the issue meant to be the first step towards implementing it?

doctest already runs as part of the CI, it's only about telling people how to run it locally

> is there a way to still use backslash as a continuation line when the docstring is in raw format

I don't know... I would just shorten the line

hazrulakmal (commented):
What should I do with this?

hazrulakmal:
> I don't know... I would just shorten the line

the thing is, that sentence explains a list of acceptable parameters for that argument and what the default parameter. cant really shorten it else, users won't know what to change in that argument to control the function behaviour. 

Anyway, I may have missed your requested changes.  😅 Sorry about that. missed the opportunity to get coauthor-commit badges. 

> (well, funnily, this got more complicated than thought. What a fiddly thing, that backslash...)

yeah lol.

fkiraly:
> I guess I missed the opportunity to get coauthor-commit badges.

Not sure what badge that is?

You can get a `doc` badge for this in the `all-contributorsrc` file if you want, though!

fkiraly:
>> I guess I missed the opportunity to get coauthor-commit badges.

> Not sure what badge that is?

I see - well, in that case, I took the liberty to commit directly to your branch with an alternative docstring field that does not go over the character limit, this should also get you the badge :-)

fkiraly (approved):

Issue #4566, opened 2023-05-09
[BUG] recent failure to install `tensorflow-probability` on `python 3.7`


fkiraly:
The recent update of `tensorflow-probability` (0.20.0) seems to causes an error at the installation stage, on `python 3.7`:

```
INTERNALERROR> ImportError: This version of TensorFlow Probability requires TensorFlow version >= 2.12; Detected an installation of version 2.11.0. Please upgrade TensorFlow to proceed.
```

FYI @yarnabrina as we hae discussed python 3.7 recently.

fkiraly:
https://github.com/sktime/sktime/pull/4567 bounds `tensorflow-probability<0.20.0` - once python 3.7 EOL is reached, we can release the bound.

yarnabrina:
Not sure if it exists already, but may be we can create a issue tracker of things to do after EOL? adding pins like this, may be running `pyupgrade --py38-plus`, etc.?

---

Also, I thought not pinning will download latest supported version for each package, and it's dependencies are changed accordingly. Didn't know it fails this way.

fkiraly:
> Not sure if it exists already, but may be we can create a issue tracker of things to do after EOL?

No, it doesn't - great idea!

Do you want to open it? Could be an issue, or a milestone. Perhaps umbrella issue is easier to manage.

yarnabrina:
Created #4569. Please take a look.

fkiraly:
excellent, then we can close this and keep track of the todos there.

PR #4567, opened 2023-05-09
[MNT] bound `tensorflow-probability` to `<0.20.0`

fkiraly:
Fixes https://github.com/sktime/sktime/issues/4566 temporarily by bounding `tensorflow-probability` to `<0.20.0`.


fkiraly:
failure on main, therefore merging

PR #4572, opened 2023-05-11
[DOC] docstring for `SimpleRNNClassifier`

### LINKED ISSUE #4558: [DOC] write docstring for RNN estimators
The estimators `SimpleRNNRegressor` and `SimpleRNNClassifier` are practically empty, these should be added.

wasup-yash:
#### Reference Issues/PRs
Towards #4558

Added DocStrings for the `[SimpleRNNClassifier]`(https://github.com/sktime/sktime/blob/main/sktime/classification/deep_learning/rnn.py#L20)

### DIFF
--- a/sktime/classification/deep_learning/rnn.py
+++ b/sktime/classification/deep_learning/rnn.py
@@ -22,7 +22,6 @@ class SimpleRNNClassifier(BaseDeepClassifier):
     
     Parameters
     ----------
-    should inherited fields be listed here?
     n_epochs       : int, default = 2000
         the number of epochs to train the model
     batch_size      : int, default = 16
--- a/sktime/classification/deep_learning/rnn.py
+++ b/sktime/classification/deep_learning/rnn.py
@@ -19,6 +19,36 @@ _check_dl_dependencies(severity="warning")
 
 class SimpleRNNClassifier(BaseDeepClassifier):
     """Simple recurrent neural network.
+    
+    Parameters
+    ----------
+    should inherited fields be listed here?
+    n_epochs       : int, default = 2000
+        the number of epochs to train the model
+    batch_size      : int, default = 16
+        the number of samples per gradient update.
+    kernel_size     : int, default = 7
+        the length of the 1D convolution window
+    avg_pool_size   : int, default = 3
+        size of the average pooling windows
+    n_conv_layers   : int, default = 2
+        the number of convolutional plus average pooling layers
+    filter_sizes    : array of shape (n_conv_layers) default = [6, 12]
+    random_state    : int or None, default=None
+                      Seed for random number generation.
+    verbose         : boolean, default = False
+                      whether to output extra information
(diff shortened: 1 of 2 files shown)


yarnabrina:
@fkiraly just a quick question: why do most of `deep_learning` directory codes have a different docustring style from rest? These extra spaces before `:` are not really common in other `sktime` modules or even in `numpydoc` etc.

Not that it's objectively bad or anything, just asking because of inconsistent (if it matters for docstrings) style from most of other modules.

fkiraly:
> @fkiraly just a quick question: why do most of `deep_learning` directory codes have a different docustring style from rest? These extra spaces before `:` are not really common in other `sktime` modules or even in `numpydoc` etc.

Because it has been without maintenance for a period of time - most deep learning estimators lived originally in `sktime-dl`, which got abandoned as its maintainers did not keep up with the professionalisation of the software architecture in the main package (which also included better docstrings as a side effect).

As content has been migrated to `sktime`, the interfaces have been updated, but the docstrings often still are in the old style.

sktime/classification/deep_learning/rnn.py
```diff
@@ -19,6 +19,36 @@
 
 class SimpleRNNClassifier(BaseDeepClassifier):
     """Simple recurrent neural network.
+    
+    Parameters
+    ----------
+    should inherited fields be listed here?
```
fkiraly, line 25: what do you mean by that? Like `random_state`?

Yes, all args of `__init__`.

fkiraly (changes requested):
Thanks, this looks great!

Could you kindly remove the "should inherited fields" question from the docstring? Sth like this is better asked in the PR thread.

wasup-yash:
> 

> Thanks, this looks great!
> 
> Could you kindly remove the "should inherited fields" question from the docstring? Sth like this is better asked in the PR thread.

ok I am gonna remove it, thanks for the review

fkiraly (changes requested):
now waitasec, I just noticed - and apologies for not noticing earlier - the defaults in the docstring don't actually agree with those in the `__init__`!

Can you kindly fix that? Also, where the default is `None` (e.g., ,`optimizer`), you should check what it is set to internally, e.g., is it `Adam`?

wasup-yash:
> now waitasec, I just noticed - and apologies for not noticing earlier - the defaults in the docstring don't actually agree with those in the `__init__`!
> 
> Can you kindly fix that? Also, where the default is `None` (e.g., ,`optimizer`), you should check what it is set to internally, e.g., is it `Adam`?

oh, sorry for not noticing it from my side too, I'll work on it and update it for the review.

fkiraly (approved):
Fixed outstanding items for the release

wasup-yash:
> Fixed outstanding items for the release

thanks a lot @fkiraly for the fixes, I was indulged in uni exams and thought gonna work on it but you already fixed the docs, Really appreciate it.

PR #4575, opened 2023-05-11
[ENH] changing Union[int | float] to float as per issue #4379

### LINKED ISSUE #4379: [ENH] simplify type annotations for float
As per [PEP 484](https://peps.python.org/pep-0484/#the-numeric-tower), the type annotations do not have to say `Union[int | float]`, just a float is OK.

mdsaad2305:
#### Reference Issues/PRs

Fixes #4379 

in type annotations, replaces all instances of `Union[int | float]` (either way round) with `float`

This is my first contribution.
replacing Union[int | float] to just float

### DIFF
--- a/sktime/distances/_lower_bounding_numba.py
+++ b/sktime/distances/_lower_bounding_numba.py
@@ -244,7 +244,7 @@ def numba_create_bounding_matrix(
     x: np.ndarray,
     y: np.ndarray,
     window: float = -1.0,
-    itakura_max_slope: Union[float, int] = -1.0,
+    itakura_max_slope: float = -1.0,
 ) -> np.ndarray:
     """Numba compiled way of creating bounding matrix.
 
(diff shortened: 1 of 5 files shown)


fkiraly (changes requested):
Thanks!

There are a lot of other occurrences of `Union[int, float]`, so it would be great if you could also take care of those - simply use vs code or pycharm search all files functionality.

mdsaad2305:
Will make the necessary changes

mdsaad2305:
I think these are the only occurrences of `Union[float, int]` and `Union[int, float]`. Furthermore, I recall the flake8 test failing during pre-commit due to the import statements, so I removed them

fkiraly:
> I think these are the only occurrences of `Union[float, int]` and `Union[int, float]`.

Yes, I think you got all of them, thanks!

fkiraly (approved):
Thank you!

PR #4251, opened 2023-02-20
[ENH] Add ForecastingBayesianSearchCV using Skopt 

hazrulakmal:
Fixes #4188. 
See also #3390 #3359 

#### What does this implement/fix? Explain your changes.

This PR aims to integrate BayesianCV optimization method using skopt under the hood into sktime ecosystem 

#### Does your contribution introduce a new dependency? If yes, which one?

Yes, scikit-optimize

#### What should a reviewer concentrate their feedback on?

Under development. Will mention someone when the code is ready for review. 

#### Did you add any tests for the change?

Will do it soon. 

#### Any other comments?

Under Active development.

##### Note to self
- [ ] Look through BaseGridSearch, ForecastingGridSearch etc works
- [ ] Look through how to use skopt
- [ ] think of the best way to integrate skopt, inherit from `BaseGridSearch` or from `_DelegatedForecaster`
##### For all contributions
- [ ] I've added myself to the [list of contributors](https://github.com/sktime/sktime/blob/main/.all-contributorsrc).
- [ ] Optionally, I've updated sktime's [CODEOWNERS](https://github.com/sktime/sktime/blob/main/CODEOWNERS) to receive notifications about future changes to these files.
- [ ] The PR title starts with either [ENH], [MNT], [DOC], or [BUG] indicating whether the PR topic is related to enhancement, maintenance, documentation, or bug.

##### For new estimators
- [ ] I've added the estimator to the online documentation.
- [ ] I've updated the existing example notebooks or provided a new one to showcase how my estimator works.

PR #4580, opened 2023-05-12
[ENH] SkoptForecastingCV Integration

### LINKED ISSUE #3390: [ENH] ForecastingBayesianSearchCV, skopt interface
Would it be possible to add support for Bayesian search, e.g. make skopt.BayesSearchCV accessible through ForecastingBayesSearchCV? Similar to how ForecastingGridSearchCV provides an interface to GridSearchCV from sklearn. 

This would be a great help to tune more complex models with large parameter space.

### LINKED ISSUE #4188: [ENH] advanced hyper-parameter tuning for forecasting
We should add a module in `sktime` that is to `ForecastingGridSearchCV` as `tune-sklearn` is to `sklearn`.

This should not be too hard if we inherit from `BaseGridSearch` and exchange the tuning engine.

Optimization methods:

- [ ] `hyperopt` native
- [ ] `optuna` native
- [ ] `skopt` native
- [ ] `tune-sklearn` (ray back-end for the above)

hazrulakmal:
#### Reference Issues/PRs
Fixes #3390 
See also #4188 & #3359
Continuation from #4251 

Implementing hyperparameter optimisation algorithm from `scikit-optimize` into sktime forecastingCV. This is a draft PR so works are in progress. 

Yes, this PR introduces soft dependency, `scikit-optimize` 

In the process of creating tests. 

#### Architecture design decision
_Sktime Side_
Decision: subclassing `BaseGridCV` rather than directly from `_DelegatedForecaster`
- to inherit usable methods like `_get_fitted_params` ,  `_update` and all the `__init__` setup
- `_fit` method has to be overwritten because the `BaseGridCV` is restrive nature, it assumes that the collection of hyperparameter combinations is defined beforehand. this is not possible for iterative or sequential algorithm where the the collection of hyperparameter combinations depends on previous iterations.

other things to consider:
1. ensure that implementation also supports `pipeline` object.

_Skopt Side_
there are two possibilities
1. (high-level abstraction)  a high-level interface to various [pre-configured optimizers](https://scikit-optimize.github.io/stable/modules/minimize_functions.html). i.e `gp_minimize` etc
3. (low-level abstaction) a low-level interface that controls `optimizer` directly. `optimizer` is a black box algorithm that tells which hyperparameter combo to search for.

Decision: go with the second option so that sktime users can interact directly with the optimizer and control how it behaves.

#### To-do lists

Some notes for myself. 
- [x] get the optimizer initialisation ready for each dict
- [x] `_evaluate_step` method - for fitting and evaluating a sample of hyperparameter from `optimizer` at each iteration. keep updating the cv result for every step. 
- [x] `_run_search` method  - running sequential hyperparameter optimisation. at each iteration, `optimizer` will suggest the next-best hyperparameters to search.
- [x] `_fit` method - reranking and updating the attributes
- [x] handle optimisation direction internally 
- [x] pytest
- [x] docstring writing 
- [x] quality_check on param_distribution
- [x] add an interface for custom iteration for each parameter space (when users give a list of dicts and want to customise the num of iterations for each parameter space)


sktime/forecasting/model_selection/_tune.py
```diff
@@ -719,3 +720,250 @@ def get_test_params(cls, parameter_set="default"):
...
+        refit=True,
+        verbose=0,
+        return_n_best_forecasters=1,
+        pre_dispatch="2*n_jobs",
+        backend="loky",
+        update_behaviour="full_refit",
+        error_score=np.nan,
+        optimizer_kwargs=None,
+    ):
+        _check_soft_dependencies(
```
fkiraly, line 756: non-blocking side note: I suppose we should think whether we can cover this with a minor update to the tag system?
We have three instances now, `scikit-learn`/`sklearn`, `dtw-python`/`dtw_python`, and this

hazrulakmal: I'm not really familiar with the tag system yet. could you point me to where to look to understand how this tagging system?. Also, to verify, did I use the `_check_soft_dependecies` and import the dependency in the class correctly?

fkiraly: oh yes, you did it correctly!
Sorry if I wasn't clear, well done inferring the pattern from the few examples (or guessing) where import path does not equal the package name.

What I meant, maybe you have a suggestion how this could be improved for developers as it's a bit clunky and an odd special case.

Regarding general information on the tag system, it's not in the developer guide, but the best references are:

* the extension templates, they contain a full documentation of the tags and what they do
* as a tutorial on "writing your own estimator", the 2 hour workshop here: https://github.com/sktime/sktime-workshop-pydata-london-2022 (on video), which also contains explanations of the tag system and the most important tags

pyproject.toml
```diff
@@ -86,6 +86,7 @@ all_extras = [
     "tslearn>=0.5.2; python_version < '3.11'",
     "xarray",
     "seasonal",
+    "scikit-optimize=0.9.0"
```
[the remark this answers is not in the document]
fkiraly: hm, not sure. Maybe it's a step that isn't run?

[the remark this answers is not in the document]
fkiraly: the typo is still in the PR, do you want to push an updated version to see whether the CI runs through?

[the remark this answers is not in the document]
hazrulakmal: not yet. in the middle of the exam rn. will do that soon. just put down my first paper today. three more to go!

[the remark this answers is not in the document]
fkiraly: all the best for your exams!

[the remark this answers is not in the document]
hazrulakmal: Thanks! I think this has been solved! the CI is not throwing an error about it anymore.

sktime/forecasting/model_selection/_tune.py
hazrulakmal: I solved the issues with softdeps tests by including a `python_dependecies` tag to the class but I'm still would like to know what kind of test the `get_test_params` does.

hazrulakmal:
Hi @yarnabrina, I'd like to bring this PR to your attention and may need some help. 

So it seems that the soft-dependency package, `scikit-optimize`, is using a depreciated attribute from numpy in their codebase and so whenever tests are run, there is a warning message about it. Although there's no harm to it, it's a bit annoying to see this warning message every time I run the pytest locally, is there a way to silence the warning message or is that not a good thing to do?

Also, I'm not quite sure why the tests only fail in version 3.11. I suspect the package is not yet compatible with python 3.11. can that be the reason?

fkiraly:
Hm, I see!
Also see my comment here: https://github.com/sktime/sktime/issues/4188#issuecomment-1585713141

I think a way to deal with this would be to limit the python version or the numpy version then in the tags (not sure what is better).

The tags are `python_version` and `python_dependencies`.

For estimator specific tests, we can use `pytest.skipif` with `_check_estimator_deps`.

sktime/forecasting/model_selection/tests/test_tune.py
```diff
@@ -3,7 +3,7 @@
 # copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
 """Test grid search CV."""
 
-__author__ = ["mloning", "fkiraly"]
+__author__ = ["mloning", "fkiraly", "hazrul"]
```
fkiraly, line 6: @hazrulakmal, no?

hazrulakmal: ops sorry about that didn't exactly remember why I put my name there. fixed it

fkiraly: no, I meant, it should be @hazrulakmal, not hazrul, because @hazrulakmal is your GitHub user name (not that you should remove your name). It's fine to add yourself as an author if you wrote significant code in the module.

hazrulakmal: I thought you meant authors are for the original author who started the module. my bad for misinterpreting. anyway, It's fine, I can add it later.

hazrulakmal:
@yarnabrina I believe that a recent change was made to the pre-commit that is not being captured in this working branch. This is causing a failure in the CI, although the pre-commit ran successfully on my local machine.

Would merging the current main branch into this working branch solve this issue? I am considering doing something like this:
```
git fetch upstream
git merge upstream/bayescv
```

yarnabrina:
`pre-commit` changes in one branch do not affect another branch, and GHA does not depend on codes in different branches. `pre-commit` fails in this PR as your code can be modified using `black`, as you can see in the logs. It is not affected by #4680.

The commands you shared tries to merge the `bayescv` branch of `upstream` remote in your local branch. Assuming you're using standard default, `origin` should be your own fork and `upstream` should be that of `sktime` repo. `bayescv` branch is unlikely to exist this repo, and will exist only in `origin`.

I'll recommend you to do the following:

1. update `main` branch of your fork to `main` of current `sktime` repo, if it's not done. You should get a [sync option from the UI itself](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/syncing-a-fork).
2. Fetch the changes in your forked repo locally with `git fetch`.
3. Merge up-to-date `main` branch of your forked repo into your `bayescv` branch, what can be done with `git merge origin/main`. This may need you to solve some conflicts.
4. Run `pre-commit` locally on all files using `pre-commit run --all-files`.
5. Commit and push changes.

Let me know if it doesn't solve the problems.

For future, I'll suggest to install `pre-commit` locally (if not done already) with `pre-commit install` so that all commits run all the hooks automatically.

sktime/forecasting/model_selection/tests/test_tune.py
fkiraly: well, strictly speaking, there is ownership per estimator (by file), the authors line in tests etc is more for acknowledgment.

pyproject.toml
```diff
@@ -86,6 +86,7 @@ all_extras = [
     "tslearn>=0.5.2; python_version < '3.11'",
     "xarray",
     "seasonal",
+    "scikit-optimize; python_version < '3.11'",
```
yarnabrina, line 89: Note: `numpy` version `1.24` does support `Python versions 3.8-3.11`. For some reason (pinning in other dependencies I guess), we are getting `1.24` only in `3.11`. When that reason is no longer applicable (i.e. when other dependencies upgrade their `numpy` pinning), how shall `sktime` be affected and what can we to resolve?

fkiraly: I think that's an interesting but separate question. We have accumulated so many soft dependencies that they are interacting with each other in the `all_extras` set - how can we untangle this? As said, separate imo (unless we are introducing a worse bound here)

fkiraly:
FYI, I added a `_check_python_version` at the start of `__init__` to deal with the 3.11 failures.

We should probably think how to address the malfunction of the dependency system, this happens whenever (a) the package and import string are different, and (b) a python version bound is given.

hazrulakmal:
> _check_python_version at the start of __init__ to deal with the 3.11 failures.

This is weird. I was baffled as to why this estimator did not capture the Python version dependency even after specifying it in the tags before you added `_check_python_version`. Shouldn't Python version dependencies be checked by `_check_estimator_deps`, which was placed in the parent class `BaseForecaster`? 

My suspicion (not entirely sure) is that it has to do with the placement of `_check_estimator_deps` in `BaseForecaster`. Shouldn't it be placed before calling the parent classes of `BaseForecaster` (i.e., before `super()` is called)?

Anyway, it seems like the reason for the failure of the test-window (3.8) was not `SkoptForecastingCV`, but rather `MovingBlockBootstrapTransformer`. or did I miss something here? Also, any idea why there is no test log for test-window (3.9)?

fkiraly:
@hazrulakmal, not sure which failures you are referring to - two CI elements are failing, but that seemed unrelated to your PR, one of the sporadic failures (pytest-xdist or memory related)

hazrulakmal:
> not sure which failures you are referring to

I was referring to CI failures in [56b6db9](https://github.com/sktime/sktime/pull/4580/commits/56b6db94e60322c11efc80c9ba324f2be86418bd) that failed due to python version dependency which then was resolved (thanks to you) after you made some edits here [d0d97ee](https://github.com/sktime/sktime/pull/4580/commits/d0d97ee724de19f8c6dc5555594d803a5e131d5a) by adding the python dependency check manually in the `__init__`.

fkiraly:
> I was referring to CI failures in [56b6db9](https://github.com/sktime/sktime/pull/4580/commits/56b6db94e60322c11efc80c9ba324f2be86418bd) that failed due to python version dependency in test python 3.11

Ah, I see.
Indeed, the base class should take care of that, but unfortunately it doesn't work the intended way if the package has an alias.

Have we opened an issue for this general problem? I vaguely remember we wanted to but cannot find it.

hazrulakmal:
I think my previous sentence may have misled you, my bad. What I meant was *python version*, not *python dependencies*. Upon reviewing the test log report, I found that the issue was not with scikit-optimize, but rather with the detection of the Python version. The error message is as follows:

> FAILED sktime/tests/test_softdeps.py::test_python_error[ForecastingSkoptSearchCV] - RuntimeError: Estimator ForecastingSkoptSearchCV has python version bound <3.11 according to tags, but does not raise an appropriate error message on __init__ for incompatible python environments. Likely reason is that __init__ does not call super(cls).__init__.

this is what makes me wonder, shouldn't this python version bound be handled by base class via the tags? or did I miss something here?

> but unfortunately it doesn't work the intended way if the package has an alias.

Agreed. The current workaround is to add `_check_soft_dependency` to handle that alias manually, rather than relying on the tagging system. This is more of a patch to the tagging system's inability to handle packages with aliases - something we want to discuss further.  I haven't yet raised an issue for this because the initial discussion was done at the end of the week before I left for holiday.

fkiraly:
Just updated from `main`, we should check that this does not constrain `pandas < 2` on 3.8-3.10 (or other unexpected constraints).

hazrulakmal:
it seems that scikit-optimize is not being installed as part of dependencies in unix-pandas1, do you happen to know why is that the case?

pyproject.toml
yarnabrina: Tracked in #4837

fkiraly:
Hm, could there be a clash with soft dependencies that imply `numpy>=1.24`? As `skopt` implies `numpy<1.24`.

sktime/forecasting/model_selection/_tune.py
hazrulakmal: resolved here #4832 and #4867

PR #4586, opened 2023-05-13
[ENH] Extend forecasting benchmarking framework to multiple metrics, add test coverage

hazrulakmal:

I recently looked through the changes made to the benchmarking module in sktime because I'm interested in extending sktime benchmarking with `kotsu`.  As I went through the codebase, I noticed a comment below in the code 

> TODO re-write evaluate to allow multiple scorers, to avoid recomputation

that there was a need to rewrite `evaluate` to handle multiple metrics but `evaluate` indeed already supports this. we just need to give the scoring argument as a list of `BaseMetric` and it will compute multiple metrics without having to refit a model n number of times for n metrics.

I've also added an extra test coverage to ensure this works properly. 

@alex-hh @DBCerigo @TNTran92 @fkiraly 
Please confirm that this changes tackle the `todo` comment

### DIFF
--- a/sktime/benchmarking/forecasting.py
+++ b/sktime/benchmarking/forecasting.py
@@ -36,10 +36,9 @@ def forecasting_validation(
     """
     y = dataset_loader()
     results = {}
+    scores_df = evaluate(forecaster=estimator, y=y, cv=cv_splitter, scoring=scorers)
     for scorer in scorers:
         scorer_name = scorer.name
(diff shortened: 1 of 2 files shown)


fkiraly (approved):
Thanks, @hazrulakmal!

Good spot - looks like a straightforward passthough.

hazrulakmal:
Thanks :)

Issue #4587, opened 2023-05-13
[BUG] AutoETS `predict_quantile` and `predict_interval` fail in sktime 0.17.2


ngupta23:
**Describe the bug**

AutoETS `predict_quantiles` and `predict_interval` fail in sktime 0.17.2, These were working in 0.17.1. Originally observed when we tried to upgrade pycaret to use sktime 0.17.2 (see: https://github.com/pycaret/pycaret/issues/3523).

**To Reproduce**

```python
!pip install sktime==0.17.2

from sktime.datasets import load_airline
from sktime.forecasting.ets import AutoETS

y = load_airline()
forecaster = AutoETS(auto=True, n_jobs=-1, sp=12)  
forecaster.fit(y)  

# Gives error
y_pred = forecaster.predict_quantiles(fh=[1,2,3])

# Also gives error
y_pred = forecaster.predict_interval(fh=[1,2,3])
```

```python-traceback
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
[<ipython-input-4-b9df560b190c>](https://localhost:8080/#) in <cell line: 1>()
----> 1 y_pred = forecaster.predict_quantiles(fh=[1,2,3])

4 frames
[/usr/local/lib/python3.10/dist-packages/sktime/forecasting/base/_base.py](https://localhost:8080/#) in predict_quantiles(self, fh, X, alpha)
    559         # we call the ordinary _predict_quantiles if no looping/vectorization needed
    560         if not self._is_vectorized:
--> 561             quantiles = self._predict_quantiles(fh=fh, X=X_inner, alpha=alpha)
    562         else:
    563             # otherwise we call the vectorized version of predict_quantiles

[/usr/local/lib/python3.10/dist-packages/sktime/forecasting/base/_base.py](https://localhost:8080/#) in _predict_quantiles(self, fh, X, alpha)
   2077 
   2078                 # compute quantile forecasts corresponding to upper/lower
-> 2079                 pred_a = self._predict_interval(fh=fh, X=X, coverage=[coverage])
   2080                 pred_int = pd.concat([pred_int, pred_a], axis=1)
   2081 

[/usr/local/lib/python3.10/dist-packages/sktime/forecasting/base/adapters/_statsmodels.py](https://localhost:8080/#) in _predict_interval(self, fh, X, coverage)
    189         valid_indices = fh.to_absolute(self.cutoff).to_pandas()
    190 
[...]

PR #4588, opened 2023-05-13
[ENH] add more test parameter sets to `AutoETS`

fkiraly:
This adds a second test parameter set to `AutoETS`.

Towards https://github.com/sktime/sktime/issues/3429

Related: https://github.com/sktime/sktime/issues/4587, as the second set has `auto=True`

### DIFF
--- a/sktime/forecasting/ets.py
+++ b/sktime/forecasting/ets.py
@@ -466,12 +466,14 @@ class AutoETS(_StatsModelsAdapter):
         -------
         params : dict or list of dict
         """
+        # default setting, non-auto
         params1 = {}
-        params2 = {
-            "sp": 2,
-            "auto": True,
+        # "auto-ets"
+        params2 = {"sp": 2, "auto": True}
+        # ets (non-auto) with some non-default parameters
+        params3 = {
             "information_criterion": "bic",
             "trend": "add",
             "damped_trend": True,
         }
-        return [params1, params2]
+        return [params1, params2, params3]
--- a/sktime/forecasting/ets.py
+++ b/sktime/forecasting/ets.py
@@ -450,3 +450,28 @@ class AutoETS(_StatsModelsAdapter):
         https://www.statsmodels.org/dev/examples/notebooks/generated/ets.html
         """
         return self._fitted_forecaster.summary()
+
+    @classmethod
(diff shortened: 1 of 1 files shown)


fkiraly:
FYI @ngupta23, @yarnabrina - we expect this to fail with the bug that @ngupta23 reported in #4587. We can then develop against the failing tests.

fkiraly:
surfaces the error as expected.

yarnabrina:
While the `auto` error needs to be fixed, I've some doubts on the added test case.

1. `sp` is seasonal periodicity right? So, `sp=2` means half-yearly? When random data is generated during testing, is it possible to ensure they will be half-yearly?
2. If `auto=True` is passed, all the other 3 arguments, `trend` and `damped_trend` are ignored. Basically, it'll run all valid values and chhose best, if I understand correctly. So, may be we can split this test case into 2, one with just `auto` as `True`, and the other with remaining parameters?

And then a question regarding `auto`. I haven't seen the `R` code, but to me it seems to be the case that if user passes a parameter along with `auto=True`, user does not want to test variation of that parameter itself. e.g. if I pass `trend` as `mul`, I'd expect `auto` to find best among rest hyperparameters keeping `trend` as `mul`. Will appreciate your view on this.

fkiraly:
> 1. sp is seasonal periodicity right? So, sp=2 means half-yearly? When random data is generated during testing, is it possible to ensure they will be half-yearly?

I think it does not really matter whether it matches the data, what is important is just that the code is run with data of the right shape. Usually, `sp=1` runs a different logic branch than `sp>1`.

> 2. If `auto=True` is passed, all the other 3 arguments, `trend` and `damped_trend` are ignored. Basically, it'll run all valid values and chhose best, if I understand correctly. So, may be we can split this test case into 2, one with just `auto` as `True`, and the other with remaining parameters?

Is that so? If yes, it is nowhere properly documented!
This should be clarified in the docstring.
It is clear from the code, but I looked at the docstring only.

In general, parameters that are ignored if another is set to a value, it should be said "ignored if (condition)".

I will split the test case, would you like to open an issue on the problem with the docstring? (I can also do it, but since it came out of your observation)

fkiraly:
There should probably be a warning too, if the parameters are set to non-defaults.

PR #4590, opened 2023-05-15
[BUG] BaseGridCV ranking sorted wrongly

hazrulakmal:

While I was working on my `skopforecastingCV`, I noticed a little tiny bug in BaseGridCV. On line 221, a rank column is populated based on scoring tag `lower_is_better`. A few lines of code later, BaseGridCV updates its n_best_scores and n_best_forecasters. so to do this, the same table is then sorted based on the rank column created in line 221, but with ascending order following the scoring tag again. This works well for metrics where lower is better, which is the case for all forecasting metrics as far as I know lol  (minimum is better). However, it would be incorrect for the opposite case. To give you an example:

| column_1  | rank |
| ----------- | ---- |
| 0.527870   | 2.0  |
| 0.677122   | 1.0  |
|  0.060541  | 3.0  |

if ascending = False, the tag for when higher is better, the algorithm is fitting the n **worst**  forecaster rather. 
| column_1  | rank |
| ----------- | ---- |
|  0.060541  | 3.0  |
| 0.527870   | 2.0  |
| 0.677122   | 1.0  |

### DIFF
--- a/sktime/forecasting/model_selection/_tune.py
+++ b/sktime/forecasting/model_selection/_tune.py
@@ -243,7 +243,8 @@ class BaseGridSearch(_DelegatedForecaster):
 
         # Sort values according to rank
         results = results.sort_values(
-            by=f"rank_{scoring_name}", ascending=scoring.get_tag("lower_is_better")
+            by=f"rank_{scoring_name}",
+            ascending=True,
         )
         # Select n best forecaster
         self.n_best_forecasters_ = []


fkiraly (approved):
Oh, I see - doing this twice would cancel each other out, leaving the case of "higher is better" with the wrong sign.

Excellent spot - I wonder how you noticed, if as you say none of the common metrics are "higher is better".

hazrulakmal:
I spotted it as I went through the code logic cus I used the same block of code for `skoptforecastingCV` too.

fkiraly:
I see, @hazrulakmal.
Maybe a thing to note, we should try avoid copy-pasting code (the so-called DRY principle). Of course we can try to reduce the duplication in a second go rather than from the start.

hazrulakmal:
yeah, I had that in mind. I was thinking about once this current work is completed. I will put another PR to refactor _tune.py. I don't want to directly make changes in the original `BaseSearchCV` in this PR yet. May wanna think about making it more abstract cus the current shape only fits well with the grid search only.

fkiraly:
> May wanna think about making it more abstract cus the current shape only fits well with the grid search only.

Yes, that's an excellent plan.

I have also been thinking about how to refactor the tuners so it fits all, this needs to account for iterative search, not just independent/grid search.
Would you like to open an issue with the topic to refactor the tuning algorithms, we can develop designs there.

hazrulakmal:
@fkiraly Certainly! I will open an issue to initiate a comprehensive discussion on enhancing and refactoring the sktime tuners. Additionally, I think it is worth exploring possibilities for extending the tuner's capabilities to encompass other tasks such as classification (which I noticed is currently in progress by you), clustering, and any other relevant areas as deemed appropriate.

PR #4596, opened 2023-05-17
[ENH] grid search for time series classification

### LINKED ISSUE #4129: [ENH] `GridSearchCV` for time series classification
There are some good reasons to write an `sktime` native grid search for time series classification (and regression):

* discussion in https://github.com/sktime/sktime/issues/4128 which highlights that classifiers not compatible with `pickle` persistence cannot be combined with `sklearn` grid search (but the `sktime` serialization interface could replace this)
* input/output checks would wokr properly, and more mtypes such as `pd-multiindex` would become compatible with the grid search
* this would provide a basis for scaling and distributed computing, e.g., via `dask`, see https://github.com/sktime/sktime/issues/4013

Opening this for design discussion.

Of course, `sklearn` `GridSearchCV` would remain compatible with `sktime` when used with `sklearn` compatible mtypes and estimators that are persistence compatible with `sklearn` contracts.

fkiraly:
This adds grid search for TSC, "quick hack" implementation that wraps `sklearn` `GridSearchCV`.

Partially towards https://github.com/sktime/sktime/issues/4129 - I do not consider this the ultimate form as this is a hacky interface, so we should not close #4129.

Internally, converts the data to a "rows = instances" format, so the `sklearn` grid search can cope with it.

Note: `sklearn` grid search can already be used if the data is passed in such a format, but it loses inheritance from `sktime` `BaseClassifier` and therefore all the boilerplate.

Depends on https://github.com/sktime/sktime/pull/5697, https://github.com/sktime/sktime/pull/5700


achieveordie:
Is this up for grabs or are you still working on it?

fkiraly:
up for grabs for now, please take it - I'm working on the releases now.

sktime/classification/model_selection/_tune.py
```diff
@@ -0,0 +1,404 @@
...
+            for specifications, see examples/AA_datatypes_and_datasets.ipynb
+
+        Returns
+        -------
+        y : 2D array of shape [n_instances, n_classes] - predicted class probabilities
+            1st dimension indices correspond to instance indices in X
+            2nd dimension indices correspond to possible labels (integers)
+            (i, j)-th entry is predictive probability that i-th instance is of class j
+        """
+        if not self.refit:
```
achieveordie, line 347: @fkiraly Is there a reason we're directly calling the base estimator's method and not `estimator`'s ? (return `self.estimator._predict_proba(X)`)

achieveordie, line 347: @fkiraly Is there a reason we're directly calling the base estimator's method and not estimator's? (`return self.estimator._predict_proba(X)`)

fkiraly: if `refit=False`, then the `sklearn` grid search (see the same parameter there) does not compute probabilistic predictions and will raise an exception. In this case, we instead need to carry out the default logic in `BaseClassifier`, which lives in `_predict_proba`.

Or, `predict_proba`, there may be a minor issue in calling the private, not the public method, but I must have thought the private one is ok.

achieveordie: Let me rephrase my question, let's say that `refit=False`. In this case, if we were to call `predict()` (or `predict_proba()`) on it, shouldn't it simply call the underlying estimator's `predict` (or `predict_proba()`)? In sklearn's implementation, `predict()` only happens when `refit=True` so we're already diverging from the expected contract - we at least need to add more documentation if we're making a conscious decision to diverge.

achieveordie:
> Let me rephrase my question, let's say that refit=False. In this case, if we were to call predict() (or predict_proba()) on it, shouldn't it simply call the underlying estimator's predict (or predict_proba())? In sklearn's implementation, predict() only happens when refit=True so we're already diverging from the expected contract - we at least need to add more documentation if we're making a conscious decision to diverge.

@fkiraly I'm still waiting for a response - please Let me know what's your opinion about it so I could resume working on this PR.

sktime/classification/model_selection/_tune.py
fkiraly: apologies for the delay in responding. pydata, travels etc.

Re predict dispatch, I'm not sure - as said, if `refit=False`, a call to `sklearn`  `predict_proba` will raise an exception, so we have to do sth else.

A call to `predict` will dispatch to the underlying `predict`, as you suggest.

So, is that not what you think should happen? Maybe let's discuss in one of the dev meet-ups.

fkiraly:
apologies, I did not spot this. Thanks for the reminder, see my answer above.

sktime/classification/model_selection/_tune.py
fkiraly: from discussion with @achieveordie on discord: @achieveordie was right, `refit=False` impacts not just `predict_proba` but also `predict` - it was my mistake and confusion to think it only affects `predict_proba`.

Solved now by not testing `refit=False`, as that will violate the classifier contract (non-functional `predict`)

PR #4597, opened 2023-05-17
[DOC] `AutoETS` docstring - clarify conditional ignore of parameters dependent on `auto`

### LINKED ISSUE #4592: [DOC] behaviour of `auto=True` to cause suppressing other user inputs is not documented in `AutoETS` docstring
If `auto=True` is provided in `AutoETS`, then `error`, `trend`, `seasonal` and `damped` is ignored. Based on other parameters, all possible models are trained and automatically the best model is decided based on `information_criterion`.

The docstring should be updated to explicitly note this behaviour, as user passing a subset of these 4 parameters along with `auto=True` may expect that `auto` is being applied on remaining parameters and their provided subset is being respected.

We can consider throwing an warning to the user as well. This can be done always at start, or only in case best model varies from user provided subset.

See #4588 for related discussion.

luca-miniati:
#### Reference Issues/PRs
Fixes #4592

- added docstring for auto=True
- added a warning for passing auto=True alongside error, trend, dampen_trend, and/or seasonal

No

First pull request here, any feedback is greatly appreciated!

No official tests, but I tested the functionality of the warning locally (it worked)

#### Any other comments?
I couldn't figure out how to add myself to the list of contributors, how can I do that?


fkiraly:
> I couldn't figure out how to add myself to the list of contributors, how can I do that?

Add yourself in the file `.all-contributorsrc` in the root folder, follow the pattern. This PR qualifies you for the `doc` and the `code` badge :-)

I'll also update the PR template with that information, since it seems this is not a rare problem (i.e. "how to")

fkiraly:
Done, here: https://github.com/sktime/sktime/pull/4599
hope the explanation helps.

fkiraly (changes requested):
Thanks, great!

Left minor suggestions for change above.

luca-miniati (commented):
Ok, fixed

sktime/forecasting/ets.py
```diff
@@ -238,7 +239,12 @@ def __init__(
         self.n_jobs = n_jobs
 
         super(AutoETS, self).__init__(random_state=random_state)
-
+        
+        if self.auto and (any(param for param in [trend, damped_trend, seasonal]) or error=="add"):
+            warnings.warn(
+                "Warning: The user-specified parameters provided alongside auto=True in AutoETS may not be respected. The AutoETS function automatically selects the best model based on the information criterion, ignoring the error, trend, seasonal, and damped parameters when auto=True is set. Please ensure that your intended behavior aligns with the automatic model selection."
```
yarnabrina, line 245: Small typo: the argument name for `AutoETS` is `damped_trend`, not `damped`.

yarnabrina, line 245: lint comment: this line is quite long, and my guess is that it'll trigger code quality check, especially `flake8` etc. Can we split this into multiple lines?

E.g. these two are equivalent:

```python3
text_1 = "assume it is a long sentence"
text_2 = (
    "assume it is"
    " a long "
    "sentence"
)

assert text_1 == text_2
```

sktime/forecasting/ets.py
```diff
@@ -119,7 +119,8 @@ class AutoETS(_StatsModelsAdapter):
     return_params : bool, default=False
         Whether or not to return only the array of maximizing parameters.
     auto : bool, default=False
-        Set True to enable automatic model selection.
+        Set True to enable automatic model selection. If auto=True, then error,
+        trend, seasonal and trend_damped are ignored.
```
yarnabrina, line 123: Small typo: the argument name for AutoETS is `damped_trend`, not `trend_damped`.

fkiraly (changes requested):
Thanks for the changes.

I think there is a small (but consequential) mistake in the condition related to `error`, can you check?

luca-miniati (commented):
Thanks for the help, fixed typos and errors

sktime/forecasting/ets.py
```diff
@@ -238,7 +239,15 @@ def __init__(
         self.n_jobs = n_jobs
 
         super(AutoETS, self).__init__(random_state=random_state)
-
+        
+        if self.auto and (any(param for param in [trend, damped_trend, seasonal]) or error!="add"):
```
yarnabrina, line 243: May be I am missing something, can you please explain this check? What is the goal here - whether user has explicitly passed these 4 parameters, or something else?

yarnabrina: I'll explain my confusion. If I am not wrong, these parameters will always be present whether or not user explicitly pass the values. If user specifically pass a value, then I felt it's wrong for `auto` to change it without informing user. Otherwise, if user don't pass anything and use the default settings, do we still need to throw a warning?

This may be some design discussion, as I am not familiar how this sort of thing is being handled elsewhere in `sktime`, so cc @fkiraly.

fkiraly: I also got confused first, but here is what I think it checks (disjunction of):

* one of `trend`, `seasonal` are not `None`
* `damped_trend=True`
* `error` is not `"add"`

For each of these parameters, that's the condition "is not the default", so it would fire if the user has provided a non-default parameter to any of the affected parameters that `auto` overrides/ignores.

luca-miniati: This is what I was intending, is there a different way it should be written?

fkiraly: (I thought it was fine, just obscure)

yarnabrina: One minor code suggestion: `any(<list>)` should be enough I think, `any(element for element in <list>)` is probably not required. Please check.

But mainly I have some feedback from user point of view. As a user, if I specify `damped_trend=False` explictly in function call, I expect that to be respected. It may be default value as well, but I want to ensure that it does not get modified because I also pass `auto=True`. If that happens, I'd at least want to be informed.

Now I understand that whether user has passed `False` or it's taken from default will be difficult to check (may be even impossible, I have never tried). But this was my consideration for the feedback.

sktime/forecasting/ets.py
```diff
@@ -244,7 +244,7 @@ def __init__(
             # If auto=True, check if trend, damped_trend, seasonal, or error have been passed in by user
             if any([trend, damped_trend, seasonal, error]):
                 warnings.warn(
-                    "Warning: The user-specified parameters provided alongside auto=True in AutoETS may not be respected. "
```
luca-miniati: "Warning:" is redundant

sktime/forecasting/ets.py
```diff
@@ -180,9 +180,9 @@ class AutoETS(_StatsModelsAdapter):
 
     def __init__(
         self,
-        error="add",
+        error=None,
```
luca-miniati, line 183: Here I changed default values for error and damped_trend, so that the block of logic later can tell whether or not they have been passed in by the user.

sktime/forecasting/ets.py
```diff
@@ -242,7 +242,7 @@ def __init__(
         
         if self.auto:
             # If auto=True, check if trend, damped_trend, seasonal, or error have been passed in by user
-            if any([trend, damped_trend, seasonal, error]):
+            if any(param is not None for param in [trend, damped_trend, seasonal, error]):
```
luca-miniati, line 245: Accounts for user passing in False for certain params.

fkiraly: I disagree with this now, as this will also raise a warning for some default settings - this would confuse the user.

fkiraly (changes requested):
Thanks!

Though, this would now raise warnings for just setting `auto=True`, I think? That should be fixed.

luca-miniati:
> Though, this would now raise warnings for just setting `auto=True`, I think? That should be fixed.
See comment on commit, I ran tests that seem to pass using auto=True

sktime/forecasting/ets.py
```diff
@@ -35,13 +35,13 @@ class AutoETS(_StatsModelsAdapter):
...
     ----------
     error : str, default="add"
-        The error model. Takes one of "add" or "mul".
+        The error model. Takes one of "add" or "mul". Ignored if auto=True. 
     trend : str or None, default=None
-        The trend component model. Takes one of "add", "mul", or None.
-    damped_trend : bool, default=False
-        Whether or not an included trend component is damped.
+        The trend component model. Takes one of "add", "mul", or None. Ignored if auto=True. 
+    damped_trend : bool, default=None
```
fkiraly, line 41: even if internally translated, for the user it should be explained as `False`

sktime/forecasting/ets.py
```diff
@@ -238,7 +239,29 @@ def __init__(
...
+                warnings.warn(
+                    "The user-specified parameters provided alongside auto=True in AutoETS may not be respected. "
+                    "The AutoETS function automatically selects the best model based on the information criterion, ignoring "
+                    "the error, trend, seasonal, and damped_trend parameters when auto=True is set. Please ensure that your "
+                    "intended behavior aligns with the automatic model selection."
+                )
+        else:
+            # If auto=False, check if damped_trend has been passed by user
+            if damped_trend:
+                self.damped_trend = damped_trend
```
fkiraly, line 255: this will break the `sktime` interface contract.
If you read the extender contract, attributes of self should always have the same value as the `__init__` inputs.

fkiraly (changes requested):
Thanks!

What is very important though, you need to adhere to the `sklearn`/`sktime` interface contracts, and one important part of this is that `self.paramname` must always be the same value as `paramname` as passed to `__init__`. What you have most recently changed would break this contract (and fail the tests, if you use `check_estimator`) - kindly don't do that.

An alternative would be writing a conditional value to `self._damped_trend` etc and use that in `statsmodels`.

sktime/forecasting/ets.py
fkiraly: good idea and sth we often do, but see below on the "don't change init params" requirement

sktime/forecasting/ets.py
fkiraly: ah, yes, that makes sense.

yarnabrina:
It seems my feedback of distinguishing between user passed and default values made things unnecessarily complicated. Sorry for that, but I felt users may want to know that as there is no other (easy) way (as far as I know) to detect what combination the final model was actually trained on.

Anyway, since @fkiraly was okay with changes, and so far no one ever complained about their passed values not being respected, I do not feel very strongly about my feedbacks.

So, I suppose if @luca-miniati now reverts to the check+warning as it was before, it'd be all good again?

fkiraly:
> It seems my feedback of distinguishing between user passed and default values made things unnecessarily complicated.

Well, at least it turns it from a documentation task into an implementation task which invites the complications, yes.

I'd be fine with either solution, i.e., having the weaker condition (non-default rather than user specified) covered by a warning.

luca-miniati:
> I would prefer 1., as it does not change the logic too much.

Got it, thanks for the help everyone

luca-miniati:
Ok, I believe everything is now in order.

fkiraly (approved):
Yes, looks good!

luca-miniati:
I'm confused by the checks, what are they and how do I fix them?

fkiraly:
> I'm confused by the checks, what are they and how do I fix them?

These are linting aka code formatting checks.
This means, the code is not formatted according to the `sktime` conventions.

See here for the guide on how to set up automatic code formatting on your computer:

You can also go on "details" in the remote CI, i.e., next to the failing test, and at the bottom it should tell you what is failing.

fkiraly:
(hint: it has to do with whitespaces and newline spacings)

luca-miniati:
Thanks for the help, I got pre-commit up and running so it should be good now.

fkiraly (approved):
indeed, all good! Thanks for your contribution!

fkiraly:
anything from your side, @yarnabrina?

yarnabrina (approved):
Looks good 👍

PR #4600, opened 2023-05-17
Revert "[MNT] bound `holidays` to avoid error in `Prophet`"

### LINKED ISSUE #4595: [BUG] incompatibility of `Prophet` with `holidays 0.25.0`
`Prophet` with `holidays 0.25.0` crashes in the CI since the `holidays` release - the location of the error or its nature is unclear.

For now, this has been solved by bounding `holidays`, but we should investigate the issue and potentially fix it, if it is localized in `sktime`. At least, it could be caught instead of leaving to an error in contsructing the estimator.

Reference: https://github.com/sktime/sktime/pull/4594

yarnabrina:
Reverts sktime/sktime#4594.
Fixes #4595.

`prophet` released a new version [`1.1.3`](https://github.com/facebook/prophet/releases/tag/v1.1.3) today. It should help to release this temporary pin of `holidays`.


yarnabrina:
@fkiraly Not sure why it says some checks are incomplete. https://github.com/sktime/sktime/pull/4600/commits/f73f95dbdf459a45265f1486b03295b5c64e2589 passed all but read the docs workflow, which was auto-cancelled. But that was succesful in https://github.com/sktime/sktime/pull/4600/commits/dc22cd2c872b6d5b62b98300d73eda679be009ac.

So, a bit confused on this, but based on my understanding, it should be ready to merge now with no more need of pinning holidays.

fkiraly:
I believe it is some interaction between the "automatically update contributors" workflow and the auto-cancel workflow, which prevents the tesets from being displayed properly.

See https://github.com/sktime/sktime/pull/3807 and discussion therein.

fkiraly:
same problem here, btw https://github.com/sktime/sktime/pull/4583

PR #4604, opened 2023-05-17
[DOC] Add SVG version of the sktime logo

marrov:
Tiny PR that adds a vectorized version of the sktime logo (previously only png was available).


fkiraly (approved):
thanks

PR #4608, opened 2023-05-17
[ENH] `BaseObject` boilerplate to set parameters to `self` at start of `__init__`

fkiraly:
This PR simplifies `__init__` boilerplate and extension templates as follows.

The `sklearn` and `sktime` contracts requires that all `__init__` parameters are written to `self` in `__init__`, and never be mutated, e.g., `self.a=a`, `self.b=b` etc.

This can become cumbersome, error-prone, and hard to review if there are dozens of parameters.
That is not a hypothetical case, but the state of affairs for many, if not most `sktime` estimators, especially those interfacing `statsmodels` or time series classifiers that are highly composite.

This PR proposes to reduce boilerplate and simplifies the extender contract by using the same line everywhere, `self._set_params_from(locals())`, which never needs to be changed.

As a demonstration, this refactors a larger number of problematic forecasters, and changes the extension templates for forecasters.

If this is agreed, I'd finish the work for other learning tasks, and eventually we may like to port this functionality into `scikit-base`.

### DIFF
--- a/extension_templates/forecasting.py
+++ b/extension_templates/forecasting.py
@@ -165,13 +165,16 @@ class MyForecaster(BaseForecaster):
         # estimators should precede parameters
         #  if estimators have default values, set None and initalize below
 
-        # todo: write any hyper-parameters and components to self
-        self.est = est
-        self.parama = parama
-        self.paramb = paramb
-        self.paramc = paramc
+        # this should be the first line and not change
+        self._set_params_from(locals())
+        # convenience function that does self.est = est, self.parama = parama, etc
+        # parameters of same name as __init__ args should never be overwritten!
+        # if parameters are generated, write to other attr names, such as self._parama
+        # estimators of same name as __init__ args should never be mutated!
+        # for internal use, make a clone later, e.g., self._est = est.clone()
 
         # todo: change "MyForecaster" to the name of the class
+        # this should be the second line and not be changed except the class name
         super(MyForecaster, self).__init__()
 
         # todo: optional, parameter checking logic (if applicable) should happen here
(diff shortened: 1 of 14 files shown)


fkiraly:
Hm, this does not seem to work when calling `super.__init__` - why that?

yarnabrina:
I just came acros this PR by accident today. If editing the `__init__` methods of all estimators is an option, can I suggest using `attrs` (already a soft-dependency) or `dataclasses` (part of standard library)?

Not familiar with every single feature of it, but repetitive `self.a = a` etc. problem is definitely solved by it.

Ref. https://docs.python.org/3/library/dataclasses.html

fkiraly:
`dataclass` was already one of the ideas that came up, however:

* as far as I remember, the decorator interacts problematically with some access patterns to the class, although I do not remember which/if. I vaguely remember it was use of `abc` in some parts of the library.
* the `sktime` `__init__` patterm assumes a `super.__init__` call after the attr writing, and potentially further code after that in relation to parameter validation and tag setting. I think that's not compatible with `dataclass`?

fkiraly:
Btw, @yarnabrina, do you understand what happens in `ThetaForecaster`, in the `super.__init__` call? It does not seem to write the parameters coming from `ExponentialSmoothing` to the `self` of `ThetaForecaster`.

fkiraly:
found the issue - it was because we were writing only parameters in `self.get_param_names` - which for an estimator that inherits, might not actually agree with the parameters of the class the method is called from!
Fixed by removing that condition.

fkiraly:
old failure was due to writing not enough to `self` (writes from parent class).
new failure is writing too much - any `self` attributes.

I start thinking, this cannot be made compatible with inheritance...

PR #4621, opened 2023-05-19
[ENH] better documentation and test coverage for custom estimators and parameters in `DrCIF`

Taise228:
Update: this PR improves documentation and test coverage around custom estimators and estimator parameters in `DrCIF`.

Previously, it was not very clear to users how to set custom estimators and/or tune them, and the custom estimator case was not tested.

This PR improves the docstring of `DrCIF` in that respect, and adds a test case with a custom estimator.

Original PR description by @Taise228:

---

#### Reference Issues/PRs

None.

This PR is to make it possible to set the maximum depth to each tree of DrCIF.   
Currently we cannot set max_depth to them, and it will be set to ```sys.maxsize``` if using ```ContinuousIntervalTree``` (see https://github.com/sktime/sktime/blob/cc6085ca8d11096ab9df0f06063150d81a40eb3b/sktime/classification/sklearn/_continuous_interval_tree.py#L88).   
Setting max_depth to each tree is important to avoid overfitting.   
After this PR, you will be able to set ```max_depth``` as   
```model = DrCIF(max_depth=10)```   
This will set ```max_depth=10``` for all trees contained in model.

No

```sktime/classification/interval_based/_drcif.py```

No.

##### For all contributions

##### For new estimators

### CODE REVIEW

**Review by fkiraly — CHANGES_REQUESTED** [review]
Thanks for the addition!
I see the problem that it is unclear how to set or tune the parameters of the CIT, it is a good idea to try addressing this.

However, there is already a way to do that, namely you pass the `ContinuousIntervalTree(max_depth=your_max_depth)` as the parameter `base_estimator`.

However:

* this is not clear from the docstring. Might be worth adding a note: "if you want to set or tune parameteres of `base_estimator`, it is recommended to pass an `sklearn` or `sktime` estimator rather than a string". Might also be useful to add, in the description of "DTC" and "CIT", the precise locations in those packages
* There are also some issues with how the parameters are set. If `base_estimator` is an estimator, then `self._base_estimator` should be a clone rather than a reference
* the logic for setting parameters could be moved to after the `super` call in `__init__`, it would already raise the error at construction rathere than later at fit.

What do you think?

**Comment by fkiraly** [comment]
Oh, and code quality checks are failing.
This means, the code is not formatted according to the `sktime` conventions.

See here for the guide on how to set up automatic code formatting on your computer:

You can also go on "details" in the remote CI, i.e., next to the failing test, and at the bottom it should tell you what is failing.

**Comment by Taise228** [comment]
Thanks for reviewing!

Actually I didn't realize that I can pass parameters to base trees by directly passing an instance to ```base_estimator```.
It should not be desirable to only deal with ```max_depth```, so I agree with you that the current implementation is better. Simply adding some docstring will be enough.

However now I realize that if I pass a BaseEstimator instance to ```base_estimator```, then ```AttributeError: 'ContinuousIntervalTree' object has no attribute 'lower'``` will be called. So we have to process it correctly.

Also as you mentioned, if an instance of BaseEstimator is passed to ```base_estimator```, ```self._base_estimator``` should be clone. 

Considering those, I'll push a new commit.

**Comment by Taise228** [comment]
How about this?

I made it to have ```self._base_estimator``` in __init__ right after super call

**Comment by fkiraly** [comment]
> Actually I didn't realize that I can pass parameters to base trees by directly passing an instance to ```base_estimator```.

Yes, that's what I meant that it is not obvious.

First, I was thinking, well, you should be able to pass the estimator rather than the string.
Then I noticed you can actually do that.
Therefore I thought, well, the docstring should explain this better.

> However now I realize that if I pass a BaseEstimator instance to ```base_estimator```, then ```AttributeError: 'ContinuousIntervalTree' object has no attribute 'lower'``` will be called. So we have to process it correctly.

Sounds like a bug.

Could you add a test parameter set to `get_test_params` where `base_estimator` is an estimator, so this gets caught by the tests?

File: sktime/classification/interval_based/_drcif.py [hunk 1]
```diff
@@ -193,26 +195,26 @@ def __init__(
...
 
         super(DrCIF, self).__init__()
 
+        if isinstance(base_estimator, str):
+            if base_estimator.lower() == "dtc":
+                self._base_estimator = DecisionTreeClassifier(criterion="entropy")
+            elif base_estimator.lower() == "cit":
+                self._base_estimator = ContinuousIntervalTree()
+        elif isinstance(base_estimator, BaseEstimator):
+            self._base_estimator = copy.deepcopy(base_estimator)
```
fkiraly, line 207: using `_clone_estimator` `sktime.base._base` should be safer here

fkiraly:
> How about this?
> I made it to have ```self._base_estimator``` in ```__init__``` right after super call

Yes, that's what I meant. Although using `clone` is safer, as `deepcopy` also copies the state and you only want the blueprint copied.

Btw, you can also get the `bug` badge because of finding the bug if an estimator (not string) is passed.

sktime/classification/interval_based/_drcif.py
Taise228: I changed here

Taise228:
Ok I pushed a new commit which changed

- use ```_clone_estimator``` instead of ```copy.deepcopy```
- added ```base_estimator``` key to the return of ```get_test_params```

is it good now?

fkiraly:
yes, if the tests run through.

fkiraly (changes requested):
Thanks for the changes.

I just went through the code and would have some requests, apologies for not mentioning this earlier.

* in the docstring, can you mention the exact module locations in `sktime`, `sklearn`, for DTC, CIT?
* in the `get_test_params`, please add a new parameter set instead of overwriting the old one. That is, in the "else" branch, add the new dict in a 2-element list together with the old dict.

Taise228:
Now I pushed my changes at

- [docstring of main](https://github.com/Taise228/sktime/blob/4012b739c914c3f802083fc77f8836297921efdf/sktime/classification/interval_based/_drcif.py#L64)
- [return of get_test_params](https://github.com/Taise228/sktime/blob/4012b739c914c3f802083fc77f8836297921efdf/sktime/classification/interval_based/_drcif.py#L638)

If I am not getting what you meant, please tell me. Anyway I appreciate your kind instruction so much. Thanks!

fkiraly (approved):
That's exactly what I meant, thanks!

PR #4589, opened 2023-05-13
[BUG] in `_StatsModelsAdapter`, avoid passing `exog` to `get_prediction` of `statsmodels` in `_predict_interval` if parameter is not supported

yarnabrina:
Fixes #4587.
Depends on #4588.

Avoid passing exogenous variables to `get_prediction` if it is not supported.

PR #4623, opened 2023-05-19
[MNT] address passing of unsupported `random_state` parameter in `_StatsModelsAdapter`

fkiraly:
This PR addresses a deprecation warning from `statsmodels`.

The `_StatsModelsAdapter` may pass a `random_state` parameter that does nothing. This was accepted as `kwargs`, but will raise an exception in the future, and raises a deprecation warning currently.

This PR prevents it from being passed by `_StatsModelsAdapter` if it isn't a recognized parameter.

### DIFF
--- a/sktime/forecasting/base/adapters/_statsmodels.py
+++ b/sktime/forecasting/base/adapters/_statsmodels.py
@@ -190,12 +190,14 @@ class _StatsModelsAdapter(BaseForecaster):
 
         get_prediction_arguments = {"start": start, "end": end}
 
-        if hasattr(self, "random_state"):
+        def _pred_has_arg(x):
+            get_pred_args = inspect.signature(self._fitted_forecaster.get_prediction)
+            return x in get_pred_args.parameters.keys()
+
+        if hasattr(self, "random_state") and _pred_has_arg("random_state"):
             get_prediction_arguments["random_state"] = self.random_state
 
-        if inspect.signature(self._fitted_forecaster.get_prediction).parameters.get(
-            "exog"
-        ):
+        if _pred_has_arg("exog"):
             get_prediction_arguments["exog"] = X
 
         prediction_results = self._fitted_forecaster.get_prediction(
--- a/sktime/forecasting/structural.py
+++ b/sktime/forecasting/structural.py
@@ -277,7 +277,18 @@ class UnobservedComponents(_StatsModelsAdapter):
         self.flags = flags
         self.low_memory = low_memory
 
-        super(UnobservedComponents, self).__init__(random_state=random_state)
+        if random_state is not None:
+            from warnings import warn
+
+            msg = (
+                "the random_state parameter of UnobservedComponents does not "
+                "have any effect, it is deprecated in sktime and will be removed "
+                "in 0.20.0, "
+                "as statsmodels UnobservedComponents does not have this parameter"
+            )
+            warn("", stacklevel=2)
+
+        super(UnobservedComponents, self).__init__()
 
     def _fit_forecaster(self, y, X=None):
         """Fit to training data.
--- a/sktime/forecasting/structural.py
+++ b/sktime/forecasting/structural.py
@@ -276,22 +276,8 @@ class UnobservedComponents(_StatsModelsAdapter):
         self.optim_hessian = optim_hessian
         self.flags = flags
         self.low_memory = low_memory
-        self.random_state = random_state
 
-        # todo 0.20.0: remove random_state parameter
-        # also docstring references and logic
-        if random_state is not None:
-            from warnings import warn
-
-            msg = (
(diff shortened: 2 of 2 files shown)


fkiraly:
FYI @yarnabrina, requesting review as you have recently worked on this.

yarnabrina (commented):
Wasn't aware of the deprecation. Changes in `_statsmodels.py` looks good.

How is dynamic time warping related to `random_state`? A bit confused.

fkiraly:
oops, it's not - commit accident by having too many vs code windows open and simultaneous PR going on. Reverted.

fkiraly:
removed the accident, would appreciate re-review, @yarnabrina

yarnabrina (commented):
Looks good to me 👍🏽 

However, just to clarify a confusion, please see https://github.com/sktime/sktime/issues/2388. I'm not certain of deterministic behaviour of `ETSModel`, but it was specifically requested to add a `random_state`. If you see [this](https://www.statsmodels.org/stable/_modules/statsmodels/tsa/exponential_smoothing/ets.html), `random_state` is indeed used for simulation, though I'm not sure if we are indirectly using this method or not while calling `get_prediction`.

If we are using the `simulate` method indirectly, there is a problem. `random_state` is absent in arguments of `get_prediction`, abstracted away by `**kwargs`, so `inspect` won't detect it.

```pycon
>>> 
>>> inspect.signature(ETSModel(load_airline()).fit(disp=False).get_prediction)
<Signature (start=None, end=None, dynamic=False, index=None, method=None, simulate_repetitions=1000, **simulate_kwargs)>
>>> 
>>> "random_state" in inspect.signature(ETSModel(load_airline()).fit(disp=False).get_prediction).parameters.keys()
False
>>> 
```

I'm probably being a bit too cautious, but I'm not confident on this part for `AutoETS`: `_pred_has_arg("random_state")`. And, similar behaviour can be present for other inherited estimators as well. May be we can ask others (e.g. @ngupta23 who raised that issue) if they are more certain of non-deterministic nature?

fkiraly:
I stumbled upon this when a warning was raised - I did not dig down which of the `statsmodels` estimators it really was, but it will cause an exception once `statsmodels` updates.

Is there an easy way to inspect which models have the arg?

yarnabrina:
> Is there an easy way to inspect which models have the arg?

I'm not aware of.

I'll take a look at #4589 logs, I'm guessing that's where you found the deprecation warnings. If I can think of something, will update.

fkiraly:
any good ideas, @yarnabrina?

I think you are right, we cannot merge this as-is since the argument inspection does not work (so the argument is never passed)

yarnabrina:
I do not, but here's a thought:

> if hasattr(self, "random_state") and _pred_has_arg("random_state"):

Is it "safe" to assume that two check correspond to one another, i.e. if `self.random_state` exists, the algorithm allows `random_state` as input and should not fail if passed to `get_prediction` (and vice versa)?

If yes, we can drop `random_state` argument from `structural.py`, as it's missing in both [`__init__`](https://www.statsmodels.org/stable/generated/statsmodels.tsa.statespace.structural.UnobservedComponents.html#statsmodels.tsa.statespace.structural.UnobservedComponents) and [`fit`](https://www.statsmodels.org/stable/generated/statsmodels.tsa.statespace.structural.UnobservedComponents.fit.html#statsmodels.tsa.statespace.structural.UnobservedComponents.fit) of `statsmodels` documentation. Then does it solve the problem in the sense of not passing arbitrary keyword arguments?

Also, may be related, I don't quite follow why `_StatsModelsAdapter` take a `random_state` argument. Is it supposed to be (or was it) present for all forecasting algorithms from `statsmodels` previously?

fkiraly:
> Is it "safe" to assume that two check correspond to one another, i.e. if `self.random_state` exists, the algorithm allows `random_state` as input and should not fail if passed to `get_prediction` (and vice versa)?

I don't think that's safe to assume right now, but I think the higher-order idea of the developer encoding that information is the right one. The specific idea could also work: that is, I assume you were thinking of, (a) implenting the logic along those lines, and (b) then removing `random_state` from the `__init__` of all estimators which raise a warning or error.

While this works, here is a variant of the same idea: make this a tag `statsmodels:args` or `statsmodels:arg:random_state` or similar. This determines the logic of the adapter, by default the arg is passed.

> I don't quite follow why `_StatsModelsAdapter` take a `random_state` argument. Is it supposed to be (or was it) present for all forecasting algorithms from `statsmodels` previously?

I don't know, that part is such an old piece of code. I don't even remember if I had design input or code input into it.

fkiraly:
@yarnabrina, thanks for reminding me again of this - how shall we proceed with the `random_state` issue? Should we close this, or can it be salvaged?

Issue #4598, opened 2023-05-17
[BUG] difference in predictions between `AutoARIMA` of `statsforecast` and `StatsForecastAutoARIMA`


yarnabrina:
(Reported by `Nasreddine` on discord)

If one dataset is trained once directly with `statsforecast` model and another time through `sktime` interface, user expectes to get identical predictions given both are trained with same configurations. But it is not the case.

Also it does not seem to be because of random variation. If one replicates any of the two models multiple times, the results remain identical.

```python
import numpy
from statsforecast.models import AutoARIMA
from statsforecast.utils import AirPassengers

from sktime.forecasting.statsforecast import StatsForecastAutoARIMA

# 476.00650924 482.84620599
statsforecast_original_predictions = AutoARIMA().fit(AirPassengers).predict(h=2)

# 464.3571524 487.38348364
sktime_adapter_predictions = StatsForecastAutoARIMA().fit(AirPassengers).predict(fh=[1, 2])  

# fails
numpy.testing.assert_allclose(
    statsforecast_original_predictions["mean"], sktime_adapter_predictions.ravel()
)
```

Results are found on Ubuntu WSL on Windows 11, Python `3.9.16`, commit e63af2e2 on `main`.

yarnabrina:
@FedericoGarza Can you please help regarding this ☝🏽?

Few argument names in [`statsforecast` implementation](https://nixtla.github.io/statsforecast/models.html#autoarima) and [`sktime` adaptation](https://www.sktime.net/en/latest/api_reference/auto_generated/sktime.forecasting.statsforecast.StatsForecastAutoARIMA.html#statsforecastautoarima) vary, along with lots of difference in argument order. This makes it difficult to compare/track over time, but I think that the difference is coming out of different default values of few arguments.

As far as I can see, `trend`, `with_intercept`, `approximation` defaults vary. Can you please confirm what should be the desired defaults?

yarnabrina:
Also, it seems two models in `statsforecast` gives differen results. Can you please explain the difference between `statsforecast.models.AutoARIMA` and `statsforecast.arima.AutoARIMA`? They have different output formats, and different defaults leading to different results.

```pycon
>>> from statsforecast.utils import AirPassengers as ap
>>> from statsforecast.models import AutoARIMA as _models_ARIMA
>>> from statsforecast.arima import AutoARIMA as _arima_ARIMA
>>> _models_ARIMA().fit(ap).predict(2)
[...]

PR #4629, opened 2023-05-20
[ENH] rewrite of `_StatsForecastAdapter` in a generic way to support other models than `AutoARIMA`

yarnabrina:
#### Reference Issues/PRs

See #4447 for discussions.

The existing `_StatsForecastAdapter` was too specific for `AutoARIMA`, and it was not possible to use same base adapter class for other estimators, e.g. `AutoTheta`. This PR attempts to rewrite in a general way to avoid this problem.

No.

* `copyright` part - not sure what to use
* `__author__` part - replaced `FedericoGarza` as I rewrote entirely, but since it's Nixtla's implementation, not sure if that's okay
* specification of tags `X-y-must-have-same-index` and `enforce_index_type` - not sure what is appropritae, skipped to use default
* introduced `fh_type`, which should only be either `in-sample` or `out-of-sample` - should this be an Enum, or normal check is fine?
* modified import for `StatsForecastAutoARIMA`, as `statsforecast.models` is publicly documented (and changed argument name as appropriate) - please look at #4598
* other code readability issues, if any
* other code generalisation issues, if any

No.

#### Any other comments?

I tested the functionally locally by running this:

```sh
pytest -k "StatsForecastAutoARIMA" sktime/forecasting/tests/test_all_forecasters.py
```

All tests passed successfully. To test generalisation of the base adapter, I took the liberty of merging this PR branch into PR #4539 branch locally. Afterwards, all of these passed locally:

```sh
pytest -k "StatsForecastAutoTheta" sktime/forecasting/tests/test_all_forecasters.py
```

Hence, updated that branch remotely as well, and that PR can be checked to test generic nature of this adapter.

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/forecasting/base/adapters/_statsforecast.py
+++ b/sktime/forecasting/base/adapters/_statsforecast.py
@@ -1,37 +1,33 @@
 # -*- coding: utf-8 -*-
 # copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
-"""Implements adapter for StatsForecast forecasters to be used in sktime framework."""
-
-__author__ = ["FedericoGarza"]
-__all__ = ["_StatsForecastAdapter"]
-
-
-import pandas as pd
+"""Implements adapter for StatsForecast models."""
+import pandas
 
 from sktime.forecasting.base import BaseForecaster
-from sktime.forecasting.base._base import DEFAULT_ALPHA
+
+__all__ = ["_StatsForecastAdapter"]
+__author__ = ["yarnabrina"]
 
 
 class _StatsForecastAdapter(BaseForecaster):
-    """Base class for interfacing StatsForecast."""
+    """Base adapter class for StatsForecast models."""
 
     _tags = {
-        "scitype:y": "univariate",  # which y are fine? univariate/multivariate/both
-        "ignores-exogeneous-X": False,  # does estimator ignore the exogeneous X?
-        "handles-missing-data": False,  # can estimator handle missing data?
-        "y_inner_mtype": "pd.Series",  # which types do _fit, _predict, assume for y?
-        "X_inner_mtype": "pd.DataFrame",  # which types do _fit, _predict, assume for X?
-        "requires-fh-in-fit": False,  # is forecasting horizon already required in fit?
-        "X-y-must-have-same-index": False,  # can estimator handle different X/y index?
-        "enforce_index_type": None,  # index type that needs to be enforced in X/y
-        "capability:pred_int": True,  # does forecaster implement predict_quantiles?
-        "capability:pred_int:insample": True,
-        "python_dependencies": "statsforecast",
+        "y_inner_mtype": "pd.Series",
+        "X_inner_mtype": "pd.DataFrame",
+        "scitype:y": "univariate",
+        "requires-fh-in-fit": False,
+        # "X-y-must-have-same-index": True,  # TODO: need to check (how?)
+        # "enforce_index_type": None,  # TODO: need to check (how?)
+        "handles-missing-data": False,
+        "python_version": ">=3.7",  # TODO: change to 3.8 (when?)
+        "python_dependencies": ["statsforecast"],
     }
 
     def __init__(self):
+        super().__init__()
+
         self._forecaster = None
-        super(_StatsForecastAdapter, self).__init__()
 
     def _instantiate_model(self):
         raise NotImplementedError("abstract method")
@@ -46,7 +42,8 @@ class _StatsForecastAdapter(BaseForecaster):
 
         Parameters
         ----------
-        y : guaranteed to be of a type in self.get_tag("y_inner_mtype")
+        y : sktime time series object
+            guaranteed to be of an mtype in self.get_tag("y_inner_mtype")
             Time series to which to fit the forecaster.
             if self.get_tag("scitype:y")=="univariate":
                 guaranteed to have a single column/variable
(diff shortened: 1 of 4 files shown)


yarnabrina:
@fkiraly This is the first draft for rewrite of `statsforecast` adapter. Please take a look and suggest revisions as required.

The points I am not confident of or need someone else to also take a look are mentioned in the PR description.

fkiraly:
> `__author__` part - replaced `FedericoGarza` as I rewrote entirely, but since it's Nixtla's implementation, not sure if that's okay

In general:
For estimators that have an owner, we can't change them without consent of the owners, in this case @FedericoGarza, unless they become unresponsive, see rules here: https://www.sktime.net/en/stable/get_involved/governance.html#algorithm-maintainers

FYI @marrov (who also looks after rules on the CoC committee).

Now, about this specifically:

* I've met @FedericoGarza at pydata Seattle and let them know that we were going to refactor this, and he was fine with it, and also offered to give advice - so we'll just ping him here.
* this PR touches the *estimator* only on the maintenance level, it's the *adapter* that it changes which is an intermediate framework part. Framework does not fall under ownership rules, but this is package specific, so a grey area. I'd still err on the side of asking @FedericoGarza for approval.
* kindly leave him as an author, at least, unless he becomes unresponsive. You can add yourself if he agrees.
* if @FedericoGarza does not react within 10 days, a "clean" approach would be to write a second adapter and do a rename switch.

sktime/forecasting/base/adapters/_statsforecast.py
```diff
@@ -226,47 +220,35 @@ def _predict_interval(self, fh, X=None, coverage=0.90):
...
-        # both in-sample and out-of-sample values (we reach this line only then)
-        # in this case, we additionally need to concat in and out-of-sample returns
-        _, y_ins_pred_int = self._predict_in_sample(fh_ins, **kwargs)
-        _, y_oos_pred_int = self._predict_fixed_cutoff(fh_oos, **kwargs)
-        for ins_int, oos_int, a in zip(y_ins_pred_int, y_oos_pred_int, coverage):
-            pred_int[("Coverage", a, "lower")] = pd.concat([ins_int, oos_int])["lower"]
-            pred_int[("Coverage", a, "upper")] = pd.concat([ins_int, oos_int])["upper"]
-
-        return pred_int
+        X_predict_input = X if X is None else X.to_numpy(copy=False)
```
fkiraly, line 231: it would be slightly less obscure if you say: `if X is not None: X_predict_input = X.to_numpy, else: X_predict_input = None`

sktime/forecasting/base/adapters/_statsforecast.py
```diff
@@ -57,19 +54,92 @@ def _fit(self, y, X=None, fh=None):
...
+        X_fit_input = X.to_numpy(copy=False) if X is not None else X
+
+        self._forecaster.fit(y_fit_input, X=X_fit_input)
 
         return self
 
+    def _predict_in_or_out_of_sample(self, fh, fh_type, X=None, levels=None):
+        maximum_forecast_horizon = fh.to_relative(self.cutoff)[-1]
+
+        absolute_horizons = fh.to_absolute(self.cutoff).to_pandas()
```
fkiraly, line 79: preferred here: `to_absolute_index`, does the same

sktime/forecasting/base/adapters/_statsforecast.py
```diff
@@ -57,19 +54,92 @@ def _fit(self, y, X=None, fh=None):
...
+            interval_predictions[
+                ("Coverage", level, "lower")
+            ] = lower_interval_predictions[horizon_positions]
+            interval_predictions[
+                ("Coverage", level, "upper")
+            ] = upper_interval_predictions[horizon_positions]
+
+        return interval_predictions
+
+    def _split_horizon(self, fh):
```
fkiraly, line 137: it occurs to me that almost always when we call one of the methods in there, we do this. Would it make sense to add this to `ForecastingHorizon`? Out of scope for this PR, a non-blocking comment.

sktime/forecasting/base/adapters/_statsforecast.py
```diff
@@ -57,19 +54,92 @@ def _fit(self, y, X=None, fh=None):
...
-        self._forecaster.fit(y.values, X.values if X is not None else X)
+
+        y_fit_input = y.to_numpy(copy=False)
+        X_fit_input = X.to_numpy(copy=False) if X is not None else X
+
+        self._forecaster.fit(y_fit_input, X=X_fit_input)
 
         return self
 
+    def _predict_in_or_out_of_sample(self, fh, fh_type, X=None, levels=None):
```
fkiraly, line 76: I wonder if point and interval predictions could be separated better here, though I can't see a good way given how intertwined it is in `statsforecast`.

fkiraly (changes requested):
Content-wise, this looks good to me!

As said, we need to check with @FedericoGarza and not remove him as an owner.

fkiraly:
> `copyright` part - not sure what to use

`sktime` developers, as usual. The interface is part of `sktime`, the interfaced algorithm (subject to nixtla copyright) is external.

> specification of tags `X-y-must-have-same-index` and `enforce_index_type` - not sure what is appropritae, skipped to use default

Yes, I think that's the right setting.

> introduced `fh_type`, which should only be either `in-sample` or `out-of-sample` - should this be an Enum, or normal check is fine?

Looks fine like this to me. Enum would complicate usage slightly imo.

> modified import for `StatsForecastAutoARIMA`, as `statsforecast.models` is publicly documented (and changed argument name as appropriate) - please look at #4598

Yes, that's odd. Why are there two of it? Maybe @FedericoGarza can shed light on this.

sktime/forecasting/base/adapters/_statsforecast.py
yarnabrina: Yes.

In fact, getting end of horizon, splitting horizon in in-sample and out-of-sample, call separate method for both cases - all of this seems to be repeated for `pmdarima` and `statsforecast` (and also `tbats`?), and for both `predict` and `predict_interval`. A way to avoid these duplications will be useful.

sktime/forecasting/base/adapters/_statsforecast.py
yarnabrina: Do you mean ask for only point or only interval predictions? I don't think that'd be possible, given asking for interval will always give point in `statsforecast`.

OR

What was there in old adapter was for each level, a new call was made to `predict`. Are you talking about this? I felt that was unnecessary comuputation because it supports prediction for multiple levels simultaneously.

yarnabrina:
I went for the third suggestion of clean approach for now: created a new adapter keeping the old adapter and author unmodified. So, the only modification of old codebase in that I changed the base class for `StatsForecastAutoARIMA` to use the new adapter.

fkiraly:
Ok, I think that's a fair approach!
I's maintenance & renaming, and we're not changing any downstream functionality in the concrete class.

@FedericoGarza can comment in his time.

fkiraly (approved):
All blocking comments have been addressed - we could merge if this weren't a draft.

yarnabrina:
@fkiraly Changed PR status from draft to ready. Will create an issue for the `fh` enhancements.

fkiraly:
great, let's merge this then.

Based on earlier discussion, I assume this will be used for https://github.com/sktime/sktime/pull/4539 later, which possibly also includes minor adaptations to the adapter.

FedericoGarza:
hey @yarnabrina, @fkiraly! I want to remain an author if possible. The PR is already merged but LGTM. 

@yarnabrina let me know if you need anything from my side.

fkiraly:
Re authorship, I'm going to spell out the rules first and then my recommendation.
http://www.sktime.net/en/latest/get_involved/governance.html#algorithm-maintainers

here's two cases: on the existing estimator, we *need* to keep @FedericoGarza based on the rules we have (since he's sent the note). On new estimators, the author (here: @yarnabrina) *can* add another person but doesn't need to. Of course another person could make a forked copy that they maintain.

Now the *nice* thing to do is obviously add @FedericoGarza as an author on all the new `statsmodels` facing estimators, and (according to the same rules as linked above) this comes with the commitment of his to help maintaining the interfaces :-)

fkiraly:
Regarding folder structure:

in the past we've followed a loose rule of, estimators can live in one module if they are "semantically related", and clearly this is the case for all `statsforecast` facing estimators - although one could argue, maybe ARIMA should be put to ARIMA, reduction to reduction etc (that is a different taxonomy).

When too many estimators end up in a file, it becomes clunky and difficult to handle, in such a case it may make sense to split them up into a module. To avoid deprecation, there should be an init that still exports them in the same location, but instead of a single file you then have a folder and multiple files with less estimators per file.
An example PR of this kind is here: https://github.com/sktime/sktime/pull/4368

PR #4632, opened 2023-05-21
[BUG] allow unfitted estimators as fitted parameters in `get_fitted_params`

### LINKED ISSUE #4574: [BUG] `get_fitted_params` fails on a tuned reconciler forecaster
Reported by @daniel-torresc in https://github.com/sktime/sktime/discussions/4101.

A `get_fitted_params` call on a highly nested combination of reconciliation and tuning fails where it should work as per interface specification.

reproducible example, fails on 3.10, windows, 0.18.0 for me too:

```python
from sktime.forecasting.exp_smoothing import ExponentialSmoothing
from sktime.forecasting.model_selection import ForecastingGridSearchCV, ExpandingWindowSplitter
from sktime.forecasting.reconcile import ReconcilerForecaster
from sktime.forecasting.compose import ForecastByLevel

y = _make_hierarchical()
agg = Aggregator()
y_agg = agg.fit_transform(y) 

param_grid = [{"forecaster": [ExponentialSmoothing()],
               "forecaster__trend": ['add','mul']
              },
              {"forecaster": [PolynomialTrendForecaster()],
               "forecaster__degree": [1,2]}
              ]

    ("forecaster", ExponentialSmoothing())])

N_cv_fold = 2
step_cv = 1
fh = [1,2]

initial_window_cv_len = len(y_agg.index.get_level_values(2).unique()) - (N_cv_fold - 1) * step_cv - fh[-1]

cv = ExpandingWindowSplitter(
            initial_window = initial_window_cv_len,
            step_length = step_cv,
            fh = fh)

        forecaster=pipe,
        param_grid=param_grid,
        cv=cv,
        n_jobs=-1,
        verbose = 1
        )

gscv_bylevel = ForecastByLevel(gscv, 'local')

reconciler = ReconcilerForecaster(gscv_bylevel, method="ols")

reconciler.fit(y_agg)

reconciler.get_fitted_params()  # here we get the error when accessing ".get_fitted_params()"
```

fkiraly:
The `get_fitted_params` method would terminate with an exception if a fitted parameter was an estimator, but not fitted - as it would attempt to call its `get_fitted_params` which would then error out.

However, it is perfectly sensible - even if rare - for a fitted parameter to be an estimator, so this should be possible.

This PR solves this by not raising an exception when `get_fitted_params` is called on an unfitted estimator, but returning an empty `dict`.

Deprecation is not necessary, as it removes an exception case upon nothing hinges in downstream code.

Reviewers should focus on whether this interface change is sensible.

Fixes #4574, and adds it as a test case.

### DIFF
--- a/sktime/base/_base.py
+++ b/sktime/base/_base.py
@@ -386,7 +386,7 @@ class BaseEstimator(BaseObject):
         """Get fitted parameters.
 
         State required:
-            Requires state to be "fitted".
+            Requires state to be "fitted". If not, returns empty dict.
 
         Parameters
         ----------
@@ -412,12 +412,11 @@ class BaseEstimator(BaseObject):
               all parameters of `componentname` appear as `paramname` with its value
             * if `deep=True`, also contains arbitrary levels of component recursion,
               e.g., `[componentname]__[componentcomponentname]__[paramname]`, etc
+
+            If estimator is not fitted, will return an empty dict.
         """
         if not self.is_fitted:
-            raise NotFittedError(
-                f"estimator of type {type(self).__name__} has not been "
-                "fitted yet, please call fit on data before get_fitted_params"
-            )
+            return {}
 
         # collect non-nested fitted params of self
         fitted_params = self._get_fitted_params()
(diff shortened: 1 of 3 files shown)


fkiraly:
I was wrong in my diagnosis, I think the root cause is https://github.com/sktime/sktime/pull/4632.

PR #4634, opened 2023-05-21
[DOC] adding doctest to the documentation

### LINKED ISSUE #4582: [DOC] Include `doctest` section in developer guide
The developer guide should include steps for running `doctest` locally. This will help new developers understand how to run a `doctest` to catch any potential doc errors before making commits to a pull request. Currently, the guide only covers `pytest`.

A suggestion is to add it as a subsection after `Full test suite runs via pytest`.
Here's from `doctest` on running the doctests locally: https://docs.python.org/3/library/doctest.html

mdsaad2305:
#### Reference Issues/PRs
Fixes #4582 

Included doctest documentation within the testing and continuous integration section.


mdsaad2305:
Should I make any additional changes to the file to pass the code-quality section?

fkiraly (changes requested):
Great, thanks!

Some small requests:

* linting fails in this PR, please fix that
* can you say explicitly how to run all doctests in `sktime`?

mdsaad2305:
Hey, just needed some help before I implement these changes. 

1) what kind of extension/package are we supposed to use for linting errors?
2) When you mentioned running "all doctests," did you mean running doctest on all the files? I just want to make sure I understand that part correctly

fkiraly:
> what kind of extension/package should we use to detect linting errors?

If you follow the link to "coding standards" on the page you are editing, I hope you'll find a guide that's explicit about this.
Let me know if it's not easy to find, then we may have to think about structure etc.

> When you mentioned running "all doctests," did you mean running doctest on all the files? I just want to make sure I understand that part correctly

Yes, that's what I meant, thanks for asking.

mdsaad2305:
Could you let me know if the code for running all the files is correct? It seems to be working fine for me, but I hope I don't end up confusing the newcomers.

fkiraly:
Hm, I would have done `pytest --doctest-modules`, does that not work?

mdsaad2305:
Might be an issue with my system but does using pytest normally take a long time?

mdsaad2305:
is this alright?

yarnabrina:
> Hm, I would have done `pytest --doctest-modules`, does that not work?

`doctest` is a standard library, it's always present. And it runs only the doctests. `pytest` is not standard, and will be available only if someone installs `dev` dependencies. And it is going to run all unit tests along with doctests by default, so that immediately increases the time duration.

So, my suggestion is to include ways to run with both, and specify that one can try with the `pytest` version only if `dev` dependencies are installed. And may be they can use `pytest` to run just doctests in case they pass `-k "not test_"` or similar (not tested).

fkiraly:
> So, my suggestion is to include ways to run with both, and specify that one can try with the pytest version only if dev dependencies are installed. And may be they can use pytest to run just doctests in case they pass -k "not test_" or similar (not tested).

Ah, I see - should we then mention both ` find . -name "*.py" -print0 | xargs -0 python -m doctest -v` (on unix, no `pytest` installed) and `pytest --doctest-modules` (if you have `pytest` installed)?

yarnabrina:
Exactly.

@mdsaad2305: can you also try to find a similar command for Windows users? It'll be good to document as well along with @fkiraly's above summary, if possible.

mdsaad2305:
Sure, I’ve some work till Friday, so I’ll try to finish it during the weekend

fkiraly (approved):
Nice! Thanks, great compilation!

Issue #4054, opened 2023-01-03
[ENH] rolling window (out of sample) detrender


fkiraly:
It would be good to have a rolling window detrender that allows out-of-sample forecast detrending, i.e., computing rolling window out-of-sample residuals rather than in-sample residuals.

The current `Detrender` computes in-sample residuals, when naively applied (via `fit_transform`, which is called in a forecasting pipeline).

Not sure whether this should be an option in the class or a separate class - it should probably take a splitter (inheriting from `BaseSplitter`) as an argument. There is also some thought to be had around what should happen if the series seen in `transform` is not entirely identical nor entirely different from the one in `fit`.

If someone wants to take this up soon, it is perhaps best to base this on the simplified logic in https://github.com/sktime/sktime/pull/4053 rather than the current `main`.

blazingbhavneek:
Hey! I would like to take it, seems a little intimidating to read, but can you suggest some files to read to get idea of what's happening? What i am trying to change?

fkiraly:
Well, it's more of an extension than a change - there should be an option for "out of sample", and those are computed by the `update_predict` method (with a splitter inside).

I would recommend to read up on `update_predict` in the forecasting tutorial (part 1) and see how it compares to the current logic of `Detrender` which just uses `predict`.

blazingbhavneek:
Hey! @fkiraly (sorry for the delay, I had exams in college)
I am new to these terminologies, so I had a hard time making sense, but as far as I could understand, I have to implement something like this, right?

![Rolling window out of sample detrender](https://user-images.githubusercontent.com/86467990/218276843-18f998c3-5e5e-4ec0-a33f-8654ffbb70e4.png)

Or am I missing something?

fkiraly:
Yes, exactly - and we use the union of the out-of-sample forecasts to detrend, rather than in-sample forecasts.

fkiraly:
> 1. What is the minimum size of the dataset required to train a forecaster initially? (To be on a safe side) Before we start to `update_predict` it.

There is no minimum that the interface enforces (i.e., 1 data point); most forecasters will break though if you start going below 10. The minimum may also be influenced by seasonality.

There is also an open design discussion here: https://github.com/sktime/sktime/issues/3901
with @bethrice44 and @chillerobscuro, FYI

> 2. What should be the `fh`? Should it depend on whether the forecaster requires `fh` or not?

[...]

Issue #4610, opened 2023-05-17
[BUG] windows python 3.9 checks memouts and hangs


fkiraly:
The CI on python 3.9, windows:

* takes a long time to terminate, if it does
* frequently crashes with memouts or other problems

It is unclear why exactly that combination has problems, but it could be due to a problematic estimator that appears only on 3.9 in windows and uses up a lot of memory and compute, and due to the lower specs of the windows VM.

(recall: estimators are distributed across versions for a given operating system, with a periodicity of 3, so an estimator in 3.7 would reappear on 3.10, but 3.9 does not appear elsewhere since the range is 3.7 to 3.11)

fkiraly:
seems like it's coming from `FCNClassifier`?

Jay-sanjay:
Hey ,
I want to contribute to the issue can you it a bit more precisely and provide
 hints to debug it

fkiraly:
unfortunately, that's as precise as I know - CI hangs on occasion, and sometimes there are vague error messages implicating `FCNClassifier`.

You can run tests locally by doing:
```python
from sktime.utils.estimator_checks import check_estimator
from sktime.classification.deep_learning.fcn import FCNClassifier

check_estimator(FCNClassifier, raise_exceptions=True)
```

Jay-sanjay:
As a beginner towards open source contribution  I wanted some hints that how the issue can be fixed

Jay-sanjay:
also what actually does FCNClassifier means

fkiraly:
> As a beginner towards open source contribution I wanted some hints that how the issue can be fixed

Sure, of course.
Please note that this is not labelled a "good first issue", so it's not a recommended beginner issue, since it might get difficult to diagnose.

Good first issues are here: https://github.com/sktime/sktime/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22
(you can also get here by clicking on the "good first issue" button at the top of the issue tracker)

The suggestion, i.e., the hint to get started was the code pasted above - that is how you run the full test suite for the classifier, called `FCNClassifier` (it's the name of an estimator, a classifier)

fkiraly:
@yarnabrina, I notice that windows 3.9 is now *almost always* failing - any good ideas on this?

yarnabrina:
[...]

PR #4642, opened 2023-05-24
Rolling window (out of sample) detrender

blazingbhavneek:
#### Reference Issues/PRs
[ENH] rolling window (out of sample) detrender #4054

Added rolling window detrender that allows out-of-sample forecast detrending, i.e., computing rolling window out-of-sample residuals rather than in-sample residuals.

Nil

### DIFF
--- a/sktime/transformations/series/detrend/_detrend.py
+++ b/sktime/transformations/series/detrend/_detrend.py
@@ -6,10 +6,13 @@
 __all__ = ["Detrender"]
 __author__ = ["mloning", "SveaMeyer13", "KishManani", "fkiraly"]
 
+from warnings import warn
+
 import pandas as pd
 
 from sktime.datatypes import update_data
 from sktime.forecasting.base._fh import ForecastingHorizon
+from sktime.forecasting.model_selection import ExpandingWindowSplitter
 from sktime.forecasting.trend import PolynomialTrendForecaster
 from sktime.transformations.base import BaseTransformer
 
@@ -84,9 +87,10 @@ class Detrender(BaseTransformer):
         "transform-returns-same-time-index": True,
     }
 
-    def __init__(self, forecaster=None, model="additive"):
+    def __init__(self, forecaster=None, model="additive", out_of_sample=False):
         self.forecaster = forecaster
         self.model = model
+        self.oos = out_of_sample
 
         super(Detrender, self).__init__()
 
@@ -116,8 +120,20 @@ class Detrender(BaseTransformer):
         -------
         self: a fitted instance of the estimator
         """
+        if len(X) <= 10:
+            self.oos = False
+            warn(
+                """
+                    Length of X is less than 11, forecasters tend to break
+                    below length 10, using in-sample Detrender now
+                    """
+            )
+
         if not self.forecaster_.get_tag("requires-fh-in-fit", True):
-            self.forecaster_.fit(y=X, X=y)
+            if not self.oos:
+                self.forecaster_.fit(y=X, X=y)
+            else:
+                self.forecaster_.fit(y=X[:10], X=y)
         else:
             self._X = X
             self._y = y
@@ -136,7 +152,11 @@ class Detrender(BaseTransformer):
         if self.forecaster_.get_tag("requires-fh-in-fit", True):
             X = update_data(self._X, X)
             y = update_data(self._y, y)
-            forecaster = self.forecaster_.clone().fit(y=X, X=y, fh=fh)
+            if not self.oos:
+                forecaster = self.forecaster_.clone().fit(y=X, X=y, fh=fh)
+            else:
+                forecaster = self.forecaster_.clone().fit(y=X[:10], X=y, fh=fh)
+
         else:
             forecaster = self.forecaster_
         return forecaster
@@ -161,7 +181,14 @@ class Detrender(BaseTransformer):
         fh = self._get_fh_from_X(X=X)
         forecaster = self._get_fitted_forecaster(X=X, y=y, fh=fh)
 
-        X_pred = forecaster.predict(fh=fh, X=y)
+        if not self.oos:
+            X_pred = forecaster.predict(fh=fh, X=y)
+        else:
+            cv = ExpandingWindowSplitter()
+            X_pred1 = forecaster.predict(fh=fh, X=y[:10])
+            X_pred2 = forecaster.update_predict(X=y, cv=cv)
+            X_pred2 = X_pred.drop_duplicates()
+            X_pred = pd.concat([X_pred1, X_pred2])
 
         if self.model == "additive":
             return X - X_pred


blazingbhavneek:
@fkiraly I am facing a weird issue with skbase, i don't know what to do?
.
.
Error message:
.
AttributeError                            Traceback (most recent call last)
[d:\Open](file:///D:/Open) Source Repos\sktime\examples\01_forecasting.ipynb Cell 279 in ()
      [1](vscode-notebook-cell:/d%3A/Open%20Source%20Repos/sktime/examples/01_forecasting.ipynb#Y544sZmlsZQ%3D%3D?line=0) transformer = Detrender()
----> [2](vscode-notebook-cell:/d%3A/Open%20Source%20Repos/sktime/examples/01_forecasting.ipynb#Y544sZmlsZQ%3D%3D?line=1) y_hat = transformer.fit_transform(X=y[-11:-2])
      [3](vscode-notebook-cell:/d%3A/Open%20Source%20Repos/sktime/examples/01_forecasting.ipynb#Y544sZmlsZQ%3D%3D?line=2) print(y_hat.size)

File [D:\Open](file:///D:/Open) Source Repos\sktime\sktime\transformations\base.py:620, in BaseTransformer.fit_transform(self, X, y)
    555 """Fit to data, then transform it.
    556 
    557 Fits the transformer to X and y and returns a transformed version of X.
   (...)
    616         Example: i-th instance of the output is the i-th window running over `X`
    617 """
    618 # Non-optimized default implementation; override when a better
    619 # method is possible for a given algorithm.
--> 620 return self.fit(X, y).transform(X, y)

File [D:\Open](file:///D:/Open) Source Repos\sktime\sktime\transformations\base.py:410, in BaseTransformer.fit(self, X, y)
    373 """Fit transformer to X, optionally to y.
    374 
    375 State change:
   (...)
    407 self : a fitted instance of the estimator
    408 """
    409 # if fit is called, estimator is reset, including fitted state
--> 410 self.reset()
    412 # skip everything if fit_is_empty is True and we do not need to remember data
    413 if self.get_tag("fit_is_empty") and not self.get_tag("remember_data", False):

File [c:\Users\Bhavneek\anaconda3\envs\sktime-dev\lib\site-packages\skbase\base\_base.py:129](file:///C:/Users/Bhavneek/anaconda3/envs/sktime-dev/lib/site-packages/skbase/base/_base.py:129), in BaseObject.reset(self)
    107 """Reset the object to a clean post-init state.
    108 
    109 Using reset, runs __init__ with current values of hyper-parameters
   (...)
    126 call, self is equal in value to `type(self)(**self.get_params(deep=False))`
    127 """
    128 # retrieve parameters to copy them later
--> 129 params = self.get_params(deep=False)
    131 # delete all object attributes in self
    132 attrs = [attr for attr in dir(self) if "__" not in attr]

File [c:\Users\Bhavneek\anaconda3\envs\sktime-dev\lib\site-packages\skbase\base\_base.py:288](file:///C:/Users/Bhavneek/anaconda3/envs/sktime-dev/lib/site-packages/skbase/base/_base.py:288), in BaseObject.get_params(self, deep)
    260 def get_params(self, deep=True):
    261     """Get a dict of parameters values for this object.
    262 
    263     Parameters
   (...)
    286           e.g., `[componentname]__[componentcomponentname]__[paramname]`, etc
    287     """
--> 288     params = {key: getattr(self, key) for key in self.get_param_names()}
    290     if deep:
    291         deep_params = {}

File [c:\Users\Bhavneek\anaconda3\envs\sktime-dev\lib\site-packages\skbase\base\_base.py:288](file:///C:/Users/Bhavneek/anaconda3/envs/sktime-dev/lib/site-packages/skbase/base/_base.py:288), in (.0)
    260 def get_params(self, deep=True):
    261     """Get a dict of parameters values for this object.
    262 
    263     Parameters
   (...)
    286           e.g., `[componentname]__[componentcomponentname]__[paramname]`, etc
    287     """
--> 288     params = {key: getattr(self, key) for key in self.get_param_names()}
    290     if deep:
    291         deep_params = {}

AttributeError: 'Detrender' object has no attribute 'out_of_sample'

sktime/transformations/series/detrend/_detrend.py
```diff
@@ -84,9 +87,10 @@ class Detrender(BaseTransformer):
         "transform-returns-same-time-index": True,
     }
 
-    def __init__(self, forecaster=None, model="additive"):
+    def __init__(self, forecaster=None, model="additive", out_of_sample=False):
         self.forecaster = forecaster
         self.model = model
+        self.oos = out_of_sample
```
fkiraly, line 93: the issue is here: you need to write the parameter `out_of_sample` to `self.out_of_sample` - this is a convention from `sklearn`, that all init params need to be written to `self`, with the same name.

sktime/transformations/series/detrend/_detrend.py
```diff
@@ -136,7 +152,11 @@ def _get_fitted_forecaster(self, X, y, fh):
         if self.forecaster_.get_tag("requires-fh-in-fit", True):
             X = update_data(self._X, X)
             y = update_data(self._y, y)
-            forecaster = self.forecaster_.clone().fit(y=X, X=y, fh=fh)
+            if not self.oos:
+                forecaster = self.forecaster_.clone().fit(y=X, X=y, fh=fh)
+            else:
+                forecaster = self.forecaster_.clone().fit(y=X[:10], X=y, fh=fh)
```
fkiraly, line 158: hm, should this not be fitted to all the data, rather than just `X[:10]`?

sktime/transformations/series/detrend/_detrend.py
blazingbhavneek: Got it!

sktime/transformations/series/detrend/_detrend.py
blazingbhavneek: Because I think the whole point of this feature was to feed the data into forecaster in parts and get prediction in parts like the diagram I made in the [discussion](https://github.com/sktime/sktime/issues/4054#issuecomment-1426856960) we had earlier
image

fkiraly: I see - but wouldn't we want to do it then only in `transform`, rather than in `fit`?

blazingbhavneek:
@fkiraly what is this error? I don't understand it

fkiraly:
Sorry about not answering.

The error is totally unrelated to your contribution, some estimator is crashing specifically on windows 3.9 and we are currently hunding for it. It must be one of the deep learning estimators recently merged, but that's a conjecture.

See https://github.com/sktime/sktime/issues/4610

sktime/transformations/series/detrend/_detrend.py
```diff
@@ -116,8 +123,20 @@ def _fit(self, X, y=None):
         -------
         self: a fitted instance of the estimator
         """
+        if len(X) <= 10:
+            self.out_of_sample = False
```
fkiraly, line 127: estimator parameters should never be changed after `__init__`, so you need to replace this with a local variable condition.

Alternatively, we just drop this block and warn the user only.

fkiraly (changes requested):
Thanks!

* the failure is unrelated to the addition
* have you seen my question above? Why do we fit only to the first 10 data points? I would do the `update_predict` in `transform` only. This may require re-fitting in case the `transform` data is not all after the `fit` data.

sktime/transformations/series/detrend/_detrend.py
blazingbhavneek: I am not sure what you mean by this, kindly elaborate, please.

sktime/transformations/series/detrend/_detrend.py
blazingbhavneek: Done! I will see to it.

sktime/transformations/series/detrend/_detrend.py
blazingbhavneek: I see! I was looking at the use of this functionality in a different way, I got it now! I will make the necessary changes now.

blazingbhavneek:
@fkiraly Anything else I can/should change?

fkiraly (changes requested):
Thanks, this is great!

Now, unfortunately, you are not testing the estimator - you need to use `check_estimator`, either after updating `get_params`, or with the new parameters set.

That way, you could detect that the inverse is not actually inverse to the transform.

I've added a `get_params` case and made some changes trying to address the failures. I haven't finished entirely and need to work on other things now, but hopefully I am making sense in where this is going? Feel free to revert if not.

Issue #4602, opened 2023-05-17
[BUG] make_reduce with unequal index set panels raises ValueError: Input X contains NaN 


EduardoPach:
**Describe the bug**

When using a panel set with unequal DatetimeIndex raises ValueError: Input X contains NaN. When changing so that the indexes are all the same no errors are raised.

**To Reproduce**

```python
import numpy as np
import pandas as pd
from sklearn.pipeline import make_pipeline
from sklearn.ensemble import RandomForestRegressor
from sktime.transformations.series.date import DateTimeFeatures
from sktime.transformations.series.summarize import WindowSummarizer
from sktime.forecasting.model_selection import temporal_train_test_split
from sktime.forecasting.compose import ForecastingPipeline, make_reduction

def create_df(cat: str, dates: pd.DatetimeIndex) -> pd.DataFrame:
    df = pd.DataFrame({
        'category': [cat]*len(dates),
        'date': dates,
        'value': np.random.randn(len(dates))
    })

    return df


dates_a = pd.date_range(start='2018-01-01', end="2019-01-01", freq='MS')
dates_b = pd.date_range(start='2018-04-01', end="2019-01-01", freq='MS')
dates_c = pd.date_range(start='2018-05-01', end="2019-01-01", freq='MS')

cats = ["a" ,"b", "c"]
dates = [dates_a, dates_b, dates_c]


df = pd.concat([create_df(cat, date) for cat, date in zip(cats, dates)], ignore_index=True)
df.set_index(["category", "date"], inplace=True)

train, test = temporal_train_test_split(df, test_size=0.2)

regressor = make_pipeline(RandomForestRegressor(random_state=1))

kwargs = {
    "lag_feature": {
        "lag": [1, 2, 3],
        "mean": [[1, 3]],
        "std": [[1, 3]]
    }
}

forecaster = make_reduction(
    regressor,
    transformers=[WindowSummarizer(**kwargs, n_jobs=1)],
    window_length=None,
    strategy="recursive",
    pooling="global",
    scitype="tabular-regressor"
)

pipe = ForecastingPipeline(
    steps=[
        ("daily_season", DateTimeFeatures(ts_freq="M")),
        ("forecaster", forecaster)
    ]
)

pipe.fit(train);
```

**Expected behavior**
That the pipeline would have been fitted without any error.

**Additional context**
The specific problem has only different starts for each panel, i.e. series start at different timepoints, but end at the same timepoint.

**Versions**
<details>

</details>

[...]

PR #4644, opened 2023-05-25
[BUG] fix for `make_reduction` with unequal panels time index for `global` pooling

kbpk:
This PR partially fixes a problem with the unequal panels' time index described in #4602.

However, the problem requires further investigation, especially how is `window_length` determined. Getting an index by `get_time_index` does not seem to be the proper way. There should not be a limit for window size dependent on data size, at most there will be no data (truncate). Also, `get_time_index` should return a min/max time interval over all panels (hierarchical data), then every split should be performed on its basis.


fkiraly (changes requested):
Thanks! This looks like a really nice improvement!

Based on inspection, the changes make sense, although some test cases show a problem with zero samples - not sure whether that is coming from.

Further, the logic for `tr.truncate is None` seems a bit hacky - it changes the behaviour in an existing case, so it would require deprecation.
Can we avoid the `n_cut=0` case?

fkiraly:
FYI @danbartl - any insight on the issue with the test cases?

kbpk:
Currently, according to the docs, we can only have `WindowSummarizer` transformers. Thus, I've included the n=0 case. If you wish, I can remove that fragment. Just confirm.

fkiraly:
ok, I suppose @danbartl is not going to respond. Apologies for the wait, @kbpk.

Primarily, we should ensure that the tests pass (and without hacks).

So, if the tests pass without the `n=0` fragments, I would be fine with that.

fkiraly:
hm, it seems this is failing on the 3.7 versions?

Now 3.7 support will be removed as 3.7 will soon reach end of life - but it is still worth checking here if everything is correct and as intented with the logic.

fkiraly:
scheduled for after 0.20.0 and 3.7 EOL due to 3.7 specific failures

fkiraly:
updated from `main` - will merge if tests pass on 3.8-3.11

Issue #4643, opened 2023-05-25
[ENH][MNT] umbrella issue: interface estimators (other than `AutoARIMA`) from `statsforecast` into `sktime`


yarnabrina:
This issue lists different tasks to interface more estimators from `statsforecast`. Till `0.18.0`, `AutoARIMA` (present in `statsforecast.arima`) was already interfaced as `StatsForecastAutoARIMA` inheriting more `ARIMA` focused base adapter `_StatsForecastAdapter`. As part of #4629, a more general base class has been attempted to conform to the general pattern of `statsforecast.models.*` estimators, and `StatsForecastAutoARIMA` has been updated accordingly. Given new general adapter, one can now add other `statsforecast` estimators inheriting from `_GeneralisedStatsForecastAdapter`.

Estimators to add:

1. `Automatic Foreacsting` models (more popular)
  1. [`AutoETS`](https://nixtla.github.io/statsforecast/models.html#autoets)
  2. [`AutoCES`](https://nixtla.github.io/statsforecast/models.html#autoces)
  3. [`AutoTheta`](https://nixtla.github.io/statsforecast/models.html#autotheta) - #4539
2. Other models listed [here](https://nixtla.github.io/statsforecast/models.html)

Expected tasks for each PR to add one of the models:

1. create a new concrete estimator class inheriting from `_GeneralisedStatsForecastAdapter`
2. add appropriate tags, especially `ignores-exogeneous-X`, `capability:pred_int` and `capability:pred_int:insample`
3. add all positional and keyword arguments required by the underlying estimator in `__init__` method, instead of using `*args` or `**kwargs`
4. create an instance of underlying `statsforecast` model in `_instantiate_model` method (indicated as abstact class in adapter)
5. add at least one, and preferably more (but not too high to reduce CI time), test cases in `get_test_params` method

To start with, this model specific classes can live in `sktime/forecasting/statsforecast.py`. Please keep @FedericoGarza as the author of the underlying estimators. The new class should be exported in `__all__` and also should be mentioned in `docs/source/api_reference/forecasting.rst`.

After (most of) these models are added, a maintainenace PR can split into different model into individual modules inside a directory (sub-package) of same name as current module (i.e. `sktime/forecasting/statsforecast/`) to help maintenance and avoid deprecation lag.

yarnabrina:
[...]

PR #4648, opened 2023-05-27
[ENH] statsforecast AutoETS direct interface estimator

yarnabrina:
Part of #4643.

This PR adds an interface to the `AutoETS` model provided in `statsforecast`. It's exposed to users as `StatsForecastAutoETS`, which inherits from `_GeneralisedStatsForecastAdapter` added in #4629.


fkiraly (approved):
Straightforward given the adapter class, great!

PR #4649, opened 2023-05-27
[ENH] statsforecast AutoCES direct interface estimator

yarnabrina:
Part of #4643.
Depends on #4648.

This PR adds an interface to the `AutoCES` model provided in `statsforecast`. It's exposed to users as `StatsForecastAutoCES`, which inherits from `_GeneralisedStatsForecastAdapter` added in #4629.


fkiraly (approved):
Straightforward given the adapter class, great!

PR #4652, opened 2023-05-29
 [ENH] Graphical Pipelines for any learning task (polymorphic)

benHeid:
(PR description by @fkiraly)

This PR implements the first prototype for graphical pipelines, see #4281 .

The design adopted is interface compatibility, rather than the polymorphic base class design.

Covers all learning tasks currently in sktime.

Should be considered experimental until consolidated.


benHeid:
FYI @fkiraly

benHeid:
This PR should enable a better async discussion on this topic and the concrete implementation. 

### Open TODOs:
* [ ] Test edge cases
* [ ] Implement Unit Tests
* [ ] Implement Subsetting/FeatureUnion Use-Cases
* [ ] Add additional Graphpipeline examples that are more complex
* [ ] Discuss the `mode==proba`  stuff that is currently implemented outside of the transformers
* [ ] Add documentation
* [ ] Add Example Jupyter Notebook

sktime/pipeline/computation_setting.py
```diff
@@ -0,0 +1,8 @@
+
+
+class ComputationSetting:
```
fkiraly, line 3: quick Q: what is this for - is this later going to be a builder class or config pattern?

sktime/pipeline/pipeline.py
```diff
@@ -0,0 +1,164 @@
...
+        # TODO do here subsetting using Subsetter
+        #   ColumnSelect Transformer
+        if name in self.steps:
+            return self.steps[name]
+
+        raise Exception("Required Input does not exist")
+
+    def add_step(self, skobject, name, edges, **kwargs):
+        """
+        TODO
```
fkiraly, line 36: :-)

sktime/pipeline/pipeline.py
```diff
@@ -0,0 +1,164 @@
...
+        self._last_step_name = name
+        return step
+
+
+    def fit(self, X, y=None, **kwargs):
+        # Fits the pipeline
+        self.steps["X"].buffer = X
+        self.steps["y"].buffer = y
+
+        self.computation_setting.method_resolution_order = ["transform", "predict"]
```
fkiraly, line 65: is this sth we need to hardcode in the `computation_setting`? I'd rather do dispatch on type or similar, but not sure if that has domino effects

sktime/pipeline/tests/regression_tests/test_pipeline_regression.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+def test_classifier_regression():
+    np.random.seed(42)
+    X, y = load_arrow_head(split="train", return_X_y=True)
+    clf_pipe = ExponentTransformer() * KNeighborsTimeSeriesClassifier()
+    clf_pipe.fit(X, y)
+    result = clf_pipe.predict(X)
+    general_pipeline = Pipeline()
+    for step in [{"skobject" : ExponentTransformer(),
```
fkiraly, line 44: note for later: we may like to think about simplified specification syntax, e.g.:

* if name not provided, generate a unique name
* if edges not provided, assume sensible defaults, i.e., last transformer given

further potential syntactic sugar:

* dunders `*` and `+` that work with the `Pipeline` and do the expected thing (`*` adds sequential after the last element added; `+` adds parallel after the last element's predecessor if unique)

I would do the above at the very last though. I think it is great for the user experience, but best done when the logic and structure internally are consolidated.

fkiraly (changes requested):
I think this is great, the design is lean and I can see how it would work in general.

For now, I would suggest:

* let's focus on getting the various `predict`-like methods to work and see how that looks like
* you may want to enable linting and/or precommits. The problem won't get smaller...
   * and failing linting prevents the regression tests from running.

sktime/pipeline/computation_setting.py
benHeid: This class is for communication between the pipeline and its step. The pipeline and all of its steps share the same instance, and if a step is executed, it can look up in this instance for specific information that is only available after calling a method on the pipeline. E.g., consider the following scenario:


```python
pipeline = Pipeline()

pipeline.add_step(transformer, ...)
pipeline.add_step(forecaster, ...)

pipeline.fit()

pipeline.predict() # <- After this line, it is clear that on the forecaster, predict needs to be called

pipeline.predict_interval() # <- After this line, it is clear that on the forecaster, predict_interval needs to be called
```

In this scenario, the information is available after calling a method. To share such information, I introduced this class. So it is more a config class or something like that.
Also, all keyword arguments, the pipeline received during predict/transform etc., are shared via this class with its steps.  

Probably, the naming can be improved. Perhaps, there is also a better way of sharing such information.

sktime/pipeline/pipeline.py
benHeid: I have introduced the method_resolution_order due to the `predict_interval` method and the assumption that there might be a forecaster that is only deterministic, and by specific postprocessing steps, the forecast become probabilistic.
But perhaps using dispatching is better and more general. I need to think about it

sktime/pipeline/computation_setting.py
benHeid: The only alternatives that come in my mind are:

* The pipeline sets this information in each step explicitly. A disadvantage of this solution is that these values stay there until they are deleted again. Potentially error might arise if, somewhere, the old value is not deleted after execution or not overridden in a new execution. Furthermore, we must remind where these values are set if a new element is needed, then we have to adapt it everywhere.
* We pass all values of the computation setting explicitly when calling the `get_result`. Advantage, we get rid off the computation setting class, and the validity is automatically restricted to one pipeline execution (`predict`, `fit`, .. etc. call). Furthermore, it is more explicit and clearer from where the values originate. A disadvantage is that there might be a lot of arguments in the end. However, in total, I think this is currently my preferred solution, and I also prefer this over the current solution.

fkiraly: the last solution looks better from a structural perspective to me, but it breaks the "iron rule" in the sklearn-like design that `predict` and `transform` calls should not mutate the estimator - do I understand that correctly?

benHeid: I pass these arguments now through the pipeline instead of using computation_setting.

benHeid:
> I think this is great, the design is lean and I can see how it would work in general.
> 
> For now, I would suggest:
> 
> * let's focus on getting the various `predict`-like methods to work and see how that looks like
> * you may want to enable linting and/or precommits. The problem won't get smaller...
>   
>   * and failing linting prevents the regression tests from running.

Regarding the predict_* methods, currently I am not sure how I should implement them, since the inverse_transform method, for example, fails if it is called on input_data that are from a predict_proba call. 

I think there are two options:
* Extract the probability information from a call of predict_quantiles (Default behaviour of sktime BaseForecaster).
* Check if all steps in the pipeline can handle the output of a predict_proba call as input, then execute it. If not, use default behaviour (Seems to be complicated to implement, since I have no idea how I should do the checking...)

fkiraly:
hm, not sure what to do with `predict_proba` / `inverse_transform`.

I think the current default is indeed applying to `predict_quantiles` and getting a distribution from that.

fkiraly (changes requested):
I've reviewed this, and I must say - a serious amount of work (design and implementation)! Really nice!

The errors are "first pass" issues with interface violation of the basic `sklearn` contract that `sktime` follows. The contract requires that:

* all args of `__init__` are written to `self`-attributes, with values *as passed*. Let's call these parameter attributes.
* after that, the *parameter attributes are never changed at any point in time* (other attributes can change, but this is specifically about the parameter attributes).

The violation is due to `step_information`, which is mutated and in general not identical to what was passed.

The basic pattern to get around that is to use a copy such as `self._step_information` for initialization logic, and an attribute `self.step_information_` for any mutation after `fit`, if needed (important: attributes mutated in or past `fit` should end in underscore, also an `sklearn` contract requirement).

So, in the `__init__` specifically, `step_information` should be written to `self.step_information` and that's the last time we change that attribute. Anything that happens to `self.step_information` currently should be done with `self._step_information`.

fkiraly:
from pipeilnes meeting on Aug 11:

* it's not 100% clear how to satisfy this requirement while also having `add_step`
* consensus from discussion: `add_step` should return a *new* `Pipeline` with `step_information`; in `__init__`, we only do validation, which is the "checking" part of current `add_step`; checking from current `add_step` is refactored to `_check_stepinformation` (or similar)

Minor comment: perhaps it's better to rename the arg `step_informations` to `steps` - that's in line with `sklearn` pipelines and much more convenient to type...

benHeid:
> from pipeilnes meeting on Aug 11:
> 
> * it's not 100% clear how to satisfy this requirement while also having `add_step`
> * consensus from discussion: `add_step` should return a _new_ `Pipeline` with extended `step_information`; in `__init__`, we only do validation, which is the "checking" part of current `add_step`; checking from current `add_step` is refactored to `_check_stepinformation` (or similar), and that's called in `__init__` (therefore, `add_step` will trigger the check via `__init__` of the newly constructed object)
> 
> Minor comment: perhaps it's better to rename the arg `step_informations` to `steps` - that's in line with `sklearn` pipelines and much more convenient to type...

I have fixed the errors. Also, I ensured that it works together with the ForecastingGridSearch (with additional few changes  in #5175 )

sktime/tests/test_all_estimators.py
```diff
@@ -894,6 +895,8 @@ def test_inheritance(self, estimator_class):
                 f"estimator: {estimator_class} has fit method, but"
                 f"is not a sub-class of BaseEstimator."
             )
+        if isinstance(Pipeline(), estimator_class):
```
benHeid, line 898: @fkiraly, any ideas on how to fix this test in a prettier way? 

The problem is that the Pipeline inherits from BaseEstimator but not from a more specific class as BaseForecaster, etc. This results in the failing of the assertion about the number of base classes.

docs/source/api_reference/pipeline.rst
```diff
@@ -0,0 +1,17 @@
+.. _regression_ref:
+
+Graphical Pipeline
```
fkiraly, line 3: I think this should also list `make_pipeline`, so there's two objects here - the linear and the graphical one.

sktime/pipeline/pipeline.py
```diff
@@ -0,0 +1,584 @@
...
+
+class Pipeline(BaseEstimator):
+    """Implementation of a Graphpipeline.
+
+    This class is a generalized graph pipeline. Generalized means that it can
+    contain forecasters, classifiers, etc. The graph pipeline mean that the structure
+    is not linear. I.e., the each element of the pipeline can be the input of multiple
+    other steps and not only one sucessors.
+
+    **Acknowledgements**
```
fkiraly, line 33: acknowledgments should be at the bottom. You may also like to cite `pywatts` in a `References` section.

sktime/pipeline/pipeline.py
```diff
@@ -0,0 +1,584 @@
...
+    This graphical pipeline is inspired by pyWATTS that is developed by the Institute
+    for Automation and Applied Informatics (IAI) at Karlsruhe Institute of Technology.
+    The implementation is supported by IAI and the author benHeid is funded by
+    HelmholtzAI.
+
+    `fit(y, X, *args)` - changes state by running `fit` on all sktime estimators and
```
fkiraly, line 39: should this include polymorphism, or are we primarily focused on forecasters for now?

sktime/pipeline/pipeline.py
```diff
@@ -0,0 +1,584 @@
...
+
+        Example 1: Simple sequential pipeline of transformers using the generalized
+        non-sequential pipeline implementation
+    >>> y, X = load_longley()
+    >>> general_pipeline = Pipeline()
+    >>> for step in [
+    ...     {"skobject": ExponentTransformer(), "name": "exp", "edges": {"X": "X"}},
+    ...     {"skobject": BoxCoxTransformer(), "name": "box", "edges": {"X": "exp"}},
+    ...     ]:
+    ...     general_pipeline = general_pipeline.add_step(**step)
```
fkiraly, line 103: this seems cumbersome - should it not be possible to simply construct the pipeline with the `step` argument, and have `add_step` as a secondary option?

fkiraly (changes requested):
Overall, looks great!

I have the following blocking requests - for discussion or change:

* the arg `step_informations` should be called `steps` to be consistent with pipelines in `sklearn`, `sktime`, etc
* it should be possible to construct the `Pipeline` with `steps` dicts like in the examples, directly. `add_step` should be considered a shortcut for a `set_params` call rather than the primary construction interface. I think this one is pretty important.
* once this is possible, the examples should become simpler.
* I would also credit @ViktorKaz, at least with an independent design - the design in https://github.com/sktime/sktime/pull/856 turns out to be very similar imo, given that the `edges` arg appears there im similar form.

fkiraly:
(I have no requests or suggestions for improvement re implementation - this looks great! change requests are mostly cosmetic and user facing API design, but imo important for consistency)

sktime/pipeline/pipeline.py
benHeid: The pipeline is also working for classifiers. Compare `test_classifier_regression` in test_pipeline_regression

sktime/pipeline/pipeline.py
benHeid: You mean passing the list to the creation of the pipeline?  It is almost possible, I will enable it. Would you then change all tests?

fkiraly: yes, that's what I meant. I would still allow both types of syntax though.

sktime/pipeline/pipeline.py
fkiraly: ah, great!

sktime/tests/test_all_estimators.py
fkiraly: well, this works for now. We should later revisit the test.

fkiraly (approved):
My comments are all addressed, thanks.

The main point I see is the wider question how we ensure consistency with the boilerplate layer and the respective contracts, and that we run `check_estimator` with examples that claim to be forecasters or classifiers with consistent interfaces - but I suppose we can revisit that in the next iteration of the design.

For now, we should treat it as an exciting experimental construct. Perhaps some warnings in the docstrings might be appropriate...

PR #4673, opened 2023-06-04
[MNT] resolve pre-commit issues on `main`

yarnabrina:
Running `pre-commit` on main branch finds a few linting issues. These are probably introduced because of old PR's before adding the linters, or because of some skipped CI's. This PR attempts to solve these.

Steps followed:

1. Ran `pre-commit run --all-files` on `main` branch.
2. Got 3 issues regarding misisng module docstring, trailing whitespace and missing newline.
3. Fixed one by one.


yarnabrina:
A related note is that most hook versions in `.pre-commit-config.yaml` is outdated. If updated to latest versions using `pre-commit autoupdate`, it leads to changes to ~140 files, and a lot of file blame will be modified. Skipped for now, but this may be worth considering in future.

fkiraly:
> Skipped for now, but this may be worth considering in future.

I would recommend, open an issue with the "maintenance" tag

fkiraly (approved):
Thanks!

odd that nothing complained when we merged these, some of these files were changed recently.

PR #4679, opened 2023-06-06
[DOC] correct some failing doctests

mdsaad2305:

In some cases while using doctest, the expected outputs for certain items were incorrect, resulting in the failure of the corresponding doctests.

### DIFF
--- a/sktime/classification/interval_based/_stsf.py
+++ b/sktime/classification/interval_based/_stsf.py
@@ -82,7 +82,7 @@ class SupervisedTimeSeriesForest(BaseClassifier):
     >>> X_test, y_test = load_unit_test(split="test", return_X_y=True)
     >>> clf = SupervisedTimeSeriesForest(n_estimators=5)
     >>> clf.fit(X_train, y_train)
-    SupervisedTimeSeriesForest(...)
+    SupervisedTimeSeriesForest(n_estimators=5)
     >>> y_pred = clf.predict(X_test)
     """
 
--- a/sktime/classification/interval_based/_tsf.py
+++ b/sktime/classification/interval_based/_tsf.py
@@ -79,7 +79,7 @@ class TimeSeriesForestClassifier(
     >>> X_test, y_test = load_unit_test(split="test", return_X_y=True)
     >>> clf = TimeSeriesForestClassifier(n_estimators=5)
     >>> clf.fit(X_train, y_train)
-    TimeSeriesForestClassifier(...)
+    TimeSeriesForestClassifier(n_estimators=5)
     >>> y_pred = clf.predict(X_test)
     """
 


fkiraly:
That's odd - could you explain why this is a failure?

The doctests are not failing on main, and the `...` should catch any expression therein as correct.

I wouldn't mind merging this as it passes too, but I want to understand your rationale more precisely.

yarnabrina:
`...` needs to enable the `` flag.

Does this help?

```python3
import dataclasses


@dataclasses.dataclass
class Calculator:
    """Dummy calculator class.

    Parameters
    ----------
    x : float
        first number
    y : float
        second number
    """
    x: float
    y: float

    def add(self: "Calculator") -> str:
        """Add ``self.x`` and ``self.y``.

        Returns
        -------
        float
            sum of ``self.x`` and ``self.y``

        Examples
        --------
        >>> calculator = Calculator(1, 1)  # should pass
        >>> calculator  # should pass
        Calculator(x=1, y=1)
        >>> calculator.add()  # should pass
        sum of self.x=1 and self.y=1 is self.x + self.y=2.
        >>> calculator = Calculator(2, 3)  # should pass
        >>> calculator  # should fail
        Calculator(...)
        >>> calculator.add()  # should fail
        ...
        >>> calculator = Calculator(3, 2)  # should pass
        >>> calculator  # should pass # doctest: +ELLIPSIS
        Calculator(...)
        >>> calculator.add()  # should pass # doctest: +ELLIPSIS
        sum...3...2...5.
        """
        print(f"sum of {self.x=} and {self.y=} is {self.x + self.y=}.")


if __name__ == "__main__":
    import doctest

    doctest.testmod(verbose=True)

```

I believe CI uses does not call `doctest` directly, and use through `pytest`. `pytest` adds `ELLIPSIS` as a default option, so it'll be unable to detect the failure.

Ref. https://github.com/pytest-dev/pytest/blob/main/src/_pytest/doctest.py#L70-L75

yarnabrina:
@mdsaad2305, please feel free to add more details to my above comment.

A question: did you check if `doctest` fails only for the two modules you listed or not? I just ran the follwing, which shows quite a lot more failures, including some import errors. If you also get those, can you handle those in the same PR?

> find sktime -name "*.py" | xargs -n 1 python -m doctest

mdsaad2305:
I thought the ```...``` was a mistake since I've been getting this error
```
Failed example:
    clf.fit(X_train, y_train)
Expected:
    SupervisedTimeSeriesForest(...)
Got:
    SupervisedTimeSeriesForest(n_estimators=5)
```
    
But if that was intentional, apologies for the inconvenience.

mdsaad2305:
@yarnabrina 
Initially I was using the ```-v``` flag with doctest which showed the passed test along with the few failing tests. I might have overlooked some of the failing tests.
 
I'll try to fix these errors in this PR. Any tips for fixing these issues?

fkiraly:
> I don't know the amount of places it may affect

A quick search for `(...)` gives 124 instances in 94 files, so pretty substantial as it looks like.

PR #4681, opened 2023-06-07
[DOC] adds a banner for non-latest branches in read-the-docs

### LINKED ISSUE #4677: [DOC] change default documentation version from `latest` (`main` branch) to `stable` (latest tag) in read-the-docs configuration
Currently the [documentation website](https://www.sktime.net) defaults to `latest`, i.e. the `main` branch. So, it is possible that it shows some estimators or arguments which are not yet released and won't be available for the general users who download from `PyPI` or `conda-forge`.

Read-the docs allow to change the default version, and it can be considered to switch to `stable` to refer to [latest semantic version](https://docs.readthedocs.io/en/stable/versions.html). This is likely to be what's used by most of the direct users of `sktime` (i.e. not through another framework whose dependency is `sktime`).

To make the change, the read-the-docs owner can log in, go to `Admin`, go to `Advanced Settings`, change `Default version` and save the changes. It's likely to trigger a new build and then will be in effect once it's complete.

yarnabrina:
Closes #4677 

Adds a new dependency `sphinx-version-warning` as part of `docs` optional dependency. The changes in this PR follows `elyra` example almost unchanged, though `ray`, `marshmallow`, etc. also seem to have same format.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -49,6 +49,7 @@ extensions = [
     "myst_parser",
     "sphinx_design",
     "sphinx_issues",
+    "versionwarning.extension",
 ]
 
 # Recommended by sphinx_design when using the MyST Parser
@@ -435,3 +436,15 @@ intersphinx_mapping = {
 
 # -- Options for _todo extension ----------------------------------------------
 todo_include_todos = False
+
+# sphinx-version-warning config
+versionwarning_messages = {
+    "latest": (
+        "This document is for the development version. "
+        'For the stable version documentation, see <a href="/en/stable/">here</a>.'
+    )
+}
+
+# Show warning at top of page
+versionwarning_body_selector = "div.document"
+versionwarning_banner_title = ""
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -128,6 +128,7 @@ docs = [
     "sphinx_issues==1.2.0",
     "sphinx-gallery==0.6.0",
     "sphinx-design==0.3.0",
+    "sphinx-version-warning",
     "sphinx==4.1.1",
     "tabulate",
 ]


yarnabrina:
@fkiraly The documentation corresponding to this PR will not show the banner, as by configuration the banner will be shown only against `latest` tag, which corresponds to default branch, i.e. `main` for `sktime`. For the same reason, I'm unable to test locally as this feature seems to be specific to read the docs hosting.

Requesting to consider merging it to see if it works there, if not we can revert.

fkiraly (approved):
Sure - let's try and consider demerge/revert

yarnabrina:
@humitos will it be possible for you to please take a look at this PR?

I want to show a banner informing users that by default they are in latest instead of stable. I tried to apply the similar configuration as I saw in other repos, but it didn't work. Can you please suggest how to resolve this?

humitos:
Hi @yarnabrina. I seems the banner is not being injected because the default CSS selector is not found in the page. You can try using the config `versionwarning_body_selector` to define the correct one: https://sphinx-version-warning.readthedocs.io/en/latest/configuration.html#confval-versionwarning_body_selector. You can start trying with `.bd-article` for example.

yarnabrina:
Hi @humitos, thanks for the response. I tried changing `div.document` to `div.bd-article` locally (after setting `versionwarning_project_version = "latest"`), but unfortunately couldn't see any banner.

I can not tell if it's because `sphinx-version-warning` works only on Read the Docs environment, or needs a specific Sphinx version (we have a pin on `4.1.1`, but in the process of removing it), or etc. Given I'm not a CSS expert (almost zero knowledge), I guess it may take quite a few trial and error to find the suitable selector and will be easier if I could test locally. Can you please suggest some ways to identify what to use and how to check the functionality locally?

humitos:
> I tried changing `div.document` to `div.bd-article`

I think it's not a `div`. I'd suggest you to use only `.bd-article` first as I mentioned in my previous comment.

> I can not tell if it's because `sphinx-version-warning` works only on Read the Docs environment, or needs a specific Sphinx version

IIRC, this extension only works when building on Read the Docs, yeah. So, you should make your tests there.

PR #4682, opened 2023-06-07
[MNT] except some DL and `numba` based estimators from tests to prevent memory overload

fkiraly:
Update: removes some recent esetimators from the tests.

Reference to the memory issue is added to the comments in the config files, so we can get back to this, e.g., when we change the testing framework.

---

Suspecting `MACNNClassifier` to be beind the memory related crashes on windows, python 3.9 (https://github.com/sktime/sktime/issues/4610), seeing what happens if we except it from the tests.

Suspicion is due to it being a recent addition. FYI @achieveordie.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -196,7 +196,7 @@ jobs:
           mkdir -p testdir/
           cp .coveragerc testdir/
           cp setup.cfg testdir/
-          python -m pytest -v -n 1
+          python -m pytest -v
 
       - name: Publish code coverage
         uses: codecov/codecov-action@v3
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -196,7 +196,7 @@ jobs:
           mkdir -p testdir/
           cp .coveragerc testdir/
           cp setup.cfg testdir/
-          python -m pytest
+          python -m pytest -v
 
       - name: Publish code coverage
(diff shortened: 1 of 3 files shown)


fkiraly:
that wasn't it - windows 3.9 is still breaking.

fkiraly:
ok, now removing the last DL estimators added before that - `SimpleRNNClassifier`, `SimpleRNNRegressor`

fkiraly:
we should be able to infer which estimator is the culprit by observing whether 3.9 or 3.10 fails - because estimators are deterministically spread across versions using a modulo 3 grid pattern, with estimators in lexicographic order. I.e., an estimator will be tested on windows 3.7 and 3.10, or on windows 3.8 and 3.11, or on windows 3.9.

If 3.9 fails at removal of one estimator, then the failing estimator must be before it in the lexicographic order (since is removal did not change the position of the culprit)
If 3.8 and 3.11 fail at removal of one estimator, then the failing estimator must be after it in the lexicographic order (because its removal changed the position of the culprit)

But now I've removed one and 3.10 fails, but 3.7 doesn't. This is confusing. Is the failure therefore stochastic across versions?

fkiraly:
so, I changed this to a single worker to allow better diagnosis of what is failling, and now all tests pass :-(

fkiraly (commented):
hm, seems like this fixes it

Issue #4683, opened 2023-06-07
[BUG] column-vectorized forecaster predictions are in wrong order if input is `DataFrame` with non-lexicographically ordered columns


joaovitorfranco:
Update by @fkiraly:

The minimal bug is actually in vectorization, if columns are not lexicographically ordered. This will then happen in *any* estimator.

Example code:

```python
from sktime.datasets import load_macroeconomic
from sktime.forecasting.naive import NaiveForecaster

X = load_macroeconomic()

f = NaiveForecaster()
f.fit(y=X, fh=[1, 2, 3, 4])
f.predict()
```

The bug is likely in `vectorize_est`, and possibly introduced with #4195 - FYI @hoesler.
Update to the update: indeed coming from there and introduced with #4195, but actually coming from `DataFrame.pivot`.

The interface tests do not pick this up since the return is correct from a typing perspective, with column and row indices as expected. We should therefore add a test with the `NaiveForecaster`.

---

**Describe the bug**
The method `_get_forecaster_X_prediction` in the `ForecastX` class predicts wrong values probably due to incorrect column names and order. 

**To Reproduce**
```python
import pandas as pd
import numpy as np

from sklearn.ensemble import RandomForestRegressor
from sktime.datasets import load_macroeconomic
from sktime.forecasting.compose import make_reduction

from sktime.forecasting.compose import ForecastX
from sktime.forecasting.arima import AutoARIMA
from sktime.forecasting.arima import ARIMA


data = load_macroeconomic()
y = data["unemp"]
X = data.drop(columns=["unemp"])

regressor = RandomForestRegressor(random_state=42)

forecaster = make_reduction(
    regressor,
    window_length=12,
    strategy="recursive",
    pooling="global",
)

pipe = ForecastX(forecaster_X=forecaster, forecaster_y=forecaster)

pipe.fit(X=X, y=y, fh=[1,2,3,4,5])

pipe._get_forecaster_X_prediction()
```

**Expected behavior**
Comparing X with the predicted future values of X and they should be close to each other. This is not happening in this example.

**Version**
0.18.0

fkiraly:
This is odd. We had a problem with incorrectly ordered columns before - issue #3986 - and I thought we fixed it in this PR: https://github.com/sktime/sktime/pull/3987

This was definitely merged before 0.18.0. Let me have a look, as I've written the bugfix in #3987, perhaps I missed sth.

fkiraly:
[...]

PR #4684, opened 2023-06-07
&hoesler [BUG] Fix `vectorize_est` returning jumbled columns for column vectorization, `pd.DataFrame` return, if column names were not lexicographically ordered

fkiraly:
Fixes https://github.com/sktime/sktime/issues/4683.

The fix is adding some lines of code that correct the behaviour of `Dataframe.pivot` used in the code (see #4633) by adding some more lines that prevent the columns from being lexicographically sorted, by manually reverting the sorting that `pivot` does.

Further:

* adds a test, indirectly through the minimal forecaster case that failed in #4683.
* fixes an unreported bug in forecasting metrics that could end up with the `name` attr being `None`. This was masked by how `pivot` was previously applied and would result in a vectorization result with all `None` columns.

### DIFF
--- a/sktime/datatypes/_vectorize.py
+++ b/sktime/datatypes/_vectorize.py
@@ -593,18 +593,8 @@ class VectorizedDF:
             ret.append((group_name, col_name, est_i_result))
 
         if return_type == "pd.DataFrame":
-            df_long = pd.DataFrame(ret)
-            cols_right_order = df_long.loc[:, 1].unique()
-
-            df = df_long.pivot(index=0, columns=1, values=2)
-            # DataFrame.pivot sorts the columns (is this a bug? see #4683)
-            # either way, we need to fix this:
-            df = df.loc[:, cols_right_order]
-
-            # remove "0" and "1" from index/columns name
-            df.index.names = [None] * len(df.index.names)
-            df.columns.name = None
-
+            df = pd.DataFrame(ret)
+            df = df.pivot(index=0, columns=1, values=2).reindex(df[1], axis=1)
             # TODO: add test case for tuple index
             try:
                 df.index = pd.MultiIndex.from_tuples(df.index)
--- a/sktime/datatypes/_vectorize.py
+++ b/sktime/datatypes/_vectorize.py
@@ -593,7 +593,18 @@ class VectorizedDF:
             ret.append((group_name, col_name, est_i_result))
 
         if return_type == "pd.DataFrame":
-            df = pd.DataFrame(ret).pivot(index=0, columns=1, values=2)
+            df_long = pd.DataFrame(ret)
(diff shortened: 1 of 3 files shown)


fkiraly:
FYI @hoesler, given this is changing code you most recently edited, would appreciate your eyes on this.

sktime/datatypes/_vectorize.py
```diff
@@ -593,7 +593,18 @@ def _to_iter(e):
...
+
+            df = df_long.pivot(index=0, columns=1, values=2)
+            # DataFrame.pivot sorts the columns (is this a bug? see #4683)
+            # either way, we need to fix this:
+            df = df.loc[:, cols_right_order]
+
+            # remove "0" and "1" from index/columns name
+            df.index.names = [None] * len(df.index.names)
+            df.columns.name = None
+
```
hoesler, line 607: What about this? pivot supports a sort arg since 1.3 (https://github.com/pandas-dev/pandas/pull/40954)

```suggestion
            df = pd.DataFrame(ret)
            try:  # pandas >= 1.3
                df = df.pivot(index=0, columns=1, values=2, sort=False)
            except TypeError:
                df = df.pivot(index=0, columns=1, values=2).reindex(df[1], axis=1)
```

hoesler: Ah, sorry. That only applies to pivot_table. So it would be just 

```suggestion
            df = pd.DataFrame(ret)
            df = df.pivot(index=0, columns=1, values=2).reindex(df[1], axis=1)
```

fkiraly: ah, thanks - that's much nicer.

Quick questions:

* would that work on `pandas` 1.1 too? we try to be downwards compatible where that is still possible
* there's also the matter of the spurious column/level labels "0" and "1" that appear through using `pivot`, they should optimally be replaced by the labels that were there to start with in the data, or removed
    * for an example of this malfunction (bit ugly but ultimatly inconsequential), search for `forecasters_` in this notebook: https://github.com/sktime/sktime-tutorial-ODSC-Europe-2023/blob/main/notebooks/03_forecasting.ipynb

fkiraly: thanks for the swift review btw! 🚀

fkiraly:
FYI @hoesler, I combined your suggestion with mine:

* the `reindex` works, but we first need to make the index unique. Just taking row 1 will have non-unique elements and leads to column duplication.
* we still need to reset the level names, that is not solved by your solution.

hoesler:
> * we first need to make the index unique

But if we have duplicate column names, e.g. `a, b, a`, how does your version preserve the correct order if you make col names unique?
Which scenario can lead to duplicate names in the first place? Probably we should add a test case for that.

PR #4689, opened 2023-06-10
[DOC] developer guide on remote setup of test suite

fkiraly:
This doc adds a section to the developer guide on the `sktime` testing framework that describes the remote setup of the test suite, including the subsampling.

It is a bit compact, so a case could be made to extend it into its own section that also explains the GHA setup (not sure if that's duplicative).

FYI @yarnabrina who requested/suggested this - review appreciated.


yarnabrina (approved):
Looks great 👍🏽

achieveordie (approved):

PR #4705, opened 2023-06-15
[BUG] fix dead source link for UEA datasets

### LINKED ISSUE #4703: [BUG] CI failures when loading from UEA time series classification data repo
The notebook CI seems to have started failing without any related code change prior.

The failure occurs when attempting to load the `Yoga` dataset, with exception

```
Invalid dataset name =Yoga is not available on extract path =None. Nor is it available on [https://timeseriesclassification.com/.](https://timeseriesclassification.com/)
```

The code snippet that causes the error seems to be

```python
from sktime.datasets import load_UCR_UEA_dataset

X, y = load_UCR_UEA_dataset(name="Yoga", return_X_y=True)
```

but this seems to work fine locally.

Update: most datasets no longer work locally either, see below.

fkiraly:
UEA have changed the location of their classification datasets, which breaks the loaders that directly access the URL.

This PR fixes the issue temporarily, for the mid term we want to think about mirrors that make the data loaders less dependent on individual data sources that can deprecate without warning.

Fixes #4703

### DIFF
--- a/sktime/datasets/_data_io.py
+++ b/sktime/datasets/_data_io.py
@@ -195,7 +195,10 @@ def _load_dataset(name, split, return_X_y, return_type=None, extract_path=None):
         ):
             # Dataset is not already present in the datasets directory provided.
             # If it is not there, download and install it.
-            url = "https://timeseriesclassification.com/Downloads/%s.zip" % name
+            url = (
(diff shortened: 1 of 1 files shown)


fkiraly:
failure is on main, so merging quickly

PR #4712, opened 2023-06-16
[BUG] Clustering Estimators to accept a callable / string from `sktime`, `sklearn` and `scipy` to calculate distances

### LINKED ISSUE #4697: [BUG] Time series DBSCAN crashes on str distance key
**Describe the bug**
I have a numpy array of size (1560, 2) where column 0 is integer data and column 1 is time data in unix format.
So when I try to use the .fit(array) method, it gives the following output:

```
T>>> tsdbscan = TimeSeriesDBSCAN('euclidean')
>>> tsdbscan.fit(X)
  File "<stdin>", line 1, in <module>
  File "/tmp/env/lib/python3.11/site-packages/sktime/clustering/base.py", line 119, in fit
    self._fit(X)
  File "/tmp/env/lib/python3.11/site-packages/sktime/clustering/dbscan.py", line 124, in _fit
    distmat = distance(X)
              ^^^^^^^^^^^
TypeError: 'str' object is not callable
```

**To Reproduce**

```python
from sktime.clustering.dbscan import TimeSeriesDBSCAN as TSDBSCAN

X = np.random.randint(0, 100, (100, 2))

tsdbscan = TSDBSCAN(distance='dwt')
tsdbscan.fit(X)
```

**Expected behavior**
Expecting to work with distance as str.

**Additional context**

**Versions**
0.19.1

achieveordie:
#### Reference Issues/PRs

Fixes #4697

I've added the functionality to convert strings into callables in `BaseClustering` since this functionality can be used in other clustering estimators where `distance` or `metric` plays a significant role. It also allows for custom callable apart from the three main modules that provide most distance calculating methods.
 
None

In its current state, passing `sktime`'s distance kernels as strings will fail (eg: 'dtw'). @fkiraly I think the best and most straightforward strategy will be to adopt a global dictionary which maps a string to the corresponding kernel. What's your opinion?

I've edited the `get_test_params()` method to include:
- A `sklearn` callable.
- A string corresponding to `sklearn`'s metric module
- A string corresponding to `scipy.spatial.distance` module

The current state has been tested using `check_estimator()`.
Notice how `params3` has been commented out, this is the string that would map to `sktime`'s distance kernel.

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/clustering/base.py
+++ b/sktime/clustering/base.py
@@ -4,10 +4,14 @@ __author__ = ["chrisholder", "TonyBagnall"]
 __all__ = ["BaseClusterer"]
 
 import time
+from inspect import getmodule
 from typing import Any, Union
 
 import numpy as np
 import pandas as pd
+from scipy.spatial.distance import _METRICS as SCIPY_DISTANCE_METRICS
+from sklearn.metrics import pairwise_distances
+from sklearn.metrics.pairwise import distance_metrics
 
 from sktime.base import BaseEstimator
 from sktime.datatypes import check_is_scitype, convert_to
@@ -17,6 +21,7 @@ from sktime.utils.validation._dependencies import _check_estimator_deps
 
 # Valid input types for clustering
 TimeSeriesInstances = Union[pd.DataFrame, np.ndarray]
+SKLEARN_DISTANCE_METRICS = distance_metrics()
 
 
 class BaseClusterer(BaseEstimator):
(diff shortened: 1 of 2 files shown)


fkiraly:
I think this might be going not quite in the right direction?

The distances we want to address are the ones in `dists_kernels` for panels, not the scalar ones?

achieveordie:
Well you're correct - there's my lesson to not code during fever, lol. Let me correct this in the upcoming days.

fkiraly:
Ouch - get well soon! (unless you already are, of course)

fkiraly:
The idea of a "resolver dictionary" makes sense - you just encode with strings either estimator instances or callables.

I have a slight preference for estimator instances as they are modular and plug & play, but in the few instances where we are already doing this, it's callables (from the `distances` module). In parts this is due to legacy designs, and it also causes some secondary issues (e.g., data format handling).

fkiraly:
nice! Btw, I was thinking about the str-to-estimator idea, and thought - why not do this in full generality?

That is, there might be some "shorthands", but should it not also be possible to store estimator configurations as serializable objects?

Here: https://github.com/sktime/sktime/pull/4738

Issue #4698, opened 2023-06-12
[ENH] design discussion - `fh` related contract if passed in both `fit` and `predict`


yarnabrina:
update by @fkiraly: the below is mainly a design discussion related to the `BaseForecaster` contract in a case where both `fit` and `predict` have a non-trivial `fh` argument passed, specifically in some cases where the user might expect different behaviour from status quo.

The behaviour of `ForecastX` is a special case, in fact all forecasters behave like that. (it is non-obvious from the example that it is a special case that generalizes)
That special behaviour is particularly unintuitive for `ForecastX` though, more obviously so than for other forecasters.

---

original post by @yarnabrina:

---

As of now, `ForecastX` seems to allow prediction with only the identical `fh` provided in `fit` or a permutation of it, and of nothing else. Even a subset of the fit fails, so does in-sample predictions.

For example, the following composite estimator is trained with `fh=[1, 2, 3]`. As an user, I expect to be able to predict for just `1` or just `[2, 3]` etc., as these observations must be already predicted by `forecaster_X`. Also, since the `fit` already contained the training data for `X`, I expect to be able to predict in-sample for all (negative) horizons. However, both of these cases fail.

## Reproducible Example

```python3
from sktime.datasets import load_longley
from sktime.forecasting.compose import ForecastX
from sktime.forecasting.naive import NaiveForecaster

y, X = load_longley()
pipe = ForecastX(NaiveForecaster(), NaiveForecaster())
pipe = pipe.fit(y, X=X, fh=[1, 2, 3])

# predicting future with identical fh - works
pipe.predict([1, 2, 3])

# predicting future with shuffled fh - works
pipe.predict([3, 1, 2])

# predicting future with subset of fh - fails
pipe.predict([1, 2])

# attempting in-sample prediction - fails
pipe.predict([-3])
```

## Error Message

```pycon
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/path/to/venv/lib/python3.11/site-packages/sktime/forecasting/base/_base.py", line 396, in predict
    fh = self._check_fh(fh)
         ^^^^^^^^^^^^^^^^^^
  File "/path/to/venv/lib/python3.11/site-packages/sktime/forecasting/base/_base.py", line 1693, in _check_fh
    raise ValueError(
[...]

PR #4716, opened 2023-06-17
[ENH] support passing horizons as `range` object in `ForecastingHorizon` and in `fit` and `predict` methods

yarnabrina:
Related #4698.

Before this PR, `ForecastingHorizon` did not support passing `range` objects. Since asking for all horizons till a specific one is quite common, one had to do `ForecastingHorizon(values=list(range(12)))` to get next `12` predictions, or use `numpy.arange`.

This PR adds support for `range` (and tests for the same), so that one can do `ForecastingHorizon(values=range(12))` to do the above. This also updates the documentations slightly to explicitly note that list/array passed to create horizon must be of integers.

(note: also added few contribution types to my name as part of this PR. can be removed by reverting commit https://github.com/sktime/sktime/pull/4716/commits/f8d9429c2ff3b32b3a176b7b124c788efaf871af if inappropriate.)

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -2145,6 +2145,10 @@
       "contributions": [
         "bug",
         "code",
+        "doc",
+        "ideas",
+        "maintenance",
+        "review",
         "test"
       ]
     },
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -2148,6 +2148,8 @@
         "doc",
         "ideas",
         "maintenance",
+        "mentoring",
+        "question",
         "review",
         "test"
       ]
--- a/sktime/forecasting/base/_fh.py
+++ b/sktime/forecasting/base/_fh.py
@@ -80,7 +80,7 @@ def _check_values(values: Union[VALID_FORECASTING_HORIZON_TYPES]) -> pd.Index:
 
     Parameters
     ----------
-    values : int, list of int, array of int, certain pd.Index types
+    values : int, range, list of int, array of int, certain pd.Index types
         Forecasting horizon with steps ahead to predict.
 
     Raises
@@ -107,6 +107,11 @@ def _check_values(values: Union[VALID_FORECASTING_HORIZON_TYPES]) -> pd.Index:
     elif is_int(values):
         values = pd.Index([values], dtype=int)
 
+    # convert range object to pandas.RangeIndex
+    # range has to be for integers, no need to separate check
+    elif isinstance(values, range):
+        values = pd.Index(values)
+
     elif is_timedelta_or_date_offset(values):
         values = pd.Index([values])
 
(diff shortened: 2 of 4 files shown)


.all-contributorsrc
```diff
@@ -2145,6 +2145,10 @@
       "contributions": [
         "bug",
         "code",
+        "doc",
+        "ideas",
```
fkiraly, line 2149: you could also add `"mentoring"`, `"question"`

fkiraly (approved):
Thanks, great!

I think we can merge this as-is, although I have some suggestions:

* I think we should explicitly test `range(1, sth)`, since the typical `fh` start with `1`, not with `0`.
* the coercions can probably be simplified - first coerce to iterable; second coerce to `pd.Index`
* I think if none of the cases applies, we should delegate to the coercion of `pd.Index` rather than raising an error. Catch the exception and say, it should be one of the types already listed, or `pd.Index` coercible.
    * that way, we do not have to explicitly maintain most coercions but can defer to `pandas`.

fkiraly:
Also, question: this would also allow passing of `range`-type `fh` in `fit` and `predict`, right?

If so:

* title of the PR should reflect that
* maybe an integration test with `fit` etc is worth carrying out? (not per estimator but overall)

.all-contributorsrc
yarnabrina: Wasn't aware, will add while updating the test as you suggested below.

yarnabrina:
> I think we should explicitly test range(1, sth), since the typical fh start with 1, not with 0.

Updated `test_standard_range_in_fh` locally from `range(5)` to `range(1, 5 + 1)`. Will push.

> * the coercions can probably be simplified - first coerce to iterable; second coerce to `pd.Index`
> 
> * I think if none of the cases applies, we should delegate to the coercion of `pd.Index` rather than raising an error. Catch the exception and say, it should be one of the types already listed, or `pd.Index` coercible.
>   
>   * that way, we do not have to explicitly maintain most coercions but can defer to `pandas`.

Agreed, but may be out of scope (based on title) for this PR? Can we track in another issue? This one may need to more thorough testing to see no existing functionality breaks.

>  question: this would also allow passing of range-type fh in fit and predict, right?

Yes, and verified locally. Skipped for now as I thought that's just a natural consequence of creating `ForecastingHorizon` successfully. Are there situations where it may not be the case?

In that case, can you please suggest how to test it? Are you suggesting to add tests in `TestAllForecasters`? Or, will adding one/two more tests in `test_fh.py` with random choice of forecasters be enough?

fkiraly:
> Agreed, but may be out of scope (based on title) for this PR? Can we track in another issue? This one may need to more thorough testing to see no existing functionality breaks.

Of course - I have approved, so that usually means you decide whether to add here or track separately.

fkiraly:
> In that case, can you please suggest how to test it? Are you suggesting to add tests in TestAllForecasters? Or, will adding one/two more tests in test_fh.py with random choice of forecasters be enough?

I would add it in `test_base` (in `forecasting.base`) which tends to have these kind of integration tests.

fkiraly (dismissed):
(still approve)

fkiraly (approved):

Issue #4569, opened 2023-05-10
[MNT] umbrella issue: actions at python 3.7 EOL (June 2023)


yarnabrina:
Python 3.7 reaches end of life on 27th June, 2023. (Ref. https://endoflife.date/python)

This issue is to track actions to carry out at and after python 3.7 EOL.

* there are some changes in the codebase to support 3.7, and not necessary for later versions. After EOL is reached, the changes can be reverted.
* some changes are blocked by incompatibility with 3.7. These can be merged after 3.7 EOL

0. deprecation: announce moving away from python 3.7 support in next release, for 0.20.0; remove 3.7 in 0.20.0 https://github.com/sktime/sktime/pull/4717
1. pinned package versions (e.g. #4566)
   * revert https://github.com/sktime/sktime/pull/4567
2. PR blocked by 3.7 specific failures
   * https://github.com/sktime/sktime/pull/4463
4. updated type hints (e.g. `list` instead of `typing.List`, support of `Literal`/`TypedDict`, etc.)
5. new coding styles (e.g. assignment expressions with walrus operator, positional/keyword only parameters, etc.)

(Please update the list with more tasks as required.)

Note: a majority of update of coding styles can be done using [`pyupgrade`](https://github.com/asottile/pyupgrade), with flag `--py38-plus`.

fkiraly:
@yarnabrina, I added #4463 to the list, and a clarification that the outstanding item from #4566 is the revert of #4567

PR #4717, opened 2023-06-17
[MNT] 0.20.0 release action - remove python 3.7 support

fkiraly:
Scheduled release action for 0.20.0 - removing python 3.7 due to end of life. See https://github.com/sktime/sktime/issues/4569

Removed from:

* package metadata, support
* estimator metadata
* CI and tests
* documentation

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -164,7 +164,7 @@ jobs:
     strategy:
       fail-fast: false  # to not fail all combinations if just one fail
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
+        python-version: ['3.8', '3.9', '3.10', '3.11']
 
     steps:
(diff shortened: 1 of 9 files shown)


yarnabrina:
Do you want me to apply the 4th point in #4569? In this PR or in a separate one?

> new coding styles (e.g. assignment expressions with walrus operator, positional/keyword only parameters, etc.)

Using `pyupgrade`, either once running locally or through new `pre-commit` hook will do this hopefully. walrus, updated type hint formats, etc. should behandled. Not sure if you want positional/keywork only behaviour, but that has to be manual I think.

I expect a big amount of changes, one quick run shows changes in 800+ files, and most are with respect to the `# -*- coding: utf-8 -*-` comment present in almost all files (ref. https://github.com/asottile/pyupgrade/issues/89). Please let me know whether/how to proceed.

PR #4720, opened 2023-06-19
[MNT] 0.20.0 release action - support for numpy 1.25

jorenham:
https://numpy.org/doc/stable/release/1.25.0-notes.html


fkiraly:
Thanks! If passes, let's schedula as a release action for 0.20.0

jorenham:
Fails on the (unsupported) python 3.7 versions

fkiraly:
yes, sktime will also drop 3.7 support in 0.20, so that's fine

PR #4722, opened 2023-06-19
[MNT] remove python 3.7 tests from CI

fkiraly:
Removes testing on python 3.7 on CI.

Part of https://github.com/sktime/sktime/pull/4717, but in case 3.7 tests start failing due to dependencies.


fkiraly:
merging this already since the mac 3.7 has started failing due to dependencies - otherwise that failure will block release.

PR #4725, opened 2023-06-20
[MNT] 0.20.0 release action - remove initial utf comments in all python modules which are unnecessary in python 3

yarnabrina:
All python modules in `sktime` were starting with the comment `# -*- coding: utf-8 -*-`. However, this is unnecessary in Python 3. Please refer to https://github.com/asottile/pyupgrade/issues/89 for related discussions.

This PR adds the argument to the existing hook to revert the behaviour (force removal instead of force presence), as suggested in the [hooks documentation](https://github.com/pre-commit/pre-commit-hooks#fix-encoding-pragma).

> To remove the coding pragma pass --remove (useful in a python3-only codebase)

Notes:

* This PR should not affect the CI tests, as it's just removal of comments.
* This PR is not necessarily part of `3.7` deprecation, even though it is discovered while doing that.


fkiraly (approved):
Agreed - let's do all of this in one batch at 0.20.0

PR #4726, opened 2023-06-20
[MNT] 0.20.0 release action - upgrade to coding style of python 3.8 and above using `pyupgrade`

yarnabrina:
Part of #4569.
Depends on #4722.
Depends on #4725.

This PR adds a new `pre-commit` hook [`pyupgrade`](https://github.com/asottile/pyupgrade). As an effect of the configured argument `--py38-plus`, all python codes in the `sktime` repository which have updated/newer syntax in newer python versions (`3.8` and above) are modified automatically.

Notes:

1. Apart from adding the new hook, this PR contains no manual change.
2. Most of the changes are coming from `super` statements. In pytthon 3, `super(class_name, self)` is same and recommended over `super()` inside same class definition, and most of `sktime` codebases were not using this form.


yarnabrina:
> to avoid bad surprises at merge, have you checked whether this does not conflict with https://github.com/sktime/sktime/pull/4722?

I copied the changes from #4722 and committed locally. Without these, I thought `pyupgrade` may make changes which will fail in `<3.8` versions. Does same changes by different committer cause conflicts? I don't know, didn't check.

fkiraly (approved):
Makes sense - I assume we can use this later, say, when we upgrade to min 3.9 etc.

PR #4738, opened 2023-06-21
[ENH] estimator crafter aka deserialization of estimator spec from string

fkiraly:
This PR adds a blueprint deserialization aka "crafter" for composite estimators, from string - the function `craft`.

`craft` uses the registry to construct an estimator from a string - interpreted as serialized object or crafting spec - **without the necessity for imports** (as long as the imports are within the scope of `all_estimators`).

Together with a full frozen requirements file, this should make the object constructed unique.

Together with `str` printing, of an object, the `craft` function forms a serialization/deserialization pair.

More precisely: for an in-memory estimator, string print will always result in a valid spec. Namely,

```python
spec = str(my_est)
new_est = craft(spec)
```

has the same effect as `new_est = clone(spec)`.

### Use cases

I think there are a number of compelling use cases:

* serialization/deserialization of estimator blueprints (as opposed to fitted estimators), in a deployment scenario
* use in benchmarking, e.g., to define reference estimators or standard choices - FYI @tcuongd, @ngupta23, @hazrulakmal
* use for quick string specification in estimator construction, see, e.g., https://github.com/sktime/sktime/pull/4712
* use in reproducible science, as a reference to define the precise estimator used (together with env version stamps)

Noting that some of these may be improved by also having an alias system, i.e., exposing any of the estimator blueprints by a shorter alias. The alias system may be similar to what @achieveordie is building in #4712.

As none of the above is time series specific, we may want to move this to `scikit-base` once user tested.

#### Example usage 1: simple parametric object
```python
from sktime.registry import craft

f = craft("VAR(trend='ct')")
```

creates an instance `f` of `VAR` with parameter `trend='ct'`.

#### Example usage 2: complex composite with variable assignments

more complex definitions can use assignments, the estimator itself needs to be returned by a `return` statement (as if this were a function body)

```python
pipe_spec = """
    ("imputer", Imputer()),
    ("forecaster", NaiveForecaster())])
cv = ExpandingWindowSplitter(
    initial_window=24,
    step_length=12,
    fh=[1, 2, 3])

# this estimator/object is being defined
return ForecastingGridSearchCV(
    forecaster=pipe,
    param_grid=[{
        "forecaster": [NaiveForecaster(sp=12)],
        "forecaster__strategy": ["drift", "last", "mean"],
    },
    {
        "imputer__method": ["mean", "drift"],
        "forecaster": [ThetaForecaster(sp=12)],
    },
    {
        "imputer__method": ["mean", "median"],
        "forecaster": [ExponentialSmoothing(sp=12)],
        "forecaster__trend": ["add", "mul"],
    },
    ],
    cv=cv,
    n_jobs=-1)
"""

pipe = craft(pipe_spec)
```

does what one would expect it should.

#### Example usage 3: with dunders

we can use dunders and any composition syntax available in `sktime`, e.g.,:

```python
dunder_spec = "Detrender(ExponentialSmoothing(sp=12)) * ARIMA()"
f = craft(dunder_spec)
```


fkiraly:
update: example 2 does not work because of the assignments in it. Thinking about whether this can be salvaged, otherwise there are going to be lots of nested terms.

fkiraly:
suggestion from discussion with @ctfallon and @mgazian000: there should be also an `imports` utility or similar, which prints out all the import statements for a spec

PR #4741, opened 2023-06-22
[ENH] replace load of UEA deprecated `Yoga` dataset from `AA_datatypes_and_datasets.ipynb`

fkiraly:
It seems that UEA have deprecated or removed the `Yoga` dataset from their data repository (without forwarning?)

This dataset was only used in the data format and loading tutorial in `AA_datatypes_and_datasets.ipynb` as an arbitrary choice.
It has now been replaced by another arbitrary choice which still exists.


fkiraly:
failure is on main, already, so merging this.

PR #4749, opened 2023-06-23
[ENH] loading of `sktime` datasets from GitHub LFS buffer mirror instead of direct upstream at UEA

fkiraly:
This PR changes the primary download location of datasets distributed with `sktime` from the (recently unreliable) UEA datasets repository to a mirror of it, the `sktime-datasets` repository, which is using GitHub LFS.

The `sktime-datasets` repository acts as a downstream buffer from the UEA datasets repository, and prevents that abrupt changes to file paths, file contents, etc reach the user instantly (and surprisingly).

Mid-term, we can set up cron CI around `sktime-datasets` to check integrity and consistency with the upstream.

### DIFF
--- a/sktime/datasets/_data_io.py
+++ b/sktime/datasets/_data_io.py
@@ -193,9 +193,8 @@ def _load_dataset(name, split, return_X_y, return_type=None, extract_path=None):
         ):
             # Dataset is not already present in the datasets directory provided.
             # If it is not there, download and install it.
-            url = (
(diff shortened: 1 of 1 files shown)


fkiraly:
(this seems to work but requires a GitHub LFS storage plan, as the free storage space is only 3GB - in any case, this is ready to go in case we need it)

fkiraly:
superseded by https://github.com/sktime/sktime/pull/5260 which uses a mirror priority system

PR #4758, opened 2023-06-24
[BUG] allows probabilistic predictions in `DynamicFactor` in presence of exogenous variables

### LINKED ISSUE #4744: [BUG] `DynamicFactor` cannot predict quantiles in presence of exogenous variables
**Describe the bug**

`DynamicFactor` has `"ignores-exogeneous-X": False` and `"capability:pred_int": True`, but fails to predict quantiles when `X` is present.

**To Reproduce**

```python
_, df = load_longley()

y = df[["GNPDEFL", "GNP"]]
X = df[["UNEMP", "POP"]]

from sktime.forecasting.dynamic_factor import DynamicFactor
forecaster = DynamicFactor().fit(y[:10], X=X[:10])

forecaster.predict_quantiles(fh=[1, 2, 3, 4, 5, 6], X=X[10:])
```

**Expected behavior**

No error. Currently it fails with this:

> ValueError: Out-of-sample operations in a model with a regression component require additional exogenous values via the `exog` argument.

**Additional context**

This is likely failing due to some issue in check of `exog`. Since `predict` works by checking on `self._forecaster`, failure in `_predict_interval` can be because of using `self._fitted_forecaster`.

**Versions**
<details>

```pycon
>>> from sktime import show_versions; show_versions()

System:
    python: 3.9.16 (main, Mar  8 2023, 14:00:05)  [GCC 11.2.0]
executable: /path/to/venv/bin/python
   machine: Linux-5.15.90.1-microsoft-standard-WSL2-x86_64-with-glibc2.35

Python dependencies:
          pip: 23.1.2
       sktime: 0.19.2
      sklearn: 1.2.2
       skbase: 0.4.0
        numpy: 1.24.3
        scipy: 1.10.1
       pandas: 1.5.3
   matplotlib: 3.7.1
       joblib: 1.2.0
  statsmodels: 0.13.5
        numba: 0.57.1
     pmdarima: 2.0.3
      tsfresh: 0.20.0
   tensorflow: 2.12.0
tensorflow_probability: 0.19.0
>>> 
```

</details>

yarnabrina:
Closes #4744.

Main Changes

* passes exogeneous variables always (`None` causes no problem) for prediction
* adds tests

Extra Changes

* fix type hinting

### DIFF
--- a/sktime/forecasting/dynamic_factor.py
+++ b/sktime/forecasting/dynamic_factor.py
@@ -1,5 +1,7 @@
 # copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
 """Implements DynamicFactor Model as interface to statsmodels."""
+from typing import List, Union
+
 import numpy as np
 import pandas as pd
 
@@ -219,7 +221,7 @@ class DynamicFactor(_StatsModelsAdapter):
             )
         return y_pred.loc[fh.to_absolute_index(self.cutoff)]
 
-    def _predict_interval(self, fh, X=None, coverage: [float] = None):
+    def _predict_interval(self, fh, X=None, coverage: Union[float, List[float]] = None):
         """Compute/return prediction quantiles for a forecast.
 
         private _predict_interval containing the core logic,
--- a/sktime/forecasting/dynamic_factor.py
+++ b/sktime/forecasting/dynamic_factor.py
@@ -1,14 +1,11 @@
 # copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
 """Implements DynamicFactor Model as interface to statsmodels."""
-
-import inspect
-
 import numpy as np
 import pandas as pd
 
 from sktime.forecasting.base.adapters import _StatsModelsAdapter
 
-_all_ = ["DynamicFactor"]
+__all__ = ["DynamicFactor"]
 __author__ = ["Ris-Bali", "lbventura"]
 
 
--- a/sktime/forecasting/dynamic_factor.py
+++ b/sktime/forecasting/dynamic_factor.py
@@ -211,10 +211,8 @@ class DynamicFactor(_StatsModelsAdapter):
         # beginning of the training series when passing integers
         start, end = fh.to_absolute_int(self._y.index[0], self.cutoff)[[0, -1]]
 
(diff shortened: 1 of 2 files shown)


fkiraly:
Hm, looks good - but can you explain why this was non captured by the automatic contract tests (e.g., via `check_estimator`)?

This may hint at a gap in the test coverage, not testing exogeneous variables could be consequential.

yarnabrina:
I am confused, are you asking why this happens? I thought it's because it's considering only `TestAllForecasters` for forecasters, and indivual `test_<algorithm>.py` is not supported in that file.

```pycon
>>> check_estimator(DynamicFactor, tests_to_run="test_DynamicFactor_without_exogenous_variables")
All tests PASSED!
{}
```

Is this not expected and/or is it not by design?

(I always test with `pytest` CLI directly, i.e. `pytest -k "<algorithm> and <test>" sktime/forecasting/<module_path>`, and don't really use `check_estimator` myself.)

fkiraly:
sorry, I should have been more specific.

I meant, why is the failure not certified by a test in `TestAllForecasters` already? That is, why did the standard tests in there not catch the bug?

yarnabrina:
As mentioned here (https://github.com/sktime/sktime/issues/4744#issuecomment-1605384672), I'm not very clear on this. But as far as I can see, estimators are trained with exogenous data only in two tests: `test_predict_time_index_with_X` and `test_hierarchical_with_exogeneous`. And for both of these, only `predict` is tested and not other probabilistic versions.

Unless I am missing something, this probably means exogenous probabilistic predictions are untested for all estimators as part of `TestAllForecasters`?

fkiraly:
yes, I think you are right. I went through the tests and see now:

* the tests are not using scenarios but hard-coded data
* exogenous data are not tested in `test_predict_time_index_with_X`, as you say

I think the reason is, when we updated the logic in `test_predict_time_index` to not use try/except and also cover probabilistic methods, it was missed in `test_predict_time_index_with_X`.

The other question is, should these two tests be pulled together and using scenarios (or maybe not as the hassle might be bigger than the benefit)

fkiraly:
A quick win would be to turn `test_predict_time_index_with_X` into a logic copy of `test_predict_time_index` except with `X`.

PR #4759, opened 2023-06-24
[ENH] Implementing SplitterSummarizer Transformer

### LINKED ISSUE #4664: [ENH] `SplitterSummarizer` which applies a series-to-primitive transformer per split
From discussion with @abishalom on discord.

I infer that it would be great to have a "splitter summarizer", which is a series-to-series transformer.

It takes as a parameter a series-to-primitives transformer `t` and a splitter `cv`.

It transforms a single series `X` into a single series `Xt`, such that: the `i`-th row of `Xt` is `t.fit_transform(cv.split_y(X)[i][0])` (i.e., the `t`-transform of the `i`-th train split of `X`), where all series are `pd.DataFrame` (otherwise equivalent index). The `fit` is empty.

Optionally, there is a parameter that determines the index of `Xt`, it could be the integer index, or the last index of the train split.

FYI @abishalom - if you pipeline `SplitterSummarizer(PCA(), cv=ExpandingWindowSplitter())` with `ARIMA`, it should be what you want?

That is, the forecaster you want (if I understand right?) is

`SplitterSummarizer(PCA(), cv=ExpandingWindowSplitter()) * ARIMA()`

BensHamza:
#### Reference Issues/PRs

Fixes #4664 .

This PR introduces a new series-to-series transformer SplitterSummarizer.

No

#### Any other comments?

The testing isn't working 

##### For all contributions

##### For new estimators


yarnabrina:
Thanks for your contribution.

> The testing isn't working

Unfortunately, your code style is a bit different the rest of the codebase, and this is being checked by the configured hooks in `.pre-commit-config.yaml` file.

What you can do are the following:

1. install `pre-commit` python package from in your local environment (`pip install pre-commit`)
2. install configured git hooks (`pre-commit install`)
3. run configured git hooks manually once now for all files (`pre-commit run --all-files`)
4. stage all of the changes (`git add .`)
5. commit the staged changes (`git commit -m "code quality changes"`)
6. push the committed changes (`git push`)

Please review the changes before you stage. Ideally, they should match what you see [here](https://github.com/sktime/sktime/actions/runs/5365227837/jobs/9734028999?pr=4759).

Also, feel free to use any commit message of your choice.

For your next commits, all the configured git hooks will run automatically for the staged files while you run `git commit` in future. You may read up more about `pre-commit` [here](https://pre-commit.com/).

fkiraly:
do you still need help on this, or input, @BensHamza?

BensHamza:
> do you still need help on this, or input, @BensHamza?

Sorry for the time it takes and thank you for the follow-up. I needed some time to get used to the codebase.

I had first a problem understanding how to make the pipeline mentionned in #4664 work since `PCA()` is a series-to-series transformer and we first wanted a series-to-primitives transformer as param. From what I understood in discord, we want SplitterSummarizer to work in a general context ( eg.  `SplitterSummarizer(SummaryTransformer(),ExpandingWindowSplitter())` ). 

This last code version works fine on this last example, Can you please tell me if it is the right guess? Also, there are surely some improvements to this version. I would appreciate any feedback. Thank you !

fkiraly:
> I had first a problem understanding how to make the pipeline mentionned in #4664 work since `PCA()` is a series-to-series transformer and we first wanted a series-to-primitives transformer as param.

That's an excellent point!

I made a mistake here, I was thinking of a (hypothetical) PCA transformer that produces PCA loadings probably (a primitive/row vector), rather than the actual PCA transforme that is currently in the code base (this produces scores, column vectors).

> From what I understood in discord, we want SplitterSummarizer to work in a general context ( eg. `SplitterSummarizer(SummaryTransformer(),ExpandingWindowSplitter()) )`.

Yes, you understand correctly imo!

sktime/transformations/series/summarize.py
```diff
@@ -750,3 +750,140 @@ def get_test_params(cls, parameter_set="default"):
         params4 = {"summary_function": None, "quantiles": (0.1, 0.2, 0.25)}
 
         return [params1, params2, params3, params4]
+
+
+class SplitterSummarizer(BaseTransformer):
+    """Create summary values of a time series' splits.
+
+    A series-to-series transformer that applies a series-to-primitives transformer
```
fkiraly, line 758: Good description - I would add a little more detail, e.g., what are the resulting indices, and what is in the i-th row, etc.

sktime/transformations/series/summarize.py
```diff
@@ -750,3 +750,140 @@ def get_test_params(cls, parameter_set="default"):
...
+
+    It splits the series using a specified splitter and transforms each resulting
+    split using the provided transformer.
+
+    This transformer aims to provide a summarization of the series based on the
+    transformer and splitter given.
+
+    Attributes
+    ----------
+    transformer : series-to-primitives transformer used to convert series
```
fkiraly, line 769: in the docstring style we use, the first row would precisely specify the type (`sktime` transformer inheriting from `BaseTransformer`), the second and subsequent rows explain what it is and how it impacts the outputs of the methpds.

sktime/transformations/series/summarize.py
```diff
@@ -750,3 +750,140 @@ def get_test_params(cls, parameter_set="default"):
...
+        super().__init__()
+
+    def _transform(self, X, y=None):
+        """Transform X and return a transformed version.
+
+        Transforms the series `X` into a new series `Xt` by applying the
+        series-to-primitives transformer `t` on each train split of `X`,
+         which is created using the splitter `cv`.
+
+         The `i-th` row of `Xt` is equivalent to
```
fkiraly, line 826: This should be in the docstring of the *class*, this is good!

The docstrings of the private methods are not displayed in the API docs so not visible to the user.

sktime/transformations/series/summarize.py
```diff
@@ -750,3 +750,140 @@ def get_test_params(cls, parameter_set="default"):
...
+        "scitype:transform-output": "Series",
+        # what scitype is returned: Primitives, Series, Panel
+        "scitype:instancewise": True,  # is this an instance-wise transform?
+        "X_inner_mtype": ["pd.DataFrame", "pd.Series"],
+        # which mtypes do _fit/_predict support for X?
+        "y_inner_mtype": "None",  # which mtypes do _fit/_predict support for X?
+        "fit_is_empty": True,
+    }
+
+    def __init__(self, transformer, splitter, index="last"):
```
fkiraly, line 812: does the `index` arg to anything? Not mentioned in docstring

sktime/transformations/series/summarize.py
```diff
@@ -750,3 +750,140 @@ def get_test_params(cls, parameter_set="default"):
...
+            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+            `create_test_instance` uses the first (or only) dictionary in `params`
+        """
+        from sktime.forecasting.model_selection import ExpandingWindowSplitter
+
+        params1 = {
+            "transformer": SummaryTransformer(),
+            "splitter": ExpandingWindowSplitter(),
+        }
+
```
fkiraly, line 888: Generally, it's a good idea to have at least two test examples.

You should try a different setting for `index`, and maybe another combination of transformer and splitter.

A good idea is to cover transformers that need to be fitted..

fkiraly (dismissed):
@BensHamza, looks great! This is exactly what we were looking for!

Let's see whether the tests run through, nice contribution!

I left some comments above. We could merge as is, but it would be great to straighten out the docstrings and clarify the `index` parameter.

Additional features we can leave for later, e.g., an option to remember data seen in `fit` and use the union of data seen in `fit` / `transform` - that might be important for forecasting use cases.

sktime/transformations/series/summarize.py
fkiraly, line 812: maybe `ExpandingWindowSplitter` is the most sensible default for `splitter`?

If we want this default, then the `__init__` default should be `None`, and the default is written to `self._splitter`.

BensHamza: I have added `index` in class docstring

sktime/transformations/series/summarize.py
BensHamza: I moved it to class' docstring

sktime/transformations/series/summarize.py
BensHamza: I guess it is fixed now

sktime/transformations/series/summarize.py
BensHamza: Added

sktime/transformations/series/summarize.py
BensHamza: Made changes regarding splitter's default object with  `ExpandingWindowSplitter` . I am thinking maybe `SlidingWindowSplitter(window_length=10, step_length=10, fh=1, start_with_window=True)` would be more appropriate for a default splitter here ?

fkiraly: Hm, let's think carefully in terms of user expectations. A good default has no "surprising" choices, i.e., it agrees with the user guess if you don't tell them what the choice is.

* I agree that sliding window is more in line with user expectation than expanding.
* however, sliding has the problem of having to specify window and step length.
* I disagree with `step_length=10` because it is an arbitrary choice, and default user expectation imo is `step_length=1`.
* `window_length` is also arbitrary, but consistent with the default of `SlidingWindowSplitter`. Perhaps not overriding the default is the best choice here.

fkiraly:
Re test failures, this is due to the test parameters using the catch22 transformer, which has some soft dependencies (`numba` I think). Optimally, the test parameters do not rely on additional packages - so you could replace it with the `SummaryTransformer`.

fkiraly (dismissed):
Looks great!

Would approve if all tests pass.

sktime/transformations/series/summarize.py
fkiraly: yes, of course - this is feature creep, so a potential separate project.

fkiraly (approved):
Really great, good to go!

I've made two minor changes to expedite this for the release:

* added the estimator to the API reference, so users can find it easily
* fixed the docstring - for multi-line cells, use `... ` and not `>>> `

(see the latest commits for this)

sktime/transformations/series/summarize.py
fkiraly: I think you did not change this?

fkiraly:
FYI, I think you did not change the default from `ExpandingWindowSplitter` to `SlidingWindowSplitter` as you suggested?

BensHamza:
> I've made two minor changes to expedite this for the release

Thank you @fkiraly for these changes and the explanations.

> FYI, I think you did not change the default from ExpandingWindowSplitter to SlidingWindowSplitter as you suggested?

My bad. I did the changes.

I was having a problem passing the `test-nosoftdeps-full` , I didn't get where it fails.

fkiraly:
> I was having a problem passing the test-nosoftdeps-full , I didn't get where it fails.

sure - let's check where/if it still fails.

fkiraly:
FYI, I shortened the windows in some test cases in `get_test_params`, as the test data is short and longer window length will raise errors.

BensHamza:
Thank you @fkiraly , I learned a lot in the process.

fkiraly:
well, @BensHamza, hopefully this is just the start of a long and fruitful collaboration!

PR #4760, opened 2023-06-24
[MNT] Temporarily skip all DL Estimators

achieveordie:
#### Reference Issues/PRs

In reference to #4610, skipping all DL estimators that are causing several CI runs to crash till we find a root cause and hence a solution to this.

### DIFF
--- a/sktime/tests/_config.py
+++ b/sktime/tests/_config.py
@@ -37,6 +37,13 @@ EXCLUDE_ESTIMATORS = [
     "SimpleRNNClassifier",
     "SimpleRNNRegressor",
     "EditDist",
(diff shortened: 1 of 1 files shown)


sktime/tests/_config.py
```diff
@@ -37,6 +37,13 @@
     "SimpleRNNClassifier",
     "SimpleRNNRegressor",
     "EditDist",
```
fkiraly, line 39: I notice, EditDist is not a DL estimator - do we want to remove it from the list or write a separate comment?

fkiraly (approved):
Yes, that's probably the way to go until we find out separately what is causing the hangs.

Issue #4743, opened 2023-06-22
[BUG] `VECM` can not predict quantiles for Longley dataset


yarnabrina:
**Describe the bug**

`VECM` has `"capability:pred_int": True`, and is being tested in CI, but does not seem to work for Longley dataset.

**To Reproduce**

```python
from sktime.datasets import load_longley
_, y = load_longley()

from sktime.forecasting.vecm import VECM
forecaster = VECM().fit(y)

forecaster.predict_quantiles(fh=[1, 2, 3])
```

**Expected behavior**

No error. Currently fails with this:

> TypeError: Input must be list-like

**Additional context**

It is most likely failing as `var_names` is not being created in the correct format. I'm not clear why multivariate model needs to check the `index` instead of using columns always.

**Versions**
<details>

```pycon
>>> from sktime import show_versions; show_versions()

System:
    python: 3.9.16 (main, Mar  8 2023, 14:00:05)  [GCC 11.2.0]
executable: /path/to/venv/bin/python
   machine: Linux-5.15.90.1-microsoft-standard-WSL2-x86_64-with-glibc2.35

Python dependencies:
          pip: 23.1.2
       sktime: 0.19.2
      sklearn: 1.2.2
       skbase: 0.4.0
        numpy: 1.24.3
        scipy: 1.10.1
       pandas: 1.5.3
   matplotlib: 3.7.1
       joblib: 1.2.0
  statsmodels: 0.13.5
        numba: 0.57.1
     pmdarima: 2.0.3
      tsfresh: 0.20.0
   tensorflow: 2.12.0
tensorflow_probability: 0.19.0
>>> 
```

</details>

fkiraly:
confirmed on current `main`, windows, python 3.10

PR #4763, opened 2023-06-25
[ENH] add tests for predict_interval, predict_quantiles

yarnabrina:
This PR locates all tests of `predict` in `sktime/forecasting/tests/test_all_forecasters.py`, and whenever it makes sense, it adds checks for `predict_interval` and `predict_quantiles`.

Depends on #4758.
See also https://github.com/sktime/sktime/pull/4758#issuecomment-1605783891.


yarnabrina:
@fkiraly This is not only for the exogenous variables as we were discussing in #4758, added all tests that I felt missing. If you feel we should restrict this only to that, please let me know what to edit.

fkiraly (approved):
Excellent, good thinking!

(seeing this, I wonder whether there is a more structured way of doing this that does not incur a lot more `fit` calls, but that´s another topic)

I think this indeed catches some other things as well!

I do not see any problem with the tests, but of course we should not merge failing tests to main.

Previously, when adding coverage to suite tests and that would detect a batch of bugs, I would do one of the following two:

1. report & except

* report all bugs in issue tracker or link/add them to existing issues
* add exceptions to the tests
* deal with the issues later (and they have a note that the exception should be reverted in the fix)

2. fix immediately, if the bugs look easy:

* deal with them now in separate PR
* make this PR depend on those PR and ensure the tests pass
* then merge the PR in sequence of dependency

yarnabrina:
Opting for "report and except".

All unique failing tests:

- hierarchical_with_exogeneous - VECM - TypeError: Input must be list-like
- hierarchical_with_exogeneous - DynamicFactor - ValueError: Out-of-sample operations in a model with a regression component require additional exogenous values via the `exog` argument.
- predict_time_index_with_X - SquaringResiduals - KeyError: 0
- predict_time_index_with_X - DynamicFactor - ValueError: Out-of-sample operations in a model with a regression component require additional exogenous values via the `exog` argument.
- predict_time_index_in_sample_full - ARIMA - AssertionError
- predict_time_index_in_sample_full - DynamicFactor - ValueError: Prediction must have `end` after `start`.

Tracking Issues:

- DynamicFactor - ValueError: Prediction must have `end` after `start`. - #4744
- VECM - TypeError: Input must be list-like - #4743
- Rest - #4765 - without reproducible examples as of now

fkiraly:
I see, so there are two known ones and three new ones 👍

yarnabrina:
@fkiraly Almost all (except 2 mentioned below) are coming from `scipy`, which should be solved by #4770 and are out of scope (not disabling transformer tests in forecasting PR). Requesting review.

---

@achieveordie `test-windows3.8` fails in two classifier tests, and the names of the tests look related to serialisation. The traceback also shows that few test workers crashed with fatal exceptions, which may or may not be related to our memory error related issues. Can you please take a look?

CI job link: https://github.com/sktime/sktime/actions/runs/5376135078/jobs/9752982340?pr=4763
Test Failure Summary:

> FAILED sktime/tests/test_all_estimators.py::TestAllEstimators::test_save_estimators_to_file[InceptionTimeClassifier-2-ClassifierFitPredict-predict_proba]
> FAILED sktime/tests/test_all_estimators.py::TestAllEstimators::test_persistence_via_pickle[InceptionTimeClassifier-0-ClassifierFitPredictMultivariate-predict]

fkiraly (approved):
Yes, happy with this! Thanks!

fkiraly:
(failures resolved either by being sporadic or by being fixed on `main`)

Issue #4769, opened 2023-06-25
[BUG] `BoxCoxTransformer` fails with `scipy.optimize._optimize.BracketError` since `scipy 1.11.0`


fkiraly:
`scipy 1.11.0` was released today, this is leading to failures along the following line:

`scipy.optimize._optimize.BracketError: The algorithm terminated without finding a valid bracket. Consider trying different initial points`

This impacts the `BoxCoxTransformer`, amongst others.

Minimal reproducing example:

```python
from sktime.transformations.series.boxcox import BoxCoxTransformer
from sktime.utils._testing.panel import _make_panel_X

X = _make_panel_X(
    n_instances=7, n_columns=1, n_timepoints=10, random_state=42
)

est = BoxCoxTransformer()

est.fit(X)
```

Condition: `scipy 1.11.0` and any recent `sktime` version, e.g., 0.20.0

fkiraly:
Update: this happens only for negative values, which should not be passed to `BoxCoxTransformer` (at least the naive version).

The deeper reason is that some test cases (for interface conformance) were using dummy data with negative values, which led to a failure of the `scipy` optimizers on all previous versions of `scipy`, but this was masked by a bug in `scipy` that would prevent this from being flagged by warning or exception.

This is therefore not a bug and shouldn't impact sensible usage, but still should be dealt with.

PR #4770, opened 2023-06-25
[BUG] fix `BoxCoxTransformer` failure after `scipy` 1.11.0

fkiraly:
Fixes https://github.com/sktime/sktime/issues/4769 by using `scipy`'s on-board box-cox-fitter for methods `"mle"` and `"pearsonr"` and enforcing positive values.

The reason for the failure was use of negative values in some interface test cases, which previously would produce garbage results but `scipy` would not complain. As of 1.11.0, it complains (albeit with a confusing error message that makes it hard to track down that it is in fact from negative values in some test cases). See further discussion here: https://github.com/scipy/scipy/issues/18761

This is now solved by introducing an `enforce_positive` argument to `BoxCoxTransformer`, which ensures positive values before fitting. This can be turned off to obtain the previous behaviour, but for positive values it will be the same (although with an additional `np.sign` and `np.abs` calculation that should be fast).

Also makes the following changes, while we're cleaning this up:

* The `sktime` box-cox-fitter in `BoxCoxTransformer` was not using `scipy`'s since the latter did not have settable bounds when the former was originally implemented. I have hence replaced it with `scipy`'s own box-cox-fitter and using a bounded optimizer if bounds are passed, from `scipy` 1.7.0 on where it exists in the form we need. Pre-1.7.0, I've left the old behaviour.
* improved docstring - general improvements, removed some wrong math statements, made clear what the choice of method means and when it interfaces `scipy`
* added a `"fixed"` method, which allows to use the transformer with a fixed parameter - or tune it via grid search.
* input checks on `method`

Note: the fix is behaviour changing, but imo does not require deprecation because it changes some behaviour from buggy/misleading (only in the sub-case where negative values were passed) to sensible.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,7 +50,7 @@ dependencies = [
     "packaging",
     "scikit-base<0.6.0",
     "scikit-learn>=0.24.0,<1.3.0",
-    "scipy<1.11.0,>=1.2.0",
+    "scipy<2.0.0,>=1.2.0",
 ]
 
 [project.optional-dependencies]
--- a/sktime/transformations/series/boxcox.py
+++ b/sktime/transformations/series/boxcox.py
@@ -263,9 +263,9 @@ class BoxCoxTransformer(BaseTransformer):
         params1 = {"method": "pearsonr"}
         params2 = {"method": "mle"}
         params3 = {"method": "guerrero", "sp": 2}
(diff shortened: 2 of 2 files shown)


fkiraly:
update: just learnt that an initial bracket need not contain bounds - so I simply set the bracket to bounds if provided, while passing bounds if provided also to the optimizer (constrained)

PR #4771, opened 2023-06-26
[MNT] bound `scipy<1.11.0` after bracket failures

fkiraly:
This bounds `scipy` away from the new release, given `BracketError` failures #4769, which may or may not be due to a bug in `scipy` - in any case they seem unclear and hard to fix.

We may also want to consider making the defensive bounds in `scipy` more granular in the future.


fkiraly:
merging due to the failure on `main`

Issue #4585, opened 2023-05-13
[ENH] Scikit-Learn 1.3 Support


celestinoxp:
**Additional context**
Should prepare support for Scikit-Learn 1.3.

** How to reproduce erros **
Install scikit-learn dev: pip install --pre --extra-index https://pypi.anaconda.org/scipy-wheels-nightly/simple scikit-learn
Try use some code with sktime

celestinoxp:
More Details: https://github.com/pycaret/pycaret/issues/3540

fkiraly:
great idea, @celestinoxp!

Would you like to open a diagnostic PR that installs the preview version? I.e., changing this in `.github/workflows/test.yml`.

(we can also do this, but we always encourage new contributors to join, first!)

celestinoxp:
> great idea, @celestinoxp!
> 
> Would you like to open a diagnostic PR that installs the preview version? I.e., changing this in `.github/workflows/test.yml` and marking it with "do not merge".
> 
> (we can also do this, but we always encourage new contributors to join, first!)

I apologize, but my knowledge is limited and I don't know how to do it...
I just gave the idea, because I think that in the future, when the final version of scikit-learn 1.3 comes out, sktime already supports it, instead of having a delay to make the necessary changes...
Of course, dev versions shouldn't be put into production, so I thought that sktime could also have a dev version, or it's probably not necessary, as it can be installed via pip install https://github.com/sktime /sktime/archive/refs/heads/main.zip

I don't know if you agree... if you don't, please ignore this comment...

fkiraly:
ok, I've tried to set up a diagnostic PR here:
https://github.com/sktime/sktime/pull/4611

Feel free to add yourself with an `ideas` contributor badge to the `sktime` contributor file, @celestinoxp - great idea!

fkiraly:
There is a weird error re optimization failure, `test_panel_in_panel_out_not_supported_but_series - scipy.optimize._optimize.BracketError: The algorithm terminated without finding a valid bracket.`

any ideas?

celestinoxp:
File ~\anaconda3\lib\site-packages\sktime\forecasting\compose\__init__.py:48
     41 from sktime.forecasting.compose._multiplexer import MultiplexForecaster
     42 from sktime.forecasting.compose._pipeline import (
     43     ForecastingPipeline,
     44     ForecastX,
     45     Permute,
     46     TransformedTargetForecaster,
     47 )
[...]

PR #4778, opened 2023-06-27
[MNT] 0.21.0 release action - update `sklearn` bound to `<1.4.0`

fkiraly:
Fixes https://github.com/sktime/sktime/issues/4585

According to pre-release tests, `sktime` is now compatible with `sklearn 1.3.0`: https://github.com/sktime/sktime/pull/4611

We should therefore increase the bound at the next MINOR update.

Depends on https://github.com/sktime/sktime/pull/4860 for fixing remaining issues.


fkiraly:
Remaining failure in `MatrixProfileClassifier` is due to a bug in `sklearn` (`KNeighborsClassifier`) that seems to be fixed in the next release 1.3.1: https://github.com/scikit-learn/scikit-learn/issues/26768

Issue #4434, opened 2023-04-04
[DOC] Mismatch in the method default and method documentation


yarnabrina:
#### Describe the issue linked to the documentation

There is a minor typo in documentation of `_predict_interval` for `ForecastX`.

The method definition has default of `0.90`:

https://github.com/sktime/sktime/blob/d9a038f9f259637aee5180bf7b0b420c93300e91/sktime/forecasting/compose/_pipeline.py#L1401

while docstring says `0.95`:

https://github.com/sktime/sktime/blob/d9a038f9f259637aee5180bf7b0b420c93300e91/sktime/forecasting/compose/_pipeline.py#L1414

The same mismatch is there for `BaseForecaster` as well:

https://github.com/sktime/sktime/blob/d9a038f9f259637aee5180bf7b0b420c93300e91/sktime/forecasting/base/_base.py#L1984

https://github.com/sktime/sktime/blob/d9a038f9f259637aee5180bf7b0b420c93300e91/sktime/forecasting/base/_base.py#L1997

#### Suggest a potential alternative/fix

We can keep `0.95` everywhere, given it's most popular and `DEFAULT_ALPHA = 0.05` already.

fkiraly:
Ah, I see!

This is very subtle, since:

* this concerns private methods which are not shown in the docs (I think)
* the private default is *always* overridden by the value from the public method, `predict_interval`, which is 0.90

There's also the principle of "don't change public facing logic without deprecation", because code in production could start breaking at the next release otherwise.
So, when there's a clash between docs and code, in a case that could be public facing, it's preferable to change the code as an immediate measure. Of course, if the logic also needs change, it may require deprecation, i.e., announcements and warnings towards the user that this piece of logic will change.

Another subtle point being, even if it's in a private method, it could still impact public facing logic. In this case, it actually doesn't matter what the default is, but I would argue it should be consistent with the public method default (0.90) to avoid confusion.

fkiraly:
Looking at it, a lot of these docstrings are wrong in subtle ways! Including `_predict_quantiles`, in many estimators, including the base class.

It's mostly the private ones so minimal danger to the user, but it still shouldn't be that way.
Bulk-replace (with a bit of caution) should do the trick

PR #4780, opened 2023-06-27
[ENH] replace `"Coverage"` and `"Quantiles"` default variable name in univariate case with variable name

### LINKED ISSUE #4709: [BUG] Column names `predict_interval` output inconsistent between univariate and multivariate case
**Describe the bug**

When calling `predict_interval` on a pd.Series or pd.Dataframe with only one column, the column name in the result is `Coverage` and not the original column's name

**To Reproduce**

Using a series
```python
y, X = load_longley()
pipe = Differencer() * ExponentTransformer() ** SARIMAX()
pipe.predict_interval(X=X_test)
```

Output: 

```
          Coverage              
               0.9
             lower         upper
1959  66024.326892  68465.109904
1960  66001.646041  70923.894181
1961  65438.989794  72844.058934
1962  66857.619049  76745.554752
```

Using a dataframe with two columns:

```python
y, X = load_longley()
y = pd.concat([y] * 2, ignore_index=True, axis=1)
pipe = Differencer() * ExponentTransformer() ** SARIMAX()
pipe.predict_interval(X=X_test)
```
Output
```
                   0                           1              
                 0.9                         0.9              
               lower         upper         lower         upper
Period                                                        
1959    66024.714431  68466.243863  66024.714431  68466.243863
1960    66002.607821  70926.326488  66002.607821  70926.326488
1961    65440.810411  72848.072287  65440.810411  72848.072287
1962    66860.203253  76751.053796  66860.203253  76751.053796

```

**Expected behavior**
The forecaster should reuse the same column also in the single column case. However, I am unsure if this is a bug or expected behavior.

fkiraly:
Fixes #4709 - in `predict_interval` and `predict_quantiles` return, replaces `"Coverage"` and `"Quantiles"` default variable name in the univariate case with the variable name, to make this consistent with the multivariate case. If no variable name is available, the variable name used is the integer 0.

This is done with a deprecation horizon or two MINOR cyles, by introducing a `legacy_interface` argument to public and private `predict_interval`, `predict_quantiles`. A value `True` (default) gives current behaviour, `False` gives future behaviour. In 0.22.0, we switch the default to `False`, and remove the argument in 0.23.0.

This includes tests for `legacy_interface` both `True` and `False`, with deprecation notes on removing this later.


fkiraly:
FYI @benheid

sktime/forecasting/base/_base.py
```diff
@@ -2305,6 +2347,27 @@ def _predict_moving_cutoff(
...
+        valid only in the univariate case. Can be extended later.
+        """
+        if legacy_interface:
+            var_name = default
+        else:
+            y = self._y
+            if isinstance(y, pd.Series):
+                var_name = self._y.name
+            elif isinstance(y, pd.DataFrame):
+                return y.columns
```
benHeid, line 2363: Personally, I prefer if functions have only one return or if each if-else clause has its own return.  Would it be possible to do something like that:

``python`
if isinstance(y, pd.Series):
  return [y.name] if y.name is not None else [0] # <- Probably if y.name is not None could also be part of the outer if case
elif isinstance(y, pd.DataFrame):
  return y.columns
else:
 return [0]

```

File: sktime/forecasting/base/_base.py [hunk 2]
```diff
@@ -2305,6 +2347,27 @@ def _predict_moving_cutoff(
...
+        """Return variable column for DataFrame-like returns.
+
+        Developer note: currently a helper for predict_interval, predict_quantiles,
+        valid only in the univariate case. Can be extended later.
+        """
+        if legacy_interface:
+            var_name = default
+        else:
+            y = self._y
+            if isinstance(y, pd.Series):
```
benHeid, line 2360: I personally prefer if a function has only one return value or each if-else clause has its own. Perhaps, we can change it as follows:

```python
if isinstance(y, pd.Series):
  return [y.name] if y.name is not None else [0] # <- if y.name is not None could probably also moved to the outer if clause.
elif isinstance(y, pd.DataFrame):
  return y.columns
return [0] # Or raise an exception, unsure if this case should ever happen. However, probably linting requires that line.
```

sktime/forecasting/base/adapters/_statsmodels.py
```diff
@@ -147,7 +147,9 @@ def _extract_conf_int(prediction_results, alpha) -> pd.DataFrame:
 
         raise NotImplementedError("abstract method")
 
-    def _predict_interval(self, fh, X=None, coverage=0.95):
+    # todo 0.22.0 - switch legacy_interface default to False
+    # todo 0.23.0 - remove legacy_interface arg and logic using it
+    def _predict_interval(self, fh, X=None, coverage=0.95, legacy_interface=True):
```
benHeid, line 152: Side comment, not blocking. I just realised that depending on the method, the default value of the coverage differs. It is a bit inconsistent but I am not sure if we should create an issue on that...

benHeid (changes requested):
In the description of the PR there is a TODO left about tests for the legacy interface. I could not find these tests. So I suppose this TODO is still open?

# TODOs
* [x] Tests
* [ ] Utility function design (get_var_names)

fkiraly:
Ah, yes, you are right!

* tests for both legacy and new
* looking at this, there should be a warning too (deprecation) that is raised when a single-variable frame is returned.

sktime/forecasting/base/_base.py
fkiraly: yes, that seems better style

sktime/forecasting/base/adapters/_statsmodels.py
fkiraly: is that so? Can you give two examples?

Either way, it should never reach the user, as the `predict_interval` default overrides any `_predict_interval` default.

But it still would be best to homogenize this, to avoid confusion.

benHeid: I think `stats_models` is the exception. This has the default value 0.95, however, the other forecasters adapted by this PR has 0.9 (e.g. `fbpropet`)

fkiraly: FYI @yarnabrina, due to `statsmodels`

yarnabrina: If I remember correctly, we discussed previously on this. I kept the same default as it was in the base docstring while creating the statsmodels adapter wrapper, which differs with the function signature.

Ref. https://github.com/sktime/sktime/issues/4434

fkiraly: Maybe it is safest if the private methods do *not* have a default at all? Because the public method must pass an argument to all private methods, and defaults should be global.

So, to avoid any accidents, it is perhaps best to remove all defaults from all private methods?

fkiraly: I opened a design issue for this:
https://github.com/sktime/sktime/issues/4809

I would consider this out of scope for this PR, but we should deal with it. That should also resolve #4434 ultimately.

fkiraly:
ready for re-review, @benheid - tests and warning added.

fkiraly:
Quick question, under todos, what did you mean with "utility function design"? Was this addressed in one of the comments?

sktime/forecasting/base/_base.py
```diff
@@ -2302,6 +2408,27 @@ def _predict_moving_cutoff(
...
+                var_name = self._y.name
+            elif isinstance(y, pd.DataFrame):
+                return y.columns
+            else:
+                var_name = 0
+            if var_name is None:
```
benHeid, line 2427: This if-clause is unnecessary because it it should be either a pd.Series, pd.Dataframe or 0. Or do I miss something?

benHeid (changes requested):
Thank you for the revison, I have only a short question regarding one if statement.

benHeid:
> Quick question, under todos, what did you mean with "utility function design"? Was this addressed in one of the comments?

Yes this is solved :)

sktime/forecasting/base/_base.py
fkiraly: if can happen that it's a `y: pd.Series` with `y.name` being `None`, that case may lead to pain.

benHeid: Ok thank you :) I approve this PR then.

benHeid (approved):

PR #4781, opened 2023-06-27
Set the matplotlib backend to agg to avoid that a GUI is triggered.

### LINKED ISSUE #2066: [BUG] Sporadic test failure in `test_plotting`, of `_tkinter.TclError` on Windows, Python 3.7 CI/CD

benHeid:
Set the backend of matplotlib to 'agg' to hopefully deactivate the tkinter/gui...
#### Reference Issues/PRs
fixes #2066

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/utils/tests/test_plotting.py
+++ b/sktime/utils/tests/test_plotting.py
@@ -48,14 +48,14 @@ def valid_data_types():
     return valid_data_types
 
 
-# todo: remove skip when issue #2066 has been fixed
-@pytest.mark.skip(reason="sporadic failure on win CI/CD, see issue 2066")
 @pytest.mark.parametrize("series_to_plot", series_to_test)
 def test_plot_series_runs_without_error(series_to_plot):
     """Test whether plot_series runs without error."""
     _check_soft_dependencies("matplotlib")
+    import matplotlib
     import matplotlib.pyplot as plt
 
+    matplotlib.use("agg")
     _plot_series(series_to_plot)
     plt.gcf().canvas.draw_idle()
 
@@ -69,8 +69,6 @@ def test_plot_series_runs_without_error(series_to_plot):
     plt.close()
 
 
-# todo: remove skip when issue #2066 has been fixed
-@pytest.mark.skip(reason="sporadic failure on win CI/CD, see issue 2066")
 @pytest.mark.parametrize("series_to_plot", invalid_input_types)
 def test_plot_series_invalid_input_type_raises_error(series_to_plot, valid_data_types):
     """Tests whether plot_series raises error for invalid input types."""
@@ -88,8 +86,6 @@ def test_plot_series_invalid_input_type_raises_error(series_to_plot, valid_data_
             _plot_series(series_to_plot)
 
 
-# todo: remove skip when issue #2066 has been fixed
-@pytest.mark.skip(reason="sporadic failure on win CI/CD, see issue 2066")
 @pytest.mark.parametrize(
     "series_to_plot", [(y_airline_true, y_airline_test.reset_index(drop=True))]
 )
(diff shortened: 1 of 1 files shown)


fkiraly (changes requested):
hm, but matplotlib is also used in `_plot_series`, which can potentially lead to crashes, no?

benHeid:
> hm, but matplotlib is also used in `_plot_series`, which can potentially lead to crashes, no?

Matplotlib uses global variables as far as I know. Thus, if in plot_series no other backend is set, it should still be agg.

fkiraly:
what I mean, there are tests such as `test_plot_series_invalid_marker_kwarg_len_raises_error` where the variable is not set prior to going into `_plot_series` - the different tests could be distributed across different `pytest-xdist` workers, an executed in any sequence.

fkiraly:
We should probably retest a couple of times to see that the original error does not sporadically reoccur.

Regarding the failure of `test-nosoftdeps-full`, you need to conditionally skip the test if the soft dependencies are not installed - search for `pytest.skipif` for examples of this.

fkiraly:
yes, that's how it works - except you forgot some of the tests

fkiraly:
>  Furthermore, I assumed that these tests are about checking if the correct exception is raised they do not call show or something like that and thus do not trigger tkinter.

You might be right - I remember back then I skipped everything in the file since it was like a game of whack-a-mole - first this test was failing and then that one.

fkiraly (approved):
Hm, seems to work!

PR #4782, opened 2023-06-27
[MNT] clean-up of `CODEOWNERS`

fkiraly:
This cleans up the `CODEOWNERS` file as follows:

* orders `CODEOWNERS` by alphabet
* sets the general owners to the currently four core developers with daily/weekly contribution frequency - @achieveordie, @benHeid , @fkiraly, @yarnabrina - this does not have a formal meaning of ownership, it only means we receive review requests
* removes codeowners of base and some general framework building modules - these are "commonly owned" and not an individual's
* rectifies ownership where it was missing or incorrect
    * adds missing `DateTimeTransformer` (@danbartl, @KishManani)
    * adds missing `ClearSkyTransformer` (@ciaran-g)
    * adds missing `HierarchyEnsembleForecaster` (@VyomkeshVyas)
    * added missing owners in the annotation module - @miraep8, @katiebuc

There are probably others, I wonder whether it's worth updating it here - notifications will in general not be sent to individuals without write access. So, should owners (if applicable) not rather be attached via a tag instead?

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,10 +1,7 @@
 # The file lists sktime's algorithm maintainers as specified in GOVERNANCE.md.
 # Each line is a file pattern followed by one or more owners.
 
-* @fkiraly @aiwalter @guzalbulatova
-
-
-sktime/regression/dummy/ @badrmarani
+* @achieveordie @benheid @fkiraly @yarnabrina
 
 sktime/classification/dictionary_based/_boss.py @patrickzib @MatthewMiddlehurst @TonyBagnall
 sktime/classification/dictionary_based/_cboss.py @patrickzib @MatthewMiddlehurst @TonyBagnall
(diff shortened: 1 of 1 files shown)


yarnabrina (commented):
This is not a review per se, just noting that Github severely complains of the `CODEOWNERS` file as it is currently. I guess it needs all the owners to have write access.

fkiraly:
Yes, correct, @yarnabrina, indeed the problem is with the write access - which clashes with the current governance model where you need to be a core developer or subject to an exception policy.
Afaik there is also no way to grant the write access by file.

That's why I am suggesting, perhaps move away from `CODEOWNERS` as a register of ownsership (where applicable) and use the tag system instead, to tag an estimator with owners (where applicable) and authors?
The historical `CODEOWNERS` was quite messy anyway, and the intended workflow was not really followed properly anyway.

Besides, the key practical consequence is that this PR notifies the most active current core devs.

yarnabrina:
Makes sense 👍🏽 

To confirm my understanding, is this what you'll do in this PR?

1. Only L4 will remain in `CODEOWNERS`, which will be in sync with active core developers
2. For all concrete public estimators, there will be a tag (say "authors") with value being the original contributor
3. For all other base classes and/or modules, `__author__` will acknowledge all contributors

---

> using tags for owners of estimators would make it easier to write a CI element that checks whether a change is made that would require ownerapproval

I'm really curious how to achieve this. How can we detect programmatically (with git/shell) if contents of a class got modified?

fkiraly:
> Only L4 will remain in CODEOWNERS, which will be in sync with active core developers

Yes, or "very active", or aligned with a special role, tbd.

> For all concrete public estimators, there will be a tag (say "authors") with value being the original contributor

Yes, or two tags - one for authors, one for owner (which might be different?)

> I'm really curious how to achieve this. How can we detect programmatically (with git/shell) if contents of a class got modified?

Here's how I'd do it:

1. get files modified via `git log`
2. use `all_estimators` to find all estimators in those files
3. use `get_tags` to retrieve the owners
4. add them as reviewers to the PR

PR #4784, opened 2023-06-28
[ENH] in `CNNNetwork`, add options to control `padding` and `filter_size` logic

alan191006:
#### Reference Issues/PRs
No PR associated.

Change padding and filter_size logic at sktime/networks/cnn.py
* Remove hardcoded the 60 threshold for padding type with a dynamic one based on input size and kernel size.
* Change the if-elif chain for filter_sizes a one-liner using python's extend function.

No

Code correctness if there are any.

No additional test

#### Any other comments?
None

##### For all contributions

##### For new estimators


fkiraly (changes requested):
Thanks!

May I kindly ask for a quick explanation what this fixes or changes?

The previous code looks odd, but I'm a bit at a loss when trying to understand how this got fixed.

Perhaps some quick comparison before/after with an explainer? That would be much appreciated!

fkiraly:
FYI @achieveordie

alan191006:
Hi @fkiraly,

Sorry for my late reply, upon revising my changes, my padding logic don't make sense to me... The previous threshold is to deal with Italy Power On-demand dataset. I think we should let the user decided which type of padding to use instead of we decide for them and the default will be `valid`.

As for the second change, I tried to just improve the readability a little with python's `extend` function instead of manual concatenating array.

I will commit another changes addressing the first issue is that alright?

alan191006:
Hi @fkiraly,

Thank you for your response. I have added a new commit that changes the padding to a parameter and includes a docstring. Additionally, I have modified the logic for filter_sizes as the previous implementation did not work as expected.

fkiraly (changes requested):
Thanks! Looks great!

Can you kindly add more details to the docstring what the new parameter does?
E.g., what are the possible values for `padding`, and how do they impact the NN structure?

alan191006:
Hi @fkiraly,

I have already made the requested change. Could you please take a look and let me know your thoughts?

sktime/networks/cnn.py
```diff
@@ -44,10 +51,12 @@ def __init__(
         avg_pool_size=3,
         n_conv_layers=2,
         activation="sigmoid",
+        padding="valid",
```
achieveordie, line 54: An interesting point regarding the implementation is that the authors of this paper have not mentioned anything about padding ([link](https://ieeexplore.ieee.org/stamp/stamp.jsp?arnumber=7870510) to the paper, Ctrl + F to search for pad, padding etc.). "valid" seems to be the choice of the authors of the code.

The hard-coded "same" might be to pass the CI tests since this architecture might result in negative dims for smaller data sizes. We could face the same problem which makes me question what should our default value be.

"valid" seems like the more apt choice (this is also the default value of the underlying Conv1D) but this would mean passing "same" into test configs (class method called `get_test_params()`) of `CNNClassifier` and `CNNRegressor`.

I'd love your and @fkiraly's opinion on it.

sktime/networks/cnn.py
```diff
@@ -22,6 +22,13 @@ class CNNNetwork(BaseDeepNetwork):
     filter_sizes    : array of int, shape = (nb_conv_layers)
     activation      : string, default = sigmoid
         keras activation function
+    padding         : string, default = "valid"
+        The padding type for the convolutional layers.
+
+        - 'valid': No padding is applied. The input is not padded, and the output size is reduced.
+        - 'same': The input is padded with zeros so that the output size is the same as the input size.
+
+        Note: This parameter determines how the input is padded or not padded during the convolution operation. It affects the output size of the convolution layers.
```
achieveordie, line 31: This is probably too verbose, isn't it? One of the additional values a user can have is "causal" (see the padding argument section in [Conv1D](https://www.tensorflow.org/api_docs/python/tf/keras/layers/Conv1D) for reference).

We'd probably be better off mentioning that we allow all values as allowed by TensorFlow's `Conv1D`, so we won't be required to update this docstring if they choose to add/remove values.

achieveordie (changes requested):
@alan191006 Thanks for bringing this up! I had this on my mental to-do for quite a while, I'm glad you went ahead to rectify it.

achieveordie:
We'd also need to temporarily enable testing of `CNNClassifier` and `CNNRegressor`. (You can do this by removing the names of these estimators from [`_config.py`](https://github.com/sktime/sktime/blob/main/sktime/tests/_config.py#L40)).

We are excluding all DL estimators owing to sporadic errors in the CI, see #4610 for reference.

sktime/networks/cnn.py
alan191006: Hi @achieveordie,

I wanted to share my thoughts on the default value for padding. Firstly, thank you for providing the link to the paper. After considering the options, I believe using `valid` padding as the default choice may not be ideal. This is because, as you mentioned, it can lead to negative dimensions and potentially leaving out important edge information in certain cases.

On the other hand, `same` padding appears to be a more rational choice. Although it may increase resources consumption and complexity, it ensures that the dimensions remain consistent, avoiding the issue of negative dimensions and preserving information at the edges.

However, I must admit that my expertise in CNNs for time series prediction is limited. I would greatly appreciate hearing your and @fkiraly thoughts on this matter.

sktime/networks/cnn.py
alan191006: Thank you for the respond, I forgot the `casual` is even there. Anyway, I will add it and make the whole thing more concise.

fkiraly: I think you haven't added it, I updated the dostring.

fkiraly:
I updated the docstring and the logic so the behaviour does not change in the current default setting.

fkiraly (approved):
@alan191006, thanks!

I've made some minor changes to ensure we need no deprecation, and this is ready for the release.

I notice that the parameters are now in `CNNNetwork`, but they are not yet addressable from `CNNClassifier` and `CNNRegressor`. I suggest to add that in a separate PR.

alan191006:
Hi @fkiraly ,
Sorry for not responding, I have quite a lot of exam these day. I appreciate your efforts in handling the PR changes and merges.

fkiraly:
And thank you for your effort in making this PR in the first place!

Issue #4785, opened 2023-06-28
[BUG] ColumnTransformer passthrough doesn't always work 


davidgilbertson:
**Describe the bug**
When I have a `ColumnTransformer` with two transforms operating on the same column (one is `Id()` one is `WindowSummarizer`) then `ColumnTransformer` forgets to pass through the remainder of the columns.

**To Reproduce**
This code works as expected:
```python
import numpy as np
import pandas as pd
from sktime.transformations.compose import Id
from sktime.transformations.panel.compose import ColumnTransformer
from sktime.transformations.series.summarize import WindowSummarizer

df = pd.DataFrame(
    dict(
        A=np.random.normal(size=5),
        B=np.random.normal(size=5),
        C=np.random.normal(size=5),
    )
)

transformer= ColumnTransformer(
    [
        ("col1", Id(), ["A"]),
        (
            "col2",
            WindowSummarizer(
                lag_feature={"lag": [1, 2]},
                truncate="bfill",
                n_jobs=1,
            ),
            ["B"],
        ),
    ],
    remainder="passthrough",
)

print(transformer.fit_transform(df))

```

And outputs this (note column C is there):
```txt
        A  B_lag_1  B_lag_2       C
0 -0.1583  -1.5775  -1.5775 -0.6509
1 -0.4888  -1.5775  -1.5775 -0.9251
2  1.6321  -0.7883  -1.5775  0.1432
3 -0.4012  -1.1564  -0.7883  0.9755
4  1.0094   0.4105  -1.1564 -0.5998
```

But what I really want is to keep A, _and_ get some lags. If I change the `WindowSummarizer` to operate on column `A`:
```py
import numpy as np
import pandas as pd
from sktime.transformations.compose import Id
from sktime.transformations.panel.compose import ColumnTransformer
from sktime.transformations.series.summarize import WindowSummarizer

df = pd.DataFrame(
    dict(
        A=np.random.normal(size=5),
        B=np.random.normal(size=5),
        C=np.random.normal(size=5),
    )
)

transformer= ColumnTransformer(
    [
        ("col1", Id(), ["A"]),
        (
            "col2",
            WindowSummarizer(
                lag_feature={"lag": [1, 2]},
                truncate="bfill",
                n_jobs=1,
            ),
            ["A"],
        ),
    ],
    remainder="passthrough",
)

print(transformer.fit_transform(df))

```

Now the output doesn't have C any more:
```txt
[...]

PR #4789, opened 2023-06-28
[ENH] `ColumnEnsembleTransformer` - bugfix and `remainder` argument

fkiraly:
This fixes `ColumnEnsembleTransformer` which looks like a merge accident upon an incomplete implementation.
The buggy estimator was not detected as it was not added to the module export, so the CI did not run its tests.

This now fixes the `ColumnEnsembleTransformer`:

* it should work now as intended
* adds a `remainder` argument to `ColumnEnsembleTransformer`, similar to `ColumnTransformer` of `sklearn`. This replicates functionality available in the semi-deprecated `ColumnTransformer` to allow a move to the `sktime` native one.
* adds a new `feature_names_out` parameter which allows to control the naming of the output columns (if `pandas` based)

Related: this bug https://github.com/sktime/sktime/issues/4785

Depends on https://github.com/sktime/sktime/pull/4793


sktime/transformations/compose/_column.py
```diff
@@ -145,6 +185,42 @@ def _transformers(self, value):
...
+        # handle remainder
+        remainder = self.remainder
+
+        if remainder == "passthrough" or isinstance(remainder, BaseTransformer):
+            if isinstance(remainder, BaseTransformer):
+                rem_t = self.remainder.clone()
+            elif remainder == "passthrough":
+                from sktime.transformations.compose import Id
+
+                rem_t = Id()
```
yarnabrina, line 212: To reduce the code complexity, here's a non-blocking suggestion:

1. After L204, check if remainder is `drop` or not. If yes, return.
2. Rest of the logic can go unindented skipping L206 check.

yarnabrina (commented):
Looks okay to me, left one non-blocking suggestion. Requesting @davidgilbertson / @benheid to take a look as well.

Question: if there are multiple parallel (in the sense of being independent in DAG or at least lead to different leaves) transformers, will `feature_names_out` give an option to identify the source? I didn't get a chance to try this branch myself yet, so not sure.

sktime/transformations/compose/_column.py
fkiraly: yes, changed to that.

fkiraly:
> Question: if there are multiple parallel (in the sense of being independent in DAG or at least lead to different leaves) transformers, will feature_names_out give an option to identify the source? I didn't get a chance to try this branch myself yet, so not sure.

As long as they live in the same `ColumnEnsembleTransformer`, the names will be unique. The only case where they could be non-unique is the `"original"` setting, in which case an error would be raised (from `rename_columns` utility)

sktime/transformations/compose/_column.py
yarnabrina: Personally I'm in favour of positive checks when possible, and what I meant is this:

```python
if remainder == "drop" or remainder is None:
```

But this is equivalent I suppose (please let me know if that's not the case), so I've no further comments.

```python
if remainder != "passthrough" and not isinstance(remainder, BaseTransformer):
```

fkiraly: welllll, my reasoning in favour of the negative check is to maximize the scope of the default behaviour in case of an input check failure.

Suppose in some future scenario where the input check in `__init__` gets removed or breaks accidentally.

Then the default `"drop"` happens only for a positive condition that you have to specifically hit.

Whereas, currently, the default happens whenever you do not hit the narrow negative condition, which makes the default more robust to accidents or changes.

fkiraly: (but, assuming the input check in `__init__` has passed and nothing modifies `self.remainder`, the two conditions are afaik indeed equivalent)

Issue #3895, opened 2022-12-06
[BUG] Installation of pycatch22 ends with exception for Python 3.8 and 3.9


aiwalter:
**Describe the bug**
`pycatch22` makes issues on Python 3.8, Python 3.7 works well.
```
Python interpreter will be restarted.
Collecting sktime[all_extras]==0.14.1
  Using cached sktime-0.14.1-py3-none-any.whl (15.9 MB)
Requirement already satisfied: numba>=0.55 in /databricks/python3/lib/python3.8/site-packages (from sktime[all_extras]==0.14.1) (0.55.1)
Requirement already satisfied: pandas<1.6.0,>=1.1.0 in /databricks/python3/lib/python3.8/site-packages (from sktime[all_extras]==0.14.1) (1.2.4)
Collecting deprecated>=1.2.13
  Using cached Deprecated-1.2.13-py2.py3-none-any.whl (9.6 kB)
Requirement already satisfied: statsmodels>=0.12.1 in /databricks/python3/lib/python3.8/site-packages (from sktime[all_extras]==0.14.1) (0.12.2)
Requirement already satisfied: scipy<2.0.0 in /databricks/python3/lib/python3.8/site-packages (from sktime[all_extras]==0.14.1) (1.6.2)
Requirement already satisfied: scikit-learn<1.2.0,>=0.24.0 in /databricks/python3/lib/python3.8/site-packages (from sktime[all_extras]==0.14.1) (0.24.1)
Collecting numpy<1.23,>=1.21.0
  Using cached numpy-1.22.4-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (16.9 MB)
Collecting gluonts>=0.9.0
  Using cached gluonts-0.11.4-py3-none-any.whl (1.0 MB)
Collecting pycatch22
  Using cached pycatch22-0.4.2.tar.gz (49 kB)
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Getting requirements to build wheel: finished with status 'done'
  Installing backend dependencies: started
  Installing backend dependencies: finished with status 'done'
    Preparing wheel metadata: started
    Preparing wheel metadata: finished with status 'done'
  WARNING: Generating metadata for package pycatch22 produced metadata for project name unknown. Fix your #egg=pycatch22 fragments.
[...]

PR #4790, opened 2023-06-28
[MNT] add back `pycatch22` as a soft dependency post python 3.7 EOL

fkiraly:
Reverts the temporary fix https://github.com/sktime/sktime/pull/3917/files which removed `pycatch22` as a soft dependency (and prevented dependent estimator tests to run).

Potentially fixes https://github.com/sktime/sktime/issues/3895 if the install is without problems now, post-3.7 EOL.


fkiraly:
Note to self: we need to check whether `pycatch` implies `pandas < 2` or not. Only if not we can merge as-is.

Issue #4701, opened 2023-06-13
[ENH]Improve Feature Importances Attribute For TimeSeriesForest


JonathanBechtel:
**Is your feature request related to a problem? Please describe.**
The current implementation of the TimeSeriesForest returns feature importances for each time interval for each summary statistic, but it is unclear what each feature importance refers to.  As such, it's all but impossible to quickly discern what inputs are influencing prediction unless you look at the source code. There is also a discrepancy with regards to how the feature is presented in the original research paper and how it's displayed in the library.  

The original paper for the TimeSeriesForest introduces the algorithm as using "temporal curves", which displays how entropy gain for each summary statistic changes throughout the time series.  That is, there are three feature importances, one for each summary statistic over time.  You can see the formula in page 10 here:  https://arxiv.org/pdf/1302.2277.pdf.   However the current implementation uses the standard feature importance that's inherited from SKLearn, and displays the values in a purely cross-sectional format.

**Describe the solution you'd like**
I think the TimeSeriesForest implementations should have two separate attributes not currently included:

 - feature_labels_, which simply provides a reference for what each of the values in feature_importances_ refers to, and would have the format `mean_0`, `mean_1`, `mean_2`, `std_0`, `std_1`, `slope_0`, `slope_1`, and so on.  
 - temporal_curves_, which would be truer to the original idea of the algorithm.  This would be a dictionary of 3 arrays, each one with the feature importance for each summary statistic over each interval organized temporally.  This would allow people to quickly see how changes in the summary statistics of the data influenced the prediction, and would be true to the original intent of the technique.  

In the original paper it's mentioned that temporal curves are one of the primary features of the technique to make it easier to diagnose where problems might occur in a process, and there's not a feasible way to do that now.  The original paper suggested using raw entropy gain at each time interval for the calculation, however I don't see why you wouldn't be able to do something similar from the values returned by `feature_importances_` after fitting.  These values would be scaled, whereas using raw Entropy would be unscaled, but otherwise the information contained therein would be the same.  I actually see it being more user friendly to see scaled data since the y_axis would always refer to the same thing across all datasets.

**Additional context**
There was additional discussion about this issue in this pull request:  #3517, it mentions some of the confusion about the labeling that's associated with it, but I'm separating this into its own issue to make it more visible.  Can also start working on this with others relatively soon.

fkiraly:
FYI @YHallouard, have you seen this? and #4792 of @CTFallon?

YHallouard:
[...]

PR #4792, opened 2023-06-28
[ENH] Added temporal importance curve generation and plotting to BaseTimeSeriesForest

CTFallon:
#### Reference Issues/PRs
Resolves issues discussed in #4701.

Added two bits of functionality.
- in `BaseTimeSeriesForest` class (`series_as_features/base/estimators/interval_based/_tsf.py`), added `_temporal_curves` function which correctly calculates the importance of each feature for every timestamp. This new function is automatically called during `_fit` and creates four numpy.ndarray attributes to `BaseTimeSeriesForest`.
- adds function `plot_TSF_temporal_importance_curve` to `classification/plotting/temporal_importance_diagram.py` which takes a fitted estimator that inherits from `BaseTimeSeriesForest` and creates a plot of the three features' importance curves.

No.

We are fairly new to sktime, so overall program paradigm. To enumerate a few questions:
- The plotting function currently saves the plot, should it instead return the plot?
- For large forests, the `_temporal_curves` function might take a while: should it be removed from `_fit`, and perhaps changed into a public function that the user can call explicitly, and be called in the plotting function?
- Calculating the actual curves can probably be optimized, but that is not my strong suit.

No.

##### For all contributions

##### For new estimators

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -2197,7 +2197,8 @@
       "avatar_url": "https://avatars.githubusercontent.com/u/19725980?v=4",
       "profile": "https://github.com/CTFallon",
       "contributions": [
-        "doc"
+        "doc",
+        "code"
       ]
     },
     {
@@ -2206,7 +2207,8 @@
       "avatar_url": "https://avatars.githubusercontent.com/mgazian000",
       "profile": "https://github.com/mgazian000",
       "contributions": [
-        "doc"
+        "doc",
+        "code"
       ]
     }
   ]
--- a/sktime/classification/plotting/temporal_importance_diagram.py
+++ b/sktime/classification/plotting/temporal_importance_diagram.py
@@ -5,6 +5,9 @@ __author__ = ["MatthewMiddlehurst"]
 import numpy as np
 
 from sktime.classification.interval_based import CanonicalIntervalForest
+from sktime.series_as_features.base.estimators.interval_based import (
+    BaseTimeSeriesForest,
+)
 from sktime.transformations.panel import catch22
 from sktime.utils.validation._dependencies import _check_soft_dependencies
 
@@ -75,3 +78,25 @@ def plot_cif(cif, normalise_time_points=False, top_curves_shown=None, plot_mean=
         top_curves_shown=top_curves_shown,
         plot_mean=plot_mean,
     )
+
+
+def plot_TSF_temporal_importance_curve(tsf, file_name):
+    """Temporal Importance curve diagram generator for TSF."""
+    import matplotlib.pyplot as plt
+
+    if not isinstance(tsf, BaseTimeSeriesForest) or not tsf._is_fitted:
+        raise ValueError("Input must be a fitted object that inherits from BaseTSF")
+
+    curves = {
+        "Mean": tsf.mean_curve_,
+        "StDev": tsf.stdev_curve_,
+        "Slope": tsf.slope_curve_,
+    }
+
+    for curve_name, curve in curves.items():
+        plt.plot(curve, label=curve_name)
+
+    plt.legend()
+
+    plt.savefig(file_name)
+    plt.clf()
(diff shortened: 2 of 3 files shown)


fkiraly (changes requested):
Thanks, this is great addition that users will like!

> The plotting function currently saves the plot, should it instead return the plot?

Yes, I was just going to say - a good design separates business logic and data I/O in layers, and a lowest-level function shouldn't do both.

I would recommend that the function returns `fig, ax` objects, and optionally you can also pass them as args in which case the args will be modified (this is useful for tile plots and sub-plots).

> should it be removed from _fit, and perhaps changed into a public function that the user can call explicitly, and be called in the plotting function?

Yes, I would strongly recommend that. We do not have a unified interface design for plots - there aren't that many - so feel free to take any unreserved name.

> Calculating the actual curves can probably be optimized, but that is not my strong suit.

Before I help you there, I would strongly suggest you write a docstring for the function, in which you also reference the paper from which these feature importances are coming (I assume this is not original research but implements from some source?) This would help me understand the math and help you with speeding things up.

fkiraly:
Btw, if you are looking at the class `TimeSeriesForestClassifier`: it isn't really great how it inherits from multiple parents. It would be better if it doesn't do the `super().fit` etc, instead of inheritance it would have been better to compose in hindsight, see https://en.wikipedia.org/wiki/Composition_over_inheritance

Not for this PR, but FYI - in case that's something you can see how to do easily given that you have dived into the class.

CTFallon:
Thanks for the review. We'll get started on the first two points, and updating some docstrings

As for the third, we followed section 4.4 of https://arxiv.org/pdf/1302.2277.pdf, the original paper on TSFs. This hadn't yet been implemented in sktime.

The basic idea is to, for each timestamp, sum each feature's importance from each individual estimator within the forest. Right now we do so as three `for` loops (once over the (estimators, list of intervals) pairs, once over the individual intervals, and once over the timestamps within that interval). My guess would be the last loop is unnecessary, and that numpy has a function that can be used to add a value to a range of elements in an array. If not, we could figure out a way to do it with masks.

fkiraly:
Great! The test failure looks like a failure of the doctest.

By default, the examples in the docstrings are executed, checking whether the code actually runs.
The problem seems to be that an "import" line ends abruptly. I suppose you had problems with the character limit per line, so you have to find a way around it while ensuring that the code runs.

If you want statements to continue over multiple lines, you need to combine use of `...` (instead of `>>>`) for 2nd, 3rd line etc, and possibly line wrap characters if there is no way you can use brackets or similar, e.g., 
```python
>>> from long_stiff import (
...     this_gets_imported
....    )
```

fkiraly:
> the plotting function return the figure, instead of directly saving it to file.

On that note, I was suggesting the signature `plot_sth(args, ax=None)` ... `return ax)`. If `ax` is passed

Otherwise, you cannot use the plotting in a subplot.

(or am I mistaken? my `matplotlib` is a little rusty)

CTFallon:
@fkiraly That makes sense. I've used matplotlib in the past, but not super extensively. We've modified the plotting function so now it follows matplotlib's recommended call signature (i..e, requires ax to be passed, and returns it). This does mean there is more work to be done by the user, but I think the example we wrote is illustrative enough to help.

fkiraly (changes requested):
Great!

Re "more work to be done by the user" - what I tried to suggest was a compromise between now and your first implementation. Apologies if I have been giving confusing directions.

My suggestion was to make it an *option* to pass `ax`:

* if not passed, `fig`, `ax` objects are created and returned.
* if passed, the `ax` input is modified and returned.

Have a look at `plot_series` for an example of this behaviour.

This design allows you to ahve your cake and eat it - no work for the user if they don't want to, and you still can use the function in subplots.

fkiraly:
`torch` related failures are due to this, don't panic:
https://github.com/sktime/sktime/pull/4819
fixed and merged

fkiraly:
@CTFallon, apologies, I missed this - and it conflicts with some recent refactoring, and might intersect with @YHallouard's work.

@YHallouard, would you mind looking at this briefly and check re intersections and interfaces?

Issue #4794, opened 2023-06-29
[ENH] resolve `pandas` restriction to `<2` in CI for 3.8-3.10


fkiraly:
In the CI for 3.8-3.10, `pandas 1.5.3` is installed.

This is due to the `all_extras` together implying 1.5.3.

I would suggest to do the following:

* create a `pandas2_extras` dependency set, only for tests
* move the extras which are not `pandas 2` compatible to the `pandas1` dependency set
* do not change `all_extras`

The dependencies in question:

* `statsforecast` an its dependency `plotly-resampler`

PR #4795, opened 2023-06-29
[MNT] ensure CI for python 3.8-3.10 runs on `pandas 2`

fkiraly:
Fixes https://github.com/sktime/sktime/issues/4794

This is done by removal of packages conflicting with `pandas 2` from the dependency set that is tested.

The `all_extras` set does not change, instead a smaller `all_extras_pandas2` is tested.


fkiraly:
Works as expected - the python 3.8-3.10 CI now use `pandas 2.0.3`, while `statsforecast` is present and being tested on the `pandas1` CI elements.

PR #4805, opened 2023-07-01
[ENH][MNT] Interface StatsForecast MSTL

luca-miniati:
#### Reference Issues/PRs

Umbrella issue #4643

Interface for StatsForecast MSTL

Tests found in `get_test_params`, passed all tests in `check_estimator` 

##### For all contributions

##### For new estimators

### DIFF
--- a/docs/source/developer_guide/dependencies.rst
+++ b/docs/source/developer_guide/dependencies.rst
@@ -42,9 +42,8 @@ Estimators with a soft dependency need to ensure the following:
    or a ``list`` of ``str``, of import dependencies. Exceptions will automatically raised when constructing the estimator
    in an environment without the required packages.
 *  In a case where the package import differs from the package name, i.e., ``import package_string`` is different from
-   ``pip install different-package-string`` (usually the case for packages containing a dash in the name), the ``_check_soft_dependencies``
-   utility should be called in ``__init__`` before the ``super.__init__`` call, with ``severity="error"``.
-   The ``package_import_alias`` argument should be used to pass the information on package and import strings.
+   ``pip install different-package-string`` (usually the case for packages containing a dash in the name), the ``python_dependecies_alias`` tag
+   should be populated to pass the information on package and import strings as ``dict`` such as ``{"scikit-learn":"sklearn"}``.
 *  If the soft dependencies require specific python versions, the ``python_version``
    tag should also be populated, with a PEP 440 compliant version specification ``str`` such as ``"<3.10"`` or ``">3.6,~=3.8"``.
 *  If including docstring examples that use soft dependencies, ensure to skip doctest. To do this add a ``# doctest: +SKIP`` to the end of each
--- a/sktime/alignment/dtw_python.py
+++ b/sktime/alignment/dtw_python.py
@@ -51,6 +51,7 @@ class AlignerDTW(BaseAligner):
         "capability:distance": True,  # does compute/return overall distance?
         "capability:distance-matrix": True,  # does compute/return distance matrix?
         "python_dependencies": "dtw-python",
+        "python_dependecies_alias": {"dtw-python": "dtw"},
     }
 
     def __init__(
@@ -65,13 +66,6 @@ class AlignerDTW(BaseAligner):
         """Construct instance."""
         # added manually since dtw-python has an import alias
         # default check from super.__init__ does not allow aliases
-        _check_soft_dependencies(
-            "dtw-python",
-            package_import_alias={"dtw-python": "dtw"},
-            severity="error",
-            obj=self,
-            suppress_import_stdout=True,
-        )
         super().__init__()
 
         self.dist_method = dist_method
--- a/sktime/registry/_tags.py
+++ b/sktime/registry/_tags.py
@@ -372,6 +372,13 @@ ESTIMATOR_TAG_REGISTER = [
         "python dependencies of estimator as str or list of str",
     ),
     (
+        "python_dependencies_alias",
+        "estimator",
+        "dict",
+        "should be provided if import name differs from package name, \
+        key-value pairs are package name, import name",
+    ),
+    (
         "requires_cython",
         "estimator",
         "bool",
--- a/sktime/utils/validation/_dependencies.py
+++ b/sktime/utils/validation/_dependencies.py
@@ -378,10 +378,13 @@ def _check_estimator_deps(obj, msg=None, severity="error"):
     compatible = compatible and _check_python_version(obj, severity=severity)
 
     pkg_deps = obj.get_class_tag("python_dependencies", None)
+    pck_alias = obj.get_class_tag("python_dependencies_alias", None)
     if pkg_deps is not None and not isinstance(pkg_deps, list):
         pkg_deps = [pkg_deps]
     if pkg_deps is not None:
-        pkg_deps_ok = _check_soft_dependencies(*pkg_deps, severity=severity, obj=obj)
+        pkg_deps_ok = _check_soft_dependencies(
+            *pkg_deps, severity=severity, obj=obj, package_import_alias=pck_alias
+        )
         compatible = compatible and pkg_deps_ok
 
     return compatible


fkiraly (changes requested):
Great start!

Couple change requests:

* all parameters must be passed explicitly to an `sklearn` / `sktime` transformer, otherwise they will not be visible to tuning algorithms
* I would also make the individual parameters in `autoets_kwargs` available as parameters in `sktime`
* further, I think `MSTL` takes a forecaster internally - optimally, it should be possible to use any `sktime` forecaster here. For that, it would be useful to write a "back-adapter" to `statsforecast`. Not sure how difficult that is though
* not sure what to do with the `alias` argument - I think we don't really need it?

luca-miniati:
So for a back adapter, I would write a class that inherits from `_TS`, the `statsforecast` base class. Then, I update all the methods and attributes with the `sktime` estimator's interface?

fkiraly:
> Then, I update all the methods and attributes 

That's one possible way - not sure whether there is an easier way to pass an `sktime` forecaster to be used in `statsforecast` `MSTL`.

sktime/forecasting/statsforecast.py
```diff
@@ -628,14 +644,165 @@ def get_test_params(cls, parameter_set="default"):
...
+                    "Trend forecaster should not adjust seasonal models. Please pass "
+                    "`sp=1` to your trend forecaster"
+                )
+
+    def __repr__(self):
+        return "StatsForecastBackAdapterMSTL"
+
+    def fit(self, y, X=None):
+        self.estimator.fit(y=y, X=X)
+        self.in_sample_fh = -np.arange(len(y))
```
luca-miniati, line 717: Here I've set `in_sample_fh` to values [-len(y)...0] to represent in-sample forecast horizon

sktime/forecasting/statsforecast.py
```diff
@@ -628,14 +644,165 @@ def get_test_params(cls, parameter_set="default"):
...
+            **{f"hi-{_l}":
+                pred_int[("Coverage", c, "upper")].values
+                for c, _l in zip(coverage, level)},
+        }
+
+    def predict_in_sample(self, level=None):
+        fitted = self.estimator.predict(
+            fh=self.in_sample_fh,
+            X=self.in_sample_X
+        )[:, 0]
```
luca-miniati, line 747: When testing on `load_airline` dataset, `fitted` is way off (first value is -68, when it should be 112), because `ThetaForecaster` computes a negative drift. Drift should probably positive due to positive trend of airline dataset?

sktime/forecasting/statsforecast.py
```diff
@@ -552,3 +555,255 @@ def get_test_params(cls, parameter_set="default"):
...
+
+        if hasattr(self.estimator, "sp"):
+            if self.estimator.sp > 1:
+                raise Exception(
+                    "Trend forecaster should not adjust seasonal models. Please pass "
+                    "`sp=1` to your trend forecaster"
+                )
+
+    def __repr__(self):
+        return "StatsForecastBackAdapterMSTL"
```
yarnabrina, line 713: I'm not sure if this is necessary. Can you please share the `repr` output in presence and absence of this?

sktime/forecasting/statsforecast.py
```diff
@@ -552,3 +555,255 @@ def get_test_params(cls, parameter_set="default"):
...
+        "capability:pred_int": True,
+        "capability:pred_int:insample": True,
+    }
+
+    from sktime.forecasting.ets import AutoETS
+
+    def __init__(
+        self,
+        season_length: Union[int, List[int]],
+        trend_forecaster=AutoETS(),
```
yarnabrina, line 608: This changes the default from `statsforecast`. Can you consider using `StatsForecastAutoETS`?

sktime/forecasting/statsforecast.py
```diff
@@ -552,3 +555,255 @@ def get_test_params(cls, parameter_set="default"):
...
+        self.season_length = season_length
+        self.trend_forecaster = trend_forecaster
+
+        super().__init__()
+
+    def _instantiate_model(self):
+        """Create underlying forecaster instance."""
+        from statsforecast.models import MSTL
+
+        from sktime.forecasting.base._base import BaseForecaster
```
yarnabrina, line 619: `_base` is unnecessary, and given it's already imported above, I'm not sure why is it being imported again.

sktime/forecasting/statsforecast.py
```diff
@@ -552,3 +555,255 @@ def get_test_params(cls, parameter_set="default"):
...
+
+        if issubclass(type(self.trend_forecaster), BaseForecaster):
```
yarnabrina, line 621:
```suggestion
        if isinstance(self.trend_forecaster, BaseForecaster):
```

sktime/forecasting/statsforecast.py
```diff
@@ -552,3 +555,255 @@ def get_test_params(cls, parameter_set="default"):
...
+            self.trend_forecaster = StatsForecastBackAdapterMSTL(self.trend_forecaster)
+
```
yarnabrina, line 623: What happens if the check fails? Is `MSTL` capable of handling any forecaster from any library?

If not, it should raise an error, or at least throw an warning.

sktime/forecasting/statsforecast.py
```diff
@@ -552,3 +555,255 @@ def get_test_params(cls, parameter_set="default"):
...
+            `create_test_instance` uses the first (or only) dictionary in `params`
+        """
+        from sktime.forecasting.theta import ThetaForecaster
+
+        del parameter_set  # to avoid being detected as unused by ``vulture`` etc.
+
+        params = {
+            "season_length": [3, 12],
+            "trend_forecaster": ThetaForecaster(),
+        }
```
yarnabrina, line 655: Can you please add few more test configurations?

1. corresponding to `season_length` being `int` instead of `typing.List[int]`
2. corresponding to the case where forecaster is multivariate in nature, e.g. `VAR`

sktime/forecasting/statsforecast.py
```diff
@@ -552,3 +555,255 @@ def get_test_params(cls, parameter_set="default"):
...
+
+
+class StatsForecastBackAdapterMSTL(BaseForecaster):
+    """StatsForecast Back Adapter for StatsForecast MSTL.
+
+    StatsForecastBackAdapter is a wrapper for
+
+    Parameters
+    ----------
+    estimator :
```
yarnabrina, line 667: It's missing. Apart from a general documentation, I think it should contain the expected properties of the estimator, like it must have these methods taking these arguments and so on.

sktime/forecasting/statsforecast.py
```diff
@@ -552,3 +555,255 @@ def get_test_params(cls, parameter_set="default"):
...
+
+        return {
+            **{f"lo-{_l}":
+                pred_int[("Coverage", c, "lower")].values
+                for c, _l in zip(reversed(coverage), level)},
+            **{f"hi-{_l}":
+                pred_int[("Coverage", c, "upper")].values
+                for c, _l in zip(coverage, level)},
+        }
+
```
yarnabrina, line 767: Please remove the unused commented code below.

sktime/forecasting/statsforecast.py
```diff
@@ -552,3 +555,255 @@ def get_test_params(cls, parameter_set="default"):
...
+    def predict_in_sample(self, level=None):
+        fitted = self.estimator.predict(
+            fh=self.in_sample_fh,
+            X=self.in_sample_X
+        )[:, 0]
+
+        if level is None:
+            return {"fitted": fitted}
+
+        coverage = [round(1-(_l/100), 2) for _l in level]
```
yarnabrina, line 752: Does `level` have to be sorted? If not, will the column order in output be compatible with `statsforecast`?

yarnabrina (changes requested):
Thanks for this contribution. I made a few comments, let me know if any clarification is required.

Apart from the comments, please setup `pre-commit` locally and make the linting related changes. Only that'll allow the CI to move on to unit tests and we'll see if it passes all checks.

And in case I am missing something, can you please the idea/goal of the back adaper class? Since not all forecasters have `predict_interval` capability, and some forecasters do not have in-sample capability, I'm a bit confused.

Also, if I'm not wrong, `sktime` forecasters capabale of in-sample can go beyond the earliest available data, like `fh < -len(y)`. I'm really unsure when that's desirable, but if it's part of contract, your `in_sample_fh` is probably breaching that contract.

sktime/forecasting/statsforecast.py
```diff
@@ -552,3 +555,255 @@ def get_test_params(cls, parameter_set="default"):
...
+
+        params = {
+            "season_length": [3, 12],
+            "trend_forecaster": ThetaForecaster(),
+        }
+
+        return params
+
+
+class StatsForecastBackAdapterMSTL(BaseForecaster):
```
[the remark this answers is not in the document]
fkiraly: agree - if it works, it should be in one of the adapter modules as it is of general use. The `_generalised_statsforecast` is probably the best one.

sktime/forecasting/statsforecast.py
```diff
@@ -552,3 +555,255 @@ def get_test_params(cls, parameter_set="default"):
...
+                raise Exception(
+                    "Trend forecaster should not adjust seasonal models."
+                )
+
+        if hasattr(self.estimator, "sp"):
+            if self.estimator.sp > 1:
+                raise Exception(
+                    "Trend forecaster should not adjust seasonal models. Please pass "
+                    "`sp=1` to your trend forecaster"
+                )
```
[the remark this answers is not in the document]
fkiraly: I would simply remove?

I think it is perfectly fine, methodologically, that both the MSTL and the inner forecaster have a seasonal component.

fkiraly:
CI fails due to code quality issues, see details

sktime/forecasting/statsforecast.py
luca-miniati: Internally, MSTL tries to access attributes of `trend_forecaster` that won't exist if `trend_forecaster` is a back adapter. Here's a snippet from the StatsForecast source code for MSTL:
```python
# check ETS model doesnt have seasonality
if repr(trend_forecaster) == "AutoETS":
    if trend_forecaster.model[2] != "N":
        raise Exception(
            "Trend forecaster should not adjust " "seasonal models."
        )
```
Since not all estimators have `model`, I thought this would be messy to implement. 
I thought adding `__repr__` would solve the problem, but I understand that it is confusing. Is there a better way to do this?
(BTW, `repr` output in absence of this is: StatsForecastBackAdapterMSTL(estimator=AutoETS()))

luca-miniati:
To clarify the goal of the back adapter, @yarnabrina, it was to make it possible to pass any `sktime` forecaster for use in the MSTL interface. So, since some forecasters don't have in-sample/pred_int capability, perhaps should the back adapter raise exceptions when the StatsForecast MSTL tries to call such methods internally (if forecaster doesn't support method)?

sktime/forecasting/statsforecast.py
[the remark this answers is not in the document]
luca-miniati: As of now, MSTL seems to be the only composite model in StatsForecast. In case they add more, I moved it to `_generalised_statsforecast` and renamed it from `StatsForecastBackAdapterMSTL` to `StatsForecastBackAdapter`.

sktime/forecasting/statsforecast.py
luca-miniati: Gotcha, done

PR #4806, opened 2023-07-02
[ENH] add `RecursiveReductionForecaster` to module exports, tests, and documentation

fkiraly:
This adds `RecursiveReductionForecaster` to module exports of `forecasting.compose`, and to the API docs.

This should also ensure that tests are run, potentially uncovering bugs.


fkiraly:
note from tests: the forecaser currently does not seem to work on hierarchical data properly.

Perhaps code along these lines could help:
```python
# Example: If you have a MultiIndex, split and operate only on the date part:
if isinstance(y_plus_preds.index, pd.MultiIndex):
    date_index = y_plus_preds.index.get_level_values(-1)  # Assuming date is in the last level
    y_plus_preds.index = date_index
    y_plus_preds = y_plus_preds.asfreq(self.fh.freq)
```

Issue #4809, opened 2023-07-02
[ENH] managing defaults in public and private interfaces


fkiraly:
From this discussion https://github.com/sktime/sktime/pull/4780#discussion_r1246621774, FYI @benHeid, @yarnabrina.

*Context*: `sktime` has a public/private design using the template pattern, e.g., `fit` / `_fit`, where `fit` calls `_fit`.

Defaults are currently present in the public (e.g., `fit`) and the private (e.g., `_fit`) method - and it turns out the defaults in the private methods are inconsistent.
Whereas there is only one single default for the public method, and one location for those (the base class).

However, the private methods should not be directly called and usually get all parameters passed, so there is in-principle no need for default in the private methods (and no harm caused from the private defaults being inconsistent).

The *question* is how to deal with the defaults in the private methods, and the current inconsistency of those defaults.

Some options on this I can see:

1. ensure the private defaults are always consistent with the public defaults, and change the current inconsistency to that end state.
2. remove any defaults from private methods, and remove any current defaults from private methods.

My personal opinion is that 2 is better than 1, since 2 is better maintainable. Suppose a default changes somewhere, then this requires a change only in one place (the base class), not in 100s of classes including classes that may live in third party packages (as would be required in option 1). The important argument imo is that option 1 would require maintenance action by all extending users/developers, while option 2 would limit the maintenance scope to a single line of code (in the base class).

benHeid:
I agree with your opinion and also prefer option 2.

murriee:
Hi @fkiraly,

I’d like to work on removing the default parameters from the private methods as discussed in #4810 and #4780. To start,I will   update the _transform method (please let me know if there’s a preference for a different method) and then extend the changes to the remaining methods. Could you please assign this issue to me?

Thanks!

PR #4810, opened 2023-07-02
[ENH] remove private defaults in forecasting module

fkiraly:
This removes the private defaults in the forecasting module for the methods `_fit`, `_predict`, `_predict_interval`, `_predict_quantiles`.

Also see https://github.com/sktime/sktime/issues/4809.

Interestingly, this uncovered some bugs:

* in `AutoARIMA`, and `_TbatsAdapter` `_update`, `_fit` was invoked but `fh` was not passed (if it was already passed), but the default `None . This might have led to bugs down the line.

### DIFF
--- a/extension_templates/forecasting.py
+++ b/extension_templates/forecasting.py
@@ -193,7 +193,7 @@ class MyForecaster(BaseForecaster):
         #   self.clone_tags(est2, ["enforce_index_type", "handles-missing-data"])
 
     # todo: implement this, mandatory
-    def _fit(self, y, X=None, fh=None):
+    def _fit(self, y, X, fh):
         """Fit forecaster to training data.
 
         private _fit containing the core logic, called from fit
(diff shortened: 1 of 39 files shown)


yarnabrina (approved):
Looks good, but I have two comments.

Now almost (?) all docstrings are outdated though, as they all document these as optional keyword arguments. It'd be too much of a change I guess, but will it be possible to try once for `BaseForecaster` and do a bulk find+replace?

Also, there are lot more private methods in lots of classes, including the base one, e.g. `_update`, `_predict_var`, `_predict_proba`, etc. Are these all out of scope of this PR?

fkiraly:
> Are these all out of scope of this PR?

Yes, I just wanted to see first (a) whether anything serious breaks, and (b) what you and others think, before going on the more tedious exhaustive journey with other methods and docstrings.

Issue #3018, opened 2022-07-15
[ENH] forecasting horizon conditional parameter wrapper


fkiraly:
A common situation in forecasting multiple forecasting horizons is that one wants to do something different per horizon element.

For instance, if `fh=[1, 2, 3]`, I may like to lag my exogeneous data by 1, when forecasting 1 step ahead, lag my exogeneous data by 2, when forecasting 2 ahead, and lag my exogeneous data by 3, when forecasting 3 ahead. Or, I might like to lag only *some* exogeneous variables, this way, so do a FeatureUnion with `ColumnSelect` with `Lag`

This is done internally, in the "reduction" transformers, but that is not very flexible.
We should have generic functionality for this, it could look like as follows:

```python
class FhConditionalForecast(BaseForecaster):
    """Uses different parameters by forecasting horizon element.

    When provided with forecasting horizon [f1, f2, ..., fn],
    will fit forecaster with fh=f1 and parameters fh_params[f1] to forecast f1,
    forecaster with fh=f2 and parameters fh_params[f1] to forecast f2, etc.

    Parameters
    ----------
    forecaster : sktime compatible forecaster
    fh_params : dict, with keys = fh elements, values = param dict for forecaster
    """
```

Example usage:
```python
forecaster = ForecastingPipeline([Lag(), ARIMA()])
fc_by_lag = FhConditionalForecast(
    forecaster,
    fh_params = {
        1: {"Lag__lags": 1}
        2: {"Lag__lags": 2}
    }
)

fc_by_lag.fit(y, X, fh=[1, 2])
# internally, calls:
# forecaster.set_params(Lag__lags=1).fit(y, X, fh=1)
# forecaster.set_params(Lag__lags=2).fit(y, X, fh=2)

#then, we can do this, do not need X_pred!
fc_by_lag.predict()
```

aiwalter:
proposal to do this with a transformer:
```python
class XpredTransformer(BaseTransformer):
    # fh is int for simplification here, but we can do this probably with fh.to_indexer() method
    #  there can be also a param "column: Union[int, str, List] to select columns."
    def __init__(self,transformer, fh: int=3):
        self.transformer = transformer
        self.fh = fh

    def _fit(self, X, y=None):
        X_lagged = Lag().fit_transform(X=X, y=y)
        self._X_fit = X_lagged.iloc[:-self.fh]
        self._X_pred = X_lagged.iloc[:self.fh]
[...]

Issue #4776, opened 2023-06-27
[BUG] Can't combine `make_reduction` with `HistGradientBoostingRegressor.categorical_features`


davidgilbertson:
**Describe the bug**

scikit-learn's `HistGradientBoostingRegressor` accepts an argument `categorical_features`, instructing it to treat certain columns (in my exogenous data) as categorical. But the `make_reduction` function changes the columns so all the options for providing categorical features fail. To be specific:
* Providing a list of columns names: fails because a NumPy array is passed to `.fit()`, not the original dataframe
* Providing a boolean mask: fails because it should match the number of columns, but this is different after `make_reduction` has done its magic
* Providing the indexes of the columns: fails (sometimes silently) because the order of columns changes

**To Reproduce**

```python
from sktime.datasets import load_longley
from sktime.forecasting.compose import make_reduction
from sktime.forecasting.model_selection import temporal_train_test_split
from sklearn.ensemble import HistGradientBoostingRegressor

y, X = load_longley()
X["Year"] = X.index.year - 1900 # bad example, just to get a value that can be considered categorical
y_trn, y_tst, X_trn, X_tst = temporal_train_test_split(y, X, test_size=5)

forecaster = make_reduction(
    HistGradientBoostingRegressor(
        # categorical_features=["Year"],  # Strings
        # categorical_features=X.columns.isin(["Year"]),  # Boolean mask
        categorical_features=X.columns.get_indexer(["Year"]),  # Column positions
    ),
    window_length=2,
)

forecaster.fit(
    fh=[1, 2, 3],
    y=y_trn,
    X=X_trn,
)
```

**Expected behavior**

In a dream world, the column names I define in my DataFrame would still be there when `X` is eventually passed to `HistGradientBoostingRegressor.fit`.

Two solutions I can think of (with no regard for complexity or feasibility):
1. sktime only adds columns to the _end_ of the dataframe that the user provides as X (including adding y), so that we can at least rely on the column indexes remaining the same.
2. sktime keeps the dataframe and column names, only adding new columns.

2 Would be ideal as it solves other problems. For example XGBoost has the (experimental) ability to deal with Pandas categorical columns (`enable_categorical`), but as far as I can tell this wouldn't work with sktime because that information is lost in the conversion to NumPy.

**Additional context**

**Versions**
<details>

System:
    python: 3.10.8 (main, Oct 12 2022, 19:14:26) [GCC 9.4.0]
executable: /home/davidg/.virtualenvs/learning/bin/python
[...]

Issue #3224, opened 2022-08-09
[ENH] reducer design for rework


fkiraly:
I think we need a small design document for the rework of the reducers. FYI @danbartl, given recent discussions.

Cases we should cover with modular design:

* lagging `X` and lagging `y`, lagging `y` but not lagging `X`
* use of contemporary `X`, use of shifted `X` (by forecasting horizon)
* use of transformers on `X`, e.g., window summarizer
* removing missing values created due to shifting/windowing
* treating missing values created due to shifting/windowing (e.g. filling)
* treating missing values from shifting/windowing differently from pre-existing missing values, including options to leave the latter be

ltsaprounis:
One more point from me:
- Use of transformers that are applied on the index such as the `DateTimeFeatures` transformer and other transformers that could be applied to the hierarchy features of hierarchical time series.

ltsaprounis:
> * use of contemporary `X`, use of shifted `X` (by forecasting horizon)
Would that force passing `fh` on fit?

fkiraly:
> Would that force passing `fh` on fit?

I suppose so?

fkiraly:
PR with a prototype design for direct and recursive forecasting:
https://github.com/alan-turing-institute/sktime/pull/3333

Solves the above points as follows:

>lagging X and lagging y, lagging y but not lagging X

reducers by default *do not apply any transformations (incl lagging) to `X`*, only to `y`.
* to lag `X`, the `Lag` transformer needs to be applied in a pipeline to exogeneous data.
* to lag `X` and not `y`, apply the `Lag` trafo to exog, and leave the window length parameter at zero.
* lagging `y` an not `X` is the standard behaviour.

 > use of contemporary X, use of shifted X (by forecasting horizon)

In direct reduction, these are options, via parameter `X_treatment` that can be `"concurrent"` or `"shifted"`. In recursive reduction, concurrent is default. Shifted is obtained by lagging `X` with a lag of 1 (we could turn thsi into a parameter, too).

> use of transformers on X, e.g., window summarizer

Transformers for `X` are not passed to the reducers, but need to be applied via a pipeline.
This is more modular, and avoids duplication and "god class" like parameter handling.

> removing missing values created due to shifting/windowing

done by parameter `impute` which is looped through to `Imputer`.
All options in `Imputer` are available in reduction.

`Imputer` has no missing value removal, but that could be added as an option, and would also add the option to the reducers, via the estimator modular dependency.

[...]

PR #4811, opened 2023-07-02
[ENH] by-horizon forecaster, for different estimator/parameter per horizon

fkiraly:
Fixes https://github.com/sktime/sktime/issues/3018

Implements a forecaster `FhPlexForecaster` that allows to specify different parameters per forecasting horizon.

To specify different forecasters, combine with `MultiplexForecaster` or `MultiplexTransformer`.

Potentially useful in https://github.com/sktime/sktime/issues/4776#issuecomment-1616804513.

FYI @davidgilbertson, would appreciate feedback.

Currently does not have as features (but should have?)

* multiple `fh` indices per forecaster
* inheritance from `_HeterogenousMetaEstimator` - I already prepared the `_forecasters` attr for that but didn't yet test it with inheritance. Hopefully works out of the box.

### DIFF
--- a/docs/source/api_reference/forecasting.rst
+++ b/docs/source/api_reference/forecasting.rst
@@ -50,6 +50,7 @@ Pipelines can also be constructed using ``*``, ``+``, and ``|`` dunders.
     ForecastByLevel
     Permute
     HierarchyEnsembleForecaster
+    FhPlexForecaster
 
 Reduction
 ---------
--- a/sktime/forecasting/compose/__init__.py
+++ b/sktime/forecasting/compose/__init__.py
@@ -2,8 +2,6 @@
 # copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
 """Implements composite forecasters."""
 
-__author__ = ["mloning"]
-
 __all__ = [
     "HierarchyEnsembleForecaster",
     "ColumnEnsembleForecaster",
@@ -24,6 +22,7 @@ __all__ = [
     "MultiplexForecaster",
     "make_reduction",
     "BaggingForecaster",
+    "FhPlexForecaster",
     "ForecastByLevel",
     "Permute",
     "YfromX",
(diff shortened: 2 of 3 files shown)


davidgilbertson:
I don't have any strong opinions on this. It does look like it could solve the problem I have, but requires quite a lot of user code. 

I think/hope I might be confused because I thought what I'm doing was pretty much stock-standard time series analysis and would be how the default, simple, inflexible would work. Can you confirm whether or not this particular use case should be currently possible?

* I have `X` and `y` time series data, with multiple instances.
* I want to put, say, 2 lags of `y` data into my `X` dataframe (two new columns) and pass that to a particular estimator's `.fit()` method.
* I want to predict, say, 14 timesteps in the future

Everything I've tried fails in some way:
* `make_reduction` by default lags every column in `X` - a flag to control whether the user wants to lag X, y, both, or some named columns would solve this.
* `make_reduction` passing `transformers` and `window_length=None` can only do 'global' calculations, not one-estimator-per-instance.
* `YfromX` leaks future values of `y` into the dataframe passed to my estimator during `.predict()`
* `DirectReductionForecaster` takes far too long to run (more than 30x slower)

So is this use case (X and y data, want to put lags of y into X) actually impossible at the moment, and this PR resolves that? Or is there something I've missed.

fkiraly:
> actually impossible at the moment, and this PR resolves that? Or is there something I've missed.

Yes - the idea is to make it possible first (the combination of unlagged `X`, lagged `y`, global, which for an odd reason seems not supported) and then optimize.

Tbh the off-shelf reducers should work here.

Btw, it would be great if you could paste some code for `DirectReductionForecaster` in this issue that shows a long runtime: https://github.com/sktime/sktime/issues/3224 - I have been mostly diagnosing the `RecursiveReductionForecaster`.

davidgilbertson:
OK I'm glad you agree its odd that it's not supported, that means I'm not going crazy :)

Actually it seems I've just discovered a way. I _think_ I can wrap my whole pipeline in `ForecastByLevel`, and then use the reduction method that only works with global pooling:
```py
make_reduction(
    HistGradientBoostingRegressor(),
    transformers=[
        WindowSummarizer(
            lag_feature={"lag": range(1, target_lags + 1)},
            truncate="bfill",
            n_jobs=n_jobs,
        ),
    ],
    pooling="global",
    window_length=None,
)
```

Haven't tested thoroughly, and it still has a tiny leak due to `bfill` reaching into the future, but it seems that it at least allows me to use sktime.

fkiraly:
> OK I'm glad you agree its odd that it's not supported, that means I'm not going crazy :)

No, not at all.

It is not uncommon that features are added consecutively on estimators (e.g., global, use of `X`) so that combination cases don't work properly.

That's why we wanted to start with a re-design in 2022, which has the "right user interface" imo, it ended up working, but being computationally too expensive. So the next step is keeping the interface but making it faster.

PR #4823, opened 2023-07-04
[MNT] unpin `sphinx` and plugins, with defensive upper bounds

fkiraly:
This PR replaces all the pins from `sphinx` and plugins with defensive upper bounds - next MAJOR release if the package is post-0, next MINOR release if the package is on zero-point.

---

original post: Diagnostic PR to see what happens if we remove all the pins from sphinx and plugins.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -157,11 +157,11 @@ docs = [
     "nbsphinx>=0.8.6",
     "numpydoc",
     "pydata-sphinx-theme",
-    "sphinx_issues<4.0.0",
+    "sphinx-issues<4.0.0",
     "sphinx-gallery<0.14.0",
     "sphinx-design<0.5.0",
     "sphinx-version-warning",
-    "sphinx<8.0.0",
+    "Sphinx<8.0.0",
     "tabulate",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -157,11 +157,11 @@ docs = [
     "nbsphinx>=0.8.6",
     "numpydoc",
     "pydata-sphinx-theme",
-    "sphinx_issues==1.2.0",
-    "sphinx-gallery==0.6.0",
-    "sphinx-design==0.3.0",
+    "sphinx_issues",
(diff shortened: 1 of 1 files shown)


yarnabrina:
Read the docs workflow seems to be working fine, so we can remove pins?

(Assuming other workflows are all independent of `docs` dependency, and hence don't matter.)

fkiraly:
> Assuming other workflows are all independent of `docs` dependency, and hence don't matter.

Afaik this assumption is correct, I just checked the `yml`-s and could not find a `docs` set.

> Read the docs workflow seems to be working fine, so we can remove pins?

Maybe upper bound to next major (or minor if zero-point), for safety?

yarnabrina:
I doubt it'll have changes that'll break `sktime`, but as you said, it makes sense for safety. Let's do in this PR only.

(You can commit with [skip ci] in commit message I think, it'll skip all Github workflows, but read-the-docs one will still run. )

fkiraly:
> You can commit with [skip ci] in commit message I think, it'll skip all Github workflows

Thanks for the reminder and learning moment!

Although in this specific case, I think it's worth to check carefully whether it has any interactions with the other dependencies by a way we don't foresee.

yarnabrina (approved):
As long as read the docs workflow passes, I'm happy.

One minor non-blocking suggestion: use `sphinx-issues` instead of `sphinx_issues` as hyphen one is the [PyPI name](https://pypi.org/project/sphinx-issues/).

fkiraly:
> One minor non-blocking suggestion: use `sphinx-issues` instead of `sphinx_issues` as hyphen one is the [PyPI name](https://pypi.org/project/sphinx-issues/).
> Similarly, `Sphinx` instead of `sphinx`. I don't think these differences matter though, as `pip` checks only for sanitised versions.

Updated. I did not know that `pip` did some kind of closest name resolution.

I knew it doesn't allow package names which are too close, e.g., only dashes or capitalization away.

yarnabrina (approved):

Issue #4695, opened 2023-06-12
[ENH] Add real world datasets into dataset module


hazrulakmal:
To conduct a real-world benchmarking study, we need access to real-world datasets. However, the current dataset module only contains a collection of simple/toy forecasting datasets. To address this limitation, I suggest incorporating more complex datasets from two public repositories: 

1. [Monash Forecasting](https://forecastingdata.org/)
2. [Time Series Extrinsic Regression](http://tseregression.org/)

The goal is to create a function similar to `load_UCR_UEA_dataset` for classification tasks. This function would fetch datasets from the public repositories, download them to the local machine, and load them into memory into a format that is compatible with estimators.

Monash Forecasting has several panel datasets that the current module does not include see #3465 , one panel dataset example from the repo is  *London Smart Meters*.  The other repository is for a regression task, so it should also address issue #4314. Personally, I think it would be better to have a function for users to download the datasets, as this would reduce the memory footprint of the package and offer many more datasets for users to choose from.

This feature would not only benefit the benchmarking framework, but also users who are interested in working with real-world time series examples. With just a single line of code to download and load the dataset, users can focus on modeling rather than data collection side of things.

hazrulakmal:
I believe that those functionalities like `make_time_series_dataset` is not  **real-world** datasets, but rather synthetic ones generated based on some characteristics. Therefore, I suggest creating a separate issue for this matter.

fkiraly:
> If I'm not mistaken, there is a separate issue for this matter.

yes, afaik it´s this one: https://github.com/sktime/sktime/issues/4098

Issue #4754, opened 2023-06-23
[ENH] file hosting for buffering or mirroring upstream data repositories


fkiraly:
Design issue for finding a solution to mitigating abrupt changes or failures of upstream data repositories in `sktime`. Currently, such changes directly propagate to the user, and they are not unlikely as the data repositories are usually academic and not professionally maintained.

From a design perspective, I see four options:

1. status quo
2. drop data set loaders entirely
3. mirroring: mirror the upstream and use the mirror as primary source
4. fallback/buffer: mirror the upstream but use the mirror only if the upstream fails

Options 3 and 4 require hosting the files, which in the case of most data repositories are in the order of 100s of files at a total file volume in the order of GB.

Hosting mechanisms discussed so far:

* GitHub LFS, an implementation of option 3 with LFS is here: https://github.com/sktime/sktime/pull/4749
* AWS S3 with http redirect from sktime.net to the S3 bucket (proposed by @kirilral and @JonathanBechtel)
* other cloud providers than Amazon same principle

hazrulakmal:
In my opinion, I will vote for options 3 and 4, but I will not suggest hosting datasets on platforms (AWS, GCP etc) that require costs to maintain them, to avoid the financial burden since we are a non-profit entity. Instead, I suggest storing datasets in sktime's Google Drive. Every Google account comes with a [free 15GB](https://one.google.com/about/plans?hl=en_GB), which I hope should suffice for our case. @fkiraly, do you have any ideas on how big datasets in GB can be, since you already made the first move of storing all datasets from the classification website in a GitHub repository? If the allocated storage of 15GB is not sufficient, we can do two things:
1. Create a secondary sktime email account to get another 15GB of storage.
2. Be more selective about which datasets to store in Google Drive.

The Google Drive API has [rate limits](https://developers.google.com/drive/api/guides/limits) (number of requests allowed for a given time), allowing only 20,000 requests per 100 seconds. I think it should be sufficient to support sktime current and future users who use data loaders. To implement downloading files from Google Drive, we can closely follow this [StackOverflow](https://stackoverflow.com/questions/38511444/python-download-files-from-google-drive-using-url). Additionally
To further enhance this limit, I think it's important to choose when to use options 3 and 4 carefully.

The decision to use option 3 or 4 depends on the capacity of the upstream repository, specifically whether or not the API link is professionally maintained and capable of handling high-frequency requests and queries.

[...]

PR #4826, opened 2023-07-04
[ENH] add Monash Forecasting Repository data loader

hazrulakmal:
#### Reference Issues/PRs
See #4695. Since this PR implement one of the two repos mentioned in the reference PR, it should'nt be closed.

Add a `fetch_forecasting` dataloader that retrieves forecasting datasets from the Monash Time Series Forecasting Repository. This dataloader first looks for the dataset in the extract_path. If the dataset is not present, it attempts to download the data from *https://forecastingdata.org/* and saves it to the extract_path.

To ensure consistency with the implementation of `load_UCR_UEA_dataset`, a dataloader that does the same  for classification datasets, `load_forecastingdata` closely follows it and uses some of the private functions that it relies on.

Discussed: A naming fetch instead of load would be better to distinguish it from other dataloaders that load data from sktime pre-installed datasets as per suggestion in #4314. For now, we've called it `load_forecastingdata` for consistency reasons. 

**Work in progress**

- [ ] `fetch_forecastingorg` function
- [ ] tests to account the changes


hazrulakmal:
1. I have made some changes to `_list_available_datasets` to also look for datasets from the forecastingorg repo, which has a different format than the one from classification. FYI, classification datasets from `timeseriesclassification.com` and regression datasets from `tseregression.org` have a similar format of `<dataset name>._TRAIN.ts` and `<dataset name>._TEST.ts`, while forecasting datasets are in the tsf format. I added some tests to ensure this works correctly.
2. I have modified `fetch_forecastingorg` to search for datasets in the directory first to load them to memory. If the dataset is not found, `fetch_forecastingorg` will download it from the repo. By default, `fetch_forecastingorg` saves the datasets in the `data` directory together with other pre-installed datasets, which is slightly different than `load_UCR_UEA_dataset`, which creates a new directory `local_data` to save the datasets. Parking downloaded datasets in the `data` directory, in my opinion, is cleaner and results in simpler code, as there is no need to sequentially search in `data` and then in `local_data` before deciding to download the datasets - search can be done in just a single dir (reduce redundancy).
3. I have added a test to check if the URL exists and is downloadable without downloading the zip file (for speed) for all available datasets in `test_single_problem_loaders`. This helps us quickly identify if the upstream repo has deleted a dataset or changed the URL (although the former is unlikely to happen, as the datasets are saved in a third-party, well-maintained storage platform). However, I'm unsure if checking the existence of a dataset every time sktime developers run a CI test is desirable. Another option is to trigger a CI test for this periodically (weekly, monthly, etc.). I'm not very familiar with CI workflows, so if this needs to be implemented, I need some guidance on how to do it.

tagging @achieveordie, since you worked in dataset module before.

yarnabrina:
I think these are related to your 3rd point. I'm not aware if any decision was taken in dev or council calls.

https://github.com/sktime/sktime/issues/4754
https://github.com/sktime/sktime/pull/4749

fkiraly:
side note, @hazrulakmal - given that you have looked at this, do you have any suggestions for https://github.com/sktime/sktime/issues/4754 ?

fkiraly:
> I'm not aware if any decision was taken in dev or council calls.

I see we were thinking of the same reference, @yarnabrina - no, no decision has been taken so far, the last was that we would engage with #4754 and come up with options (which has not happened yet - I assume due to holiday weeks).

achieveordie (commented):
The code is excellent and so are the tests (we still need to think about the 3rd point you raised). My general concern right now is that we might be collecting some technical debts if we were to be directly inspired from the classification loaders code.

As you know, the current implementation for forecasting and classification have a lot of steps in common, resulting in a bloated codebase. If we choose to add a separate repository for a different task, we would unnecessarily increase it even further.

So I suppose we should be asking ourselves some fundamental questions about designing data loaders and fetchers so that extending it remains simple and we avoid amassing any amount of tech debt for us to deal with in the future. 

(This could very well be outside of the scope of this PR, but we need to start thinking in this context to avoid the aforementioned problems.)

hazrulakmal:
> As you know, the current implementation for forecasting and classification have a lot of steps in common, resulting in a bloated codebase. If we choose to add a separate repository for a different task, we would unnecessarily increase it even further.

I have personally looked at the dataset module codebase and I hope our understanding of *bloated codebase* is the same. However, I don't think that adding another repository, in this case for a forecasting task, is the root cause of the problem. Instead, I believe it's more about how the module is maintained and structured.  Currently, all kinds of dataset functionality (readers, loaders, writers, etc.) are stored in one file, `_data_io.py`. This can be a headache for new developers and should be separated into different files for long-term maintainability. Common steps can always be remedied by gluing them together with utility functions. Refactoring the module should be able to resolve all these issues, which is a separate issue from this pull request.

In regard to the third point, I have made some comments on the issue. Please do look at it.

hazrulakmal:
> .. the current design is not suitable for extension ..

I agree with you on this matter. The inherent flaw in the dataset design enforces repeatable codes for the same processing steps. In my opinion, the current design is more of a research code.

> designing data loaders and fetchers so that extending it remains simple 

Seconding this. As the dataset collection continues to grow to account for more complicated cases, a major redesign of the dataloader abstraction is something worth considering - I'm thinking about datasets loader something similar to deep-learning libraries like pytorch datasets. Your point above should be an important factor to consider. I think this will be a great addition to work-in-progress enhancements in the kotsu benchmarking framework. Would you be willing to collaborate on this? I believe some early design work is currently being carried out here. #4333 

> Should we add more repositories that download from a different source into a different format, we would need to repeat the same process. This isn't on new developers that are trying to extend, but an inherent flaw of the API design

Moving forward, I believe it would be best to open a central issue to discuss the shortfalls in the dataset module, as we have been discussing them sporadically here and there across multiple dataset-related issues. Regarding this specific issue, I will change the interface `fetch_forecasting` to `load_forecasting_dataset` to maintain consistency with the current implementation for fetches. As you pointed out, we may not want to extend the dataset using the current design so any direct improvement to existing code can be discarded, yet this PR still offers a great feature to users. what do you think?

fkiraly (dismissed):
Agree with the reasoning above - let's merge this as it is useful functionality.

I'll just rename it to `load_forecastingdata` and add it to the API reference.

fkiraly:
(please let me know, @hazrulakmal, if you'd rather have a different name etc)

hazrulakmal:
> (please let me know, @hazrulakmal, if you'd rather have a different name etc)

I'm ok with the naming :)

PR #4831, opened 2023-07-05
[DOC] improved guide for soft dependencies

fkiraly:
From a few discussions I had, it turns out that the guide on soft dependencies is slightly confusing, as it has content about handling existing soft dependencies and adding one, but both in a section called "adding a soft dependency".

To resolve the confusion, I've split the existing section in two:

* a section about handling soft dependencies
* a section about adding soft dependencies

I've also updated the documentation with how pandas 1 and pandas 2 conditional testing of soft dependencies is currently handled in `pyproject.toml`


achieveordie (approved):
LGTM

PR #4832, opened 2023-07-05
[ENH] Improve handling of dependencies with alias

### LINKED ISSUE #4820: [ENH] Extend tagging system to handle dependencies with alias
Currently, estimator tagging system failed to handle soft dependency that has alias. For example, `AlignerDTW` (dtw-python/dtw) from the alignment module and #4580 (scikit-optimize/skopt). 

The current solution is to use a private validation utility function, `_check_soft_dependencies` before calling `super().__init__()` to handle the shortfall. 

**Solution Descriptions**
Other solution that I have in mind is to add an additional tag, *dependencies_alias* and extend the `_check_estimator_dep` to check for alias and pass the alias to `_check_soft_dependecies` that is used under the hood. A quick code solution may look like

```python
class Estimator(BaseEstimator):

    _tags = {
        ...
        "python_dependencies": ["numpy", "pandas", "scipy", "scikit-learn"],
        "python_dependencies_alias": {"scikit-learn": "sklearn"},
    }

    def __init__(...):
        ...

# Improvement to check_estimator_dep 
# since _check_soft_dependecies already handle the alias, check_estimator_dep should be easy to extend the case.

def _check_estimator_deps(obj, msg=None, severity="error"):
    compatible = True

    ... object iteration check ...
    ... python version check ...

    pkg_deps = obj.get_class_tag("python_dependencies", None)
    pck_alias = obj.get_class_tag("python_dependencies_alias", None)
    if pkg_deps is not None and not isinstance(pkg_deps, list):
        pkg_deps = [pkg_deps]
    if pkg_deps is not None:
        pkg_deps_ok = _check_soft_dependencies(*pkg_deps, severity=severity, obj=obj, package_import_alias=pck_alias)
        compatible = compatible and pkg_deps_ok

    return compatible
```

hazrulakmal:
#### Reference Issues/PRs
Fixes #4820

This PR improves the tagging system to account for dependencies with aliases. The following changes have been made to achieve this:
- [x] _check_estimator_deps has been extended to pass alias information.
- [x]  `AlignerDTW` estimators have been updated to reflect the changes.
- [x] The dependencies documentation has been updated.

Are there anything else (estimator, etc) that I should be aware of in order to reflect these new implementations?


fkiraly:
FYI, you need to add new tags in `registry._tags` or it will throw an error.

fkiraly:
This looks fine now! Would be keen to merge, as we need `scikit-learn!=1.3.0` on one of the estimators (could you also add the `sklearn` alias in the `BaseObject`?)

fkiraly (dismissed):
Would approve as-is, seems done and as intended, but this is still a draft (so not done)?

I will soon need the alias `"scikit-learn": "sklearn"` in another PR (to deal with a bug from `sklearn 1.3.0`) so merge would be great.

fkiraly:
ah yes, the docs - good catch!

fkiraly (approved):
Great addition, thanks!

> Are there anything else (estimator, etc) that I should be aware of in order to reflect these new implementations?

Not that I can think of.

Issue #4837, opened 2023-07-05
[MNT] how to deal with soft dependencies that imply bounds on core dependencies


fkiraly:
I noticed a problem with our CI: some soft dependencies imply bounds on core dependencies. For example, `numba` implies `numpy<1.25`.

In consequence, the bulk of the tests is not run with maximal version core dependencies, but the minimum bound of the soft dependencies.
(although the base tests are run with soft dependencies only).

As we accumulate more and more soft dependencies, this is bound to happen.

How should we deal with this? I can think of a couple options.

A. leave as-is

B. the main matrix of tests (OS and python version) is run without soft dependencies. That excludes `statsmodels` and `numba`. The soft dependencies run in their own test batch, perhaps a smaller matrix.

C. we have multiple matrices of tests, we could have three tiers:

* important packages that always tend to be up to date, e.g., `statsmodels`
* important packages that are slow and lagging, e.g., `numba` or `statsforecast` (the latter implies `pandas<2`)
* packages used only in one or two estimators

D. factor out models into their own repos which are imported from `sktime`. Tests are run there, for similar batches of models.

Any thoughts?

PR #4838, opened 2023-07-05
DIAGNOSTIC [MNT] run test suite with `numpy 1.25`

fkiraly:
Removes `numba` from the test dep set to ensure it runs with `numpy 1.25`

This is done by removing packages that imply `numba` from the `all_extras_pandas2` dependency set:

* `numba` itself
* `pyod`
* `tsfresh`

Related: https://github.com/sktime/sktime/issues/4837


fkiraly:
sth is *still* implying `numba` 😢

Issue #4842, opened 2023-07-06
[BUG] `evaluate` subroutine `_split` splits multiindex dataframes incorrectly when `X` has more timepoints than `y`


felipeangelimvieira:
**Describe the bug**
`forecasting.model_evaluation._functions._split` uses `y` integer-location to select the rows of X. However, if the dataframes are multiindex and X has more timepoints than y, it selects rows of X corresponding to the wrong dates.  This affects `evaluate` when using the splitters to backtest a forecasting model.

**To Reproduce**

```python

from sktime.transformations.hierarchical.aggregate import Aggregator
from sktime.forecasting.model_selection import ExpandingWindowSplitter
from sktime.forecasting.model_evaluation._functions import _split
from sktime.utils.validation.series import check_equal_time_index
import pandas as pd
import numpy as np

# A multiindex with level 0 with as "A", and "B" as dates from 2020-01-01 to 2020-01-10
df = pd.DataFrame(
    index=pd.MultiIndex.from_product(
        [["A", "B"], pd.date_range("2020-01-01", "2020-01-10").to_period("D")]
    ),
    data={"target": np.arange(20) % 10},
).sort_index()
df = Aggregator().fit_transform(df)

y = df[df.index.get_level_values(-1) < "2020-01-08"]
X = df.copy()
cv = ExpandingWindowSplitter(initial_window=2, fh=[1], step_length=1)


train, test = next(cv.split(y))
y_train, y_test, X_train, X_test = _split(y=y, X=X, train=train, test=test)
check_equal_time_index(y_train, X_train)
check_equal_time_index(y_test, X_test)

```

**Expected behavior**

_split should have returned y/X pairs with the same time index.

**Additional context**

Although this error can be circumvented by filtering the dates of X before passing it to evaluate, it's common to have an X dataframe with future values and maybe this shouldn't cause trouble when backtesting. If this is the expected behaviour, the error message is not clear.

**Versions**
<details>

System:
    python: 3.8.16 (default, Dec  7 2022, 01:36:11)  [Clang 14.0.0 (clang-1400.0.29.202)]
executable: /Users/felipe/Library/Caches/pypoetry/virtualenvs/app-2GAkMlzy-py3.8/bin/python
   machine: macOS-13.4.1-x86_64-i386-64bit

Python dependencies:
          pip: 23.0.1
       sktime: 0.19.2
      sklearn: 1.2.2
       skbase: 0.4.6
        numpy: 1.22.3
        scipy: 1.7.3
[...]

PR #4851, opened 2023-07-08
[ENH] splitter that replicates `loc` of another splitter

fkiraly:
This adds a splitter that takes a splitter and some data, and always replicates the same `loc` references for splits of another splitter.

Related issue and discussion: https://github.com/sktime/sktime/issues/4842

More general form of `temporal_train_test_split`, particularly useful for the case where we want to split `X` and `y`, and primarily `y`, and `X` can have different indices from `y`.

This ability partially addresses #4842, FYI @felipeangelimvieira

### DIFF
--- a/sktime/forecasting/model_selection/_split.py
+++ b/sktime/forecasting/model_selection/_split.py
@@ -333,6 +333,10 @@ class BaseSplitter(BaseObject):
         Single step ahead or array of steps ahead to forecast.
     """
 
+    _tags = {"split_hierarchical": False}
+    # split_hierarchical: whether _split supports hierarchical types natively
+    # if not, splitter broadcasts over instances
+
     def __init__(
         self,
         fh: FORECASTING_HORIZON_TYPES = DEFAULT_FH,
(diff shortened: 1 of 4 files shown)


fkiraly:
this is weird, equality tests fails only on windows, as array type seems to be `int64` vs `int`. Why only on windows?

sktime/forecasting/model_selection/_split.py
```diff
@@ -1301,6 +1308,112 @@ def get_test_params(cls, parameter_set="default"):
...
+        for y_train_loc, y_test_loc in cv.split_loc(y_template):
+            y_train_iloc = y.get_indexer(y_train_loc)
+            y_test_iloc = y.get_indexer(y_test_loc)
+            yield y_train_iloc, y_test_iloc
+
+    def get_n_splits(self, y: Optional[ACCEPTED_Y_TYPES] = None) -> int:
+        """Return the number of splits.
+
+        Since this splitter returns a single train/test split,
+        this number is trivially 1.
```
benHeid, line 1369: Same comment as in #4862 it's okay if it is answered there or only here:

I do not understand this comment, what is the referring to? The return value can be any positive integer depending on the parameters of the inner cv.

benHeid (changes requested):
Only one comment, I am also fine if this comment is answered on #4862 since it is exactly the same comment

sktime/forecasting/model_selection/_split.py
fkiraly: oh, that's a typo and wrong. Will fix.

benHeid (approved):
Thank you for addressing the comments :)

PR #4821, opened 2023-07-04
[MNT] update readthedocs env to python 3.11 and ubuntu 22.04

fkiraly:
This PR updates the readthedocs env to the current [official recommendations](https://docs.readthedocs.io/en/stable/config-file/v2.html), namely python 3.11 and ubuntu 22.04. That should also, hopefully, speed up the docs build.

PR #4852, opened 2023-07-08
Dependabot Setup

yarnabrina:
Ref. https://docs.github.com/en/code-security/dependabot/dependabot-version-updates

Previusly discussed in https://github.com/sktime/sktime/pull/4821#issuecomment-1620336961.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,20 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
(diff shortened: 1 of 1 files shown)


yarnabrina:
Note for admins: hopefully this file should be enough to set it up correctly, but if not, it may need enabling it from the repository settings. The details can be followed in the link in the PR description.

fkiraly (approved):
Seems simple enough, let's see how this works.

Thanks!

PR #4860, opened 2023-07-09
&hazrulakmal [MNT] fix remaining `sklearn 1.3.0` compatibility issues

fkiraly:
There are some remaining compatibility issues from the `sklearn 1.3.0` release which the pre-release testing did not catch. See https://github.com/sktime/sktime/pull/4778. This PR fixes those.

* `ElasticEnsemble` instantiated `sklearn` `RandomizedSearchCV` with `float` `n_iter`, which sets off the parameter validation since 1.3.0. Fixed by coercing to `int` (rounding down)
* `MatrixProfileClassifier` fails due to `sklearn` bug https://github.com/scikit-learn/scikit-learn/issues/26768. This is 1.3.0 specific and appears fixed for (not yet released) 1.3.1, so added an `sklearn!=1.3.0` version requirement for `MatrixProfileClassifier`

Depends on the bugfix https://github.com/sktime/sktime/pull/4867 to ensure the specifier `sklearn!=1.3.0` is handled correctly.

### DIFF
--- a/sktime/base/_base.py
+++ b/sktime/base/_base.py
@@ -73,6 +73,9 @@ class BaseObject(_BaseObject):
     Extends skbase BaseObject with additional features.
     """
 
+    # global dependency alias tag for sklearn dependency management
+    _tags = {"python_dependecies_alias": {"scikit-learn": "sklearn"}}
+
     def __eq__(self, other):
         """Equality dunder. Checks equal class and parameters.
 
--- a/sktime/base/_base.py
+++ b/sktime/base/_base.py
@@ -74,7 +74,7 @@ class BaseObject(_BaseObject):
     """
 
     # global dependency alias tag for sklearn dependency management
-    _tags = {"python_dependecies_alias": {"scikit-learn": "sklearn"}}
+    _tags = {"python_dependencies_alias": {"scikit-learn": "sklearn"}}
 
     def __eq__(self, other):
         """Equality dunder. Checks equal class and parameters.
--- a/sktime/base/_base.py
+++ b/sktime/base/_base.py
@@ -73,9 +73,6 @@ class BaseObject(_BaseObject):
     Extends skbase BaseObject with additional features.
     """
 
-    # global dependency alias tag for sklearn dependency management
-    _tags = {"python_dependencies_alias": {"scikit-learn": "sklearn"}}
-
     def __eq__(self, other):
         """Equality dunder. Checks equal class and parameters.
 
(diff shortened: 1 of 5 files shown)


fkiraly:
FYI @hazrulakmal, this is where I needed your PR https://github.com/sktime/sktime/pull/4832

Bit ironic given that a few days ago we were talking about how rare those aliases were...

sktime/base/_base.py
```diff
@@ -73,6 +73,9 @@ class BaseObject(_BaseObject):
     Extends skbase BaseObject with additional features.
     """
 
+    # global dependency alias tag for sklearn dependency management
+    _tags = {"python_dependencies_alias": {"scikit-learn": "sklearn"}}
+
```
hazrulakmal, line 78: I suggest placing the _tags  for python_dependencies_alias in `BaseEstimator` instead of `BaseObject`, since the tagging is designed for estimators anyway. Also, based on the test created in *test_base.py*, `BaseObject` requires a clean tag (i.e., a class with no tag at all) for all test cases to succeed. If you still want to place the tag in `BaseObject`, you may need to directly modify the test fixtures directly.

hazrulakmal: A small request from me 😅 (ignore this if somehow burdens you - I don't mind 😇), if you think the suggestion is great and place the _tag in `BaseEstimator` later, could you make the [commit as *commit  with coauthors*](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors). will be great to collect [github achievement badges](https://github.com/drknzz/GitHub-Achievements) here. hehe

hazrulakmal (changes requested):

sktime/base/_base.py
fkiraly: hm, never done that manually - how about you PR to this branch, that would count, right?

(agree with moving tag to `BaseEstimator`)

hazrulakmal: > how about you PR to this branch, that would count for the badge once we merge this branch to `main`, right?

not quite sure but I think it's worth trying - at least get to learn to make a PR to a different branch other than `main`

fkiraly: (I think it counts because there are commits from different people in the branch - they get squashed to a single commit to `main` authored by multiple contributors)

hazrulakmal: alright, thanks!! :)

PR #4861, opened 2023-07-09
[ENH] cleaner `evaluate` routine, fix for case where `y` and `X` are not equal length

fkiraly:
This refactors the `evaluate` forecasting benchmark tool and adds features:

* refactors the internal `_split` to use the `BaseSplitter` interface instead of separate logic.
* to handle `X`, we use the `SameLocSplitter` from https://github.com/sktime/sktime/pull/4851 and the `TestPlusTrainSplitter` from https://github.com/sktime/sktime/pull/4862
* As #4851 is more general and allows `y` and `X` of different length, this fixes https://github.com/sktime/sktime/issues/4842
* new argument to `evaluate` which allows the user to pass the splitter for `X` explicitly as `cv_X`. If not passed, defaults to the `SameLocSplitter`, i.e., `X` split indices are same as from `y`. (requires no deprecation as added as last arg, and default is existing behaviour)
* proper docstring

Note: this changes behaviour in cases where `y` and `X` had same length but different index set. However, I would contend (is this correct?) that in these cases behaviour was generally unexpected or buggy, so no deprecation is needed.

Depends on:
https://github.com/sktime/sktime/pull/4851
https://github.com/sktime/sktime/pull/4862

Related to, and indirectly fixes https://github.com/sktime/sktime/issues/4842

Includes MRE from #4842 as an integration test.


fkiraly:
I understood what's wrong - the `SameLocSplitter` does not play well with hierarchical data and broadcasting from `BaseSplitter`.

For now I've hacked the base class to see what the tests say, but the principled way would be:

* introduce a broadcasting tag (does the splitter need to broadcast?) for splitters`
* if the tag says "don't broadcast", carry out what is now the class specific hack
* add the tag to `SameLocSplitter` in the respective pull request.

fkiraly:
> For now I've hacked the base class to see what the tests say, but the principled way would be:

The principled way is now part of #4851, and this depends on it. Hack is gone.

PR #4862, opened 2023-07-09
[ENH] test-plus-train splitter compositor

fkiraly:
This PR adds a composite splitter which takes any splitter and changes its test splits to be the union of respective test plus train split.

Related issues and PR:
https://github.com/sktime/sktime/issues/4842
https://github.com/sktime/sktime/pull/4851
https://github.com/sktime/sktime/pull/4861

### DIFF
--- a/sktime/forecasting/model_selection/_split.py
+++ b/sktime/forecasting/model_selection/_split.py
@@ -1358,8 +1358,8 @@ class TestPlusTrainSplitter(BaseSplitter):
     def get_n_splits(self, y: Optional[ACCEPTED_Y_TYPES] = None) -> int:
         """Return the number of splits.
 
-        Since this splitter returns a single train/test split,
-        this number is trivially 1.
+        This will always be equal to the number of splits
+        of ``self.cv`` on ``y``.
 
         Parameters
         ----------
--- a/sktime/forecasting/model_selection/_split.py
+++ b/sktime/forecasting/model_selection/_split.py
@@ -1391,10 +1391,8 @@ class TestPlusTrainSplitter(BaseSplitter):
             `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
             `create_test_instance` uses the first (or only) dictionary in `params`
         """
-        from sktime.datasets import load_airline
         from sktime.forecasting.model_selection import ExpandingWindowSplitter
 
-        y = load_airline()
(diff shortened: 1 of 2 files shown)


sktime/forecasting/model_selection/_split.py
```diff
@@ -1301,6 +1302,104 @@ def get_test_params(cls, parameter_set="default"):
...
+        train : 1D np.ndarray of dtype int
+            Training window indices, iloc references to training indices in y
+        test : 1D np.ndarray of dtype int
+            Test window indices, iloc references to test indices in y
+        """
+        cv = self.cv
+
+        for y_train_inner, y_test_inner in cv.split(y):
+            y_train_self = y_train_inner
+            y_test_self = set(y_train_inner).union(y_test_inner)
```
benHeid, line 1354: I would prefer to use `np.union1d` here, since you use always 1d numpy arrays here. A short test on my laptop says that this is three times faster.

sktime/forecasting/model_selection/_split.py
```diff
@@ -1301,6 +1302,104 @@ def get_test_params(cls, parameter_set="default"):
...
+            y_test_self = np.array(list(y_test_self))
+            yield y_train_self, y_test_self
+
+    def get_n_splits(self, y: Optional[ACCEPTED_Y_TYPES] = None) -> int:
+        """Return the number of splits.
+
+        Since this splitter returns a single train/test split,
+        this number is trivially 1.
```
benHeid, line 1362: I do not understand this comment, what is the referring to? The return value can be any positive integer depending on the parameters of the inner cv.

benHeid (approved):
Thank you for the PR. I have a change request and a question.

sktime/forecasting/model_selection/_split.py
fkiraly: typo and incorrect, copy-paste accident. Will fix.

sktime/forecasting/model_selection/_split.py
fkiraly: well, I did not know this existed! Thanks for the learning moment and improvement suggestion.

fkiraly:
review comments addressed, kindly re-review

benHeid (approved):
Thank you for addressing the comments :)

PR #4865, opened 2023-07-11
[ENH][MNT] Interface StatsForecast MSTL

luca-miniati:
#### Reference Issues/PRs
Accidentally closed #4805, this PR replaces it

Towards `statsforecast` umbrella issue #4643

Interface for StatsForecast MSTL

Tests found in `get_test_params`, passed all tests in `check_estimator`


docs/source/api_reference/forecasting.rst
```diff
@@ -211,6 +211,18 @@ All "ARIMA" and "Auto-ARIMA" models below include SARIMAX capability.
 
     ARIMA
 
+.. currentmodule:: sktime.forecasting.statsforecast
```
fkiraly, line 214: I think this is a copy-paste accident - I moved the estimators around in a previous PR. MSTL belongs with STL rather than ARIMA.

sktime/forecasting/base/adapters/_generalised_statsforecast.py
```diff
@@ -254,3 +255,184 @@ def _predict_interval(self, fh, X, coverage):
         final_interval_predictions = pandas.concat(interval_predictions, copy=False)
 
         return final_interval_predictions
+
+
+class StatsForecastBackAdapter(BaseForecaster):
```
fkiraly, line 260: I think the backadapter should inherit from the `statsforecast` forecaster rather than the `sktime` one? If it makes a difference.

sktime/forecasting/statsforecast.py
```diff
@@ -552,3 +554,124 @@ def get_test_params(cls, parameter_set="default"):
...
+
+        Returns
+        -------
+        params : dict or list of dict, default = {}
+            Parameters to create testing instances of the class
+            Each dict are parameters to construct an "interesting" test instance, i.e.,
+            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+            `create_test_instance` uses the first (or only) dictionary in `params`
+        """
+        from sktime.forecasting.theta import ThetaForecaster
```
fkiraly, line 658: FYI, this is what raises the current exception, as `ThetaForecaster` and `VAR` have a `statsmodels` dependency.

You can work with `_check_soft_dependencies` to have them as part of the test param set only if `statsmodels` is available, or replace with an estimator that doesn't have the `statsmodels` dependency, e.g., `NaiveForecaster`.

sktime/forecasting/base/adapters/_generalised_statsforecast.py
luca-miniati: Ok, I changed it to inherit from `statsforecast` base forecaster. In both nosoftdeps and nodevdeps, however, it raises exceptions of `statsforecast` not found. Do I make the same check that I did in this commit? -> [update get_test_params StatsForecastMSTL](https://github.com/sktime/sktime/pull/4865/commits/d41565e48ca076b43b724a7ddf238bd9a01e75ef)
Further, did that commit match what you meant by using `_check_soft_dependencies` for nosoftdeps exception?

fkiraly: > In both nosoftdeps and nodevdeps, however, it raises exceptions of statsforecast not found.

Ah, that has to do with checks for dependency isolation when `statsforecast` is not present.

Imagine a user of `sktime` who has no dependencies (and no `statsforecast`, they would now experience an import exception upon using `sktime`, because this one module imports `statsforecast`. We have to isolate soft dependencies, to avoid users of `sktime` without all the dependencies installed get a crash.

The general procedure is to ensure `statsforecast` gets imported only if it is in the environment. Estimators take care of this in the base class automatically if we use the tags correctly.

This is now an unusual case though, since we import a class and inherit from it, so we need to get a little creative.

Here's one way we could do it, although it might not be the bets way (it is not a standard situation).

1. introduce a preamble using `_check_soft_dependencies` (with severity being `"none"`). If `statsforecast` is present, import the class. If not, define a class of the same name.
2. this will result in the class we want if `statsforecast` is present, and a class inheriting from a dummy that does not import `statsforecast` on module load. Now this class will likely be broken, but it will also not be used anywhere (because `statsforecast` is missing).

What do you think about this solution? It is similar to what we are doing for `numba`/.
`numba` uses decorators heavily, so they sit flat in the module. If `numba` is not present, we replace the decorators by dummies.

luca-miniati: Gotcha, that solution makes a lot of sense. Thanks!

fkiraly:
FYI, the failure in the test generation is since you used `set_tags` not as per the docstring - you should pass a kwargs dict.

docs/source/api_reference/forecasting.rst
```diff
@@ -211,6 +219,17 @@ All "ARIMA" and "Auto-ARIMA" models below include SARIMAX capability.
...
+
+.. autosummary::
+    :toctree: auto_generated/
+    :template: class.rst
+
+    StatsForecastAutoARIMA
+    StatsForecastAutoCES
+    StatsForecastAutoETS
+    StatsForecastAutoTheta
+
```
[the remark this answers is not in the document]
yarnabrina:
```suggestion
```

sktime/forecasting/base/adapters/_generalised_statsforecast.py
```diff
@@ -254,3 +271,197 @@ def _predict_interval(self, fh, X, coverage):
...
+            Clean time series of shape (t, ).
+        X : typing.Optional[numpy.ndarray], default=None
+
+        Returns
+        -------
+        self : returns an instance of self.
+        """
+        from sklearn.base import clone
+
+        self.in_sample_fh = np.flip(-np.arange(len(y)))
```
yarnabrina, line 331: Repeating my previous comment from previous PR, as far as I know `sktime` allows in-sample prediction even before the oldest known data point. This is probably enforcing a different behaviour than that, and I'm not sure if that is okay.

(I personally can't think of why someone will need that, but not sure if it's part of any contract or not. CC @fkiraly.)

sktime/forecasting/base/adapters/_generalised_statsforecast.py
```diff
@@ -254,3 +271,197 @@ def _predict_interval(self, fh, X, coverage):
...
+        estimator = clone(self.estimator)
+        clone_self = self.new()
+        clone_self.estimator = estimator.fit(y=y, X=X)
```
yarnabrina, line 334: Clarifying a doubt: at this point, `self.estimator` is not trained then? Only, `clone_self.estimator` is fitted? Is that correct understanding?

sktime/forecasting/base/adapters/_generalised_statsforecast.py
```diff
@@ -254,3 +271,197 @@ def _predict_interval(self, fh, X, coverage):
...
+        mean = self.estimator.predict(fh=range(1, h + 1), X=X)[:, 0]
+        if level is None:
+            return {"mean": mean}
+
+        level = sorted(level)
+        coverage = [round(1 - (_l / 100), 2) for _l in level]
+
+        pred_int = self.estimator.predict_interval(
+            fh=range(1, h + 1), X=X, coverage=coverage
+        )
```
yarnabrina, line 365: If `self.estimator` is not fitted (based on above clarification), I'm a bit confused how is it working here. Can you please explain?

sktime/forecasting/base/adapters/_generalised_statsforecast.py
```diff
@@ -254,3 +271,197 @@ def _predict_interval(self, fh, X, coverage):
...
+
+class StatsForecastBackAdapter(_TS):
+    """StatsForecast Back Adapter.
+
+    StatsForecastBackAdapter is a wrapper for sktime forecasters to be used in
+    StatsForecast composite models.
+
+    Parameters
+    ----------
+    estimator : sktime estimator
```
yarnabrina, line 284:
```suggestion
    estimator : sktime forecaster
```

Non-forecaster estimators will not work here.

sktime/forecasting/base/adapters/_generalised_statsforecast.py
```diff
@@ -254,3 +271,197 @@ def _predict_interval(self, fh, X, coverage):
...
+        return {
+            "mean": mean,
+            **{
+                f"lo-{_l}": pred_int[("Coverage", c, "lower")].values
+                for c, _l in zip(reversed(coverage), reversed(level))
+            },
+            **{
+                f"hi-{_l}": pred_int[("Coverage", c, "upper")].values
+                for c, _l in zip(coverage, level)
+            },
```
yarnabrina, line 376: So, I believe this part works as of now only for univariate data? As far as I know, `Coverage` will be present only in case of univairate, and not if there are multiple targets. Have you tried passing multivariate data? How does it work then?

Also, there is a PR which changes this "Coverage" behaviour soon. So, it's probably better to change to at least that format.

sktime/forecasting/base/adapters/_generalised_statsforecast.py
```diff
@@ -254,3 +271,197 @@ def _predict_interval(self, fh, X, coverage):
...
+        level : typing.Optional[typing.Tuple[int]]
+            Confidence levels (0-100) for prediction intervals.
+
+        Returns
+        -------
+        y_pred : dict
+            Dictionary with entries mean for point predictions and level_* for
+            probabilistic predictions.
+        """
+        if level and not hasattr(self.estimator, "predict_interval"):
```
yarnabrina, line 393: Can a `sktime` forecaster lack `predict_interval`?

Also, if you really want to check, you need to check for `callable` as well.

sktime/forecasting/base/adapters/_generalised_statsforecast.py
```diff
@@ -254,3 +271,197 @@ def _predict_interval(self, fh, X, coverage):
...
+            raise Exception(
+                "The provided trend_forecaster does not support interval prediction."
+                "Please remove the 'level' parameter or use a different model that"
+                "supports it."
+            )
```
yarnabrina, line 398: We should not raise such a broad exception class. Please raise `ValueError` or something that is more appropriate.

sktime/forecasting/base/adapters/_generalised_statsforecast.py
```diff
@@ -254,3 +271,197 @@ def _predict_interval(self, fh, X, coverage):
...
+
+        fitted = self.estimator.predict(fh=self.in_sample_fh, X=self.in_sample_X)[:, 0]
```
yarnabrina, line 400: Q1. If you are unsure about `predict_interval`, I think you should worry about `predict` as well. Or, don't worry about either.

Q2. Why are you picking only the first column from the predictions?

sktime/forecasting/base/adapters/_generalised_statsforecast.py
```diff
@@ -254,3 +271,197 @@ def _predict_interval(self, fh, X, coverage):
...
+            "fitted": fitted,
+            **{
+                f"fitted-lo-{_l}": pred_int[("Coverage", c, "lower")].values
+                for c, _l in zip(reversed(coverage), reversed(level))
+            },
+            **{
+                f"fitted-hi-{_l}": pred_int[("Coverage", c, "upper")].values
+                for c, _l in zip(coverage, level)
+            },
+        }
```
yarnabrina, line 421: Same comment as above "Coverage" area.

Also, can you create an internal method to reduce duplication?

sktime/forecasting/base/adapters/_generalised_statsforecast.py
```diff
@@ -254,3 +271,197 @@ def _predict_interval(self, fh, X, coverage):
...
+
+        Returns
+        -------
+        params : dict or list of dict, default = {}
+            Parameters to create testing instances of the class
+            Each dict are parameters to construct an "interesting" test instance, i.e.,
+            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+            `create_test_instance` uses the first (or only) dictionary in `params`
+        """
+        from sktime.utils.validation._dependencies import _check_soft_dependencies
```
yarnabrina, line 442: This is already imported in L13, if I'm not wrong.

sktime/forecasting/base/adapters/_generalised_statsforecast.py
```diff
@@ -1,12 +1,29 @@
...
+else:
+
+    class _TS:
+        def __init__(self):
+            warnings.warn(
+                "The 'statsforecast' module is not available. Please ensure that"
+                f"'statsforecast' is installed to use {type(self)}.",
+                stacklevel=1,
+            )
+
```
yarnabrina, line 26: Can you please explain why is this required? Why is the tag insufficient?

sktime/forecasting/base/adapters/_generalised_statsforecast.py
```diff
@@ -254,3 +271,197 @@ def _predict_interval(self, fh, X, coverage):
         final_interval_predictions = pandas.concat(interval_predictions, copy=False)
 
         return final_interval_predictions
+
+
+class StatsForecastBackAdapter(_TS):
```
yarnabrina, line 276: Related to above comment, is it necessary to inherit from `_TS`?

First of all it's private I think (from starting _), plus you're only using `new` method from it as far as I can see, and I did not understand why is that necessary. Can you please explain?

yarnabrina (changes requested):

sktime/forecasting/statsforecast.py
```diff
@@ -552,3 +555,150 @@ def get_test_params(cls, parameter_set="default"):
...
+        Number of observations per unit of time. For multiple seasonalities use a
+        list.
+    trend_forecaster : estimator, optional, default=StatsForecastAutoETS()
+        Sktime estimator used to make univariate forecasts. Multivariate estimators are
+        not supported.
+
+    References
+    ----------
+    .. [1]
+        https://nixtla.github.io/statsforecast/src/core/models.html#multiple-seasonalities
```
yarnabrina, line 583:
```suggestion
        https://nixtla.github.io/statsforecast/src/core/models.html#mstl
```

It's better to link the model directly.

sktime/forecasting/statsforecast.py
```diff
@@ -552,3 +555,150 @@ def get_test_params(cls, parameter_set="default"):
...
+    >>> model = StatsForecastMSTL(season_length=[3,12]) # doctest: +SKIP
+    >>> fitted_model = model.fit(y=y) # doctest: +SKIP
+    >>> y_pred = fitted_model.predict(fh=[1,2,3]) # doctest: +SKIP
+    """
+
+    _tags = {
+        "ignores-exogeneous-X": True,
+        "capability:pred_int": True,
+        "capability:pred_int:insample": True,
+        "python_dependencies": ["statsmodels", "statsforecast"],
```
yarnabrina, line 600: Can you please again why does it need `statsmodels`?

sktime/forecasting/statsforecast.py
```diff
@@ -552,3 +555,150 @@ def get_test_params(cls, parameter_set="default"):
...
+    ):
+        super().__init__()
+
+        if _check_soft_dependencies("statsmodels", severity="none"):
+            from statsforecast.models import _TS
+        else:
+            raise Exception(
+                "StatsForecastMSTL requires module `statsforecast` to be installed. "
+                "Please install `statsforecast` in your environment."
+            )
```
yarnabrina, line 616: I am not sure how `_TS` is being used here, but checking for `statsmodels` and importing `statsforecast` is not right.

yarnabrina: Also, one should never raise such a broad exception. If I'm not wrong, `_check_soft_dependencies` allow raising error itself.

sktime/forecasting/statsforecast.py
```diff
@@ -552,3 +555,150 @@ def get_test_params(cls, parameter_set="default"):
...
+
+        self.trend_forecaster = trend_forecaster
+        self.season_length = season_length
+        if trend_forecaster:
+            self._trend_forecaster = trend_forecaster
+        else:
+            self._trend_forecaster = StatsForecastAutoETS(model="ZZN")
```
yarnabrina, line 623: I think `self.trend_forecaster` should always match used trend forecaster.

sktime/forecasting/statsforecast.py
```diff
@@ -552,3 +555,150 @@ def get_test_params(cls, parameter_set="default"):
...
+
+        # checks if trend_forecaster is already wrapped with
+        # StatsForecastBackAdapterMSTL
```
yarnabrina, line 626: Where does this class exist? Or, is it a typo?

sktime/forecasting/statsforecast.py
```diff
@@ -552,3 +555,150 @@ def get_test_params(cls, parameter_set="default"):
...
+                self._trend_forecaster = StatsForecastBackAdapter(
+                    self._trend_forecaster
+                )
+            # if trend_forecaster is not StatsForecast forecaster
+            elif not isinstance(self._trend_forecaster, _TS):
+                raise Exception(
+                    "The provided forecaster is not compatible with MSTL. Please ensure"
+                    " that the forecaster you pass into the model is a sktime "
+                    "forecaster or statsforecast forecaster."
+                )
```
yarnabrina, line 639: Too broad exception class. Use `TypeError`.

sktime/forecasting/statsforecast.py
```diff
@@ -552,3 +555,150 @@ def get_test_params(cls, parameter_set="default"):
...
+        # checks if trend_forecaster is already wrapped with
+        # StatsForecastBackAdapterMSTL
+        if not isinstance(self._trend_forecaster, StatsForecastBackAdapter):
+            # if trend_forecaster is sktime forecaster
+            if isinstance(self._trend_forecaster, BaseForecaster):
+                self._trend_forecaster = StatsForecastBackAdapter(
+                    self._trend_forecaster
+                )
+            # if trend_forecaster is not StatsForecast forecaster
+            elif not isinstance(self._trend_forecaster, _TS):
```
yarnabrina, line 634: Can you please explain why `_TS` is required?

sktime/forecasting/statsforecast.py
```diff
@@ -552,3 +555,150 @@ def get_test_params(cls, parameter_set="default"):
...
+            season_length=self.season_length,
+            trend_forecaster=self._trend_forecaster,
+        )
+
+    def __getattr__(self, name):
+        """Return self._trend_forecaster when self.trend_forecaster is accessed."""
+        if name == "trend_forecaster":
+            return self._trend_forecaster
+        else:
+            return self.__getattribute__(name)
```
yarnabrina, line 655: If this is the behaviour, why not assign as such in the `__init__` directly? I'm not following the logic, can you please explain?

yarnabrina (changes requested):

sktime/forecasting/base/adapters/_generalised_statsforecast.py
luca-miniati: I just assumed it was necessary to do this, I think you're right that it's not necessary

sktime/forecasting/base/adapters/_generalised_statsforecast.py
```diff
@@ -254,3 +271,197 @@ def _predict_interval(self, fh, X, coverage):
...
+    >>> model = StatsForecastMSTL( # doctest: +SKIP
+            season_length=[3,12],
+            trend_forecaster=trend_forecaster
+        )
+    >>> fitted_model = model.fit(y=y) # doctest: +SKIP
+    >>> y_pred = fitted_model.predict(fh=[1,2,3]) # doctest: +SKIP
+    """
+
+    _tags = {
+        "python_dependencies": ["statsforecast", "statsmodels"],
```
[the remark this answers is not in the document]
luca-miniati: Yeah I put it for testing, I'm removing it

sktime/forecasting/base/adapters/_generalised_statsforecast.py
```diff
@@ -254,3 +271,197 @@ def _predict_interval(self, fh, X, coverage):
...
+    def __repr__(self):
+        return "StatsForecastBackAdapter"
+
+    def fit(self, y, X=None):
+        """Fit to training data.
+
+        Parameters
+        ----------
+        y : ndarray
+            Clean time series of shape (t, ).
```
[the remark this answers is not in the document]
luca-miniati: I copy pasted that from statsforecast, I'm guessing it means no missing values? Should it be instead:
y : ndarray
    Time series of shape (t, ) without missing values

sktime/forecasting/base/adapters/_generalised_statsforecast.py
luca-miniati: Yes, I did this so as to not overwrite `estimator`, a parameter of the adapter. However, if those tests (checking if parameters are overwritten) are not run on the adapter because it doesn't inherit from any base sktime class, I agree that `self.estimator` should be trained. I'll try and see if training `self.estimator` works.

sktime/forecasting/base/adapters/_generalised_statsforecast.py
luca-miniati: Yes, this is definitely wrong. Thank you for pointing this out!

sktime/forecasting/base/adapters/_generalised_statsforecast.py
luca-miniati: Internally, `statsforecast` MSTL doesn't support multivariate data, it simply takes first column if it's given multivariate data. Also, what PR are you referencing for the updated behavior of `Coverage`? If it's #4880, I think this won't be a problem, as multivariate forecasts are not supported?

sktime/forecasting/base/adapters/_generalised_statsforecast.py
luca-miniati: Yes, method for this makes sense

sktime/forecasting/statsforecast.py
luca-miniati: You're right, it definitely does not

sktime/forecasting/statsforecast.py
luca-miniati: I did this because `statsforecast` models are still compatible with `statsforecast` MSTL. Should I get rid of this, to disallow the option of using a `statsforecast` model here?

sktime/forecasting/statsforecast.py
luca-miniati: I did this because fitting `self.trend_forecaster` would be overwriting class params, which afaik is not allowed because of a contract

sktime/forecasting/statsforecast.py
luca-miniati: See explanation [below](https://github.com/sktime/sktime/pull/4865/files/54f9efa607a5b6173e8b6c70a8aba41d6df9503c#r1266093273)

sktime/forecasting/statsforecast.py
luca-miniati: Typo on the `statsmodels` check
Also now using "error" `severity` of `_check_soft_dependencies`

sktime/forecasting/base/adapters/_generalised_statsforecast.py
luca-miniati: Q1: Gotcha, missed that. Adding check for `predict`

Q2: Output of `predict` gives output of shape [t, 1] which doesn't match expected shape of [t] 
(I'm getting this from `statsforecast` docs, their example has this shape. Not sure if this is an important detail or not)

sktime/forecasting/base/adapters/_generalised_statsforecast.py
luca-miniati: This is related to [this comment](https://github.com/sktime/sktime/pull/4865/files/54f9efa607a5b6173e8b6c70a8aba41d6df9503c#r1266092620)
I put this in case they use `statsforecast` forecasters, which don't have `predict_interval`. Just realizing that I would just use `predict` in this case though. 

So I propose one of 2 options:
          1. remove option for use of `statsforecast` forecaster as `trend_forecaster`
          2. if `trend_forecaster` is `statsforecast` forecaster, use `predict`

sktime/forecasting/base/adapters/_generalised_statsforecast.py
luca-miniati: Not sure why I did this, now the adapter uses `estimator._y` and `estimator._X` for in-sample values

sktime/forecasting/statsforecast.py
fkiraly: @yarnabrina is right, I think this is dangerous to do, and will break the `set_params` / `get_params` logic! The tests would tell you if `get_test_params` would return two parameter sets (I'd recommend, add a second one so all `set_params` / `get_params` tests run).

The "standard protocol" would be:

* in `__init__`, write `trend_forecaster` to `self.trend_forecaster`
* in `__init__` (or anytime before the logic in `_fit`), write a clone of `trend_forecaster` to `self._trend_forecaster`
* if you need a default, e.g., for `trend_forecaster = None`, write the default to `self._trend_forecaster` in the `None` case
* never access `self.trend_forecaster` after this
* whenever you need the estimator that needs to mutate state, e.g., by fitting, use `self._trend_forecaster`

sktime/forecasting/base/adapters/_generalised_statsforecast.py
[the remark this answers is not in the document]
yarnabrina: That seems much better to me 👍🏽

sktime/forecasting/base/adapters/_generalised_statsforecast.py
yarnabrina: The way I'm thinking right now is that if a user created a direct `statsforecast` forecaster, they are much more likely to use `MSTL` directly instead of the back-adapter from `sktime`. So, I'll suggest to assume `trend_forecaster` being only those which inherits from `BaseForecaster`. This can be checked even in `__init__`, and fail otherwise. Afterwards, then there will be no more need of checking existence of `predict` or `predict_interval` or etc.

@fkiraly Let me know if you disagree.

sktime/forecasting/base/adapters/_generalised_statsforecast.py
yarnabrina: When I read this code, I imeddiately thought may be this is why you're getting just one column in case of multivariate. I'll do some validation on my end to give an updated comment, but won't have time before weekend for work pressure.

sktime/forecasting/statsforecast.py
luca-miniati: Gotcha, thanks guys!

sktime/forecasting/base/adapters/_generalised_statsforecast.py
luca-miniati: Ok, I've set the logic to only allow `sktime` forecasters, and I've removed the checks on `predict` and `predict_interval`.

sktime/forecasting/base/adapters/_generalised_statsforecast.py
luca-miniati: Ok, if this helps, here is the code that led me to believe `MSTL` only takes univariate: 
https://github.com/Nixtla/statsforecast/blob/main/statsforecast/mstl.py, L28-L29

fkiraly: if only univariate is supported, the tag `scitype:y` should be set to `"univariate"` - no extra logic to deal with it is needed

fkiraly: `sktime` will per default broadcast across columns

fkiraly (approved):
Looks ready to go! Excellent!

fkiraly:
@yarnabrina, are your comments resolved?

yarnabrina:
Sorry, I didn't have time for a thorough review after all the changes, but since all tests are passing, no new comments from my side.

There is one open issue regarding use of `Coverage` and `Quantiles`. If you're sure that it's not going to cause any problems post deprecation ( regarding `legacy_interface` argument), it's all good 👍🏽

fkiraly:
> There is one open issue regarding use of Coverage and Quantiles. If you're sure that it's not going to cause any problems post deprecation ( regarding legacy_interface argument)

Ah yes, I'll deal with that in a separate PR.

PR #4867, opened 2023-07-11
& fkiraly [MNT] Extend softdep error message tests support for packages with version speciefier and alias

hazrulakmal:
In response to #4820 and maintainance in #4860,

The current test_softdepds does not account for package versions and aliases. It only checks if dependencies can be imported using the package name via `import_modules` that only captures cases like "pmdarima" but not "pmdarmia<1.3". This resulted in failures if the **python_dependencies** tag had a package version bound or a different import name. This PR leverages `_check_soft_dependencies` to account for the drawbacks.


fkiraly (approved):
Thanks! This finishes what I started.

PR #4869, opened 2023-07-11
[MNT] fix test for soft dependency checks to account for PEP 440 specifier tags

fkiraly:
This fixes an unreported bug where soft dependency handling checks in `test-nosoftdeps` did not account for full PEP 440 specifiers. The estimators would correctly raise an error, but it would not be detected as an expected error, due to string mismatch with the error message.

This is fixed by upgrading the checks for expected strings in the error message to the case of PEP 440 specifiers.

Also updates the error message raised by the test.


fkiraly:
superseded by https://github.com/sktime/sktime/pull/4867, opened by accident/confusion

PR #4871, opened 2023-07-12
[ENH] Deprecate pseudo-private DL deserializing methods in favour of public interface

### LINKED ISSUE #4740: [ENH] Deprecate misleading DL deserializing methods in favour of private methods and improve their documentation
We want to avoid cases like #4706 where users inadvertently use methods like `load_from_path` and `load_from_serial` to deserialize their estimators. We also found that the docstrings for these methods are not complete and need to be updated.

The most obvious way is to add privatized versions of these functions and follow the deprecation policy to remove the current public ones. 

I'd want things to stay Pythonic - deserializing estimators should have only public API since having these lower-level methods as public only serves the purpose of confusing users. I'm open to diverging opinions if anyone has any.

achieveordie:
#### Reference Issues/PRs

Fixes #4740

The two class methods will be deprecated in 0.20.2. I think this should provide the users time to upgrade their code accordingly.
 
None

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/base/_base.py
+++ b/sktime/base/_base.py
@@ -152,6 +152,30 @@ class BaseObject(_BaseObject):
         deserialized self resulting in output `serial`, of `cls.save(None)`
         """
         import pickle
+        from warnings import warn
+
+        warn(
+            "`load_from_serial()` will be deprecated in 0.20.2 in the favor of load() "
+            "which is the standard way of deserializing estimators.",
+            category=DeprecationWarning,
+            stacklevel=2,
+        )
+
+        return pickle.loads(serial)
+
+    @classmethod
+    def _load_from_serial(cls, serial):
+        """Private helper method to load object from serialized memory container.
+
+        Parameters
+        ----------
+        serial : 1st element of output of `cls.save(None)`
+
+        Returns
+        -------
+        deserialized self resulting in output `serial`, of `cls.save(None)`
+        """
+        import pickle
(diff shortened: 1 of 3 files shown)


fkiraly (changes requested):
Thanks!

Instead of copy-pasting multiple times, can you kindly let the public method point to the private one in the base, and rename everywhere else?

achieveordie:
Excuse me for the messy history, I was switching to a different working environment.

@fkiraly It is not exactly the change you requested, but an experimental one, do take a look at it.

The concept is simple, for every distinct deprecation call (`warnings.warn()`), we create a function that is to be used as a decorator in the actual methods. For now, I've only added it to base deep classifiers to demonstrate how it will be used. 

There are two main advantages of using such a system:
1. All the upcoming deprecations are consolidated in a single file (how it is done currently?).
2. We avoid writing multiple deprecations warnings/methods in case the same deprecation is done over multiple objects (this case).

Should we choose to use such a pipeline, I shall properly write more docstring (for future developers) and port other deprecation warnings into a single file (in a future PR).

achieveordie:
(I will be closing this PR and will open a new one either way)

fkiraly:
@achieveordie, I thought this was perfectly fine (private methods, warning in public methods, no decorator) before I assume an accident happened?

We need to push this to 0.22.0 then?

Issue #4848, opened 2023-07-07
[ENH] prioritizing `make_reduction` related items from davidgilbertson use case


fkiraly:
@davidgilbertson, thanks a lot for the various bug reports, improvement ideas, and code snippets!

May we kindly ask, if you were to set a priority from the entire space, would it be still this? https://github.com/sktime/sktime/issues/4776

That is, ensure that the following use case is supported:

* direct reduction with arbitrary `fh`
* with exogeneous data `X`
* `X` features including categorical variables
* user specified lags on `X`, or no lags on `X` (choice)
* user specified feature set used in the `sklearn` estimator
* global and local should both work
* optinally: option to apply further transformations to `X`

requirements:
* variable names of transformed variables (or equivalent information) must be easily specifiable, optimally also inspectable
* performance must be decent (comparable to current `make_reduce` with similar estimator without specifying categorical variables
* `get_fitted_params` should be inspectable with clear, identifiable, reference to transformed variables, e.g., when `LinearRegressor` is used

The above being a combination of feature requests (e.g., variable name inspection) and bugfix (potential ways to treat this combination case don't fully work yet).

fkiraly:
Oh, @davidgilbertson, the following would also be much appreciated:

* if you would be available for a 15-30min slot during the 2023 community & developer days at europython to explain your primary use case, that would be great (I assume virtual is preferred, but in-person in Prague is also possible), that would be much appreciated. We'll also take your user feedback form with suggestions obviously, even if you can't attend.
* are you on discord? Would be great if you would join https://discord.com/invite/54ACzaFsn7 for some unstructured tech chat around your use case. If you look into the meet-up channel and the hackmd reference today, you can see some discussion around the column names issue.

davidgilbertson:
I think you've summed up my use case well.

As for whether it's a priority... As I've dived deeper and tried different things I've been creating more and more components of my own, to get certain things done. This is a pity because I then lose a lot of the sktime magic (for example I got my own `LagYIntoX` working fine, but then wanted to try `pooling="global"` but of course that doesn't work because I haven't implemented it).

I know there's been discussions around callbacks, and I think these would open up sktime to many more use cases, without having to explicitly allow for those cases.

[...]

PR #4876, opened 2023-07-13
[ENH] set `ForecastX` missing data handling tag to `True` to properly cope with future unknown variables

fkiraly:
This PR sets the `ForecastX` missing data handling tag to `True` to properly cope with future unknown variables.

An edge case - or, arguably, a general case if none of the future variables are known - is showcased in this issue discussion:
https://github.com/sktime/sktime/issues/4848

More precisely, `ForecastX` should always allow for `nans` in future `X`, as these may be in places where there is genuinely no knowledge about the future. Alternatively, dummy values have to be provided which sounds odd.

FYI @davidgilbertson

### DIFF
--- a/sktime/forecasting/compose/_pipeline.py
+++ b/sktime/forecasting/compose/_pipeline.py
@@ -1234,6 +1234,7 @@ class ForecastX(BaseForecaster):
         "ignores-exogeneous-X": False,
         "capability:pred_int": True,
         "capability:pred_int:insample": True,
+        "handles-missing-data": True,
     }
 
     def __init__(


fkiraly:
quick merge for release

yarnabrina:
@fkiraly Not opposing the merge (at least not yet), but asking for a clarification about the tag below.

Based on `extension_templates/forecasting.py`, this is what the tag means.

```python
        # handles-missing-data = can estimator handle missing data?
        "handles-missing-data": False,
        # valid values: boolean True (yes), False (no)
        # if False, raises exception if y or X passed contain missing data (nans)
```

(If it's not the official tag reference for users or for developers, kindly share the proper reference.)

From the tag definition above, I interpret that if it's set as True, it should be able to handle missing value in either `y` or `X`, and not just `X`. Based on my understanding of `ForecastX`, that is not the case. Even if I ignore the `y` case leaving up to the forecaster of choice, it does need all values to be present for columns that are not covered by `columns`, assuming that is non-empty and does not cover all columns of `X`.

So, based on my interpretation, either addition of this tag gives a possibly wrong information to users, or the tag definition is not quite clear. Can you please confirm which one is it? Or, if I'm just completely off, please let me know.

fkiraly:
@yarnabrina, happy to unmerge this - my quick merge was based on that it is functionally inconsequential in cases that work as present, while unlocking what I felt was a common use case, important enough to have it in the release quickly.

To reply to your points:

* you are right as to that the extension template is currently the best reference for the tags.
* the crux is that the tag specification is valid only for atomic (non-composite) estimators, and may become misleading in the case of composites.
    * for composites, bug reports have repeatedly indicated that in case of ambiguity, tags should be set to the most permissive value (i.e., generating least, not most exceptions), which is a soft policy now (but nowhere written down)
    * why: imagine users build some composite which, from the methodological perspective should work - such as the example in #4848 which imo isn't that uncommon. However, the compositor blocks this by raising an exception based on tag/data checks.
    * the deeper reason is that the tag inference would need to be based on properties of components, and the actual property may depend in very complex ways not just on the tags of the components, but also on parts of the logic which are not inspectable as tags

yarnabrina:
I understand your point of allowing as much as possible to the composite estimators in order to not avoid failures from composite logic and leave that to the actual non-composite estimators. It makes sense, and so there's no need to unmerge.

However, personally I'd like no tags at all for composites in that case, as you said it doesn't make sense. I do not have any idea as of now on how to achieve that based on current base classes, so let's keep it as is. If I can think of anything, will create a new design issue later for discussion.

fkiraly:
> However, personally I'd like no tags at all for composites in that case, as you said it doesn't make sense.

I think that's stronger than the statement I wanted to make. What I tried to convey: the properties the tags indicate may depend on parameters (but that does not make them nonsensical overall)

The current design tries to set the tags dynamically, based on the parameters, for objects, and leaves the tag at the most general setting for classes.

> I do not have any idea as of now on how to achieve that based on current base classes, so let's keep it as is. If I can think of anything, will create a new design issue later for discussion.

Ok, would be appreciated - I think it's an interesting question.

Though that concludes the discussion for this PR at least as far as I understand.

Issue #4873, opened 2023-07-12
[ENH] Enhancements to kotsu Benchmarking Framework 


hazrulakmal:
A great initial effort was made by @DBCerigo @TNTran92 @alex-hh to create a user-friendly sktime benchmarking interface for quick univariate forecasting benchmarks. The decision to include Kotsu as part of the benchmarking module was made in issue #2777.

After the first pull request was merged, the work was then put on hold, but there are still future tasks to complete, as suggested by the original author in issue #2804.  I plan to continue the excellent work that has been done, and this issue is intended to keep track of what needs to be done.

Additionally, there were some issues and suggestions for improving the current implementation, which were discussed by @fkiraly , @marrov, and myself previously [here](https://hackmd.io/@GjA0yi6JRRO3HnLEUL3mNw/ry01xt_wn). Below is a summary of the discussion - please do add anything if you I miss something out.

1) `ForecastBenchmark` force writes results to the hard drive, but it should allow users to save results to different storage files (cloud) or formats (e.g. Parquet).
2) `ForecastBenchmark` return results in aggregated fold metrics, without providing intermediate results. Users have no control over further evaluation processing steps, such as adding new metrics to the benchmark after it has run. visualizing predictions, compute different aggregations like median.
3) `ForecastBenchmark` datasets loader assumes a callable from `sktime.datasets`. does this mean it only accept datasets that exist in the `sktime.datasets` ? is this intented? what about external datasets from the users?
4) The `estimator_id` is restricted to certain strings without clear user signposting. Should all strings be allowed?
5) Currently, the `add_estimator` method only takes a single estimator per call. Can be modified to accept a list of estimators.

My initial thoughts on the problems in chronological order

1) The Kotsu framework itself is responsible for results storage and forces results to be saved as CSV, no other option asfaik. Making it more flexible should be done within the Kotsu framework rather than in sktime.
2) I believe that `evaluate` itself has an option to return raw predictions, so we may need to make some tweaks to the current forecasting benchmarking to allow for this option.
3) Should we instead accept sktime in-memory data formats (scitype, mtype)?
4) The `estimator_id` naming convention should follow the format `<name>-v<version number>` (e.g. -v1.2), as per required by Kotsu. To allow for all kinds of strings, this has to be done in Kotsu.
5) I agree with this part!

Moving forwads this is what I wanna do
- [ ] implement storing estimator predictions for intermediate results
- [ ] support for univariate with exogenous
[...]

PR #4877, opened 2023-07-13
[ENH] input checks for `BaseBenchmark`, allow `add_estimator` to accept multiple estimators

hazrulakmal:
See #4873 

The current ForecastingBenchmark interface requires manually looping over multiple estimators to benchmark multiple models, as the add_estimator interface only accepts one estimator per call. This PR provides an option for users to add multiple models with a single add_estimator call while maintaining the existing interface. The interface for multiple models looks as follows:

```python
from sktime.benchmarking.forecasting import ForecastingBenchmark

benchmark = ForecastingBenchmark()

# for auto naming
models = [NaiveForecaster(), AutoARIMA()]
benchmark.add_estimator(models)

# for custom naming
models ={"Arima-v1":AutoARIMA(),
        "Naive-v1":NaiveForecaster()}
benchmark.add_estimator(models)

# existing interface still works
benchmark.add_estimator(NaiveForecaster())
benchmark.add_estimator(AutoARIMA())
```

### Questions
1. are there any other preferences in terms of the interface?
2. In sktime, are there any best practices regarding where to place util functions within the module? Should I create another folder called *_utils.py* to place any benchmarking util functions, or is the existing solution sufficient?

FYI @fkiraly @marrov

### DIFF
--- a/sktime/benchmarking/benchmarks.py
+++ b/sktime/benchmarking/benchmarks.py
@@ -10,7 +10,7 @@ from sktime.base import BaseEstimator
 from sktime.utils.validation._dependencies import _check_soft_dependencies
 
 
-def coer_estimator_and_id(estimators, estimator_id=None):
+def coerce_estimator_and_id(estimators, estimator_id=None):
     """Coerce estimators to a dict with estimator_id as key and estimator as value.
 
     Parameters
@@ -75,7 +75,7 @@ class BaseBenchmark:
         estimator_id : str, optional (default=None)
             Identifier for estimator. If none given then uses estimator's class name.
         """
-        estimators = coer_estimator_and_id(estimator, estimator_id)
+        estimators = coerce_estimator_and_id(estimator, estimator_id)
         for estimator_id, estimator in estimators.items():
             estimator = estimator.clone()  # extra cautious
             self.estimators.register(id=estimator_id, entry_point=estimator.clone)
--- a/sktime/benchmarking/benchmarks.py
+++ b/sktime/benchmarking/benchmarks.py
@@ -10,19 +10,8 @@ from sktime.base import BaseEstimator
 from sktime.utils.validation._dependencies import _check_soft_dependencies
 
 
-def is_initalised_estimator(estimator):
-    """Check if the estimator is an initialised BaseEstimator object.
-
-    Parameters
-    ----------
-    estimator : object
-        The estimator to check.
-
-    Returns
-    -------
-    is_estimator : bool
-        Whether the estimator is an initialised BaseEstimator object or not.
-    """
+def is_initalised_estimator(estimator: BaseEstimator) -> bool:
+    """Check if estimator is initialised BaseEstimator object."""
     if isinstance(estimator, BaseEstimator):
         return True
     return False
--- a/sktime/benchmarking/benchmarks.py
+++ b/sktime/benchmarking/benchmarks.py
@@ -68,8 +68,8 @@ class BaseBenchmark:
         ----------
         estimator : Dict, List or BaseEstimator object
             Estimator to add to the benchmark.
-            If Dict, keys are estimator_ids and values are estimators, use to customise
-            the identifier ID.
+            If Dict, keys are estimator_ids and values are estimators,
+            use to customise the identifier ID.
             If List, each element is an estimator. estimator_ids are generated
             automatically using the estimator's class name.
         estimator_id : str, optional (default=None)
--- a/sktime/benchmarking/benchmarks.py
+++ b/sktime/benchmarking/benchmarks.py
@@ -10,6 +10,40 @@ from sktime.base import BaseEstimator
 from sktime.utils.validation._dependencies import _check_soft_dependencies
 
 
+def is_initalised_estimator(estimator):
+    """Check if the estimator is an initialised BaseEstimator object.
+
+    Parameters
+    ----------
+    estimator : object
+        The estimator to check.
+
+    Returns
+    -------
+    is_estimator : bool
+        Whether the estimator is an initialised BaseEstimator object or not.
+    """
+    if isinstance(estimator, BaseEstimator):
+        return True
+    return False
+
+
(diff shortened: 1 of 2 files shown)


hazrulakmal:
> We should also test for it though?

Done add test for checking registered multiple estimators here [75d4d93](https://github.com/sktime/sktime/pull/4877/commits/75d4d93d0c5b729bfd787726e9f54f20dab1bafc)

hazrulakmal:
Hi @fkiraly @benHeid @yarnabrina @achieveordie, can anyone retrigger the CI workflow or merge this PR if there's no objection? Thanks!

the failure is unrelated to this PR.

sktime/benchmarking/benchmarks.py
```diff
@@ -10,6 +10,39 @@
 from sktime.utils.validation._dependencies import _check_soft_dependencies
 
 
+def coer_estimator_and_id(estimators, estimator_id=None):
```
yarnabrina, line 13: Real nitpicky non-blocker suggestion: can you rename it to `coerce` from `coer` (in all relevant places)? Personally as a user/developer I'll appreciate seeing 2 more characters to readily understand what the function is supposed to do.

sktime/benchmarking/benchmarks.py
```diff
@@ -10,6 +10,39 @@
...
+    estimator_id : str, optional (default=None)
+        Identifier for estimator. If none given then uses estimator's class name.
+
+    Returns
+    -------
+    estimators : dict
+        Dict with estimator_id as key and estimator as value.
+    """
+    VERSION_ID = "-v1"
+    if isinstance(estimators, dict):
```
yarnabrina, line 29: Is it worth checking that all values are `BaseEstimator` objects, as per docstring?

sktime/benchmarking/benchmarks.py
```diff
@@ -10,6 +10,39 @@
...
+        return estimators
+    elif isinstance(estimators, list):
```
yarnabrina, line 31: Same as above.

Side comment: Pydantic helps to validate these things on it's own just by adding decorators, but I don't think we're ready to start using it. May be this can be discussed in one of the Friday calls.

sktime/benchmarking/benchmarks.py
```diff
@@ -10,6 +10,39 @@
...
+            f"{estimator.__class__.__name__ + VERSION_ID }": estimator
+            for estimator in estimators
+        }
+    elif isinstance(estimators, BaseEstimator):
+        estimator_id = estimator_id or f"{estimators.__class__.__name__+ VERSION_ID}"
+        return {estimator_id: estimators}
+    else:
+        raise ValueError(
+            f"estimator must be of a type a dict, list or\
+            BaseEstimator object but received {type(estimators)}"
```
yarnabrina, line 42:
```suggestion
            "estimator must be of a type a dict, list or "
            f"BaseEstimator object but received {type(estimators)}"
```

sktime/benchmarking/benchmarks.py
```diff
@@ -10,6 +10,39 @@
...
+    elif isinstance(estimators, list):
+        return {
+            f"{estimator.__class__.__name__ + VERSION_ID }": estimator
+            for estimator in estimators
+        }
+    elif isinstance(estimators, BaseEstimator):
+        estimator_id = estimator_id or f"{estimators.__class__.__name__+ VERSION_ID}"
+        return {estimator_id: estimators}
+    else:
+        raise ValueError(
```
yarnabrina, line 40:
```suggestion
        raise TypeError(
```

yarnabrina (changes requested):

sktime/benchmarking/benchmarks.py
hazrulakmal: ofcourse! you are right!

sktime/benchmarking/benchmarks.py
hazrulakmal: I added a validation function to check for the estimators object. Pls kindly review them again. 

I think most of the validations in sktime is done natively. adding pydantic means we are proposing adding pydantic as a dev dependency, right? Most likely need to go through meeting(s) with other core developers if im not mistaken.

yarnabrina: Agreed. I love to use pydantic, but for sktime it'd be a super big effort for sure. That's why made it a side note.

yarnabrina (approved):
Looks good to me. Added 2 minor non-blocking comments.

sktime/benchmarking/benchmarks.py
```diff
@@ -10,6 +10,63 @@
...
+    ------
+    TypeError
+        If any of the estimators are not BaseEstimator objects.
+    """
+    if isinstance(objs, BaseEstimator):
+        objs = [objs]
+    items = objs.values() if isinstance(objs, dict) else objs
+    compatible = all([is_initalised_estimator(estimator) for estimator in items])
+    if not compatible:
+        raise TypeError("Estimator must be an initialised BaseEstimator object.")
```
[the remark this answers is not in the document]
hazrulakmal: a bit difficult to do that in this form. How about providing the users with a better error message like "One or many estimator(s) is not an initialised BaseEstimator object(s). Please instantiate the estimator(s) first."

sktime/benchmarking/benchmarks.py
```diff
@@ -10,6 +10,63 @@
...
+
+    Raises
+    ------
+    TypeError
+        If any of the estimators are not BaseEstimator objects.
+    """
+    if isinstance(objs, BaseEstimator):
+        objs = [objs]
+    items = objs.values() if isinstance(objs, dict) else objs
+    compatible = all([is_initalised_estimator(estimator) for estimator in items])
```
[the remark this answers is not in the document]
hazrulakmal: yup it does. Thanks for pointing it out!

fkiraly (approved):
Thanks, request re tests has been addressed.

PR #4878, opened 2023-07-13
[MNT] add `statsforecast` to the `pandas2` compatible dependency set

fkiraly:
According to @yarnabrina, `statsforecast` is now compatible with `pandas 2`, so should be added to the respective dependency set in `pyproject.toml`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -109,6 +109,7 @@ all_extras_pandas2 = [
     "pyod>=0.8.0; python_version < '3.11'",
     "scikit_posthocs>=0.6.5",
     "seaborn>=0.11.0",
+    "statsforecast>=0.5.2; python_version < '3.11'",
     "statsmodels>=0.12.1",
     "stumpy>=1.5.1; python_version < '3.11'",
     "tbats>=1.1.0",


fkiraly:
Note: we need to check in the 3.8-3.10 tests whether the `pandas` version is `>=2`.

yarnabrina (approved):
Based on "Show dependencies" step, it seems to have worked.

One not-quite-blocking comment: for `statsforecast` and few others, lower bounds are often a specific patch version. Is it because we don't support anything till the previous patch version of same major/minor release, or is it because it's tested on that? As it stands, it looks like `sktime` is incompatible with `statsforecast==1.5.0` for example, which I don't think is the case.

fkiraly:
> and few others, lower bounds are often a specific patch version. Is it because we don't support anything till the previous patch version of same major/minor release, or is it because it's tested on that? 

I think it may be a combination of cases of "dependency fixed a critical bug" and "arbitrary lower bound".

I was starting to build a single CI element that tests on minimal dependency versions.
https://github.com/sktime/sktime/pull/4016
I got it to the point where only individual estimators were failing, the next (tedious) step would be to find out exactly the requirement that is making them fail and add that to the `python_dependencies` tag...

It's all yours if you want, otherwise I might return to it one day.

yarnabrina:
Yes, I'm terribly afraid of those tedious tests for every single version as well. Here's a slightly lazier approach, not sure if it's safe/too restrictive/etc. or not.

1. For soft dependencies with lower bounds, create one PR with pinning at corresponding minor versions (1.2.3 -> 1.2.0). If it succeeds, update those as lower bound and upper bound being the one above latest version on PyPI. In future, dependabot can take care of increasing upper bound automatically. Not sure how to continue checking the lower bound or increasing that in future.
2. For soft dependencies without lower bounds, again create a PR with pin at latest minor (4.5.6 -> 4.5) and then same as previous point.
3. In future, ensure >,>=,<,<= type restrictions are only on minor (may be major for very stable libraries but not confident on that). The only allowed patch restrictions will be of != type, unless there's an exceptional case where we need == type.
4. Just noting that this assumes pinnings on different soft dependencies will not affect one another. I'm not at all confident that will be the case, but not sure how to handle that properly.

Let me know what do you think. Requesting @achieveordie, @benHeid, @hazrulakmal for your comments as well.

(Now that I think of it, it may be an issue of it's own probably.)

fkiraly:
> Let me know what do you think

I think this makes sense generally for the framework depenencies - we can do the same thing for estimator dependencies, although the gradual approach seems tedious to carry out for many estimators.

It shouldn't be too hard to identify which package an estimator specific failure is coming from, we could put work through bounds on that only?

PR #4880, opened 2023-07-14
[ENH] replace `"Coverage"` and `"Quantiles"` default variable name in univariate case with variable name

### LINKED ISSUE #4709: [BUG] Column names `predict_interval` output inconsistent between univariate and multivariate case

fkiraly:
Copy of https://github.com/sktime/sktime/pull/4780 which was reverted to ensure it ships with 0.21.0, not 0.20.1.

Fixes #4709 - in `predict_interval` and `predict_quantiles` return, replaces `"Coverage"` and `"Quantiles"` default variable name in the univariate case with the variable name, to make this consistent with the multivariate case. If no variable name is available, the variable name used is the integer 0.

This is done with a deprecation horizon or two MINOR cyles, by introducing a `legacy_interface` argument to public and private `predict_interval`, `predict_quantiles`. A value `True` (default) gives current behaviour, `False` gives future behaviour. In 0.22.0, we switch the default to `False`, and remove the argument in 0.23.0.

This includes tests for `legacy_interface` both `True` and `False`, with deprecation notes on removing this later.

### DIFF
--- a/sktime/datatypes/_proba/_convert.py
+++ b/sktime/datatypes/_proba/_convert.py
@@ -85,8 +85,8 @@ def convert_pred_interval_to_quantiles(y_pred, inplace=False):
     idx = y_pred.columns
     var_names = idx.get_level_values(0)
 
+    # todo 0.22.0 - predict_interval new interface - remove this
     # treat univariate default name
-    # todo: maybe not a good idea, remove this...
     # here because it's in the current specification
     var_names = ["Quantiles" if x == "Coverage" else x for x in var_names]
 
@@ -152,8 +152,8 @@ def convert_pred_quantiles_to_interval(y_pred, inplace=False):
     idx = y_pred.columns
     var_names = idx.get_level_values(0)
 
+    # todo 0.22.0 - predict_interval new interface - remove this
     # treat univariate default name
-    # todo: maybe not a good idea, remove this...
     # here because it's in the current specification
     var_names = ["Coverage" if x == "Quantiles" else x for x in var_names]
 
--- a/sktime/forecasting/base/_base.py
+++ b/sktime/forecasting/base/_base.py
@@ -37,6 +37,7 @@ __author__ = ["mloning", "big-o", "fkiraly", "sveameyer13", "miraep8"]
 __all__ = ["BaseForecaster"]
 
 from copy import deepcopy
+from inspect import getfullargspec
 from itertools import product
 from warnings import warn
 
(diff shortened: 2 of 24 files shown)


fkiraly:
Failure on merge to main (after a new forecaster was added) indicates that this is not safe to merge for the extender contract, i.e., keeping third party estimators safe and raising a warning rather than crashing them so we need to think about how we deal with this before a merge.

From discussion with @benHeid:

The high-level principle being, no one should get hit by an exception without one minor cycle of forewarning
(but that's exactly what happened with the new forecaster!) 

I also have an idea how it could work: we use configs.

We tell extenders, 0.21.X will raise no exception, but they are warned that 0.22.X will, for the old style. Until 0.23.0, they can select testing behaviour with a config, so they can set the config that under 0.21.X. it will test like under the new interface contract, or that 0.22.X will test like the old. After 0.23.0, only the new contract is permitted.

The config would sit in the `BaseForecaster`, and be removed at 0.23.0.

In the tests, we do the following:

* if the config is on `"auto"`, and the private `_predict_etc` has the `legacy_interface` arg, we test both values
* if the config is on `"auto"` and the private does not have the arg, we test old on 0.21.X, and new on 0.22.X.
* if the config is `"new"` or `"old"`, we test for what this indicates, between 0.21.0 and 0.22.X
* from 0.23.0, the config will have no effect, and we test the new interface only, non-compliance raises error.

fkiraly:
I think there's an important lesson learnt here, both for me (who wrote/designed this) and the merger who reviewed.

We have two contracts that we need to be careful of - the user contract and the extender contract. As a developer, it's usually much simpler to deal with the user contract than the extender contract, the latter might actually require some careful thinking.

fkiraly:
implemented - the config is called `"pred_int:legacy_interface:testcfg"`, and the draft deprec instructions for extenders are in the release notes https://github.com/sktime/sktime/pull/4894

PR #4886, opened 2023-07-14
&whoknowsB [BUG] fix `ExpandingWindowSplitter` for timedelta index missing last fold

### LINKED ISSUE #4885: [BUG]
**Describe the bug**
Splitters miss a fold at the end: 

**To Reproduce**

```python
import pandas as pd
from sktime.utils.plotting import plot_windows

temp = pd.DataFrame({
    'y': range(7*24),
    'objectid': [1] * 7 * 24,
    'timestamp': pd.date_range('2022-07-01 00:00:00', '2022-07-7 23:00:00', freq='H')
}).set_index(['objectid', 'timestamp'])


fh = pd.timedelta_range(start=0, end=pd.Timedelta(days=1), closed='right', freq='H')
initial_window = pd.Timedelta(days=1)
step_length = pd.Timedelta(days=1)
splitter = ExpandingWindowSplitter(fh=fh, initial_window=initial_window, step_length=step_length)

plot_windows(splitter, temp)
```

**Expected behavior**
The output number of splits is 5, it should be 6

**Additional context**
image

**Versions**
'0.20.0'

</details>

fkiraly:
Fixes #4885 using the bugfix suggested by @whoknowsB

Includes tests for the failure case in #4885.

### DIFF
--- a/sktime/forecasting/model_selection/_split.py
+++ b/sktime/forecasting/model_selection/_split.py
@@ -1022,7 +1022,7 @@ class BaseWindowSplitter(BaseSplitter):
         else:
             offset = step_length if start == 0 else pd.Timedelta(0)
             start_date = y[y < y[start] + offset][-1]
-            end_date = y[end - 1] - step_length if end <= len(y) else y[-1]
+            end_date = y[end - 1] if end <= len(y) else y[-1]
             date_cutoffs = pd.date_range(
                 start=start_date, end=end_date, freq=step_length
             )
--- a/sktime/forecasting/model_selection/tests/test_split.py
+++ b/sktime/forecasting/model_selection/tests/test_split.py
@@ -28,6 +28,7 @@ from sktime.forecasting.tests._config import (
     TEST_YS,
     VALID_INDEX_FH_COMBINATIONS,
 )
+from sktime.utils._testing.deep_equals import deep_equals
 from sktime.utils._testing.forecasting import _make_fh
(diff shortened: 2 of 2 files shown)


fkiraly:
@whoknowsB, this is really strange - initially, your fix seems to have worked.

Now i've updated from `main`, after 0.21.0, and everything seems to be failing around splitters. The only difference I can think of that makes a difference is `sklearn` now being 1.3.0.

fkiraly:
@whoknowsB, the longer I look into this, I wonder whether your propose fix works in all the cases? The last numbers of the folds seem to be wrong.

whoknowsB:
@fkiraly I did not have the time to test it extensively, it make a more extensive investigation. Rn I'm really busy with other work :(

fkiraly:
no worries, we'll get this sorted!

hazrulakmal:
>  @whoknowsB, this is really strange - initially, your fix seems to have worked.

could you show me how did you know the fix worked?. [a24b0f2](https://github.com/sktime/sktime/pull/4886/commits/a24b0f28525205b9d258dd39f7ad8b3ad48a9ef6) seems to only run docs check tho. I think the proposed change may have unintentionally change other behaviors of the splitter. Hence, I dont think the problem lies on changes in dependency. will look into this problem more deeply.

fkiraly:
> could you show me how did you know the fix worked?

Ah, you are right! - it actually never worked!

That makes things simpler in some respect...

PR #4891, opened 2023-07-16
[ENH] LTSFLinearForecaster, LTSFLinearNetwork, BaseDeepNetworkPyTorch

luca-miniati:

Issue reference: https://github.com/sktime/sktime/issues/4939

Implements `LTSFLinearForecaster`, `LTSFLinearNetwork`, using a base class for `pytorch`, `BaseDeepNetworkPyTorch`

LTSFLinearForecaster
 * highest level, takes hyperparameters
 
`LTSFLinearNetwork`
 * copy/pasted from https://github.com/cure-lab/LTSF-Linear/blob/main/models/Linear.py
 * changed `enc_in` to `in_channels`, idk if that makes more sense
 
 `BaseDeepNetworkPyTorch`
  * base class for pytorch networks
  * implements train loop

`torch`, as a soft dep in testing and the `dl` dep set (only used in testing)

##### For all contributions
- [x] Optionally, I've added myself and possibly others to the [CODEOWNERS](https://github.com/sktime/sktime/blob/main/CODEOWNERS) file - do this if you want to become the owner or maintainer of an estimator you added.

##### For new estimators
- [x] If the estimator relies on a soft dependency, I've set the `python_dependencies` tag and ensured


examples/ltsf/ltsf_example.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def train_test_split(dataset_path, input_output_len, test_size):
+    """Shuffle data for LTSF-Linear."""
+    data = pd.read_csv(dataset_path).drop("date", axis=1)
+    data = [
+        data[i : i + input_output_len] for i in range(0, len(data), input_output_len)
+    ]
+    data = [x for x in data]
+    random.shuffle(data)
```
llllkihn, line 17: Hey, I am here from sktime discord (lkihn) and I was wondering whether this train-test separation was done the same way in paper or is this draft part that is to be rewritten.

I am asking as this can lead to misleading results due to overfitting compared to time based split.
Thanks!

fkiraly: I think this is probably just a draft to test functionality - a temporal split is done in 19/20
but @luca-miniati would know best

sktime/forecasting/ltsf.py
```diff
@@ -0,0 +1,74 @@
+"""Deep Learning Forecasters using LTSF-Linear Models."""
+
+import torch
+
+from sktime.forecasting.deep_learning.base import BaseDeepNetworkPyTorch
```
pgryko, line 5: Do we want the BaseModel under forecasting or under network, where the existing CCN are?

examples/ltsf/ltsf_example.py
llllkihn: I found everything I needed here: https://github.com/cure-lab/LTSF-Linear/blob/main/data_provider/data_loader.py#L235-L241

Thanks.

sktime/forecasting/ltsf.py
luca-miniati: I was using the `sktime` classifiers as an example, their `BaseDeepClassifier` is found in `sktime/classification/deep_learning/base.py`

sktime/forecasting/deep_learning/base.py
```diff
@@ -58,5 +58,81 @@ def _fit(self, y, X, fh):
 
     def _predict(self, X, **kwargs):
         """Predict with fitted model."""
-        self.network.eval()
-        return self.network(X)
+        dataloader = self.build_pytorch_dataloader(X)
+        y_pred = []
+        for x, _ in dataloader:
+            y_pred.append(self.network(x).detach().numpy())
+        return np.concatenate(y_pred)
```
luca-miniati, line 65: This creates 3D np.array of shape [n, pred_len, in_channels]. When this return is passed to `convert_to` in the internal `predict` of `BaseForecaster`, the error is thrown.

fkiraly: would like to see the message, that should not happen?

I assume it happens in the default contract tests, so if you would kindly fix the linting issues, this should be visible.

fkiraly:
FYI, the current errors are about dependency isolation.

Soft dependencies should not be importanted outside functions or classes, or users who do not have then will encounter import errors, see here:
https://www.sktime.net/en/stable/developer_guide/dependencies.html#handling-soft-dependencies

There's probably a stray `import torch` somewhere.

sktime/forecasting/ltsf.py
```diff
@@ -0,0 +1,128 @@
...
+        self.criterion = criterion
+        self.optimizer = optimizer
+        self.lr = lr
+        self.num_epochs = num_epochs
+        self.custom_dataset_train = custom_dataset_train
+        self.custom_dataset_pred = custom_dataset_pred
+        self.batch_size = batch_size
+        self.scale = scale
+        self.shuffle = shuffle
+        self.network = LTSFLinearNetwork(
```
benHeid, line 78: I would move this initialisation to the fit method. Or move it to an extra method that is called by _fit.
The reason would be that then the pred_len could be reflect the desired forecasting horizon that is passed in fit.

sktime/networks/base.py
```diff
@@ -25,3 +28,154 @@ def build_network(self, input_shape, **kwargs):
...
+
+        Changes to state:
+            writes to self._network.state_dict
+
+        Parameters
+        ----------
+        X : iterable-style or map-style dataset
+            see (https://pytorch.org/docs/stable/data.html) for more information
+        """
+        dataloader = self.build_pytorch_train_dataloader(y)
```
benHeid, line 54: I would propose that the creation of the dataloader considers the passed fh. This would also enable the user to create a forecaster that predicts the next 4 to 8 values instead of having to predict all the next 8 values. This would probably also solve the issue with your test and the forecast that does not fit to the fh.

sktime/networks/base.py
```diff
@@ -25,3 +28,154 @@ def build_network(self, input_shape, **kwargs):
...
+
+class BaseDeepNetworkPyTorch(BaseForecaster, ABC):
+    """Abstract base class for deep learning networks using torch.nn."""
+
+    _tags = {
+        "python_dependencies": "torch",
+    }
+
+    def __init__(self):
+        super().__init__()
```
benHeid, line 41: Can you add some default values for all attributes used by this class and defined only in subclasses? This would enable the intellisense to work properly.

sktime/networks/base.py
```diff
@@ -25,3 +28,154 @@ def build_network(self, input_shape, **kwargs):
...
+                loss = self._criterion(y_pred, y)
+                self._optimizer.zero_grad()
+                loss.backward()
+                self._optimizer.step()
+
+        # self._fh = self.pred_len
+
+    def _predict(self, X, **kwargs):
+        """Predict with fitted model."""
+        from torch import cat
```
benHeid, line 69: If an fh is passed here that does not fit the fh provided in `fit`, we should probably raise an exception or at least a warning, since they have to match. @fkiraly what would be the best way to do it in sktime? 

Furthermore, future work could relax this a bit I adding the possibility of calling the network in an autoregressive way.

fkiraly: forecasters methods (public and private) should not have kwargs, although this is a network, so the syntax is completely open. Having said that, a `fh` in a network doesn't feel right, design-wise.

luca-miniati:
So the vectorized fit does not instantiate the network. The network is instantiated in regular fit (in base class), should this behavior be called from the vectorized version?

fkiraly:
FYI, the failures are related to the unstable downloads - they seem unrelated to the content of this PR.

fkiraly:
can you kindly update from `main` and resolve merge conflicts?

fkiraly:
@yarnabrina - I think this one is interesting, old CI all passes, new CI elements fail. No clear explanation, timeouts.

yarnabrina:
> @yarnabrina - I think this one is interesting, old CI all passes, new CI elements fail. No clear explanation, timeouts.

I'm not sure if I agree. As per this, all failures are from `networks`.

https://github.com/sktime/sktime/actions/runs/6684228637

`pytest` failed to collect any unit tests in that sub-directory in the configured timeout, and it failed as expected.

fkiraly:
I'm merging https://github.com/sktime/sktime/pull/5511 into this, that should prevent test failures from the `networks` module.

fkiraly:
all pass, is this ready for review?

luca-miniati:
Yes, sorry I forgot to mark

luca-miniati:
Currently working on other networks - will open a separate PR

examples/ltsf/test_ltsf.py
```diff
@@ -0,0 +1,8 @@
+from sktime.forecasting.ltsf import LTSFLinearForecaster
+from sktime.utils.estimator_checks import check_estimator
+
+res = check_estimator(
```
fkiraly, line 4: this should not be done like this - the test will be carried out automatically if the estimator is collected.

fkiraly (changes requested):
Really nice, thanks!

A couple change requests:

* the `sktime` extras in `pyproject.toml` do not include `torch`. As a consequence, the estimator is not tested by `TestAllForecasters`. I recommend to solve this as follows: add `pytorch` to the `dl` dependency set. In `test-full`, change `pip install .[all_extras_pandas2,dev]` to `pip install .[all_extras_pandas2,dev,dl]`
* The `BaseDeepNetworkPyTorch` is actually not a network class but an intermediate forecaster class. It should therefore be in the forecasting module, I would suggest `forecasting.base.adapters._pytorch`.

examples/ltsf/test_ltsf.py
fkiraly: test skips for the suite are configured globally, in `tests._config`

fkiraly (approved):
Looks like it works now!

I made some changes in `pyproject` and API ref which you may like to translate to your new PR.

fkiraly (approved):
Still approve.

sktime/networks/tests/test_all_networks.py
```diff
@@ -0,0 +1,33 @@
...
+    """
+
+    # note: this should be separate from TestAllNetworks
+    #   additional fixtures, parameters, etc should be added here
+    #   TestAllNetworks should contain the tests only
+
+    estimator_type_filter = "network"
+
+
+class TestAllNetworks(NetworkFixtureGenerator, QuickTester):
```
benHeid, line 27: I do not understand this test :)

benHeid (approved):

Issue #1675, opened 2021-12-01
[ENH] Add holiday transformer 


mloning:
**Is your feature request related to a problem? Please describe.**
While there is a `DateTimeFeatures` transformer to get date/time features from a time series (see https://github.com/alan-turing-institute/sktime/issues/1273), we don't have a transformer for handling holidays. 

**Describe the solution you'd like**
A `HolidaysFeatures` transformer that creates dummies for holidays based on a date-time index.

mloning:
I'm working on this.

mloning:
Someone else can take this if interested, some preliminary work is in #1722

VyomkeshVyas:
I'm working on this.

PR #4893, opened 2023-07-16
[ENH] Holiday indicator transformers by country or market based on `holidays` package

yarnabrina:
Fixes https://github.com/sktime/sktime/issues/1675

`sktime` lacks a holidays transformer as far as I can find in the documentation. This PR adds `CountryHolidaysTransformer` and `FinancialHolidaysTransformer` based on `holidays` package by @dr-prodigy.


yarnabrina:
@fkiraly This is the first draft of the transformers, but will appreciate if you can take a look. I need help in the following items:

1. if the tags are appropriate
2. what tests can be added? and how? added doctests that pass, but it seems `pytest` fails to discover any other unit test (at least in my local)
3. what features can be added? I suppose one may be interested in pre/post effects of holidays, or number of holidays if index is higher than days (weeks/months/etc.), but seemed complex
4. anything else that I missed

fkiraly:
FYI, @VyomkeshVyas has posted in #1675 recently that he is also working on a holiday transformer.
I think there's no reason not to have multiple, as there are multiple packages that produce holiday indicators, and implementation can differ. Just pinging to enable alignment.

sktime/transformations/series/holiday.py
```diff
@@ -0,0 +1,310 @@
...
+
+    This implementation wraps over holidays [1]_ by dr-prodigy.
+
+    Parameters
+    ----------
+    country : str
+        An ISO 3166-1 Alpha-2 country code.
+    subdiv : str, optional
+        The subdivision (e.g. state or province); not implemented for all countries
+        (see documentation).
```
fkiraly, line 22: which documentation? Link would be useful.

Perhaps in the preamble, link to the original package documentation for country code register and subivisions?

sktime/transformations/series/holiday.py
```diff
@@ -0,0 +1,310 @@
...
+from sktime.transformations.base import BaseTransformer
+
+__all__ = ["CountryHolidaysTransformer", "FinancialHolidaysTransformer"]
+__author__ = ["yarnabrina"]
+
+
+class CountryHolidaysTransformer(BaseTransformer):
+    """Country Holidays Transformer.
+
+    This implementation wraps over holidays [1]_ by dr-prodigy.
```
fkiraly, line 14: might be worth saying here what it does in `transform`. Does it produce a df with columns? If yes, which names, and order? Which rows?

sktime/transformations/series/holiday.py
```diff
@@ -0,0 +1,310 @@
...
+    observed : bool, optional
+        Whether to include the dates of when public holiday are observed (e.g. a
+        holiday falling on a Sunday being observed the following Monday). False may not
+        work for all countries.
+    language : str, optional
+        The language which the returned holiday names will be translated into. It must
+        be an ISO 639-1 (2-letter) language code. If the language translation is not
+        supported the original holiday names will be used.
+    categories : Tuple[str], optional
+        requested holiday categories.
```
fkiraly, line 37: where do I look up valid strings?

sktime/transformations/series/holiday.py
```diff
@@ -0,0 +1,310 @@
...
+        The year(s) to pre-calculate public holidays for at instantiation.
+    expand : bool, optional
+        Whether the entire year is calculated when one date from that year is
+        requested.
+    observed : bool, optional
+        Whether to include the dates of when public holiday are observed (e.g. a
+        holiday falling on a Sunday being observed the following Monday). False may not
+        work for all countries.
+    language : str, optional
+        The language which the returned holiday names will be translated into. It must
```
fkiraly, line 33: translated into, as a column name?

fkiraly: and what are valid language codes?

sktime/transformations/series/holiday.py
```diff
@@ -0,0 +1,310 @@
...
+        )
+
+        Xt = X.copy(deep=True)
+
+        if isinstance(Xt.index, pandas.DatetimeIndex):
+            dates = Xt.index
+        elif isinstance(Xt.index, pandas.PeriodIndex):
+            dates = Xt.index.to_timestamp()
+        else:
+            raise TypeError(f"index type is unsupported: {type(Xt.index)}")
```
fkiraly, line 131: this might upset the tests...

* if index of `X` is integer, one might just assume "days since start of 1970" (unix convention)
* shouhld the index of `Xt` not be the same as that of `X` (unless `expand=True`), or am I missing sth?

fkiraly (changes requested):
Tags look good to me!

Some comments to resolve or discuss:

* these docstrings are hard to write, I know ... I think it's important to always state
    * not just type but set of valid values, how to check the full range
    * what `transform` exactly produces conditional on the parameter - I assume a `pd.DataFrame`, but I can't really tell from the docstring how columns are named, what they contain, and how the params influence this
* I would suggest to think about more robust index handling - e.g., allowing int, and ensuring the type is not changed by `transform`. For discussion what makes most sense.
* `get_test_params` should cover a max of meaningful params (multiple settings incl default per param) while not proliferating the number of test sets too much

fkiraly:
> what tests can be added? and how? 

The most immediate thought I had, schema output tests? I.e., "we expect these columns, rows, index type given these parameters and this data".

> added doctests that pass, but it seems pytest fails to discover any other unit test (at least in my local)

what do you mean by that? I don't see any `pytest`-tests in the PR.

sktime/transformations/series/holiday.py
yarnabrina: May be I'm not following you, `Xt` does have same index as that of `X`. The conversion is only for `dates` which is local to this method, and not stored in either `X` or `Xt`.

---

Not sure about first point either. If indices are integers, are you suggesting to assume that they mean date difference since unix epoch? I'm not sure if that'd be a safe assumption, and I'd prefer to let it fail preferring explicit over implicit.

yarnabrina: I took a look at `DateTimeFeatures`, which looked similar.

https://github.com/sktime/sktime/blob/b29e147b54959a53cc96e5be9c3f819717aa38e7/sktime/transformations/series/date.py#L186-L192

yarnabrina:
First of all, i apologize for the PR. I wasn't aware of that issue and that @VyomkeshVyas is already assigned and working on it. If that's almost ready,  I'll happy to withdraw this one. I created as I will need to use this feature soon for my office work.

Second, in reponse to your docstring comments @fkiraly, those details are picked up from `holidays` directly, and not my addition. I'll add specific links to all the codes in relevant sections over next days.

Regarding `pytest`, what I meant is whenever I added a forecaster before, I saw `pytest -k "forecaster_name"` executes few tests based on test parameters. That didn't happen with these two transformers, and `pytest` complained that no tests exist.

Now I know that I didn't add specific units test yet, but expected that transformers will have some generic tests validating tags, indices, etc. like forecasters have in the `test_all_forecasters.py`, and those will be run always even in absence of estimator specific tests. May be it's handled differently for transformers?

(I am asking blindly here, I have no idea.)

fkiraly:
> First of all, i apologize for the PR. I wasn't aware of that issue and that @VyomkeshVyas is already assigned and working on it. If that's almost ready, I'll happy to withdraw this one. I created as I will need to use this feature soon for my office work.

As said, no harm in multiple holiday transformers, especially if the upstream dependency is different. But even if not, still fine - we have the "library" ethos. There are at least five significantly different versions of Hugo's "Les Misérables" (the book) by the author himself, and they are all great.

sktime/transformations/series/holiday.py
fkiraly: > May be I'm not following you, `Xt` does have same index as that of `X`. The conversion is only for dates which is local to this method, and not stored in either `X` or `Xt`.

My bad - I thought I saw you writing to the `index`, but I was mistaken.

Still, the error is raised for any non-datelike index, that much is right, no?

fkiraly: > Not sure about first point either. If indices are integers, are you suggesting to assume that they mean date difference since unix epoch? I'm not sure if that'd be a safe assumption, and I'd prefer to let it fail preferring explicit over implicit.

Well, if the tests get upset about this. I thought they would?

Minor comment: in an exception like this: it is always good to say which estimator the `ValueError` is coming from; what the "found" an "expected" is.

fkiraly (changes requested):
I noticed two more things while replying:

* `handles-missing-data` should be `True`, no? Values are all ignored.
* it is better to only return the new feature columns than keep the old ones too. It is easier to get the "other" behaviour if you do that (via `FeatureUnion` rather than subsetting to columns whose names are not easy to retrieve). If you want both in the estimator, suggest a parameter that controls this and set the default to the first.

yarnabrina:
@fkiraly can you please guude regarding the two present test failures?

`test-nosoftdeps` fail complaining about  `holidays`, but I did add `python_dependencies` tag and imported inside `_transform`. What am I missing?

`test-windows (3.11)` is also failing with `scipy`'s `BracketError`, which I thought was fixed by #4770.

fkiraly:
>`test-nosoftdeps` is failing after complaining about `holidays`, but I did add `python_dependencies` tag and imported inside `_transform`. What am I missing?

as far as I can see, the failures are from the doctests, where the soft dependency has not been isolated.

https://www.sktime.net/en/latest/developer_guide/dependencies.html#handling-soft-dependencies

point "if including docstring examples" should be the relevant one

> `test-windows (3.11)` is also failing with `scipy`'s `BracketError`, which I thought was fixed by #4770.

might be sporadic?

fkiraly:
looks ready to merge, but still a draft?

yarnabrina:
I didn't get enough time to go through `holidays` codebase in details to update all parts of the documentation. If you are okay with the current docstring (I updated with ISO links etc.), then code wise I have nothing new to add. Marking it ready for review.

fkiraly (changes requested):
Looks great!

Just some small blocking requests:

* could we move the estimators into a module, i.e., `holiday/_country` etc, with a joint export locus via `__init__`, in anticipation of https://github.com/sktime/sktime/pull/4909?
* can you be more explicit in the docstring description? I.e., what exactly from `holidays` does it provide to the user? What does `fit`/`transform` precisely do? I presume, given country and market codes, produces a column, but what is in that column? Boolean, 1/0, sth else?

fkiraly:
PS: I also suggested a new PR title by change to be more descriptive, please feel to change if you don't like it.

Just thinking ahead, as the title is usually what goes in the changelog (if clear enough) and, potentially, the "highlights" section of the release notes.

yarnabrina:
Re-requesting review.

1. Tried to address the blocking suggestions.
2. Regarding non-blocking suggestions
    1. I prefer to not give a name from `sktime` side. The estimator does not know (without explicitly checking) what other columns is already present in the dataset, and I'll prefer to give no name and keep that to user's control entirely.
    2. I am not sure how common it is to use multiple countries/markets, though I suppose in EU this may be very common. This extension can be done, but I'd skip for now if it's okay. We can track in a issue of you want, will come back to it when I have a bit more time.

fkiraly:
> I'll prefer to give no name and keep that to user's control entirely.

If you want it or not, the column will get a name when in a multivariate `pd.DataFrame`. I think in the case it will simply be `0`.

> The estimator does not know (without explicitly checking) what other columns is already present in the dataset

In a `ColumnEnsembleTransformer`, the names are automatically made unique.
Design-wise, it is perfectly admissible - and desired - that transformers produce outputs with descriptive names.

fkiraly:
> This extension can be done, but I'd skip for now if it's okay. We can track in a issue of you want, will come back to it when I have a bit more time.

Well, I said it's non-blocking.

Indeed it's not uncommon to have holidays for close-by countries in the mix - not just for, but also for sales of units close to a national border.

fkiraly:
another non-blocking remark: would it be helpful to give the user a choice of boolean and float? Most sklearn-like toolboxes expect float and it tends to be less buggy, so it might be even better as a default.

fkiraly (dismissed):
Blocking requests were addressed, thanks!

Leaving this open for a few days in case you still want to comment on the other discussion points, @yarnabrina.

yarnabrina:
Updates:

1. removed `language` argument, as boolean indicator is unaffected by it.
2. added `name` argument which can allow user to control name, and will auto-populate using country/subdic/market information

Comments:

1. Multi-country/market is really out of scope for me now because of bandwidth (and personal prioritisation of darts adapter). Will come back to it, if not added by others already.
2. I am not clear on float vs boolean argument. Python booleans are basically 0 and 1, and should support all arithmetic operations, and should take less storage than `int` or `float` (or `uint`). Can you please tell what is the problem, and which sklearn-like functionality have issues with it? Skipped for now, can add `astype` if necessary.

fkiraly (approved):
All points addressed.

Regarding the boolean type, should be ok, I follow your argument.

Issue #4803, opened 2023-07-01
[BUG] objects are being documented while they are not present in correct location or absent altogether


yarnabrina:
Currently, `make docs`lead to the following warnings during build process:

```shell
WARNING: [autosummary] failed to import 'sktime.classification.deep_learning.MACNNClassifier': no module named sktime.classification.deep_learning.MACNNClassifier
WARNING: [autosummary] failed to import 'sktime.performance_metrics.forecasting.count_error': no module named sktime.performance_metrics.forecasting.count_error
WARNING: [autosummary] failed to import 'sktime.performance_metrics.forecasting.hausdorff_error': no module named sktime.performance_metrics.forecasting.hausdorff_error
WARNING: [autosummary] failed to import 'sktime.performance_metrics.forecasting.prediction_ratio': no module named sktime.performance_metrics.forecasting.prediction_ratio
WARNING: [autosummary] failed to import 'sktime.regression.deep_learning.ResNetRegressor': no module named sktime.regression.deep_learning.ResNetRegressor
WARNING: [autosummary] failed to import 'sktime.transformations.panel.compose.SeriesToPrimitivesRowTransformer': no module named sktime.transformations.panel.compose.SeriesToPrimitivesRowTransformer
WARNING: [autosummary] failed to import 'sktime.transformations.panel.compose.SeriesToSeriesRowTransformer': no module named sktime.transformations.panel.compose.SeriesToSeriesRowTransformer
WARNING: [autosummary] failed to import 'sktime.transformations.panel.compose.make_row_transformer': no module named sktime.transformations.panel.compose.make_row_transformer
```

I have checked for each of these, and most of these are valid issues:

1. `SeriesToPrimitivesRowTransformer`, `SeriesToSeriesRowTransformer`, `make_row_transformer` are absent in codebase but are being listed in `docs/source/api_reference/transformations.rst`. The first two are also present in `sktime/tests/_config.py`
2. `count_error`, `hausdorff_error`, `prediction_ratio` are present in `annotation` but are being listed in `docs/source/api_reference/performance_metrics.rst` as part of `forecasting`
3. `MACNNClassifier` is defined in `sktime/classification/deep_learning/macnn.py`, but not imported and exposed through `sktime/classification/deep_learning/__init__.py` file, even though listed as such in `docs/source/api_reference/classification.rst`
[...]

PR #4895, opened 2023-07-16
[DOC] remove outdated references from transformers API

fkiraly:
The row transformers (compositors that add broadcasting that is now standard for all transformers) have been removed a while ago from `sktime` - this PR removes the dead references to them from the transformers API.


yarnabrina (approved):
Looks good. I believe this addresses part of #4803, so not adding a "Fixes" keyword.

PR #4901, opened 2023-07-17
[BUG] #4900 60 seconds per core of unnecessary delay for pytest --col…

### LINKED ISSUE #4900: [BUG] 60 seconds per core of unnecessary delay for pytest --collect-only execution
**Describe the bug**
I followed the contributor installation instructions and executed `python -m pytest -n0 --collect-only` and the command took 1m30s to execute on may fairly fast computer. This is also related to: https://github.com/sktime/sktime/issues/2890#issuecomment-1188833581 and a little to: https://github.com/sktime/sktime/issues/2530

Upon profiling 60s of that was spent on line
https://github.com/sktime/sktime/blob/b61f2de7dc792db7d437f65a2dd65f51832011b4/sktime/utils/_testing/scenarios.py#L265
creating a string which is then thrown away.

test4

**To Reproduce**
```
python -m pytest -n0 --collect-only
```

**Expected behavior**
Let's not create msg until it's needed. 
```
diff --git a/sktime/utils/_testing/scenarios.py b/sktime/utils/_testing/scenarios.py
index 449889d4..0949f0c6 100644
--- a/sktime/utils/_testing/scenarios.py
+++ b/sktime/utils/_testing/scenarios.py
@@ -262,11 +262,9 @@ def _check_dict_of_dict(obj, name="obj"):
     ------
     TypeError if obj is not dict of dict, with str keys
     """
-    msg = f"{obj} must be a dict of dict, with str keys"
-    if not isinstance(obj, dict):
-        raise TypeError(msg)
-    if not np.all([isinstance(x, dict) for x in obj.values()]):
-        raise TypeError(msg)
-    if not np.all([isinstance(x, str) for x in obj.keys()]):
-        raise TypeError(msg)
+
+    if not (isinstance(obj, dict) and
+            np.all([isinstance(x, dict) for x in obj.values()]) and
+            np.all([isinstance(x, str) for x in obj.keys()])):
+        raise TypeError(f"{obj} must be a dict of dict, with str keys")
     return obj
```  

**Additional context**

**Versions**
System:
    python: 3.9.16 | packaged by conda-forge | (main, Feb  1 2023, 21:38:11)  [Clang 14.0.6 ]
executable: /Users/tibor/miniforge3/envs/sktime-dev/bin/python3
   machine: macOS-13.3.1-arm64-arm-64bit

Python dependencies:
          pip: 23.2
       sktime: 0.20.1
      sklearn: 1.1.3
       skbase: 0.4.6
        numpy: 1.22.4
        scipy: 1.11.1
       pandas: 1.5.3
   matplotlib: 3.7.1
       joblib: 1.3.0
  statsmodels: 0.14.0
        numba: 0.57.1
     pmdarima: 1.8.5
      tsfresh: 0.20.0
   tensorflow: 2.13.0
tensorflow_probability: None

tarpas:
#### Reference Issues/PRs
Fixes #4900.

Removes unecessary creation of the string.

No.

Not really, It's a coupole of lines.

I only changed test collection I think.

##### For all contributions
- [X] The PR title starts with either [ENH], [MNT], [DOC], or [BUG]. [BUG] - bugfix, [MNT] - CI, test framework, [ENH] - adding or improving code, [DOC] - writing or improving documentation or docstrings.


sktime/utils/_testing/scenarios.py
```diff
@@ -262,11 +262,8 @@ def _check_dict_of_dict(obj, name="obj"):
...
-    msg = f"{obj} must be a dict of dict, with str keys"
-    if not isinstance(obj, dict):
-        raise TypeError(msg)
-    if not np.all([isinstance(x, dict) for x in obj.values()]):
-        raise TypeError(msg)
-    if not np.all([isinstance(x, str) for x in obj.keys()]):
-        raise TypeError(msg)
+    if not (isinstance(obj, dict) and
+            np.all([isinstance(x, dict) for x in obj.values()]) and
+            np.all([isinstance(x, str) for x in obj.keys()])):
```
yarnabrina, line 267: Questions (mainly to @fkiraly I guess):

1. Do we need `np.all` here, is `all` not enough?
2. `all` can do a short circuit directly if I'm not wrong, but you need to skip the internal list. Can we change that?

yarnabrina (commented):
Thanks. I've faced slow collection for so long, but didn't know there's a tool to find what's causing such things. I'll use it almost everywhere now :)

Made a minor comment to clarify one of my personal doubts.

sktime/utils/_testing/scenarios.py
fkiraly: Not sure - I think `all` is enough.

I also think, there's no harm in removing the check entirely, if it is so consequential!
This is so far developer-facing (not even extender-facing) that it doesn't matter if we get rid of it entirely!

fkiraly:
FYI, @tarpas, see https://github.com/sktime/sktime/issues/4900#issuecomment-1640098525 - the message should have called repr of a short string, `name`, rather than of the `DataFrame`-s (`obj`)

Could you change the `obj` in the message to `name`.

I got confused when I wrote this, since the default for `name` was the string `"obj"`.

fkiraly (approved):
Thanks!

I approve if you're hppy with changing `obj` to `name`.

Also, from my side non-blocking, but I'd agree with @yarnabrina that `np.all` might not be needed.

sktime/utils/_testing/scenarios.py
yarnabrina: Are you asking between `np.all` and `all`? My guess is `all` as elements are basic python objects and not `numpy` objects.

Ref. https://stackoverflow.com/a/43382396/11117265

(Probably not much difference though, my comment was mainly from "numpy-not-required-here" perspective.)

sktime/utils/_testing/scenarios.py
```diff
@@ -262,11 +262,10 @@ def _check_dict_of_dict(obj, name="obj"):
...
-        raise TypeError(msg)
-    if not np.all([isinstance(x, dict) for x in obj.values()]):
-        raise TypeError(msg)
-    if not np.all([isinstance(x, str) for x in obj.keys()]):
-        raise TypeError(msg)
+    if not (
+        isinstance(obj, dict)
+        and np.all([isinstance(x, dict) for x in obj.values()])
+        and np.all([isinstance(x, str) for x in obj.keys()])
+    ):
```
yarnabrina, line 269:
```suggestion
    if not (
        isinstance(obj, dict)
        and all(isinstance(x, dict) for x in obj.values())
        and all(isinstance(x, str) for x in obj.keys())
    ):
```

Suggesting 2 changes:

1. `np.all` to `all`
2. allow short-circuit

I'm suggesting from Github site, code's not tested.

sktime/utils/_testing/scenarios.py
fkiraly: ok, let's remove numpy then

yarnabrina (approved):
Approved, with two non-blocking suggestions from my side.

fkiraly (approved):
This is the most collaborative PR in a while!

* I've merged @yarnabrina's suggestion of not using `numpy`
* I've done this in another parallel location, and fixed the `name`/`obj` confusion in the same location

I think this is now good to go!

PR #4903, opened 2023-07-18
[ENH] private `split_loc` and tag to control dispatch of `split_series` to `split` vs `split_loc`

fkiraly:
This PR introduces, for splitters (`BaseSplitter` descendants):

* an optional private `_split_loc` to implement, which can differ from the default of looking up `loc` from `iloc` indices
* a tag `split_series_uses`, which determines whether `split_series` dispatches to `split_loc` or `split` (iloc). This may be later extended to have splitters where only `split_series` exists, e.g., which are not just determined by rows

This PR also changes implementation of two splitters which in sum make the forecasting benchmark `evaluate` more performant:
* for the `SameLocSplitter`, `split_series` should dispatch to `split_loc` for performance gain.
* the `TestPlusTrainSplitter` should dispatch to the same as the wrapped, for performance gain.

An alternative approach to https://github.com/sktime/sktime/pull/4902 to speed up `evaluate`, but both could be merged.

### DIFF
--- a/sktime/forecasting/model_selection/_split.py
+++ b/sktime/forecasting/model_selection/_split.py
@@ -1490,6 +1490,10 @@ class TestPlusTrainSplitter(BaseSplitter):
         self.cv = cv
         super().__init__()
 
+        # dispatch split_series to the same split/split_loc as the wrapped cv
+        # for performance reasons
+        self.clone_tags(cv, "split_series_uses")
+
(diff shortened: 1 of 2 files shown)


hazrulakmal:
is changes in this PR what you refer to as **iloc windowing**?

fkiraly:
> are the changes in this PR what you refer to as iloc windowing?

No, this is just a delegator switch.

PR #4911, opened 2023-07-18
[ENH] k-means clusterer based on `sklearn` `KMeans` and `sktime` distances

### LINKED ISSUE #4863: [ENH] Initial centroid guess for kmeans
This package is great, thanks for all the hard work.

**Is your feature request related to a problem?**
No

It would be nice to be able to provide the `TimeSeriesKMeans` class with an initial guess of clusters similar to `KMeans` in `sklearn`, https://scikit-learn.org/stable/modules/generated/sklearn.cluster.KMeans.html. In `sklearn` you can pass a `numpy` array representing your initial guess to the `init` argument.

This would mean updating the `init_algorithm` argument for `TimeSeriesKMeans` to accept an array containing the initial centroid guesses. The dimensions of the array would be the same as the `cluster_centers_` attribute.

Alternatively, since an array of centroids is not an algorithm, perhaps it would make more sense add a new argument called `init_centroids` instead of adding it as an option to the `init_algorithm` argument. Or, the `init_algorithm` argument could be renamed to `init` which would bring it more inline with `sklearn`.

Alex-JG3:
#### Reference Issues/PRs

Closes #4863.

Adds a version of KMeans that uses the sklearn KMeans implementation. This implementation closely follows the `TimeSeriesDBSCAN` class.

Moves the current KMeans class into its own subpackage.

No

##### For all contributions

##### For new estimators


sktime/clustering/k_means/k_means_sklearn.py
```diff
@@ -0,0 +1,264 @@
...
+            Time series instances to predict cluster indexes for
+        y: ignored, exists for API consistency reasons
+
+        Returns
+        -------
+        np.ndarray (1d array of shape (n_instances,))
+            Index of the cluster each time series in X belongs to
+        """
+        # if X is the same as seen in _fit, simply return the labels
+        if X is self._X:
```
fkiraly, line 192: I think we don't need to do this, because the sklern estimator can handle different `X` in `predict`

fkiraly (commented):
Great start, thanks for the contrib!

Left one comment above.

sktime/clustering/k_means/k_means_sklearn.py
Alex-JG3: Thanks for starting the review. Addressed in commit bfd0be0. Let me know if that change is what you are looking for.

fkiraly: yes, exactly!

The DBSCAN logic was just a crutch, since `sklearn` apparently allows estimators to fail in `predict` - we don't...

fkiraly:
Failures are unrelated, probably a `dask` or `pandas` update, see https://github.com/sktime/sktime/issues/4925

sktime/clustering/k_means/k_means_sklearn.py
```diff
@@ -0,0 +1,253 @@
...
+        "verbose",
+        "random_state",
+        "copy_x",
+        "algorithm",
+    ]
+    DELEGATED_FITTED_PARAMS = ["cluster_centers_", "labels_"]
+
+    def __init__(
+        self,
+        distance,
```
Alex-JG3, line 111: @fkiraly, should the `distance` parameter be a string or a callable like the current `metric` parameter in the `TimeSeriesKMeans`?

The docstring for the `TimeSeriesDBSCAN` class (which the `TimeSeriesKMeansSklearn` class follows) says the `distance` parameter can be a string or a callable. However, from what I can tell there is no code to parse strings in the `distance` parameter?

```python
>>> from sktime.clustering.dbscan import TimeSeriesDBSCAN
>>> import numpy as np
>>> x = np.random.normal(size=(50, 2))
>>> model = TimeSeriesDBSCAN(distance="euclidean")
>>> model.fit(x)
  File "<stdin>", line 1, in <module>
  File "/home/alex/documents/sktime/sktime/clustering/base.py", line 118, in fit
    self._fit(X)
  File "/home/alex/documents/sktime/sktime/clustering/dbscan.py", line 122, in _fit
    distmat = distance(X)
TypeError: 'str' object is not callable
>>> from sktime.dists_kernels import DtwDist
>>> model = TimeSeriesDBSCAN(distance=DtwDist())
>>> model.fit(x)
TimeSeriesDBSCAN(distance=DtwDist())
```
Am I missing something in my code example above or is the docstring for  `TimeSeriesDBSCAN` slightly out of date?

fkiraly: yes, the docstring is incorrect - but it is out of date in the sense that it is the "future desired state" which has not been implemented by accident.

I.e., in the end state, we want all distance based learners to accept strings, and some logic internally converts it to the right callable.

For now, I'd say, you don't need to add that logic because it will be centrally done for all distance based learners - but for correctness' sake, you may want to consider removing the reference to `str` from the docstring until that's implemented.

fkiraly: for reference, there's a draft PR here: https://github.com/sktime/sktime/pull/4712
although the end result might look different

fkiraly (approved):
Looks like it's done - but it's still a draft, which indicates you're still working on it, @Alex-JG3?

Feel free to add yourself to the `.allcontributorsrc` file - I think this qualifies for `code` and `ideas`.

Alex-JG3:
@fkiraly, on second thought, I am not sure we can use the sklearn implementation of `KMeans` in the way we we want to:

1. You cannot pre-compute the distances in `KMeans` like you can in `dbscan`

Alex-JG3:
@fkiraly, on second thought, I don't think we can use sklearns `KMeans` algorithm in the way we want to.

I assumed you can pass a pre-computed matrix of distances to `sklearn.cluster.KMeans` like you can `sklearn.cluster.DBCAN`. I don't think you can do this because in KMeans the distances from each point to the centroids have to be recalculated at each step. I am not too familiar with how DBSCAN works but I don't believe it uses centroids in the same way.

The documentation for `sklearn.cluster.DBSCAN.fit` says that `X` can be,

> Training instances to cluster, or distances between instances if metric='precomputed'. ...

Whereas the documentation for `sklearn.cluster.KMeans.fit` says that `X` can be,

> Training instances to cluster. ...

It doesn't mention you can use distances. Also, currently `sklearn.cluster.KMeans` does not support metrics other than euclidean distance. Here is a related stack overflow thread https://stackoverflow.com/questions/5529625/is-it-possible-to-specify-your-own-distance-function-using-scikit-learn-k-means. So, I don't think we can pass the pre-computed distance matrix to `sklearn.cluster.KMeans`.

## How did I realise there was a problem?

I tried using the example in the `TimeSeriesKMeans` docstring with the new class, `TimeSeriesKMeansSklearn`. 

```python
>>> from sktime.datasets import load_arrow_head
>>> from sktime.clustering.k_means.k_means_sklearn import TimeSeriesKMeansSklearn
>>> from sktime.dists_kernels import DtwDist
>>> X_train, y_train = load_arrow_head(split="train")
>>> X_test, y_test = load_arrow_head(split="test")
>>> X_train.shape
(36, 1)
>>> X_test.shape
(175, 1)
>>> clusterer = TimeSeriesKMeansSklearn(distance=DtwDist(), n_clusters=3)
>>> clusterer.fit(X_train)
TimeSeriesKMeansSklearn(distance=DtwDist(), n_clusters=3)
>>> clusterer.predict(X_test)
  File "<stdin>", line 1, in <module>
  File "/home/alex/documents/sktime/sktime/clustering/base.py", line 142, in predict
    return self._predict(X)
  File "/home/alex/documents/sktime/sktime/clustering/k_means/k_means_sklearn.py", line 206, in _predict
    return self.kmeans_.predict(distmat)
  File "/home/alex/documents/sktime/.venv/lib/python3.10/site-packages/sklearn/cluster/_kmeans.py", line 1097, in predict
    X = self._check_test_data(X)
  File "/home/alex/documents/sktime/.venv/lib/python3.10/site-packages/sklearn/cluster/_kmeans.py", line 950, in _check_test_data
    X = self._validate_data(
  File "/home/alex/documents/sktime/.venv/lib/python3.10/site-packages/sklearn/base.py", line 625, in _validate_data
    self._check_n_features(X, reset=reset)
  File "/home/alex/documents/sktime/.venv/lib/python3.10/site-packages/sklearn/base.py", line 414, in _check_n_features
    raise ValueError(
ValueError: X has 175 features, but KMeans is expecting 36 features as input.
```
The training data array has shape `(36,1)` so the distance matrix will have have shape `(36,36)`. Passing the distance matrix straight into `sklearn.cluster.KMeans.fit` means the algorithm is treating the array as having 36 samples and 36 features. Hence, it gets confused when we pass it an array with shape (175, 175) as it treats this as having 175 features.

So I don't think this implementation works how we want it and since it doesn't look like we can pass a pre-computed distance matrix to `sklearn.cluster.KMeans` like we can to `sklearn.cluster.DBSCAN`, we we may not be able to use `sklearn.cluster.KMeans` in the same way.

fkiraly:
oh, I see, @Alex-JG3  - it seems like you are right.

I think there is a subtle point which prevented me from realizing there was a problem in the first place - in kernel-k-means you *can* pass a precomputed matrix, but it is a kernel, not a distance matrix. Plain k-means being a special case, and distances-kernels being translatable to each other, it was plausible that there is a way to use precomputed matrices!

Now, I see four ways to go from here:

* use the kernel k-means from sklearn instead! - this is the same as spectral clustering, mathematically, so what should work is https://scikit-learn.org/stable/modules/generated/sklearn.cluster.SpectralClustering.html, and perhaps call the estimator spectral clustering.
* only use training distances in predict - that will give rise to a valid algorithm, ut it is not the same as k-means with the time series distance. It is a valid k-means though, and this could e explained in the docstring
* go ack to #4863 and the option to exted the initial centroid
* try to modify the `tslear` kmeas, here https://tslearn.readthedocs.io/en/stable/gen_modules/clustering/tslearn.clustering.TimeSeriesKMeans.html

The first two are low hanging fruit imo, the other two are more involved, ut nothing keeps us from trying multiple?

fkiraly:
side note: option 2 would be the same as a pipeline of `DistanceFeatures` and `sklearn` `KMeans`

fkiraly:
doing some more reading:

* there are multiple algorithms which can e used to compute a k-means clustering. The minimand - both for fitting and cluster assignment can be expressed completely in terms of pairwise distances
* however, not every algorithm does have a natural generalization in terms of distances, that includes Lloyd's algorithm which is claimed to e the one in the current implementation. The docstring does not say which generalization is implemented, it should.
    * on a very subtle point, more algorithms can e expressed in terms of kernels than distances (including the naive algorithm)
* looking at the current implementation, whatever it's doing, it does use precomputed distance matrices, so it should easily be convertible to a general implementation, because precomputing is all what it is doing in `fit`.

Alex-JG3:
Happy to try option 1 to begin with since, as you say, it sounds like low hanging fruit.

fkiraly:
ok - feel free to use this PR or open another one.

Alex-JG3:
Looks like sklearn's 'SpectralCluster' does not have a 'SpectralClustering.predict' method. Is that okay or would it be better to explore some another approach?

fkiraly:
> Is that okay

In-principle yes - we can use the logic from `DBSCAN`.

Though it is a bit unsatisfying.
I'll have a look whether I can see an easy way to adapt one of the k-means algorithms (tslearn, sklearn, sktime)

Alex-JG3:
Closing since this was address in #5001.

Issue #4904, opened 2023-07-18
[BUG] `mlflow` tests failing with `AssertionError: Expected 'register_model' to be called once. Called 0 times.`


fkiraly:
Since today or yesterday, the `mlflow` CI is failing with `AssertionError: Expected 'register_model' to be called once. Called 0 times.`.

FYI @benjaminbluhm - is this a problem in `sktime`, or upstream in `mlflavors`?

benjaminbluhm:
Doesn't look like sktime problem to me, might be related to latest mlflow release: https://pypi.org/project/mlflow/2.5.0/

Unfortunately I do not have capacity to look into this in more detail

fkiraly:
I `xfail`-ed it for now, let's track it here. Maybe it goes away with 2.5.1 ...

PR #4913, opened 2023-07-19
[MNT] xfail `mlflow` failure #4904 until debugged, gitignore for `py-spy`

fkiraly:
Small maintenance actions:

* `xfail`-s the `mlflow` failure in #4904 until debugged
* add `py-spy` outputs to `.gitignore`, see #4900


fkiraly:
`mlflow` failure is in main and multiple people are working with `pyspy` (see #4900), so quickmerge

PR #4914, opened 2023-07-19
[DOC] 0.21.0 release action - update deprecation guide to reflect deprecation of use of `deprecated`

fkiraly:
In preparation of deprecation of `deprecated` as a dependency and for use in the deprecation process scheduled for 0.22.0 https://github.com/sktime/sktime/pull/4822, the deprecation guide is updated to reflect the post-deprecation state.

### DIFF
--- a/docs/source/developer_guide/deprecation.rst
+++ b/docs/source/developer_guide/deprecation.rst
@@ -27,7 +27,8 @@ Our current deprecation policy is as follows:
 * deprecation warnings must be included for at least one full MINOR version cycle before change or removal.
   Therefore, typically, the change or removal happens at the *second* next MINOR release.
 
-Example process:
+Example timeline:
+
 1. developer A resolves, at current state v0.9.3, to remove functionality X at some point in the near future.
 2. therefore, by the above, we should introduce a deprecation message, visible from next release (e.g., v0.9.4),
   which says that functionality will be removed at v0.11.0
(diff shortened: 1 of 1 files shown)


achieveordie (commented):
While this has not yet been merged, I'd like to point to an alternative process that might be easier due to the two reasons mentioned [here](https://github.com/sktime/sktime/pull/4871#issuecomment-1639723843):
> 1. All the upcoming deprecations are consolidated in a single file(...).

> 2. We avoid writing multiple deprecations warnings/methods in case the same deprecation is done over multiple objects.

fkiraly:
> While this has not yet been merged, I'd like to point to an alternative process that might be easier

Yes, I saw that, although I haven't spent enough time understanding it well enough (was focused on the release).

What this PR does - in my intention at least - it just removes `deprecated` related content from the deprecation guide without modifying current practice. It is to prevent that contributors make PR with a dependency on `deprecated`.

Your proposal afaik is a suggested *change* to current practice, and in-principle we can move there. Should we discuss in an issue? I'd do that later though than 0.21.0, as it is an additive or a change, rather than a removal. Pre-0.22.0 is an option.

What do you think?

If you are against removing `deprecated` as a dependency, then we can leave this out from 0.21.0 and revert the dependency change, just say so.

achieveordie:
Yes, we can discuss it post-release, and I don't have any issue with `deprecated` package being removed as a dependency. In fact, it aligns with our current philosophy of not having needless dependencies, so it's all good.

Issue #4827, opened 2023-07-05
[BUG] make_forecasting_scorer and functools.partial don't mix.


davidgilbertson:
**Describe the bug**
I have a scoring function that is created using `functools.partial` and the scikit-learn `mean_squared_log_error` function. I'm wrapping this with `make_forecasting_scorer` to use with sktime. There's some logic in sktime that checks the parameters of this function, but there's a mismatch because of the partial application.

**To Reproduce**

```python
import functools
import pandas as pd
from sktime.performance_metrics.forecasting import make_forecasting_scorer
from sklearn.metrics import mean_squared_log_error

rmsle = functools.partial(mean_squared_log_error, squared=False)

scorer = make_forecasting_scorer(rmsle, name="RMSLE")

scorer.evaluate(pd.Series([1, 2, 3]), pd.Series([1, 2, 4]))
```

This produces the error:
```txt
Traceback (most recent call last):
  File "/home/davidg/.pycharm_helpers/pydev/pydevconsole.py", line 364, in runcode
    coro = func()
  File "<input>", line 10, in <module>
  File "/home/davidg/.virtualenvs/learning/lib/python3.10/site-packages/sktime/performance_metrics/forecasting/_classes.py", line 214, in evaluate
    out_df = self._evaluate(y_true=y_true_inner, y_pred=y_pred_inner, **kwargs)
  File "/home/davidg/.virtualenvs/learning/lib/python3.10/site-packages/sktime/performance_metrics/forecasting/_classes.py", line 582, in _evaluate
    params = {key: params[key] for key in func_params}
  File "/home/davidg/.virtualenvs/learning/lib/python3.10/site-packages/sktime/performance_metrics/forecasting/_classes.py", line 582, in <dictcomp>
    params = {key: params[key] for key in func_params}
KeyError: 'squared'
```

**Expected behavior**

**Additional context**
The code that raises the error:
https://github.com/sktime/sktime/blob/80907258adb7df7d3b7e0356c1264ee40bd7d986/sktime/performance_metrics/forecasting/_classes.py#L575-L578

`func_params` contains `squared`, but `params` doesn't. This code should allow for the fact that the `signature` params might not be in the `.get_params()` result.

**Versions**
<details>

</details>

davidgilbertson:
I should add, the workaround is simple enough - don't use `partial`.
```py
def rmsle(y_true, y_pred):
    return mean_squared_log_error(y_true, y_pred, squared=False)


[...]

PR #4915, opened 2023-07-19
[BUG] 0.21.0 release bugfix - fix interaction of `sklearn 1.3.0` with dynamic error metic based on `partial` in `test_make_scorer`

fkiraly:
This fixes an interaction of new input checking logic in `sklearn 1.3.0` with dynamic forecasting error metics based on `partial` that was uncovered `test_make_scorer`, regression test for #4827.


fkiraly:
fast merge due to issue on `main`

Issue #4889, opened 2023-07-15
[ENH] consolidating splitters as their own module with systematic tests and extension template


fkiraly:
Most recently, there have been a number of improvements, feature requests, and bug reports related to splitters.

This has highlighted the need to be able to add new splitters easily, and the relatively low systematic test coverage of interfaces, compared to other modules (e.g., forecasting). References below.

I think we should move the splitters on the same state as other mature modules, especially since they are frequently used.
This means:

- [ ] move them out of `model_selection`, and give them a proper module structure - no longer everything in a single file. Will require deprecation while it's moving.
- [ ] test class based on `TestAllObjects` and conditional fixture generation. The test class should cover all public API points, including `split`, `split_series`, `split_loc`, and `get_n_splits`.
- [ ] extension template for extenders who want to implement a custom splitter

#### Bugs and issues

* `ExpandingWindowSplitter` did not properly work for time-based index in an edge case https://github.com/sktime/sktime/pull/4886 
* unexpected behaviour of the forecasting `evaluate` benchmark, as it split time series internally but was not using splitters throughout https://github.com/sktime/sktime/issues/4842
* `get_n_splits` does not work for the window-based splitters for certain index combinations, see discussion in https://github.com/sktime/sktime/pull/4886

#### additions and feature requests

* the fix to #4842 which introduce two new splitters https://github.com/sktime/sktime/pull/4861
* the end-anchored expanding window splitter suggested by @davidgilbertson https://github.com/sktime/sktime/issues/4874 (works for partial type coverage, implemented using override, not template, due to lack of extension template)
* splitter control for `iloc` and `loc` indexing behaviour https://github.com/sktime/sktime/issues/4887, this will require proper testing
* `CutoffSplitter` with varying horizon suggested by @gianlucaguarrosonos https://github.com/sktime/sktime/issues/3682

#### FK's thoughts on further mid/low-priority improvements

* consider replacing broadcasting in splitters with native multi-index functionality (via the `"split_hierarchical"` tag)
* splitter delegator, similar to `_DelegatedForecaster` for quick definition of composite aliases
* splitter union
* splitter subsetting of index, e.g., generator generates only the i-th, j-th, etc for i, j in a subsetting index
* splitter compositor that producse only the splits where train set intersects with a given index set

FYI @hazrulakmal, @benheid, @davidgilbertson, @whoknowsB, @benshamza, @felipeangelimvieira, due to recent interest or work in splitters or being affected by the issues.

[...]

PR #4917, opened 2023-07-20
[ENH] add expanding greedy splitter

### LINKED ISSUE #4874: [ENH] Simplified expanding window splitter anchored at time series end
This is not so much a feature request but me sharing a thing I did that works nicely for me, in case others find it useful. Feel free to close.

The idea is that when defining cross validation, most of the time I just care about two things
1. the number of folds (which drives run time and variance of results) and
2. the horizon/number of test items 
 
By default, the splittier should use as much data as possible.

So this is another splitter that takes those two args:
```py
ExpandingGreedySplitter(horizon=7, folds=5)
```

This will create 5 folds, each with 7 days in the test set, arranged such that the test sets in total cover the last 35 days with no overlaps, and each train set is all data before the test set.

(I'm aware that having `horizon` as an int is not the sktime way but this is what works for me and how I think about the size of the validation set in cross validation, so I'm just sharing as is.)

It also takes a `step_length`, which defaults to the same as horizon for simplicity since that's what I want most of the time.

My work flow tends to be to set a low number for folds and horizon in the early stages of grid search to get a ballpark idea of what values work, then increase the folds and horizon to get more refined results. I found with the existing options this was a bit fiddly.

This is the class. A bit hacky in places, certainly not accounting for all use cases, but again, gets the job done for my uses.
```py
class ExpandingGreedySplitter(BaseSplitter):
    def __init__(self, horizon, folds=5, step_length=None):
        super().__init__()
        self.folds = folds
        self.horizon = horizon
        self.fh = np.arange(horizon) + 1
        self.step_length = step_length or horizon

    def split(self, y):
        y = self._coerce_to_index(y)
        if isinstance(y, pd.MultiIndex):
            dates = y.get_level_values(y.names[-1])
        else:
            dates = y

        if isinstance(dates, pd.PeriodIndex):
            freq = dates.freq
            dates = dates.to_timestamp()
        else:
            freq_str = pd.infer_freq(dates)
            freq = pd.tseries.frequencies.to_offset(freq_str)

        if freq is None:
            print("Couldn't work out frequency, assuming 'Day'")
            freq = pd.tseries.offsets.Day()

        trn_ends = pd.date_range(
            end=dates.max() - freq * self.horizon,
            periods=self.folds,
            freq=freq * self.step_length,
        )

        for trn_end in trn_ends:
            tst_end = trn_end + freq * self.horizon
            trn_indices = np.flatnonzero(dates <= trn_end)
            tst_indices = np.flatnonzero((dates > trn_end) & (dates <= tst_end))
            yield trn_indices, tst_indices
```

This assumes all instances _end_ at the same date but can have any start date (again, works for my case, not sure how often this assumption holds for others).

I originally wrote a `make_expanding_window_cv` function that used `ExpandingWindowSplitter` but this didn't work when my instances had different lengths.

davidgilbertson:
#### Reference Issues/PRs

Fixes #4874

This allows a user to define cross-validation splits in terms of `test_size` and number of folds.

Nope

I wasn't sure about the name of `test_size`, it could also be called `horizon` but I thought that was a bit misleading since it expects an `int` unlike 'forecast horizon'. Open to suggestions.

 - [ ] I just learned from this template about the contributors file, I'll do this tomorrow and update the PR.

Yes, but I wasn't sure about the existing machinery like `@pytest.mark.parametrize` or `get_test_params`. If I should be using these please let me know.

#### Any other comments?
My first ever Python PR, be gentle :)

##### For all contributions

##### For new estimators


fkiraly (commented):
Nice! Welcome to `sktime` pull requests!

What woul be great if you could add this to the API reference in `docs/source/api_reference/forecasting` next to the expanding window splitter, so users can find it easily!

fkiraly (changes requested):
Looks like it should work - the failure is due to the automated testing framework requiring a `get_test_params` function that returns valid parameter sets for parameters without a default.

There's currently no extension guide for splitters - this is tracked in issue https://github.com/sktime/sktime/issues/4889 - but you can have a look at any other extension guide, e.g., forecasters in `extension_templates/forecasting_simple`. I'd suggest we return 2 parameter sets, one where only `test_size` is set, and one where `folds` and `step_length` are additionally set to a non-default (while `test_size` is sth else than in the 1st set).

fkiraly:
@davidgilbertson, I hope it's ok that I added the `get_test_params` - just wanted to see the test execution results.
Feel free to revert.

fkiraly (dismissed):
I added the `get_test_params`, and it passes the tests now - so I'm fine with it, if you are fine with my changes.

Nice addition, should make "chop from the end" validation easier, especially if series are different lengths!

davidgilbertson:
Sorry for the delay, PR updated, should be ready for review/merge now.

fkiraly (approved):
Thanks! Looks good.

PR #4932, opened 2023-07-22
[ENH] improve `SARIMAX` test parameter coverage

janpipek:
#### Reference Issues/PRs

Fixes (partly) #3429 

This adds test_params for the SARIMAX forecaster.

No.

That the params make sense.

Nope.

#### PR checklist


.all-contributorsrc
```diff
@@ -2227,6 +2227,16 @@
...
       ]
+    },
+    {
+      "login": "janpipek",
+      "name": "Jan Pipek",
+      "avatar_url": "https://avatars.githubusercontent.com/janpipek",
+      "profile": "https://github.com/janpipek",
+      "contributions": [
+        "code",
+        "bug"
```
janpipek, line 2238:
```suggestion
```

sktime/forecasting/sarimax.py
```diff
@@ -221,3 +221,47 @@ def _extract_conf_int(prediction_results, alpha) -> pd.DataFrame:
...
+            {
+                "order": (2, 1, 2),
+                "trend": "ct",
+                "enforce_stationarity": False,
+                "enforce_invertibility": False,
+                "concentrate_scale": True,
+                "use_exact_diffuse": True,
+            },
+            {
+                "order": [1, 0, 1],
```
benHeid, line 261: Why do you uses here lists instead of tuples? In the previous dicts you are using tuples.

sktime/forecasting/sarimax.py
```diff
@@ -221,3 +221,47 @@ def _extract_conf_int(prediction_results, alpha) -> pd.DataFrame:
...
+        Returns
+        -------
+        params : dict or list of dict, default = {}
+            Parameters to create testing instances of the class
+            Each dict are parameters to construct an "interesting" test instance, i.e.,
+            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+            `create_test_instance` uses the first (or only) dictionary in `params`
+        """
+        return [
+            {
```
benHeid, line 245: To reduce the load on the testing system, It would be better to use only the last two more complex parameter sets, since the first two are almost covered by the others. @fkiraly what is your opinion?

benHeid (changes requested):
Thank you for the PR. I have only two small remarks. Feel free to add your opinions on that :)

sktime/forecasting/sarimax.py
janpipek: It is actually intentional :-)

The docs describe the orders as iterables, not tuple/list. So either should go and it is good to try.

fkiraly: yes, better test coverage! both tuple and list case are tested this way!

sktime/forecasting/sarimax.py
fkiraly: agreed (maybe ensure that list/tuple is also tested both)

fkiraly (changes requested):
Thanks!

The failures look like actual bugs, and we can´t merge failing tests to main, so I would suggest:

* report the bug - optimally, with a minimal reproducing example that involves no `check_estimator`, but a test run will also do
* comment out the offending parameter set with a reference to the bug report issue

fkiraly:
FYI @yarnabrina re bugs, and since this is coming from a `statsmodels` based estimator.

yarnabrina:
From traceback, it can be an upstream issue in statsmodels itself. I don't see `description` key is being passed by `sktime` anywhere.

fkiraly:
well, we found a bug in `statsmodels` then!

Way to go, @janpipek!

PR #4938, opened 2023-07-22
[ENH] interface to `statsforecast` ARCH family estimators

eyjo:
#### Reference Issues/PRs
Towards issue #4643.

direct interfaces to the ARCH family of models from Nixtla: ARCH and GARCH

No new dependency. It relies on statsforecast

This is my first attempt at implementing a forecast module for sktime. 

No

##### For all contributions

##### For new estimators

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -2227,6 +2227,16 @@
         "code",
         "bug"
       ]
-    }
+    },
+    {
+    "login": "eyjo",
+    "name": "Eyjólfur Sigurðsson",
+    "avatar_url": "https://avatars.githubusercontent.com/eyjo",
+    "profile": "https://github.com/eyjo",
+    "contributions": [
+      "code",
+      "doc"
+    ]
+  }
   ]
 }
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -2229,14 +2229,23 @@
       ]
     },
     {
-    "login": "eyjo",
-    "name": "Eyjólfur Sigurðsson",
-    "avatar_url": "https://avatars.githubusercontent.com/eyjo",
-    "profile": "https://github.com/eyjo",
-    "contributions": [
-      "code",
-      "doc"
-    ]
-  }
+      "login": "eyjo",
+      "name": "Eyjólfur Sigurðsson",
+      "avatar_url": "https://avatars.githubusercontent.com/eyjo",
+      "profile": "https://github.com/eyjo",
+      "contributions": [
+        "code",
+        "doc"
+      ]
+    },
+    {
+      "login": "julia-kraus",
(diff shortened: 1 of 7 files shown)


fkiraly (changes requested):
uh-oh, there are some linter/precommit failures!

But don't worry, this is only a small issue, I'll fix them so we can wrap up today with a geat success!

fkiraly:
additional note: there are a lot more parameters in `forecast` - let's deal with these in a separate PR.

@yarnabrina, does the generalized adapter class cover parameters in the `statsforecast` `forecast` method?

yarnabrina:
> does the generalized adapter class cover parameters in the `statsforecast` `forecast` method?

It does not. It uses `predict` and not `forecast`, and the two function signatures are quite different (only prediction vs train and prediction).

```python
    def predict(
        self,
        h: int,
        X_df: Optional[Union[pd.DataFrame, pl.DataFrame]] = None,
        level: Optional[List[int]] = None,
    ):
```

vs

```python
    def forecast(
        self,
        h: int,
        df: Optional[Union[pd.DataFrame, pl.DataFrame]] = None,
        X_df: Optional[Union[pd.DataFrame, pl.DataFrame]] = None,
        level: Optional[List[int]] = None,
        fitted: bool = False,
        sort_df: bool = True,
        prediction_intervals: Optional[ConformalIntervals] = None,
    ):
```

Ref. https://github.com/Nixtla/statsforecast/blob/main/statsforecast/core.py

fkiraly:
@eyjo, looks like this almost works, but you are passing an `approximation` param in the `get_test_params`, which the `__init__`-s don't have?

Should the inits have that parameter, or is this coming from sth else?

fkiraly (approved):
Thanks for fixing!

Looks all green now, we should be able to merge soon!

yarnabrina:
It's already merged, but I do have a small feedback here: I'd have suggested to have this in `statsforecast.py` instead of a new module. @fkiraly What do you think?

Or, if too big module is an issue, we can create a `statsforecast` directory, have a `__init__` and then have separate modules for separate set of estimators grouped together. Grouping logic can be same as the way Nixtla groups theire algorithms. This way, all `statsforecast` models can be imported easily in a single import.

(I just realised that this is different from how `statsmodels` forecasters are listed, and this PR follows that, but dependency wise grouping seemed like a nice idea to me.)

fkiraly:
> (I just realised that this is different from how statsmodels forecasters are listed, and this PR follows that, but dependency wise grouping seemed like a nice idea to me.)

That's a good point, @yarnabrina, although I don't know what the best ordering system is. I'm opening a new issue on the general point.

Regarding this specific estimator's location, we can move it anytime efore the next release.

PR #4946, opened 2023-07-23
[ENH] rewrite test_probabilistic_metrics using proper pytest fixtures

### LINKED ISSUE #4907: [ENH] rewrite `test_probabilistic_metrics` using proper `pytest` fixtures, or directly generate `pd.DataFrame`.
From profiling `pytest` test collection runs, we've discovered that, at current, major increase to runtime comes from `test_probabilistic_metrics` which fits and predicts every time upon module import.

This doesn't follow best `pytest` practice where fixture generation should not happen every time in the module, only at test time. The fit/predict also seems redundant, as we are only using the output.

There is a "software engineer" and a "data scientist" way to fix this, doing at least one should solve the problem - but it would be better to do both.

* the "software" appraoch would be moving the generation of the lists of `DataFrame`-s in lines 63 - 89 (`uni_data` etc) to `pytest` fixtures.
* the "data science" way would eliminate the need to fit/predict using the forecaster, after all we only use the output as a fixture, and it would be more robust to create these directly (i.e., using `pd.DataFrame` constructor) rather than getting them from an estimator's fit/predict

Good first issue for anyone who knows `pytest` or `pandas` - and it would help a lot of developers as the `pytest` test construction time is long.

julia-kraus:
#### Reference Issues/PRs
Fixes #4907 

Moved the sample data generation inside fixtures, so it doesn't run with each import, but only when testing is performed. 
Moreover refactored the tests such that the output tests for interval predictions and quantile predictions are separated.

Reason: With only one parametrized test were 96 parameter combinations, which made the test very slow and CPU heavy. 
Also, some parameter combinations are not necessary (quantile metrics belong to quantile forecast only and interval metrics only to interval forecast) - please correct me if I'm wrong. 
Separating the tests speeds up the testing process considerably and makes debugging easier.

No

I left the original `test_output` with the many parameter combinations in working and refactored state commented out. 
Decide if you still need it and if not, delete the commented out function.

Added the test `test_sample_data()` to check if the sample data generating fixture `sample_data()` is working correctly.

#### Any other comments?
The output test function contains many nested if - else statements. In my opinion, these should be separate tests with separate parametrized inputs, instead of putting all inputs in one test, and then distinguishing with if-else statements.

##### For all contributions

##### For new estimators


fkiraly (changes requested):
NICE! This speeds up test collection time massively for me.

Minor but important thing - linting is failing.

See here for linting & precommit:

fkiraly:
> Also, some parameter combinations are not necessary (quantile metrics belong to quantile forecast only and interval metrics only to interval forecast) - please correct me if I'm wrong.

Hm, I think they should work at least the way round that quantile metrics should be applicable to quantile for interval forecasts - this makes mathematical sense since interval upper/lower define quantile forecasts. If this was tested before and didn't raise errors, it means it's actually implemente that way.

julia-kraus:
ok, then I'd leave the original big function in the hope that the test server can handle it :)
thumbs up?

fkiraly (dismissed):
Thanks!

On details regarding what exactly to test, I think we ought to revisit this anyway with a class inheriting from `TestAllObjects`, so I'd not spend too much effort on the details of what is tested.

As far as I can see, your code is an 1:1 translation.

julia-kraus:
@fkiraly 
So I kept the two different functions for the tests for intervals and quantiles, but run all metrics on both.

The tests finish green on my machine. However, both take an hour to complete. This should be fine, but if you're looking for performance improvements in the future, that might be the place to look :)

fkiraly:
> There's a memory error on the test runner as well.. I guess the vectorized tests from before

This is unrelated to your changes - it is an instance of this sporadic error
https://github.com/sktime/sktime/issues/4610
which we haven't managed to properly diagnose yet.

PR #4948, opened 2023-07-24
[DOC] correct docstring of `BaseForecaster.score`, reference to use of non-symmetric MAPE

### LINKED ISSUE #4947: [DOC] BaseForecaster score function uses MAPE instead of sMAPE as metric

The [score function](https://github.com/sktime/sktime/blob/ed4efe9e47fa998ad6b21d63b0ac58f932ab1b3f/sktime/forecasting/base/_base.py#L1234C1-L1234C1) of the BaseForecaster suggests the metric that is used for evaluation is sMAPE:

```
Returns

        -------

        score : float
            sMAPE loss of self.predict(fh, X) with respect to y_test.
```
            
The code calls the [mean_absolute_percentage_error function](https://github.com/sktime/sktime/blob/ed4efe9e47fa998ad6b21d63b0ac58f932ab1b3f/sktime/performance_metrics/forecasting/_functions.py#L1448) without any further parameters. 

```     
from sktime.performance_metrics.forecasting import (
        )

        return mean_absolute_percentage_error(y, self.predict(fh, X))
``` 

But the default values of the [mean_absolute_percentage_error function ](https://github.com/sktime/sktime/blob/ed4efe9e47fa998ad6b21d63b0ac58f932ab1b3f/sktime/performance_metrics/forecasting/_functions.py#L1448) are: 

```
def mean_absolute_percentage_error(
    y_true,
    y_pred,
    horizon_weight=None,
    multioutput="uniform_average",
    symmetric=False,
    **kwargs,
):
```

and the doc states:

```
... 
    symmetric : bool, default=False
        Calculates symmetric version of metric if True.
...
```

Thus I conclude that instead of the sMAPE, MAPE is returned.

Either adjust the documentation such that it states that the MAPE metric is used or adjust the implementation by adding the symmetric parameter. I did not investigate if there are other places with the documentation error. ;) Many thanks!!

MBristle:
#### Reference Issues/PRs
Fixes #4947

The default value of the mean_absolute_percentage_error function changed to the non-symmetric MAPE. This has been made explicit, and the doc-string is adjusted.

 none 

Was the change from symmetric  - MAPE to non-symmetric MAPE intended for the score function of the BaseForecaster? 
How are down-stream tasks affected in either way?

Functionality was not changed.

##### For all contributions

##### For new estimators

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -2246,6 +2246,15 @@
       "contributions": [
         "doc"
       ]
+    },
+    {
+      "login": "MBristle",
+      "name": "Mirko Bristle",
+      "avatar_url": "https://avatars.githubusercontent.com/MBristle",
+      "profile": "https://github.com/MBristle",
+      "contributions": [
+        "doc"
+      ]
     }
   ]
 }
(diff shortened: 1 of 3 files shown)


fkiraly (changes requested):
Thanks for the fix!

One change request - the file `CONTRIBUTORS.md` should not be changed directly, this is automatically populated from `.all-contributorsrc` (so your change would be reverted at the next auto-update). Please change the latter, the `all-contributorsrc`, that's the "ground truth".

fkiraly (approved):
Looks good now!

fkiraly:
> Was the change from symmetric - MAPE to non-symmetric MAPE intended for
the score function of the BaseForecaster?

That's a good question - I think the default changed simply since the non-symmetric version is more common, and more in line with the user expectation when reading the class name. As long as the docstring is consistent, all should be fine?

Downstream tasks - like tuning - afaik are not affected as the user always specifies the metric themselves, and there the docstrings are correct.

PR #4956, opened 2023-07-25
[DOC] remove duplication of troubleshooting 'matches_not_found' in install instructions

### LINKED ISSUE #4955: [DOC] addressing 'matches not found' error in the installation guide 'setting up a dev environment'

In the issue #4933, the statement, `docs/source/installation.rst` in the 'setting up a development environment' section,
the 'matches not found error was addressed as a MacOS specific problem. However, it might just as well be specific to newer versions of the zsh shell. 

https://github.com/sktime/sktime/blob/f4815d7fcd15e4d8936dce770066062e36ed82a8/docs/source/installation.rst?plain=1#L180C1-L182C55

As the same error was already addressed in the troubleshooting section, and might also appear with other OS or shell combinations, it would make sense to leave the comment out and instead just refer to the more imprecise description in the troubleshooting section.

just refer to the trouble shooting

julia-kraus:
the comment referenced an issue that is addressed also in troubleshooting below. The troubleshooting also is more general.

#### Reference Issues/PRs
Fixes #4955 

In the developer setup, the 'matches_not_found' comment for installation on Mac was unnecessary, because it's already addressed in the troubleshooting in a more general manner.

no

no

#### Any other comments?
no

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -177,24 +177,26 @@ In the ``anaconda prompt`` terminal:
 
 5. Activate the environment: :code:`conda activate sktime-dev`
 
-6. Build an editable version of sktime :code:`pip install -e .[dev]`
-In order to install only the dev dependencies, use :code:`pip install -e .[dev]`
+6. Build an editable version of sktime.
+In order to install only the dev dependencies, :code:`pip install -e .[dev]`
 If you also want to install soft dependencies, install them individually, after the above,
 or instead use: :code:`pip install -e .[all_extras,dev]` to install all of them.
+
+    .. note::
+
+        If this step results in a "no matches found" error, it may be due to how your shell handles special characters.
+
+        - Possible solution: use quotation marks:
+
+            .. code-block:: bash
+
+                pip install -e ."[dev]"
 
 7. If everything has worked you should see message "successfully installed sktime"
 
 Some users have experienced issues when installing NumPy, particularly version 1.19.4.
 
-.. note::
 
-    If step 4. results in a "no matches found" error, it may be due to how your shell handles special characters.
-
-    - Possible solution: use quotation marks:
-
-        .. code-block:: bash
-
-            pip install -e ."[dev]"
 
 .. note::
 
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -179,7 +179,6 @@ In the ``anaconda prompt`` terminal:
 
 6. Build an editable version of sktime :code:`pip install -e .[dev]`
 In order to install all soft dependencies, use :code:`pip install -e .[dev]`
-If you are on Mac, use :code:`pip install -e ".[dev]"`
 If you also want to install soft dependencies, install them individually, after the above,
 or instead use: :code:`pip install -e .[all_extras,dev]` to install all of them.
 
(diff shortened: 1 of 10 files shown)


fkiraly (approved):
Yes, I think that addresses the issue - FYI @yarnabrina

docs/source/installation.rst
```diff
@@ -179,7 +179,6 @@ In the ``anaconda prompt`` terminal:
 
 6. Build an editable version of sktime :code:`pip install -e .[dev]`
 In order to install all soft dependencies, use :code:`pip install -e .[dev]`
```
yarnabrina, line 181: I've a small objection to the use of word `all`. Just skipping it, or changing it explaining that it's for code quality and testing dependencies, will be preferable to me.

docs/source/installation.rst
```diff
@@ -179,7 +179,6 @@ In the ``anaconda prompt`` terminal:
...
-If you are on Mac, use :code:`pip install -e ".[dev]"`
```
yarnabrina, line 182: I think it was a very good thought to add the point earlier, becuase a lot of users may face that error. I'll suggest to add it back without specifying Mac, or may be by adding a note something on the lines of:

> In case you get an error saying `no matches found: sktime[dev]`, use `pip install -e ".[dev]"`

yarnabrina (changes requested):

docs/source/installation.rst
fkiraly: looking at this, the command doesn't actually install the soft dependencies. Is this a copy-pastes mistake? Should this not include `all_extras`?

docs/source/installation.rst
fkiraly: it's already in the troubleshooting section, we could have it twice though (at the point where it occurs).
Or just point to the troubleshooting section

use -> try?

julia-kraus: sure!

fkiraly (approved):
Looks good to me, thanks!

docs/source/installation.rst
```diff
@@ -177,25 +177,26 @@ In the ``anaconda prompt`` terminal:
...
+
+            .. code-block:: bash
 
-    If step 4. results in a "no matches found" error, it may be due to how your shell handles special characters.
+                pip install -e ."[dev]"
 
-    - Possible solution: use quotation marks:
+7. If everything has worked you should see message "successfully installed sktime"
+
+Some users have experienced issues when installing NumPy, particularly version 1.19.4.
```
yarnabrina, line 197: Non-blocking suggestion: This can probably be removed, as `sktime` now needs `numpy>1.21`.

https://github.com/sktime/sktime/blob/b1921da2d613a900b9a9f6e0ce1319fe27b46f1b/pyproject.toml#L48

yarnabrina (approved):

docs/source/installation.rst
fkiraly: hm, didn't notice! Interesting, when/why did we introduce that lower bound?

yarnabrina: As per blame, it was done in #1921. Way before I was even aware of `sktime`, no idea.

fkiraly: ah, I asked the same question in #1921. It was said there were security risks below that bound, so I just took that on face value without checking. Probably true.

yarnabrina: Hi @julia-kraus, will you remove this line then? it's outdated and no more required. Everything else is in this PR looks good, and it can be merged then.

(not strictly related to the PR title, but probably it isn't worth a new PR just for one line removal.)

Issue #4925, opened 2023-07-21
[ENH] `test_reconstruct_identical` fails with hierarchical `dask` fixtures


fkiraly:
Since recently, `test_reconstruct_identical` is failing with hierarchical `dask` based fixtures.

(no further info at the moment)

PR #4957, opened 2023-07-25
[MNT] remove bound `dask<2023.7.1`

fkiraly:
Reverts sktime/sktime#4928, removes the temporary bound `dask<2023.7.1`.

`dask` `2023.7.2` was released, to diagnose if bug #4925 is still present.


fkiraly:
obsolete, as done by dependabot

PR #4958, opened 2023-07-25
[BUG] Add Informative Error messages for incompatible `scitype` in `BaseClusterer`

### LINKED ISSUE #4898: [BUG] Non informative Exception when fitting TimeSeriesKMeans with pd.Dataframe
**Describe the bug**
The fitting of TimeSeriesKMeans with pd.Dataframe produces an TypeError that is not informative. 

**To Reproduce**

```python
from sktime.clustering.k_means import TimeSeriesKMeans
import numpy as np
import pandas as pd
data = np.random.normal(0,1, (50, 12))
data_df = pd.DataFrame(data)
TimeSeriesKMeans().fit(data)
TimeSeriesKMeans().fit(data_df) # <- This line raises the TypeError
```

The resulting error message is:

```
TypeError: X is not of a supported input data type.X must be of type np.ndarray or pd.DataFrame, found <class 'pandas.core.frame.DataFrame'>Use datatypes.check_is_mtype to check conformance with specifications.
```

**Expected behavior**
Based on the discussion in Discord: The problem is that the the error message from `check_is_mtype` is not passed on. The problem is not that the python type is wrong, but that no panel specification is complied with (e.g., `pd-multiindex` needs 2 `pd.MultiIndex` levels). The error message should be more specific about the problem, e.g., by passing on the `check_is_mtype` error messages, such as in forecasting.

achieveordie:
#### Reference Issues/PRs

Fixes #4898

As discovered by @benHeid, the error messages were quite unhelpful in the case there was a mismatch between the supported and provided type. I've updated the code to be a slightly more generalized validation function and more information is provided in case of an error.

I also found a very minor bug in forecasting's validation function and took the liberty to correct it.

Finally, `clusterer` was not added by the original authors into the registry so I added that as well.

None

There are a bunch of `TODOs in `BaseForecaster`, we might want to take care of that in the upcoming PRs.

No test is required for this change.

##### For all contributions

##### For new estimators

Sorry for the delay!

### DIFF
--- a/sktime/clustering/base.py
+++ b/sktime/clustering/base.py
@@ -1,5 +1,5 @@
 """Base class for clustering."""
-__author__ = ["chrisholder", "TonyBagnall"]
+__author__ = ["chrisholder", "TonyBagnall", "achieveordie"]
 __all__ = ["BaseClusterer"]
 
 import time
@@ -9,7 +9,7 @@ import numpy as np
 import pandas as pd
 
 from sktime.base import BaseEstimator
-from sktime.datatypes import check_is_scitype, convert_to
+from sktime.datatypes import check_is_scitype, convert_to, scitype_to_mtype
 from sktime.utils.sklearn import is_sklearn_transformer
 from sktime.utils.validation import check_n_jobs
 from sktime.utils.validation._dependencies import _check_estimator_deps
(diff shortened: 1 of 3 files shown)


fkiraly (approved):
Looks nice!

Non-blocking comments, perhaps for other PR:

* I think there is a bit of copy-paste here (from forecaster or transformers?), so this now seems like a natural starting point for a refactor that moves out the error message generation, e.g., into the `datatypes` module (new utility?)
* the classifiers and regressors are very similar to the clusterer, the same places could be improved there, too - I think?

achieveordie:
There is a general question about how do we manage similar sanity checks across modules. In its current state, every module seems to implement its own version of validation on input data leading to repeated code, as you rightfully mentioned. Moving the validations into a separate entity brings the question of how modular we want our validations to be useful across modules.

Moving out the more obvious repetitions into a common utility seems trivial, but if we address the bigger issue [of how we centralize input validation] we might be able to get away with other surrounding issues.

fkiraly:
I would start with the error messages for now, coming from `check_is_mtype`, and only after that go for the validation logic.
I think it needs to be more modular, but I feel it might be a bit tricky at the moment, especially with error handling, error messages, and vectorization mixed together.

fkiraly:
Thanks for fixing the error message bug with the forecasting base class, btw.

Issue #4952, opened 2023-07-24
[BUG] evaluate returns incorrect results with multiple instances


davidgilbertson:
**Describe the bug**
The `evaluate` function appears to do the following steps, for a single fold of the passed CV:
* Split data into instances
* fit, predict, get score for each instance
* average the scores

I think this is wrong, it should first concatenate the predictions from all the instances and _then_ calculate the score.

Note that this is fine with a score like MSE, but not RMSE.

**To Reproduce**
The below two snippets should in theory produce the same results:

Manual train/test split, fit, predict:

```python
y_train, y_test, X_train, X_test = temporal_train_test_split(
    X=df.drop(columns=["Target"]),
    y=df[["Target"]],
    test_size=horizon,
)

pipe.fit(
    fh=fh,
    y=y_train,
    X=X_train,
)

y_pred = pipe.predict(X=X_test)

score = rmsle(y_test, y_pred)
```

Using `evaluate` and a splitter with just 1 fold:
```py
eval_results = evaluate(
    forecaster=pipe,
    cv=ExpandingGreedySplitter(folds=1, horizon=horizon),
    X=df.drop(columns=["Target"]),
    y=df[["Target"]],
    scoring=make_forecasting_scorer(rmsle, name="RMSLE"),
)
eval_score = eval_results.test_RMSLE.mean()
```

And they do match, if there's only one instance, but with more they start to vary.

Either I'm missing an option that allows me to bypass this behaviour or this is quite troublesome as it produces incorrect results.

Problem code is here, where it vectorizes an evaluator in the same way it vectorizes a forecaster:
https://github.com/sktime/sktime/blob/f4815d7fcd15e4d8936dce770066062e36ed82a8/sktime/performance_metrics/forecasting/_classes.py#L264-L282

**Versions**
<details>

System:
    python: 3.10.8 (main, Oct 12 2022, 19:14:26) [GCC 9.4.0]
executable: /home/davidg/.virtualenvs/learning/bin/python
   machine: Linux-5.15.90.1-microsoft-standard-WSL2-x86_64-with-glibc2.31
Python dependencies:
          pip: 23.2
       sktime: 0.21.0
      sklearn: 1.3.0
       skbase: 0.4.6
        numpy: 1.23.5
        scipy: 1.10.1
       pandas: 2.0.3
   matplotlib: 3.7.0
       joblib: 1.3.1
  statsmodels: 0.13.5
        numba: None
     pmdarima: 2.0.3
      tsfresh: None
[...]

PR #4960, opened 2023-07-26
[ENH] applying forecasting metrics disregarding index - docstrings and tests

fkiraly:
Fixes https://github.com/sktime/sktime/issues/4952 as follows:

It turns out that "pooled" application of a metric, i.e., ignoring index and passing the panel data as single 2D array was alreay possile, using `multilevel="uniform_average_time"`.

This is perhaps a bit of a misnomer, either way the docstrings did not inform the user of the option.

This has now been rectified:

* docstrings of (point forecast) metrics now all explain the multilevel argument and its possible values
* a test has been added to ensure that the parameter indeed corresponds to "ignoring level index"
* removes "hidden" input checks in some of the metric functions - these are covered by the base class, and inconsistently present in the functions. As they prevent the `"univorm_average_time"` logic in the four metrics they were still present, they have been removed.

### DIFF
--- a/sktime/performance_metrics/forecasting/_classes.py
+++ b/sktime/performance_metrics/forecasting/_classes.py
@@ -680,7 +680,7 @@ def make_forecasting_scorer(
         If array-like, values used as weights to average the errors.
         If 'raw_values', returns a full set of errors in case of multioutput input.
         If 'uniform_average', errors of all outputs are averaged with uniform weight.
-    multilevel : {'raw_values', 'uniform_average', 'uniform_average_time}
+    multilevel : {'raw_values', 'uniform_average', 'uniform_average_time'}
         Defines how to aggregate metric for hierarchical data (with levels).
         If 'uniform_average' (default), errors are mean-averaged across levels.
         If 'uniform_average_time', metric is applied to all data, ignoring level index.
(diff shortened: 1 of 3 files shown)


yarnabrina (commented):
Not a review, but a cursory look suggested that there is no example of how `multilevel` affects the results. May be that will be useful to users?

fkiraly:
> Not a review, but a cursory look suggested that there is no example of how `multilevel` affects the results. May be that will be useful to users?

Where would that example live?

fkiraly:
> It can be present in Examples sections like this:

All of them? There are dozens of metrics.

sktime/performance_metrics/forecasting/_classes.py
```diff
@@ -680,9 +680,10 @@ def make_forecasting_scorer(
         If array-like, values used as weights to average the errors.
         If 'raw_values', returns a full set of errors in case of multioutput input.
         If 'uniform_average', errors of all outputs are averaged with uniform weight.
-    multilevel : {'raw_values', 'uniform_average'}
+    multilevel : {'raw_values', 'uniform_average', 'uniform_average_time}
```
davidgilbertson, line 683: is this missing a closing quote?

fkiraly: yes, thanks for the pointer!

PR #4962, opened 2023-07-26
[MNT] speed up `test_gscv_proba` test

fkiraly:
This PR speeds up the `test_gscv_proba` test by cutting down on the data size and therefore the number of folds in the tuning.

Each run of the test takes ca 1min, at a total of 18 fixture combinations.


benHeid (approved):
Thank you :)

PR #4970, opened 2023-07-27
[ENH] Interface statsmodels MSTL

luca-miniati:

Interfaces statsmodels MSTL

##### For new estimators
- [x] If the estimator relies on a soft dependency, I've set the `python_dependencies` tag and ensured

### DIFF
--- a/sktime/forecasting/mstl.py
+++ b/sktime/forecasting/mstl.py
@@ -54,6 +54,10 @@ class MSTL(_StatsModelsAdapter):
     >>> plt.show()
     """
 
+    _tags = {
+        "python_dependencies": "statsmodels",
+    }
+
     def __init__(
         self,
         periods=None,
--- a/sktime/forecasting/mstl.py
+++ b/sktime/forecasting/mstl.py
@@ -1,0 +1,110 @@
+# !/usr/bin/env python3 -u
+# copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+"""Implements MSTL."""
+
+__all__ = ["MSTL"]
+__authors__ = ["luca-miniati"]
+
+from typing import Optional
+
+from sktime.forecasting.base.adapters import _StatsModelsAdapter
+
+
+class MSTL(_StatsModelsAdapter):
+    """Season-Trend decomposition using LOESS for multiple seasonalities.
+
+    Direct interface for `statsmodels.tsa.seasonal.MSTL`.
+
+    Parameters
+    ----------
+    endog : array_like
+        Data to be decomposed. Must be squeezable to 1-d.
+    periods : {int, array_like, None}, optional
+        Periodicity of the seasonal components. If None and endog is a pandas Series or
+        DataFrame, attempts to determine from endog. If endog is a ndarray, periods
+        must be provided.
+    windows : {int, array_like, None}, optional
+        Length of the seasonal smoothers for each corresponding period. Must be an odd
+        integer, and should normally be >= 7 (default). If None then default values
+        determined using 7 + 4 * np.arange(1, n + 1, 1) where n is number of seasonal
+        components.
+    lmbda : {float, str, None}, optional
+        The lambda parameter for the Box-Cox transform to be applied to endog prior to
+        decomposition. If None, no transform is applied. If “auto”, a value will be
+        estimated that maximizes the log-likelihood function.
+    iterate : int, optional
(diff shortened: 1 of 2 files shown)


sktime/forecasting/mstl.py
```diff
@@ -0,0 +1,114 @@
+# !/usr/bin/env python3 -u
```
yarnabrina, line 1: What is this shebang for? Can you please explain why is it required?

sktime/forecasting/mstl.py
```diff
@@ -0,0 +1,114 @@
...
+
+    _tags = {
+        "python_dependencies": "statsmodels",
+    }
+
+    def __init__(
+        self,
+        periods=None,
+        windows=None,
+        lmbda=None,
```
yarnabrina, line 65: Since you are type hinting the other arguments, do you want to do these as well for consistency?

sktime/forecasting/mstl.py
```diff
@@ -0,0 +1,114 @@
...
+        iterate=Optional[int],
```
yarnabrina, line 66:
```suggestion
        iterate: Optional[int] = None,
```

Did you mean this? Not sure if `Optional[int]` can be a valid default or not.

sktime/forecasting/mstl.py
```diff
@@ -0,0 +1,114 @@
...
+        stl_kwargs=Optional[dict],
```
yarnabrina, line 67:
```suggestion
        stl_kwargs: Optional[dict] = None,
```

Same as above, not sure if type hint can be a default value.

sktime/forecasting/mstl.py
```diff
@@ -0,0 +1,114 @@
...
+        self.lmbda = lmbda
+        self.iterate = iterate
+        self.stl_kwargs = stl_kwargs
+
+    def _fit_forecaster(self, y, X=None):
+        from statsmodels.tsa.seasonal import MSTL as _MSTL
+
+        self._forecaster = _MSTL(
+            y, self.periods, self.windows, self.lmbda, self.iterate, self.stl_kwargs
+        )
```
yarnabrina, line 82:
```suggestion
        self._forecaster = _MSTL(
            y, periods=self.periods, windows=self.windows, lmdba=self.lmbda, iterate=self.iterate, stl_kwargs=self.stl_kwargs
        )
```

`statsmodels` implementation has only one positional argument: `endog`, and rest are keyword arguments. Even though neither `statsmodels` not `sktime` use positional-only or keyword-only arguments so far, it may be better to pass these as `statsmodels` intended, to avoid later surprises.

The code suggestion is not `black` formatted, will fail code quality check.

sktime/forecasting/mstl.py
yarnabrina: Also, do you want to continue `2` as the default value as in `statsmodels`?

sktime/forecasting/mstl.py
```diff
@@ -0,0 +1,114 @@
...
+            `create_test_instance` uses the first (or only) dictionary in `params
+        """
+        params1 = {}
+        params2 = {
+            "periods": [1, 12],
+            "windows": 9,
+            "lmbda": "auto",
+            "iterate": 10,
+            "stl_kwargs": {"trend_deg": 0},
+        }
```
yarnabrina, line 112: Can you add another test case with `periods` being a single integer, and not an array of integer(s)? You can vary `windows` as well.

yarnabrina (changes requested):
There are few failures, which I think are being caused by using type hints as default values. Please take a look at fixing that.

Other comments are mostly non-blocking, minor suggestions.

fkiraly (changes requested):
Nice start!

I think there is a fundamental issue with the `statsmodels` MSTL estimator - it is not a forecaster afaik, but a (series-to-series) transformer.

We can turn this into both a transformer or a forecaster:

* a transformer would be a direct interface, similar to `STLTransformer`
* a forecaster needs to be structured similar to `STLForecaster`, which does the decomposition, makes individual forecasts for the components (the seasonal component just repeats), and then creates a forecasts out of the components

Either way, I think the adapter won't work here (is that correct, @yarnabrina)?

sktime/forecasting/mstl.py
luca-miniati: Not sure, I think I must have copy/pasted it from the LTSF code. Removing it now

PR #4976, opened 2023-07-29
DRAFT [ENH] runtime optimizations for reducer forecasters

fkiraly:
DRAFT PR with partial work to speed up the reworked reducers, towards https://github.com/sktime/sktime/issues/4359.

Current state:
I think I have identified the bottleneck that causes the long runtime in the example in #4359.

It is repeated calls to the `lag.fit_transform`, specifically in the loops in `_predict` - each call takes only 0.03 s, but in the example they add up to minutes, in the original version, and now to 50s in a slightly modified version.

This is still much worse than the 1s in the old version, but should be possibly to reduce - equivalent numpy calls take near zero seconds.

The offending line is marked, it is a call of `Lag` on a single row, which produces, in the example, a diagonal matrix of five lags.

FYI @davidgilbertson


ksharma6:
Hi @fkiraly  could a possible solution here be checking/converting X and y to np. arrays when when implementing fit_transform() in the BaseTransformer class? Are there any potential conflicts with that data type elsewhere I should be mindful of?

fkiraly:
But in general - any speed-up and help with getting this over hte finish line is papreciated!

fkiraly:
> This feels to be the most productive avenue to go down at this time.

👍 excellent - should we try in a separate PR then? It might be important to have a few tests in place for multivariate and hierarchical for input/output consitency before and after the change.

> For timing, is there a built in `sktime` function I should use or would the time python library suffice?

A while ago, I wrote a utility for classifiers, in `utils.profiling`, but there is no such utility for forecasters or transformers.
I have only unconsolidated thoughts here, my top three:

* it would be nice to have sth like this for forecasters and transformers, so we could compare timings by a common standard
* however, in the reducers, timings of subroutines are probably going to be important, so a profiler like that will not necessarily tell us where the time gets lost
* @tarpas used `py-spy` a while ago to profile test execution time in a very granular way, this could be combined with the idea of a profiler function to avoid manual `time` hacking. Recipe and python command are in this discussion: https://github.com/sktime/sktime/issues/4900

ksharma6:
Yes, let's do that

> @tarpas used py-spy a while ago to profile test execution time in a very granular way, this could be combined with the idea of a profiler function to avoid manual time hacking. Recipe and python command are in this discussion: [BUG] 60 seconds per core of unnecessary delay for pytest --collect-only execution  #4900

Somthing like this would be super helpful and provides the granulatiry needed to look at subroutines

fkiraly:
you mean, a profiling utility for transformers? That would be a neat addition, and we can directly test it with `Lag`!

(I'd split the `profiling` module up, into a folder with private submodules `_classifier`, `_transformer`, etc)

fkiraly:
> is this a bug somewhere else?

Not sure, we'd need to see your code or your environment versions.

ksharma6:
Hi @fkiraly 

I had a chance to do more trouble shooting and I am making no progress. Here is the code that triggers my bug:
pandas_bug

**What I've Tried**
- Updated branch to make current with main 
- Updated branch and ran `pip install --editable .` to update my venv for sktime

Below is my requirements.txt
[requirements.txt](https://github.com/sktime/sktime/files/14765934/requirements.txt)

I feel like I am missing something obvious here, please let me know your thoughts whenever you have a moment.

fkiraly:
@ksharma6, it seems your environment is broken, or uses old versions? That would be my educated guess. You can raise this in the weekly meet-up where we help people getting set up with their python environments etc.

ksharma6:
will do @fkiraly I should be able to stop by this week and we can go over it. 

There were the docs I referred to when making a new venv. I used the instructions for the development versions, please let me know if there is something else to refer to instead:
https://www.sktime.net/en/stable/installation.html#step-1-clone-the-git-repository

fkiraly:
FYI @hliebert, may I kindly ask you to look at this PR, it contains some speedups, but does not resolve all speed issues.

There are some diagnostic prints in it currently, but if removed the result should already be a speedup.

PR #4978, opened 2023-07-29
[ENH] enable use of `TabularToSeriesAdaptor` with feature selectors, and passing of `y`

fkiraly:
This PR extends `TabularToSeriesAdaptor` so it is compatible with feature selectors such as `SelectKBest` or `VarianceThreshold`, and allows passing of `y`.

Addresses feature request in https://github.com/sktime/sktime/discussions/4967.

* the default behaviour for an inner estimator that requiesa `y` in `fit` (i.e., having a defaultless named argument `y`) is to pass on any `y` seen in `fit`, and only pass `X` otherwise (i.e., if the inner estimator does not have an arg `y`, or one with a default); and, never passing `y` to `transform`.
* the behaviour can be modified by the new `pass_y` argument to pass never, always (including to `transform`), or to `fit` (covering the case where `y` can be passed but is not a named argument)

No deprecation is needed as far as I see, because the estimator would crash in cases where `y` was required, previously.

Also fixes a minor docstring hiccup where the `fit_in_transform` arg was erroneously in the "attributes" section, and not the "parameters" section where it should have been.

### DIFF
--- a/sktime/transformations/series/adapt.py
+++ b/sktime/transformations/series/adapt.py
@@ -73,16 +73,16 @@ class TabularToSeriesAdaptor(BaseTransformer):
         whether transformer_ should be fitted in transform (True), or in fit (False)
             recommended setting in forecasting (single series or hierarchical): False
             recommended setting in ts classification, regression, clustering: True
-    pass_y : str, optional, one of "auto" (default), "fit", "fit_transform", "no"
+    pass_y : str, optional, one of "auto" (default), "fit", "always", "never"
         Whether to pass y to transformer methods of the ``transformer`` clone.
         "auto": passes y to methods fit, transform, fit_transform, inverse_transform,
             if and only if y is a named arg of either method.
             Note: passes y even if it is None
         "fit": passes y to method fit, but not to transform.
             Note: passes y even if it is None, or if not a named arg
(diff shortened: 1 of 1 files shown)


hliebert:
Sorry for being late with feedback. I've tested this quite a bit, both the original draft and the more conservative default and it works fine for me. Is there a reason for the more conservative choice to not pass `y` if there is a default value? I think if there is a default for `y`, it is typically `None` in most places (e.g. `SelectFromModel` in `sklearn`).

fkiraly:
> I've tested this a bit, both the original draft and the more conservative default and it works fine for me. Is there a reason for the more conservative choice to not pass `y` if there is a default value?

The high-level reason was, the `TabularToSeriesAdaptor` is used in many downstream cases, including in `sktime`, some of these apparently with a `y` that has a default of `None`.

This can happen even for `sklearn` estimators that have `y` just for interface compatibility, which will fail if anything else than `y=None` is passed to them - so that would make all those cases fail!

That is, it would make existing code break, and the only guarantee to not make existing code break is pass the `y` only where the current code would break.

fkiraly:
(very pragmatically, I noticed when tests started breaking with the original draft default, which showed me that there are already cases we rely upon, where `y` is present but has a default, and we pass nothing to `y`)

PR #4984, opened 2023-07-29
[ENH] fix `Differencer` for integer index

fkiraly:
This fixes an unreported bug in `Differencer._inverse_transform` for integer index.

In `_inverse_transform`, `Differencer` would call the `shift` method of the index which is not defined for integer based index - this should have used the internal `_shift` method.

The tests did not catch this as all tests were datetime-index based.
This has now been rectified.

### DIFF
--- a/sktime/transformations/series/tests/test_differencer.py
+++ b/sktime/transformations/series/tests/test_differencer.py
@@ -42,8 +42,11 @@ def test_differencer_produces_expected_results(na_handling):
 
 @pytest.mark.parametrize("y", test_cases)
 @pytest.mark.parametrize("lags", lags_to_test)
-def test_differencer_same_series(y, lags):
+@pytest.mark.parametrize("index_type", ["int", "datetime"])
+def test_differencer_same_series(y, lags, index_type):
     """Test transform against inverse_transform."""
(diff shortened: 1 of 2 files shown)


sktime/transformations/series/tests/test_differencer.py
```diff
@@ -42,8 +42,11 @@ def test_differencer_produces_expected_results(na_handling):
 
 @pytest.mark.parametrize("y", test_cases)
 @pytest.mark.parametrize("lags", lags_to_test)
-def test_differencer_same_series(y, lags):
+@pytest.mark.parametrize("index_type", ["int", "datetime"])
```
benHeid, line 45: Not sure if this is better. As an alternative, we could add the time series with resetted index also to the `test_cases` list.

fkiraly: that would double the number of test cases, and fixture paramterization logic in the module root - I think there is already too much logic in the module root and not in `pytest.fixture`-s ...

fkiraly: if one wants to be really clean, it should be a parametric pytest fixture

PR #4987, opened 2023-07-30
[BUG] use correct arguments in `geometric_mean_absolute_error`

### LINKED ISSUE #4986: [BUG] `geometric_mean_squared_error` and `geometric_mean_absolute_error` uses undefined keyword argument `sample_weight`
Both of `geometric_mean_squared_error` and `geometric_mean_absolute_error` uses `_weighted_geometric_mean` underneath when `horizon_weight` is passed, these weights are passed to `sample_weight` argument. However, that argument does not exist, and probably `weights` is the indented argument.

**To Reproduce**

```pycon
>>> from sktime.performance_metrics.forecasting import geometric_mean_squared_error
>>> geometric_mean_squared_error([1, 2, 3], [6, 5, 4], horizon_weight=[7, 8, 9])
```

This leads to:

> TypeError: _weighted_geometric_mean() got an unexpected keyword argument 'sample_weight'

yarnabrina:
Fixes #4986

### DIFF
--- a/sktime/performance_metrics/forecasting/_functions.py
+++ b/sktime/performance_metrics/forecasting/_functions.py
@@ -1302,9 +1302,7 @@ def geometric_mean_absolute_error(
     else:
         check_consistent_length(y_true, horizon_weight)
         output_errors = _weighted_geometric_mean(
-            np.abs(errors),
-            sample_weight=horizon_weight,
-            axis=0,
+            np.abs(errors), weights=horizon_weight, axis=0
         )
 
     if isinstance(multioutput, str):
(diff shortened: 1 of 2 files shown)


fkiraly (changes requested):
Looks straightforward!

I'm worried that no test caught this - would you mind adding a test along the lines of #4986?
Bonus would be underrstanding why the suite tests didn't catch this, but not blocking if we don't understand this yet.

yarnabrina:
> underrstanding why the suite tests didn't catch this

Assuming `sktime/performance_metrics/forecasting/tests/test_metrics.py` and `sktime/performance_metrics/tests/test_performance_metrics_forecasting.py` are running all the metrics tests, there is no test that uses weights. Not just for these two metrics, but for all metrics in general.

fkiraly (approved):
Thanks.

We should probably introduce more systematic testing for the metrics later.

PR #4990, opened 2023-07-30
[MNT] speed up `test_stat` benchmarking test

fkiraly:
This speeds up the `test_stat` benchmarking test from 60s to in the order of seconds, by using a smaller dataset.

Related to: https://github.com/sktime/sktime/issues/2890


achieveordie (commented):
Is there any reason to have even 7 datapoints, if our tests are independent of the dataset size used, we might as well move with the smallest allowable subset.

fkiraly:
No specific reason for 7, I just think below there is a too high risk of degeneracy for a 2-class classification dataset. That means at most 3 in the smaller class, any less will make variance computations (or model computations) fail, possibly sporadically.

achieveordie (approved):
Makes sense! LGTM

PR #4991, opened 2023-07-30
[ENH] ensure robustness of `StatsForecastBackAdapter` w.r.t. change of `predict_interval` return format

fkiraly:
This PR changes the `StatsForecastBackAdapter` to ensure it is not affected by the change of `predict_interval` return format.

This is done by simply dropping the first column index level, which should therefore avoid any impact of the default change.

Note to @yarnabrina, @luca-miniati - it sems the adapter assumes a univariate forecaster. Is this intentional, and is this sufficient?

### DIFF
--- a/sktime/forecasting/base/adapters/_generalised_statsforecast.py
+++ b/sktime/forecasting/base/adapters/_generalised_statsforecast.py
@@ -384,14 +384,16 @@ class StatsForecastBackAdapter:
     def format_pred_int(self, y_pred_name, y_pred, pred_int, coverage, level):
         pred_int_prefix = "fitted-" if y_pred_name == "fitted" else ""
 
+        pred_int_no_lev = pred_int.droplevel(0, axis=1)
+
         return {
             y_pred_name: y_pred,
             **{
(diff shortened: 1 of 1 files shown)


yarnabrina:
I do not know. I noticed it before and raised at least twice:

1. https://github.com/sktime/sktime/pull/4865#discussion_r1265561260
2. https://github.com/sktime/sktime/pull/4865#issuecomment-1652917339

Unfortunately, I didn't get much time to test after merge into main, so I suppose only @luca-miniati can answer.

luca-miniati:
This behavior was intentional, and I think it is sufficient for now. Currently, `statsforecast` doesn't have any composite models that take multivariate forecasters, and for that matter, they don't even have any multivariate forecasters. So, I don't see how we can write an adapter when we don't know the behavior that the `statsforecast` base model (e.g `MSTL`) will expect from the multivariate ensemble model (e.g. `VAR`). We can infer the behavior, but I think it's likely that we would have to take a second look to ensure compatibility. Therefore, I think that the adapter should support multivariate data in the event that `statsforecast` releases both of these:
 - a base estimator that takes a multivariate forecaster
 - multivariate forecaster

Thoughts?

PR #4998, opened 2023-08-01
&MCRE-BE [BUG] Fix `Differencer.inverse_transform` on train data if `na_handling=fill_zero`

### LINKED ISSUE #4965: [BUG] 'Differencer' returning strange results
**Describe the bug**

When calling fit_transform and inverse_transform sequentially, I don't get the original series back when applying one or several lags. It seems to work fine with a PeriodIndex (like the example in the documentation), but does not work with a datetimeindex. I reproduced the issue with one of the default datasets included with SKTime (but I need to run a simple groupby to get it to daily level to match my own).

**To Reproduce**
```python
from sktime.datasets import load_solar
y = load_solar()
y = y.groupby(by=pd.Grouper(level=0, freq="D")).sum()
display(y.plot())
```
image

```python
transformer = Differencer(lags=[1, 7])
y_transform = transformer.fit_transform(y)
display(y_transform.plot())
```
image

```python
y_revert = transformer.inverse_transform(y_transform)
y_revert = pd.DataFrame(y_revert, index=y_transform.index)
y_revert.plot()
```
image

**Expected behavior**
The returned series should be identical (or very close)

**Additional context**
I have the same problem with a much longer timeseries.

**Versions**
<details>
System:
    python: 3.11.4 | packaged by conda-forge | (main, Jun 10 2023, 17:59:51) [MSC v.1935 64 bit (AMD64)]
executable: [c:\ProgramData\Miniconda3\envs\DLL_Forecast\python.exe](file:///C:/ProgramData/Miniconda3/envs/DLL_Forecast/python.exe)
   machine: Windows-10-10.0.19042-SP0

Python dependencies:
          pip: 23.2
       sktime: 0.21.0
      sklearn: 1.3.0
       skbase: 0.4.6
        numpy: 1.24.4
        scipy: 1.11.1
       pandas: 2.0.3
   matplotlib: 3.7.2
       joblib: 1.3.0
  statsmodels: 0.14.0
        numba: 0.57.1
     pmdarima: None
      tsfresh: None
   tensorflow: None
tensorflow_probability: None

</details>

### LINKED ISSUE #4993: [BUG] Wrong result for `predict_residuals` of`TransformedTargetForecaster` on training data
**Describe the bug**
If `predict_residuals` is called  of a `TransformedTargetForecaster` on the training data, the results seems to be wrong. My guess it that internally, the  `self._y` of the `TransformedTargetForecaster` is used to calculate the residuals instead of the transformed target data. 

**To Reproduce**

```python
from datasets import load_longley
from forecasting.model_selection import temporal_train_test_split
from forecasting.sarimax import SARIMAX
from transformations.series.difference import Differencer

method = "predict_residuals"
y, X = load_longley()
pipe = Differencer() * SARIMAX()
pipe.fit(y=y_train.to_frame(), X=X_train, fh=[1, 2, 3, 4])
result = pipe.predict_residuals()
result_test = pipe.predict_residuals(y=y_test, X=X_test)
```

In this example, result_test looks like residuals. However, result is far higher and not the residuals.

**Expected behavior**

Also on the training data the residuals should be calculated of the TransformedTargetForecaster applied on the training data.

benHeid:
Set first lag values to nan if it has the same index as the train time series. This should fix the strange results in inverse_transform of the Differencer if fill_zero is set.

#### Reference Issues/PRs
Fixes #4993 and fixes #4965.

Regarding both issues, I am not sure if there are additional problems... 

The fix sets the first lag values of the input time series to nan if the index aligns with the index of the train time series. Thus, the  combine_first operation will set it to the values of the original time series and cumsum reproduces the original time series.

No

* Is the solution understandable and maintainable or looks it like black magic.
* Regarding both linked issues, are they solved with this single bugfix or are additional problems there.. Currently, I am not sure about it. 

I added a test for reproducing the original bug.

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/transformations/series/difference.py
+++ b/sktime/transformations/series/difference.py
@@ -147,7 +147,7 @@ def _inverse_diff(X, lags, X_diff_seq=None):
         X_update = X_diff_orig.loc[X_ix_shift.intersection(X_diff_orig.index)]
         X.loc[
             X_diff_orig.index.difference(
-                _shift(X_diff_orig.index, lag_last)
+                _shift(X_diff_orig.index, sum(lags) + lag_last)
             ).intersection(X.index)
         ] = np.nan
         X = X.combine_first(X_update)
--- a/sktime/transformations/series/difference.py
+++ b/sktime/transformations/series/difference.py
@@ -142,9 +142,19 @@ def _inverse_diff(X, lags, X_diff_seq=None):
 
     # invert last lag index
     if X_diff_seq is not None:
+        # Get the train time series before the last difference
         X_diff_orig = X_diff_seq[len(lags)]
+        # Shift the differenced time series index by the last lag
+        # to match the original time series index
         X_ix_shift = _shift(X.index, -lag_last)
+        # Get the original time series values for the intersecting
+        # indices between the shifted index and the original index
         X_update = X_diff_orig.loc[X_ix_shift.intersection(X_diff_orig.index)]
+        # Set the values of the differenced time series to nan for all indices
+        # that are in the indices of the original and the by the sum of all lags
+        # shifted original time series that are available in the differenced time
+        # series (intersection). These are the indices for which no valid differenced
+        # values exist.
         X.loc[
             X_diff_orig.index.difference(
                 _shift(X_diff_orig.index, sum(lags) + lag_last)
--- a/sktime/transformations/series/difference.py
+++ b/sktime/transformations/series/difference.py
@@ -145,7 +145,11 @@ def _inverse_diff(X, lags, X_diff_seq=None):
         X_diff_orig = X_diff_seq[len(lags)]
         X_ix_shift = _shift(X.index, -lag_last)
         X_update = X_diff_orig.loc[X_ix_shift.intersection(X_diff_orig.index)]
-        X.loc[X.index.intersection(X_diff_orig.index[:lag_last])] = np.nan
+        X.loc[
+            X_diff_orig.index.difference(
+                _shift(X_diff_orig.index, lag_last)
+            ).intersection(X.index)
+        ] = np.nan
         X = X.combine_first(X_update)
 
     X_diff_last = X.copy()
--- a/sktime/transformations/series/difference.py
+++ b/sktime/transformations/series/difference.py
@@ -145,7 +145,7 @@ def _inverse_diff(X, lags, X_diff_seq=None):
         X_diff_orig = X_diff_seq[len(lags)]
         X_ix_shift = _shift(X.index, -lag_last)
         X_update = X_diff_orig.loc[X_ix_shift.intersection(X_diff_orig.index)]
-
+        X.loc[X.index.intersection(X_diff_orig.index[:lag_last])] = np.nan
         X = X.combine_first(X_update)
 
     X_diff_last = X.copy()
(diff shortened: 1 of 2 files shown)


sktime/transformations/series/difference.py
```diff
@@ -145,7 +145,7 @@ def _inverse_diff(X, lags, X_diff_seq=None):
         X_diff_orig = X_diff_seq[len(lags)]
         X_ix_shift = X.index.shift(-lag_last)
         X_update = X_diff_orig.loc[X_ix_shift.intersection(X_diff_orig.index)]
-
+        X.loc[X.index.intersection(X_diff_orig.index[:lag_last])] = np.nan
```
fkiraly, line 148: instead of taking "first lag_last", should we work with `_shift` instead?

Because the index might not be contiguous. That is, instead of the intersection, take the difference with `_shift` of `lag_last`?

Also see https://github.com/sktime/sktime/pull/4984

benHeid: Good point. I need to test it with the shift. Probably the best is if I rebase on the branch of the linked PR.

Thank you for linking the other PR.

benHeid: Ok, I rebased. 
Happy to add a comment if this line is too complicated.

fkiraly: yes, I think comments in the code would be helpful!
I had myself some problems reading this again after a year.

fkiraly:
@benHeid, while we're at it, could you check whether this fixes https://github.com/sktime/sktime/issues/4965? If yes, add it as a test case? If not, can we try to understand why not?

benHeid:
> @benHeid, while we're at it, could you check whether this fixes #4965? If yes, add it as a test case? If not, can we try to understand why not?

I can add the same code as in the issue if you like. However, I am convinced that the added test already covers the code in the issue. For further explanation, see the code snippet below, which extends the code snippet from the issue with comments to explain why the test case should cover this code snippet. 

```python
from sktime.datasets import load_solar
import pandas as pd
y = load_solar()
y = y.groupby(by=pd.Grouper(level=0, freq="D")).sum() # Aggregating time series makes no difference for the Differencer.
# display(y.plot())

transformer = Differencer(lags=[1, 7]) # The lags is variable in the added test case since we use parametrize.
y_transform = transformer.fit_transform(y) # In the added tests, this line is combined with the inverse_transform
# display(y_transform.plot()) # Plots are not required in the test

y_revert = transformer.inverse_transform(y_transform)
# y_revert = pd.DataFrame(y_revert, index=y_transform.index) <- This line is not required in the tests, it just tests if pandas works correctly.
# y_revert.plot()

assert_array_almost_equal(y_revert, y)
```

fkiraly:
> However, I am convinced that the added test already covers the code in the issue. 

Yes, I see - because the `lags` includes a double lag, `[1, 12]`. Makes sense, in this case I don't think we need to add another test.

sktime/transformations/series/difference.py
fkiraly: addressed

fkiraly (approved):
Thanks!

Doesn't look like much in terms of number of lines, but I want to express my appreciation since this part of the code was really hard to understand when I recently investigated. Not easy!

fkiraly:
added credit to @MCRE-BE for investigation in #4965

PR #5001, opened 2023-08-01
[ENH] Handle array centroid guesses for TimeSeriesLloyds

### LINKED ISSUE #4863: [ENH] Initial centroid guess for kmeans

Alex-JG3:
#### Reference Issues/PRs

Closes #4863.

Allows numpy arrays representing the initial centroid guesses to be passed into `init_algorithm` for `TimeSeriesLoyds`.

No

Not yet

##### For all contributions

##### For new estimators

### DIFF
--- a/sktime/clustering/partitioning/_lloyds.py
+++ b/sktime/clustering/partitioning/_lloyds.py
@@ -267,7 +267,7 @@ class TimeSeriesLloyds(BaseClusterer, ABC):
         if isinstance(self.init_algorithm, str):
             if self.init_algorithm not in self._init_algorithms.keys():
                 raise KeyError(
-                    f"The string provided for init_algorim: {self.init_algorithm} is "
+                    f"The string provided for init_algorithm: {self.init_algorithm} is "
                     f"invalid. The following are a list of valid init algorithms "
                     f"strings: {list(self._init_algorithms.keys())}"
                 )
@@ -278,7 +278,7 @@ class TimeSeriesLloyds(BaseClusterer, ABC):
             self._init_algorithm = self.init_algorithm
         else:
             raise ValueError(
-                f"The value provided for init_algorim: {self.init_algorithm} is "
+                f"The value provided for init_algorithm: {self.init_algorithm} is "
                 f"invalid. The following are a list of valid init algorithms strings: "
                 f"{list(self._init_algorithms.keys())}"
             )
--- a/sktime/clustering/partitioning/_lloyds.py
+++ b/sktime/clustering/partitioning/_lloyds.py
@@ -265,14 +265,13 @@ class TimeSeriesLloyds(BaseClusterer, ABC):
         self._random_state = check_random_state(self.random_state)
 
         if isinstance(self.init_algorithm, str):
-            try:
-                self._init_algorithm = self._init_algorithms[self.init_algorithm]
-            except KeyError as e:
+            if self.init_algorithm not in self._init_algorithms.keys():
                 raise KeyError(
                     f"The string provided for init_algorim: {self.init_algorithm} is "
                     f"invalid. The following are a list of valid init algorithms "
                     f"strings: {list(self._init_algorithms.keys())}"
-                ) from e
+                )
+            self._init_algorithm = self._init_algorithms[self.init_algorithm]
         elif isinstance(self.init_algorithm, np.ndarray):
             self._init_algorithm = self.init_algorithm
         elif isinstance(self._init_algorithm, Callable):
--- a/sktime/clustering/partitioning/_lloyds.py
+++ b/sktime/clustering/partitioning/_lloyds.py
@@ -274,6 +274,17 @@ class TimeSeriesLloyds(BaseClusterer, ABC):
                 )
             self._init_algorithm = self._init_algorithms[self.init_algorithm]
         elif isinstance(self.init_algorithm, np.ndarray):
+            # if self.init_algorithm.ndim != 3:
+            #     raise ValueError(
+            #         "The array provided to init_algorithm is invalid. It must be 3D."
+            #     )
+            # if self.init_algorithm.shape[1] != X.shape[1]:
+            #     raise ValueError(
+            #         "X and init_algorithm must be 3D arrays of shape "
+            #         "(n_instances, n_dimensions, series_length) such that
+            # n_dimensions "
+            #         "is the same for both arrays."
+            #     )
             self._init_algorithm = self.init_algorithm
         elif isinstance(self._init_algorithm, Callable):
             self._init_algorithm = self.init_algorithm
--- a/sktime/clustering/partitioning/_lloyds.py
+++ b/sktime/clustering/partitioning/_lloyds.py
@@ -162,9 +162,10 @@ class TimeSeriesLloyds(BaseClusterer, ABC):
     n_clusters: int, defaults = 8
         The number of clusters to form as well as the number of
         centroids to generate.
-    init_algorithm: str, defaults = 'forgy'
-        Method for initializing cluster centers. Any of the following are valid:
-        ['kmeans++', 'random', 'forgy']
+    init_algorithm: str, np.ndarray (3d array of shape (n_clusters, n_dimensions,
+        series_length)), defaults = 'forgy'
+        Method for initializing cluster centers or an array of initial cluster centers.
+        Any of the following strings are valid: ['kmeans++', 'random', 'forgy']
     metric: str or Callable, defaults = 'dtw'
         Distance metric to compute similarity between time series. Any of the following
         are valid: ['dtw', 'euclidean', 'erp', 'edr', 'lcss', 'squared', 'ddtw', 'wdtw',
--- a/sktime/clustering/partitioning/_lloyds.py
+++ b/sktime/clustering/partitioning/_lloyds.py
@@ -242,10 +242,6 @@ class TimeSeriesLloyds(BaseClusterer, ABC):
         self._init_algorithm = None
         self._initial_cluster_centers = None
 
-        if isinstance(self.init_algorithm, np.ndarray):
-            self._initial_cluster_centers = self.init_algorithm
-            self.init_algorithm = lambda *args, **kwargs: self._initial_cluster_centers
-
         self._distance_params = distance_params
         if distance_params is None:
             self._distance_params = {}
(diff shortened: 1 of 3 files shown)


sktime/clustering/partitioning/_lloyds.py
```diff
@@ -240,6 +240,11 @@ def __init__(
 
         self._random_state = None
         self._init_algorithm = None
+        self._initial_cluster_centers = None
+
+        if isinstance(self.init_algorithm, np.ndarray):
+            self._initial_cluster_centers = self.init_algorithm
+            self.init_algorithm = lambda *args, **kwargs: self._initial_cluster_centers
```
fkiraly, line 247: I think this should be `self._init_algorithm` (with a leading underscore).

Also, to avoid spreading this logic out, either move this into `_check_params`, or move the `_init_algorithm` logic from `_check_params` here.

Alex-JG3, line 247: I could use some feedback on whether there is a better way than using this lambda function to return the initial cluster centers.

I cannot use a class method called something like `return_initial_cluster_centers` because that causes a recursion error. For example, suppose this class has the method
```python
def return_initial_cluster_centers(self, *args, **kwargs):
      return self._initial_cluster_centers
```
And the lambda function is replaced with the function above. Then, I get the following reccursion error.
```python
>>> from sktime.clustering.k_means import TimeSeriesKMeans
>>> import numpy array as np
>>> model = TimerSeriesKMeans(init_algorithm=np.array([0, 0]))
>>> model
  File "/home/alex/documents/sktime/.venv/lib/python3.10/site-packages/skbase/base/_pretty_printing/_pprint.py", line 389, in _safe_repr
    params = _changed_params(obj)
  File "/home/alex/documents/sktime/.venv/lib/python3.10/site-packages/skbase/base/_pretty_printing/_pprint.py", line 34, in _changed_params
    params = base_object.get_params(deep=False)
  File "/home/alex/documents/sktime/.venv/lib/python3.10/site-packages/skbase/base/_base.py", line 359, in get_params
    params = {key: getattr(self, key) for key in self.get_param_names()}
  File "/home/alex/documents/sktime/.venv/lib/python3.10/site-packages/skbase/base/_base.py", line 311, in get_param_names
    parameters = cls._get_init_signature()
  File "/home/alex/documents/sktime/.venv/lib/python3.10/site-packages/skbase/base/_base.py", line 283, in _get_init_signature
    init_signature = inspect.signature(init)
  File "/usr/lib/python3.10/inspect.py", line 3247, in signature
    return Signature.from_callable(obj, follow_wrapped=follow_wrapped,
  File "/usr/lib/python3.10/inspect.py", line 2995, in from_callable
    return _signature_from_callable(obj, sigcls=cls,
  File "/usr/lib/python3.10/inspect.py", line 2383, in _signature_from_callable
    _get_signature_of = functools.partial(_signature_from_callable,
RecursionError: maximum recursion depth exceeded while calling a Python object
```
I am guessing this is something to do with the `BaseClusterer` class but I am not sure exactly.

fkiraly: how about you skip the idea of passing a function entirely - if `init_algorithm` is `numpy.array`, in `_check_params`, just get the array, instead of calling a function.

fkiraly:
would you like me to give it a try?

Alex-JG3:
> would you like me to give it a try? (if you're not sure where to put things?)

I'll give it a go.

sktime/clustering/partitioning/_lloyds.py
Alex-JG3: Moved the logic to `_check_params` in commit 109bbc3.

Alex-JG3: Switched to avoiding passing around function in commit 109bbc3.

sktime/clustering/partitioning/_lloyds.py
```diff
@@ -264,11 +265,19 @@ def _check_params(self, X: np.ndarray) -> None:
         self._random_state = check_random_state(self.random_state)
 
         if isinstance(self.init_algorithm, str):
-            self._init_algorithm = self._init_algorithms.get(self.init_algorithm)
-        else:
+            try:
```
fkiraly, line 268: I recommed not using try/except if the condition can be explicitly checked.
I would recommed `isinstance(init_algorithm, np.ndarray)`

fkiraly: thanks!

sktime/clustering/partitioning/_lloyds.py
```diff
@@ -264,11 +265,18 @@ def _check_params(self, X: np.ndarray) -> None:
...
+                    f"strings: {list(self._init_algorithms.keys())}"
+                )
+            self._init_algorithm = self._init_algorithms[self.init_algorithm]
+        elif isinstance(self.init_algorithm, np.ndarray):
             self._init_algorithm = self.init_algorithm
-
-        if not isinstance(self._init_algorithm, Callable):
+        elif isinstance(self._init_algorithm, Callable):
+            self._init_algorithm = self.init_algorithm
+        else:
```
fkiraly, line 279: I think the "else" should now say, it should be str, callable, or 3D np.ndarray?

sktime/clustering/partitioning/_lloyds.py
```diff
@@ -264,11 +265,18 @@ def _check_params(self, X: np.ndarray) -> None:
...
-            self._init_algorithm = self._init_algorithms.get(self.init_algorithm)
-        else:
+            if self.init_algorithm not in self._init_algorithms.keys():
+                raise KeyError(
+                    f"The string provided for init_algorim: {self.init_algorithm} is "
+                    f"invalid. The following are a list of valid init algorithms "
+                    f"strings: {list(self._init_algorithms.keys())}"
+                )
+            self._init_algorithm = self._init_algorithms[self.init_algorithm]
+        elif isinstance(self.init_algorithm, np.ndarray):
```
fkiraly, line 275: do you also want to check that it's 3D? and/or that it has the same number of variables (axis 1) as the data `X` seen here?

fkiraly (changes requested):
Looks like it should work!

Left some non-blocking comments above.

Two blocking comments:

* the option should be described in the docstring of `TimeSeriesKMeans` and `TimeSeriesKMedoids`.
* the new option needs to be covered by tests. Details below.

Details on testing:

It looks like, current tests don't cover the init options, that should also be fixed.

The place to fix this would be the `get_test_params` method of `TimeSeriesKMeans`.

There we should add one or two other parameter sets, at least one having the fixed init option. While adding parameter sets, we should also make sure we have a variety of combinations for the arguments covered.

Alex-JG3:
>

fkiraly:
did you mean to request a review instead of closing, @Alex-JG3?

sktime/clustering/partitioning/_lloyds.py
Alex-JG3: I have add a check for `init_algorithm` begin 3D but I am not sure the best way to go about checking the same number of dimensions for X and `init_algorithm` without loads of `if` statements.

sktime/clustering/partitioning/_lloyds.py
```diff
@@ -264,15 +266,32 @@ def _check_params(self, X: np.ndarray) -> None:
...
+            if self.init_algorithm.ndim != 3:
+                raise ValueError(
+                    "The array provided to init_algorithm is invalid. It must be 3D."
+                )
+            if self.init_algorithm.shape[0] != self.n_clusters:
```
Alex-JG3, line 281: Add a check to make sure the `n_clusters` is the same as the number of centers in `init_algorithm`. If they are not then after fitting the shape of `self.cluster_centers_` may be different to the shape of the array provided to `init_algorithm` which is maybe confusing?

Alex-JG3:
> did you mean to request a review instead of closing, @Alex-JG3?

Yes I meant to request a review, closing the PR was a mistake :)

sktime/clustering/partitioning/_lloyds.py
fkiraly: that's not really blocking, fine with me.

I was simply thinking about `X.shape[1] == init_algorithm.shape[1]`, but perhaps that's "over-checking".

sktime/clustering/partitioning/_lloyds.py
fkiraly: good idea

sktime/clustering/partitioning/_lloyds.py
fkiraly: I would still list the possible string keys though, these are helpful.

fkiraly (changes requested):
Looks great!

One blocking request:
The docstring needs to be clearer - a user might not be able to guess what the 3D array means (and, e.g., that it needs to have as many centers as `n_clusters`)

It should contain two blocks like:

"If string, any of the following ..."
"If 3D np.ndarray, initializes the cluster centers with ..."

Left some comments in addition, but these are non-blocking.

fkiraly (approved):
Looks like it's done?

You didn't request a review, so I'm not sure whether you are still working on it?

sktime/clustering/partitioning/_lloyds.py
fkiraly: addresed

sktime/clustering/partitioning/_lloyds.py
fkiraly: addressed

Alex-JG3:
@fkiraly, yes I think this is done. I have requested a rev view now.

fkiraly (approved):
Great, thanks!

All requests were addressed!

Alex-JG3:
@fkiraly, thanks for your help and for being so responsive. This has been a fun experience!

fkiraly:
> This has been a fun experience!

Likewise, thanks for contributing!

Even if the way we arrived here was a bit labyrinthine due to the various options, but perhaps that was what made it fun :-)

PR #5004, opened 2023-08-02
[MNT] separate CI element for testing external data downloaders

fkiraly:
This PR moves out tests of downloaders from external data sources into one single CI element:

* whether a test is testing for a data download is controlled by a new `pytest` mark `datadownload`, which informs GHA which tests to run exclusively in that new CI element.
* The CI element runs once, on ubuntu and python 3.9.
* all tests with the mark have been moved to a single test module in `datasets`, although that could change in the future as long as the mark `datadownload` is used.

These changes constitute an improvement, as they remove any dependencies on the status of external URLs of the default test suite. It also makes it easier for contributors to see whether failures are die to temporary failures of external download sources as opposed to a genuine failure in the contributed code. As the number of download dependencies increases, so does the frequency of failures of this kind.

In addition, this PR adds a test for `load_solar` - which depends on an external data source that was not tested - and fixes the docstrings for `load_solar`.


benHeid (approved):

fkiraly:
@benHeid, before you approve, have you seen the test durations?

This somehow overrides the "test only changes" or the matrix setup, so we end up at six hours.

benHeid:
Oh sorry.. I just examined if the different tests are working and not took a look on the times. This is strange...

fkiraly:
@yarnabrina, any ideas why the `-m` parameter seems to overide the flags (e.g., incremetal testing)?

yarnabrina:
Sorry, I have no idea. I'd also have expected no other tests to run except framework ones and data loader ones.

fkiraly:
It must be something about parameter override logic between the `make`-file and the `setup.cfg`, imo

yarnabrina:
@fkiraly, @benHeid I've a silly question. Can you please explain how are arguments being parsed here?

> python -m pytest -m "not datadownload"

The flag `-m` is being used twice. How is it correctly being inferred that that first one is indicating run `pytest` as module to the `python` command's argument parsing and the second one is indicating marker filter to that module's argument parsing capability, and not incorrectly as both are being passed to `python` and the second one overriding first one?

(This question may not have anything to do with the PR objective and/or incremental testing issue, but wanted to clarify a personal doubt.)

fkiraly:
superseded by https://github.com/sktime/sktime/pull/5437

PR #5012, opened 2023-08-03
[BUG] fix BaseWindowSplitter get_n_split method for hierarchical data

### LINKED ISSUE #4971: [BUG] get_n_splits of window splitter does not reflect the number of splits for hierarchical data
**Describe the bug**

`get_n_splits` method returns the wrong number of splits when hierarchical data is passed into window based splitter. 

**To Reproduce**

```python
from sktime.forecasting.model_selection import ExpandingWindowSplitter, SlidingWindowSplitter

y_hierarchical = _make_hierarchical(hierarchy_levels=(2, 3), min_timepoints=12, max_timepoints=12, random_state=0)
fh = [1]
step_length = 1
window_length = 10

# each unique series (lowest hierachical level) should have two splits
cv = ExpandingWindowSplitter(initial_window=window_length, fh=fh, step_length=step_length)
sliding = SlidingWindowSplitter(fh=fh, window_length=window_length, step_length=step_length, start_with_window=True)

n_splits_cv = cv.get_n_splits(y_hierarchical)
n_splits_sliding = sliding.get_n_splits(y_hierarchical)

print(f"Expanding Window splits : {n_splits_cv}")
print(f"Sliding Window Number of splits : {n_splits_sliding}")
```

Output
```
Expanding Window Number of splits : 62
Sliding Window Number of splits : 62
```
**Expected behavior**

 
Shouldnt the number of split is based on either of these two?:
1. Return number of splits based on datatype train/test splits which is two in this case
```python
print(f"Expanding Window splits: {len(list(cv.split(y_hierarchical)))}")
print(f"Sliding Window splits: {len(list(sliding.split(y_hierarchical)))}")
```
2. Return number of splits based on splits of the lowest hierachical level series. In this case it will be 6 * 2 = 12

**Versions**
<details>

System:
python: 3.9.17 (main, Jul  5 2023, 20:47:11) [MSC v.1916 64 bit (AMD64)]
machine: Windows-10-10.0.19045-SP0

Python dependencies:
pip: 23.2.1
sktime: 0.21.0
sklearn: 1.3.0
skbase: 0.4.6
numpy: 1.24.3
scipy: 1.11.1
pandas: 2.0.3
matplotlib: 3.7.2
joblib: 1.3.1
statsmodels: 0.14.0
numba: 0.57.1
pmdarima: 2.0.3
tsfresh: 0.20.1
tensorflow: 2.13.0
tensorflow_probability: None

</details>

hazrulakmal:
Fixes #4971

The logic of the `BaseWindowSplitter.get_n_split` method does not work for hierarchical cases. This is mainly because it counts cutoffs on y, assuming that it is either a **single** or a **multivariate** time series. Therefore, when hierarchical data is passed, the logic breaks because it counts cutoff as if the "pd_multiindex_hier" is a single series when it is not.

Solution
Using the existence implementation, for when pd_multiindex_hier datatype is passed, the cutoff point is counted based on the first instance of the lowest level time series hierarchy. 

Other comments
I also found out that other splitters, like `CutoffSplitter`, may return the wrong number of splits because they use the same logic as `BaseWindowSplitter` to count the splits. Also, It seems that the original implementation for hierarchical splitter #2599 may not consider how `get_n_splits` method should return in the case of hierarchical.


sktime/forecasting/model_selection/_split.py
```diff
@@ -996,12 +996,26 @@ def get_n_splits(self, y: Optional[ACCEPTED_Y_TYPES] = None) -> int:
...
+
         if y is None:
             raise ValueError(
                 f"{self.__class__.__name__} requires `y` to compute the "
                 f"number of splits."
             )
-        return len(self.get_cutoffs(y))
+
+        # n_splits based on the first instance of the lowest level series cutoffs
+        if check_is_scitype(y, scitype="Hierarchical"):
```
fkiraly, line 1008: this should also cover `"Panel"`, which is separate (e.g., 2-level `MultiIndex`)

sktime/forecasting/model_selection/tests/test_split.py
```diff
@@ -711,3 +711,20 @@ def test_same_loc_splitter_hierarchical():
...
+
+
+@pytest.mark.parametrize("CV", [SlidingWindowSplitter, ExpandingWindowSplitter])
+@pytest.mark.parametrize("fh", [*TEST_FHS, *TEST_FHS_TIMEDELTA])
+@pytest.mark.parametrize("window_length", TEST_WINDOW_LENGTHS)
+@pytest.mark.parametrize("step_length", TEST_STEP_LENGTHS)
+def test_windowbase_splitter_get_n_split_hierarchical(
+    CV, fh, window_length, step_length
+):
+    """Test that WindowBaseSplitter.get_n_splits works for hierarchical data."""
```
fkiraly, line 723: should we also look at unequal index sets to be sure?

fkiraly (changes requested):
Looks like a pertinent fix!

Main comments:

* the fix and tests should also cover the `Panel` scitype
* should this perhaps be in the `BaseSplitter`, for non-hierarchical splits?

hazrulakmal:
> should this perhaps be in the BaseSplitter, for non-hierarchical splits?

I think we can as long as the splitter child implements the .get_cuttoff method.

sktime/forecasting/model_selection/tests/test_split.py
hazrulakmal: added here [d0dc47f](https://github.com/sktime/sktime/pull/5012/commits/d0dc47f329cce2b98f7c0aef62a60aeb1ce573f5). created another test to strike a balance between coverage and speed.

sktime/forecasting/model_selection/_split.py
hazrulakmal: same as below. Kindle review them again to check if it's ok

hazrulakmal:
@fkiraly tagging you again just in case you missed this

fkiraly:
>  I suggest we don't move it to `BaseSplitter` and maintain the status quo as an abstract method cus the implementation may not be general enough for arbitrary splitters.

Yes, I think you are right.

> Perhaps, we can perhaps do lazy compute at `BaseSplitter` something like

Yes, makes sense - and specific splitters should override this.

In the final test framework that @BensHamza is working on, I think we should have a test that compares the `get_n_splits` return to the various method returns, e.g., `split`.

We should also make notes of all the splitters we're changing, for @BensHamza's refactpr PR.
If the files register as "moved", all should be fine, but it might get a mess if the commits registered it as a change rather than a move.

hazrulakmal:
> I think we should have a test that compares the get_n_splits return to the various method returns, e.g., split.

Agreed. Currently, the state checks only look at the n_split of the `split` method for a single series case. This can be expanded to check for other method returns, such as `split_series` and `split_loc`

> Yes, makes sense - and specific splitters should override this.

ok, we can do this on a separate PR as it is a separate issue from what this PR aims to do.

This is the final state of the PR.  I have made changes to account for the requested changes. all good to go if there; 's no more blocking.

sktime/forecasting/model_selection/_split.py
```diff
@@ -1051,12 +1051,33 @@ def get_n_splits(self, y: Optional[ACCEPTED_Y_TYPES] = None) -> int:
         n_splits : int
             The number of splits.
         """
+        from sktime.datatypes import check_is_scitype, convert_to
+
         if y is None:
```
fkiraly, line 1056: hm, this is an odd one. `y` has a default, ut that default always throws this error.

Also, only some splitters require `y`. Perhaps we should deal with this with a tag that tells us whether?

Out of scope for this PR though.

sktime/forecasting/model_selection/_split.py
```diff
@@ -1051,12 +1051,33 @@ def get_n_splits(self, y: Optional[ACCEPTED_Y_TYPES] = None) -> int:
...
             raise ValueError(
                 f"{self.__class__.__name__} requires `y` to compute the "
                 f"number of splits."
             )
-        return len(self.get_cutoffs(y))
+
+        scitype = ["Hierarchical", "Panel"]
+        is_non_single, _, metadata = check_is_scitype(y, scitype, True)
```
fkiraly, line 1063: to esure we compute only the metadata we need, the `return_metadata` arg should e the empty list

fkiraly (changes requested):
Thanks!

The only locking request is the one regarding minimizing the metadata computations.

sktime/forecasting/model_selection/_split.py
hazrulakmal: empty list? but the return_metadata arg accepts a bool?

fkiraly: you can also pass a list of str, which is as in the docstring. If you pass the empty list, only the types are returned.

fkiraly (dismissed):
I fixed it now in the interest of time, for the release and as @BensHamza is also dependent on this.

sktime/forecasting/model_selection/_split.py
hazrulakmal: resolved [3623c7f](https://github.com/sktime/sktime/pull/5012/commits/3623c7ffb8ab791aace27ce32655c9c9375853a8)

hazrulakmal:
> I fixed it now in the interest of time

Thank you @fkiraly. Really appreciate the help.

PR #5016, opened 2023-08-03
[MNT] updated sphinx intersphinx links for other libraries

### LINKED ISSUE #4802: [DOC] update configurations for `sphinx.ext.intersphinx` extension to use up to date links for latest stable version
As of now, `make docs` leads to the following warnings during documentation building process locally. 

```shell
intersphinx inventory has moved: https://docs.scipy.org/doc/numpy/objects.inv -> https://numpy.org/doc/stable/objects.inv
intersphinx inventory has moved: https://matplotlib.org/objects.inv -> https://matplotlib.org/stable/objects.inv
intersphinx inventory has moved: https://docs.scipy.org/doc/scipy/reference/objects.inv -> https://docs.scipy.org/doc/scipy/objects.inv
```

The links in `docs/source/conf.py` in `intersphinx_mapping` should be updated appropriately.

If there are other links to update as well, for example changing from `latest` to `stable` if available, adding a new link, always using python 3 instead of unnecessarily using `sys.version_info`, etc. can be useful.

yarnabrina:
Fixes #4802

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -425,9 +425,9 @@ Generated using nbsphinx_. The Jupyter notebook can be found here_.
 # Example configuration for intersphinx: refer to the Python standard library.
 intersphinx_mapping = {
     "python": (f"https://docs.python.org/{sys.version_info.major}", None),
-    "numpy": ("https://docs.scipy.org/doc/numpy/", None),
-    "scipy": ("https://docs.scipy.org/doc/scipy/reference", None),
-    "matplotlib": ("https://matplotlib.org/", None),
+    "numpy": ("https://numpy.org/doc/stable/", None),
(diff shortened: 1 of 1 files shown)


yarnabrina:
Note to the reviewers: the change affects only the links in documentation, and has (and can have) no effect whatsoever on the main functionalities. Skipped GitHub CI testing for that reason, but ReadTheDocs CI was successful.

fkiraly (approved):
makes sense - although I'm not sure how I would check this programmatically

yarnabrina:
> how I would check this programmatically

If you run `make docs` locally, after-PR should have less warnings than before-PR, which will have same warnings as in the linked issue. And since the warning showed the link has moved, I thought (don't really know for sure) may be they will simply fail to work sometime in future.

PR #5017, opened 2023-08-03
[ENH] consolidating splitters as their own module with systematic tests and extension

### LINKED ISSUE #4889: [ENH] consolidating splitters as their own module with systematic tests and extension template

BensHamza:
#### Reference Issues/PRs

Fixes #4889 

- [x] Create a new module, `sktime.split`, which will contain all the splitters
- [x] Create submodule `base` which only has the `BaseSplitter` and `BaseWindowSplitter`
- [x] Move the splitters into individual files inside `split`. `__init__` should export them all. The old location should be subject to deprecation - imports should still work but raise a warning.
- [x] Create a test module, `split.tests`. All current tests should move there.
- [ ] Add an extension template for splitters, in `extension_templates` - this will inform tests below as it will make concrete the key methods.
- [ ] Create a new test class, `tests::TestAllSplitters`. This should inherit from `TestAllObjects`, and be similar to `TestAllAligners`, `TestAllDistributions`, and similar. It should also be registered in `check_estimators` (see there for pattern)
- [ ] The generic tests should be migrated to `TestAllObjects`, as long as it is a test for all splitters.
- [ ] Think about new tests, e.g., "test that `get_n_splits` is always the same as length of `split`". All methods of splitters should be covered.
- [ ] It may make sense to add test scenarios, although this may equally be covered by simple fixtures.

No

Still no test

#### Any other comments?

This work is still on progress, but open to discussions, remarks and reviews.

##### For all contributions

##### For new estimators


BensHamza:
Thank you @hazrulakmal for your comment. Indeed I pushed the code without `pre-commit` configured. I make the changes right now.

sktime/forecasting/model_selection/_split.py
```diff
@@ -67,6 +68,12 @@
 SPLIT_GENERATOR_TYPE = Iterator[SPLIT_ARRAY_TYPE]
 PANDAS_MTYPES = ["pd.DataFrame", "pd.Series", "pd-multiindex", "pd_multiindex_hier"]
 
+warnings.warn(
```
fkiraly, line 71: if you want to avoid that this warning is raised when running crawling utility `all_estimator`, this should be added to the excepted module list there.

Also, replace "this module" by the full name, as users receiving the warning from an indirect import may get confused.

BensHamza: Thank you for the feedback!
I've added the `_split` and `test_split` modules to the `EXCEPTED_MODULES` list in the `all_estimator` crawling utility as suggested and I've updated the deprecation warning to use the full module name using the `"__name__"` attribute for clarity.
Could you please review the updated changes?

hazrulakmal:
Notes: some other PRs that may affect this PR.

1. #5012 
2. #5009

fkiraly:
Had a quick look - the failing test `test_inheritance` seems failing as the check fails that the splitters inherit from `BaseSplitter`.

I suppose this is due to a duplication of classes - the registry should contain the correct class, and the old class should be removed.

BensHamza:
> Had a quick look - the failing test `test_inheritance` seems failing as the check fails that the splitters inherit from `BaseSplitter`.
> 
> I suppose this is due to a duplication of classes - the registry should contain the correct class, and the old class should be removed.

Thank you @fkiraly for identifying the issue. I made the changes in `registry`.

fkiraly:
hm, it's still not identifying the inheritance tree correctly - one of the two sides on the comparison must be mismatched.

I would advise, try to run the test manually and see how the mismatch comes to pass. I can't grok from naive inspection what's the root cause. If you can't find it, I'm happy to check in a screenshare in the Fri meet-up or mentoring meeting, if you can make it this week.

BensHamza:
I will try tomorrow before the meeting but I think your help is really appreciated here @fkiraly, thank you ! 😁

fkiraly:
ok, let's look at it later in a screenshare.
I would also advise, let's cut the plan up in multiple PR, and try to merge the "move-refactor" soon, and look at the test class based on that

fkiraly:
from discussion on Sep 8:

* splitters are currently duplicated, in the old and the new module
* that is the deeper reason behind the test failures
* instead, the old module should be removed, with imports pointing to the new module so users can still import from the old location
* a warning should be raised when a user imports from the old location, pointing them to the new import location
* deprecation horizon should be one minor cycle ahead, right now that would be 0.24.0
* a todo comment should be left in the old module that starts with `# todo 0.24.0`, with explicit instructions what the release manager should do at 0.24.0 (if we merge after 0.23.0, this needs to change accordingly)

fkiraly (changes requested):
Great, I looked through this - almost done!

One thing I´d still like to ask, we need to make sure that we do not break code of downstream users, who may still import splitters from `sktime.forecasting.model_selection`.

For that, we need to follow the typical deprecation cycle:

* until 0.25.0, allow imports from that location, but still raise a warning. This can be achieved by the old location importing from the new location.
* at 0.25.0, remove the old location.
* leave `todo 0.25.0` messages in the code that tells the release manager for 0.25.0 which files or content to remove.

Let me know if you need further details!

BensHamza:
> Great, I looked through this - monumental work! We're almost done!

Thank you for your enormous help and dedication ! I learned a lot trough the process.

> * until 0.25.0, allow imports from that location, but still raise a warning.

Importing splitters from  `sktime.forecasting.model_selection` or `sktime.forecasting.model_selection._split` works. the deprecation warning is also raised for both.

> * leave `todo 0.25.0` messages in the code that tells the release manager for 0.25.0 which files or content to remove.

I left the todo comment for the `sktime.forecasting.model_selection._split` module, the only module that is affected by . The old location of splitters' test is deleted (transferred with adaptation to `sktime.split.tests`) in this PR.

fkiraly:
> the deprecation warning is also raised for both.

Great! The warning should say though that the import will stop working in 0.25.0.

> I left the todo comment 

I see! For search purposes, the comment should start literally with `todo 0.25.0`.

BensHamza:
> The warning should say though that the import will stop working in 0.25.0.

I added it in my last commit, here is the actual message : 
```
Please import sktime splitters from `sktime.split`. Importing splitters from `sktime.forecasting.model_selection` is deprecated and will be removed in release 0.25.0
```
> the comment should start literally with todo 0.25.0.

Sorry I didn't mention it correctly. I added the comment `todo 0.25.0 : Please ensure this file is deleted.` in the specified deprecated file with a quick description above it.

I can change the message or else if needed. Thank you !

fkiraly (approved):
Excellent, thanks! Now it is very clear!

PR #5023, opened 2023-08-04
DIAGNOSTIC [MNT] testing #5019 with dummy changes

fkiraly:
This is a diagnostic PR to test the changes introduced in #5019.

It makes changes in the modules of two estimators, `NaiveForecaster` and `DummyClassifier`.

If the changes in #5019 work, it should pick up these changes and run suite tests for, and only for, these two estimators.


fkiraly:
no longer needed for testing

PR #5028, opened 2023-08-04
[MNT] speed up `deep_equals` by avoiding string-coercion of pandas objects

fkiraly:
This PR speeds up the ubiquitous testing utility `deep_equals` by avoiding string-coercion of pandas objects unless they are really necessary.

### DIFF
--- a/sktime/utils/_testing/deep_equals.py
+++ b/sktime/utils/_testing/deep_equals.py
@@ -57,7 +57,7 @@ def deep_equals(x, y, return_msg=False):
             != - call to generic != returns False
     """
 
-    def ret(is_equal, msg):
+    def ret(is_equal, msg=""):
         if return_msg:
             if is_equal:
                 msg = ""
@@ -65,7 +65,7 @@ def deep_equals(x, y, return_msg=False):
         else:
             return is_equal
 
-    if type(x) != type(y):
+    if type(x) is not type(y):
         return ret(False, f".type, x.type = {type(x)} != y.type = {type(y)}")
 
     # compute delayed objects (dask)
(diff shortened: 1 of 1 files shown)


sktime/utils/_testing/deep_equals.py
```diff
@@ -93,7 +93,12 @@ def ret(is_equal, msg):
                 msg = ""
             return ret(index_equal and values_equal, msg)
         else:
-            return ret(x.equals(y), f".series_equals, x = {x} != y = {y}")
+            # this construction is needed so costly str(pd.Series) is only
```
benHeid, line 96: Alternatively, we could try to make `ret` lazy evaluating capable. The idea would be something as: 

```python
  def ret(is_equal, msg="", string_arguments=None):
      string_arguments = [] if len(string_arguments) == 0 else string_arguments
      if return_msg:
          if is_equal:
              msg = ""
          elif len(string_arguments) > 0:
              msg = msg.format(*string_arguments)
          return is_equal, msg
      else:
          return is_equal
```

this would enable to call `ret(x.equals(y), ".index_equals, x = {} != y = {}", [x, y])`. Consequently, we also would not require this construction.

benHeid (changes requested):
Thank you for the PR and speed up the `deep_equals` method.  I proposed a possible alternative that probably would avoid the introduced construction. Happy to discuss it, if it really works as intended :)

sktime/utils/_testing/deep_equals.py
fkiraly: hm, I see! Great idea!

My main worry is that it doesn't actually work precisely as described, there rare a lot of case distinctions, e.g., `string_arguments = None`, and calling `len(None)`, etc.

I don't mind if you try to make this change, we need to check that the negative case is also tested somewhere.

benHeid: I can try to change it. Should I create a new PR for this? Or work on this?

fkiraly: either is fine with me, as you prefer!

benHeid: I added a new PR based on this branch so that we could merge this PR, if we prefer this solution.

New PR is here: #5044

fkiraly:
closed in favour of https://github.com/sktime/sktime/pull/5044

PR #5030, opened 2023-08-05
Test only modified changes in CI

yarnabrina:
Related to #5019

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -169,6 +169,10 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      - run: git remote set-branches origin 'main'
+
+      - run: git fetch --depth 1
+
       - uses: conda-incubator/setup-miniconda@v2
         with:
           auto-update-conda: true
@@ -215,6 +219,10 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      - run: git remote set-branches origin 'main'
+
+      - run: git fetch --depth 1
+
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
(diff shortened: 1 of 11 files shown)


fkiraly:
I assume this is for experienting?

Once it works, this makes most sense as a PR to #5019, no?
(for now not, as it won't trigger CI)

yarnabrina:
Yes, this PR is trying to replace the depth 0 thing (full commit history for all tags and for all branches) by checking out only the latest commit on `main`, along with that on the PR branch.

I made one extra commit by adding a new test case of `{}`. Based on the `Show available branches` step, this seems to be successful.

I suppose this can be merged into your branch in #5019, and then that into `main` later, but I don't know how to change the target branch after creation of PR.

fkiraly:
> but I don't know how to change the target branch after creation of PR.

image
:-)

yarnabrina:
@fkiraly I was trying the edit next to description 🤦🏽 

Updated base branch. Ready for review.

fkiraly (changes requested):
Thanks!

I was wondering how you get only `main`.

Now, I'm confused though, there are many more changes showing up in the diff than I thought there would be, some unrelated from recent merges to main. Do you have any idea what is going on there? I've tried updating both branches from `main`, but it still looks like this.

yarnabrina:
> I was wondering how you get only main.

This part does it, and I got it from the stack overflow link I shared yesterday in #5019.

```shell
git remote set-branches origin 'main'  # tells shallow clone that there is a main branch in remote
git fetch --depth 1  # gets all known branch's latest state, which is only for `main` because of shallow
```

(The comments are my understanding from git documentation, not sure if it's exactly the case or not.)

> Do you have any idea what is going on there?

I've no clue. I created a new branch from yours, made changes in `test.yaml` and `statsforecast.py`, and that's it. I did some debugging, but didn't go beyond these files.

Complete random guess: can it be the case that the diff shows against `main` even though base is `test-only-changes`? I see green Squash and merge button though.

fkiraly:
hm, let's see what happens if we merge. We can always revert.

PR #5031, opened 2023-08-05
[MNT] upgrade CI runners to latest stable images

### LINKED ISSUE #4727: [MNT] upgrade CI machines to `latest` versions
Currently, `sktime` CI tests mainly runs on following machines:

* `macos-11`
* `ubuntu-20.04`
* `windows-2019`

The available machines are documented [here](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources). We can consider switching to the `latest` versions, which will check against the operating system versions most users are on.

* `macos-latest`
* `ubuntu-latest`
* `windows-latest`

yarnabrina:
Fixes #4727 

Ref. https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources


fkiraly:
the failures all seem sporadic/probabilistic - I think the `ExponentialSmoothing` is reported, but the `ConformalIntervals` not yet.

FYI @bethrice44 in case you can make sense of the `ConformalIntervals` sporadic failure

fkiraly (approved):

fkiraly (changes requested):
Quick question - it seems like the names of the CI elements have all been changed to `Install and Test / CI element name` - why is that the case?

yarnabrina:
image

This is from #5019. It is the exact same pattern there.

What I think is being shown here is that it's running the CI jobs according to this branch, and since `main` do not have these jobs, CI is failing to know that equivalent of these jobs have already been run so it's waiting for these to be executed. Kind of similar what happened when I committed by adding `skip ci`.

(It's sort of an educated guess, can't prove it.)

fkiraly (approved):
makes sense - so we expect it to go away, right?

yarnabrina:
Are you okay to merge this, now that feature freeze is over?

fkiraly:
yes, thanks for the reminder.

I suppose I should merge since I approved.

PR #5033, opened 2023-08-05
[MNT] mark and skip slow non-suite tests

fkiraly:
This PR marks and skips individual, slow non-suite tests.

A non-suite test may be marked as slow if it has a runtime of 20sec or larger, and it is unclear how to speed it up.

The proposal inherent is to skip these tests, and track them via the "slow" mark, potentially speeding them up as a good first issue.

We could run the expensive tests in a CRON or at release.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -196,7 +196,7 @@ jobs:
           mkdir -p testdir/
           cp .coveragerc testdir/
           cp setup.cfg testdir/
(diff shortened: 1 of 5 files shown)


fkiraly:
... why do test times go up to 4-5 hours here?

fkiraly:
closed in favour of the more selective strategy https://github.com/sktime/sktime/pull/5084

PR #5034, opened 2023-08-05
[MNT] fix duplication of `pytest` `durations` parameter in CI

fkiraly:
Fixes the duplication of the `pytest` `durations` parameter where `setup.cfg` would override the `durations` passed later on.


fkiraly:
that's strange, it still shows only 10?

fkiraly:
There was a duplication, so removed the duplication which ensures we see 20 test times now.

PR #5035, opened 2023-08-05
[ENH] speed up `Lag` transformer

fkiraly:
This PR carries out various speedups to the `Lag` transformer, the general idea is to avoid checks or `combine_first` which is costly, and adds up, e.g., in the new reducer class which uses `Lag`.

* adds `remember_data` option which can be turned off if using `fit_transform` only, that turns off `update_data` and `combine_first` from remembering data
* replaces `combine_first` calls with `reindex` when dealing with the resulting index set

Depends on https://github.com/sktime/sktime/pull/5069 for tests.


benHeid (approved):
Thank you for the speed-ups. The PR looks fine to me.

fkiraly:
@benHeid, I just noticed a bug with the `"shift"` setting, so would appreciate another review - sorry.

Turns out this was not covered by tests sufficiently, I rectified this in #5069.

PR #5037, opened 2023-08-05
[ENH] interface to all `tslearn` clusterers

fkiraly:
This PR adds:

* interfaces to all `tslearn` clusterers 
* ensures all existing classes use the adapter from https://github.com/sktime/sktime/pull/4992

Depends on https://github.com/sktime/sktime/pull/4992

Note: the default of `tslearn`'s k-means clusterer has been changed from `"k-means++"` to `"random"`, as there seems to be a genuine bug in `tslearn` with `"k-means++"` - it causes an internal private function fail on expecting `sample_weights`. The exception raised is `TypeError: _kmeans_plusplus() missing 1 required positional argument: 'sample_weight'`


fkiraly:
quick question, @Alex-JG3 - have you, by chance, tried to use `tslearn`'s k-means clusterer? Any idea of what is going on with the `"k-means++"` init method?

Alex-JG3:
> quick question, @Alex-JG3 - have you, by chance, tried to use `tslearn`'s k-means clusterer?

@fkiraly, I haven't unfortunately.

PR #5039, opened 2023-08-05
[ENH] `tslearn` distances and kernels including adapter

fkiraly:
This PR adds interfaces to all `tslearn` distances and kernels to `sktime`.

Includes a general adapter to `tslearn` distanes and kernels.

Depends on https://github.com/sktime/sktime/pull/5038

Adds the following distances (all interfaced from `tslearn`):

* dynamic time warping distance
* soft-dtw
* canonical time warping
* generalized alignment kernel
* longest common subsequence


fkiraly:
While I have your kind attention, @rtavenar, @yanncabanes, would appreciate a comment on:

* are these all distances and kernels in `tslearn`?
* do you think credit is appropriately given? Docstrings refer to `tslearn` explicitly. In the search utility and API reference, these will appear with the `tslearn` branding.
* would you like to be owners of the adapter and/or the individual distances/kernels as per https://www.sktime.net/en/stable/get_involved/governance.html#algorithm-maintainers?

Issue #3641, opened 2022-10-24
[BUG] enforce_index_type doesn't raise an exception for invalid index types.


ltsaprounis:
**Describe the bug**
The enforce_index_type tag is currently not working as intended.
The issue was discovered in #3605. The FourierFeatures transformer should only allow a pandas PeriodIndex.
However if a DateTimeIndex is passed the transformer will not raise an exception (expected behaviour based on extension contract), nor it will convert the index to period index in before it reaches _fit (expected behaviour based on my understanding of the tag's name).

**To Reproduce**

same code as in #3605
```python
from sktime.datasets import load_airline
from sktime.transformations.series.fourier import FourierFeatures
from sktime.utils.plotting import plot_series

df = load_airline()
df.index = df.index.to_timestamp(freq="M")
from sktime.transformations.series.fourier import FourierFeatures
F_transform = FourierFeatures(sp_list = [12], fourier_terms_list=[2]) # should be yearly
sp1 = F_transform.fit_transform(df)

# Convert seasonality to nanoseconds to get correct result
F_transform = FourierFeatures(sp_list = [12*30*24*3600000000000], fourier_terms_list=[2])
sp2 = F_transform.fit_transform(df)
```

**Expected behavior**

An exception should be raised if the tag is not included in the list in the tag.

I would suggest we also add another tag to convert the index to the specified index, obviously, converting back to the original index type for any public "facing" results. This will **really** simplify development in a few cases e.g. LagTransformer.

fkiraly:
> I would suggest we also add another tag to convert the index to the specified index, obviously, converting back to the original index type for any public "facing" results.

Yes, I think this would be great. I was consiering lifting some of the functionality from the `_ProphetAdapter`, but then again there is a limited number of hours per day.

Maybe open a new issue for this? That's a new (useful) feature, above and beyond the current bug.

fkiraly:
I have discovered the reason: `_check_X_y`, where this should happen, does nothing of the sort.
In fact, nothing at all reads the tag anywhere.

This must be the result of a refactor accident, and lack of testing of said functionality.
(probably my fault)

There should be a few lines in `_check_X_y` that, at the end of conversion:
* check whether the converted-to format is `pandas`
* check whether the index type is in the pass list given by the tag, if not raise an error

fkiraly:
[...]

PR #5043, opened 2023-08-06
[ENH] `darts` adapter

### LINKED ISSUE #1624: [ENH] Adapter for DARTS
Currently models from DARTS are not supported in sktime.

Since the DARTS models use a unified fit, predict format, it may be beneficial to have a DARTS adapter built into `sktime`. The catch seems to be that DARTS works on a proprietary format of data (`TimeSeries`), so the adapter will need to handle the conversion from `TimeSeries` to pd.Series and vice versa.

```python
from darts import TimeSeries
```

I tried creating the adapter outside of sktime. It trains OK, but the prediction does not work. Not sure what the issue is. Can someone help out please?

**[Notebook with DARTS adapter](https://gist.github.com/ngupta23/d876b55fdfc47e5573b209d3aa5be222)**

yarnabrina:
Closes #1624

### DIFF
--- a/sktime/forecasting/base/adapters/_darts.py
+++ b/sktime/forecasting/base/adapters/_darts.py
@@ -48,6 +48,8 @@ class _DartsAdapter(BaseForecaster):
         self.past_covariates = [] if past_covariates is None else past_covariates
         self.num_samples = num_samples
 
+        self._forecaster = None
+
     @staticmethod
     def convert_dataframe_to_timeseries(dataset: pandas.DataFrame):
         """Convert dataset for compatibility with ``darts``.
@@ -102,9 +104,8 @@ class _DartsAdapter(BaseForecaster):
 
         return future_known_dataset, future_unknown_dataset
 
-    @property
     @abc.abstractmethod
-    def _forecaster(self: "_DartsAdapter"):
+    def _create_forecaster(self: "_DartsAdapter"):
         """Create Darts model."""
 
     def _fit(
@@ -140,6 +141,8 @@ class _DartsAdapter(BaseForecaster):
 
         endogenous_actuals = self.convert_dataframe_to_timeseries(y)
         unknown_exogenous, known_exogenous = self.convert_exogenous_dataset(X)
+
+        self._forecaster = self._create_forecaster()
 
         self._forecaster.fit(
             endogenous_actuals,
--- a/sktime/forecasting/base/adapters/_darts.py
+++ b/sktime/forecasting/base/adapters/_darts.py
@@ -1,0 +1,241 @@
+# copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+"""Implements adapter for Darts models."""
+import abc
+from typing import List, Optional
+
+import pandas
+
+from sktime.forecasting.base import BaseForecaster, ForecastingHorizon
+
+__author__ = ["yarnabrina"]
+
+
+class _DartsAdapter(BaseForecaster):
+    """Base adapter class for Darts models.
+
+    Parameters
+    ----------
+    past_covariates : Optional[List[str]], optional
+        column names in ``X`` which are known only for historical data, by default None
+    num_samples : Optional[int], optional
+        Number of times a prediction is sampled from a probabilistic model, by default
+        1000
+
+    Notes
+    -----
+    If unspecified, all columns will be assumed to be known during prediction duration.
+    """
+
+    _tags = {
+        "y_inner_mtype": "pd.DataFrame",
+        "X_inner_mtype": "pd.DataFrame",
+        "requires-fh-in-fit": False,
+        "enforce_index_type": pandas.DatetimeIndex,
+        "handles-missing-data": False,
+        "capability:insample": False,
+        "python_version": ">=3.8",
+        "python_dependencies": ["u8darts"],
+        "python_dependecies_alias": {"u8darts": "darts"},
+    }
+
+    def __init__(
+        self: "_DartsAdapter",
+        past_covariates: Optional[List[str]] = None,
+        num_samples: Optional[int] = 1000,
+    ) -> None:
+        super().__init__()
+
+        self.past_covariates = [] if past_covariates is None else past_covariates
+        self.num_samples = num_samples
+
+    @staticmethod
+    def convert_dataframe_to_timeseries(dataset: pandas.DataFrame):
+        """Convert dataset for compatibility with ``darts``.
+
+        Parameters
+        ----------
+        dataset : pandas.DataFrame
+            source dataset to convert from
+
+        Returns
+        -------
+        darts.TimeSeries
+            converted target dataset
+        """
+        import darts
+
+        return darts.TimeSeries.from_dataframe(dataset)
+
+    def convert_exogenous_dataset(
+        self: "_DartsAdapter", dataset: Optional[pandas.DataFrame]
+    ):
+        """Make exogenous features to ``darts`` compatible, if available.
+
+        Parameters
+        ----------
+        dataset : Optional[pandas.DataFrame]
+            available data on exogenous features
+
+        Returns
+        -------
+        Tuple[darts.TimeSeries, darts.TimeSeries]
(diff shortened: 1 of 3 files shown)


sktime/forecasting/base/adapters/_darts.py
```diff
@@ -0,0 +1,241 @@
...
+    _tags = {
+        "y_inner_mtype": "pd.DataFrame",
+        "X_inner_mtype": "pd.DataFrame",
+        "requires-fh-in-fit": False,
+        "enforce_index_type": pandas.DatetimeIndex,
+        "handles-missing-data": False,
+        "capability:insample": False,
+        "python_version": ">=3.8",
+        "python_dependencies": ["u8darts"],
+        "python_dependecies_alias": {"u8darts": "darts"},
```
yarnabrina, line 38: @hazrulakmal, @fkiraly Can you please help why is this failing complaining `u8darts` is not installed, even though it is installed? Is the alias specification format incorrect?

(This is with respect to my local environment.)

sktime/forecasting/base/adapters/_darts.py
```diff
@@ -0,0 +1,241 @@
...
+    Notes
+    -----
+    If unspecified, all columns will be assumed to be known during prediction duration.
+    """
+
+    _tags = {
+        "y_inner_mtype": "pd.DataFrame",
+        "X_inner_mtype": "pd.DataFrame",
+        "requires-fh-in-fit": False,
+        "enforce_index_type": pandas.DatetimeIndex,
```
yarnabrina, line 33: @fkiraly Can you please tell if this will be ensured during tests, or tests can pass datasets with any index type leading to failures?

yarnabrina (commented):
This is a very very basic first draft, and I don't expect it to pass any tests. Started the draft PR to clarify few doubts, and to get some initial feedbacks/suggestions.

sktime/forecasting/base/adapters/_darts.py
fkiraly: this implies that you install with `pip install u8darts`, and import with `import darts` - is that correct?

sktime/forecasting/base/adapters/_darts.py
fkiraly: Not sure - I think the tag has no effect. There is an issue somewhere that states this, and asks to implement coercion.

sktime/forecasting/base/adapters/_darts.py
yarnabrina: Yes, exactly. And it's not working.

> ModuleNotFoundError: DartsXGBModel requires package 'u8darts' to be present in the python environment, but 'u8darts' was not found. 'u8darts' is a soft dependency and not included in the base sktime installation. Please run: `pip install u8darts` to install the u8darts package. To install all soft dependencies, run: `pip install sktime[all_extras]`

But I do have it.

```sh
❯ pip freeze | grep darts
u8darts==0.25.0
```

If I use `darts` as `python_dependencies`, then it starts working. But actual `darts` package in PyPI needs so much more than `u8darts` (lots of system dependencies that are not straightforward to install on all OS - as per [Darts' own documentation](https://github.com/unit8co/darts/blob/master/INSTALL.md)) that I don't want to enforce that.

sktime/forecasting/base/adapters/_darts.py
yarnabrina: Ok, found it https://github.com/sktime/sktime/issues/3641.

What will be better to do here? Fix this issue, or add an if-else to check index type and raise TypeError? Or, both?

sktime/forecasting/base/adapters/_darts.py
fkiraly: > Yes, exactly. And it's not working.

Well, that's strange.

Last person working on this and looking into the code of the `python_dependencies` handling was @hazrulakmal, I think - any ideas?

sktime/forecasting/base/adapters/_darts.py
fkiraly: yes, that was it

> What will be better to do here?

I would do it in a separate PR, but yes.

I also would strongly recommend to think carefully about consequences of starting to raise an exception - this may cause estimators that have previously "over-restricted" the index type to fail, so it may break working user code!

I've added more details with my now-preferred solution here: https://github.com/sktime/sktime/issues/3641#issuecomment-1667237858 (this is based on making `Prophet` work for any index type)

yarnabrina:
I interpreted this as a consequence of the soft dependency check failure, i.e. the checker (wrongly) detected `u8darts` is missing and skipped all tests on new estimator. Is it not the case? 🤔

fkiraly:
> I interpreted this as a consequence of the soft dependency check failure, i.e. the checker (wrongly) detected u8darts is missing and skipped all tests on new estimator. Is it not the case? 

Oh, I see - you are right!
Well, sorry for being overoptimistic (that's usually not me...)

sktime/forecasting/base/adapters/_darts.py
hazrulakmal: There's a typo, it's 'python_dependencies_alias' not 'python_dependecies_alias'. That can be a reason.

yarnabrina: Thank you, fixed it and now we should see plenty of failures. 

But I wonder why is there no error when I used a non-existing tag. I'm not super familar with tag management system, but do we not have any check to validate if correct tags are provided or if supported valid values are provided? Can I suggest to add that enhancement?

sktime/forecasting/base/adapters/_darts.py
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    def __init__(
+        self: "_DartsAdapter",
+        past_covariates: Optional[List[str]] = None,
+        num_samples: Optional[int] = 1000,
+    ) -> None:
+        super().__init__()
+
+        self.past_covariates = [] if past_covariates is None else past_covariates
```
fkiraly, line 48: this line would violate the sklearn contract, `self.paramname` should always have same value as the `__init__` arg `paramname`

yarnabrina: Also, I've found another issue that probabilistic capability is probably dependent on algorithm and passed configuration. Like `XGBModel` is capable of doing so, but not unless you pass `likelihood` parameter. Is it okay to set tags based on passed arguments values? Will it cause any problem if `predict_quantiles` is expected to work only in few cases and not always?

sktime/forecasting/base/adapters/_darts.py
fkiraly: >  I'm not super familar with tag management system, but do we not have any check to validate if correct tags are provided or if supported valid values are provided?

That check exists, but it applies only to concrete estimators - private classes such as those starting in underscore or `Base` are excluded from the check.

The test would capture it once you inherit in a concrete class.

sktime/forecasting/base/adapters/_darts.py
fkiraly: > Is it okay to set tags based on passed arguments values?

Yes, for that you need to do a `set_tags` after the `super().__init__` call in `__init__`.

fkiraly: > Can you please suggest how will you approach this?

Use an attribute with a different name.

That is, internally, you use `self._paramname`. For a default, you set `paramname=None`, and you do `if paramname = None: self_paramname = []` or similar.

>Also, I'll probably get time for this again in the weekend, so if you can send me any reference for sklearn contract designs, I can read it up as it's coming again and again as part of different threads.

That would be here:
https://scikit-learn.org/stable/developers/develop.html

It is noteworthy that we don't follow all those conventions.

The expectations for developers/extenders are all outlined in the extension templates afaik.

fkiraly:
superseded by https://github.com/sktime/sktime/pull/6454

yarnabrina:
@@fkiraly I'll delete my branch in fork, but I think this is not superceded by @fnhirwa 's PR. That one replaced #5997 instead.

PR #5044, opened 2023-08-06
[MNT] Speed-up `deep_equals` - lazy evaluation of costly error message string coercions

benHeid:
This PR speeds `deep_equals` by string-coercion of pandas objects unless necessary using the idea of lazy evaluation. This is an alternative to #5028.

### TODO
* [ ] Which alternative is preferrable?
* [ ] Is this alternative working in each case? 
* [ ] Is this alternative as fast as the PR #5028

### DIFF
--- a/sktime/utils/_testing/deep_equals.py
+++ b/sktime/utils/_testing/deep_equals.py
@@ -57,10 +57,13 @@ def deep_equals(x, y, return_msg=False):
             != - call to generic != returns False
     """
 
-    def ret(is_equal, msg=""):
+    def ret(is_equal, msg="", string_arguments: list = None):
+        string_arguments = [] if string_arguments is None else string_arguments
         if return_msg:
             if is_equal:
                 msg = ""
+            elif len(string_arguments) > 0:
+                msg = msg.format(*string_arguments)
             return is_equal, msg
         else:
             return is_equal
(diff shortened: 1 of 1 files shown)


fkiraly (commented):
Seems much more elegant!

Have you checked whether it works?

benHeid:
> Seems much more elegant!
> 
> Have you checked whether it works?

I used the following lines of code for testing:

```python
from sktime.utils._testing.deep_equals import deep_equals
import pandas as pd
import numpy as np
import time

df = pd.DataFrame(np.random.normal(0,1, 10000))
df1 = df.copy(deep=True)
times = []
for _ in range(100):
    start_time = time.time()
    deep_equals(df, df1, return_msg=True)
    times.append(time.time() - start_time)
print("Equal: ", sum(times)/ len(times))

df1 = pd.DataFrame(np.random.normal(0, 1, 10000))
times = []
for _ in range(100):
    start_time = time.time()
    deep_equals(df, df1, return_msg=True)
    times.append(time.time() - start_time)
print("Unequal: ", sum(times)/ len(times))

```

the results were: 
```
Equal:  0.00020035743713378907
Unequal:  0.005350635051727295
```

So I suppose that it is working. However, I am unsure if this is the best way of testing. If you have a better way of testing, I am happy to try it out.

fkiraly:
> However, I am unsure if this is the best way of testing.

Yes, I think it is!
I also meant just that it works as intended (without any runtime measurement) in the negative case.

fkiraly (approved):

PR #5049, opened 2023-08-07
[ENH] Add DropNA transformer

### LINKED ISSUE #4977: [BUG] Using IndexSubset together with PandasTransformAdaptor(method="dropna") in a pipeline fails
**Describe the bug**
Using `IndexSubset` and `PandasTransformAdaptor(method="dropna")` together in a pipeline fails in some orderings. `PandasTransformAdaptor` appears to be operating on the untransformed/not subset `X`. 

**To Reproduce**
The example below fails. It works with the `PandasTransformAdaptor` commented out.  

The example differences `y`, lags `X`, then sets `y` and `X` to the same index (effectively removing the missing values from `X`). Dropping `nan` values should not have any effect. 

```python
import numpy as np
from sktime.forecasting.compose import (
    ForecastingPipeline,
    TransformedTargetForecaster,
    RecursiveTabularRegressionForecaster,
)
from sktime.transformations.series.adapt import TabularToSeriesAdaptor, PandasTransformAdaptor
from sktime.transformations.series.boxcox import LogTransformer
from sktime.transformations.series.subset import IndexSubset
from sktime.transformations.series.lag import Lag

y, X = load_longley()
# X.loc["1955", "GNPDEFL"] = np.nan
# X.loc["1947", "POP"] = np.nan

horizon = ForecastingHorizon(np.arange(1, 4), is_relative=True)

pipe = TransformedTargetForecaster(
    [
        LogTransformer(),
        Differencer(lags=1, na_handling="drop_na"),
        ForecastingPipeline(
            [
                Lag(lags=[0, 1], index_out="original"),
                IndexSubset(index_treatment="remove"),
                PandasTransformAdaptor(method="dropna", kwargs={"axis": "columns", "how": "any"}),
                TabularToSeriesAdaptor(MinMaxScaler()),
                RecursiveTabularRegressionForecaster(RandomForestRegressor(), window_length=2),
            ],
        ),
    ]
)
pipe.fit(y, X, fh=horizon)
pipe.predict(fh=horizon, X=X)

pipe._X
pipe.forecaster_.forecaster_._X
```

**Expected behavior**
Drop only features in `X` that are actually missing at this step in the pipeline.

**Versions**
<details>
System:
    python: 3.10.6 (main, May 29 2023, 11:10:38) [GCC 11.3.0]
executable: /usr/bin/python3.10
   machine: Linux-5.19.0-50-generic-x86_64-with-glibc2.35
Python dependencies:
          pip: 22.0.2
       sktime: 0.21.0
      sklearn: 1.1.2
       skbase: 0.4.6
        numpy: 1.22.4
        scipy: 1.8.1
       pandas: 1.4.4
   matplotlib: 3.5.3
       joblib: 1.1.0
  statsmodels: 0.13.2
        numba: 0.56.0
     pmdarima: 2.0.1
      tsfresh: 0.19.0
   tensorflow: None
tensorflow_probability: None
</details>

hliebert:
#### Reference Issues/PRs
Fixes #4977.

Implements a `DropNA` transformer that saves the dropped index/column names from `fit`.

No.

Any feedback is appreciated. This is mostly a thin wrapper around `pd.DataFrame.dropna`, accepting the arguments `axis`, `how`, and `thresh`, defaulting to `axis=0` and `how=any`. With univariate series, only `axis=0` will work. 

I am validating the inputs to some degree and I've extended the `thresh` argument to also accept a fraction of non-NA values (pandas only accepts counts). I've always found the pandas default of specifying the threshold in terms of non-NA values strange (as opposed to specifying NA values) , but I've left it as is to be consistent. Thoughts on this?

I've added a couple of tests for the functionality and to make sure the parameter validation works. I haven't added any tests with a univariate series so far.

#### Any other comments?
Any feedback is appreciated.

##### For all contributions

### DIFF
--- a/sktime/transformations/series/dropna.py
+++ b/sktime/transformations/series/dropna.py
@@ -108,13 +108,18 @@ class DropNA(BaseTransformer):
 
     def _check_thresh(self, thresh, how):
         """Check thresh parameter, should be a valid value as per docstring."""
-        if not isinstance(thresh, self.VALID_THRESH_TYPES):
+        if not isinstance(thresh, self.VALID_THRESH_TYPES) or isinstance(thresh, bool):
             raise TypeError(
                 f'invalid thresh parameter value encountered: "{thresh}", '
-                f"thresh must be of type: {self.VALID_HOW_VALUES}"
+                f"thresh must be of type: {self.VALID_THRESH_TYPES}"
             )
-        if isinstance(thresh, float) and not (0 < thresh < 1):
-            raise ValueError("thresh must be a fraction between zero and one")
+        if (isinstance(thresh, int) and not (thresh > 0)) or (
+            isinstance(thresh, float) and not (0 < thresh < 1)
+        ):
+            raise ValueError(
+                "thresh must be positive integer or a fraction between zero and one"
+            )
+
         if (how is not None) and (thresh is not None):
             raise TypeError("thresh cannot be set together with how")
 
--- a/sktime/transformations/series/dropna.py
+++ b/sktime/transformations/series/dropna.py
@@ -33,6 +33,11 @@ class DropNA(BaseTransformer):
          If int, require that many non-NA values (as in pandas.dropna). If
          float, share of non-NA values for rows or columns to be retained.
          Cannot be combined with how.
+
+    remember : bool, default False if axis==0, True if axis==1
+        If True, drops the same rows/columns in transform as in fit. If false,
+        drops rows/columns according to the NAs seen in transform (equivalent
+        to PandasTransformAdaptor(method="dropna")).
     """
 
     _tags = {
@@ -51,14 +56,16 @@ class DropNA(BaseTransformer):
     VALID_AXIS_VALUES = [0, "index", 1, "columns"]
     VALID_HOW_VALUES = [None, "any", "all"]
     VALID_THRESH_TYPES = (type(None), int, float)
+    VALID_REMEMBER_TYPES = (type(None), bool)
 
-    def __init__(self, axis=0, how=None, thresh=None):
+    def __init__(self, axis=0, how=None, thresh=None, remember=None):
         self.axis = self._check_axis(axis)
         self.how = self._check_how(how)
         self.thresh = self._check_thresh(thresh, how)
+        self.remember = self._check_remember(remember)
         super().__init__()
 
-        # use numeric axis internally
+        # axis, use numeric axis internally, default rows/index
         if self.axis == "index":
             self._axis = 0
         elif self.axis == "columns":
@@ -66,13 +73,18 @@ class DropNA(BaseTransformer):
         else:
             self._axis = self.axis
 
-        # set default for criterion, default to "any" if neither how nor thresh passed
+        # criterion (how/thresh), default to "any" if neither how nor thresh passed
         if (self.how is None) and (self.thresh is None):
             self._how = "any"
         else:
             self._how = how
-
         self._thresh = self.thresh
+
+        # remember, default to remember dropped columns but not rows
+        if self.remember is None:
+            self._remember = bool(self._axis)
+        else:
+            self._remember = self.remember
 
     def _check_axis(self, axis):
         """Check axis parameter, should be a valid string as per docstring."""
@@ -108,6 +120,16 @@ class DropNA(BaseTransformer):
 
         return thresh
 
+    def _check_remember(self, remember):
+        """Check remember parameter, should be a valid type as per docstring."""
+        if not isinstance(remember, self.VALID_REMEMBER_TYPES):
+            raise ValueError(
+                f'invalid remember parameter value encountered: "{remember}", '
+                f"remember must be of type: {self.VALID_REMEMBER_TYPES}"
+            )
+
+        return remember
+
     def _fit(self, X, y=None):
         """Fit transformer to X and y.
 
@@ -126,18 +148,19 @@ class DropNA(BaseTransformer):
         self: reference to self
         """
         self.dropped_index_values_ = None
-
-        agg_axis = 1 - self._axis
+        self._agg_axis = 1 - self._axis
 
         mask = None
         if self._how == "any":
-            mask = X.isna().any(axis=agg_axis)
+            mask = X.isna().any(axis=self._agg_axis)
         elif self._how == "all":
-            mask = X.isna().all(axis=agg_axis)
+            mask = X.isna().all(axis=self._agg_axis)
         elif isinstance(self._thresh, int):
-            mask = X.count(axis=agg_axis) < self._thresh
+            mask = X.count(axis=self._agg_axis) < self._thresh
         elif isinstance(self._thresh, float):
-            mask = X.count(axis=agg_axis).div(X.shape[agg_axis]) < self._thresh
+            mask = (
+                X.count(axis=self._agg_axis).div(X.shape[self._agg_axis]) < self._thresh
+            )
 
         if mask is not None:
             self.dropped_index_values_ = mask.index[mask].to_list()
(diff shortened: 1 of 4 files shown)


sktime/transformations/series/dropna.py
```diff
@@ -0,0 +1,210 @@
...
+        Returns
+        -------
+        params : dict or list of dict, default = {}
+            Parameters to create testing instances of the class
+            Each dict are parameters to construct an "interesting" test instance, i.e.,
+            `MyClass(**params)` or `MyClass(**params[i])` creates a valid test instance.
+            `create_test_instance` uses the first (or only) dictionary in `params`
+        """
+        params = [
+            {"axis": 0, "how": "any", "thresh": None},
```
fkiraly, line 194: that's a lot of parameter combinations - looks exhaustive.
Might be worth checking the runtime and reducing this if it takes longer than, say, 20 sec.

fkiraly (approved):
Nice! Thanks for the contribution.

I've started the tests, and would be happy to merge if the CI runs through.

Non-blocking comments:

* we should add this to the API reference so other users can find it too. Can be done in another PR, or here. The location would be `docs/source/api_reference/transformations`. I would change the header "missing value imputation" to "missing value treatment" and add it there.
* currently, if you have a panel/hierarchical dataset, and use the setting `axis=1`, this estimator would remove variables by instance, which could result in columns being there for some individual series or not. When merging back, in some data representations this will result in all-NA columns for individual series. Is this intended?

hliebert:
> Nice! Thanks for the contribution.
> 
> I've started the tests, and would be happy to merge if the CI runs through.
> 
> Non-blocking comments:
> 
>     * we should add this to the API reference so other users can find it too. Can be done in another PR, or here. The location would be `docs/source/api_reference/transformations`. I would change the header "missing value imputation" to "missing value treatment" and add it there.

I'll do it here later today or tomorrow. 

>     * currently, if you have a panel/hierarchical dataset, and use the setting `axis=1`, this estimator would remove variables by instance, which could result in columns being there for some individual series or not. When merging back, in some data representations this will result in all-NA columns for individual series. Is this intended?

Not intended, I haven't used sktime with panel or hierarchical data, so I didn't actually consider it.  What should happen in that case? Only merge back the intersection present for all? How can I best distinguish between the input types?

Two more questions I forgot:

- I am saving the dropped index values in `fit` for both `axis=0` and `axis=1`. Does this matter? Should it only be done for columns?

- Should there be some validation for `axis=1` if the user passes a univariate series or is it sufficient to mention it in docstring?

fkiraly:
> I am saving the dropped index values in fit for both axis=0 and axis=1. Does this matter? Should it only be done for columns?

Why would we need to save if for `axis=0`? I get `axis=1`.
Maybe that's an option for the user to control, i.e., whether in `transform` to drop the same columns or base it on the NAs seen there - for `axis=1` I would  have a default of "remember", for `axis=0` I would have the opposite default.

Maybe this is too complicated though, not sure.

> Should there be some validation for axis=1 if the user passes a univariate series or is it sufficient to mention it in the docstring?

What validation? Do you mean, raising an exception and telling the user, you might end up with 0 columns?

I think that might be a valid use case, e.g., if the transformer is used as part of a `ColumnEnsembleTransformer` (it should return a 0-column `pd.DataFrame` then?)

> Not intended, I haven't used sktime with panel or hierarchical data, so I didn't actually consider it. What should happen in that case? Only merge back the intersection present for all?

That would be my intuition. To do this, you need to set `X_inner_mtype = ["pd.DataFrame", "pd-multiindex", "pd_multiindex_hier"]`, so no broadcasting takes place and hierarchical data frames are also seen by `_transform`.

fkiraly:
> I've added the remember option

Yes, that's probably a better solution than what I had in mind, makes sense.

> if univariate Series are always 1d pd.DataFrames, that case should never occur

Well, they will always be *converted* to 1d `pd.DataFrames` before they are seen by `_transform`, so it should be fine.

sktime/transformations/series/dropna.py
```diff
@@ -0,0 +1,239 @@
...
+    how : {'any', 'all'}, default 'any'
+        Determine if row or column is removed from DataFrame, when we have
+        at least one NA or all NA.
+
+        * 'any' : If any NA values are present, drop that row or column.
+        * 'all' : If all values are NA, drop that row or column.
+
+    thresh : int or float, optional
+         If int, require that many non-NA values (as in pandas.dropna). If
+         float, share of non-NA values for rows or columns to be retained.
```
yarnabrina, line 34: Will it make sense to add "at least" for `int` case? Just in case someone misinterpretes it to be exact value.

For `float` case, I found this line a bit difficult to follow. Is `retained` the intended word, because it sounds like `thresh=0.3` mean 30% non-null values will be kept and rest will be removed?

sktime/transformations/series/dropna.py
```diff
@@ -0,0 +1,239 @@
...
+        if self._how == "any":
+            mask = X.isna().any(axis=self._agg_axis)
+        elif self._how == "all":
+            mask = X.isna().all(axis=self._agg_axis)
+        elif isinstance(self._thresh, int):
+            mask = X.count(axis=self._agg_axis) < self._thresh
+        elif isinstance(self._thresh, float):
+            mask = (
+                X.count(axis=self._agg_axis).div(X.shape[self._agg_axis]) < self._thresh
+            )
```
yarnabrina, line 163: Why not pass this task to `pandas`?

```suggestion
        mask = X.dropna(axis=self._agg_axis, how=self._how, thresh=self._thresh)
```

yarnabrina (changes requested):

sktime/transformations/series/tests/test_dropna.py
```diff
@@ -0,0 +1,172 @@
...
+    with pytest.raises(TypeError):
+        DropNA(axis=0, how=str(how), thresh=thresh)
+
+
+@pytest.mark.parametrize("how", ["any", "all", "invalid_value"])
+@pytest.mark.parametrize("thresh", [3, 0.9, True, False, "invalid_value", [1, 2, 3]])
+def test_dropna_invalid_arguments(how, thresh):
+    """Test that invalid arguments or combinations thereof are not accepted."""
+    with pytest.raises((TypeError, ValueError)):
+        DropNA(axis=0, how=str(how), thresh=thresh)
```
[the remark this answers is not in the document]
hliebert: I'll adjust this.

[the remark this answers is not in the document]
hliebert: I separated the tests per your recommendation.

sktime/transformations/series/dropna.py
yarnabrina: Yes, that seems much better to me.

Can you please also mention about being strict proper fractions, 0/1 exclusive part?

sktime/transformations/series/dropna.py
```diff
@@ -0,0 +1,239 @@
...
+        return how
+
+    def _check_thresh(self, thresh, how):
+        """Check thresh parameter, should be a valid value as per docstring."""
+        if not isinstance(thresh, self.VALID_THRESH_TYPES):
+            raise ValueError(
+                f'invalid thresh parameter value encountered: "{thresh}", '
+                f"thresh must be of type: {self.VALID_HOW_VALUES}"
+            )
+        if isinstance(thresh, float) and not (0 < thresh < 1):
```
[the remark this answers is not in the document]
yarnabrina: > `thresh=1.0` is just the same as `how="any"`

This was not immediately clear to me, may be it's worth adding in `Notes` section of docstring?

Unfortunately, I do not follow the equivalence of `thresh=0.0` and `how="all"` cases. Can you please explain that part?

For `how="all"` and `axis="index"`, it means drop a row (index) if all columns is null in it. How do I even interpret `axis="index"` and `thresh=0.0`? Proportion of columns with non-null values has to be at least 0, can not be negative, so the condition to drop will never hold good. And hence I interpreted that `(0, 1.0]` is the valid range for `thresh`, assuming I'm not making a super silly mistake. Please correct my understanding, sorry if I'm missing something obvious.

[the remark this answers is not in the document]
hliebert: I've added this to the docstring. Sorry for the confusion, the `thresh="0.0"` equivalence was a mistake, I responded too late in the evening. I've thought about it and I think you're right, allowing the interval for thresh to be right-open makes sense. I've adjusted the logic and the docstring accordingly.

sktime/transformations/series/dropna.py
hliebert: I've adjusted it as discussed.

fkiraly:
PS: if you think you're done with requests and want to notify reviewers, @hliebert, the "standard protocol" is to click on the "re-review" button top right:

image
(the circling arrows)

hliebert:
> PS: if you think you're done with requests and want to notify reviewers, @hliebert, the "standard protocol" is to click on the "re-review" buttons top right:

Thanks, I didn't know this.

fkiraly:
> Thanks, I didn't know this.

Well, it's very non-obvious...

fkiraly (approved):
all addressed, and tests passing.

Issue #4518, opened 2023-04-28
[ENH] implementing `sktime` probability distributions and simulators


fkiraly:
Umbrella issue for implementing `sktime` probability distributions.

Distributions will move to `skpro`, mid-term, see issue with more distributions to implement, there:
https://github.com/sktime/skpro/issues/22

PR are appreciated to either repository (until the refactor, start in one repo, e.g., `sktime` or `skpro`, and once approved, we will ask you to open the same PR to the other repo)

High priority:

- [x] laplace - https://github.com/sktime/sktime/pull/5161
- [x] empirical incl delta - https://github.com/sktime/sktime/pull/5094
- [ ] multivariate normal

mid priority:

- [x] t - https://github.com/sktime/sktime/pull/5050
- [x] Cauchy (special case of t) - https://github.com/sktime/sktime/pull/5050
- [ ] mixture composition
- [ ] truncated, compositor

Should extend `BaseDistribution`, a good template is `Normal`.

Also see https://github.com/sktime/skpro/issues/22 - contributions can be made to either repository, and should be copied over to the other once approved/merged, until the modules are merged into one.

fkiraly:
Great! This is still a bit experimental, so we appreciate any advice on improving the interfaces!

dhivyeshrk:
Hey @fkiraly! I would like to work on Laplace and Empirical Distributions

fkiraly:
@dhivyeshrk, apologies for the confusion, we forgot to check the boxes.
These are already implemented, the PR were merged.

The distribution functionality will also move to `skpro` mid-term, there are more here:
https://github.com/sktime/skpro/issues/22

How about uniform, or truncated distribution?

fkiraly:
closing since distributions are now in https://github.com/sktime/skpro/issues/22

Issue #1742, opened 2021-12-10
[ENH] wishlist of interval and probabilistic foreasters


fkiraly:
This issue is to create a wishlist of interval forecasters and probabilistic forecasters.

Interval forecasters have `predict_quantile` or `predict_interval`.
(or `predict_var` but that's more variance forecaster)

Probabilistic forecasters will have `predict_proba`, i.e., predict full distributions.

We should also include compositors, especially those converting point foreceasters into interval or probabilistic forecasters.

Inspiration can be the M5 competition, but we should also make sure we have "simple baselines" such as "constant variance = back-test aggregate variance" etc.

ltsaprounis:
1. [Amazon's Level Set Forecaster (LSF)](https://www.amazon.science/publications/probabilistic-forecasting-a-level-set-approach) - very promising results from their presentation in the M5 conference. In short they took the accuracy competition winning method, passed it through the LSF algorithm and they showed that they beat the winner of the uncertainty competition. This is already implemented in [gluon-ts](https://github.com/awslabs/gluon-ts/blob/35c5e6c862f88b8a912d6f34c8b146cd921c8062/src/gluonts/model/rotbaum/_model.py#L68)
2. [Prediction Intervals from Bootstrapped residuals](https://otexts.com/fpp3/prediction-intervals.html#prediction-intervals-from-bootstrapped-residuals)

fkiraly:
From my side:
* `NaiveForecaster` should provide various strategies for estimating forecast variances: squared differences, smoothing residuals, etc. Few shape/scale distributions (normal, laplace). And, going forward, for variance/scale: constant, or "number time steps forward times one step forward".
* naive strategy: take a window over last `n` time points, forecast the empirical distribution corresponding to this.
* variant of this: instead of empirical distribution, use a kernel density estimator on top of it. May require `skpro`, can wrap a kernel density estimator.
* there should be a "squared residuals" (or "absolute residuals") wrapper, "squaring residuals" and related strategies are common compositors to add a volatility/uncertainty prediction. This could use in-sample residuals, or forecast residuals. A forecaster can be applied to the squared or absolute residuals, so it would be a `ResidualUncertaintyForecaster(point_forecaster, uncertainty_forecaster)`..
* taking quantile forecaster(s) and composing a distribution based on forecast quantiles. `QuantileForecasterComposer` or similar.
[...]

PR #5050, opened 2023-08-07
[ENH] Student's t-distribution

Alex-JG3:
#### Reference Issues/PRs

Related to #4518

Add t-distribution to the `proba` subpackage.

No

Not yet.

##### For all contributions

### DIFF
--- a/sktime/proba/t.py
+++ b/sktime/proba/t.py
@@ -69,7 +69,8 @@ class TDistribution(BaseDistribution):
         r"""Return expected value of the distribution.
 
         Let :math:`X` be a random variable with the distribution of `self`.
-        Returns the expectation :math:`\mathbb{E}[X]`
+        Returns the expectation :math:`\mathbb{E}[X]`. The expectation,
+        :math:`\mathbb{E}[X]`, as infinite if :math:`\nu \le 1`.
 
         Returns
         -------
--- a/sktime/proba/t.py
+++ b/sktime/proba/t.py
@@ -32,8 +32,8 @@ class TDistribution(BaseDistribution):
     """
 
     _tags = {
-        "capabilities:approx": ["pdfnorm", "energy", "ppf"],
-        "capabilities:exact": ["mean", "var", "pdf", "log_pdf", "cdf"],
+        "capabilities:approx": ["pdfnorm", "energy"],
+        "capabilities:exact": ["mean", "var", "pdf", "log_pdf", "cdf", "ppf"],
         "distr:measuretype": "continuous",
     }
 
--- a/sktime/proba/t.py
+++ b/sktime/proba/t.py
@@ -110,26 +110,26 @@ class TDistribution(BaseDistribution):
     def pdf(self, x):
         """Probability density function."""
         d = self.loc[x.index, x.columns]
-        pdf_arr = gamma((d.df + 1) / 2)
-        pdf_arr = pdf_arr / (np.sqrt(np.pi * d.df) * gamma(d.df / 2))
-        pdf_arr = pdf_arr * (1 + x**2 / d.df) ** (-(d.df + 1) / 2)
+        pdf_arr = gamma((d._df + 1) / 2)
+        pdf_arr = pdf_arr / (np.sqrt(np.pi * d._df) * gamma(d._df / 2))
+        pdf_arr = pdf_arr * (1 + x**2 / d._df) ** (-(d._df + 1) / 2)
         return pd.DataFrame(pdf_arr, index=x.index, columns=x.columns)
 
     def log_pdf(self, x):
         """Logarithmic probability density function."""
         d = self.loc[x.index, x.columns]
-        lpdf_arr = loggamma((d.df + 1) / 2)
-        lpdf_arr = lpdf_arr - 0.5 * np.log(d.df * np.pi)
-        lpdf_arr = lpdf_arr - loggamma(d.df / 2)
-        lpdf_arr = lpdf_arr - ((d.df + 1) / 2) * np.log(1 + x**2 / d.df)
+        lpdf_arr = loggamma((d._df + 1) / 2)
+        lpdf_arr = lpdf_arr - 0.5 * np.log(d._df * np.pi)
+        lpdf_arr = lpdf_arr - loggamma(d._df / 2)
+        lpdf_arr = lpdf_arr - ((d._df + 1) / 2) * np.log(1 + x**2 / d._df)
         return pd.DataFrame(lpdf_arr, index=x.index, columns=x.columns)
 
     def cdf(self, x):
         """Cumulative distribution function."""
         d = self.loc[x.index, x.columns]
-        cdf_arr = x * gamma((d.df + 1) / 2)
-        cdf_arr = cdf_arr * hyp2f1(0.5, (d.df + 1) / 2, 3 / 2, -(x**2) / d.df)
-        cdf_arr = 0.5 + cdf_arr / (np.sqrt(np.pi * d.df) * gamma(d.df / 2))
+        cdf_arr = x * gamma((d._df + 1) / 2)
+        cdf_arr = cdf_arr * hyp2f1(0.5, (d._df + 1) / 2, 3 / 2, -(x**2) / d._df)
+        cdf_arr = 0.5 + cdf_arr / (np.sqrt(np.pi * d._df) * gamma(d._df / 2))
         return pd.DataFrame(cdf_arr, index=x.index, columns=x.columns)
 
     def ppf(self, p):
@@ -142,12 +142,12 @@ class TDistribution(BaseDistribution):
 
         mask1 = (p.values < 0.5) & (p.values > 0)
         mask2 = (p.values < 1) & (p.values > 0.5)
-        ppf_arr[mask1] = 1 / betaincinv(0.5 * d.df[mask1], 0.5, 2 * ppf_arr[mask1])
+        ppf_arr[mask1] = 1 / betaincinv(0.5 * d._df[mask1], 0.5, 2 * ppf_arr[mask1])
         ppf_arr[mask2] = 1 / betaincinv(
-            0.5 * d.df[mask2], 0.5, 2 * (1 - ppf_arr[mask2])
+            0.5 * d._df[mask2], 0.5, 2 * (1 - ppf_arr[mask2])
         )
         ppf_arr[mask1 | mask2] = np.sqrt(ppf_arr[mask1 | mask2] - 1)
-        ppf_arr[mask1 | mask2] = np.sqrt(d.df[mask1 | mask2]) * ppf_arr[mask1 | mask2]
+        ppf_arr[mask1 | mask2] = np.sqrt(d._df[mask1 | mask2]) * ppf_arr[mask1 | mask2]
         ppf_arr[mask1] = -ppf_arr[mask1]
         return pd.DataFrame(ppf_arr, index=p.index, columns=p.columns)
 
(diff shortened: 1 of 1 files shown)


sktime/proba/t.py
```diff
@@ -0,0 +1,160 @@
+# copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
+"""Student's t-distribution."""
+
+__author__ = ["fkiraly"]
```
fkiraly, line 4: well, you're the author, not I!

sktime/proba/t.py
```diff
@@ -0,0 +1,160 @@
...
+    >>> n = TDistribution(mu=[[0, 1], [2, 3], [4, 5]], sigma=1, df=10)
+    """
+
+    _tags = {
+        "capabilities:approx": ["pdfnorm"],
+        "capabilities:exact": ["mean", "var", "energy", "pdf", "log_pdf", "cdf", "ppf"],
+        "distr:measuretype": "continuous",
+    }
+
+    def __init__(self, mu, sigma, df, index=None, columns=None):
```
fkiraly, line 38: you'll either need to provide `df` in the `get_test_params`, or set a default here.

Would `df=1` be a good default, that gives Cauchy?

sktime/proba/t.py
```diff
@@ -0,0 +1,160 @@
...
+        ----------
+        x : None or pd.DataFrame, optional, default=None
+            if pd.DataFrame, must have same rows and columns as `self`
+
+        Returns
+        -------
+        pd.DataFrame with same rows as `self`, single column `"energy"`
+        each row contains one float, self-energy/energy as described above.
+        """
+        if x is None:
```
fkiraly, line 84: these formulae look suspicious - should `df` not figure in here? Or is this just a copy-paste from normal?

sktime/proba/t.py
```diff
@@ -0,0 +1,160 @@
...
+
+        if index is None:
+            index = pd.RangeIndex(shape[0])
+
+        if columns is None:
+            columns = pd.RangeIndex(shape[1])
+
+        super().__init__(index=index, columns=columns)
+
+    def _get_bc_params(self):
```
fkiraly, line 56: hm, maybe we should factor this out later

sktime/proba/t.py
```diff
@@ -0,0 +1,160 @@
...
+
+        Let :math:`X` be a random variable with the distribution of `self`.
+        Returns :math:`\mathbb{V}[X] = \mathbb{E}\left(X - \mathbb{E}[X]\right)^2`
+
+        Returns
+        -------
+        pd.DataFrame with same rows, columns as `self`
+        variance of distribution (entry-wise)
+        """
+        sd_arr = self._sigma
```
fkiraly, line 120: that's not right? It should be df/(df+2)

Alex-JG3:
@fkiraly, my apologies, this is basically a copy paste from the normal currently. I have only changed the `pdf` function at the moment. Still working through the rest. I should have made this clearer.

fkiraly:
just posting this here in case you're not aware
https://xkcd.com/1347/

sktime/proba/t.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def pdf(self, x):
+        """Probability density function."""
+        d = self.loc[x.index, x.columns]
+        pdf_arr = gamma((d.df + 1) / 2)
+        pdf_arr = pdf_arr / (np.sqrt(np.pi * d.df) * gamma(d.df / 2))
+        pdf_arr = pdf_arr * (1 + x**2 / d.df) ** (-(d.df + 1) / 2)
+        return pd.DataFrame(pdf_arr, index=x.index, columns=x.columns)
+
+    def log_pdf(self, x):
```
Alex-JG3, line 145: Couldn't find a formula for the log of the pdf. I had a go at doing it myself and tried to use scipy's `loggamma` function as much as possible.

Alex-JG3:
@fkiraly, I couldn't find a formula for energy, if you have one I am happy to implement it. Also, from what I have read it sounds like there isn't an analytic solution for probability point function for the t-distribution. Are you aware of an analytic solution for the ppf?

fkiraly:
> I couldn't find a formula for energy, if you have one I am happy to implement it. 

Thanks! It's nasty though, so don't do it if it feels to unpleasant.

Appendix A.2 of ["evaluating forecasts with `scoringRules`"](https://www.jstatsoft.org/article/view/v090i12) has a few explicit formulae for the energy, including the t-distribution. The expression is hard to track in implementation, so I would advise comparing against the Monte-Carlo default if you implement it.

> Are you aware of an analytic solution for the ppf?

The inverse t cdf tends to appear, I'll have a look.

sktime/proba/t.py
fkiraly: nice!

Alex-JG3:
>  here's what Wolfram Alpha says: https://www.wolframalpha.com/input/?i=inverse+cdf+student%27s+t+distribution

Just added the ppf using the Wolfram Alpha link. Thanks for sharing that link.

fkiraly:
Have you checked that it's indeed inverse to the cdf?

This should probably be part of the tests, too, not to self.

Alex-JG3:
>Have you checked that it's indeed inverse to the cdf?

Here is an example showing the output of the `ppf` is the inverse of the `cdf`.

```python
>>> import pandas as pd
>>> from sktime.proba.t import TDistribution
>>> n = TDistribution(mu=0, sigma=1, df=[[1, 1], [1, 1], [1, 1]])
>>> p = pd.DataFrame({"col1": [0.1, 0.3, 0.5], "col2": [0.7, 0.9, 1.0]})
>>> n.cdf(n.ppf(p))
   col1  col2
0   0.1   0.7
1   0.3   0.9
2   0.5   NaN
```

Alex-JG3:
@fkiraly, I am unsure where to start with the energy function. I see the section for student's t-distribution in A.2 of ["evaluating forecasts with scoringRules"](https://www.jstatsoft.org/article/view/v090i12) but, since there isn't a flashing neon sign saying "this is the energy function!" I am unsure of what needs to be implemented.

I assume the continuous ranked probability score (CRPS) is related to energy but it is not obvious to me how.

Would it be possible to get a review and perhaps merge in what is currently implemented and leave energy function for a different PR unless you want to take a look at energy on this PR?

fkiraly:
> Would it be possible to get a review and perhaps merge in what is currently implemented and leave energy function for a different PR unless you want to take a look at energy on this PR?

Sure!

Have you checked why the tests are failing?
There's something going on with indexing. Happy to have a more detailed look if you're not sure what is going on.

fkiraly:
> I assume the continuous ranked probability score (CRPS) is related to energy but it is not obvious to me how.

Yes, the general formula is `CRPS(d.cdf, y) = d.energy(y) - 0.5 * d.energy()`. Unfortunately the reference only gives CRPS, but the scaling behaviour of the terms makes me think that the positive terms are `d.energy(y)`, and then negative term is ` - 0.5 * d.energy()`. If not, `d.energy(y)` should be the positive terms minus a constant multiple of the negative term (the coefficient should be "simple" and guessable by comparison with the Monte Carlo estimate).

fkiraly:
Btw @Alex-JG3, thanks a lot!
What would also be useful is feedback on the implementer experience - e.g., is there boilerplate that is unclear, or interface elements that you think would be better changed?

sktime/proba/t.py
```diff
@@ -0,0 +1,164 @@
...
+    >>> n = TDistribution(mu=[[0, 1], [2, 3], [4, 5]], sigma=1, df=10)
+    """
+
+    _tags = {
+        "capabilities:approx": ["pdfnorm", "energy"],
+        "capabilities:exact": ["mean", "var", "pdf", "log_pdf", "cdf", "ppf"],
+        "distr:measuretype": "continuous",
+    }
+
+    def __init__(self, mu, sigma, df=2, index=None, columns=None):
```
Alex-JG3, line 40: @fkiraly, I have change the default to `df` to 2. This is because the variance tests were failing because [`_check_output__format`](https://github.com/sktime/sktime/blob/main/sktime/proba/tests/test_all_distrs.py#L143) checks to see if all variances are greater than or equal to zero. Variance for the t-distribution is undefined when the degrees of freedom is 1 or less. Do you think this test needs to be updated?

fkiraly: I see, interesting.
Perhaps we need to adapt the tests.

Although, are you sure the implementation is correct? I thought even for `df=2` the variance was infinite, so there might be something not right.

For `df=1`, the mean is undefined.

One thought: can we allow nans in the test?

Alex-JG3: >Although, are you sure the implementation is correct? I thought even for df=2 the variance was infinite, so there might be something not right.

Here is an example for the current implementation.

```python
>>> import pandas as pd
>>> from sktime.proba.t import TDistribution
>>> d = TDistribution(mu=0, sigma=1, df=[[0, 1], [2, 3], [4, 5]])
>>> x = pd.DataFrame([[0.0, 1.0], [1.5, 2.0], [3.0, 5.0]])
>>> d.var()
     0         1
0  NaN       NaN
1  inf  3.000000
2  2.0  1.666667
```
Does this look right to you?

>For df=1, the mean is undefined.

Currently that is not happening. I will make the change.

>One thought: can we allow nans in the test?

Could we use `np.greater_equal` and use the `where` argument to ignore `nans`?

fkiraly: > Does this look right to you?

Hmmm, yes, though shouldn't the first row not also be `inf`?
Also, what were you using `x` for? (were you going to write another line?)

> Could we use `np.greater_equal` and use the `where` argument to ignore nans?

That would work, although replacing nan-s with `np.inf` might be a more principled solution? All univariate distributions have a variance, if you allow for infinity (to mean divergence)

sktime/proba/t.py
```diff
@@ -0,0 +1,164 @@
...
+from sktime.proba.base import BaseDistribution
+
+
+class TDistribution(BaseDistribution):
+    """Student's t-distribution (sktime native).
+
+    Parameters
+    ----------
+    mean : float or array of float (1D or 2D)
+        mean of the normal distribution
```
fkiraly, line 19: I think the word "normal" is a copy-paste accident?

fkiraly (approved):
Excellent, really nice!

I haven't checked all the formulae, but I did spot checks and it looked good.

Would you be so kind and also open a PR on `skpro`?
See https://github.com/sktime/skpro/issues/22
Just copy-paste this in the appropriate folder and change the `BaseDistribution` import location, the test framework will automatically do the rest.

If you're interested, you could:

* implement more distributions
* join on the next phease refinement of the distributions interface https://github.com/sktime/skpro/issues/21
* contribute probabilistic regressors in `skpro` - https://github.com/sktime/skpro/issues/7
* contribute probabilistic forecasters in `sktime` - https://github.com/sktime/sktime/issues/1742

(happy to chat on discord for planning!)

sktime/proba/t.py
fkiraly: superseded - was just a draft

sktime/proba/t.py
fkiraly: captured in https://github.com/sktime/skpro/issues/21

sktime/proba/t.py
fkiraly: superseded, was preliminary draft

Alex-JG3:
> Is there boilerplate that is unclear, or interface elements that you think would be better changed?

@fkiraly, generally I found it quite clear how to adapt what has been done already to the t-distribution. The implementation of the normal distribution is a great example. Some things that were less clear to me:

* It wasn't obvious to me that methods that are not implemented are estimated numerically but this is a very cool feature!
* I don't quite understand the benefit using pandas-like syntax as opposed to numpy-like indexing. I see that you can pass your own index/columns but I myself would probably avoid doing that.

PR #5055, opened 2023-08-08
[ENH] Extend benchmarking to accept sktime-in-memory data formats

hazrulakmal:
#### Reference Issues/PRs
partly towards #4873 

PR to allow kotsu benchmarking to accept sktime-in-memory container (which is what is widely used across sktime as data ingestion input) listed in MTYPE_REGISTER (mostly from pd.DataFrame) for Series, Multivariate and Hierarchical time series. 

In addition, also allows users to access estimator predictions and training and test time points.

I added  `test_dataset_different_format` to check for output consistency.

#### Any other comments?

There is a design question that I would like to discuss. The functions `forecasting_validation` and `_factory_forecasting_validation` are currently created as separate functions outside of the `ForecastingBenchmark` class. I propose that we can convert them into methods instead, to make `forecasting_validation` an abstract validation method, say, called `.prepare_forecasting_validation`.  By doing this, we allow users to substitute the evaluation procedure if they do not want to use the default `evaluate` method. Instead, they can use their own evaluation steps by subclassing `ForecastingBenchmark` and replacing the `prepare_forecasting_validation` method with their own evaluation steps.

are they any specific reasons why they were created as functions rather than method in the first place?

fyi @fkiraly @DBCerigo @alex-hh


fkiraly:
> are they any specific reasons why they were created as functions rather than method in the first place?

This is not my design - but I would agree that methods make more sense.

I also find the nested argument design where you pass functions to functions quite strange. Surely this can be solved more cleanly, e.g., with passing classes and dispatching?

hazrulakmal:
## SKTIME MTYPE

The enhancement made is to accept dataset_loader as a tuple of the format (y, X), where y and X are in sktime MTYPE. When X is not available, we expect dataset_loader as (y, None).

For a callable dataset_loader, previously it is assumed to return a single value, to accept a target variable. Hence, there's no way to allow X exogenous variable(s) in. This PR also makes an adjustment to account for a callable that returns two values, so that benchmarking can now accept exogenous variables as well when a callable is given.

Now to input X variable(s), users can do either of the following

```python
from sktime.benchmarking.forecasting import ForecastingBenchmark

y, X = load_longley() # an abstract pre-processed MTYPE of any datasets
benchmark = ForecastingBenchmark()
benchmark.add_task(
    (y, X),
    CV,
    scores
)

```

or

```python
from sktime.benchmarking.forecasting import ForecastingBenchmark

benchmark = ForecastingBenchmark()
benchmark.add_task(
    load_longley,
    CV,
    scores
)

```

## Estimator Predictions

Storing estimator predictions can be quite tricky. My initial solution is to simply pass the predictions, train and test sets from the `evaluate` output to the benchmark results dictionary. However, this will result in redundancy (in the order of estimators) of the train and test sets in the final output. this is a serious problem when the dataset is huge. 

Anyway, the reason for saving train and test sets on top of the prediction is to allow users to perform further post-processing analysis, such as statistical tests, visualizations etc, after running the benchmarks.

Currently, I am considering saving only the estimator predictions and finding another way to save the train and test sets for each fold.

Saving the estimator predictions can pose a problem when the resulting dataframe is saved as a CSV file. If a user wants to look at the benchmark result and loads the CSV file back into memory, the saved predictions may not load properly because they were saved as MTYPE (such as pd.dataframe). This is a general drawback of CSV files. One solution is to serialize the dataframe as a pickle, which can then be deserialized properly to the original form without issues that are present with CSV files. However, this also means that an override is needed to change the format in which the data is saved.

fkiraly:
Regarding design questions, here are some ideas:

* I think the cleanest way would be to give `evaluate` a suitable signature, e.g., allowing to pass multiple estimators, and add the estimator name as another index. This would still multiplicate the data, but in a second step we could move/deprecate to a format where the true splits are returned in a separate output. Going with non-nested data containers would be better anyway.
* Now, a full change of `evaluate` might be too much for doing it in one step. In the interim, we could just discard anything that's multiplied, e.g., get data only from the first estimator?
* regarding storing the output: I think a clean way would be an option to pass a reference to an object, which is, if passed, where outputs are written. Or, more generally, a config such as `{"path": "path/to/file", "serializer": "pickle"}`. Going about it this way would allow to retain the status quo as a default until deprecation, and would allow more reasonable options like pickle, or using the object's own serializer (the `save` method!)

hazrulakmal:
> I think a clean way would be an option to pass a reference to an object

It seems that Kotsu has limitations when saving a nested container because it uses the default CSV saving option. Changing or extending this option would require a significant override of the Kotsu run function. Therefore, I have decided to deal with this issue separately in another PR to avoid cluttering this PR.

fkiraly (changes requested):
Apologies for missing this. Nice! I thikn this is good to go up to a documentation issue, to avoid user confusion.

Blocking:

* I think we also need to document clearly the case where there is no exogenous data - as far as I understand the code, this is supported, but not documened in the docstring.

not blocking:

* in such a case, we should think whether plain `y` should also be possible to pass. Might bring with it type checking issues though

Issue #5020, opened 2023-08-04
[DOC] in `ForecastingGridSearchCV` and forecaster tuner docstrings, improve description of `scoring` parameter


TeoSkondras:
#### In ForecastingGridSearchCV, it is not well specified what the scoring attribute possible values are...

For example:

gscv = ForecastingGridSearchCV(
    forecaster=pipe_x,
    param_grid=[{
        "forecaster__forecaster__estimator__n_components": [1,2,3,4]
    }
    ],
    cv=cv,
    n_jobs=-1,
    scoring="neg_mean_absolute_error")

TypeError: scoring parameter of ForecastingGridSearchCV must be an sktime metric, descendant ofBaseMetric, or a callable with signature (y_true: 1D np.ndarray, y_pred: 1D np.ndarray) -> float, assuming np.ndarrays being of the same length, and lower being better.

fkiraly:
Done, would appreciate a look whether this adds clarity, @TeoSkondras.
I also included a line of code that lets you list all the metrics available.

fkiraly:
Other question: since you expected strings to work - should we make strings a valid argument type, too?

TeoSkondras:
@fkiraly Your pull request fixes the doc perfectly; thanks!
Regarding the string functionality, I think that in sklearn this is their approach regarding param grid scoring, if you want to follow the same principle its fine, if not I am perfectly ok with the solution you gave. It is up to you

TeoSkondras:
@fkiraly another good thing would be to allow more than one metric, ex. r2 and mae. t would be really useful!

hazrulakmal:
> @fkiraly another good thing would be to allow more than one metric, ex. r2 and mae. t would be really useful!

relates to #3508

fkiraly:
> Regarding the string functionality, I think that in sklearn this is their approach regarding param grid scoring, if you want to follow the same principle its fine, if not I am perfectly ok with the solution you gave. It is up to you

Thinking about it

fkiraly:
@TeoSkondras, draft PR for alias strings in the tuners are here: https://github.com/sktime/sktime/pull/5058

PR #5058, opened 2023-08-09
[ENH] sktime str/object aliasing registry mechanism, alias strings for `scoring` argument in tuners

fkiraly:
This PR adds a registry-based aliasing mechanisms for strings, and, as a proof-of-concept, implements aliasing for the `scoring` argument in forecasting tuners such as `ForecastingGridSearchCV`.

The aliasing mechanism is implemented in a new aliasing resolution endpoint `registry.resolve_alias`, which is internally based on:

* the `craft` utility, as a means for general string aliasing
* a new `ALIAS_DICT` dictionary which allows definition of custom aliases, which take priority over `craft` inside `resolve_alias`

Related:

* discussion with @teoskondras on string aliasing of `scoring` in tuners: https://github.com/sktime/sktime/issues/5020#issuecomment-1667323583
* PR https://github.com/sktime/sktime/pull/4712 on string aliasing for distances in clusterers

Depends on https://github.com/sktime/sktime/pull/5066 for a bugfix to `craft`.


fkiraly:
FYI @achieveordie, this introduces a single registry endpoint for string resolution. Would like to hear your opinion, especially in the context of #4712.

Any logic in #4712 could tie in here under the hood, although I'm unsure about the exact details, and/or whether we want to modify the logic within `resolve_alias` to account for additional logic in your PR.

PR #5059, opened 2023-08-09
[MNT] add more soft dependencies to `show_versions`

fkiraly:
This adds more soft dependencies to a standard `show_versions` return:

* `pytorch` which is increasingly common as a DL back-end
* `tslearn` for clustering, classification, distances
* `statsforecast` from the nixtla-verse

This is to ensure that standard version dumps in bug reports cover common cases.


yarnabrina (approved):
Approved, obviously.

Two completely non-blocking suggestions:

1. Since it has the capability to handle `ImportError`, can we consider adding all soft dependencies? I can not think of any harm in that.
2. Can the list of packages be sorted alphabetically?

fkiraly:
> I can not think of any harm in that.

Well, the harm I'm thinking of is an extremely long list in bug reports, that might make it harder for both reporter and diagnoser to read. It's more low-level stochastic harm by noise, really, than serious harm.

PR #5060, opened 2023-08-09
[ENH] add speed test for negative `check_is_mtype` outcomes

fkiraly:
This adds a max duration check on `test_check_negative`, which tests uses of `check_is_mtype` with an expected negative outcome, on other `mtype`-s.

The max duration is set to `1e-5` seconds, as on examples of other mtypes the negative result should be identifiable from one or few type checks which should take orders of magnitude less.

Also removes what looks like an accidental merge-duplication in `test_check_negative` that led to carrying out the test twice.

### DIFF
--- a/sktime/datatypes/tests/test_check.py
+++ b/sktime/datatypes/tests/test_check.py
@@ -2,6 +2,8 @@
 
 __author__ = ["fkiraly"]
 
+from time import time
+
 import numpy as np
 
 from sktime.datatypes._check import (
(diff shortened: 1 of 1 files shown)


yarnabrina:
A proposal: use `time.perf_counter()` instead of `time.time()`

Ref. https://docs.python.org/3/library/time.html#time.perf_counter

fkiraly:
some of these take slightly longer than `1e-5` - any idea why? I tracked the logic behind some of these, and they should almost all fail at an initial `isinstance` check.

PR #5068, opened 2023-08-10
[DOC] Improve Docstring for PinballLoss

Ram0nB:
#### Reference Issues/PRs

Issue #4303

Enhances the PinballLoss docstring: example added

No

#### Any other comments?

This is my first contribution. Please let me know where I can improve - also for setting up this Pull Request. Many thanks!


fkiraly (changes requested):
Thanks a lot!

Might I request the following change - for expressions going over multiple rows, like `y_pred = ` etc, from the second row on you need to use `... ` rather than `>>> `, or this will give problems with testing and display in the docs.

yarnabrina (changes requested):
Thanks for the contribution. Apart from what @fkiraly has said above already, did you mean to edit `_BaseProbaForecastingErrorMetric` class? Your example, and title of this PR, seem to be more suitable for `PinballLoss` class?

Ram0nB:
First of all, thanks for your quick replies!

> Might I request the following change - for expressions going over multiple rows, like `y_pred = ` etc, from the second row on you need to use `... ` rather than `>>> `, or this will give problems with testing and display in the docs.

Change processed.

> Thanks for the contribution. Apart from what @fkiraly has said above already, did you mean to edit `_BaseProbaForecastingErrorMetric` class? Your example, and title of this PR, seem to be more suitable for `PinballLoss` class?

I indeed meant to edit the `PinballLoss` class. I now moved the example to the `PinballLoss` class - sorry about this.

Also, the checklist for all contributors state that:

Should I also add myself to the https://github.com/sktime/sktime/blob/main/.all-contributorsrc file in this pull request? If not - how do I add myself?

fkiraly:
> Should I also add myself to the https://github.com/sktime/sktime/blob/main/.all-contributorsrc file in this pull request?

Yes, with a `doc` contribution!

fkiraly:
Minor linting issue, you have trailing whitespaces.

See here for general coding style and how to set up automated checking locally:

(or, just go on the failing test and see where the problem is)

fkiraly (approved):
Fixed it now.

yarnabrina:
The `numpy` issue is out of scope for this PR, but I don't understand why there are so many fatal bad revision in the logs.

@fkiraly, you updated this with latest `main` right? I see no such fatal revision stuff in `main` CI logs, but quite a lot here. Any idea what is causing this?

fkiraly:
@yarnabrina, there are two separate issues:

* the fatal revisions are due to the `git_diff` utility running in `test-nosoftdeps-full` but the `main` branch not being present. This is fixed in https://github.com/sktime/sktime/pull/5084 (would appreciate review)!
* the `numpy` equality failures are due to equality comparisons to `np.ndarray` having deprecated at some point and we haven't tested it, making the weighted versions of metrics fail. The problem and the lack of tests are fixed in https://github.com/sktime/sktime/pull/5086 (would appreciate review)

Swift reviews for both would be appreciated if you have time, then we can put this in the 0.21.1 release. Especially the failure of weighted metrics seems important.

fkiraly:
@Ram0nB, thanks a lot for this PR - the manually generated `y_pred` is very useful for testing, and this has led us to discover some issues with the weighted losses! (including that this was not properly tested)

Starting from the explicit `y_pred`: we're currently trying to cut down on test time, see https://github.com/sktime/sktime/issues/2890. We're close to the 10min target!

What we've discovered, the fixture generation for `y_pred` uses actual forecasters' predictions, and that takes ca 3 sec per test run for dozens of tests, so we could get far by replacing the fixtures with explicit `y_pred`!

Would you be inclined to help with that?

It is just replacing `y_pred = forecaster.predict(etcetc)` with an explicit generation of `y_pred` like in your doctest.

The precise location is `performance_metrics.forecasting.probabilistic.tests.test_probabilistic_metrics`, and in there the fixture called `sample_data`. As said, any logic that calls `NaiveForecaster` to generate the forecasts should be replaced with an explicit `y_pred = pd.DataFrame(etc)` (or possibly `pd.Series`)

fkiraly:
FYI, my commits are coming from testing this with the branch #5104 - they shouldn't be in the diff in the end.

PR #5069, opened 2023-08-10
[BUG] fix wrong logic for `index_out="shift"` in `Lag` transformer

fkiraly:
Fixes an unreported bug: due to a confusion between strings `"shift"` and `"shifted"`, the `Lag` transformer would run incorrect logic for `index_out="shift"`.

This has been rectified, and the test has been sharpened to check for this.

### DIFF
--- a/sktime/transformations/series/lag.py
+++ b/sktime/transformations/series/lag.py
@@ -248,7 +248,7 @@ class Lag(BaseTransformer):
                 Xt = Xt.loc[X_orig_idx]
             # sub-set to shifted index, if "shifted"
             # this is necessary, because we added indices from _X above
-            if index_out == "shifted":
+            if index_out == "shift":
                 Xt = Xt.loc[X_orig_idx_shifted]
 
             Xt_list.append(Xt)
--- a/sktime/transformations/series/lag.py
+++ b/sktime/transformations/series/lag.py
@@ -246,7 +246,7 @@ class Lag(BaseTransformer):
             # sub-set to original, if "original"
             if index_out == "original":
                 Xt = Xt.loc[X_orig_idx]
-            # sub-set to shifted index, if "shifted"
+            # sub-set to shifted index, if "shift"
             # this is necessary, because we added indices from _X above
             if index_out == "shift":
                 Xt = Xt.loc[X_orig_idx_shifted]
--- a/sktime/transformations/series/tests/test_lag.py
+++ b/sktime/transformations/series/tests/test_lag.py
@@ -47,6 +47,23 @@ def test_lag_fit_transform_out_index(X, index_out):
 
 @pytest.mark.parametrize("X", X_fixtures)
 @pytest.mark.parametrize("index_out", index_outs)
+def test_lag_fit_transform_out_values(X, index_out):
+    """Test that index sets of fit_transform output behave as expected."""
+    t = Lag(2, index_out=index_out)
+    X_fit = X[:2]
+    X_trafo = X[2:]
+    Xt = t.fit(X_fit).transform(X_trafo)
+
+    if index_out in ["original", "extend"]:
+        assert all(Xt.iloc[0].values == X_fit.iloc[0].values)
+        assert all(Xt.iloc[2].values == X_trafo.iloc[0].values)
+
+    elif index_out == "shift":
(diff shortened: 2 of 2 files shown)


sktime/transformations/series/lag.py
```diff
@@ -248,7 +248,7 @@ def _transform(self, X, y=None):
                 Xt = Xt.loc[X_orig_idx]
             # sub-set to shifted index, if "shifted"
```
benHeid, line 249: In the comment, there is still a "shifted".

sktime/transformations/series/tests/test_lag.py
```diff
@@ -42,6 +42,27 @@ def test_lag_fit_transform_out_index(X, index_out):
...
+    """Test that index sets of fit_transform output behave as expected."""
+    t = Lag(2, index_out=index_out)
+    if isinstance(X, pd.Series):
+        X = pd.DataFrame(X)
+    X_fit = X[:2]
+    X_trafo = X[2:]
+    Xt = t.fit(X_fit).transform(X_trafo)
+
+    if index_out in ["original", "extend"]:
+        assert all(Xt.iloc[0].values == X_fit.iloc[0].values)
```
benHeid, line 60: An alternative here would be to use:

```python
np.testing.assert_equal(Xt.iloc[2].values, X_fit.iloc[0].values)
```

or an equivalent of pandas instead of the assert statement. This should have the advantage that `np.array([np.nan])`  would be asserted to be equal with `np.array([np.nan])`, while `assert (np.array([np.nan]) == np.array([np.nan])).all()` would raise an exception.

sktime/transformations/series/tests/test_lag.py
```diff
@@ -42,6 +42,27 @@ def test_lag_fit_transform_out_index(X, index_out):
...
+        if len(Xt) > 2:
+            assert all(Xt.iloc[2].values == X_trafo.iloc[0].values)
```
benHeid, line 62: See comment above

sktime/transformations/series/tests/test_lag.py
```diff
@@ -42,6 +42,27 @@ def test_lag_fit_transform_out_index(X, index_out):
...
+
+    elif index_out == "shift":
+        assert (Xt.values == X_trafo.values).all()
```
benHeid, line 65: See comment above

benHeid (changes requested):
Thank you for your PR.  There are two different remarks on the PR. Happy to discuss them :)

sktime/transformations/series/tests/test_lag.py
fkiraly: good point - is that just removing `values`?

benHeid: No, it is replacing line 60 with the proposed line of code. (If I got the question right)

fkiraly: Oh, I see., `np.testing`. Makes sense.

I meant, do you know what is the "equivalent of pandas"?

sktime/transformations/series/lag.py
fkiraly: fixed

sktime/transformations/series/tests/test_lag.py
fkiraly: changed to `np.testing`

sktime/transformations/series/tests/test_lag.py
fkiraly: changed to `np.testing.assert_equal`

sktime/transformations/series/tests/test_lag.py
fkiraly: changed to `np.testing.assert_equal`

sktime/transformations/series/tests/test_lag.py
benHeid: pd.testing.assert_frame_equal or pd.testing.assert_series_equal is the corresponding pandas testing function. However, If I rethink it, this would probably not work since we change the index when applying the Lag. Thus, it is probably better to stay with numpy.testing here.

benHeid (approved):
Thank you :)

Issue #1634, opened 2021-11-21
Technical debt in aligner module


Lovkush-A:
During the review process of #1264, a few items of work came up that were better dealt with in separate PRs.

1. creating a notebook demo 
2. having more comprehensive unittests
  * e.g. covering cases with multiple alignment
  * e.g. testing the dtw aligner's contract somehow. (note, checking for identity alignment is not sound. I think one option would be to create a mock object for the dtw package we wrap)
3. (maybe). have more detailed documentation for `_fit` in dtw wrapper classes, beyond 'core logic of fit'. this suggest is intended as develoepr documentation rather than end-user documentation.

PR #5074, opened 2023-08-10
[DOC] API reference for time series aligners

fkiraly:
Adds API reference for the time series alignment module.
Looks like this was missing (even though linked from README).

Towards https://github.com/sktime/sktime/issues/1634


fkiraly:
merging quickly, as this is an important doc page missing on main, with dead link.

PR #5078, opened 2023-08-11
[DOC] improves formatting of docstring examples

### LINKED ISSUE #5013: [DOC] Improve documentation formatting by replacing empty lines with comments
Consider the [documentation](https://www.sktime.net/en/latest/api_reference/auto_generated/sktime.param_est.seasonality.SeasonalityACF.html) of `SeasonalityACF` as an example:

image

This is not very readable, and not at all copy-paste friendly.

My guess is this is caused by the new lines in the `Examples` section of the docstring:

https://github.com/sktime/sktime/blob/f5399cd5c6e9839661526d1983f0a223b5da23ec/sktime/param_est/seasonality.py#L58-L85

A way to solve will be to remove the empty line and make the headings into comments.

yarnabrina:
Fixes #5013.

Adds empty lines to docstrings examples missing new lines before or after comments in examples sections.

### DIFF
--- a/sktime/base/_serialize.py
+++ b/sktime/base/_serialize.py
@@ -35,6 +35,7 @@ def load(serial):
     Examples
     --------
     Example 1: saving an estimator as pickle and loading
+
     >>> from sktime.datasets import load_airline
     >>> from sktime.forecasting.naive import NaiveForecaster
     >>>
@@ -55,6 +56,7 @@ def load(serial):
     >>> y_pred = forecaster_loaded.predict()
 
     Example 2: saving a deep learning estimator on the hard drive and loading
+
     >>> import numpy as np
     >>> from sktime.classification.deep_learning import CNNClassifier
     >>>
--- a/sktime/classification/compose/_pipeline.py
+++ b/sktime/classification/compose/_pipeline.py
@@ -87,6 +87,7 @@ class ClassifierPipeline(_HeterogenousMetaEstimator, BaseClassifier):
     >>> y_pred = pipeline.predict(X_test)
 
     Alternative construction via dunder method:
+
     >>> pipeline = PCATransformer() * TimeSeriesForestClassifier(n_estimators=5)
     """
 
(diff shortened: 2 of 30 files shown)


fkiraly (approved):
Thanks! Looks like hard work!

hazrulakmal (changes requested):
**Suggestion**: Should we make the separation sentence, for example `Example 3: using the dunder method` consistent throughout all docstring examples? some are indented, which make them appear as highlighted object while other do not. 

this is how indented seperation sentence looks like 
image

while this is when it didnt
image

Thanks for the hardwork, @yarnabrina! wondering did you do this automatically? or go one by one?

yarnabrina:
> Should we make the separation sentence, for example Example 3: using the dunder method consistent throughout all docstring examples?

I did notice this, and few other issues. For examples few examples don't contain `>>>` or `...`, few have inconsistent headers, etc. But since these were not strictly affecting the issue I mentioned in #5013, I skipped all of those and focused only on those.

> wondering did you do this automatically? or go one by one?

Both. Filtered by regex lines not starting with `>>>` or `...` under `Examples` section before `"""` and then one-by-one.

hazrulakmal:
just wondering, should we add [numpydoc validation precommit hook](https://numpydoc.readthedocs.io/en/latest/validation.html) into `.pre-commit-config.yaml` file? do you happen to know if it will help validating the docstring before making commit?

this is a different issue but I also found a weird display for [`evaluate` function](https://www.sktime.net/en/stable/api_reference/auto_generated/sktime.forecasting.model_evaluation.evaluate.html) on the API document. could adding the precomit hook help solving this issue?

yarnabrina:
We did discuss about adding using these validations in https://github.com/sktime/sktime/issues/4803, but did you try to use it locally? I've a bad (even though completely untested) feeling that it'll create a huge list of issues and warnings by trying to check for every single ideal feature in every single docstring.

But my guess (again, completely untested) is this issue won't be auto-fixed. Whether user wants to split examples or not is difficult to detect programmatically. But if it works for you locally, we can definitely add it.

---

By the way, do you still want me to address non-5013 issues in this PR? I am not very enthusiastic, mainly because I don't know what regex to search filtering for in this case.

PR #5084, opened 2023-08-12
[ENH] dependency and diff test switch for individual estimators to decorate non-suite tests

fkiraly:
This PR introduces a test switch for individual estimators that can be used to decorate non-suite pytest tests.

The test switch reads the `pytest` parameters, and if used in a `skipif` decorator will run the test iff all soft dependencies are present, and if the estimator module has changed (if the `ONLY_CHANGED_MODULES` switch is on).

This should allow us to minimize test time, by gradually removing all estimator specific costly tests using this decorator.

As a proof of concept, this PR decorates the most costly non-suite tests in the code base (mostly various heavy time series classifier ensembles and such) to be executed only upon module change.

Includes the following changes:

* also checks out `main` on the `test-nosoftdeps-full` CI element so it can compute a non-trivial diff to `main`
* moves the runtime intensive `ConformalIntervals` doctest into separate `pytest` tests to allow it to be decorated
* adds missing `"numba"` dependency tag to the HIVE-COTE classifiers

As both the module of `ConformalIntervals` and the HIVE-COTE classifiers change (see above), we can consider these test cases, as we would expect the specific non-suite tests for the respective estimators to run in this CI element, while we would not expect the other decorated tests to run.


sktime/forecasting/tests/test_conformal.py
yarnabrina, line 1: Skipped this file for review, not much familiarity with this based on #5089.

sktime/tests/test_switch.py
```diff
@@ -0,0 +1,43 @@
...
+    -------
+    bool : True if class should be tested, False otherwise
+        if cls was a list, is True iff True for at least one of the classes in the list
+    """
+    if isinstance(cls, list):
+        return any(run_test_for_class(x) for x in cls)
+
+    from sktime.tests.test_all_estimators import ONLY_CHANGED_MODULES
+    from sktime.utils.git_diff import is_class_changed
+    from sktime.utils.validation._dependencies import _check_estimator_deps
```
yarnabrina, line 33: Can these not be at module level?

If I am not wrong, these don't depend on any soft dependencies, and top level imports are usually recommended by linters, style guides, etc.

sktime/tests/test_switch.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def run_test_for_class(cls):
+    """Check if test should run for a class.
+
+    This checks the following conditions:
+
+    * if required dependencies are not present, does not run the test
+    * if ONLY_CHANGED_MODULES setting is on, runs the test
+      if and only if the class has changed according ot is_class_changed
```
yarnabrina, line 14:
```suggestion
      if and only if the module containing the class has changed according to is_class_changed
```

Minor text change suggestion. Unless this is intentional with future plan of improving `is_class_changed` to actually detect change in a single class?

yarnabrina (dismissed):
Added 2 minor non-blocking suggestions, and skipped conformal testing.

sktime/tests/test_switch.py
fkiraly: no, it is meant to refer to the module, indeed.

In the future, I was thinking to optionally retrieve parent classes, i.e., test all classifiers if we change the `BaseClassifier`.

sktime/forecasting/tests/test_conformal.py
fkiraly: this is just the former doctests from `ConformalIntervals` moved into `pytest` tests

sktime/tests/test_switch.py
fkiraly: what do you mean by "these" - the `validation._dependencies`?

Yes, I suppose so, but this is "how it has always been".
In `skbase` it's moved to the top already, and in `sktime` it would be a fiddly refactor.

yarnabrina: I just meant to move the 3 import lines from inside the function definition to between module docstring and `__author__`. If it's complicated or likely to have side-effects, please ignore.

fkiraly: yes, worried about side effects in the module. Probably over-worrying though.

PR #5086, opened 2023-08-12
[ENH] tests and fixes for `numpy` weights in performance metrics

fkiraly:
This PR:

* adds systematic testing for using `numpy` weights in `multioutput`
* fixes a post-deprecation breakage where comparisons of str to `np.ndarray` raise an exception
* adds the ability of using `numpy` weights in probabilistic metrics which previously was not possible, apparently

Depends on https://github.com/sktime/sktime/pull/5097 to avoid failures due to the number of test parameters.


fkiraly:
The failures are all due to the `get_test_params` check to have more than 1 parameter setting. This has been activated with the incremental testing - it is not in the scope of this PR, or changed by it, though.

yarnabrina (approved):

PR #5087, opened 2023-08-12
[ENH] speed up `check_estimator` tests by using simpler forecaster

fkiraly:
This PR speeds up the `check_estimator` tests by using a purpose-built, simplified `NaiveForecaster` which much less settings, using only the `"last"` strategy.


fkiraly:
ok, for some reason this seems much worse.

PR #5097, opened 2023-08-14
[ENH] metrics classes - add testing parameters

fkiraly:
This PR adds a second set of test parameters to those metrics classes that did not have these, but had non-reserved parameters that could be set to at least two values.


yarnabrina (approved):
Just confirming, this is to handle errors in #5086 right?

fkiraly:
> Just confirming, this is to handle errors in https://github.com/sktime/sktime/pull/5086 right?

Yes.

PR #5098, opened 2023-08-14
[ENH] speed up tests in `test_fh`

fkiraly:
This PR speeds up tests in `test_fh` by:

* making tests relying on costly estimators conditional on `run_test_for_class`
* adding copies of the same tests using `NaiveForecaster` to ensure coverage in unconditional runs

### DIFF
--- a/sktime/forecasting/base/tests/test_fh.py
+++ b/sktime/forecasting/base/tests/test_fh.py
@@ -638,7 +638,7 @@ def test_auto_arima_case_with_naive():
 
     fh_ = ForecastingHorizon(X.index[5:], is_relative=False)
 
-    a_clf = NaiveForecaster(start_p=2, start_q=2, max_p=5, max_q=5)
+    a_clf = NaiveForecaster()
     clf = a_clf.fit(X=X[:5], y=y[:5])
     y_pred_sk = clf.predict(fh=fh_, X=X[5:])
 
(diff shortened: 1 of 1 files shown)


sktime/forecasting/base/tests/test_fh.py
```diff
@@ -578,6 +621,50 @@ def test_auto_arima():
     )
 
 
+def test_auto_arima_case_with_naive():
```
yarnabrina, line 624: I think this also needs the `pmdarima` soft dependency check. The same decorator from `test_auto_arima` is probably enough.

fkiraly:
> Made a comment with respect to the failing test.

> I am not sure I understand these tests: `test_<forecaster>_case_with_naive`. Few (most actually) don't even use the `<forecaster>`, and `auto_arima` one uses both `<forecaster>` and `naive`. Apologies if it's obvious...but what are these supposed to test?

To offer an explanation - this is to the best of my knowledge as I am not the author of these tests.

I think these are regression tests for failure cases, specific to logic in the `BaseForecaster` around the handling of `ForecastingHorizon`.

As such, what is being tested should not depend on the forecaster in there, but to me it appears the person who added these simple used the forecaster from the bug reports.

Therefore, the regression tests should cover the same logic and regression case if the forecaster is replaced with `NaiveForecaster` (which will be much faster than, say, `AutoARIMA` or `AutoETS`).

Now, since I'm also slightly confused by the existence and form of the tests, I had the suspicion that they could still be specific to the forecasters in there - perhaps they also test a case of non-compliance in the adapter or similar.

To also account for that possibility and not reduce test coverage, I also left the test in the original, but made sure the original would only be executed if the specific forecaster is changed.

sktime/forecasting/base/tests/test_fh.py
fkiraly: fixed, the undecorated test should be with the `NaiveForecaster`, see below.

PR #5104, opened 2023-08-15
[ENH] tests and fixes for `numpy` weights in performance metrics - probabilistic metrics

fkiraly:
Continuation-fix of https://github.com/sktime/sktime/pull/5086 for probabilistic metrics.

This PR:

* fixes the ability of using `numpy` weights in probabilistic metrics - this was added in #5086 but not tested
* adds test for weighted probabilistic metrics. This is a generalized version of the doctest in https://github.com/sktime/sktime/pull/5068


fkiraly:
bugfix to https://github.com/sktime/sktime/pull/5086 which made it on `main` - quick merge to repair #5086

PR #5110, opened 2023-08-16
[BUG] Fix `vectorize_est` returning jumbled rows for row vectorization, `pd.DataFrame` return, if row names were not lexicographically ordered

### LINKED ISSUE #5108: [BUG] forecasters broadcast predictions on wrong instances when label as str
**Describe the bug**

When a panel datatype is given where the instances are labelled as str such that a random name followed by some int, e.g *user1*, *series1* , *T1* etc. the prediction that was made on the instances sometimes gets switched. below is a minimal example to show the problem. 

I suppose this occurs for all forecasters, albeit the test I did wasn't exhaustive, only on `TrendForecaster`, `NaiveForecaster` and `ExponentialSmothing`. I haven't looked at the logic of how forecasters made predictions for hierarchical or panel datatypes but I suspect they were all inherited from the same logic when making predictions for each single series. 

 

**To Reproduce**

```python
import pandas as pd
from sktime.forecasting.model_selection import ExpandingGreedySplitter

n_instances=3
n_points=5

temp = pd.DataFrame({
    'y': [i for i in range(n_points*n_instances)],
    'id':["T1"]*n_points + ["T2"]*n_points+ ["T11"]*n_points,  
    # bug still occurs even when timestamp is int-like
    'timestamp': pd.date_range(start='2022-07-01', periods=n_points*n_instances, freq='D') 
}).set_index(['id', 'timestamp'])

fh = [1,2]
train_test_splitter = ExpandingGreedySplitter(test_size=len(fh), folds=1)
forecaster = NaiveForecaster(strategy="last")  # NaiveForecaster() or TrendForecaster()
for train, test in train_test_splitter.split_series(temp): 
    forecaster.fit(train)
    y_pred = forecaster.predict(fh)
    print("=====TEST=====")
    print(test.head(15))
    print("=====PREDICT=====")
    print(y_pred.head(15))
    print("=====TRAIN=====")
    print(train.head(15))
```

**Output** - predictions for T2 and T11 got **switched** as shown in the output below
```
=====TEST=====
                 y
id  timestamp     
T1  2022-07-04   3
    2022-07-05   4
T2  2022-07-09   8
    2022-07-10   9
T11 2022-07-14  13
    2022-07-15  14
=====PREDICT=====
                   y
id  timestamp       
T1  2022-07-04   2.0
    2022-07-05   2.0
T2  2022-07-14  12.0  
    2022-07-15  12.0
T11 2022-07-09   7.0 
    2022-07-10   7.0 
=====TRAIN=====
                 y
id  timestamp     
T1  2022-07-01   0
    2022-07-02   1
    2022-07-03   2
T2  2022-07-06   5
    2022-07-07   6
    2022-07-08   7
T11 2022-07-11  10
    2022-07-12  11
    2022-07-13  12
```

**Expected behavior**

Trivially, the predictions should be broadcasted to the corresponding instance based on the trained data.

**Additional context**

This bug is spotted when trying to fit naive forecaster in M4 weekly dataset as per in this notebook #5051

**Versions**
<details>

sktime: 0.21.0
sklearn: 1.3.0
skbase: 0.4.6
numpy: 1.24.3
pandas: 2.0.3
</details>

@fkiraly @benHeid

fkiraly:
Fixes #5108.

Similar to the fix in #4684, although `pandas` needs to be prevented from sorting columns in two places:

* in the `groupby` call in `items`
* in/after the `pivot` call in `vectorize_est`, same problem as #4684.

Adds a compact version of the failure case reported in https://github.com/sktime/sktime/issues/5108 as a test.

Further, removed the `statsmodels` dependency in the related column test by replacing a `statsmodels` dataset with an equivalent, randomly generated one (also with string, non-ordered columns).

### DIFF
--- a/sktime/datatypes/_vectorize.py
+++ b/sktime/datatypes/_vectorize.py
@@ -261,7 +261,7 @@ class VectorizedDF:
         if is_self_iter:
             yield from _iter_cols(self.X_multiindex)
         else:
-            for name, group in self.X_multiindex.groupby(level=iter_levels):
+            for name, group in self.X_multiindex.groupby(level=iter_levels, sort=False):
                 yield from _iter_cols(group.droplevel(iter_levels), group_name=name)
 
     def _iter_levels(self, iterate_as):
@@ -593,11 +593,14 @@ class VectorizedDF:
         if return_type == "pd.DataFrame":
             df_long = pd.DataFrame(ret)
             cols_right_order = df_long.loc[:, 1].unique()
+            rows_right_order = df_long.loc[:, 0].unique()
 
             df = df_long.pivot(index=0, columns=1, values=2)
-            # DataFrame.pivot sorts the columns (is this a bug? see #4683)
+            # DataFrame.pivot sorts the rows & columns
+            # (is this a bug? see #4683 and #5108)
             # either way, we need to fix this:
             df = df.reindex(cols_right_order, axis=1)
+            df = df.reindex(rows_right_order, axis=0)
 
             # remove "0" and "1" from index/columns name
             df.index.names = [None] * len(df.index.names)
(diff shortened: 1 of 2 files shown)


fkiraly:
consequential bug, merging to include in 0.21.1

PR #5112, opened 2023-08-16
[ENH] make `statsforecast` adapter compatible with optional `predict` `level` arguments, and different init param sets

### LINKED ISSUE #5024: [BUG] StatsForecastAutoCES: predict() got an unexpected keyword argument 'level'
**Describe the bug**

StatsForecastAutoCES model complains about `TypeError: predict_in_sample() got an unexpected keyword argument 'level'` when `.predict()` is called after `.fit()`

**To Reproduce**
```python
from sktime.forecasting.statsforecast import StatsForecastAutoCES

y = load_airline()

forecaster = StatsForecastAutoCES(season_length=12)
forecaster.fit(y)

forecaster.predict(fh=[1,2,3])

```

**Additional context**

```
  File "/lib/python3.9/site-packages/IPython/core/interactiveshell.py", line 3508, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-17-4d83c1149447>", line 1, in <module>
    forecaster.predict(fh=[1,2,3])
  File "/lib/python3.9/site-packages/sktime/forecasting/base/_base.py", line 404, in predict
    y_pred = self._predict(fh=fh, X=X_inner)
  File "/lib/python3.9/site-packages/sktime/forecasting/base/adapters/_generalised_statsforecast.py", line 192, in _predict
    out_of_sample_point_predictions = self._predict_in_or_out_of_sample(
  File "/lib/python3.9/site-packages/sktime/forecasting/base/adapters/_generalised_statsforecast.py", line 92, in _predict_in_or_out_of_sample
    predictions = self._forecaster.predict(
TypeError: predict() got an unexpected keyword argument 'level'
```

**Versions**
statsforecast 1.4.0 version
M1 Mac chip.

@yarnabrina

arnaujc91:
#### Reference Issues/PRs
Fixes #5024 and #5207:

* Versions previous to v1.5.0 of `statsforecast` do not have the `level`
argument in the `predict` or `predict_in_sample` methods of the
forecasters.
* `statsforecast` removed the `parallel` argument in some estimator since v1.6.0

Details:

Now before calling `predict` or `predict_in_sample`  we make sure the methods contain the `level` argument otherwise we do not pass it.

To handle the `level` argument we will set tags dynamically at instantiation time to see if they are there in the methods `predict_in_sample` and `predict`. 

Also newer versions of `statsforecast` removed parameters that were available in previous versions. We will take care of it by throwing a warning if the user passes a keyword argument that is not supported by the current version of `statsforecast` that they might be using. We will throw a warning if the user passes the keyword argument which is non-existent in its current version of `statsforecast`.

If a parameter is set by default and NOT passed by the user, the parameter will also be ignored but NO warning will be thrown as the user did not intend to use that parameter in the first place, so he should not care.

No

Make sure it will work for any version of `stasforecast`

No

##### For all contributions

### DIFF
--- a/sktime/forecasting/arch.py
+++ b/sktime/forecasting/arch.py
@@ -48,14 +48,13 @@ class StatsForecastGARCH(_GeneralisedStatsForecastAdapter):
 
         super().__init__()
 
-    def _instantiate_model(self):
-        # import inside method to avoid hard dependency
+    def _get_statsforecast_class(self):
         from statsforecast.models import GARCH as _GARCH
 
-        return _GARCH(
-            p=self.p,
-            q=self.q,
-        )
+        return _GARCH
+
+    def _get_statsforecast_params(self):
+        return {"p": self.p, "q": self.q}
 
     @classmethod
     def get_test_params(cls, parameter_set="default"):
@@ -110,13 +109,13 @@ class StatsForecastARCH(_GeneralisedStatsForecastAdapter):
 
         super().__init__()
 
-    def _instantiate_model(self):
-        # import inside method to avoid hard dependency
+    def _get_statsforecast_class(self):
         from statsforecast.models import ARCH as _ARCH
 
-        return _ARCH(
-            p=self.p,
-        )
+        return _ARCH
+
+    def _get_statsforecast_params(self):
+        return {"p": self.p}
 
     @classmethod
     def get_test_params(cls, parameter_set="default"):
--- a/sktime/forecasting/base/adapters/_generalised_statsforecast.py
+++ b/sktime/forecasting/base/adapters/_generalised_statsforecast.py
@@ -28,6 +28,14 @@ class _GeneralisedStatsForecastAdapter(BaseForecaster):
     def __init__(self):
         super().__init__()
 
+        forecaster = self._instantiate_model()
+
+        if "level" not in signature(forecaster.predict_in_sample).parameters.keys():
+            self.set_tags(**{"capability:pred_int": False})
+
+        if "level" not in signature(forecaster.predict).parameters.keys():
+            self.set_tags(**{"capability:pred_int:insample": False})
+
         self._forecaster = None
 
     def _instantiate_model(self):
--- a/sktime/forecasting/base/adapters/_generalised_statsforecast.py
+++ b/sktime/forecasting/base/adapters/_generalised_statsforecast.py
@@ -42,6 +42,9 @@ class _GeneralisedStatsForecastAdapter(BaseForecaster):
     def _get_statsforecast_class(self):
         raise NotImplementedError("abstract method")
 
+    def _get_statsforecast_params(self):
+        raise NotImplementedError("abstract method")
+
     def _get_init_statsforecast_params(self):
         statsforecast_class = self._get_statsforecast_class()
         return list(signature(statsforecast_class.__init__).parameters.keys())
@@ -63,7 +66,9 @@ class _GeneralisedStatsForecastAdapter(BaseForecaster):
         return sktime_params
 
     def _instantiate_model(self):
-        raise NotImplementedError("abstract method")
+        cls = self._get_statsforecast_class()
+        params = self._validate_init_params(**self._get_statsforecast_params())
+        return cls(**params)
 
     def _fit(self, y, X, fh):
         """Fit forecaster to training data.
--- a/sktime/forecasting/base/adapters/_generalised_statsforecast.py
+++ b/sktime/forecasting/base/adapters/_generalised_statsforecast.py
@@ -30,7 +30,15 @@ class _GeneralisedStatsForecastAdapter(BaseForecaster):
         super().__init__()
 
         self._forecaster = self._instantiate_model()
-        self._check_supports_pred_int()
+        (
+            self._support_pred_int_in_sample,
+            self._support_pred_int,
+        ) = self._check_supports_pred_int()
+
+        if not self._support_pred_int_in_sample:
+            self.set_tags(**{"capability:pred_int:insample": False})
+        if not self._support_pred_int:
+            self.set_tags(**{"capability:pred_int": False})
 
     def _instantiate_model(self):
         raise NotImplementedError("abstract method")
@@ -284,22 +292,28 @@ class _GeneralisedStatsForecastAdapter(BaseForecaster):
 
         Check if `level` argument is available in `predict_in_sample` and
         `predict` methods from the `statsforecast` forecaster.
-        If that is not the case it will set the corresponding tags
-        (`capability:pred_int:insample`,`capability:pred_int`) to False as well as
-         the corresponding class attributes `_support_pred_int_in_sample`
-         and `_support_pred_int`.
+        A tuple of booleans (`support_pred_int_in_sample`, `support_pred_int`)
+        is returned where the user is informed which of the two, if any,
+        support interval predictions.
 
          Furthermore, will throw a warning to let the user know that he should consider
          upgrading statsforecast versio as both or one of the two methods might
          not be able to produce confidence intervals.
 
+        Returns
+        -------
+        Tuple of bool
+            A tuple containing two boolean values:
+            - `support_pred_int_in_sample`: True if prediction intervals are supported
+              in `predict_in_sample`, False otherwise.
+            - `support_pred_int`: True if prediction intervals are supported
+              in `predict`, False otherwise.
         """
         if (
             "level"
             not in signature(self._forecaster.predict_in_sample).parameters.keys()
         ):
-            self.set_tags(**{"capability:pred_int:insample": False})
-            self._support_pred_int_in_sample = False
+            support_pred_int_in_sample = False
             import statsforecast
 
             warn(
@@ -309,11 +323,10 @@ class _GeneralisedStatsForecastAdapter(BaseForecaster):
                 f"Consider upgrading to a newer version."
             )
         else:
-            self._support_pred_int_in_sample = True
+            support_pred_int_in_sample = True
 
         if "level" not in signature(self._forecaster.predict).parameters.keys():
-            self.set_tags(**{"capability:pred_int": False})
-            self._support_pred_int = False
+            support_pred_int = False
             import statsforecast
 
             warn(
@@ -323,7 +336,9 @@ class _GeneralisedStatsForecastAdapter(BaseForecaster):
                 f"Consider upgrading to a newer version."
             )
         else:
-            self._support_pred_int = True
+            support_pred_int = True
+
+        return support_pred_int_in_sample, support_pred_int
 
 
 class StatsForecastBackAdapter:
--- a/sktime/forecasting/base/adapters/_generalised_statsforecast.py
+++ b/sktime/forecasting/base/adapters/_generalised_statsforecast.py
@@ -1,6 +1,7 @@
 # copyright: sktime developers, BSD-3-Clause License (see LICENSE file)
 """Implements adapter for StatsForecast models."""
 from inspect import signature
+from warnings import warn
 
 import pandas
 
@@ -28,15 +29,8 @@ class _GeneralisedStatsForecastAdapter(BaseForecaster):
     def __init__(self):
         super().__init__()
 
-        forecaster = self._instantiate_model()
-
-        if "level" not in signature(forecaster.predict_in_sample).parameters.keys():
-            self.set_tags(**{"capability:pred_int:insample": False})
-
-        if "level" not in signature(forecaster.predict).parameters.keys():
-            self.set_tags(**{"capability:pred_int": False})
-
-        self._forecaster = None
+        self._forecaster = self._instantiate_model()
+        self._check_supports_pred_int()
 
     def _instantiate_model(self):
         raise NotImplementedError("abstract method")
@@ -73,8 +67,6 @@ class _GeneralisedStatsForecastAdapter(BaseForecaster):
         """
         del fh  # avoid being detected as unused by ``vulture`` like tools
 
-        self._forecaster = self._instantiate_model()
-
         y_fit_input = y.to_numpy(copy=False)
 
         X_fit_input = X
@@ -100,9 +92,7 @@ class _GeneralisedStatsForecastAdapter(BaseForecaster):
             # Before v1.5.0 (from statsforecast) not all foreasters
             # have a "level" keyword argument in `predict_in_sample`
             level_kw = (
-                {"level": level_arguments}
-                if ("level" in signature(predict_method).parameters.keys())
-                else {}
+                {"level": level_arguments} if self._support_pred_int_in_sample else {}
             )
             predictions = predict_method(**level_kw)
             point_predictions = predictions["fitted"]
@@ -110,11 +100,7 @@ class _GeneralisedStatsForecastAdapter(BaseForecaster):
             predict_method = self._forecaster.predict
             # Before v1.5.0 (from statsforecast) not all foreasters
             # have a "level" keyword argument in `predict`
-            level_kw = (
-                {"level": level_arguments}
-                if ("level" in signature(predict_method).parameters.keys())
-                else {}
-            )
+            level_kw = {"level": level_arguments} if self._support_pred_int else {}
             predictions = predict_method(maximum_forecast_horizon, X=X, **level_kw)
             point_predictions = predictions["mean"]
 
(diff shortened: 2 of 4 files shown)


fkiraly (dismissed):
Looks good to me!

I'll wait for @yarnabrina though. One question that came up, how should we test this, given that all estimators at max version do have the `level` arg.

yarnabrina (changes requested):
Thanks for your contribution.

Based on first look, I don't expect it to work on older versions.

You have adjusted to pass `level` argument only when the method supports it. No problem so far, but in that case the response will not have other columns but `mean`, if I'm not wrong. The rest of the code handling probabilistic predictions is going to fail, and my guess is it'll be a `KeyError`.

Is it working for you in an environment with `statsforecast` version `1.4.0`, if you use `StatsForecastAutoCES`, same as your original bug report?

---

Some other changes are also required. For examples estimators lacking `level` argument should have `False` for the tag indicating probabilistic capabilities etc.

arnaujc91:
Hi @yarnabrina, I tried with 1.4.0 and it all worked fine. I can give you an output of both methods:

```python

self._forecaster.predict_in_sample
level_kw = {}


{'fitted': array([ 93.27810669, 109.55554199, 130.16934967, 124.91722107,
       116.33216858, 136.15605927, 152.71878815, 149.94748688,
       135.57047272, 117.08627319, 105.00260925, 132.33216858, ... ])}

self._forecaster.predict
level_kw = {}

{'mean': array([453.03417969, 429.34039307, 488.64471436])}

```

arnaujc91:
> Some other changes are also required. For examples estimators lacking level argument should have False for the tag indicating probabilistic capabilities etc

I will have to look at what to do there, any ideas?

fkiraly:
> I will have to look at what to do there, any ideas?

The mechanics of setting the tag dynamically would have to happen in `__init__`, after the `super().__init__` call, as `self.set_tags(**{"tagname": tagvalue})`. The tags would be `"capability:pred_int"`, `"capability:pred_int:insample", and, possibly, `"capability:insample"`.

yarnabrina:
Hi @arnaujc91, I'm not following how it's working for you in 1.4.0 as well. Your results show only `fitted` and `mean` keys in the direct results from `statsforecast`, right?

What happens in lines 142 and 148?

https://github.com/sktime/sktime/blob/c06fae6475905bff673907defb0c5c42884405f4/sktime/forecasting/base/adapters/_generalised_statsforecast.py#L142

https://github.com/sktime/sktime/blob/c06fae6475905bff673907defb0c5c42884405f4/sktime/forecasting/base/adapters/_generalised_statsforecast.py#L148

Is there any change that will ensure that these lines will not be reached if `level` argument is missing? Until you handle the tag issue, I'm not following how it's working. May be I'm missing something obvious, can you please point it out?

arnaujc91:
Hi @yarnabrina, the thing the code that i posted in #5024 


> ```python
> from sktime.forecasting.statsforecast import StatsForecastAutoCES
> from sktime.datasets import load_airline
> 
> y = load_airline()
> 
> forecaster = StatsForecastAutoCES(season_length=12)
> forecaster.fit(y)
> 
> forecaster.predict(fh=[1,2,3])
> ```
> 
> **Additional context**
> 
> ```
> Traceback (most recent call last):
>   File "/lib/python3.9/site-packages/IPython/core/interactiveshell.py", line 3508, in run_code
>     exec(code_obj, self.user_global_ns, self.user_ns)
>   File "<ipython-input-17-4d83c1149447>", line 1, in <module>
>     forecaster.predict(fh=[1,2,3])
>   File "/lib/python3.9/site-packages/sktime/forecasting/base/_base.py", line 404, in predict
>     y_pred = self._predict(fh=fh, X=X_inner)
>   File "/lib/python3.9/site-packages/sktime/forecasting/base/adapters/_generalised_statsforecast.py", line 192, in _predict
>     out_of_sample_point_predictions = self._predict_in_or_out_of_sample(
>   File "/lib/python3.9/site-packages/sktime/forecasting/base/adapters/_generalised_statsforecast.py", line 92, in _predict_in_or_out_of_sample
>     predictions = self._forecaster.predict(
> TypeError: predict() got an unexpected keyword argument 'level'
> ```


apparently does not include `levels` so now the code is not breaking but is also not reaching the lines of code you posted, because when `levels` is `None` you get this:

https://github.com/sktime/sktime/blob/c06fae6475905bff673907defb0c5c42884405f4/sktime/forecasting/base/adapters/_generalised_statsforecast.py#L120
https://github.com/sktime/sktime/blob/c06fae6475905bff673907defb0c5c42884405f4/sktime/forecasting/base/adapters/_generalised_statsforecast.py#L121

Therefore it did not break. So if i find a place where you pass `levels` (like using some confidence interval somewhere), i guess `predict_intervals` or something similar, i assume there it will break. What about i propose the following fix:

```python
if "level" in signature(predict_method).parameters.keys():
    level_kw = {"level": level_arguments}
else:
    level_kw = {}
    levels = None
```

Probably i need some Warning logs here to tell the user `levels` is not available.

yarnabrina:
Hi @arnaujc91, let me tell you the way I was thinking.

As @fkiraly pointed out above, you need to add those tag handling. So,
after you do that, user can not use predict_interval or predict_quantile
methods, since they will receive a not implemented error from base class.

My idea is that it's enough, and you don't need anything else. Once you add
it, since the private probabilistic variants will never be called, so
there's no need of separate handling. And as you pointed out,
non-probabilistic variant will always explicitly specify None so we should
be good as well. When users call predict, they are not looking for
probabilistic forecasts so I think throwing warning may not be necessary.

So, in summary, my suggestion is you implement the tag handling, and that
should lead to a working code. This is what I tried to indicate by "until
you handle the tag issue", but may be that was too obscure? Anyways, please
let me know what you think of this approach.

fkiraly:
yes, I would agree with @yarnabrina's suggestion let's have the base class handle exceptions raised if proba forecasts are not available

arnaujc91:
Ok, one comment here. So I have seen these tags:

```python
# AutoETS class

    _tags = {
        "ignores-exogeneous-X": True,
        "capability:pred_int": True,
        "capability:pred_int:insample": True,
    }
```

and i can interpret that it means that `predict_intervals` is available in that case. The thing is the following, this tag should be dependent on the version of `statsforecast` that the user is using, i.e. depending on the version they are using `predict_intervals` is or is not available. If you want to have the tag at the class level, this is of course independent of the version (of `statsforecast`) the user is using, so `predict_intervals` will not work for any user regardless of his version of `statsforecast` if the Tag is `False`. So we should agree at which `statsforecast` version do we start. The older the version, the more models do not have the `levels` argument. 

Am i missing something?

fkiraly:
> Am i missing something?

Yes, I think you missed my comment above here?
https://github.com/sktime/sktime/pull/5112#issuecomment-1682072801

You can set the tags dynamically in `__init__`, e.g., after a version check of `statsforecast` (which you could do with `_check_soft_dependencies`, or direct version number read) - see my comment.

For that, you need to use the method `set_tag` as described, and leave the class tag as it is.

arnaujc91:
@fkiraly , @yarnabrina , updated the PR. Let me know :)

fkiraly (approved):
Nice!

Minor request, not blocking: I would make the call to `_check_supports_pred_int` pure, so a developer understands at line 33 what is happening. That is, ensure that `_check_supports_pred_int` neither reads nor writes `self` attributes, so line 33 would be:
`can_pred_int, can_pred_int_ins = self._check_supports_pred_int(self._forecaster)`,
then have the attr and `set_tag` directly after.

This might seem more clunky, but imo it is more transparent since dynamic tag sets an init attribute sets are usually visible in the `__init__` directly.

Blocking: I would like to wait for a review by @yarnabrina and make my approval conditional on his.

fkiraly (changes requested):
Tests should pass, of course.

sktime/forecasting/base/adapters/_generalised_statsforecast.py
```diff
@@ -26,7 +29,16 @@ class _GeneralisedStatsForecastAdapter(BaseForecaster):
     def __init__(self):
         super().__init__()
 
-        self._forecaster = None
+        self._forecaster = self._instantiate_model()
```
yarnabrina, line 32: I'll prefer if you avoid this. We are initialising the model during fit for all concrete estimators now, and I'd prefer to keep it that way. Here's my suggestion:

1. create a new property storing the model class in all concrete estimators as a string, e.g. "AutoARIMA", "GARCH", etc.
2. in both of the new methods, check if "level" is in `predict` or `predict_in_sample` of `getattr(statsforeast.models, self.model_class)` or not, without initiating an estimator of that class

This is the line failing in tests for MSTL class. I believe not instantiating the estimator will solve that as well, can you confirm @luca-miniati?

sktime/forecasting/base/adapters/_generalised_statsforecast.py
```diff
@@ -26,7 +29,16 @@ class _GeneralisedStatsForecastAdapter(BaseForecaster):
...
+        (
+            self._support_pred_int_in_sample,
+            self._support_pred_int,
+        ) = self._check_supports_pred_int()
+
+        if not self._support_pred_int_in_sample:
+            self.set_tags(**{"capability:pred_int:insample": False})
+        if not self._support_pred_int:
+            self.set_tags(**{"capability:pred_int": False})
```
yarnabrina, line 41: As of now, these tags are being passed in concrete classes. If you handle these here, totally fine, but may be we can remove from there? And we can always (both True/False cases) set tags as `self.set_tags(**{"capability:pred_int:insample": self._support_pred_int_in_sample})` or similarly I guess.

sktime/forecasting/base/adapters/_generalised_statsforecast.py
```diff
@@ -86,12 +96,20 @@ def _predict_in_or_out_of_sample(
         level_arguments = None if levels is None else [100 * level for level in levels]
 
         if fh_type == "in-sample":
-            predictions = self._forecaster.predict_in_sample(level=level_arguments)
+            predict_method = self._forecaster.predict_in_sample
```
yarnabrina, line 99: I believe you also raised the point that this method itself may be absent for older versions? We can handle it as either a specific lower bound, or via another tag check. As earlier, my first vote will still be to use `1.5.0` as lower bound as that's when the generalised class and new models were supported, and I'm reasonably optimistic that Nixtla will ensure this in future.

But @fkiraly please share your opinion as well.

sktime/forecasting/base/adapters/_generalised_statsforecast.py
fkiraly: hm, this seems a bit clunky to me. May I suggest a variation on the idea which avoids a new property?

What I would do is, instead of a string which we have to resolve later, to deal with it similar in the other adapter, `_TslearnAdapter` - using a `_get_statsforecast_class` method, which returns the class, the output can then be used -possibly in another method `_supports_proba` - to inspect the method signatures.

From the extender contract perspective, changing a class import in `get_statsforecast_class` is just as much work as setting a string attribute, and it seems more flexible and transparent.

yarnabrina: Makes sense, and I suppose initiate forecaster can use this method as well to intiate the instance. It'd be consistent with `tslearn` adapter then.

(I remember a open `statsmodels` issue with `random_state`. Don't remember it very well, and can't find now, but will this sort of handling help it as well?)

sktime/forecasting/base/adapters/_generalised_statsforecast.py
fkiraly: hm, I usually err on the side of greater compatibility if it is not too much of a hassle.

How about: we have a checker function as suggested, `_supports_proba`, which returns two booleans. At the very top, we have a `hasattr` check for the method?

sktime/forecasting/base/adapters/_generalised_statsforecast.py
fkiraly: > Don't remember it very well, and can't find now, but will this sort of handling help it as well?

Here, there wasn't an issue, as it is subsumed under generic deprecation message handling:

https://github.com/sktime/sktime/pull/4623

I don't think the problem is solved, since in `statsmodels` the `random_state` arg can be passed to `kwargs`, which cannot be determined by a signature inspection.

luca-miniati: > This is the line failing in tests for MSTL class. I believe not instantiating the estimator will solve that as well, can you confirm @luca-miniati?

@yarnabrina I agree, init params in `StatsForecastMSTL` are set after the call to `_GeneralisedStatsForecastAdapter.__init__`. Should work if `_instantiate_model` is moved back to `fit`

arnaujc91:
@fkiraly , @yarnabrina , I think right now my PR is really ready for review. Let me know if you would change anything in the code.

Taken care of the `predict` argument and checking if any kwargs passed to the instantiator of the statsforecast model are not there, if that is the case they are removed and a warning is thrown to the user to check his statsforecast version (such that he can adjust the the API).

sktime/forecasting/arch.py
```diff
@@ -48,14 +48,13 @@ def __init__(
...
+    def _get_statsforecast_class(self):
         from statsforecast.models import GARCH as _GARCH
 
-        return _GARCH(
-            p=self.p,
-            q=self.q,
-        )
+        return _GARCH
+
+    def _get_statsforecast_params(self):
```
fkiraly, line 56: we could also make this abstract, and default to `return self.get_params()` - but leave the implementer an option to override.

sktime/forecasting/base/adapters/_generalised_statsforecast.py
```diff
@@ -27,10 +31,46 @@ def __init__(self):
...
+        return list(signature(statsforecast_class.__init__).parameters.keys())
+
+    def _validate_init_params(self, **sktime_params):
+        sktime_params = sktime_params.copy()
+        statsforecast_params = self._get_init_statsforecast_params()
+
+        for sktime_param in sktime_params.keys():
+            if sktime_param not in statsforecast_params:
+                sktime_params.pop(sktime_param)
+                warn(
```
fkiraly, line 60: I think the warning should be raised only if the user sets a non-default in `sktime_param`. You can check with via the method `self.get_param_defaults()` which every `skbase` object has.

sktime/forecasting/base/adapters/_generalised_statsforecast.py
```diff
@@ -27,10 +31,46 @@ def __init__(self):
...
+                    f" not found in the __init__ method "
+                    f"from {self._get_statsforecast_class()}. "
+                    f"Check your statsforecast version"
+                    f"to find out the right API parameters."
+                )
+        return sktime_params
+
+    def _instantiate_model(self):
+        cls = self._get_statsforecast_class()
+        params = self._validate_init_params(**self._get_statsforecast_params())
```
fkiraly, line 71: if we always call `_validate_init_params` with `**self._get_statsforecast_params()`, should it have an arg at all?

E.g., you could remove the arg, and rename the `_validate_init_params` to `_get_validated_statsforecast_params`

fkiraly (approved):
Some change requests above, but these are non-blocking.

In addition, can you ensure you update the PR title and description to match the scope before we merge?

* there are things added to set tags dynamically.
* the title and description only refers to the `level` arg
* we now also deal with the param mismatch issue, which should also be linked

sktime/forecasting/base/adapters/_generalised_statsforecast.py
arnaujc91: It can also be the case where we have a sktime_param that is not found in the installed version of statsforecast of the user. Not sure what you want to happen if that is the case. So my thought was to skip that parameter and inform the user that this is the case. As they keep changing the API it might be possible that the statsforecast version of the user is too new and maybe some parameters have disappeared eventhough they are in the documentation of sktime.

fkiraly: > Not sure what you want to happen if that is the case.

In that case, I would like to:

* warn the user if they pass a non-default
* do not warh the user if they pass a default (e.g., if they do not explicitly set the parameter)

Why: imagine the long-term issue where there is exactly one unique parameter per minor `statsforecast` version, 1.5.0, 1.6.0, 1.7.0, etc, which then disappears again.

In this case, with the current logic, any user would *always* see a warning about the parameters that are no longer even applicable. That seems excessive...

With the proposed logic, they would see warnings only if they actually type out a deprecated parameters (although not "if and only if")

sktime/forecasting/base/adapters/_generalised_statsforecast.py
arnaujc91: done

sktime/forecasting/arch.py
arnaujc91: done

fkiraly (approved):
LGTM now, and @yarnabrina closed his comments, so I assume we can merge for the release.

PR #5114, opened 2023-08-17
[MNT] 0.32.0 release action - revert temporary skip `get_test_params` number check for 0.21.1 and 0.22.0 release

fkiraly:
Reverts sktime/sktime#5113,

and adds all non-compliant estimators to a list of exceptions, which can be gradually reduced.


fkiraly:
Failures are unrelated - known timeouts etc.

PR #5116, opened 2023-08-17
[BUG] fix `numba` based aligners

fkiraly:
This PR fixes a bug in the `numba` based aligners which were detected by the release action.

The bug was unreported, and causes `transform` to break at call, due to the `X_inner_mtype` tag and the handling of `X` in `transform` being inconsistent. This has now been rectified, assuming `numpy3D` in both tag and `_transform` logic.

Odd that we did not catch this earlier.

FYI @yarnabrina, any idea why the differential testing did not catch this?

### DIFF
--- a/sktime/alignment/base.py
+++ b/sktime/alignment/base.py
@@ -88,7 +88,14 @@ class BaseAligner(BaseEstimator):
 
         self._fit(X=X_inner, Z=Z)
 
-        self._X = X_inner
+        # convert X to df-list for use in get_aligned, get_alignment_loc
+        if X_inner_mtype != "df-list":
+            self._X = convert(
+                X, from_type=X_mtype, to_type="df-list", as_scitype="Panel"
+            )
+        else:
+            self._X = X_inner
+
         self._Z = Z
 
         self._is_fitted = True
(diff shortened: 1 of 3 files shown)


yarnabrina:
I have no familiarity with aligner components (admitted I don't even know what do these do), so don't usually check these PR's. But based on number of tests, it looks like not all tests are running.

> why the differential testing did not catch this?

Can you please tell what are you expecting?

(By the way, the logs suggest an actual assert failure.)

fkiraly:
> Can you please tell what are you expecting?

I was expecting that the failures in the release action here: https://github.com/sktime/sktime/actions/runs/5889230714

would have appeared in at least one previous PR.

I have now been fixing the failure:
https://github.com/sktime/sktime/pull/5116

To me it looks like it was introduced as an interaction of multiple PR, which individually passed.

This highlights the following problem: if multiple PR are merged, and none of the PR is updated to the sum of changes, there will not be a CI element testing the sum of changes until the release - this is since on `main` the tests specific to the most recently merged change are not executed!

I think we need to think about how to change that - or, at least, enforce that all PR are up-to-date with main before merge.

yarnabrina:
> enforce that all PR are up-to-date with main before merge

Agreed. This is what we do for all our work PR's by using the Fast Forward merges only settings in Gitlab. Not sure if this option exists in Github or not.

Issue #5117, opened 2023-08-17
[BUG] readthedocs build failing with `AttributeError("'TocTree' object has no attribute 'get_toctree_ancestors'")`


fkiraly:
Readthedocs builds are erroring out with `sphinx.errors.ThemeError: An error happened in rendering the page about.
Reason: AttributeError("'TocTree' object has no attribute 'get_toctree_ancestors'")`, since yesterday, I think?

I can identify no immediate reason why they would.

This seems high priority as it prevents us from updating documentation.

fkiraly:
Might be due to the `sphinx 7.2.0` release 12 hours ago, so attempted fix adds an upper bound: https://github.com/sktime/sktime/pull/5118

fkiraly:
Update: that fixes it, though we should still understand what we need to do to make our doc build compatible with 7.2.0 and above.

yarnabrina:
This is definitely caused by the new Sphinx release: https://github.com/sphinx-doc/sphinx/issues/11608

There's a new release as well, going to try in a new MNT PR.

PR #5118, opened 2023-08-17
[MNT] restrict `sphinx < 7.2.0` to address build failure

fkiraly:
Temporary fix for #5117, which is likely caused by the 7.2.0 release 12 hours ago.


fkiraly:
Yes, this fixes it - temporarily.

PR #5125, opened 2023-08-18
[ENH] Interface statsmodels MSTL - transformer

luca-miniati:
Interfaces statsmodels MSTL, replaces #4970

#### For new estimators
- [x] I've added the estimator to the API reference - in docs/source/api_reference/taskname.rst, follow the pattern.
- [x]  I've added one or more illustrative usage examples to the docstring, in a pydocstyle compliant Examples section.
- [x]  If the estimator relies on a soft dependency, I've set the python_dependencies tag and ensured


docs/source/api_reference/forecasting.rst
```diff
@@ -158,6 +158,14 @@ Trend forecasters
 
     StatsForecastMSTL
 
+.. currentmodule:: sktime.forecasting.mstl
```
fkiraly, line 161: this should be in the transformations API reference

fkiraly (changes requested):
Nice, great!

A question, not necessarily a change request:

In the `statsmodels` example here
https://www.statsmodels.org/devel/generated/statsmodels.tsa.seasonal.MSTL.html

it seems that `MSTL` returns have as many seasonal components (named `Seasonal_[period]` in the plot) as there are entries to the iterable `periods`. Whereas, this transformer seems to add the seasonal components all up?

The result must allow to access these somehow, since the plots displays the components separately. Although there's no clear documentation in `statsmodesl` that I can find...

fkiraly:
Another comment, there is also this PR, seems same name?
https://github.com/sktime/sktime/pull/4970

It makes sense to have one PR with the transformer, and one (the other one?) with a forecaster - where the forecaster makes forecasts for the seasonal components and the trend component, with defaults being `NaiveForecaster(sp=period)` (one per period) and `TrendForecaster`.

fkiraly (changes requested):
It looks like that in this commit:
https://github.com/sktime/sktime/pull/5125/commits/c5df14c76f0c960deee63c27df5b499e3161bdc2

you accidentally deleted the file containing the MSTL transformer?

luca-miniati:
> The result must allow to access these somehow, since the plots displays the components separately. Although there's no clear documentation in `statsmodesl` that I can find...

I've split the seasonal components as shown in the `statsmodels` example. Hopefully, this should pass tests, as that weird parsing error seems to be gone.

luca-miniati:
> It looks like that in this commit: [c5df14c](https://github.com/sktime/sktime/commit/c5df14c76f0c960deee63c27df5b499e3161bdc2)
> 
> you accidentally deleted the file containing the MSTL transformer?

Yes, I had moved the file and forgot to add the "new" file. Should be there now under `sktime/transformations/mstl`

docs/source/api_reference/transformations.rst
```diff
@@ -236,6 +236,7 @@ Series-to-series transformers transform individual time series into another time
 
 When applied to panels or hierarchical data, individual series are transformed.
 
+
```
fkiraly, line 239: look like an acciental newline?

sktime/transformations/series/detrend/mstl.py
```diff
@@ -0,0 +1,218 @@
...
+import pandas as pd
+
+from sktime.transformations.base import BaseTransformer
+
+
+class MSTL(BaseTransformer):
+    """Season-Trend decomposition using LOESS for multiple seasonalities.
+
+    Direct interface for `statsmodels.tsa.seasonal.MSTL`.
+
```
fkiraly, line 18: I think it would be crucial to tell the user what it does in `transform`, i.e., what components it returns, with what column names, etc.

fkiraly (changes requested):
Looks great!

Blocking change request: can you explain to the user in the docstring what happens in `transform` (and/or `fit`)?

Non-blocking change requests:

* stray newline in API reference
* should there be an inverse transform? That should just sum up the three components back to one, as far as I see.

sktime/transformations/series/detrend/mstl.py
```diff
@@ -0,0 +1,248 @@
...
+    _tags = {
+        "scitype:transform-input": "Series",
+        "scitype:transform-output": "Series",
+        "scitype:instancewise": True,
+        "X_inner_mtype": "pd.Series",
+        "y_inner_mtype": "pd.Series",
+        "transform-returns-same-time-index": True,
+        "univariate-only": True,
+        "fit_is_empty": False,
+        "python_dependencies": "statsmodels",
```
fkiraly, line 101: since you decided to implement the `_inverse_transform`, you should add the tags or it will always be skipped.

here is the relevant excerpt:

```
        # capability:inverse_transform = is inverse_transform implemented?
        "capability:inverse_transform": False,
        # valid values: boolean True (yes), False (no)
        # if True, _inverse_transform must be implemented
        # if False, exception is raised if inverse_transform is called,
        #   unless the skip-inverse-transform tag is set to True
```

fkiraly (changes requested):
Great! Looks all good!

Now, since you decided to add the `_inverse_transform` (very useful!) you should ensure the tags are set that turn it on, see above.

fkiraly:
Ok, the failures seem to be the same as what PR https://github.com/sktime/sktime/pull/5166 addresses - that some `inverse_transform` are not exact, so the test that checks that should not run.

Otherwise all seems fine.

fkiraly (approved):
I've updated the tags after merging the respective framework PR, so the one outstanding point should now be addressed!

